Issue #3697, opened 2022-05-08
add `Random.shuffle`


JonathanStarup:
I'm not sure what properties are needed or if it should be on a typeclass but I'd like a shuffle function
```
def shuffle(r: Random.Random, c: Collection): Collection = ???
```
where Collection is at least Array and List

magnus-madsen:
@ninaandrup Can you add an **impure** function `Array.shuffle` that shuffles an `Array[a]`.

Then, we should add `shuffle` impure functions to the various namespaces that essentially do `x |> toArray |> shuffle |> toX`.

JonathanStarup:
> I think something simple like: https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle should do.
> 
> It essentially just swapping each `a[i]` with a random j.

It's important here what range j is from. I remembered this and selected any j which gave biased results. I just wanted to note that it's easy to make the wrong choice if you don't look at reference material.

ninaandrup:
If I have to make a `shuffle` function for the rest of the collections I have to make `toArray` and `toX` functions as well. Do we want `shuffle` for more collections?

magnus-madsen:
> If I have to make a `shuffle` function for the rest of the collections I have to make `toArray` and `toX` functions as well. Do we want `shuffle` for more collections?

Which collections are missing a shuffle? Sets and maps  cannot be shuffled because they are always sorted :)

ninaandrup:
> > If I have to make a `shuffle` function for the rest of the collections I have to make `toArray` and `toX` functions as well. Do we want `shuffle` for more collections?
> 
> Which collections are missing a shuffle? Sets and maps cannot be shuffled because they are always sorted :)

I think it is only `Nec` and `Nel`

PR #3948, opened 2022-06-13
feat: add `{Array, Iterator, Nec}.toMutDeque`

ninaandrup:
Closes #3761

PR #3955, opened 2022-06-14
feat: add `{Chain, List, MutDeque}.shuffle`

ninaandrup:
Related to https://github.com/flix/flix/issues/3697


ninaandrup:
@magnus-madsen it is not very convenient to have `r: Random.Random` when working with regions. What would be a better name for `Random`?

ninaandrup:
Waiting for #3948 to get merged so we have a implemented `Array.toMutDeque` with tests

PR #3958, opened 2022-06-14
feat: improve auto-completion for pipelines

### LINKED ISSUE #3957: Better auto complete within a pipeline
When you type a pipeline e.g. xs |> List.len and auto-complete we should drop the last argument to the function.

paulbutcher:
Fixes #3957

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
@@ -232,8 +232,14 @@ object CompletionProvider {
       s"$name(${args.mkString(", ")}): $retTpe & $eff"
   }
 
-  private def getApplySnippet(name: String, fparams: List[TypedAst.FormalParam]): String = {
-    val args = fparams.zipWithIndex.map {
+  /**
+    * Generate a snippet which represents calling a function.
+    * Drops the last argument in the event that the function is in a pipeline
+    * (i.e. is preceeded by `|>`)
+    */
+  private def getApplySnippet(name: String, fparams: List[TypedAst.FormalParam])(implicit context: Context): String = {
+    val fparamsUsed = if (context.previousWord == "|>") fparams.dropRight(1) else fparams
+    val args = fparamsUsed.zipWithIndex.map {
       case (fparam, idx) => "$" + s"{${idx + 1}:${fparam.sym.text}}"
     }
     s"$name(${args.mkString(", ")})"
@@ -472,7 +478,7 @@ object CompletionProvider {
    * @param uri          Source file URI (from client)
    * @param range        Start and end position of the word underneath (or alongside) the cursor
    * @param word         The word underneath (or alongside) the cursor
-   * @param previousWord The word before the above
+   * @param previousWord The word before the above (note that this may be on either the current or previous line)
    * @param prefix       The text from the start of the line up to the cursor
    */
   private case class Context(uri: String, range: Range, word: String, previousWord: String, prefix: String)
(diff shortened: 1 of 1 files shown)


jaschdoc:
Maybe this should also apply to `||>` and `!>`?

paulbutcher:
Done. And in the process, I also stopped it from adding parentheses if the argument list ends up being empty.

paulbutcher:
Converting to draft because I just realised I need to do the same for the type class case.

paulbutcher:
> Converting to draft because I just realised I need to do the same for the type class case.

False alarm, back out of draft.

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -490,15 +503,22 @@ object CompletionProvider {
...
+      for(line <- lines.slice(y, y + 1).toList.headOption) yield {
         val (prefix, suffix) = line.splitAt(x)
         val reversedPrefix = prefix.reverse
         val wordStart = reversedPrefix.takeWhile(isWordChar).reverse
         val wordEnd = suffix.takeWhile(isWordChar)
         val word = wordStart + wordEnd
         val start = x - wordStart.length
         val end = x + wordEnd.length
-        val previousWord = reversedPrefix.dropWhile(isWordChar).dropWhile(_.isWhitespace).takeWhile(isWordChar).reverse
+        val prevWord = reversedPrefix.dropWhile(isWordChar).dropWhile(_.isWhitespace).takeWhile(isWordChar).reverse
```
magnus-madsen, line 515: Perhaps a helper function for this line would be useful. Just because I can't quite parse what it does :)

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -490,15 +503,22 @@ object CompletionProvider {
...
+        val previousWord = if (prevWord.nonEmpty) {
+          prevWord
+        } else lines.slice(y - 1, y).toList.headOption match {
```
magnus-madsen, line 518: Could this slice y - 1 every be out of bounds?

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -490,15 +503,22 @@ object CompletionProvider {
...
+          case None => ""
+          case Some(s) => s.reverse.dropWhile(_.isWhitespace).takeWhile(isWordChar).reverse
```
magnus-madsen, line 520: This is just our own version of `trim` right? I guess it could also be in a helper function.

magnus-madsen (commented):
This looks fine to me. I just added some small suggestions.

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -232,11 +232,24 @@ object CompletionProvider {
...
-    val args = fparams.zipWithIndex.map {
+  /**
+    * Generate a snippet which represents calling a function.
+    * Drops the last one or two arguments in the event that the function is in a pipeline
+    * (i.e. is preceeded by `|>`, `!>`, or `||>`)
+    */
+  private def getApplySnippet(name: String, fparams: List[TypedAst.FormalParam])(implicit context: Context): String = {
+    val paramsToDrop = context.previousWord match {
+      case s if (s == "||>") => 2
+      case s if (s == "|>" || s == "!>") => 1
```
jaschdoc, line 243: Is there any reason not to directly match against the string pattern? I.e.
```scala
context.previousWord match {
      case "||>" => 2
      case "|>" | "!>" => 1
      case _ => 0
}
```

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
paulbutcher: Will do

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
paulbutcher: Yes, it could. But in that case it returns an empty list, and `headOption` therefore returns `None`, so everything should work correctly I believe.

magnus-madsen:
Looks good, just waiting for CI

magnus-madsen:
Thanks!

PR #3973, opened 2022-06-15
feat: add support for evaluating declarations and expressions

magnus-madsen:
…e shell.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser.scala
@@ -145,6 +145,10 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
       Declarations.Effect
   }
 
+  def DeclarationEOI: Rule1[ParsedAst.Declaration] = rule {
+    Declaration ~ EOI
+  }
+
   def UseDeclarations: Rule1[Seq[ParsedAst.Use]] = rule {
     // It is important for documentation comments that whitespace is not consumed if no uses are present
     (optWS ~ oneOrMore(Use ~ optWS ~ ";").separatedBy(optWS)) | push(Seq.empty)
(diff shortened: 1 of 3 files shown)


magnus-madsen:
@stephentetley @paulbutcher back by popular demand!

Its still very limited. You cannot bind variables and refer to them later.

Also its very untested, so there probably many bugs, but please do kick the tires :)

magnus-madsen:
This works on a purely textual basis. This means that its just keeping a buffer of declarations it has seen so far. There is no sharing of state between calls. Also, you cannot let-bind variables and then use them in subsequent expressions.

magnus-madsen:
Also, it shows the compilation progress bar after entering an expression. I don't know if that is good or bad?

paulbutcher:
It doesn't seem to have access to the project it's running within. So I have a project containing `Main.flix`:

```
def pbTest(x: Int32): Int32 =
  x + 42

def main(): Unit & Impure =
  println("")
```

```
% java -jar flix.jar --interactive
     __   _   _
    / _| | | (_)            Welcome to Flix v0.28.0
   | |_  | |  _  __  __
   |  _| | | | | \ \/ /     Enter a command and hit return.
   | |   | | | |  >  <      Type ':help' for more information.
   |_|   |_| |_| /_/\_\     Type ':quit' or press 'ctrl + d' to exit.
      
flix> pbTest(3)
                                                                                
-- Resolution Error -------------------------------------------------- <shell>

>> Undefined name 'pbTest'.

3 | pbTest(3)
    ^^^^^^
    name not found

Tip: Possible typo or non-existent definition?
flix> No main function to run.
flix> 
```

paulbutcher:
And similarly (in the same project):

```
flix> :run                                                                      
No main function to run.                                                        
flix> 
```

magnus-madsen:
If you pass the file arguments, e.g. `java -jar flix.jar foo.flix bar.flix --interactive` then the files are loaded. 

I am not sure if we should load in the files in the current folder (and recursively) by default. Right now we have the usual "compiler behavior" were you (or your tooling) has to pass the filenames.

paulbutcher:
Oh! That honestly never even occurred to me. If you look at the commands supported by the command line (init, check, build, etc.) every single one of them operates on a project. It never occurred to me that the interactive "REPL" would be the exception.

I've just tried, and can see that if I pass the filenames of my project on the command line, then I can access the code within them. But I can't see any way to do (what seems to me to be) the the obvious thing, which is "load the project and then let me play with it".

PR #3976, opened 2022-06-15
fix: allow whitespace just after and before multiuse braces

### LINKED ISSUE #3975: use does not allow whitespace before initial multiuse
This does not parse
```
use Namespace/Inner.{ somethingA, somethingB}
```

JonathanStarup:
Fixes #3975.


magnus-madsen:
Thanks!

PR #3977, opened 2022-06-15
feat: Array get optional

JonathanStarup:


magnus-madsen:
I am not so convinced whether this is a good idea nor what the name should be? (Possibly `at`?)

JonathanStarup:
> I am not so convinced whether this is a good idea nor what the name should be? (Possibly `at`?)

I suppose another option would be a inbounds function
```
pub def inbounds(i: Int32, a: Array[a, r]): Bool = 0 <= i and i < length(a)
```

magnus-madsen:
I am not very keen on adding `getOpt`. I am not even sure we should have `get`.

We could have an `indices` and a range object. (which would give you inbounds indirectly).

PR #3983, opened 2022-06-16
feat: add support for type completions

### LINKED ISSUE #2247: Add support for completion of type names (enums and type aliases).

paulbutcher:
Closes #2247


main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -481,6 +483,87 @@ object CompletionProvider {
...
+    * Format type params in the right form to be displayed in the list of completions
+    */
+  private def formatTParams(tparams: List[TypedAst.TypeParam]): String = {
+    tparams match {
+      case Nil => ""
+      case _ => tparams.map(_.name).mkString("[", ", ", "]")
+    }
+  }
+
+  val simpleTypeNames = List(
```
magnus-madsen, line 508: Lets move this to the top of the file and call it `BuiltinTypes` or `BuiltinTypeNames` 

Also add a comment like:

// This list manually maintained. If a new built-in type is added, it must be extended.
// Built-in types are typically descrbed in `TypeConstructor`, `Namer` and `Resolver`.

(SimpleType has a very specific meaning elsewhere in the compiler.)

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -481,6 +483,87 @@ object CompletionProvider {
...
+    "Unit",
+    "Null",
```
magnus-madsen, line 510: Null is really specially and should only be used for interop. I would not suggest it as a completion because it would rarely be meaningful.

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -481,6 +483,87 @@ object CompletionProvider {
...
+    "Int8",
+    "Int16",
+    "Int32",
+    "Int64",
+    "BigInt",
+    "String",
+    "Array",
+    "Ref",
+    "Channel",
+    "Lazy"
```
magnus-madsen, line 524: We also have types like tuples and records, but I guess we cannot easily complete those?

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -481,6 +483,87 @@ object CompletionProvider {
     }.toList
   }
 
+  /**
+    * Format type params in the right form to be inserted as a snippet
```
magnus-madsen, line 487: Could you add an example, e.g. Returns a string of the form `Name[Tparame1, ..]` (I think its like that?)

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -481,6 +483,87 @@ object CompletionProvider {
...
+        val name = t.sym.name
+        CompletionItem(label = s"$name${formatTParams(t.tparams)}",
+          sortText = Priority.tname(name),
+          textEdit = TextEdit(context.range, s"$name${formatTParamsSnippet(t.tparams)}"),
+          documentation = Some(t.doc.text),
+          insertTextFormat = InsertTextFormat.Snippet,
+          kind = CompletionItemKind.Enum)
+    }
+
+    val simpleTypes = simpleTypeNames map { name =>
```
magnus-madsen, line 557: builtinTypes

magnus-madsen (commented):
This looks good to me. I only have some minor suggestions.

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
paulbutcher: That makes sense. Should I also add a comment in one or more of TypeConstructor etc. pointing back here?

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
paulbutcher: I can't think of any easy way to do so, no, but please do let me know if you can think of one!

paulbutcher:
Comments addressed, I believe. Let me know about whether I should add a comment to one/all of TypeConstructor etc.

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
magnus-madsen: You could, yes!

magnus-madsen:
OK; I merged it. (We can add that comment later, no worries).

Issue #3771, opened 2022-05-16
Roadmap to disallowing non-unit expressions in statements and adding discard


JonathanStarup:
* [x] Add the discard keyword #3673.
    * [x] Use the discard keyword #3984.
* [x] Make channel retrieval return unit #3683.
    * [x] Simplify parser to disallow chained <- operations #4093.
* [x] Make channel operations return unit in the new standard lib implementation #3775.
    * [x] #3880
* Remove non-unit statement expressions in standard library.
    * [x] #3877
    * [x] #3879
    * [x] #3882
* There are changes that are not needed but do support this design decision.
    * [x] Some functions want to use zipWithIndex for array, chain, delaylist, nec but use a fold instead #3682.
        * Rewrite some folds to use zipWithIndex and foreach instead of folds
            * [x] #3881.
            * [x] #3878
    * [x] Add the fluent builder operator to remove return values on destructive methods #3664 and #3694.
        * [x] #3883
* [x] Add the redundancy check to disallow non-unit statement expressions #3667.

Issues related to this:
* #2432
    * A question for this is: should we add an ignore function to preamble? This would allow `something |> doSomething |> ignore`

JonathanStarup:
@magnus-madsen I think this should capture all of the discussion around the topic and lay out the steps needed

PR #3984, opened 2022-06-16
refactor: use discard in non-unit impure statements

JonathanStarup:
Related to #3771.

### DIFF
--- a/main/src/library/Array.flix
+++ b/main/src/library/Array.flix
@@ -603,7 +603,7 @@ namespace Array {
                         let pos2 = arrayWrites(pos1, a, out);
                         (pos2, i+1)
                     }};
-                foldLeft(overwrite, (0, 0), arrs);
+                discard foldLeft(overwrite, (0, 0), arrs);
                 out
         }
 
@@ -889,7 +889,7 @@ namespace Array {
             case None => []
             case Some(x) => {
                 let out = new(x, len, Static);
-                foldLeft((pos, a) -> arrayWrites(pos, a, out), 0, arrs);
+                discard foldLeft((pos, a) -> arrayWrites(pos, a, out), 0, arrs);
                 out
             }
         }
--- a/main/src/library/File.flix
+++ b/main/src/library/File.flix
@@ -310,9 +310,9 @@ namespace File {
 
             let javaFile = newFile(f);
             let stream = newInputStream(javaFile);
-            skip(stream, opts.offset);
+            discard skip(stream, opts.offset);
             let bytes = [0i8; opts.count];
-            read(stream, bytes, 0, opts.count);
+            discard read(stream, bytes, 0, opts.count);
             let charSet = forName(opts.charSet);
             let returnString = newString(bytes, charSet);
             Ok(returnString)
@@ -486,7 +486,7 @@ namespace File {
 
             let javaFile = newFile(f);
             let stream = newInputStream(javaFile);
-            skip(stream, opts.offset);
+            discard skip(stream, opts.offset);
             let bytes = [0i8; opts.count];
             let numberRead = read(stream, bytes, 0, opts.count);
 
@@ -957,7 +957,7 @@ namespace File {
             let srcPath = toPath(srcFile);
             let dstFile = newFile(dst);
             let dstPath = toPath(dstFile);
-            move(srcPath, dstPath, []);
+            discard move(srcPath, dstPath, []);
             Ok()
         } catch {
             case ex: ##java.io.IOException =>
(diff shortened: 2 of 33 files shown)


main/src/library/Array.flix
```diff
@@ -603,7 +603,7 @@ namespace Array {
                         let pos2 = arrayWrites(pos1, a, out);
                         (pos2, i+1)
                     }};
-                foldLeft(overwrite, (0, 0), arrs);
+                discard foldLeft(overwrite, (0, 0), arrs);
```
magnus-madsen, line 606: Use foreach?

main/src/library/Array.flix
```diff
@@ -889,7 +889,7 @@ namespace Array {
             case None => []
             case Some(x) => {
                 let out = new(x, len, Static);
-                foldLeft((pos, a) -> arrayWrites(pos, a, out), 0, arrs);
+                discard foldLeft((pos, a) -> arrayWrites(pos, a, out), 0, arrs);
                 out
```
magnus-madsen, line 893: Use foreach?

main/src/library/File.flix
```diff
@@ -310,9 +310,9 @@ namespace File {
 
             let javaFile = newFile(f);
             let stream = newInputStream(javaFile);
-            skip(stream, opts.offset);
+            discard skip(stream, opts.offset);
```
magnus-madsen, line 313: Does this suggest there is a bug? What does skip do? Can it fail?

main/src/library/File.flix
```diff
@@ -1018,7 +1018,7 @@ namespace File {
 
             let srcFile = newFile(src.src);
             let srcPath = toPath(srcFile);
-            move(srcPath, dstPath, [getOverwrite() as ##java.nio.file.CopyOption]);
+            discard move(srcPath, dstPath, [getOverwrite() as ##java.nio.file.CopyOption]);
```
magnus-madsen, line 1021: This also looks unsafe, no?

main/test/ca/uwaterloo/flix/library/TestDelayList.flix
```diff
@@ -429,15 +429,15 @@ namespace TestDelayList {
     @test
     def mapFusion01(): Bool & Impure =
         let l = ref Nil;
-        (1 :: 2 :: 3 :: Nil) |> List.toDelayList |>
+        discard (1 :: 2 :: 3 :: Nil) |> List.toDelayList |>
```
magnus-madsen, line 432: I wonder if this is the binding we expect? Maybe discard should be more closely?

main/test/flix/Test.Exp.ArraySlice.flix
```diff
@@ -3,203 +3,203 @@ namespace Test/Exp/ArraySlice {
...
-        let _ = x[2..3];
+        discard x[2..3];
         ()
     }
 
     @test
     def testArraySlice07(): Unit = region r {
         let x = [false, false, false, false, true] @ r;
-        let _ = x[1..3];
+        discard x[1..3];
```
magnus-madsen, line 48: I am surprised? I would think that slicing a mutable array has an effect?

main/test/flix/Test.Exp.ArrayLit.flix
```diff
@@ -2,523 +2,523 @@ namespace Test/Exp/ArrayLit {
 
     @test
     def testArrayLit01(): Unit = region r {
-        let _ = [];
+        discard [];
```
magnus-madsen, line 5: Does this not have an effect? In particular `r`?

magnus-madsen:
I have some high-level concerns that should be discussed. We can do it here and/or at the meeting.

main/test/flix/Test.Exp.ArraySlice.flix
JonathanStarup: It does have an effect, but it is non-unit, so you have to discard it. discarding something pure would be an error.

main/test/flix/Test.Exp.ArrayLit.flix
JonathanStarup: Same comment as the above, it is non-unit, so you have to discard it.

main/src/library/Array.flix
JonathanStarup: The function needs to keep track of how big the nested expressions are so I think its best as a fold. It could be a zipWithIndex and then a fold but a fold none the less.

main/src/library/File.flix
JonathanStarup: I've made an issue #4006

magnus-madsen:
My plan is still to have this merged, but I might wait until after the meeting, to ensure that we get the right design.

PR #3985, opened 2022-06-16
feat: add `RedundancyError.UnderAppliedFunction`

### LINKED ISSUE #3758: Quality of life: useless expression errors should include types
This program
```
def main(): Unit & Impure = {
    let num = ref 2;
    addRefs(num);
    println(deref num)
}
def addRefs(n1: ScopedRef[Int32, r], n2: ScopedRef[Int32, r]): Unit & r = {
    n1 := deref n2
}
```
Gives a redundancy error
image

The error here is that I forgot to fully apply the function. In the full example where this happened I used some regions so I thought that the errors related to regions. Reporting the type of the expression would have helped me to spot the under-application.

@magnus-madsen this could also be a suitable place to specialize the error if the useless expression is of function type

A function like this
```
def justCallingAddRefs(n1: ScopedRef[Int32, r1], n2: ScopedRef[Int32, r2]): Unit & r1 and r2 = {
    addRefs(n1)
}
```
Also gives the error of the purity before the result type error
```
❌ -- Type Error -------------------------------------------------- file:...

>> The inferred effect: 'Pure' cannot be generalized to 'r1 and r2'.

8 | def justCallingAddRefs(n1: ScopedRef[Int32, r1], n2: ScopedRef[Int32, r2]): Unit & r1 and r2 = {
        ^^^^^^^^^^^^^^^^^^
        unable to generalize the effect.
```
I'd argue that the type error should come before the effect error because its a) easier to understand and b) the type fix could change the effect

JonathanStarup:
Fixes #3758.
Included in this
* UselessFunctionExpression (and tests)
* Added type information to UselessExpression


magnus-madsen:
If so, should we check that the function has an effect?

JonathanStarup:
> What's the intuition here? I am guessing when you under apply a function?

yes

> If so, should we call the error something closer to that?

UselessUnderAppliedFunction? The error could still be that you underapplied a effectful function but forgot to letbind it. 

> If so, should we check that the function has an effect?

We could yes, then an underapplied pure function would just give a regular useless expression error?

main/src/ca/uwaterloo/flix/language/errors/RedundancyError.scala
```diff
@@ -312,15 +312,17 @@ object RedundancyError {
...
     */
-  case class UselessExpression(loc: SourceLocation) extends RedundancyError {
+  case class UselessExpression(tpe: Type, loc: SourceLocation) extends RedundancyError {
     def summary: String = "Useless expression."
 
     def message(formatter: Formatter): String = {
       import formatter._
       s"""${line(kind, source.name)}
          |>> Useless expression: It has no side-effect(s) and its result is discarded.
+         |>> The expression has type '${FormatType.formatWellKindedType(tpe)}'
```
magnus-madsen, line 325: This is some extra information. I would move this after the code part (and remove the >>).

main/src/ca/uwaterloo/flix/language/errors/RedundancyError.scala
```diff
@@ -338,6 +340,38 @@ object RedundancyError {
...
+    * @param loc the location of the expression.
+    */
+  case class UselessFunctionExpression(tpe: Type, loc: SourceLocation) extends RedundancyError {
+    def summary: String = "Useless function expression."
+
+    def message(formatter: Formatter): String = {
+      import formatter._
+      s"""${line(kind, source.name)}
+         |>> Useless function expression: It has no side-effect(s) and its result is discarded.
+         |>> The function has type '${FormatType.formatWellKindedType(tpe)}'
```
magnus-madsen, line 356: Move later

main/src/ca/uwaterloo/flix/language/errors/RedundancyError.scala
```diff
@@ -338,6 +340,38 @@ object RedundancyError {
     })
   }
 
+  /**
+    * An error raised to indicate that a function expression is useless.
```
magnus-madsen, line 344: If we want to have this error message, we should try to give some more information.

main/src/ca/uwaterloo/flix/language/errors/RedundancyError.scala
```diff
@@ -338,6 +340,38 @@ object RedundancyError {
...
+    *
+    * @param tpe the type of the expression.
+    * @param loc the location of the expression.
+    */
+  case class UselessFunctionExpression(tpe: Type, loc: SourceLocation) extends RedundancyError {
```
magnus-madsen, line 349: UselessFunctionExpression -> UnderAppliedFunction

main/src/ca/uwaterloo/flix/language/errors/RedundancyError.scala
```diff
@@ -338,6 +340,38 @@ object RedundancyError {
...
+    def summary: String = "Useless function expression."
```
magnus-madsen, line 350: Under applied function. Missing function argument(s)?

main/src/ca/uwaterloo/flix/language/errors/RedundancyError.scala
```diff
@@ -338,6 +340,38 @@ object RedundancyError {
...
+
+    def message(formatter: Formatter): String = {
+      import formatter._
+      s"""${line(kind, source.name)}
+         |>> Useless function expression: It has no side-effect(s) and its result is discarded.
```
magnus-madsen, line 355: Under applied function. Missing function argument(s)?

main/src/ca/uwaterloo/flix/language/errors/RedundancyError.scala
```diff
@@ -338,6 +340,38 @@ object RedundancyError {
...
+         |>> The function has type '${FormatType.formatWellKindedType(tpe)}'
+         |
+         |${code(loc, "useless function expression.")}
```
magnus-madsen, line 358: the function is not fully-applied and hence has no effect.

main/src/ca/uwaterloo/flix/language/phase/Redundancy.scala
```diff
@@ -390,9 +390,13 @@ object Redundancy {
       val us2 = visitExp(exp2, env0, rc)
 
       // Check for useless pure expressions.
-      if (exp1.eff == Type.Pure)
```
magnus-madsen, line 393: Introduce helper predicates:

- isUseless(exp1)
- isUnderAppliedFunction(exp1)

Note isUnderAppliedFunction should check both for purity, but also that the function actually has an effect!

main/src/ca/uwaterloo/flix/language/errors/RedundancyError.scala
magnus-madsen: Maybe we should go further? Missing argument(s) of type: tpe1, tpe2... 

Lets try that?

main/src/ca/uwaterloo/flix/language/errors/RedundancyError.scala
JonathanStarup: So the useless part is left implicit? I think the name implies that it is an error to partially apply something

magnus-madsen:
I think a more specific error message would be useful.

JonathanStarup:
@magnus-madsen feedback is implemented. I noticed an issue regarding curried functions so some additional work on types had to be done to recursively find all information needed.

main/src/ca/uwaterloo/flix/language/errors/RedundancyError.scala
```diff
@@ -338,6 +341,51 @@ object RedundancyError {
...
+    * An error raised to indicate that an impure function expression is useless
+    * is statement position.
+    *
+    * @param tpe the type of the expression.
+    * @param loc the location of the expression.
+    */
+  case class UnderAppliedFunction(tpe: Type, loc: SourceLocation) extends RedundancyError {
+    def summary: String = "Under applied function. Missing function argument(s)?"
+
+    private def applicationAdvice(tpe: Type): String = {
```
magnus-madsen, line 354: Move helper function last in this class.

main/src/ca/uwaterloo/flix/language/errors/RedundancyError.scala
```diff
@@ -338,6 +341,51 @@ object RedundancyError {
...
+      val arguments = tpe.curriedArrowArgTypes
+      if (arguments.isEmpty) { // fallback message
+        "Missing function argument(s)?"
+      } else {
+        val argumentStrings = arguments.map(t => s"'${FormatType.formatWellKindedType(t)}'").mkString(", ")
+        val pluralS = if (arguments.sizeIs == 1) "" else "s"
+        s"Missing argument$pluralS of type: $argumentStrings."
```
magnus-madsen, line 361: How about the format of this error message just becomes:

Missing argument(s) of type: Foo, Bar, Baz

?

main/src/ca/uwaterloo/flix/language/ast/Type.scala
```diff
@@ -160,26 +160,85 @@ sealed trait Type {
...
+    * NB: Assumes that `this` type is an arrow.
+    */
+  def curriedArrowEffectType: Type = {
+    val resType = arrowResultType
+    resType.typeConstructor match {
+      case Some(TypeConstructor.Arrow(_)) => resType.curriedArrowEffectType
+      case _ => arrowEffectType
+    }
+  }
+
```
magnus-madsen, line 241: I understand that you need these functions. It also makes sense to put them on Type.

However, I would prefer to move them to where they are actually used. 

The reason is that they are quite special and brittle, and we *do not* other parts of the compiler to depend on them.

magnus-madsen: In other words. Type tries hard to only expose a few core functions that a widely used.

magnus-madsen:
Thanks!

PR #3990, opened 2022-06-19
refactor: delete functions from `Prelude`

### LINKED ISSUE #3945: Proposal to slim down `Prelude`
Drop:

- `<==>`
- `==>`
- `curry`
- `uncurry`
- `constant`

I think these are rarely used.

ninaandrup:
Closes #3945

### DIFF
--- a/main/src/library/Prelude.flix
+++ b/main/src/library/Prelude.flix
@@ -34,11 +34,6 @@ pub enum ReifiedType with Eq, Order, ToString {
 ///
 pub def identity(x: a): a = x
 
-///
-/// Returns the constant function with return value `y`.
-/// That is, returns `f` such that `f(x) = y` for all inputs `x`.
-///
-pub def constant(y: a): b -> a = _ -> y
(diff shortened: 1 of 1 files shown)


magnus-madsen:
I did not realize `==>` was being used.

Here is my new proposal. Let us move it into `Bool.==>`  and then that can be imported by the type classes.

ninaandrup:
I do know why the tests are failing :(

main/src/library/Eq.flix
```diff
@@ -47,7 +49,7 @@ pub lawful class Eq[a] {
     ///
     /// x != y is logically equivalent to not (x == y).
     ///
-    law inverseNeq: forall(x: a, y: a). (x != y) <==> (not (x == y))
```
magnus-madsen, line 50: I realize I would also like to bring back `<==>` but keep it in the `Bool` namespace and import it.

Sorry for the confusion and thanks!

PR #3995, opened 2022-06-20
Towards implementing object

paulbutcher:
Incomplete plumbing to get the new object syntax working. I believe that these are the right changes to the parser, weeder, namer, resolver, and kinder. Next up is the typer, but it would be good to get confirmation (or otherwise) that I'm heading in the right direction.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/ParsedAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/ParsedAst.scala
@@ -657,6 +657,15 @@ object ParsedAst {
     case class LetImport(sp1: SourcePosition, op: ParsedAst.JvmOp, exp: ParsedAst.Expression, sp2: SourcePosition) extends ParsedAst.Expression
 
     /**
+      * NewObject (create an anonymous object which implements a Java interface).
+      *
+      * @param sp1 the position of the first character in the expression.
+      * @param fqn the name of the Java interface.
+      * @param sp2 the position of the last character in the expression.
+      */
+    case class NewObject(sp1: SourcePosition, fqn: Seq[String], sp2: SourcePosition) extends ParsedAst.Expression
+
+    /**
       * Static Region Expression.
       *
       * @param sp1 the position of the first character in the expression.
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser.scala
@@ -805,6 +805,10 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
       }
     }
 
+    def NewObject: Rule1[ParsedAst.Expression] = rule {
+      SP ~ keyword("object") ~ WS ~ atomic("##") ~ Names.JavaName ~ SP ~ WS ~ "{" ~ optWS ~ "}" ~> ParsedAst.Expression.NewObject
+    }
+
     def Match: Rule1[ParsedAst.Expression.Match] = {
       def Rule: Rule1[ParsedAst.MatchRule] = rule {
         keyword("case") ~ WS ~ Pattern ~ optWS ~ optional(keyword("if") ~ WS ~ Expression ~ optWS) ~ atomic("=>") ~ optWS ~ Stm ~> ParsedAst.MatchRule
--- a/main/src/flix/test/TestInterface.java
+++ b/main/src/flix/test/TestInterface.java
@@ -1,0 +1,4 @@
+package test.flix;
+
+public interface TestInterface {
+}
--- a/main/test/flix/Test.Exp.Jvm.Object.flix
+++ b/main/test/flix/Test.Exp.Jvm.Object.flix
@@ -1,0 +1,7 @@
+namespace Test/Exp/Jvm/Object {
+
+  @test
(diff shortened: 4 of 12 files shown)


main/src/flix/test/TestInterface.java
```diff
@@ -0,0 +1,4 @@
+package test.flix;
+
+public interface TestInterface {
```
magnus-madsen, line 3: We have had some trouble with tests depending on Java sources.

Perhaps you could simply use:

- java.lang.Object
- java.io.Serializable
- java.lang.Cloneable

paulbutcher: Can do, but I can see that there are tests in, for example, `flix/main/test/flix/Test.Exp.Jvm.GetField.flix` which depend upon `test.flix.TestClass`? Can you explain the "boundaries" of the problems?

magnus-madsen: > Can do, but I can see that there are tests in, for example, `flix/main/test/flix/Test.Exp.Jvm.GetField.flix` which depend upon `test.flix.TestClass`? Can you explain the "boundaries" of the problems?

Its possible it has been resolved :)

magnus-madsen: The problem is/was how Java source code is compiled and made available to the Flix test suite.

paulbutcher:
I've now, I think, plumbed NewObject all the way through, and all that remains is code generation.

I've made some educated guesses in places which I wouldn't be at all surprised to discover I'd got wrong, so please shout if I have.

main/src/ca/uwaterloo/flix/language/ast/ParsedAst.scala
```diff
@@ -656,6 +656,15 @@ object ParsedAst {
       */
     case class LetImport(sp1: SourcePosition, op: ParsedAst.JvmOp, exp: ParsedAst.Expression, sp2: SourcePosition) extends ParsedAst.Expression
 
+    /**
+      * NewObject (create an anonymous object which implements a Java interface).
```
magnus-madsen, line 660: ... implements a Java interface or extends a Java class.

main/src/ca/uwaterloo/flix/language/ast/ParsedAst.scala
```diff
@@ -656,6 +656,15 @@ object ParsedAst {
...
+      *
+      * @param sp1 the position of the first character in the expression.
+      * @param fqn the name of the Java interface.
```
magnus-madsen, line 663: the fully-qualified name of the Java interface or class.

magnus-madsen (commented):
Looks good!

magnus-madsen:
The backend is still a bit of a mess (compared to the rest of the compiler) and is gradually being replaced by something nicer. 

I think what the backend needs to do is:

- Collect a list of anon. classes that must be generated (essentially every NewObject expression will give rise to one). We need to give these a name (somehow).
- Implement a new `GenAnonClasses` which emits a class that implements the interface for each collected anon. class.
- Modify `GenExpression` to emit a code that constructs a new object of the right class.

magnus-madsen:
I might be nicer to collect a map of anon. classes much earlier in the compiler pipeline. At least we should think about it.

magnus-madsen:
Before we do the class generation, you could just have `GenExpression` emit ACONST_NULL and then we can merge this PR in with only a few more lines.

magnus-madsen:
> > I might be nicer to collect a map of anon. classes much earlier in the compiler pipeline. At least we should think about it.
> 
> Any suggestions for where exactly? Perhaps the namer? And we can generate a unique name for each while we're at it?

I would propose we hold of on a bit. We don't have enough information just yet, so I would continue your current path.

paulbutcher:
Working (for small values of "working" 😉)

magnus-madsen:
Should I review and merge?

paulbutcher:
I believe it's ready for that, yes. I'll start looking at code generation.

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
```diff
@@ -805,6 +805,10 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
       }
     }
 
+    def NewObject: Rule1[ParsedAst.Expression] = rule {
+      SP ~ keyword("object") ~ WS ~ atomic("##") ~ Names.JavaName ~ SP ~ WS ~ "{" ~ optWS ~ "}" ~> ParsedAst.Expression.NewObject
```
magnus-madsen, line 809: Two things:

- You insist on hard whitespace before the first brace? We could get away with optional whitespace.
- The second SP should be after the last brace (thus the two SPs span the entire range of the object expression).

main/test/flix/Test.Exp.Jvm.Object.flix
```diff
@@ -0,0 +1,7 @@
+namespace Test/Exp/Jvm/Object {
```
magnus-madsen, line 1: This file should probably be called `...Jvm.NewObject.flix` and its namespace updated.

paulbutcher:
> OK, only minor issues.  

Done

magnus-madsen:
Merged. Thanks!

PR #3999, opened 2022-06-22
Only display code lenses if AST is current

### LINKED ISSUE #200: Implement multi-line error reporting
We currently have error reporting ala:

```
def foo: Bool = 0 == (1, 2, 3, 4)
```

Gives us:

```
-- TYPE ERROR -------------------------------------------------- foo.flix

>> Expected equal types 'Int32' and '(Int32, Int32, Int32, Int32)'.

1|def foo: Bool = 0 == (1, 2, 3, 4)
                  ^
1|def foo: Bool = 0 == (1, 2, 3, 4)
                       ^^^^^^^^^^^^
```

but if we split the code over multiple lines:

```
def foo: Bool = 0 == (1, 
                      2, 
                      3, 
                      4)
```

We only get the first line:

```
-- TYPE ERROR -------------------------------------------------- foo.flix

>> Expected equal types 'Int32' and '(Int32, Int32, Int32, Int32)'.

1|def foo: Bool = 0 == (1,
                  ^
1|def foo: Bool = 0 == (1,
                       ^^^
```

paulbutcher:
Fixes #200

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/LanguageServer.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/LanguageServer.scala
@@ -250,9 +250,12 @@ class LanguageServer(port: Int) extends WebSocketServer(new InetSocketAddress("l
     case Request.Check(id) => processCheck(id)
 
     case Request.Codelens(id, uri) =>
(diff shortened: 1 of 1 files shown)


magnus-madsen:
Thanks! Nice to have it solved :)

Issue #3896, opened 2022-06-05
Refactor Datalog solver to use regions


magnus-madsen:
We should refactor the Datalog engine to use regions (while collecting information about the refactoring for the case study).

I propose the following approach:

- [ ] Declare a new region in `solveWithStratification` (I think that's the entry point, right?)
- [ ] Add a `r: Region[r]` parameter to every function in `Fixpoint` which appears to need such an argument. (Need to discuss if it should be first or last.)
- [ ] Refactor the Arrays in `Fixpoint.Ram` to use regions.
- [ ] Refactor the Arrays in `Fixpoint.Ast` to use regions (may require tweaks to `Lowering`.

Each should be a separate PR. Possibly with sub-PRs, if feasible.

While doing this work, keep notes of interesting observations. E.g. is the refactoring straightforward? Are there interesting choices to be made? Are some restructuring required or not?

magnus-madsen:
We should also ask ourselves, is there any reason for the Ram AST to use arrays? Maybe we could just use a list.

Then we would only need a region for the actual database. I would probably suggest that.

But do take a look at the code.

JonathanStarup:
As far as I can see arrays are used for 1) random access and 2) array reuse for efficiency

1) could be solved by an immutable vector
2) seems like the cost of functional code here.

JonathanStarup:
Additional todos:
* [ ] Actually impure toString methods should not implement ToString typeclass

magnus-madsen:
I would suggest to break this into smaller tasks than can be merged one by one.

JonathanStarup:
> I propose to change the data structures ie ast and ram to use lists (then in the future they can use vectors) and to keep using arrays for the search env and to use mutable collections for the databases. This makes sense and seems like a useful use case for regions.

So this means that lookups will be linear? as a temporary performance cost?

magnus-madsen:
> > I propose to change the data structures ie ast and ram to use lists (then in the future they can use vectors) and to keep using arrays for the search env and to use mutable collections for the databases. This makes sense and seems like a useful use case for regions.
> 
> So this means that lookups will be linear? as a temporary performance cost?

I don't think so, because we don't do any lookups for the AST/RAM enums.

magnus-madsen:
All the performance critical stuff is in searchEnv and db and there we will continue to use arrays/mut maps (but with regions).

magnus-madsen:
How far are we? :)

JonathanStarup:
[...]

PR #4000, opened 2022-06-22
refactor: change `RelOp` to use lists

JonathanStarup:
Related to #3896.

### DIFF
--- a/main/src/library/Fixpoint/Interpreter.flix
+++ b/main/src/library/Fixpoint/Interpreter.flix
@@ -93,7 +93,7 @@ namespace Fixpoint {
             case RelOp.Query(RowVar.Index(i), ramSym, query, body) =>
                 let (tupleEnv, latEnv) = env;
                 // TODO: Unsafe effect cast
-                MutMap.queryWith(evalQuery(env, 0, query) as Fixpoint.Tuple[v] -> Comparison, t -> l -> {
+                MutMap.queryWith(evalQuery(env, query) as & Pure -> Comparison, t -> l -> {
                     tupleEnv[i] = t;
                     latEnv[i] = l;
                     evalOp(db, env, body)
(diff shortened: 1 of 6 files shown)


main/src/library/Fixpoint/Interpreter.flix
```diff
@@ -102,16 +101,20 @@ namespace Fixpoint {
...
-                            ()
-                        else
-                            MutMap.putWith!(lub, key, lat, rel)
+                        // assume that length(terms) > 0
+                        let len = List.length(terms);
+                        let keyList = terms |> List.map(evalTerm(env));
+                        let (relKeys, latValList) = List.splitAt(len-1, keyList);
+                        let key = Tuple(List.toArray(relKeys, Static));
+                        let latVal = match List.head(latValList) {
+                            case None => ?emptyListOfTerms
```
magnus-madsen, line 113: Should use `bug!`

main/src/library/Fixpoint/Interpreter.flix
```diff
@@ -102,16 +101,20 @@ namespace Fixpoint {
...
-                        if (lat `leq` bot)
-                            ()
-                        else
-                            MutMap.putWith!(lub, key, lat, rel)
+                        // assume that length(terms) > 0
+                        let len = List.length(terms);
+                        let keyList = terms |> List.map(evalTerm(env));
+                        let (relKeys, latValList) = List.splitAt(len-1, keyList);
+                        let key = Tuple(List.toArray(relKeys, Static));
+                        let latVal = match List.head(latValList) {
```
magnus-madsen, line 112: Could we use a non-empty list? `Nel`?

JonathanStarup: Possibly yes. It would be a bit cumbersome though since the required element is the last item of the list so we could perhaps store the terms in reverse. Should I do this?

magnus-madsen: Does Nel not expose a `last` and `init` which is what you want?

magnus-madsen: I would do it now, or at least try, and see how it goes.

JonathanStarup: Ah yes, thats true. Ill give it a go

JonathanStarup: yes id believe so too. ill look it over again

magnus-madsen:
image

😆

Issue #2591, opened 2021-11-09
New Backend


JonathanStarup:
## Rewriting the backend to use a frame transformer DSL
* [x] Make ErasedAst be a copy of FinalAst
* [x] Make the eraser transform FinalAst to ErasedAst but return FinalAst
* [x] Change JvmBackend to use ErasedAst
* [ ] Add explicit cast expressions and introduce them as needed in the eraser
* [x] Add instruction transformations
* [ ] Change the (internal) compilation method to return a frame transformer
* [ ] Collect jvm type information in `BackendObjType` objects. This will reduce the `GenXYZClass` work to calling `BackendObjXYZ.generateBytecode(...)`
    * [x] Fields #3932
    * Methods/constructors/genByteCode methods
        * [x] #4001
        * [x] #4015
        * [x] #4016
        * [x] #4019
        * [x] #4072
        * [x] #4094
        * [x] #4098
* [ ] Rewrite cases one by one 
* [ ] Rewrite gen classes
  * [x] #2604
  * [x] #2612
  * [x] #2637
  * [x] #2639
  * [x] #2660
  * [x] #2698
  * [x] #2721
  * [x] #2782
  * [x] #2797
  * [x] #2845
  * [x] #2891
  * [x] #2898
  * [x] #3884
  * [ ] Function classes
  * [ ] Closures
  * [ ] Lazy
  * [x] ...
  * [ ] Enum Interfaces
  * [ ] Main
  * [ ] namespaces?
  * [ ] tags
  * [ ] tuples
  * [ ] tuple interfaces
  * [ ] Make documentation, naming, and format consistent

## Adding type safety
* [ ] Add PType/RType to model the jvm types
* [ ] Add type index to erased expressions
* [ ] Add types for dsl instructions and cases (this probably has to be done in one go)

## Additional Issues:
* [ ] Maximal parallelism #3531
* [ ] Type string generation and memoization (store strings in the type values?)
* [x] #2417
* [x] #2715 
* [ ] #2359
* [ ] Support of nested array types (#3332)
* [ ] Type set collection in the eraser to avoid multiple AST passes (collection of things like the set of all types)

But I think these are separate and muddled the previous attempt

magnus-madsen:
We can start building the DSL based on the `GenUnit` and similar simple classes and then go from there.

magnus-madsen:
Use DSL in:

- [ ] `GenUnit`
- [ ] `GenGlobalCounterClass`
[...]

PR #4001, opened 2022-06-22
refactor: collect finality/publicity/code in method objects

JonathanStarup:
Part of #2591.


magnus-madsen:
For these PRs its hard for me to check for correctness, so I mostly just assume that you know what you are doing 👍  and that it works :)

JonathanStarup:
> For these PRs its hard for me to check for correctness, so I mostly just assume that you know what you are doing 👍 and that it works :)

It is a bit messy right now but the idea of these PR's is to essentially remove all of the genXClass objects. Instead, the whole jvm class is declared on the BackendObjType and generation just happens by calling .genClass on the types collected in the backend. Generating ref classes will be

```
refTypes.map(_.genClass).toMap
```

Issue #3963, opened 2022-06-15
Numeric functions and subject last


JonathanStarup:
It seems to me that many of the numeric functions, like `Int64.clampToInt` does not adhere to subject last.
```
pub def clampToInt32(x: Int64, minimum: Int32, maximum: Int32): Int32
```

stephentetley:
Potentially `minimum` and `maximum` should be in a record as well to simulate named arguments.

ninaandrup:
@magnus-madsen I have made the functions
```
clampToInt*(minimum: Int32, maximum: Int32, x: Int*): Int*
```
to
```
clampToInt*(minimum: {minimum :: Int32}, maximum: {maximum :: Int32}, x: Int*): Int*
```
do we maybe want to use `min` and `max` instead such that it is not too long?

stephentetley:
> clampToInt*(min: {min :: Int32}, max: {max :: Int32}, x: Int*): Int*

I wonder if this is a nicer signature:

> clampToInt*(range: {min :: Int32, max :: Int32}, x: Int*): Int*

My wager is that no-one would ever partially apply just min so we might as we make a tangible range.

magnus-madsen:
Agree with Stephen. Let's do that.

JonathanStarup:
related to this #3977, it could be cool to have range objects that the std lib could use. Minimum, maximum, and maximum inclusive/exclusive. Cool but maybe also too cumbersome.

magnus-madsen:
I agree a Range type or type class (?) would be nice. Not sure what the best practice is. Suggestions welcome.

ninaandrup:
This should be closed. All numeric functions have been refactored to subject last :-)

PR #4005, opened 2022-06-22
refactor: `Int32` functions to subject last and use record

ninaandrup:
Closes to https://github.com/flix/flix/issues/3963


main/src/library/Int32.flix
```diff
@@ -102,38 +102,38 @@ namespace Int32 {
     /// A zero is shifted into the leftmost position regardless of sign extension.
     ///
     @Time(1) @Space(1)
-    pub def logicalRightShift(x: Int32, distance: Int32): Int32 =
-        if (x < 0 and distance rem size() != 0)
-            ((x >>> 1) &&& maxValue()) >>> (distance - 1)
+    pub def logicalRightShift(distance: {distance :: Int32}, x: Int32): Int32 =
```
ninaandrup, line 105: Since both arguments are `Int32` i made `distance` a record. Do we want the whole word or something like `dist`?

main/src/library/Int32.flix
```diff
@@ -242,32 +242,32 @@ namespace Int32 {
     /// The bits of x have positions: 0 (rightmost bit) - 31 (leftmost bit)
     ///
     @Time(1) @Space(1)
-    pub def getBit(x: Int32, position: Int32): Int32 =
-        if ((x >>> position) rem 2 == 0) 0 else 1
+    pub def getBit(position: {position :: Int32}, x: Int32): Int32 =
```
ninaandrup, line 245: Since both arguments are `Int32` i made `position` a record. Do we want the whole word or something like `pos`?

ninaandrup (commented):
I'm not sure about the naming of the records

main/src/library/Int32.flix
magnus-madsen: Yes, lets do it!

main/src/library/Int32.flix
magnus-madsen: Yes, lets do it!

PR #4008, opened 2022-06-22
refactor: generalize Array functions to use Arrays with a polymorphic region rather than the Static region

stephentetley:
This PR generalizes various Array functions (`flatMap`, `flatten`, etc.) to use a polymorphic region rather than the Static region. 

Presumably these functions were the first to be converted to use regions and got missed when the `Scoped` class was introduced allowing allowed more generalization.

### DIFF
--- a/main/src/library/Array.flix
+++ b/main/src/library/Array.flix
@@ -275,7 +275,7 @@ namespace Array {
     /// Return the positions of the all the occurrences of `a` in `arr`.
     ///
     @Time(length(arr)) @Space(length(arr))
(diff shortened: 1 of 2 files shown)


magnus-madsen:
Thanks-- I would think the tests should also be updated to use a region? I think we have done that everywhere else.

Does this eliminate the last use of Static in Array?

stephentetley:
Hi Magnus, I've updated the tests to use a local region.

I think this eliminates `Static` from the API of Array. `Static` is used internally in `copyOfRange`.

magnus-madsen:
Thanks!-- I wonder how close we are to everything being regionified?

PR #4009, opened 2022-06-23
feat: code generation for `NewObject`

paulbutcher:
This is now successfully creating the bytecode for NewObject, creating an instance of it, and calling the constructor.

No support for method implementations yet, or naming the anonymous class.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/jvm/GenAnonymousClasses.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/jvm/GenAnonymousClasses.scala
@@ -18,14 +18,15 @@ package ca.uwaterloo.flix.language.phase.jvm
 
 import ca.uwaterloo.flix.api.Flix
 import ca.uwaterloo.flix.language.ast.ErasedAst._
-import ca.uwaterloo.flix.language.phase.jvm.JvmName.RootPackage
+import ca.uwaterloo.flix.language.phase.jvm.JvmName.{MethodDescriptor, RootPackage}
 import ca.uwaterloo.flix.util.ParOps
+import org.objectweb.asm
 import org.objectweb.asm.Opcodes._
+import org.objectweb.asm.{ClassWriter, Label}
 
 /**
   * Generates bytecode for anonymous classes (created through NewObject)
   */
-
 object GenAnonymousClasses {
 
   /**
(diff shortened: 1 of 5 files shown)


JonathanStarup:
Looks good to me 👍

magnus-madsen:
This looks reasonable to me 👍 

We should add more tests - now or later - e.g. with a few interfaces and trying to call `tostring` (as done) but also hashCode and equals (and verifying that give correct results).

magnus-madsen:
What I mean is:

- Use multiple interfaces.
- Test that we can call toString, equals and hashCode,
- Check that toString equals and hashCode work.
- Construct multiple objects and check that equals and hashCode work as expected.

paulbutcher:
100% agree that we need (a lot!) more in the way of tests. But it strikes me that those become most useful when we add the ability to implement methods (so we will need tests to test all the different types of argument/result those methods might take/return). This is part of why it would be helpful for me to be able to create my own interface (because I can ensure that it defines methods with all the different signatures I care about, rather than trawling the Java standard library to find things that happen to exercise what I want to test).

magnus-madsen:
Agreed, but we can already test the default java.lang.Object methods.

paulbutcher:
So I've added some additional tests, and they pass.

But, TBH, I'm kinda surprised that they pass given that I'm generating the the same anonymous class with the same name over and over again. In any case they're misleading, because on the face of things I'm instantiating different interfaces, but there's actually only the one `HardcodedAnon` class under the hood.

So they're doing what they seem to be, and won't until we have the naming side of things sorted.

main/src/ca/uwaterloo/flix/language/phase/jvm/GenAnonymousClasses.scala
```diff
@@ -0,0 +1,79 @@
...
+  def gen(objs: Set[Expression.NewObject])(implicit root: Root, flix: Flix): Map[JvmName, JvmClass] = {
+    //
+    // Generate an anonymous class for each object and collect the results in a map.
+    //
+    ParOps.parAgg(objs, Map.empty[JvmName, JvmClass])({
+      case (macc, obj) =>
+        val className = JvmName(RootPackage, "HardcodedAnon")
+        flix.subtask(className.toInternalName, sample = true)
+
+        macc + (className -> JvmClass(className, genByteCode(className, obj)))
```
JonathanStarup, line 44: Since this is the same `className` each time, only the last class get produced in the end

paulbutcher: Indeed. This will obviously change once we've worked out what to do about naming, but @magnus-madsen suggested that I don't think about that yet. I think we have to next though!

JonathanStarup:
I was considering how this would work
```
def test(x: Int32): ##java.lang.Object = {
    object ##java.lang.Object {
        override def hashCode(): Int32 = x
    }
}
```
This can't just be a static class since x is not statically known. So this would presumably be translated to something like
```
def test(x: Int32): ##java.lang.Object = {
    def hashCode() = x;
    new test$local(hashCode);
}
```
```
class test$local extends ##java.lang.Object {
    test$local(clo: <clo type>) {
        this.clo=clo;
    }

    override int hashCode() {
        return <call clo>;
    }
}
```

paulbutcher:
Yes, that makes sense, @JonathanStarup 👍

Issue #3589, opened 2022-04-24
Refactor `Iterable` and `Iterator` to use regions


magnus-madsen:
Starting point:

```
pub class Iterable2[t: Type -> (Bool -> Type)] {

    pub def iterator[t: Type -> (Bool -> Type), a: Type, r: Region](t: t[a, r]): Iterator2[a, r] & r

}

pub enum Iterator2[a: Type, r: Bool] {
    case Iterator2(Done2[r], Next2[a, r])
}

type alias Done2[ef: Bool] = Unit -> Bool & ef

type alias Next2[a: Type, ef: Bool] = Unit -> a & ef

```

The ugly signature on `iterator` seems necessary.

jaschdoc:
What do we do with instances of `Iterable` with kind `Type -> Type`? Remove the instances?

magnus-madsen:
> What do we do with instances of `Iterable` with kind `Type -> Type`? Remove the instances?

Good question. @mlutze Thoughts?

mlutze:
We discussed this a bit, but I'm not sure we came to any satisfying conclusion. The main things we considered were to only support `Iterable` for mutable types, or to introduce an alternative `Iterable` for pure types.

<details>
<summary> (MPTCs might help though) </summary>

```
pub class Iterable[reg: Bool, elm: Type, a: Type] {
    pub def iterator(t: a): Iterator[elm, reg] & reg
}

pub instance Iterable[Pure, a, List[a]] {
    pub def iterator(l: List[a]): Iterator[a, Pure]
}
```

</details>

magnus-madsen:
> 

Will this be the end of my resistance? 🌥️

magnus-madsen:
Note that this:

```
pub instance Iterable[Pure, a, List[a]] {
    pub def iterator(l: List[a]): Iterator[a, Pure]
}
```

is not sound because an iterator is not pure.

magnus-madsen:
We can continue to implement Iterable for mutable collections, but I don't see how we can make it work for immutable collections.

jaschdoc:
Another (perhaps slightly worse) solution would be to have `List` and `IterableList` which is just a wrapper around `List` i.e.
```flix
opaque type IterableList[a: Type, r: Region] = ScopedRef[List[a], r]
```
or simply
```flix
opaque type IterableList[a: Type, _: Region] = List[a]
```

magnus-madsen:
Maybe this?:

```
[...]

PR #4011, opened 2022-06-23
refactor: introduce new region-based `Iterator`

jaschdoc:
Related to #3589


jaschdoc:
RFR

jaschdoc:
Here's a brief overview of this (big) PR:
- Renamed old non-region based `Iterator` -> `StaticIterator`.
- Kept `StaticIterator` related functions in a few places to avoid breaking compatibility with `Iterable` TC until that issue is resolved (i.e. refactored to use regions).
- Renamed `RegionIterator` and related functions to `Iterator`.
- Note that *any* (region-based) `Iterator`-related function that also takes a region as an argument now has the region as the first argument. I.e. `pub def toIterator(r: Region[r], l: List[a]): ...`

magnus-madsen:
Massive! Thanks :)

Issue #3731, opened 2022-05-12
Removing Scheme (mostly)


mlutze:
We carry a scheme from `NamedAst` through to `TypedAst`, where thereafter we carry 2 schemes (inferred and declared).
The scheme is a useful abstraction for managing typing, but is a burden for everything else, as it is an extra AST node that is redundant with the parameters + return type + effect.

I think we can remove Scheme at least from the early ASTs without doing any harm. I think it's also likely that we can remove it even from later ASTs, only using it as needed in the Typer.

magnus-madsen:
OK, agreed.

magnus-madsen:
Status?

mlutze:
Still WIP. Taking it incrementally.

magnus-madsen:
Status?

mlutze:
I think we're in a much better spot than we were. There's still a bit of weirdness around enums, but I don't plan on revisiting it unless it annoys me again :)

PR #4020, opened 2022-06-24
refactor: remove scheme from specs in pre-typer phases

mlutze:
related to #3731


magnus-madsen:
Thanks!- nice cleanup

Issue #3582, opened 2022-04-23
let rec: remove redundant inner functions


jaschdoc:
When the function is recursive without a helper function it should just call itself and let the compiler do its magic, rather than having a `loop` function in its body.

- [x] Array
- [x] Chain
- [x] Iterator
- [x] DelayList
- [x] List
- [x] Map
- [x] MutList
- [x] Nel
- [x] Option
- [x] Result
- [x] Set
- [x] String
- [x] Validation
- [x] RedBlackTree

jaschdoc:
@magnus-madsen `Iterator` is a special case here since each function has to destructure the iterator on each recursive call. So which implementation do you prefer (with `foreach` as an example of directly recursive functions`)?
```flix
    pub def foreach(f: a -> Unit & ef, iter: Iterator[a]): Unit & Impure =
        let Iterator(done, next) = iter;
        if (done())
           ()
        else {
            f(next());
            foreach(f, iter)
        }
```
or
```flix
    pub def foreach(f: a -> Unit & ef, iter: Iterator[a]): Unit & Impure =
        let Iterator(done, next) = iter;
        def loop() = {
            if (done())
                ()
            else {
                f(next());
                loop()
            }
        };
        loop()
```

PR #4023, opened 2022-06-24
chore: prettify `MutList.flix`

jaschdoc:
I was going over `MutList` while working on #3582 and decided to reduce some of the verbosity in `MutList` that had been introduced when adding regions.


magnus-madsen:
Thanks; this is much appreciated!

PR #4026, opened 2022-06-24
refactor: clean up PatternExhaustiveness

### LINKED ISSUE #3426: Maximally productize Validations
In several places we misuse the syntactic sugar of `Validation.flatMap`. As a result, we don't maximally collect the errors when they occur.

For example, in the Weeder we have this:
```scala
  /**
    * Performs weeding on the given class declaration `c0`.
    */
  private def visitClass(c0: ParsedAst.Declaration.Class)(implicit flix: Flix): Validation[List[WeededAst.Declaration.Class], WeederError] = c0 match {
    case ParsedAst.Declaration.Class(doc0, ann0, mods0, sp1, ident, tparam0, superClasses0, lawsAndSigs, sp2) =>
      val loc = mkSL(sp1, sp2)
      val doc = visitDoc(doc0)
      val laws0 = lawsAndSigs.collect { case law: ParsedAst.Declaration.Law => law }
      val sigs0 = lawsAndSigs.collect { case sig: ParsedAst.Declaration.Sig => sig }
      for {
        ann <- visitAnnotations(ann0)
        mods <- visitModifiers(mods0, legalModifiers = Set(Ast.Modifier.Public, Ast.Modifier.Sealed, Ast.Modifier.Lawless))
        sigs <- traverse(sigs0)(visitSig)
        laws <- traverse(laws0)(visitLaw)
        tparam = visitTypeParam(tparam0)
        superClasses <- traverse(superClasses0)(visitTypeConstraint)
      } yield List(WeededAst.Declaration.Class(doc, ann, mods, ident, tparam, superClasses, sigs.flatten, laws.flatten, loc))
  }
```

Because we use the `for` syntax, which desugars to `flatMap`, if there's an error with the annotations and with the mods, for example, then we only retain the error for the annotations.

We should go through all the front end phases and identify places where these `flatMap`s can be replaced by `product`s.

We may also want to consider removing `flatMap` from `Validation` altogether (since it's not really a monad).

- [x] ~Reader~
- [x] ~Parser~
- [x] Weeder
- [x] Namer
- [x] Resolver
- [x] Kinder
- [x] ~Deriver~
- [x] Typer
- [x] ~Statistics~
- [x] Instances
- [x] Stratifier
- [ ] PatternExhaustiveness
- [x] Redundancy
- [x] Safety

### LINKED ISSUE #4024: Exhaustivity not checked in match head
```scala
def f(): String = {
    match (match Nil { case _ :: _ => Nil}) {
        case Nil => "it is Nil"
        case _ => "it is not Nil"
    }
}
```

This should throw an error because the nested match is not exhaustive. Instead it crashes at runtime.

mlutze:
Refactors the PatternExhaustiveness phase:
- uses `Validation[Unit, _]` instead of passing around unused AST nodes
- productizes validations so that we can catch multiple errors at once (fixes #3426)
- changes some names to follow convention
- removes some unused vars
- marks private defs private
- checks inside nested patterns (fixes #4024)

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/PatternExhaustiveness.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/PatternExhaustiveness.scala
@@ -23,11 +23,8 @@ import ca.uwaterloo.flix.language.ast.TypedAst.{Expression, Pattern}
 import ca.uwaterloo.flix.language.ast.ops.TypedAstOps
 import ca.uwaterloo.flix.language.ast.{Type, TypeConstructor, TypedAst}
 import ca.uwaterloo.flix.language.errors.NonExhaustiveMatchError
-import ca.uwaterloo.flix.language.phase.PatternExhaustiveness.Exhaustiveness.{Exhaustive, NonExhaustive}
 import ca.uwaterloo.flix.util.Validation._
 import ca.uwaterloo.flix.util.{InternalCompilerException, Validation}
-
-import scala.Function.const
 
 /**
   * The Pattern Exhaustiveness phase checks pattern matches for exhaustiveness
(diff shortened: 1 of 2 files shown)


mlutze:
blocked by #4027

PR #4028, opened 2022-06-24
refactor: use DeepMap in NamedAst

### LINKED ISSUE #3726: NamedAst: introduce `NestedMap` or `NamespaceMap` or some such structure
In the Namer we have a big block where many times we have to access and modify a map within a map, leading to code that looks like 
```
                    prog0.copy(
                      classes = prog0.classes + (ns0 -> (classes0 + (ident.name -> clazz))),
                      defsAndSigs = prog0.defsAndSigs + (sigNs -> defsAndSigs))
```

It's difficult to read and easy to make a mistake. I'd like to introduce some new structure that makes the nested map easier to manipulate.

mlutze:
fixes #3726


mlutze:
@magnus-madsen RFR

magnus-madsen:
I convinced myself this is a bridge too far. We can revisit if you feel strongly about it.

PR #4037, opened 2022-06-25
feat: add `{DelayList, MutList}.shuffle`

ninaandrup:
Related to https://github.com/flix/flix/issues/3697


magnus-madsen:
Iterator- due to its nature - should not have a shuffle.

magnus-madsen:
Otherwise looks fine.

ninaandrup:
These are the rest of the collections, where both the collection has a `toArray` function, and `Array` has a function that converts to the specific collection. Should it close the ticket then?

magnus-madsen:
> These are the rest of the collections, where both the collection has a `toArray` function, and `Array` has a function that converts to the specific collection. Should it close the ticket then?

What do you mean? I don't understand :)

PR #4049, opened 2022-06-27
release 0.29.0

magnus-madsen:
```
Version 0.29.0:

- Language: Added syntactic sugar for singleton enums (thanks Matt!)
- Language: Added a specific error message for when a function is under applied (thanks Jonathan!)
- Language: Added a specific error message about missing dependencies (thanks Matt!)
- Compiler: Improved type and effect formatting (thanks Matt!)
- Library: Refactored integer namespaces to use subject-last (thanks Nina!)
- Library: Refactored all sort functions to internally use arrays (thanks Nina!)
- Library: Refactored more functions to be effect polymorphic (thanks Nina!)
- LSP: Added support for auto-completion of types (thanks Paul!)
- LSP: Improved auto-completion in pipelines (thanks Paul!)
- LSP: Improved auto-completion prioritization (thanks Paul!)
- Tooling: Resurrected the REPL with more improvements to come (thanks Magnus!)
```

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Changelog
 
+Version 0.29.0:
+
+- Language: Added support for `opaque` enums (thanks Matt!)
+- Language: Added a specific error message for when a function is under applied (thanks Jonathan!)
+- Language: Added a specific error message about missing dependencies (thanks Matt!)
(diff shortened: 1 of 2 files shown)


magnus-madsen:
@mlutze @JonathanStarup Release notes OK?

mlutze:
I would wait on the `opaque` keyword until next release, when we know it works as expected, since I think I only had negative tests for it.

PR #4053, opened 2022-06-27
In-progress REPL simplification/rationalisation

paulbutcher:
This is definitely not ready to be merged, but I'm creating this PR because I'm confused and would appreciate guidance.

I thought that incremental compilation would mean that performing a second compilation after the first would not re-compile a file on disk that had changed unless it was re-added to the `Flix` object with `addSourcePath`. It seems that I'm mistaken about this because with this code if I run `flix repl` to load the current project and then change a file on disk, that file is recompiled.

Looking at the code within `Flix.scala`, I'm not surprised that this doesn't work. I can see that there's a `changeSet` maintained, but files are only ever added to it: nothing ever marks a file as compiled and therefore unchanged.

Am I understanding the intent correctly, but incremental compilation is broken? Or am I misunderstanding the intent?

### DIFF
--- a/main/src/ca/uwaterloo/flix/Main.scala
+++ b/main/src/ca/uwaterloo/flix/Main.scala
@@ -107,10 +107,10 @@ object Main {
       options = options.copy(progress = false)
     }
 
+    val cwd = Paths.get(".").toAbsolutePath.normalize()
+
     // check if command was passed.
     try {
-      val cwd = Paths.get(".").toAbsolutePath.normalize()
-
       cmdOpts.command match {
         case Command.None =>
         // nop, continue
@@ -149,6 +149,12 @@ object Main {
           val result = Packager.test(cwd, o)
           System.exit(getCode(result))
 
+        case Command.Repl =>
+          val source = if (cmdOpts.files.isEmpty) Left(cwd) else Right(cmdOpts.files)
+          val shell = new Shell(source, options)
+          shell.loop()
+          System.exit(0)
+
         case Command.Install(project) =>
           val o = options.copy(progress = false)
           val result = Packager.install(project, cwd, o)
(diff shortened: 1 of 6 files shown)


magnus-madsen:
OK; so I think this PR is trying to do too much. Better to break it down into smaller pieces where we can discuss each issue in detail.

magnus-madsen:
The changeSet contains knowledge about which source inputs have changed and which have not. Its important to maintain it to ensure you get incremental compilation.

Note Flix only does incremental compilation in the same JVM process; not on the file system. That is, you always have to do a full build (typically on startup) and then only after that is the build incremental.

paulbutcher:
So ignore the PR for a bit (it's in-progress and not in a fit state to be reviewed - you're seeing the sausage in the middle of being made).

I'm asking about the behaviour of `changeSet` in `Flix.scala`. I can see that it's *intended* to support incremental compilation, but I don't see how it's actually doing that:

* It's initialised to `ChangeSet.Everything`: https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/api/Flix.scala#L59
* `changeSet.markChanged` is called in `addInput`: https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/api/Flix.scala#L371
* It's passed to various phases within `check`: https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/api/Flix.scala#L464

And ... that's it. Nothing ever removes a file from `changeSet`. So once a file is marked as changed (which every file will be because it's added via `addInput`) it's always changed.

So whatever the intent of `changeSet` is, I can't see how it's achieving it?

paulbutcher:
I could certainly make things work as intended by reading the file and calling `addSourceCode` (this is what the VSCode plugin does too). But I didn't want to go down that road until I understood what was going on. Which I don't.

magnus-madsen:
Right, so we don't fully use it yet:

https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/ast/ChangeSet.scala#L47

What this means is that we always recompile the user source code, but not the library code (once it has been compiled once, because the library code cannot change).

paulbutcher:
Ah! Right. Ok, that makes sense.

So that does mean that incremental compilation isn't quite as incremental as I hoped, and the right solution is to read the source files and call `addSourceCode`, not `addSourcePath`.

magnus-madsen:
The idea is (which is of course not supposed to be the end goal): we only cache files that are guaranteed never the change (i.e. the big standard library). The reason for this is that our current "incremental strategy" (which should really be called a "don't recompile things that cannot change-strategy") does not work because it is insufficient to recompile one file since other (already compiled) files may depend on it incompatible ways.

magnus-madsen:
Not to ruin your flow :) But I think I would prefer if we have: one PR to remove :watch, another PR to introduce `repl` and remove `--interactive`, and then a more extensive PR to resolve the issue with reload/incremental compilation.

PR #4057, opened 2022-06-27
feat: add effects to types

mlutze:
related to #4050


mlutze:
> Can you write a very high level description of the changes?

- Add `Empty` type constructor to denote empty effect sets
- Change arrow types to include effect. Arrow is now `Bool -> Effect -> Type -> Type -> Type`
- Change type formatting to support both purity and effect
- Do some little temporary hacks to make it all work, like setting the effect to `Empty` everywhere

magnus-madsen:
> > Can you write a very high level description of the changes?
> 
>     * Add `Empty` type constructor to denote empty effect sets
> 
>     * Change arrow types to include effect. Arrow is now `Bool -> Effect -> Type -> Type -> Type`
> 
>     * Change type formatting to support both purity and effect
> 
>     * Do some little temporary hacks to make it all work, like setting the effect to `Empty` everywhere

Any negative performance impact (before I merge)?

mlutze:
My darn computer isn't stable enough to have reliable tests, but the spread seems roughly the same.

PR #4061, opened 2022-06-28
chore: remove text argument to remSourceCode

paulbutcher:
remInput just blanks the text of the given input, so requiring the previous text to be passed is unnecessary.


magnus-madsen:
Thanks!

PR #4067, opened 2022-06-28
refactor: set up for effect inference

mlutze:
Adds the inference logic to the Typer for the effect sets.

This currently doesn't do anything because a hack is added to type unification which makes it succeed automatically if the types have kind `Effect`.

~depends on #4057~
related to #4050


magnus-madsen:
I suggest at some point we go through the type rules one-by-one.

PR #3667, opened 2022-05-04
fix: disallow ignoring impure expression values.

JonathanStarup:
Related to #3771.
This contains
* A new error for impure non-unit values in statements (and tests thereof)
* Prettyprinting (unused code I think?)
* A few missing `discard` uses

Sidenote @magnus-madsen, I think its interesting to see what `let _ = ...` uses remain. Some look unsafe (tests might be reduced to nothing by optimization)  and the other part is places of unsafe effect use.

PR #4074, opened 2022-06-28
fix: `List.shuffle` tests

JonathanStarup:
The tests were written like shuffle is destructive. @magnus-madsen this was caught by the check in #3667, for our upcoming discussion


magnus-madsen:
> The tests were written like shuffle is destructive. @magnus-madsen this was caught by the check in #3667, for our upcoming discussion

Thanks!

PR #4076, opened 2022-06-28
experiment: region metrics

JonathanStarup:


magnus-madsen:
(Closed-- but we still have this saved for later).

Issue #3950, opened 2022-06-13
Add `foreach` loop construct


magnus-madsen:

magnus-madsen:
@jaschdoc This should be a fun programming exercise for the internship.

jaschdoc:
Next up:

- [ ] `foreach (x <- xs; y <- ys) exp` which desugars into `foreach(x <- xs) foreach (y <- ys) exp`.
- [ ] `foreach (x <- xs; if x > 5) exp` which desugars into `ForEach.foreach(match x -> if (x > 5) exp else ())`.
- [ ] Use `Iterator.foreach` (refactor tests to use `List.toIterator(r, l)`.
- [ ] Rename `<-` to `in`. The word `<-` is used for channels.
- [ ] Idiomatic way to enumerate over collection with index? `collection.iterator |> Iterator.zipWithIndex`?

magnus-madsen:
I think we should leave `<-` as is for now. We can debate later whether to use a keyword. Lets both foreach and for and then we can see :)

JonathanStarup:
> 

Shouldn't there be some ambiguity with channels (`for (a <- b <- c)`)? Not that it is a very sensible combination. I guess you could retrieve an iterator from a channel `for (a <- <- c)`

jaschdoc:
> we should probably change the syntax of channel ops anyway

Ahh didn't know. Matt and I wanted to rename `<-` to `in` for foreach, so it wouldn't clash / be confused with channel operators.

jaschdoc:
Also, should we add a `collection.enumerate` which just calls `collection.iterator |> Iterator.zipWithIndex`? That's a bit shorter when writing foreach loops.

jaschdoc:
What do you think of this?

```flix
namespace List {
    // ...

    ///
    /// Returns an iterator over `l` along with the index of `x_i`
    /// i.e. the iterator returns `(x_i, i)` where `x_i` is the ith
    /// element in `l`.
    ///
    pub def enumerate(l: List[a]): List[(a, Int32)] =
        List.zipWithIndex(l)
        // Should be `List.toIterator(r) |> Iterator.zipWithIndex`
        // and the function should take a region parameter.

    // ...
}

namespace TestList {

    // ...

    /////////////////////////////////////////////////////////////////////////////
    // enumerate                                                               //
    /////////////////////////////////////////////////////////////////////////////

    @test
    def enumerate01(): Bool = region r {
        let z = ref Nil @ r;
[...]

PR #4092, opened 2022-06-29
feat: add `foreach` expression

jaschdoc:
Related to #3950


main/test/flix/Test.Exp.Foreach.flix
```diff
@@ -0,0 +1,7 @@
+namespace Test/Exp/Foreach {
+
+    @test
+    def testForeach01(): Int32 = // Dummy program that just checks compiler phases
+        foreach (x <- xs) x + 1
```
magnus-madsen, line 5: Now you can write some test cases to experiment with what you think should be allowed :)

main/src/ca/uwaterloo/flix/language/phase/Weeder.scala
```diff
@@ -699,6 +699,10 @@ object Weeder {
         case e => WeededAst.Expression.Discard(e, loc)
       }
 
+    case ParsedAst.Expression.ForEach(sp1, pat, exp1, exp2, sp2) =>
+      val loc = mkSL(sp1, sp2)
```
magnus-madsen, line 703: For simplicity, you can do an unsafe pattern match on `pat` to only handle the case where it is a variable. E.g. something like:

```
val varSym = pat match {
  case Pattern.Var(...) => ..
  case _ => ???
```

Then you recursively visit exp1 and exp2.

Then you need to construct a call `List.foreach(varSym -> exp1, exp2)` which means you need to construct some applies nodes and a Def node.

main/src/ca/uwaterloo/flix/language/ast/ParsedAst.scala
```diff
@@ -704,6 +704,17 @@ object ParsedAst {
...
+    /**
+      * ForEach Expression.
+      *
+      * @param sp1     the position of the first character in the expression.
+      * @param pat     the pattern on the left hand side.
+      * @param genExp  the generator expression.
+      * @param bodyExp the body expression.
+      * @param sp2     the position of the last character in the expression.
+      */
+    case class ForEach(sp1: SourcePosition, pat: ParsedAst.Pattern, genExp: ParsedAst.Expression, bodyExp: ParsedAst.Expression, sp2: SourcePosition) extends ParsedAst.Expression
```
magnus-madsen, line 716: By convention we just call these exp1 and exp2

magnus-madsen:
Very cool!

main/test/flix/Test.Exp.Foreach.flix
```diff
@@ -0,0 +1,63 @@
...
+        let z = ref 0 @ r;
+        let q = ref 1 @ r;
+        let l = (1, 2) :: (3, 4) :: Nil;
+        foreach ((x, y) <- l) {
+            q := deref q + x;
+            z := deref z + x + y
+        };
+        deref z + deref q == 15
+    }
+
```
magnus-madsen, line 62: How about some nested foreach? And how about some more tests to ensure the parsing is as expected?

main/test/flix/CompilerSuite.scala
```diff
@@ -5,6 +5,8 @@ import ca.uwaterloo.flix.util.{FlixSuite, Options}
 class CompilerSuite extends FlixSuite(incremental = true) {
   implicit val options: Options = Options.TestWithLibAll
 
-  mkTestDir("main/test/flix/")
+  // REVERT ME
+  //mkTestDir("main/test/flix/")
```
magnus-madsen, line 9: Dummy code

main/src/ca/uwaterloo/flix/language/phase/Weeder.scala
```diff
@@ -1622,6 +1628,21 @@ object Weeder {
       }
   }
 
+  private def mkLambdaMatch(sp1: SourcePosition, p: WeededAst.Pattern, e: WeededAst.Expression, sp2: SourcePosition)(implicit flix: Flix) = {
```
magnus-madsen, line 1631: Add a comment that explains what this does.

jaschdoc:
RFR

PR #4095, opened 2022-06-29
refactor: use `List` in `HeadPredicate`

JonathanStarup:
related to #3896.

It seems pretty easy from here to make `query` return a list. Should we try this although it will break a lot of things?


magnus-madsen:
> related to #3896.
> 
> It seems pretty easy from here to make `query` return a list. Should we try this although it will break a lot of things?

Let's review this and then make a new PR?

Issue #3989, opened 2022-06-18
Simple improvents to the REPL


magnus-madsen:
- [x] Allow `\` as a line terminator (i.e. for multi-line inputs).
- [x] Allow `\\` as syntax for starting and ending a multi-line input (i.e. with having to type `\` on each individual line).
- [x] Do not trigger a compilation before an expression is entered. 
- [x] Do not print a progress bar (set option on Flix object).
- [x] Fix `Input input cannot be ...`
- [ ] User defined `~/.flix/Prelude.flix`?
- [x] VSCode open REPL thingy.
- [x] Update `Unknown` command text to not suggest `:run` (since it does not exist anymore.)
- [x] Determine whether to start in "file mode", "project model" or "empty mode".
- [x] Add a command that allows adding/reloading a specific file.
- [x] Rename `Shell` to `repl`? Rename command to launch it?
- [x] Add build commands. 
- [x] Maybe show progress bar on first compile?
- [x] Add `:doc`:

paulbutcher:
Will do, although I'll try to get my head around the code generation side of `object` first?

magnus-madsen:
> Will do, although I'll try to get my head around the code generation side of `object` first?

Sure. No rush on either! :)

magnus-madsen:
Oh- we could also allow the user to run build commands, e.g. `build-jar` from the repl!

magnus-madsen:
> > Fix Input input cannot be ...
> 
> @magnus-madsen what did you have in mind here?

The word "input" repeats.

paulbutcher:
> The word "input" repeats.

🤣 I was assuming that there was something deeper going on. NP 👍

paulbutcher:
> Allow \ as a line terminator (i.e. for multi-line inputs).

I've been looking at this, and I can't make up my mind whether JLine is going to be on our side or against us.

The default parser (which is what we're currently using) provides [`eofOnEscapedNewLine`](https://github.com/jline/jline3/blob/1315fc0bde9325baff8bc4035dbf29184b0b79f7/reader/src/main/java/org/jline/reader/impl/DefaultParser.java#L57) which on the face of things does exactly what we want. Here's how the REPL works with it set:

```
flix> 1 + \                                                                     
> 2
3
```

It also provides `eofOnUnclosedQuote` and `eofOnUnclosedBracked` which do something similar with quotes and brackets, for example:

```
flix> "this is a
dquote> multiline string"
this is a                                                                       
multiline string
```

[...]

PR #4097, opened 2022-06-29
chore: don't display progress within REPL apart from :reload

paulbutcher:
Towards #3989


magnus-madsen:
Question: The first compile is pretty slow. Maybe we should show it just for first/initial compile?

magnus-madsen:
We should it on first compile, but not again.

Not when typing an expression or when typing reload.

Issue #3906, opened 2022-06-06
Add `Graph` namespace


magnus-madsen:
TBD

Define an API for a `Graph` namespace backed by Datalog:

```scala
///
/// Returns the transitive closure of the graph `g`.
///
def closure(g: List[(String, String)]): List[(String, String)] = ???

///
/// Returns the vertices that are reachable from the origin `o` in the graph `g`.
///
def reachable(o: String, g: List[(String, String)]): Set[String] = ???

///
/// Returns the vertices that are unreachable from the origin `o` in the graph `g`.
///
def unreachable(o: String, g: List[(String, String)]): Set[String] = ???

///
/// Returns `true` if there is a path from `src` to `dst` in the graph `g`.
///
def isConnected(src: String, dst: String, g: List[(String, String)]): Set[String] = ???

///
/// Returns the strongly connected components of the graph `g`.
///
def stronglyConnectedComponents(g: List[(String, String)]): List[Set[String]] = ???

///
/// Returns the topological sort of the graph `g`.
///
def topologicalSort(g: List[(String, String)]): Option[List[String]] = ???

///
/// ...
///
def isCyclic(g: ...): Bool = ???

///
/// Returns the shortest distance from `src` to `dst` in the graph `g`.
///
def distance(src: String, dst: String, g: List[(String, d, String)]): d with LatticeStuff[d] = ???

///
/// Returns the shortest distance from `o` to every other vertex in the graph `g`.
///
def distances(o: String, g: List[(String, d, String)]): Map[String, d] with LatticeStuff[d] = ???

///
/// Returns the given graph `g` as a GraphViz dot string.
///
def toGraphViz(g: List[(String, d, String)]): String = ???

///
/// Returns the undirected graph of the given directed graph.
///
def toUndirected(...)


/// Minimum spanning tree?
```

In the above examples I have used `List` and `String` but really this should work for any `m[t]` and `t`.

The distance functions probably require some more explanation, but the TLDR is that they require a lattice and a sum function. Maybe this requires a type class too, like:

```
class Distance[t] with AllTheLatticeClassesHere[t] {
  def add(x: t, y: t): t
}
```

Ideas:
- SCC?

We should use `m[(t, t)] with Foldable[m]` as the main representation of a graph (edges).

magnus-madsen:
@JonathanStarup @mlutze @stephentetley @jaschdoc @ninaandrup 

Do you have ideas for additional functions that we would want?

[...]

PR #4099, opened 2022-06-30
feat: add `Graph` namespace

ninaandrup:
Related to #3906


main/test/ca/uwaterloo/flix/library/TestGraph.flix
```diff
@@ -0,0 +1,34 @@
+namespace TestGraph {
+
+    /////////////////////////////////////////////////////////////////////////////
+    // closure                                                                 //
+    /////////////////////////////////////////////////////////////////////////////
+
+    @test
+    def closure01(): Bool & Impure =
```
magnus-madsen, line 8: I was wondering if there would be a way to visualize the graphs?

magnus-madsen: Jonathan did this: https://github.com/flix/flix/blob/master/examples/datalog/connect-graph.flix#L74 but maybe that is too much work.

main/src/library/Graph.flix
```diff
@@ -0,0 +1,92 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+namespace Graph {
+
+    ///
+    /// Returns the transitive closure of the graph `g`.
+    ///
+    pub def closure(g: m[(t, t)]): Set[(t, t)] & Impure with Foldable[m], Boxable[t] =
```
ninaandrup, line 21: Should the result type be `List[(t, t)]` ? I made it `Set[(t, t)]`, because then it is easier to test, but we could just convert to `Set` in the tests themselves.

main/src/library/Graph.flix
```diff
@@ -0,0 +1,92 @@
...
+            Reachable(n2, n2) :- Edge(_, n2).
+            // `n1` can reach `n2` by using an edge.
+            Reachable(n1, n2) :- Edge(n1, n2).
+            // if `n1` can reach `m` and `m` can reach `n2` then `n1` can
+            // also reach `n2`. This adds node pairs to the relational
+            // that is reachable using any number of nodes.
+            Reachable(n1, n2) :- Reachable(n1, m), Reachable(m, n2).
+        };
+        let x = query edges, reachability select (src, dst) from Reachable(src, dst);
+        Array.toSet(x)
```
ninaandrup, line 36: The type of `x` is `Array[(t, t), False]`, which makes the functions `Impure`. I guess that is not what we want, but I was not sure how to wrap it in a region. This does not work:
```
region r {
    let x = (query edges, reachability select (src, dst) from Reachable(src, dst)) @ r;
    ...
}
```

ninaandrup (commented):
@magnus-madsen I have some questions for `Graph`

main/test/ca/uwaterloo/flix/library/TestGraph.flix
ninaandrup: I could try doing something like that :-)
Then I can always stop, if it takes too long

main/src/library/Graph.flix
magnus-madsen: Currently query returns array (which is impure). That will change, but in the mean time, I suggest to use a cast on the  whole thing.

main/test/ca/uwaterloo/flix/library/TestGraph.flix
ninaandrup: I stole some of Jonathans work and made arrows for directions. Is that what you had in mind?

magnus-madsen:
@JonathanStarup Can you review the first iteration?

JonathanStarup:
> @JonathanStarup Can you review the first iteration?

yes, i got distracted by making query return list so this could be easier.

main/src/library/Graph.flix
```diff
@@ -0,0 +1,57 @@
...
+    } as & Pure                         // TODO : delete cast when query uses regions
+
+    ///
+    /// Returns the vertices that are reachable from the origin `o` in the graph `g`.
+    ///
+    pub def reachable(o: t, g: m[(t, t)]): Set[t] with Foldable[m], Boxable[t] = {
```
JonathanStarup, line 41: Maybe its just me but I'd prefer `origin: t` instead of `o`.

main/src/library/Graph.flix
```diff
@@ -0,0 +1,57 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+namespace Graph {
+
+    ///
+    /// Returns the transitive closure of the graph `g`.
+    ///
+    pub def closure(g: m[(t, t)]): Set[(t, t)] with Foldable[m], Boxable[t] = {
```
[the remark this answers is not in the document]
magnus-madsen: I think set is fine. The cost of datalog eats the log n factor anyway.

main/src/library/Graph.flix
magnus-madsen: Ok. Let's call it origin.

main/src/library/Graph.flix
```diff
@@ -0,0 +1,57 @@
...
+    ///
+    /// Returns the vertices that are reachable from the origin `o` in the graph `g`.
+    ///
+    pub def reachable(o: t, g: m[(t, t)]): Set[t] with Foldable[m], Boxable[t] = {
+        let edges = inject g into Edge;
+        let reachability = #{
+            // A node can reach itself.
+            Reachable(o).
+            // Directly reachable.
+            Reachable(n) :- Edge(o, n).
```
JonathanStarup, line 47: This rule is already covered by the next rule since `o` is reachable. I think these examples were made to be clear so they are a bit redundant.

main/test/ca/uwaterloo/flix/library/TestGraph.flix
```diff
@@ -0,0 +1,116 @@
...
+    def graph04(): Set[(Int32, Int32)] =
+        let _graphString =
+            " ┌─┐   ┌─┐ " ::
+            " │1├─→─┤2│ " ::
+            " └─┘   └┬┘ " ::
+            "        ↓  " ::
+            " ┌─┐   ┌┴┐ " ::
+            " │4├─←─┤3│ " ::
+            " └─┘   └─┘ " ::
+            Nil;
```
magnus-madsen, line 44: Looks amazing! 👍

main/test/ca/uwaterloo/flix/library/TestGraph.flix
```diff
@@ -0,0 +1,116 @@
...
+        Set#{(1, 2), (2, 3), (3, 4)}
+
+
```
magnus-madsen, line 47: I think it would be nice to have one or two more graphs that are a bit bigger and have several larger components.

magnus-madsen:
Upwards and onwards 🚀

main/test/ca/uwaterloo/flix/library/TestGraph.flix
mlutze: Hackathon project to generate these?

magnus-madsen: Y

main/test/ca/uwaterloo/flix/library/TestGraph.flix
ninaandrup: I think they will come, when I have to do the strongly connected components and such :-)

PR #4100, opened 2022-06-30
feat: perform type inference for effects

mlutze:
related to #4050

Depends on (hoping this list auto-updates with Github magic **(nope)**):
- #4105


magnus-madsen:
Looks like the fun part has started :)

PR #4102, opened 2022-06-30
test: add region tests

### LINKED ISSUE #3367: Add test cases for regions
Add `Test.Exp.Region.flix` which tries to use regions in various ways.

jaschdoc:
Closes #3367


jaschdoc:
Not sure about the last test. Thought we should have something ridiculous at the end just to have it.

PR #4108, opened 2022-06-30
refactor: make `RedBlackTree.size` tail recursive

### LINKED ISSUE #3582: let rec: remove redundant inner functions

jaschdoc:
Going over the `RedBlackTree` I made some minor stylistic changes like adding white space where needed. Also refactored `size` to run in parallel for large trees.
Closes #3582

### DIFF
--- a/main/src/library/RedBlackTree.flix
+++ b/main/src/library/RedBlackTree.flix
@@ -314,7 +314,7 @@ namespace RedBlackTree {
     ///
     pub def foldRightWithCont(f: (k, v, Unit -> b & ef) -> b & ef, z: b, t: RedBlackTree[k, v]): b & ef = match t {
         case Node(_, a, k, v, b) => foldRightWithCont(f, f(k, v, _ -> foldRightWithCont(f, z, b)), a)
-        case _ => z
+        case _                   => z
     }
 
     ///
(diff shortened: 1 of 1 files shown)


main/src/library/RedBlackTree.flix
```diff
@@ -121,10 +121,16 @@ namespace RedBlackTree {
     /// Returns the number of nodes in `t`.
     ///
     @Time(size(t)) @Space(Int32.log2(size(t)))
-    pub def size(t: RedBlackTree[k, v]): Int32 = match t {
-        case Node(_, a, _, _, b) => 1 + size(a) + size(b)
-        case _                   => 0
-    }
+    pub def size(t: RedBlackTree[k, v]): Int32 =
```
jaschdoc, line 124: Should probably have a parallel annotation

jaschdoc:
Alright. The thought had occurred to me :)

jaschdoc:
RFR

main/src/library/RedBlackTree.flix
```diff
@@ -121,10 +121,12 @@ namespace RedBlackTree {
     /// Returns the number of nodes in `t`.
     ///
     @Time(size(t)) @Space(Int32.log2(size(t)))
-    pub def size(t: RedBlackTree[k, v]): Int32 = match t {
```
magnus-madsen, line 124: Technically, this does not have to be tail recursive because the tree has height O(log n) which means the stack can never be blown.

jaschdoc:
We can undo it if you wish? :)

magnus-madsen:
Nah, whatever.

PR #4110, opened 2022-06-30
feat: stop jline from handling backslashes

paulbutcher:
This stops jline from treating a backslash as an escape character. Before this change:

```
flix> "\n"                                                                      
n                                                                               
flix> def sub(x: Int32, y: Int32): Int32 \ { } = ??? 
Error: Input cannot be parsed.
```

After this change:

```
flix> "\n"                                                                      
                                                                                

flix> def sub(x: Int32, y: Int32): Int32 \ { } = ???
Ok. 
```


magnus-madsen:
Thanks!

PR #4111, opened 2022-06-30
feat: single backslash at the end of the line acts as continuation

paulbutcher:
I will freely admit that this is heading towards "voodoo programming". Jline does an awful lot, and the documentation is unhelpfully sparse, so I've cobbled this together by reading the source and looking at examples of other projects that use jline.

But, it does seem to do the right thing:

```
flix> 1 + \                                                                     
> 2
3                                                                               
flix> def myFunction(x: Int32): Int32 = \
> x + 1
Ok.                                                                             
flix> myFunction(1)
2 
```

### DIFF
--- a/main/src/ca/uwaterloo/flix/runtime/shell/ShellParser.scala
+++ b/main/src/ca/uwaterloo/flix/runtime/shell/ShellParser.scala
@@ -22,7 +22,14 @@ import org.jline.reader.Parser.ParseContext
 import scala.jdk.CollectionConverters._
 
 /**
-  * Minimal implementation of `ParsedLine`, necessary to keep jline happy
+  * Minimal implementation of `ParsedLine`, necessary to keep jline happy.
+  * 
+  * JLine uses these values to implement syntax highlighting, line continuation on 
+  * unclosed string, brackets, etc. Because we have all of this functionality 
+  * switched off, we just need to return something that conforms to the `ParsedLine`
+  * interface.
+  * 
+  * https://github.com/jline/jline3/blob/master/reader/src/main/java/org/jline/reader/ParsedLine.java
   */
 class Parsed(s: String) extends ParsedLine {
 
--- a/main/src/ca/uwaterloo/flix/runtime/shell/ShellParser.scala
+++ b/main/src/ca/uwaterloo/flix/runtime/shell/ShellParser.scala
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2022 Paul Butcher
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ 
+package ca.uwaterloo.flix.runtime.shell
+
+import org.jline.reader.{EOFError, ParsedLine, Parser}
+import org.jline.reader.Parser.ParseContext
+
+import scala.jdk.CollectionConverters._
+
+/**
+  * Minimal implementation of `ParsedLine`, necessary to keep jline happy
+  */
+class Parsed(s: String) extends ParsedLine {
+
+  def wordIndex(): Int = -1
+  def word() = ""
+  def wordCursor(): Int = -1
+  def words() = Nil.asJava
+  def cursor() = -1
+  def line() = s
+}
+
+/**
+  * Implementation of jline's `Parser` to trigger jline's line continuation functionality
+  */
+class ShellParser extends Parser {
+  def parse(s: String, cursor: Int, context: Parser.ParseContext): ParsedLine = {
+    
+    // If the last character in the string is a backslash, throw `EOFError` to trigger line continuation
+    if (s.lastOption == Some('\\'))
+      throw new EOFError(-1, -1, "Escaped new line", "newline")
+
+    new Parsed(s)
+  }
+}
--- a/main/src/ca/uwaterloo/flix/runtime/shell/Shell.scala
+++ b/main/src/ca/uwaterloo/flix/runtime/shell/Shell.scala
@@ -62,6 +62,14 @@ class Shell(sourceProvider: SourceProvider, options: Options) {
   private var isFirstCompile = true
 
   /**
+    * Remove any backslashes from escaped new lines from the given string
+    */
+  def unescapeLine(s: String) = {
+    val escapedLineEnd = raw"\\\n".r
+    escapedLineEnd.replaceAllIn(s, "\n")
+  }
+
+  /**
     * Continuously reads a line of input from the terminal, parses and executes it.
     */
   def loop(): Unit = {
@@ -78,6 +86,7 @@ class Shell(sourceProvider: SourceProvider, options: Options) {
     val reader = LineReaderBuilder
       .builder()
       .appName("flix")
+      .parser(new ShellParser)
       .terminal(terminal)
       .build()
     reader.setOpt(LineReader.Option.DISABLE_EVENT_EXPANSION)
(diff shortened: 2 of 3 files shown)


paulbutcher:
Because jline provides all kinds of nice stuff such as changing the prompt to indicate that you're on a continuation line, handling the history correctly if you hit the up arrow and you have multiline entries in the history, editing the current multiline input if you want to edit the line "before" the line you're currently on, etc. We'd lose all of this if we adopt the approach you suggest.

magnus-madsen:
> Because jline provides all kinds of nice stuff such as changing the prompt to indicate that you're on a continuation line, handling the history correctly if you hit the up arrow and you have multiline entries in the history, editing the current multiline input if you want to edit the line "before" the line you're currently on, etc. We'd lose all of this if we adopt the approach you suggest.

Ah, I see. I never considered all that!

magnus-madsen:
!!!jar

magnus-madsen:
Is it supposed to change the prompt now because I don't see anything that indicates more input is expected?

paulbutcher:
Really? You're seeing something different from me in that case.

What I see is that the prompt is initially `flix>` but if you end that line with a backslash it changes to `>`. Is that not what you see?

magnus-madsen:
image

paulbutcher:
😳

I wonder if this is a difference between Mac and Windows? I'll fire up my Windows machine and see if I see the same thing...

magnus-madsen:
If it works on Mac/Linux that's fine. Can't expect everything to work on Windows :P

(But it is surprising, given that this is the new Windows Terminal.

paulbutcher:
I've confirmed that I do see the same under PowerShell (and the old fashioned Windows Command Prompt too). I've converted this PR to draft and will see if I can work out whether this is a limitation of jline (which honestly I doubt - it's too widely used and I can't imagine that it doesn't support Windows fully) or (more likely) it's me misusing it somehow.

magnus-madsen:
By the way, I am a big fan of Linux/Mac too, and I am happy to add stuff even if only for the benefit of users on those platforms-- as long as it does not break anything on Windows (and in this case there is no breakage, afaik).

magnus-madsen:
Could you paste a screenshot from a platform were it works so I can see? :)

magnus-madsen:
> I notice that quite a few of our dependencies are out of date (jline 3.5.1 dates from 2017, for example) and the most recent (3.21.0) also gives this error. Is there a reason why we're sticking to such elderly libraries?

No specific reason. Stability (questionable?) and Java 11 support (which is our current target). 

But you say that even with the newest version there is no luck? I mean it still works on Windows, so maybe its fine. The experience is just better on Mac/Linux-- thats fine too.

paulbutcher:
So this is turning into a much wider question than the prosaic "how do we support backslashes at the end of a line" 🤣

Where does the requirement for Java 11 support come from? And is it a requirement for Java 11 for the compiler? Or for the output of the compiler? Presumably we could require something more recent for the compiler without ending up generating bytecode that requires a more recent JVM?

My default approach (although this comes from a background in commercial software engineering, which has different tradeoffs from compiler construction) has always been to try to stay pretty close to the most recent versions of all dependencies. The reason for this is that if you discover a bug or a missing feature in a dependency, the (very reasonable!) response from the maintainer is likely to first be: upgrade to the latest release. And if there is a problem and it does get fixed, the fix is unlikely to be back-ported to the old version of the dependency. Sticking with elderly dependencies tends to paint you into a corner where you both miss out on useful features/fixes, and upgrading when you do finally get to it is a major headache (because you're applying many years worth of changes) whereas if you stick close to the most recent versions it becomes a background task with each upgrade typically being relatively easy and constrained.

Regarding the output of the compiler, do we have good reason to suspect that Flix users will not be able to run on more recent JVM versions? One of the nice things about a new language is that we don't have a great deal of legacy to worry about, so what would be the downside of saying you need a recent JVM to run Flix binaries?

Finally, I hear your point about Windows support, but given that jline does support Windows, and the only reason why it doesn't currently do so in our case is because we're on an old JVM version, it seems a pity not to allow Windows to be a first class citizen of the Flix community?

magnus-madsen:
Ah! A philosophical treatise!

First, I don't really hold strong opinions about any of these choices, so I can be swayed. Also, remember some of these choices are historical accidents. 

> Where does the requirement for Java 11 support come from? And is it a requirement for Java 11 for the compiler? Or for the output of the compiler? Presumably we could require something more recent for the compiler without ending up generating bytecode that requires a more recent JVM?

That it is Java 11 is partly historical. The other part is that Debian stable, on which our infrastructure runs, has Java 11. But the idea is to ensure that (1) the JVM bytecode output works in Java 11 and (2) the Flix standard library does not depend on features from newer class libraries. (3) The Flix compiler itself runs on Java 11. That said, the plan for last years has always been that when Java X with Project Loom/Green threads is released we will immediately jump that version X to take advantage. But this is dragging out (on the  Java side). 

> My default approach (although this comes from a background in commercial software engineering, which has different tradeoffs from compiler construction) has always been to try to stay pretty close to the most recent versions of all dependencies.

(1) We try to have as few dependencies as possible. We already have to many IMO. (2) My *major concern* is silent breakage. Despite all our tests, upgrading one of our few libraries could break something that we don't know about and only find out much later (when someone was stopped from using Flix!). Hence upgrading libraries is something we do from time to time, but there is no attempt to follow the latest release. 

I think the rest follows from these observations :)

paulbutcher:
> Ah! A philosophical treatise!

Apologies—I do have a habit of waxing philosophical on occasion.

Understood. I've marked this PR as ready for review again and will leave Windows as a second class citizen for now.

I will just note that in my experience I've found the risk and pain associated with frequent, small dependency upgrades to be significantly lower than that associated with infrequent, large dependency upgrades. But happy to stick with the consensus.

main/src/ca/uwaterloo/flix/runtime/shell/ShellParser.scala
```diff
@@ -0,0 +1,49 @@
...
+  */
+class Parsed(s: String) extends ParsedLine {
+
+  def wordIndex(): Int = -1
+  def word() = ""
+  def wordCursor(): Int = -1
+  def words() = Nil.asJava
+  def cursor() = -1
+  def line() = s
+}
```
magnus-madsen, line 35: All of the above look a bit scary. I hope they cannot lead to IndexOutOfBounds?

main/src/ca/uwaterloo/flix/runtime/shell/ShellParser.scala
```diff
@@ -0,0 +1,49 @@
...
+import org.jline.reader.{EOFError, ParsedLine, Parser}
+import org.jline.reader.Parser.ParseContext
+
+import scala.jdk.CollectionConverters._
+
+/**
+  * Minimal implementation of `ParsedLine`, necessary to keep jline happy
+  */
+class Parsed(s: String) extends ParsedLine {
+
```
magnus-madsen, line 28: extra newline

main/src/ca/uwaterloo/flix/runtime/shell/ShellParser.scala
magnus-madsen: Does JLIne have some documentation that we could borrow here?

main/src/ca/uwaterloo/flix/runtime/shell/ShellParser.scala
paulbutcher: Forgive me, I'm not sure whether this is asking me to add an extra newline, or telling me that there's a newline here that you would like removed?

main/src/ca/uwaterloo/flix/runtime/shell/ShellParser.scala
paulbutcher: The honest answer is I don't know. The jline documentation is frustratingly sparse so I've worked most of this out by reading the jline code and the code of other projects that make use of jline.

However, I *think* that none of the above is ever actually used. It's used by the highlighting code within jline, and the code which performs line continuation if there are unclosed quotes, brackets, etc. Because we have all the above switched off, we just need to return something (anything) that conforms to the `ParsedLine` interface.

At some point in the future, if we could populate the above with something derived from the parboiled parser, then we could switch on syntax highlighting etc. within jline, which might be nice.

I'll add a comment to the above effect, and a link to the relevant portion of the jline source.

main/src/ca/uwaterloo/flix/runtime/shell/ShellParser.scala
magnus-madsen: Just noticed a stray newline.

main/src/ca/uwaterloo/flix/runtime/shell/ShellParser.scala
magnus-madsen: OK. Thanks for the information 👍

magnus-madsen:
Merged!

PR #4112, opened 2022-06-30
refactor: `query` returns list

### LINKED ISSUE #1867: Datalog: What should `query` return?
It currently returns an array, but other options are: `List` or `LazyList`.

Edit it could return some "abstract value" that implements Foldable.

JonathanStarup:
Fixes #1867.


JonathanStarup:
> Question/thought: Should we be using Foldable.XYZ instead of List.XYZ in this refactoring?

Ah 😪 yes, good point

magnus-madsen:
I don't think you can return `m[v] with Foldable[m]`. 

The question is whether use Foldable or not. Advantages: We can change things later. Disadvantage: We teach programmers they have to use Foldable.

magnus-madsen:
Let's sleep on it :)

magnus-madsen:
Wanted to merge, but I merged Graph first leading to some conflicts. Sorry!

JonathanStarup:
> 

Fixed

magnus-madsen:
Thanks!!

Issue #4047, opened 2022-06-26
Review standard library for use of `Static` and `Impure`


magnus-madsen:
We should review the standard library to ensure that every use of `Static` and/or `Impure` is necessary, 
i.e. if regions can be used instead, they should be. 

We are already refactoring the Datalog engine in: https://github.com/flix/flix/issues/3896

But we still have all these occurrences left:

![image](https://user-images.githubusercontent.com/2287953/175834008-67d1b8ab-d6d4-4456-b72d-83022878cbc6.png)

and

![image](https://user-images.githubusercontent.com/2287953/175834020-8679ff72-aa1b-4a2d-ac62-e61f10e2d820.png)

magnus-madsen:
@JonathanStarup Ill make you the "Static and Impure" czar. Basically, if you could figure out which of these can be fixed (and what it would take), and then we can hopefully delegate those tasks to Jakob or Nina. 

Obviously a lot of `Impure` annotations are fine and should stay. For `Static` I would expect most can disappear. Agree?

jaschdoc:
Same issue as #3841?

jaschdoc:
> [ ] (Static import, Impure import) `getArgs` should instead import as `Array[String, r]` in a local region.

The impure import is correct, no? Should be impure to interact with a stateful environment.

magnus-madsen:
> > (Static import, Impure import) `getArgs` should instead import as `Array[String, r]` in a local region.
> 
> The impure import is correct, no? Should be impure to interact with a stateful environment. Do you want it to write to a local region instead?

Function has to be impure, import is correct, but could return a List.

JonathanStarup:
But the program arguments are fixed for the whole program duration, so it should always return an array with the same content (the array returned is even a copy so you can't mess that part up). Thats why it shouldn't be impure imo but just be a regioned array

magnus-madsen:
What's the overall picture now?

JonathanStarup:
## `as & Pure`
22/62 occurrences
* `DelayMap` thread related RBT calls
* `Map` thread related RBT calls
* `Set` thread related RBT calls
36/62 occurrences
* Benchmarking (28)
* Fixpoint (8)

## Impure

magnus-madsen:
I suggest we:

- Leave the DelayMap and RBT stuff alone.
- Come up with a refactoring of Benchmarking (will you propose something?)
- Leave the File stuff alone.
- Leave the Channel stuff alone.
[...]

PR #4119, opened 2022-07-01
refactor: use regions is `File`

jaschdoc:
Related to #4047


jaschdoc:
I'm not sure if this is correct. It doesn't really make sense to have region parameters, when the functions are impure anyway. Thoughts?

magnus-madsen:
I don't think this is the right approach. Instead I think we need a vector type.

PR #4121, opened 2022-07-01
feat: add `UnorderedFoldable` type class

### LINKED ISSUE #3067: Type Classes: Add UnorderedFoldable
This class probably only has a few methods and must make use of commutativity.

jaschdoc:
Closes #3067

Is this what you had in mind?


magnus-madsen:
Look at https://github.com/typelevel/cats/blob/main/core/src/main/scala/cats/UnorderedFoldable.scala

magnus-madsen:
I think there is only one fold: `unorderedFold`.

magnus-madsen:
The unordered part means there is no direction, hence its much more limited what you can do.

jaschdoc:
Alright, I've added some empty functions but I'm not quite sure what to do with these since they all return `Bool` and we have no instance of `CommutativeMonoid` for `Bool`. I've looked at cats and it seems like they have an instance where the `empty` function (the neutral element) returns `false` and the `combine` function is `x or y` but I can't find the instance. What do you reckon we should do here? :thinking: 

Also, do you want the `sum` and `product` functions in here too? We have `CommutativeMonoid` instances for numbers so it seems like functionality we may as well put in here now that `Foldabl` also has those functions.

jaschdoc:
Oh... the boolean monoid instances are right there in the same file! Maybe I'm just blind :joy:

jaschdoc:
RFR

Do we need other instances?

main/src/library/UnorderedFoldable.flix
```diff
@@ -0,0 +1,114 @@
...
+ *  limitations under the License.
+ */
+
+///
+/// A type class for unordered data structures that can be folded.
+///
+pub class UnorderedFoldable[t : Type -> Type] with ForEach[t] {
+
+    ///
+    /// Unordered fold of `t`.
```
magnus-madsen, line 23: Add line break.

main/src/library/UnorderedFoldable.flix
```diff
@@ -0,0 +1,114 @@
...
+    ///
+    pub def count(f: a -> Bool & ef, t: t[a]): Int32 & ef =
+        UnorderedFoldable.foldMap(a -> if (f(a)) 1 else 0, t)
+
+}
+
+namespace UnorderedFoldable {
+
+    @Internal
+    pub enum OrBool(Bool)
```
magnus-madsen, line 84: I would move this (and its instances) into `SemiGroup` and rename:

`OrBool` -> `SemiGroup.Any`
`AndBool` -> `SemiGroup.All`

(Following Haskell).

main/src/library/UnorderedFoldable.flix
```diff
@@ -0,0 +1,114 @@
...
+    ///
+    /// Returns the result of applying `CommutativeMonoid.combine` to all the elements in `t`.
+    ///
+    pub def fold(t: t[a]): a with CommutativeMonoid[a] =
+        UnorderedFoldable.foldMap(identity, t)
+
+    ///
+    /// Returns true if and only if `t` is empty.
+    ///
+    pub def isEmpty(t: t[a]): Bool =
```
magnus-madsen, line 38: I think most instances should override this for efficiency.

magnus-madsen:
I think pretty much every collection can be an instance, right?

jaschdoc:
RFR

magnus-madsen:
Thanks!

PR #4122, opened 2022-07-01
refactor: remove rigidity from type var sym

### LINKED ISSUE #3923: Remove rigidity from TypeVarSym

mlutze:
I replaced it with an `isRegion` parameter to support the `Regions` phase, but at least this makes things less ambiguous and eliminates all the unused `InstantiateMode` stuff. Fixes #3923


magnus-madsen:
A reasonable approach 👍

PR #4138, opened 2022-07-04
feat: add `{unreachable, isConnected}` to `Graph` namespace

ninaandrup:
Related to #3906


magnus-madsen:
@JonathanStarup Will you review?

main/src/library/Graph.flix
```diff
@@ -50,4 +50,35 @@ namespace Graph {
...
+            UnReachable(n) :- Edge(n, _), not Reachable(n).
+        };
+        let x = query edges, reachability, unreachablility select dst from UnReachable(dst);
+        Array.toSet(x)
+    } as & Pure                         // TODO : delete cast when query uses regions
+
+    ///
+    /// Returns `true` if there is a path from `src` to `dst` in the directed graph `g`.
+    ///
+    pub def isConnected(src: {src :: t}, dst: {dst :: t}, g: m[(t, t)]): Bool with Foldable[m], Boxable[t] =
```
JonathanStarup, line 81: We should probably only use origin or src

JonathanStarup:
If we want bigger tests without manually typed expected answers we could test quickcheck kind of properties.

for all src, dst, g: `isConnected(src = src, dst = dst, g) == not Set.contains(dst, unreachable(src, g))`

or

for all origin, g: `reachable(origin, g) == reachable(origin, closure(g))`

etc

magnus-madsen:
> If we want bigger tests without manually typed expected answers we could test quickcheck kind of properties.
> 
> for all src, dst, g: `isConnected(src = src, dst = dst, g) == not Set.contains(dst, unreachable(src, g))`
> 
> or
> 
> for all origin, g: `reachable(origin, g) == reachable(origin, closure(g))`
> 
> etc

I think we can move to that later. For now I am happy with manual tests.

main/src/library/Graph.flix
magnus-madsen: Right, ok. So lets rename the earlier use of `origin` to `src`.

main/src/library/Graph.flix
```diff
@@ -50,4 +50,35 @@ namespace Graph {
...
+            Reachable(origin).
+            // If `origin` can reach `m` and there is an edge from `m` to `n`
+            // then `origin` can also reach `n`. This adds all node pairs to
+            // the relational that are reachable using any number of nodes.
+            Reachable(n) :- Reachable(m), Edge(m, n).
+        };
+        let unreachablility = #{
+            // If there is an Edge to a node that is not Reachable from origin
+            // then it is unreachable.
+            UnReachable(n) :- Edge(_, n), not Reachable(n).
```
magnus-madsen, line 69: I would compute `Node(x) :- Edge(x, _).` and the dual and then use that when writing the UnReachable rule.

main/test/ca/uwaterloo/flix/library/TestGraph.flix
```diff
@@ -1,6 +1,5 @@
 namespace TestGraph {
 
-
     def graph01(): Set[(Int32, Int32)] =
```
JonathanStarup, line 3: It would probably be good to test everything with an empty graph.

JonathanStarup, line 3: It would be good to test everything with an empty graph

main/test/ca/uwaterloo/flix/library/TestGraph.flix
```diff
@@ -85,32 +107,87 @@ namespace TestGraph {
...
+
+    @test
+    def isConnected04(): Bool =
+        let s1 = graph03();
+        Graph.isConnected(src = 1, dst = 3, s1)
+
+    @test
+    def isConnected05(): Bool =
+        let s1 = graph04();
+        Graph.isConnected(src = 2, dst = 4, s1)
```
JonathanStarup, line 190: There should be some tests where isConnected returns false

main/src/library/Graph.flix
```diff
@@ -31,23 +31,51 @@ namespace Graph {
...
         List.toSet(x)
-    }
+
+    ///
+    /// Returns the vertices that are unreachable from the `src` in the directed graph `g`.
+    ///
+    pub def unreachable(src: t, g: m[(t, t)]): Set[t] with Foldable[m], Boxable[t] =
+        let edges = inject g into Edge;
+        let nodes = #{
+            Node(x) :- Edge(x, _).
```
JonathanStarup, line 57: I was wondering about src and dst that are not nodes in any edge and it seems to work fine for all of them. Here `Node(src)` should technically be a fact but it doesn't influence the output. Maybe there should be tests for these cases too?

JonathanStarup (changes requested):

main/src/library/Graph.flix
ninaandrup: Good idea. I didn't even think of that

ninaandrup: Right now `Graph.reachable(1, Set#{}: Set[(Int32, Int32)])` returns `Set#{1}`. Should it return `Set#{}` instead?

magnus-madsen: I think that is correct. The initial node is reachable.

magnus-madsen:
@JonathanStarup Good to go?

magnus-madsen:
🚀

PR #4139, opened 2022-07-04
chore: make resume take a single parameter

mlutze:


magnus-madsen:
Conflicted.

magnus-madsen:
Ah yes, `resume` only takes one argument because its the return of the `do`!

mlutze:
> Ah yes, `resume` only takes one argument because its the return of the `do`!

Yes! For this reason it's very `return`-y. I am wondering at this point whether it wouldn't be nicer to use a `return`-y syntax rather than a functiony syntax. But maybe that obscures too much that it returns a value as well...

PR #4143, opened 2022-07-04
fix: improve let-match error msg

### LINKED ISSUE #3926: Improve error message in let-match
for a program like this:
```
pub def example(): Bool = {
  let (a, _b, _c) = List.foldLeft(
    match (b, acc) -> _i -> (b, acc),
    (true, Nil),
    List.range(0, 100)
  );
  a
}
```
where its easy to get confused about types, it is annoying when the error is actually in the pattern of the let. The error here is:
```
>> Unable to unify the types: '(?, ?)' and '(t173131314, ?, ?)'.

336 |>         let (a, _b, _c) = List.foldLeft(
337 |>             match (b, acc) -> _i -> (b, acc),
338 |>             (true, Nil),
339 |>             List.range(0, 100)
340 |>         );
341 |>         a

mismatched types.

Type One: (Bool, List[t173517959])
Type Two: (t173131314, t173131316, t173131317)
```

I would like it to be clearer that the error is in the pattern match and not the expression itself.

JonathanStarup:
Fixes #3926.

I've made an attempt influencing pattern matches with one case. I didn't use expectTpyeM since I didn't feel like i could blame one or the other.

---

```
pub def example(): Bool = {
  let (a, _b, _c) = List.foldLeft(
    match (b, acc) -> _i -> (b, acc),
    (true, Nil),
    List.range(0, 100)
  );
  a
}
```
```
>> Unable to unify the types: '(t51244, ?, ?)' and '(?, ?)'.

2 |   let (a, _b, _c) = List.foldLeft(
          ^^^^^^^^^^^
          mismatched types.

Type One: (t51244, t51245, t51246)
Type Two: (Bool, List[t126005])
```
The previous error shows the whole match expression.

---

```
pub def example(): Bool = {
  match List.foldLeft(
    match (b, acc) -> _i -> (b, acc),
    (true, Nil),
    List.range(0, 100)
  ) {
    case (a, _b, _c) => a
  }
}
```
```
>> Unable to unify the types: '(t65102, ?, ?)' and '(?, ?)'.

7 |     case (a, _b, _c) => a
             ^^^^^^^^^^^
             mismatched types.

Type One: (t65102, t65103, t65104)
Type Two: (Bool, List[t127449])
```
The previous error shows the whole match expression.

---

```
pub def example(): Bool = {
  match List.foldLeft(
    match (b, acc) -> _i -> (b, acc),
    (true, Nil),
    List.range(0, 100)
  ) {
    case (a, _b, _c) => a
    case (a, _b, _c, _d) => a
  }
}
```
```
>> Unable to unify the types: '(?, ?)' and '(t59417, ?, ?)'.

2 |>   match List.foldLeft(
3 |>     match (b, acc) -> _i -> (b, acc),
4 |>     (true, Nil),
5 |>     List.range(0, 100)
6 |>   ) {
7 |>     case (a, _b, _c) => a
8 |>     case (a, _b, _c, _d) => a
9 |>   }

mismatched types.

Type One: (Bool, List[t127185])
Type Two: (t59417, t59418, t59419)
```
This is unchanged. Fixing this is more complicated because all cases have to unify with the matched expression. I could make it show the first case on which unification fails but maybe that could be misleading.


JonathanStarup:
rfr @magnus-madsen

magnus-madsen:
I think this is a hack. We need a general reusable combinator that we can use throughout the Typer.

magnus-madsen:
I concluded this is not the "right way" to do it.

PR #4145, opened 2022-07-04
feat: port boolean unification for sets

mlutze:
related to #4050


magnus-madsen:
RFR?

magnus-madsen:
> > RFR?
> 
> Yes, at least for the conceptual level. There's a lot of messiness where I haven't replaced every "bool" word with "set" for example. But since lots of adjustments will need to be made, I have not yet gone through those.

OK, so I merge this so you can move on :)

mlutze:
Feel free to add to #4146 (or just tell me to) if there's anything important but not immediately necessary

PR #3954, opened 2022-06-14
Incremental compilation PoC

paulbutcher:
This is definitely not ready to be merged! 

But it is a proof of concept for the quick-and-dirty incremental compilation approach discussed on Gitter recently. The impact is minimal (just a few very minor changes to the language server, no changes to the compiler whatsoever).

And it works:

Screenshot 2022-06-14 at 08 55 27

PR #4159, opened 2022-07-05
feat: use `ResolutionError.UndefinedName` to improve auto-completion

magnus-madsen:


magnus-madsen:
Yikes, IDEA destroyed some formatting :/

paulbutcher:
> Yikes, IDEA destroyed some formatting :/

It did, but it looks like it did that because some of my whitespace was out for some reason. I'll sort to avoid similar issues in the future.

paulbutcher:
OK, I think I see what you have in mind. I'll have a crack at turning this into something that's in a fit state to be merged.

paulbutcher:
Ah, you're thinking of this as an alternative to the `???` (I thought it was a both/and, not an either/or) 👍

magnus-madsen:
> Ah, you're thinking of this as an alternative to the `???` (I thought it was a both/and, not an either/or) 👍

I think this style will eventually be able to cover the scenarios ??? cover and more precisely. (At least if we only consider auto-complete)

paulbutcher:
Understood, and makes sense. In which case ignore my recent comments in #3954

magnus-madsen:
image

magnus-madsen:
I should probably have written something like `closestBefore`

PR #4170, opened 2022-07-06
refactor: make `Fixpoint.Tuple` opaque

JonathanStarup:
Related to #3896.


magnus-madsen:
What's the plan here? I don't see any thing with regions.

JonathanStarup:
> What's the plan here? I don't see any thing with regions.

It is to collect all of the unsafe casts in the tuple namespace instead of all of the use-sites.

main/src/library/Fixpoint/Interpreter.flix
```diff
@@ -16,27 +16,53 @@
 
 use Fixpoint/Ram.{RamStmt, RelOp, RamSym, RamTerm, BoolExp, RowVar, toDenotation};
 use Fixpoint/Ast.Denotation;
+use Fixpoint/Tuple.Tuple;
 
 namespace Fixpoint {
+    namespace Tuple {
```
magnus-madsen, line 22: Why the extra namespace?

magnus-madsen:
> > What's the plan here? I don't see any thing with regions.
> 
> It is to collect all of the unsafe casts in the tuple namespace instead of all of the use-sites.

But will you add a region to the tuple or what?

JonathanStarup:
> > > What's the plan here? I don't see any thing with regions.
> > 
> > 
> > It is to collect all of the unsafe casts in the tuple namespace instead of all of the use-sites.
> 
> But will you add a region to the tuple or what?

I wanted to but it is used as a map key so it has to be pure. It should be a vector.

main/src/library/Fixpoint/Interpreter.flix
JonathanStarup: I thought that Fixpoint.new/Fixpoint.toList would be ambigous and when Vector arrives, it will be easier to replace

PR #4172, opened 2022-07-06
docs: add foreach-loop example

jaschdoc:
Related to #3950


magnus-madsen:
I think we should have two examples that fit in the box on https://flix.dev/ i.e. they need to be a lot shorter.

jaschdoc:
I'll make another one tonight :)

jaschdoc:
Something with arrays and string builders

JonathanStarup:
Maybe there should also be a quick section on Static and how to avoid it?

magnus-madsen:
> Maybe there should also be a quick section on Static and how to avoid it?

I think in the docs not for flix.dev

The purpose of flix.dev is to show cool stuff.

JonathanStarup:
> > Maybe there should also be a quick section on Static and how to avoid it?
> 
> I think in the docs not for flix.dev
> 
> The purpose of flix.dev is to show cool stuff.

I was just thinkink of a short note about regions so curious people know what to look for

jaschdoc:
Oh so we still want to merge this? :jack_o_lantern:

examples/iterating-over-arrays-with-foreach.flix
```diff
@@ -0,0 +1,31 @@
...
+/// Flix supports writing code in an imperative
+/// style if it is preferable to you.
+///
+/// foreach-loops are great for working with
+/// mutable data structures.
+///
+/// This example implements copying an array
+/// in an imperative style.
+///
+def copy!(in: { in :: Array[a, r1] }, out: { out :: Array[a, r2] }): Unit \ { Read(r1), Write(r2) } =
```
jaschdoc, line 11: Should we remove the records to reduce complexity?

jaschdoc:
Added some more examples but I'm not sure how interesting they are. Any ideas?

magnus-madsen:
I think the example should be short; so they fit in the box (or maybe 1.5 of the box).

I think they should show the features, so both a simple loop, use of pattern matching, use if nested-loops, use of two loops in one, and use of the if construct.

magnus-madsen:
For example maybe something like:

```
def main(): Unit & Impure =
    use Array.toIterator;
    region r { 
    let xs = [1, 2, 3, 4] @ r;
    let ys = [1, 2, 3, 4] @ r;
    foreach (x <- toIterator(r, xs); 
             y <- toIterator(r, ys))
        println(x + y)
    }
```

with some comments thrown in.

jaschdoc:
Okay I wasn't sure about the level of complexity I had to demonstrate the feature at. The above example should be doable :)

jaschdoc:
What do you think of these examples (excluding `imperative-style-foreach-loops.flix`)?

magnus-madsen:
Ill take a look, but we can let this brew while you work on the other for loop :)

magnus-madsen:
Do you want to take another stab at these examples? I would just try to simplify the long one and then we can merge them in :)

jaschdoc:
Hmm... the long example just explains the functionality of `foreach` so maybe we should just delete it, as it does not showcase an example but serves more as documentation of the feature (i.e. something we could put in the book).

examples/imperative-style-foreach-loops.flix
```diff
@@ -0,0 +1,184 @@
...
+/// Flix supports writing code in an imperative
+/// style if it is preferable to you.
+///
+/// One of the features related to this is a
+/// foreach-loop.
+///
+/// In this example we explore how this loop
+/// works in Flix.
+///
+def main(): Unit & Impure =
```
magnus-madsen, line 11: should be IO

jaschdoc: Yes, but this is the one I'm suggesting we remove (that's why I didn't touch it) :)

magnus-madsen:
I will merge this once CI completes, but think about what examples you would want for flix.dev and what you would want for doc.flix.dev.

PR #4175, opened 2022-07-06
chore: cleanup whitespace to avoid clashing with IDEA

paulbutcher:
Fix whitespace to avoid unnecessary diffs (e.g. #4159)

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
@@ -77,10 +77,15 @@ object CompletionProvider {
   //
   object Priority {
     def high(name: String) = "1" + name
+
     def boost(name: String) = "2" + name
+
     def snippet(name: String) = "4" + name
+
     def local(name: String) = "5" + name
+
     def normal(name: String) = "7" + name
+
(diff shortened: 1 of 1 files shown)


paulbutcher:
BTW, have we thought about using [Scalafmt](https://scalameta.org/scalafmt/) to standardise our formatting? (I don't have any strong feelings about this either way, just wondering whether it's something that's been considered).

magnus-madsen:
> BTW, have we thought about using [Scalafmt](https://scalameta.org/scalafmt/) to standardise our formatting? (I don't have any strong feelings about this either way, just wondering whether it's something that's been considered).

The informal rule is that we all use Intellij IDEAs formatting (we have open source licenses) but of course there is no CI way to enforce that. Any chance you could want to use IDEA too? :)

paulbutcher:
I did briefly try, and found it heavyweight and slow compared to VSCode (which is where I've spent most of the last 5 years, and for which my fingers are well trained). I've recently been experimenting with the Scala Metals extension for VSCode which seems to give me most of what IDEA would.

But I shouldn't let personal preference get in the way of being a good citizen, so if IDEA is the way to go, I'll force myself to get used to it...

magnus-madsen:
I can give you an open source license for your work on Flix. (DM me the e-mail you want to use on Gitter).

PR #4176, opened 2022-07-06
feat: add `Graph.stronglyConnectedComponents`

ninaandrup:
Related to #3906


main/src/library/Graph.flix
```diff
@@ -50,4 +50,47 @@ namespace Graph {
...
+        };
+
+        let allNodes = query edges, nodes select x from Node(x);
+
+        // Find every node that is part of the strongly connected component with `x`.
+        let f = (x) -> {
+            query edges, reachability, connected select dst from Connected(src, dst) where src == x
+        };
+
+        let f1 = (acc, x) -> {
```
JonathanStarup, line 84: I'm less certain on how to do this part in Datalog. Maybe we could use the same trick just on the sets
```
Scc(;Set#{scc}) :- Node(a), Connected(a; scc)
```
Now the facts of `Scc` will have a single fact which is the set of set of nodes that correspond to the sccs, Maybe `Set[Set[t]]` is also suited for the return type since the components are not inherently ordered.

I'm not sure how slow/fast this is

main/src/library/Graph.flix
```diff
@@ -50,4 +50,47 @@ namespace Graph {
...
+            // If `n1` can reach `n2` and `n2` can reach `n1`
+            // then they are part of the same strongly connected component.
+            Connected(n1, n2) :- Reachable(n1, n2), Reachable(n2, n1).
+        };
+
+        let allNodes = query edges, nodes select x from Node(x);
+
+        // Find every node that is part of the strongly connected component with `x`.
+        let f = (x) -> {
+            query edges, reachability, connected select dst from Connected(src, dst) where src == x
```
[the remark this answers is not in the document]
ninaandrup: Ah, nice! I was wondering how to do this in Datalog, but I just made this such that I could write some tests :-)

main/src/library/Graph.flix
```diff
@@ -50,4 +50,47 @@ namespace Graph {
...
+    ///
+    /// Returns the strongly connected components of the directed graph `g`.
+    ///
+    pub def stronglyConnectedComponents(g: m[(t, t)]): List[Set[t]] with Foldable[m], Boxable[t] =
+        let edges = inject g into Edge;
+        let nodes = #{
+            Node(x) :- Edge(x, _).
+            Node(x) :- Edge(_, x).
+        };
+        let reachability = #{
```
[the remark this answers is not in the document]
ninaandrup: Yes, I was thinking about doing this too, but waited for the other PR to be merged to better structure the code after. But I think using a private helper method for the dublicate code would be a good idea

main/src/library/Graph.flix
ninaandrup: I'll try it out and maybe read up on the lattice stuff on monday

main/src/library/Graph.flix
```diff
@@ -78,4 +79,16 @@ namespace Graph {
...
+    /// Returns the strongly connected components of the directed graph `g`.
+    ///
+    pub def stronglyConnectedComponents(g: m[(t, t)]): List[Set[t]] with Foldable[m], Boxable[t] =
+        let edges = inject g into Edge;
+        let connected = #{
+            // If `n1` can reach `n2` and `n2` can reach `n1`
+            // then they are part of the same strongly connected component.
+            Connected(n1; Set#{n2}) :- Reachable(n1, n2), Reachable(n2, n1).
+        };
+        query edges, nodes(), reachability(), connected select x from Connected(_; x)
```
ninaandrup, line 92: When I query it like this and do not care about the first argument of `Connected(_; x)`, then I only get one instance of each strongly connected component. Is this slow?

main/src/library/Graph.flix
```diff
@@ -15,61 +15,55 @@
...
+    def reachability(): #{ Edge(t, t), Node(t), Reachable(t, t) | r } with Boxable[t] = #{
+        // All nodes can reach themselves.
+        Reachable(n, n) :- Node(n).
+        // If `n1` can reach `m` and there is an edge from `m` to `n2`
+        // then `n1` can also reach `n2`. This adds all node pairs to
+        // the relational that are reachable using any number of nodes.
+        Reachable(n1, n2) :- Reachable(n1, m), Edge(m, n2).
+    }
+
+    def reachabilityFromSrc(src: t): #{ Edge(t, t), Node(t), Reachable(t) | r } with Boxable[t] = #{
```
JonathanStarup, line 32: Node is not required in the type

JonathanStarup, line 32: Node is not required in the type

main/src/library/Graph.flix
JonathanStarup: Oh yeah this is nice, I didn't think of that. I think that this is the same as the set of sets solution though I'm not sure. I will test the speed in a bit.

JonathanStarup (changes requested):

main/src/library/Graph.flix
JonathanStarup: Using this rule to query from `Components(s) :- fix Connected(n; s), if Some(n) == Set.minimum(s).` seems a good bit faster (both using a fully disconnected graph and a fully connected graph)

main/src/library/Graph.flix
```diff
@@ -15,61 +15,55 @@
  */
 namespace Graph {
 
+    def nodes(): #{ Edge(t, t), Node(t) | r } with Boxable[t] = #{
```
ninaandrup, line 18: I'm not sure about the return type of these

main/src/library/Graph.flix
```diff
@@ -78,4 +72,21 @@ namespace Graph {
     pub def isConnected(src: {src :: t}, dst: {dst :: t}, g: m[(t, t)]): Bool with Foldable[m], Boxable[t] =
         reachable(src.src, g) |> Set.exists(x -> dst.dst == x)
 
+    ///
+    /// Returns the strongly connected components of the directed graph `g`.
+    ///
+    pub def stronglyConnectedComponents(g: m[(t, t)]): List[Set[t]] with Foldable[m], Boxable[t] =
```
ninaandrup, line 78: Should the return type be `Set[Set[t]]` for consistency?

JonathanStarup: I'd say so yeah

JonathanStarup:
Looks good 👍

main/src/library/Graph.flix
magnus-madsen, line 18: - Add documentation comments.
- Private functions belong at the bottom of the file or immediately after the public function where they are used :)

ninaandrup: It makes most sense to have the private functions at the bottom of the file, when they are used by multiple public functions, right?

magnus-madsen: Yep

main/src/library/Graph.flix
```diff
@@ -20,62 +20,85 @@ namespace Graph {
...
             // If there is a node that is not Reachable from src then it is unreachable.
             UnReachable(x) :- Node(x), not Reachable(x).
         };
-        let x = query edges, nodes, reachability, unreachablility select dst from UnReachable(dst);
+        let x = query edges, nodes(), reachabilityFromSrc(src), unreachablility select dst from UnReachable(dst);
         List.toSet(x)
 
+    ///
+    /// Helper functions for `reachable` and `unreachable`.
+    /// Returns reachable facts as `Reachable(n)` where nodes `n` that are reachable from `src`.
```
ninaandrup, line 48: I'm not sure about the comments of these private functions.

magnus-madsen: Maybe: `Returns a Datalog program which when give a set of `Edge` facts computes the reachable nodes.`

magnus-madsen:
Thanks!

PR #4179, opened 2022-07-06
fix: add Impure to shuffles set effect

JonathanStarup:
This is why mutlist was counted as region polymorphic @magnus-madsen


magnus-madsen:
OK; can you check the rest of the shuffle functions? "Search in path"?

Issue #3888, opened 2022-06-03
Interoperability: Add an expression that allows the (anon.) implementation of a Java interface/class extension


magnus-madsen:
# Overall Goal: Add new anonymous class/object expression

For example:

```
object ##java.lang.Iterator {
  def hasNext(): Bool = true
  def next(): String = "Hello"
}
```

which will generate a new class called say `Anon$1` which implements the `Iterator` interface and has two methods `hasNext` and `next`. The `object` expression is a Java value of type `##java.lang.Iterator`. (We ignore generics for now-- they are erased anyway.)

I propose to work towards this goal through several milestones:

# Milestone 1

- [ ] Add a new ` NewObject` expression (right after PutStaticField) that takes a list of `1` class and `n` interfaces. 

For example, we should be able to compile and run these expressions (trivial as they may be):

Ex1:
```
object ##java.lang.Object {  }
```

Ex2:
```
object ##java.lang.Serializable {  }
```

Ex3:
```
object ##java.lang.Object, ##java.lang.Serializable, ##java.util.RandomAccess {  }
```

# Milestone 2

TBD

mlutze:
How about `object` for the keyword? Too confusing wrt Scala?

magnus-madsen:
> How about `object` for the keyword? Too confusing wrt Scala?

Could work.

JonathanStarup:
maybe implement, like `implement ##java.lang...`.

also, its only java classes here right? so maybe no ##

paulbutcher:
For reference, this is how Clojure does it (which is very nice and simple):

https://clojure.org/reference/java_interop#_implementing_interfaces_and_extending_classes

paulbutcher:
(makes it very clear that what's going on is very different from Java's `extend`.

magnus-madsen:
@paulbutcher I wrote a first proposal. Very simple, but would still require a principled restructuring throughout the compiler.

paulbutcher:
It is very simple 👍

Playing devil's advocate: What are the use cases where we would want to create an object that implements more than one interface or extends a single class without any interfaces?

The vast majority of cases I can think of would only require a single interface (and in the rare case where you might need to implement more than one, you could always implement a class in Java which implements the interfaces you're interested in, and then extend that class?).

I don't know whether that would make things simpler? But if it would, perhaps worth considering?

magnus-madsen:
[...]

PR #4180, opened 2022-07-06
refactor: temporary solution to naming NewObjects

paulbutcher:
As discussed in #3888

Given that this is a temporary solution, I did the quick-and-dirty thing of using `hashCode` rather than trying to derive something from source line and position. I hope that that's OK.


magnus-madsen:
Thanks!

PR #4186, opened 2022-07-06
feat: show "Not Current" in HoverProvider

### LINKED ISSUE #4040: LSP: Hover (etc) should show (`not current`) if the information is outdated
If compilation is currently failing, but we are showing "old" information from the last successful compilation, we could include that in e.g. hover messages.

jaschdoc:
Closes #4040

What else should display not current? I glanced over the other provider objects and it doesn't seem like anything else should inject "Not Current".
Below is a screenshot of a hover.

Screenshot from 2022-07-06 22-29-25


jaschdoc:
RFR

magnus-madsen:
I guess one question is where it is best to show the text.

magnus-madsen:
Could we add it add the end of the message: `"String & Pure (information may not be current)"`

jaschdoc:
Screenshot of new message and layout

Screenshot from 2022-07-11 20-19-43

jaschdoc:
RFR

Issue #3951, opened 2022-06-13
Add `for-yield` loop construct


magnus-madsen:
Add support for:

```scala
for (pat <- exp) yield exp_body
```

as a `ForYield` node in the AST after `Choose`.

Desugar as follows in the `Weeder`:

```scala
for (pat <- exp) yield exp_body
  ==>
Functor.map(match pat -> exp, exp_body)

for (pat1 <- exp1; pat2 <- exp2) yield exp_body
  ==>
Monad.flatMap(match pat1 -> 
    Functor.map(match pat2 -> exp_body, exp2), exp1)
```

(Note: We should discuss whether to use a different type class than foreach).

TODO: Add `if` construct.

magnus-madsen:
@jaschdoc This should be a fun programming exercise for the internship.

jaschdoc:
Do we want to support multiple yield expressions?

magnus-madsen:
I propose:

```
for (x <- xs; if exp1; y <- ys; if exp2) exp3
```

becomes

```
List.flatMap(x -> 
    if (exp1)
        List.flatMap(y -> 
            if (exp2) 
                Applicative.point(exp3) 
            else 
                MonadZero.empty(), l2)
    else 
        MonadZero.empty(), l1);
```

Does that make sense @jaschdoc ?

jaschdoc:
Yes I believe so :)

jaschdoc:
So no arbitrary nesting of guards?

magnus-madsen:
What do you mean?

jaschdoc:
Nevermind, I thought about it and the question doesn't make sense hehe :)

jaschdoc:
Related issue(s) (will be updated when other issues are created):
#4214

magnus-madsen:
Fixed in master (some issues remain).

PR #4191, opened 2022-07-07
feat: add `for-yield` loop

jaschdoc:
Related to #3951


jaschdoc:
I feel like we could re-use the fragments for the foreach loop

PR #4195, opened 2022-07-07
feat: enable effect hovering

mlutze:
related to #4050 

Depends on:
- #4194 

Some caveats:
1. This cannot yet be tested with VSCode because of https://github.com/flix/vscode-flix/issues/208
2. Effect-related expressions are still erased in TypedAst, so their effects will not be shown

Addressing 2 is next on my TODO list. I'll see if there's an easy fix for 1.


magnus-madsen:
Can we rebase so the diff is easier on the eyes?

magnus-madsen:
!!!jar

Issue #4146, opened 2022-07-05
Effect cleanup


mlutze:
- [ ] Use schemes rather than specialized logic for typing effect operations
- [ ] Figure out what to do about the split-off type variables (e.g. do they go in the displayed scheme or not?)
- [x] Reconcile Bool unification and minimizations for sets and bools
- [x] Bring effects through Typer for VSCode hover, etc.
- [x] Support eff and op syms in VSCode
- [ ] Fix up TODOs
- [ ] Double check pretty printed output in VSCode/API

mlutze:
Superseded by #4236

PR #4197, opened 2022-07-07
feat: add `do`, `trywith`, `without`, and `resume` to `TypedAst`

mlutze:
related to #4050 

Depends on
- #4195


magnus-madsen:
What does this do and what's with the TODOs?

mlutze:
It adds effect-related expressions to the TypedAst. Previously they were erased down to `Unit`, which of course is unhelpful for getting the hover information.

The TODOs are nonessential LSP features, such as highlighting references to an effect, etc. What's important to us right now is just being able to see the effect of an expression.

magnus-madsen:
Could you fix those TODOs. I don't think it should be too hard?

magnus-madsen:
(Just afraid otherwise we forget one or two in the future).

mlutze:
Can I do it as an immediate follow-up? This PR touches a lot of files and I'd prefer a more focused one for the LSP changes.

mlutze:
(I've added to the effect cleanup #4146 to-do list, just in case)

mlutze:
Looking at it more, the LSP stuff is more substantial as we need to add the effect and op declarations (here we have only the expressions)

magnus-madsen:
> Looking at it more, the LSP stuff is more substantial as we need to add the effect and op declarations (here we have only the expressions)

Okay

PR #4198, opened 2022-07-07
feat: add `JavaMethod`  to `NewObject`

paulbutcher:
This is the first step towards the next phase of #3888. I've stubbed out `Weeder` so that things compile, but before wiring this through all the phases, I wanted to check that I'm handling method names correctly.

You suggested that the method name should be of type `Name.Ident`. I've used `String` because `Names.JavaIdentifier` is a `String`.

If we wanted to use `Name.Ident` instead, I could change the name to be `Names.Definition`, but that wouldn't strictly conform to the Java naming rules. Or I could define a new version of `Names.JavaIdentifier` which returns `Name.Ident`? Or...?

I've not included `Documentation`, `Annotations`, or `Modifiers` for methods because I don't think that they're necessary? But please let me know if you disagree. Similarly I've left out `TypeParams` and `OptTypeConstraintList`. We could perhaps add these in eventually, but they seemed like more complication than we need right now (given that we don't strictly need them thanks to type erasure).


magnus-madsen:
> You suggested that the method name should be of type `Name.Ident`. I've used `String` because `Names.JavaIdentifier` is a `String`.

I think either could work, but if you use a `String` you will not have any source locations to report an error with later! (E.g. when you try to implement `toSting` and you discover there is no such function then we will lack a way to place the error under `toSting`. This I suggest to use `Name.Ident`.

> If we wanted to use `Name.Ident` instead, I could change the name to be `Names.Definition`, but that wouldn't strictly conform to the Java naming rules. Or I could define a new version of `Names.JavaIdentifier` which returns `Name.Ident`? Or...?

I would do the following: Add a `Names.JavaMethod` to `Parser` which returns just a `Name.Ident`. I think that is best. 

> I've not included `Documentation`, `Annotations`, or `Modifiers` for methods because I don't think that they're necessary? But please let me know if you disagree. Similarly I've left out `TypeParams` and `OptTypeConstraintList`. 

I agree. I don't think any annotations or modifiers will become relevant until much further down the road.

magnus-madsen:
I will merge this in once CI clears. I propose the following next steps (which take a very incremental approach).

- Step 1: Continue with JavaMethod through all the ASTs all the way until the backend (but do not do any checks or transformations). Do not generate any backend code yet.
- Step 2: In the backend, generate JVM methods for with the ***right signatures*** but where their bodies always `throw null`. 
- Step 3: Refactor to use LambdaLifting on JavaMethod.
- Step 4: Refactor backend to now generate code to call the lifted function.
- Step 5: Add checks to ensure that `NewObject` expressions are legal (e.g. all methods are provided, signatures match, etc.)

PR #4199, opened 2022-07-07
feat: two backslashes to enable multiple line input in REPL

paulbutcher:
Another one of the suggestions in #3989


magnus-madsen:
It seems to work correctly most of the time, e.g.:

image

but what happens here:

image

Why does it repeat the last line?

magnus-madsen:
(Just realized I was testing the "wrong" feature!!)

paulbutcher:
> Why does it repeat the last line?

I've just checked: I see the same behaviour as you on Windows, but it doesn't do this on OSX/Linux. So my guess is that this is another artefact of the fact that we're not really using jline properly on Windows (see the discussion in #4111).

The more I think about this, the more uneasy I become about Windows being a second-class citizen. We should either do the work to allow us to use the most recent version of JLine, or ditch JLine completely, I think.

paulbutcher:
> If you type \\ can you then successfully copy and paste a program from flix.dev into the repl and end with \\?

I can successfully copy and paste multi-line declarations or expressions from flix.dev in to the REPL this way, yes. But not complete programs. 

I don't think that this is a surprise? The code within `Shell.scala` only knows how to handle a single declaration or expression at the moment?

https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/runtime/shell/Shell.scala#L209

PR #4200, opened 2022-07-08
feat: add `--Xno-set-effects`

mlutze:
Now effects are enabled by default.

Related to https://github.com/flix/vscode-flix/issues/208

This will make prototyping a lot easier. We can revisit before release.

### DIFF
--- a/main/src/ca/uwaterloo/flix/Main.scala
+++ b/main/src/ca/uwaterloo/flix/Main.scala
@@ -95,7 +95,7 @@ object Main {
       xnobooltable = cmdOpts.xnobooltable,
       xstatistics = cmdOpts.xstatistics,
       xstrictmono = cmdOpts.xstrictmono,
-      xnoeffects = cmdOpts.xnoeffects
+      xnoseteffects = cmdOpts.xnoseteffects
     )
 
     // Don't use progress bar if benchmarking.
@@ -284,7 +284,7 @@ object Main {
                      xnobooltable: Boolean = false,
                      xstatistics: Boolean = false,
                      xstrictmono: Boolean = false,
-                     xnoeffects: Boolean = false,
+                     xnoseteffects: Boolean = false,
                      files: Seq[File] = Seq())
 
   /**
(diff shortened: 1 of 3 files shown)


magnus-madsen:
Rename to `Xno-set-effects` thanks!

magnus-madsen:
Ill hold of merging this before a few more things are ready and we have that ticket information.

magnus-madsen:
@stephentetley 

We thinking our new effect system is fairly stable now.

Does your projects work with master with `--Xeffects` or alternatively with this PR?

magnus-madsen:
!!!jar

magnus-madsen:
@mlutze Build failure.

magnus-madsen:
Maybe we should release a new Flix stable before flipping the flag?

mlutze:
> Maybe we should release a new Flix stable before flipping the flag?

Yes, good thought

stephentetley:
> @stephentetley
> 
> We thinking our new effect system is fairly stable now.
> 
> Does your projects work with master with `--Xeffects` or alternatively with this PR?

My lower level libs work with effects `--Xeffects` . I'm still working on an refactoring that's got somewhat out of hand, so my higher level libraries are broken due to generally being out-of-date.

magnus-madsen:
> > @stephentetley
> > We thinking our new effect system is fairly stable now.
> > Does your projects work with master with `--Xeffects` or alternatively with this PR?
> 
> My lower level libs work with effects `--Xeffects` . I'm still working on an refactoring that's got somewhat out of hand, so my higher level libraries are broken due to generally being out-of-date.

Oh no. Did we break something fundamental or is it just due to our many recent changes?

stephentetley:
> Oh no. Did we break something fundamental or is it just due to our many recent changes?

Moving to regions was quite a lot of work. But the substantive problem has been the code was up to two years old, so Flix has moved on quite a bit, I'm more proficient with Java so I'd do "wrapper" libraries differently, and of course I'm better at Flix now as well - so there are general design decisions I'd do differently.

magnus-madsen:
@mlutze Conflicted, but I think we are ready.

magnus-madsen:
This PR together with https://github.com/flix/flix/issues/4340 will make us available to move forward 🚀

magnus-madsen:
I would like to enable both effect systems by default and add two (mutually exclusive??) flags to toggle between them.

Perhaps: `--Xno-bool-effects` and `--Xno-set-effects`. 

Maybe they don't have to be exclusive?

PR #4203, opened 2022-07-08
feat: store local env in `UndefinedName`

magnus-madsen:


magnus-madsen:
> It seems reasonable to me. Do we not need to carry this env in `Expression.Var` as well?

We can only have an Expression.Var if it was actually found. In this way `defOrSig` is a bit of a misnomer because it means: Def or Sig or NotFound.

PR #4204, opened 2022-07-08
Further towards NewObject methods

paulbutcher:
Further towards #3888

A very small PR, just wanted to check that I'm heading along the right lines.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/WeededAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/WeededAst.scala
@@ -193,7 +193,7 @@ object WeededAst {
 
     case class PutStaticField(className: String, fieldName: String, exp: WeededAst.Expression, loc: SourceLocation) extends WeededAst.Expression
 
-    case class NewObject(className: String, loc: SourceLocation) extends WeededAst.Expression
+    case class NewObject(className: String, methods: List[JvmMethod], loc: SourceLocation) extends WeededAst.Expression
 
     case class NewChannel(exp: WeededAst.Expression, tpe: WeededAst.Type, loc: SourceLocation) extends WeededAst.Expression
 
@@ -415,6 +415,8 @@ object WeededAst {
   case class CatchRule(ident: Name.Ident, className: String, exp: WeededAst.Expression)
 
   case class HandlerRule(op: Name.Ident, fparams: List[WeededAst.FormalParam], exp: WeededAst.Expression)
+
+  case class JvmMethod(ident: Name.Ident, fparams: List[WeededAst.FormalParam], exp: WeededAst.Expression, tpe: WeededAst.Type, purAndEff: PurityAndEffect, loc: SourceLocation)
 
   case class ChoiceRule(pat: List[WeededAst.ChoicePattern], exp: WeededAst.Expression)
 
--- a/main/src/ca/uwaterloo/flix/language/phase/Namer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Namer.scala
@@ -973,7 +973,7 @@ object Namer {
         case e => NamedAst.Expression.PutStaticField(className, fieldName, e, loc)
       }
 
-    case WeededAst.Expression.NewObject(className, loc) =>
+    case WeededAst.Expression.NewObject(className, methods, loc) =>
       NamedAst.Expression.NewObject(className, loc).toSuccess
 
     case WeededAst.Expression.NewChannel(exp, tpe, loc) =>
@@ -1528,7 +1528,7 @@ object Namer {
     case WeededAst.Expression.PutField(_, _, exp1, exp2, _) => freeVars(exp1) ++ freeVars(exp2)
     case WeededAst.Expression.GetStaticField(_, _, _) => Nil
     case WeededAst.Expression.PutStaticField(_, _, exp, _) => freeVars(exp)
-    case WeededAst.Expression.NewObject(_, _) => Nil
+    case WeededAst.Expression.NewObject(_, _, _) => Nil
     case WeededAst.Expression.NewChannel(exp, _, _) => freeVars(exp)
     case WeededAst.Expression.GetChannel(exp, _) => freeVars(exp)
     case WeededAst.Expression.PutChannel(exp1, exp2, _) => freeVars(exp1) ++ freeVars(exp2)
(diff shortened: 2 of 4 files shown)


main/src/ca/uwaterloo/flix/language/ast/WeededAst.scala
```diff
@@ -416,6 +416,8 @@ object WeededAst {
 
   case class HandlerRule(op: Name.Ident, fparams: List[WeededAst.FormalParam], exp: WeededAst.Expression)
 
+  case class JvmMethod(ident: Name.Ident, fparams: List[WeededAst.FormalParam], exp: WeededAst.Expression, tpe: WeededAst.Type, purAndEff: PurityAndEffect, loc: SourceLocation)
```
magnus-madsen, line 419: Lets place this immediately before `CatchRule` in all of the following ASTs.

magnus-madsen (commented):
This looks correct  to me.

paulbutcher:
In which case, hold off merging this, I'll add more phases to this PR 👍

paulbutcher:
@magnus-madsen, I've pushed changes up to the Resolver. I'm not 100% confident that I'm handling regions correctly in the resolver (TBH, I'm not sure I understand regions yet), so please let me know what you think.

paulbutcher:
And now changes to Kinder. I'm definitely getting into the realms of voodoo programming and just following the pattern of what I see elsewhere.

But hey, it typechecks, so it must be right. Right?

magnus-madsen:
Its just about calling the usual functions :)

main/src/ca/uwaterloo/flix/language/phase/Namer.scala
```diff
@@ -1762,6 +1764,21 @@ object Namer {
       }
   }
 
+  /**
+    * Translates the given weeded JvmMethod to a named JvmMethod.
+    */
+  private def visitJvmMethod(method: WeededAst.JvmMethod, uenv: UseEnv, tenv: Map[String, Symbol.UnkindedTypeVarSym])(implicit flix: Flix): Validation[NamedAst.JvmMethod, NameError] = {
+    flatMapN(traverse(method.fparams)(visitFormalParam(_, uenv, tenv))) {
+      case fparams =>
+        val exp = visitExp(method.exp, getVarEnv(fparams), uenv, tenv)
```
magnus-madsen, line 1773: Thus here you would take the enviroment `env` and extend it with the result of getVarEnv. `(++)`

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -1188,6 +1190,16 @@ object Resolver {
 
       }
 
+      def visitJvmMethod(method: NamedAst.JvmMethod, taenv: Map[Symbol.TypeAliasSym, ResolvedAst.TypeAlias], ns0: Name.NName, root: NamedAst.Root)(implicit flix: Flix): Validation[ResolvedAst.JvmMethod, ResolutionError] = {
```
magnus-madsen, line 1193: Add a doc comment.

magnus-madsen:
I suggest to fix these issues and then we can do a merge :)

paulbutcher:
I've pushed changes which I believe address your code review comments.

I've been trying to get my head around the typer. I'll persevere for a while longer, but may need you to answer some questions for me.

magnus-madsen:
@mlutze Can you do a quick pass to check I did not miss anything that might bite us later?

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -1188,6 +1190,19 @@ object Resolver {
...
       }
 
+    /**
+      * Performs name resolution on the given JvmMethod `method` in the namespace `ns0`.
+      */
+      def visitJvmMethod(method: NamedAst.JvmMethod, taenv: Map[Symbol.TypeAliasSym, ResolvedAst.TypeAlias], ns0: Name.NName, root: NamedAst.Root)(implicit flix: Flix): Validation[ResolvedAst.JvmMethod, ResolutionError] = {
+        val fparams = resolveFormalParams(method.fparams, taenv, ns0, root)
+        val exp = visitExp(method.exp, None)
+        val tpe = resolveType(method.tpe, taenv, ns0, root)
+        val purAndEff = semiResolvePurityAndEffect(method.purAndEff, ns0, root)
```
mlutze, line 1200: We should be fully resolving this, not semi-resolving it.

mlutze (commented):
One important comment about `semiResolveX` I added inline.

Some less-vital stylistic things:
* In several places there is deconstructing of a case class by accessing its fields by name (e.g. `method.tpe`. In general we prefer to pattern match.
* The naming convention in the front-end is to name validations as `xVal`, then map over all the `xVal`s.

With those two style points, most of our recent code has approximately this shape:

```scala
def visitX(x: A.X): Validation[B.X, Error] = x match {
  case A.X(p, q0, r0) =>
    val qVal = visitQ(q0)
    val rVal = visitR(r0)
    mapN(qVal, rVal) {
      case (q, r) => B.X(p, q, r)
    }
}
```

main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
```diff
@@ -1193,6 +1195,16 @@ object Kinder {
       }
   }
 
+  private def visitJvmMethod(method: ResolvedAst.JvmMethod, kenv: KindEnv, senv: Map[Symbol.UnkindedTypeVarSym, Symbol.UnkindedTypeVarSym], taenv: Map[Symbol.TypeAliasSym, KindedAst.TypeAlias], henv: Option[(Type.KindedVar, Type.KindedVar)], root: ResolvedAst.Root)(implicit flix: Flix) = method match {
+    case ResolvedAst.JvmMethod(ident, fparams, exp, _, purAndEff, loc) =>
+      val fparamsVal = traverse(fparams)(visitFormalParam(_, kenv, senv, taenv, root))
+      val expVal = visitExp(exp, kenv, senv, taenv, henv, root)
+      val purAndEffVal = visitPurityAndEffect(method.purAndEff, kenv, senv, taenv, root)
+      val tpeVal = Type.freshVar(Kind.Star, loc.asSynthetic)
```
paulbutcher, line 1203: I'm not convinced that the above is correct (I'm ignoring type type that's in `ResolvedAst.JvmMethod`?).

If this is right, can someone explain to me what's going on. And if it's wrong, what should it be instead?

mlutze: Indeed, you should be visiting the type of the JvmMethod node. Something like `val tpeVal = visitType(tpe, Kind.Star, <all the envs>)`

paulbutcher (commented):
Think we might finally be there.

Issue #2163, opened 2021-08-22
Formatting of types with unit


magnus-madsen:
We should fix this somehow:

![image](https://user-images.githubusercontent.com/2287953/130350950-9d812a42-1855-4e2b-842e-1c04d80035c5.png)

magnus-madsen:
I think we need `FormatFormalParams`.

magnus-madsen:
@mlutze Did we solve this now?

mlutze:
> @mlutze Did we solve this now?

Looks like it is fixed for hover, but not autocomplete.

magnus-madsen:
Maybe @nicdard could take a peek into `HoverProvider` and `CompleteProvider` and see how they differ?

PR #4207, opened 2022-07-09
add `:doc` command to REPL

paulbutcher:
This implements another of the suggestions within #3989

Draft for now because there are some unanswered questions.

This works reasonably nicely:

Screenshot 2022-07-09 at 11 19 13

But...

I haven't found a way to display the type of a class (so all that's displayed is the documentation):

Screenshot 2022-07-09 at 11 21 27

And it would probably make sense to display its signatures (in the same way as does api.flix.dev).

I *have* managed to find a way to display the signature of an enum, but that relies on using `tpeDeprecated` which probably isn't a good thing?

And the formatting can get messy:

Screenshot 2022-07-09 at 11 24 34

So, to do this properly, we probably need to:

* Do something about formatting markdown nicely for terminal output
* Extract some of the code within the documentor phase to re-use here. Or run the documentor when compiling within the repl and make the results available?

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/Symbol.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/Symbol.scala
@@ -136,6 +136,14 @@ object Symbol {
   }
 
   /**
+    * Returns the class symbol for the given fully qualified name
+    */
+  def mkClassSym(fqn: String): ClassSym = split(fqn) match {
+    case None => new ClassSym(Nil, fqn, SourceLocation.Unknown)
+    case Some((ns, name)) => new ClassSym(ns, name, SourceLocation.Unknown)
+  }
+
+  /**
     * Returns the hole symbol for the given name `ident` in the given namespace `ns`.
     */
   def mkHoleSym(ns: NName, ident: Ident): HoleSym = {
@@ -162,6 +170,15 @@ object Symbol {
     */
   def mkTypeAliasSym(ns: NName, ident: Ident): TypeAliasSym = {
     new TypeAliasSym(ns.parts, ident.name, ident.loc)
+  }
+
+  /**
+    * Returns the type alias symbol for the given fully qualified name
+    */
+
+  def mkTypeAliasSym(fqn: String): TypeAliasSym = split(fqn) match {
+    case None => new TypeAliasSym(Nil, fqn, SourceLocation.Unknown)
+    case Some((ns, name)) => new TypeAliasSym(ns, name, SourceLocation.Unknown)
   }
 
   /**
--- a/main/src/ca/uwaterloo/flix/language/ast/Symbol.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/Symbol.scala
@@ -175,7 +175,6 @@ object Symbol {
   /**
     * Returns the type alias symbol for the given fully qualified name
     */
-
   def mkTypeAliasSym(fqn: String): TypeAliasSym = split(fqn) match {
     case None => new TypeAliasSym(Nil, fqn, SourceLocation.Unknown)
     case Some((ns, name)) => new TypeAliasSym(ns, name, SourceLocation.Unknown)
--- a/main/src/ca/uwaterloo/flix/runtime/shell/Command.scala
+++ b/main/src/ca/uwaterloo/flix/runtime/shell/Command.scala
@@ -36,7 +36,7 @@ object Command {
   case object Reload extends Command
 
   /**
-    * Displays documentation
+    * Displays documentation about the fqn s
     */
   case class Doc(s: String) extends Command
 
--- a/main/src/ca/uwaterloo/flix/runtime/shell/Command.scala
+++ b/main/src/ca/uwaterloo/flix/runtime/shell/Command.scala
@@ -36,6 +36,11 @@ object Command {
   case object Reload extends Command
 
   /**
+    * Displays documentation
+    */
+  case class Doc(s: String) extends Command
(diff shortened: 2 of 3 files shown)


magnus-madsen:
Looks like a good start! 

We need a collection of helper functions, probably to be placed in `flix/language/fmt` for formatting various things. (https://github.com/flix/flix/issues/2163)

We should not use `tpeDeprecated`. Trying to hard to remove that. There should be a `Scheme sc` that can be used.

magnus-madsen:
We could start with just def/sig/enum support.

magnus-madsen:
Btw. does Clojure have something like this? Does it use colors?

paulbutcher:
> Btw. does Clojure have something like this? Does it use colors?

It does, and no it doesn't use colours:

Screenshot 2022-07-10 at 21 17 44

When I said that we should "Do something about formatting markdown nicely for terminal output", I wasn't thinking of colours, I was more thinking about layout (see the mess that `:doc Applicative` creates above). But it looks like Clojure gets around that by not having long lines in the source documentation (whereas we typically have lines longer than 100 characters).

paulbutcher:
> There should be a Scheme sc that can be used.

I'm not seeing it for an Enum? As far as I can see, Specs are only associated with Defs and Sigs?

paulbutcher:
I've removed the reference to `tpeDeprecated` and marked this as ready for review. It's definitely a step forwards, and we can enhance as and when we implement #2163.

Should I create a "reformat documentation comments" issue (that will be a fun task for someone 😳)

main/src/ca/uwaterloo/flix/language/ast/Symbol.scala
```diff
@@ -164,6 +172,15 @@ object Symbol {
     new TypeAliasSym(ns.parts, ident.name, ident.loc)
   }
 
+  /**
+    * Returns the type alias symbol for the given fully qualified name
+    */
+
```
magnus-madsen, line 178: Spurious newline.

main/src/ca/uwaterloo/flix/runtime/shell/Command.scala
```diff
@@ -35,6 +35,11 @@ object Command {
     */
   case object Reload extends Command
 
+  /**
+    * Displays documentation
```
magnus-madsen, line 39: Prints documentation about the definition or signature with fqn `s`.

main/src/ca/uwaterloo/flix/runtime/shell/Shell.scala
```diff
@@ -172,17 +180,57 @@ class Shell(sourceProvider: SourceProvider, options: Options) {
...
+    * Displays documentation for the given identifier
+    */
+  private def execDoc(s: String)(implicit terminal: Terminal): Unit = {
+    val w = terminal.writer()
+    val classSym = Symbol.mkClassSym(s)
+    val defnSym = Symbol.mkDefnSym(s)
+    val enumSym = Symbol.mkEnumSym(s)
+    val aliasSym = Symbol.mkTypeAliasSym(s)
+
+    for(r <- root) {
```
magnus-madsen, line 205: I would prefer if this was an explicit match on whether root is None or Some :) (for clarity)

main/src/ca/uwaterloo/flix/runtime/shell/Shell.scala
```diff
@@ -172,17 +180,57 @@ class Shell(sourceProvider: SourceProvider, options: Options) {
...
+        w.println(FormatDoc.asMarkDown(enumDecl.doc))
+
+      } else if (r.typeAliases.contains(aliasSym)) {
+
+        val aliasDecl = r.typeAliases(aliasSym)
+        w.println(FormatType.formatWellKindedType(aliasDecl.tpe))
+        w.println
+        w.println(FormatDoc.asMarkDown(aliasDecl.doc))
+
+      } else {
```
magnus-madsen, line 230: Remove the extra newlines above.

main/src/ca/uwaterloo/flix/runtime/shell/Shell.scala
```diff
@@ -198,11 +246,12 @@ class Shell(sourceProvider: SourceProvider, options: Options) {
...
     val w = terminal.writer()
 
-    w.println("  Command       Purpose")
+    w.println("  Command       Arguments     Purpose")
     w.println()
-    w.println("  :reload :r    Recompiles every source file.")
-    w.println("  :quit :q      Terminates the Flix shell.")
-    w.println("  :help :h :?   Shows this helpful information.")
+    w.println("  :reload :r                  Recompiles every source file.")
+    w.println("  :doc :d       <identifier>  Displays documentation for <identifier>")
```
magnus-madsen, line 252: Maybe we should say `<fqn>` instead of `identifier` ?

magnus-madsen (commented):
Overall looks good. I just have some nitpicks.

main/src/ca/uwaterloo/flix/runtime/shell/Command.scala
paulbutcher: Done, but it does more than just definitions and signatures (it also handles enums and aliases).

magnus-madsen:
Dang, forgot the `feat:` part :/

magnus-madsen:
Oh well

magnus-madsen:
> > Oh well
> 
> Should we make a bot to nag people about this? :D

Haha. The return of the bots. Sure why not?

PR #4208, opened 2022-07-09
refactor: add consistent naming to `foreach`

jaschdoc:
`ForEach` was not properly formatted.


mlutze:
As long as we are bikeshedding, I'd argue that since we treat the keyword as one word (not `forEach`), the AST nodes should be one-wordy too: `Foreach`

magnus-madsen:
> As long as we are bikeshedding, I'd argue that since we treat the keyword as one word (not `forEach`), the AST nodes should be one-wordy too: `Foreach`

But, we do have `ForYield` also, and `Foryield` would be strange, no?

mlutze:
Yes, but in `ForYield`, they are two separate keywords :)

PR #4209, opened 2022-07-09
refactor: update effect syntax

### LINKED ISSUE #4189: Use set effect syntax in stdlibrary
Change every effect like `ef1 and ef2` to `{ef1, ef2}`

jaschdoc:
Closes #4189

### DIFF
--- a/main/src/library/Nel.flix
+++ b/main/src/library/Nel.flix
@@ -405,7 +405,7 @@ namespace Nel {
     /// Applies `g` to the initial element of `l` and combines it
     /// with the remainder of `l` using `f` going from right to left.
     ///
-    pub def reduceRightTo(f: (a, b) -> b & ef1 \ ef1_, g: a -> b & ef2 \ ef2_, l: Nel[a]): b \ { ef1, ef2 } =
+    pub def reduceRightTo(f: (a, b) -> b & ef1 \ ef1_, g: a -> b & ef2 \ ef2_, l: Nel[a]): b \ { ef1, ef2, ef1_ + ef2_ } =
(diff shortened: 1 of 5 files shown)


main/src/library/Nel.flix
```diff
@@ -409,14 +409,14 @@ namespace Nel {
     /// Applies `g` to the initial element of `l` and combines it
     /// with the remainder of `l` using `f` going from right to left.
     ///
-    pub def reduceRightTo(f: (a, b) -> b & ef1 \ ef1_, g: a -> b & ef2 \ ef2_, l: Nel[a]): b & (ef1 and ef2) \ (ef1_ + ef2_) =
+    pub def reduceRightTo(f: (a, b) -> b & ef1 \ ef1_, g: a -> b & ef2 \ ef2_, l: Nel[a]): b \ { ef1, ef2 } =
```
mlutze, line 412: For this one, we need to keep the weirdness. Otherwise we throw away the set effects from the parameters.

jaschdoc:
@mlutze is this what you intended?

mlutze:
> @mlutze is this what you intended?

No, the `reduceRightTo` shouldn't be changed in this PR.

mlutze:
> > > @mlutze is this what you intended?
> > 
> > 
> > No, the `reduceRightTo` shouldn't be changed in this PR.
> 
> What did you intend :)

Everything else is perfect. It's just `reduceRightTo` that needs special treatment because of its cast.

jaschdoc:
> Do you want to drop the Boolean `and` syntax from the Parser?

Is this a question for @mlutze or me? :)

PR #4211, opened 2022-07-09
refactor: add region to `Fixpoint.Database`

JonathanStarup:
Related to #3896.

Lots of pure casts and impure functions gone in this one 🎉


magnus-madsen:
Nice!

PR #4215, opened 2022-07-11
feat: add `MonadZero`

### LINKED ISSUE #4214: Add `MonadZero` and use it in `For-Yield`
Add a type class:

```
///
/// A type class for Monads that have a zero element.
///
class MonadZero[m: Type -> Type] with Monad[m] {
    pub def empty(): m[t]
}
```

Instances for `Option`, `Result` (?), `List`, `Chain` (what else?)

jaschdoc:
Adds `MonadZero` and instances to `List`, `DelayList`,`Chain`, `Option`.

I tried experimenting with adding `Applicative` and `Monad` instances to `Result` but it seems the `point` function causes problems with the type variable `e` in `Result[t, e]`. Besides, there is no empty element for `Result` so it wouldn't make sense to add a `MonadZero` instance either.

Also, removed a todo from `List` and changed `& e` to `& ef`.

Closes #4214

### DIFF
--- a/main/src/library/MonadZero.flix
+++ b/main/src/library/MonadZero.flix
@@ -1,0 +1,22 @@
+/*
+ *  Copyright 2022 Magnus Madsen
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


main/src/library/MonadZero.flix
```diff
@@ -0,0 +1,22 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions not
+ *  limitations under the License.
+ */
+
+///
+/// A type class for Monads that have a zero element.
+///
+class MonadZero[m: Type -> Type] with Monad[m] {
+    pub def empty(): m[t]
```
magnus-madsen, line 21: ///
/// Returns the zero element of the monad.
///

magnus-madsen:
Looks good.

PR #4233, opened 2022-07-11
feat: add `JvmMethod` up to `TypedAst`

paulbutcher:
I'll leave this as draft, but can you take a look at my changes to the typer to see if I'm along the right lines?


magnus-madsen:
> I'll leave this as draft, but can you take a look at my changes to the typer to see if I'm along the right lines?

Looks OK so far. There are two phases in the Typer:

- The difficult phase with type inference, and
- The easier phase where the AST is reassembled with all type information.

In the type inference phase, you will need to visit all the JavaMethods and do type inference for their body expressions. Then you should check that type of the body is the same as the declared return type. (We could of course leave this check for later, like several other checks that are pending). 

But you will have to do inference on the expression bodies.

paulbutcher:
Aha! The failing test suggests that it is doing at least something close to the right thing 😊

paulbutcher:
Can someone advise what would be the easiest way to check that FormatExpression is generating sensible output?

mlutze:
> Can someone advise what would be the easiest way to check that FormatExpression is generating sensible output?

Probably the easiest thing is to add a little hack at the end of `Typer` to print out the AST. If you run with `--Xlib nix` then you avoid including the standard library.

FormatExpression is used exclusively for debugging, so making it pretty isn't a priority :)

paulbutcher:
Monomorph changes pushed, but I'm not especially confident I understand what's going on here, so please let me know if I'm along the right lines.

main/src/ca/uwaterloo/flix/api/lsp/Indexer.scala
```diff
@@ -322,8 +322,11 @@ object Indexer {
     case Expression.PutStaticField(_, exp, _, _, _, _) =>
       visitExp(exp) ++ Index.occurrenceOf(exp0)
 
-    case Expression.NewObject(_, _, _, _, _) =>
-      Index.occurrenceOf(exp0)
+    case Expression.NewObject(_, _, _, _, methods, _) =>
+      Index.occurrenceOf(exp0) ++ traverse(methods) {
+        case JvmMethod(_, fparams, exp, _, _, _, _) =>
+          Index.traverse(fparams)(visitFormalParam) ++ visitExp(exp)
```
mlutze, line 328: The JVM method type, effect, and purity are all written syntactically, right? In that case you should also visit those.

main/src/ca/uwaterloo/flix/language/dbg/FormatJvmMethod.scala
```diff
@@ -0,0 +1,15 @@
+package ca.uwaterloo.flix.language.dbg
```
mlutze, line 1: Add a license to the top of this (copy-paste from any other file and change the name and date at the top)

main/src/ca/uwaterloo/flix/language/phase/Redundancy.scala
```diff
@@ -598,8 +598,14 @@ object Redundancy {
     case Expression.PutStaticField(_, exp, _, _, _, _) =>
       visitExp(exp, env0, rc)
 
-    case Expression.NewObject(_, _, _, _, _) =>
-      Used.empty
+    case Expression.NewObject(_, _, _, _, methods, _) =>
```
mlutze, line 601: I think it may be worthwhile to add a test for this. There are two things to check for: shadowed variables and unused variables.

Are we at the point where we could add a couple tests to `TestRedundancy.scala`?

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -244,7 +244,7 @@ object Simplifier {
         val e = visitExp(exp)
         SimplifiedAst.Expression.PutStaticField(field, e, tpe, simplifyPurity(pur), loc)
 
-      case TypedAst.Expression.NewObject(clazz, tpe, pur, eff, loc) =>
+      case TypedAst.Expression.NewObject(clazz, tpe, pur, eff, methods, loc) =>
```
mlutze, line 247: This field is ignored on purpose, right?

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1439,11 +1439,22 @@ object Typer {
...
-        val resultTyp = getFlixType(clazz)
-        val resultPur = Type.Impure
-        val resultEff = Type.Empty
-        liftM(List.empty, resultTyp, resultPur, resultEff)
+      case KindedAst.Expression.NewObject(clazz, methods, loc) => {
+        def visitJvmMethod(method: KindedAst.JvmMethod): InferMonad[(List[Ast.TypeConstraint], Type, Type, Type)] = method match {
+          case KindedAst.JvmMethod(ident, fparams, exp, returnTpe, pur, eff, loc) =>
+            for {
+              (constrs, bodyTpe, bodyPur, bodyEff) <- visitExp(exp)
+              _ <- expectTypeM(expected = returnTpe, actual = bodyTpe, exp.loc)
```
mlutze, line 1447: What is the plan for dealing with Purity/Effects in the method bodies?

mlutze (changes requested):

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
magnus-madsen: We have not discussed it yet. I think, as a first step, we should just allow any effect.

magnus-madsen:
Ill let you look over Matt's comments then Ill do a pass and then I think we should indeed to a merge :)

main/src/ca/uwaterloo/flix/language/dbg/FormatJvmMethod.scala
paulbutcher: Done, but I left it out because none of the other files here have a license (I thought that there might be some reason why they weren't used here?)

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
paulbutcher: For now, yes. I'll not ignore it once I add JvmMethod to SimplfiedAst.

main/src/ca/uwaterloo/flix/language/dbg/FormatJvmMethod.scala
mlutze: Because someone forgot :)

main/src/ca/uwaterloo/flix/language/phase/Redundancy.scala
paulbutcher: So I've added a couple of tests, but the shadowed variable test is failing with:

ShadowedVar.NewObject.01 - Expected an error of type ShadowedVar, but found class ca.uwaterloo.flix.language.errors.RedundancyError$UnusedVarSym.

I'm unsure whether that's because:

a) My test is wrong in some way
b) The test is exposing a bug in the code I've added to `Redundancy.scala`
c) We need to finish working through the remaining phases before this test can work.

Advice gratefully received!

main/test/ca/uwaterloo/flix/language/phase/TestRedundancy.scala
```diff
@@ -451,6 +465,17 @@ class TestRedundancy extends FunSuite with TestUtils {
...
 
+  test("UnusedFormalParam.NewObject.01") {
+    val input =
+      """
+        |def f(): ##java.lang.Comparable & Impure =
+        |   object ##java.lang.Comparable {
+        |     def compareTo(x: ##java.lang.Object, _y: ##java.lang.Object): Int32 =
+        |       0
+        |   }
+        """.stripMargin
```
mlutze, line 476: Don't forget to compile and expectError :)

paulbutcher: D'oh!

main/src/ca/uwaterloo/flix/language/phase/Redundancy.scala
paulbutcher: Now that my other test is actually a test, it is also failing, but with an error that makes more sense to me:

UnusedFormalParam.NewObject.01 - Expected an error of type UnusedFormalParam, but found class ca.uwaterloo.flix.language.errors.RedundancyError$UnusedVarSym.

What I'm not sure about is whether the right thing to do is change the test (because that's the right error) or change the code to generate a different error.

mlutze: `UnusedFormalParam` is only created via a call to `findUnusedFormalParams`, which is only called for sigs and defs. I suspect that if you change the signature of this function from taking a `Spec` to taking a `List[FormalParam]`, you should be able to generate the appropriate fparam error.

For the `ShadowedVar` error, you will need to explicitly check for shadowing on each formal parameter with the `shadowing` function. Look at `Expression.Let` for how this is done.

This requires some extra thought since this is the first time we have formal parameters in an environment that already contains other symbols.

paulbutcher: See new version just pushed. The tests pass, which is a good start!

mlutze:
Passing the review baton to @magnus-madsen

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1439,11 +1439,22 @@ object Typer {
...
+            } yield (constrs, returnTpe, bodyPur, bodyEff)
+        }
+
+        for {
+          (constrs, _, _, _) <- seqM(methods map visitJvmMethod).map(unzip4)
+          resultTyp = getFlixType(clazz)
+          resultPur = Type.Impure
+          resultEff = Type.Empty
```
magnus-madsen, line 1455: We can discuss later whether object creation is actually impure or not.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1439,11 +1439,22 @@ object Typer {
...
         } yield (valueConstrs, resultTyp, resultPur, resultEff)
 
-      case KindedAst.Expression.NewObject(clazz, _, loc) =>
-        val resultTyp = getFlixType(clazz)
-        val resultPur = Type.Impure
-        val resultEff = Type.Empty
-        liftM(List.empty, resultTyp, resultPur, resultEff)
+      case KindedAst.Expression.NewObject(clazz, methods, loc) => {
+        def visitJvmMethod(method: KindedAst.JvmMethod): InferMonad[(List[Ast.TypeConstraint], Type, Type, Type)] = method match {
+          case KindedAst.JvmMethod(ident, fparams, exp, returnTpe, pur, eff, loc) =>
```
magnus-madsen, line 1444: @mlutze Do we need to do something here to ensure the formal parameter's type matches or do we get it for free?

mlutze: You're right, we need to check that they match the fparams from the class.

(Probably should also have a test for this :) )

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
paulbutcher: Surely it has to be impure? We're creating a new object, which has a different identity. [This test](https://github.com/flix/flix/blob/master/main/test/flix/Test.Exp.Jvm.NewObject.flix#L32) demonstrates that it has to be impure, yes?

magnus-madsen: Yes. Ok you are right.

PR #4238, opened 2022-07-12
fix: `File.{readWith, readBytesWith}` bug caused by `skip`

### LINKED ISSUE #4006: Issue regarding `File.readWith` and its use of `skip`
As discussed in #3984 `readWith` might not give data from the correct offset since `skip` can possibly not `skip` all of the requested bytes. From doc: ["The skip method may, for a variety of reasons, end up skipping over some smaller number of bytes, possibly 0. The actual number of bytes skipped is returned."](https://docs.oracle.com/javase/7/docs/api/java/io/FileInputStream.html#skip(long)).

Options to fix this are:
1) loop until the correct amount of bytes have been skip (this could be infinite?)
2) If the skip is not of the requested size, return `Err`
3) combine 1 and 2 by having an upper bound of allowed loops before returning `Err`. This loop could be some low constant but could also be a fraction of the requested bytes (a 1000th maybe?).

ninaandrup:
Closes #4006


main/src/library/File.flix
```diff
@@ -309,19 +309,29 @@ namespace File {
...
+                        let returnString = newString(bytes, charSet);
+                        Ok(returnString)
+                    } else {
+                        loop(i - 1)
+                    }
+                } else {
+                    Err("An error occurred when reading the file: '${f}'.")
+                }
+            };
+            loop(opts.count / 1000)
```
ninaandrup, line 329: I am not sure how many times we want to try to skip the correct number of bytes

main/src/library/File.flix
```diff
@@ -309,19 +309,29 @@ namespace File {
...
+                if (0 <= i) {
+                    let stream = newInputStream(javaFile);
+                    let skipped = skip(stream, opts.offset);
+                    if (skipped == opts.offset) {
+                        let bytes = [0i8; opts.count];
+                        discard read(stream, bytes, 0, opts.count);
+                        let charSet = forName(opts.charSet);
+                        let returnString = newString(bytes, charSet);
+                        Ok(returnString)
+                    } else {
```
JonathanStarup, line 322: In this case some `bytes < opts.offset` might have been skipped, which means that the next loop should only try to skip `opts.offset-bytes`. `loop` should probably have a `bytesToSkip` argument that can be decreased.

main/src/library/File.flix
JonathanStarup: I think it should use `opts.offset` instead of `opts.count` since offset is the argument to skip.

I have no idea what number to divide by but `opts.offset/1000` will be a retry for every kilobyte which is.. maybe a bit high? maybe 10.000 or 100.000.

main/src/library/File.flix
```diff
@@ -485,14 +495,23 @@ namespace File {
...
+            def loop(i) = {
+                if (0 <= i) {
+                    let stream = newInputStream(javaFile);
+                    let skipped = skip(stream, opts.offset);
+                    if (skipped == opts.offset) {
+                        let bytes = [0i8; opts.count];
+                        let numberRead = read(stream, bytes, 0, opts.count);
+                        let res = readBytesWithHelper(bytes, numberRead, opts.count);
+                        Ok(res)
+                    } else {
```
JonathanStarup, line 507: Given that this loop-skip is used more than once it should probably be its own `def skip(f, bytes): Result` kind of thing

JonathanStarup (changes requested):

ninaandrup:
I noticed the tests for `readWith` only checks that the result has the requested length. Should we make some tests for the offset?

JonathanStarup:
> I noticed [the tests for `readWith`](https://github.com/flix/flix/blob/0ad7b293d930a96518690cb96300ed24b7a906ab/main/test/ca/uwaterloo/flix/library/TestFile.flix#L172) only checks that the result has the requested length. Should we make some tests for the offset?

Sounds good, maybe just testing that the content looks as expected?

main/src/library/File.flix
```diff
@@ -299,29 +299,53 @@ namespace File {
...
+            }
+        } catch {
+            case ex: ##java.io.IOException =>
+                import java.lang.Throwable.getMessage(): String & Impure;
+                Err(getMessage(ex))
+        }
+
+    ///
+    /// Helper function for `readWith` and `readBytesWith`.
+    /// Tries to skip the requested number of bytes `bytes to skip` in `f` over `i` iterations.
```
JonathanStarup, line 330: i+1 iterations

main/test/ca/uwaterloo/flix/library/TestFile.flix
```diff
@@ -229,6 +229,13 @@ namespace TestFile {
             case _ => false
         }
 
+    @test
+    def testReadBytesWith03(): Bool & Impure =
+        match readBytesWith({offset = 10i64, count = 4}, "./LICENSE.md") {
```
JonathanStarup, line 234: There should also be a test with offset 0 and one with count 0 (maybe also some with negative numbers)

JonathanStarup (commented):
Reviewed from my phone so excusive the brief messages

main/src/library/File.flix
```diff
@@ -520,37 +554,41 @@ namespace File {
...
-                    bytes := [0i8; chunkSize];
-                    let accRead = deref numberRead;
-                    numberRead := read(stream, deref bytes);
-                    Ok(readBytesWithHelper(chunk, accRead, chunkSize))
-                } catch {
-                    case ex: ##java.io.IOException => {
-                        numberRead := -1;
-                        import java.lang.Throwable.getMessage(): String & Impure;
-                        Err(getMessage(ex))
+        if (0 < chunkSize) {
```
ninaandrup, line 558: We should force the user to not use chunks of size 0, right?

main/src/library/File.flix
```diff
@@ -299,29 +301,57 @@ namespace File {
...
             let stream = newInputStream(javaFile);
-            discard skip(stream, opts.offset);
-            let bytes = [0i8; opts.count];
-            discard read(stream, bytes, 0, opts.count);
-            let charSet = forName(opts.charSet);
-            let returnString = newString(bytes, charSet);
-            Ok(returnString)
+            match skip((opts.offset / 10_000i64), opts.offset, stream, f) {
+                case Ok(_)    => {
+                    if (0 <= opts.count) {
```
ninaandrup, line 316: Does it make sense to check for this before even starting the function? So checking as the first thing?

main/src/library/File.flix
```diff
@@ -481,18 +511,23 @@ namespace File {
...
-            discard skip(stream, opts.offset);
-            let bytes = [0i8; opts.count];
-            let numberRead = read(stream, bytes, 0, opts.count);
-
-            let res = readBytesWithHelper(bytes, numberRead, opts.count);
-
-            Ok(res)
+            match skip((opts.offset / 10_000i64), opts.offset, stream, f) {
+                case Ok(_)    => {
+                    if (0 <= opts.count) {
```
ninaandrup, line 520: Same as above, checking this as the first thing in the function

ninaandrup (commented):
Some more bugs

main/src/library/File.flix
JonathanStarup: I'd say so yes, then you can also return `""` for `count=0`

main/src/library/File.flix
JonathanStarup: I think an empty iterator should be returned

ninaandrup: What do you mean with empty? The return type still needs to be `Iterator[Result[Array[Int8, Static], String], Static]`. My solution was :
```
Iterator.singleton(Static, Err("The argument `chunkSize` has to be bigger than 0 when calling `File.readChunks`."))
```
but I'm not sure this is what you mean by empty.

JonathanStarup: `Iterator.new(r)` such that it is instantly empty instead of an error element since nothing went wrong there is just no information

main/test/ca/uwaterloo/flix/library/TestFile.flix
```diff
@@ -226,23 +243,66 @@ namespace TestFile {
...
+    @test
+    def testReadChunks03(): Bool & Impure =
+        match readChunks(0, "./LICENSE.md") {
+            case Iterator(done, _) => done()
+        }
+
+    @test
+    def testReadChunks04(): Bool & Impure =
+        match readChunks(-10, "./LICENSE.md") {
+            case Iterator(done, _) => done()
```
ninaandrup, line 304: is this the expected result for `chunkSize < 0`

JonathanStarup: I don't think so no, I only meant to return `Iterator.new(Static)` if `chunkSize == 0`, the error before was good for `chunksize < 0`

ninaandrup:
RFR

Issue #4236, opened 2022-07-12
Polish Set Effects


magnus-madsen:
- [x] Non-public op: The *explanation* should state: An effect operation must be public. 
- [x] Introduce a distinct error message for duplicate op. (If there is already one, its text should be updated).
- [ ] Update error message: "Unexpected effect parameter(s)". Also, it should not say tip in the explanation.
- [x] Undefined operation has extra quotes. 
- [x] Add auto-complete support for `do ...`.
- [x] Add Goto-def for operations.
- [x] The `Redundancy` phase rejects a stm exp `do op; exp2`.
- [x] Effect sets are formatted as `Print + Throw`, but should it not be `{Print, Throw}`?
- [x] Add `without` to VSCode.
- [x] Add `without` to auto-complete suggestions.
- [x] Allow sugar `without Throw,NonDet,Print`
- [x] ~Its very confusing that we use `&` for two things. Proposal (temporarily) rename `&` to `&&` and `|` to `||` for the set part.~
- [ ] Cleanup typer.
- [x] Use effects in monomorphization (?)
- [x] Explore if we can pretty print `ef & (not ConcreteEffect)` as `ef - ConcreteEffect`.
- [x] Ensure Redundancy phase runs for effect handlers
- [ ] Add a metric truckload of tests (both positive and negative)
- [ ] Check for unused effects

#### Bugs:
- [x] Effect.Set (`& { }`)
- [ ] Formulas are not always simplified: `String -> Unit \ (Throw + (e3713740 & Print)) & (~Print)`.
- [ ] Semantic highlighting is sometimes broken: `let _a = f(u -> do Throw.throw(u));`.

#### Future work:
- [ ] Add rename support for effect operations.
- [ ] Add rename support for effects.

magnus-madsen:
I suggest to do small targeted PRs for better reviews instead of everything at once :)

mlutze:
> Semantic highlighting is sometimes broken: `let _a = f(u -> do Throw.throw(u));`.

Unable to reproduce

mlutze:
Renaming Effs and Ops is a little tricky. We do the same namespacing trick as Classes, using a companion namespace, so changing an Eff name is tantamount to changing a namespace name. We don't currently support renaming classes or sigs because of this difficulty.

mlutze:
> Allow sugar `without Throw,NonDet,Print`

[...]

PR #4239, opened 2022-07-12
fix: check for set difference in BoolFormula

mlutze:
related to #4236


main/src/ca/uwaterloo/flix/language/phase/unification/BoolFormula.scala
```diff
@@ -155,6 +155,7 @@ object BoolFormula {
     case Type.Apply(Type.Cst(TypeConstructor.Complement, _), tpe1, _) => Neg(fromEffType(tpe1, m))
     case Type.Apply(Type.Apply(Type.Cst(TypeConstructor.Intersection, _), tpe1, _), tpe2, _) => Conj(fromEffType(tpe1, m), fromEffType(tpe2, m))
     case Type.Apply(Type.Apply(Type.Cst(TypeConstructor.Union, _), tpe1, _), tpe2, _) => Disj(fromEffType(tpe1, m), fromEffType(tpe2, m))
+    case Type.Apply(Type.Apply(Type.Cst(TypeConstructor.Difference, _), tpe1, _), tpe2, _) => Conj(fromEffType(tpe1, m), Neg(fromEffType(tpe2, m)))
```
magnus-madsen, line 158: Is the plan to get rid of this or?

magnus-madsen: By keeping it aren't we allowing sugar to leak into the types?

mlutze: Sure. I just didn't want a user's written signature to be unrecognizable after typing. But if we add the resugaring for displaying differences, it should be OK

Issue #4224, opened 2022-07-11
Polish `foreach`


magnus-madsen:
- [x] Add `foreach` as keyword snippet to `CompletionProvider`.
- [x] Add `foreach` as a keyword to https://github.com/flix/vscode-flix.
- [ ] Figure out what goes wrong with the syntactic highlight of the second foreach: ![image](https://user-images.githubusercontent.com/2287953/178248163-8b5100dd-5085-4b6a-9d99-e80ebc71df25.png)
- [x] Use `foreach` in all our mutable `ToString` instances. 
- [x] Ensure every collection has an `iterator` (e.g. `MutList` seems to lack one).
- [x] https://github.com/flix/flix/issues/4225
- [x] Refactor the de-sugaring to call `Iterator.foreach(f, Iterable.iterator(X))`.
- [x] Add `iterator` to prelude as an alias for `Foldable.iterator`.
- [x] Ditty for `enumerator`.
- [ ] Add custom compiler message for missing instance of `Iterable`.

jaschdoc:
- [ ] Add foreach as keyword snippet to CompletionProvider.
- [ ] Add foreach as a keyword to https://github.com/flix/vscode-flix.
- [ ] Figure out what goes wrong with the syntactic highlight of the second foreach: image
- [ ] Use foreach in all our mutable ToString instances.
- [ ] Ensure every collection has an iterator (e.g. MutList seems to lack one).
- [ ] https://github.com/flix/flix/issues/4225
- [ ] Refactor the de-sugaring to call Iterator.foreach(f, Iterable.iterator(X)).
- [ ] Add iterator to prelude as an alias for Foldable.iterator.
- [ ] Ditty for enumerator.
- [ ] Add custom compiler message for missing instance of Iterable.

jaschdoc:
Hmm not sure about your vscode thing. Mine works just fine.
![Screenshot from 2022-07-14 14-56-06](https://user-images.githubusercontent.com/60233376/178987617-6fe61614-450d-476f-912f-413badbe9aa1.png)

jaschdoc:
Do we want enumerator on maps to return an iterator of type `Iterator[(k, v, Int32), r]`?
Instead of `Iterator[((k, v), Int32), r]

magnus-madsen:
> Do we want enumerator on maps to return an iterator of type `Iterator[(k, v, Int32), r]`? Instead of `Iterator[((k, v), Int32), r]`

[...]

PR #4242, opened 2022-07-12
refactor: use `foreach`-loop in `ToString` instances

### LINKED ISSUE #4162: Add `MultiMap.iterator`

jaschdoc:
Related to #4224
Closes  #4162


jaschdoc:
RFR

jaschdoc:
I'm not quite sure why the compiler suite is failing. Something to do with the purity casts?

jaschdoc:
@mlutze do you have any idea what's up with this?

mlutze:
> @mlutze do you have any idea what's up with this?

Is the `ForEach` class in the core library?

jaschdoc:
> Is the `ForEach` class in the core library?

Yes, but we're using `Iterable` (which is in the library).

mlutze:
> > Is the `ForEach` class in the core library?
> 
> Yes, but we're using `Iterable` (which is in the library).

Ah, but StringBuilder is not in there.

mlutze:
> Do we want `StringBuilder` in the core library? Also, Iterable and Iterator were not in the core library but I've moved them there now (I thought std lib = core lib).

Yeah, probably fine.

jaschdoc:
I tried running tests locall with `StringBuilder`, `Iterable` and `Iterator` in the core lib and the tests still fail. How is the following test related to `ToString` instances?

```scala
  test("Test.IllegalEntryPointArg.Main.01") {
    val input =
      """
        |def main(blah: Array[String, _]): Unit & Impure = ??? as & Impure
        |""".stripMargin
    val result = compile(input, Options.TestWithLibMin)
    expectError[EntryPointError.IllegalEntryPointArgs](result)
  }
```

jaschdoc:
We're already using `StringBuilder` in the `MultiMap` `ToString` instance and that's on the main branch. Intuitively, adding `ToString` shouldn't affect the entry point tests. I've tried reverting the array instance so it doesn't use stringbuilder but just simple strings (and not including stringbuilder in the core lib) and it still doesn't pass the tests.

mlutze:
It looks like there are actually a bunch of transitive dependencies:

StringBuilder, Iterator, Newable, Iterable, Scoped, Option, Map, and probably a bunch more once the whole list is exhausted.

For the entrypoint tests, let's just change from `TestWithLibMin` to `TestWithLibAll`

It's likely that the other tests can be similarly fixed.

magnus-madsen:
I think there was some confusion about the goal. But first, reflect on the fact that we should never add casts to pure. Doing so is inherently unsafe. Everyone is really treating this too lightly :-) If I ask you to do that you should immediately challenge me :)

Now, what I actually wanted for this PR, which I explained poorly, was to use our new `foreach` loop inside our `ToString` instances-- inside those that are pure, but where we use a string builder (which should be all of them, AFAIK).

jaschdoc:
RFR

I've also added an iterator override to the foldable instances so we avoid converting the collection to an array when using the top level `iterator` and `enumerator` functions.

jaschdoc:
RFR

PR #4246, opened 2022-07-12
refactor: use regions in `SearchEnv`

### LINKED ISSUE #3896: Refactor Datalog solver to use regions

JonathanStarup:
Fixes #3896.

With this PR the debugging messages will now be erased by the optimizer. This is probably just the way it is for now.

The remaining impure methods are related to debug messages and the use of `Global.newId`.

The remaining unsafe Pure casts are from the above and then the unsafe implementation of `Tuple`

### DIFF
--- a/main/src/library/Fixpoint/Interpreter.flix
+++ b/main/src/library/Fixpoint/Interpreter.flix
@@ -76,20 +76,20 @@ namespace Fixpoint {
     }
 
     type alias Database[v: Type, r: Region] = MutMap[RamSym[v], MutMap[Tuple[v], v, r], r]
-    type alias SearchEnv[v] = (Array[Tuple[v], Static], Array[v, Static])
+    type alias SearchEnv[v: Type, r: Region] = (Array[Tuple[v], r], Array[v, r])
 
-    def interpret(r: Region[r], stmt: RamStmt[v]): Database[v, r] & Impure with Order[v], ToString[v] =
+    def interpret(r: Region[r], stmt: RamStmt[v]): Database[v, r] & r with Order[v], ToString[v] =
         interpretWithDatabase(new MutMap(r), stmt)
 
-    def interpretWithDatabase(db: Database[v, r], stmt: RamStmt[v]): Database[v, r] & Impure with Order[v], ToString[v] =
-        notifyPreInterpret(stmt);
+    def interpretWithDatabase(db: Database[v, r], stmt: RamStmt[v]): Database[v, r] & r with Order[v], ToString[v] =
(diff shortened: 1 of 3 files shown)


main/src/library/Fixpoint/Interpreter.flix
```diff
@@ -76,20 +76,20 @@ namespace Fixpoint {
...
+    type alias SearchEnv[v: Type, r: Region] = (Array[Tuple[v], r], Array[v, r])
 
-    def interpret(r: Region[r], stmt: RamStmt[v]): Database[v, r] & Impure with Order[v], ToString[v] =
+    def interpret(r: Region[r], stmt: RamStmt[v]): Database[v, r] & r with Order[v], ToString[v] =
         interpretWithDatabase(new MutMap(r), stmt)
 
-    def interpretWithDatabase(db: Database[v, r], stmt: RamStmt[v]): Database[v, r] & Impure with Order[v], ToString[v] =
-        notifyPreInterpret(stmt);
+    def interpretWithDatabase(db: Database[v, r], stmt: RamStmt[v]): Database[v, r] & r with Order[v], ToString[v] =
+        let _ = notifyPreInterpret(stmt) as & Pure;
```
magnus-madsen, line 85: If the cast is not on the call to notifyPre... but on the whole body of the function then it should not be removed, right?

magnus-madsen: But then of course that cast has to be to r.

JonathanStarup:
@magnus-madsen fixed

magnus-madsen:
I notice that you use `r` as an effect, but you should use `Read(r)` and/or `Write(r)` since thats the future.

JonathanStarup:
> I notice that you use `r` as an effect, but you should use `Read(r)` and/or `Write(r)` since thats the future.

I've changed it when regions are used but not for Impure. Is this consistent with other places, or should & be entirely removed?

PR #4249, opened 2022-07-12
feat: add `iterator` and `enumerator` to Prelude

jaschdoc:
Related to #4224


jaschdoc:
Hmm I'm not sure this is a good idea. This introduces a lot of dependencies into the core lib (i.e. all that `Iterator` depends on).

magnus-madsen:
Cheap trick: Instead of adding them to add `Prelude` add them to `Iterator` but outside the namespace; at the top.

jaschdoc:
RFR

magnus-madsen:
Thanks!

PR #4251, opened 2022-07-12
allow sets in without expressions

mlutze:
related to #4326


magnus-madsen:
Why retain the list versus desugaring it into multiple without expressions?

magnus-madsen:
I think desugaring into multiple without would be better since it keeps us closer to the calculus.

mlutze:
I was going with the "less sugar" philosophy :) But I don't have a strong reason either way.

magnus-madsen:
> I was going with the "less sugar" philosophy :) But I don't have a strong reason either way.

I think, in the interest of the calculus, we should desugar to multiple without expressions.

PR #4252, opened 2022-07-13
feat: use && for intersection types

mlutze:
related to #4236


magnus-madsen:
I know I asked for it, but it also feels like a step back :) Is there any other way we could make it easier to distinguish the effects?

magnus-madsen:
It is more the separation between the purity part and the effect part that is the problem, I think.

mlutze:
Vertical pipe `|` is unused. Could that work?

magnus-madsen:
Let's leave it.

PR #4254, opened 2022-07-13
fix: simplify constant sets in SVE

mlutze:
There was an issue with the application of SVE to effect sets:
`{F} - ({E, F} - {E})` should always equal the empty set
But its boolean formula counterpart `F and not ((E or F) and not E)` is not equal to False.

Per discussion, the argument to the base case of the SVE algorithm must be fully reduced: in a simple set form with no free variables.

In order to arrive at this form, I've introduced a simplification function. It represents an effect as a either a positive or negative set of effects and rigid variables, (e.g. `{a, b, c}` or `~{d, e, f}`. Then it converts it back to a regular for use in the SVE algorithm.

With this simplification in place, things _appear_ to work well, in that nothing has broken since introducing the simplification, but I am not sure it is sound. In particular, I am not sure that it is appropriate to treat the type variables as singleton effect constants, when in reality they represent sets of effects.


mlutze:
Updated with some explanation

PR #4269, opened 2022-07-14
feat: add `upcast` function

### LINKED ISSUE #4248: Add safe upcast functions?
What if we added:

```
def upcast(f: a -> b): a -> b & ef = x -> f(x) as & ef
```

and used them in places where effect casts are safe? 

- I assume the above signature could be generalized to work with both effect systems?
- I assume we could also have `upcast2`, `upcast3`, as needed.

jaschdoc:
Closes #4248


jaschdoc:
I'm not sure about what to do with `Nec.mapAccumLeft` and `...right` and `Traversable.mapAccumLeft` and `...right`. Nothing really seems to work. I especially don't understand why nothing else works on `Nec.map.Accum` as there's nothing impure (only effectful). Not even passing tuples to identity works. Been arguing with the compiler for a while hehe

jaschdoc:
```

-- Type Error -------------------------------------------------- Nec.flix

>> Unable to unify the types: 's' and 'Nec[b] -> (s, Nec[b]) & ef \ ef''.

693 |         loop(start, c, upcast(identity))
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
              mismatched types.

Type One: (s -> (Nec[b] -> (s, Nec[b]) & ef \ ef')) -> (s, Nec[b]) & ?pur126087 \ ?eff126088
Type Two: (?result126389 -> ?result126389 & b126387 \ e126388) -> ?result126328 & ?pur126329 \ ?eff126330
```

```

-- Type Error -------------------------------------------------- Nec.flix

>> Unable to unify the types: 's' and '(s, Nec[b])'.

683 |>         def loop(s1: s, c1: Nec[a], k: (s, Nec[b]) -> (s, Nec[b]) & ef) = match viewLeft(c1) {
684 |>             case OneLeft(x)      => {
685 |>                 let (s2, x1) = f(s1, x);
686 |>                 k((s2, NecOne(x1)))
687 |>             }
688 |>             case SomeLeft(x, rs) => {
689 |>                 let (s2, x1) = f(s1, x);
690 |>                 loop(s2, rs, match (s3, ks) -> k(s3, cons(x1, ks)))
691 |>             }
692 |>         };
693 |>         loop(start, c, upcast(identity))

mismatched types.

Type One: s -> (Nec[b] -> (s, Nec[b]) & ef \ ef')
Type Two: ((s, Nec[b])) -> (Nec[b] -> ?result124967 & ?pur124968 \ ?eff124969) & ?pur124971 \ ?eff124972
```

magnus-madsen:
> I'm pretty sure it's because it's a binary function. You need an `upcast2` or some such for binary functions.

Yes, @jaschdoc Try to add an upcast2 function.

magnus-madsen:
@jaschdoc How many `as` are left after this? Just approx?

magnus-madsen:
Thanks!

PR #4270, opened 2022-07-14
docs: remove old todo

jaschdoc:
The todo comment had been there since 2017. I think it's okay to call it `Modulo`.


jaschdoc:
True, but it isn't really used anywhere except for pretty printing so perhaps we should introduce `rem` and `mod`? We can rename this at least.

PR #4271, opened 2022-07-14
feat: add lsp support for type classes

### LINKED ISSUE #1525: LSP: Add Support for Traits and Instances

jaschdoc:
- [ ] Rename classes
- [ ] Type constraints (Semantic tokens for these or what is meant by this?)

Test program
```flix
def main(): Unit & Impure =
    Tester.p();
    println("${A.b(5)}")

pub class A[t] {
    pub def b(t: t): t with ToString[t] = t
}

instance A[Int32]
```
other file
```flix
namespace Tester {

    pub def p(): Unit & Impure =
        println("${A.b(15)}")

}
```

Closes #1525


mlutze:
Also please test this with namespaces. Source locations are tricky when you have e.g. `instance N.C[a]`

magnus-madsen:
I will break the original task down into smaller tasks.

PR #4274, opened 2022-07-15
fix: use DNF conversion for effect set unification

mlutze:


magnus-madsen:
I would like some Flix tests. In particular, tests where the declared type is smaller than the inferred type.

Also, presumably some tests with `without`.

magnus-madsen:
(But we can merge first if you prefer).

magnus-madsen:
Ill merged it, but lets get a battery of tests :)

PR #4275, opened 2022-07-15
feat: add `upcast` expression pt. 1

### LINKED ISSUE #4264: Add `upcast` expression
Add an expression `upcast exp` (after `Cast`). The `upcast` expression enables safe casts and should be used without concern.

The `upcast exp` works as follows:

- The `Typer` allows `exp` and `upcast exp` to have any types.
- Post type inference, in `Safety, we check that the upcast is sound.

The soundness rules include:

- Casting a pure expression to an effect polymorphic or impure expression.
- Casting a Java object to one of its super-types.

This can then be generalized for functions: 
- Casting a pure function to an effect polymorphic or impure function.
- Casting a function that returns a Java object to one that returns a subtype (and so forth).

I suggest to start with simple rules and then extend them gradually. 

The plan is to start with https://github.com/flix/flix/issues/4248 to see if that approach is feasible and to identify the locations where `upcast` is useful. This ticket is then about extending that functionality.

jaschdoc:
Related to #4264


magnus-madsen:
I would suggest a test-driven approach:

- Tests that upcasts from pure to effect polymorphic.
- Tests that upcasts from pure to impure.
- Tests that upcasts Java objects to their superclass.
- Tests that upcasts pure functions to effect polymorphic
- ...
- ...
- and soon until you have complicated functions :)

jaschdoc:
Yes that's exactly what I was thinking. Right now, though, I'm trying to hit the redundancy error case. Problem is that the type variables from the upcast are still variables and not a specific type after the Typer phase. Any thoughts on how to fix that?

jaschdoc:
@magnus-madsen There's a problem: after the the typer the `upcast` expression still only has type variables in its types so there's no concrete type. I asked @mlutze to take a look at it. Maybe he can explain what's going on :)

magnus-madsen:
Maybe the substitution is not applied.

mlutze:
The substitution is applied, but this doesn't always give use all the information we need because we don't do bidirectional type checking.

```scala
    @test
    def upcast02(): Unit = // Should trigger redundancy error.
        upcast ()
```

Here, the expression is inferred to have type `forall a . a`.

`Unit` is expected, and is the type of the inner expression, but the fact that `Unit` is expected is not registered during type inference, since it comes from the declared return type.

When we do `Scheme.checkLessThanEqual(declared, inferred)`, we get back a substitution. I imagine we can apply this substitution during type reconstruction and get the expected type that we want, but I'm not sure about the consequences of this, especially with respect to type hovering, etc. showing an overly-precise type.

jaschdoc:
Yes but there is still going to be a problem during the `Safety` phase as there is no way to know what upcast should cast to (the types are still variables at that point)

magnus-madsen:
> Yes but there is still going to be a problem during the `Safety` phase as there is no way to know what upcast should cast to (the types are still variables at that point)

What I mean is: If there are type variables in an upcast in the Safety phase then we flag the upcast as incorrect and reject the program. That is, we introduce a SafetyError: `IllegalUpcast` and return that instead.

magnus-madsen:
I think it might be helpful to start with simpler tests.

For example:

```
def test01(): Unit = 
  let f = x -> x + 1;
  let g = x -> {x + 1 _ as & Impure};
  let _ = if (true) upcast f ele g
  ()
```

jaschdoc:
> > Yes but there is still going to be a problem during the `Safety` phase as there is no way to know what upcast should cast to (the types are still variables at that point)
> 
> What I mean is: If there are type variables in an upcast in the Safety phase then we flag the upcast as incorrect and reject the program. That is, we introduce a SafetyError: `IllegalUpcast` and return that instead.

I've tried doing this, but the problem also arises when checking safe upcasts. When the types are still variables we cannot check if the cast is legal or *even what the type should be cast to*.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1299,6 +1296,14 @@ object Typer {
           resultEff = declaredEff.getOrElse(actualEff)
         } yield (constrs, resultTyp, resultPur, resultEff)
 
+      case KindedAst.Expression.Upcast(exp, _, _, _, loc) =>
```
magnus-madsen, line 1299: @jaschdoc 

Here is a small mistake:

You should already have a tvar, pvar, and evar from the Upcast. These are the ones you want to unify with. You don't want to create fresh ones here. 

Then you want to do:

```
(constrs, tpe, pur, eff) <- visitExp(exp)
_ <- unifyM(tvar, tpe)
_ <- unifyM(pvar, pur)
_ <- unifyM(evar, eff)
````

Comment (jaschdoc) [reply main/src/ca/uwaterloo/flix/language/phase/Typer.scala:1299]: Yes but now the upcast expression gets the same type as the inner `exp`. The way I understand it is that we want the upcast expression to have the expected types of the environment (read: the expression enclosing the upcast expression) and the inner `exp` to just have its inferred type. In other words, upcast can be whatever is necessary.

In the test below `upcast f` should have type `Int32 -> Int32 & Impure` whereas `f` should have `Int32 -> Int32` because `if (true) (upcast f) else g` has type `Int32 -> Int32 & Impure`.

```flix
    def testUpcast03(): Unit =
      let f = x -> x + 1;
      let g = x -> { x + 1 as & Impure };
      let _ = if (true) (upcast f) else g;
      ()
```

Am I overlooking something, is there any way in the typer to unify with something else than the inner expression of upcast?

magnus-madsen: Ah right- I was not thinking. In that case, drop the unification calls and just return the type variables (tvar, pvar, evar) etc.

jaschdoc:
Okay I've tried something bonkers (which I'm not finished with but maybe you can stop me from finishing this):
1. During reassembly we maintain a set of upcast expressions
2. After reassembly we traverse the AST one more time to "patch" all the upcast expressions to the parent types by checking for each node if its children is in the set of upcast expressions.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -2711,4 +2731,133 @@ object Typer {
     }
     t.write(new PrintWriter(System.out))
   }
+
+  private def patchUpcasts(exp0: TypedAst.Expression): TypedAst.Expression = {
```
magnus-madsen, line 2735: What is this supposed to do?

magnus-madsen:
> 

No, this is not going to work.

jaschdoc:
Okay good to know :)

magnus-madsen:
Our test case is this program:

```
    @test
    def testUpcast03(): Unit =
      let f = x -> x + 1;
      let g = x -> { x + 1 as & Impure };
      let _ = if (true) (upcast f) else g;
      ()
```

What happens when you try to type check and then safety check that?

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1299,6 +1296,14 @@ object Typer {
           resultEff = declaredEff.getOrElse(actualEff)
         } yield (constrs, resultTyp, resultPur, resultEff)
 
+      case KindedAst.Expression.Upcast(exp, _, _, _, loc) =>
+        for {
```
magnus-madsen, line 1300: I think this should be:

```
       for {
          (constrs, _tpe, _pur, _eff) <- visitExp(exp)
        } yield {
          (constrs, tvar, pvar, evar)
        }
```

where tvar, pvar, and evar come from the upcast.

jaschdoc:
```

-- Safety Error -------------------------------------------------- main/test/flix/Test.Exp.Upcast.flix

>> Unsafe upcast

18 |       let _ = if (true) (upcast f) else g;
                              ^^^^^^^^
                              the upcast occurs here.
```

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
magnus-madsen: Then in safety you have two things. The Upcast expression now knows the "expected" type, purity and effect of the "outer" whereas the "inner" expression has the "actual" type, purity, and effect.

jaschdoc: Alright, with this change it works!

jaschdoc: Hopefully, it also works when the safety error is fully implemented hehe.

magnus-madsen: OK, and then in safety you can extract "actualTpe, actualPur, actualEff" and "expectedTpe, ..." and compare the,

jaschdoc:
Okay so here's a twist. The program below should trigger an error
```
@test
def testUpcast02(): Unit = // Should trigger error
    upcast ()
```
because of this implementation in `Safety`.
```scala
    case Expression.Upcast(exp, tpe, pur, eff, loc) =>
      println(s"tpe    : $tpe")
      println(s"pur    : $pur")
      println(s"eff    : $eff")
      println(s"exp.tpe: ${exp.tpe}")
      println(s"exp.pur: ${exp.pur}")
      println(s"exp.eff: ${exp.eff}")
      val x = tpe match {
        // case _: Type.Var => List(UnsafeUpcast(loc))
        case _ => Nil
      }
      val ps = (pur, exp.pur) match {
        case (Type.Pure, _) =>
          // Nothing can be cast to pure
          List(UnsafeUpcast(loc))

        case (_, Type.Impure) =>
          // An Impure expression cannot be cast to anything else
          List(UnsafeUpcast(loc))

        case _ => Nil
      }
      visitExp(exp) ::: ps ::: x
```
However, the types when printed are:
```
tpe    : t88002
pur    : b88003
eff    : e88004
exp.tpe: Unit
exp.pur: true
exp.eff: Empty
```

I know that for this test it doesn't matter that it is cast to pure but there may be situations where the type variables "slip past" the safety checks.

This test, however, of course does not pass.
```flix
    @test
    def testUpcast02(): Unit & Impure =
        upcast (() as & Impure)
```

Of course this problem can easily be solved in the safety phase but I'm just not sure I'd trust it to **always** be sound.

magnus-madsen:
Suggestion btw: Lets merge this when it has the same functionality as the upcast function-- then in the next PR we can significantly expand what it permits.

magnus-madsen:
You can start by generating an error if any of the tpe, pur, and eff are variables.

jaschdoc:
Alright, but this is still broken for some reason:
```flix
    @test
    def testUpcast02(): Unit & Impure =
        upcast ()
```
:exploding_head:

magnus-madsen:
> Alright, but this is still broken for some reason:
> 
> ```
>     @test
>     def testUpcast02(): Unit & Impure =
>         upcast ()
> ```
> 
> types
> 
> ```
> tpe    : t59083
> pur    : b59084
> eff    : e59085
> exp.tpe: Unit
> exp.pur: true
> exp.eff: Empty
> ```
> 
> error
> 
> ```
> -- Safety Error -------------------------------------------------- main/test/flix/Test.Exp.Upcast.flix
> 
> >> Unsafe upcast
> 
> 12 |         upcast ()
>              ^^^^^^^^^
>              the upcast occurs here.
> ```

I think its okay for that to be "broken" because of the way return types are treated.

magnus-madsen:
Lets polish this up (possibly remove the tests), remove any debugging printlns, add comments etc, and then we can merge it as a stepping stone.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1546,6 +1551,9 @@ object Lowering {
       val e = substExp(exp, subst)
       Expression.Cast(e, declaredType, declaredPur, declaredEff, tpe, pur, eff, loc)
 
+    case Expression.Upcast(_, _, _, _, _) =>
```
magnus-madsen, line 1554: Move to the end of the match. We usually group the impossible cases there.

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -177,6 +177,36 @@ object Safety {
     case Expression.Cast(exp, _, _, _, _, _, _, _) =>
       visitExp(exp)
 
+    case Expression.Upcast(exp, tpe, pur, eff, loc) =>
```
magnus-madsen, line 180: Dramatically simplify this and remove the printlns so we can do a merge :)

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -206,6 +206,10 @@ object Simplifier {
         val e = visitExp(exp)
         SimplifiedAst.Expression.Cast(e, tpe, simplifyPurity(pur), loc)
 
+      case TypedAst.Expression.Upcast(exp, tpe, pur, _, loc) =>
+        val e = visitExp(exp)
+        SimplifiedAst.Expression.Cast(e, tpe, simplifyPurity(pur), loc)
```
magnus-madsen, line 211: I think here we should just translate to the underlying expression.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -222,9 +222,6 @@ object Typer {
       // Compute the stale and fresh definitions.
       val (staleDefs, freshDefs) = changeSet.partition(root.defs, oldRoot.defs)
 
-      // println(s"Stale = ${staleDefs.keySet}")
-      // println(s"Fresh = ${freshDefs.keySet.size}")
```
magnus-madsen, line 226: Keep these please, I am currently using them, sorry :)

main/test/flix/Test.Exp.Upcast.flix
```diff
@@ -0,0 +1,14 @@
+namespace Test/Exp/Upcast {
+
```
magnus-madsen, line 2: Ensure the tests pass.

magnus-madsen (commented):
I added some comments so we can get this merge. Once done, all the difficult work can happen in Safety while the rest of the phases require no changes.

jaschdoc:
RFR

PR #4288, opened 2022-07-16
test: add comprehensive set effect tests

mlutze:


mlutze:
@magnus-madsen Any other vectors I should be hitting or should I just keep expanding coverage along these lines?

magnus-madsen:
> @magnus-madsen Any other vectors I should be hitting or should I just keep expanding coverage along these lines?

Type schemes that are less general than their inferred type.

For example: 

```
def compose(f: a -> a \ ef, g: a -> a \ ef): a -> c \ ef = x -> f(g(x))
```

and things like that.

mlutze:
This isn't as comprehensive as would be ideal, but I will mark it as "ready" since it's better than sitting on it.

PR #4289, opened 2022-07-16
feat: add `JvmMethod` to the backend

paulbutcher:
I'll leave this draft, but I suspect it would make sense to review.

I'm particularly unsure whether I'm doing the right thing in `OccurrenceAnalyzer`

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/ErasedAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/ErasedAst.scala
@@ -266,7 +266,7 @@ object ErasedAst {
 
   case class Case(sym: Symbol.EnumSym, tag: Name.Tag, tpeDeprecated: MonoType, loc: SourceLocation)
 
-  case class JvmMethod(ident: Name.Ident, fparams: List[ErasedAst.FormalParam], exp: ErasedAst.Expression, retTpe: MonoType, loc: SourceLocation)
+  case class JvmMethod(ident: Name.Ident, fparams: List[ErasedAst.FormalParam], retTpe: MonoType, loc: SourceLocation)
 
   case class CatchRule(sym: Symbol.VarSym, clazz: java.lang.Class[_], exp: ErasedAst.Expression)
 
--- a/main/src/ca/uwaterloo/flix/language/ast/ErasedAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/ErasedAst.scala
@@ -175,7 +175,7 @@ object ErasedAst {
 
     case class PutStaticField(field: Field, exp: ErasedAst.Expression, tpe: MonoType, loc: SourceLocation) extends ErasedAst.Expression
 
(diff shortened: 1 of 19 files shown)


paulbutcher:
@magnus-madsen when you have a moment, can you sketch out what you have in mind for lambda lifting? The more I think about it, the less certain I am about whether we need to lambda lift methods at all, given that they'll never be called (or at least will never be directly called) by Flix code?

But it's very likely that I'm missing something.

paulbutcher:
I assume that we don't want to do anything special to support `this`? So, we expect the user to explicitly provide it as a first argument, i.e.:

```
  object ##java.lang.Comparable {
    def compareTo(this: ##java.lang.Comparable, o: ##java.lang.Object): Int32 =
      ...
  }
```

And not:

```
  object ##java.lang.Comparable {
    def compareTo(o: ##java.lang.Object): Int32 =
      ...
  }
```

Having said that, `this` is almost useless for our purposes (given that we're not providing any means to add fields to an object).

paulbutcher:
Do we want to allow one method within a `object` to call other methods within the same `object`?

magnus-madsen:
> I assume that we don't want to do anything special to support `this`? So, we expect the user to explicitly provide it as a first argument, i.e.:
> 
> ```
>   object ##java.lang.Comparable {
>     def compareTo(this: ##java.lang.Comparable, o: ##java.lang.Object): Int32 =
>       ...
>   }
> ```

Yes, let us start with `this` being the first argument.

magnus-madsen:
> @magnus-madsen when you have a moment, can you sketch out what you have in mind for lambda lifting? The more I think about it, the less certain I am about whether we need to lambda lift methods at all, given that they'll never be called (or at least will never be directly called) by Flix code?
> 
> But it's very likely that I'm missing something.

Suggestion: After Lambda Lifting: Retain `JvmMethod` but now is has no expression body. 

This should allow you to proceed to the backend and generate methods that throw exceptions. Once that works, we will discuss the details of lambda lifting. (Possibly we should talk about this over Zoom).

magnus-madsen:
Our current focus should be to generate correct JVM classes with correct superclasses and signatures (but with "empty", exception throwing methods).

paulbutcher:
I just stumbled across [`ClassMaker`](https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/jvm/ClassMaker.scala). Should I be using this within [`GenAnonymousClasses`](https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/jvm/GenAnonymousClasses.scala)?

paulbutcher:
I've just pushed changes which implement empty exception throwing methods.

magnus-madsen:
I think we can merge this as-is (unless you have more changes) and then start on the next iteration.

paulbutcher:
> I think we can merge this as-is (unless you have more changes) and then start on the next iteration.

Works for me (I've marked it as ready for review).

Issue #4281, opened 2022-07-16
Check set effects for purity reflection


mlutze:
Currently we only look at the binary effects.

magnus-madsen:
binary -> boolean

mlutze:
I thought boolean was ambiguous :D

PR #4300, opened 2022-07-18
fix: check for set effects in reifyEff

mlutze:
related to #4236
fixes https://github.com/flix/flix/issues/4281


magnus-madsen:
(This is a bit important so we don't break purity reflection)

PR #4306, opened 2022-07-19
feat: add `Graph.isCyclic`

ninaandrup:
Related to #3906


main/src/library/Graph.flix
```diff
@@ -89,16 +103,19 @@ namespace Graph {
...
+        let edges = inject g into Edge;
+        let reachability = #{
+            // Reachability given the edges.
+            Reachable(n1, n2) :- Edge(n1, n2), if n1 != n2.
+            // If `n1` can reach `m` and there is an edge from `m` to `n2`
+            // then `n1` can also reach `n2`. This adds all node pairs to
+            // the relational that are reachable using any number of nodes.
+            Reachable(n1, n2) :- Reachable(n1, m), Edge(m, n2).
+        };
+        let selfReach = query edges, reachability select (x, y) from Reachable(x, y) where x == y;
```
JonathanStarup, line 118: I think `()` can be selected here since the length of the result is the only thing used.

PR #4307, opened 2022-07-19
feat: add UnusedEffectSym error

mlutze:
related to #4236

Uses the same rule as enum: use in a type is not enough; it must be used in an expression.


magnus-madsen:
Lets do it :)

Issue #4310, opened 2022-07-19
Simplification of set effects


magnus-madsen:
This:

```
def twice(f: a -> a \ ef - Console, x: a): a \ ef - Console = 
    let r = f(f(x)) without { Console };
    do Console.echo();
    r
```

correctly, but confusingly, yields:

```
❌ -- Type Error -------------------------------------------------- file:///c%3A/tmp/flix/src/Main.flix

>> The type scheme: '∀(a, ef, e1703899). (a -> a \ ef - Console) -> (a -> a \ {({ef, Console} & {e1703899, ~Console}) - Console, Console})' cannot be generalized to '∀(a, ef). (a -> a \ ef - Console) -> (a -> a \ ef - Console)'.

30 | def twice(f: a -> a \ ef - Console, x: a): a \ ef - Console = 
         ^^^^^
         unable to generalize the type scheme.

The declared type does not match the inferred type:

  Declared: ∀(a, ef). (a -> a \ ef - Console) -> (a -> a \ ef - Console)
  Inferred: ∀(a, ef, e1703899). (a -> a \ ef - Console) -> (a -> a \ {({ef, Console} & {e1703899, ~Console}) - Console, Console})

```

Issue #4235, opened 2022-07-12
Refactor Boolean Infrastructure


magnus-madsen:
- [ ] Rename `BoolTable` to `BoolAlgTable` or BoolAlgebraTable?
- [x] Ensure `BoolTable` is completely independent of `Type`. (Maybe introduce `BoolMinimization` class?
- [ ] Rename `BoolFormula` to `BoolAlgebra`
- [ ] Redocument `BoolFormula` so it is agnostic about formula vs set expression.
- [ ] Rename `True/False` to Top/Bot or maybe Universe/Empty or ?

mlutze:
> Redocument `BoolFormula` so it is agnostic about formula vs set expression.

What terminology should I use?
- Latticelike (`join`, `meet`) or (`lub`, `glb`)
- Nontechnical (`both`, `either`)
- Something else?

magnus-madsen:
> 

What does the literature use?

magnus-madsen:
I suggest to use the lattice terms:

- Top, Bot, Join, Meet.

PR #4314, opened 2022-07-20
refactor: introduce TypeMinimization.scala

mlutze:
related to #4235 

sets up for #4310


main/src/ca/uwaterloo/flix/language/phase/unification/BoolTable.scala
```diff
@@ -56,12 +54,12 @@ object BoolTable {
   /**
     * The size a formula (but represented as a type) must have before we try to minimize it.
     */
-  private val Threshold: Int = 10
+  val Threshold: Int = 10
```
magnus-madsen, line 57: Maybe these should be moved? But I suppose they have dependencies so they cannot?

PR #4315, opened 2022-07-20
chore: tidy up some closure-related cruft

### LINKED ISSUE #4313: chore: closure tidyup
As per discussion with @magnus-madsen, remove unused `freeVars` member of `SimplifiedAst.Expression.Closure` and `tpe` argument to `getClosureClassType`.

paulbutcher:
Fixes #4313

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/SimplifiedAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/SimplifiedAst.scala
@@ -138,7 +138,7 @@ object SimplifiedAst {
     }
 
     @IntroducedBy(LambdaLift.getClass)
-    case class Closure(sym: Symbol.DefnSym, freeVars: List[FreeVar], tpe: Type, loc: SourceLocation) extends SimplifiedAst.Expression {
+    case class Closure(sym: Symbol.DefnSym, tpe: Type, loc: SourceLocation) extends SimplifiedAst.Expression {
       def purity: Purity = Pure
     }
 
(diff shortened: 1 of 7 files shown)


paulbutcher:
@magnus-madsen, this didn't allow me to get rid of `getFunctionTypeTemporaryToBeRemoved` as I hoped, as it is used elsewhere.

In fact, the more I look at it, the more I'm unsure why we think this can be removed? Don't we need something that encodes the argument and return types of the closure?

magnus-madsen:
> @magnus-madsen, this didn't allow me to get rid of `getFunctionTypeTemporaryToBeRemoved` as I hoped, as it is used elsewhere.
> 
> In fact, the more I look at it, the more I'm unsure why we think this can be removed? Don't we need something that encodes the argument and return types of the closure?

I don't recall, but the name suggests there is a 'structural' issue with that function.

paulbutcher:
So after some more code spelunking, I think we can get rid of it and replace it with something that encodes just the return type of the closure. The argument types are encoded within the `closureArgs` member of `Expression.Closure`?

paulbutcher:
But we couldn't get rid of it entirely.

magnus-madsen:
I think the concern with the function is indeed the duplication of information; there should be one source of ground truth. Thus your proposal makes sense, but lets do it in yet another PR :)

paulbutcher:
> lets do it in yet another PR

👍

paulbutcher:
Actually, this turns out to be much easier than I thought: there is already a `tpe` member in `Expression.Closure`. I have another PR ready to roll as soon as this one is merged.

magnus-madsen:
Ill merge this one and then I have to leave the PC :)

PR #4322, opened 2022-07-20
fix: actually pass along Xeffects flag

mlutze:


magnus-madsen:
should have added a TestMain test :P

mlutze:
> should have added a TestMain test :P

True, though IMO the `Default.copy(...)` pattern is dangerous so I'm only 97% to blame.

magnus-madsen:
> > should have added a TestMain test :P
> 
> True, though IMO the `Default.copy(...)` pattern is dangerous so I'm only 97% to blame.

Yeah, true, maybe we should not use it.

PR #4329, opened 2022-07-21
feat: object code generation

paulbutcher:
This is definitely not ready to be merged, but I've hit a problem that I think I need assistance with.

I'm halfway through and you're definitely seeing how the sausage gets made. The structure isn't right, the method is still throwing an exception, and I've not written any comments, so please don't worry too much about those things.

I've fixed a couple of problems that became apparent as I was working through this:

* The type of the closure needed to be set to an arrow type
* The `NewObject` case within `closuresOf` was returning `Set.empty` (blush).

Given the following Flix:

```
@test
def f():Int32 & Impure =
    import java.lang.Comparable.compareTo(##java.lang.Object): Int32 & Impure;
    let x = 0;
    let anon = object ##java.lang.Comparable {
        def compareTo(_this: ##java.lang.Comparable, _that: ##java.lang.Object) : Int32 = x + 1
    };
    compareTo(anon, () as ##java.lang.Object)
```

This crashes with `Exception in thread "main" java.util.NoSuchElementException` on this line:

https://github.com/paulbutcher/flix/blob/object-codegen/main/src/ca/uwaterloo/flix/language/phase/jvm/GenClosureClasses.scala#L97

I guess that I should be doing something to add the symbol to `root.defs`? But I'm not sure what?

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/ClosureConv.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/ClosureConv.scala
@@ -266,8 +266,9 @@ object ClosureConv {
     case Expression.NewObject(clazz, tpe, purity, methods0, loc) =>
       val methods = methods0 map {
         case JvmMethod(ident, fparams, exp, retTpe, purity, loc) =>
-          val closure = mkLambdaClosure(fparams, exp, retTpe, loc)
-          JvmMethod(ident, fparams, closure, retTpe, purity, loc)
+          val cloType = Type.mkPureUncurriedArrow(fparams.map(_.tpe), retTpe, loc)
+          val clo = mkLambdaClosure(fparams, exp, cloType, loc)
+          JvmMethod(ident, fparams, clo, retTpe, purity, loc)
       }
       Expression.NewObject(clazz, tpe, purity, methods, loc)
 
--- a/main/src/ca/uwaterloo/flix/language/phase/ClosureConv.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/ClosureConv.scala
@@ -266,7 +266,7 @@ object ClosureConv {
     case Expression.NewObject(clazz, tpe, purity, methods0, loc) =>
       val methods = methods0 map {
         case JvmMethod(ident, fparams, exp, retTpe, purity, loc) =>
-          val cloType = Type.mkPureUncurriedArrow(fparams.map(_.tpe), retTpe, loc)
+          val cloType = Type.mkImpureUncurriedArrow(fparams.map(_.tpe), retTpe, loc)
           val clo = mkLambdaClosure(fparams, exp, cloType, loc)
           JvmMethod(ident, fparams, clo, retTpe, purity, loc)
       }
--- a/main/src/ca/uwaterloo/flix/language/phase/jvm/GenAnonymousClasses.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/jvm/GenAnonymousClasses.scala
@@ -57,7 +57,7 @@ object GenAnonymousClasses {
 
     compileConstructor(superClass, visitor)
 
-    obj.methods.foreach(compileMethod(_, visitor))
+    obj.methods.zipWithIndex.foreach{ case (m, i) => compileMethod(m, i, visitor) }
 
     visitor.visitEnd()
     visitor.toByteArray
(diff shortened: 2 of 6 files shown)


magnus-madsen:
Do you visit the jvm method exp in lambda lift?

paulbutcher:
> Do you visit the jvm method exp in lambda lift?

Yes: https://github.com/paulbutcher/flix/blob/object-codegen/main/src/ca/uwaterloo/flix/language/phase/LambdaLift.scala#L377

paulbutcher:
> Some method, somewhere, is not visiting correctly.

OK, that's given me another thread to pull on: I'll check that `m: TopLevel` contains the newly generated closure after lambda lifting...

magnus-madsen:
I am on my phone so I am just guessing without looking at the code

paulbutcher:
That seems to have done the trick. Thank you!

paulbutcher:
This is now mostly there I think: enough that it's worth reviewing and merging.

Still to do (but I'll do the following in another PR):

* Lots (lots!) more tests.
* I'm pretty sure that this won't work properly for methods that take arguments of a type which occupy more than a single stack location.

But still, a very nice milestone to have reached.

paulbutcher:
Other things remaining (have I missed anything)?

* Relax requirements to support classes with zero-argument constructors.
* Replace the interim naming hack.

main/src/ca/uwaterloo/flix/language/phase/ClosureConv.scala
```diff
@@ -266,8 +266,9 @@ object ClosureConv {
     case Expression.NewObject(clazz, tpe, purity, methods0, loc) =>
       val methods = methods0 map {
         case JvmMethod(ident, fparams, exp, retTpe, purity, loc) =>
-          val closure = mkLambdaClosure(fparams, exp, retTpe, loc)
-          JvmMethod(ident, fparams, closure, retTpe, purity, loc)
+          val cloType = Type.mkPureUncurriedArrow(fparams.map(_.tpe), retTpe, loc)
```
magnus-madsen, line 269: I think it should be an Impure arrow type. (Otherwise the optimizer may wreck havoc).

main/src/ca/uwaterloo/flix/language/phase/jvm/GenAnonymousClasses.scala
```diff
@@ -66,12 +67,21 @@ object GenAnonymousClasses {
...
 
+    // Invoke the superclass constructor
     constructor.visitVarInsn(ALOAD, 0)
     constructor.visitMethodInsn(INVOKESPECIAL, superClass.toInternalName, JvmName.ConstructorMethod,
       MethodDescriptor.NothingToVoid.toDescriptor, false)
+
+    // For each method, compile the closure which implements the body of that method and store it in a field
+    methods.zipWithIndex.foreach { case (m, i) => 
+      constructor.visitVarInsn(ALOAD, 0)
+      GenExpression.compileExpression(m.clo, constructor, currentClass, Map(), new Label())
```
magnus-madsen, line 81: I am debating whether or not we should call into GenExpression or rather just copy the 10 or so lines from GenExpression that are relevant. I leaning towards the latter (code duplication, but at the benefit of ensuring there is no dependency between these two parts).

magnus-madsen:
I am OK with merging this except the Pure -> Impure function should be fixed (otherwise it may come back to bite us, maybe).

main/src/ca/uwaterloo/flix/language/phase/jvm/GenAnonymousClasses.scala
paulbutcher: Yes, I went through the same thought process. There are several points within `GenExpression` with similar code. There may be some value in bringing them together, but probably not as part of this PR.

paulbutcher:
> Replace the interim naming hack.

@magnus-madsen: I'd welcome your thoughts on what the right long-term solution is here?

magnus-madsen:
> 

A good solution, at least for a while, is to call `flix.genSym.freshId()` and just call the classes `Anon$17653` and so forth.

paulbutcher:
> A good solution, at least for a while, is to call flix.genSym.freshId() and just call the classes Anon$17653 and so forth.

I guess the right place to do this is in the namer, and then plumb it all the way through?

magnus-madsen:
> > A good solution, at least for a while, is to call flix.genSym.freshId() and just call the classes Anon$17653 and so forth.
> 
> I guess the right place to do this is in the namer, and then plumb it all the way through?

Actually, since the name is never referred to, we could probably just do it in GenAnonClass.

magnus-madsen:
If we want to be more fancy then we should probably introduce a symbol and do it in Namer, but I would start by doing it in GenAnonClass.

paulbutcher:
> Actually, since the name is never referred to, we could probably just do it in GenAnonClass.

Except it's also referred to in `GenExpression`:

https://github.com/paulbutcher/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala#L940

magnus-madsen:
Right, of course!

Lets start by generating a name `Anon$ID` of type String in namer and use that. I think this can be an isolated PR.

PR #4339, opened 2022-07-23
feat: add `NonDet` and `IO` effects

magnus-madsen:


magnus-madsen:
@JonathanStarup Heads-up, we will need to account for this in your scripts. I don't think it should be too difficult, just look for `NonDet` and `IO` and treat them as-if they were `Impure`.

Issue #4230, opened 2022-07-11
Allow multiple collections in `for-yield`


magnus-madsen:

stephentetley:
Could a `MonadZip` class help?

This was added to Haskell for monad comprehensions.

magnus-madsen:
> Could a `MonadZip` class help?
> 
> This was added to Haskell for monad comprehensions.

Do you have a link?

stephentetley:
I've implemented Haskell's `MonadZip` class in Flix - shall I make a PR to add it to the stdlib?

magnus-madsen:
> I've implemented Haskell's `MonadZip` class in Flix - shall I make a PR to add it to the stdlib?

Yes, please! I cannot guarantee we will use it, but I can take a look and try to understand it :)

magnus-madsen:
Paper looks interesting. Just need to find the time :)

stephentetley:
I've created a PR #4347 - feat: add a MonadZip class, potentially helping issue #4230

I'm looking at adding some more `Applicative` / `Monad` traversals to the stdlib so I've included a generic `zipWithA` function that isn't in the Haskell class.

magnus-madsen:
I am still intrigued by this, I just have not gotten around to it yet.

magnus-madsen:
I read both papers with great enjoyment. I think a lot of the stuff is a bit too advanced to add to Flix (at least for the moment), but I was happy to see that we are pretty close to where Haskell is at. (I did find one improvement we can make, reflected in https://github.com/flix/flix/issues/4429).

I think we can add `MonadZip` and syntax for it, e.g. (`for (x <- xs | y <- ys)`, but I am less convinced of whether it would be worth it. (Perhaps we can add the type class, but no comprehension syntax). In any case, none of these papers seems to suggest to me a way to get the flexibility that Scala has, but possible such a system is simply in-expressible in a language like Flix/Haskell. 

All this to say, I am not sure exactly what should be "next" for the for-yield/comprehension syntax. Possibly we just leave it where it is at the moment.

jaschdoc:
I think this can be done using `Iterable` along with a type class like below. `Materializable.materialize` would be wrapped around the entire expression in the weeder (along with a region so it remains pure). That way we're just manipulating / appending iterators which finally materialize to the type we expect.
This is partly inspired by Scala's way of handling it via inheritance and just defaulting to `Iterable` most of the time.

The downside is that we completely abandon the monad implementation.

```flix
[...]

PR #4347, opened 2022-07-24
feat: add a MonadZip class, potentially helping issue #4230

stephentetley:
This PR adds a MonadZip class. 

Design issues:

Superclass is `Monad` - although `MonadZip` class might help adding `for-yield` I don't think the superclass should be MonadZero. The superclass of Haskell's `MonadZip` is `Monad` and there are non-empty types that can have a valid `MonadZip` that can't have `MonadZero` (`Identity` , `Nel`, `Nec`).

I've added `zipWithA` which isn't in the Haskell class. Haskell does not have an overloaded `zipWithA` (or `zipWithM` as it's called in Haskell or historic reasons). Adding `zipWithA` potentially makes the class useful outside supporting comprehensions (if it is useful for de-sugaring for `for-yield`).

I've not followed Haskell's convention to prefix the members of the class with "m" - this is consistent with Flix's `MonadZero` class which defines `empty` rather than `mempty`.

I haven't added functions into the `Identity` and `Option` namespaces although I've written instances for them. I can do this - at the moment they seem perhaps obscure functions to have inside the modules.


magnus-madsen:
Thanks Stephen. 

I am merging this, but I will hold off on whether to add it to the `for-yield` syntax. I would like to understand better if it is commonly used.

PR #4348, opened 2022-07-24
feat: NewObject tests

paulbutcher:
I'm not done yet, but I think that there's enough here to be worth a review and merge.

This PR implements tests which confirm that `object` can:

* Implement methods taking and returning primitive types.
* Implement overloaded methods.
* Implement interfaces with default methods.
* Implement generic interfaces, and generic methods within an interface.
* Implement methods taking boxed primitive types (within the test for generics).
* Implement classes with a default constructor.

These are all "happy path" tests (i.e. they only test things that should work, not that things that shouldn't work fail correctly.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Safety.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Safety.scala
@@ -512,8 +512,10 @@ object Safety {
   }
 
   /**
-    * Given a class, get sets representing both the methods that can be implemented,
-    * and the methods that must be implemented.
+    * Given a class or interface, returns a pair of the methods that can be implemented and that must be implemented.
+    *
+    * * A (non-static) method can be implemented if it exists in the class.
+    * * A (non-static) method must be implemented if it is abstract.
     */
   private def getJavaMethods(clazz: java.lang.Class[_]): (Set[MethodSignature], Set[MethodSignature]) = {
     val methods = clazz.getMethods().toList.filterNot(m => java.lang.reflect.Modifier.isStatic(m.getModifiers()))
--- a/main/src/ca/uwaterloo/flix/language/phase/Safety.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Safety.scala
@@ -490,23 +490,6 @@ object Safety {
   case class MethodSignature(name: String, retTpe: Type, paramTypes: List[Type])
 
   /**
-    * Convert a list of Java methods to a set of MethodSignatures. Returns a map to allow subsequent reverse lookup.
-    */
-  private def getJavaMethodSignatures(methods: List[java.lang.reflect.Method]): Map[MethodSignature, java.lang.reflect.Method] = {
-    methods.foldLeft(Map.empty[MethodSignature, java.lang.reflect.Method]) {
-      case (acc, m) =>
-        if (!java.lang.reflect.Modifier.isStatic(m.getModifiers())) {
-          val signature = MethodSignature(m.getName(), 
-            getFlixType(m.getReturnType),
-            m.getParameterTypes().toList.map(getFlixType))
-          acc + (signature -> m)
-        } else {
-          acc
-        }
-    }
-  }
-
-  /**
     * Convert a list of Flix methods to a set of MethodSignatures. Returns a map to allow subsequent reverse lookup.
     */
   private def getFlixMethodSignatures(methods: List[JvmMethod]): Map[MethodSignature, JvmMethod] = {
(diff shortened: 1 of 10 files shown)


paulbutcher:
Remaining to be done:

* Functional Interfaces
* Exceptions
* Arrays
* Varargs

The first two I expect to fall out in the wash. Arrays and varargs, on the other hand, I think will require considerably more work as I can't see anything in Flix which handles native arrays yet? Forgetting `object` entirely, I don't think it's possible to call a Java method which takes a native array, is it?

Or am I missing something?

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -514,53 +501,66 @@ object Safety {
...
+    MethodSignature(method.getName(), 
+      getFlixType(method.getReturnType),
+      method.getParameterTypes().toList.map(getFlixType))
+  }
+
+  /**
+    * Given a class, get sets representing both the methods that can be implemented,
+    * and the methods that must be implemented.
+    */
+  private def getJavaMethods(clazz: java.lang.Class[_]): (Set[MethodSignature], Set[MethodSignature]) = {
```
magnus-madsen, line 518: Given a class or interface, returns a pair of the methods that *can* be implemented and that *must* be implemented. 

- A (non-static) method *can* be implemented if it exists in the class.
- A (non-static) method *must* be implemented if it is abstract.

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -514,53 +501,66 @@ object Safety {
...
+    val methods = clazz.getMethods().toList.filterNot(m => java.lang.reflect.Modifier.isStatic(m.getModifiers()))
+    val mustImplement = if (clazz.isInterface()) {
+        methods.filterNot(_.isDefault())
+      } else {
+        methods.filter(m => java.lang.reflect.Modifier.isAbstract(m.getModifiers()))
+      }
+    (methods.map(getJavaMethodSignature).toSet, mustImplement.map(getJavaMethodSignature).toSet)
```
magnus-madsen, line 525: I am bit concerned about the `toSet`. Could it be expensive and is it necessary?

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -514,53 +501,66 @@ object Safety {
...
-            Some(IllegalThisType(tpe, fparams.head.tpe, ident.name, methodLoc))
-          } else {
-            None
-          }
-      }
+    val thisErrors = methods.flatMap {
+      case JvmMethod(ident, fparams, _, _, _, _, methodLoc) => 
+        if (fparams.length < 1)
+          Some(MissingThis(tpe, ident.name, methodLoc))
+        else if (fparams.head.tpe != tpe) {
```
magnus-madsen, line 539: This check is not exactly correct, but I guess that will be discovered/refactored later.

main/test/flix/Test.Exp.Jvm.NewObject.flix
```diff
@@ -52,22 +42,87 @@ namespace Test/Exp/Jvm/NewObject {
...
-    compareTo(anon, anon as ##java.lang.Object) == 0
+  def testClassWithDefaultConstructor02(): Bool & Impure =
+    import flix.test.TestClassWithDefaultConstructor.abstractMethod(Int32): Int32;
+    import flix.test.TestClassWithDefaultConstructor.concreteMethod(String): String;
+    let anon = object ##flix.test.TestClassWithDefaultConstructor {
+      def abstractMethod(_this: ##flix.test.TestClassWithDefaultConstructor, x: Int32): Int32 = x + 1
+      def concreteMethod(_this: ##flix.test.TestClassWithDefaultConstructor, y: String): String = "flix: ${y}"
+    };
+    abstractMethod(anon, 1) == 2 and
+      concreteMethod(anon, "bar") == "flix: bar"
```
magnus-madsen, line 127: As a general comment, I worry whether these tests are too complicated.

See for example:

https://github.com/flix/flix/blob/master/main/test/flix/Test.Exp.Jvm.PutStaticField.flix

(Of course those tests can get away with being simpler, but it might still be nice to split some tests up into "smaller" tests). 

Also, I am worried about the use of `runTest`. Don't get me wrong, such helper functions are usually fine in normal test cases, but when writing a compiler, and a thing goes wrong, it can become extremely messy if helper functions are used.

That's why most other tests try to use as few constructs as humanly possible. 

For example, I would be more inclined towards e.g.:

```
public interface TestInt32Interface {
  int get();
  void set(int x);
}
```

and then a batch of test cases that use just this interface.

Repeat 7 (8?) x for all types.

I know this is very tedious to write-- (possibly we could ask a student programmer)-- but it is much better for when things go wrong. Otherwise what can happen is that **all** test cases fail when we make just a small breakage, for say Int32, or whatever.

magnus-madsen: We don't have to address this specific comment before I merge this PR, as work in progress.

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
paulbutcher: This may be where my rusty Scala knowledge is letting me down. Is there an equivalent of `Set.diff` that works on sequences?

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
paulbutcher: Can you give me a hint as to what's not correct? 🤣

main/test/flix/Test.Exp.Jvm.NewObject.flix
paulbutcher: I have no objection to splitting this into multiple smaller tests (understand the motivation for doing so, and no need to palm it off on a student!).

paulbutcher: The motivation behind `runTest` was that I wanted to test "from" Java. I worry that tests written in Flix could easily depend upon vagaries of Flix?

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
magnus-madsen: Ah okay. You need it for diff. Hmm. Maybe its okay then.

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
magnus-madsen: Yes, but its complicated. `==` only checks for structural equality, but there are types which are equivalent but not structurally equivalent. For example the function type `Int32 -> Int32 \ {r1, r2}` != `Int32 -> Int32 \ {r2, r1}`. But, I don't think this will start to hurt until much later :)

magnus-madsen: Actually, maybe this check should happen in the Typer?

main/test/flix/Test.Exp.Jvm.NewObject.flix
magnus-madsen: Yes, but you depend on strings, and on the `ToString` type class which means suddenly you touch a lot of the language inadvertently :)

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
paulbutcher: Well, I originally suggested doing this in the typer, but you suggested I shouldn't 🤣

Remember that the types here will only ever be Java types. So `Int32 -> Int32 \ {r1, r2}` (and similar) will never be relevant?

main/test/flix/Test.Exp.Jvm.NewObject.flix
paulbutcher: Right. But that's deliberate. It was only by doing this that I uncovered the bug that you fixed in #4342. We need to have *some* test that tests this kind of thing?

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
magnus-madsen: Possibly, not sure. But I think its fine for now

paulbutcher:
@magnus-madsen, when you have a moment, do you have any thoughts on the issue with native arrays?

magnus-madsen:
> @magnus-madsen, when you have a moment, do you have any thoughts on the issue with native arrays?

I know they are broken. There is a ticket somewhere. We may have to ask @JonathanStarup

main/test/flix/Test.Exp.Jvm.NewObject.flix
magnus-madsen: But could we not call each method one at a time and check that it returns 42 (if it is an int) and so forth?

paulbutcher: > But could we not call each method one at a time and check that it returns 42 (if it is an int) and so forth?

Yes, of course we could.

There are two reasons why I'm reticent to do so:

Firstly, calling Flix from Flix feels like cheating. What we're implementing here is something that's intended to support Java interop. Just because Flix can call Flix doesn't necessarily mean that Java can call Flix?

Secondly, avoiding the use of `ToString` or similar would mean that I wouldn't have found the bug that you fixed in #4342.

Nevertheless, if this is what you would prefer, I'll defer to your judgement.

paulbutcher: (unless you're talking about splitting it out into multiple separate tests, in which case I've already said that I agree?)

magnus-madsen: Oh- maybe we are talking past each other then.

paulbutcher: OK, so just to make sure that we're on the same page:

I'll split the current tests into multiple separate tests, but still retain the Java `runTest` method?

paulbutcher: (and still retain the use of `ToString` in the test case that tests multiple method parameters—I want to retain this because without it, there's no test to ensure that we're calculating parameter stack offsets correctly)

magnus-madsen: OK, I miss-read what runTest does. We can have something like that, but I would have one for each interface.

magnus-madsen: I would just like to have tests that are very incremental:

- Implement a method and check its return value.
- Implement a method and pass an argument to it.
- Implement a method where I use captured variables, type classes, etc.

paulbutcher: Gotcha 👍

I'll see about breaking things down like this tomorrow.

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
magnus-madsen: If you update this I can do a merge now 👍

magnus-madsen:
> > @magnus-madsen, when you have a moment, do you have any thoughts on the issue with native arrays?
> 
> I know they are broken. There is a ticket somewhere. We may have to ask @JonathanStarup

Partially broken that is. I think `Array[String]` and `Array[Int32]` should work, but I am less sure about `Array[Integer]` and multi-dimensional arrays.

JonathanStarup:
> > @magnus-madsen, when you have a moment, do you have any thoughts on the issue with native arrays?
> 
> I know they are broken. There is a ticket somewhere. We may have to ask @JonathanStarup

Arrays are essentially treated as generic lists in that they are either treated as `Array[Primitive]` or `Array[Object]` **in most places**. Things like `[[true], [false]]` should I believe have its precise type, but `[[true]; 2]` will not, so in general, I think it's best to assume the previous rule.

This is due to be fixed in the new backend (#2591), but it keeps being bumped down my priority list. There is a lot of typing left but hopefully not too much thinking.

PR #4355, opened 2022-07-25
feat: add `upcast` expression pt. 2

jaschdoc:
Related to #4264


jaschdoc:
It seems like the purity and effect set is still not inferred from the context. Rejecting any program with purity and effect variables would also reject a valid program like this. Any ideas?
```flix

    @test
    def testUpcast01(): Unit =
        let _ =
            if (true)
                upcast x -> x + 1
            else
                x -> { println("x"); x + 1 };
        ()
```
with error
```
-- Safety Error -------------------------------------------------- main/test/flix/Test.Exp.Upcast.flix

>> The following upcast is unsafe and not allowed.

7 |                 upcast x -> x + 1
                    ^^^^^^^^^^^^^^^^^
                    the upcast occurs here.

Actual type  : Int32 -> Int32
Actual purity: true
Actual effect: Empty

Expected type  : Int32 -> Int32 & Impure
Expected purity: b86490
Expected effect: e86491
```

magnus-madsen:
Ah yes. I think we should just ignore the pvar and evar for now and focus just on the type.

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -181,7 +181,13 @@ object Safety {
       visitExp(exp)
 
     case Expression.Upcast(exp, tpe, pur, eff, loc) =>
-      throw InternalCompilerException("Not implemented")
+      val es = (tpe, pur, eff) match {
+        case (_: Type.Var, _, _) => List(UnsafeUpcast(exp.tpe, exp.pur, exp.eff, tpe, pur, eff, loc))
+        case (_, _: Type.Var, _) => List(UnsafeUpcast(exp.tpe, exp.pur, exp.eff, tpe, pur, eff, loc))
```
magnus-madsen, line 186: So allow this case.

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -181,7 +181,13 @@ object Safety {
...
+        case (_, _, _: Type.Var) => List(UnsafeUpcast(exp.tpe, exp.pur, exp.eff, tpe, pur, eff, loc))
```
magnus-madsen, line 187: And this case.

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -180,8 +180,67 @@ object Safety {
     case Expression.Cast(exp, _, _, _, _, _, _, _) =>
       visitExp(exp)
 
-    case Expression.Upcast(exp, tpe, pur, eff, loc) =>
-      throw InternalCompilerException("Not implemented")
+    case e@Expression.Upcast(exp, tpe, pur, eff, loc) =>
```
magnus-madsen, line 183: This looks like a lot of fun :) :)

But try write a recursive function on that matches on `expected` and `actual` simultaneously. 

You will find this a lot easier, I think.

magnus-madsen: Like this:

```
def isSafeUpcast(expected: Type, actual: Type): Bool = (expected, actual) match {
  case (Type.Cst(...), Type.Cst(...)) => // do checks
  case (Type.Apply(...), Type.Apply(...) => // recurse on the parts
/// and so forth
```

magnus-madsen: So you line them up one at a time

jaschdoc:
Am I on the right track with this?

magnus-madsen:
> Am I on the right track with this?

Not really no 😆 Maybe we can talk tomorrow :)

jaschdoc:
Ok I had some help from @mlutze so this should be good for the purity checks.

mlutze:
$e_1 < e_2 \iff \exists e_3 . (e_1 \wedge e_3) \equiv_B e_2$

Is that true and should we use it?

jaschdoc:
RFR

PR #4362, opened 2022-07-25
test: additional `NewObject` tests

paulbutcher:
This concludes, I believe, all the "happy path" tests for NewObject (shout if you think I've missed any)

### DIFF
--- a/main/src/flix/test/TestFunctionalInterface.java
+++ b/main/src/flix/test/TestFunctionalInterface.java
@@ -1,0 +1,10 @@
+package flix.test;
+
+@FunctionalInterface
+public interface TestFunctionalInterface {
+  int testMethod(int x);
+
+  static boolean runTest(TestFunctionalInterface obj) {
+    return obj.testMethod(1) == 2;
+  }
+}
--- a/main/src/flix/test/TestThrowingInterface.java
+++ b/main/src/flix/test/TestThrowingInterface.java
@@ -1,0 +1,10 @@
+package flix.test;
+
+public interface TestThrowingInterface {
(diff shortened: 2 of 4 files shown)


magnus-madsen:
> This concludes, I believe, all the "happy path" tests for NewObject (shout if you think I've missed any)

I think this covers everything I can think of.

magnus-madsen:
The next steps would be abstract classes?

magnus-madsen:
Oh- what happened to interfaces with generics?

paulbutcher:
> Oh- what happened to interfaces with generics?

https://github.com/paulbutcher/flix/blob/d588b4862350541f7ec6cc1dc9072f8b91cbaf29/main/test/flix/Test.Exp.Jvm.NewObject.flix#L219-L237

paulbutcher:
> The next steps would be abstract classes?

https://github.com/paulbutcher/flix/blob/d588b4862350541f7ec6cc1dc9072f8b91cbaf29/main/test/flix/Test.Exp.Jvm.NewObject.flix#L274-L284

magnus-madsen:
Goodie. Do we have any tests with generics and *primitive values*. I guess they would negative tests because Java does not allow that?

paulbutcher:
> Goodie. Do we have any tests with generics and _primitive values_. I guess they would negative tests because Java does not allow that?

I'm not sure how I would write such a test (given that it wouldn't compile)?

magnus-madsen:
> > Goodie. Do we have any tests with generics and _primitive values_. I guess they would negative tests because Java does not allow that?
> 
> I'm not sure how I would write such a test (given that it wouldn't compile)?

Indeed.

PR #4369, opened 2022-07-26
feat: add support for top-level imports

magnus-madsen:
NB: This desugars to type aliases. Later we will want to avoid desugarings.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/ParsedAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/ParsedAst.scala
@@ -32,11 +32,11 @@ object ParsedAst {
     * A collection of imports and declarations.
     *
     * @param sp1   the position of the first character in the source.
-    * @param uses  the uses in the abstract syntax tree.
+    * @param usesOrImports  the uses in the abstract syntax tree.
     * @param decls the declarations in the abstract syntax tree.
     * @param sp2   the position of the last character in the source.
     */
-  case class CompilationUnit(sp1: SourcePosition, uses: Seq[ParsedAst.Use], decls: Seq[ParsedAst.Declaration], sp2: SourcePosition)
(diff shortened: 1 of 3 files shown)


paulbutcher:
Is there a reason why we require a semicolon at the end of a top-level import? This seems out of step with other top-level constructs?

paulbutcher:
> Is there a reason why we require a semicolon at the end of a top-level import? This seems out of step with other top-level constructs?

Hmm: I see that we require a semicolon after a `use` statement too.

But not after (e.g.) a `type alias`.

This seems inconsistent to me? Or am I missing something?

magnus-madsen:
> > Is there a reason why we require a semicolon at the end of a top-level import? This seems out of step with other top-level constructs?
> 
> Hmm: I see that we require a semicolon after a `use` statement too.
> 
> But not after (e.g.) a `type alias`.
> 
> This seems inconsistent to me? Or am I missing something?

Its inconsistent. We just discussed in the office. 

One day we plan to have semi-colon inference, but because it is so brittle (and dangerous) it makes sense to add it as late as possible.

I am not sure what to do about the inconsistency today. Possibly we could drop it for top-level imports/uses.

paulbutcher:
I agree about semicolon inference. In fact, I personally would recommend against even trying.

But this is a different question: this is whether the grammar requires a semicolon between (or more accurately, at the end of) top-level constructs. And for some (`use`, `import`) we do. For others (`type alias`, `enum`, ...) we don't. It strikes me that we should decide whether top-level constructs need a semicolon or not?

magnus-madsen:
> I agree about semicolon inference. In fact, I personally would recommend against even trying.
> 
> But this is a different question: this is whether the grammar requires a semicolon between (or more accurately, at the end of) top-level constructs. And for some (`use`, `import`) we do. For others (`type alias`, `enum`, ...) we don't. It strikes me that we should decide whether top-level constructs need a semicolon or not?

We should and I think that semi-colons should not be required at the top-level.

PR #4387, opened 2022-07-28
chore: Use imports in NewObject tests

### LINKED ISSUE #4365: Refactor `NewObject` tests to use imports/type aliases

paulbutcher:
Fixes #4365


main/test/flix/Test.Exp.Jvm.NewObject.flix
```diff
@@ -1,285 +1,313 @@
...
     runTest(anon)
 
   @test
   def testInt32ArrayInterface01(): Bool & Impure =
-    import static flix.test.TestInt32ArrayInterface.runTest(##flix.test.TestInt32ArrayInterface): Bool;
-    let anon = object ##flix.test.TestInt32ArrayInterface {
-      def testMethod(_this: ##flix.test.TestInt32ArrayInterface, xs: Array[Int32, false]): Array[Int32, false] = 
+    import static flix.test.TestInt32ArrayInterface.runTest(TestInt32ArrayInterface): Bool;
+    let anon = object TestInt32ArrayInterface {
+      def testMethod(_this: TestInt32ArrayInterface, xs: Array[Int32, false]): Array[Int32, false] =
```
magnus-madsen, line 180: I just saw this now, but the `false` should really be `Static`.

paulbutcher: I will make the change. But...

If these types are wrong, then the error message currently output with `--explain` currently suggests `Array[Int32, false`):

```
-- Safety Error -------------------------------------------------- main/test/flix/Test.Exp.Jvm.NewObject.flix

>> No implementation found for method 'testMethod' of superclass 'TestInt32ArrayInterface'.

148 |>     let anon = object ##flix.test.TestInt32ArrayInterface {
149 |>       def testMethod(_this: ##flix.test.TestInt32ArrayInterface, xs: Array[Int32, Static]): Array[Int32, _] = 
150 |>         xs |> Array.map(x -> x + 1)
151 |>     };

the object occurs here.

 Try adding a method with the following signature:

 def testMethod(Array[Int32, false]): Array[Int32, false]
```

This is because the string generated by `FormatType.formatWellKindedType` says `false` not `Static`.

Is this a bug? Or..?

magnus-madsen: That's a bug, or rather its because we are in a longer process. In `Prelude.flix` you can see that `Static = False`, but we should use Static everywhere so we can change the definition of it later.

paulbutcher:
Eugh - not sure what's happened here, but the merge with master has got completely confused.

paulbutcher:
I'm going to kill this PR and recreate it, which is probably easier than trying to work out what I need to to do address the merge issue.

Issue #4377, opened 2022-07-27
Refactor `& Impure` to `\ IO` in Standard Library and Tests


magnus-madsen:
See https://github.com/flix/flix/issues/4346

magnus-madsen:
@ninaandrup @jaschdoc This is a big refactoring I hope you can both help with :) I suggest to work by namespaces. Most likely a search and replace can be useful with minor manual adjustments.

ninaandrup:
I thought the compiler had already been refactored to support the new syntax - I should probably have read the other ticket a bit more thoroughly. I'll just leave the PR for when it becomes relevant :-)

magnus-madsen:
> I thought the compiler had already been refactored to support the new syntax - I should probably have read the other ticket a bit more thoroughly. I'll just leave the PR for when it becomes relevant :-)

I think it should work-- or-- @mlutze

mlutze:
It _should_ work

magnus-madsen:
This is mostly refactored now. What we need to do is update the documentation at flix.dev and doc.flix.dev.

@ninaandrup Can you help with that? :)

ninaandrup:
> This is mostly refactored now. What we need to do is update the documentation at flix.dev and doc.flix.dev.
> 
> @ninaandrup Can you help with that? :)

Is it a new repository for doc.flix.dev instead of [this repository](https://github.com/flix/programming-flix) after updating to the mdbook ?

magnus-madsen:
Yes, its flix/book

PR #4391, opened 2022-07-28
refactor: `& Impure` to `\ IO` in namespaces `Prelude`

ninaandrup:
Related to #4377


magnus-madsen:
@mlutze But some tests seem to fail?

mlutze:
`println` does not have the `IO` effect, only `Impure`.

We have `IO => Impure` but we don't have `Impure => IO`.

magnus-madsen:
> `println` does not have the `IO` effect, only `Impure`.
> 
> We have `IO => Impure` but we don't have `Impure => IO`.

To use other words, you mean that Nina should change println to have IO effect somehwere?

mlutze:
> > `println` does not have the `IO` effect, only `Impure`.
> > We have `IO => Impure` but we don't have `Impure => IO`.
> 
> To use other words, you mean that Nina should change println to have IO effect somehwere?

Yes

ninaandrup:
> `println` does not have the `IO` effect, only `Impure`.
> 
> We have `IO => Impure` but we don't have `Impure => IO`.

Makes sense :-)

main/src/library/Benchmark.flix
```diff
@@ -20,7 +20,7 @@ namespace Benchmark {
     /// impure since timing is considered a side-effect in this context.
```
ninaandrup, line 20: Should this like be `IO since timing is considered a side-effect in this context.` now ?

ninaandrup:
Can we remember to later change this [line in `TestRedBlackTree`](https://github.com/flix/flix/blob/ee597c91a1a95dd4f0404edff7af42ae8429a921/main/test/ca/uwaterloo/flix/library/TestRedBlackTree.flix#L251) when `MutList(Static)` gets `IO` effect instead of `Impure` ? Or should/can I do something to change it in this PR?

ninaandrup:
And should [this line not be deleted](https://github.com/flix/flix/blob/ee597c91a1a95dd4f0404edff7af42ae8429a921/main/src/library/RedBlackTree.flix#L568), since `f` has been changed to have effect `ef2` ?

main/src/library/Benchmark.flix
magnus-madsen: Yes

magnus-madsen:
Maybe we should do prelude and some other name spaces first?

ninaandrup:
I have run into a problem when chainging the `Prelude` namespace from `& Impure` to `\ IO`. Most of the examples have a `main` function with a `println`. But when I change `main` to
```
def main(): Unit \ IO =
    println(area(Rectangle(2, 4)))
```
in `algebraic-data-types-and-pattern-matching.flix` I get the error:
```
#
#
#   Unable to unify: 't6994 -> Unit & Impure \ IO' and 'Unit -> Unit & Impure'.
#
# This is a bug in the Flix compiler. Please report it here:
...
```

magnus-madsen:
Hmm. Matt? Is this because we have not yet made it default?

magnus-madsen:
@mlutze

mlutze:
It has to do with the `printUnlessPure` function and its assumed type in `EntryPoint`.

Since I can't really fix that without fixing the changes in this PR, I just added a commit to Nina's branch.

magnus-madsen:
> It has to do with the `printUnlessPure` function and its assumed type in `EntryPoint`.
> 
> Since I can't really fix that without fixing the changes in this PR, I just added a commit to Nina's branch.

I am bit unsure of the steps needed to proceed.

mlutze:
> > It has to do with the `printUnlessPure` function and its assumed type in `EntryPoint`.
> > Since I can't really fix that without fixing the changes in this PR, I just added a commit to Nina's branch.
> 
> I am bit unsure of the steps needed to proceed.

Nina should be fine to proceed as normal. I made a change to `EntryPoint.scala` which should unblock progress.

magnus-madsen:
By changing the examples to use `\ IO` instead of `& Impure`, I presume.

mlutze:
> By changing the examples to use `\ IO` instead of `& Impure`, I presume.

Yep

magnus-madsen:
@mlutze Can I go ahead with this?

mlutze:
Yes, we should be good to go. But we must be aware that once we pull the `--Xeffects` trigger, we'll break almost all client code. Community build passes because `--Xeffects` is off by default.

magnus-madsen:
@ninaandrup You can proceed :)

PR #4397, opened 2022-07-29
feat: Allow more characters within user operators

paulbutcher:
As discussed on Gitter.

I've added `/`, `:`, and `.` to the characters allowed in user operators.

Created as draft because I had to make one, very minor, change to `ReservedName.Law.01` in the Weeder tests.

The problem is that with this change, this

```
class C[a] {
    law ^^^: forall (x: a) . true
}
```

Now parses `^^^:` as one token (i.e. a law with the name `^^^:` rather than a law with the name `^^^` followed by a colon. So the error generated is a parse error, not a `ReservedName` error.

The fix is to change the above to:

```
class C[a] {
    law ^^^ : forall (x: a) . true
}
```

(note the space before the colon).

This is only an issue with laws because other definitions (e.g. `def`) don't allow a colon immediately after the symbol being defined.

I think that this is OK, but we should probably think about it, hence marking this as draft.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser.scala
@@ -102,7 +102,7 @@ object Parser {
     /**
       * An operator letter.
       */
-    val OperatorLetter: CharPredicate = CharPredicate("+-*<>=!&|^$")
+    val OperatorLetter: CharPredicate = CharPredicate("+-*<>=!&|^$:/.")
 
     /**
       * a (upper/lower case) letter, numeral, greek letter, or other legal character.
--- a/main/test/ca/uwaterloo/flix/language/phase/TestWeeder.scala
+++ b/main/test/ca/uwaterloo/flix/language/phase/TestWeeder.scala
@@ -596,7 +596,7 @@ class TestWeeder extends FunSuite with TestUtils {
     val input =
       """
         |class C[a] {
-        |    law ^^^: forall (x: a) . true
+        |    law ^^^ : forall (x: a) . true
         |}
         |""".stripMargin
     val result = compile(input, Options.TestWithLibNix)
--- a/main/test/flix/Test.Exp.Jvm.NewObject.flix
+++ b/main/test/flix/Test.Exp.Jvm.NewObject.flix
@@ -1,3 +1,33 @@
+import java.lang.Object;
+import flix.test.TestBoolArrayInterface;
+import flix.test.TestBoolInterface;
+import flix.test.TestCharArrayInterface;
+import flix.test.TestCharInterface;
+import flix.test.TestClass;
+import flix.test.TestClassWithDefaultConstructor;
+import flix.test.TestClassWithNonDefaultConstructor;
+import flix.test.TestDefaultMethods;
+import flix.test.TestFloat32ArrayInterface;
+import flix.test.TestFloat32Interface;
+import flix.test.TestFloat64ArrayInterface;
+import flix.test.TestFloat64Interface;
+import flix.test.TestFunctionalInterface;
+import flix.test.TestGenericInterface;
+import flix.test.TestGenericMethod;
+import flix.test.TestInt8ArrayInterface;
+import flix.test.TestInt16ArrayInterface;
+import flix.test.TestInt16Interface;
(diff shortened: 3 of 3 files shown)


magnus-madsen:
I will need some time to think about this. I am worried about future consequences of allowing `:` at the end of an identifier (because of its rule in so many places`. What happens to this: `let x:Int32`? What about `let x::Int32`?

Anyway, let me stew on it :)

paulbutcher:
I understand your reticence (hence why this is draft).

But...

> What happens to this: let x:Int32? What about let x::Int32?

This is fine, because the rules for names don't allow a mix of operator characters and non-operator characters. So `x:` isn't a valid identifier.

magnus-madsen:
What about `let +:Int32=...`? (It seems we don't allow `let +` but surely that is a mistake since we allow `def +(...)` and that would be irregular).

paulbutcher:
> What about `let +:Int32=...`? (It seems we don't allow `let +` but surely that is a mistake since we allow `def +(...)` and that would be irregular).

Yes, that would be a problem.

I can think of two possible approaches (other than just not allowing this):

* In this case, the user has to type `let + : Int32`
* Forbid `:` at the end of an identifier

These kinds of ambiguities are inevitable, and already exist in the Flix grammar, for example:

```
flix> 1+1
2                                                                               
flix> 1 + -1
0                                                                               
flix> 1+-1
-- Resolution Error -------------------------------------------------- <shell>  

>> Undefined name '+-'.

2 | println(1+-1)
             ^^
             name not found

Tip: Possible typo or non-existent definition?
```

So we should think carefully about them, but they'll never be completely eliminable.

Unless you fancy turning Flix into a Lisp? 😜

paulbutcher:
Other characters that aren't allowed within operators which might be worth considering: `~` & `%`, neither of which are used within Flix currently. Also `?` which is only used for holes?

paulbutcher:
@magnus-madsen, have you had a chance to give this any thought?

magnus-madsen:
I will give you some more detailed thoughts later, but I am afraid of symbolic operators turning into quicksand. Both for the language (e.g. symbol soup) and for parsing (if in the future we should switch parsing technology). I am especially worried about allowing `:`, even though I know we use it for Cons.

paulbutcher:
> I will give you some more detailed thoughts later, but I am afraid of symbolic operators turning into quicksand. Both for the language (e.g. symbol soup) and for parsing (if in the future we should switch parsing technology). I am especially worried about allowing `:`, even though I know we use it for Cons.

@magnus-madsen have you had a chance to come up with those more detailed thoughts?

magnus-madsen:
> > I will give you some more detailed thoughts later, but I am afraid of symbolic operators turning into quicksand. Both for the language (e.g. symbol soup) and for parsing (if in the future we should switch parsing technology). I am especially worried about allowing `:`, even though I know we use it for Cons.
> 
> @magnus-madsen have you had a chance to come up with those more detailed thoughts?

I did not work anything out in detail, but essentially I think that if we allow characters that are used as "punctuation" then this forces us to require white space around operators. For example, `a+b` might not be the same as `a + b`. I am very worried about such consequences. For example, I don't want a situation where `x \newline +y` is different from say `x+y` etc. 

"Most" other programming languages also disallow their punctuation and keywords to be used as identifiers, right?

paulbutcher:
> I did not work anything out in detail, but essentially I think that if we allow characters that are used as "punctuation" then this forces us to require white space around operators. For example, a+b might not be the same as a + b. I am very worried about such consequences. For example, I don't want a situation where x \newline +y is different from say x+y etc.

I 100% agree that that would be undesirable. And I can see how this might be an issue if we allowed user-defined operators which comprised a mix of operator letters and latin letters. But Flix doesn't: user-defined operators have to be at least 2 characters long, and can only consist of operator letters.

We already have a number of operators which share characters with "punctuation" in the language. For example: `::`, `:::`, `:=`, `..`, and `/`, plus all the operators which use `=`. Why don't these cause the problems you're concerned about?

I understand why we want to be careful here, I'm definitely not suggesting any kind of Scala-style "free for all".

magnus-madsen:
I will open a ticket to discuss symbolic operators.

Meanwhile I hope its OK I close this PR (which is outdated anyway, I guess) to clear the queue, so to speak.

PR #4398, opened 2022-07-29
fix: correct source position for import statements

### LINKED ISSUE #4363: Incorrect source location for import statement
Within the following source code, the source location associated with `import` statements is incorrect. For example, the import for `join` has the location: `Range(Position(15,4),Position(33,16))`.

This leads to confusing highlighting within VSCode.

```
type alias Request = ##org.eclipse.jetty.server.Request
type alias HttpServletRequest = ##jakarta.servlet.http.HttpServletRequest
type alias HttpServletResponse = ##jakarta.servlet.http.HttpServletResponse
type alias Server = ##org.eclipse.jetty.server.Server
type alias AbstractHandler = ##org.eclipse.jetty.server.handler.AbstractHandler
type alias Handler = ##org.eclipse.jetty.server.Handler
type alias AbstractLifeCycle = ##org.eclipse.jetty.util.component.AbstractLifeCycle
type alias HandlerWrapper = ##org.eclipse.jetty.server.handler.HandlerWrapper
type alias ServletResponse = ##jakarta.servlet.ServletResponse

// The main entry point.
def main(): Unit & Impure =
    import new org.eclipse.jetty.server.Server(Int32): Server & Impure as newServer;
    import org.eclipse.jetty.server.handler.HandlerWrapper.setHandler(Handler): Unit & Impure;
    import org.eclipse.jetty.util.component.AbstractLifeCycle.start(): Unit & Impure;
    import org.eclipse.jetty.server.Server.join(): Unit & Impure;
    import org.eclipse.jetty.server.Request.setHandled(Bool): Unit & Impure;
    import jakarta.servlet.http.HttpServletResponse.setStatus(Int32): Unit & Impure;
    import jakarta.servlet.ServletResponse.getWriter(): ##java.io.PrintWriter & Impure;
    import jakarta.servlet.ServletResponse.setContentType(String): Unit & Impure;
    import java.io.PrintWriter.print(String): Unit & Impure;

    let server = newServer(8080);
    let handler = object ##org.eclipse.jetty.server.handler.AbstractHandler {
        def handle(_this: ##org.eclipse.jetty.server.handler.AbstractHandler, _target: String, jettyRequest: ##org.eclipse.jetty.server.Request, _request: ##jakarta.servlet.http.HttpServletRequest, response: ##jakarta.servlet.http.HttpServletResponse): Unit =
            setHandled(jettyRequest, true);
            setStatus(response, 200);
            setContentType(response as ServletResponse, "text/html; charset=UTF-8");
            let writer = getWriter(response as ServletResponse);
            print(writer, "<h1>Hello from Flix!</h1>")
    };
    setHandler(server as HandlerWrapper, handler as Handler);
    start(server as AbstractLifeCycle);
    join(server)
```

paulbutcher:
Fixes #4363

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/ParsedAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/ParsedAst.scala
@@ -676,7 +676,7 @@ object ParsedAst {
       * @param exp the body expression.
       * @param sp2 the position of the last character in the expression.
       */
-    case class LetImport(sp1: SourcePosition, op: ParsedAst.JvmOp, exp: ParsedAst.Expression, sp2: SourcePosition) extends ParsedAst.Expression
+    case class LetImport(sp1: SourcePosition, op: ParsedAst.JvmOp, sp2: SourcePosition, exp: ParsedAst.Expression) extends ParsedAst.Expression
 
     /**
       * NewObject (create an anonymous object which implements a Java interface or extends a Java class).
--- a/main/src/ca/uwaterloo/flix/language/ast/ParsedAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/ParsedAst.scala
@@ -676,7 +676,7 @@ object ParsedAst {
       * @param exp the body expression.
       * @param sp2 the position of the last character in the expression.
       */
-    case class LetImport(sp1: SourcePosition, op: ParsedAst.JvmOp, sp2: SourcePosition, exp: ParsedAst.Expression) extends ParsedAst.Expression
+    case class LetImport(sp1: SourcePosition, op: ParsedAst.JvmOp, spIntermediate: SourcePosition, exp: ParsedAst.Expression, sp2: SourcePosition) extends ParsedAst.Expression
 
     /**
       * NewObject (create an anonymous object which implements a Java interface or extends a Java class).
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser.scala
@@ -859,7 +859,7 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
       }
 
       rule {
-        SP ~ Import ~ optWS ~ ";" ~ optWS ~ Stm ~ SP ~> ParsedAst.Expression.LetImport
+        SP ~ Import ~ optWS ~ ";" ~ SP ~ optWS ~ Stm ~> ParsedAst.Expression.LetImport
       }
     }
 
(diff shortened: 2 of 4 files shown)


magnus-madsen:
We have an invariant that the source location must span the entire expression. Thus I think the real fix is to have two source locations around here: https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/Weeder.scala#L843 and then use them appropriately. Probably around here: https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/Weeder.scala#L885

magnus-madsen:
But because of all the desugaring its not really so clear cut exactly what to do.

magnus-madsen:
Gosh. That's alot of code to look at :O

paulbutcher:
Are the changes I've just pushed what you have in mind?

(I've only done the `Method` case - if they are, I'll do the other cases too)

magnus-madsen:
Not really. Going waaay back. I think the problem is that we have:

```
def JavaName: Rule1[Seq[String]]
```

But we should really have an abstraction `Name.Java` (??) that has the Seq[String] **and** source positions.

paulbutcher:
Ah, got you.

An oddity that I don't understand and which may cast some light on this?:

The source position for `import new` statements is not broken. But the source position for method and static method imports is (I've not checked getters and setters yet). 

Why does `import new` work?

magnus-madsen:
> Ah, got you.
> 
> An oddity that I don't understand and which may cast some light on this?:
> 
> The source position for `import new` statements is not broken. But the source position for method and static method imports is (I've not checked getters and setters yet).
> 
> Why does `import new` work?

No idea.. Probably because the source locations are properly managed (??).

The real, long-term issue is that we should *move all desugaring __after__ type inference*. Thus all the way from the Weeder to Lowering. That way we preserve the program exactly as the programmer wrote it, and only start to break it down for compilation.

paulbutcher:
> No idea.. Probably because the source locations are properly managed (??).

Worked it out. It's not that it works for constructors. It works if we specify a name for the import.

So this works:

```
import ... as Foo
```

But without the `as`, we construct a synthetic identifier using `sp1` and `sp2`.

So yes, the solution is to have a source position associated with `Java.Name`

paulbutcher:
I'm going to close this PR as it's clearly not the right way forwards. I might take a look at adding source position to Java.Name in a separate PR.

magnus-madsen:
> I'm going to close this PR as it's clearly not the right way forwards. I might take a look at adding source position to Java.Name in a separate PR.

Thanks!

Issue #3720, opened 2022-05-11
CLI: commands and options


mlutze:
The CLI has the following **commands**:
* `init`
* `check`
* `build`
* `build-jar`
* `build-pkg`
* `run`
* `test`
* `install`

and the following **options**:
* `args`
* `benchmark`
* `doc`
* `explain`
* `help`
* `interactive`
* `json`
* `listen`
* `lsp`
* `output`
* `test`
* `threads`
* `version`

and the following **experimental options**:
* `Xbenchmark-code-size`
* `Xbenchmark-incremental`
* `Xbenchmark-phases`
* `Xbenchmark-throughput`
* `Xdebug`
* `Xlib`
* `Xstatistics`
* `Xstrictmono`

It seems to me that several of the "options" are really **commands**, since they don't modify anything but rather indicate what action to take:
* `benchmark`
* `doc`
* `help`
* `interactive`
* `lsp`
* `test`
* `version`
* `Xbenchmark-code-size`(?)
* `Xbenchmark-incremental`
* `Xbenchmark-phases`
* `Xbenchmark-throughput`

magnus-madsen:
I think it is pretty much only `doc` that remains and then to maybe hide the `--X` options?

magnus-madsen:
Ignoring the experimental options, I think we have come pretty close:

```
The Flix Programming Language 0.42.0
Usage: flix [init|check|build|build-jar|build-pkg|doc|run|benchmark|test|repl|lsp] [options] <args>...

Command: init
  creates a new project in the current directory.
Command: check
  checks the current project for errors.
Command: build
  builds (i.e. compiles) the current project.
Command: build-jar
  builds a jar-file from the current project.
Command: build-pkg
  builds a fpkg-file from the current project.
Command: doc
  generates API documentation.
Command: run
  runs main for the current project.
Command: benchmark
  runs the benchmarks for the current project.
Command: test
  runs the tests for the current project.
Command: repl
  starts a repl for the current project, or provided Flix source files.
Command: lsp port
  starts the LSP server and listens on the given port.
  port

  --args <a1, a2, ...>     arguments passed to main. Must be a single quoted string.
  --entrypoint <value>     specifies the main entry point.
  --explain                provides suggestions on how to solve a problem.
  --github-key <value>     API key to use for GitHub dependency resolution.
  --help                   prints this usage information.
  --json                   enables json output.
  --listen <port>          starts the socket server and listens on the given port.
[...]

PR #4401, opened 2022-07-29
experiment: principled CLI

mlutze:
Properly separates commands from options

This should help with several issues:
* #3717
* #3720
* an issue I can't find where options are not passed to LSP

### DIFF
--- a/main/src/ca/uwaterloo/flix/Main.scala
+++ b/main/src/ca/uwaterloo/flix/Main.scala
@@ -112,9 +112,6 @@ object Main {
     // check if command was passed.
     try {
       cmdOpts.command match {
-        case Command.None =>
-        // nop, continue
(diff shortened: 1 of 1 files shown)


magnus-madsen:
I am not so sold on this approach. I would rather deal with one "option" at a time and figure out how it should work. I think we need more discussion, possibly offline.

magnus-madsen:
We also have a big problem with "file" versus "project" mode that needs to be resolved.

magnus-madsen:
We don't have to close, but lets discuss more.

mlutze:
I've added this as a discussion topic for the next language design meeting.

magnus-madsen:
I am OK with this, but I want to do it gradually, flag by flag.

Issue #4395, opened 2022-07-29
Method import doesn't check return type


paulbutcher:
`import` checks that parameter types are correct, but doesn't check the return type, leading to subsequent issues.

For example, the following compiles:

```
import java.util.ArrayList;
import java.lang.Object;
import java.util.AbstractCollection;

// The main entry point.
def main(): Unit & Impure =
    import new java.util.ArrayList(): ArrayList & Impure as newArrayList;
    import java.util.ArrayList.add(Object): Unit & Impure;
    import java.util.AbstractCollection.toString(): String & Impure;
    let al = newArrayList();
    add(al, "Hello World!" as Object);
    println(toString(al as AbstractCollection))
```

But when run crashes with:

```
java -jar /Users/paulbutcher/Projects/flix-playground/flix.jar --entrypoint main /Users/paulbutcher/Projects/flix-playground/src/Main.flix /Users/paulbutcher/Projects/flix-playground/test/TestMain.flix 
paulbutcher@Pauls-MBP flix-playground % java -jar /Users/paulbutcher/Projects/flix-playground/flix.jar --entrypoi
nt main /Users/paulbutcher/Projects/flix-playground/src/Main.flix /Users/paulbutcher/Projects/flix-playground/tes
t/TestMain.flix 
Exception in thread "main" java.lang.VerifyError: Bad type on operand stack     
Exception Details:
  Location:
    Def%main%173003.invoke()LCont%Obj; @27: checkcast
  Reason:
    Type integer (current frame, stack[0]) is not assignable to 'java/lang/Object'
  Current Frame:
    bci: @27
    flags: { }
    locals: { 'Def%main%173003', 'java/lang/Object', 'java/lang/Object' }
    stack: { integer }
  Bytecode:
    0000000: 2ab4 000e 4c2a b400 114d 2bc0 0013 c000
    0000010: 132c c000 15c0 0015 b600 19c0 001b 2a5f
    0000020: b500 1e01 b0                           

        at java.base/java.lang.Class.getDeclaredMethods0(Native Method)
```

paulbutcher:
This isn't entirely straightforward to fix because `WeededAst.InvokeMethod` doesn't preserve the return type (making it impossible to check in the resolver.

@magnus-madsen: I think the solution requires adding the return type to `WeededAst.InvokeMethod` and then plumbing it through to the resolver. Sound reasonable?

magnus-madsen:
[...]

PR #4403, opened 2022-07-30
feat: check import return type

paulbutcher:
This PR is draft because definitely not yet complete. But I need some advice as to how to proceed from here.

I've plumbed the return type through the AST to the resolver. The check on this line is clearly wrong, but I'm not sure what the right way to check that the declared return value matches the method is?

https://github.com/paulbutcher/flix/blob/import-retval/main/src/ca/uwaterloo/flix/language/phase/Resolver.scala#L2507

Also not done yet: the same functionality for constructors and field access.


magnus-madsen:
Looks ok to me, but lets ask @mlutze if he agrees.

paulbutcher:
> Looks ok to me, but lets ask @mlutze if he agrees.

Well, it causes (as you can see!) plenty of failed tests, so I assume that there's a better approach...

magnus-madsen:
It should at least be a subtype check, I guess. But more seems to be wrong than that?

mlutze:
We're importing methods that return arrays with regions. A normal (regionless) array is `Array[a, Static]` so even a subtype check would be insufficient because `Static` is not a subtype of `r: Region`.

I'm not sure what the best solution is. In principle, we ought not to allow `import XYZ: Array[Int32, r]`, but it seems like a very convenient feature.

magnus-madsen:
Ah- I did not think of that complication.

paulbutcher:
Until I understand regions properly (which I currently don't), I think I need to rely on one of you to tell me what the right way forwards is here?

magnus-madsen:
Lets be practical and incremental. How about we start by doing the check for: (1) any primitive type and (2) any Java type. For all other types (arrays, enums, other Flix types) we skip the check. Once this works, we can think deeper about it or defer, since we will catch most issues anyway.

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -2491,18 +2493,21 @@ object Resolver {
...
         val method = clazz.getDeclaredMethod(methodName, sig: _*)
 
         // Check if the method should be and is static.
-        if (static == Modifier.isStatic(method.getModifiers))
-          method.toSuccess
-        else
+        if (static != Modifier.isStatic(method.getModifiers))
           throw new NoSuchMethodException()
+        // Check that the return type of the method matches the declared type
+        else if (Type.getFlixType(method.getReturnType) != retTpe)
```
magnus-madsen, line 2507: Lets special case this an only consider Java types to begin with (and maybe primitive types).

magnus-madsen:
We can add the more advanced checks once we have actually figured out how to integrate (1) generics, (2) regions, (3) effects, (4) function types with Java. For now a 90% solution seems like a great improvement.

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -2486,18 +2488,28 @@ object Resolver {
...
         val method = clazz.getDeclaredMethod(methodName, sig: _*)
 
         // Check if the method should be and is static.
-        if (static == Modifier.isStatic(method.getModifiers))
-          method.toSuccess
-        else
+        if (static != Modifier.isStatic(method.getModifiers))
           throw new NoSuchMethodException()
+        else 
+          // Check that the return type of the method matches the declared type.
```
magnus-madsen, line 2502: Overall approach looks: 👍 

A few things we should resolve:

- We have to call Type.eraseAliases on retTpe (I think).

- In Java, a method is uniquely determined by its name and formal parameter types, but not its return type. Hence we should have an new error message "MismatchedReturnType" that says we found the method, but the declared return type does not match the actual return type.

- We should do an explicit pattern match on the pair `((Type.getFlixType(method.getReturnType), Type.erase..(retTpe)` and handle these cases: `(Int8, Int8) => // OK` and so forth for all of these types: Bool, Char, FloatX, IntX, BigInt, String, and Native. Everything else goes through unchecked.

paulbutcher: Regarding "We have to call Type.eraseAliases on retTpe (I think)." what is the test that I should write to confirm/deny that this is the right thing to do?

mlutze: > Regarding "We have to call Type.eraseAliases on retTpe (I think)." what is the test that I should write to confirm/deny that this is the right thing to do?

You can define a type alias and use it as the return type. If Magnus is right, then you'll get an error. For example,

```
type alias MyString = ##java.lang.String
...
import new ##java.lang.String(): MyString \ Pure as newString; // This is valid because MyString = String but might not work if we don't erase aliases
```

paulbutcher: That makes sense. Thanks 👍

magnus-madsen: If you make the above changes then I think we can merge this 👍

paulbutcher: So I've added a test for aliases, and it passes with no code changes:

https://github.com/flix/flix/blob/332d8bd6c8fe700b0a0230f9f398df73b0c64ea6/main/test/flix/Test.Exp.Jvm.InvokeMethod.flix#L28

Which worries me somewhat. Have I screwed the test up somehow? If not, how is it passing without aliases being erased?

Would welcome thoughts.

magnus-madsen: I see that. @mlutze Any idea?

paulbutcher: I think I've worked it out: it's my fault. The match on `Type.Cst(_, _)` doesn't match a type alias, so it's assumed to succeed.

It should therefore be fixed when I make the change that @magnus-madsen recommended to the matching.

paulbutcher:
This is now ready to be merged, I think.

magnus-madsen:
Merged! Thanks

paulbutcher:
Fixes #4395

PR #4404, opened 2022-07-30
refactor: remove Type.Tag

mlutze:
related to #3731


magnus-madsen:
Any this does not break the correct typing of patterns and/or tag expressions?

mlutze:
> Any this does not break the correct typing of patterns and/or tag expressions?

Nope, changes them to a regular function type and extracts the tag directly from the case instead of from the type.

PR #4409, opened 2022-07-30
refactor: use lists for Instance phase

mlutze:


magnus-madsen:
All this flatMapping is worse than folding right?

mlutze:
Hmm, maybe. Though all the lists should be empty except in the error case, right?

PR #4412, opened 2022-07-30
fix: preserve array expression polymorphism in Typer

### LINKED ISSUE #4190: ArrayLoad should be effect-polymorphic
Currently in the Typer its purity is set to `Impure` in the reconstruction step

mlutze:
fixes #4190


magnus-madsen:
Conflicted.

PR #4416, opened 2022-08-01
Only display inlay hints when AST is current

### LINKED ISSUE #4415: Inlay hints broken during editing
Inlay hints are cool, but they're broken for source code that won't parse (which is most of the time when editing source code).

For example, imagine I want to turn this:

Screenshot 2022-07-31 at 21 41 45

into:

Screenshot 2022-07-31 at 21 44 03

The inlay hint displays just fine at the start and end points. But here's what it looks like while halfway through editing:

Screenshot 2022-07-31 at 21 44 59

The more complex the source, the worse this gets.

I believe that we should only display inlay hints if the AST is current (rather like this change which only displays code lenses if the AST is current #3999).

paulbutcher:
Fixes #4415


magnus-madsen:
Thanks!

PR #4421, opened 2022-08-02
refactor: `& Impure` to `\ IO` in namespace `File`

ninaandrup:
Related to #4377


ninaandrup:
@mlutze It is right, that there is a conflict with `Regions` and `IO` effect? I get a conflict when I try to refactor the functions in `File` that uses `Iterator`. For example if I refactor the helper function `fromJavaIter`:
```
def fromJavaIter(iter: ##java.util.Iterator): Iterator[a, Static] =
    import java.util.Iterator.hasNext(): Bool \ IO;
    import java.util.Iterator.next(): a \ IO as getNext;

    let done = () -> not hasNext(iter);
    let next = () -> {
        if(hasNext(iter)) {
            getNext(iter)
        } else {
            bug!("Empty iterator.")
        }
    };
    Iterator(done, next)
```
I get the error message:
```
-- Type Error -------------------------------------------------- File.flix

>> Unable to unify the Boolean formulas: 'Empty' and 'IO'.

984 |         Iterator(done, next)
              ^^^^^^^^^^^^^^^^^^^^
              mismatched boolean formulas.

Type One: ((Done[b137178], Next[t137177, b137178])) -> Iterator[t137177, b137178]
Type Two: ((Unit -> Bool & Impure \ IO, Unit -> ?result137173 & Impure \ IO)) -> t41842
```
I'm not sure, if it has something to do with the regions, because it seems like the compiler does not know, that the `done` is of type `pub type alias Done[ef: Region] = Unit -> Bool & ef` and the same for `next`.

If this is a big refactor, we can wait with the functions in `File` using `Iterator` for another PR :-)

ninaandrup:
I'm also having some problems with this helper function:
```
def readBytesWithHelper(bytes: Array[Int8, Static], numberRead: Int32, count: Int32): Array[Int8, Static] \ IO =
    let nothingRead = (numberRead == -1);
    if (nothingRead) {
        []
    } else {
        let countBiggest = numberRead < count;
        if (countBiggest) {
            Array.slice(0, numberRead, bytes)
        } else {
            bytes
        }
    }
```
I get the error:
```
-- Type Error -------------------------------------------------- File.flix

>> The type scheme: 'Array[Int8, false] -> (Int32 -> (Int32 -> Array[Int8, false] & Impure))' cannot be generalized to 'Array[Int8, Static] -> (Int32 -> (Int32 -> Array[Int8, Static] & Impure \ IO))'.

545 |     def readBytesWithHelper(bytes: Array[Int8, Static], numberRead: Int32, count: Int32): Array[Int8, Static] \ IO =
              ^^^^^^^^^^^^^^^^^^^
              unable to generalize the type scheme.

The declared type does not match the inferred type:

  Declared: Array[Int8, Static] -> (Int32 -> (Int32 -> Array[Int8, Static] & Impure \ IO))
  Inferred: Array[Int8, false] -> (Int32 -> (Int32 -> Array[Int8, false] & Impure))
```

mlutze:
The issue is that this declaration
```scala
pub type alias Done[ef: Region] = Unit -> Bool & ef
```
is only polymorphic in the boolean world since we explicitly mark the type variable `ef` as a Region (= Bool).

We will have to change the definition to be polymorphic in both the boolean world and the set world by using a type variable for each:
```scala
pub type alias Done[r: Region, ef: Eff] = Unit -> Bool & r \ ef
```

Some upcasts might also be necessary but this is a step in the right direction.

mlutze:
> I'm also having some problems with this helper function:
> 
> ```
> def readBytesWithHelper(bytes: Array[Int8, Static], numberRead: Int32, count: Int32): Array[Int8, Static] \ IO =
>     let nothingRead = (numberRead == -1);
>     if (nothingRead) {
>         []
>     } else {
>         let countBiggest = numberRead < count;
>         if (countBiggest) {
>             Array.slice(0, numberRead, bytes)
>         } else {
>             bytes
>         }
>     }
> ```
> 
> I get the error:
> 
> ```
> -- Type Error -------------------------------------------------- File.flix
> 
> >> The type scheme: 'Array[Int8, false] -> (Int32 -> (Int32 -> Array[Int8, false] & Impure))' cannot be generalized to 'Array[Int8, Static] -> (Int32 -> (Int32 -> Array[Int8, Static] & Impure \ IO))'.
> 
> 545 |     def readBytesWithHelper(bytes: Array[Int8, Static], numberRead: Int32, count: Int32): Array[Int8, Static] \ IO =
>               ^^^^^^^^^^^^^^^^^^^
>               unable to generalize the type scheme.
> 
> The declared type does not match the inferred type:
> 
>   Declared: Array[Int8, Static] -> (Int32 -> (Int32 -> Array[Int8, Static] & Impure \ IO))
>   Inferred: Array[Int8, false] -> (Int32 -> (Int32 -> Array[Int8, false] & Impure))
> ```

This one is because reading in the static region does not imply `IO`

I'm not totally sure what the right move is for this one. @magnus-madsen ?

magnus-madsen:
Hm.. I guess we need to discuss more.

@ninaandrup Can you try to look at some of the other namespaces and see if they are more agreeable?

magnus-madsen:
@mlutze Would it be better if you did the Impure -> IO refactor (after your deadlines)? I think its just a search and replace, but you need to fix several things along the way probably.

magnus-madsen:
I think this is on-hold. @mlutze Will take a look, IIRC.

magnus-madsen:
(We need to ignore IO in the set-system).

mlutze:
> I think this is on-hold. @mlutze Will take a look, IIRC.

Yes. #4425

magnus-madsen:
Fixed in master.

PR #4428, opened 2022-08-04
fix: bug in set DNF

mlutze:


magnus-madsen:
Where is my test cases :)

PR #4439, opened 2022-08-09
feat: make IO imply pure in SetUnification

mlutze:


magnus-madsen:
Can we deprecated the old syntax after this? Or rather, only allow it if you write `& foo \ bar`?

magnus-madsen:
Does this enable set effects? or not yet?

mlutze:
> Can we deprecated the old syntax after this? Or rather, only allow it if you write `& foo \ bar`?

I believe so.



> Does this enable set effects? or not yet?

I haven't touched the `--Xeffects` flag.

PR #4452, opened 2022-08-10
refactor: rewrite `Tester`

magnus-madsen:
This rewrites the `Tester`. No new features are added, but the architecture is changed.

### DIFF
--- a/main/src/ca/uwaterloo/flix/tools/Tester.scala
+++ b/main/src/ca/uwaterloo/flix/tools/Tester.scala
@@ -33,18 +33,25 @@ object Tester {
     * Runs all tests.
     */
   def run(compilationResult: CompilationResult)(implicit flix: Flix): Unit = {
+    //
+    // Find all test cases (both active and ignored).
+    //
+    val tests = getTestCases(compilationResult)
+
+    // Start the TestRunner and TestReporter.
     val queue = new ConcurrentLinkedQueue[TestEvent]()
-    val reporter = new TestReporter(queue, compilationResult)
-    val runner = new TestRunner(queue, compilationResult)
+    val reporter = new TestReporter(queue)
+    val runner = new TestRunner(queue, tests)
     reporter.start()
     runner.start()
 
+    // Wait for everything to complete.
     reporter.join()
     runner.join()
   }
 
   // TODO: DOC
-  private class TestReporter(queue: ConcurrentLinkedQueue[TestEvent], compilationResult: CompilationResult)(implicit flix: Flix) extends Thread {
+  private class TestReporter(queue: ConcurrentLinkedQueue[TestEvent])(implicit flix: Flix) extends Thread {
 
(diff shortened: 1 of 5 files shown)


main/src/ca/uwaterloo/flix/util/Duration.scala
```diff
@@ -43,7 +43,7 @@ object Duration {
 /**
   * A simple class to format a time duration.
   */
-class Duration(val d: Long) {
+case class Duration(val d: Long) {
```
mlutze, line 46: Redundant `val` :)

mlutze: Would be nice if this said what `d` represents (units)

main/src/ca/uwaterloo/flix/tools/Tester.scala
```diff
@@ -1,125 +1,208 @@
...
+    * A common super-type for test events.
+    */
+  sealed trait TestEvent
+
+  object TestEvent {
+
+    /**
+      * A test event emitted immediately before a test case is executed.
+      */
+    case class Before(sym: Symbol.DefnSym) extends TestEvent
```
mlutze, line 190: What's this for?

main/src/ca/uwaterloo/flix/tools/Tester.scala
```diff
@@ -1,125 +1,208 @@
...
+      writer.flush()
+
+      // Main event loop.
+      var passed = 0
+      var failed = 0
+
+      var finished = false
+      while (!finished) {
+        queue.poll() match {
+          case TestEvent.Success(sym, elapsed) =>
```
mlutze, line 80: A minor suggestion for your consideration:

It would be nice to show the active test, _then_ show whether or not it succeeded.
For example:
```
testDoesTheCoolStuff // no newline
// wait half a second
OK
```
so we end up with
```
test01 OK
test02 OK
test03 OK
test04
```

And we see we're waiting on test04. Maybe that one causes an infinite loop?

mlutze: Maybe that doesn't work well with concurrency though

main/src/ca/uwaterloo/flix/util/Duration.scala
magnus-madsen: Done!

main/src/ca/uwaterloo/flix/tools/Tester.scala
magnus-madsen: To print that we are "working on" the test. But not used yet.

main/src/ca/uwaterloo/flix/tools/Tester.scala
magnus-madsen: Agreed. Upcoming.

Issue #4443, opened 2022-08-09
Traversable.mapAccumLeft


magnus-madsen:
Question 1:

Is this code equivalent to `Traversable.mapAccumLeft`?:

```
pub def mapAccumLeft2(f: (acc, a) -> (acc, b) & ef, start: acc, t: t[a]): (acc, t[b]) & ef with Traversable[t] =
    region r {
        let acc = ref start @ r;
        let tb = t |> Functor.map(x -> {
            let (nacc, b) = f(deref acc, x);
            acc := nacc;
            b
        });
        (deref acc, tb)
    }
```

Question 2:

Does this mean `mapAccumLeft` only needs `Functor` and should be moved to a different namespace?

magnus-madsen:
@stephentetley Thoughts?

@JonathanStarup Can we use this as an example for the region paper? Can you try to look at `mapAccumRight` and see if it can be expressed in a similar way?

stephentetley:
Question 1

Yes (I think) if the functor implementation of `map` goes left to right (caveat below).

Question 2

Probably not. The documentation for Haskell's `Traversable` class makes the strong requirement that the traversal should be left-to-right, I missed this off when I implemented the Flix version and should have included it.  

`mapAccum` effectively carries a trace of previous states which can be a summary of nodes visited, so it is essential to the idea of the function that the visit order is defined. `Functor.map` 's traversal order is unspecified so we can't safely rely on it for a `mapAccumLeft` or `mapAccumRight`.

magnus-madsen:
Hmm. What about using Foldable? There we have fold left and right

stephentetley:
> Foldable

I don't think so - you want both the container (rebuilt / mapped) and the summary value. You can't rebuild the container with just `Foldable` the fold step would need to know about a `cons` operation.

magnus-madsen:
Oh right

magnus-madsen:
What happens if I traverse with `Identity`? Would that work? But then I guess I would not know how to accum from the right. 

Do you use these two functions?

(I bring these things up because the implementation is causing us a few problems with the effect system).

stephentetley:
I've made a version of `Traversable` where `ContStateLeft` and `ContStateRight` thread a region rather than IO so the effects of `map` and `ap` can be internally cast to the region but pushed up again to the outer-level so they aren't lost (this is the same trick to track the effects as before but with a region rather than IO/Impure). 

[...]

PR #4460, opened 2022-08-10
refactor: Traversable - ContStateLeft and ContStateRight to use a region rather than Impure effect #4443

stephentetley:
Please see issue #4443.

This PR changes Traversable's `ContStateLeft` and `ContStateRight` applicatives to thread a region effect rather than IO (Impure). We can internally cast the effects on `map` of Functor and `ap` of Applicative to the region and use an outer cast to push the effect back out the function's scope. 

Maybe this trick is not sufficiently safe as per the note in #4443 so I am on the fence about whether it merits merging into the mainline.

### DIFF
--- a/main/src/library/Traversable.flix
+++ b/main/src/library/Traversable.flix
@@ -15,7 +15,8 @@
  */
 
 ///
-/// A type class for data structures that can be traversed with an applicative functor.
+/// A type class for data structures that can be traversed in left-to-right
+/// order with an applicative functor.
 ///
 pub class Traversable[t : Type -> Type] with Functor[t], Foldable[t] {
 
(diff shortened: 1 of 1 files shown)


magnus-madsen:
I think we should refactor `mapAccumLeft` to use regions and remove `mapAccumRight`.

stephentetley:
@magnus-madsen Yes, I'm happy with that decision.

magnus-madsen:
> I think we should refactor `mapAccumLeft` to use regions and remove `mapAccumRight`.

Do you want to do that? :-)

stephentetley:
> Do you want to do that? :-)

Yes, I will do.

stephentetley:
Oh no, looking at this has revealed that `List.traverse` actually goes from right-to-left. So the change will have to fix that (and other implementations if they copied List's code?) as well...

magnus-madsen:
> Oh no, looking at this has revealed that `List.traverse` actually goes from right-to-left. So the change will have to fix that (and other implementations if they copied List's code?) as well...
> 
> [Edit - strictly speaking it does go from left to right but only it fires effects when it is gone to the right end and then runs them backwards]

Oh. That's not good. I would think that the effects should occur from left to right.

stephentetley:
> Oh. That's not good. I would think that the effects should occur from left to right.

Definitely! I'm just about to make a PR to fix `List.traverse`. Then I'll look at the other instances (e.g Chain, RedBlackTree...). Once they are working I'll do the Traversable PR.

magnus-madsen:
Is this PR still relevant?

stephentetley:
> Is this PR still relevant?

No, It can be rejected / discarded. With the recent `traverse` fixes I should be able to make a new PR for the mutable state version of `mapAccumLeft`.

PR #4466, opened 2022-08-12
feat: turn effects on by default

mlutze:


mlutze:
!!!benchmark even though we all know it's unreliable

mlutze:
Wow, benchmark bot is broken :(

magnus-madsen:
Maybe run them locally?

mlutze:
My machine says that the master branch is 50% slower than the feature branch :P

mlutze:
A rerun says master is 10% faster

magnus-madsen:
@stephentetley This enables the new effect system by default. Let us know if you encounter breakages. (Those would most likely involve enums with effect types).

stephentetley:
Thanks Magnus, I'm switching my code over to use the effect set notation. I'll see if there are any breakages.

Issue #4286, opened 2022-07-16
Add `Par` Expression


magnus-madsen:
I want to add two new constructs: `ParApply` and `ParTag`. The latter is very similar to the former, but I suggest to implement `ParApply` fully before tackling `ParTag`. 

Both constructs should exists until the Lowering phase where they are translated. 

We allow a programmer to write:

```
par exp1(exp2, exp3, ...)
```

The type system treat this as a normal call, except that `exp2, ... expn` must be pure.

During the lowering phase, we translate `par exp1(exp2, exp3, ...)` to:

```
{
let ch1 = channel 1;
let ch2 = channel 1;
let ch3 = channel 1;
spawn ch1 <- exp1;
spawn ch2 <- exp2;
spawn ch3 <- exp3;
let tmp1 = <- ch1;
let tmp2 = <- ch2;
let tmp3 = <- ch3;
(tmp1)(tmp2, tmp3)
}  as & * the effect of tmp1 to tmp3*
```

magnus-madsen:
@jaschdoc I think this should be a fun task (after/during upcast work) :)

athas:
Why mix it up with function application at all, instead of just returning a tuple?

magnus-madsen:
> Why mix it up with function application at all, instead of just returning a tuple?

It's a good point! If you think about it, parallelism is just the ability to write `exp1 || exp2 || exp3` (possibly with some mechanism for communication). Taking this a step further, you would want a "parallel tuple": `par (exp1, exp2, exp3)`. Then, if you were in Standard ML, you could write `f (par (exp1, exp2, exp3))` and you would have what I propose... But unfortunately Flix has multi-argument functions (and not functions taking tuples), hence this would not be as elegant. Now why do it for calls? I don't have a good answer, except its often were funny stuff happens: call-by-value, call-by-name, etc. etc. Finally, I believe it will allow me to write a recursive parallel List.map function is a straightforward manner and explain it to 1st year student. 

Anyway, its still an experiment! If its no good, we can take it out.

athas:
Surely just having parallel tuples also lets you write a parallel Lisp.map?  Pseudocode:

```
let (y,ys) = par(f(x),List.map(f,xs))
Cons(y,ys)
```

Although if this `par` construct always returns a tuple, you might as well just write ``Cons(par(f(x),List.map(f,xs)))` (maybe you'll need some curry/uncurry stuff; I can never keep track of how tuples interact with using parens for function application).

[...]

PR #4469, opened 2022-08-18
feat: add `Par` expression for tuples

jaschdoc:
Related to #4286


jaschdoc:
RFR

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1345,6 +1352,51 @@ object Lowering {
...
+    */
+  private def mkLetSym(prefix: String, loc: SourceLocation)(implicit flix: Flix): Symbol.VarSym = {
+    val name = prefix + Flix.Delimiter + flix.genSym.freshId()
+    Symbol.freshVarSym(name, BoundBy.Let, loc)
+  }
+
+  /**
+    * Returns a tuple expression that is evaluated in parallel.
+    */
+  private def mkParTuple(tuple: Expression.Tuple)(implicit flix: Flix): Expression = {
```
mlutze, line 1368: It would be nice to have a comment that details the transformation. E.g. `par (a, b, c)` transforms into `[the lines that are generated]`

mlutze: And maybe an inline comment for each chunk of expressions you build

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1345,6 +1352,51 @@ object Lowering {
...
+    val Expression.Tuple(elms, tpe, pur, eff, loc) = tuple
+    val chanSymsWithElms = elms.map(e => (mkLetSym("channel", e.loc.asSynthetic), e))
+
+    // make wait expressions
+    val resultTuple = Expression.Tuple(chanSymsWithElms.map {
```
mlutze, line 1373: I think we should pull this map expression into its own variable

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1345,6 +1352,51 @@ object Lowering {
...
+          Expression.Var(sym, Type.mkChannel(e.tpe, loc), loc),
+          e.tpe, Type.Impure, e.eff, loc)
+    }, tpe, pur, eff, loc.asSynthetic)
+
+    // make spawn expressions
+    val spawns = chanSymsWithElms.foldRight(resultTuple: Expression) {
+      case ((sym, e), acc) =>
+        val loc = e.loc.asSynthetic
+        val e1 = Expression.Var(sym, Type.mkChannel(e.tpe, loc), loc)
+        val e2 = Expression.PutChannel(e1, e, Type.Unit, Type.Impure, Type.mkUnion(e.eff, e1.eff, loc), loc)
```
mlutze, line 1386: High-level question. We introduce `Impure` here for an expression that may not have been `Impure` before. Do we do this elsewhere? Is there reason to be concerned?

mlutze (commented):
This is _so much cleaner_ than before. Well done.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
jaschdoc: My thinking is that it is an Impure expression in itself, but the caller wraps it in a Cast expression which casts it to pure. This is okay since we've checked during the typer that the original expression was pure.

jaschdoc:
Okay added some comments now

PR #4470, opened 2022-08-18
feat: add `Par` expression for function application

jaschdoc:
Related to #4286


jaschdoc:
RFR

magnus-madsen:
Conflicted.

magnus-madsen:
But I do think we want it for apply.

jaschdoc:
Merged

magnus-madsen:
Nonexhaustive match :)

jaschdoc:
Should work now

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -573,6 +573,9 @@ object Lowering {
         val t = visitType(tpe)
         val parExp = mkParApply(Expression.Apply(e, es, t, pur, eff, loc1))
         Expression.Cast(parExp, None, Some(Type.Pure), Some(Type.Empty), t, pur, eff, loc0)
+
+      case _ =>
```
magnus-madsen, line 577: We don't use wildcards, because then we don't get warnings when we change the AST. We should probably add that to STYLE.md

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -574,8 +574,8 @@ object Lowering {
         val parExp = mkParApply(Expression.Apply(e, es, t, pur, eff, loc1))
         Expression.Cast(parExp, None, Some(Type.Pure), Some(Type.Empty), t, pur, eff, loc0)
 
-      case _ =>
-        throw InternalCompilerException("Unexpected par expression")
+      case e =>
```
magnus-madsen, line 577: No, I mean, you should write it out `case Exp.Par => `.

We want it so that when we add an new expression `Foo` then we get "Unmatched Foo`-- which the wildcard is blocking :)

jaschdoc: I'm not sure what you mean. Do you mean that we should add a case for each expression within in the par case (we could add a `visitParExp` helper function for this of course)?

jaschdoc: I.e. an exhaustive match

magnus-madsen: You should have an explicit case for Par (like all the other cases).

jaschdoc: Oh so you want to handle the case `par par f(x)`?

magnus-madsen:
@jaschdoc Oh god no, that's not what I had in mind :)

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
magnus-madsen: No, thats illegal.

magnus-madsen:
What you had before was fine, except you needed `Expression.Par(..) => throw ...`

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
magnus-madsen: Okay, I get it now. I think maybe I was wrong.

magnus-madsen:
Ill take a look at this with fresh eyes later.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -560,11 +560,23 @@ object Lowering {
...
+        Expression.Cast(e, None, Some(Type.Pure), Some(Type.Empty), t, pur, eff, loc0)
+
+      case Expression.Apply(exp, exps, tpe, pur, eff, loc1) =>
+        val e = visitExp(exp)
+        val es = visitExps(exps)
+        val t = visitType(tpe)
+        val parExp = mkParApply(Expression.Apply(e, es, t, pur, eff, loc1))
+        Expression.Cast(parExp, None, Some(Type.Pure), Some(Type.Empty), t, pur, eff, loc0)
+
+      case Expression.Par(_, _) =>
```
magnus-madsen, line 577: OK, I see now that this has to be a wildcard. Thats OK.

PR #4472, opened 2022-08-18
feat: add `Par` expression for tags

jaschdoc:
Related to #4286

Depends on #4470 

Does not currently evaluate the tag symbol in parallel. Not quite sure how to deal with that. Any input would be appreciated :)


magnus-madsen:
I don't think we want this for tags.

PR #4473, opened 2022-08-19
feat: add `UnorderedTraversable`

### LINKED ISSUE #3068: Type Classes: Add UnorderedTraversable
This class probably only has a few methods and must make use of commutativity.

jaschdoc:
Closes #3068


jaschdoc:
Hmm I'm not quite sure how to write the commutative law for `CommutativeApplicative`. Maybe it's because I'm not completely with how `Applicative` works in the first place. Any help would be appreciated :)

jaschdoc:
Is `Set` applicative and commutative? I see Cats has an unordered traverse instance and they use a commutative applicative instance on `G` to implement the unordered traverse instance.
https://github.com/typelevel/cats/blob/791c65d3b13715e2c3a0de6e1d19045b8ff34b11/core/src/main/scala/cats/instances/set.scala#L34

magnus-madsen:
My current thinking of just that we should drop these Unordered* classes.

PR #4482, opened 2022-08-23
refactor: make Validation typeclassy

mlutze:
I think I've discovered the way to make Scala return the right type, so that Validation.sequence over an option will give back an option.


mlutze:
does not work with options because: `Option <: IterableOnce` and 

> IterableOnce does not extend [IterableOnceOps](https://www.scala-lang.org/api/current/scala/collection/IterableOnceOps.html). This is different than the general design of the collections library

Issue #4445, opened 2022-08-09
Add `--Xno-bool-effects` and `--Xno-set-effects`


magnus-madsen:

mlutze:
This should affect:
* [ ] typechecking
* [ ] redundancy checks
* [ ] error messages
* [ ] LSP information

PR #4484, opened 2022-08-26
feat: add Xno-bool-effects

mlutze:
related to #4445 

* [x] typechecking
* [ ] redundancy checks
* [ ] error messages
* [ ] LSP information
* [ ] documentation?


mlutze:
I'm marking this as ready. It only partially addresses the issue, but the formatting stuff is a bit more complicated since it seems natural to merge it with the `Audience` field, which I think calls for a bigger refactor.

PR #4496, opened 2022-08-30
refactor: `Formatter.scala` to be consistent with `Console.flix`

### LINKED ISSUE #4459: Refactor `Formatter.scala` to be consistent with `Console.flix`
- Offer the same set of color functions, e.g. `red` and `bgRed`, etc.
- Use the same colors in Scala as in Flix.

ninaandrup:
Closes #4459


main/src/ca/uwaterloo/flix/util/Formatter.scala
```diff
@@ -131,33 +208,71 @@ object Formatter {
 
     override def black(s: String): String = Console.BLACK + s + Console.RESET
 
+    override def brightBlack(s: String): String = fgColor(128, 128, 128, s)
```
ninaandrup, line 211: @magnus-madsen  Should the colors be added to AnsiColor.scala instead of using rgb?

magnus-madsen: What do you mean?

ninaandrup: Like for black, where I can write:
```
  override def black(s: String): String = Console.BLACK + s + Console.RESET
```
The `Console.BLACK` part.

magnus-madsen: Oh. I would avoid using Console and just use rgb-- for consistency.

ninaandrup: Should I change the ones using `Console` to `rgb`? Or just leave them as is

magnus-madsen: Change them to make it consistent

magnus-madsen:
Thanks!

PR #4497, opened 2022-08-30
fix: Make List.traverse effects run left-to-right 

stephentetley:
While looking at changing `Traversable.mapAccumLeft` to use local mutable state I discovered that `List.traverse` does not run its effects in left-to-right order. Instead, it traverses the List, builds up a stack of effects to run and performs them backwards (the rightmost effect is top of the stack so it is performed first). 

This PR fixes `List.traverse` to perform the effects in left to right order and adds tests so we can see it working.

Counter-intuitively `List.sequence` for all appearances does the same traversal as the original `List.traverse` but it performs the effects in the correct left to right order (I have added tests for `List.sequence` but the implementation is unchanged). 

The new implementation `List.traverse` is a bit tricky - to get left-to-right effects and to not escape the applicative context, the implementation is not in CPS. Instead it it builds an "applicative DList" so we can build the answer list in the correct order and avoid doing a reverse in a post-processing step.

### DIFF
--- a/main/src/library/List.flix
+++ b/main/src/library/List.flix
@@ -1377,16 +1377,19 @@ namespace List {
         List.foldRight((x, acc) -> DelayMap.insert(fst(x), snd(x), acc), DelayMap.empty(), l)
 
     ///
-    /// Helper function for `traverse` and `sequence`.
+    /// Helper function for `sequence`.
     ///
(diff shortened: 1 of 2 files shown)


magnus-madsen:
Thanks Stephen :)

PR #4499, opened 2022-08-31
chore: Use LibMin where necessary in tests

paulbutcher:
Some tests used `IO` with `LibNix`. This doesn't cause the tests to fail, but is confusing during debugging (because `IO` requires `LibMin`).


magnus-madsen:
Does this impact the time it takes to run tests? What's the time before/after?

magnus-madsen:
@mlutze Why does `IO` require LibMin? Does LibNix not include anything at all?

paulbutcher:
The last PR I created ran the tests in 10:23. Let's see how long this takes once it's complete (although I expect it will make a very small difference).

It's worth mentioning that there are many tests that use `IO`, and only a few that didn't already use `LibMin` (I just happened to be working in the Resolver, which had almost all of them).

mlutze:
> @mlutze Why does `IO` require LibMin? Does LibNix not include anything at all?

`IO` is in the prelude. LibNix includes no Flix files whatsoever.

paulbutcher:
BTW, not sure why all those merge requests have started appearing in my PRs: I've clearly got my fork confused somehow. I've done a `reset --hard` on my fork which will hopefully avoid them in the future.

PR #4503, opened 2022-08-31
fix: consistent fromString signature

### LINKED ISSUE #4420: Inconsistent fromString return type
`Int32.fromString` (and other numeric types) returns `Result`, whereas `FromString.fromString` returns `Option`.

paulbutcher:
Fixes #4420


magnus-madsen:
Thanks!

Issue #4392, opened 2022-07-28
Polish `import`


magnus-madsen:
Tasks:
- [x] Allow import of multiple classes/interfaces in one declaration, i.e. `import package.{A, B, C}`.
- [x] Allow an import to rename a class or interface `import package.{A => B}` (note the name `B` *must* be uppercase).
- [x] Allow ***top-level*** imports and uses without semi-colons.
- [x] Do not desugar imports to type aliases, but keep them in a new import `ienv`.
- [ ] Method imports should default to `IO` (old `Impure`) effect. 
- [x] Add a specific type error for when two Java types do not unify and suggest to use `upcast`.
- [ ] Support generic types at the Flix type level, i.e. also use the kinds of classes and interfaces.

Discuss:
- [x] Rename `object ClassOrInterface {...}` to `new ClassOrInterface {...}`
- [x] Allow overriding non-abstract methods from classes? (Why not? - It seems this feature would be free for us.)

Potential missing checks:
- [x] For `NewObject`, the default constructor must not only be empty, but also public.
- [x] For `NewObject`, ensure type aliases are erased for both the NewObject type and the parameter type.

LSP:
- [ ] Given the prefix `object ...` only suggest types (or give preference to types).
- [ ] Add auto-complete for import statements, e.g. `import package.<auto complete>`.
- [ ] Add auto complete for `object Foo {` which fills in all signatures in foo (This would be awesome!).

Experiments:
- [ ] Experiment with generic Java types. Can we use `ArrayList<T>`? Can we extend `List<T>`?
- [ ] ~Experiment with method imports without signatures provided that only one variant exists.~ (Broken)
- [ ] Experiment with `import X.Y` where `X` has already been imported. (Not before desugaring has been refactored).

Implementation stuff:
- [x] I think Safety.isAbstract can be simplified to just `Modifier.isAbstract`. 
- [x] `hasDefaultConstructor` should be `hasPublicZeroArgConstructor` and check for public.
- [x] Add `isStatic` and `isPublic` to be inline with `isAbstract`?

As usual, I much prefer small PRs that target one bullet at a time.

Off the table (per previous discussions):
- Wildcards imports (per previous discussions about ambiguity).

JonathanStarup:
* The rename syntax for use is `use List.{toArray => listToArray}` so we should probably also use `=>` here.
* I like this syntax `new ClassOrInterface {...}`

mlutze:
###### Question 1
[...]

PR #4507, opened 2022-09-01
feat: make semicolons optional for imports and uses

paulbutcher:
Related to #4392

This simply removes the semi colon from top-level use and import. I didn't see any particular value in allowing it to be optional, but let me know if you think I should.


paulbutcher:
The community build should pass when https://github.com/mlutze/fcwg/pull/69 is merged.

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
```diff
@@ -151,7 +151,7 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
 
   def UsesOrImports: Rule1[Seq[ParsedAst.UseOrImport]] = rule {
     // It is important for documentation comments that whitespace is not consumed if no uses are present
-    (optWS ~ oneOrMore((Use | Import) ~ optWS ~ ";").separatedBy(optWS)) | push(Seq.empty)
+    optWS ~ oneOrMore(Use | Import).separatedBy(optWS) | push(Seq.empty)
```
JonathanStarup, line 154: I'm not sure what the precedence of ~ and | is, but its should be `(a ~ b) | c` not `a ~ (b | c)` otherwise the docs comments after `use` is consumed. maybe just keep the paranthesis for clarity, whatever the precedence is?

magnus-madsen:
Ill look over the changes soon, in the interim, I think we need to allow a period where both the old and new syntax is supported (at least one stable version), since both Jaco and Stephen has a lot of code this would break.

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
paulbutcher: Agreed.

paulbutcher:
> I think we need to allow a period where both the old and new syntax is supported 

Agreed. Done.

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
```diff
@@ -151,7 +151,7 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
 
   def UsesOrImports: Rule1[Seq[ParsedAst.UseOrImport]] = rule {
     // It is important for documentation comments that whitespace is not consumed if no uses are present
-    (optWS ~ oneOrMore((Use | Import) ~ optWS ~ ";").separatedBy(optWS)) | push(Seq.empty)
+    (optWS ~ oneOrMore((Use | Import) ~ optWS ~ optional(";")).separatedBy(optWS)) | push(Seq.empty)
```
magnus-madsen, line 154: This is not completely correct because it would allow two uses next to eachother without whitespace, e.g. `use FoouseBar`.

magnus-madsen: We want something like either there is optWS and a semicolon or there is hard WS.

paulbutcher: Good point. Fixed, I think.

JonathanStarup: `oneOrMore((Use | Import) ~ optWS ~ optional(";")).separatedBy(optWS)` this sequence could also end on optWS and thus eat the doc comments

paulbutcher: @JonathanStarup I think my change and your comment crossed in the aether. Does what I've just pushed address your concern?

paulbutcher:
Hmm - that clearly doesn't do the trick - looking.

paulbutcher:
Well that was fiddlier than I expected. Finally there now, I think.

magnus-madsen:
Thanks!

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
JonathanStarup: The final version looks good 👍

PR #4508, opened 2022-09-01
fix: make traverse implementations run left-to-right

stephentetley:
This PR fixes the `traverse` implementations for `DelayList`, `Nec` and `RedBlackTree` (called `mapAWithKey`) to go left-to-right (this is a prelude to implementing `Traverse.mapAccumLeft` with mutable state).
    
It also simplifies the `List.traverse` fix from two days ago so the helper function `consAD` is no longer needed, and it adds `sequence` and `traverse` tests for the container types than provide them in their APIs.
    
The `sequence` tests for `List` from two days ago overstated their "effectiveness" (pun intended). The effects `sequence01` etc. were supposed to show were actually being performed by the `List.map` calls that preprocessed the input and not by the call to `List.sequence`. To show `List.sequence` running effects in left-to-right order, we would need a monad that does something effective unlike Identity. For the time being I've just simplified the tests so they show Identity being "lifted" from each element in the input list to the outer type constructor of the result list. To show that effects are run in left-to-right order we will need an imperative monad in the stdlib or implement a local one in the test suite.

This caveat for the `sequence` tests doesn't apply to the `traverse` tests because `traverse` is supplied with a function to run in the traversal and this function can be effective.

### DIFF
--- a/main/src/library/DelayList.flix
+++ b/main/src/library/DelayList.flix
@@ -1315,8 +1315,8 @@ namespace DelayList {
     pub def traverse(f: a -> m[b] \ ef, l: DelayList[a]): m[DelayList[b]] \ ef with Applicative[m] =
         def loop(ll, k) = match ll {
             case ENil         => k(Applicative.point(ENil))
-            case ECons(x, xs) => loop(xs, ks -> k(consA(f(x), ks)))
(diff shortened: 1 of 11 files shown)


magnus-madsen:
Thanks- I don't fully understand the details, but I will take your word and test cases for it :)

PR #4512, opened 2022-09-02
fix: reject `par` `Apply` instead of crashing

### LINKED ISSUE #4491: `Par` expression crashes
This program:

```
def main(): Unit \ IO = 
    List.range(1, 100) |> parMap(x -> x + 1) |> println

def parMap(f: a -> b, l: List[a]): List[b] = match l {
    case Nil            => Nil
    case Cons(x, xs)    => Cons(par f(x), parMap(f, xs))
}
```

should be rejected, but not crash the compiler. (Probably an easy fix).

jaschdoc:
Closes  #4491

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Safety.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Safety.scala
@@ -255,9 +255,6 @@ object Safety {
     case Expression.Par(exp: Expression.Tuple, _) =>
       visitExp(exp)
 
-    case Expression.Par(exp: Expression.Apply, _) =>
-      visitExp(exp)
-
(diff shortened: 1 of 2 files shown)


main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -255,9 +255,6 @@ object Safety {
     case Expression.Par(exp: Expression.Tuple, _) =>
```
magnus-madsen, line 255: I would like to have just one Par case where you match on the inner expression and then a comment that explains what is happening :)

jaschdoc:
fixed

PR #4515, opened 2022-09-02
refactor: reimplement Traversable.mapAccumLeft to use mutable state and remove mapAccumRight

stephentetley:
This PR re-implements `Traversable.mapAccumLeft` with mutable state. The previous version used a cont-state monad that used casts that were causing problems.

The PR also removes `Traversable.mapAccumRight` - unlike `mapAccumLeft` this cannot be written in terms of `Traverse.traverse`.

### DIFF
--- a/main/src/library/Traversable.flix
+++ b/main/src/library/Traversable.flix
@@ -15,7 +15,8 @@
  */
 
 ///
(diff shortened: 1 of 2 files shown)


magnus-madsen:
Thanks!

PR #4521, opened 2022-09-04
feat: Disallow semi-colons

paulbutcher:
Follow-on for #4507, this can be merged whenever we decide we want to move from optional semi-colons after top-level uses and imports to no semi-colons.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser.scala
@@ -151,7 +151,7 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
 
   def UsesOrImports: Rule1[Seq[ParsedAst.UseOrImport]] = rule {
     // It is important for documentation comments that whitespace is not consumed if no uses are present
-    oneOrMore(optWS ~ (Use | Import) ~ ((optWS ~ ";") | WS)) | push(Seq.empty)
+    oneOrMore(optWS ~ (Use | Import) ~ WS) | push(Seq.empty)
   }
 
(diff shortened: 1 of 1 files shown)


magnus-madsen:
We will merge this once the we have had one or two stable releases that allow optional semi colons.

magnus-madsen:
I am now thinking that we should simply allow optional semicolons. I think ultimately that is what we would want anyway.

Now we have it for imports and uses, but not yet generally.

paulbutcher:
When you say allow optional semi colons, do you mean optional everywhere that they're currently disallowed? Or are you suggesting that we should add semi colon inference within functions?

magnus-madsen:
> When you say allow optional semi colons, do you mean optional everywhere that they're currently disallowed? Or are you suggesting that we should add semi colon inference within functions?

I am suggesting that ultimately it would be nice to have semi-colon inference (almost) everywhere.

Note that imports and uses can already occur within functions. 

Thus is simply propose that we stick with the current implementation (which brings us some semi colon "inference").

paulbutcher:
OK, got you. Semi colon inference is famously a can of worms: do you have any thoughts on how it might be best approached within Flix?

magnus-madsen:
> OK, got you. Semi colon inference is famously a can of worms: do you have any thoughts on how it might be best approached within Flix?

Very carefully.

magnus-madsen:
That was a joke :) But anyway, I think it is something we should add a bit later, once the language has matured even more, but we can allow it for imports.

magnus-madsen:
I plan to continue to allow semicolons

PR #4520, opened 2022-09-04
feat: Allow inherited methods to be invoked

paulbutcher:
As per @mlutze suggestion in #4392.

Note: This only addresses calling methods. We'll need to make similar changes for fields and static methods, but I wanted to get this reviewed before going too far.

PR #4529, opened 2022-09-05
feat: Support get and set for inherited fields

paulbutcher:
Follow-on from #4520 (and related to #4392).

Turns out that calling inherited static methods, and setting and getting inherited fields already worked, but I've added tests to confirm.


magnus-madsen:
Thanks!

PR #4534, opened 2022-09-06
feat: turn --doc into doc

### LINKED ISSUE #3717: Don't launch REPL on `--doc`

mlutze:
fixes #3717, related to #3720

Currently requires the user to be in a Flix project, which I assume is too annoying.

### DIFF
--- a/main/src/ca/uwaterloo/flix/Main.scala
+++ b/main/src/ca/uwaterloo/flix/Main.scala
@@ -174,6 +174,8 @@ object Main {
           System.exit(0)
 
         case Command.Doc =>
+          val result = Packager.doc(cwd, options)
+          System.exit(getCode(result))
 
 
       }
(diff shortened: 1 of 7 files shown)


magnus-madsen:
> Currently requires the user to be in a Flix project, which I assume is too annoying.

Yes, because how to generate for std lib.

magnus-madsen:
This kind of blocks the way I build the documentation.

magnus-madsen:
Should all package commands just work without being in a package format?

mlutze:
> This kind of blocks the way I build the documentation.

Can you elaborate?



> Should all package commands just work without being in a package format?

Maybe someday, but for now there's a difference between `--test` and `test`, and `build-pkg` doesn't really work for non-package projects.

magnus-madsen:
> 

I would work on unifying these. (Not sure about the build-pkg etc.)

magnus-madsen:
We will revisit.

PR #4543, opened 2022-09-07
refactor: use formulas for Boolean unification

### LINKED ISSUE #4539: Refactor BoolUnification and SetUnification to use `Formula`
Possibly need to split `Formula` into two. 

I think by immediately translating from types into something simpler will significantly simplify the implementation.

mlutze:
fixes #4539


magnus-madsen:
I think working on formiulas and unifiying those formulas into one is two separate things. I would prefer to do the formulas first and only think about unification later. (In fact, I don't think we want to unify them because then we cannot exploit BDDs for the non-set formulas).

magnus-madsen:
So I guess this is not exactly what I had in mind :)

mlutze:
I see why it looks like this is meant for both sets and bools, since there's some eff stuff about. But my target here was only the Bool side of things.

magnus-madsen:
- What was the conclusion on the variable order issue?
- Can you provide a high-level description of this refactoring? "I don't get it"

main/src/ca/uwaterloo/flix/language/phase/unification/BoolAlgebraSubstitution.scala
```diff
@@ -0,0 +1,125 @@
...
+
+  /**
+    * Returns `true` if `this` is the empty substitution.
+    */
+  val isEmpty: Boolean = m.isEmpty
+
+  /**
+    * Applies `this` substitution to the given type `tpe0`.
+    */
+  def apply(f00: BoolAlgebra): BoolAlgebra = {
```
magnus-madsen, line 55: I really would prefer not to use this naming. I think it goes way too far :)

magnus-madsen:
- This seems to unify Bool and Set or at least use similar names, but that is not what I intended, because I want to experiment with different representations, but only for the Booleans.

magnus-madsen:
> * This seems to unify Bool and Set or at least use similar names, but that is not what I intended, because I want to experiment with different representations, but only for the Booleans.

OK, I think I understand my confusion now: We already previously introduced `BoolAlgebra` but rather than being a BoolAlgebra its hardcoded to be the two-valued Boolean algebra. (i.e. its not an interface, its an implementation). It would be a bit like calling `List` for `Monoid`.

mlutze:
> * What was the conclusion on the variable order issue?

Not really much of a conclusion. I just don't reverse the order of the variables and it magically works.
It seems to me that this is to do with the problems we faced when trying to add Region syms, where substitutions would allow some variables to go un-purified.

magnus-madsen:
> 

Hmm. That's a bit worrisome.

magnus-madsen:
What do you think of the following:

- Introduce a `BoolAgebra` trait.
- Rename the current BoolAlgebra to `BoolFormula` (or whatever).
- Introduce a `Substitution[a]` generic class (which possibly uses `BoolAlgebra`).

mlutze:
> What do you think of the following:
> 
>     * Introduce a `BoolAgebra` trait.
> 
>     * Rename the current BoolAlgebra to `BoolFormula` (or whatever).
> 
>     * Introduce a `Substitution[a]` generic class (which possibly uses `BoolAlgebra`).

I am concerned that generifying `Substitution` won't be that easy. It's not just `Type` vs `Formula`. Substitution also needs:
- specific `mkNot` etc. smart constructors.
- some kind of `traverse` function
- knowledge of the `id`/`sym` of the variables

magnus-madsen:
> * specific `mkNot` etc. smart constructors.
> * some kind of `traverse` function
> * knowledge of the `id`/`sym` of the variables

(1) Would come from the BoolAlg trait.
(2) Which traverse is this? Would it not come from BoolAlg also?
(3) Not sure.

magnus-madsen:
Does my overall point about naming and interfaces vs. impl make sense?

magnus-madsen:
I think I can state more precisely now what I would like:

- I want to abstract over two-valued Boolean algebras (but not yet arbitrary Boolean algebras).
- This means I need a trait `BoolAlg` with `mkTrue`, `mkFalse`, `mkAnd` etc. 
- We also need a substitution defined in terms of `BoolAlg`.
- Then we express SVE using `BoolAlg` and `BoolSubst`.
- Finally, we will have one type, `BoolFormula` that implements `BoolAlg`.
- This sets us up to implement `BoolAlg` with `BddFormula` in the future. 

Everything about sets and other BoolAlgs should be left separate.

magnus-madsen:
I think, at minimum, we need to undo some renaming:

- `BoolAlgebra` -> `BoolFormula`
- `Top` -> `True`, `Bot -> False`. etc.

mlutze:
Ok, I now understand what you mean.

mlutze:
It's looking like this is better modeled as a type class. Are we okay with using Scala's pseudo-type class mechanisms?

mlutze:
> It's looking like this is better modeled as a type class. Are we okay with using Scala's pseudo-type class mechanisms?

Or I guess we just have a separate "factory" object that implements the trait.

magnus-madsen:
Can be a type class

mlutze:
Currently (before this PR) we have a `Formula` which is only used for minimization, shared between Set and Bool effects.

If I understand correctly, that `Formula` need _not_ implement the `BoolFormulaTrait`. Instead we have a new `BoolFormula`.

Is that corrrect?

If so, we will need additional functions for converting `MyBoolImpl <-> FormulaForMinimization`

So far the trait looks like this:

```scala
// F for formula
// I for ID of vars
// E for conversion environment
// (can get rid of I and E if we're always going to use KindedTypeVarSyms)
trait BoolAlgTrait[F, I, E] {
  def mkTrue: F

  def mkFalse: F

  def mkAnd(f1: F, f2: F): F

  def mkOr(f1: F, f2: F): F

  def mkNot(f1: F): F

  def map(g: I => F, f: F): F

  def getEnv(fs: List[F]): E

  def toType(f: F, env: E): Type

  def fromType(t: Type, env: E): F

  // and here we add
  def toFormulaForMinimization(f: F): (FormulaForMinimization, SomeEnv)

  def fromFormulaForMinimization(f: FormulaForMinimization, env: SomeEnv): F
}
```

magnus-madsen:
- I think the trait looks OK, but I would drop the I and E parameters. Variables should just be integers, I think.
- We probably need a `freeVars`.
- I think formula can be renamed to `BoolFormula` and *can* implement the interface. 
- What if we just add a `minimize` to the trait? I think that would be useful, more than conversions.

magnus-madsen:
- I think fromType can be implemented on the trait.

mlutze:
Ok, ready for another cursory look.

Names are all still terrible and confusing, but to summarize:
1. `BoolAlgTrait` is the type class for two-valued boolean algebra-like things
2. `BinaryBoolAlgebra` is isomorphic to `BoolAlgebra` (used for minimization) but uses names like `True` and `And` instead of `Top` and `Join`. It has an instance of `BoolAlgTrait`
3. `BoolUnification2` uses SVE to unify instances of `BoolAlgTrait`
4. `BoolAlgebraSubstitution` manages subsitutions on instances of `BoolAlgTrait`
5. `BoolTypes` holds the smart constructors for booleans
6. `BoolUnification` delegates to `BoolUnification2`
7. `BoolTable` now checks whether the type variables in question are a subset of `{0, 1, 2, 3}` instead of counting them, since there's no guarantee that a formula of `n` variables will contain no variable `>= n`
8. We now have a duplication of the smart constructors: one for formulas and one for proper types.

magnus-madsen:
> Ok, ready for another cursory look.
> 
> Names are all still terrible and confusing, but to summarize:
> 
>     1. `BoolAlgTrait` is the type class for two-valued boolean algebra-like things

Looks good. But I would like this to be just `BoolAlgbra`.

>     2. `BinaryBoolAlgebra` is isomorphic to `BoolAlgebra` (used for minimization) but uses names like `True` and `And` instead of `Top` and `Join`. It has an instance of `BoolAlgTrait`

What's with this duplication?

> 
>     3. `BoolUnification2` uses SVE to unify instances of `BoolAlgTrait`

And this duplication?
> 
>     4. `BoolAlgebraSubstitution` manages subsitutions on instances of `BoolAlgTrait`

Looks good.

> 
>     5. `BoolTypes` holds the smart constructors for booleans

I think they should go where the types are declared.

> 
>     6. `BoolUnification` delegates to `BoolUnification2`

Why this duplication? :D :D :D 

> 
>     7. `BoolTable` now checks whether the type variables in question are a subset of `{0, 1, 2, 3}` instead of counting them, since there's no guarantee that a formula of `n` variables will contain no variable `>= n`

What matters is the number of variables, not their numbers (???)

> 
>     8. We now have a duplication of the smart constructors: one for formulas and one for proper types.

:O

magnus-madsen:
Perhaps we should break this into smaller pieces? We could start with the two traits?

mlutze:
> > ```
> > 2. `BinaryBoolAlgebra` is isomorphic to `BoolAlgebra` (used for minimization) but uses names like `True` and `And` instead of `Top` and `Join`. It has an instance of `BoolAlgTrait`
> > ```
> 
> What's with this duplication?
> 

This duplication is because the original `BoolAlgebra` is used for both kinds of effects, for the purposes of minimization. We have `BoolEffect -> BoolAlgebra` and `SetEffect -> BoolAlgebra`.

> > ```
> > 3. `BoolUnification2` uses SVE to unify instances of `BoolAlgTrait`
> > ```
> 
> And this duplication?
>

`BoolUnification2` operates only on formulas, whereas `BoolUnification` operates on `Type`s

> > ```
> > 5. `BoolTypes` holds the smart constructors for booleans
> > ```
> 
> I think they should go where the types are declared.
>

We currently have two different smart constructors. The `BoolTypes` ones are much more aggressive than the ones in `Type` Should we use those everywhere?

> > ```
> > 6. `BoolUnification` delegates to `BoolUnification2`
> > ```
> 
> Why this duplication? :D :D :D
> 

See 3

> > ```
> > 7. `BoolTable` now checks whether the type variables in question are a subset of `{0, 1, 2, 3}` instead of counting them, since there's no guarantee that a formula of `n` variables will contain no variable `>= n`
> > ```
> 
> What matters is the number of variables, not their numbers (???)
> 

The reduction function requires the variables to be named appropriately, with no numbers `>= 4`. Otherwise renaming has to be done.



We can discuss this all later today.

mlutze:
Summary from discussion:

- trait is called `BoolFormula`
- only one explicit BoolFormula called `ExplicitFormula`
- Sets using `ExplicitFormula` for minimization is a hack: we don't need generic names

mlutze:
Outstanding issues:
- how to add `isSatisfiable` to the class when it depends on SVE
- what to do about variable order
- what to do about `BoolTypes`

magnus-madsen:
> Outstanding issues:
> 
>     * how to add `isSatisfiable` to the class when it depends on SVE

Let's do it in a separate PR.

>     * what to do about variable order

Separate PR.

>     * what to do about `BoolTypes`

This is for types? Where are these smart constructors used on _types_?

mlutze:
> This is for types? Where are these smart constructors used on _types_?

image

magnus-madsen:
> > This is for types? Where are these smart constructors used on _types_?

What about having them just call `Type.mkXYZ` (which does not do more than rudimentary simplification)?

main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
```diff
@@ -1212,7 +1212,7 @@ object Kinder {
             // Case 2: There is no necessarily impure effect. Build the formula
             case None =>
               purs.reduceOption({
-                case (t1, t2) => BoolUnification.mkAnd(t1, t2)
+                case (t1, t2) => BoolTypes.mkAnd(t1, t2)
```
magnus-madsen, line 1215: Type.mkAnd

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -951,10 +951,10 @@ object Typer {
               acc
             case (acc, ((isAbsentVar, _), KindedAst.ChoicePattern.Present(_, _, _))) =>
               // Case 2: A `Present` pattern forces the `isAbsentVar` to be equal to `false`.
-              BoolUnification.mkAnd(acc, Type.mkEquiv(isAbsentVar, Type.False, loc))
+              BoolTypes.mkAnd(acc, Type.mkEquiv(isAbsentVar, Type.False, loc))
```
magnus-madsen, line 954: Type.mkAnd

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -951,10 +951,10 @@ object Typer {
...
             case (acc, ((_, isPresentVar), KindedAst.ChoicePattern.Absent(_))) =>
               // Case 3: An `Absent` pattern forces the `isPresentVar` to be equal to `false`.
-              BoolUnification.mkAnd(acc, Type.mkEquiv(isPresentVar, Type.False, loc))
+              BoolTypes.mkAnd(acc, Type.mkEquiv(isPresentVar, Type.False, loc))
```
magnus-madsen, line 957: Type.mkAnd

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -971,18 +971,18 @@ object Typer {
               acc
             case (acc, ((isAbsentVar, _), KindedAst.ChoicePattern.Absent(_))) =>
               // Case 2: An `Absent` pattern may match if the `isAbsentVar` is `true`.
-              BoolUnification.mkAnd(acc, isAbsentVar)
+              BoolTypes.mkAnd(acc, isAbsentVar)
```
magnus-madsen, line 974: Type.mkAnd

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -971,18 +971,18 @@ object Typer {
...
             case (acc, ((_, isPresentVar), KindedAst.ChoicePattern.Present(_, _, _))) =>
               // Case 3: A `Present` pattern may match if the `isPresentVar` is `true`.
-              BoolUnification.mkAnd(acc, isPresentVar)
+              BoolTypes.mkAnd(acc, isPresentVar)
```
magnus-madsen, line 977: Type.mkAnd

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -971,18 +971,18 @@ object Typer {
...
           }
 
         /**
           * Constructs a disjunction of the constraints of each choice rule.
           */
         def mkOuterDisj(m: List[List[KindedAst.ChoicePattern]], isAbsentVars: List[Type.Var], isPresentVars: List[Type.Var]): Type =
           m.foldLeft(Type.False) {
-            case (acc, rule) => BoolUnification.mkOr(acc, mkUnderApprox(isAbsentVars, isPresentVars, rule))
+            case (acc, rule) => BoolTypes.mkOr(acc, mkUnderApprox(isAbsentVars, isPresentVars, rule))
```
magnus-madsen, line 985: Types.mkXYZ

main/src/ca/uwaterloo/flix/language/phase/unification/BoolFormula.scala
```diff
@@ -0,0 +1,90 @@
...
+import ca.uwaterloo.flix.util.InternalCompilerException
+import ca.uwaterloo.flix.util.collection.Bimap
+
+import scala.collection.immutable.SortedSet
+
+/**
+  * A type class for Boolean Formulas.
+  */
+trait BoolFormula[F] {
+  def mkTrue: F
```
magnus-madsen, line 28: Add comments.

Also, I would like the order to be: true, false, var, not, or, and.

main/src/ca/uwaterloo/flix/language/phase/unification/BoolFormula.scala
```diff
@@ -0,0 +1,90 @@
...
+
+  def mkNot(f1: F): F
+
+  def mkVar(id: Int): F
+
+  /**
+    * Maps the function over the formula.
+    * The function is executed for each variable in the formula.
+    */
+  def map(formula: F)(f: Int => F): F
```
magnus-madsen, line 44: Applies the function `fn` to every variable occurrences in the formula `f`.

mlutze:
> > > This is for types? Where are these smart constructors used on _types_?
> 
> What about having them just call `Type.mkXYZ` (which does not do more than rudimentary simplification)?

We can try, but I think these were changed to the more complex simplification because the basic simplification caused explosions.

main/src/ca/uwaterloo/flix/language/phase/unification/BoolFormula.scala
```diff
@@ -0,0 +1,90 @@
...
+
+  /**
+    * Returns an environment built from the given types.
+    *
+    * This environment is used to convert between formulas and types.
+    */
```
magnus-madsen, line 50: I don't understand the comment.

main/src/ca/uwaterloo/flix/language/phase/unification/BoolFormula.scala
```diff
@@ -0,0 +1,90 @@
...
+  def getEnv(fs: List[Type]): Bimap[Symbol.KindedTypeVarSym, Int]
+
+  /**
+    * Converts the given formula into a type.
```
magnus-madsen, line 54: given formula `f`.

main/src/ca/uwaterloo/flix/language/phase/unification/BoolFormula.scala
```diff
@@ -0,0 +1,90 @@
...
+    * Returns a rigidity environment on formulas that is equivalent to the given one on types.
+    */
+  def liftRigidityEnv(renv: RigidityEnv, env: Bimap[Symbol.KindedTypeVarSym, Int]): SortedSet[Int] = {
+    renv.s.flatMap(env.getForward)
+  }
+
+  /**
+    * Returns the set of free variables in the formula.
+    */
+  def freeVars(f: F): SortedSet[Int]
```
magnus-madsen, line 84: This seems like an important op, so I would put it before map

main/src/ca/uwaterloo/flix/language/phase/unification/BoolFormulaSubstitution.scala
```diff
@@ -0,0 +1,135 @@
...
+import ca.uwaterloo.flix.util.collection.Bimap
+
+/**
+  * Companion object for the [[BoolFormulaSubstitution]] class.
+  */
+object BoolFormulaSubstitution {
+  /**
+    * Returns the empty substitution.
+    */
+  def empty[F]: BoolFormulaSubstitution[F] = BoolFormulaSubstitution(Map.empty)
```
magnus-madsen, line 29: val, for performance.

main/src/ca/uwaterloo/flix/language/phase/unification/BoolFormulaSubstitution.scala
```diff
@@ -0,0 +1,135 @@
...
+
+  /**
+    * Returns `true` if `this` is the empty substitution.
+    */
+  val isEmpty: Boolean = m.isEmpty
+
+  /**
+    * Converts this formula substitution into a type substitution
+    */
+  def toTypeSubstitution(env: Bimap[Symbol.KindedTypeVarSym, Int])(implicit alg: BoolFormula[F]): Substitution = {
```
magnus-madsen, line 57: I would put this function last in the class.

main/src/ca/uwaterloo/flix/language/phase/unification/BoolFormulaSubstitution.scala
```diff
@@ -0,0 +1,135 @@
...
+      BoolFormulaSubstitution(
+        this.m ++ that.m.filter(kv => !this.m.contains(kv._1))
+      )
+    }
+  }
+
+  /**
+    * Returns the composition of `this` substitution with `that` substitution.
+    */
+  def @@(that: BoolFormulaSubstitution[F])(implicit alg: BoolFormula[F]): BoolFormulaSubstitution[F] = {
```
magnus-madsen, line 104: since there is no coherence, should the implicit boolformula actually be an implicit of the environment and not of the function?

magnus-madsen: I guess that does not work with empty.. OK.

main/src/ca/uwaterloo/flix/language/phase/unification/BoolFormulaTable.scala
```diff
@@ -66,19 +71,19 @@ object BoolAlgebraTable {
...
     * Returns the same formula or a smaller formula.
     */
-  def minimizeFormula(f: BoolAlgebra): BoolAlgebra = {
+  def minimizeFormula(f: ExplicitFormula): ExplicitFormula = {
     // Compute the number of free variables.
     val numVars = f.freeVars.size
 
     // Determine whether to minimize once or recursively.
-    val result = if (numVars <= MaxVars) {
+    val result = if (f.freeVars.subsetOf(VarSet)) {
```
magnus-madsen, line 86: I still don't get this :)

main/src/ca/uwaterloo/flix/language/phase/unification/BoolTypes.scala
```diff
@@ -0,0 +1,235 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package ca.uwaterloo.flix.language.phase.unification
+
+import ca.uwaterloo.flix.language.ast._
+
+import scala.annotation.tailrec
+
+object BoolTypes {
```
magnus-madsen, line 22: I am hoping this can be gone.

magnus-madsen: Replaced by Type.mkXYZ

main/src/ca/uwaterloo/flix/language/phase/unification/FormulaUnification.scala
```diff
@@ -0,0 +1,138 @@
...
+        case BooleanUnificationException => false
+      }
+    }
+  }
+
+
+  /**
+    * To unify two Boolean formulas p and q it suffices to unify t = (p ∧ ¬q) ∨ (¬p ∧ q) and check t = 0.
+    */
+  private def mkEq[F](p: F, q: F)(implicit alg: BoolFormula[F]): F = alg.mkOr(alg.mkAnd(p, alg.mkNot(q)), alg.mkAnd(alg.mkNot(p), q))
```
magnus-madsen, line 137: Lets move into BoolFormula.

magnus-madsen: It can have a default impl liek the above.

main/src/ca/uwaterloo/flix/language/phase/unification/Substitution.scala
```diff
@@ -65,9 +65,9 @@ case class Substitution(m: Map[Symbol.TypeVarSym, Type]) {
           val y = visit(t2)
           visit(t1) match {
             // Simplify boolean equations.
-            case Type.Cst(TypeConstructor.Not, _) => BoolUnification.mkNot(y)
-            case Type.Apply(Type.Cst(TypeConstructor.And, _), x, _) => BoolUnification.mkAnd(x, y)
-            case Type.Apply(Type.Cst(TypeConstructor.Or, _), x, _) => BoolUnification.mkOr(x, y)
+            case Type.Cst(TypeConstructor.Not, _) => BoolTypes.mkNot(y)
+            case Type.Apply(Type.Cst(TypeConstructor.And, _), x, _) => BoolTypes.mkAnd(x, y)
```
magnus-madsen, line 69: Can we use Type.mkXYZ?

main/src/ca/uwaterloo/flix/language/phase/unification/Unification.scala
```diff
@@ -367,15 +367,15 @@ object Unification {
 
       case TypeConstructor.Not =>
         val List(t) = tpe.typeArguments
-        BoolUnification.mkNot(purify(tvar, t))
+        BoolTypes.mkNot(purify(tvar, t))
```
magnus-madsen, line 370: Type.mkXYZ?

magnus-madsen:
I guess we could merge before changing to use Type.mkXYZ? 
Then just fix the other things.

main/src/ca/uwaterloo/flix/language/phase/unification/BoolFormulaTable.scala
mlutze: The formula table expects that the variables in the formula have the numbers 0, 1, 2, 3
Previously, it was true by construction that if you had 4 variables in your formula, then they must be 0, 1, 2, 3
This is no longer true, due to the way Type -> BoolFormula works. Therefore, we need to perform alpha conversion unless we happen to be in the nice world of 0, 1, 2, 3.

That said, I need to revisit this since needing to alpha convert != needing to minimize recursively

magnus-madsen: But why does Type -> BoolFormula not name them 0 to n?

mlutze: It does. But say we have 10 variables total. We do transformations in the SVE and then we have to minimize a type with vars 1, 3, 5, 9. So we have to do alpha-renaming.

In the old implementation, we do Type -> BoolFormula at this point that includes the mapping 1, 3, 5, 9 -> 0, 1, 2, 3

In the new implementation, we already have a Boolean formula at this point, so we need an extra alpha-renaming step.

main/src/ca/uwaterloo/flix/language/phase/unification/BoolFormulaSubstitution.scala
mlutze: Can you do a polymorphic val?

main/src/ca/uwaterloo/flix/language/phase/unification/BoolFormulaTable.scala
magnus-madsen: I see OK.

magnus-madsen: Could you add some comments.

main/src/ca/uwaterloo/flix/language/phase/unification/BoolFormulaSubstitution.scala
magnus-madsen: Like Nil?

mlutze: Scala doesn't like it :D

magnus-madsen: Hmm okay then :)

magnus-madsen: val empty: BoolFormula[Nothing] = ... ??

mlutze:
> RFM?

Ready, unless `val empty` is a blocker.

PR #4548, opened 2022-09-10
refactor: move simple cli use to new file

mlutze:
related to #3720 

This moves the default (non-package) functionality to a new file, making the `main` function simpler. Should make #4534 easier to implement.

### DIFF
--- a/main/src/ca/uwaterloo/flix/Main.scala
+++ b/main/src/ca/uwaterloo/flix/Main.scala
@@ -115,7 +115,8 @@ object Main {
     try {
       cmdOpts.command match {
         case Command.None =>
-        // nop, continue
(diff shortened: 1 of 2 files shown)


magnus-madsen:
Does moving this into a separate file really simplify things? Or just make it even harder to understand how `Main` works?

mlutze:
This is just one step along the path. The significant difference here is that we go from a big match statement where some cases fall through, to a big match where each case dispatches to some other function.

In the longer term, this function will be simplified as we get rid of a lot of those if-statements, since those flags will return into commands.

mlutze:
My intent is for this new file to be a collection of the commands that do not use the packager, similarly to how we have `Packager.scala` for the packager commands and `Benchmarker.scala` for benchmark "commands". This will have at least `run` and `doc` functions for these commands.

PR #4549, opened 2022-09-10
feat: add new example using datalog

### LINKED ISSUE #4513: Add Example 13 from Answer Set Programming: A Primer
I think this example would be nice to have in Flix.

ninaandrup:
Closes #4513


examples/railroad-network.flix
```diff
@@ -0,0 +1,79 @@
...
+        // Indirectly linked stations.
+        Connected(a, b) :- Linked(a, b).
+        Connected(a, b) :- Connected(a, c), Linked(c, b).
+
+        // There exist som station x that is a cutpoint between a and b.
+        Cutpoint(x, a, b) :- Connected(a, b), Station(x), not Circumvent(x, a, b).
+
+        // ???
+        Circumvent(x, a, b) :- Linked(a, b), if x != a, Station(x), if x != b.
+        Circumvent(x, a, b) :- Circumvent(x, a, c), Circumvent(x, c, b).
```
ninaandrup, line 21: I'm not sure how to describe this set of facts.

JonathanStarup: `(x, a, b)` describes that `a` can reach `b` without using `x`.

examples/railroad-network.flix
```diff
@@ -0,0 +1,79 @@
...
+
+        // Check the cutpoint is neither station a nor b.
+        HasICutPoint(a, b) :- Cutpoint(x, a, b), if x != a, if x != b.
+        // Stations are safely connected if there is no cutpoints.
+        SafelyConnected(a, b) :- Connected(a, b), not HasICutPoint(a, b).
+        // The set of stations not safely connected are unsafely connected.
+        UnsafelyConnected(a, b) :- Station(a), Station(b), not SafelyConnected(a, b).
+        // All stations.
+        Station(x) :- Linked(x, _).
```
JonathanStarup, line 30: We also need `Station(src).` as an additional fact. I might not actually be needed but I think its clearer if we just include it.

examples/railroad-network.flix
```diff
@@ -0,0 +1,79 @@
...
+    };
+    let model = solve db <+> pr;
+    let safe = query model select b from SafelyConnected(src, b);
+    let unsafe = query model select b from UnsafelyConnected(src, b);
+    {safe = safe, unsafe = unsafe}
+
+// Define some example graph to test the function `safeConnects`.
```
JonathanStarup, line 37: `//` should be `///`.

Wouldn't the graph comment be better suited as a comment rather than an unused string?

examples/railroad-network.flix
```diff
@@ -0,0 +1,79 @@
+// Define a function that decides whether pairs of railroad stations in `links` are safely connected.
```
JonathanStarup, line 1: `//` should be `///`

examples/railroad-network.flix
```diff
@@ -0,0 +1,79 @@
...
+        ("icsi" , "mamuk" ),
+        ("mamuk", "quincy"),
+        ("mamuk", "clote" ),
+        ("mamuk", "quater"),
+        ("mamuk", "dalte" ),
+        ("clote", "quincy"),
+        ("dalte", "quater")
+    }
+
+// Test which of the stations are safely and unsafely connected from `bis`.
```
JonathanStarup, line 75: `//` should be `///`

examples/railroad-network.flix
```diff
@@ -0,0 +1,79 @@
...
+def test(): Unit \ IO =
+    let rec = safeConnects("bis", exampleGraph());
+    rec.safe |> List.sort |> List.map(Console.green) |> List.foreach(println);
```
JonathanStarup, line 78: Maybe there should be a "header" - like `println("Safe:")` before safe and similarly `"Unsafe:"` before the unsafe ones.

JonathanStarup (changes requested):

Issue #4541, opened 2022-09-07
UnkindedType followup


mlutze:
- [ ] move `Resolver.mkXYZ` to `UnkindedType.mkXYZ`
- [ ] format UnkindedType
- [ ] remove that TODO

magnus-madsen:
"Remove that TODO"-- I love it

magnus-madsen:
Ah-- the passage of time.

PR #4550, opened 2022-09-11
feat: use bool and eff deactivation options for type formatting

mlutze:
related to #4445 

(incidentally fixes part of #4541)


mlutze:
Still need to add a couple tests

PR #4569, opened 2022-09-15
refactor: restructure Boolean Unification

magnus-madsen:


main/src/ca/uwaterloo/flix/language/phase/unification/BoolFormula.scala
```diff
@@ -25,59 +25,89 @@ import scala.collection.immutable.SortedSet
...
 
   /**
-    * Returns the formula representing f1 ∧ f2.
+    * Returns a representation of the conjunction of `f1` and `f2`.
     */
   def mkAnd(f1: F, f2: F): F
 
   /**
-    * Returns the formula f such that f == False <=> f1 == f2.
+    * Returns a representation of the formula `f1 == f2`.
```
mlutze, line 70: But this is backwards from what you would expect. Because if they are equal then it returns false.

magnus-madsen: What do you mean? I thought:

```
x == y
x <=> y
x => y and y => x
(not x or y) and (y or notx)
```

mlutze: Hmmmm I guess I was confused.

PR #4575, opened 2022-09-17
fix: disallow for-loops only containing guard fragments

### LINKED ISSUE #4570: Fix foreach / for-yield fragment parsing
Currently allows `foreach (x > 2) exp` i.e. only a guard.

jaschdoc:
Closes #4570

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/ParsedAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/ParsedAst.scala
@@ -31,10 +31,10 @@ object ParsedAst {
     *
     * A collection of imports and declarations.
     *
-    * @param sp1   the position of the first character in the source.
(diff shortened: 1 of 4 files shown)


magnus-madsen:
The solution seems pretty complex. What's the advantage? You want to disallow this:

```
def main(): Unit \ IO = 
    let x = 1232;
    foreach (if x >= 2) 
        println("Hello World!")
```

but is that really bad?

magnus-madsen:
Is there a more harmful example?

jaschdoc:
No, but it seems like it could lead to awkward bugs where a loop could be used as an if statement but where it wasn't intentional. Arguably, that would be undefined behavior.

magnus-madsen:
> No, but it seems like it could lead to awkward bugs where a loop is used as an if expression but where it wasn't intentional. Arguably, that would be undefined behavior.

Do you have an example?

PR #4578, opened 2022-09-18
feat: Set release version at build time

helpermethod:
This change allows setting the release version at build time via the `version` property. If no version is set, a `dev` version (0.0.0) is assumed.

This step is a prerequisite in creating a release pipeline.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -57,8 +57,11 @@ sourceSets {
 }
 
 jar {
+    inputs.property("releaseVersion", findProperty("releaseVersion"))
+
     manifest {
         attributes 'Main-Class': 'ca.uwaterloo.flix.Main'
+        attributes 'Implementation-Version': inputs.properties['releaseVersion'] ?: ''
     }
 
     from {
@@ -66,7 +69,7 @@ jar {
         configurations.runtimeClasspath.collect { it.isDirectory() ? it : zipTree(it) }
         configurations.compileClasspath.collect { it.isDirectory() ? it : zipTree(it) }
     }
-    
+
     from('main') {
         include '**/*.flix'
         include '**/*.zip'
--- a/main/src/ca/uwaterloo/flix/api/Version.scala
+++ b/main/src/ca/uwaterloo/flix/api/Version.scala
@@ -20,7 +20,11 @@ object Version {
   /**
     * Represents the current version of Flix.
     */
-  val CurrentVersion: Version = Version(major = 0, minor = 31, revision = 0)
+  val CurrentVersion: Version = {
+    val version = Option(getClass).map(_.getPackage).map(_.getImplementationVersion).getOrElse("0.0.0")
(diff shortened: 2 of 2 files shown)


magnus-madsen:
Can we make it so that if no version is set the entire build fails? I think that would be much safer?

helpermethod:
Hi @magnus-madsen, I've made the `releaseVersion` an required input to the JAR task, so it's not possible anymore to trigger the task without providing a `releaseVersion`.

    ./gradlew jar -PreleaseVersion=1.2.3

works, whereas 

    ./gradlew jar

will complain that the `releaseVersion` is missing

```
❯ ./gradlew jar
> Task :jar FAILED

FAILURE: Build failed with an exception.

* What went wrong:
A problem was found with the configuration of task ':jar' (type 'Jar').
  - In plugin 'org.gradle.api.plugins.BasePlugin' type 'org.gradle.api.tasks.bundling.Jar' property 'releaseVersion' doesn't have a configured value.
    
    Reason: This property isn't marked as optional and no value has been configured.
    
    Possible solutions:
      1. Assign a value to 'releaseVersion'.
      2. Mark property 'releaseVersion' as optional.
```

magnus-madsen:
This seems reasonable. We use the `jar` command alot (for CI and local testing). Perhaps we should keep it and add a `release` command that works as you describe?

magnus-madsen:
Is that even possible with Gradle?

(EDIT: Also, could you add yourself to `AUTHORS.md` or state here you agree to open source your contributions under Apache 2.0?)

helpermethod:
Sounds good! I've just had a look at the JAR on demand workflow, and this wouldn't quite work with the changes to the jar command ;).

Yes, it's possible to create another task named release of type JAR. This would indeed be the cleaner solution.

Is it okay if the JAR on demand uses some default version number (like 0.0.0)? It's not a release after all :).

As an alternative we could write the version into a file and update that during the build. Wdyt?

magnus-madsen:
> Is it okay if the JAR on demand uses some default version number (like 0.0.0)? It's not a release after all :).

Yes, but `999.999.999` might be better (because often we build from master which is newer than whatever the current release is)

magnus-madsen:
We could have: `0.0.0` -- something is broken, `999.999.999` build by jar command.

mlutze:
We could do a prerelease version based on the latest version plus a git SHA: like `0.31.0-35BAF134B5`

helpermethod:
@mlutze Good idea, I was thinking of something similar.

Honestly I think that having the version in a VERSION file committed to the repository is a better solution than reading it from a property.

That way building the JAR on demand and building the JAR locally would still work without any changes.

Was a little busy the last few days but I hope to show something by the end of the day 😄 .

magnus-madsen:
I think for now we are most comfortable with managing our releases manually. 

That does not mean we do not want to ship Flix with homebrew etc. etc.

PR #4594, opened 2022-09-20
fix: add error for wildcard types in illegal positions

### LINKED ISSUE #4560: Disallow wildcards in type alias and enum definitions
What is the use-case for this?

```scala
type alias TA = Map[_, _]

enum E {
    case C(_)
}
```

This does not seem meaningful and complicates the typing as we cannot assume the set of type variables in the alias definition or enum case is a subset of those in the declaration.

mlutze:
fixes #4560


magnus-madsen:
What's the idea here: Its a syntactic check of where `_` are allowed in types?

mlutze:
> What's the idea here: Its a syntactic check of where `_` are allowed in types?

Yes. Because in some cases we assume all type variables are explicit, but with wildcard variables they are implicit.

PR #4607, opened 2022-09-23
fix: make Xno-optimizer work

### LINKED ISSUE #4604: --Xno-optimizer has no impact
because missing:

```
xnooptimizer = cmdOpts.xnooptimizer,
```

mlutze:
fixes #4604


main/src/ca/uwaterloo/flix/Main.scala
```diff
@@ -90,13 +90,19 @@ object Main {
...
       xnoseteffects = cmdOpts.xnoseteffects,
       xnobooleffects = cmdOpts.xnobooleffects,
-      xvirtualthreads = cmdOpts.xvirtualthreads
+      xvirtualthreads = cmdOpts.xvirtualthreads,
+      xnooptimizer = cmdOpts.xnooptimizer,
+      incremental = Options.Default.incremental,
+      target = Options.Default.target,
+      test = Options.Default.test,
+      loadClassFiles = Options.Default.loadClassFiles,
+      xallowredundancies = Options.Default.xallowredundancies
```
magnus-madsen, line 105: What's the point of having default if we are going to do this?

I think either:

- We remove default (forcing you always to choose)
- Or we keep the current system where we start with Options.Default.copy

mlutze: `Default` organizes the defaults well and is used as the basis for `DefaultForTesting` etc.

magnus-madsen: Yes, but still, whats the point of not using it here?

mlutze: So that we get a compilation error if we add an option and forget to update it.

More generally, I think there are two cases we are concerned about:
1. We are specifying almost all of the options (as in this case)
2. We want the defaults except a particular option (as in the `DefaultTest` case)

For 1. it's not appropriate to use `Default` due to lack of compilation protections, but for 2. it's more appropriate since it's less likely that a change to the API would necessitate a change in `DefaultTest`.

magnus-madsen: I am not entirely convinced, but I will take it.

magnus-madsen: Can I then be pedantic and ask you to keep them in the same order as in options? :D

PR #4608, opened 2022-09-23
feat: Track import environment within Namer

paulbutcher:
Follow-on from #4538 and as per #4392

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Namer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Namer.scala
@@ -2065,6 +2065,9 @@ object Namer {
     def addTag(s: String, n: Name.QName, t: Name.Ident): UseEnv = copy(tags = tags + (s -> (n, t)))
   }
 
+  /**
+    * Merges the given `imports` into the given import environment `ienv0`.
+    */
   private def mergeImportEnvs(imports: List[WeededAst.Import], ienv0: ImportEnv): Validation[ImportEnv, NameError] = {
     Validation.fold(imports, ienv0) {
       case (ienv1, WeededAst.Import.Import(name, alias, loc)) =>
@@ -2075,5 +2078,8 @@ object Namer {
     }
   }
 
+  /**
+    * Represents an environment of "imported" Java classes or interfaces
+    */
   private type ImportEnv = Map[String, Name.JavaName]
 }
--- a/main/src/ca/uwaterloo/flix/language/phase/Namer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Namer.scala
@@ -55,15 +55,15 @@ object Namer {
 
     // collect all the declarations.
     val declarations = mapN(traverse(program.units.values) {
-      case root => mapN(mergeUseEnvs(root.uses, UseEnv.empty)) {
-        case uenv0 => root.decls.map(d => (uenv0, d))
+      case root => mapN(mergeUseEnvs(root.uses, UseEnv.empty), mergeImportEnvs(root.imports, Map.empty)) {
+        case (uenv0, ienv0) => root.decls.map(d => (uenv0, ienv0, d))
       }
     })(_.flatten)
 
     // fold over the top-level declarations.
     flatMapN(declarations) {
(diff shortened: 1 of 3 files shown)


mlutze:
Before I review, is this up-to-date with respect to discussion in #4392 ?

magnus-madsen:
I think so, but it is also just intended as a small step. (The real action still happens in the Weeder AFAIK)

main/src/ca/uwaterloo/flix/language/phase/Namer.scala
```diff
@@ -2065,4 +2065,21 @@ object Namer {
...
+          case None => (ienv1 + (alias.name -> name)).toSuccess
+          case Some(otherName) => mkDuplicateNamePair(alias.name, loc, SourceLocation.mk(otherName.sp1, otherName.sp2))
+        }
+    }
+  }
+
+  /**
+    * Represents an environment of "imported" Java classes or interfaces
+    */
+  private type ImportEnv = Map[String, Name.JavaName]
```
mlutze, line 2084: @magnus-madsen doesn't usually like type aliases :)

mlutze (approved):

main/src/ca/uwaterloo/flix/language/phase/Namer.scala
magnus-madsen: Indeed. We try to avoid type aliases.

mlutze (approved):

main/src/ca/uwaterloo/flix/language/phase/Namer.scala
```diff
@@ -2065,4 +2065,30 @@ object Namer {
...
+    * Companion object for the [[ImportEnv]] class.
+    */
+  private object ImportEnv {
+    val empty: ImportEnv = ImportEnv(Map.empty)
+  }
+
+  /**
+    * Represents an environment of "imported" Java classes or interfaces
+    */
+  private case class ImportEnv(imports: Map[String, Name.JavaName]) {
```
magnus-madsen, line 2091: Oh no. This is even worse! :)

My problem with this and with type aliases is that we already have "too many types". I want meaningful types.  This is just a wrapper around a meaningful type. There is no reason not to just write `Map[String, Name.JavaName]` other than its a few chars. longer than a type alias.

Flix contains 1,822 case classes (!!!!)

If we add `ImportEnv` or even just a type alias then it raises the bar for a future student/contributor/programmer to understand what is going on. If we have ` Map[String, Name.JavaName]` then they have a chance 👍

magnus-madsen: @mlutze For rationale

paulbutcher: > Oh no. This is even worse! :)

I'm so glad you think so! I hated it while I was writing it, but thought that's what you meant by avoiding type aliases.

I understand your logic, however, consider what comes next. The PR that I have queued up to follow this "plumbs" `ienv` through the namer so that it can be used within the `WeededAst.Type.Ambiguous` case of `visitType`. This works, but it involves adding an `ienv` parameter to just about every single method within the namer. This requires literally hundreds of small changes (every method needs the additional parameter, and every method call).

Right now, a typical example looks like:

```
```

mlutze: Do we have something in the contributing docs about this?

magnus-madsen: I see we already have UseEnv, so your proposal would be consistent.

What does `UseEnv` look like?

paulbutcher: `UseEnv` is what I copied to create `ImportEnv` as it currently exists:

https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/Namer.scala#L2041-L2066

magnus-madsen: Then let's keep it as is. Will merge when at a real computer.

Issue #4609, opened 2022-09-24
Re-design `par` construct


magnus-madsen:
I would like to re-design the `par` construct as follows:

- [x] We keep the `par(exp1, exp2, ... expn)` expression.
  - [x] The compiler should be smart enough that if `exp_i` is a variable or literal, no thread is spawned.
  - [ ] The compiler should only spawn `n - 1` threads.
- [x] We drop the `par exp1(exp2, exp3, ..., expn)` expression (i.e. `ParApply`).
- [x] We add a new expression of the form `par (x <- exp1, y <- exp2, ..., expn) yield exp`. 
  - [ ] The compiler should spawn `n - 1` threads.

I suggest to implement each of these changes in small and incremental PRs.

EDIT: I suggest to start by removing features and tidying up. Then adding new stuff.

magnus-madsen:
@mlutze Acceptable? :)

magnus-madsen:
@jaschdoc Meaningful?

jaschdoc:
Should `x` be a pattern in `par (x <- exp, ...) yield e`?

mlutze:
> Should `x` be a pattern in `par (x <- exp, ...) yield e`?

I vote "yes"

magnus-madsen:
if it can be in `for` then yes

jaschdoc:
I've been toying with the idea of having a `par-select` expression, or just adding a variant of the select expression  that would accommodate the "throw-more-threads-at-the-problem" pattern.
This would look something like
```flix
par (pat <- exp; ... ) select {
    pat => exp
    ...
    pat => exp
}
```
We could potentially also rename this to `select-yield`.

I'm not sure if this would really make sense for pure expressions within the parentheses or it would require or at least allow for a `NonDet` effect.

The only example I can come up with right now, is the case where you want to find prime numbers. Of course, this would be a nondeterministic function, but you could easily imagine, throwing more threads, say 3, at the problem, and then selecting the first case that returns.

The obvious problem is that it's not wildly different from the current select expression that we already have, however, I could see this being more convenient to write but then again, this may only be a niche use case.
The only real upside is that it spares the programmer from writing code that has to set up channels.

Do you have any thoughts on this?

magnus-madsen:
Question: What problem are you trying to solve?

jaschdoc:
[...]

PR #4621, opened 2022-09-27
feat: add `ParYield` expression

jaschdoc:
Related to #4609

- [x] Add tests that check `a <- declare a here` is valid
- [x] Add tests that check nesting is valid
- [x] Add tests that check a pattern cannot be used in another fragment
- [x] Add test that checks re-declaring a variable bound in a pattern in the yield exp is invalid
- [x] Implement missing phases (see changed files)
- [x] Add documentation
- [x] Wait for review


jaschdoc:
@magnus-madsen thoughts on on having two par expressions in the parser?

magnus-madsen:
I think we should have two expressions all the way until Lowering.

jaschdoc:
@magnus-madsen I think I need your opinion on this:
```
par (pat1 <- exp1, pat2 <- exp2, pat3 <- exp3) yield expn
```
is desugared into
```
match exp1 {
case pat1 => match exp2 {
case pat2 => match exp3 {
case pat 3 => expn
}}}
```
in the weeder. This is easier to implement throughout all of the phases and the goal is to adjust this expression in the lowering phase to
```
{
let ch1 = chan 1;
let ch2 = chan 1;
let ch3 = chan 1;
spawn ch1 <- exp1;
spawn ch2 <- exp2;
spawn ch3 <- exp3;
match <-ch1 {
case pat1 => match <-ch2 {
case pat2 => match <-ch3 {
case pat 3 => expn
}}}
} as \ Pure
```
also at a later stage it would not spawn `exp3`. This approach, however, has the downside that if `expn` is a match-case expression, the lowering phase would not be able to distinguish `expn` from the expressions in the par (exps) so it would spawn more threads than specified. Are there any obvious solutions to this or should I do something completely different?

magnus-madsen:
Are we doing too many things at once? I don't think allowing pattern matches is super important.

magnus-madsen:
Has this gone of the rails? Perhaps we could start with just supporting variable patterns?

jaschdoc:
Not necessarily off the rails, just slowly progressing because I wasn't sure what the right course of action was.
I'll refactor to just support variables for now.

magnus-madsen:
What's the status on this? :)

magnus-madsen:
Here I am just interested in the simple solution which allows me to write e.g. `par (a <- 1, b <- 2, c <- 3) yield a + b + c`.

jaschdoc:
I think I had writer's block on this, but luckily I figured out how to handle it :) I'll ping you when I have questions or (hopefully) an initial version is ready.

jaschdoc:
@magnus-madsen I think the main part is ready for review. All tests are passing (locally) but some documentation is stilling missing and a few non-essential phases.
It doesn't yet spawn `n-1` threads but I think we should handle that in a separate PR.

main/src/ca/uwaterloo/flix/language/phase/Namer.scala
```diff
@@ -1074,6 +1074,27 @@ object Namer {
...
+    case WeededAst.Expression.ParYield(frags, exp, loc) =>
+      // Visit patterns and collect an env
+      val (pats, finalEnv) = frags.foldRight((List.empty[NamedAst.Pattern], env0)) {
+        case (WeededAst.ParYield.Fragment(pat, _, _), (pats, envAcc)) =>
+          val (p, env1) = visitPattern(pat, uenv0)
+          (p :: pats, envAcc ++ env1)
+      }
+
+      // Fragments exps are visited in env0 and paired with the previously visited pattern
+      val fragsVal = traverse(frags.zip(pats)) {
```
mlutze, line 1086: I wonder how it would look just to traverse everything once and to have a `var finalEnv` that gets updated. What do you thing?

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
```diff
@@ -1020,6 +1020,17 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
...
     }
 
+    def ParYield: Rule1[ParsedAst.Expression.ParYield] = {
+
+      def Fragment: Rule1[ParsedAst.ParYield.Fragment] = rule {
+        SP ~ Pattern ~ optWS ~ atomic("<-") ~ optWS ~ Expression ~ SP ~> ParsedAst.ParYield.Fragment
+      }
+
+      rule {
+        SP ~ keyword("par") ~ optWS ~ "(" ~ zeroOrMore(Fragment).separatedBy(optWS ~ "," ~ optWS) ~ ")" ~ optWS ~ keyword("yield") ~ WS ~ Expression ~ SP ~> ParsedAst.Expression.ParYield
```
mlutze, line 1030: Should this be `oneOrMore`?

main/src/ca/uwaterloo/flix/language/phase/Redundancy.scala
```diff
@@ -746,6 +787,14 @@ object Redundancy {
       Used.of(sym) ++ visitExp(exp1, env0, rc) ++ visitExp(exp2, env0, rc) ++ visitExp(exp3, env0, rc)
   }
 
+  private def findShadowedVarSyms(freeVars: Set[Symbol.VarSym], env0: Env): Used = {
```
mlutze, line 790: docs

main/src/ca/uwaterloo/flix/language/phase/Redundancy.scala
```diff
@@ -746,6 +787,14 @@ object Redundancy {
...
+    freeVars.map(sym => shadowing(sym, env0)).foldLeft(Used.empty)(_ ++ _)
+  }
+
+  private def findUnusedVarSyms(freeVars: Set[Symbol.VarSym], usedSyms: Used): Set[UnusedVarSym] = {
```
mlutze, line 794: docs

main/src/ca/uwaterloo/flix/language/phase/Redundancy.scala
```diff
@@ -706,6 +706,47 @@ object Redundancy {
...
       visitExp(exp, env0, rc)
 
+    case Expression.ParYield(frags, exp, _, _, _, _) =>
+      val (used, env1, fvs) = frags.foldLeft((Used.empty, env0, Set.empty[Symbol.VarSym])) {
+        case ((usedAcc, envAcc, fvsAcc), ParYield.Fragment(p, e, _)) =>
+          // Find free vars in pattern
+          val fvs = freeVars(p)
+
+          // Extend env
+          val extendedEnv = envAcc ++ fvs
```
mlutze, line 716: Move below `shadowedVars` so Matt doesn't get confused

main/src/ca/uwaterloo/flix/language/phase/Redundancy.scala
```diff
@@ -706,6 +706,47 @@ object Redundancy {
     case Expression.Par(exp, _) =>
       visitExp(exp, env0, rc)
 
+    case Expression.ParYield(frags, exp, _, _, _, _) =>
+      val (used, env1, fvs) = frags.foldLeft((Used.empty, env0, Set.empty[Symbol.VarSym])) {
+        case ((usedAcc, envAcc, fvsAcc), ParYield.Fragment(p, e, _)) =>
```
mlutze, line 711: It might be nice to pull this into a helper. I feel like there's a lot going on here.

main/src/ca/uwaterloo/flix/language/phase/Redundancy.scala
```diff
@@ -706,6 +706,47 @@ object Redundancy {
...
+          (allUsed, extendedEnv, fvsAcc ++ fvs)
+      }
+
+      val usedYield = visitExp(exp, env1, rc)
+      val unusedVarSyms = findUnusedVarSyms(fvs, usedYield)
+      (usedYield -- fvs) ++ unusedVarSyms ++ used
+
+    /*
+
+    val fvs = usedFrags.foldLeft(Set.empty[Symbol.VarSym]) {
```
mlutze, line 739: commented code

main/src/ca/uwaterloo/flix/language/phase/Regions.scala
```diff
@@ -370,6 +370,16 @@ object Regions {
     case Expression.Par(exp, loc) =>
       flatMapN(visitExp(exp))(_ => checkType(exp.tpe, loc))
 
+    case Expression.ParYield(frags, exp, tpe, _, _, loc) =>
+      val fragsVal = traverse(frags) {
+        case ParYield.Fragment(_, e, _) => flatMapN(visitExp(e)) {
```
mlutze, line 375: Maybe I'm confused, but can't this just be `visitExp(e)`?

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -337,6 +337,8 @@ object Simplifier {
       case TypedAst.Expression.Par(_, _) =>
         throw InternalCompilerException(s"Unexpected expression: $exp0.")
 
+      case TypedAst.Expression.ParYield(frags, exp, tpe, pur, eff, loc) => ???
```
mlutze, line 340: Unexpected expression blah blah

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1645,6 +1645,20 @@ object Typer {
...
+        val patterns = frags.map(_.pat)
+        val parExps = frags.map(_.exp)
+        for {
+          (constrs, tpe, pur, eff) <- visitExp(exp)
+          patternTypes <- inferPatterns(patterns, root)
+          (parExpConstrs, parExpTypes, parExpPurs, parExpEffs) <- seqM(parExps map visitExp).map(unzip4)
+          _ <- InferMonad.point(patternTypes.zip(parExpTypes).map(ts => unifyTypeM(List(ts._1, ts._2), loc)))
+          actualPurs = Type.mkAnd(pur :: parExpPurs, loc)
+          actualEffs = Type.mkUnion(eff :: parExpEffs, loc)
+          resultPur <- expectTypeM(expected = Type.Pure, actual = actualPurs, loc)
```
mlutze, line 1658: I think we can get better errors than this by expecting each individual fragment expression to be pure, rather than expecting their `And` to be pure.

Also, do we really need `exp` to be pure?

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1645,6 +1645,20 @@ object Typer {
...
         } yield (constrs, tpe, resultPur, resultEff)
 
+      case KindedAst.Expression.ParYield(frags, exp, loc) =>
+        val patterns = frags.map(_.pat)
+        val parExps = frags.map(_.exp)
+        for {
+          (constrs, tpe, pur, eff) <- visitExp(exp)
+          patternTypes <- inferPatterns(patterns, root)
+          (parExpConstrs, parExpTypes, parExpPurs, parExpEffs) <- seqM(parExps map visitExp).map(unzip4)
+          _ <- InferMonad.point(patternTypes.zip(parExpTypes).map(ts => unifyTypeM(List(ts._1, ts._2), loc)))
```
mlutze, line 1655: This looks weird. Isn't there a `InferMonad.traverse` or something?

(Also, `xyz._1` is evil)

main/src/ca/uwaterloo/flix/language/phase/extra/CodeHinter.scala
```diff
@@ -295,6 +295,8 @@ object CodeHinter {
     case Expression.Par(exp, _) =>
       visitExp(exp)
 
+    case Expression.ParYield(frags, exp, tpe, pur, eff, loc) => ???
```
mlutze, line 298: Seems like it might cause a crash :)

main/test/ca/uwaterloo/flix/language/phase/TestRedundancy.scala
```diff
@@ -1278,4 +1278,47 @@ class TestRedundancy extends FunSuite with TestUtils {
     val result = compile(input, Options.TestWithLibNix)
     expectError[RedundancyError.RedundantUpcast](result)
   }
+
+  test("TestParYield.01") {
```
mlutze, line 1282: Let's also test for unused var and shadowed var where the shadowing is inside a fragment.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1580,6 +1590,32 @@ object Lowering {
     }
   }
 
+  def mkLetMatch(exp: Expression, pat: Pattern, body: Expression): Expression = {
```
mlutze, line 1593: docs

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1580,6 +1590,32 @@ object Lowering {
...
+    val expLoc = exp.loc.asSynthetic
+    val rule = List(MatchRule(pat, Expression.True(pat.loc.asSynthetic), body))
+    val pur = Type.mkAnd(exp.pur, body.pur, expLoc)
+    val eff = Type.mkUnion(exp.eff, body.eff, expLoc)
+    Expression.Match(exp, rule, body.tpe, pur, eff, expLoc)
+  }
+
+  def mkBoundParWaits(patSymExps: List[(Pattern, Symbol.VarSym, Expression)], exp: Expression): Expression =
```
mlutze, line 1601: docs

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1580,6 +1590,32 @@ object Lowering {
...
+    patSymExps.map {
+      case (p, sym, e) =>
+        val loc = e.loc.asSynthetic
+        val chExp = mkChannelExp(sym, e.tpe, loc)
+        (p, mkGetChannel(chExp, e.tpe, Type.Impure, e.eff, loc))
+    }.foldRight(exp) {
+      case ((pat, chan), e) => mkLetMatch(chan, pat, e)
+    }
+
+  def mkParYield(frags: List[(Pattern, Expression, SourceLocation)], exp: Expression, tpe: Type, pur: Type, eff: Type, loc: SourceLocation)(implicit flix: Flix): Expression = {
```
mlutze, line 1611: docs

main/src/ca/uwaterloo/flix/language/ast/KindedAst.scala
```diff
@@ -349,4 +351,10 @@ object KindedAst {
   case class SelectChannelRule(sym: Symbol.VarSym, chan: KindedAst.Expression, exp: KindedAst.Expression)
 
   case class TypeParam(name: Name.Ident, sym: Symbol.KindedTypeVarSym, loc: SourceLocation)
+
+  object ParYield {
```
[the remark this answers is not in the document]
jaschdoc: Sometimes. But since there's no other sub classes I've removed it

main/src/ca/uwaterloo/flix/language/phase/Namer.scala
jaschdoc: I guess we could do that. That saves us a fold. I think I'm just allergic to mutability :joy:

jaschdoc: I think `traverse` does some parallel stuff. If we have mutable var we would need a mutex around it, yeah?
```scala
  /**
    * A fast implementation of traverse.
    */
  private def fastTraverse[T, S, E](xs: Iterable[T])(f: T => Validation[S, E]): Validation[List[S], E] = {
    // Check if the sequence is empty.
    if (xs.isEmpty)
      return Validation.SuccessNil

    // Two mutable arrays to hold the intermediate results.
    val successValues = mutable.ArrayBuffer.empty[S]
    val failureStream = mutable.ArrayBuffer.empty[LazyList[E]]

    // Apply f to each element and collect the results.
    for (x <- xs) {
      f(x) match {
        case Success(v) => successValues += v
        case Failure(e) => failureStream += e
      }
    }

    // Check whether we were successful or not.
    if (failureStream.isEmpty) {
      Success(successValues.toList)
    } else {
      Failure(failureStream.foldLeft(LazyList.empty[E])(_ #::: _))
    }
  }
```

jaschdoc:
Redeclaring a variable from a pattern inside a new fragment is currently legal since each fragment is independent of the other and conceptually nothing "escapes" out of the right hand side of the arrow (i.e. the exp in `pat <- exp`):
`par (a <- 5, b <- let a = 1; a) yield a + b`
Should this be okay?
Note that `par (a <- exp, a <- exp) yield exp` is NOT valid.

mlutze:
> Redeclaring a variable from a pattern inside a new fragment is currently legal since each fragment is independent of the other and conceptually nothing "escapes" out of the right hand side of the arrow (i.e. the exp in `pat <- exp`): `par (a <- 5, b <- let a = 1; a) yield a + b` Should this be okay? Note that `par (a <- exp, a <- exp) yield exp` is NOT valid.
> 
> Note that I've allowed this since it wouldn't make sense if a fragment had dependencies on a previous fragment. This would defeat the purpose of evaluating them in parallel.

Yes, this is all correct.

My concern was a case like `par (a <- let a = ???; a) yield a` where the `a` is shadowing.

main/src/ca/uwaterloo/flix/language/phase/Namer.scala
mlutze: It doesn't look like this is parallel. Just a for-loop, no?

jaschdoc:
> My concern was a case like `par (a <- let a = ???; a) yield a` where the `a` is shadowing.

Ah but there's a test that explicitly allows this. We also allow `let x = { let x = 5; x };`

mlutze:
> > My concern was a case like `par (a <- let a = ???; a) yield a` where the `a` is shadowing.
> 
> Ah but there's a test that explicitly allows this. We also allow `let x = { let x = 5; x };`
> 
> ```
> def main(): Unit \ IO =
>     let x = { let x = 1 ; x };
>     println(x)
> ```

Hmmmmmmm seems fishy :D

jaschdoc:
> > > My concern was a case like `par (a <- let a = ???; a) yield a` where the `a` is shadowing.
> > 
> > 
> > Ah but there's a test that explicitly allows this. We also allow `let x = { let x = 5; x };`
> > ```
> > def main(): Unit \ IO =
> >     let x = { let x = 1 ; x };
> >     println(x)
> > ```
> 
> Hmmmmmmm seems fishy :D

Hehe I see why, but it does make sense. When evaluating the body of the let-binding, `x` is not yet in scope so we can do whatever we want here.

main/src/ca/uwaterloo/flix/language/phase/Namer.scala
jaschdoc: Ahh yes, I totally had it mixed up with something else... oops :eyes:

jaschdoc:
@mlutze ready for another review :)

magnus-madsen:
Ha! There is a button "re-request" review :D

mlutze (commented):
Looks good, just one missing implementation

jaschdoc:
Should be good to go now

mlutze (approved):

jaschdoc:
@magnus-madsen ready for merge

main/src/ca/uwaterloo/flix/language/phase/Namer.scala
```diff
@@ -1074,6 +1074,24 @@ object Namer {
         case e => NamedAst.Expression.Par(e, loc)
       }
 
+    case WeededAst.Expression.ParYield(frags, exp, loc) =>
+      // Create a mutable env that will be updated during traverse
+      var finalEnv = env0
```
magnus-madsen, line 1079: @mlutze is this OK?

mlutze: Yes, I was the one who proposed it :)

Issue #4614, opened 2022-09-25
Measure hashing performance


mlutze:
https://archive.clojure.org/design-wiki/display/design/Better%2Bhashing.html

mlutze:
I've replicated some of the tests from the clojure study. Our numbers are slightly better, but these tests are tailored to the weaknesses in Clojure, not Flix.

<details> 
  <summary>Testing script</summary>

```scala
//
// ======= MAIN RUNNER ==========
//

def main(): Unit \ IO = {
    testMillions();
    testPosPairLists();
    testNegPairLists();
    testSets();
    testMaps();
    ()
}

//
// ========== TESTS ==========
//

def testMillions(): Unit \ IO = {
    describe("Ints in the range [-1_000_000, 1_000_000]");
    let l = List.range(-1_000_000, 1_000_001);
    testHash(l)
}

def testPosPairLists(): Unit \ IO = {
    describe("Two-element lists of integers in the range [0, 999]");
    let ints = List.range(0, 1000);
    let l = { for (x <- ints; y <- ints) yield (x :: y :: Nil) };
    testHash(l)
}

def testNegPairLists(): Unit \ IO = {
    describe("Two-element lists of integers in the range [-500, 499]");
    let ints = List.range(-500, 500);
    let l = { for (x <- ints; y <- ints) yield (x :: y :: Nil) };
    testHash(l)
}

def testSets(): Unit \ IO = {
    describe("Subsets of the set {0, 1, ... 19}");
    let l = Set.subsets(Set.range(0, 20)) |> Set.toList;
    testHash(l)
}

def testMaps(): Unit \ IO = {
    describe("Singleton maps of from int to int in the range [0, 999]");
    let ints = List.range(0, 1000);
    let l = { for (x <- ints; y <- ints) yield Map.singleton(x, y) };
    testHash(l)
}

//
// ========== HELPERS ============
//

def describe(s: String): Unit \ IO = {
    println("===================================================");
    println(s)
}

def testHash(l: List[a]): Unit \ IO with Hash[a] = {
    // total number of objects hashed
    let totalObjs = List.length(l);

    // map of Hash -> number of occurrences
    let hashCount = l |> List.map(Hash.hash) |> toOccMap;

    // total number of distinct hash values
    let totalHashes = Map.size(hashCount);

    // maximum occurrences for a hash value
    let maxColls = Map.valuesOf(hashCount) |> List.maximum |> unsafeGet;

    // average occurrences per hash value
[...]

PR #4625, opened 2022-09-28
feat: improve hashing

mlutze:
related to #4614


mlutze:
I'm wondering whether we could add some unit tests for hash collisions.

PR #4630, opened 2022-09-30
feat: avoid use of 0 for hashing

mlutze:
related to #4614


JonathanStarup:
Isn't 1 also bad? with all this multiplication

PR #4633, opened 2022-10-03
feat: allow named arguments in enums and tuples

### LINKED ISSUE #4622: Named arguments in enums
Should be a fairly easy implementation and would be especially useful with #4553

mlutze:
fixes #4622. I am not sure about supporting it for tuples since it may lead to confusing error messages for classic typos like if(x = 42)


magnus-madsen:
Can you give some examples?

magnus-madsen:
How about all of the above :) But just starting with one or two examples for me!

mlutze:
```scala
enum Person({first = String}, {last = String})

...
let ada = Person(first = "Ada", last = "Lovelace");
let Person(f, l) = ada;
println("My name is ${f.first} ${l.last}")
```

```scala
enum PeopleData({ids = Map[Person, Int32]}, {ages = Map[Person, Int32]})
```

magnus-madsen:
I am OK with this, but would request some more tests and an example.

magnus-madsen:
@mlutze If you add some test cases and an example then we can merge :)

PR #4636, opened 2022-10-03
fix: allow set effects in entry point

### LINKED ISSUE #4635: `NonDet` unusable
Because main doesn't like it

mlutze:
fixes #4635


magnus-madsen:
Hm. Is this the right fix?

mlutze:
> Hm. Is this the right fix?

I can attest that it "works", in that I can run a NonDet entrypoint function. I agree that it's not pretty but I do not know of a better solution under the circumstances. I'd really rather not do additional text-level code generation to get types right in `Shell.scala`

PR #4639, opened 2022-10-04
feat: add 'invalid' to possible test results

### LINKED ISSUE #4287: Crash when putting `@test` on non-unit function
```scala
@test
def testOne(x: Int32): Int32 = x
```

Running this test causes a crash, but we should instead issue a compilation error or test failure

mlutze:
fixes #4287

### DIFF
--- a/main/src/ca/uwaterloo/flix/runtime/CompilationResult.scala
+++ b/main/src/ca/uwaterloo/flix/runtime/CompilationResult.scala
@@ -57,8 +57,19 @@ class CompilationResult(root: Root,
   def getTests: Map[Symbol.DefnSym, TestFn] = {
     defs.collect {
       case (sym, run) if root.defs(sym).ann.isTest =>
-        val skip = root.defs(sym).ann.isSkip
-        (sym -> TestFn(sym, skip, run))
+        val prop = {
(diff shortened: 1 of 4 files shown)


magnus-madsen:
Why not catch this in the type system? And why is it invalid to have a Unit test? (If you execute it for its effect?)

mlutze:
> Why not catch this in the type system?

Per some ticket for a principle on annotations, annotations should not have an effect on compilation (or at least this is how I interpreted it)

> And why is it invalid to have a Unit test? (If you execute it for its effect?)

This is not concerning return type but rather the argument type. If the argument is not a unit then we can't execute the test.

magnus-madsen:
> Per some ticket for a principle on annotations, annotations should not have an effect on compilation (or at least this is how I interpreted it)

That's true, but is it still not much better to have a compile-time check? I mean, the implementation would be shorter and less brittle!

mlutze:
> > Per some ticket for a principle on annotations, annotations should not have an effect on compilation (or at least this is how I interpreted it)
> 
> That's true, but is it still not much better to have a compile-time check? I mean, the implementation would be shorter and less brittle!

I totally agree, I just didn't want to go against principle :)

magnus-madsen:
Let's add a small compiler check and not tell anyone.

PR #4641, opened 2022-10-04
feat: lsp completions for exhaustive match

Daniel-Anker-Hermansen:


magnus-madsen:
Can you post a screenshot of what it currently gives?

magnus-madsen:
!!!jar

magnus-madsen:
It seems to work pretty well.

mlutze:
Is this the exhaustive match generator I've been praying for?

magnus-madsen:
Two things:

- We should decide what enums to include. Probably not all from the library.
- I think we should fill out with temp variables, e.g `_x` and so forth. Because otherwise the program will not compile until complete.

Daniel-Anker-Hermansen:
I was considering making it so only local types and imports would be considered, but I do not know if that is too restrictive.
I think filling out with temp variables is a good idea. I had not thought about that possibility.

magnus-madsen:
> 

Lets do a new version with temp variables and some form of restriction on what types are considered (I leave that to you)

magnus-madsen:
@Daniel-Anker-Hermansen What's the status on this?

Daniel-Anker-Hermansen:
It now does not complete on enums with less than two cases. It also only considers enums withing the same source uri. It also places temp variables.

magnus-madsen:
I see conflicts

magnus-madsen:
!!!jar

magnus-madsen:
A few issues that should be resolved:

- The variables are sometimes named `elem` and other times `_elem`. Stick to one.
- The cases are prefixed by the enum name. I don't think that should be the default.
- It would be nice to suggest all public enums (but give priority to the one in the current file. Enums from other files would be given lower priority).

magnus-madsen:
Thanks!

PR #4643, opened 2022-10-05
fix: avoid using Scala types in bytecode

### LINKED ISSUE #4640: Runtime error: `java.lang.NoClassDefFoundError: scala/math/package$`
```
Exception in thread "main" java.lang.NoClassDefFoundError: scala/math/package$
```

The triggering code is fairly complex so I don't have a minimal example.

mlutze:
fixes #4640


magnus-madsen:
Wonder how this happened in the first place.

PR #4644, opened 2022-10-05
feat: add --Xflexible-regions flag

mlutze:


magnus-madsen:
Does this ensure that no rigid Boolean variables are ever introduced? (Except for with type scheme checking?)

mlutze:
> Does this ensure that no rigid Boolean variables are ever introduced? (Except for with type scheme checking?)

Yes, as far as I can tell.

There are only 3 places `markRigid` is used:
1. here
2. type scheme checking
3. safety checks

Note that activating this flag will cause boolean formula explosions for some reason.

magnus-madsen:
> Note that activating this flag will cause boolean formula explosions for some reason.

To the extent that type checking does not complete?

mlutze:
> > Note that activating this flag will cause boolean formula explosions for some reason.
> 
> To the extent that type checking does not complete?

Correct.

PR #4647, opened 2022-10-06
feat: add `debug` construct

magnus-madsen:


JonathanStarup:
I agree with `dbg` and not `dbg!`

magnus-madsen:
```
def main(): Unit \ IO =
    println(foo())

def foo(): Int32 =
    let x = 123;
    let y = 345;
    debug("x = ${x}, y = ${y}");
    x + y
```

```
[foo.flix:8] x = 123, y = 345
468
```

magnus-madsen:
!!!jar

magnus-madsen:
Give it a spin

magnus-madsen:
Do we want just the file name and not the whole path?

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -294,9 +294,15 @@ object Simplifier {
         val e = visitExp(exp)
         SimplifiedAst.Expression.Force(e, tpe, loc)
 
-      case TypedAst.Expression.Default(_, _) => throw InternalCompilerException(s"Unexpected expression: $exp0.")
+      case TypedAst.Expression.Debug(exp, tpe, pur, eff, loc) =>
+        val e = visitExp(exp)
+        ???
```
magnus-madsen, line 299: Woops

mlutze:
```
def square(x: Int32): Int32 = {
    debug("squaring");
    x * x
}

def foo(): Int32 = {
    let _ = square(123);
    1324
}
```

Doesn't print "squaring" even though I might expect that.

main/src/library/Prelude.flix
```diff
@@ -129,6 +129,17 @@ pub def println(x: a): Unit \ IO with ToString[a] =
...
     x |> ToString.toString |> println(getOut())
 
+///
+/// Prints the given string `x`.
+///
+/// Note: For internal use only. The actual `debug(exp)` expression is parsed and ultimately rewritten to call this function.
+///
+@Internal
+def debug(x: String): Unit \ IO =
+    import static get java.lang.System.out: ##java.io.PrintStream \ IO as getOut;
```
mlutze, line 139: Can we instead print to StdErr?

main/src/ca/uwaterloo/flix/language/phase/Weeder.scala
```diff
@@ -357,7 +357,7 @@ object Weeder {
   /**
     * Performs weeding on the given enum case `c0`.
     */
-  private def visitCase(c0: ParsedAst.Case, enum: Name.Ident)(implicit flix: Flix): WeededAst.Case = c0 match {
+  private def visitCase(c0: ParsedAst.Case, ident: Name.Ident)(implicit flix: Flix): WeededAst.Case = c0 match {
```
mlutze, line 360: Shadows `ParsedAst.Case(...ident...) below, no?

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -2315,6 +2324,13 @@ object Typer {
         val pur = Type.mkAnd(e1.pur, e2.pur, e3.pur, loc)
         val eff = Type.mkUnion(List(e1.eff, e2.eff, e3.eff), loc)
         TypedAst.Expression.ReifyEff(sym, e1, e2, e3, tpe, pur, eff, loc)
+
+      case KindedAst.Expression.Debug(exp, loc) =>
+        val e = visitExp(exp, subst0)
+        val tpe = Type.Unit
+        val pur = Type.Impure
+        val eff = Type.Empty
```
mlutze, line 2332: These should get the `pur` and `eff` from `exp` I think.

main/src/ca/uwaterloo/flix/language/ast/SourceLocation.scala
```diff
@@ -89,10 +89,29 @@ case class SourceLocation(input: Option[ParserInput], source: Source, locationKi
...
+    */
+  def formatWithLine: String = s"${source.name}:$beginLine"
+
+  /**
+    * Returns a string representation of `this` source location with the line and column numbers.
     */
   def format: String = s"${source.name}:$beginLine:$beginCol"
 
+  /**
+    * Returns the source text of the source location.
```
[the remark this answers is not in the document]
magnus-madsen: actually this function is not used.

main/src/ca/uwaterloo/flix/language/ast/SourceLocation.scala
```diff
@@ -89,10 +89,29 @@ case class SourceLocation(input: Option[ParserInput], source: Source, locationKi
...
+    */
+  def text: Option[String] = {
+    if (isMultiLine) {
+      None
+    } else {
+      val line = lineAt(beginLine)
+      val b = Math.min(beginCol - 1, line.length)
+      val e = Math.min(endCol - 1, line.length)
```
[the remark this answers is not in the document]
magnus-madsen: Who knows. I am being careful :)

main/src/ca/uwaterloo/flix/language/phase/Weeder.scala
magnus-madsen: Dang, good catch.

magnus-madsen: But it seems the name is not used?

main/src/library/Prelude.flix
magnus-madsen: Not sure I like that. Whats the pros and cons?

magnus-madsen:
> 

Yes, but that seems unavoidable, sadly. The function is pure and so gets removed.

magnus-madsen:
Its good it does not print squaring! It shows you the function was removed by the Optimizer :)

main/src/library/Prelude.flix
mlutze: Pros: 
1. We don't interfere with StdOut. If I have a Flix program as part of a pipeline and I'm debugging, I don't want debugging information going out with the actual output.
2. Principle: debugging information is not generally primary output from a program

Cons:
1. Depending on where things are flushed, might not print to output when we expect, with respect to other println calls

Incidentally, we should probably add a flush so this occurs right away instead of hanging out in the buffer.

magnus-madsen: I believe println flushes.

Issue #4646, opened 2022-10-06
Add new debugging facilities


magnus-madsen:

mlutze:
I was thinking of introducing an alternative string interpolation like `"the value of myVar is: #{myVar}"` where it calls `mkDebugString(myVar)` or whatever instead of `toString(myVar)`

magnus-madsen:
@JonathanStarup Can you sketch how we can even implement this magic toString? I don't see it myself

magnus-madsen:
@JonathanStarup @mlutze @jaschdoc 

I have merged support for `debug(expr` into `master`. Please try it out! :)

JonathanStarup:
So what parts of debug works now? I thought it would be a keyword, but its both a keyword and a function? 

So missing aspects are this magic toString and optimizer behaviour?

JonathanStarup:
> @JonathanStarup Can you sketch how we can even implement this magic toString? I don't see it myself

My idea was just to call java.lang.object.toString, but then we of course need to know object vs primitive

JonathanStarup:
you mean like this? (i forgot syntax of type case)
```
def debug(x: a): Unit = match reifyType(x) {
    case Obj => $magicToStringObj$(x as ##java.lang.object)
    case _ => ...
} as \ {}
```

mlutze:
Yes, or

> you mean like this? (i forgot syntax of type case)
> 
> ```
> def debug(x: a): Unit = match reifyType(x) {
>     case Obj => $magicToStringObj$(x as ##java.lang.object)
>     case Int32 => $magicToStringInt32$(x as Int32)
>     case _ => ...
> } as \ {}
> ```

Yes, though we don't need magic ToString, right? Just Object.toString and Integer.toString etc.

JonathanStarup:
> though we don't need magic ToString, right? Just Object.toString and Integer.toString etc.

Ah good point, yes, the backend objects just need to implement sensible implementations of toString yeah. So it's actually just
```
def debug(x: a): Unit =
    import java.lang.object.toString(##java.lang.object): String;
    match reifyType(x) {
    case Obj => toString(x as ##java.lang.object)
    case Int32 => Int32.toString(x as Int32)
    case _ => ...
} as \ {}
```

magnus-madsen:
> but a lot of the cases in the match give their own weird runtime/compiletime errors

Can you be more specific? Are we generating the wrong code? 

@mlutze Are we even allowed to use the ToString type class in the above scenario?

[...]

PR #4650, opened 2022-10-08
feat: allow `debug` for all types

JonathanStarup:
Anything can be printed, the string isn't to meaningful for tags, tuples, etc, but that can be improved in the backend. Related to #4646.

### DIFF
--- a/examples/named-arguments.flix
+++ b/examples/named-arguments.flix
@@ -1,0 +1,36 @@
+// When functions have two arguments of the same type,
+// it can be difficult to remember the right order.
+// We can use named arguments to be sure we are applying the function correctly.
+// The parameter type becomes a singleton record,
+// and a syntactic sugar allows us to elide the braces when calling a function with a named argument.
+
+///
+/// Returns true if the `sublist` is contained within the list `l`.
+///
+pub def contains(sublist: {sublist = List[a]}, l: List[a]): Bool with Eq[a] = match (sublist.sublist, l) {
+    case (Nil, _) => true
+    case (_ :: _, Nil) => false
+    case (h1 :: t1, h2 :: t2) if h1 == h2 => contains(sublist = t1, t2) or contains(sublist = sublist.sublist, t2)
+    case (sub, _ :: t2) => contains(sublist = sub, t2)
+}
+
+// We often leave one of the parameters unnamed to allow for piping.
+
+///
+/// Returns true if the list contains my secret code.
+///
+pub def hasSecretCode(l: List[Int32]): Bool = {
+    l |> contains(sublist = 1 :: 2 :: 3 :: 4 :: Nil)
+}
+
+// We can also use named arguments in enums, where fields are otherwise unnamed.
+
+///
+/// Represents a Flixball player.
+///
+pub enum Player({id = Int32}, {score = Int32})
+
+///
+/// Creates a new player from the given ID.
+///
+pub def initializePlayer(id: Int32): Player = Player(id = id, score = 0)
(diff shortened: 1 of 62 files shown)


JonathanStarup:
<details>
<summary> Example and output </summary>

```scala
enum Test(Int32)

def main(): Unit \ IO = {
    let _ = debug("hello");
    let _ = debug(Test(2));
    let _ = debug(());
    let _ = debug(true);
    let _ = debug('a');
    let _ = debug(32.5f32);
    let _ = debug(64.5f64);
    let _ = debug(8i8);
    let _ = debug(16i16);
    let _ = debug(32i32);
    let _ = debug(64i64);
    let _ = debug([1, 2, 3]);
    let _ = debug(x -> x+1);
    let _ = debug((1, 2, "hello"));
    ()
}
```

Output
```
[main\src\personal\test.flix:4] hello
[main\src\personal\test.flix:5] Test
[main\src\personal\test.flix:6] ()
[main\src\personal\test.flix:7] true
[main\src\personal\test.flix:8] a
[main\src\personal\test.flix:9] 32.5
[main\src\personal\test.flix:10] 64.5
[main\src\personal\test.flix:11] 8
[main\src\personal\test.flix:12] 16
[main\src\personal\test.flix:13] 32
[main\src\personal\test.flix:14] 64
[main\src\personal\test.flix:15] [I@25a6944c
[main\src\personal\test.flix:16] Clo%main%175449@5e1fa5b1
[main\src\personal\test.flix:17] Tuple3
```
</details>

JonathanStarup:
!!!jar

magnus-madsen:
- I am OK with two exps in `Debug`
- I am not so OK with all the test cases (this seems to cost 4minutes!)
- I am not sure how this can handle enums (will expand in other ticket).

JonathanStarup:
> I am not so OK with all the test cases (this seems to cost 4minutes!)

It's not so nice. It makes sense that debug uses these toString methods, though, unless we inline them in the prelude.

The other option is to have `$Magic$` in the prelude and let the backend differentiate the types (with the aid of monomorphization)

magnus-madsen:
Isnt e.g. `Float32.toString` just a call to a Java method anyway?

mlutze:
You can improve array output with Java's `Arrays.deepToString`

If we just use Java functions for everything then we can avoid using the full lib in the tests.

magnus-madsen:
I like this solution. I think its a pretty elegant given the complexity of the problem.

One question: Do we want to separate "magic" from the "debug"? Because then you can do:

```
debug("complex string x = ${magic(x))
```

Of course it should not be called magic but something else.

JonathanStarup:
@magnus-madsen enums are "nice" now (multi arity just display the ugly tuple string)
```
enum Test {
    case TestA(Int32)
    case TestB(Test, Test)
}

enum Poly[a, b] {
    case PolyA(a)
    case PolyB(b)
}

def main(): Int32 =
    let _ = debug(TestB(TestA(1), TestA(2)));
    let _ = debug(PolyA(42));
    let _ = debug(PolyB(true));
    42
```

output

```
[main\src\personal\test.flix:12] Test.TestB(Tuple2)
[main\src\personal\test.flix:13] Poly.PolyA(42)
[main\src\personal\test.flix:14] Poly.PolyB(true)
42
```

JonathanStarup:
> Looks good. We could merge and iterate? (See my comments about having two functions.)

It could be done, the current debug function just has to be written in jvm instead

JonathanStarup:
> Looks good. We could merge and iterate? (See my comments about having two functions.)

Here's your magic string function

PR #4652, opened 2022-10-08
feat: Abbreviate instead of hide long inlay hints

paulbutcher:
This modifies inlay hints so that instead of hiding long hints, they're now abbreviated. So instead of this:

Screenshot-2022-10-08-at-16 08 04

We now have this:

Screenshot 2022-10-08 at 16 48 12

A couple of things to perhaps consider before merging this:

* I've left the length of a "long" inlay hint at 14 characters. Other languages [apparently use 30 characters](https://github.com/ionide/ionide-vscode-fsharp/issues/1694), we might want to follow suit?
* I've (almost) copied the `truncate` function from `ProgressBar.scala`, which might not be the right thing to do. Is there somewhere where this kind of utility should live?
* The difference between my copy of `truncate` and the one in `ProgressBar.scala` is that I'm using the Unicode ellipsis character `…` instead of three periods `...`. This strikes me as the right thing to do given that space is at a premium in inlay hints, but I'm open to being persuaded otherwise.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
@@ -53,10 +53,10 @@ object InlayHintProvider {
         val label = ": " + FormatType.formatType(minType)
 
         // Hide long inlay hints.
-        if (isTypeVar(minType) || label.length >= 14)
+        if (isTypeVar(minType))
           None
         else
-          Some(InlayHint(pos, label, Some(InlayHintKind.Type), Nil, ""))
+          Some(InlayHint(pos, abbreviate(label, 14), Some(InlayHintKind.Type), Nil, ""))
     }
   }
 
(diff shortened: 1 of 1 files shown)


magnus-madsen:
@mlutze @JonathanStarup Any comments? Otherwise I think we can just merge it.

I think 30 chars might be a bit too much? I don't really have a good solution though. We could also try to something fancy like `List[...]` and stuff like that (i.e. showing the type going from the outside in).

mlutze:
> We could also try to something fancy like List[...] and stuff like that (i.e. showing the type going from the outside in).

Scala does this and I think it's prettier. This is a good enough start.

> I've (almost) copied the truncate function from ProgressBar.scala, which might not be the right thing to do. Is there somewhere where this kind of utility should live?

We have a `Utils` folder. Maybe something like `StringOps` would be a good place for it.

> The difference between my copy of truncate and the one in ProgressBar.scala is that I'm using the Unicode ellipsis character … instead of three periods .... This strikes me as the right thing to do given that space is at a premium in inlay hints, but I'm open to being persuaded otherwise.

Clever, I like it.

magnus-madsen:
Thanks!

PR #4655, opened 2022-10-09
fix: Towards fixing duplicate names

### LINKED ISSUE #4649: Ambiguous types compile erroneously
Some examples of programs which compile and which shouldn't:

```
import java.sql.Statement;
enum Statement

def foo(): Statement = ???
```

```
enum Statement
enum Foo
type alias Statment = Foo

def foo(): Statement = ???
```

```
import java.sql.Statement;
use Foo.Statement;

namespace Foo {
    pub enum Statement
}

def foo(): Statement = ???
```

```
use Foo.Statement;

namespace Foo {
 pub enum Statement
}
enum Statement
def foo(): Statement = ???
```

Split from #4392

paulbutcher:
Part of  #4649

This fixes cases where the names of classes, enums, aliases, or effects clash with the names of used or imported types. It doesn't fix the case where used types clash with imported types (see commented out test in `TestNamer` for an example).

I think that the right way to address the latter is to merge `ImportEnv` into `UseEnv`, but that involves largely undoing the change I made a while ago to plumb `ImportEnv` through the namer. Given that that means hundreds of lines changing (eugh!) I wanted to check this in separately.

A note about testing: To test this exhaustively, we would have to test every possible combination of classes, enums, aliases, effects, uses and imports, with various different levels of nesting. I've only added a subset of the missing combinations: I'm not sure how much value there is in adding all the combinations, but will do so if we think it useful.

### DIFF
--- a/main/test/ca/uwaterloo/flix/language/phase/TestNamer.scala
+++ b/main/test/ca/uwaterloo/flix/language/phase/TestNamer.scala
@@ -921,18 +921,18 @@ class TestNamer extends FunSuite with TestUtils {
     expectError[NameError.DuplicateUpperName](result)
   }
 
-  // test("DuplicateUpperName.23") {
-  //   val input =
-  //     """
-  //       |import java.sql.Statement
-  //       |use A.Statement
-  //       |namespace A {
-  //       |    enum Statement
-  //       |}
-  //       |""".stripMargin
-  //   val result = compile(input, Options.TestWithLibNix)
-  //   expectError[NameError.DuplicateUpperName](result)
-  // }
+  ignore("DuplicateUpperName.23") {
+    val input =
+      """
+        |import java.sql.Statement
+        |use A.Statement
+        |namespace A {
+        |    enum Statement
+        |}
+        |""".stripMargin
+    val result = compile(input, Options.TestWithLibNix)
+    expectError[NameError.DuplicateUpperName](result)
+  }
 
   test("DuplicateUpperName.24") {
     val input =
--- a/main/test/ca/uwaterloo/flix/language/phase/TestNamer.scala
+++ b/main/test/ca/uwaterloo/flix/language/phase/TestNamer.scala
@@ -901,6 +901,52 @@ class TestNamer extends FunSuite with TestUtils {
     expectError[NameError.DuplicateUpperName](result)
   }
 
+  test("DuplicateUpperName.21") {
+    val input =
+      """
+        |import java.sql.Statement
+        |enum Statement
+        |""".stripMargin
+    val result = compile(input, Options.TestWithLibNix)
+    expectError[NameError.DuplicateUpperName](result)
+  }
+
+  test("DuplicateUpperName.22") {
+    val input =
+      """
+        |enum Statement
+        |type alias Statement = Int
+        |""".stripMargin
+    val result = compile(input, Options.TestWithLibNix)
+    expectError[NameError.DuplicateUpperName](result)
+  }
+
+  // test("DuplicateUpperName.23") {
+  //   val input =
+  //     """
+  //       |import java.sql.Statement
+  //       |use A.Statement
+  //       |namespace A {
+  //       |    enum Statement
+  //       |}
+  //       |""".stripMargin
+  //   val result = compile(input, Options.TestWithLibNix)
+  //   expectError[NameError.DuplicateUpperName](result)
+  // }
+
+  test("DuplicateUpperName.24") {
+    val input =
+      """
+        |use A.Statement
+        |namespace A {
+        |    enum Statement
+        |}
+        |enum Statement
+        |""".stripMargin
+    val result = compile(input, Options.TestWithLibNix)
+    expectError[NameError.DuplicateUpperName](result)
+  }
+
   test("SuspiciousTypeVarName.01") {
     val input =
       s"""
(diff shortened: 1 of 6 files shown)


paulbutcher:
OK, so that clearly didn't work as well as I hoped. All of the failures seem to be related to `Fixpoint/Interpreter.flix` which contains code of the form:

```
use Fixpoint/Tuple.Tuple

namepace Fixpoint {
    namespace Tuple {
        pub enum Tuple
    }
}
```

This is definitely being rejected according to what I had in mind while writing the code: I didn't realise that we had code which used this pattern though.

I'm not sure whether the solution is to change `Fixpoint/Interpreter.flix` or to change the compiler. If the latter, how?

magnus-madsen:
Hmm. Surprising. So (the programmer) imports Tuple at the top-into the same file where he declares it. In this case it seems strange, but are there legitimate use cases?

magnus-madsen:
Does removing line 19 in interpreter solve the problem?

magnus-madsen:
What if you have:

```
namespace Foo {
  namespace Bar { enum Qux }

  // Want to refer to Qux, can I import Foo/Bar.Qux at the top or not?
}
```

paulbutcher:
> Does removing line 19 in interpreter solve the problem?

No, because `Tuple` is used within the `Fixpoint` namespace.

> What if you have:
>
> ```
> namespace Foo {
>   namespace Bar { enum Qux }
> 
>   // Want to refer to Qux, can I import Foo/Bar.Qux at the top or not?
> }
> ```

Not with this PR, no. What will work is:

```
namespace Foo/Bar { enum Qux }

namespace Foo {
  use Foo/Bar.Qux
}
```

ASIDE: Unrelated to this specific issue, I'm a little surprised that the following doesn't compile?

```
namespace Foo/Bar { enum Qux }

namespace Foo {
  use Bar.Qux
}
```

paulbutcher:
It looks like there's one other file other than `Fixpoint/Interpreter.flix` with the same issue:

https://github.com/flix/flix/blob/master/main/test/flix/Test.Use.Type.flix

magnus-madsen:
OK; I understand the issue now. The question is what to do about it. The ambiguity occurs inside Fixpoint/Tuple but not inside Fixpoint.

paulbutcher:
> OK; I understand the issue now. The question is what to do about it. The ambiguity occurs inside Fixpoint/Tuple but not inside Fixpoint.

For reference, I've just pushed changes which make the tests pass with the PR as it currently stands.

paulbutcher:
@stephentetley, as someone with a significant corpus of existing code, it would be great to get your thoughts on this:

1. What do you think about it from a philosophical POV?
2. Does this PR break any of your code?

stephentetley:
Thanks for the heads up Paul.

I don't think it would break any of my code - I've switched to using `use` inside namespaces rather than at the top of a file. Maybe I have some "main" modules that use top level `use` but if they break it wouldn't be a hassle to update them.

magnus-madsen:
(I have not forgotten, but I want to check how this works in other languages).

magnus-madsen:
I am OK with this. Will you fix the commented test case?

magnus-madsen:
> Thanks for the heads up Paul.
> 
> I don't think it would break any of my code - I've switched to using `use` inside namespaces rather than at the top of a file. Maybe I have some "main" modules that use top level `use` but if they break it wouldn't be a hassle to update them.

If you have time, could you expand a bit on what you think of the current namespace system? The mechanism a bit old and I was never convinced we got it 100% right. But I don't know what other language to look at for inspiration.

paulbutcher:
> I am OK with this. Will you fix the commented test case?

Yes, but in a subsequent PR (it'll involve unwinding the changes I made to plumb `ImportEnv` through the namer, which will mean hundreds of lines changing 😟

magnus-madsen:
> > I am OK with this. Will you fix the commented test case?
> 
> Yes, but in a subsequent PR (it'll involve unwinding the changes I made to plumb `ImportEnv` through the namer, which will mean hundreds of lines changing 😟

I am a fan of incremental development! :-)

Once we have rewritten something about three times then it tends to be good 👍

paulbutcher:
Cool. I'll push ahead with it as soon as this is merged.

main/test/ca/uwaterloo/flix/language/phase/TestNamer.scala
```diff
@@ -901,6 +901,52 @@ class TestNamer extends FunSuite with TestUtils {
...
+    val input =
+      """
+        |enum Statement
+        |type alias Statement = Int
+        |""".stripMargin
+    val result = compile(input, Options.TestWithLibNix)
+    expectError[NameError.DuplicateUpperName](result)
+  }
+
+  // test("DuplicateUpperName.23") {
```
magnus-madsen, line 924: Whats the plan for this test?

paulbutcher: As I mentioned at the top of this PR:

> I think that the right way to address the latter is to merge ImportEnv into UseEnv, but that involves largely undoing the change I made a while ago to plumb ImportEnv through the namer. Given that that means hundreds of lines changing (eugh!) I wanted to check this in separately.

(so, when we add a new upperName to UseEnv, we'll check that it doesn't clash with the imports, and when we add a new import, we'll check that it doesn't clash with an upperName).

magnus-madsen: OK, can you mark the test as ignored then?

paulbutcher: Instead of commented out, you mean? Sure, once I work out how to do that 🤣

magnus-madsen: I think you just change test(".. to ignore("...

paulbutcher: I think that this is ready to be merged.

magnus-madsen:
Thanks!

Issue #4653, opened 2022-10-08
Essential Variable Followup


mlutze:
Followup from https://github.com/flix/flix/pull/4645

- [x] make helper for `t[x -> true]` and `t[x -> false]` transformations
- [ ] retry  introducing `RegionSym`
- [ ] performance

magnus-madsen:
- [ ] Performance

PR #4660, opened 2022-10-09
refactor: introduce RegionSym

mlutze:
related to #4653

### DIFF
--- a/examples/named-arguments.flix
+++ b/examples/named-arguments.flix
@@ -1,0 +1,36 @@
+// When functions have two arguments of the same type,
+// it can be difficult to remember the right order.
+// We can use named arguments to be sure we are applying the function correctly.
+// The parameter type becomes a singleton record,
+// and a syntactic sugar allows us to elide the braces when calling a function with a named argument.
+
+///
+/// Returns true if the `sublist` is contained within the list `l`.
+///
+pub def contains(sublist: {sublist = List[a]}, l: List[a]): Bool with Eq[a] = match (sublist.sublist, l) {
+    case (Nil, _) => true
+    case (_ :: _, Nil) => false
+    case (h1 :: t1, h2 :: t2) if h1 == h2 => contains(sublist = t1, t2) or contains(sublist = sublist.sublist, t2)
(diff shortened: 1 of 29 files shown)


mlutze:
This seems to have the same issue we faced before:

We build up a substitution that includes something like
```
a -> a and r
```

The type of our expression is inferred to be
```
a and r
```

We purify `r` and the type becomes
```
a
```

and then we apply the substitution and `r` returns:
```
a and r
```

magnus-madsen:
So, we need more information in the substitution or we need to split it somehow.

magnus-madsen:
Should we apply the substitution first and then purify?

magnus-madsen:
> > Should we apply the substitution first and then purify?
> 
> I don't think that can work.
> 
> ```scala
>       case KindedAst.Expression.Scope(sym, regionVar, exp, pvar, loc) =>
>         for {
>           // don't make the region var rigid if the --Xflexible-regions flag is set
>           _ <- if (flix.options.xflexibleregions) InferMonad.point(()) else rigidifyM(regionVar)
>           _ <- unifyTypeM(sym.tvar, Type.mkRegion(regionVar, loc), loc)
>           (constrs, tpe, pur, eff) <- visitExp(exp)
>           purifiedPur <- purifyEffM(regionVar, pur)
>           resultPur <- unifyTypeM(pvar, purifiedPur, loc)
>           _ <- noEscapeM(regionVar, tpe)
>           resultTyp = tpe
>           resultEff = eff
>         } yield (constrs, resultTyp, resultPur, resultEff)
> ```
> 
> The `resultPur <- ...` triggers another substitution and cannot be done before purifying.

I am not saying this is the right solution, but does the purify need to happen in purifyM, what if we do it in the yield?

mlutze:
> I am not saying this is the right solution, but does the purify need to happen in purifyM, what if we do it in the yield?

I think that still breaks because `pvar` needs to be purified too.

magnus-madsen:
A lot of conflicts. Should we close and revisit?

mlutze:
Yep

PR #4665, opened 2022-10-10
feat: Improve tests and error messages for static nested classes

### LINKED ISSUE #4663: Support for static nested Java classes/interfaces
Split out from #4392.

A static nested class/interface is of the form:

```
class OuterClass {
    static class StaticNestedClass { }
}
```

(note that this is not the same as an inner class)

paulbutcher:
Fixes #4663

It turns out that this already works: the trick is to use '$' to separate the nested class instead of '.'.

I've added a few tests to confirm that this works, and improved the error messages to make this more discoverable.


JonathanStarup:
Oh yeah, I remember discovering this too and forgetting about it again. It really should be documented somewhere

paulbutcher:
> Oh yeah, I remember discovering this too and forgetting about it again. It really should be documented somewhere

https://github.com/flix/book/pull/55

magnus-madsen:
Thanks!

Issue #4664, opened 2022-10-10
Improve `debug`


magnus-madsen:
A concrete list of improvements:

- [x] Add support for Unit (JLS)
- [x] Add support for `toString` on records
- [x] Add support for `toString` on closures
- [x] Refactor `Test.Exp.Debug` to test `debugString` directly (ML)
- [x] Add support for debug interpolation `"x = #{x}` should desugar to a call to `debugString` (ML)
- [x] Fix redundancy issue (?) (ML)
- [x] Add debug variants.
- [x] Make debug work like the identity function.
- [x] Format floating point with suffix (e.g. `f64`) (ML)
- [x] Format integers with suffix (e.g. `Int32`) (ML)
- [x] Format strings with quotes " (ML)
- [x] Format arrays with `Arrays.deepToString` (ML)
- [ ] #4679
- [x] #4680
- [ ] #4698

JonathanStarup:
Also add custom redundancy hoops right? to allow `debug(Test(42)); 2`

magnus-madsen:
> Also add custom redundancy hoops right? to allow debug(Test(42)); 2

Yes, does that not work already?

> At the backend we only have Int32 -> Bool, and only object for non-primitive things. I am also a bit sceptical about the usefulness. This is all in the type? isn't it better just to keep it brief for readability.

Good point! Not sure then. We could keep just the purity? (Maybe that information is gone too?)

> I feel the same about number suffixes. Its all known statically

Programmers could be confused. The point of the debug statement is to be very precise.

magnus-madsen:
Another reason for keeping the suffixes: https://en.wikipedia.org/wiki/Homoiconicity

JonathanStarup:
> We could keep just the purity? (Maybe that information is gone too?)

Maybe we could find the name sometimes, but I think `Fun#18y4271` has all the information you need.

> Then you can copy-paste a printed value into a program.

hmm. but the same logic could be said about normal `toString` right? and I think its a good thing that it isn't `42.5f32` by flix syntax you can't paste it anywhere else than flix

mlutze:
@JonathanStarup can we also fix unit type and paramless enums in the back-end?

```
debug(Card(Hearts, N(5)))
```
gives
```
[main/test/ca/uwaterloo/flix/library/TestPrelude.flix:233] Card(Hearts(dev.flix.runtime.Unit@46cc127b), N(5))
```

magnus-madsen:
This feature is the most exciting of the year!

[...]

PR #4667, opened 2022-10-10
test: move debug tests to testPrelude

mlutze:
related to #4664


JonathanStarup:
I ruined your tests, now magnus can decide who has to correct them 😅

mlutze:
Dang, I wish you did that an hour ago :P

magnus-madsen:
Huh. What happened?

mlutze:
Jonathan just improved the output for tuples etc. so my tests that look for `Tuple2` instead of `(abc, 123)` are invalid.

magnus-madsen:
Ah. Well it seems you should fix the tests then ;)

magnus-madsen:
Its interesting that by separating the implementer from the tester you found some bugs/improvements.

PR #4669, opened 2022-10-10
feat: introduce debug interpolation

mlutze:
related to #4664. I went with % instead of # since there were some parsing weirdnesses related to schemas and we use # for e.g. Map#{

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/ParsedAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/ParsedAst.scala
@@ -2049,6 +2049,15 @@ object ParsedAst {
       */
     case class StrPart(sp1: SourcePosition, chars: Seq[ParsedAst.CharCode], sp2: SourcePosition) extends InterpolationPart
 
+    /**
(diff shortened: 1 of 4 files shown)


magnus-madsen:
I will merge this, but I actually liked `#`. What's the parsing problem? Will `%` not be confused for some format specifier?

mlutze:
> I will merge this, but I actually liked `#`. What's the parsing problem? Will `%` not be confused for some format specifier?

I wasn't able to figure out exactly what the parsing problem was, only that changing the sigil fixed it. `%` was an arbitrary choice so I'm fine going with something else.

magnus-madsen:
Rust uses `?` or `#?` am I reading that right?

https://doc.rust-lang.org/std/fmt/trait.Debug.html

mlutze:
> Rust uses `?` or `#?` am I reading that right?
> 
> https://doc.rust-lang.org/std/fmt/trait.Debug.html

Yes, seems like it.

magnus-madsen:
Failed :(

PR #4670, opened 2022-10-11
fix: mark debug as impure

magnus-madsen:


JonathanStarup:
Oh it was me that broke it, my bad

PR #4671, opened 2022-10-11
feat: improve `java.lang.Object.toString` of `Unit`

JonathanStarup:
Related to #4664.

* Unit toString is `()`
* Nullary enum values `enum Test { case TestA }` prints `TestA`, not `TestA()` or `TestA(())`


magnus-madsen:
@JonathanStarup Looks good, but lets extend the test cases in Test.Debug (or what was it)

JonathanStarup:
@magnus-madsen It's really nice coding in the new parts of the backend vs. the old parts. It's good to feel the payoff

PR #4685, opened 2022-10-11
feat: improve array formatting for stringify

mlutze:
related to #4664


magnus-madsen:
All that code and only 3 test cases. Inflation must be hitting hard!

magnus-madsen:
Also could you not just call deepToString in all cases?

mlutze:
> All that code and only 3 test cases. Inflation must be hitting hard!

Yeah, turned a lot bigger than I thought it would be. I'll go make more tests :(

> Also could you not just call deepToString in all cases?

Nope. `deepToString` takes a `Object[]` so it cannot be used for e.g. `Array[Bool, Static]`

JonathanStarup:
> > All that code and only 3 test cases. Inflation must be hitting hard!
> 
> Yeah, turned a lot bigger than I thought it would be. I'll go make more tests :(
> 
> > Also could you not just call deepToString in all cases?
> 
> Nope. `deepToString` takes a `Object[]` so it cannot be used for e.g. `Array[Bool, Static]`

It's funny I also just coped this in jvm buts its not really reusable. deepToString must also have this kind of logic inside. A showcase of a pain point of the java type system.

magnus-madsen:
RFM?

PR #4691, opened 2022-10-13
feat: escape strings for debug output

mlutze:
related to #4664. Note that this does not work for strings inside of arrays or other objects

### DIFF
--- a/main/src/library/Prelude.flix
+++ b/main/src/library/Prelude.flix
@@ -135,71 +135,17 @@ pub def println(x: a): Unit \ IO with ToString[a] =
 /// Escapes the given string per Flix's escaping rules.
 ///
 @Internal
-pub def escape(s: String): String = region r {
-    import java.lang.String.charAt(Int32): Char \ {};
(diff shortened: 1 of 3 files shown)


main/src/library/Prelude.flix
```diff
@@ -131,6 +131,77 @@ pub def println(x: a): Unit \ IO with ToString[a] =
...
+                        discard append(sb, "\\u" + code);
+                        visit(i + 1)
+                    } else {
+                        discard append(sb, "${c}");
+                        visit(i + 1)
+                    }
+            }
+        }
+
+    };
```
magnus-madsen, line 200: I think this is overly complex. I would only focus on replacing \n and \r etc. and using some String.replaceAll.

The more fancy stuff we can add later and by need.

JonathanStarup:
A fun implementation could also be to take the source program representation of the string

magnus-madsen:
I would really just escape the escape chars.

magnus-madsen:
I mean the opposite, replace newlines by escape chars etc.

PR #2696, opened 2021-11-22
Channel lowering

JonathanStarup:
part of #2169

Issue #2169, opened 2021-08-23
Tidy Backend


magnus-madsen:
These are some minor changes we should make to the existing backend before it is replaced by the new backend.

The purpose of these changes is to make the generated code independent of (1) the Flix compiler and (2) the flix-runtime.jar.

- [x] #2170
- [x] #2171
- [x] #2172
- [x] #2173
- [x] #2174
- [x] #2175
- [x] #1654
- [x] #2270
- [x] #2271
- [x] Re-implement channels in Flix. See in particular #1724 and #1840. I suggest to take #1904 as a starting point and break it down into the following sub-tasks
  - [x] #2176
  - [x] Backend: Concurrency: Implement a CyclicBuffer (Maybe #2031 can be used?)
  - [x] #2177
  - [x] #2178
  - [ ] Backend: Concurrency: Implement the Channel namespace using #1904.
  - [ ] Backend: Concurrency: Implement a Lowering phase to the Channel namespace.
  - [ ] Backend: Concurrency: Remove old, unused code.

**These changes should be kept minimal and independent of the new backend.**

justinfargnoli:
I created issue #2221 for the "Backend: Concurrency: Implement a CyclicBuffer (Maybe Add MutDeque #2031 can be used?)" checkbox.

justinfargnoli:
@magnus-madsen are #2177 and #2178 saying introduce a Flix namespace into the Flix standard library where all of the functions are just wrappers to the Java methods of the `ReenterantLock` and `Condition` type?

magnus-madsen:
> 
> 
> @magnus-madsen are #2177 and #2178 saying introduce a Flix namespace into the Flix standard library where all of the functions are just wrappers to the Java methods of the `ReenterantLock` and `Condition` type?

Yes

JonathanStarup:
@magnus-madsen maybe just close this?

PR #4692, opened 2022-10-13
feat: remove `Boxed` from `Channel.flix`

JonathanStarup:
Related to #2169.

Based on disucssion in #2696

### DIFF
--- a/main/src/library/Concurrent/Channel.flix
+++ b/main/src/library/Concurrent/Channel.flix
@@ -19,6 +19,9 @@
 namespace Concurrent/Channel {
     use Concurrent/ReentrantLock.lock
     use Concurrent/ReentrantLock.ReentrantLock
+    use Concurrent/ReentrantLock.newLock
+    use Concurrent/Condition.Condition
+    use Concurrent/ReentrantLock.newCondition
 
     ///
     /// An enum that represents a multiple-producer, multiple-consumer (mpmc)
@@ -74,6 +77,7 @@ namespace Concurrent/Channel {
         MutDeque[a, Static] // elementDeque
     )
 
+    /// Returns the `MpmcAdmin` of `c`.
     pub def mpmcAdmin(c: Mpmc[a]): MpmcAdmin = {
         let Mpmc(admin, _) = c;
         admin
@@ -85,8 +89,8 @@ namespace Concurrent/Channel {
         Bool, // unBuffered
         Int32, // maxSize
         Ref[Int32, Static], // size
-        MutList[(ReentrantLock, Concurrent/Condition.Condition), Static], // waitingGetters
-        Concurrent/Condition.Condition // waitingPutters
+        MutList[(ReentrantLock, Condition), Static], // waitingGetters
+        Condition // waitingPutters
     )
 
     ///
@@ -121,7 +125,7 @@ namespace Concurrent/Channel {
         import static dev.flix.runtime.Global.newId(): Int64 \ IO;
         let _bufferCheck = if (bufferSize < 0) bug!("bufferSize < 0") else ();
         let unBuffered = bufferSize == 0;
-        let reentrantLock = Concurrent/ReentrantLock.newLock(false);
+        let reentrantLock = newLock(false);
         let size = ref 0;
         Mpmc(
             MpmcAdmin(
@@ -131,7 +135,7 @@ namespace Concurrent/Channel {
                 if (unBuffered) 1 else bufferSize,
                 size,
                 new MutList(Static),
-                Concurrent/ReentrantLock.newCondition(reentrantLock)
+                newCondition(reentrantLock)
             ),
             new MutDeque(Static)
         )
@@ -187,6 +191,10 @@ namespace Concurrent/Channel {
         unlock(channelLock);
         element
 
+    ///
+    /// Receives an element from the channel `c`. This will crasah if the
+    /// channel is empty.
+    ///
     @Internal
     pub def assertiveGet(c: Mpmc[a]): a \ IO = {
         let Mpmc(admin, elementDeque) = c;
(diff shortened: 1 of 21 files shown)


magnus-madsen:
Overall looks good. I will propose some stylistic changes.

Question question: Why do you need eq and order? I think they are dangerous.

JonathanStarup:
> Overall looks good. I will propose some stylistic changes.
> 
> Question question: Why do you need eq and order? I think they are dangerous.

locking order. the functions could also be inlined but then its not so easy to call `Array.sort`.

magnus-madsen:
> > Overall looks good. I will propose some stylistic changes.
> > Question question: Why do you need eq and order? I think they are dangerous.
> 
> locking order. the functions could also be inlined but then its not so easy to call `Array.sort`. Its only on the private admin enum not on the mpmc

What about using Array.sortBy?

JonathanStarup:
> What about using Array.sortBy?

Of yeah that works. Why are the instances dangerous?

magnus-madsen:
Actually they are not because you only refer to immutable data 👍 

I had not though about that.

Even so, I would prefer a few well-chosen sortBy.

main/src/library/Concurrent/Channel.flix
```diff
@@ -67,59 +72,50 @@ namespace Concurrent/Channel {
     /// conditions in `waitingPutters` comes from locks created by the get
     /// function or the select function.
     ///
```
magnus-madsen, line 74: As always, I would ask for some documentation here to explain that Mpmc now consists of two parts.

main/src/library/Concurrent/Channel.flix
```diff
@@ -67,59 +72,50 @@ namespace Concurrent/Channel {
...
-    /// Order is based on the channel `id` and is used by `select` to lock
-    /// in a global ordering.
-    ///
-    instance Order[Mpmc] {
-        pub def compare(x: Mpmc, y: Mpmc): Comparison =
-        let Mpmc(a_id, _, _, _, _, _, _) = x;
-        let Mpmc(b_id, _, _, _, _, _, _) = y;
-        a_id <=> b_id
-    }
+    enum MpmcAdmin(
```
magnus-madsen, line 86: I wish we had a better name, but I don't have any suggestions.

main/src/library/Concurrent/Channel.flix
```diff
@@ -171,30 +169,71 @@ namespace Concurrent/Channel {
         unlock(channelLock);
         element
 
+    ///
+    /// Receives an element from the channel `c`. This will crasah if the
```
magnus-madsen, line 173: crash

main/src/library/Concurrent/Channel.flix
```diff
@@ -67,59 +72,50 @@ namespace Concurrent/Channel {
...
-    /// Equality is based on the channel `id` and is needed for the `Order`
-    /// instance.
-    ///
-    instance Eq[Mpmc] {
-        pub def eq(x: Mpmc, y: Mpmc): Bool =
-        let Mpmc(a_id, _, _, _, _, _, _) = x;
-        let Mpmc(b_id, _, _, _, _, _, _) = y;
-        a_id == b_id
+    /// Returns the `MpmcAdmin` of `c`.
+    pub def mpmcAdmin(c: Mpmc[a]): MpmcAdmin = {
```
magnus-madsen, line 81: You have this function, but don't use it consistently. Perhaps drop it?

main/src/library/Concurrent/Channel.flix
```diff
@@ -171,30 +169,71 @@ namespace Concurrent/Channel {
...
+    /// Receives an element from the channel `c`. This will crasah if the
+    /// channel is empty.
+    ///
+    @Internal
+    pub def assertiveGet(c: Mpmc[a]): a \ IO = {
+        let Mpmc(admin, elementDeque) = c;
+        let MpmcAdmin(_, channelLock, _, _, size, _, _) = admin;
+        lock(channelLock);
+
+        // Try to get the element (which could already be taken by someone
```
magnus-madsen, line 182: Lets phrase this as: Gets an element from the channel `c` which must be non-empty.

main/src/library/Concurrent/Channel.flix
```diff
@@ -171,30 +169,71 @@ namespace Concurrent/Channel {
         unlock(channelLock);
         element
 
+    ///
+    /// Receives an element from the channel `c`. This will crasah if the
+    /// channel is empty.
+    ///
+    @Internal
+    pub def assertiveGet(c: Mpmc[a]): a \ IO = {
```
magnus-madsen, line 177: How about `unsafeGet`?

main/src/library/Concurrent/Channel.flix
magnus-madsen: "Assumes the channel to be non-empty".

main/src/library/Concurrent/Channel.flix
```diff
@@ -171,30 +169,71 @@ namespace Concurrent/Channel {
...
+        let MpmcAdmin(_, channelLock, _, _, size, _, _) = admin;
+        lock(channelLock);
+
+        // Try to get the element (which could already be taken by someone
+        // else)
+        let optionalElement = MutDeque.popFront(elementDeque);
+        size := deref size - 1;
+
+        match optionalElement {
+            case None => bug!("assertiveGet could not retrieve channel element")
```
magnus-madsen, line 188: Explain this cannot happen.

main/src/library/Concurrent/Channel.flix
```diff
@@ -240,44 +281,30 @@ namespace Concurrent/Channel {
...
-        let element = MutDeque.popFront(elementDeque);
-
-        // If there was an element, signal waiting setters
-        Option.foreach(_ -> signalPutters(c), element);
-
-        unlock(channelLock);
-        element
-
     ///
     /// Recursive helper function for select, returns the first channel that
```
magnus-madsen, line 289: Docs.

magnus-madsen (commented):
All my comments are nitpicks; I am just asking for better documentation and naming.

magnus-madsen:
(This implementation could be an interesting target for verification.)

main/src/library/Concurrent/Channel.flix
JonathanStarup: It was meant for lowering (`pub def`), It seems annoying to unpack the ADT in the compiler

JonathanStarup: I also though about making the admin a record type since it could be nice to access fields directly

magnus-madsen: I think we can afford to use some records, but that could be in another iteration.

magnus-madsen:
Other than the build, I think we could try to merge this.

PR #4694, opened 2022-10-13
fix: Prevent duplicate upper names

### LINKED ISSUE #4649: Ambiguous types compile erroneously

paulbutcher:
Fixes #4649

Buckle up, this was less straightforward than I thought it was going to be.

The basic problem was that neither uses nor imports were checking to see if they clashed with pre-existing upper case names (enums, classes, aliases, effects). The solution is easy enough: call `lookupUpperName` within both `mergeUseEnvs` and `mergeImportEnvs`.

The problem is that that requires the namespace and compilation unit root, which need to be plumbed all the way through, which amounts to a lot of changes.

The parameter lists are starting to get uncomfortably long. Given that just about every function takes the same set of parameters now, we might like to think about packaging them up into a single data structure (but not as part of this change).

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/errors/NameError.scala
+++ b/main/src/ca/uwaterloo/flix/language/errors/NameError.scala
@@ -124,32 +124,6 @@ object NameError {
   }
 
   /**
-    * An error raised to indicate that the given uppercase `name` is used twice.
-    *
-    * @param name the clashing name.
-    * @param loc1 the location of the first use.
-    * @param loc2 the location of the second use.
-    */
-  case class DuplicateUseUpper(name: String, loc1: SourceLocation, loc2: SourceLocation) extends NameError {
-    def summary: String = s"Duplicate use of '$name'."
-
-    def message(formatter: Formatter): String = {
-      import formatter._
-      s"""${line(kind, source.name)}
-         |>> Duplicate use of the type or class '${red(name)}'.
-         |
-         |${code(loc1, "the first use was here.")}
-         |
-         |${code(loc2, "the second use was here.")}
-         |""".stripMargin
-    }
-
-    def explain(formatter: Formatter): Option[String] = None
-
-    def loc: SourceLocation = loc1
-  }
-
-  /**
     * An error raised to indicate that the given `tag` is used twice.
     *
     * @param name the clashing name.
--- a/main/src/ca/uwaterloo/flix/language/phase/Namer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Namer.scala
@@ -55,9 +55,13 @@ object Namer {
 
     // collect all the declarations.
     val declarations = mapN(traverse(program.units.values) {
-      case root => mapN(mergeUseEnvs(root.uses, UseEnv.empty), mergeImportEnvs(root.imports, ImportEnv.empty)) {
-        case (uenv0, ienv0) => root.decls.map(d => (uenv0, ienv0, d))
-      }
+      case root => 
+        flatMapN(mergeUseEnvs(root.uses, Name.RootNS, UseEnv.empty, ImportEnv.empty, prog0)) {
+          case uenv0 => 
+            mapN(mergeImportEnvs(root.imports, ImportEnv.empty, uenv0, Name.RootNS, prog0)) {
+              case ienv0 => root.decls.map(d => (uenv0, ienv0, d))
+            }
+        }
     })(_.flatten)
 
     // fold over the top-level declarations.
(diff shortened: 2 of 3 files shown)


magnus-madsen:
> The parameter lists are starting to get uncomfortably long. Given that just about every function takes the same set of parameters now, we might like to think about packaging them up into a single data structure (but not as part of this change).

Agree and agree (i.e. lets think about it and then execute in separate PR).

magnus-madsen:
@mlutze Lets do two reviews. Can you start? :)

magnus-madsen:
Questions (but not for this PR):

- Where did you need ns0?
- Would the next step likely be to merge usenv and ienv? (Possibly with ns0)?
- How do we feel about making `prog0` implicit? (We use implicit vary sparingly, but we do often use them for the AST root or in this case prog.)

magnus-madsen:
I guess making it implicit does not shorten the parameter lists only the argument lists.

mlutze (commented):
Generally looks good. I agree that we should combine `UseEnv` and `ImportEnv`.

magnus-madsen:
Merged. Thanks!

JonathanStarup:
@magnus-madsen @paulbutcher A consequence of this seemingly is that the following program does not compile, and neither does it if you include the `use`. You have to fully name the functions of outside definitions (in the file only)
```
namespace TestA {
    pub enum CoolEnum(Int32)
    namespace TestB {
        //use TestA.CoolEnum
        def generate(): CoolEnum = ???
    }
}
```

magnus-madsen:
OK; I am not sure if that was intended.

I am unsure if CoolEnum should be in scope in the inner namespace, but at least I think you should be permitted to use it.

What do you think?

paulbutcher:
This is surprising, but I don't believe that's a consequence of this change. This is with 0.29.0 from July:

```
paulbutcher@Pauls-MBP flix-playground % java -jar flix.jar --version
The Flix Programming Language v0.29.0
paulbutcher@Pauls-MBP flix-playground % java -jar flix.jar build    
-- Resolution Error -------------------------------------------------- /Users/paulbutcher/Projects/Flix/flix-playground/src/Main.flix

>> Undefined type 'CoolEnum'.

5 |         def generate(): CoolEnum = ???
                            ^^^^^^^^
                            type not found.

Tip: Possible typo or non-existent type?
```

magnus-madsen:
So it probably never worked, but how should it work?

paulbutcher:
I'll admit that I was assuming that anything in the outer namespace was in scope in the inner namespace.

Apparently that's not the case. 

Without that being true, then (some of) the errors I've introduced don't make sense.

magnus-madsen:
I assume we had this feature, so I am also a bit baffled.

magnus-madsen:
In Java, a class in package X.Y is not in scope of a class located in X.Y.Z.

magnus-madsen:
I think we have it at the level of visibility? But not scoping?

JonathanStarup:
this is why i want a rewamp of the namespaces! me and Matthew has some thorns to remove

It definitely worked recently, using master broke the code in Flixball

paulbutcher:
Sounds like we have some missing tests? 😟

mlutze:
Clearly we need Flixball in the community build!

(But we can't add it easily because it has dependencies)

paulbutcher:
Forgive my ignorace, but what is Flixball?

mlutze:
A little side-project that Jonathan and I work on: https://github.com/mlutze/flixball/issues

magnus-madsen:
> Forgive my ignorance, but what is Flixball?

Flixball is a AAA+ immersive, massive online multiplayer game being developed by @mlutze and @JonathanStarup. I hear its the next thing after the metaverse.

mlutze:
> > Forgive my ignorance, but what is Flixball?
> 
> Flixball is a AAA+ immersive, massive online multiplayer game being developed by @mlutze and @JonathanStarup. I hear its the next thing after the metaverse.

We'll quote you for our ad campaign.

JonathanStarup:
> This is surprising, but I don't believe that's a consequence of this change. This is with 0.29.0 from June:
> 
> ```
> paulbutcher@Pauls-MBP flix-playground % java -jar flix.jar --version
> The Flix Programming Language v0.29.0
> paulbutcher@Pauls-MBP flix-playground % java -jar flix.jar build    
> -- Resolution Error -------------------------------------------------- /Users/paulbutcher/Projects/Flix/flix-playground/src/Main.flix
> 
> >> Undefined type 'CoolEnum'.
> 
> 5 |         def generate(): CoolEnum = ???
>                             ^^^^^^^^
>                             type not found.
> 
> Tip: Possible typo or non-existent type?
> ```

sorry, i was unspecific, the version of the program with the use statement worked recently

magnus-madsen:
@JonathanStarup Can you open a new ticket and identify precisely what worked recently? :)

JonathanStarup:
> @JonathanStarup Can you open a new ticket and identify precisely what worked recently? :)

Here it is https://github.com/flix/flix/issues/4722 @magnus-madsen

PR #4695, opened 2022-10-13
feat: improve `toString` of function objects

JonathanStarup:
Realted to #4664.

Functions print their (java, i.e. no String or Unit) erased type, so `Int32 -> Int32`, `Obj -> Bool` etc. I've programmed other arities `() -> Int32` and `(Int32, Float32) -> Char` but I don't think any can exist

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
@@ -153,6 +153,7 @@ object BackendObjType {
 
       cm.mkConstructor(Constructor)
       args.indices.foreach(argIndex => cm.mkField(ArgField(argIndex)))
+      cm.mkMethod(ToStringMethod)
 
       cm.closeClassMaker()
     }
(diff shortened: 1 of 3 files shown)


JonathanStarup:
I think github timed out

main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
```diff
@@ -164,6 +165,28 @@ object BackendObjType {
     ))
 
     def ArgField(index: Int): InstanceField = InstanceField(this.jvmName, IsPublic, NotFinal, s"arg$index", args(index))
+
+    def ToStringMethod: InstanceMethod = {
```
magnus-madsen, line 169: How about we compute the string at compile-time instead? :)

JonathanStarup: Oh yeah of course, I'm stuck in jvm

JonathanStarup: Actually I was totally optimizing the amount of string constants of the class

magnus-madsen: But one string would be shortest right? No instructions to save :)

JonathanStarup: oh yeah theyre different classes. cant even find a made up excuse 😄

JonathanStarup:
A future idea might be that closures could show their captured variables

magnus-madsen:
> A future idea might be that closures could show their captured variables. It's easy on the backend, but maybe too messy to show.

Its interesting. We should think about it!

PR #4699, opened 2022-10-14
feat: add suffixes for numbers and escapes for characters

mlutze:
related to #4664

Note again that numbers nested inside objects are unaffected.

### DIFF
--- a/main/src/library/Prelude.flix
+++ b/main/src/library/Prelude.flix
@@ -174,29 +174,28 @@ pub def stringify(x: a): String = {
         case ReifiedBool =>
             (b -> if (b) "true" else "false") as a -> String
         case ReifiedChar =>
-            import static java.lang.Character.toString(Char): String \ {};
-            toString as a -> String
+            (c -> "\'" + escape("${c: Char}") + "\'") as a -> String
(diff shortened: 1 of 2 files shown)


main/src/library/Prelude.flix
```diff
@@ -174,29 +174,28 @@ pub def stringify(x: a): String = {
         case ReifiedBool =>
             (b -> if (b) "true" else "false") as a -> String
         case ReifiedChar =>
-            import static java.lang.Character.toString(Char): String \ {};
-            toString as a -> String
+            (c -> "\'" + escape("${c: Char}") + "\'") as a -> String
```
magnus-madsen, line 177: One question: Why are we returning a function? Would it not be better to simply return the string??

mlutze: Something weird with monomorph, I guess. @JonathanStarup may have more information.

JonathanStarup: because we dont eliminate the reification match at compile time, the verifier complains that the irrelevant cases cast to a wrong type. ype reification is a bit limited right now as a consequence thereof

PR #4706, opened 2022-10-14
fix: Disallow java names which aren't also Flix names

### LINKED ISSUE #4666: Require rename in import when Java name isn't valid Flix name
It's possible to import Java classes/interfaces with names that aren't valid Flix names, e.g.:

```
import java.util.Locale$Builder
```

In this case we should error and require that the user renames the class, e.g.:

```
import java.util.{Locale$Builder => Builder}
```

paulbutcher:
Fixes #4666

This works, but I'm not 100% happy with it: I've had to effectively re-implement the definition of `UpperCaseName`. I'd love to use the same code, but I can't see how I could reuse a portion of the parser?

If there is a way, I'd love to hear about it.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
+++ b/main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
@@ -865,6 +865,9 @@ object WeederError {
     })
   }
 
+  /**
+    * An error raised to indicate that an imported Java name is not a valid Flix identifier
+    */
   case class IllegalJavaClass(name: String, loc: SourceLocation) extends WeederError {
     def summary: String = "${name} is not a valid Flix identifer."
 
@@ -873,7 +876,7 @@ object WeederError {
       s"""${line(kind, source.name)}
          |>> ${red(name)} is not a valid Flix identifer.
          |
-         |${code(loc, "illegal identifer")}
+         |${code(loc, "identifer")}
          |
          |""".stripMargin
     }
--- a/main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
+++ b/main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
@@ -864,4 +864,27 @@ object WeederError {
          |""".stripMargin
     })
   }
+
+  case class IllegalJavaClass(name: String, loc: SourceLocation) extends WeederError {
+    def summary: String = "${name} is not a valid Flix identifer."
+
+    def message(formatter: Formatter): String = {
+      import formatter._
+      s"""${line(kind, source.name)}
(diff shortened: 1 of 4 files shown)


main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
```diff
@@ -864,4 +864,27 @@ object WeederError {
          |""".stripMargin
     })
   }
+
```
mlutze, line 867: We put Scaladoc on error classes.

main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
```diff
@@ -864,4 +864,27 @@ object WeederError {
...
+  case class IllegalJavaClass(name: String, loc: SourceLocation) extends WeederError {
+    def summary: String = "${name} is not a valid Flix identifer."
+
+    def message(formatter: Formatter): String = {
+      import formatter._
+      s"""${line(kind, source.name)}
+         |>> ${red(name)} is not a valid Flix identifer.
+         |
+         |${code(loc, "illegal identifer")}
```
mlutze, line 876: I think we're avoiding the word 'illegal' to sound more friendly. Not sure what the best replacement is though.

main/src/ca/uwaterloo/flix/language/phase/Weeder.scala
```diff
@@ -457,18 +457,19 @@ object Weeder {
   private def visitImport(i0: ParsedAst.Import): Validation[List[WeededAst.Import], WeederError] = i0 match {
     case ParsedAst.Imports.ImportOne(sp1, name, sp2) =>
       val loc = mkSL(sp1, sp2)
-      val ident = Name.Ident(sp1, name.fqn.last, sp2)
-      List(WeededAst.Import.Import(name, ident, loc)).toSuccess
+      val alias = name.fqn.last
+      if (raw"[A-Z][A-Za-z0-9_?]*".r matches alias) {
```
mlutze, line 461: Do we actually allow `?` in our names?

paulbutcher: Oops! That was meant to be `!` (blush). A good example of why it would be nice to be able to reuse the code in the Parser (I don't suppose you're aware of any way I could do this?)

mlutze: I don't think the Parboiled API makes it easy to do that. The macros seem to enforce some kind of self-containment of the parser which helps performance but hurts reusability.

PR #4711, opened 2022-10-15
fix: Allow new with protected constructor

paulbutcher:
Requiring that the type passed to `new` has a public constructor is too strict: a protected constructor will do just fine.


magnus-madsen:
Thanks!

Issue #4681, opened 2022-10-11
Introduce `Mask`?


magnus-madsen:
Proposal to have:

- Ascribe: Asks the typer to verify a type and effect is as expected. Safe.
- Cast: Tells the typer that an expression has the given type and effect. Unsafe.
- Mask: Is type checked as pure, but treated as impure post type inference. Unsafe.

We can then possibly refactor Debug to use `Mask`.

JonathanStarup:
Is the use case custom logging where you want the same behavior? If so, isn't that supposed to be solved by effects?

magnus-madsen:
> Is the use case custom logging where you want the same behavior? If so, isn't that supposed to be solved by effects?

The use case is to have a more general mechanism (internally) than a very specific "Debug" expression.

mlutze:
What's the syntax?

Where should the typing change come in? It seems like it needs to come after regions but before redundancy.

magnus-madsen:
Not sure if there is any surface syntax. The magic happens in Typer and Redundancy.

magnus-madsen:
The syntax could be `mask e` which is typed as pure (no matter the effect of e), but which is treated as impure after type inference.

magnus-madsen:
Its a bit hard to use in the surface language though. I mean you could do `mask println()`, but you could not write a function that does that (because it would be pure).

PR #4714, opened 2022-10-16
feat: add mask expression

mlutze:
related to #4681


magnus-madsen:
Can we try to refactor debug in this same PR?

magnus-madsen:
Does it appear to work? :)

mlutze:
Yep, did some very basic testing in my sandbox and output seems to be as expected.

magnus-madsen:
Also, if it is simple, perhaps we could even do the currying of debug etc in this PR? (or we can do it later)

mlutze:
> Could I ask you to move `Mask` right after `Cast` in the ASTs and pattern matches? I think it belongs there and would be good style.

Done



> Also, if it is simple, perhaps we could even do the currying of debug etc in this PR? (or we can do it later)

I'm not yet totally sure how that should be done, so let's save it for another PR.

magnus-madsen:
Also in the pattern matches? :) :)

magnus-madsen:
Thanks!

Issue #4713, opened 2022-10-16
Rename `FunctorFilter` to `Filterable` ?


stephentetley:
Now we have a `Witherable` class a la PureScript and Haskell should we rename `FunctorFilter` to `Filterable` ?

[1] https://hackage.haskell.org/package/witherable-0.4.2/docs/Witherable.html

[2] https://pursuit.purescript.org/packages/purescript-filterable/2.1.0

magnus-madsen:
I think that sounds reasonable. 

I forget if FunctorFilter is even used anywhere?

stephentetley:
> I forget if FunctorFilter is even used anywhere?

It's a superclass of `Witherable`. It can be filled out a bit more e.g. with `mapMaybe` and `catMaybe` to make it more useful.

I'll make a PR with just the renaming for the time being...

stephentetley:
PR #4715 submitted: 
refactor: rename FunctorFilter to Filterable, issue #4713

PR #4715, opened 2022-10-16
refactor: rename FunctorFilter to Filterable, issue #4713

stephentetley:
This PR renames the class `FunctorFilter` to `Filterable`. `Filterable` is the name Haskell and PureScript use for the equivalent class.


magnus-madsen:
Thanks!

Issue #4568, opened 2022-09-15
add `Bool.fromString` for consistensy


JonathanStarup:

magnus-madsen:
Blocked because of min lib.

magnus-madsen:
I think its better not to have this. One can easily do `x == "true"` or whatever anyways.

PR #4718, opened 2022-10-16
feat: add `Bool.fromString`

jaschdoc:
Do we want this behavior which is easy to reason about or do we want the `java.lang.Boolean.parseBoolean` behavior which returns `true` if the string was `"true"` and returns `false` on all other input?

https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/lang/Boolean.html#parseBoolean(java.lang.String)

Closes https://github.com/flix/flix/issues/4568

### DIFF
--- a/main/src/library/Bool.flix
+++ b/main/src/library/Bool.flix
@@ -84,10 +84,17 @@ namespace Bool {
     /// A successful parse is wrapped with `Some(x)`, a parse failure is indicated by `None`.
     ///
     @Time(1) @Space(1)
-    pub def fromString(s: String): Option[Bool] = match String.trim(s) {
-        case "true"  => Some(true)
-        case "false" => Some(false)
-        case _       => None
-    }
+    pub def fromString(s: String): Opt[Bool] =
+        import java.lang.String.strip(): String \ {};
+        match strip(s) {
+            case "true"  => S(true)
+            case "false" => S(false)
+            case _       => N
+        }
 
+}
+
+pub enum Opt[a] {
+    case S(a)
+    case N
 }
--- a/main/src/library/Bool.flix
+++ b/main/src/library/Bool.flix
@@ -87,14 +87,14 @@ namespace Bool {
     pub def fromString(s: String): Opt[Bool] =
         import java.lang.String.strip(): String \ {};
         match strip(s) {
-            case "true"  => S(true)
-            case "false" => S(false)
-            case _       => N
+            case "true"  => Opt.Some(true)
+            case "false" => Opt.Some(false)
+            case _       => Opt.None
         }
 
-}
+    pub enum Opt[a] {
+        case Some(a)
+        case None
+    }
 
-pub enum Opt[a] {
-    case S(a)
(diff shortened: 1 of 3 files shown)


mlutze:
> Do we want this behavior which is easy to reason about or do we want the `java.lang.Boolean.parseBoolean` behavior which returns `true` if the string was `"true"` and returns `false` on all other input?

This behavior for sure. Java's parsing seems insane to me.

jaschdoc:
> > Do we want this behavior which is easy to reason about or do we want the `java.lang.Boolean.parseBoolean` behavior which returns `true` if the string was `"true"` and returns `false` on all other input?
> 
> This behavior for sure. Java's parsing seems insane to me.

Agreed, seems crazy.

jaschdoc:
Adding this function creates a dependency between the std lib and the core lib so we would have to move array, option, string etc. into the core library. I think this is why we haven't done this before. What's the correct thing to do here? @magnus-madsen

jaschdoc:
Also a fully qualified tag throws a parse error when used in a match-case expression.

```flix
namespace Bool {
    pub def fromString(s: String): Opt[Bool] =
        import java.lang.String.strip(): String \ {};
        match strip(s) {
            case "true"  => Opt.Some(true)
            case "false" => Opt.Some(false)
            case _       => Opt.None
        }

    pub enum Opt[a] {
        case Some(a)
        case None
    }
}

namespace Option {
    pub def fromOpt(o: Bool.Opt[a]): Option[a] = match o {
        case Bool.Opt.Some(a) => Some(a)
        case Bool.Opt.None    => None
    }
}
```

```

-- Parse Error -------------------------------------------------- Option.flix

>> Parse Error: Invalid input '.', expected LegalLetter, Tuple, ::, "if" or "=>" (line 618, column 22):
        case Bool.Opt.Some(a) => Some(a)
                     ^
```

JonathanStarup:
You cant write `Ns.type.enum`, it seems you have to `use Ns.type` and then use `type.enum`

jaschdoc:
Why do we have this restriction?

magnus-madsen:
@jaschdoc Can you check what they other fromString do?

jaschdoc:
This is the `Int32` which I took inspiration from.

```flix
    ///
    @Time(1) @Space(1)
    pub def fromString(s: String): Option[Int32] = try {
        import java.lang.String.strip(): String \ {};
        import static java.lang.Integer.parseInt(String): Int32 \ {};
        Some(s |> strip |> parseInt)
    } catch {
        case _: ##java.lang.NumberFormatException => None
    }
```

jaschdoc:
But Ints don't really have the problem with the Option type since they are part of the std lib (not the core lib)

magnus-madsen:
I see. We can't have Bool.fromString unless we make Option part of the core :/

magnus-madsen:
I will close this for now (but leave the ticket open).

This cannot be resolved before we have some new module system in place that allows a better division of the standard library.

Issue #4721, opened 2022-10-16
Allow closures to implement suitable functional interfaces (e.g. `IntFunction`)


magnus-madsen:

JonathanStarup:
I can do the backend if somebody can do typer implementation

magnus-madsen:
You can test by starting with a cast. Then I will find someone to do the frontend. (@mlutze?)

mlutze:
I can do the front end if somebody can give me more details about what's needed :)

magnus-madsen:
> I can do the front end if somebody can give me more details about what's needed :)

Yes. When we lookup at Java type and see e.g. `IntFunction` we should translate it/treat it as `Int32 -> Int32`

magnus-madsen:
Maybe we can treat `##java.util.function.IntFunction` as pseudo-type alias for `Int32 -> Int32`?

mlutze:
In `Resolver` we already translate `java.lang.String` to `String` and `java.lang.BigInteger` to `BigInt`, so it seems like we can do a similar translation there.

Do we want to curry e.g. `BiFunction` or make it a tuple?

magnus-madsen:
> Do we want to curry e.g. BiFunction or make it a tuple?

Not sure

But lets start with `IntFunction` and see if we can make it work! :)

magnus-madsen:
@mlutze @JonathanStarup 

Perhaps what would be most sensible now is that you add support for e.g. `IntUnaryOperator`, refactor the code to be generalizable, and then @jaschdoc can add support for most of the stuff in https://docs.oracle.com/javase/8/docs/api/java/util/function/package-summary.html ?

magnus-madsen:
Its kind of crazy what they have there.

stephentetley:
Thanks Magnus, I've updated my library code with the new `IntNumber` and it passes `check`, I haven't got to writing / running tests yet.

magnus-madsen:
> Thanks Magnus, I've updated my library code with the new `IntNumber` and it passes `check`, I haven't got to writing / running tests yet.

Let us know how it goes :)

stephentetley:
I've got a PR almost ready that adds `IntPredicate`.

magnus-madsen:
@stephentetley Thanks! Perhaps you could be persuaded to add the other functions that you deem useful?

(I don't think we necessarily need everything in `java.util.Function`, but a few more would surely be useful?)

stephentetley:
[...]

PR #4724, opened 2022-10-18
feat: interpret Intfunction as Int32 -> Object

mlutze:
related to #4721

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
@@ -2509,26 +2509,27 @@ object Resolver {
   private def lookupJvmMethod(clazz: Class[_], methodName: String, signature: List[UnkindedType], retTpe: UnkindedType, static: Boolean, loc: SourceLocation)(implicit flix: Flix): Validation[Method, ResolutionError] = {
     // Lookup the signature.
     flatMapN(lookupSignature(signature, loc)) {
-      sig => try {
-        // Lookup the method with the appropriate signature.
-        val method = clazz.getMethod(methodName, sig: _*)
+      sig =>
+        try {
+          // Lookup the method with the appropriate signature.
+          val method = clazz.getMethod(methodName, sig: _*)
 
-        // Check if the method should be and is static.
-        if (static != Modifier.isStatic(method.getModifiers)) {
-          throw new NoSuchMethodException()
-        } else {
-          // Check that the return type of the method matches the declared type.
-          // We currently don't know how to handle all possible return types,
-          // so only check the straightforward cases for now and succeed all others.
-          // TODO move to typer
-          val erasedRetTpe = UnkindedType.eraseAliases(retTpe)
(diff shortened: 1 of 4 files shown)


magnus-madsen:
It should be possible to write some tests that result in ClassCastException (until the backend is fixed).

Perhaps adding all the different functions could be done by a student programmer? (If there is a "template" in place).

mlutze:
> It should be possible to write some tests that result in ClassCastException (until the backend is fixed).

Wouldn't those tests be better written by the backend developer? :P

> Perhaps adding all the different functions could be done by a student programmer? (If there is a "template" in place).

Yes, the pattern I've used here ought to work for the rest of the function types.

magnus-madsen:
I think the test actually needs to pass the function into something in Java. E.g. something with streams.

magnus-madsen:
> Wouldn't those tests be better written by the backend developer? :P

I think there should be tests written by both, no? :)

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -2653,21 +2657,31 @@ object Resolver {
...
         case TypeConstructor.Union => ResolutionError.IllegalType(tpe, loc).toFailure
 
+        case t: TypeConstructor.Arrow => throw InternalCompilerException(s"unexpected type: $t")
+        case t: TypeConstructor.Enum => throw InternalCompilerException(s"unexpected type: $t")
 
       }
 
-      // Case 2: Enum. Return an object type.
+      // Case 2: Arrow. Convert to Java function interface
+      case UnkindedType.Arrow(_, _, _) =>
```
magnus-madsen, line 2666: What's being converted here? Why are we converting TO IntFunction?

magnus-madsen: Makes sense 👍

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -1853,6 +1853,12 @@ object Resolver {
       fqn match {
         case "java.math.BigInteger" => UnkindedType.Cst(TypeConstructor.BigInt, loc).toSuccess
         case "java.lang.String" => UnkindedType.Cst(TypeConstructor.Str, loc).toSuccess
+        case "java.util.function.IntFunction" =>
```
magnus-madsen, line 1856: Could this be one line, since there is going to be many lines?

magnus-madsen: Perhaps these constants should be elsewhere?

magnus-madsen:
Is this ready to be merged and handed-off to Jakob and Jonathan?

mlutze:
> Is this ready to be merged and handed-off to Jakob and Jonathan?

Yes

magnus-madsen:
@JonathanStarup Matt did something for the frontend.

PR #4729, opened 2022-10-20
fix: use `String.valueOf`

### LINKED ISSUE #4726: `stringify` crash on `null`
```
pub def main(): String = stringify(null)
```

JonathanStarup:
Fixes #4726


mlutze:
Would it be simpler just to check for null before checking the type?

magnus-madsen:
> Would it be simpler just to check for null before checking the type?

Does not work if its a primitive?

mlutze:
> > Would it be simpler just to check for null before checking the type?
> 
> Does not work if its a primitive?

Ok, fix `isNull` and then use it. Maybe a little less simple.

JonathanStarup:
> > > Would it be simpler just to check for null before checking the type?
> > 
> > Does not work if its a primitive?
> 
> Ok, fix `isNull` and then use it. Maybe a little less simple.

I think thatd be better but we cant without expaning the minimum library. Im not sure what the least clumsy solution. I had really forgotten how much null suck

mlutze:
> > > > Would it be simpler just to check for null before checking the type?
> > > 
> > > 
> > > Does not work if its a primitive?
> > 
> > 
> > Ok, fix `isNull` and then use it. Maybe a little less simple.
> 
> I think thatd be better but we cant without expaning the minimum library. Im not sure what the least clumsy solution. I had really forgotten how much null suck

ah the dang minlib issue...

magnus-madsen:
Superseded by #4744

JonathanStarup:
@magnus-madsen I don't see how it's fixed by typematch, this is about not calling `obj.toString` and instead call `String.valueOf(obj)` and things like that

magnus-madsen:
> @magnus-madsen I don't see how it's fixed by typematch, this is about not calling `obj.toString` and instead call `String.valueOf(obj)` and things like that. Can you reopen this and ill remove the non-backend aspects?

Where else does the problem occur?

JonathanStarup:
> > @magnus-madsen I don't see how it's fixed by typematch, this is about not calling `obj.toString` and instead call `String.valueOf(obj)` and things like that. Can you reopen this and ill remove the non-backend aspects?
> 
> Where else does the problem occur?

Enums just call `value.toString` on their value which fails with `null`.  I think there were other places

magnus-madsen:
OK, I did not grasp that initially.

JonathanStarup:
@magnus-madsen ready to merge. There's some commented tests for null fix in the flix part

magnus-madsen:
Thanks!

PR #4732, opened 2022-10-20
fix: define isNull for primitives

### LINKED ISSUE #4727: `Object.isNull` crash on primitive values
```
pub def main(): Bool = Object.isNull(42)
```

mlutze:
fixes #4727

### DIFF
--- a/main/src/library/Object.flix
+++ b/main/src/library/Object.flix
@@ -20,7 +20,21 @@ namespace Object {
     ///
     pub def isNull(x: a): Bool =
         import static java.util.Objects.isNull(##java.lang.Object): Bool \ {};
-        isNull(x as ##java.lang.Object)
+        match reifyType a {
+            case ReifiedType.ReifiedUnit => false
+            case ReifiedType.ReifiedBool => false
+            case ReifiedType.ReifiedChar => false
+            case ReifiedType.ReifiedFloat32 => false
+            case ReifiedType.ReifiedFloat64 => false
+            case ReifiedType.ReifiedInt8 => false
+            case ReifiedType.ReifiedInt16 => false
+            case ReifiedType.ReifiedInt32 => false
+            case ReifiedType.ReifiedInt64 => false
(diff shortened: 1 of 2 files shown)


main/src/library/Object.flix
```diff
@@ -20,7 +20,21 @@ namespace Object {
...
+            case ReifiedType.ReifiedUnit => false
+            case ReifiedType.ReifiedBool => false
+            case ReifiedType.ReifiedChar => false
+            case ReifiedType.ReifiedFloat32 => false
+            case ReifiedType.ReifiedFloat64 => false
+            case ReifiedType.ReifiedInt8 => false
+            case ReifiedType.ReifiedInt16 => false
+            case ReifiedType.ReifiedInt32 => false
+            case ReifiedType.ReifiedInt64 => false
+            case ReifiedType.ReifiedBigInt => isNull(x as ##java.lang.Object)
```
JonathanStarup, line 33: Does this work :O I though it would break when isNull is called with cat2 types

magnus-madsen: We need to fix (something) so that we eliminate all the other cases.

magnus-madsen: Perhaps we need a dedicated pattern match construct?

mlutze: It doesn't work :D

mlutze: I tried something with it but it ended up needing GADTs...

magnus-madsen: Hardcoded into the compiler which monomorph guarantees to remove.

JonathanStarup: I think for now it might work by eta-expanding the isNull function and casting it
```
((y -> isNull(y)) as a -> Bool)(x)
```

magnus-madsen: Jeezus. We should fix this instead!

magnus-madsen:
Once #4744 is merged, this can be refreshed and merged. Then we can use `isNull` inside `stringify`.

PR #4733, opened 2022-10-21
feat: add IntUnaryOperator to translated functional interfaces

mlutze:
related to #4721


magnus-madsen:
@JonathanStarup Ready for backend.

PR #4730, opened 2022-10-20
First steps towards Channel lowering

paulbutcher:
This is nowhere near ready yet, but I wanted to get @JonathanStarup to take a look.

Jonathan, I think I'm some way towards getting `NewChannel`, `PutChannel` and `GetChannel` working (I've not yet looked at select). With these changes, the following program compiles successfully:

```
pub def main(): Unit \ IO = 
    let c = chan Int32 100;
    c <- 42;
    println(<- c)
```

But when I run it, I get:

```
Exception in thread "main" java.lang.VerifyError: Bad local variable type       
Exception Details:
  Location:
    Def%main%.invoke()LCont%Obj; @1346: iload
  Reason:
    Type 'IBoxed' (current frame, locals[59]) is not assignable to integer
```

So I guess that I need to do something about unboxing the value returned from `Concurrent/Channel.get`? But I'm not sure what this should be?

PR #4737, opened 2022-10-21
chore: Simplify Channel API

paulbutcher:
Simplify Channel API as discussed in #4730

### DIFF
--- a/main/src/library/Concurrent/Channel.flix
+++ b/main/src/library/Concurrent/Channel.flix
@@ -189,7 +189,7 @@ namespace Concurrent/Channel {
     /// Assumes the channel to be non-empty.
     ///
     @Internal
-    pub def getAndUnlock(c: Mpmc[a], locks: List[ReentrantLock]): a \ IO = {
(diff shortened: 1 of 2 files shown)


main/src/library/Concurrent/Channel.flix
```diff
@@ -185,11 +185,11 @@ namespace Concurrent/Channel {
         element
 
     ///
-    /// Retrieves an element from the channel `c`.
+    /// Retrieves an element from the channel `c` and unlocks `locks` after doing so.
     /// Assumes the channel to be non-empty.
     ///
     @Internal
-    pub def unsafeGet(c: Mpmc[a]): a \ IO = {
+    pub def getAndUnlock(c: Mpmc[a], locks: List[ReentrantLock]): a \ IO = {
```
magnus-madsen, line 192: Its still unsafe right? So I would prefer `unsafeGetAndUnlock`

paulbutcher: Agreed. Done

magnus-madsen:
Looks good!

PR #4738, opened 2022-10-21
chore: Channel lowering

paulbutcher:
This is a replacement for #4730.

This now successfully compiles the following:

```
def meow(c: Channel[String]): Unit \ IO = c <- "Meow!"; ()

def woof(c: Channel[String]): Unit \ IO = c <- "Woof!"; ()

def main(): Unit \ IO =
    let c1 = chan String 1;
    let c2 = chan String 1;
    spawn meow(c1);
    spawn woof(c2);
    select {
        case m <- c1 => m
        case m <- c2 => m
    } |> println
```

But when I run  it, I get:

```
Exception in thread "main" dev.flix.runtime.MatchError: Non-exhaustive match at ../flix-playground/src/Main.flix:10:5
```

Which I'm at a loss to explain: I definitely (as far as I can see?) have a wildcard match rule in the match expression I'm generating.

@magnus-madsen, @JonathanStarup, what am I missing?


paulbutcher:
Ah, I think I'm passing the wrong type to `Pattern.Wild`? Let's see if that fixes it...

paulbutcher:
So I tried this, but it made no difference:

```
val unreachableMatch = MatchRule(Pattern.Wild(Type.freshVar(Kind.Star, loc, text = Ast.VarText.FallbackText("wild")), loc), Expression.False(loc), unreachableExp)
```

I think I may need you to point me in the right direction?

magnus-madsen:
What does `--Xdebug` show?

magnus-madsen:
Is it necessary to desugar to a pattern match? Select does not permit patterns.

paulbutcher:
This is `main`, let me know if you need more:

```
def main(_unit%175203: Unit, ): Unit & Impure = 
    let c1%175204: Mpmc[String] = Concurrent/Channel.newChannel%175205(1i32, ));
  let c2%175206: Mpmc[String] = Concurrent/Channel.newChannel%175205(1i32, ));
  let _%176438: Unit = spawn Closure(main%177092, [c1%175204, ]);
  let _%176440: Unit = spawn Closure(main%177093, [c2%175206, ]);
  |>%175207*(let matchVar%%176442: (Int32, List[ReentrantLock]) = Concurrent/Channel.selectFrom([Concurrent/Channel.mpmcAdmin%175214(c1%175204, )),Concurrent/Channel.mpmcAdmin%175214(c2%175206, )),], true, ));
  branch {  jumpto case%176445
      case%176445:
      let innerElm%%176448: Int32 = matchVar%%176442[0];
      let innerElm%%176450: List[ReentrantLock] = matchVar%%176442[1];
      if (0i32 == innerElm%%176448) {
        let locks%175123%175208: List[ReentrantLock] = innerElm%%176450;
        if (false) {
          let m%175209: String = Concurrent/Channel.unsafeGetAndUnlock%175210(c1%175204, locks%175123%175208, ));
          m%175209
        } else {
          jumpto case%176446
        }
      } else {
        jumpto case%176446
      }
      case%176446:
      let innerElm%%176452: Int32 = matchVar%%176442[0];
      let innerElm%%176454: List[ReentrantLock] = matchVar%%176442[1];
      if (1i32 == innerElm%%176452) {
        let locks%175126%175211: List[ReentrantLock] = innerElm%%176454;
        if (false) {
          let m%175212: String = Concurrent/Channel.unsafeGetAndUnlock%175210(c2%175206, locks%175126%175211, ));
          m%175212
        } else {
          jumpto case%176447
        }
      } else {
        jumpto case%176447
      }
      case%176447:
      if (false) {
        unreachable!%175213())
      } else {
        jumpto default%176444
      }
      default%176444:
      MatchError
  }
  , Closure(main%177094, []), )
```

magnus-madsen:
Is it because selectFrom returns a pair?

magnus-madsen:
In all your MatchRules you probably want the guard to be true and not false.

paulbutcher:
Oh! I'm passing `false` as the guard. I guess I should be passing `true`? 🤦‍♂️

paulbutcher:
Snap!

paulbutcher:
```
Meow!
```

🎉🍾🥳

JonathanStarup:
> I'll investigate, but @JonathanStarup may have some idea what's going on?

Hmm that means that even though selectFrom says that a element is available on channel i, the unsafeGet could not get an element.

paulbutcher:
So I think I know what's going on. This is an instance where the channel being selected from is being created within the `select`, not being passed in from outside. The generated `main` function looks like this:

```
def main(_unit%175125: Unit, ): Bool & Impure =               
    let matchVar%%176498: (Int32, List[ReentrantLock]) = Concurrent/Channel.selectFrom([Concurrent/Channel.mpmcAdmin%175130(Concurrent/Channel/Timer.days(0i64, )), )),], true, ));
  branch {  jumpto case%176501
      case%176501:
      let innerElm%%176503: Int32 = matchVar%%176498[0];
      let innerElm%%176505: List[ReentrantLock] = matchVar%%176498[1];
      if (0i32 == innerElm%%176503) {
        let locks%175090%175126: List[ReentrantLock] = innerElm%%176505;
        if (false) {
          let _%175127: Unit = Concurrent/Channel.unsafeGetAndUnlock%175128(Concurrent/Channel/Timer.days(0i64, )), locks%175090%175126, ));
          true
        } else {
          jumpto case%176502
        }
      } else {
        jumpto case%176502
      }
      case%176502:
      if (false) {
        unreachable!%175129*()
      } else {
        jumpto default%176500
      }
      default%176500:
      MatchError
  }
```

Which is (sorta) equivalent to:

```
def main(): Bool \ IO =
    match selectFrom([mpmcAdmin(Concurrent/Channel/Timer.days(0i64))], true) {
        case (0, locks) =>
            unsafeGetAndUnlock(Concurrent/Channel/Timer.days(0i64), locks)
        case _ => unreachable!()
    }
```

Note that the timer channel is being created *twice* (so it's unsurprising that we get the error).

I guess I need to hoist the evaluation of the `chan` member of the `SelectChannelRule`s outside the `Match`.

JonathanStarup:
> I guess I need to hoist the evaluation of the chan member of the SelectChannelRules outside the Match.

Ah yes, didn't not think about that, but yes, the value begin matches might also have effects so it should not be duplicated.

paulbutcher:
Whew, that was a marathon. I can't say that @magnus-madsen didn't warn me.

Anyway, this seems to be working now 😊

paulbutcher:
Once this is merged, I'll create a followup PR to remove the now-unused AST elements and code generation

magnus-madsen:
> Whew, that was a marathon. I can't say that @magnus-madsen didn't warn me.
> 
> Anyway, this seems to be working now 😊

Great work on making it through!

magnus-madsen:
Ill review before/after Matt, but two things come to mind:

- I think it would be nice with a comment that demonstrates the code transformation (e.g. if you have this select then it is de-sugared to this code).
- The select case is hard to read. I wonder if we can introduce some helper functions to construct each part more incrementally.

paulbutcher:
> * I think it would be nice with a comment that demonstrates the code transformation (e.g. if you have this select then it is de-sugared to this code).

Yup, that would make sense, I'll put something together

> * The select case is hard to read. I wonder if we can introduce some helper functions to construct each part more incrementally.

I thought about this and decided against it because there are quite a lot of cross-dependencies between the different sections of the Select case (and little chance for reuse).

But I understand the desire. I'll see if I can come up with something that's easier to read and you can see if you think it's better than what's there.

magnus-madsen:
I suggest to follow the AST structure, so you could have `visitSelectCase` and `visitDefaultCase` etc. 
Try to look at the rest of the phase for inspiration.

magnus-madsen:
Here is a nice test case for you:

```
    @test
    def testSelectSideEffecting02(): Bool \ IO = {
        let c1 = chan Int32 10;
        let c2 = chan Int32 10;
        let c3 = chan Int32 10;

        select {
            case x <- {c3 <- 1; c1} => x + (<- c2) + (<- c3) == 6
            case x <- {c2 <- 2; c2} => x + (<- c1) + (<- c3) == 6
            case x <- {c1 <- 3; c3} => x + (<- c1) + (<- c2) == 6
        }
    }
```

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -555,24 +555,58 @@ object Lowering {
...
+    //     let ch2 = ?ch2;
+    //     match selectFrom([mpmcAdmin(c1), mpmcAdmin(c2)]) @ Static, false) {  // true if no default
+    //         case (0, locks) =>
+    //             let x = unsafeGetAndUnlock(c1, locks);
+    //             ?handlech1
+    //         case (1, locks) =>
+    //             let y = unsafeGetAndUnlock(c2, locks);
+    //             ?handlech2
+    //         case (-1, _) =>                                                  // omitted if no default
+    //             ?default                                                     //
```
magnus-madsen, line 606: Do we unlock in this case?

paulbutcher: No, in the non-blocking case, the unlock is done for us by `selectFrom`.

So, in the blocking case, the unlocking is done by `unsafeGetAndUnlock`. In the non-blocking case, the unlocking is done by `selectFrom`, and then again by `unsafeGetAndUnlock` (but this is a no-op).

I did think about whether we should have both `unsafeGet` and `unsafeGetAndUnlock`, and use the first in the non-blocking case, the second in the blocking case, but thought that this was probably getting too complicated. I'm happy to be persuaded otherwise though.

paulbutcher: Sorry - ignore what I wrote above. Let me try again...

paulbutcher: No, in the non-blocking case, the unlock is done for us by selectFrom, when the default case holds.

paulbutcher:
I've pushed changes which split the body of SelectChannel out into helper functions. I'm happy to go with this if you think it's an improvement (although personally I prefer the original version).

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -538,27 +555,73 @@ object Lowering {
       val ms = methods.map(visitJvmMethod)
       Expression.NewObject(name, clazz, t, pur, eff, ms, loc)
 
+    // New channel expressions are rewritten as follows:
+    //     chan Int32 10
+    // becomes:
+    //     newChannel[Int32](10)
```
magnus-madsen, line 561: How about we write:

becomes a call to the standard library function:
```
  Channel.newChannel(10)
```

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -538,27 +555,73 @@ object Lowering {
...
       val e = visitExp(exp)
       val t = visitType(tpe)
-      Expression.NewChannel(e, t, pur, eff, loc)
+      val newChannel = Expression.Def(Defs.ChannelNew, Type.mkImpureArrow(e.tpe, t, loc), loc)
+      Expression.Apply(newChannel, e :: Nil, t, pur, eff, loc)
 
+    // Channel get expressions are rewritten as follows:
+    //     <- c
+    // becomes:
+    //     get(c)
```
magnus-madsen, line 572: becomes a call to the standard library function:
```
  Channel.get(c)
```

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -538,27 +555,73 @@ object Lowering {
...
-      Expression.GetChannel(e, t, pur, eff, loc)
-
-    case Expression.PutChannel(exp1, exp2, tpe, pur, eff, loc) =>
+      val getChannel = Expression.Def(Defs.ChannelGet, Type.mkImpureArrow(e.tpe, t, loc), loc)
+      Expression.Apply(getChannel, e :: Nil, t, pur, eff, loc)
+
+    // Channel put expressions are rewritten as follows:
+    //     c <- 42
+    // becomes:
+    //     put(42, c)
```
magnus-madsen, line 582: becomes a call to the standard library function:
```
  Channel.put(42, c)
```

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -538,27 +555,73 @@ object Lowering {
...
+    //     let ch2 = ?ch2;
+    //     match selectFrom([mpmcAdmin(ch1), mpmcAdmin(ch2)]) @ Static, false) {  // true if no default
+    //         case (0, locks) =>
+    //             let x = unsafeGetAndUnlock(ch1, locks);
+    //             ?handlech1
+    //         case (1, locks) =>
+    //             let y = unsafeGetAndUnlock(ch2, locks);
+    //             ?handlech2
+    //         case (-1, _) =>                                                  // omitted if no default
+    //             ?default                                                     //
```
magnus-madsen, line 606: Maybe add a comment that unlock is handled by selectFrom?

(Separately: I find this a strange design decision, but maybe it is required?)

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -538,27 +555,73 @@ object Lowering {
...
+    //         case x <- ?ch1 => ?handlech1
+    //         case y <- ?ch2 => ?handlech2
+    //         case _ => ?default
+    //     }
+    // becomes:
+    //     let ch1 = ?ch1;
+    //     let ch2 = ?ch2;
+    //     match selectFrom([mpmcAdmin(ch1), mpmcAdmin(ch2)]) @ Static, false) {  // true if no default
+    //         case (0, locks) =>
+    //             let x = unsafeGetAndUnlock(ch1, locks);
```
magnus-madsen, line 600: Why not translate to `Expression.Stm`?

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -538,27 +555,73 @@ object Lowering {
...
+    //         case _ =>
+    //             unreachable!()
+    //     }
     case Expression.SelectChannel(rules, default, tpe, pur, eff, loc) =>
       val rs = rules.map(visitSelectChannelRule)
       val d = default.map(visitExp)
       val t = visitType(tpe)
-      Expression.SelectChannel(rs, d, t, pur, eff, loc)
+
+      val channels = (rs map { case SelectChannelRule(_, c, _) => (mkLetSym("chan", loc), c) }).toArray
```
magnus-madsen, line 615: We tend to never use arrays because they are mutable. Is there a particular reason you need an array over a List or Seq?

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1242,6 +1309,75 @@ object Lowering {
     mkTag(Enums.HeadTerm, s"App$arity", innerExp, Types.HeadTerm, loc)
   }
 
+  /**
+    * Make the array of MpmcAdmin objects which will be passed to `selectFrom`
+    */
+  private def mkChannelAdminArray(rs: List[SelectChannelRule], channels: Array[(Symbol.VarSym, Expression)], pur: Type, eff: Type, loc: SourceLocation): Expression = {
```
magnus-madsen, line 1315: You pass in `pur` and `eff`, but as far as I can tell everything return here should be pure, i.e. we can just use Type.Pure and Type.Empty (I think its called).

magnus-madsen: (In fact its wrong to use `pur` and `eff` if you apply a function that is pure (i.e. `Type.mkPureArrow`).

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1242,6 +1309,75 @@ object Lowering {
...
+    val admins = rs.zipWithIndex map {
+      case (SelectChannelRule(_, c, _), i) =>
+        val admin = Expression.Def(Defs.ChannelMpmcAdmin, Type.mkPureArrow(c.tpe, Types.ChannelMpmcAdmin, loc), loc)
+        Expression.Apply(admin, List(Expression.Var(channels(i)._1, c.tpe, loc)), Types.ChannelMpmcAdmin, pur, eff, loc)
```
magnus-madsen, line 1319: Instead of `channels(i)._1`, could we zip together `rs`, `channels`, and `zipWithIndex`?

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1242,6 +1309,75 @@ object Lowering {
...
+    val locksType = Types.mkList(Types.ConcurrentReentrantLock, loc)
+
+    val selectRetTpe = Type.mkTuple(List(Type.Int32, locksType), loc)
+    val selectTpe = Type.mkImpureUncurriedArrow(List(adminArray.tpe, Type.Bool), selectRetTpe, loc)
+    val select = Expression.Def(Defs.ChannelSelectFrom, selectTpe, loc)
+    val blocking = default match {
+      case Some(_) => Expression.False(loc)
+      case None => Expression.True(loc)
+    }
+    Expression.Apply(select, List(adminArray, blocking), selectRetTpe, pur, eff, loc)
```
magnus-madsen, line 1337: Same comment about `pur` and `eff` being "wrong" (here they should be `Impure`).

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1242,6 +1309,75 @@ object Lowering {
...
+    * Construct a sequence of MatchRules corresponding to the given SelectChannelRules
+    */
+  private def mkChannelCases(rs: List[SelectChannelRule], channels: Array[(Symbol.VarSym, Expression)], retTpe: Type, pur: Type, eff: Type, loc: SourceLocation)(implicit flix: Flix): List[MatchRule] = {
+    val locksType = Types.mkList(Types.ConcurrentReentrantLock, loc)
+
+    rs.zipWithIndex map {
+      case (SelectChannelRule(sym, chan, exp), i) =>
+        val locksSym = mkLetSym("locks", loc)
+        val pat = Pattern.Tuple(List(Pattern.Int32(i, loc), Pattern.Var(locksSym, locksType, loc)), retTpe, loc)
+        val getTpe = chan.tpe match {
```
magnus-madsen, line 1350: @mlutze Is this safe here or not?

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1242,6 +1309,75 @@ object Lowering {
...
+        val getExp = Expression.Apply(get, List(Expression.Var(channels(i)._1, chan.tpe, loc), Expression.Var(locksSym, locksType, loc)), getTpe, pur, eff, loc)
+        val e = Expression.Let(sym, Ast.Modifiers.Empty, getExp, exp, exp.tpe, pur, eff, loc)
+        MatchRule(pat, Expression.True(loc), e)
+    }
+  }
+
+  /**
+    * Construct additional MatchRules to handle the (optional) default and unreachable cases
+    */
+  private def mkChannelDefaultCases(default: Option[Expression], t: Type, retTpe: Type, pur: Type, eff: Type, loc: SourceLocation)(implicit flix: Flix): List[MatchRule] = {
```
magnus-madsen, line 1364: ~I think we should split this function into two. Earlier you have:~
```
cases ++ extraCases
```

~I propose to make this:~

```
cases ++ defaultCase ++ unreachableCase
```

~then I would split `mkChannelDefaultCases` into two functions. One that just returns a list of length zero or one.  And a function `mkUnreachableCase()`.~

magnus-madsen: Strike all that. I now realize that the `Simplifier` will insert `MatchError`. Hence just construct a pattern match that is not exhaustive. The `Simplifier` will take care of it. 

Thus this code becomes much simpler 👍

magnus-madsen (commented):
I added lots of comments.

They may seem like nitpicks, but I think they are worth it.

One serious issue is that we have to be much more careful with `pur` and `eff`. 

On the plus side, I found a significant simplification with the unreachable case.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
magnus-madsen: Obviously because `x` is used inside the exp.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
JonathanStarup: its to avoid unlockLocks as a public function

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
JonathanStarup: I think its just lefter code in channels. It could be changed in Channels.flix

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
magnus-madsen: OK, but in Lowering we dont have to respect visibility any more. But yeah, IDK.

JonathanStarup: Oh, so everything in Channel could be private?

magnus-madsen: Yes, expect the Redundancy phase will complain :D (Cant win, cant ya :))

JonathanStarup: Ah, just mark it as `@test`. 

Actually why doesn't `@Internal` just make it okay to be unused like `@test`?

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
paulbutcher: It's because I wanted to access it by numeric index, i.e. channels(i).

I could have achieved the same by zipping (as per later comment) but thought that this was simpler/easier to understand.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
magnus-madsen: It could.

magnus-madsen: We could also add `@Unused`

paulbutcher:
I've pushed changes which I think address all the comments.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -538,27 +553,75 @@ object Lowering {
...
       val rs = rules.map(visitSelectChannelRule)
       val d = default.map(visitExp)
       val t = visitType(tpe)
-      Expression.SelectChannel(rs, d, t, pur, eff, loc)
+
+      val channels = rs map { case SelectChannelRule(_, c, _) => (mkLetSym("chan", loc), c) }
+      val adminArray = mkChannelAdminArray(rs, channels, loc)
+      val selectExp = mkChannelSelect(adminArray, d, loc)
+      val cases = mkChannelCases(rs, channels, selectExp.tpe, pur, eff, loc)
+      val extraCases = mkChannelDefaultCase(d, t, selectExp.tpe, pur, eff, loc)
```
magnus-madsen, line 619: extraCases -> defaultCase

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1242,6 +1309,71 @@ object Lowering {
...
+        }
+        val get = Expression.Def(Defs.ChannelUnsafeGetAndUnlock, Type.mkImpureUncurriedArrow(List(chan.tpe, locksType), getTpe, loc), loc)
+        val getExp = Expression.Apply(get, List(Expression.Var(channel._1, chan.tpe, loc), Expression.Var(locksSym, locksType, loc)), getTpe, pur, eff, loc)
+        val e = Expression.Let(sym, Ast.Modifiers.Empty, getExp, exp, exp.tpe, pur, eff, loc)
+        MatchRule(pat, Expression.True(loc), e)
+    }
+  }
+
+  /**
+    * Construct additional MatchRule to handle the (optional) default case
```
magnus-madsen, line 1362: Add comment.

NB: Does not need to unlock because that is handled inside `Concurrent/Channel.selectFrom`.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1242,6 +1309,71 @@ object Lowering {
...
+    */
+  private def mkChannelDefaultCase(default: Option[Expression], t: Type, retTpe: Type, pur: Type, eff: Type, loc: SourceLocation)(implicit flix: Flix): List[MatchRule] = {
+    val locksType = Types.mkList(Types.ConcurrentReentrantLock, loc)
+    default match {
+      case Some(d) =>
+        val locksSym = mkLetSym("locks", loc)
+        val pat = Pattern.Tuple(List(Pattern.Int32(-1, loc), Pattern.Var(locksSym, locksType, loc)), retTpe, loc)
```
magnus-madsen, line 1369: The locks is not used here, so could we use a wildcard pattern instead?

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1242,6 +1309,71 @@ object Lowering {
...
+    }
+  }
+
+  /**
+    * Construct additional MatchRule to handle the (optional) default case
+    */
+  private def mkChannelDefaultCase(default: Option[Expression], t: Type, retTpe: Type, pur: Type, eff: Type, loc: SourceLocation)(implicit flix: Flix): List[MatchRule] = {
+    val locksType = Types.mkList(Types.ConcurrentReentrantLock, loc)
+    default match {
+      case Some(d) =>
```
magnus-madsen, line 1367: d -> defaultExp

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1242,6 +1309,71 @@ object Lowering {
...
+  /**
+    * Construct a sequence of MatchRules corresponding to the given SelectChannelRules
+    */
+  private def mkChannelCases(rs: List[SelectChannelRule], channels: List[(Symbol.VarSym, Expression)], retTpe: Type, pur: Type, eff: Type, loc: SourceLocation)(implicit flix: Flix): List[MatchRule] = {
+    val locksType = Types.mkList(Types.ConcurrentReentrantLock, loc)
+
+    rs.zip(channels).zipWithIndex map {
+      case ((SelectChannelRule(sym, chan, exp), channel), i) =>
+        val locksSym = mkLetSym("locks", loc)
+        val pat = Pattern.Tuple(List(Pattern.Int32(i, loc), Pattern.Var(locksSym, locksType, loc)), retTpe, loc)
```
magnus-madsen, line 1349: I was confused about whether using `retTpe` for the pattern was correct because I thought retTpe was the return type of the select. But actually its the type of the select expression (which is (Int32, LocksType)).

magnus-madsen: I think it would be better if instead of taking retTpe as an argument we construct the argument here.

Thus I propose to introduce a helper function `mkPairExp(exp1, exp2, loc): Exp` which builds the `Pattern.Tuple` (and then uses the types of exp1 and exp2 to construct the right tuple type.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
magnus-madsen: Also, we can use mkPairPat here. And maybe add a mkWildPat

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1242,6 +1309,71 @@ object Lowering {
...
+        val getExp = Expression.Apply(get, List(Expression.Var(channel._1, chan.tpe, loc), Expression.Var(locksSym, locksType, loc)), getTpe, pur, eff, loc)
+        val e = Expression.Let(sym, Ast.Modifiers.Empty, getExp, exp, exp.tpe, pur, eff, loc)
+        MatchRule(pat, Expression.True(loc), e)
+    }
+  }
+
+  /**
+    * Construct additional MatchRule to handle the (optional) default case
+    */
+  private def mkChannelDefaultCase(default: Option[Expression], t: Type, retTpe: Type, pur: Type, eff: Type, loc: SourceLocation)(implicit flix: Flix): List[MatchRule] = {
```
magnus-madsen, line 1364: retTpe, pur, and eff can all be dropped (see below :))

magnus-madsen, line 1364: mkChannelDefaultCase -> mkSelectDefaultCase?

magnus-madsen (commented):
I added some more comments. Found two simplifications :)

paulbutcher:
Thanks @magnus-madsen, that's made things significantly nicer.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1242,6 +1309,71 @@ object Lowering {
...
+        MatchRule(pat, Expression.True(loc), e)
+    }
+  }
+
+  /**
+    * Construct additional MatchRule to handle the (optional) default case
+    * NB: Does not need to unlock because that is handled inside Concurrent/Channel.selectFrom.
+    */
+  private def mkSelectDefaultCase(default: Option[Expression], t: Type, loc: SourceLocation)(implicit flix: Flix): List[MatchRule] = {
+    val locksType = Types.mkList(Types.ConcurrentReentrantLock, loc)
```
magnus-madsen, line 1366: Unused :)

magnus-madsen:
I think this is looking good now. Lets see if @mlutze has any comments.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -768,6 +831,10 @@ object Lowering {
         case _ => tpe0
       }
 
+      // Special case for Channel[_], which is rewritten to Concurrent/Channel.Mpmc
+      case Type.Cst(TypeConstructor.Channel, loc) =>
```
mlutze, line 835: Looks like we should add `@EliminatedBy(Lowering.class)` annotation to `TypeConstructor.Channel`.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1242,6 +1309,70 @@ object Lowering {
...
   }
 
+  /**
+    * Make the array of MpmcAdmin objects which will be passed to `selectFrom`
+    */
+  private def mkChannelAdminArray(rs: List[SelectChannelRule], channels: List[(Symbol.VarSym, Expression)], loc: SourceLocation): Expression = {
+    val admins = rs.zip(channels) map {
+      case (SelectChannelRule(_, c, _), channel) =>
+        val admin = Expression.Def(Defs.ChannelMpmcAdmin, Type.mkPureArrow(c.tpe, Types.ChannelMpmcAdmin, loc), loc)
+        Expression.Apply(admin, List(Expression.Var(channel._1, c.tpe, loc)), Types.ChannelMpmcAdmin, Type.Pure, Type.Empty, loc)
```
mlutze, line 1319: I prefer to avoid `_1` syntax when possible. Can we just deconstruct the var directly?

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
mlutze: I think we need to call `eraseTopAliases` here.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
magnus-madsen: OK, sure

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
magnus-madsen: Right. This is important otherwise things could go badly.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
mlutze: Some of us still believe in that annotation :D

magnus-madsen:
@paulbutcher Matt has three proposed changes, once they are in, we can merge this.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
paulbutcher: Is there a test that I could write before making this change to confirm that a) it's a problem as it currently stands and b) that I've fixed it?

magnus-madsen: Yes, if you do something like:

type alias MyChannel = Channel[Int32]
def foo(c: MyChannel): ... /* use select on c)

paulbutcher: Hmm, so this passes. What am I missing?

```
    type alias MyChannel = Channel[Int32]

    @test
    def testSelectAliasedChannel(): Bool \ IO = {
        let c = chan Int32 1;

        def useChan(ch: MyChannel) : Bool =
            select {
                case x <- ch => x == 42
            };

        c <- 42;
        useChan(c)
    }
```

paulbutcher: Oh! But I am getting failures elsewhere (not sure why this wasn't happening before?)

Investigating...

paulbutcher: Ah! Because Magnus recently changed it to `Receiver` and `Sender`!

paulbutcher:
OK, I think we're there now.

magnus-madsen:
Yay!

PR #4740, opened 2022-10-21
feat: add `IntUnaryOperator` to respective function objects

JonathanStarup:
Fixes part of #4721.

Also fixes a bug where IntFunction did not properly use its argument

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
@@ -24,7 +24,7 @@ import ca.uwaterloo.flix.language.phase.jvm.ClassMaker.Final.{IsFinal, NotFinal}
 import ca.uwaterloo.flix.language.phase.jvm.ClassMaker.Visibility.{IsPrivate, IsPublic}
 import ca.uwaterloo.flix.language.phase.jvm.ClassMaker._
 import ca.uwaterloo.flix.language.phase.jvm.JvmName.MethodDescriptor.mkDescriptor
-import ca.uwaterloo.flix.language.phase.jvm.JvmName.{DevFlixRuntime, JavaLang, JavaUtil, IntFunction, MethodDescriptor, RootPackage}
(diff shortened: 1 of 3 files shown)


JonathanStarup:
@mlutze Must IntUnaryOperator be Impure?

mlutze:
> @mlutze Must IntUnaryOperator be Impure? is it the same for IntFunction?

I think so. If some java function returns a `IntUnaryOperator` then we have to assume it is impure.

JonathanStarup:
> > @mlutze Must IntUnaryOperator be Impure? is it the same for IntFunction?
> 
> I think so. If some java function returns a `IntUnaryOperator` then we have to assume it is impure.

Wait is it a doublesided equality? that sounds like it can crash the backend

mlutze:
> > > @mlutze Must IntUnaryOperator be Impure? is it the same for IntFunction?
> > 
> > 
> > I think so. If some java function returns a `IntUnaryOperator` then we have to assume it is impure.
> 
> Wait is it a doublesided equality? that sounds like it can crash the backend

Yes it is. Good thought...

magnus-madsen:
You mean if someone returns a IntUnaryOperator  to us?

mlutze:
> You mean if someone returns a IntUnaryOperator to us?

Yes, it gets translated by the typer to `Int -> Int \ IO`

JonathanStarup:
> You mean if someone returns a IntUnaryOperator to us?

yeah, if you call someone else's Function object, then backend will try to `obj.arg0 = 2` or whatever.

I don't think these should be proper equalities

magnus-madsen:
I think we can assume, for now, that we will not be getting Java closures back.

mlutze:
Maybe we can make it upcastable instead of a direct translation?

JonathanStarup:
Unless, we specialize function call on these specific types 😄(doesn't work for tailcall elimination though)

magnus-madsen:
> Maybe we can make it upcastable instead of a direct translation?

Let's not over complicate. When was the last time you GOT a Java function back from somewhere?

JonathanStarup:
> Maybe we can make it upcastable instead of a direct translation?

Yeah maybe this. Maybe we can abuse the unify method with an expected/required type?

> I think we can assume, for now, that we will not be getting Java closures back.

I just feel like we will all forget about this in a week 😄

PR #4741, opened 2022-10-21
fix: Calling a constructor that takes an array argument

paulbutcher:
This fixes calling a constructor which takes an array argument (calling methods and static methods with array arguments already works).

The problem was that the code to put the arguments on the stack for constructors was different from the equivalent code for methods and static methods. This PR pulls that code out into a separate method to ensure that it's always done the same way.


JonathanStarup:
This looks fine, it seems like a fix that wasn't put at all the copied sites

Issue #4739, opened 2022-10-21
Improved ReifyType


mlutze:
Option 1:

We can change reifyType to be essentially a type-case expression:

```
typematch x {
    case y: Int32 => ...
    case y: Array[String, Static] => ...
    case y: Map[String, Bool] => ...
    case _ => ...
}
```

Then monomorph replaces the expression with 
```
let y = x;
...
```

The nice thing is that here we don't have to cast anything anymore. Plus we can match on any type we want.
But, erased types might cause complications; we can likely get around this with safety checks.
Matching on any type you want might lead to unprincipled design.

Option 2:

Add a parameter to reifyType cases;
```
pub enum ReifiedType2 {
    case ErasedType(##java.lang.Object),
    case ReifiedUnit(Unit),
    case ReifiedBool(Bool),
    case ReifiedChar(Char),
    case ReifiedFloat32(Float32),
    case ReifiedFloat64(Float64),
    case ReifiedInt8(Int8),
    case ReifiedInt16(Int16),
    case ReifiedInt32(Int32),
    case ReifiedInt64(Int64),
    case ReifiedBigInt(BigInt),
    case ReifiedString(String),
    case ReifiedArray(ReifiedType, ##java.lang.Object)
}
```

And do magic replacements as in Option 1.

```
typematch x {
    case ReifiedInt32(y) => ...
    case ReifiedArray(String, y) => ...
    case _ => ...
}
```

Some casts are still needed since `ReifiedArray` can be of any type.

magnus-madsen:
I think we need option 1 with the requirement that its eliminated in monomorph.

magnus-madsen:
Plus:
- The requirement there is a default cause (because there are infinitely many types (?)).
- The ability to use wildcards `Array[t, _]`.
- (Unsure about overlapping cases. Maybe we don't care for now).
- Implemented an entirely new construct `TypeMatch` with `TypeMatchCase` etc.

magnus-madsen:
(Clever use of this construct can probably loop the compiler; but this feature is intended for internal use anyway.)

magnus-madsen:
@mlutze Do you want to sketch an implementation by showing how TypedAst would be extended?

magnus-madsen:
What should happen for:

```
typematch Nil
```

?

JonathanStarup:
What is the usecase for arbitrary types? It seems overly complicated for current usecases

magnus-madsen:
[...]

PR #4744, opened 2022-10-21
feat: introduce typematch expression

mlutze:
related to #4739


mlutze:
Next step is to convert reifyType everywhere to use this new expression instead.

magnus-madsen:
> Next step is to convert reifyType everywhere to use this new expression instead.

In a separate PR?

mlutze:
> > Next step is to convert reifyType everywhere to use this new expression instead.
> 
> In a separate PR?

Yes.

> RFR?

Yes.

main/test/flix/Test.Exp.MatchType.flix
```diff
@@ -0,0 +1,287 @@
+namespace Test/Exp/ReifyType {
```
magnus-madsen, line 1: These test cases don't seem to use much polymorphism. I think we want to call functions that take type parameters and reflect on those, right?

magnus-madsen: I think at minimum we should send the values through the identity function.

magnus-madsen: We want to see that the Monomorph correctly clones functions and eliminates cases with a healthy mix of different types.

main/test/flix/Test.Exp.MatchType.flix
```diff
@@ -0,0 +1,287 @@
...
+
+    @test
+    def matchTypeUnit01(): Bool = matchType () {
+        case _: Unit => true
+        case _: _ => false
+    }
+
```
magnus-madsen, line 8: I was also excepting to see a lot more pattern matches with wildcards.

magnus-madsen: And function types

main/src/ca/uwaterloo/flix/language/dbg/FormatExpression.scala
```diff
@@ -43,6 +43,7 @@ object FormatExpression {
     case TypedAst.Expression.Stm(exp1, exp2, _, _, _, _) => s"Stm($exp1, $exp2)"
     case TypedAst.Expression.Discard(exp, _, _, _) => s"Discard($exp)"
     case TypedAst.Expression.Match(exp, rules, _, _, _, _) => s"Match($exp, ${rules.mkString(", ")})"
+    case TypedAst.Expression.MatchType(exp, rules, _, _, _, _) => s"Match($exp, ${rules.mkString(", ")})"
```
magnus-madsen, line 46: MatchType

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
```diff
@@ -905,6 +905,16 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
       }
     }
 
+    def MatchType: Rule1[ParsedAst.Expression.MatchType] = {
+      def Rule: Rule1[ParsedAst.MatchTypeRule] = rule {
+          keyword("case") ~ WS ~ Names.Variable ~ optWS ~ ":" ~ optWS ~ Type ~ optWS ~ atomic("=>") ~ optWS ~ Stm ~> ParsedAst.MatchTypeRule
+      }
+
+      rule {
+        SP ~ keyword("matchType") ~ WS ~ Expression ~ optWS ~ "{" ~ optWS ~ oneOrMore(Rule).separatedBy(CaseSeparator) ~ optWS ~ "}" ~ SP ~> ParsedAst.Expression.MatchType
```
magnus-madsen, line 914: I rather liked the keyword `typematch`. I don't think we should have camel case keywords.

main/src/ca/uwaterloo/flix/language/phase/PatternExhaustiveness.scala
```diff
@@ -170,6 +170,11 @@ object PatternExhaustiveness {
         val rulesErrs = checkRules(exp, rules, root)
         expsErrs ::: rulesErrs
 
+      case Expression.MatchType(exp, rules, _, _, _, _) =>
+        val ruleExps = rules.map(_.exp)
+        val expsErrs = (exp :: ruleExps).flatMap(visitExp(_, root))
+        expsErrs
```
magnus-madsen, line 176: Kind of strange, but okay.

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -136,6 +137,17 @@ object Safety {
         visit(exp) :::
           rules.flatMap { case MatchRule(_, g, e) => visit(g) ::: visit(e) }
 
+      case Expression.MatchType(exp, rules, _, _, _, _) =>
+        // check whether the last case in the
```
magnus-madsen, line 141: comment :)

main/src/ca/uwaterloo/flix/language/phase/Monomorph.scala
```diff
@@ -394,6 +399,26 @@ object Monomorph {
         }
         Expression.Match(visitExp(exp, env0), rs, subst0(tpe), pur, eff, loc)
 
+      case Expression.MatchType(exp, rules, tpe, pur, eff, loc) =>
+        // make the tvars in `exp`'s type rigid
+        // so that Nil: List[x%123] can only match List[_]
+        val renv = exp.tpe.typeVars.foldLeft(RigidityEnv.empty) {
```
magnus-madsen, line 405: Why is `subst0` not applied to `exp.tpe` first?

magnus-madsen: What about `Nil: List[a]`? What patterns should match that? Only `List[_]` or more than that?

main/src/ca/uwaterloo/flix/language/phase/Monomorph.scala
```diff
@@ -394,6 +399,26 @@ object Monomorph {
...
+          case (acc, Type.Var(sym, _)) => acc.markRigid(sym)
+        }
+        rules.collectFirst {
+          case MatchTypeRule(sym, t, body0)
+            // use the non-strict substitution
+            // to allow free type variables to match with anything
+            if Unification.unifiesWith(subst0.nonStrict(exp.tpe), subst0.nonStrict(t), renv) =>
```
magnus-madsen, line 412: This does a check, but it does not compute a new substitution. What if I have:

```
typematch "abc" :: Nil {
  case x: List[a] => let y: a = head(x)
}
```

then `y` is never given the right type?

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -893,6 +893,19 @@ object Typer {
           resultEff = Type.mkUnion(eff :: guardEffs ::: bodyEffs, loc)
         } yield (constrs ++ guardConstrs.flatten ++ bodyConstrs.flatten, resultTyp, resultPur, resultEff)
 
+      case KindedAst.Expression.MatchType(exp, rules, loc) =>
+        val bodies = rules.map(_.exp)
+
+        for {
+          (constrs, tpe, pur, eff) <- visitExp(exp)
+          // unify each rule's variable with its type
+          _ <- seqM(rules.map(rule => unifyTypeM(rule.sym.tvar, rule.tpe, rule.sym.loc)))
```
magnus-madsen, line 902: I think we have to be careful with rigidity. Imagine if one writes:

```
typematch x {
  case v: List[a] => v == 123 :: Nil
}
```

main/src/ca/uwaterloo/flix/language/phase/Monomorph.scala
```diff
@@ -394,6 +399,26 @@ object Monomorph {
...
+              // Generate a fresh symbol for the let-bound variable.
+              val freshSym = Symbol.freshVarSym(sym)
+              val env1 = env0 + (sym -> freshSym)
+              val e = visitExp(exp, env0)
+              val body = visitExp(body0, env1)
+              Expression.Let(freshSym, Modifiers.Empty, e, body, subst0(tpe), pur, eff, loc)
```
magnus-madsen, line 418: It would be more precise/correct to use the purity/effect of the selected body (plus matchExp)?

main/src/ca/uwaterloo/flix/language/phase/Monomorph.scala
```diff
@@ -394,6 +399,26 @@ object Monomorph {
...
+
+        }.getOrElse(throw InternalCompilerException("Unexpected matchType failure."))
```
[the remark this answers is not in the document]
mlutze: Because of cloning, yes.

main/src/ca/uwaterloo/flix/language/phase/Monomorph.scala
mlutze: You're right. I need to apply non-strict `subst0` to get the renv.

The important thing that `subst0.nonStrict(exp.tpe)` can contain is free variables, like in your Nil example.

So `Nil: List[a]` where `a` is unbound (i.e. not in the function signature) will _only_ match with `List[_]`

main/src/ca/uwaterloo/flix/language/phase/Monomorph.scala
mlutze: My intention is not to introduce new type variables here. In your example you'd get an error saying `no such variable "a"`.

I think this is the right choice because by ensuring the type variables are previously bound, we can do things like:

```scala
def sameType(x: a, y: b): Bool = typematch x {
    case _: b => true
    case _: _ => false
}
```

main/src/ca/uwaterloo/flix/language/phase/Monomorph.scala
mlutze: Yes, good point. That will help with optimization in later phases.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
mlutze: `a` is necessarily in the `renv` here, so this will not compile. I'll add a test.

main/src/ca/uwaterloo/flix/language/phase/Monomorph.scala
mlutze: Either `a` is not in the signature -- résolution error -- or it is in the signature and has been monomorphed to something so it actually looks like `case x: List[MyCoolEnum]`

mlutze:
(RFR) @magnus-madsen

magnus-madsen:
Merged. Thanks!

magnus-madsen:
I guess now we can get rid of most of the other constructs?

mlutze:
> I guess now we can get rid of most of the other constructs?

Yes, I think so.

michelou:
What about this code ?!
```scala
enum Weekday {
    case Monday, Tuesday, Wednesday, Thursday, Friday, Saturday, Sunday
}

def f(d: Weekday): Bool = typematch d {
    case _: Weekday.Monday => true
    case _: _ => false
}
```
I get the following resolution error (tested with nightly build "flix-2022-10-27_cc36fbc.jar") : 
```text
-- Resolution Error --------------------------------- [...]\type-matches\src\Main.flix

>> Undefined type 'Weekday.Monday'.

6 |     case _: Weekday.Monday => true
                ^^^^^^^^^^^^^^
                type not found.

Tip: Possible typo or non-existent type?
```

mlutze:
@michelou The `typematch` expression is for matching on type only. In your example you are matching on Enum cases. In Flix (unlike e.g. Scala 2) these are not types in themselves but only cases of the `Weekday` type.

michelou:
@mlutze Here is a second (*valid*) code example : 
```scala
enum Bottle[a] {
    case Unopened,
    case Opened(a)
}
def f(s: Bottle[a]): Bool = typematch s {
    case _: Bottle[Int32] =>
        match s as Bottle[Int32] { case Opened(x) => x == 1 case _ => false }
    case _: _ =>
        false
}
def main(): Unit \ IO =
    f(Bottle.Unopened) |> println;    // false
    f(Bottle.Opened(1)) |> println;   // true
    f(Bottle.Opened(2)) |> println;   // false
    f(Bottle.Opened('a')) |> println  // false
```
It is ok to say that the `typematch` feature is primarily useful together with [polymorphic types](http://localhost:3000/enums.html?highlight=polymorphic#polymorphic-types), i.e. `enum Bottle[a]` ?

And &ndash; to keep it short &ndash; is the following semantics correct ?
| Flix | Scala |
|:----|:-----|
| `case _: Bottle[Int32]` | `s.isinstanceof[Bottle[Int32]]` |
| `s as Bottle[Int32]` | `s.asinstanceof[Bottle[Int32]]` |

magnus-madsen:
> It is ok to say that the typematch feature is primarily useful together with [polymorphic types](http://localhost:3000/enums.html?highlight=polymorphic#polymorphic-types), i.e. enum Bottle[a] ?

Yes, but I would go further. You should not really be using typematch. Its mostly intended for the compiler writes and less for end-users. Ordinary pattern matches would be the way to go.

> And – to keep it short – is the following semantics correct ?

Approximately yes.

Issue #4736, opened 2022-10-21
BigDecimal


mlutze:
It would be nice to have a `BigDecimal` type in the standard library.

See also #719

stephentetley:
I've taken a look at this as it's a feature I've wanted for a long time. 

I've got support working in the compiler, I need to do the tests and stdlib updates.

magnus-madsen:
Compiler support could be one PR with library support to follow.

stephentetley:
> Compiler support could be one PR with library support to follow.

Okay

First discussion point - what to use for the literal suffix? If we use "ff" as per `BigInt` 's "ii" there is a problem that "ff" might be read as a hex number, although `Float32` and `Float64` already have this problem.

magnus-madsen:
> > Compiler support could be one PR with library support to follow.
> 
> Okay
> 
> First discussion point - what to use for the literal suffix? If we use "ff" as per `BigInt` 's "ii" there is a problem that "ff" might be read as a hex number, although `Float32` and `Float64` already have this problem.

I never liked `ii`. I am open to suggestions.

stephentetley:
I'm largely agnostic?

Does @mlutze have an opinion as he does a lot of work on the parser?

magnus-madsen:
Do we know of any other languages that use some established suffix?

mlutze:
I'm partial to making it the default, but maybe it's too early for that.

magnus-madsen:
> 

BigDecimal, right (not BigInt). But in any case, we would want a suffix to be precise. 

And yes, I think its a bit too early.

stephentetley:
Looking at hex numbers they will always have an "0x" prefix, so the "ff" suffix for BigDecimal is probably not a headache for the parser. So I'm now leaning towards "ff" especially if it is always lowercase.

stephentetley:
Compiler changes are now submitted: 

PR #4748 - feat: add compiler support for BigDecimal (related to #4736)

stephentetley:
And stdlib changes submitted: 

PR #4752 - feat: add stdlib support for BigDecimal (related to #4736)

PR #4748, opened 2022-10-23
feat: add compiler support for BigDecimal (related to #4736)

stephentetley:
This PR adds compiler support for BigDecimal. The new type is called `BigDecimal` and the suffix for literals is "ff".

Support for `BigDecimal` is directly patterned after the existing support for `BigInt`, basically I ran the following two `greps` on the Flix compiler codebase and copied the code for `BigInt` changing it to support `BigDecimal`:

~~~ 

find . -name "*.scala" | xargs grep -n "BigInt"

find . -name "*.flix" | xargs grep -n "BigInt"

~~~

`BigInt` has support for logical operations and bit shifting which I didn't copy as `BigDecimal` is obviously a floating point type.

I've left out general support for `BigDecimal` in the stdlib but have included some support (e.g. `Eq` and `Boxable` instances) so that `BigDecimal` can support the same tests as `BigInt` in the `test.flix` test suite (a further PR will extend stdlib support).

There is one non- `BigDecimal` code change in `main/src/ca/uwaterloo/flix/language/dbg/FormatExpression.scala` where formatting `BigInt` used the `Int64` suffix which I took to be an error and corrected.

~~~

-    case TypedAst.Expression.BigInt(lit, _) => s"${lit}i64"
+    case TypedAst.Expression.BigInt(lit, _) => s"${lit}ii"

~~~


mlutze:
I haven't looked at this, but I know that `BigDecimal.equals` in Java compares scale as well as value. How is/should this be handled?

stephentetley:
@mlutze - good spot.

I'd lean towards leaving the Flix `Eq.eq` as Java's `equals` but adding a function to the `BigInteger` library for "numerically equals" i.e allowing different scales.

magnus-madsen:
There are three things I would like to discuss (post merge):

- The choice of suffix.
- What is `0.0ff / 0.0ff`?
- `BigDecimal` or `BigInt`?

magnus-madsen:
@stephentetley Merged! Thanks for all the hard work :) It must have been quite an adventure to modify the compiler.

magnus-madsen:
If you have time, would you be willing to update: https://doc.flix.dev/primitive-types.html (there is a nice edit link top right) :)

stephentetley:
> @stephentetley Merged! Thanks for all the hard work :) It must have been quite an adventure to modify the compiler.

Yes the only other time I've done it was a one-liner.

stephentetley:
> If you have time, would you be willing to update: https://doc.flix.dev/primitive-types.html (there is a nice edit link top right) :)

Yes I'll look into it after I've submitted the stdlib changes PR.

PR #4749, opened 2022-10-23
feat: remove `ToString` requirement from `Boxed`/`Boxable`

### LINKED ISSUE #4679: Refactor `Boxable/Boxed` to use `stringify` instead of `ToString`

JonathanStarup:
Fixes #4679


JonathanStarup:
@magnus-madsen It seems to me that it would make things simpler in the type if we lowered some things earlier? If we lowered fixpoint stuff before the typer then there would be no special treatment of these expressions. Why is the lowering so late currently?

magnus-madsen:
Two reasons: 

- We cannot desugar before because we need special types rules.
- We WANT to delay desugaring until after typing for reasons of IDE support.

In the future more and more desugaring should take place in Lowering, not in Weeder.

JonathanStarup:
> Two reasons:
> 
> * We cannot desugar before because we need special types rules.
> * We WANT to delay desugaring until after typing for reasons of IDE support.
> 
> In the future more and more desugaring should take place in Lowering, not in Weeder.

Hmm okay. Its just annoying that the typer hardcodes type constraints that hopefully match the typeconstraints after lowering

magnus-madsen:
Imagine the error messages if we did lowering before.

Issue #4554, opened 2022-09-13
Improvements to auto-complete


magnus-madsen:
- [x] Auto-complete `case T...` (with priority for enums in the same file)
- [x] Auto-generate complete pattern matches (?) (E.g. with `match Type` with template on `Type`)
- [x] Auto-complete predicate names `P...`.
- [x] Match should sort by source location for the tags- see ReifiedType
- [x] Auto-complete field names `r.X...`.
- [x] Auto-complete uses `use X.Y...`
- [x] Auto-complete imports `import ...` (with fqn class name)
- [x] Complete of unit functions.
- [x] Use hardcoded list of packages/classes from `classlist` (but heavily filtered).
- [x] Improve suggestions when completions should be provided (e.g. after `def _`).
- [x] Improve suggestions in context `[...]`.
- [x] Refactor `CompletionProvider` to not have a separate argument for `javaClasses`.
- [x] Built-in names for Arrays, Sender/Receiver, etc. (Remove Channel).
- [x] Remove `Int8` and other rarely used types from completions (?) / or give them lowest priority.
- [x] Improve prioritization (e.g. `Int32` is much more likely than `Identity`). 
- [x] Remove auto-completion of tags. 
- [ ] (Use `TypedAst.Root` modules for `use` completion.)
- [x] Idea: "A new type of hole that refers to an expression" and then provide auto-complete for things "relevant" to that expressions.
- [x] (Magnus) Prompt on `?` in VSCode extension.

Situations that should *NOT* trigger auto-complete:

- [x] Prefix is `def`
- [x] Prefix is `enum`
- [x] Prefix is `type alias`
- [x] Prefix is `class`
- [x] Prefix is `let` or `let*`
- [x] Prefix is `namespace`

paulbutcher:
Completion isn't currently namespace aware. So if I have:

```
namespace Foo {
    pub def frobnicate(x: Int32): Int32 = ???
}

namespace Bar {
    import Foo.frobnicate;

    pub def munge(): Int32 =
        frob⁁
}
```

And accept the completion, I get:

```
namespace Foo {
    pub def frobnicate(x: Int32): Int32 = ???
}

namespace Bar {
    import Foo.frobnicate;

    pub def munge(): Int32 =
        Foo.frobnicate(?x)
}
```

Where I probably don't want the `Foo.` prefix.

paulbutcher:
Completion doesn't currently handle functions that take no arguments well. Given:

```
def foo(): Unit = ???

[...]

PR #4750, opened 2022-10-23
feat: order match cases in declaration order in match completions

Daniel-Anker-Hermansen:
related to #4554


main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -613,7 +613,15 @@ object CompletionProvider {
     else {
       return None
     }
-    val (completion, _) = enm.cases.foldLeft("", 1)({
+    val (completion, _) = enm.cases.toList.sortWith{
```
magnus-madsen, line 616: SourceLocations are already ordered. I think that can be reused out of the box. (e.g. `sortBy(_.loc)`)

Daniel-Anker-Hermansen: That is correct. I missed that implementation.

PR #4751, opened 2022-10-23
Experiment: Investigate Re-Entracy and `MutList`

AStenbaek:
### PR
Comments with suggested typing to higher-order functions, which take functions with possible effects as an argument. These assume the option to "blacklist" certain effects, which in the case of `MutList.flix` restrict the use of `Write(r)` annotated functions.

---
## Discussion
---
### Motivating Examples
Consider the following setup:  
```scala
enum T(Int32) with ToString, Eq

def getList(r: Region[r]): MutList[T][r] \ Write(r) =
    let l = MutList.new(r);
    MutList.pushAll!(6::28::496::8128::33550336::Nil)(l);
    MutList.map(x -> T(x),l)
```
---
#### Example 1:
Using the `exists` function, as described above, we can run into problems if we provide a function that pops elements from the `MutList`:
```scala
region r {
    let l = getList(r);
    let f = x -> {
        let g = b -> { 
            let _ = MutList.pop!(l);
            b
        };
        match x {
            case T(n) => n == 0 |> g |> g |> g
        }
    };
    MutList.exists(f, l)
} |> println
```
resulting in a null pointer exception:
```
java.lang.NullPointerException: Cannot invoke "T%T.getValue()" because "<local2>" is null
        at Clo%main%176392.invoke(Unknown Source)
        at Cont%Bool.unwind(Cont%Bool)
        at MutList.Clo%exists%176371.invoke(Unknown Source)
        at Cont%Bool.unwind(Cont%Bool)
        at MutList.Def%exists%175460.invoke(Unknown Source)
        at Cont%Bool.unwind(Cont%Bool)
        at Def%main%.invoke(Unknown Source)
        at Cont%Obj.unwind(Cont%Obj)
        at Ns.m_main%(Unknown Source)
        at Main.main(Main)
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
        at java.base/java.lang.reflect.Method.invoke(Method.java:578)
        at ca.uwaterloo.flix.language.phase.jvm.Bootstrap$.mainFunction$1(Bootstrap.scala:79)
        at ca.uwaterloo.flix.language.phase.jvm.Bootstrap$.$anonfun$bootstrap$9(Bootstrap.scala:88)
        at ca.uwaterloo.flix.language.phase.jvm.Bootstrap$.$anonfun$bootstrap$9$adapted(Bootstrap.scala:88)
        at ca.uwaterloo.flix.tools.SimpleRunner$.run(SimpleRunner.scala:97)
        at ca.uwaterloo.flix.Main$.main(Main.scala:125)
        at ca.uwaterloo.flix.Main.main(Main.scala)
```
This result is due to the missing `T` data, defaulting to `Null` in JVM. We would ideally like to prevent such behavior from library abstractions such as `MutList`.

---
#### Example 2:
Based on the previous example consider the following:
```scala
region r {
    let l = getList(r);
    let f = x -> {
        match x {
            case T(n) => { MutList.clear!(l); n == 0 }
        }
    };
    MutList.exists(f, l)
} |> println
```
Resulting in:
```
false
```
While initially surprising, this is a result of `clear!` reassigning the `Array` reference of the `MutList` to a new empty `Array`, without modifications to the old data. Looking at the implementation of `exists`, we can see this data is dereferenced prior to `clear!` being called, which explains why it does not crash.

---
### Region Reentrancy
So how does the above two examples relate to the notion of reentrancy? 
We see in both cases, that a called function, reenters the region of its caller, 
which in turn allows the callee to break the assumptions of the caller.  
Based on this I will be thinking of this in terms of _region reentrancy_, and considering how we can use effects on regions to control this.

---
### Preventing Region Rentrancy
Let us consider the following application rule:
$$\frac{\Gamma \vdash e_1 : \tau_1 \xrightarrow{ef_1}\tau_2 \quad \Gamma \vdash e_2 : \tau_1 \setminus ef_2}{\Gamma \vdash e_1 \; e_2 : \tau_2 \setminus ef_1 \cup ef_2}$$
For now, let `Write(r)` and `Read(r)` reduce to the set $\{ \mathsf{r} \}$.
Based on examples above, we modify the rule as follows:
$$\frac{\Gamma \vdash e_1 : \tau_1 \xrightarrow{ef_1}\tau_2 \quad \Gamma \vdash e_2 : \tau_1 \setminus ef_2 \quad {\color{red}ef_1 \cap ef_2 = \varnothing}}{\Gamma \vdash e_1 \; e_2 : \tau_2 \setminus ef_1 \cup ef_2}$$
This rule would prevent the calls to both `pop!` and `clear` in the examples above, due to the overlap of regions with the caller.  
This rule is however too restrictive in practice, as it would also reject reading in the region.

___
#### Example 3:
Consider the following (reasonable?) example of the inner function having a `Read(r)` effect:
```scala
region r {
    let l = getList(r);
    let f = x -> match x { case T(n) => T(n*2) };
    MutList.exists(x -> MutList.exists(y -> x == f(y), l),l)
} |> println
```
This use would be rejected by the above disjointness criteria, which is not ideal.

---
### Current Status
I am currently looking into the possibilities for defining a typing rule:
$$\frac{\Gamma \vdash e_1 : \tau_1 \xrightarrow{ef_1}\tau_2 \quad \Gamma \vdash e_2 : \tau_1 \setminus ef_2 \quad {\color{red}ef_1 \cdot ef_2 \neq \bot}}{\Gamma \vdash e_1 \; e_2 : \tau_2 \setminus ef_1 \cdot ef_2}$$
Where $\cdot$ is intended to define an operation to determine validity of a given effect combination (where $\bot$ indicates invalid), such that:
$$\mathsf{Write(r)} \cdot \mathsf{Write(r)} = \bot$$
while for instance
$$\mathsf{Read(r)} \cdot \mathsf{Read(r)} = \mathsf{Read(r)}$$

Open questions I have related to this:
1. Can these kinds of constraints be expressed as a boolean unification problem?
2. What do we require of the $\cdot$ operation? How can it be defined?
   - Seems to me this might require us to infer if effects are external.
3. Assuming the above two. Can we define a general (maybe algebraic) structure we would expect effects to satisfy?
   - Such structure would of course need to be able to encapsulate our current way of thinking of effect.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
+++ b/main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
@@ -153,29 +153,6 @@ object WeederError {
   }
 
   /**
-    * An error raised to indicate an illegal array length.
-    *
-    * @param loc the location where the illegal array length occurs.
-    */
-  case class IllegalArrayLength(loc: SourceLocation) extends WeederError {
-    def summary: String = "Illegal array length"
-
-    def message(formatter: Formatter): String = {
-      import formatter._
-      s"""${line(kind, source.name)}
-         |>> Illegal array length.
-         |
-         |${code(loc, "illegal array length.")}
-         |""".stripMargin
-    }
-
-    /**
-      * Returns a formatted string with helpful suggestions.
-      */
-    def explain(formatter: Formatter): Option[String] = None
-  }
-
-  /**
     * An error raised to indicate that a negative atom is marked as fixed.
     *
     * @param loc the location where the illegal fixed atom occurs.
--- a/main/src/ca/uwaterloo/flix/language/phase/Weeder.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Weeder.scala
@@ -466,12 +466,11 @@ object Weeder {
           // NB: We only use the source location of the identifier itself.
           WeededAst.Expression.Ambiguous(qname, ident.loc).toSuccess
 
-        // Case 2: actually a record access/array length
+        // Case 2: actually a record access
         case ident :: fields =>
           // NB: We only use the source location of the identifier itself.
           val base = WeededAst.Expression.Ambiguous(Name.mkQName(prefix.map(_.toString), ident.name, ident.sp1, ident.sp2), ident.loc)
           fields.foldLeft(base: WeededAst.Expression) {
-            case (acc, field) if field.name == "length" => WeededAst.Expression.ArrayLength(acc, field.loc) // TODO NS-REFACTOR should use better location
             case (acc, field) => WeededAst.Expression.RecordSelect(acc, Name.mkField(field), field.loc) // TODO NS-REFACTOR should use better location
           }.toSuccess
       }


main/src/library/MutList.flix
```diff
@@ -354,6 +370,8 @@ namespace MutList {
     ///
     /// The result is a new mutable list.
     ///
+    /// pub def map(f: a -> b \ { ef - Write(r) }, v: MutList[a, r]): MutList[b, r] \ { ef, Read(r), Write(r) }
```
magnus-madsen, line 373: Map is more interesting. 

Here the (arbitrary?) design choices seems to have been that `map` allocates a new `MutList` but in the *same* region. 

Question 0: Argue for/against this design choice.

Question 1: What would `map` look like if it took a region parameter and allocated them new `MutList` in that region? Write the signature!

Question 2: Does your answer change depending on whether `r` may alias `r2`. How would you deal with this?

Question 3: Since `map` actually allocates a new array (l366) does that change your answer whether `f` can have an effect in `r`/`r2`?

main/src/library/MutList.flix
```diff
@@ -801,7 +835,9 @@ namespace MutList {
     ///
     /// Modifying `l` while using an iterator has undefined behavior and is dangerous.
     ///
-    pub def iterator(l: MutList[a, r]): Iterator[a, r] \ Write(r) =
+    /// Andreas: "Function reads from len, so I have added Read(r) effect.
```
magnus-madsen, line 838: Question 1: Since the length of an array is fixed (once created) should this operation be pure or impure?
Question 2: What does Flix do?

main/src/library/MutList.flix
```diff
@@ -801,7 +835,9 @@ namespace MutList {
...
+    ///
+    pub def iterator(l: MutList[a, r]): Iterator[a, r] \ { Read(r), Write(r) } =
```
magnus-madsen, line 840: Question 1: What does the interface of `Iterator` look like? 

Question 2: Given that we can create an iterator over MutList, what prevents re-entrancy by exploiting the iterator? Is that a property of (a) `MutList`, (b) `def iterator()`, (c) both, (d) something else?

main/src/library/MutList.flix
```diff
@@ -833,6 +871,8 @@ namespace MutList {
     ///
     /// Applies `f` to all the elements in `v`.
     ///
+    /// pub def foreachWithIndex(f: (Int32, a) -> Unit \ { ef - Write(r) }, v: MutList[a, r]): Unit \ { ef, Read(r) }
+    ///
```
magnus-madsen, line 875: I would like to see the rest of the file annotated. In particular, I am interested in if the private functions change anything.

(Maybe not, if none of them are higher-order, but what if they were?)

main/src/library/MutList.flix
```diff
@@ -833,6 +871,8 @@ namespace MutList {
...
     pub def foreachWithIndex(f: (Int32, a) -> Unit \ ef, v: MutList[a, r]): Unit \ { ef, Read(r) } =
         let MutList(ra, rl) = v;
         let a = deref ra;
```
magnus-madsen, line 878: A MutMap has an adjust function:

```
pub def adjust!(f: v -> v \ ef, k: k, m: MutMap[k, v, r]): Unit \ { ef, Read(r), Write(r) } with Order[k] =
```

Question 1: What does the corresponding function look like for `MutList`?
Question 2: Is it interesting?
Question 3: What if we change `f` to be `v -> Option[v]` where `None` would be interpreted as dropping that index in the `MutList`?

magnus-madsen (commented):
@AStenbaek I added some questions. I would like you to go over each of them and think and give me your reaction :)

magnus-madsen:
One more: What are your thoughts on `MutMap.mergeWithKey!`:

https://github.com/flix/flix/blob/master/main/src/library/MutMap.flix#L252-L255

magnus-madsen:
What are your thoughts on this program:

```
region r {
  let l = List.range(1, 100);
  let it1 = List.iterator(l, r);
  let it2 = List.iterator(l, r);
  Iterator.foreach(x -> Iterator.foreach(y -> ..., it1), it2)
|
```

main/src/library/MutList.flix
AStenbaek: 0. Making a new array of the new type makes sense for type-checking.  
It is sensible unless we _really_ care about it being in-place, which I feel would require some shenanigans to type correctly and not create issue for previous references to the underlying array.

1. 
    ```scala 
    pub def map(f: a -> b \ { ef - Write(r1) }, v: MutList[a, r1], r2: Region[r2]): MutList[b, r2] \ { { ef - Write(r1) } , Read(r1), Write(r2) }
    ```

2. At the face of it, I think above suggestion would work, even with aliasing.

3. `f` having an effect in `r2` should be fine, as I do not believe it can reference the memory we are working on in `r2`.

__Counter Question:__  
Since we are making a new array anyway. Why not make use the functional implementation on lists, whereby `f` can do whatever it wants without issue?

```scala
pub def map(f: a -> b \ ef, v: MutList[a, r]): MutList[b, r] \ { ef, Read(r), Write(r) } =
        let r = Scoped.regionOf(v);
        if (isEmpty(v))
            new MutList(r)
        else {
            let MutList(ra, rl) = v;
            let a = toList(deref ra);
            let l = deref rl;
            let b = List.map(f, a);
            List.toMutList(r, b)
        }
```

main/src/library/MutList.flix
AStenbaek: 1. Looking at `Iterator`, I am not sure _why_ the `new` method needs to have any effect on the region `r` at all. 
2. (a) `def iterator()` breaks the abstraction, so not really on `MutList` at this point.  
    (b) This function has `Write(r)` effect - there might be a discussion on if we should allocate `i` in the same region. Outside of that I can be convinced the iterator somewhat guards us from re-entrancy.

main/src/library/MutList.flix
AStenbaek: 1. Would say something along the lines of:  
    ```scala
    def adjust!(f: a -> a \ ef, i: Int32, l: MutList[a, r]): Unit \ { ef, Read(r), Write(r) } = {
        let g: (Int32, a) -> a \ ef = (j,x) -> {
            if (j == i) {
                f(x)
            } else {
                x
            }
        };
        MutList.transformWithIndex!(g, l)
    }
    ```
    This however make me realize, `MutList.transformWithIndex!` expects pure function, which might be somewhat odd - what if we wanted to read from some region and replace the element at `i` with that?
2. This could perhaps be an example of a benign write to the MutList
3. This would instead make it dangerous, similar to pop!

main/src/library/MutList.flix
AStenbaek: Looking through the file, most private functions appear in the shape of `loop`. I have not annotated all of these cases, so let me summarize the most interesting.

---
__General Case:__  
Consider the following from the code (`map` to be precise):
```scala
def loop(i) = { // Int32 -> Unit \ { ef, Read(r), Write(r) }
                if (i >= l)
                    ()
                else {
                    b[i] = f(a[i]); // { ef, Read(r), Write(r) }
                    loop(i + 1)
                }
            }
```
`loop` can be thought of as a higher order function under eta-expansion (binding `f` which is an argument to the `map` function) which would have the type `(a -> b \ { { ef - Write(r) }, Read(r) }) -> Int32 -> Unit \ { ef, Read(r), Write(r) }`.

---
__Continuation__  
Consider the `loop` from `foldRightWithCont`:
```scala
def loop(i) = { // { ef, Read(r) }
            if (i >= l)
                z
            else {
                f(a[i], _ -> loop(i + 1)) // { ef, Read(r) }
            }
        }
```
I do not think this case, in and of itself is that interesting, since we can require `f` not having a `Write(r)` effect. But imagine a `mapWithCont`(whatever that would be). If we base it on the previous example, `loop` would have a `Write(r)` effect - but how do we then prevent `f` from performing writes outside of invocation of `loop`?  
Doing the `{ ef - Write(r) }` would not cut it, since the function may invoke `loop`.

---
__Iterator:__  
Lastly, let us look at `iterator`
```scala
    pub def iterator(l: MutList[a, r]): Iterator[a, r] \ { Write(r) } =
        let r = Scoped.regionOf(l);
        let MutList(a, len) = l;
        let i = ref 0 @ r; // Write(r)
        let done = () -> deref i == deref len; // Read(r)
        let next = () -> { // { Read(r), Write(r) }
            let x = (deref a)[deref i]; // Read(r)
            i := deref i + 1;  // { Read(r), Write(r) }
            x 
        };
        Iterator(done, next)
```
I we discussed, I find `next` having a `{ Read(r), Write(r) }` effect to be somewhat misleading,
given the `Write(r)` does not actually affect the underlying `MutList`, but rather the `i` variable, which might as well be allocated in a different region.  
This consideration of course, does not depend on effect restriction, so should be expressible already.

PR #4752, opened 2022-10-23
feat: add stdlib support for BigDecimal (related to #4736)

stephentetley:
This PR adds support for the new `BigDecimal` type to the stdlib. 

* The `BigDecimal` module has been fleshed out. 

* The set of instances has been extended to match those of the other numeric types.

* Conversion functions to `BigDecimal` hasve been added to the other numeric types.

* A new test module for `BigDecimal` is added and tests for the conversion functions and instances have been added to existsing modules.

There is more code that should be wrapped from `java.math.BigDecimal` (particularly `setScale`) but the current set of functions now in `BigDecimal` is a useful start.


main/src/library/BigDecimal.flix
```diff
@@ -14,8 +14,50 @@
...
+    ///
+    /// Returns the larger of `x` and `y`.
+    ///
+    pub def max(x: BigDecimal, y: BigDecimal): BigDecimal = if (x >= y) x else y
+
+    ///
+    /// Returns the absolute value of `x`.
+    ///
+    pub def abs(x: BigDecimal): BigDecimal =
+        import java.math.BigDecimal.abs(): ##java.math.BigDecimal \ {};
```
magnus-madsen, line 39: Does it not work here to write the type as `BigDecimal` instead of `##java.math.BigDecimal`?

magnus-madsen:
I am merging this, but I think a search replace for `##java.math.BigDecimal` would be useful.

stephentetley:
> I am merging this, but I think a search replace for ##java.math.BigDecimal would be useful.

Thanks Magnus, I'll look a doing a find-replace tidy up PR tonight.

PR #4758, opened 2022-10-24
chore: find and replace ##java.math.BigDecimal to BigDecimal

stephentetley:
This PR tidies up the occurrences of `##java.math.BigDecimal` that slipped through the BigDecimal stdlib PR.

`##java.math.BigDecimal` in import signatures is replaced with `BigDecimal`.


magnus-madsen:
Thanks!

PR #4775, opened 2022-10-27
fix: Don't require that java.lang.Object methods are implemented

### LINKED ISSUE #4759: `checkObjectImplementation` should be relaxed about Object methods (see module ca.uwaterloo.flix.language.phase.Safety)
The `checkObjectImplementation` check in `ca.uwaterloo.flix.language.phase.Safety` is more strict than it should be. The `mustImplement` set of methods ideally should not include methods of the class `Object` as these do not need to be provided when implementing a functional interface - see the notice in the Java Language Spec: 

https://docs.oracle.com/javase/specs/jls/se18/html/jls-9.html#jls-9.8

Concretely this becomes an issue trying to define an instance of Comparator, the following code "works" but the mandatory implementation of `equals` is bogus: 

~~~

pub def makeComparator(cmp: (a, a) -> Int32 \ ef): Comparator[a] \ ef = 
    let compare1 = (o1, o2) -> cmp(o1 as a, o2 as a);
    let comparator1 = 
        new ##java.util.Comparator {
            def compare(_this: ##java.util.Comparator, t: ##java.lang.Object, u: ##java.lang.Object): Int32 = compare1(t, u)
            def equals(_this: ##java.util.Comparator, _t: ##java.lang.Object): Bool = false
            // equals should be discounted because it is an Oject method - currently it isn't.
        } as \ ef;
    Comparator(comparator1)

~~~

In Java you only need to provide an implementation of `compare`:

~~~

public static void test03() throws Exception {
    Comparator<String> cmp = new Comparator<String> () {
        public int compare(String s, String t) {
            return s.toLowerCase().compareTo(t.toLowerCase());
        }
    };
    System.out.println("one ?? ONE: " + cmp.compare("one", "ONE"));
    System.out.println("one ?? only: " + cmp.compare("one", "only"));
}

~~~    

In some cases you may want to provide implementations of methods found in `Object` but certainly for `Comparator` a non-bogus implementation of `equals` is not trivial. In terms of the check, methods of `Object` should be considered `mayImplement`

paulbutcher:
Fixes #4759

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Safety.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Safety.scala
@@ -641,8 +641,9 @@ object Safety {
     val implemented = flixMethods.keySet
 
     val javaMethods = getJavaMethodSignatures(clazz)
+    val objectMethods = getJavaMethodSignatures(classOf[Object]).keySet
     val canImplement = javaMethods.keySet
-    val mustImplement = canImplement.filter(m => isAbstractMethod(javaMethods(m)))
+    val mustImplement = canImplement.filter(m => isAbstractMethod(javaMethods(m)) && !objectMethods.contains(m))
 
     //
     // Check that there are no unimplemented methods.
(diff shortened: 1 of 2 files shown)


magnus-madsen:
Thanks!

PR #4778, opened 2022-10-27
feat: lsp completions for ```import (static) ... ```

Daniel-Anker-Hermansen:
related to #4554 
image
image
image
image


magnus-madsen:
Thanks!

magnus-madsen:
@Daniel-Anker-Hermansen There is a bug where the "this" object should be the first argument.

For example:

This is obviously wrong:

```
import java.lang.String.toUpperCase(): String \ IO;
```

You cannot call toUpperCase with an object. It should be

```
import java.lang.String.toUpperCase(String): String \ IO;
```

Daniel-Anker-Hermansen:
@magnus-madsen
```
pub def foo(): Unit \ IO =
    import java.lang.String.toUpperCase(): String \ IO;
    "Hello, world!" |> toUpperCase |> println
```
compiles, runs and produces the expected result while
```
pub def foo(): Unit \ IO =
    import java.lang.String.toUpperCase(String): String \ IO;
    "Hello, world!" |> toUpperCase |> println
```
does not compile.
If the receiver this is supposed to be typed explicitly this too is a bug in the compiler.

magnus-madsen:
Ah right! I don't even understand my own language! 😛

Issue #4755, opened 2022-10-24
Re-Engineer Channels


magnus-madsen:
This ticket supersedes previous tickets.

- [x] Merge #4738
- [x] Add `$CHANNEL_GET(exp)$` to internal syntax (see e.g. `$INT32_ADD$`). Use in `Channel.recv`.
- [x] Add `$CHANNEL_PUT(exp1, exp2)$` to internal syntax. Use in `Channel.send`.
- [x] Add `$CHANNEL_NEW(exp)` to internal syntax. Use in `Channel.{buffered,unbuffered}`.
- [x] Remove surface syntax `chan`
- [x] Remove surface syntax `<- exp` and replace by `Channel.recv(exp)` everywhere (library, tests...)
- [x] Remove surface syntax `exp1 <- exp2` and replace by `Channel.send(exp2, exp1)` everywhere (library, tests...)
- [x] Split the type `Channel[t]` into `Sender[t]` and `Receiver[t]`. Drop `Channel[t]`.
- [x] Support sub-millisecond sleep in `Thread.sleep`.
- [x] Update documentation.
- [x] Remove channel code downstream of `Lowering`.
- [x] Remove old channel code in `runtime/interpreter`.
- [x] Remove `MonoType.Channel` (if possible?)
- [x] Investigate if `elmType` can be dropped from `ChannelNew` expression
- [x] Update select syntax to `case x <- recv(exp)` (and optionally `case x <- Channel.recv(exp)`).
- [x] Decide if `recv` and `send` should have keyword status in VSCode
- [x] #4845
- [x] #4846
- [x] #4847
- [x] #4848
- [x] Rename `Immutable` to `Sendable`. 
- [x] Figure out where to check instances of `Imm`
- [x] #4850

Later (requires other things to progress):
- [ ] What effect should spawn have?

paulbutcher:
Also: remove channel code downstream of Lowering (made unnecessary by https://github.com/flix/flix/pull/4738

paulbutcher:
Grr, and as I've just discovered, we also need to rewrite the `par` code which is creating old-style channel ASTs during lowering 😟

paulbutcher:
We seem to have lost the suggestion of changing the channel creation syntax to `Channel.bounded` or `Channel.unbounded`?

(although, thinking about it, perhaps `Channel.buffered` and `Channel.unbuffered` would be better?)

paulbutcher:
(or is the intent that `chan Int32 10` returns a pair?)

[...]

PR #4784, opened 2022-10-28
chore: Switch par code to new style channels

paulbutcher:
Related to #4755


magnus-madsen:
Thinks looks OK to me; especially with the useful helpers. Ill merge it.

@jaschdoc Maybe you can take a second look, just in case.

PR #4786, opened 2022-10-28
chore: Remove ca.uwaterloo.flix.runtime.interpreter

paulbutcher:
Related to #4755


magnus-madsen:
Thanks!

magnus-madsen:
In `JvmName` we have:

```
  
  val DevFlixRuntime: List[String] = List("dev", "flix", "runtime")

  val ProxyObject: JvmName = JvmName(DevFlixRuntime, "ProxyObject")
  val SelectChoice: JvmName = JvmName(List("ca", "uwaterloo", "flix", "runtime", "interpreter"), "SelectChoice")

  // Deprecated: Should not be used in new code.
  val Channel: JvmName = JvmName(List("ca", "uwaterloo", "flix", "runtime", "interpreter"), "Channel")
```

I wonder if these can now also be removed?

PR #4788, opened 2022-10-28
fix: Class loader improvements

### LINKED ISSUE #4728: Class loader thoughts
While looking at package management, I've been looking into how the compiler uses class loaders. Here is what I believe the current situation to be:

* The compiler itself runs under the system class loader (this is the default for "normal" Java code).
* The compiled code runs under [`FlixClassLoader`](https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/jvm/FlixClassLoader.scala).
* Any jar files are loaded by [`ExternalJarLoader`](https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/util/ExternalJarLoader.scala).

Both `FlixClassLoader` and `ExternalJarLoader` delegate to the system class loader (this is the default if you don't provide an explicit parent when constructing a class loader, which neither of them do).

The consequence of this is that everything within the Flix compiler is available to compiled Flix code, e.g.:

```
pub def main(): Unit \ IO = 
    import static ca.uwaterloo.flix.util.Toucan.leToucan(): String \ {};
    println(leToucan())
```

```
flix> :eval main()                                                              
                                                                                
░░░░░░░░▄▄▄▀▀▀▄▄███▄░░░░░░░░░░░░░░
░░░░░▄▀▀░░░░░░░▐░▀██▌░░░░░░░░░░░░░
░░░▄▀░░░░▄▄███░▌▀▀░▀█░░░░░░░░░░░░░
░░▄█░░▄▀▀▒▒▒▒▒▄▐░░░░█▌░░░░░░░░░░░░
░▐█▀▄▀▄▄▄▄▀▀▀▀▌░░░░░▐█▄░░░░░░░░░░░
░▌▄▄▀▀░░░░░░░░▌░░░░▄███████▄░░░░░░
░░░░░░░░░░░░░▐░░░░▐███████████▄░░░
░░░░░le░░░░░░░▐░░░░▐█████████████▄
░░░░toucan░░░░░░▀▄░░░▐█████████████▄
░░░░░░has░░░░░░░░▀▄▄███████████████
░░░░░arrived░░░░░░░░░░░░█▀██████░░


()
```

Right now, this is necessary because the channel code needs to access classes such as `ca.uwaterloo.flix.runtime.interpreter.Channel`, but it causes problems:

* Firstly, it allows compiler internals to "leak" into Flix code, which might not be discovered until you create a Jar file and then discover that it's missing dependencies.
* More seriously, if a Fix program depends upon a Jar file which the compiler also depends upon, there will be uncontrolled conflicts.
  * This is a problem today, but will become a much larger problem when we integrate Maven, because Maven has *lots* of dependencies, many of which are in very common use (I've already hit this while using Jetty to write a webserver in Flix: Maven depends upon many of the same libraries as Jetty).

The solution is simple, I think: stop both `FlixClassLoader` and `ExternalJarLoader` from delegating to the system class loader.

Before we do so, we will need to move the channel code into the runtime, but I believe that that's the only code currently within the compiler that's legitimately accessed by compiled Flix code.

One other thing we should do: we're not currently setting the thread context class loader. This causes problems for libraries that rely upon it such as Log4j, so we should do so.

Does anyone see any issues with the above?

paulbutcher:
Fixes #4728

There are a couple of wrinkles to be aware of:

* We can't delete Global.java because the compiler has to load it in order to check its methods' signatures. So we special-case it in ExternalJarLoader.
* I've also special cased classes in flix.test when running tests. A true fix would probably be to build these into a separate jar file and pass this jar file to the compiler when building tests, but that would require significant reworking of the gradle build script so this seemed the pragmatic choice.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/Flix.scala
+++ b/main/src/ca/uwaterloo/flix/api/Flix.scala
@@ -42,7 +42,7 @@ object Flix {
 /**
   * Main programmatic interface for Flix.
   */
-class Flix(testing: Boolean = false) {
+class Flix {
 
   /**
     * A sequence of inputs to be parsed into Flix ASTs.
@@ -287,7 +287,7 @@ class Flix(testing: Boolean = false) {
   /**
     * A class loader for loading external JARs.
     */
-  val jarLoader = new ExternalJarLoader(testing)
+  val jarLoader = new ExternalJarLoader
 
   /**
     * Adds the given string `s` to the list of strings to be parsed.
--- a/main/src/ca/uwaterloo/flix/api/Flix.scala
+++ b/main/src/ca/uwaterloo/flix/api/Flix.scala
@@ -42,7 +42,7 @@ object Flix {
 /**
   * Main programmatic interface for Flix.
   */
-class Flix {
+class Flix(testing: Boolean = false) {
 
   /**
     * A sequence of inputs to be parsed into Flix ASTs.
@@ -287,7 +287,7 @@ class Flix {
   /**
     * A class loader for loading external JARs.
     */
-  val jarLoader = new ExternalJarLoader
+  val jarLoader = new ExternalJarLoader(testing)
 
   /**
     * Adds the given string `s` to the list of strings to be parsed.
--- a/main/src/ca/uwaterloo/flix/language/phase/jvm/FlixClassLoader.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/jvm/FlixClassLoader.scala
@@ -25,6 +25,9 @@ import scala.collection.mutable
   * A custom class loader to load generated class files.
   *
   * @param classes A map from internal names (strings) to JvmClasses.
+  * 
+  * We pass null as the `ClassLoader` parent to avoid it delegating to the system classloader
+  * (otherwise compiled Flix code has access to all classes within the compiler)
   */
 class FlixClassLoader(classes: Map[String, JvmClass])(implicit flix: Flix) extends ClassLoader(null) {
 
--- a/main/src/ca/uwaterloo/flix/language/phase/jvm/FlixClassLoader.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/jvm/FlixClassLoader.scala
@@ -26,7 +26,7 @@ import scala.collection.mutable
   *
   * @param classes A map from internal names (strings) to JvmClasses.
   */
-class FlixClassLoader(classes: Map[String, JvmClass])(implicit flix: Flix) extends ClassLoader {
+class FlixClassLoader(classes: Map[String, JvmClass])(implicit flix: Flix) extends ClassLoader(null) {
 
   /**
     * An internal cache of already loaded classes.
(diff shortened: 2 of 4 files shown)


main/src/ca/uwaterloo/flix/api/Flix.scala
```diff
@@ -42,7 +42,7 @@ object Flix {
 /**
   * Main programmatic interface for Flix.
   */
-class Flix {
+class Flix(testing: Boolean = false) {
```
magnus-madsen, line 45: I think this belongs in the Options class.

magnus-madsen: But really, now that I look through the other changes, I think we should hardcode the other classes inside the class loader.

main/src/ca/uwaterloo/flix/util/ExternalJarLoader.scala
```diff
@@ -29,4 +31,22 @@ class ExternalJarLoader extends URLClassLoader(Array.empty) {
...
+      super.findClass(name)
+    } catch {
+      case e: ClassNotFoundException => 
+        // Special case for dev.flix.runtime.Global
+        // This is never used at runtime, but we need to be able to load it at compile
+        // time in order to check method signatures
+        if (name == "dev.flix.runtime.Global")
+          super.findSystemClass(name)
+        // Special case for testing to allow us to load test classes
+        else if (name.startsWith(("flix.test.")))
```
magnus-madsen, line 46: I think this is much better than relying on the flag.

Ultimately, I would hope we could just use Javas library for tests.

paulbutcher:
@magnus-madsen Well that's embarrassing, that's not the code I *intended* to write at all (blush).

What I intended was that I should only special case the `flix.test` namespace if `testing` was true. As it currently stands the testing flag can be lost.

I didn't want to pollute the namespace with `flix.test` in the general case, because it's just an artefact of our build process. Whereas polluting it with `dev.flix.runtime.Global` is less of an issue because that's always there in production code too.

But in this case I can see that perhaps the pragmatic choice is to just special case `flix.test` all the time: the chances of it clashing with user code is pretty small.

Let me know what you think.

magnus-madsen:
Yes, also, if we used `dev.flix.test` (instead of `flix.test`) we own the namespace; no-one else should be using it.

magnus-madsen:
Lets go for the pragmatic choice; no flag.

main/src/ca/uwaterloo/flix/language/phase/jvm/FlixClassLoader.scala
```diff
@@ -26,7 +26,7 @@ import scala.collection.mutable
   *
   * @param classes A map from internal names (strings) to JvmClasses.
```
magnus-madsen, line 27: I think we need a comment to explain the null.

paulbutcher:
> Yes, also, if we used `dev.flix.test` (instead of `flix.test`) we own the namespace; no-one else should be using it.

OK, that makes sense. We should also move it out of `main/src` and into `main/test` (so it's not included in production builds). I'll do that in a separate PR.

paulbutcher:
@magnus-madsen should be ready to be merged now I hope

magnus-madsen:
Thanks!

PR #4791, opened 2022-10-29
chore: Move flix tests into dev.flix.test and out of production jar

paulbutcher:
Follows on from #4788

This both moves Flix tests from the flix.test namespace into the dev.flix.test namespace, and move test classes out of the production source tree (so they're not included in the production build).


magnus-madsen:
Seems reasonable. Thanks!

PR #4793, opened 2022-10-30
refactor: `Graph.flix` formatting

JonathanStarup:
* Reduce comments to 80char width
* put private methods after public methods
* normalize naming of query result `let res = query ...`
* write `query .. select ... from ... where ...` on separate lines

### DIFF
--- a/main/src/library/Graph.flix
+++ b/main/src/library/Graph.flix
@@ -13,7 +13,7 @@
  *  See the License for the specific language governing permissions and
  *  limitations under the License.
  */
-namespace Graph {
+namespace GGraph {
 
     ///
     /// Returns the transitive closure of the directed graph `g`.
@@ -21,8 +21,8 @@ namespace Graph {
     pub def closure(g: m[(t, t)]): Set[(t, t)] with Foldable[m], Boxable[t] =
         let edges = inject g into Edge;
         let res = query edges, nodes(), reachability()
-                select (src, dst)
-                from Reachable(src, dst);
+                  select (src, dst)
+                  from Reachable(src, dst);
         List.toSet(res)
 
     ///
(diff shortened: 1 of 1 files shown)


main/src/library/Graph.flix
```diff
@@ -20,57 +20,41 @@ namespace Graph {
...
-    /// Returns the vertices that are reachable from the `src` in the directed graph `g`.
+    /// Returns the vertices that are reachable from the `src` in the directed
+    /// graph `g`.
     ///
     pub def reachable(src: t, g: m[(t, t)]): Set[t] with Foldable[m], Boxable[t] =
         let edges = inject g into Edge;
-        let x = query edges, reachabilityFromSrc(src) select dst from Reachable(dst);
-        List.toSet(x)
+        let res = query edges, reachabilityFromSrc(src)
+                select dst
```
magnus-madsen, line 35: Indent off by one space?

JonathanStarup: fixed

JonathanStarup:
What is the Flix guideline to reduce to width of lines such as
```
pub def distance(src: { src :: t }, dst: { dst :: t }, g: m[(t, Int32, t)]): Option[Int32] with Foldable[m], Boxable[t] =
```

magnus-madsen:
> What is the Flix guideline to reduce to width of lines such as
> 
> ```
> pub def distance(src: { src :: t }, dst: { dst :: t }, g: m[(t, Int32, t)]): Option[Int32] with Foldable[m], Boxable[t] =
> ```

I think Flix is OK with lines more than 80 chars. I think 120 chars is fine, for example.

main/src/library/Graph.flix
```diff
@@ -13,64 +13,48 @@
  *  See the License for the specific language governing permissions and
  *  limitations under the License.
  */
-namespace Graph {
```
magnus-madsen, line 16: Unintended?

JonathanStarup: oops, workflow leak

JonathanStarup:
> 

Point stands! its 123 chars 😄

magnus-madsen:
RFM?

Issue #4746, opened 2022-10-22
Re-Engineer Namespaces


magnus-madsen:
The purpose of this ticket is to streamline the namespace mechanism. 

The current namespace mechanism is a bit idiosyncratic and has some strange corner cases.

By carefully refactoring, inspired by Rust's namespace mechanism, I think we can arrive in a great place.

We need these (breaking) changes:

- [ ] Declarations in a parent namespace are not visible in a sub namespace (but must be explicitly imported).
  - Detail: The root namespace is always visible (e.g. Prelude).
- [ ] Opening a new namespace will clear all `use`s and `import`s. 
- [ ] A name cannot refer to more than one entity:
  - [ ] It is illegal to shadow a declaration or type with a `use` or `import`.
  - [ ] It is illegal to shadow a `use` or `import` with a declaration or type.
- [ ] We unify the namespace of Flix types and Java types (i.e. uenv and ienv should ultimately be merged). 

I believe these changes can be implemented one-at-a-time with some changes to the standard library required.

JonathanStarup:
A question: If i have two files with the same namespace, what is the scope of imports and uses? Does the below example compile?
```
// File 1
namespace B { pub def function(x: Int32): Int32 = x }
namespace A { use B.function }
```
```
// File 2
namespace A {
    pub def example(x: Int32): Int32 = function(x)
}
```

JonathanStarup:
Another question: Now that uses/imports are namespace local, what does it mean to use/import something at the top level?

magnus-madsen:
> Another question: Now that uses/imports are namespace local, what does it mean to use/import something at the top level?

It means they are only visible at the top-level. In other words, it would only be useful for small programs.

magnus-madsen:
> A question: If i have two files with the same namespace, what is the scope of imports and uses? Does the below example compile?

Imports and uses are per file (like currently), hence your program does not compile.

In the future we may disallow declaring the same namespace twice, and then it would even more not compile.

JonathanStarup:
> > Another question: Now that uses/imports are namespace local, what does it mean to use/import something at the top level?
> 
> It means they are only visible at the top-level. In other words, it would only be useful for small programs.

Top level imports could also mean, import this for the whole file, no matter the namespace. This would be nice for very general imports like `ToString.toString`

magnus-madsen:
If you think about a namespace as more of a module then the nice thing about (having to write):

```
namespace Foo {
    use Bar.XYZ;
[...]

PR #4796, opened 2022-10-30
fix: Clear uses and imports when opening namespace

paulbutcher:
This implements the "Opening a new namespace will clear all uses and imports." portion of #4746

Note: the substantive changes are to Namer.scala and TestNamer.scala. All the remaining changes are fixes to make things work with this new setup.


main/src/ca/uwaterloo/flix/language/phase/Namer.scala
```diff
@@ -82,9 +82,9 @@ object Namer {
        * Namespace.
        */
       case WeededAst.Declaration.Namespace(ns, uses, imports, decls, loc) =>
-        flatMapN(mergeUseEnvs(uses, ns0, uenv0, ienv0, prog0)) {
```
magnus-madsen, line 85: Add comments:

// Note: Opening a new namespace clears all current imports and uses.
// Hence we pass empty import and use environments.

main/test/ca/uwaterloo/flix/language/phase/TestNamer.scala
```diff
@@ -20,6 +20,7 @@ import ca.uwaterloo.flix.TestUtils
 import ca.uwaterloo.flix.language.errors.NameError
 import ca.uwaterloo.flix.util.Options
 import org.scalatest.FunSuite
+import ca.uwaterloo.flix.language.errors.ResolutionError
```
magnus-madsen, line 23: If there are resolution errors then these tests should be in `TestResolver`.

Or perhaps something is wrong with the tests?

paulbutcher: Yes, this is where I originally put them, in fact, but it felt wrong to have tests that only pass because of code that's in the namer.

But I can put them back there if you prefer?

main/src/ca/uwaterloo/flix/language/phase/Namer.scala
```diff
@@ -82,9 +82,10 @@ object Namer {
...
+        // Note: Opening a new namespace clears all current imports and uses. // Hence we pass empty import and use environments.
```
magnus-madsen, line 85: Two lines :)

magnus-madsen:
@stephentetley @mlutze @JonathanStarup 

This one causes breakages because now uses/imports must be "under" a namespace to be visible. (See the changes in the PR).

main/src/ca/uwaterloo/flix/language/phase/Namer.scala
```diff
@@ -82,9 +82,11 @@ object Namer {
        * Namespace.
        */
       case WeededAst.Declaration.Namespace(ns, uses, imports, decls, loc) =>
-        flatMapN(mergeUseEnvs(uses, ns0, uenv0, ienv0, prog0)) {
+        // Note: Opening a new namespace clears all current imports and uses.
+        // Hence we pass empty import and use environments.
+        flatMapN(mergeUseEnvs(uses, ns0, UseEnv.empty, ImportEnv.empty, prog0)) {
```
magnus-madsen, line 87: Question: What is actually merged here if UseEvn and ImportEnv is empty? Can we omit the merge or no?

main/test/ca/uwaterloo/flix/language/phase/TestNamer.scala
magnus-madsen: Import can be removed :)

main/src/ca/uwaterloo/flix/language/phase/Namer.scala
paulbutcher: For `use`, I think we definitely need to stick with merging, because use environments *can* be merged, for example:

```
namespace A {
    pub enum Foo
    pub enum Bar
    pub def foo(): Foo = ???
    pub def bar(x: Foo): Bar = ???
}

namespace B {
    use A.Bar
    use A.bar

    pub def baz(): Bar = {
        use A.Foo;
        use A.foo;
        let x: Foo = foo();
        bar(x)
    }
}
```

So the use environment within `baz` is a combination of the use environment within the `B` namespace, plus the use environment within the function.

And given that we're merging use environments in this case, then it makes sense to create a "fresh" use environment by merging into an empty environment as the code does above, and also [here](https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/Namer.scala#L59).

For `import` the situation is different, however, because (as things currently stand, at least) use environments are now never actually merged.

But...

I still think that this is a mistake. I think that users will be confused that it's legitimate to use a function at namespace scope, but not to import a method at namespace scope. I.e.:

```
namespace A {
    use A.foo // this compiles
    import x.y.Z.foo(): Unit \ IO // this does not
```

And that it's legitimate to use a type at function scope, but not to import a type at function scope. I.e.:

```
def foo() =
    use A.Foo; // this compiles
    import x.y.Foo; // this does not
    ....
```

If we do unify the way that use and import work, then the same argument will apply to merging the UseEnv as the ImportEnv.

magnus-madsen: > So the use environment within baz is a combination of the use environment within the B namespace, plus the use environment within the function.

OK, yes. This is the key.

magnus-madsen:
Thanks!

PR #4799, opened 2022-10-30
chore: Add $CHANNEL_GET$ to internal syntax and use in Channel.recv.

paulbutcher:
This addresses the "Add $CHANNEL_GET(exp)$" task of #4755

I've also changed a number of the channel gets in the tests to use `Channel.recv` (not all of them, but enough to be confident that this works).


magnus-madsen:
Thanks!

PR #4801, opened 2022-10-31
chore: Add CHANNEL_PUT to internal syntax and use in Channel.send

paulbutcher:
This fixes the "Add CHANNEL_GET to internal syntax and use in Channel.send" task within #4755


magnus-madsen:
You mean channel put :)

magnus-madsen:
Thanks!

PR #4802, opened 2022-10-31
feat: add support for `IntConsumer` to closures

stephentetley:
*This is a draft PR and is not ready for merging.*

I'm trying to implement the `IntConsumer` functional interface but have got stuck. 

`IntConsumer` should be like `IntFunction` which we already have in place but it is a problem to implement `IntConsumer` because it's type and value appears to overlap with `Intfunction` in the backend code.

In `Resolver.scala` circa line 2697 I've managed to distinguish between `Unit` (i.e. Void) and `Object` with the `isBaseTypeUnit` test because we have more type information from `UnkindedType` at this point so I think this bit works though I imagine there are better ways of doing it.

In `BackendObjType.scala` circa line 221 - I've not been able to distinguish `Unit` from `Object` in the function `specialization`. Currently `IntConsumer` has this bogus match which overlaps with `IntProducer` and only fires because it is encountered first:

~~~

case (BackendType.Int32 :: Nil, BackendType.Reference(BackendObjType.JavaObject))  =>  // SPT temp
  Some(IntConsumer)

~~~

With this bogus match I can get a demo to run which shows the codegen for `IntConsumer` working (BackendObjType.scala circa line 196 the `case IntConsumer => ...`) but any `IntFunction` functional interfaces will then fail. 

Codegen itself needed a change to the return type for `mkDescriptor` and also `RETURN()` as the last instruction compared to the codegen for `IntFunction`.

The demo code is:

~~~

import java.util.function.IntConsumer

def main(): Unit \ IO = 
    println("Running testIntConsumer...");
    testIntConsumer();
    ()

def testIntConsumer(): Unit \ IO = 
    import static java.util.stream.IntStream.range(Int32, Int32): ##java.util.stream.IntStream \ IO;
    import java.util.stream.IntStream.forEach(IntConsumer): Unit \ IO;
    let st = ref 0;
    let stream0 = range(0, 9);
    forEach(stream0, i -> {st := i; ()});
    let ans = deref st;
    println("Ans: ${ans}")

~~~

As far as I can tell from stepping through with the debugger at this level `Unit` is represented as a `BackendType.Reference(BackendObjType.JavaObject)` and there isn't a way to get a more specific class name than `Object`, so I don't know how to implement a valid version of `specialization`.


JonathanStarup:
One quick way this could be solved is just to make every `Int32 -> JavaObject` have both the int function and the int consumer interface & function. Then hopefully, by the welltypedness of the program, the int consumer function is only called on the functions that do return unit.

magnus-madsen:
It is hard for me to understand. @JonathanStarup Can you explain what the underlying issue is?

stephentetley:
Thanks Jonathan - changing `specialization` to return both `IntFunction` and `IntConsumer` passes the Compiler Suite tests.

JonathanStarup:
> It is hard for me to understand. @JonathanStarup Can you explain what the underlying issue is?

In the backend `Int32 -> Unit` and `Int32 -> ##java.lang.Object` are both represented as the erased type `Int32 -> ##java.lang.Object`. This means that we can't find out what's what when the bytecode gets generated.

Solution is to put both interfaces on all `Int32 -> ##java.lang.Object` which means that all `Int32 -> Unit` functions will have the correct interface (and also more functions than that, which wont matter since they wouldn't be called).

JonathanStarup:
> > It is hard for me to understand. @JonathanStarup Can you explain what the underlying issue is?
> 
> In the backend `Int32 -> Unit` and `Int32 -> ##java.lang.Object` are both represented as the erased type `Int32 -> ##java.lang.Object`. This means that we can't find out what's what when the bytecode gets generated.
> 
> Solution is to put both interfaces on all `Int32 -> ##java.lang.Object` which means that all `Int32 -> Unit` functions will have the correct interface (and also more functions than that, which wont matter since they wouldn't be called). 

an addendum to this is that this erasure happens first in the collection of types in the backend phase itself, and the again (redundantly) on the arrow type. the first step could be removed and then we would see the unit return type correctly. this would probably duplicate a lot of bytecode though.

magnus-madsen:
Thanks! (I assume this was ready).

PR #4805, opened 2022-10-31
chore: Remove type from NewChannel

paulbutcher:
This is an attempt to remove the type from NewObject, as discussed in #4755

But this results in errors of the form:

```
- Test.Currying.flix *** FAILED ***
  Unable to compile. Failed with: 6 errors.
  
  -- Type Error -------------------------------------------------- RedBlackTree.flix
  
  >> Expected type: '?elm155501' but found type: 'Bool'.
  
  392 |                         spawn (chanL <- parExists((n - 2) / 2, f, a)); // We divide the rest of the threads as follows:
                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Which I'm not sure I understand.

I'd be grateful for a pointer as to what might be going on @magnus-madsen

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/KindedAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/KindedAst.scala
@@ -188,7 +188,7 @@ object KindedAst {
 
     case class NewObject(name: String, clazz: java.lang.Class[_], methods: List[KindedAst.JvmMethod], loc: SourceLocation) extends KindedAst.Expression
 
-    case class NewChannel(exp: KindedAst.Expression, tpe: Type, loc: SourceLocation) extends KindedAst.Expression
+    case class NewChannel(exp: KindedAst.Expression, loc: SourceLocation) extends KindedAst.Expression
 
     case class GetChannel(exp: KindedAst.Expression, tpe: Type.Var, loc: SourceLocation) extends KindedAst.Expression
 
--- a/main/src/ca/uwaterloo/flix/language/ast/KindedAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/KindedAst.scala
@@ -188,7 +188,7 @@ object KindedAst {
 
     case class NewObject(name: String, clazz: java.lang.Class[_], methods: List[KindedAst.JvmMethod], loc: SourceLocation) extends KindedAst.Expression
 
-    case class NewChannel(exp: KindedAst.Expression, loc: SourceLocation) extends KindedAst.Expression
+    case class NewChannel(exp: KindedAst.Expression, elemType: Type.Var, loc: SourceLocation) extends KindedAst.Expression
 
     case class GetChannel(exp: KindedAst.Expression, tpe: Type.Var, loc: SourceLocation) extends KindedAst.Expression
 
--- a/main/src/ca/uwaterloo/flix/language/ast/KindedAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/KindedAst.scala
@@ -188,7 +188,7 @@ object KindedAst {
 
     case class NewObject(name: String, clazz: java.lang.Class[_], methods: List[KindedAst.JvmMethod], loc: SourceLocation) extends KindedAst.Expression
 
-    case class NewChannel(exp: KindedAst.Expression, elemType: Type.Var, loc: SourceLocation) extends KindedAst.Expression
+    case class NewChannel(exp: KindedAst.Expression, elmType: Type.Var, loc: SourceLocation) extends KindedAst.Expression
 
     case class GetChannel(exp: KindedAst.Expression, tpe: Type.Var, loc: SourceLocation) extends KindedAst.Expression
 
--- a/main/src/ca/uwaterloo/flix/language/phase/Typer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Typer.scala
@@ -1564,7 +1564,7 @@ object Typer {
         for {
           (constrs, tpe, _, eff) <- visitExp(exp)
           _ <- expectTypeM(expected = Type.Int32, actual = tpe, exp.loc)
-          resultTyp <- liftM(Type.mkChannel(tpe, loc))
+          resultTyp <- liftM(Type.freshVar(Kind.Star, loc, text = FallbackText("channel")))
           resultPur = Type.Impure
           resultEff = eff
         } yield (constrs, resultTyp, resultPur, resultEff)
--- a/main/src/ca/uwaterloo/flix/language/phase/Typer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Typer.scala
@@ -1560,11 +1560,11 @@ object Typer {
         } yield (constrs.flatten, resultTyp, resultPur, resultEff)
 
 
-      case KindedAst.Expression.NewChannel(exp, declaredType, loc) =>
+      case KindedAst.Expression.NewChannel(exp, loc) =>
         for {
           (constrs, tpe, _, eff) <- visitExp(exp)
           _ <- expectTypeM(expected = Type.Int32, actual = tpe, exp.loc)
-          resultTyp <- liftM(Type.mkChannel(declaredType, loc))
+          resultTyp <- liftM(Type.mkChannel(tpe, loc))
           resultPur = Type.Impure
           resultEff = eff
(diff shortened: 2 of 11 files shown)


main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1560,11 +1560,11 @@ object Typer {
         } yield (constrs.flatten, resultTyp, resultPur, resultEff)
 
 
-      case KindedAst.Expression.NewChannel(exp, declaredType, loc) =>
+      case KindedAst.Expression.NewChannel(exp, loc) =>
         for {
           (constrs, tpe, _, eff) <- visitExp(exp)
           _ <- expectTypeM(expected = Type.Int32, actual = tpe, exp.loc)
-          resultTyp <- liftM(Type.mkChannel(declaredType, loc))
+          resultTyp <- liftM(Type.mkChannel(tpe, loc))
```
magnus-madsen, line 1567: Try `Type.mkChannel(tpe, loc)` -> `Type.mkFreshVar()` (or whatever its called).

The problem is that you are saying the type of the channel is the type of the expression (but the expression is of type int).

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -2253,11 +2253,11 @@ object Typer {
         val ms = methods map visitJvmMethod
         TypedAst.Expression.NewObject(name, clazz, tpe, pur, eff, ms, loc)
 
-      case KindedAst.Expression.NewChannel(exp, tpe, loc) =>
+      case KindedAst.Expression.NewChannel(exp, loc) =>
         val e = visitExp(exp, subst0)
         val pur = Type.Impure
         val eff = e.eff
-        TypedAst.Expression.NewChannel(e, Type.mkChannel(tpe, loc), pur, eff, loc)
+        TypedAst.Expression.NewChannel(e, Type.mkChannel(e.tpe, loc), pur, eff, loc)
```
magnus-madsen, line 2260: This seems OK

paulbutcher:
Hang fire on this: I think I've done something foolish. Looking.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
paulbutcher: Ah! Right...

paulbutcher:
So I've made the change you suggest in `inferExp`, and that's definitely made things better, but it's not there yet. I don't think the case in `reassembleExp` is correct yet because the type of `exp` will always be `Int32`, I think? But I'm not sure what I should replace it with.

Here's an example of the type of error I'm now getting:

```
- Test.Exp.Concurrency.Select.flix *** FAILED ***
  java.lang.ClassCastException: class Concurrent.Channel.Mpmc%Mpmc%Int32 cannot be cast to class Concurrent.Channel.IMpmc%Obj (Concurrent.Channel.Mpmc%Mpmc%Int32 and Concurrent.Channel.IMpmc%Obj are in unnamed module of loader ca.uwaterloo.flix.language.phase.jvm.FlixClassLoader @4351ed61)
 ```

File: main/src/ca/uwaterloo/flix/language/phase/Typer.scala (hunk 2 above)
Comment (magnus-madsen) [reply main/src/ca/uwaterloo/flix/language/phase/Typer.scala:2260]: Right, this is not OK.

Comment (magnus-madsen) [reply main/src/ca/uwaterloo/flix/language/phase/Typer.scala:2260]: Here is what we need to do:
- KindedAst.Expression.NewChannel(exp, loc) needs to have an extra argument `elmType` which is a Type.Var (constructed with Type.fresh... but in the previous phase (or maybe in an even earlier phase-- check some of the other uses). Anyway, it will have this `elmType`.
- Here in reassembly the type of the channel will be `mkChannel(subst0(elmType))`.

File: main/src/ca/uwaterloo/flix/language/phase/Typer.scala [hunk 3]
```diff
@@ -1560,11 +1560,11 @@ object Typer {
         } yield (constrs.flatten, resultTyp, resultPur, resultEff)
 
 
-      case KindedAst.Expression.NewChannel(exp, declaredType, loc) =>
+      case KindedAst.Expression.NewChannel(exp, loc) =>
         for {
           (constrs, tpe, _, eff) <- visitExp(exp)
           _ <- expectTypeM(expected = Type.Int32, actual = tpe, exp.loc)
-          resultTyp <- liftM(Type.mkChannel(declaredType, loc))
+          resultTyp <- liftM(Type.freshVar(Kind.Star, loc, text = FallbackText("channel")))
```
magnus-madsen, line 1567: Up here, instead of creating the variable, we will just pull it out of `Channel(exp, elmType, loc)` and unify with it. This will ensure that `elmType` has some inferred type later when we rebuild the AST.

paulbutcher:
Sorry @magnus-madsen, I feel like I'm being dense, but this still isn't there. I'm now getting this:

```
- Test.Exp.Concurrency.Buffered.flix *** FAILED ***
  Unable to compile. Failed with: 6 errors.
  
  -- Type Error -------------------------------------------------- main/test/flix/Test.Exp.Concurrency.Buffered.flix
  
  >> Equality is not defined on t. Define or derive an instance of Eq.
  
  183 |         Err("Goodbye World!") == Channel.recv(c)
                                      ^^
                                      missing Eq instance
```

and

```
- Test.Exp.Concurrency.NewChannel.flix
                                                                                
*** RUN ABORTED ***
  java.lang.VerifyError: Bad local variable type
Exception Details:
  Location:
    Test/Exp/Concurrency/Select/Def%testSelectDefault01.invoke()LCont%Bool; @137: iload
  Reason:
    Type 'dev/flix/runtime/Unit' (current frame, locals[4]) is not assignable to integer
  Current Frame:
    bci: @137
    flags: { }
    locals: { 'Test/Exp/Concurrency/Select/Def%testSelectDefault01', 'java/lang/Object', 'Concurrent/Channel/IMpmc%Obj', 'Tuple2%Int32%Obj', 'dev/flix/runtime/Unit' }
    stack: { }
```

magnus-madsen:
The first is expected because what is the type of `Ok`? This can be fixed by a type annotation.

For the second, I am not sure. I would need to see the program.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -2253,11 +2253,11 @@ object Typer {
         val ms = methods map visitJvmMethod
         TypedAst.Expression.NewObject(name, clazz, tpe, pur, eff, ms, loc)
 
-      case KindedAst.Expression.NewChannel(exp, tpe, loc) =>
+      case KindedAst.Expression.NewChannel(exp, elmType, loc) =>
         val e = visitExp(exp, subst0)
         val pur = Type.Impure
         val eff = e.eff
-        TypedAst.Expression.NewChannel(e, Type.mkChannel(tpe, loc), pur, eff, loc)
+        TypedAst.Expression.NewChannel(e, Type.mkChannel(elmType, loc), pur, eff, loc)
```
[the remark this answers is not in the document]
paulbutcher: D'oh! Of course. Sorry - I knew I was being dense.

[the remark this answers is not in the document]
magnus-madsen: No worries :) (but I am off to teach now, back later)

main/src/ca/uwaterloo/flix/language/ast/KindedAst.scala
```diff
@@ -188,7 +188,7 @@ object KindedAst {
 
     case class NewObject(name: String, clazz: java.lang.Class[_], methods: List[KindedAst.JvmMethod], loc: SourceLocation) extends KindedAst.Expression
 
-    case class NewChannel(exp: KindedAst.Expression, tpe: Type, loc: SourceLocation) extends KindedAst.Expression
+    case class NewChannel(exp: KindedAst.Expression, elemType: Type.Var, loc: SourceLocation) extends KindedAst.Expression
```
magnus-madsen, line 191: elemType -> elmType

paulbutcher:
Finally there I think (sorry it turned out to be more of a chore than it might have been!)

magnus-madsen:
Thanks! We got there in the end 👍

PR #4809, opened 2022-11-01
feat: add unreachable typematch test

mlutze:


mlutze:
> _ No description provided. _

Doesn't actually catch all unreachable stuff since something like this is not caught:
```scala
typematch x {
    case x: _ => ??? // everything goes here
    case x: List[Int32] => ???
}
```

But the only way I could think of doing that was a n^2 pairwise comparison, so I'm not bothering checking it.

main/test/ca/uwaterloo/flix/language/phase/TestRedundancy.scala
```diff
@@ -1278,4 +1278,51 @@ class TestRedundancy extends FunSuite with TestUtils {
...
+        |
+        |def f(_: a): Unit =
+        |    typematch E {
+        |        case _: E[a] => ()
+        |        case _: _ => ()
+        |    }
+        |""".stripMargin
+    val result = compile(input, Options.TestWithLibNix)
+    expectError[RedundancyError.UnreachableTypeMatchCase](result)
+  }
```
magnus-madsen, line 1327: I was expecting a new more tests given the complexities :)

magnus-madsen: few*

main/src/ca/uwaterloo/flix/language/errors/RedundancyError.scala
```diff
@@ -609,4 +609,30 @@ object RedundancyError {
     def explain(formatter: Formatter): Option[String] = None
   }
 
+  /**
```
magnus-madsen, line 612: Add over-all comment.

main/test/ca/uwaterloo/flix/language/phase/TestRedundancy.scala
```diff
@@ -1278,4 +1278,103 @@ class TestRedundancy extends FunSuite with TestUtils {
...
+  test("TestUnreachableTypeMatchCase.03") {
+    val input =
+      """
+        |enum E[_] {
+        |    case E
+        |}
+        |
+        |def f(_: a): Unit =
+        |    typematch E {
+        |        case _: E[a] => ()
```
magnus-madsen, line 1321: What's the reasoning here?

mlutze: See above. `a` is bound so it's rigid, acting like the `Int32` in the other example.

magnus-madsen:
I see a test failure.

mlutze:
> I see a test failure.

Yes, this is the problem I don't know how to solve yet :)

magnus-madsen:
> > I see a test failure.
> 
> Yes, this is the problem I don't know how to solve yet :)

Ah

magnus-madsen:
Could some subset of this be merged or?

mlutze:
What I have implemented seems to work for most cases, but I'm not confident in it. I'll close since I think the real solution will likely look different.

magnus-madsen:
OK

PR #4810, opened 2022-11-01
feat: introduce default function

mlutze:
fixes # 2428


magnus-madsen:
Reflection or Reflect?

mlutze:
Let's go with `Reflect`. It's more familiar to Java people and shorter.

main/src/library/Reflect.flix
```diff
@@ -0,0 +1,18 @@
+namespace Reflect {
+
+    @Unsafe
```
magnus-madsen, line 3: No comment on what this does? :)

https://docs.oracle.com/javase/specs/jls/se8/html/jls-4.html#jls-4.12.5

magnus-madsen:
Nice!

PR #4813, opened 2022-11-01
feat: add Long versions of translated functional interfaces (see issue #4721)

stephentetley:
This PR adds `LongFunction`, `LongPredicate`, `LongUnaryOperator` and `LongConsumer`to the supported functional interfaces.

The code is patterned after the Int functional interface code.

Tests are in the module Test.Java.Function.flix.


magnus-madsen:
Thanks!

Issue #4773, opened 2022-10-27
Refactorings post merge of `typematch`


magnus-madsen:
- [x] Rewrite `stringify` to use `typematch`
- [x] Rewrite `reifyEff` uses in the standard library to use `typematch`.
- [x] Remove `Reify` expression from compiler.
- [x] Remove `ReifyType` expression from compiler.
- [x] Remove `ReifyEff` expressiom from compiler.
- [x] Remove `ReifiedBool` from Prelude.
- [x] Remove `ReifiedType` from Prelude.
- [x] Add `typematch` keyword to VSCode.
- [x] Remove `default` keyword from VSCode.
- [x] https://github.com/flix/flix/issues/4772
- [x] https://github.com/flix/flix/issues/4726
- [x] https://github.com/flix/flix/issues/4727
- [x] Add tests for reflection on records and Datalog values.
- [x] https://github.com/flix/flix/issues/2428
- [ ] Reject type match for Channel and Schema types (they are lost).

JonathanStarup:
- [ ] Remove Boxable

magnus-madsen:
Can we remove Boxable?

JonathanStarup:
with match type can't you implement `def box(x: a): Boxed` and `def unbox(b: Boxed): a` instead of relying on the `Boxable` type class?

magnus-madsen:
> with match type can't you implement `def box(x: a): Boxed` and `def unbox(b: Boxed): a` instead of relying on the `Boxable` type class?

What happens to the Eq-Order constraints? That's the issue I think.

JonathanStarup:
> What happens to the Eq-Order constraints? That's the issue I think.

Ah yes, `def box(x: a): Boxed with Order[a], Eq[a]` and `def unbox(b: Boxed): a`

JonathanStarup:
Oh I see. I wasn't aware of that issue. Is this hard to solve or is it "just" a matter of passing the correct information along? Could the typer/monomorpher stash the needed information in the AST node?

mlutze:
Note about replacing reifyBool:

```
typematch ((x, y) -> {println(x); println(y)}) {
    case f: _ -> _ \ {} => true
    case f: _ => false
}
```

Returns `true` because of pure currying.

mlutze:
Also, `default` does not seem doable via typematch. What do we match on?

JonathanStarup:
> Also, `default` does not seem doable via typematch. What do we match on?

the return type?

[...]

PR #4818, opened 2022-11-02
refactor: use typematch for stringify function

mlutze:
related to #4773


magnus-madsen:
@mlutze What happened to the tests?

PR #4820, opened 2022-11-02
experiment: add defsym map

mlutze:


mlutze:
@magnus-madsen is this roughtly what you're looking for?

main/src/ca/uwaterloo/flix/language/ast/TypedAst.scala
```diff
@@ -33,7 +33,9 @@ object TypedAst {
                   typeAliases: Map[Symbol.TypeAliasSym, TypedAst.TypeAlias],
                   entryPoint: Option[Symbol.DefnSym],
                   sources: Map[Source, SourceLocation],
-                  classEnv: Map[Symbol.ClassSym, Ast.ClassContext])
+                  classEnv: Map[Symbol.ClassSym, Ast.ClassContext],
+                  defSyms: Map[List[String], Symbol.DefnSym]
```
magnus-madsen, line 37: I would put it first, and into one data structure  (i.e. one data structure for both defs and enums and ...)

magnus-madsen: I also would like some Ast.Namespace abstraction. (Perhaps we should do that first?)

magnus-madsen:
Can you update this to reflect https://github.com/flix/flix/issues/4890

Also, I would like these new fields to be first in TypedAst.Root

magnus-madsen:
Or maybe `modules` is first and `names` is last.

PR #4823, opened 2022-11-02
feat: add Double versions of the functional interfaces / closures (see issue #4721)

stephentetley:
This PR adds `DoubleFunction`, `DoubleConsumer`, `DoublePredicate`, and `DoubleUnaryOperator` to the supported functional interfaces.

The code is patterned after the Int / Long functional interface code.

I've re-ordered existing "functional interface" code so that the `___Function` code is usually considered first, as it seems "canonical", then the others follow alphabetical order. E.g. the code for Int functional interfaces goes `IntFunction`, `IntConsumer`, `IntPredicate`, and `IntUnaryOperator` for case matches, definitions, etc.

`JvmName.scala` is an exception as this was alphabetically ordered in the first place which I didn't notice initially.

The re-ordering adds some noise to this PR but should improve tidiness of the code base.


magnus-madsen:
Thanks! I appreciate the cleanup 👍

PR #4824, opened 2022-11-02
feat: add Object versions of the functional interfaces Function, Consumer and Predicate (see issue #4721)

stephentetley:
This PR adds `Function`, `Consumer`, `Predicate` which take an Object as the first argument.

It doesn't add `UnaryOperator` because this has the same signature as `Function` and I don't quite know how to implement it.

The code is patterned after the recent functional interface code.

Test.Java.Function.flix has been indented to four spaces, including existing code.


magnus-madsen:
Thanks!

PR #4825, opened 2022-11-02
feat: split `Channel` into `Sender` and `Receiver`

paulbutcher:
This fixes the "Split the type Channel[t] into Sender[t] and Receiver[t]. Drop Channel[t]." task of #4755


main/src/library/Channel.flix
```diff
@@ -24,8 +22,8 @@ namespace Channel {
     ///
     pub def buffered(n: Int32): (Sender[t], Receiver[t]) \ IO =
         let m = if (n < 0) 0 else n;
-        let c = $CHANNEL_NEW$(m);
-        (c, c)
+        let c: Sender[t] = $CHANNEL_NEW$(m);
+        (c as Sender[t], c as Receiver[t])
```
magnus-madsen, line 26: I am confused by the casts. Should this simply not give the right result without casts?

main/src/ca/uwaterloo/flix/language/ast/MonoType.scala
```diff
@@ -58,7 +58,9 @@ object MonoType {
 
   case class Array(tpe: MonoType) extends MonoType
 
-  case class Channel(tpe: MonoType) extends MonoType
+  case class Sender(tpe: MonoType) extends MonoType
+
+  case class Receiver(tpe: MonoType) extends MonoType
```
magnus-madsen, line 63: I think- but I am not sure- due to Lowering we can actually remove both Sender and Receiver from MonoType! (And thus simplify the later phases even more).

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1563,14 +1563,14 @@ object Typer {
         for {
           (constrs, tpe, _, eff) <- visitExp(exp)
           _ <- expectTypeM(expected = Type.Int32, actual = tpe, exp.loc)
-          resultTyp <- liftM(Type.mkChannel(elmType, loc))
+          resultTyp <- liftM(Type.mkSender(elmType, loc))
```
magnus-madsen, line 1566: This should return a pair `(Sender[t], Receiver[t]`). 

I think `Type.mkTuple` and then it should work.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -2254,7 +2254,7 @@ object Typer {
         val e = visitExp(exp, subst0)
         val pur = Type.Impure
         val eff = e.eff
-        TypedAst.Expression.NewChannel(e, Type.mkChannel(subst0(elmType), loc), pur, eff, loc)
+        TypedAst.Expression.NewChannel(e, Type.mkSender(elmType, loc), pur, eff, loc)
```
magnus-madsen, line 2257: This then is not the right type.

main/src/library/Channel.flix
magnus-madsen: Ah- its because the type rule should be updated.

main/src/ca/uwaterloo/flix/language/ast/MonoType.scala
paulbutcher: I got 99% of the way through a version where we had no `Sender` or `Receiver` at all (just using Mmpc and aliases within `Channel.flix`) but that failed because it broke all of our LibNix and LibMin tests.

So then I tried just plumbing `Sender` and `Receiver` through to lowering, but that resulted in incomplete matches. In the end this was the easiest way I could find to make things work, but if you can think of a way of rearranging things which would allow it to work, let me know.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
paulbutcher: Yes, I originally implemented that, but that then requires all of the `par` code to be rewritten. So I thought it easier to do it this way.

But, thinking about it, I could instead rewrite the `par` code so that it uses `Concurrent/Channel` directly, which would that problem.

paulbutcher:
I'll go back to the version I originally wrote which returns a pair from `$CHANNEL_NEW$` and rewrite the `par` code to use `Concurrent/Channel` directly (instead of going through `Channel`). Once I've got that working I'll have another go at removing `Sender` and `Receiver` from the post lowering code.

paulbutcher:
I think that this is now ready for review again.

I realised that I could avoid having to change the `par` code at all by leaving `Concurrent/Channel/newChannel` alone and creating the tuple in the `NewChannel` case within the lowerer. This does require plumbing the channel element type through the AST, so there are quite a few more changed files, but the changes are almost entirely mechanical.

I can't find any way to get rid of the monotypes, sadly (see the commit I tried above, which doesn't work: it hits the InternalCompilerException in the finalizer.

magnus-madsen:
I will take a look soonish.

main/src/ca/uwaterloo/flix/language/ast/MonoType.scala
magnus-madsen: I think I can explain this. 

After `Lowering` we should have eliminated all occurrences of `Channel` and instead the type should now be `Channel.Mpmc` (or whatever it is called). 

However, as I point to below, there is still one place where we retain the `Channel` type instead of translating it.

magnus-madsen: ... And I was wrong...

If we use `mkChannelTpe` everywhere, how can there still be a `Channel` left over later?

main/test/flix/Test.Exp.Concurrency.Unbuffered.flix
```diff
@@ -93,8 +93,8 @@ namespace Test/Exp/Concurrency/Unbuffered {
     @test
     def testUnbufferedChannelPutGet16(): Bool \ IO =
         let (s1, r1) = Channel.unbuffered();
-        let (s2, _) = Channel.unbuffered();
-        spawn Channel.send(s2, s1);
+        let (s2, r2) = Channel.unbuffered();
+        spawn Channel.send(r2, s1);
```
magnus-madsen, line 97: I assume the type system caught these-- which is nice

magnus-madsen:
Thanks!

Issue #4693, opened 2022-10-13
Cannot upcast String


mlutze:
String is a subclass of CharSequence, but we cannot upcast String to CharSequence

I would guess there's a similar issue with BigInteger -> Number, Array -> Object

magnus-madsen:
@jaschdoc Would this work now?

jaschdoc:
Yes we even have a test for it :)

```flix

    @test
    def testUpcast17(): Unit =
        let f = (_: ##java.lang.CharSequence) -> ();
        f(upcast "hello")

```

PR #4830, opened 2022-11-04
fix: support null in stringify

### LINKED ISSUE #4726: `stringify` crash on `null`

mlutze:
fixes #4726

### DIFF
--- a/main/src/library/Prelude.flix
+++ b/main/src/library/Prelude.flix
@@ -196,81 +196,96 @@ pub def escape(s: String): String = {
 /// Returns an automatic string representation of `x`.
 ///
 @Internal @Unsafe
-pub def stringify(x: a): String = typematch x {
-    case _: Unit => "()"
-    case b: Bool =>  if (b) "true" else "false"
-    case c: Char => "\'" + escape("${c: Char}") + "\'"
-    case y: Float32 =>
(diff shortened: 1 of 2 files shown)


main/src/library/Prelude.flix
```diff
@@ -196,81 +196,96 @@ pub def escape(s: String): String = {
...
+        case c: Char => "\'" + escape("${c: Char}") + "\'"
+        case y: Float32 =>
+            import static java.lang.Float.toString(Float32): String \ {};
+            toString(y) + "f32"
+        case y: Float64 =>
+            import static java.lang.Double.toString(Float64): String \ {};
+            toString(y)
+        case y: BigDecimal =>
+            import java.math.BigDecimal.toString(): String \ {};
+            if (isNull(y as ##java.lang.Object)) {
```
JonathanStarup, line 213: Is this cast necessary?

mlutze: Yes, unfortunately. See also https://github.com/flix/flix/issues/4693

JonathanStarup: oh sorry yes, I mistakingly thought it was isNull from flix

PR #4831, opened 2022-11-04
feat: support precision in BigDecimal literals

### LINKED ISSUE #4797: Syntax for BigDecimal precision
Part of BigDecimal is its precision, as well as its value. Currently the former cannot be expressed literally.

We may need to expand the syntax to allow for more explicit precision:
```
1.0ff // currently allowed
1ff // not allowed
1.0e23ff // not allowed
```

mlutze:
fixes #4797


magnus-madsen:
What's this precision about?

mlutze:
With BigDecimal, 1.0 is different from 1.00 because the later has more precision.

PR #4836, opened 2022-11-04
fix: Allow Ident to be marked as synthetic

paulbutcher:
As per discussion on Gitter, this allows lambda parameters to be marked as synthetic (which means that syntax colouring is correct for match lambdas).


magnus-madsen:
I think the right solution is to perform desugaring later in the compiler.

But since this work around is short, I can live with it as an intermediate step.

PR #4837, opened 2022-11-04
refactor: use safe upcast in standard library

mlutze:


magnus-madsen:
Presumably these would become `supercast`s.

PR #4843, opened 2022-11-05
chore: Remove sender and receiver monotypes

paulbutcher:
This fixes the Remove MonoType.Channel task of #4755


magnus-madsen:
Very nice!

PR #4852, opened 2022-11-06
feat: `Graph.flix` GraphViz and sinks

JonathanStarup:
Related to #3906.
* Add function to find sink nodes (`sinks`)
* Add function to compute GraphViz string of graphs


JonathanStarup:
This is actually super nice to use with a preview vscode extension

JonathanStarup:
overtaken by #4855

PR #4855, opened 2022-11-07
feat: `Graph` graphviz functions

JonathanStarup:
Related to #3906 
* Add `pub def toGraphViz(g: m[((t, t))]): String with Foldable[m], Boxable[t], ToString[t]` function
* Add `toGraphviz` tests
* Add `pub def toGraphvizWeighted(g: m[((t, Int32, t))]): String with Foldable[m], Boxable[t], ToString[t]` function
* Add `toGraphvizWeighted` tests

### DIFF
--- a/main/src/library/Graph.flix
+++ b/main/src/library/Graph.flix
@@ -176,17 +176,32 @@ namespace Graph {
     }
 
     ///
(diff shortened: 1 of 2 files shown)


magnus-madsen:
Good idea!

Issue #4766, opened 2022-10-26
Add `supercast(e)`


magnus-madsen:
The expression `supercast(e)` upcasts `e`  from a Java type to one of its super-types.

It does not cast functions nor does it cast effects. It only works on one Java type (e.g. String to Object).

EDIT: The syntax should be `supercast(e)`.

PR #4859, opened 2022-11-07
feat: add `supercast(e)` expression

jaschdoc:
Closes https://github.com/flix/flix/issues/4766

- [x] Add `RedundantSupercast` in Redundancy Error and use instead of `RedundantUpcast`
- [x] Expect `Type.native` in Typer?
- [x] Add doc to `ParsedAst`
- [x] Add qualified names to `Expression` in Ast nodes e.g. `Supercast(exp: NamedAst.Expression, ...)`
- [x] Add doc to `JavaSubTypeResult`
- [x] Add doc to new Safety Errors


jaschdoc:
> Expect `Type.native` in Typer?

@mlutze is there a way to do this? It would have to call `expectTypeM(Tpe.Cst(TypeConstructor.Native(???, ...)), ...)` so what should the actual param to `TypeConstructor.Native` which expects a class? Something like `classOf[java.lang.Object]`

There's an experiment / draft in the Typer

mlutze:
> > Expect `Type.native` in Typer?
> 
> @mlutze is there a way to do this? It would have to call `expectTypeM(Tpe.Cst(TypeConstructor.Native(???, ...)), ...)` so what should the actual param to `TypeConstructor.Native` which expects a class? Something like `classOf[java.lang.Object]`
> 
> There's an experiment / draft in the Typer (that doesn't work)

There is not a way to do this in the current system. But we don't want to do this anyway. Recall that `String` is a Flix primitive type, but it's a Java subtype of `CharSequence` so we want it to be supercastable.

jaschdoc:
Do we really want Flix primitives to be supercastable?

jaschdoc:
Do we support type parameters for java types? This test fails:
```flix
    @test
    def testSupercast05(): Unit \ Impure =
        import new java.lang.StringBuilder(): ##java.lang.StringBuilder \ Impure as newStringBuilder;
        let f = (_: ##java.lang.Comparable[StringBuilder]) -> ();
        f(supercast(newStringBuilder()))
```

jaschdoc:
Other than that, @mlutze a review would be much appreciated! :D

main/src/ca/uwaterloo/flix/api/lsp/Indexer.scala
```diff
@@ -329,6 +329,9 @@ object Indexer {
     case Expression.Upcast(exp, tpe, _) =>
       visitExp(exp) ++ visitType(tpe) ++ Index.occurrenceOf(exp0)
 
+    case Expression.Supercast(exp, tpe, _) =>
+      visitExp(exp) ++ visitType(tpe) ++ Index.occurrenceOf(exp0)
```
mlutze, line 333: The indexer should only visit types that are _visible_. Since the type is not written in the source code here, it shouldn't be visited. (Same for upcast above)

main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
```diff
@@ -433,6 +433,9 @@ object SemanticTokensProvider {
     case Expression.Upcast(exp, tpe, _) =>
       visitExp(exp) ++ visitType(tpe)
 
+    case Expression.Supercast(exp, tpe, _) =>
+      visitExp(exp) ++ visitType(tpe)
```
mlutze, line 437: Same idea here as with Indexer

main/src/ca/uwaterloo/flix/language/errors/RedundancyError.scala
```diff
@@ -540,6 +540,33 @@ object RedundancyError {
...
   }
 
+  /**
+    * An error raised to indicate that a supercast is redundant.
+    *
+    * @param loc the source location of the upcast.
+    */
+  case class RedundantSupercast(loc: SourceLocation) extends RedundancyError {
+
+    def summary: String = "Redundant supercast. The expression already has the expected type."
```
mlutze, line 550: Could be nice to see the type we're upcasting to

main/src/ca/uwaterloo/flix/language/errors/SafetyError.scala
```diff
@@ -137,14 +137,93 @@ object SafetyError {
...
+    * @param other       the other type.
+    * @param loc         the source location of the supercast.
+    */
+  case class NonJavaTypeSupercast(nonJavaType: Type, other: Type, loc: SourceLocation) extends SafetyError {
+    override def summary: String = "Unsafe supercast"
+
+    override def message(formatter: Formatter): String = {
+      import formatter._
+      s"""${line(kind, source.name)}
+         |>> The following supercast tries to cast a non-java to a java type.
```
mlutze, line 186: Capitalize Java :)

mlutze: I'm not sure "Java type" is the right verbiage. But I cannot think of something better ATM.

main/src/ca/uwaterloo/flix/language/errors/SafetyError.scala
```diff
@@ -137,14 +137,93 @@ object SafetyError {
...
+         |
+         |${code(loc, "the supercast occurs here.")}
+         |
+         |Actual type:      $actual
+         |Tried casting to: $expected
+         |""".stripMargin
+    }
+
+    override def explain(formatter: Formatter): Option[String] = Some({
+      "Did you try to supercast two types at the same time?"
```
mlutze, line 223: This explanation is a little unclear to me. Maybe an example would be helpful.

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -384,20 +382,83 @@ object Safety {
...
     case _ => tpe1 == tpe2
 
   }
 
+  /**
+    * ADT to indicate what the result of checking a java subtype was.
+    */
+  private sealed trait JavaSubtypeResult
+
+  private case object Castable extends JavaSubtypeResult
```
mlutze, line 404: We always put enums in their own namespaces:

```scala
private sealed trait JavaSubtypeResult
private object JavaSubtypeResult {
    case object Castable extends JavaSubtypeResult
    ...
}
```

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -384,20 +382,83 @@ object Safety {
...
+
+  /**
+    * Returns true if `tpe1` and `tpe2` are both java types
+    * and `tpe1` is a subtype of `tpe2`.
+    */
+  private def isJavaSubtypeOf(tpe1: Type, tpe2: Type)(implicit flix: Flix): JavaSubtypeResult = (tpe1.baseType, tpe2.baseType) match {
+    case (Type.Cst(TypeConstructor.Native(left), _), Type.Cst(TypeConstructor.Native(right), _)) =>
+      if (right.isAssignableFrom(left)) Castable else NonCastable
+
+    case (_, Type.Var(_, _)) =>
```
mlutze, line 423: Don't we need to check both right and left for this?

main/src/ca/uwaterloo/flix/language/errors/SafetyError.scala
```diff
@@ -137,14 +137,93 @@ object SafetyError {
...
+  }
+
+  /**
+    * An error raised to indicate an invalid use of supercast.
+    *
+    * @param actual   the type of the expression being supercast.
+    * @param expected the type being cast to, i.e. the type of the supercast expression itself.
+    * @param loc      the source location of the supercast.
+    */
+  case class UnsafeSupercast(actual: Type, expected: Type, loc: SourceLocation) extends SafetyError {
```
mlutze, line 155: "actual" and "expected" don't really describe the situation well

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -384,20 +382,83 @@ object Safety {
...
+    * Returns a list of errors if the the supercast is invalid.
+    */
+  private def checkSupercastSafety(exp: Expression, tpe: Type, loc: SourceLocation)(implicit flix: Flix): List[SafetyError] = {
+    val tpe1 = Type.eraseAliases(exp.tpe)
+    val tpe2 = Type.eraseAliases(tpe)
+    isJavaSubtypeOf(tpe1, tpe2) match {
+      case Castable => Nil
+      case NonCastable => UnsafeSupercast(exp.tpe, tpe, loc) :: Nil
+      case NonJavaType(t) if t == tpe1 => NonJavaTypeSupercast(exp.tpe, tpe, loc) :: Nil
+      case NonJavaType(_) => NonJavaTypeSupercast(tpe, exp.tpe, loc) :: Nil
```
mlutze, line 457: The `t == tpe1` check seems strange to me. Also, it looks like this will be backwards half the time, no?

Do we need a specific "non-java-type" error? If we do, maybe we should just split it in two: cast to non-java-type and cast from non-java-type.

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -384,20 +382,83 @@ object Safety {
...
+      case TypeVariable => TypeVariableSupercast(exp.tpe, tpe, loc) :: Nil
```
mlutze, line 458: See other comment about checking both left and right for type variables.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1343,7 +1343,12 @@ object Typer {
           (constrs, tpe, pur, eff) <- visitExp(exp)
         } yield (constrs, tpe, Type.Pure, Type.Empty)
 
-      case KindedAst.Expression.Upcast(exp, tvar, loc) =>
+      case KindedAst.Expression.Upcast(exp, tvar, _) =>
+        for {
+          (constrs, _, pur, eff) <- visitExp(exp)
```
mlutze, line 1348: Can we add a comment for each of these to point out the reason we are ignoring the type of `exp`

main/test/ca/uwaterloo/flix/language/phase/TestRedundancy.scala
```diff
@@ -1330,4 +1330,38 @@ class TestRedundancy extends FunSuite with TestUtils {
...
+        |    import new java.lang.Object(): ##java.lang.Object \ Impure as newObject;
+        |    let _ =
+        |        if (true)
+        |            supercast(newObject())
+        |        else
+        |            newObject();
+        |    ()
+        |""".stripMargin
+
+    val result = compile(input, Options.TestWithLibMin)
```
mlutze, line 1347: Do we need `MinLib`?

mlutze (changes requested):

mlutze:
> Do we really want Flix primitives to be supercastable?

I think we do. There are a lot of String-related functions that take `CharSequence`, for example, and of course many that take `Object`s. Supporting this will be important for a lot of interop without resorting to unsafe casts.

> Do we support type parameters for java types? Shouldn't Comparable take a type param like below? (as opposed to the test in the PR which just uses ##java.lang.Comparable)

We do not support Java generics. This is a hard problem for a number of reasons.  One of them is that Java generics have variance, which Flix type parameters do not support.

jaschdoc:
> > Do we really want Flix primitives to be supercastable?
> I think we do. There are a lot of String-related functions that take CharSequence, for example, and of course many that take Objects. Supporting this will be important for a lot of interop without resorting to unsafe casts.

What does @magnus-madsen have to say on this?

magnus-madsen:
You cannot casts primitives, but its OK to cast String to CharSeq.

What else do we have in mind?

main/src/ca/uwaterloo/flix/language/errors/RedundancyError.scala
jaschdoc: Isn't this explained in the `message` function?

jaschdoc:
Hmm what do you think of changing the type variable error to something like `undecidable supercast`?

magnus-madsen:
@mlutze Can you do another review, perhaps with a focus on error messages.

main/src/ca/uwaterloo/flix/language/errors/RedundancyError.scala
```diff
@@ -540,6 +540,34 @@ object RedundancyError {
...
+      s"""${line(kind, source.name)}
+         |>> Redundant supercast. The expression already has the expected type.
+         |
+         |${code(loc, "redundant supercast.")}
+         |
+         |""".stripMargin
+    }
+
+    def explain(formatter: Formatter): Option[String] = Some({
+      s"""If you try to supercast a type that is already the
```
mlutze, line 564: supercast a value that already has the expected type

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -351,20 +350,105 @@ object Safety {
...
+  /**
+    * Returns a list of errors if the the supercast is invalid.
+    */
+  private def checkSupercastSafety(exp: Expression, tpe: Type, loc: SourceLocation)(implicit flix: Flix): List[SafetyError] = {
+    val tpe1 = Type.eraseAliases(exp.tpe)
+    val tpe2 = Type.eraseAliases(tpe)
+    isJavaSubtypeOf(tpe1, tpe2) match {
+      case JavaSubtypeResult.Castable => Nil
+      case JavaSubtypeResult.NonCastable => UnsafeSupercast(exp.tpe, tpe, loc) :: Nil
+      case JavaSubtypeResult.NonJavaType(t, c) if t == tpe1 => FromNonJavaTypeSupercast(exp.tpe, c, loc) :: Nil
```
mlutze, line 445: Can't the returned value just tell which side is non-java instead of checking like this?

main/src/ca/uwaterloo/flix/language/errors/SafetyError.scala
```diff
@@ -137,14 +134,159 @@ object SafetyError {
...
+         |>> The following supercast attempts to cast to a type variable.
+         |
+         |${code(loc, "the supercast occurs here.")}
+         |
+         |Actual type:      $from
+         |Tried casting to: $tvar
+         |""".stripMargin
+    }
+
+    override def explain(formatter: Formatter): Option[String] =
```
mlutze, line 279: This also commonly occurs when we try to supercast the final value of a function, right?

```scala
def f(): ##java.lang.Object = supercast("hello")
```

main/src/ca/uwaterloo/flix/language/errors/RedundancyError.scala
```diff
@@ -540,6 +540,34 @@ object RedundancyError {
...
+         |expected type the supercast is considered redundant.
```
mlutze, line 565: Do we want a call to action? "Tip: remove the supercast." or something

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -351,20 +350,103 @@ object Safety {
...
       subTypePurity && isEffSubset && superTypeArgs && subTypeResult
 
     case _ => tpe1 == tpe2
 
   }
 
+  /**
+    * ADT to indicate what the result of checking a java subtype was.
+    */
+  private sealed trait JavaSubtypeResult
```
magnus-madsen, line 370: Is there any reason to have this intermediate data type?

PR #4862, opened 2022-11-08
refactor: introduce Validation.traverseOpt

mlutze:
depends on #4861, not because it actually depends on it, but because it introduces a bunch of Options


magnus-madsen:
Conflicted.

magnus-madsen:
Everything broke again? :)

PR #4869, opened 2022-11-09
feat: lsp completions for ```import (static) get/set```

Daniel-Anker-Hermansen:
image
image


magnus-madsen:
Thanks! Looking forward to playing with it :)

PR #4871, opened 2022-11-10
feat: improve type errors of function applications

magnus-madsen:


magnus-madsen:
@athas Hello

mlutze:
!!!jar

main/src/ca/uwaterloo/flix/language/phase/unification/Unification.scala
```diff
@@ -288,6 +289,21 @@ object Unification {
       unifyTypeM(ts, loc)
   }
 
+  /**
+    * Pairwise unifies the types in `tpes1` and `tpes2` with the corresponding source locations `locs`.
+    *
+    * Assumes that `tpes1`, `tpes2`, and `locs` have the same length.
+    */
+  def pairwiseUnifyM(tpes1: List[Type], tpes2: List[Type], locs: List[SourceLocation])(implicit flix: Flix): InferMonad[Unit] =
```
mlutze, line 297: We should communicate in the signature or function title that this is an `expectTypeM` rather than just a `unifyTypeM`

main/src/ca/uwaterloo/flix/language/phase/unification/Unification.scala
```diff
@@ -288,6 +289,21 @@ object Unification {
...
+    (tpes1, tpes2, locs) match {
+      case (x :: xs, y :: ys, loc :: rs) =>
+        for {
+          _ <- expectTypeM(expected = x, actual = y, loc)
+          _ <- pairwiseUnifyM(xs, ys, rs)
+        } yield ()
+      case (_, _, _) => InferMonad.PointUnit
```
mlutze, line 304: I guess this is better than traversing since it returns `PointUnit`?

mlutze:
This is an improvement. I notice some weirdnesses with source locations but I don't think that's to do with this PR directly.

```
❌ -- Type Error -------------------------------------------------- file:///home/matt/Downloads/tmp/Main.flix

>> Expected type: 'Unit' but found type: 'String'.

6 |     f("ji")
           ^^
```

```
❌ -- Type Error -------------------------------------------------- file:///home/matt/Downloads/tmp/Main.flix

>> Expected type: 'Bool' but found type: 'Unit'.

5 |     hjk()
        ^^^^^

```

athas:
So now the research is done, right?

magnus-madsen:
Hah

Issue #4780, opened 2022-10-28
Introduce `Ast.Constant`


magnus-madsen:
Introduce `Ast.Constant` as a new type:

```scala
    case class Unit(loc: SourceLocation) extends FinalAst.Expression {
      final val tpe = MonoType.Unit
    }

    case class Null(tpe: MonoType, loc: SourceLocation) extends FinalAst.Expression

    case class True(loc: SourceLocation) extends FinalAst.Expression {
      final val tpe = MonoType.Bool
    }

    case class False(loc: SourceLocation) extends FinalAst.Expression {
      final val tpe = MonoType.Bool
    }

    case class Char(lit: scala.Char, loc: SourceLocation) extends FinalAst.Expression {
      final val tpe = MonoType.Char
    }

    case class Float32(lit: scala.Float, loc: SourceLocation) extends FinalAst.Expression {
      final val tpe = MonoType.Float32
    }

    case class Float64(lit: scala.Double, loc: SourceLocation) extends FinalAst.Expression {
      final val tpe = MonoType.Float64
    }

    case class BigDecimal(lit: java.math.BigDecimal, loc: SourceLocation) extends FinalAst.Expression {
      final val tpe = MonoType.BigDecimal
    }

    case class Int8(lit: scala.Byte, loc: SourceLocation) extends FinalAst.Expression {
      final val tpe = MonoType.Int8
    }
/// ... rest
```

Then introduce e.g. `FinalAst.Cst` (just above Var).

magnus-madsen:
(We should not include mono type in `Ast.Constant`).

magnus-madsen:
@mlutze I was thinking you could do a sketch of this.

PR #4872, opened 2022-11-10
bring AST.Constant up to Typer

mlutze:
related to #4780


magnus-madsen:
Looks like a reasonable win so far, agree?

mlutze:
> Looks like a reasonable win so far, agree?

Yep, Typer will need some special treatment but I think it's smooth sailing after that

Issue #4870, opened 2022-11-09
Change the idiomatic name of region values from `r` to `rc`


JonathanStarup:
* This will remove duplicate naming for the region type and the value `reg: Region[r]`
* Smaller chance of naming collisions for three characters

magnus-madsen:
But, on the other hand, everything else has just one letter names.

JonathanStarup:
`ef`, `xs`, `cmp`, `len`, multiple names arguments. My comment about `r` collision stems from `Random` which would also like to be called `r`

magnus-madsen:
I am not fully yet to pull the trigger on this. In particular, I am also considering if we need two names for regions (scope vs. region capability) and/or if we want to have region variable `r` with associated rigid variable `r!`.

magnus-madsen:
An update on my thought process.

I think, when it comes to names, it isimportant to distinguish between different elements. That is why, for example, for the "namespace rework" I plan to have both modules and namespaces:

```
mod A.B.C {  // This is the module.
    ^^^^^    // This is *the namespace of the module*.
  ...        // This is the module.
}            // This is the module.
```

Thus when you say "namespace" you mean "A.B.C" and when you say module you mean the entire thing (which has a namespace associated with it).

Returning to the problem at hand:
```
region r { // r has type Region[r]
  let sb = StringBuilder.new(r);
  spawn (1+1) @ r
}
```

Here we have:
- A region which binds a variable `r`.
- The variable `r` is bound to a region value `r`.
- The type of the variable `r` is `Region[r']` for some 
- There are also uses of `r` inside the region.
The same name is used for many different things. How to untangle?

Internally the compiler uses `Scope` for the region construct and `Region` for the value of `r` and `Region` for the type of `r`. I don't find this much better since `Scope` and `Region` feels like it means pretty much the same thing.

An attempt could be:

```
region rh { // r has type RegionHandle[rh]
  let sb = StringBuilder.new(rh);
  spawn (1+1) @ rh
}
```

So now `rh` is a local variable of type `RegionHandle[rh]`. 

We can say that a `region` introduces a "region handle" which is a value of type `RegionHandle[...]`. It is still a bit ugly with the name of the region inside `RegionHandle`. 

Note that what I call `RegionHandle` is probably what many would call a capability. So that's also an option.

We could also go for `scope` and `ScopedHandle` to play into the buzz on structured concurrency.

In any case, I have not yet made a decision, but these are some thoughts. Let me know what you think.

[...]

PR #4873, opened 2022-11-10
chore: Make term level and type level region names distinct

paulbutcher:
As per discussion on Gitter, this PR changes term level region variables to `reg` (or `reg1`, ...) instead of `r` to make them distinct from type level region variables.


magnus-madsen:
I am closing this (but also adding some discussion to #4870).

PR #4877, opened 2022-11-10
ci: remove mlutze/fcwg from community build

mlutze:
This project is no longer maintained.


magnus-madsen:
Could it be? I was using it in paper because it had some code.

mlutze:
I guess so

PR #4884, opened 2022-11-11
feat: lsp completions for ```use```

Daniel-Anker-Hermansen:


main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -905,6 +906,194 @@ object CompletionProvider {
...
+      case _ => Nil
+    }
+  }
+
+  /**
+    * Gets completions for a sub namespace of a prefix namespace
+    * I.e if you have namespace A/B/C/D, then if prefix is A/B it will return a completion for A/B/C
+    */
+  private def nsCompletionsAfterPrefix(prefix: List[String])(implicit context: Context, root: TypedAst.Root): Iterable[CompletionItem] = {
+    val nss = root.defs.keySet.map(_.namespace) ++
```
magnus-madsen, line 976: I guess this could be expensive, but at the same time, you don't have the right data structure to make it cheap.

magnus-madsen:
Thanks-- looking forward to trying it out.

magnus-madsen:
@JonathanStarup You should check this out, I think it answers many of your prayers :)

PR #4888, opened 2022-11-11
feat: shorthand namespace declaration

### LINKED ISSUE #4894: Feat: Floating namespace syntax
# Feature
This allows you to write
```scala
namespace NowThisIsReallyNice

pub def example(): Unit = ()
```
where `example` is in the `NowThisIsReallNice` namespace.

**Note**: This shorthand namespace can only happen if its the first (non-comment) thing in the file and it behaves as though the rest of the file is enclosed in `{...}`.

# Pros
* Decreases indentation (most files have a 4char left margin for no reason, especially with the new namespace scoping)
* Avoids annoying and confusing parsing errors when you forget the last `}`
* This is totally optional, no breaking changes
* [Consistency with Scala](https://docs.scala-lang.org/tour/packages-and-imports.html)
# Cons
* You can't trivially see if something is in a namespace or top-level (but files/projects of any scrolling size should not use top-level IMO)
* This might make nested namespaces confusing, but we could strongly consider disallowing this. In the below example `f` lies in `A/B.f`
```scala
namespace A
namespace B {
    def f(): Unit = ()
}
```

JonathanStarup:
Fixes #4894


magnus-madsen:
I think this is a good idea 👍 . Rust has something similar: 

https://doc.rust-lang.org/rust-by-example/mod/split.html

Look at `my/nested`.

But given that you just convinced me (through several other tickets) that there are many bugs in the current namespace mechanism, I think we need further re-design and re-implementation. 

Once that is done, I would support a feature like this.

magnus-madsen:
Perhaps we should convert this into a ticket with your text?

PR #4889, opened 2022-11-11
refactor: use Ast.Constant from Typer to GenExpression

### LINKED ISSUE #4780: Introduce `Ast.Constant`

mlutze:
fixes #4780


main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -36,29 +36,29 @@ object GenExpression {
     * Emits code for the given expression `exp0` to the given method `visitor` in the `currentClass`.
     */
   def compileExpression(exp0: Expression, visitor: MethodVisitor, currentClass: JvmType.Reference, lenv0: Map[Symbol.LabelSym, Label], entryPoint: Label)(implicit root: Root, flix: Flix): Unit = exp0 match {
-    case Expression.Unit(loc) =>
+    case Expression.Cst(Ast.Constant.Unit, _, loc) =>
```
magnus-madsen, line 39: I would like, but it can be separate PR, to have a `compileCst` such that functions naturally follow the data.

PR #4891, opened 2022-11-12
chore: remove old record syntax

magnus-madsen:


magnus-madsen:
@JonathanStarup @mlutze @stephentetley 

Heads up:

This PR removes the old record syntax `{ a :: Int32 }` so that only `{ a = Int32 }` is supported.

(Thus breaking some of your code-- sorry!)

PR #4899, opened 2022-11-12
feat: Add `Immutable` type class and add common instances

### LINKED ISSUE #4845: Add `Imm` type class and add common instances
- [ ] Add a type class for immutable values:

```scala
/**
 * A type class for types that are immutable.
 */
class Imm[t]
```

- [ ] add instances for most types (see e.g. `Eq` here https://github.com/flix/flix/blob/master/main/src/library/Eq.flix#L55)

The point is that types such as `Ref`, `Array`, `MutList` etc. and everything that is mutable will *NOT* be part of this class.

paulbutcher:
Fixes #4845


magnus-madsen:
Thanks!

PR #4901, opened 2022-11-13
fix: Use platform class loader, not bootstrap class loader

### LINKED ISSUE #4898: Bug: Classloader changes stop the compiler loading `java.sql` and potentially other Java SE classes
Paul's changes to the classloader in commit 2945605 (Oct 29) have had the unfortunate consequence that `java.sql` is no longer available to the Resolver stage of Flix's compiler.

While the Flix compiler can see external jars if you pass their path to the compiler through the packager looking in the `lib` directory or on the command line you can't do this for the SE classes as they are supplied with the JDK and are `.jmod` files in Java 9 and above.

To see the problem, create a new project and add the following code as a module in the `src` directory

~~~ 

/// Time.flix

namespace Interop/Sql {
    
    pub enum Time(##java.sql.Time)


}

namespace Interop/Sql/Time {

    use Interop/Sql.Time
    use Interop/Sql.Time.Time

    pub def mkTime(i: Int64): Time = 
        import new java.sql.Time(Int64): ##java.sql.Time \ {} as newTime;
        newTime(i) |> Time



}

~~~

Invoke with `java -jar flix.jar test` ...

~~~

-- Resolution Error -------------------------------------------------- E:\coding\flix\loader-test\src\Time.flix

>> Undefined Java class 'java.sql.Time'.

20 |     pub enum Time(##java.sql.Time)
                       ^^^^^^^^^^^^^^^
                       undefined class.

-- Resolution Error -------------------------------------------------- E:\coding\flix\loader-test\src\Time.flix

>> Undefined Java class 'java.sql.Time'.

31 |>         import new java.sql.Time(Int64): ##java.sql.Time \ {} as newTime;
32 |>         newTime(i) |> Time

undefined class.

-- Resolution Error -------------------------------------------------- E:\coding\flix\loader-test\src\Time.flix

>> Undefined Java class 'java.sql.Time'.

31 |         import new java.sql.Time(Int64): ##java.sql.Time \ {} as newTime;
                                              ^^^^^^^^^^^^^^^
                                              undefined class.
                                              
~~~                                               

Unfortunately this breaks my JDBC wrapper library.

paulbutcher:
Fixes #4898

Apologies: this was me being dense.

The original code passed `null` as the parent classloader, which means the bootstrap classloader. This knows how to load built-in packages like `java.lang` and `java.util`, but not `java.sql`.

This change switches to the platform class loader which does know how to load those classes (but still isn't the full system class loader, so still doesn't give compiled code access to compiler internals).

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/jvm/FlixClassLoader.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/jvm/FlixClassLoader.scala
@@ -26,10 +26,10 @@ import scala.collection.mutable
   *
   * @param classes A map from internal names (strings) to JvmClasses.
   * 
-  * We pass null as the `ClassLoader` parent to avoid it delegating to the system classloader
+  * We pass the platform class loader as the parent to avoid it delegating to the system classloader
   * (otherwise compiled Flix code has access to all classes within the compiler)
   */
-class FlixClassLoader(classes: Map[String, JvmClass])(implicit flix: Flix) extends ClassLoader(null) {
+class FlixClassLoader(classes: Map[String, JvmClass])(implicit flix: Flix) extends ClassLoader(ClassLoader.getPlatformClassLoader()) {
 
   /**
     * An internal cache of already loaded classes.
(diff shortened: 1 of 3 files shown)


magnus-madsen:
Let's add some test cases to import to ensure that this keeps working :)

magnus-madsen:
Thanks!

PR #4903, opened 2022-11-13
refactor: qualify AST nodes in Lowering and introduce LoweredAst

mlutze:
does not yet use the LoweredAst. related to #4779


main/src/ca/uwaterloo/flix/language/ast/LoweredAst.scala
```diff
@@ -0,0 +1,548 @@
+/*
+ * Copyright 2015-2016 Magnus Madsen
```
magnus-madsen, line 2: Okaaaay

mlutze: Oops, copy-pasta

PR #4904, opened 2022-11-13
refactor: replace reifyEff with typematch in stdlib

mlutze:
related to #4773

### DIFF
--- a/main/src/library/DelayMap.flix
+++ b/main/src/library/DelayMap.flix
@@ -158,12 +158,13 @@ namespace DelayMap {
         typematch f {
             case g: (k, v) -> Bool \ {} =>
                 if (useParallelEvaluation(m))
-                    let h = (k, v) -> g((k, force v));
+                    let h = (k, v) -> g(k, force v);
                     let DMap(t) = m;
                     RedBlackTree.parCount(threads() - 1, h, t) as \ {}
                 else
                     c()
-            case _: _  => c()
+            case _: (k, v) -> Bool \ ef => c()
+            case _: _  => unreachable!()
         }
 
     ///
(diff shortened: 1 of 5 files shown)


magnus-madsen:
Just a double check to ensure no silly mistakes :)

main/src/library/DelayMap.flix
```diff
@@ -154,17 +154,17 @@ namespace DelayMap {
     ///
     @Experimental @ParallelWhenPure
     pub def count(f: (k, v) -> Bool \ ef, m: DelayMap[k, v]): Int32 \ ef =
-        let f1 = match (k, v) -> f(k, v);
         def c() = foldLeftWithKey((b, k, v) -> if (f(k, v)) b + 1 else b, 0, m);
-        reifyEff(f1) {
-            case Pure(g) =>
+        typematch f {
+            case g: (k, v) -> Bool \ {} =>
```
magnus-madsen, line 159: Does k, v here refer to the params in the signature k, v? I assume so, right?

magnus-madsen: Why do you write `\ {}` here when you did not in `DelayList`? 

I guess it makes sense to be explicit and write `\ {}` everywhere?

mlutze: > Does k, v here refer to the params in the signature k, v? I assume so, right?

Yes. No new variables can be introduced in a typematch (except wildcards `_`).



> Why do you write `\ {}` here when you did not in `DelayList`?
> 
> I guess it makes sense to be explicit and write `\ {}` everywhere?

It also nice lines up with `\ ef`.

magnus-madsen: I suggest to write `\ {}` everywhere then! :)

jaschdoc (approved):
Looks good. I agree with Magnus that there should be explicit type casts to `\ {}` :)

magnus-madsen:
Thanks!

PR #4909, opened 2022-11-14
feat: Derive Immutable

### LINKED ISSUE #4848: Add automatic derivation of `Imm` to `Deriver`
Add support for automatic derivation of `Imm` for immutable types. 

@mlutze Perhaps you can expand on what is needed. I also don't know whether to treat this as `Boxed` or more like `Eq`?

paulbutcher:
Fixes #4848


paulbutcher:
So this now works, but I'm not really happy with the code in Safety. The type matching within `visitType` in particular feels a bit voodoo-ish to me:

https://github.com/flix/flix/pull/4909/files#diff-b13a9a579163dc32918be984176756151f2b153f1e080c4eb643eb2f9217732bR69

magnus-madsen:
I have a simple proposal that avoids the need for visitType. The idea is as follows:

- Our goal is that all mutable memory will be associated with a region.
- Our goal is also not to have `Static`.

Hence if we have an enum:

```
enum A[t, s] {
}
```

and the type parameters are *NOT* of kind `Bool` (but in the future Region) then none of its cases can contain a mutable reference or array. Put different if you want one of the cases to be `Array` then you have to write `Array[Int32, r]` but where can you get the `r` from? Only by making it a parameter of `A`. If you do so, we will see it, and see that it has kind `Bool` (in the future `Region)` and we will disallow you from marking the enum as `Immutable`.

Thus we can simplify the code to check if any of the type arguments to the enum (which we still have to look up) is of kind bool.

magnus-madsen:
@mlutze CC'ed for above.

paulbutcher:
That makes sense (and in fact @mlutze suggested something similar in #4848). But that raises the question of what value `Immutable` is bringing at all? If we can determine whether a type is immutable by checking to see if it takes a region as a type parameter, then why not make the presence or absence of a region in the type parameters of the type passed to `Channel.buffered` the criterium for whether the channel can be created and cut out the middleman?

That instinctively feels to me like a more fruitful avenue to me: perhaps we ultimately have something like a "threadsafe region" which could be used by types which can be sent over channels?

mlutze:
> That makes sense (and in fact @mlutze suggested something similar in #4848). But that raises the question of what value `Immutable` is bringing at all? If we can determine whether a type is immutable by checking to see if it takes a region as a type parameter, then why not make the presence or absence of a region in the type parameters of the type passed to `Channel.buffered` the criterium for whether the channel can be created and cut out the middleman?
> 
> That instinctively feels to me like a more fruitful avenue to me: perhaps we ultimately have something like a "threadsafe region" which could be used by types which can be sent over channels?

I think the simple "check `Channel.buffered` args" idea falls apart when we run into polymorphism.

Say we have a function:
```scala
def doThing(x: a): Unit \ (some effect) = {
    buffered(23) |> send(x)   // I don't know the arguments but the point is we send x on a channel
}
```

Here, we don't know anything about `a`, so checking the call site does not help us. `doThing(456)` is OK but `doThing([1, 2, 3])` is not. If we have a type class to constrain `a` however:

```scala
def doThing(x: a): Unit \ (some effect) with Immutable[a] = {
    buffered(23) |> send(x)
}
```

Then `doThing([1, 2, 3])` is prevented.

paulbutcher:
I'm not sure I follow that? Type inference allows us to work out every type that will ever be passed to `doThing` (it has to in order to be able to reject a program which ever passes a type which doesn't implement `Immutable`). What is it about a type class which makes it different in this case?

paulbutcher:
So, thinking about this more (take what is about to follow with a massive pinch of salt because I'm still working regions out).

I can write a function which can take an argument of any type, even a type that includes a region in its type parameters, without that function having any effect or using a region. Classic example would be the identity function:

```
def identity(x: a): a = x
```

But that function can never *do* anything with the value, because doing anything at all with it would require a region. If we think about a channel, the same applies to the receiver. So we could write a "channel identity" function like this:

```
def sendIdent(x: a, c: Sender[a]): Unit =
    Channel.send(x, c)
    
def receiveIdent(c: Receiver[a]): a =
    Channel.receive(c)
```

But as soon as we want to actually do anything to the value we've received we need to have a region. Which means that both the sending and receiving threads would need to have access to the same region, and would need to specify it in their effects.

What am I missing?

mlutze:
It seems to me that the problems you are describing are precisely what type classes are meant to address :)

##### In regard to the `doThing` function:
Let's imagine `doThing` is in some library. That means we don't know everything that `doThing` will be called on, since consumers can try to call it on whatever they want. It would be wrong to say that it can take _any_ type `a` as an argument because `a` _has_ to be immutable. In other words, it belongs to a _class of types_ that are all immutable.

This is the purpose of a type class: to designate some set of types with a particular property. Often that means there's some function defined for them, but in this case it's just a property of the type, which is why it's just a marker class.

So whenever we use `doThing(x)`, the type of `x` has to be either:
1. Some concrete type that is known to be `Immutable`, or
2. Some variable or type involving a variable where we know it is `Immutable` from the context.

##### In regard to the regions stuff:

The plan is for channels to live in regions, no? So the functions will look more like this:

```scala
def sendIdent(x: a, c: Sender[a, r]): Unit \ { Write(r) }=
    Channel.send(x, c)
```

paulbutcher:
> The plan is for channels to live in regions, no?

I think that this is the heart of my problem. We've spoken vaguely about what might happen, and that that might involve regions somehow, but don't (as far as I know) have an actual plan or design.

What we're currently doing is trying to implement `Immutable` which we know isn't going to be sufficient, and I struggle to see how it helps us get to where we need to get to, in part because I don't know what the end goal looks like.

paulbutcher:
I think that this is now ready for review. I went with adding the error to the deriver (which involved some plumbing to get it to report errors which it hasn't been doing until now).

main/src/ca/uwaterloo/flix/language/errors/DerivationError.scala
```diff
@@ -0,0 +1,58 @@
...
+        |>> Cannot derive 'Immutable' for enum ${red(enum0.sym.toString())}
+        |
+        |Because it takes a type parameter of kind 'Region'.
+        |
+        |${code(loc, "unable to derive Immutable.")}
+        |
+        |""".stripMargin
+    }
+
+    def explain(formatter: Formatter): Option[String] = None
```
mlutze, line 56: Maybe we could include an example in the `explain` section?

main/src/ca/uwaterloo/flix/language/phase/Deriver.scala
```diff
@@ -744,7 +750,50 @@ object Deriver {
...
+    * }
+    * }}}
+    *
+    * yields
+    *
+    * {{{
+    * instance Immutable[E[a]] with Immutable[a]
+    * }}}
+    * 
+    * The instance is empty: checks are performed within safety.
```
mlutze, line 773: This is no longer true, right?

main/src/ca/uwaterloo/flix/language/phase/Deriver.scala
```diff
@@ -31,39 +33,43 @@ import ca.uwaterloo.flix.util.{InternalCompilerException, Validation}
...
 object Deriver {
 
-  def run(root: KindedAst.Root)(implicit flix: Flix): Validation[KindedAst.Root, Nothing] = flix.phase("Deriver") {
-    val derivedInstances = root.enums.values.flatMap {
+  def run(root: KindedAst.Root)(implicit flix: Flix): Validation[KindedAst.Root, DerivationError] = flix.phase("Deriver") {
+    val derivedInstances = (traverse(root.enums.values) {
       enum => getDerivedInstances(enum, root)
+    }).map(_.flatten)
+
+    flatMapN(derivedInstances) {
```
mlutze, line 41: Here we shouldn't have to `flatMap`, just `map` and remove the `toSuccess` in the body.

main/src/ca/uwaterloo/flix/language/phase/Deriver.scala
paulbutcher: Oops! Good point.

main/src/ca/uwaterloo/flix/language/phase/Deriver.scala
paulbutcher: Ah! Thanks, didn't know about `map`.

main/src/ca/uwaterloo/flix/language/errors/DerivationError.scala
paulbutcher: I've added an example, but I'm not sure whether I've captured what you had in mind. Please let me know if you have better wording in mind.

paulbutcher:
@magnus-madsen, @mlutze 

I've been doing some more playing with this, and I don't think we're there yet. With these changes and given the following:

```
enum MyOption[a] with Immutable {
    case MyNone,
    case MySome(a)
}

def requiresImmutable(_: a): Bool with Immutable[a] = true
```

This compiles (which is good):

```
def bar(): Bool = requiresImmutable(MySome(42))
```

But this doesn't:

```
def foo(): Bool = requiresImmutable(MyNone)
```

It gives the error:

```
-- Type Error -------------------------------------------------- ../flix-playground/src/Main.flix

>> No instance of the 'Immutable' class for the type 'a'.

8 | def foo(): Bool = requiresImmutable(MyNone)
                      ^^^^^^^^^^^^^^^^^
                      missing instance
```

This, however, compiles OK:

```
def foo(): Bool = requiresImmutable(MyNone as MyOption[Int32])
```

I'm not sure what to do about this?

mlutze:
I think this is acceptable behavior. The same thing happens with other type classes, and for good reason: we don't know what instance of the class to use.

```scala
class Example[a] {
    pub def example(): a
}

class ToString[a] {
    pub def toString(x: a): String
}

instance ToString[Option[a]] with Example[a], ToString[a] {
    pub def toString(x: Option[a]): String = match x {
        case None => "None but could have been Some(${Example.example(): a})"
        case Some(y) => "Some(${y})"
    }
}
```

Granted, `Immutable` doesn't have any definitions inside, but I think for the sake of consistency, it's OK not to special-case this behavior.

mlutze:
In my experience, the necessary ascriptions come up frequently in minimal tests cases, but fairly rarely in real-world programs, where there's more type information in the system. I guess we will see if this holds with `Immutable` as well, but for now I'm OK with it.

magnus-madsen:
!!!jar

magnus-madsen:
@mlutze 

This is allowed:

```
instance Immutable[Array[t, r]]
```

but presumable putting `sealed` on `Immutable` will be sufficient?

paulbutcher:
Does `sealed` have any meaning for a class that's at top level (which `Immutable` is)?

magnus-madsen:
> Does `sealed` have any meaning for a class that's at top level (which `Immutable` is)?

Lets hear from the expert (Matt) :)

mlutze:
> Does `sealed` have any meaning for a class that's at top level (which `Immutable` is)?

I think it has no meaning at the top level.

I'm also not sure the Instances phases is smart enough to distinguish between derived and normal instances.

magnus-madsen:
Could `sealed` not just consider only instances in the same file?

paulbutcher:
In any case, wouldn't that stop `with` from deriving `Immutable` for user-defined types? Which we definitely want to support?

So, it sounds like we need something that checks for a region-typed type parameter for `instance` just like we already have for `with`?

magnus-madsen:
I will merge this and then we can iterate.

PR #4911, opened 2022-11-14
refactor: get class names in Reader

mlutze:


magnus-madsen:
LGMT

magnus-madsen:
Add more fixing :)

PR #4912, opened 2022-11-14
feat: lsp prioritize type completions

Daniel-Anker-Hermansen:
Prioritizes type completions in following order:
1. Builtin
2. Same file
3. Root namespace
4. Others


magnus-madsen:
Thanks!

Issue #4917, opened 2022-11-14
bug: adding region to test changes result


jaschdoc:
Why does the following test break when wrapping it in a region? The expressions within the regions should still have effects, right?
```flix
    @test
    def mapWithKey06(): Bool \ IO = // This passes
        let range = List.range(0, 2000);
        let m = List.zip(range, range) |> List.toMap;
        let a = new MutList(Static);
        discard Map.mapWithKey((k, v) -> { let b = new MutList(Static); MutList.push!(k, b); MutList.append!(MutList.toList(b), a); v }, m);
        MutList.toList(a) == range
```
i.e. changing to
```
    @test
    def mapWithKey06(): Bool = region r { // This fails
        let range = List.range(0, 2000);
        let m = List.zip(range, range) |> List.toMap;
        let a = new MutList(r);
        discard Map.mapWithKey((k, v) -> { let b = new MutList(r); MutList.push!(k, b); MutList.append!(MutList.toList(b), a); v }, m);
        MutList.toList(a) == range
    }
```

jaschdoc:
There is no compilation error, just assertion error.
VSCode tells me that the lambda given to `mapWithKey` has type `Int32 -> (Int32 -> Int32 & r)` and that the discard expression has type `Unit & r`.
Using `let _ = Map.mapWithKey(...)` shows that the variable `_` has type `Map[Int32, Int32]`.
Thus `mapWithKey` proceeds to the parallel section and returns a list that isn't ordered.
Maybe the bug is with `reifyEff`?

jaschdoc:
Here's an updated test that provides a bit more information:
```flix

    @test
    def mapWithKey06(): Bool = region r {
        let range = List.range(0, 2000);
        let m = List.zip(range, range) |> List.toMap;
        let a = new MutList(r);
        let f = (k, v) -> { let b = new MutList(r); MutList.push!(k, b); MutList.append!(MutList.toList(b), a); v };
        let _ =  Map.mapWithKey(f, m);
        Assert.eq(MutList.toList(a), range)
    }
```

jaschdoc:
![Screenshot from 2022-11-14 16-31-09](https://user-images.githubusercontent.com/60233376/201700349-2e050e09-6066-4951-8fef-7c35dd06926a.png)
![Screenshot from 2022-11-14 16-31-34](https://user-images.githubusercontent.com/60233376/201700353-f3079eb1-7d77-48df-8c8c-3bed8f665afa.png)
[...]

PR #4921, opened 2022-11-14
test: use regions in tests that used static region

jaschdoc:
Related to https://github.com/flix/flix/issues/4917


magnus-madsen:
Depends on https://github.com/flix/flix/issues/4917

PR #4923, opened 2022-11-15
fix: mark region vars as Impure inside regions in Monomorph

### LINKED ISSUE #4917: bug: adding region to test changes result

mlutze:
fixes #4917 in theory


magnus-madsen:
But in practice:  java.lang.StackOverflowError :D

PR #4938, opened 2022-11-17
feat: Require `Immutable` for `Channel.send`

### LINKED ISSUE #4846: Require `Imm` instance for `Channel.send` and `Channel.recv`
- [ ] Modify the signature of `Channel.send` to require an `Imm[t]` instance.
- [x] ~Modify the signature of `Channel.recv` to require an `Imm[t]` instance.~

This ensures that channels can only be used for immutable message passing; we cannot share mutable data, hence we cannot have data races. (This is strictly not true yet, because a spawn expression could close over a shared mutable data structure-- this we have to resolve).

paulbutcher:
Fixes #4846

### DIFF
--- a/main/src/library/Boxed.flix
+++ b/main/src/library/Boxed.flix
@@ -17,7 +17,7 @@
 ///
 /// Represents a boxed value.
 ///
-pub enum Boxed {
+pub enum Boxed with Immutable {
     case BoxedBool(Bool)
     case BoxedChar(Char)
     case BoxedInt8(Int8)
@@ -31,8 +31,6 @@ pub enum Boxed {
         (##java.lang.Object, ##java.lang.Object) -> Comparison
     )
 }
-
-instance Immutable[Boxed]
 
 instance Eq[Boxed] {
     pub def eq(b1: Boxed, b2: Boxed): Bool = match (b1, b2) {
--- a/main/src/library/Boxed.flix
+++ b/main/src/library/Boxed.flix
@@ -32,6 +32,8 @@ pub enum Boxed {
     )
 }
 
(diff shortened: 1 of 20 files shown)


main/src/library/Boxed.flix
```diff
@@ -32,6 +32,8 @@ pub enum Boxed {
     )
 }
 
+instance Immutable[Boxed]
```
magnus-madsen, line 35: Any reason not to use `with Immutable`?

main/src/library/Channel.flix
```diff
@@ -41,7 +41,17 @@ namespace Channel {
...
     ///
-    pub def send(m: t, s: Sender[t]): Unit \ IO = $CHANNEL_PUT$(s, m)
+    pub def send(m: t, s: Sender[t]): Unit \ IO with Immutable[t] = $CHANNEL_PUT$(s, m)
+
+    ///
+    /// Sends the message `m` on the given channel `s`.
+    ///
+    /// Blocks until the message is enqueued.
+    ///
+    /// Identical to `send` but doesn't require `Immutable`
```
magnus-madsen, line 51: . 
It is up to programmer to ensure that race conditions cannot arise due to concurrent access to `m`.

main/src/library/Boxed.flix
paulbutcher: I guess I was just aiming to be consistent with the surrounding code, but I'm happy to switch to `with` if you prefer. Let me know.

magnus-madsen: I would propose we use maximal derivation, i.e. place it on the enum, when possible.

PR #4940, opened 2022-11-18
feat: add `Histogram`

magnus-madsen:


mlutze:
See also: `Statistics.scala`

I think there may be some duplicated work.

PR #4942, opened 2022-11-18
refactor: roll ops in with defs and sigs

mlutze:


magnus-madsen:
Upwards and onwards 🚀

PR #4943, opened 2022-11-18
experiment: add return to typematch expression

mlutze:


magnus-madsen:
Not really sold on this, but if more use cases arrive, I could change my mind.

PR #4950, opened 2022-11-19
chore: Rename Immutable to Sendable

paulbutcher:
As per discussion in #4755


magnus-madsen:
Could I ask for a project-wide search on `Immutable` just to ensure we have caught everything :) Thanks!

paulbutcher:
Already done (I don't think I've missed anthing)

PR #4951, opened 2022-11-19
feat: add `minimal-project` example TOML

paulbutcher:
As per discussion with @magnus-madsen, example projects in draft PR


magnus-madsen:
which project is the smallest? I suppose the smallest would just have a toml file and one source file in `src` right?

paulbutcher:
> which project is the smallest? I suppose the smallest would just have a toml file and one source file in `src` right?

That's basically `hello-world`. I've included a `test` directory, plus `LICENSE.md` and `README.md` too, but they could all be optional.

magnus-madsen:
GitHub lost my comments :S .

Anyway, I propose that `hello-world` should be our copy-paste project. So thats fine. 

I propose we add `minimal-project` which should have a toml with:

```
[package]
name = "hello_world" # the name of the package
version = "0.1.0"    # the semantic version of the package
```

and exactly one file: `src/Main.flix` with a main function.

Anything else? Anything else in the toml?

Oh, I also propose its simply `flix.toml`.

paulbutcher:
I'm OK with `minimal-project`, and with switching to `flix.toml` instead of `FlixProject.toml` 👍

Your suggested minimal `flix.toml` suggests that the only required values (for an application project; libraries would be different) should be `name` and `version`.

This is different from what I suggested, which was `name` and `flix` (i.e. the version of the Flix compiler).

Personally:

* I prefer that we don't make `version` required for applications: it's one of my annoyances that I'm forced to do this by some build systems when I'm putting together a "throwaway" project. I don't care what the version is, and I dislike being forced to think about it.
    * Note: this is very different from a library: for a library `version` should *definitely* be required.
* I prefer that we *do* make `flix` a requirement for applications: I want a project to "just work", even if I threw it together 6 months ago and it's broken with whatever version of the compiler I happen to have lying around today.

magnus-madsen:
I propose a compromise: we require both versions. This avoids any special cases and ensure that we always know what we are dealing with.

paulbutcher:
> I propose a compromise: we require both versions. This avoids any special cases and ensure that we always know what we are dealing with.

Done

examples/projects/minimal-project/flix.toml
```diff
@@ -0,0 +1,4 @@
+[package]
+name = "minimal-project"
+version = "0.1.0"
+flix = "0.32.1"
```
magnus-madsen, line 4: Lets format and explain each option:

```
[package]
name    = "minimal-example" # The name of the package.
version = "0.1.0"           # The semantic version of the package.
flix    = "0.33.0"          # The required version of the Flix compiler.
```

magnus-madsen:
How about we do each project archetype separately? I think we have minimal example done. Next would be hello world, I guess.

paulbutcher:
> How about we do each project archetype separately? I think we have minimal example done. Next would be hello world, I guess.

Sorry Magnus, I'm confused: I've already added `hello-world`, what is it that you think should be changed about this?

magnus-madsen:
> > How about we do each project archetype separately? I think we have minimal example done. Next would be hello world, I guess.
> 
> Sorry Magnus, I'm confused: I've already added `hello-world`, what is it that you think should be changed about this?

Remove everything else, and then we can do one project "archetype" at a time. This one is "add minimal-example project", next would would be "add hello-world example project" and so forth.

magnus-madsen:
There are too many files in this one :)

paulbutcher:
Oh, got you. Does that mean that you're thinking of mergeing this PR? I was assuming that we wouldn't do that until we had the functionality implemented?

magnus-madsen:
> Oh, got you. Does that mean that you're thinking of mergeing this PR? I was assuming that we wouldn't do that until we had the functionality implemented?

I would merge each PR as we agree to it. (i.e. everyone agrees that "this is a valid toml" and "this is what it means").

magnus-madsen:
In particular, by merging this PR, we believe that:

- project name, project version, and Flix version are the three required properties that a project should have.
- We also seem to (implicitly) agree that since nothing else is specified the compiler will look in `src` and `test` for Flix source code. 
- Presumably since no dependencies are listed the compiler will NOT look for any dependencies. 

@mlutze @JonathanStarup ping

paulbutcher:
> * We also seem to (implicitly) agree that since nothing else is specified the compiler will look in src and test for Flix source code.

I mention this in the draft documentation (see https://github.com/flix/book/pull/68) but for completeness, the intention is that we have the following defaults (which can be overridden if necessary):

```ini
[package]
source-paths = ["src"]
resource-paths = ["resources"]

[build.prod]
source-paths = ["prod"]
config = { allow-holes = false, allow-debug = false }

[build.dev]
source-paths = ["dev", "test", "bench"]
config = { allow-holes = true, allow-debug = true }
```

paulbutcher:
> * project name, project version, and Flix version are the three required properties that a project should have.

Slight correction: project name, project version and Flix version are the three required properties that an *application* project should have. Library projects could (IMHO should) be different.

magnus-madsen:
> > * project name, project version, and Flix version are the three required properties that a project should have.
> 
> Slight correction: project name, project version and Flix version are the three required properties that an _application_ project should have. Library projects could (IMHO should) be different.

OK,, that makes sense.

paulbutcher:
FWIW this is why IMHO (but I'm not going to fight about it) version shouldn't be required. It's absolutely essential for a library, but it's not always necessary for an application which will never be a dependency of anything else.

magnus-madsen:
I will wait to see if Matt or Jonathan has any comments and then merge this one.

magnus-madsen:
@mlutze @JonathanStarup Do you have any comments? Otherwise this will be the minimal project configuration.

mlutze:
Two questions:
1. Does it need a name?
2. If it needs a name, do we want a more defined naming convention? E.g. like Java/Maven does this `org.birds.Albatross`

2 can be addressed later I suppose

JonathanStarup:
Is it possible to have flix-ranges? my Flix ansi package was compatible with something like flix 0.31.0-0.33.0.

magnus-madsen:
> Is it possible to have flix-ranges? my Flix ansi package was compatible with something like flix 0.31.0-0.33.0.

We have to download a compiler, so I think its better to say which one specifically? But I take your point. 

If we had a range, which one would you pick?

mlutze:
> > Is it possible to have flix-ranges? my Flix ansi package was compatible with something like flix 0.31.0-0.33.0.
> 
> We have to download a compiler, so I think its better to say which one specifically? But I take your point.
> 
> If we had a range, which one would you pick?

Newest one that's compatible with all the packages

JonathanStarup:
The idea would be that users of the library aren't forced to upgrade the flix version just because the library wanted to also support the newest version (`version = "0.32.0` would become `0.32.0-0.33.0`) and could do so without any changes

paulbutcher:
> If it needs a name, do we want a more defined naming convention? E.g. like Java/Maven does this org.birds.Albatross

Maven does that because all artefacts are stored within a central repository (so name clashes are possible and problematic). We're planning to store artefacts on GitHub and reference them via URL (not shown in this minimal example because it's neither publishing nor using dependencies) so clashes of that nature aren't possible.

There is, of course, the possibility of name clashes within the code (e.g. two different libraries use the same `namespace` or publish the same names at the top level) which we should probably think about and come up with an approach to avoiding, but that's out of scope for this conversation I think (unless we're planning on tying the name of the published artefact to the names within it somehow?)

paulbutcher:
> Is it possible to have flix-ranges?

I wasn't planning to do that, but your comments have got me thinking.

I think that there are two different concepts that we're trying to capture here, and the right answer depends upon whether we're talking about an application (which is what this minimal project represents) or a library.

For an application the answer is simple: There is a single Flix version which should be used to compile the code of the application and any libraries that it depends upon.

For a library, there are two different Flix versions that are of interest:

1. The version of Flix that should be used while developing the library (to compile and run the tests etc.)
2. The versions of Flix that it's capable of being compiled with when incorporated into a larger project (either another library or an application).

So, I propose:

1. We keep this minimal project (which represents an application) as-is.
2. For library projects, as well as `flix` specifying the version of the compiler that should be used to compile it during development, we also add an optional `flix-min` which is the minimum version of Flix required (so it will be an error for this library to be included in an application that uses an earlier version of Flix.

Notes:

* I'm not suggesting that we have `flix-max` because
    * we're trying to make sure that the compiler is backwards compatible (or at least, we will once we hit `1.0.0`)
    * when writing a library there's no way for the author to know which putative future version might kill it
    * We don't want to artificially restrict it to just whatever the version of Flix they happened to be using at the time because that will be a maintenance nightmare.
* I'm suggesting that `flix-min` should be optional because in most instances most libraries will work just fine. The author will know if they're relying on some recently introduced language feature.

magnus-madsen:
These are reasonable discussions, but I think there is agreement that the current proposal is workable. (But maybe we will revisit and extend in the future).

PR #4952, opened 2022-11-19
feat: make lsp completions more context aware

Daniel-Anker-Hermansen:
related to #4554


magnus-madsen:
!!!jar

magnus-madsen:
Excellent. Thanks!

PR #4953, opened 2022-11-19
feat: add hello-world example TOML

paulbutcher:
This an example of what would be generated by `flix new hello-world`.

Basically the same as `minimal-project` (see #4951) plus stub tests, `LICENSE.md` and `README.md`.


magnus-madsen:
> Basically the same as minimal-project (see https://github.com/flix/flix/pull/4951) plus stub tests, LICENSE.md and README.md.

Agreed.

I am looking over the list in Rust doc to see if there is anything else we should include.

magnus-madsen:
Do we want to default to a license (e.g. Apache 2)?

Looking over https://doc.rust-lang.org/cargo/reference/manifest.html maybe we should propose (i.e. include in the toml):

- authors
- description

magnus-madsen:
> > Do we want to default to a license (e.g. Apache 2)?
> > Looking over https://doc.rust-lang.org/cargo/reference/manifest.html maybe we should propose (i.e. include in the toml):
> > 
> > * authors
> > * description
> 
> I would be OK with _suggesting_ these things. But not requiring them (for an application, that is: a library would obviously be different).

We only have three required things. The point is that we would show them in the toml file.

paulbutcher:
So now that #4951 has been merged, I think we now need agreement on this? @mlutze @JonathanStarup

mlutze:
Looks good to me

JonathanStarup:
Never really used these kind of things, so its possibly just a stupid question, but what is the description for? Is it like a single brief paragraph for context or would it contain the same things as the readme?

paulbutcher:
I think this might be good to be merged @magnus-madsen?

examples/projects/hello-world/flix.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name    = "hello-world" # The name of the package.
+description = """
+A hello-world application
+"""
+version = "0.1.0"       # The semantic version of the package.
+flix    = "0.33.0"      # The required version of the Flix compiler.
+license = "Apache-2.0"  # The license applicable to this project (see also LICENSE.md)
+authors = []            # The authors of this project
```
magnus-madsen, line 9: We need some example authors here, so you can tell the format.

examples/projects/hello-world/test/TestMain.flix
```diff
@@ -0,0 +1,2 @@
+@test
```
magnus-madsen, line 1: Both `@test` and `@Test` are supported, but we were moving towards `@Test`.

PR #4954, opened 2022-11-19
feat:

paulbutcher:
This an example of what would be generated by `flix new --template library hello-library`.

For libraries, the `license` and `homepage` fields are compulsory

### DIFF
--- a/examples/projects/hello-library/LICENSE.md
+++ b/examples/projects/hello-library/LICENSE.md
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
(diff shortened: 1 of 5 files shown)


paulbutcher:
And this? @mlutze @JonathanStarup

mlutze:
Why is homepage compulsory?

Should authors be some kind of list?

paulbutcher:
> Why is homepage compulsory?

Because the distribution mechanism for Flix libraries we're planning to use is via GitHub (so we need to know the URL of the project). Perhaps a different name from `homepage` would make this clearer though?

> Should authors be some kind of list?

Good point, yes. I'll make that change.

JonathanStarup:
> > Why is homepage compulsory?
> 
> Because the distribution mechanism for Flix libraries we're planning to use is via GitHub (so we need to know the URL of the project). Perhaps a different name from `homepage` would make this clearer though?
> 
> > Should authors be some kind of list?
> 
> Good point, yes. I'll make that change.

maybe `source`? Don't you already have to have the URL to get the package in the first place? it feels kind of separate to the package, where it is located. Maybe its multiple places?

paulbutcher:
For comparison, the Cargo format has `homepage` which is intended to be a human-readable webpage and `repository` which is intended to point to the source control repository. But Cargo doesn't use GitHub as a distribution mechanism so its requirements are slightly different.

Take your point about having to have the URL to get the package in the first place, but it's good practice to have something within the package to allow you to quickly find out where it came from. And to make sure when publishing that it's being published where it "should" be.

And yes, it could be in multiple places, but there should be a single canonical source (which is what this is intended to represent).

I'm happy with `source` as an alternative to `homepage` (perhaps we use `source` as the equivalent of Cargo's `repository`, with the added meaning of "place where artefacts are published) and `homepage` to mean "human readable website". In which case `source` would be compulsory and `homepage` optional.

paulbutcher:
I think this might be good to be merged @magnus-madsen?

Issue #4850, opened 2022-11-06
Towards Structured Concurrency


magnus-madsen:
Actionable Steps:
- [x] Modify `Sender[t]` and `Receiver[t]` types to carry a region (e.g. `Sender[t, r]`).
- [x] Modify `Channel.{buffered, unbuffered}` to take a region parameters.
- [x] Modify all examples and tests to introduce a fresh region and allocate their channels in those regions.
- [x] Modify the type rule for `ChannelGet` to use the region of the `Receiver[t, r]` as its effect.
- [x] Modify the type rule for `ChannelPut` to use the region of the `Sender[t, r]` as its effect.
- [x] Modify the type rule for `Select` to use the regions of all the `Sender[t, r]` as its effect.
- [x] Modify the spawn expression to take a region parameter. (Using what syntax)?
- [x] Ensure that the effect of spawn stays `Impure` (anything else is incorrect).
- [x] Update all documentation and website.

Actionable, but more difficult steps:
- [x] Add `Expression.Region` to all remaining ASTs. (Its currently erased at some point).
- [x] Add `Expression.Scope` to all remaining ASTs. (Its currently erased at some point).
- [x] Add a `GenRegionClasses` which generates a Region class of the following approximate shape:
```java
class Region {
    private ArrayList<Thread> threads = new ArrayList<Thread>();
    public void spawn(Runnable r) {
        Thread t = new Thread(r);
        threads.add(r);
        t.start();
    }

    public void stop() {
        for (Thread t : threads) t.join();
    }
  }
```
- [x] Modify `CodeGen` to create an instance of the above class upon entry into `Expression.Scope`, then to execute the body, and finally to call `stop` just before exit.
- [x] Modify `CodeGen` such that `Expression.Spawn` now calls `region.spawn(r)`

Later:

- [x] Systematically remove `Scoped`.
- [ ] Systematically remove `'Static` (and with it the `Region` expression).
- [ ] Exploit the above to support `Closable/AutoClosable` resources.

Not yet actionable steps:
- [ ] Decide on the effect of the argument to spawn and the effect of spawn itself.

magnus-madsen:
Further complications/questions:
- We could also register all channels in the Region object. Once the region is destroyed we could mark them as such. This would allow us to trigger a runtime error if a channel is used after its scope has expired.
- We could also trigger a runtime error if a channel is non-empty upon scope exit.

JonathanStarup:
You can have scoped channels? What about nondeterminism?

[...]

PR #4955, opened 2022-11-19
feat: Add regions to `Sender` and `Receiver`

paulbutcher:
This is the start of an attempt to implement the "Modify Sender[t] and Receiver[t] types to carry a region (e.g. Sender[t, r])." task of #4850, but definitely not there yet.

I'm aware that passing `Type.False` for the region parameter to `Type.mkSender` and `Type.mkReceiver` isn't the right thing to do, but I don't see what my options are here without adding a region parameter to `Channel.buffered` etc. because I don't have a region type variable otherwise?

I suspect I'm missing something: could someone point me in the right direction?

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
@@ -841,10 +841,10 @@ object Lowering {
 
       // Special case for Sender[_] and Receiver[_], both of which are rewritten to Concurrent/Channel.Mpmc
       case Type.Cst(TypeConstructor.Sender, loc) =>
-        Type.Cst(TypeConstructor.Enum(Enums.ChannelMpmc, Kind.Star ->: Kind.Star), loc)
+        Type.Cst(TypeConstructor.Enum(Enums.ChannelMpmc, Kind.Star ->: Kind.Bool ->: Kind.Star), loc)
 
       case Type.Cst(TypeConstructor.Receiver, loc) =>
-        Type.Cst(TypeConstructor.Enum(Enums.ChannelMpmc, Kind.Star ->: Kind.Star), loc)
+        Type.Cst(TypeConstructor.Enum(Enums.ChannelMpmc, Kind.Star ->: Kind.Bool ->: Kind.Star), loc)
 
       case Type.Cst(_, _) => tpe0
 
@@ -1512,7 +1512,7 @@ object Lowering {
     * The type of a channel which can transmit variables of type `tpe`
     */
   private def mkChannelTpe(tpe: Type, loc: SourceLocation): Type = {
-    Type.Apply(Type.Cst(TypeConstructor.Enum(Enums.ChannelMpmc, Kind.Star ->: Kind.Star), loc), tpe, loc)
+    Type.Apply(Type.Cst(TypeConstructor.Enum(Enums.ChannelMpmc, Kind.Star ->: Kind.Bool ->: Kind.Star), loc), tpe, loc)
   }
 
   /**
--- a/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
@@ -598,7 +598,7 @@ object Lowering {
     case TypedAst.Expression.NewChannel(exp, tpe, elmTpe, pur, eff, loc) =>
       val e = visitExp(exp)
       val t = visitType(tpe)
-      val chTpe = mkChannelTpe(elmTpe, Type.False, loc)
+      val chTpe = mkChannelTpe(elmTpe, loc)
       val ch = mkNewChannel(e, chTpe, pur, eff, loc)
       val sym = mkLetSym("ch", loc)
       val tuple = LoweredAst.Expression.Tuple(List(LoweredAst.Expression.Var(sym, chTpe, loc), LoweredAst.Expression.Var(sym, chTpe, loc)), t, pur, eff, loc)
@@ -839,12 +839,14 @@ object Lowering {
         case _ => tpe0
       }
 
-      // Special case for Sender[_] and Receiver[_], both of which are rewritten to Concurrent/Channel.Mpmc
-      case Type.Cst(TypeConstructor.Sender, loc) =>
-        Type.Cst(TypeConstructor.Enum(Enums.ChannelMpmc, Kind.Star ->: Kind.Bool ->: Kind.Star), loc)
+      // Special case for Sender[t, _] and Receiver[t, _], both of which are rewritten to Concurrent/Channel.Mpmc[t]
+      case Type.Apply(Type.Apply(Type.Cst(TypeConstructor.Sender, loc), tpe, _), _, _) =>
+        val t = visitType(tpe)
+        Type.Apply(Type.Cst(TypeConstructor.Enum(Enums.ChannelMpmc, Kind.Star ->: Kind.Star), loc), t, loc)
 
-      case Type.Cst(TypeConstructor.Receiver, loc) =>
-        Type.Cst(TypeConstructor.Enum(Enums.ChannelMpmc, Kind.Star ->: Kind.Bool ->: Kind.Star), loc)
+      case Type.Apply(Type.Apply(Type.Cst(TypeConstructor.Receiver, loc), tpe, _), _, _) =>
+        val t = visitType(tpe)
+        Type.Apply(Type.Cst(TypeConstructor.Enum(Enums.ChannelMpmc, Kind.Star ->: Kind.Star), loc), t, loc)
 
       case Type.Cst(_, _) => tpe0
 
@@ -1361,7 +1363,7 @@ object Lowering {
         val locksSym = mkLetSym("locks", loc)
         val pat = mkTuplePattern(List(LoweredAst.Pattern.Cst(Ast.Constant.Int32(i), Type.Int32, loc), LoweredAst.Pattern.Var(locksSym, locksType, loc)), loc)
         val getTpe = Type.eraseTopAliases(chan.tpe) match {
-          case Type.Apply(Type.Apply(_, t, _), _, _) => t
+          case Type.Apply(_, t, _) => t
           case _ => throw InternalCompilerException("Unexpected channel type found.")
         }
         val get = LoweredAst.Expression.Def(Defs.ChannelUnsafeGetAndUnlock, Type.mkImpureUncurriedArrow(List(chan.tpe, locksType), getTpe, loc), loc)
@@ -1511,15 +1513,15 @@ object Lowering {
   /**
     * The type of a channel which can transmit variables of type `tpe`
     */
-  private def mkChannelTpe(tpe: Type, reg: Type, loc: SourceLocation): Type = {
-    Type.Apply(Type.Apply(Type.Cst(TypeConstructor.Enum(Enums.ChannelMpmc, Kind.Star ->: Kind.Bool ->: Kind.Star), loc), tpe, loc), reg, loc)
+  private def mkChannelTpe(tpe: Type, loc: SourceLocation): Type = {
+    Type.Apply(Type.Cst(TypeConstructor.Enum(Enums.ChannelMpmc, Kind.Star ->: Kind.Star), loc), tpe, loc)
   }
 
   /**
     * An expression for a channel variable called `sym`
     */
   private def mkChannelExp(sym: Symbol.VarSym, tpe: Type, loc: SourceLocation): LoweredAst.Expression = {
-    LoweredAst.Expression.Var(sym, mkChannelTpe(tpe, Type.False, loc), loc)
+    LoweredAst.Expression.Var(sym, mkChannelTpe(tpe, loc), loc)
   }
 
   /**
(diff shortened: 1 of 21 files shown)


magnus-madsen:
Here: 

```
pub def buffered(n: Int32): (Sender[t, r], Receiver[t, r]
```

you need an extra argument `r: Region[r]`. 

Then you should modify `CHANNEL_NEW` to take two arguments.

magnus-madsen:
Otherwise your changes look to be going in the right direction.

paulbutcher:
> you need an extra argument r: Region[r].

That is obviously the right long-term solution, but I was trying to do what you suggested in #4850 and first have a PR which adds the region to `Sender` and `Receiver` only without modifying the signature of the channel functions. I assumed that that there was some way to do that, given that that's what you requested.

So I'll change this to do both of the first two tasks within #4850. But be aware that this is going to turn into a monster PR because it will entail changing every use of channels. And I know how much you like monster PRs 😜

magnus-madsen:
> That is obviously the right long-term solution, but I was trying to do what you suggested in https://github.com/flix/flix/issues/4850 and first have a PR which adds the region to Sender and Receiver only without modifying the signature of the channel functions. I assumed that that there was some way to do that, given that that's what you requested.

Ah yes. Uhm. Let me think.

magnus-madsen:
Can you do this instead (to avoid the monster PR):

Have buffered/unbuffered return with `Static` as the region?

paulbutcher:
I'll see if I can make that work 👍

paulbutcher:
I think that this is now ready for review.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1361,7 +1361,7 @@ object Lowering {
         val locksSym = mkLetSym("locks", loc)
         val pat = mkTuplePattern(List(LoweredAst.Pattern.Cst(Ast.Constant.Int32(i), Type.Int32, loc), LoweredAst.Pattern.Var(locksSym, locksType, loc)), loc)
         val getTpe = Type.eraseTopAliases(chan.tpe) match {
-          case Type.Apply(_, t, _) => t
+          case Type.Apply(Type.Apply(_, t, _), _, _) => t
```
paulbutcher, line 1364: @magnus-madsen I'm not really happy with this: it feels rather fragile to me. But I'm not sure that there's any better way? Suggestions welcome.

main/src/library/Concurrent/Channel.flix
```diff
@@ -63,14 +63,14 @@ namespace Concurrent/Channel {
     /// structure on which select can operate.
     ///
     @Internal
-    pub enum Mpmc[a](
+    pub enum Mpmc[a: Type, _: Region](
```
magnus-madsen, line 66: I think you have two options here:

- Either you just drop the region entirely (and ignore it) or
- You introduce the region, but then you should also use it, i.e. in the definition of MutDeque.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -598,7 +598,7 @@ object Lowering {
     case TypedAst.Expression.NewChannel(exp, tpe, elmTpe, pur, eff, loc) =>
       val e = visitExp(exp)
       val t = visitType(tpe)
-      val chTpe = mkChannelTpe(elmTpe, loc)
+      val chTpe = mkChannelTpe(elmTpe, Type.False, loc)
```
magnus-madsen, line 601: See comment below about Mpmc.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
magnus-madsen: This is acceptable. Your other alternative is `.typeArguments.head`, but what you have is fine.

main/src/library/Concurrent/Channel.flix
magnus-madsen: If you keep the region parameter then you should use the actual region available in Lowering.

magnus-madsen: The right thing would be to use the region, but the easy thing (which I lean towards), is to ignore it.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1604,14 +1604,14 @@ object Typer {
         for {
           (constrs, tpe, _, eff) <- visitExp(exp)
           _ <- expectTypeM(expected = Type.Int32, actual = tpe, exp.loc)
-          resultTyp <- liftM(Type.mkTuple(List(Type.mkSender(elmType, loc), Type.mkReceiver(elmType, loc)), loc))
+          resultTyp <- liftM(Type.mkTuple(List(Type.mkSender(elmType, Type.False, loc), Type.mkReceiver(elmType, Type.False, loc)), loc))
```
magnus-madsen, line 1607: Here you should not use Type.False but create a fresh type variable and unify with it.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1604,14 +1604,14 @@ object Typer {
...
           resultPur = Type.Impure
           resultEff = eff
         } yield (constrs, resultTyp, resultPur, resultEff)
 
       case KindedAst.Expression.GetChannel(exp, tvar, loc) =>
         val elmVar = Type.freshVar(Kind.Star, loc, text = FallbackText("elm"))
-        val channelType = Type.mkReceiver(elmVar, loc)
+        val channelType = Type.mkReceiver(elmVar, Type.False, loc)
```
magnus-madsen, line 1614: Here also you should have a regVar next to elmVar.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1623,7 +1623,7 @@ object Typer {
 
       case KindedAst.Expression.PutChannel(exp1, exp2, loc) =>
         val elmVar = Type.freshVar(Kind.Star, loc, text = FallbackText("elm"))
-        val channelType = Type.mkSender(elmVar, loc)
+        val channelType = Type.mkSender(elmVar, Type.False, loc)
```
magnus-madsen, line 1626: And regVar next to elmVar.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
paulbutcher: For this and the following similar comments, I see why this would be the right thing to do, but am I not going to then have to immediately discard that code as soon as I add a region parameter to the channel functions (because I'll be unifying against that parameter and won't have to create one here?)

Or am I misunderstanding?

magnus-madsen: It would be the opposite. (The change I propose would be forwards compatible).

The point is that only GetChannel takes a region parameter. For the other channel operations the region is inferred by the type system (hence the need for type variables that get unified.)

paulbutcher: Oh! Then I am definitely misunderstanding. 

Do you mean GetChannel? Or NewChannel? Above you were talking about adding a `r: Region[r]` parameter to `Channel.buffered`?

Forgive me, I'm confused.

magnus-madsen: Maybe I am confused. My point was for everything but NEW_CHANNEL we should infer the region (i.e. not use Static/False, but rather a type variable).

magnus-madsen: Here I am only talking about what occurs in Typer and only the AST nodes.

paulbutcher: Ah, NewChannel. That makes much more sense. OK, I think I get it, thanks.

paulbutcher:
I think that this might be there now @magnus-madsen

main/src/library/Concurrent/Channel.flix
```diff
@@ -63,7 +63,7 @@ namespace Concurrent/Channel {
     /// structure on which select can operate.
     ///
     @Internal
-    pub enum Mpmc[a](
+    pub enum Mpmc[a: Type](
```
magnus-madsen, line 66: This a: Type is no longer needed.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1601,17 +1601,20 @@ object Typer {
 
 
       case KindedAst.Expression.NewChannel(exp, elmType, loc) =>
+        val regionVar = Type.freshVar(Kind.Bool, loc, text = FallbackText("region"))
```
magnus-madsen, line 1604: This should be False.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1601,17 +1601,20 @@ object Typer {
...
           (constrs, tpe, _, eff) <- visitExp(exp)
           _ <- expectTypeM(expected = Type.Int32, actual = tpe, exp.loc)
-          resultTyp <- liftM(Type.mkTuple(List(Type.mkSender(elmType, loc), Type.mkReceiver(elmType, loc)), loc))
+          resultTyp <- liftM(Type.mkTuple(List(Type.mkSender(elmType, regionVar, loc), Type.mkReceiver(elmType, regionVar, loc)), loc))
           resultPur = Type.Impure
           resultEff = eff
         } yield (constrs, resultTyp, resultPur, resultEff)
 
       case KindedAst.Expression.GetChannel(exp, tvar, loc) =>
+        val regionVar = Type.freshVar(Kind.Bool, loc, text = FallbackText("region"))
```
magnus-madsen, line 1615: This is correct.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1622,8 +1625,9 @@ object Typer {
         } yield (constrs, resultTyp, resultPur, resultEff)
 
       case KindedAst.Expression.PutChannel(exp1, exp2, loc) =>
+        val regionVar = Type.freshVar(Kind.Bool, loc, text = FallbackText("region"))
```
magnus-madsen, line 1628: This is correct.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1645,7 +1649,7 @@ object Typer {
             case KindedAst.SelectChannelRule(sym, chan, body) => for {
               (chanConstrs, chanType, _, chanEff) <- visitExp(chan)
               (bodyConstrs, bodyType, _, bodyEff) <- visitExp(body)
-              _ <- unifyTypeM(chanType, Type.mkReceiver(sym.tvar, sym.loc), sym.loc)
+              _ <- unifyTypeM(chanType, Type.mkReceiver(sym.tvar, Type.False, sym.loc), sym.loc)
```
magnus-madsen, line 1652: This should be regionVar

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -2255,7 +2259,7 @@ object Typer {
         val e = visitExp(exp, subst0)
         val pur = Type.Impure
         val eff = e.eff
-        TypedAst.Expression.NewChannel(e, Type.mkTuple(List(Type.mkSender(elmTpe, loc), Type.mkReceiver(elmTpe, loc)), loc), elmTpe, pur, eff, loc)
+        TypedAst.Expression.NewChannel(e, Type.mkTuple(List(Type.mkSender(elmTpe, Type.False, loc), Type.mkReceiver(elmTpe, Type.False, loc)), loc), elmTpe, pur, eff, loc)
```
magnus-madsen, line 2262: I think instead of all this, we should equip NewChannel with a tvar, and the unify with that, and simply use subst(tvar) here.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
paulbutcher: Sorry Magnus, what's the "this" that should be False? Are you suggesting that I shouldn't have a `regionVar` here at all and instead pass `Type.False` to `mkSender` and `mkReceiver` below. That contradicts what you said earlier, so I guess not? But then I don't know what else you might mean?

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1601,17 +1601,20 @@ object Typer {
 
 
       case KindedAst.Expression.NewChannel(exp, elmType, loc) =>
+        val regionVar = Type.freshVar(Kind.Bool, loc, text = FallbackText("region"))
+
         for {
           (constrs, tpe, _, eff) <- visitExp(exp)
           _ <- expectTypeM(expected = Type.Int32, actual = tpe, exp.loc)
-          resultTyp <- liftM(Type.mkTuple(List(Type.mkSender(elmType, loc), Type.mkReceiver(elmType, loc)), loc))
+          resultTyp <- liftM(Type.mkTuple(List(Type.mkSender(elmType, regionVar, loc), Type.mkReceiver(elmType, regionVar, loc)), loc))
```
magnus-madsen, line 1609: Here we are allocating a new channel (and for now) that is always in the static region, hence regionVar here should really be Type.False. (This will change when you later add regions to NEW_CHANNEL)

paulbutcher:
Oh! I wasn't expect you to merge this yet 🤣. I haven't addressed your comment above: "I think instead of all this, we should equip NewChannel with a tvar, and the unify with that, and simply use subst(tvar) here."

But... I was going to say that I'm not sure that I see how this makes things better. We are already inferring `elmType`. If we infer the tuple type instead (which I think is what you're getting at) it just moves the complexity elsewhere (because we need to pick the element type out of the tuple type).

Or am I missing something?

magnus-madsen:
> Oh! I wasn't expect you to merge this yet 🤣. I haven't addressed your comment above: "I think instead of all this, we should equip NewChannel with a tvar, and the unify with that, and simply use subst(tvar) here."
> 
> But... I was going to say that I'm not sure that I see how this makes things better. We are already inferring `elmType`. If we infer the tuple type instead (which I think is what you're getting at) it just moves the complexity elsewhere (because we need to pick the element type out of the tuple type).
> 
> Or am I missing something?

I think this will become clearer later once we care about both the elmtype and region. Then we can revisit.

PR #4963, opened 2022-11-21
test: add typematch tests for records

mlutze:
related to #4773

### DIFF
--- a/main/test/flix/Test.Exp.TypeMatch.flix
+++ b/main/test/flix/Test.Exp.TypeMatch.flix
@@ -675,7 +675,7 @@ namespace Test/Exp/ReifyType {
     def testMagicList02(): Bool = magicList(3) |> List.isEmpty
 
     //////////////////////////////////////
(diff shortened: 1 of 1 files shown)


magnus-madsen:
> Doesn't work with schemas because monomorph comes after lowering :(

Indeed. What to do then? Reject such types?

mlutze:
> > Doesn't work with schemas because monomorph comes after lowering :(
> 
> Indeed. What to do then? Reject such types?

Yes, that's the only solution I see.

magnus-madsen:
That's fine, better than this behavior.

Same applies to channels then? What else?

mlutze:
> That's fine, better than this behavior.
> 
> Same applies to channels then? What else?

I think that's all. But we need to remember to check recursively, e.g. no functions involving channels either.

magnus-madsen:
Can we wrap this up? (Perhaps by dropping the scheme tests)?

PR #4965, opened 2022-11-21
refactor: combine lower and upper names into single map

mlutze:


magnus-madsen:
Alright, well, I guess we are doing it in one map. Let us see how it goes then.

magnus-madsen:
What's the next steps?

mlutze:
Next I think we change how imports and uses are handled. Maybe we move them to Resolver?

PR #4983, opened 2022-11-23
feat: Channel effects

paulbutcher:
This, I think, addresses the "Modify Typer to account for the region parameter and to place it in the effect of ChannelGet and ChannelPut." task of https://github.com/flix/flix/issues/4850

But draft because:

* I am not entirely confident that I've got the code in Typer correct
* I don't know what I should be doing with `select`

Advice gratefully received

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Typer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Typer.scala
@@ -1607,12 +1607,12 @@ object Typer {
         val regionVar = Type.freshVar(Kind.Bool, loc, text = FallbackText("region"))
         val regionType = Type.mkRegion(regionVar, loc)
         for {
-          (constrs1, tpe1, _, eff1) <- visitExp(exp1)
-          (constrs2, tpe2, _, eff2) <- visitExp(exp2)
+          (constrs1, tpe1, pur1, eff1) <- visitExp(exp1)
+          (constrs2, tpe2, pur2, eff2) <- visitExp(exp2)
           _ <- expectTypeM(expected = regionType, actual = tpe1, exp1.loc)
           _ <- expectTypeM(expected = Type.Int32, actual = tpe2, exp2.loc)
           resultTyp <- liftM(tvar)
-          resultPur = Type.Impure
+          resultPur = Type.mkAnd(pur1, pur2, regionVar, loc)
           resultEff = Type.mkUnion(eff1, eff2, loc)
         } yield (constrs1 ++ constrs2, resultTyp, resultPur, resultEff)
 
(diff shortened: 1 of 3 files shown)


magnus-madsen:
Looks reasonable.

magnus-madsen:
Select should do the same, but it is of course more complicated.

paulbutcher:
> Looks reasonable.

So does that mean that I should mark this as ready for review?

paulbutcher:
> Select should do the same, but it is of course more complicated.

My question over Select is that I don't know how to introduce the `Write` and `Read` constraints. For everything else, they come from `Channel.flix` but that won't work for `select`. And I've not been able to find an example which does this in Scala code anywhere that I can crib from?

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1674,11 +1674,11 @@ object Typer {
...
 
         for {
-          (ruleConstrs, ruleTypes, _, ruleEffs) <- traverseM(rules)(inferSelectRule).map(unzip4)
-          (defaultConstrs, defaultType, _, defaultEff) <- inferDefaultRule(default)
+          (ruleConstrs, ruleTypes, pur1, ruleEffs) <- traverseM(rules)(inferSelectRule).map(unzip4)
+          (defaultConstrs, defaultType, pur2, defaultEff) <- inferDefaultRule(default)
           resultCon = ruleConstrs.flatten ++ defaultConstrs
           resultTyp <- unifyTypeM(tvar :: defaultType :: ruleTypes, loc)
-          resultPur = Type.Impure
+          resultPur = Type.mkAnd(regionVar :: pur2 :: pur1, loc)
```
magnus-madsen, line 1681: -> `Type.mkAnd(pur1, pur2, regionVar, loc) `

magnus-madsen:
For select (which can be a subsequent PR) I think you just look here:

https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/Typer.scala#L1662

magnus-madsen:
Note that right now writing `Read(r)` and `Write(r)` is literally the same as writing just `r`. (Obviously that's not the end game, but that is what it means today.)

paulbutcher:
> Note that right now writing `Read(r)` and `Write(r)` is literally the same as writing just `r`. (Obviously that's not the end game, but that is what it means today.)

Ah! That's the bit of the picture I was missing. And explains why I couldn't fin an example anywhere.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
paulbutcher: Just to check: this is a style thing, not a functionality thing (i.e., the meaning is the same as what I initially wrote, yes?

(just want to check I understand what's going on)

paulbutcher: Ah! I knew there was a reason why I wrote it the way I did: the form you suggest doesn't compile because `pur1` is of type `List[Type]` so I think it has to be the way I wrote it? Or is there another alternative?

paulbutcher: I've changed the variable name to hopefully make this clearer.

magnus-madsen:
Thanks!

PR #4986, opened 2022-11-24
feat: Add region to select effect

paulbutcher:
This is, I think, all that's necessary to address the "Modify the type rule for Select to use the regions of all the Sender[t, r] as its effect." task of #4850?


magnus-madsen:
Seems right.

PR #4988, opened 2022-11-24
feat: add example-flix-project example TOML

paulbutcher:
This is a fuller featured project which demonstrates:

* Maven and Flix dependencies
* Additional directories: resources, dev

@mlutze @JonathanStarup


JonathanStarup:
So this is using the default src/test/dev folders? does resources do anything special or is it just ignored?

paulbutcher:
> So this is using the default src/test/dev folders? does resources do anything special or is it just ignored?

https://paulbutcher.com/book/building.html

Resources are files containing data that the program might need and which get compiled into the Jar. For a web app, these might be things like HTML, CSS, and image files. They're typically accessed via the [getResourceAsStream](https://docs.oracle.com/javase/8/docs/api/java/lang/ClassLoader.html#getResourceAsStream-java.lang.String-) method.

JonathanStarup:
> > So this is using the default src/test/dev folders? does resources do anything special or is it just ignored?
> 
> https://paulbutcher.com/book/building.html
> 
> Resources are files containing data that the program might need and which get compiled into the Jar. For a web app, these might be things like HTML, CSS, and image files. They're typically accessed via the [getResourceAsStream](https://docs.oracle.com/javase/8/docs/api/java/lang/ClassLoader.html#getResourceAsStream-java.lang.String-) method.

Its so unusual to have documentation that I didn't even think about that, Ill take a read 😄

mlutze:
fpkg` means we'll always use the GitHub API to get the dependency?

paulbutcher:
> fpkg` means we'll always use the GitHub API to get the dependency?

My intention is that GitHub will be the first place we support. But medium term, I would imagine that we would expand that to support at least GitLab and BitBucket too (and possibly other sources).

magnus-madsen:
I think this project does two things:
- It shows you can have a bunch of extra files
- It shows how you can add dependencies.
But given that we a already have `hello-world`, I would drop (1) and focus on (2)

(We can extend hello-world with a few extra files, if desired).

Thus is proposed this project becomes something like `project-with-maven-deps`.

(then we can also have `project-with-flix-deps` and `project-with-deps` (or whatever).

Assume you agree (or can be convinced) then I think this project should be a carbon copy of `hello-world` except for the dependecy sections in the toml.

magnus-madsen:
Turning to the dependencies: 

```
[package.dependencies]
mvn."org.postgresql:postgresql" = "42.3.3"
mvn."org.eclipse.jetty:jetty-server" = "11.0.11"
fpkg."com.github.paulbutcher/my-flix-library" = "0.3.1"
```

May I instead propose (because I don't like the "mvn." syntax:

```
[dependencies]
"mvn::org.postgresql:postgresql" = "42.3.3"
"mvn::org.eclipse.jetty:jetty-server" = "11.0.11"
```

paulbutcher:
We can certainly do that, but using the `mvn.` syntax leverages TOML whereas including it all within a string doesn't. This has a couple of implications:

Firstly, we would have to parse the strings ourselves to work out what type of dependency they represent (whereas with the syntax I originally proposed, TOML does this for us).

Secondly with the syntax I originally proposed, the following are both equivalent (this is just how TOML works):

```ini
[package.dependencies]
mvn."org.postgresql:postgresql" = "42.3.3"
mvn."org.eclipse.jetty:jetty-server" = "11.0.11"
fpkg."com.github.paulbutcher/my-flix-library" = "0.3.1"
```

```ini
[package.dependencies.mvn]
"org.postgresql:postgresql" = "42.3.3"
"org.eclipse.jetty:jetty-server" = "11.0.11"

[package.dependencies.mvn]
"com.github.paulbutcher/my-flix-library" = "0.3.1"
```

This flexibility would be lost with the newly proposed syntax.

I don't feel hugely strongly about this, but wanted to call out the implications of making the change.

magnus-madsen:
Perhaps 

```
[dependencies.mvn]
"org.postgresql:postgresql" = "42.3.3"
"org.eclipse.jetty:jetty-server" = "11.0.11"
```

is then what I want (dropping the package prefix). (Then you can also use the other syntax, but we would present it as the above).

paulbutcher:
> Its so unusual to have documentation that I didn't even think about that, Ill take a read 😄

@JonathanStarup did you have a chance to read the documentation? Any comments?

paulbutcher:
> Perhaps
> 
> ```
> [dependencies.mvn]
> "org.postgresql:postgresql" = "42.3.3"
> "org.eclipse.jetty:jetty-server" = "11.0.11"
> ```
> 
> is then what I want (dropping the package prefix). (Then you can also use the other syntax, but we would present it as the above).

That works for me 👍

paulbutcher:
Oh, hang on, I just read what you wrote again and realised you're also suggesting that we move dependencies out from "underneath" `package`.

Again, we can do that, but there are implications.

What you suggest is perfectly sensible for a simple project like this, but be aware that a "fully featured" project is likely to have build flavour-specific dependencies. I have proposed (not in this project, because it's supposed to be simple, but in other examples and the draft documentation) that build flavour specific settings and dependencies are specified as:

```ini
#These settings apply to all build flavours
[package]
source-paths = ["src"]
resource-paths = ["resources"]

#These settings apply only to production builds
[build.prod]
source-paths = ["prod"]
config = { allow-holes = false, allow-debug = false }

#These settings apply only to development builds
[build.dev]
source-paths = ["dev", "test", "bench"]
config = { allow-holes = true, allow-debug = true }
```

Then dependencies are specified as:

```ini
[package.dependencies]
mvn."org.postgresql:postgresql" = "42.3.3"
mvn."org.eclipse.jetty:jetty-server" = "11.0.11"
fpkg."com.github.paulbutcher/my-flix-library" = "0.3.1"

[build.dev.dependencies]
fpkg."com.github.example/debugging-helper" = "2.3.4"
```

If we move package level dependencies out of `package` then where should build flavour specific dependencies live?

magnus-madsen:
I agree with the option part, but how would the dependency part work? In particular, Flix is whole-program so it pretty much has to see all source code.

magnus-madsen:
Oh and on plus size, it does not matter if you include a package and dont use it; Flix will excellent tree shaking (compared to other languages). The only cost is compilation time.

paulbutcher:
It's good to know about the tree shaking, but..

Speaking as a professional developer, I would be uncomfortable with a system in which I couldn't be certain that dependencies I didn't want included in a production build weren't included in a production build.

Imagine, for example, that there's some library that I want to use to facilitate testing. Perhaps that library implements some (deliberately) insecure encryption to allow me to "peek" into encrypted data structures. Helpful during testing, but potentially disastrous if included within a production build.

paulbutcher:
It's worth mentioning that we will *have* to do this for Maven dependencies, even if we don't for Flix dependencies. Take, for example, Jetty. Jetty uses SLF4J for logging, but it doesn't include an SLF4J implementation. That has to be provided by another library.

What that means in practice is that for a development build, you use one SLF4J provider (e.g. `slf4j-simple` which logs to stdout), and for a release build you use another (e.g. `slf4j-log4j` which brings in Log4j).

magnus-madsen:
By the way, I used the term `archtype` earlier, but I regret that it. Its not what I meant.

I simply intend for these projects to be examples; nothing more.

paulbutcher:
> OK, I believe that. Can we do:
> 
> ```
> [dependencies.mvn]
> "org.postgresql:postgresql" = "42.3.3"
> "org.eclipse.jetty:jetty-server" = "11.0.11"
> 
> [dev-dependencies.mvn]
> "org.whatnot" = "42.3.3"
> ```
> 
> following cargo?

Again, yes. But...

I would like the set of build flavours to be under user control. So we'll provide `prod` and `dev` by default, but we should allow additional build flavours to be added for cases where that makes sense.

To give concrete examples:

* perhaps there's a "staging" flavour which is "halfway between" `prod` and `dev`.
* perhaps we're building different flavours of a web app, one of which is designed to be installed on a server, the other runs within Electron.
* perhaps we want slightly different build variants for different geographical regions.

We can't possibly anticipate all of these.

So I would prefer `[dev.dependencies.mvn]` to `[dev-dependencies.mvn]`.

And if we do that, we're essentially allowing the developer to add things to the top-level namespace. That's certainly possible, but increases the opportunity for breaking changes in the future. Which is why I originally proposed that build flavours would live under `[build]`, i.e. `[build.dev.dependencies.mvn]`.

paulbutcher:
> By the way, I used the term archtype earlier, but I regret that it. Its not what I meant.

No problem.

FWIW, I *do* intend that we will implement archetypes (as mentioned in the draft documentation here):

https://paulbutcher.com/book/projects.html

JonathanStarup:
> > Its so unusual to have documentation that I didn't even think about that, Ill take a read 😄
> 
> @JonathanStarup did you have a chance to read the documentation? Any comments?

I like it :) no notes

magnus-madsen:
What's the archetypes? I think Rust has just two: bin and lib, I would probably go for the same. (At least to start with).

paulbutcher:
> What's the archetypes? I think Rust has just two: bin and lib, I would probably go for the same. (At least to start with).

Those two to start with, yes. But very soon after we'll want to add more complex archetypes such as "webapp".

paulbutcher:
@magnus-madsen 

> And if we do that, we're essentially allowing the developer to add things to the top-level namespace. That's certainly possible, but increases the opportunity for breaking changes in the future. Which is why I originally proposed that build flavours would live under [build], i.e. [build.dev.dependencies.mvn].

Do you have any thoughts about this? ^^

Let me know and I'll push ahead with updating this example

paulbutcher:
@magnus-madsen ping?

magnus-madsen:
Ill try to look over these things in the holiday break.

magnus-madsen:
I think this has been superseded by master.

PR #4990, opened 2022-11-25
feat: improve effect error messages

magnus-madsen:


mlutze:
!!!jar

main/src/ca/uwaterloo/flix/language/fmt/FormatType.scala
```diff
@@ -263,7 +289,7 @@ object FormatType {
       case SimpleType.PureArrow(arg, ret) =>
         val argString = delimitFunctionArg(arg)
         val retString = delimit(ret, Mode.Type)
-        s"$argString -> $retString"
+        s"$argString -> $retString \\ {}"
```
mlutze, line 292: The "right" way to do this is to change the transformation to use `PolyEffArrow` with `Empty` in the eff.

main/src/ca/uwaterloo/flix/language/errors/TypeError.scala
```diff
@@ -307,7 +307,7 @@ object TypeError {
     * @param fullType2 the second optional full type in which the second boolean formula occurs.
     * @param loc       the location where the error occurred.
     */
-  case class MismatchedBools(baseType1: Type, baseType2: Type, fullType1: Option[Type], fullType2: Option[Type], loc: SourceLocation)(implicit flix: Flix) extends TypeError {
+  case class MismatchedBools(baseType1: Type, baseType2: Type, fullType1: Option[Type], fullType2: Option[Type], env: RigidityEnv, loc: SourceLocation)(implicit flix: Flix) extends TypeError {
```
mlutze, line 310: `renv` is the standard name

main/src/ca/uwaterloo/flix/language/fmt/FormatType.scala
```diff
@@ -17,14 +17,40 @@ package ca.uwaterloo.flix.language.fmt
...
+import ca.uwaterloo.flix.language.phase.unification.{Substitution, TypeMinimization}
 
 object FormatType {
   /**
     * Transforms the given well-kinded type into a string.
     */
-  def formatType(tpe: Type)(implicit flix: Flix): String = {
-    formatTypeWithOptions(tpe, flix.getFormatOptions)
+  def formatType(tpe: Type, renv: RigidityEnv = RigidityEnv.empty)(implicit flix: Flix): String = {
+    val minimized = TypeMinimization.minimizeType(tpe)
```
mlutze, line 28: Minimizing/renaming unconditionally means that the type displayed in debugger is lying to use.

main/src/ca/uwaterloo/flix/language/fmt/FormatType.scala
magnus-madsen: Are you sure? The function is pure afterall?

main/src/ca/uwaterloo/flix/language/errors/TypeError.scala
magnus-madsen: My bad, thanks!

main/src/ca/uwaterloo/flix/language/fmt/FormatType.scala
magnus-madsen: Yes, but its only a white lie :) (And also would have to right? If we were ever to have variables with less than 1 gazillion ids?)

mlutze: Maybe so, but I foresee a frustrating hour this winter, sitting in my cold office late at night with the rain hammering against my window, trying to debug something with effects and being duped by this evil `toString`...

I propose you add control for this in `FormatOptions.scala`. It already has a parameter for controlling how variable names are handled.

main/src/ca/uwaterloo/flix/language/fmt/FormatType.scala
mlutze: I think so. The idea of `SimpleType` is that it closely resembles the format of the source code type.

mlutze: Here, `a -> b \ {}` could correspond to `SimpleType.PureArrow` or `SimpleType.PolyArrowWithEff`.

It's not a huge problem since we don't parse in that way, but less principled.

main/src/ca/uwaterloo/flix/language/fmt/FormatType.scala
magnus-madsen: This apparently already works because Type calls `formatTypeWithOptions`. Is that okay or do you want to restructure?

mlutze: I'd still prefer to restructure. My thought is that the only job of `formatType` is to extract the options from the `flix` parameter.

magnus-madsen: How about I let you restructure post merge?

mlutze: > How about I let you restructure post merge?

Ok

Issue #5003, opened 2022-11-27
Use `super_cast` more in test cases


magnus-madsen:
E.g.:

- `Test.Exp.Jvm.NewObject.flix`

Issue #5002, opened 2022-11-27
Null casts should use `super_cast` not `unsafe_cast`


magnus-madsen:

magnus-madsen:
Searching for `unsafe_cast null` should be easy.

mlutze:
I'm not sure about this. `super_cast` means "the compiler will check that what I am doing is safe"

magnus-madsen:
> I'm not sure about this. `super_cast` means "the compiler will check that what I am doing is safe"

And its safe for any reference type? Or you mean, its not OK for Flix?

mlutze:
Not OK for Flix IMO. Since in general, `String` means non-nullable String.

PR #5005, opened 2022-11-27
test: use `super_cast` in more JVM tests

jaschdoc:
Is `null` treated as `unit`?

Closes https://github.com/flix/flix/issues/5003

### DIFF
--- a/main/test/flix/Test.Exp.Jvm.InvokeStaticMethod.flix
+++ b/main/test/flix/Test.Exp.Jvm.InvokeStaticMethod.flix
@@ -58,6 +58,6 @@ namespace Test/Exp/Jvm/InvokeStaticMethod {
     def testInvokeStaticMethodWithArrayParam(): Bool \ IO =
         import static java.util.Arrays.fill(Array[##java.lang.Object, Static], ##java.lang.Object): Unit \ IO;
         let a = ["this", "that"];
-        fill(super_cast a, super_cast "foo");
+        fill(unsafe_cast a as Array[##java.lang.Object, Static], super_cast "foo");
         true
 }
--- a/main/test/flix/Test.Exp.Jvm.InvokeStaticMethod.flix
+++ b/main/test/flix/Test.Exp.Jvm.InvokeStaticMethod.flix
@@ -46,7 +46,7 @@ namespace Test/Exp/Jvm/InvokeStaticMethod {
         try {
             of1("end", ["p1", "p2", "p3"])
         } catch {
-            case _: ##java.nio.file.InvalidPathException => super_cast null
+            case _: ##java.nio.file.InvalidPathException => unsafe_cast null as ##java.nio.file.Path
         }
 
     @test
--- a/main/test/flix/Test.Exp.Jvm.InvokeStaticMethod.flix
+++ b/main/test/flix/Test.Exp.Jvm.InvokeStaticMethod.flix
@@ -46,7 +46,7 @@ namespace Test/Exp/Jvm/InvokeStaticMethod {
         try {
             of1("end", ["p1", "p2", "p3"])
         } catch {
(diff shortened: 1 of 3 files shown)


magnus-madsen:
`null` and Unit are not the same, no.

main/test/flix/Test.Exp.Jvm.InvokeStaticMethod.flix
```diff
@@ -46,7 +46,7 @@ namespace Test/Exp/Jvm/InvokeStaticMethod {
         try {
             of1("end", ["p1", "p2", "p3"])
         } catch {
-            case _: ##java.nio.file.InvalidPathException => unsafe_cast null as ##java.nio.file.Path
+            case _: ##java.nio.file.InvalidPathException => super_cast null
```
magnus-madsen, line 49: This should not be a supercast, according to Matt.

jaschdoc: Ah of course because of the safety issue mentioned in https://github.com/flix/flix/issues/5002

main/test/flix/Test.Exp.Jvm.InvokeStaticMethod.flix
```diff
@@ -58,6 +58,6 @@ namespace Test/Exp/Jvm/InvokeStaticMethod {
     def testInvokeStaticMethodWithArrayParam(): Bool \ IO =
         import static java.util.Arrays.fill(Array[##java.lang.Object, Static], ##java.lang.Object): Unit \ IO;
         let a = ["this", "that"];
-        fill(unsafe_cast a as Array[##java.lang.Object, Static], unsafe_cast "foo" as ##java.lang.Object);
+        fill(super_cast a, super_cast "foo");
```
jaschdoc, line 61: is `super_cast a` safe?

It currently throws an error:

```
-- Safety Error -------------------------------------------------- main/test/flix/Test.Exp.Jvm.InvokeStaticMethod.flix

>> The following supercast is unsafe and not allowed.

61 |         fill(super_cast a, super_cast "foo");
                  ^^^^^^^^^^^^
                  the supercast occurs here.

Actual type:      Array[String, false]
Tried casting to: Array[Object, Static]
```

Should it be possible to super_cast an array of `String` to `##...Object`? Each individual element is a subtype but would that be unsafe? E.g. is the width of the types changed in memory?

JonathanStarup: Hey this is the classical question of covariance/contravariance of arrays. It is unsound because if the array now has element type object, then you can insert a list for example, now when the code that sees it as a string array retrieves the "string" there is a runtime error because its a list.

That being said, java does unsoundly allow this ([here is some random discussion](https://stackoverflow.com/questions/28570877/java-covariant-array-bad)) I don't know what we should allow

magnus-madsen: If the element types are compatible then I guess yes. But its a bit of corner case.

jaschdoc: > Hey this is the classical question of covariance/contravariance of arrays. It is unsound because if the array now has element type object, then you can insert a list for example, now when the code that sees it as a string array retrieves the "string" there is a runtime error because its a list.
> 
> That being said, java does unsoundly allow this ([here is some random discussion](https://stackoverflow.com/questions/28570877/java-covariant-array-bad)) I don't know what we should allow

Yeah that was what I thought. Maybe we should just restrict `super_cast` to have something simple that works regardless of Java's unsound rules? Then Flix also avoids the variance issues that Java has.

jaschdoc:
Should be good. Only community build that fails

Issue #4959, opened 2022-11-20
Remove supercast related `JavaSubtypeResult` in Safety


jaschdoc:
Should just return a boolean and if false then perform the same match as in the subtyping function and return the appropriate safety error.

magnus-madsen:
(Fat fingers, sorry)

PR #5006, opened 2022-11-27
refactor: remove `JavaSubtypeResult` from Safety

jaschdoc:
Adds slightly more code duplication but simplifies `isJavaSubtype` and removes redundant data structure.

Closes https://github.com/flix/flix/issues/4959


magnus-madsen:
Thanks!

Issue #4939, opened 2022-11-18
Rework Casts


magnus-madsen:
- [ ] Remove `Prelude.upcast`.
- [x] Modify the syntax of `exp as Type` to `unsafe_cast exp: Type`.
- [x] Modify the syntax of `supercast(exp)` to `super_cast exp`.
- [x] Retain the syntax `upcast exp`.

magnus-madsen:
@jaschdoc I would like to get this in place next. Let me know if you have questions.

mlutze:
We do have one non-snake keyword: `typematch`

magnus-madsen:
True

jaschdoc:
> Remove `Prelude.upcast`

I think this may be harder than just removing it. We'd have to revert to using `unsafe_cast` instead of `upcast` most places, since the boolean formulas are rather complex and that's where we got stuck last time.

magnus-madsen:
> > Remove `Prelude.upcast`
> 
> I think this may be harder than just removing it. We'd have to revert to using `unsafe_cast` instead of `upcast` most places, since the boolean formulas are rather complex and that's where we got stuck last time. So do we just go back to regular unsafe casts for this step?

I would like something very simple:

Given a pure function (possibly curried) then upcast should allow you to cast that to any function (no matter its effects), but with the same argument types.

JonathanStarup:
sorry for the late objection, but i think `unsafeCast` or `unsafecast` is more flix-like. I also don't think it should be toooo hostile since you have to use it for effect subtyping that's not caught by `upcast`.

mlutze:
Shouldn't all effect subtyping be caught by upcast?

magnus-madsen:
We have:

- `super_cast` (underscore)
- `type alias` (space)
- `typematch` (nothing)

Anything we missed? Oh snakeCase.

magnus-madsen:
> > Shouldn't all effect subtyping be caught by upcast?
> 
> if then, i dont see "valid" use of casting no

I don't understand what Matt and you are discussing :)

JonathanStarup:
I often use casts for effect subtyping in foldleft and things like that, so it seems mean to me to make cast hostile if you must use it in some valid situations.

JonathanStarup:
I needed to collect a list of functions, in which some of them printed to the terminal. Thus the pure functions needed to be impure. Upcast didn't work there because of polymorphism in the function type.

magnus-madsen:
[...]

PR #5007, opened 2022-11-27
refactor: allow pure upcast to any effect

jaschdoc:
Closes https://github.com/flix/flix/issues/4939


jaschdoc:
I'm not quite sure what's wrong here. It errors with
```flix
namespace Foldable {
// ...
    pub def foldLeftM(f: (b, a) -> m[b] \ ef, s: b, t: t[a]): m[b] \ ef with Monad[m] =
        let f1 = (x, acc) -> z -> Monad.flatMap(acc, f(z, x));
        s |> Foldable.foldRight(f1, upcast x1 -> Applicative.point(x1), t)
}
```
```
-- Type Error -------------------------------------------------- Foldable.flix

>> No instance of the 'Applicative' class for the type ''0'.

354 |         s |> Foldable.foldRight(f1, upcast x1 -> Applicative.point(x1), t)
                                                       ^^^^^^^^^^^^^^^^^^^^^
                                                       missing instance
```
so it seems like `x1` is cast to type variable which isn't substituted with the concrete type so a TC instance can be derived? Or is that just wrong? :see_no_evil:

magnus-madsen:
Can you try to leave in the parens?

jaschdoc:
> Can you try to leave in the parens?

Same result
```
-- Type Error -------------------------------------------------- Foldable.flix

>> No instance of the 'Applicative' class for the type ''0'.

354 |         s |> Foldable.foldRight(f1, upcast (x1 -> Applicative.point(x1)), t)
                                                        ^^^^^^^^^^^^^^^^^^^^^
                                                        missing instance
```

magnus-madsen:
I think we need to ask @mlutze for help.

mlutze:
This is because upcast kind of separates the expression from the rest of the function.

Roughly, the inference goes:

Visited Expression | Inferred Types | Inferred Type Constraints
--- | --- | ---
`x1` | `x1: t1` | 
`body = Applicative.point(x1)` |  `x1: t1`, `body: t2[t1]` | `Applicative[t2]`
`lam = x1 -> Applicative.point(x1)` | `x1: t1`,  `body: t2[t1]`, `lam: t1 -> t2[t1]` |  `Applicative[t2]`
`up = upcast x1 -> Applicative.point(x1)` | `x1: t1`,  `body: t2[t1]`, `lam: t1 -> t2[t1]` `up: t3`| `Applicative[t2]`

The important part is the `up: t3`. Now when we get the type of the whole expression, we have something involving `t3` but not `t1` and `t2`, which means that when we unify with the signature of the function that says that `Monad[m]`, there is nothing that connects `m` with `t2`.

I think the easy fix is to add an ascription in the expression, something like this: `upcast x1 -> Applicative.point(x1): m[b]`

jaschdoc:
Should work now! Hopefully, we can start to remove the `upcast` function from the Prelude and start to merge `super_cast` with `upcast`.

Issue #4762, opened 2022-10-25
Reject impossible casts


magnus-madsen:
The safety phase should reject clearly impossible casts:

- E.g. Cast from Int32 to Int64 cannot possibly succeed.

magnus-madsen:
@jaschdoc This one seems natural for you to look at :)

magnus-madsen:
I don't expect anything fancy with functions. Just that you cant cast Int32 to Double or whatever.

magnus-madsen:
I think some casts are actually safe; we may have to experiment. What is clear no good is say `Int32` to `String`.

That's what we want to catch. Just the simple, but stupid cases.

PR #5010, opened 2022-11-27
refactor: reject impossible casts

jaschdoc:
Closes https://github.com/flix/flix/issues/4762


jaschdoc:
Do we allow casting from `String` to `Char`?

jaschdoc:
Also, I think this might break lots of casts in the library. Should it maybe be less restrictive?

magnus-madsen:
Impossible should only be truely impossible, e.g. Int32 to String.

magnus-madsen:
> Do we allow casting from `String` to `Char`?

No, that always fails.

magnus-madsen:
I think you should keep it extremely simple:

- A primitive can never be cast to a reference type (or vise versa).
- A non-Boolean primitive cannot be cast to a boolean (or vise versa).

You never reject anything with type variables or the like. The purpose here is just to catch common really stupid things that go wrong. Not to do anything fancy.

jaschdoc:
> A non-Boolean primitive cannot be cast to a boolean (or vise versa).

So `char` for instance is allowed to be cast to `Int32` to read it as a 32-bit number?

magnus-madsen:
The kava spec risallpws that.

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -316,6 +317,52 @@ object Safety {
...
+      Type.Unit :: Type.Bool :: Type.Char ::
+        Type.Float32 :: Type.Float64 :: Type.Int8 ::
+        Type.Int16 :: Type.Int32 :: Type.Int64 ::
+        Type.Str :: Type.BigInt :: Type.BigDecimal :: Nil
+    }.toSet
+
+    (tpe1, tpe2) match {
+
+      // Reference types
+      case (t1, Some(Type.Cst(TypeConstructor.Ref, _))) if primitives.contains(t1) =>
```
magnus-madsen, line 342: Ah, sorry, but reference types I did not mean ` Ref`, but anything that is an object on the JVM.

magnus-madsen:
In other words, its really this:

A primitive can never be cast to a non-primitive type (or vise versa).
A non-Boolean primitive cannot be cast to a boolean (or vise versa).

So the logic would probably be:

if (one of the primitive types, but not bool) then can cast to one of the primitive types (but not bool).

jaschdoc:
Alright, this version only handle reference types and boolean primitives. No special rules wrt. Int32 -> Int64 or anything of the sort between primitive types except for booleans.

PR #5013, opened 2022-11-27
feat: Add region to spawn

paulbutcher:
This fixes the "Modify the spawn expression to take a region parameter." task of #4850

I think it also addresses the "Ensure that the effect of spawn stays Impure" task too (or at least, I can't think what else needs doing?


main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1732,13 +1732,17 @@ object Typer {
...
 
-      case KindedAst.Expression.Spawn(exp, loc) =>
+      case KindedAst.Expression.Spawn(exp1, exp2, loc) =>
+        val regionVar = Type.freshVar(Kind.Bool, loc)
+        val regionType = Type.mkRegion(regionVar, loc)
         for {
-          (constrs, tpe, _, eff) <- visitExp(exp)
+          (constrs1, tpe1, _, eff1) <- visitExp(exp1)
+          (constrs2, tpe2, _, eff2) <- visitExp(exp2)
+          _ <- expectTypeM(expected = regionType, actual = tpe2, exp2.loc)
```
magnus-madsen, line 1741: This should just be unifyTypeM and not expectTypeM

main/test/flix/Test.Exp.Concurrency.Spawn.flix
```diff
@@ -1,57 +1,57 @@
...
-    def testSpawn16(): Unit \ IO = spawn Err(42)
+    def testSpawn16(): Unit \ IO = spawn Err(42) @ Static
 
     @test
-    def testSpawn17(): Unit \ IO = spawn spawn 123
+    def testSpawn17(): Unit \ IO = spawn spawn 123 @ Static @ Static
 
     @test
-    def testSpawn18(): Unit \ IO = spawn spawn spawn 123
+    def testSpawn18(): Unit \ IO = spawn spawn spawn 123 @ Static @ Static @ Static
```
magnus-madsen, line 55: Ultimately these tests will have to be refactored because our goal is to get rid of Static, not to introduce more uses of it.

magnus-madsen: But that could be in a later PR.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -661,7 +661,7 @@ object Lowering {
         case ((sym, c), e) => LoweredAst.Expression.Let(sym, Modifiers.Empty, c, e, t, pur, eff, loc)
       }
 
-    case TypedAst.Expression.Spawn(exp, tpe, pur, eff, loc) =>
+    case TypedAst.Expression.Spawn(exp, _, tpe, pur, eff, loc) =>
       val e = visitExp(exp)
```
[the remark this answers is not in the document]
paulbutcher: I guess when it stops being true, we change this?

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
paulbutcher: Can you help me understand this? Why is it `unifyTypeM` here and not, for example, in the `NewChanel` case above?

main/test/flix/Test.Exp.Concurrency.Spawn.flix
paulbutcher: Agreed, but this PR was already getting out of hand size-wise, and it struck me that this would go when we do the wider Static-removal

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
magnus-madsen: They are functionally the same, but `expectTypeM` generates an error messages saying that "expected" was expected.

When working with type variables, it would be strange to get a message saying the type variable is expected, because it is not what the programmer wrote. Moreover, such unifications usually cannot fail. So its more of a stylistic choice reminding us of that fact.

magnus-madsen: For example, we use `expectTypeM` for the condition of an if-then-else because we know we want a Boolean.

magnus-madsen:
Thanks!

PR #5015, opened 2022-11-28
fix: Fix select default case bug

### LINKED ISSUE #5014: ClassCastException with select
The following:

```
def recvWithDefault(rx: Receiver[Int32, r]): Int32 \ { Read(r), Write(r) } =
    select {
        case x <- Channel.recv(rx) => x
        case _                     => 1
    }

def main(): Unit \ IO = region r {
  let (_tx1, rx1) = Channel.buffered(r, 1);
  recvWithDefault(rx1) |> println
}
```

Gives:

```
java.lang.ClassCastException: class List%Nil%Obj cannot be cast to class dev.flix.runtime.Unit (List%Nil%Obj and dev.flix.runtime.Unit are in unnamed module of loader ca.uwaterloo.flix.language.phase.jvm.FlixClassLoader @126675fd)
        at Def%main%.invoke(Unknown Source)
```

JonathanStarup:
Fixes #5014

Fresh type variables basically just mean unit in the lowering so that caused an issue


paulbutcher:
Thanks for cleaning up my mess (blush).

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1386,9 +1386,10 @@ object Lowering {
     * NB: Does not need to unlock because that is handled inside Concurrent/Channel.selectFrom.
     */
   private def mkSelectDefaultCase(default: Option[LoweredAst.Expression], t: Type, loc: SourceLocation)(implicit flix: Flix): List[LoweredAst.MatchRule] = {
+    val locksType = Types.mkList(Types.ConcurrentReentrantLock, loc)
     default match {
       case Some(defaultExp) =>
-        val pat = mkTuplePattern(List(LoweredAst.Pattern.Cst(Ast.Constant.Int32(-1), Type.Int32, loc), mkWildPattern(loc)), loc)
+        val pat = mkTuplePattern(List(LoweredAst.Pattern.Cst(Ast.Constant.Int32(-1), Type.Int32, loc), LoweredAst.Pattern.Wild(locksType, loc)), loc)
```
paulbutcher, line 1392: I think `mkWildPattern` can be deleted as this was the only place it's used?

JonathanStarup:
> Thanks for cleaning up my mess (blush).

It was very non-obvious so that's fair 😄

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
JonathanStarup: fixed

magnus-madsen:
Sorry to be bothersome, but presumably we also need some tests to ensure this does not happen again :)

magnus-madsen:
> Should I just put the issue program into a test?

That's fine. But how about you write 3 variants and put them into Test.Expression.Select?

magnus-madsen:
Thanks!

PR #5019, opened 2022-11-29
feat: add hole-with-expression expression

### LINKED ISSUE #4978: Add new type of expression hole
Add a new expression: `ident?` to the grammar.

Here is the expected TypedAst:

```
case class HoleWithExp(exp: TypedAst.Expression, tpe: Type, pur: Type, eff: Type, loc: SourceLocation) extends TypedAst.Expression
```

Notably I skipped on the symbol and instead there is always an expression.

(This would also be a Var expression but we don't capture that-- so we can generalize later).

mlutze:
fixes #4978


main/src/ca/uwaterloo/flix/language/phase/Parser.scala
```diff
@@ -1136,6 +1136,10 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
       SP ~ Names.QualifiedDefinition ~ SP ~> ParsedAst.Expression.QName
     }
 
+    def HolyName: Rule1[ParsedAst.Expression.HolyName] = rule {
```
magnus-madsen, line 1139: I saw.

magnus-madsen:
@Daniel-Anker-Hermansen Sorry for the delay, but this adds the ability to write `localVarName?` and have it parsed as a `HoleWithExp`.

PR #5035, opened 2022-12-02
feat: Plumb Region through to code generation

paulbutcher:
This fixes the "Add Expression.Region to all remaining ASTs." task of #4850


paulbutcher:
Note: I'm not sure whether `tpe` needs to be included all the way through, but wanted to get this and the similar task for `Scope` done, after which I guess this will become clearer.

magnus-madsen:
Very good.

PR #5039, opened 2022-12-04
feat: lsp completions for holes with expression

Daniel-Anker-Hermansen:
related to #4554

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
@@ -110,6 +110,12 @@ object CompletionProvider {
     * Process a completion request.
     */
   def autoComplete(uri: String, pos: Position, source: Option[String], currentErrors: List[CompilationMessage])(implicit flix: Flix, index: Index, root: TypedAst.Root): JObject = {
+    val holeCompletions = getHoleExpCompletions(pos, uri, index, root)
+    // If we are currently on a hole the only useful completion is a hole completion.
+    if (holeCompletions.nonEmpty) {
+      return ("status" -> "success") ~ ("result" -> CompletionList(isIncomplete = false, holeCompletions).toJSON)
+    }
+
     //
     // To the best of my knowledge, completions should never be Nil. It could only happen if source was None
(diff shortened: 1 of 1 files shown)


magnus-madsen:
!!!jar

magnus-madsen:
Looking good:

image

I have a few suggestions for how to improve `HoleCompletion`.

- [ ] Order the returned symbols by the number of arguments (i.e. 1 argument functions before 2 argument functions).
- [ ] Order the returned symbols by polymorphism (e.g. if last argument is polymorphic then it should come after a non-polymorphic one).

I think you can probably do (1), but @mlutze probably needs to help with (2).

I would do this in a separate PR.

You can give it a stab or ask @mlutze for help :)

magnus-madsen:
@JonathanStarup Give this JAR a spin with:

```
def main(): Unit \ IO = 
    let y = "def";
    let x = "abc" + y?;
    println("Result is: ${x}")
```

as a starting point. Hit auto-complete after the `?`. 

What more would you like? :)

JonathanStarup:
I like it! but:
1. There is possibly too many results. For `String -> String` it shows `Option.foldWithCont` but this might not be a problem with good ordering
2. I still think this feels like a code action rather than an autocomplete thing. Then maybe you could also begin typing for the result instead of writing in the source and breaking the autocomplete.
3. I didn't follow the decision of `y?` but I would also like to use this without a variable, for `???` What's the use case for `y?`?

magnus-madsen:
!!!jar

magnus-madsen:
I made `HoleCompletion` give an ordered list, but it seems to have had little effect :S

magnus-madsen:
@Daniel-Anker-Hermansen Perhaps something needs to change in the priority thingy.

JonathanStarup:
> > I didn't follow the decision of y? but I would also like to use this without a variable, for ??? What's the use case for y??
> 
> If you write `y` then we know the source type. E.g. in Java you write `object.<complete>`. This is the same. For `???` we would not know what value you want to call something "on".

That makes sense. for `???: String -> Int32` it should return `String.length` etc. this is nice if you code like:
```
l |> map(Add.add(1)) |> map(?power)

```

PR #5050, opened 2022-12-06
feat: better prioritization for holeexp lsp

Daniel-Anker-Hermansen:
Prioritizes first on whether the completion is in the same uri, then number of parameters, then number of polymorphic types.


magnus-madsen:
Does this supersede #5050?

magnus-madsen:
!!!jar

magnus-madsen:
Why does this not make use of the prioritization given by `HoleCompletion`?

magnus-madsen:
https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/api/lsp/provider/HoleCompletion.scala#L56

Daniel-Anker-Hermansen:
It does supersede #5050. It does not make use of HoleCompletion because VSCode needs the completions sorted in lexicographic order.

magnus-madsen:
> 

What does this mean? Can we not adopt the sortText to reflect the order returned by HoleCompletion?

Daniel-Anker-Hermansen:
To do so you would need to convert the index in the list to a string such that these strings are lexicographically sorted the same as the indexes.

magnus-madsen:
> To do so you would need to convert the index in the list to a string such that these strings are lexicographically sorted the same as the indexes.

Seems reasonable? Maybe with some zero pads?

magnus-madsen:
zipWithIndex

Issue #2755, opened 2021-12-01
Minor change to cast syntax


magnus-madsen:
Proposal. Instead of the current:

```
exp as & Pure
```

require:

```
exp as _ & Pure
```

to make clear that the type stays the same.

EDIT: This would also apply to ascribe.

JonathanStarup:
does this also include nested types `List[_]`. That would be nice

magnus-madsen:
> does this also include nested types `List[_]`. That would be nice

Like ensuring its a list but not caring about its element?

JonathanStarup:
Yes

magnus-madsen:
OK. I understand. That is more difficult, but we could certainly do it after this ticket.

magnus-madsen:
I think this should wait until the new import syntax is in.

magnus-madsen:
We agreed to do this.

magnus-madsen:
@jaschdoc One more cast thing. Sigh... :)

We want `unsafecast exp as _ \ IO` not just `unsafecast exp as \ IO`

jaschdoc:
Do we want this for ascriptions in addition to the cast syntax? I.e. `let x: _ \ IO`?

magnus-madsen:
> Do we want this for ascriptions in addition to the cast syntax? I.e. `let x: _ \ IO`?

No

jaschdoc:
Roger

PR #5051, opened 2022-12-06
refactor: add explicit wild type in effect casts

jaschdoc:
Closes https://github.com/flix/flix/issues/2755


magnus-madsen:
I suggest to enforce that you write either:

- `cast exp as Type \ Effect` or 
- `cast exp as _ \ Effect` 

but nothing with `&`. Can we try that?

jaschdoc:
Ok it works now. I'd forgotten about some whitespace I'd previously put in the wrong place.

magnus-madsen:
@stephentetley This changes the cast syntax to require an underscore if you don't cast the type, e.g. now you must write `unsafe_cast exp as _ \ {}` for example.

magnus-madsen:
Thanks!

stephentetley:
Thanks for the heads up Magnus, it should be an easy change now I'm already using `unsafe_cast`.

Issue #4716, opened 2022-10-16
"run" broken on tests within a namespace in VSCode


paulbutcher:
This may be a regression (I seem to recall this working before, but can't be sure). Clicking on "run" below:

<img width="522" alt="Screenshot 2022-10-16 at 11 20 56" src="https://user-images.githubusercontent.com/61163/196030071-b4cd5416-4082-4fb9-b545-a91eada2714b.png">

Results in:

```
flix> :eval JavaUtils.testFindMatches()
-- Resolution Error -------------------------------------------------- <shell>  

>> Definition 'JavaUtils.testFindMatches' is not accessible from the namespace '/'.

2 | println(JavaUtils.testFindMatches()) as \ IO
            ^^^^^^^^^^^^^^^^^^^^^^^^^
            inaccessible definition.

Tip: Mark the definition as public.
```

JonathanStarup:
I think it's been like this for some time, the entry point has to be public but vscode doesn't check

magnus-madsen:
Proposal:

- Hide "Run" for non-public functions.

(The alternative is way too cumbersome).

paulbutcher:
> * Hide "Run" for non-public functions.

This works for me.

magnus-madsen:
I think this has been fixed (?)

paulbutcher:
> I think this has been fixed (?)

I don't think so: or at least I'm still seeing it with a freshly built `flix.jar`? Which change were you expecting would have fixed this?

magnus-madsen:
I guess the change was only in my mind then :)

magnus-madsen:
A one line change somewhere in CodeLensProvider.

PR #5053, opened 2022-12-06
refactor: require entrypoint to be public in `CodeLensProvider`

jaschdoc:
Closes https://github.com/flix/flix/issues/4716


magnus-madsen:
Thanks!

Issue #4967, opened 2022-11-21
Namespace Tasks


mlutze:
Roughly in priority order (@magnus-madsen feel free to reorder)

- [x] Bring local uses through Resolver
- [x] Bring local uses through Typer
- [x] Bring namespace-level uses through Resolver
- [x] Bring namespace-level uses through Typer
- [x] Properly check uses in Resolver
- [x] Put tags in their own namespace
- [x] Support `A.B.C` syntax
- [x] Resolve local variables and type variables in Resolver
- [x] Remove `InstanceSym`
- [x] Allow mixed identifiers in `use X.Y.{Z1, z2}`
- [x] Add error for aliasing with wrong case (`use A.B.{C -> c}`)
- [ ] Ban Reopening Namespaces
- [ ] Check for shadowing uses in `Redundancy` (enable all the disabled tests)
- [ ] Allow `use` and `import` in instances
- [ ] Allow `use` and `import` in classes
- [ ] Check accessibility at `use` sites
- [ ] Create specific Namer Errors ("declaring enum E but there's already a class E over here")
- [ ] Create specific Resolver Errors
- [ ] Significantly reduce usage of `NName`
- [ ] Parallelize Namer
- [ ] Parallelize Resolver
- [ ] Introduce UseSym?
- [ ] Add LSP support for local uses
- [ ] Add LSP support for namespace level uses
- [ ] Check for unused local uses
- [ ] Check for unused namespace-level uses
- [ ] Remove unnecessary validations from Namer
- [ ] Check for `NS-REFACTOR` TODOs everywhere
- [ ] Integrate Symbol.ModuleSym
- [ ] Don't flatten `use X.Y.{Z1, Z2}`
- [ ] Rename visitors in`Resolver` to `visitX`
- [ ] Disallow unqualified `QName`s

PR #5060, opened 2022-12-07
refactor: move var resolution to resolver

mlutze:
related to #4967


magnus-madsen:
RFR?

magnus-madsen:
What is the TLDR on what this does? and what are the next steps?

mlutze:
> RFR?

Ready



> What is the TLDR on what this does? and what are the next steps?

The main thing this does is to move the resolution of variables and type variables to the Resolver. Several errors were also consequently moved to the resolver.

I think the next step is to go through all the easy TODOs and clean up the messes I have made. Then we do the painful step of moving enum cases into their own namespaces.

magnus-madsen:
Merged. Looking forward to the cleanup :)

PR #5061, opened 2022-12-07
refactor: remove redundant type variable from Var expression

mlutze:
related to #4967


magnus-madsen:
RFM?

mlutze:
> RFM?

Now it is.

Issue #4934, opened 2022-11-17
LSP Server should not use null for root, but rather Option


magnus-madsen:

magnus-madsen:
i.e. this https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/api/lsp/LanguageServer.scala#L88

should become `Option[Root]`.

jaschdoc:
Oops I linked this prematurely. This should still be open

magnus-madsen:
@jaschdoc Can you list the providers that are left?

magnus-madsen:
Nevermind, I got some new ideas.

magnus-madsen:
I think once the frontend is more robust in the face of errors, we should only (some of) the providers if the AST is available.

jaschdoc:
> @jaschdoc Can you list the providers that are left?

For the sake of completeness:
- [ ] ImplementationProvider
- [ ] RenameProvider
- [ ] SymbolProvider
- [ ] FindReferencesProvider
- [ ] SemanticTokensProvider
- [ ] InlayHintProvider
- [ ] ShowAstProvider

magnus-madsen:
Fixed.

PR #5071, opened 2022-12-09
refactor: use `Option[Root]` in LanguageServer

jaschdoc:
Related to https://github.com/flix/flix/issues/4934


main/src/ca/uwaterloo/flix/api/lsp/provider/CodeLensProvider.scala
```diff
@@ -26,19 +26,15 @@ object CodeLensProvider {
...
 
   /**
     * Returns code lenses for running entry points.
     */
-  private def getRunCodeLenses(uri: String)(implicit index: Index, root: Root): List[CodeLens] = {
-    if (root == null) {
-      return Nil
-    }
-
+  private def getRunCodeLenses(uri: String)(implicit index: Index, root: Option[Root]): List[CodeLens] = {
```
magnus-madsen, line 37: instead of this map and getOrElse business, I would just do a pattern match so the code looks similar:

root match {
  case None => Nil
  case Some(root) => ...

jaschdoc:
Should be good to go, only `unsafe_cast` that fails in community build.

magnus-madsen:
I think what I want is that "all" the functions should be called but just passed an `Option[Root]` instead of `Root`. But I want them called :)

jaschdoc:
> 

Okay but do you then want all helper functions to take `Option[Root]` instead of `Root` as well or should the immediate function just check if it's `None` or not and return an empty list if its `None`? I see there are lots of null checks and then subsequent code assuming that it's not null. Refactoring to `Option[Root]` in all functions could take quite a while.

magnus-madsen:
> Okay but do you then want all helper functions to take `Option[Root]` instead of `Root` as well or should the immediate function just check if it's `None` or not and return an empty list if its `None`? I see there are lots of null checks and then subsequent code assuming that it's not null. Refactoring to `Option[Root]` in all functions could take quite a while.

Lets pause here and talk about it in person :)

jaschdoc:
Community build fails due to http, but it works otherwise.

PR #5083, opened 2022-12-12
fix: completions after ```=``` in ```def ...```

Daniel-Anker-Hermansen:
related to https://github.com/flix/flix/issues/4554#issuecomment-1344833498

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
@@ -176,7 +176,7 @@ object CompletionProvider {
     val instanceRegex = raw"\s*instance\s+[^s]*".r
 
     // if any of the following matches we do not want any completions
-    val defRegex = raw"\s*def\s+.*".r
+    val defRegex = raw"\s*def\s+[^=]*".r
     val enumRegex = raw"\s*enum\s+.*".r
     val typeAliasRegex = raw"\s*type\s+alias\s+.*".r
     val classRegex = raw"\s*class\s+.*".r


Daniel-Anker-Hermansen:
@magnus-madsen I am only able to reproduce this if i make sure to not have the project in a compileable state at any time from launching the lsp server to asking for the suggestion, which is as expected since the root is null.

magnus-madsen:
Let me try again then :) and report back

magnus-madsen:
Still does not work. 

If I write `def main(): Unit \ IO = Lis` I get no completions (even if the program compiled correctly).

Daniel-Anker-Hermansen:
Weird. This is how it looks for me:
image

JonathanStarup:
On nightly 13dec I see the same as Daniel. The autocomplete is pretty good 👍

magnus-madsen:
> 

That's probably an artifact of how the inferred type is checked (but not unified!) with the declared type.

magnus-madsen:
> > > Random thing: I've found that x? autocomplete doesn't really know the result type if its in tail position
> > 
> > 
> > That's probably an artifact of how the inferred type is checked (but not unified!) with the declared type.
> 
> Huh, i wouldn't have guessed. What would the issue be with unifying them?

It has to be one sided. Consider:

```
def f(): List[Int32] = Nil // Yep, OK
def f(): List[a] = 1 :: Nil // Ooops, not OK
```

Makes sense?

JonathanStarup:
Ah yeah okay, but there would be no downside of the body remembering the instantiation right?

mlutze:
> > > > Random thing: I've found that x? autocomplete doesn't really know the result type if its in tail position
> > > 
> > > 
> > > That's probably an artifact of how the inferred type is checked (but not unified!) with the declared type.
> > 
> > 
> > Huh, i wouldn't have guessed. What would the issue be with unifying them?
> 
> It has to be one sided. Consider:
> 
> ```
> def f(): List[Int32] = Nil // Yep, OK
> def f(): List[a] = 1 :: Nil // Ooops, not OK
> ```
> 
> Makes sense?

If `a` is rigid then we get the appropriate error.

magnus-madsen:
Appears fixed.

PR #5085, opened 2022-12-12
feat: experiment with laziness in DelayList fold

jaschdoc:


magnus-madsen:
A description of the idea would be nice :)

jaschdoc:
Yes, of course. My idea is that maybe we could use reflection to check whether the function passed to `fold(left/right)` is pure and returns a delaylist. In that case we could apply the fold lazily, improving on the "maximally lazy" design

magnus-madsen:
> Yes, of course. My idea is that maybe we could use reflection to check whether the function passed to `fold(left/right)` is pure and returns a delaylist. In that case we could apply the fold lazily, improving the "maximally lazy" design

But is the result of a fold not always eager?

jaschdoc:
Maybe? I think I view it as a generalized `map` function when mapping back to delaylist (of course without any proof to back up this claim). I thought it would be interesting to play around with but it doesn't even compile currently :D

PR #5096, opened 2022-12-13
feat: Add backend Region class

paulbutcher:
This is the first step towards the "Add a GenRegionClasses which generates a Region class" task of #4850. It's creating a `Region` class in the backend which currently does nothing. But it does type check (finally: working out what to do with MonoTypes was a PITA but I think I've got there eventually).

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/jvm/JvmOps.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/jvm/JvmOps.scala
@@ -62,6 +62,7 @@ object JvmOps {
     case MonoType.Int64 => JvmType.PrimLong
     case MonoType.BigInt => JvmType.BigInteger
     case MonoType.Str => JvmType.String
+    case MonoType.Region => JvmType.Object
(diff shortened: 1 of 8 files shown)


main/src/ca/uwaterloo/flix/language/phase/jvm/JvmOps.scala
```diff
@@ -730,6 +731,7 @@ object JvmOps {
     case MonoType.Int64 => Type.Int64
     case MonoType.BigInt => Type.BigInt
     case MonoType.Str => Type.Str
+    case MonoType.Region => Type.mkRegion(Type.Unit, SourceLocation.Unknown) // hack
```
paulbutcher, line 734: This is, as the comment suggests, a hack. But I think it's OK?

magnus-madsen: This seems reasonable. This is really an ugly hack that should be removed by <someone>

PR #5100, opened 2022-12-13
fix: restrict completions to expressions after ```<-```, ```::``` or ```:::```

Daniel-Anker-Hermansen:
It is not perfect and still suggests some things which are not expressions but a lot less than before. Making it better than this would require a lot of changes as some types of completions can both give completions which are expressions and which are not.
Related to https://github.com/flix/flix/issues/4554#issuecomment-1344833498


main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -185,8 +185,14 @@ object CompletionProvider {
     val namespaceRegex = raw"\s*namespace\s+.*".r
     val underscoreRegex = raw"(?:(?:.*\s+)|)_[^s]*".r
 
+    // if ant of the following matches we know the next must be an expression
```
magnus-madsen, line 188: ant -> any

magnus-madsen:
Thanks!

PR #5102, opened 2022-12-13
feat: Plumb region expression through to codegen for use in spawn

paulbutcher:
This just plumbs the expression through (it's not in use yet). Re #4850

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/dbg/PrettyPrinter.scala
+++ b/main/src/ca/uwaterloo/flix/language/dbg/PrettyPrinter.scala
@@ -409,7 +409,7 @@ object PrettyPrinter {
             clazz.getName +
             methods.map(fmtJvmMethod(_, formatter)).mkString("{ ", " ", " }")
 
-        case Expression.Spawn(exp, tpe, loc) => "spawn " + visitExp(exp)
+        case Expression.Spawn(exp1, exp2, tpe, loc) => "spawn " + visitExp(exp1)
 
         case Expression.Lazy(exp, tpe, loc) => "lazy " + visitExp(exp)
 
--- a/main/src/ca/uwaterloo/flix/language/dbg/PrettyPrinter.scala
+++ b/main/src/ca/uwaterloo/flix/language/dbg/PrettyPrinter.scala
@@ -409,7 +409,7 @@ object PrettyPrinter {
             clazz.getName +
             methods.map(fmtJvmMethod(_, formatter)).mkString("{ ", " ", " }")
 
-        case Expression.Spawn(exp1, exp2, tpe, loc) => "spawn " + visitExp(exp1)
+        case Expression.Spawn(exp1, exp2, tpe, loc) => "spawn " + visitExp(exp1) + " @ " + visitExp(exp2)
 
         case Expression.Lazy(exp, tpe, loc) => "lazy " + visitExp(exp)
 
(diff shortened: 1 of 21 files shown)


main/src/ca/uwaterloo/flix/language/dbg/PrettyPrinter.scala
```diff
@@ -409,7 +409,7 @@ object PrettyPrinter {
             clazz.getName +
             methods.map(fmtJvmMethod(_, formatter)).mkString("{ ", " ", " }")
 
-        case Expression.Spawn(exp, tpe, loc) => "spawn " + visitExp(exp)
+        case Expression.Spawn(exp1, exp2, tpe, loc) => "spawn " + visitExp(exp1)
```
magnus-madsen, line 412: Pretty print both exps.

paulbutcher: I left that out because I couldn't see anything else that printed its region argument. But I guess that's because it's discarded before this point in every other case 🤣

Do you have a feeling for the best format? Should I copy the `@r` syntax? Or...?

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1558,7 +1558,7 @@ object Lowering {
         val loc = e.loc.asSynthetic
         val e1 = mkChannelExp(sym, e.tpe, loc) // The channel `ch`
         val e2 = mkPutChannel(e1, e, Type.Impure, Type.mkUnion(e.eff, e1.eff, loc), loc) // The put exp: `ch <- exp0`.
-        val e3 = LoweredAst.Expression.Spawn(e2, Type.Unit, Type.Impure, e2.eff, loc) // Spawn the put expression from above i.e. `spawn ch <- exp0`.
+        val e3 = LoweredAst.Expression.Spawn(e2, LoweredAst.Expression.Region(Type.Unit, loc), Type.Unit, Type.Impure, e2.eff, loc) // Spawn the put expression from above i.e. `spawn ch <- exp0`.
```
magnus-madsen, line 1561: This line will later become a problem, I suspect.

paulbutcher: Yeah, it's basically duplicating `@ Static`. But given that we already have lots of `@ Static` which we'll have to cope with, it struck me that this wasn't a problem for now.

main/src/ca/uwaterloo/flix/language/dbg/PrettyPrinter.scala
magnus-madsen: Just copy the ` @ r` syntax for now.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
magnus-madsen: Yes, let us just try to remind each other.

paulbutcher: I don't think that will be a problem: we're going to have to give Flix a `Static`-ectomy at some point, and this will fall out in the wash when we do so I suspect.

main/src/ca/uwaterloo/flix/language/dbg/PrettyPrinter.scala
paulbutcher: Done

PR #5111, opened 2022-12-14
feat: put enum cases in the enum's companion namespace

mlutze:
related to #4967

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser.scala
@@ -70,7 +70,7 @@ object Parser {
         (source, ast).toSuccess
       case scala.util.Failure(e: org.parboiled2.ParseError) =>
         val loc = SourceLocation(None, source, SourceKind.Real, e.position.line, e.position.column, e.position.line, e.position.column)
-        ca.uwaterloo.flix.language.errors.ParseError(stripLiteralWhitespaceChars(parser.formatError(e, new ErrorFormatter(true, true, true, true, true, 4, 200))), loc).toFailure
+        ca.uwaterloo.flix.language.errors.ParseError(stripLiteralWhitespaceChars(parser.formatError(e)), loc).toFailure
       case scala.util.Failure(e) =>
         ca.uwaterloo.flix.language.errors.ParseError(e.getMessage, SourceLocation.Unknown).toFailure
     }
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser.scala
@@ -316,7 +316,7 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
   // Uses                                                                    //
   /////////////////////////////////////////////////////////////////////////////
   def Use: Rule1[ParsedAst.Use] = rule {
-    keyword("use") ~ WS ~ (Uses.UseOneTag | Uses.UseManyTag | Uses.UseOne | Uses.UseMany)
+    keyword("use") ~ WS ~ (Uses.UseOne | Uses.UseMany)
   }
 
   object Uses {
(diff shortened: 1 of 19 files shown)


mlutze:
@magnus-madsen RFR

The big points are:
* To support `.`-separated namespaces, parsing of things that are `/`- or `.`-separated is merged into one. This includes
  * qualified names
  * tags
  * records
  * array length
  * We might be able to undo this later so that record access is a separate parsing node, but this was the easiest way forward for now.
* To support the current case behavior (accessing cases from the enum's declaring namespace), the symbol table built in `Namer` has an additional field especially for cases
  * Cases are added both under this field and the standard `symbols` field.
  * This behavior will be replaced with "default uses" in the next PR
* `RedBlackTree` is moved to use the companion namespace idea
  * This along with the special `case` handling means that the common name `Leaf` cannot be declared by top-level enums. This issue will be fixed by the upcoming "default uses" change.
* Some more tests are disabled as checks move around

magnus-madsen:
So this is still compatible?

mlutze:
> So this is still compatible?

Mostly. The `RedBlackTree` change might break clients that depend directly on it. But most users should be using `Map` anyway.

magnus-madsen:
What's the next steps briefly?

mlutze:
> What's the next steps briefly?

Next smallish step is to replace the special treatment of _all_ cases with the limited special treatment of only `List` and `Result` and `Option` cases. I predict this will be a ***major breaking change***.

Then more cleanup: regularizing the Resolver, addressing TODOs, and restoring errors, mostly in redundancy.

magnus-madsen:
How much cleanup could be done after this PR but before that breaking change?

magnus-madsen:
I won't have time to look at this until later tonight or so.

magnus-madsen:
Will it be possible that the PR with breaking changes has small changes to the compiler and only large changes in the library?

mlutze:
> How much cleanup could be done after this PR but before that breaking change?

Maybe 25%. A lot of it would be made redundant by the breaking change.

> Will it be possible that the PR with breaking changes has small changes to the compiler and only large changes in the library?

The breaking change in the compiler should mostly be removal of logic. I can try to limit the compiler change as much as possible and leave extra cleanup for later.

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
```diff
@@ -70,7 +70,7 @@ object Parser {
         (source, ast).toSuccess
       case scala.util.Failure(e: org.parboiled2.ParseError) =>
         val loc = SourceLocation(None, source, SourceKind.Real, e.position.line, e.position.column, e.position.line, e.position.column)
-        ca.uwaterloo.flix.language.errors.ParseError(stripLiteralWhitespaceChars(parser.formatError(e)), loc).toFailure
+        ca.uwaterloo.flix.language.errors.ParseError(stripLiteralWhitespaceChars(parser.formatError(e, new ErrorFormatter(true, true, true, true, true, 4, 200))), loc).toFailure
```
magnus-madsen, line 73: What is this?

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
```diff
@@ -1836,21 +1812,21 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
 
     def Class: Rule1[Name.Ident] = UpperCaseName
 
-    def QualifiedClass: Rule1[Name.QName] = UpperCaseQName
+    def QualifiedClass: Rule1[Name.QName] = QName
```
magnus-madsen, line 1815: Should we just delete these different names then?

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -2490,23 +2347,95 @@ object Resolver {
...
+      case Resolution.Declaration(eff: NamedAst.Declaration.Effect) => eff.sym.namespace :+ eff.sym.name
+    }.orElse {
+      // Then see if there's a module with this name declared in our namespace
+      root.symbols.getOrElse(ns0, Map.empty).getOrElse(name, Nil).collectFirst {
+        case Declaration.Namespace(sym, usesAndImports, decls, loc) => sym.ns
+        case Declaration.Class(doc, ann, mod, sym, tparam, superClasses, sigs, laws, loc) => sym.namespace :+ sym.name
+        case Declaration.Enum(doc, ann, mod, sym, tparams, derives, cases, loc) => sym.namespace :+ sym.name
+        case Declaration.Effect(doc, ann, mod, sym, ops, loc) => sym.namespace :+ sym.name
+      }
+    }.orElse {
```
magnus-madsen, line 2421: Hmm. So we are doing this now :)

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
magnus-madsen: This is the only change we should potentially consider for this PR.

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
mlutze: Debugging hack. Removed.

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
mlutze: Maybe. I wasn't sure whether it was best to keep things in the parser or not.

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
mlutze: A helper would probably simplify this.

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
magnus-madsen: Can we remove everything and just have QName? Why not?

mlutze: We can. It just seemed that the style was to separate everything :)

magnus-madsen:
To the 🌔 !

PR #5114, opened 2022-12-14
feat: Add spawn method to Region class and call it

paulbutcher:
Further progress towards #4850

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
@@ -859,20 +859,37 @@ object GenExpression {
     case Expression.Spawn(exp1, exp2, _, loc) =>
       addSourceLine(visitor, loc)
 
-      // Compile the expression, putting a function implementing the Runnable interface on the stack
-      compileExpression(exp1, visitor, currentClass, lenv0, entryPoint)
-      visitor.visitTypeInsn(CHECKCAST, JvmName.Runnable.toInternalName)
+      exp2 match {
+        // The expression represents the `Static` region, just start a thread directly
+        case Expression.Region(_, _) => 
 
-      // make a thread and run it
-      if (flix.options.xvirtualthreads) {
-        visitor.visitMethodInsn(INVOKESTATIC, "java/lang/Thread", "startVirtualThread", s"(${JvmName.Runnable.toDescriptor})${JvmName.Thread.toDescriptor}", false)
-        visitor.visitInsn(POP)
-      } else {
-        visitor.visitTypeInsn(NEW, "java/lang/Thread")
-        visitor.visitInsn(DUP_X1)
-        visitor.visitInsn(SWAP)
-        visitor.visitMethodInsn(INVOKESPECIAL, "java/lang/Thread", "<init>", s"(${JvmName.Runnable.toDescriptor})${JvmType.Void.toDescriptor}", false)
-        visitor.visitMethodInsn(INVOKEVIRTUAL, "java/lang/Thread", "start", AsmOps.getMethodDescriptor(Nil, JvmType.Void), false)
+          // Compile the expression, putting a function implementing the Runnable interface on the stack
+          compileExpression(exp1, visitor, currentClass, lenv0, entryPoint)
+          visitor.visitTypeInsn(CHECKCAST, JvmName.Runnable.toInternalName)
+
+          // make a thread and run it
+          if (flix.options.xvirtualthreads) {
+            visitor.visitMethodInsn(INVOKESTATIC, "java/lang/Thread", "startVirtualThread", s"(${JvmName.Runnable.toDescriptor})${JvmName.Thread.toDescriptor}", false)
+            visitor.visitInsn(POP)
+          } else {
+            visitor.visitTypeInsn(NEW, "java/lang/Thread")
+            visitor.visitInsn(DUP_X1)
+            visitor.visitInsn(SWAP)
+            visitor.visitMethodInsn(INVOKESPECIAL, "java/lang/Thread", "<init>", s"(${JvmName.Runnable.toDescriptor})${JvmType.Void.toDescriptor}", false)
+            visitor.visitMethodInsn(INVOKEVIRTUAL, "java/lang/Thread", "start", AsmOps.getMethodDescriptor(Nil, JvmType.Void), false)
+          }
+
+        case _ => 
+          // Compile the expression representing the region
+          compileExpression(exp2, visitor, currentClass, lenv0, entryPoint)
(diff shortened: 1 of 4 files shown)


paulbutcher:
Note: This doesn't currently honour the `--Xvirtual-threads` flag because to do so will require plumbing the `Flix` object through the `ClassMaker`

magnus-madsen:
> Note: This doesn't currently honour the `--Xvirtual-threads` flag because to do so will require plumbing the `Flix` object through the `ClassMaker`

I think we should that then. Or less preferably remove  `--Xvirtual-threads`. I prefer not to have a broken flag.

Is the path to plumb through really that long?

paulbutcher:
I was intending to do it immediately afterwards in a subsequent PR (I'm trying to keep PRs small!)

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -859,20 +859,37 @@ object GenExpression {
...
-        visitor.visitMethodInsn(INVOKESTATIC, "java/lang/Thread", "startVirtualThread", s"(${JvmName.Runnable.toDescriptor})${JvmName.Thread.toDescriptor}", false)
-        visitor.visitInsn(POP)
-      } else {
-        visitor.visitTypeInsn(NEW, "java/lang/Thread")
-        visitor.visitInsn(DUP_X1)
-        visitor.visitInsn(SWAP)
-        visitor.visitMethodInsn(INVOKESPECIAL, "java/lang/Thread", "<init>", s"(${JvmName.Runnable.toDescriptor})${JvmType.Void.toDescriptor}", false)
-        visitor.visitMethodInsn(INVOKEVIRTUAL, "java/lang/Thread", "start", AsmOps.getMethodDescriptor(Nil, JvmType.Void), false)
+      exp2 match {
+        // The expression represents the `Static` region, just start a thread directly
```
magnus-madsen, line 863: This assumes `Region = Static`, right?

paulbutcher: Correct (which is true, I believe)

magnus-madsen: And so once `Static` is gone, `Region` would be gone too?

paulbutcher:
And let me know if you want me to do the plumbing to get `--Xvirtual-threads` working as part of this PR.

magnus-madsen:
> And let me know if you want me to do the plumbing to get `--Xvirtual-threads` working as part of this PR.

Yes please. I assume it is not excessive?

paulbutcher:
Shouldn't be too bad, no

magnus-madsen:
A few things that can help the pain:

- [ ] Introduce the singleton pattern for `Region` class object (i.e. a static field initialized to itself).
- [ ] Change the CodeGen from the pattern match on region to match on the type. If it is false (i.e. Static) then load the singleton field (regardless of what the expression is).
- [ ] Change all ugly `Scoped` things to cast `null` to the region object. (instead of `Unit`).
- [ ] Change CodeGen to work with `null` where `null` also defaults to the singleton instances.

Later:

- [ ] Systematically remove `Scoped`.
- [ ] Systematically remove `'Static`.

magnus-madsen:
We should not do all of these in this PR, but does it make sense?

paulbutcher:
> We should not do all of these in this PR, but does it make sense?

I'm not sure yet: I think I need to read some code and work out what `regionOf` is doing

magnus-madsen:
> > We should not do all of these in this PR, but does it make sense?
> 
> I'm not sure yet: I think I need to read some code and work out what `regionOf` is doing

It is a giant hack that allows you to say: I know the type of a region `r`, now I can get the actual region value out of thin air. But of course since regions are now becoming **actual runtime values** this cannot possible work. (But we can substitute the global region for now).

paulbutcher:
Right. Let me at least start with a failing test so that we know that whatever we do works.

But before that I'll do the plumbing to get `--Xvirtual-threads` working

paulbutcher:
OK, so I've taken a look at `Scoped`/`regionOf` and I think I see what's going on.

What's the endgame here? Is it:

1. We fix all the types that implement `Scoped` so that they carry a copy of their region around with them, and return that in `regionOf`? Or...
2. We get rid of `Scoped` entirely and change the functions that currently use it so that they take a parameter of type `Region`?

paulbutcher:
So I'm trying to come up with that failing test. This program fails (as expected):

```
def main(): Unit \ IO = region r {
    let a = [1, 2, 3] @ r;
    spawn () @ Scoped.regionOf(a)
}
```

But this test passes?!

```
    @test
    def testSpawnAndRegionOf(): Unit \ IO = region r {
        let a = [1, 2, 3] @ r;
        spawn () @ Scoped.regionOf(a)
    }
```

Any idea what might be going on?

magnus-madsen:
> OK, so I've taken a look at `Scoped`/`regionOf` and I think I see what's going on.
> 
> What's the endgame here? Is it:
> 
>     1. We fix all the types that implement `Scoped` so that they carry a copy of their region around with them, and return that in `regionOf`? Or...
> 
>     2. We get rid of `Scoped` entirely and change the functions that currently use it so that they take a parameter of type `Region`?

I believe (2), but I am still thinking about it.

paulbutcher:
> I believe (2), but I am still thinking about it.

Cool. I was going to say that if the answer was 1, then we should probably have `Receiver` and `Sender` implement `Scoped`. But if it's going away, there's probably no point!

paulbutcher:
> I suggest to mark the failing test as @ignored then we merge this PR and continue with the next tasks I outlined (i.e. supporting the singleton pattern).

Erm, there doesn't seem to be an `@ignored` annotation as far as I can see? But I've commented it out.

paulbutcher:
Err, not sure why the community build is failing?

main/test/flix/Test.Exp.Concurrency.Spawn.flix
```diff
@@ -54,9 +54,9 @@ namespace Test/Exp/Concurrency/Spawn {
     @test
     def testSpawn18(): Unit \ IO = spawn spawn spawn 123 @ Static @ Static @ Static
 
-    @test
-    def testSpawnAndRegionOf01(): Unit \ IO = region r {
-        let a = [1, 2, 3] @ r;
-        spawn () @ Scoped.regionOf(a)
-    }
+//    @test
```
magnus-madsen, line 57: We can use `@Skip` :)

paulbutcher: Done

paulbutcher:
Oh. It looks like `@Skip` doesn't actually skip the test?

magnus-madsen:
> Oh. It looks like `@Skip` doesn't actually skip the test?

That's fun. I guess its a bug. In that case you can comment it.

PR #5121, opened 2022-12-15
Demonstrate channel race condition

paulbutcher:
While working on #4850 I stumbled over what I believe to be a race condition in our channel code. This PR has a modified version of one of the Channel tests `testSelectRandom05` which demonstrates the problem.

The changes I've made are:

1. Reformatting to make it easier to read the test
2. Introduction of a new channel `(txd, rxd)` which every single `spawn` sends to just before it exits
3. A loop at the end of the test to ensure that exactly as many writes are made to this channel as threads are spawned.

For me, this passes about 80% of the time, and hangs 20% of the time (indicating that at least one thread has failed to exit).

### DIFF
--- a/main/test/flix/Test.Exp.Concurrency.Select.flix
+++ b/main/test/flix/Test.Exp.Concurrency.Select.flix
@@ -259,35 +259,100 @@ namespace Test/Exp/Concurrency/Select {
         let (tx15, rx15) = Channel.buffered(r, 0);
         let (tx16, rx16) = Channel.buffered(r, 0);
         let (tx17, rx17) = Channel.buffered(r, 0);
-        spawn { Channel.send((), tx17) ; () } @ r; spawn { Channel.send((), tx16) ; () } @ r; spawn { Channel.send((), tx15) ; () } @ r; spawn { Channel.send((), tx14) ; () } @ r; spawn { Channel.send((), tx17) ; () } @ r; spawn { Channel.send((), tx16) ; () } @ r; spawn { Channel.send((), tx15) ; () } @ r; spawn { Channel.send((), tx14) ; () } @ r; spawn { Channel.send((), tx17) ; () } @ r; spawn { Channel.send((), tx16) ; () } @ r; spawn { Channel.send((), tx15) ; () } @ r; spawn { Channel.send((), tx14) ; () } @ r; spawn { Channel.send((), tx17) ; () } @ r; spawn { Channel.send((), tx16) ; () } @ r; spawn { Channel.send((), tx15) ; () } @ r; spawn { Channel.send((), tx14) ; () } @ r; spawn { select {
-        case _ <- recv(rx17) => Channel.recv(rx14) ; Channel.recv(rx15) ; Channel.recv(rx16)
-        } } @ r; spawn { select {
-        case _ <- recv(rx16) => Channel.recv(rx14) ; Channel.recv(rx15) ; Channel.recv(rx17) ; ()
-        } } @ r; spawn { select {
-        case _ <- recv(rx17) => Channel.recv(rx15) ; Channel.recv(rx16) ; Channel.recv(rx14) ; let (tx171, rx171) = Channel.buffered(r, 0);
-        let (tx172, rx172) = Channel.buffered(r, 0);
-        let (tx173, rx173) = Channel.buffered(r, 0);
-        spawn { Channel.send((), tx173) ; () } @ r; spawn { Channel.send((), tx172) ; () } @ r; spawn { Channel.send((), tx171) ; () } @ r; spawn { Channel.send((), tx173) ; () } @ r; spawn { Channel.send((), tx172) ; () } @ r; spawn { Channel.send((), tx171) ; () } @ r; spawn { Channel.send((), tx173) ; () } @ r; spawn { Channel.send((), tx172) ; () } @ r; spawn { Channel.send((), tx171) ; () } @ r; spawn { Channel.send((), tx173) ; () } @ r; spawn { Channel.send((), tx172) ; () } @ r; spawn { Channel.send((), tx171) ; () } @ r; spawn { select {
-        case _ <- recv(rx172) => Channel.recv(rx173) ; Channel.recv(rx171) ; ()
-        } } @ r; spawn { select {
-        case _ <- recv(rx171) => Channel.recv(rx173) ; Channel.recv(rx172) ; ()
-        case _ <- recv(rx173) => Channel.recv(rx172) ; Channel.recv(rx171) ; ()
-        } } @ r; spawn { select {
-        case _ <- recv(rx171) => Channel.recv(rx173) ; Channel.recv(rx172) ; ()
-        case _ <- recv(rx173) => Channel.recv(rx171) ; Channel.recv(rx172) ; ()
-        case _ <- recv(rx171) => Channel.recv(rx173) ; Channel.recv(rx172) ; ()
-        } } @ r; spawn { select {
-        case _ <- recv(rx171) => Channel.recv(rx173) ; select {
-        case _ <- recv(rx172) => ()
-        case _ <- recv(rx172) => ()
-        } ; ()
-        case _ <- recv(rx173) => Channel.recv(rx171) ; Channel.recv(rx172) ; ()
-        } } @ r; ()
-        case _ <- recv(rx14) => Channel.recv(rx17) ; Channel.recv(rx15) ; Channel.recv(rx16) ; ()
-        } } @ r; spawn { select {
-        case _ <- recv(rx17) => Channel.recv(rx16) ; Channel.recv(rx15) ; Channel.recv(rx14) ; ()
-        } } @ r;
-        ()
+
+        let (txd, rxd) = Channel.buffered(r, 100);
+
+        spawn { Channel.send((), tx17); Channel.send((), txd) } @ r; 
+        spawn { Channel.send((), tx16); Channel.send((), txd) } @ r; 
+        spawn { Channel.send((), tx15); Channel.send((), txd) } @ r; 
+        spawn { Channel.send((), tx14); Channel.send((), txd) } @ r; 
+        spawn { Channel.send((), tx17); Channel.send((), txd) } @ r; 
+        spawn { Channel.send((), tx16); Channel.send((), txd) } @ r; 
+        spawn { Channel.send((), tx15); Channel.send((), txd) } @ r; 
+        spawn { Channel.send((), tx14); Channel.send((), txd) } @ r; 
+        spawn { Channel.send((), tx17); Channel.send((), txd) } @ r; 
+        spawn { Channel.send((), tx16); Channel.send((), txd) } @ r; 
+        spawn { Channel.send((), tx15); Channel.send((), txd) } @ r; 
+        spawn { Channel.send((), tx14); Channel.send((), txd) } @ r; 
+        spawn { Channel.send((), tx17); Channel.send((), txd) } @ r; 
+        spawn { Channel.send((), tx16); Channel.send((), txd) } @ r; 
+        spawn { Channel.send((), tx15); Channel.send((), txd) } @ r; 
+        spawn { Channel.send((), tx14); Channel.send((), txd) } @ r; 
+        spawn { 
+            select {
+                case _ <- recv(rx17) => Channel.recv(rx14) ; Channel.recv(rx15) ; Channel.recv(rx16)
+            };
+            Channel.send((), txd)
+        } @ r; 
+        spawn { 
+            select {
+                case _ <- recv(rx16) => Channel.recv(rx14) ; Channel.recv(rx15) ; Channel.recv(rx17)
+            };
+            Channel.send((), txd)
+        } @ r; 
+        spawn { 
+            select {
+                case _ <- recv(rx17) => 
+                    Channel.recv(rx15) ; Channel.recv(rx16) ; Channel.recv(rx14) ; let (tx171, rx171) = Channel.buffered(r, 0);
+                    let (tx172, rx172) = Channel.buffered(r, 0);
+                    let (tx173, rx173) = Channel.buffered(r, 0);
+                    spawn { Channel.send((), tx173); Channel.send((), txd) } @ r; 
+                    spawn { Channel.send((), tx172); Channel.send((), txd) } @ r; 
+                    spawn { Channel.send((), tx171); Channel.send((), txd) } @ r; 
+                    spawn { Channel.send((), tx173); Channel.send((), txd) } @ r; 
+                    spawn { Channel.send((), tx172); Channel.send((), txd) } @ r; 
+                    spawn { Channel.send((), tx171); Channel.send((), txd) } @ r; 
+                    spawn { Channel.send((), tx173); Channel.send((), txd) } @ r; 
+                    spawn { Channel.send((), tx172); Channel.send((), txd) } @ r; 
+                    spawn { Channel.send((), tx171); Channel.send((), txd) } @ r; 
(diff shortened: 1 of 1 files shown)


paulbutcher:
@JonathanStarup you may have some insight?

paulbutcher:
FWIW, the call stack of hanging threads seems to always be:

```
Unsafe.park(boolean,long)[native method] (file:///opt/homebrew/Cellar/openjdk@17/17.0.4.1_1/libexec/openjdk.jdk/Contents/Home/lib/src.zip!/java.base/jdk/internal/misc/Unsafe.java:-1)
LockSupport.park() (file:///opt/homebrew/Cellar/openjdk@17/17.0.4.1_1/libexec/openjdk.jdk/Contents/Home/lib/src.zip!/java.base/java/util/concurrent/locks/LockSupport.java:341)
AbstractQueuedSynchronizer$ConditionNode.block() (file:///opt/homebrew/Cellar/openjdk@17/17.0.4.1_1/libexec/openjdk.jdk/Contents/Home/lib/src.zip!/java.base/java/util/concurrent/locks/AbstractQueuedSynchronizer.java:506)
ForkJoinPool.unmanagedBlock(ForkJoinPool$ManagedBlocker) (file:///opt/homebrew/Cellar/openjdk@17/17.0.4.1_1/libexec/openjdk.jdk/Contents/Home/lib/src.zip!/java.base/java/util/concurrent/ForkJoinPool.java:3463)
ForkJoinPool.managedBlock(ForkJoinPool$ManagedBlocker) (file:///opt/homebrew/Cellar/openjdk@17/17.0.4.1_1/libexec/openjdk.jdk/Contents/Home/lib/src.zip!/java.base/java/util/concurrent/ForkJoinPool.java:3434)
AbstractQueuedSynchronizer$ConditionObject.awaitUninterruptibly() (file:///opt/homebrew/Cellar/openjdk@17/17.0.4.1_1/libexec/openjdk.jdk/Contents/Home/lib/src.zip!/java.base/java/util/concurrent/locks/AbstractQueuedSynchronizer.java:1578)
Def%awaitUninterruptibly.invoke() (Unknown Source:70)
Cont%Obj.unwind() (Unknown Source:-1)
Def%awaitCondition.invoke() (Unknown Source:449)
Cont%Obj.unwind() (Unknown Source:-1)
Def%getHelper%148872.invoke() (Unknown Source:308)
Cont%Obj.unwind() (Unknown Source:-1)
Def%get%148350.invoke() (Unknown Source:192)
Cont%Obj.unwind() (Unknown Source:-1)
Def%forEach%148011.invoke() (Unknown Source:1134)
Cont%Obj.unwind() (Unknown Source:-1)
Def%testSelectRandom05.invoke() (Unknown Source:355)
Cont%Obj.unwind() (Unknown Source:-1)
Ns.m_testSelectRandom05(Object) (Unknown Source:-1)
NativeMethodAccessorImpl.invoke0(Method,Object,Object[])[native method] (file:///opt/homebrew/Cellar/openjdk@17/17.0.4.1_1/libexec/openjdk.jdk/Contents/Home/lib/src.zip!/java.base/jdk/internal/reflect/NativeMethodAccessorImpl.java:-1)
```

JonathanStarup:
Its a bit hard to wrap my head around at a glance. The channel implementation could be a nice thing to formally proof.

paulbutcher:
> Its a bit hard to wrap my head around at a glance. The channel implementation could be a nice thing to formally proof.

That would be lovely! But I suspect not going to happen quickly 🤣

I'll see if I can get my head round how it's supposed to work and see where that gets me. I may have questions...

JonathanStarup:
the pattern is to send four times and then spawn four things can always recv once no matter what? Thats why it should terminate?

paulbutcher:
No. I've taken the test as-written and added a single `(txd, rxd)` channel. Each `spawn` sends once to `txd` just before it exits and the test has a loop at the end reading from `rxd` which expects to receive exactly as many things written as calls to `spawn`.

JonathanStarup:
I was just trying to see why the test even works without that part.

If getHelper is always in the trace then it could suggest that select might not be handling its post receive handling correctly

magnus-madsen:
The tests - by construction and before any modification - are guaranteed to terminate.

paulbutcher:
So this may not be the problem, but reading the code I think I've seen at least *a* problem. As far as I can tell, we're not calling `awaitUninterruptibly` within a loop. 

I, ahem, happen to know someone who's written a book about this 💅:

Screenshot 2022-12-15 at 17 13 18

The important phrase is "when `await()` returns, it only indicates that the condition *might* be true."

(the same is true of `awaitUninterruptibly`).

This is from the Java documentation:

Screenshot 2022-12-15 at 17 15 47

JonathanStarup:
I do believe we do that, just not in the await method directly. im on my phone right now so im not 100 % sure. I have at least coded with that in mind

paulbutcher:
Ok, that's quite possible. I'll read the code more carefully (the pattern that stands out clearly in Java code may not do so so well in Flix).

JonathanStarup:
hmm reading again it might be wrong yeah

paulbutcher:
So I definitely don't claim to completely understand the channel code just yet. But here are some initial observations:

1. I can see that we do appear to be calling `await` in a loop within `awaitAvailableSpace` (which is itself called by `put`), `getHelper` (which is called by `get`), and `selectHelper` (which is called by `selectFrom`).
2. But we aren't calling it within a loop in the `unBuffered` case at the end of `put`. We should fix this, but I don't think that this is our problem: it might result in the calling thread being unblocked too soon, but I don't think it could cause this bug.
3. We maintain a `size` value for each channel:
    1. Firstly, why do we use this and not use `MutQueue.size` (which should be the same, yes?)
    2. In both `unsafeGetAndLock` and `getHelper` we call `MutQueue.popFront` and then immediately decrease `size` by one. Surely we should only do that in the case that `MutQueue.popFrom` returns `Some(_)`?
    3. (although this only really matters for `getHelper` because it's a bug for `unsafeGetAndLock` to be called with an empty channel).

paulbutcher:
And disappointingly, neither of the issues I mention above are our bug:

* Switching the test to use buffered instead of unbuffered channels doesn't make the bug go away.
* Moving the `size := deref size - 1` line to the `Some(_)` case doesn't make the bug go away.

So whatever's going on, it's neither of these (although I do think they are both issues)

paulbutcher:
I've been banging my head against this for a while and have just realised that what's happening is that occasionally one of the threads dies with an exception. But we don't ever see that because exceptions are silently discarded when testing 😳

Not yet sure what that exception is, but it'll be much easier to debug when they're not silently discarded!

#5123

magnus-madsen:
Thanks for taking the time to look into this. I have no knowledge myself, but I hope you and Jonathan can figure it out :)

paulbutcher:
I've just worked out that the test isn't demonstrating what I think it is: I hadn't correctly worked out the control flow within the test.

There's definitely still a bug in here somewhere, but this test isn't reliably bringing it to light.

I'm going to see if I can fix https://github.com/flix/flix/issues/5123 and then come back to this because right now the lack of visibility is making things unnecessarily difficult.

paulbutcher:
OK, I've worked out what's going on:

1. Ignore this "bug": it's not a real bug and arises from me misunderstanding how the test works.
2. However, I think the issues that I discovered while trying to debug this are real, i.e.:
    1. Not calling `await` in a loop in the `unBuffered` case at the end of `put`.
    2. Decrementing `size` in the `None` case in `getHelper`.
    3. https://github.com/flix/flix/issues/5123
3. @JonathanStarup: perhaps you can weigh in on the above when you have a moment and let me know if you concur?

However, the actual problem I was trying to debug *is* a problem with the `Region` code. Here's what the generated class looks like:

```
class Region {
  private ArrayList<Thread> threads = new ArrayList<Thread>();

  synchronized public void spawn(Runnable r) {
    Thread t = new Thread(r);
    threads.add(t);
    t.start();
  }

  synchronized public void exit() throws InterruptedException {
    for (Thread t: threads)
      t.join();
  }
}
```

The problem arises if a thread within the region itself tries to spawn a thread after the thread in which the region was created has called `exit`.

I suspect that the best way to fix this is to avoid `spawn` and `exit` being synchronized and switch to using `ConcurrentLinkedQueue` for `threads`, but would welcome alternatives if anyone can think of one.

JonathanStarup:
> Not calling await in a loop in the unBuffered case at the end of put.

Hmm, so you're saying that we might spuriously be work up even though no one signalled the condition? In that case, would we need a flag somewhere to be able to check if someone is actually waiting or not?

> Decrementing size in the None case in getHelper.

Oh wow yeah, embarrassing for me

> Exceptions within threads are silently discarded when testing

Yeah this is something we never got around to fixing. I think this is based on the type of thread spawned? I forgot how this should be done.

I

JonathanStarup:
> The problem arises if a thread within the region itself tries to spawn a thread after the thread in which the region was created has called exit.

The `ConcurrentLinkedQueue` sounds sensible.

magnus-madsen:
Just to chip in, exit has to be synchronized. Otherwise there is a race on threads starting.

paulbutcher:
Can you give an example of how that race condition might arise?

magnus-madsen:
Thread is spawned on exit

paulbutcher:
> Thread is spawned on exit

I'm sorry Magnus, I don't follow.

Do you mean a thread is spawned *by the thread that's calling `exit`*? I don't see how that can happen?

Or do you mean a thread is spawned by some other thread in the same region? In which case that's exactly the situation that we have here, and the deadlock is caused by the fact that `exit` and `spawn` are mutually exclusive.

paulbutcher:
> Oh wow yeah, embarrassing for me. I remember thinking about why it was okay to do before the match but it just seems wrong looking at it now.

@JonathanStarup I wouldn't be embarrassed: this stuff is tricky!

Do we need `size` at all? Can we not just use `MutQueue.size`?

magnus-madsen:
We need to consider that threads could be spawned while we wait for other threads to terminate. (In general I am not sure if this problem can be avoided.)

Perhaps we need to move to e.g. `ExecutorService`.

magnus-madsen:
Wrong button, sorry.

paulbutcher:
> We need to consider that threads could be spawned while we wait for other threads to terminate. (In general I am not sure if this problem can be avoided.)

That is *exactly* the issue that we have here. And the deadlock is caused by `spawn` and `exit` being mutually exclusive.

So the solution, I think, is to stop them from being mutually exclusive and instead use a concurrent data structure to store the threads that need to be `join`ed such as `ConcurrentLinkedQueue`.

Or we could, as you suggest, move to using an executor service.

paulbutcher:
> Hmm, so you're saying that we might spuriously be work up even though no one signalled the condition? In that case, would we need a flag somewhere to be able to check if someone is actually waiting or not?

And yes, I believe that the language semantics do allow this to happen. Whether it actually happens in practice depends on all kinds of things, but I think we have to allow for it.

JonathanStarup:
> Do we need `size` at all? Can we not just use `MutQueue.size`?

We need a homogenous typed representation of channels to have a list of them in the select method, this is the Admin part of the channel. Select needs the sizes to check available elements but does not have access to the queue. So it is a neccessry redundancy until the day flix has existential types

paulbutcher:
Thinking about it, work stealing is definitely not the way to go because it won't play nicely with threads which might block (which ours might).

So the choice is, I believe, between cached thread pool or `ConcurrentLinkedQueue`

paulbutcher:
Sorry: you're getting a stream of consciousness as I work my way through this (blush).

I think it has to be `ConcurrentLinkQueue`. We can't use `ExecutorService` because it doesn't provide us with the shutdown semantics we need.

* It has [a `shutdown` method](https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/util/concurrent/ExecutorService.html#shutdown()), but this puts the executor into a state where it will allow existing tasks to complete, but will not accept new tasks. So this will stop any calls to `spawn` from completing after `Region.exit` has been called (which is not what we want).
* It has a [an `awaitTermination` method](https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/util/concurrent/ExecutorService.html#awaitTermination(long,java.util.concurrent.TimeUnit)), but this can only be called after `shutdown` has been called.

So I think we're back to `ConcurrentLinkedQueue` being the only game in town.

Although we could still use a cached thread pool if we wanted to: we just couldn't rely on it for our exit semantics.

JonathanStarup:
I like the `ConcurrentLinkedQueue` solution. It doesn't require as much reading, shouldn't change the code much, and as I see, the only requirement for correctness is that a thread has linked its children before it terminates

magnus-madsen:
Wait a minute... I think the problem is fundamentally unsolvable. I think it is unavoidable that there could be a race on entering `spawn` at the same time as `exit`. I think the only solution is to use synchronization and once `exit` is entered it sets a flag such that any subsequent call to `spawn` will be ignored.

paulbutcher:
> Wait a minute... I think the problem is fundamentally unsolvable. I think it is unavoidable that there could be a race on entering `spawn` at the same time as `exit`. I think the only solution is to use synchronization and once `exit` is entered it sets a flag such that any subsequent call to `spawn` will be ignored.

I'm 100% certain that that's not true Magnus. How about you let me write the code and demonstrate this to you?

JonathanStarup:
> > Wait a minute... I think the problem is fundamentally unsolvable. I think it is unavoidable that there could be a race on entering `spawn` at the same time as `exit`. I think the only solution is to use synchronization and once `exit` is entered it sets a flag such that any subsequent call to `spawn` will be ignored.
> 
> I'm 100% certain that that's not true Magnus. How about you let me write the code and demonstrate this to you?

With the concurrent linked queue version, spawn and exit wouldn't be exclusive right? And if a thread is spawning a child, then it means that exit cannot be at the end of the list, thus it will see its chill once the parent (and others) successfully join and let the exit continue

paulbutcher:
> With the concurrent linked queue version, spawn and exit wouldn't be exclusive right? And if a thread is spawning a child, then it means that exit cannot be at the end of the list, thus it will see its chill once the parent (and others) successfully join and let the exit continue

Exactly

magnus-madsen:
You cannot be sure that you see the same version of the queue without synchronization.

magnus-madsen:
(on my phone>

magnus-madsen:
Show me an implementation that uses synchronized and i can be convinced.

paulbutcher:
> You cannot be sure that you see the same version of the queue without synchronization.

Yes you can. That's exactly what `ConcurrentLinkedQueue` guarantees. It's an unbounded, wait-free, and nonblocking queue.

paulbutcher:
> Show me an implementation that uses synchronized and i can be convinced.

If that's the requirement, then `ConcurrentLinkedQueue` will never solve any problem (it's lock-free).

JonathanStarup:
But its a good point that the [iterator](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentLinkedQueue.html#iterator()) shouldn't be used 

> The returned iterator is a "weakly consistent" iterator that will never throw [ConcurrentModificationException](https://docs.oracle.com/javase/7/docs/api/java/util/ConcurrentModificationException.html), and guarantees to traverse elements as they existed upon construction of the iterator, and may (but is not guaranteed to) reflect any modifications subsequent to construction.

---

> You cannot be sure that you see the same version of the queue without synchronization

The thread that is spawning something itself works as synchronization, the parent being alive is essentially the lock that the exit function is waiting to unlock with join.

paulbutcher:
See #5127

magnus-madsen:
OK, maybe I was running. I will merge the other PR with the fix in.

PR #5127, opened 2022-12-16
feat: Generate Region.exit and call it

paulbutcher:
This replaces #5122 (which I will close).

This completes the Region class tasks of https://github.com/flix/flix/issues/4850 (so the Region class is both being generated and called on scope exit)

It also includes the fix to `size` within the channel code discussed within #5121 because once I addressed the race condition discussed there, I did see occasional failures as a result of this.

With this code, I have run the tests in `Test.Exp.Concurrency.Spawn.flix` 100 times without a single failure.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
@@ -24,7 +24,7 @@ import ca.uwaterloo.flix.language.phase.jvm.ClassMaker.Final.{IsFinal, NotFinal}
 import ca.uwaterloo.flix.language.phase.jvm.ClassMaker.Visibility.{IsPrivate, IsPublic}
 import ca.uwaterloo.flix.language.phase.jvm.ClassMaker._
 import ca.uwaterloo.flix.language.phase.jvm.JvmName.MethodDescriptor.mkDescriptor
-import ca.uwaterloo.flix.language.phase.jvm.JvmName.{DevFlixRuntime, JavaLang, JavaUtil, MethodDescriptor, RootPackage}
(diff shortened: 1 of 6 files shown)


paulbutcher:
So this works, but it does result in the thread calling `exit` busy waiting. We should probably put something in place to reduce the CPU load.

(probably a condition variable which is signalled when a thread exits)

main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
```diff
@@ -994,20 +994,27 @@ object BackendObjType {
...
       thisLoad() ~ INVOKESPECIAL(JavaObject.Constructor) ~ 
-      thisLoad() ~ NEW(BackendObjType.ArrayList.jvmName) ~
-      DUP() ~ invokeConstructor(BackendObjType.ArrayList.jvmName, MethodDescriptor.NothingToVoid) ~
+      thisLoad() ~ NEW(BackendObjType.ConcurrentLinkedQueue.jvmName) ~
+      DUP() ~ invokeConstructor(BackendObjType.ConcurrentLinkedQueue.jvmName, MethodDescriptor.NothingToVoid) ~
       PUTFIELD(ThreadsField) ~
       RETURN()
     ))
 
+    // final public void spawn(Runnable r) {
```
magnus-madsen, line 1013: Any reason not to make the entire class final? (its not very important)

paulbutcher: Just that it didn't occur to me (blush)

magnus-madsen: :)

PR #5133, opened 2022-12-17
feat: Install a default UncaughtExceptionHandler to report exceptions within threads

### LINKED ISSUE #5123: Exceptions within threads are silently discarded when testing
The following test passes 😳

```
@Test
def testBug(): Unit \ IO = region r {
    spawn { bug!("Something bad happened") } @ r;
    Thread.sleep(Time/Duration.fromSeconds(1))
}
```

paulbutcher:
Fixes #5123


paulbutcher:
So this turns out to be a little more complex than it appears at first glance:

1. It's not enough to have a default `UncaughtExceptionHandler` setup in Main because we don't always have a main when running tests, so I've also created one in `FlixClassLoader` (it has to be here because the handler is specific to a thread group)
2. Its also necessary to install a shutdown hook to stop the test framework from swallowing the exception message.

An alternative (which would avoid the need to install a handler in `FlixClassLoader`) would be to setup a per-thread handler within `spawn`, but it's not really any simpler, and it adds an unnecessary overhead to thread creation.

paulbutcher:
Sorry - please ignore this PR for now. I managed to get myself into a mess (left some debugging in the code which confused matters) and it doesn't work the way I thought it did).

Issue #4962, opened 2022-11-21
Add extra region parameter?


magnus-madsen:
Add an extra region parameter to:

Array:
- [ ] Array.slice
- [ ] Array.append
- [ ] Array.map and Array.mapWithIndex
- [ ] Array.flatMap
- [ ] Array.reverse
- [ ] Array.update
- [ ] Array.replace
- [ ] Array.path
- [ ] Array.intersperse
- [ ] Array.intercalate.
- [ ] Array.transpose
- [ ] Array.flatten
- [ ] Array.filter
- [ ] Array.partition - two regions?
- [ ] Array.span - two regions?
- [ ] Array.dropLeft and Array.dropRight
- [ ] Array.dropWhileLeft and Array.dropWhileRight
- [ ] Array.takeLeft and Array.takeRight
- [ ] Array.takeWhileLeft and Array.takeWhileRight
- [ ] Array.zip
- [ ] Array.zipWith
- [ ] Array.unzip - done, has complex effects
- [ ] Array.filterMap
- [ ] Array.iterator
- [ ] Array.enumerator
- [ ] Array.updateSequence
- [ ] Array.sort*

to be used for the result.

stephentetley:
> Add an extra region parameter ... to be used for the result.

So this would be the new definition of `slice` ? 

(Ideally the type of literal slicing could be changed in the compiler to remove the need for the cast.)

~~~

pub def slice(_: Region[r1], b: Int32, e: Int32, a: Array[a, r]): Array[a, r1] \ { Read(r), Write(r1) } = 
    unsafe_cast a[b..e] as Array[a, r1] \ { Read(r), Write(r1) }
    
~~~


Should other imperative libraries (e.g. database access) change to the "answer region" idiom?

magnus-madsen:
> Should other imperative libraries (e.g. database access) change to the "answer region" idiom?

A bit too early to say. (But it does seem like it could be the way forward).

stephentetley:
Shall I take a look at this?

magnus-madsen:
@stephentetley I would like to go ahead with this. I believe all allocating functions has to take a region parameter. Moreover, we have to get rid of `Scoped`. Would you be interested in doing this? :-)

@paulbutcher CC

stephentetley:
Yes - I'm free to look at it at the weekend.

paulbutcher:
[...]

PR #5143, opened 2022-12-18
refactor: add extra region paramater to functions on Array, etc. that return new Arrays etc. (issue #4962)

stephentetley:
This PR adds an "output" region parameter to the signatures of functions on mutables (Array, MutDeque, MutList, MutMap, MutSet) that return new mutables.

Exception - `iterator` and `enumerator` functions and the `Iterable` class left untouched. Firstly, I was having to do extra copying / casting or both to fit using a new region inside the interator. Secondly and a worse problem, changing the signature of the `Iterable` class was having some strange name resolution problems that I couldn't resolve e.g. the (non-mutable) `Set.toString` stopped compiling where it uses `enumerator` in the comprehension.

The naming scheme of the new region params in the function signatures probably needs some discussion. I've wanted to avoid calling the region var `r` as `r` is usually used as a type variable in the existing Array (or MutList...) type and this looks like it would be confusing. I've given the new regions indices so they are usually `r1` but sometimes the existing mutables already use indexed type vars, e.g:

~~~

pub def transpose(r3: Region[r3], a: Array[Array[a, r1], r2]): Array[Array[a, r3], r3] \ { Read(r1, r2), Write(r3) } = ...

~~~

Maybe using `rn` as a mnemonic for "r(egion) n(ew)" would be reasonable choice for the variable name, type var names could continue to use an increasing index:

~~~

pub def transpose(rn: Region[r3], a: Array[Array[a, r1], r2]): Array[Array[a, r3], r3] \ { Read(r1, r2), Write(r3) } = ...

~~~

### DIFF
--- a/main/src/library/Array.flix
+++ b/main/src/library/Array.flix
@@ -18,9 +18,9 @@ instance Scoped[Array[a]] {
     pub def regionOf(_: Array[a, r]): Region[r] = unsafe_cast () as Region[r]
 }
 
-instance Iterable[Array] {
-    pub def iterator(a: Array[a, r]): Iterator[a, r] \ Write(r) = Array.iterator(a)
-}
+/// instance Iterable[Array] {
+///    pub def iterator(a: Array[a, r]): Iterator[a, r] \ Write(r) = Array.iterator(a)
+/// }
 
 namespace Array {
 
@@ -105,7 +105,8 @@ namespace Array {
     ///
     /// Equivalent to the expression `a[b..e]`.
     ///
-    pub def slice(b: Int32, e: Int32, a: Array[a, r]): Array[a, r] \ { Read(r), Write(r) } = a[b..e]
+    pub def slice(_: Region[r1], b: Int32, e: Int32, a: Array[a, r]): Array[a, r1] \ { Read(r), Write(r1) } =
+        unsafe_cast a[b..e] as Array[a, r1] \ { Read(r), Write(r1) }
 
     ///
     /// Returns the array `a` as a list.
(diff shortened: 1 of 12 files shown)


magnus-madsen:
> The naming scheme of the new region params in the function signatures probably needs some discussion. I've wanted to avoid calling the region var r as r is usually used as a type variable in the existing Array (or MutList...) type and this looks like it would be confusing. 

I think we will be moving towards `rh: Region[r]` for "region handle" (but I have not yet fully decided).

magnus-madsen:
OK, I see the naming issue. I think we should sort that out in the future.

main/src/library/Array.flix
```diff
@@ -105,7 +105,8 @@ namespace Array {
     ///
     /// Equivalent to the expression `a[b..e]`.
     ///
-    pub def slice(b: Int32, e: Int32, a: Array[a, r]): Array[a, r] \ { Read(r), Write(r) } = a[b..e]
+    pub def slice(_: Region[r1], b: Int32, e: Int32, a: Array[a, r]): Array[a, r1] \ { Read(r), Write(r1) } =
+        unsafe_cast a[b..e] as Array[a, r1] \ { Read(r), Write(r1) }
```
magnus-madsen, line 109: This will be refactored, but in a separate PR.

magnus-madsen:
Looks good; only issue is the complicated naming.

main/src/library/Array.flix
stephentetley: Thanks - am I reverting the code?

magnus-madsen: No its fine for now :)

stephentetley: Oh yeah - it's already been merged.

Issue #4760, opened 2022-10-25
Re-Engineer Array Syntax


JonathanStarup:
It's easier to write array literals than list literals, which (wrongly, I think) pushes users towards arrays for ease of use.

| Operation | Array | List |
| --- | --- | --- |
| empty | `[] @ r` | `Nil` |
| singleton | `[k] @ r`| `k :: Nil` |
| handful | `[a, b, c, d] @ r` | `a :: b :: c :: d :: Nil` |
| repeated* | `[k; n]` | `List.repeat` |
| slice* | `arr[1 .. 2]` | `List.slice` |
| nth* | `arr[i]` | not available |
| length | `arr.length` | `List.length` |

1. I think that it's weird that these unsafe array operations are so readily available
2. I think that it's weird that List and Array do not have the same syntax support (functional first?)
3. I think that `List.nth` should exist. I know the `argument not to include it, but there are times when "lookup index `i` in linear time" is what they want, and it feels weird to force people to implement it themselves.

Proposed actions (from least to most controversial)
- [ ] Remove `[k; n]` syntax (`Array.repeat` already exists)
- [ ] Remove `arr[1 .. 2]` syntax (`Array.slice` already exists)
- [ ] Add `List#{a, b, c}` syntax like Map/Set
- [ ] Remove `arr.length`
- [ ] Add `List.nth`
- [ ] Remove `arr[i]` syntax (`Array.nth` already exists)

mlutze:
I think we may still want an `unsafeGet` function for arrays. The main reason for using arrays is their speed, so a user should be able to opt into the possibility of OOB errors.

We do ultimately need some syntax to express these operations in the source code (otherwise how do we implement the `Array.get` in the standard library?) but hiding them behind a syntax like `$ArrayGet$(i, a)` is doable.

One thing to add to your table: `x.length` is a special syntax for arrays and it means that we cannot have a record with a field called `length`.

magnus-madsen:
Overall I agree 👍 

I think we should have vectors (which are exactly like arrays, but immutable). They could probably have some of this syntax.

magnus-madsen:
I think we can easily begin with:

- [ ] Replace `[k; n]` by `$ARRAY_REPEAT$` and `Array.repeat` 
- [ ] Replace `a[i..j]` by `$ARRAY_SLICE$` and `Array.slice` 

(The test cases would use `$ARRAY_REPEAT$` and `$ARRAY_SLICE$`

[...]

PR #5145, opened 2022-12-18
refactor: replace `[a; b]` syntax with `$ARRAY_REPEAT$`

jaschdoc:
- [x] Remove option type from `ArrayNew` and the corresponding `optional(... ~ Expression ~ ...)`

Related to https://github.com/flix/flix/issues/4760


jaschdoc:
Just community build that fails due to http

main/test/flix/Test.Exp.Regions.flix
```diff
@@ -80,7 +84,7 @@ namespace Test/Exp/Regions {
     @test
     def testRegions1Region01(): Bool =
         region r {
-            let a = [1; 8];
+            let a = $ARRAY_REPEAT$(r, 1, 8);
```
magnus-madsen, line 87: Can we use `Array.new` in this file?

main/test/flix/Test.Exp.Regions.flix
```diff
@@ -298,10 +302,10 @@ namespace Test/Exp/Regions {
     @test
     def testRegions2Regions04(): Bool \ IO =
         region r1 {
-            let a1 = [1; 8] @ r1;
+            let a1 = $ARRAY_REPEAT$(r1, 1, 8);
             a1[0] == 1 and 1 ==
             region _ {
-                let a2 = [1; 8] @ Static;
+                let a2 = $ARRAY_REPEAT$(unsafe_cast magic() as Region[false], 1, 8);
```
magnus-madsen, line 308: This file is supposed to test regions, not all kinds of ARRAY_REPEAT, unsafe cast, and magic stuff! :)

main/src/library/Array.flix
```diff
@@ -66,12 +66,10 @@ namespace Array {
...
 
     ///
-    /// Returns a fresh array of length `l` filled with the element `x`.
-    ///
-    /// Equivalent to the expression `[x; l]`.
+    /// Returns a fresh array of length `l` filled with the element `x` in the region r.
     ///
     pub def new(r: Region[r], x: a, l: Int32): Array[a, r] \ Write(r) =
-        [x; l] @ r
+        $ARRAY_REPEAT$(r, x, l)
```
magnus-madsen, line 72: Ok, so I like the name new. But then lets call it `$ARRAY_NEW$`.

main/test/flix/Test.Exp.Regions.flix
jaschdoc: Yeah but our safety rules are sort of ruining things here / making things complicated.

$ARRAY_REPEAT$ requires a term level value that carries region information. So naturally you would unsafe_cast unit to some region type but that is rejected as a safety error. Then we have to introduce a function whose return type is a type variable so the compiler accepts an unsafe_cast from a flix primitive type to a type variable.

This is all just to allocate memory in the static region.

jaschdoc: All just to do the equivalent of

[1;8] @ Static

Also, $ARRAY_REPEAT$(false, ...) doesn't work as false is a Boolean not a region (just stating the obvious here)

magnus-madsen: Can we just use a normal region?

main/test/flix/Test.Exp.Regions.flix
jaschdoc: I suppose we could. Just wanted to reduce dependency on the std lib but if we're already using mutmap and mutlist then we may as well just use array

main/test/flix/Test.Exp.Regions.flix
```diff
@@ -1,5 +1,9 @@
 namespace Test/Exp/Regions {
 
+    def magic(): a = typematch () {
```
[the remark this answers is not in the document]
jaschdoc: See comment below :)

main/test/flix/Test.Exp.Regions.flix
magnus-madsen: We are trying to get rid of static anyway

main/test/flix/Test.Exp.Regions.flix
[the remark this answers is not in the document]
jaschdoc: I agree that this is very bad 😃

main/test/ca/uwaterloo/flix/library/TestRandom.flix
```diff
@@ -2,15 +2,15 @@
 def testChoose01(): Option[Int32] \ NonDet =
     region r {
         let rnd = Random.new();
-        let arr = [42i32; 10] @ r;
+        let arr = $ARRAY_REPEAT$(r, 42i32, 10);
```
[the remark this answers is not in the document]
jaschdoc: Yes my bad

main/test/flix/Test.Exp.Regions.flix
magnus-madsen: Great

main/test/flix/Test.Exp.Regions.flix
jaschdoc: Ah perfect! That's much easier haha

jaschdoc:
I've kept the `Static` region for now so the tests are equivalent to what we had before.

magnus-madsen:
Thanks!

PR #5149, opened 2022-12-19
fix: Add rendezvous

### LINKED ISSUE #5128: Should handle spurious wakeups in `put`
As per discussion in #5121, we're currently not calling await in a loop in the unBuffered case at the end of put. We should do so to cope with potential spurious wakeups.

paulbutcher:
Fixes #5128

### DIFF
--- a/main/src/library/Concurrent/Channel.flix
+++ b/main/src/library/Concurrent/Channel.flix
@@ -56,6 +56,19 @@ namespace Concurrent/Channel {
     /// conditions in `waitingPutters` comes from locks created by the get
     /// function or the select function.
     ///
+    /// -- Difference between buffered and unbuffered channels --
+    ///
+    /// An unbuffered channel works *almost* like a buffered channel with a 
+    /// buffer size of 1. The difference is in the behaviour of the thread
+    /// executing `put`.
+    ///
+    /// For a buffered channel, `put` blocks until space is available. Once it
+    /// is, the value is added to the channel and `put` returns.
+    ///
+    /// For an unbuffered channel, the thread executing `put` needs to perform
+    /// a [synchronous rendezvous](https://en.wikipedia.org/wiki/Barrier_(computer_science))
+    /// with the thread executing `get`.
+    ///
     /// -- Mpmc / MpmcAdmin split --
     /// The select function must work on heterogeneous channel collections,
     /// which cannot happen without existential types. This means that the
@@ -95,8 +108,8 @@ namespace Concurrent/Channel {
     ///   `waitingGetters` - a set of conditions that are waiting for an
     ///       element. This set is notified and cleared after each new element.
     ///       An available element is not guaranteed when notified.
-    ///   `rendezvous` - a condition that is waiting for a rendezvous (only
-    ///       used for unbuffered channels).
+    ///   `rendezvous` - a condition that is waiting for a synchronous rendezvous
+    ///       Between putter and getter (only used for unbuffered channels).
     ///   `waitingPutters` - a condition that is waiting for an element.
     ///       The condition is notified after each new element. Available space
     ///       is not guaranteed when notified.
@@ -441,6 +454,8 @@ namespace Concurrent/Channel {
     def onGetComplete(c: MpmcAdmin): Unit \ IO =
         let MpmcAdmin(_, _, unBuffered, _, _, _, rendezvous, waitingPutters) = c;
         if (unBuffered)
+            // Perform synchronous rendezvous with the putter (the putter will 
+            // signal waiting putters once that rendezvous is complete)
             signalCondition(rendezvous)
         else 
             signalCondition(waitingPutters)
--- a/main/src/library/Concurrent/Channel.flix
+++ b/main/src/library/Concurrent/Channel.flix
@@ -440,16 +440,16 @@ namespace Concurrent/Channel {
     def onGetComplete(c: MpmcAdmin): Unit \ IO =
         let MpmcAdmin(_, channelLock, unBuffered, _, _, _, rendezvous, waitingPutters) = c;
 
+        // In the unbuffered case, rendezvous with the putter from which we just received a value.
+        if (unBuffered) discard awaitBarrier(rendezvous)
+        else ();
+
         // Signal waiting setters that the channel has space
         signalCondition(waitingPutters);
         // Since there is only one condition there is no cleaning up to do.
 
         // Unlock the channel
-        unlock(channelLock);
-
-        // In the unbuffered case, rendezvous with the putter from which we just received a value.
-        if (unBuffered) discard awaitBarrier(rendezvous)
-        else ()
+        unlock(channelLock)
 
 
     // ------------- Helper Methods --------------------------------------------
--- a/main/src/library/Concurrent/Channel.flix
+++ b/main/src/library/Concurrent/Channel.flix
@@ -22,6 +22,9 @@ namespace Concurrent/Channel {
     use Concurrent/ReentrantLock.newLock
     use Concurrent/Condition.Condition
     use Concurrent/ReentrantLock.newCondition
+    use Concurrent/CyclicBarrier.CyclicBarrier
+    use Concurrent/CyclicBarrier.newCyclicBarrier
+    use Concurrent/CyclicBarrier.awaitBarrier
 
     ///
     /// An enum that represents a multiple-producer, multiple-consumer (mpmc)
@@ -108,8 +111,9 @@ namespace Concurrent/Channel {
     ///   `waitingGetters` - a set of conditions that are waiting for an
     ///       element. This set is notified and cleared after each new element.
     ///       An available element is not guaranteed when notified.
-    ///   `rendezvous` - a condition that is waiting for a synchronous rendezvous
-    ///       Between putter and getter (only used for unbuffered channels).
+    ///   `rendezvous` - a cyclic barrier used to implement synchronous
+    ///       rendezvous between putter and getter (only used for unbuffered
+    ///       channels).
     ///   `waitingPutters` - a condition that is waiting for an element.
     ///       The condition is notified after each new element. Available space
     ///       is not guaranteed when notified.
@@ -121,7 +125,7 @@ namespace Concurrent/Channel {
         Int32, // maxSize
         Ref[Int32, Static], // size
         MutList[(ReentrantLock, Condition), Static], // waitingGetters
-        Condition, // rendezvous
+        CyclicBarrier, // rendezvous
         Condition // waitingPutters
     )
 
@@ -145,7 +149,7 @@ namespace Concurrent/Channel {
                 if (unBuffered) 1 else bufferSize,
                 size,
                 new MutList(Static),
-                newCondition(reentrantLock),
+                newCyclicBarrier(2),
                 newCondition(reentrantLock)
             ),
             new MutDeque(Static)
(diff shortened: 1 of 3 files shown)


main/src/library/Concurrent/Channel.flix
```diff
@@ -166,14 +170,12 @@ namespace Concurrent/Channel {
...
         // before continuing.
-        /* TODO: Optimization possibility: This await could be faster with a
-                 separate condition such that the hand-off signal does not have
-                 to fight for the channel lock with the threads that actually
-                 want to insert an element (unblocking from await requires that
-                 you re-aquire the lock).
-        */
-        if (unBuffered) awaitCondition(waitingPutters)
-        else ();
+        if (unBuffered) {
```
JonathanStarup, line 173: It might, at some point make sense to split the data into buffered/unbuffered

main/src/library/Concurrent/Channel.flix
```diff
@@ -420,14 +435,15 @@ namespace Concurrent/Channel {
...
-        let MpmcAdmin(_, _, _, _, _, _, waitingPutters) = c;
-        // Signal waiting setters that the channel has space
-        signalCondition(waitingPutters)
-        // Since there is only one condition there is no cleaning up to do.
+    def onGetComplete(c: MpmcAdmin): Unit \ IO =
+        let MpmcAdmin(_, _, unBuffered, _, _, _, rendezvous, waitingPutters) = c;
+        if (unBuffered)
+            signalCondition(rendezvous)
+        else 
+            signalCondition(waitingPutters)
```
JonathanStarup, line 446: Why is this not done always? `put` will call `awaitCondition(waitingPutters)` and to me it doesn't seem like anything wakes it up in the unbuffered case

main/src/library/Concurrent/Channel.flix
```diff
@@ -95,6 +95,8 @@ namespace Concurrent/Channel {
     ///   `waitingGetters` - a set of conditions that are waiting for an
     ///       element. This set is notified and cleared after each new element.
     ///       An available element is not guaranteed when notified.
+    ///   `rendezvous` - a condition that is waiting for a rendezvous (only
```
JonathanStarup, line 98: As a reader, this almost cyclic definition is a bit confusing. Maybe there's a less abstract name than rendezvous? Can you expand a bit on this?

JonathanStarup (changes requested):
I don't fully understand how rendezvous works, so I'm not sure it's correct or not. It feels like this implementation deserves its own design & explaination document (not just this change) since its so complex.

main/src/library/Concurrent/Channel.flix
paulbutcher: Yes, agreed. But I wanted to keep the changes for this as simple as possible :-)

main/src/library/Concurrent/Channel.flix
paulbutcher: So the buffered case works exactly as before: the `rendezvous` condition is completely unused in that case.

For the buffered case we have to proceed in two steps:

1. First we unblock the thread which successfully sent a value (the value that was just received)
2. Then we unblock `waitingPutters`.

This second step is performed by the successful putter [here](https://github.com/flix/flix/blob/61eeb02cb756c0a881c4acd6fb868eca06fc65e6/main/src/library/Concurrent/Channel.flix#L175)

I think this two step approach is unavoidable: we need to be sure that the single thread waiting on `rendezvous` (the one that wrote the value that's just been read) is woken up first, or the danger is that another waiting putter will be woken up first, write a new value, and now `size` is 1.

paulbutcher:
> I don't fully understand how rendezvous works, so I'm not sure it's correct or not. It feels like this implementation deserves its own design & explaination document (not just this change) since its so complex.

Agreed. Where would be the right place for this to live

paulbutcher:
@JonathanStarup I've pushed some additional documentation: let me know what you think.

paulbutcher:
@JonathanStarup hold fire: this conversation has made me realise that we can do this much more easily.

I'll ping you when this is ready for you to take another look.

paulbutcher:
@JonathanStarup I think that this is ready for you to take another look.

Our conversation above made me realise that we were using a condition where in fact this is precisely the case for which [`CyclicBarrier`](https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/util/concurrent/CyclicBarrier.html) is intended. So I've switch `rendezvous` to be a `CyclicBarrier`, the result of which is that the structure is much closer to what you originally implemented.

This is, BTW, almost exactly the optimisation you suggested in your "TODO" comment.

paulbutcher:
Damn: apparently not. That's annoying: it's run successfully many times on my machine.

I'll investigate, but if you spot anything, please do let me know.

JonathanStarup:
> > I don't fully understand how rendezvous works, so I'm not sure it's correct or not. It feels like this implementation deserves its own design & explaination document (not just this change) since its so complex.
> 
> Agreed. Where would be the right place for this to live?

I don't think there's precedent for md/txt files in the std lib so many just a long comment at the end. Even though I've implemented Channel twice now, I've forgot again how it really works. I was looking at thread A calling put, and then thread B calling get. It seemed to me this could deadlock, but it can't because maxSize is strictly positive. Its just an example on why a paper proof or explanation would be really nice.

paulbutcher:
> Wait, in put, does the thread not maintain the channel lock when waiting for available space?

Sorry @JonathanStarup, I'm not sure I understand the question? This code hasn't really changed from what was there before I don't think?

paulbutcher:
(BTW, if you edit a comment, GitHub doesn't alert, so it took me a while to see your question)

JonathanStarup:
> > Wait, in put, does the thread not maintain the channel lock when waiting for available space?
> 
> Sorry @JonathanStarup, I'm not sure I understand the question? This code hasn't really changed from what was there before I don't think?

no no I'm just reanalysing how it works in the first place. it just seems like that a buffered channel with size 1 where you call send, spawn send(now blocking), sleep, and then recv should block.

paulbutcher:
Erm, you might have to show me some code to be sure I'm thinking of the same thing you are, but if I understand what you're saying then yes, I think that that should deadlock (and always would have done)?

paulbutcher:
BTW: progress report. I've managed to get this to deadlock on my machine (although it took running the test in a tight loop repeatedly, whereas on the build server it clearly happens pretty reliably).

I think I'm going to have to instrument the code to work out what's happening. I'm not yet sure whether this is a problem I've introduced with `rendezvous` or whether this is a problem that's always been there and `rendezvous` has just surfaced it.

paulbutcher:
As is often the case, I woke up this morning with an understanding of my mistake: I was guilty of premature (and incorrect) optimisation. I thought that we could release waiting putters before the rendezvous, but this is not correct. What could happen is that one of the waiting putters could place a value on the stream and hit the rendezvous before the getter did. This would result the two putters being released and the getter stuck on the rendezvous all by itself.

I've now switched the order, and everything seems good. I've run the tests 100 times back-to-back successfully, and I've run `testSelectRandom07` in a tight loop while eating breakfast, and it ran 1147727 times without problem before I finished eating 😊.

JonathanStarup:
I think that might be good for now then.

For me previous question (unrelated to this change) I find it surprising that this program does not deadlock
```scala
def main(): Int32 \ IO = region r {
    use Channel.{buffered, send, recv};
    let (tx, rx) = buffered(r, 1);
    send(1, tx);
    spawn {send(1, tx); println("spawn done")} @ r; // this should block
    Thread.sleep(Time/Duration.fromSeconds(5));
    println("sleep done");
    discard recv(rx);
    println("recv 1 done");
    recv(rx)
}
```

I would think that spawn would be stuck on `awaitAvailableSpace(admin)` and main would be stuck in `lock(channelLock)` (since spawn holds that lock)

paulbutcher:
Ah! No, it shouldn't (and indeed doesn't!) deadlock. Let me try to explain why:

(this is a bit weird, explaining your own code to you, but you clearly understood it when you wrote it, because it's nicely done 😜)

1. The initial `send` in the main thread succeeds because the channel is buffered and has space.
2. The `send` in the spawned thread:
    1. First [locks the channel](https://github.com/flix/flix/blob/master/main/src/library/Concurrent/Channel.flix#L158).
    2. It then calls [`awaitAvailableSpace`](https://github.com/flix/flix/blob/master/main/src/library/Concurrent/Channel.flix#L267).
    3. This finds that `size == maxSize`, so calls `awaitCondition(waitingPutters)`
    4. Await atomically _unlocks the channel lock_ (I think this might be the point you've forgotten?) and waits until the condition becomes true
3. In the main thread, the `recv` succeeds because the the spawned thread has released the lock
4. The condition is now true, so the spawned thread is awoken, re-claims the lock (this all happens within `await`) and tries again.

JonathanStarup:
> Ah! No, it shouldn't (and indeed doesn't!) deadlock. Let me try to explain why:
> 
> (this is a bit weird, explaining your own code to you, but you clearly understood it when you wrote it, because it's nicely done 😜)
> 
> 1. The initial `send` in the main thread succeeds because the channel is buffered and has space.
> 2. The `send` in the spawned thread:
>    
>    1. First [locks the channel](https://github.com/flix/flix/blob/master/main/src/library/Concurrent/Channel.flix#L158).
>    2. It then calls [`awaitAvailableSpace`](https://github.com/flix/flix/blob/master/main/src/library/Concurrent/Channel.flix#L267).
>    3. This finds that `size == maxSize`, so calls `awaitCondition(waitingPutters)`
>    4. Await atomically _unlocks the channel lock_ (I think this might be the point you've forgotten?) and waits until the condition becomes true
> 3. In the main thread, the `recv` succeeds because the the spawned thread has released the lock
> 4. The condition is now true, so the spawned thread is awoken, re-claims the lock (this all happens within `await`) and tries again.

Ah yes, thank you, nice explanation! I forgot the detail that waitingPutters is tied to the channel lock

magnus-madsen:
Is this ready? Could someone summarize the change and do you feel comfortable to merge it in?

paulbutcher:
**Summary of the problem:**

This addresses a problem discovered by reading the code only (we've never seen it happen in practice). The Java spec allows threads waiting on conditions to be woken up at any time. For correct behaviour the woken thread needs to check whether the condition it's waiting for actually holds, and if not re-wait on the condition. Our code was doing this in every case other than the rendezvous at the end of `put` for the unbuffered channel case.

The effect of this (if it ever actually happened) is that a call to `Channel.send` on an unbuffered channel could return prematurely (i.e. before some other thread has called `Channel.get`).

It's possible that this has been happening (we don't have any tests which would fail if it did), and it's difficult to write a test to demonstrate the problem (even if the test never saw spurious wakeups, it doesn't mean that they won't happen).

**Summary of the fix:**

Fixing this requires differentiating between threads waiting on the channel because they want to send something, and the single thread waiting for the value that it's just sent to be received. And rather than use a condition to implement this rendezvous, we're using `CyclicBarrier` (which is designed for exactly this use case, and doesn't have the spurious wakeup problem).

Every channel now has a `CyclicBarrier` associated with it with a count of 2, only used if the channel is unbuffered. The `put` function waits on this barrier once it's sent a value, just after releasing the channel lock. The `get` and `select` code paths also wait on the same barrier just before they signal waiting putters that the channel now has space.

paulbutcher:
And yes, I believe that this is ready to be merged.

magnus-madsen:
Merged. Thanks!

Issue #5131, opened 2022-12-17
Deprecate and remove the `Scoped` type class


magnus-madsen:
- [ ] Systematically remove `Scoped`.
- [ ] Systematically remove `'Static` (and with it the `Region` expression).

paulbutcher:
@magnus-madsen Can I confirm that my understanding is correct before pushing ahead with this.

Take, for example, `Array.scanLeft`, which (after @stephentetley's changes in #5143) looks like this:

```
    pub def scanLeft(f: (b, a) -> b \ ef, s: b, arr: Array[a, r]): Array[b, r] \ { ef, Read(r), Write(r) } =
        let len = length(arr) + 1;
        let r = Scoped.regionOf(arr);
        let b = new(r, s, len);
        def loop(i, acc) = {
            if (i >= len)
                ()
            else {
                let s1 = f(acc, arr[i - 1]);
                b[i] = s1;
                loop(i + 1, s1)
            }
        };
        loop(1, s);
        b
```

I think that the right solution is to add an additional region parameter to this function (and most other functions that still use `regionOf`). Do I understand correctly?

magnus-madsen:
> I think that the right solution is to add an additional region parameter to this function (and most other functions that still use `regionOf`). Do I understand correctly?

Yes, I think so.

paulbutcher:
Thanks for confirming.

This leads to a second question about consistency. Take `MutList`. We will need to add a region parameter to [`MutList.insert!`](https://github.com/flix/flix/blob/master/main/src/library/MutList.flix#L575) because it currently uses `regionOf`. But there are other similar functions within `MutList` which don't (e.g. [`MutList.pop!`](https://github.com/flix/flix/blob/master/main/src/library/MutList.flix#L561).

Should we add a region parameter to all these functions for consistency? Or only where it's strictly necessary?

paulbutcher:
Hmm. We may be premature with this. One of the functions that will need changing is [`Iterable.iterator`](https://github.com/flix/flix/blob/master/main/src/library/Iterable.flix#L30-L31), and doing that interacts with the code generated for `foreach`.

Given that we're planning to re-do `foreach` as part of #5132, perhaps we should wait until that's out of the way before tackling this?

paulbutcher:
(or, alternatively, we could do everything else, leaving `Iterable` as-is)?

magnus-madsen:
I think some data structures- including `MutList` will need to keep their region inside them. 

[...]

PR #5156, opened 2022-12-21
chore: Remove (almost) all regionOfs from Array.flix

paulbutcher:
As per discussion in #5131

This PR removes all calls to `regionOf` from Array.flix, apart from the one in `iterator`.

I *think* I've got this right from the POV of "functions that allocate a new data structure should take a region parameter".

I think that this is mostly uncontroversial, with the possible exception of `groupBy` which used to return arrays allocated in two different regions, and I've modified it to use a single region for the new arrays it creates.


magnus-madsen:
Thanks!

PR #5162, opened 2022-12-21
feat: prepare stdlib for namespace migration

mlutze:
related to #4967. Some minor breaking changes include moving RedBlackTree and Random up a level to use companion namespacing


magnus-madsen:
RFR/RFM?

magnus-madsen:
There is also one conflict.

mlutze:
Actually, it might be easier on you if we migrate the tests separately, since that will likely be fairly large as well.

magnus-madsen:
> Actually, it might be easier on you if we migrate the tests separately, since that will likely be fairly large as well.

Ok

magnus-madsen:
Thanks!

PR #5163, opened 2022-12-21
chore: Remove regionOfs from Interpreter.flix

paulbutcher:
As per #5131

I believe that these changes are correct, but I'm not sure why `evalOp` and `evalBoolExp` have two regions (`r1` and `r2`) as it seems that a single region would suffice?


magnus-madsen:
> I believe that these changes are correct, but I'm not sure why evalOp and evalBoolExp have two regions (r1 and r2) as it seems that a single region would suffice?

Not sure either.

magnus-madsen:
Thanks!

PR #5165, opened 2022-12-21
chore: Get rid of the simple to fix instances of Static

paulbutcher:
Towards #5131

This PR addresses the "quick and easy" cases where we can get rid of Static.


magnus-madsen:
Thanks!

PR #5167, opened 2022-12-22
chore: Remove Static from File.flix

paulbutcher:
As per #5131

This goes some way towards removing `Static` from `File.flix`, on the assumption that [this](https://github.com/flix/flix/issues/5131#issuecomment-1362754812) is correct.

I've not been able to go as far as I'd like because it turns out that Java interop doesn't know how to cope with arrays in any region other than `Static`. So I'll address that next.

### DIFF
--- a/main/src/library/File.flix
+++ b/main/src/library/File.flix
@@ -106,11 +106,11 @@ namespace File {
     ///
     /// Returns the BacisFileAttributes of the file `f`.
     ///
-    def getAttributes(f: String): Result[##java.nio.file.attribute.BasicFileAttributes, String] \ IO =
+    def getAttributes(f: String): Result[##java.nio.file.attribute.BasicFileAttributes, String] \ IO = region r {
         try {
             import new java.io.File(String): ##java.io.File \ IO as newFile;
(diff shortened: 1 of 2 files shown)


paulbutcher:
> I've not been able to go as far as I'd like because it turns out that Java interop doesn't know how to cope with arrays in any region other than Static. So I'll address that next.

Hmm, I think I may have got myself confused. Hold fire reviewing this for now.

magnus-madsen:
I am not happy with the introduction of `unsafe_cast`, but I will accept it as a step on the path to removing `Scoped` and then hopefully later we can drop these casts.

paulbutcher:
> I am not happy with the introduction of `unsafe_cast`, but I will accept it as a step on the path to removing `Scoped` and then hopefully later we can drop these casts.

Your answer to [this](https://github.com/flix/flix/issues/5131#issuecomment-1362747390) question suggested that that won't be possible until the two effect systems are unified?

magnus-madsen:
Yeah and/or changes to Iterator/introduction of Stream

PR #5168, opened 2022-12-22
chore: More static removal

paulbutcher:
As per #5131


main/src/library/System.flix
```diff
@@ -30,28 +30,28 @@ namespace System {
...
 
             let br = newBufferedReader(upcast newInputStream(getSystemIn())); // up-cast to superclass
-            let nextLine = ref None;
+            let nextLine = ref None @ r;
             let done = () -> match deref nextLine {
                 case Some(line) =>
                     Object.isNull(line)
                 case None =>
-                    let n = br |> brReadLine;
+                    let n = unsafe_cast brReadLine(br) as _ \ Read(r);
```
magnus-madsen, line 46: Are  these casts needed? I would prefer not to introduce new casts.

main/test/flix/Test.Exp.Concurrency.NewChannel.flix
```diff
@@ -1,93 +1,93 @@
 namespace Test/Exp/Concurrency/NewChannel {
 
     @test
-    def testNewChannel01(): (Sender[Unit, Static], Receiver[Unit, Static]) \ IO = Channel.unbuffered(Static)
+    def testNewChannel01(): Unit \ IO = region _r { discard Channel.unbuffered(Static) }
```
magnus-madsen, line 4: I think for these tests to make sense we should:

- Use the region.
- Put an element of the appropriate type e.g. `Bool`, `String` and so forth.
- Retrieve said element.

main/src/library/System.flix
paulbutcher: This is exactly the issue we discussed [here](https://github.com/flix/flix/issues/5131#issuecomment-1362754812)

We could avoid the cast by changing the `import` line so that `brReadLine` has the effect `Read(r)` instead, but both approaches are "lies" and the more explicit cast seemed better to me (someone should come and fix this up when the effect system allows it).

main/test/flix/Test.Exp.Concurrency.NewChannel.flix
paulbutcher: I was trying to keep the change as small as possible, and follow your earlier guidance that tests should only test the thing they're intended to test (i.e. these tests are supposedly only testing `newChannel`, not `send` and `receive` which are tested elsewhere).

magnus-madsen: True, but now all the tests are the same, right?

magnus-madsen: I think the minimum delta would be to:

- Introduce region.
- Create channel in a let with a type, e.g. `let _: Channel[....]  = ...`.

magnus-madsen: But I think my first outline (which admittedly requires the tests to be rewritten) would be best.

paulbutcher:
So I've rewritten these tests, but they've now pretty much morphed into copies of the tests in `Test.Exp.Concurrent.Buffered.flix` and `Test.Exp.Concurrent.Unbuffered.flix`

Honestly, I think the right approach is to delete `Test.Exp.Concurrency.NewChannel.flix`: I don't think it serves any useful purpose any longer.

magnus-madsen:
> Honestly, I think the right approach is to delete Test.Exp.Concurrency.NewChannel.flix: I don't think it serves any useful purpose any longer.

I will merge the PR, but feel free to have a pass over the channels (and to delete tests, if you deem it appropriate).

PR #5169, opened 2022-12-22
chore: More static removal

paulbutcher:
As per #5131

In some of these cases, I've chosen to delete tests referring to `Static` rather than rewrite them becase they already seem to be covered elsewhere.

I probably could have deleted more of the tests within `Test.Exp.Regions.flix` because they seem to be explicitly testing `Static` but erred on the side of caution.

### DIFF
--- a/main/test/flix/Test.Exp.New.flix
+++ b/main/test/flix/Test.Exp.New.flix
@@ -1,22 +1,6 @@
 namespace Test/Exp/New {
 
     @test
-    def testNewMutList1Region01(): MutList[Int32, Static] \ IO =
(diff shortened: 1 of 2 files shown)


magnus-madsen:
> As per #5131
> 
> In some of these cases, I've chosen to delete tests referring to `Static` rather than rewrite them becase they already seem to be covered elsewhere.
> 
> I probably could have deleted more of the tests within `Test.Exp.Regions.flix` because they seem to be explicitly testing `Static` but erred on the side of caution.

OK

paulbutcher:
Giving up on this PR because it's turned into the merge from hell: easier to redo than merge.

PR #5144, opened 2022-12-18
refactor: change shape of `Validation.Success`

jaschdoc:
- [ ] Add tests

Related to https://github.com/flix/flix/issues/5026

Issue #5026, opened 2022-11-30
Improve Frontend Support for Errors


magnus-madsen:
This is a tracking issue to improve how the frontend copes with errors.

The goal is that compiler should be able to progress through each phase, even if errors are encountered. 

Actionable steps (to be completed in separate PRs):

- [x] Refactor `Validation.Success` to have the shape `case class Success[T, E](t: T, errors: LazyList[E])`.
- [x] Change `Main.scala`, `Shell.scala`, `Lsp.scala`, `TestFlix.scala` (etc) to treat `Success(_, someNonEmptyList)` as failure. In other words, `Success` is only success if there are no errors in the lazy list.
- [x] Introduce `Validation.toSoftFailure(T, E)` which creates a `Success` with an element `T` and an error `E`.
- [x] Modify `LanguageServer` to update root AST even in the presence of errors.
- [x] Introduce a new `case class Expression.Error(m: CompilationMessage)` into all frontend ASTs (`WeededAst` until `LoweredAst`). This should be last expression in every AST.
- [x] Incrementally modify all frontend phases from `Weeder` to just before `Lowering` to return an `Error` expression and use `toSoftFailure` instead of `toFailure`. I suggest to this in many many smaller PRs (each at most a few hundred lines). Begin with just a single error, e.g. `IllegalInt` to ensure that the pipeline works.
- [ ] Rewrite the following phases to use `Validation`:
  - [ ] `Regions`
  - [x] `PatternExhaustiveness`
  - [x] `Redundancy`
  - [ ] `Safety`

JonathanStarup:
Will this allow vscode to use the parsed AST even though the types fails for example?

magnus-madsen:
> Will this allow vscode to use the parsed AST even though the types fails for example?

It will allow most of VSCode code work even if there are some non-parsing related errors.

JonathanStarup:
Having a stupid/unsound robust parser as the final fallback could also be interesting. That would allow some highlighting and possibly better autocomplete on partially broken programs.

magnus-madsen:
> Having a stupid/unsound robust parser as the final fallback could also be interesting. That would allow some highlighting and possibly better autocomplete on partially broken programs.

There is a separate issue for that.

jaschdoc:
Should we maybe introduce a `SoftError` case class? I know this will affect the entire compiler, but might be simpler (and safer) in the long run than having to check `errors.isEmpty` at call sites.

mlutze:
[...]

PR #5174, opened 2022-12-24
feat: add `Validation.SuccessWithFailures`

jaschdoc:
Supersedes https://github.com/flix/flix/pull/5144


jaschdoc:
I've added `ap` and `curry` functions to the validation functions to keep the implementation of `mapN` simple. It may be a bit slower, however, the combinatorial explosion of each `mapN` implementation would be both very time consuming to read and write.

main/src/ca/uwaterloo/flix/util/Validation.scala
```diff
@@ -214,54 +262,38 @@ object Validation {
     */
   def mapN[T1, T2, U, E](t1: Validation[T1, E], t2: Validation[T2, E])
                         (f: (T1, T2) => U): Validation[U, E] =
-    (t1, t2) match {
-      case (Success(v1), Success(v2)) => Success(f(v1, v2))
-      case _ => Failure(t1.errors #::: t2.errors)
-    }
+    ap(mapN(t1)(curry(f)))(t2)
```
magnus-madsen, line 265: I understand you want to do this to make the code pretty, but it incurs a performance cost.

We have typically written Validation to be stupid, but fast.

Here I would *suspect* you would introduce a lot of additional overhead.

You can try with --Xbenchmark-throughput to see if that is the case.

jaschdoc: Alright, I'll try that

jaschdoc: But keep in mind that for `mapN` that takes 9 parameters, if we were to keep it naive without the performance cost then we would have $2^9 + 1= 513$ distinct combinations i.e. pattern matches (assuming we use a wildcard pattern for the failure cases).

magnus-madsen: I think we should benchmark the two new solutions versus the current implementation.

jaschdoc:
# Benchmarks

**It looks like there is no significant performance decrease.**

## This branch

The first result is slightly off. I had other programs running during this benchmark so maybe it could have affected the result.
```

Throughput (best): 12,188 lines/sec (with 8 threads.)

  min:  3,973, max: 12,188, avg:  9,558, median:  9,884

  The highest throughput was in iteration: 8 (out of 10).
  The ratio between the best and worst iteration was: 3.1x.

Finished 10 iterations on 62,768 lines of code in 72 seconds.

                                                                                
Throughput (best): 13,214 lines/sec (with 8 threads.)

  min:  4,352, max: 13,214, avg: 10,713, median: 11,359

  The highest throughput was in iteration: 9 (out of 10).
  The ratio between the best and worst iteration was: 3.0x.

Finished 10 iterations on 62,768 lines of code in 64 seconds.


Throughput (best): 12,787 lines/sec (with 8 threads.)

  min:  4,495, max: 12,787, avg: 10,259, median: 11,112

  The highest throughput was in iteration: 9 (out of 10).
  The ratio between the best and worst iteration was: 2.8x.

Finished 10 iterations on 62,768 lines of code in 66 seconds.
```

## Main

```

Throughput (best): 12,871 lines/sec (with 8 threads.)

  min:  4,731, max: 12,871, avg: 10,400, median: 11,610

  The highest throughput was in iteration: 9 (out of 10).
  The ratio between the best and worst iteration was: 2.7x.

Finished 10 iterations on 62,768 lines of code in 65 seconds.

Throughput (best): 12,222 lines/sec (with 8 threads.)

  min:  4,518, max: 12,222, avg: 10,220, median: 10,767

  The highest throughput was in iteration: 9 (out of 10).
  The ratio between the best and worst iteration was: 2.7x.

Finished 10 iterations on 62,768 lines of code in 66 seconds.

Throughput (best): 12,578 lines/sec (with 8 threads.)

  min:  4,432, max: 12,578, avg: 10,478, median: 11,560

  The highest throughput was in iteration: 9 (out of 10).
  The ratio between the best and worst iteration was: 2.8x.

Finished 10 iterations on 62,768 lines of code in 65 seconds.
```

magnus-madsen:
Ok. We could also just specialise the ones for 1 to 3 cases.

jaschdoc:
Sounds good :)

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -321,6 +321,7 @@ object Typer {
               Scheme.checkLessThanEqual(inferredSc, declaredScheme, classEnv) match {
                 // Case 1: no errors, continue
                 case Validation.Success(_) => // noop
+                case Validation.SuccessWithFailures(_, _) => ???
                 case Validation.Failure(errs) =>
```
magnus-madsen, line 325: I think -- for Validation -- this could be one of the rare cases where we use an OR pattern (if it works).

Can you write:

```scala
Validation.SuccessWithFailures(_, errs) | Validation.Failure(errs) => 
```

Or does that not work?

magnus-madsen: Alternatively, we could always write:

```
case Validation.Success(...) =>
case failure => ... failure.errors ...
```

maybe that is cleaner?

main/src/ca/uwaterloo/flix/util/Validation.scala
```diff
@@ -27,6 +27,7 @@ sealed trait Validation[+T, +E] {
     */
   final def get: T = this match {
     case Validation.Success(value) => value
+    case Validation.SuccessWithFailures(value, _) => value
```
magnus-madsen, line 30: This should also throw.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
magnus-madsen: Throughout we should consider `SuccessWithFailures` as `Failure`.

The only exception is a specific point inside the language server -- which I will identify.

main/src/ca/uwaterloo/flix/language/ast/Scheme.scala
```diff
@@ -99,7 +99,7 @@ object Scheme {
   def lessThanEqual(sc1: Scheme, sc2: Scheme, classEnv: Map[Symbol.ClassSym, Ast.ClassContext])(implicit flix: Flix): Boolean = {
     checkLessThanEqual(sc1, sc2, classEnv) match {
       case Validation.Success(_) => true
-      case Validation.Failure(_) => false
+      case _failure => false
```
jaschdoc, line 102: Now that we use this variable pattern `case _failure` I've adopted the Flix style of using an underscore as prefix to signal that the wildcard match is not used while keeping it easy to read. Is this the right way to do it or should we just use `case _ => ???` instead or is there a better Scala-specific way of doing things?

magnus-madsen: I think this is ok.

main/src/ca/uwaterloo/flix/language/phase/Weeder.scala
```diff
@@ -1933,7 +1933,7 @@ object Weeder {
               // Doing a manual flatMap to keep the function tail-recursive
               translateHexCode(code, mkSL(sp1, sp2)) match {
                 case Validation.Success(char) => visit(rest2, char :: acc)
-                case Validation.Failure(errors) => Validation.Failure(errors)
+                case failure => Validation.Failure(failure.errors)
```
jaschdoc, line 1936: Should this call recursively on `SuccessWithFailures`? That would, however, require rewriting the function slightly to keep it tail recursive.

main/test/ca/uwaterloo/flix/TestUtils.scala
```diff
@@ -72,9 +73,7 @@ trait TestUtils {
     */
   def expectSuccess(result: Validation[CompilationResult, CompilationMessage]): Unit = result match {
     case Validation.Success(_) => ()
-    case Validation.Failure(errors) =>
-      val actuals = errors.map(_.getClass)
```
jaschdoc, line 76: This variable was unused. Should it be used in below in the `fail` call?

magnus-madsen:
RFR?

main/test/ca/uwaterloo/flix/TestUtils.scala
magnus-madsen: Just remove it, if it was unused.

jaschdoc:
> RFR?

Almost, just need tests

magnus-madsen:
OK. Please also build a JAR and verify that it works correctly with VSCode when writing a program with/without errors :)

main/src/ca/uwaterloo/flix/language/phase/Weeder.scala
mlutze: Fine for now. We can revisit once we start to really use `SuccessWithFailures`

jaschdoc:
> OK. Please also build a JAR and verify that it works correctly with VSCode when writing a program with/without errors :)

Roger!

main/src/ca/uwaterloo/flix/util/Validation.scala
```diff
@@ -45,19 +47,27 @@ sealed trait Validation[+T, +E] {
...
+  // Deprecated. Use flatMapN instead.
   final def flatMap[U, A >: E](f: T => Validation[U, A]): Validation[U, A] = this match {
     case Validation.Success(input) => f(input) match {
       case Validation.Success(value) => Validation.Success(value)
+      case Validation.SuccessWithFailures(value, thatErrors) => Validation.SuccessWithFailures(value, errors #::: thatErrors)
       case Validation.Failure(thatErrors) => Validation.Failure(errors #::: thatErrors)
     }
+
+    case Validation.SuccessWithFailures(input, errors) => f(input) match {
+      case Validation.Success(value) => Validation.Success(value)
```
magnus-madsen, line 59: This is wrong. We have to keep the `errors`.

let us see if you tests find it :)

main/src/ca/uwaterloo/flix/util/Validation.scala
```diff
@@ -169,14 +200,74 @@ object Validation {
     }
   }
 
+  def flatten[U, E](t1: Validation[Validation[U, E], E]): Validation[U, E] = t1 match {
```
magnus-madsen, line 203: Missing comments.

main/src/ca/uwaterloo/flix/util/Validation.scala
```diff
@@ -169,14 +200,74 @@ object Validation {
...
+      SuccessWithFailures(t, errors)
+    case SuccessWithFailures(SuccessWithFailures(t, e), errors) =>
+      SuccessWithFailures(t, errors #::: e)
+    case SuccessWithFailures(Failure(e), errors) =>
+      Failure(errors #::: e)
+    case Failure(errors) =>
+      Failure(errors)
+  }
+
+  def ap[T1, U, E](f: Validation[T1 => U, E])(t1: Validation[T1, E]): Validation[U, E] =
```
magnus-madsen, line 220: Missing comments.

main/src/ca/uwaterloo/flix/util/Validation.scala
```diff
@@ -169,14 +200,74 @@ object Validation {
...
+
+      case (Failure(e1), Success(_)) =>
+        Failure(e1)
+      case (Failure(e1), SuccessWithFailures(_, e2)) =>
+        Failure(e1 #::: e2)
+      case (Failure(e1), Failure(e2)) =>
+        Failure(e1 #::: e2)
+    }
+
+  private def curry[T1, T2, T3, T4, T5](f: (T1, T2, T3, T4) => T5): (T1, T2, T3) => T4 => T5 =
```
magnus-madsen, line 244: Missing comments.

magnus-madsen:
@mlutze Can I ask you to do a careful look through `Validation.scala` to see that we are not missing any errors?

main/src/ca/uwaterloo/flix/util/Validation.scala
```diff
@@ -157,6 +185,9 @@ object Validation {
     for (x <- xs) {
       f(x) match {
         case Success(v) => successValues += v
+        case SuccessWithFailures(v, e) =>
+          successValues += v
+          failureStream += e
```
mlutze, line 190: We need a bit more tracking here. If we only have `SuccessWithFailures` then the whole traversal should be `SuccessWithFailures` rather than `Failure`

main/src/ca/uwaterloo/flix/util/Validation.scala
jaschdoc: Are these not deprecated in favor of `flatMapN` and `mapN`?

magnus-madsen: Yes, but SuccessWithFailure should never become Success, right?

You have done that correct everywhere else, as far as I can see :)

main/src/ca/uwaterloo/flix/util/Validation.scala
mlutze: Should we put these in a separate utility object? And maybe name them differently so we get nicer errors :)

main/src/ca/uwaterloo/flix/util/Validation.scala
```diff
@@ -295,6 +409,41 @@ object Validation {
                                 (f: (T1, T2, T3) => Validation[U, E]): Validation[U, E] =
     (t1, t2, t3) match {
       case (Success(v1), Success(v2), Success(v3)) => f(v1, v2, v3)
+      case (Success(v1), Success(v2), SuccessWithFailures(v3, e3)) => f(v1, v2, v3) match {
```
mlutze, line 412: What in blazes

mlutze: IMO unless there's a significant performance cost, we should really avoid doing this by hand.

magnus-madsen: I proposes to special case for n=1..3 to avoid the overhead of curry and apply etc.

main/src/ca/uwaterloo/flix/util/Validation.scala
magnus-madsen: Ah yes!

main/src/ca/uwaterloo/flix/util/Validation.scala
mlutze: Yeah, let's do a clean impl and see what the graph has to say about it.

jaschdoc: Up to you, I can just revert it :)

magnus-madsen: OK, lets revert it, and see the impact.

jaschdoc:
> > OK. Please also build a JAR and verify that it works correctly with VSCode when writing a program with/without errors :)
> 
> Roger!

Seems to work :)

main/src/ca/uwaterloo/flix/util/Validation.scala
```diff
@@ -149,34 +178,174 @@ object Validation {
...
+      Success(values)
     }
   }
 
+  /**
+    * Returns the combined stream of `failureStream` and `successFailureStream`.
+    *
+    * Helper function for `fastTraverse`.
+    */
+  private def combineFailureStreams[E, S, T](failureStream: ArrayBuffer[(LazyList[E], Int)], successFailureStream: ArrayBuffer[(S, LazyList[E], Int)]) = {
```
magnus-madsen, line 223: This seems a bit crazy to me...

magnus-madsen:
Will you "unspecialize" the special cases and run another benchmark throughput?

main/src/ca/uwaterloo/flix/util/Validation.scala
jaschdoc: Agreed, but how should we combine the streams if we lose the order of the errors and values?

jaschdoc:
> Will you "unspecialize" the special cases and run another benchmark throughput?

Sure :)

main/src/ca/uwaterloo/flix/util/Validation.scala
magnus-madsen: My comment was like: "I want a pony". Does not mean I can get it. I was just wondering if there is a better solution without offering any insight on how to do it.

mlutze: Can't we just keep the two failure streams and also a flag that says whether the error is fatal?

magnus-madsen: > Can't we just keep the two failure streams and also a flag that says whether the error is fatal?

Can you sketch?

magnus-madsen:
Two things left:
- The combineFailureStreams thing
- Benchmarks

main/src/ca/uwaterloo/flix/util/Validation.scala
mlutze:
```scala
    var isFatal = false
    // Apply f to each element and collect the results.
    for (x <- xs) {
      f(x) match {
        case Success(v) => successValues += v
        case SuccessWithFailures(v, e) =>
            successValues += v
            failureStream += e
        case Failure(e) => 
            failureStream += e
            isFatal = true
      }
    }


    // Check whether we were successful or not.
    if (isFatal) {
        /Failure(failureStream)
    } else if (!failureStream.isEmpty) {
      Success(successValues.toList)
    } else {
        SuccessWithFailures(successValues, failureStream)
    }
}
```

jaschdoc:
# Benchmarks

## This branch f2baf2a511cd9e207cf3c2915727d7d42336b170

```

Throughput (best): 11,150 lines/sec (with 8 threads.)

  min:  4,273, max: 11,150, avg:  8,872, median:  9,096

  The highest throughput was in iteration: 6 (out of 10).
  The ratio between the best and worst iteration was: 2.6x.

Finished 10 iterations on 62,779 lines of code in 76 seconds.

                                                                                
Throughput (best): 11,026 lines/sec (with 8 threads.)

  min:  3,648, max: 11,026, avg:  9,096, median: 10,080

  The highest throughput was in iteration: 5 (out of 10).
  The ratio between the best and worst iteration was: 3.0x.

Finished 10 iterations on 62,779 lines of code in 76 seconds.


Throughput (best): 12,435 lines/sec (with 8 threads.)

  min:  3,696, max: 12,435, avg:  9,969, median: 10,439

  The highest throughput was in iteration: 8 (out of 10).
  The ratio between the best and worst iteration was: 3.4x.

Finished 10 iterations on 62,779 lines of code in 71 seconds.
```

## Main c9dcf3b7b3fe3c7ebd8b6e5ecce9481077a3fe85

```                                                                                

Throughput (best): 13,137 lines/sec (with 8 threads.)

  min:  4,786, max: 13,137, avg: 10,707, median: 11,672

  The highest throughput was in iteration: 7 (out of 10).
  The ratio between the best and worst iteration was: 2.7x.

Finished 10 iterations on 62,779 lines of code in 63 seconds.

                                                                                
Throughput (best): 12,856 lines/sec (with 8 threads.)

  min:  4,755, max: 12,856, avg: 10,561, median: 11,597

  The highest throughput was in iteration: 8 (out of 10).
  The ratio between the best and worst iteration was: 2.7x.

Finished 10 iterations on 62,779 lines of code in 64 seconds.


Throughput (best): 11,464 lines/sec (with 8 threads.)

  min:  3,902, max: 11,464, avg:  9,268, median: 10,394

  The highest throughput was in iteration: 9 (out of 10).
  The ratio between the best and worst iteration was: 2.9x.

Finished 10 iterations on 62,779 lines of code in 74 seconds.
```

main/src/ca/uwaterloo/flix/util/Validation.scala
```diff
@@ -153,85 +182,215 @@ object Validation {
...
+      case (e, _) => e
+    }.foldLeft(LazyList.empty[E])(_ #::: _)
+  }
+
+  /**
+    * Returns the combined stream of `successValues` and `successFailureStream`.
+    *
+    * Helper function for `fastTraverse`.
+    */
+  private def combineSuccessStreams[E, S, T](successValues: ArrayBuffer[(S, Int)], successFailureStream: ArrayBuffer[(S, LazyList[E], Int)]) = {
```
magnus-madsen, line 232: so these functions are no longer used?

magnus-madsen:
Ill merge; next up, I think there is some dead code and we need to look at performance.

jaschdoc:
Hadn't benchmarked the latest commit. I think maybe we gained some performance in Matt's version of `fastTraverse`.

magnus-madsen:
OK; let us see what happens on arewefast.flix.dev :)

PR #5175, opened 2022-12-24
chore: Towards removing Static from channel code

paulbutcher:
This is very definitely a work in progress (there's a lot more to do).

I'm making this into a draft PR to check that I'm heading in the right direction. I believe that the changes I've made to `Concurrent/Channel.flix` to remove `Static` are correct, but they have knock-on effects for what we generate within Lowering. Not least, we'll have to introduce a region to the [code generated for `Select`](https://github.com/paulbutcher/flix/blob/c15a1a7524a56dd20b48207e6d36fae0f90ef94d/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala#L642) and do something similar for `Par` and `ParYield`.

Is this the right way to go, or is there some simpler approach?

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
@@ -604,10 +604,11 @@ object Lowering {
     // becomes a call to the standard library function:
     //     Concurrent/Channel.newChannel(10)
     //
-    case TypedAst.Expression.NewChannel(_, exp, tpe, pur, eff, loc) =>
-      val e = visitExp(exp)
+    case TypedAst.Expression.NewChannel(exp1, exp2, tpe, pur, eff, loc) =>
+      val e1 = visitExp(exp1)
+      val e2 = visitExp(exp2)
       val t = visitType(tpe)
-      mkNewChannelTuple(e, t, pur, eff, loc)
+      mkNewChannelTuple(e1, e2, t, pur, eff, loc)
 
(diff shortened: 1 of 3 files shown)


magnus-madsen:
( on my phone ) Why is the region needed for select? Is it because we need to allocate mutable data?

paulbutcher:
> ( on my phone ) Why is the region needed for select? Is it because we need to allocate mutable data?

Exactly: `selectFrom` takes an array of channels, which is allocated by the code generated within Safety.

paulbutcher:
We'll have the same problem with `Par` and `ParYield` though, but for a different reason: They both create channels. Right now they do so by creating channels in the `Static` region, but won't be able to do so when `Static` doesn't exist.

magnus-madsen:
Is it only for the array? Could it be an inmutahle list? (And then in the future a vector)

magnus-madsen:
Agree with thr observation about par (similar to the coming stream and foreach yield)

magnus-madsen:
Thanks phone auto complete

paulbutcher:
OK, so it sounds like I should push ahead with this approach, but possibly with additional modifications.

paulbutcher:
> Thanks phone auto complete

Which phone do you have? If it's a Samsung, my apologies.

magnus-madsen:
It's a Samsung 🤣

paulbutcher:
You are hereby granted the right to blame me for all autocorrect screwups.

PR #5176, opened 2022-12-24
chore: Switch from an array to a list in channel code

paulbutcher:
As per discussion in #5175

A PR with just this change before moving on to the remainder


magnus-madsen:
Thanks!

PR #5177, opened 2022-12-25
chore: Towards removing Static from Channel code

paulbutcher:
This is a follow-on from the (now closed) #5175, and definitely still draft.

I think I've hit a roadblock with removing `Static` from the channel code. This compiles and successfully runs the tests within `Test.Exp.Concurrency.Buffered.flix` and `Test.Exp.Concurrency.Unbuffered.flix`. The tests in `Test.Exp.Concurrency.Select.flix` however fail with:

```
#
#   Unable to unify: 'List[MpmcAdmin[b24123]] -> (Bool -> (Int32, List[ReentrantLock]) & Impure)' and 'List[MpmcAdmin] -> (Bool -> (Int32, List[ReentrantLock]) & Impure)'.
```

The problem, I believe, is the [type returned by `mkChannelAdminList`](https://github.com/paulbutcher/flix/blob/bcd47a452306de3de00898f1feb962e5cc5473e9/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala#L1337-L1350), but it's not clear to me what that type should be.

The heart of the problem is, I think, the signature of [`selectFrom`](https://github.com/paulbutcher/flix/blob/bcd47a452306de3de00898f1feb962e5cc5473e9/main/src/library/Concurrent/Channel.flix#L254) which takes an array of `MpmcAdmin`. `MpmcAdmin` now takes a region parameter (as per the recent conversation on Gitter) this represents a list of admin objects which are all in the same region. They might, however, not be in the same region.

[This](https://github.com/paulbutcher/flix/blob/bcd47a452306de3de00898f1feb962e5cc5473e9/main/test/ca/uwaterloo/flix/library/TestChannel.flix#L98-L116) new (failing) test illustrates the problem.

I'm not sure where to go from here? Suggestions gratefully received (@magnus-madsen, @JonathanStarup)?

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
@@ -132,7 +132,7 @@ object Lowering {
     lazy val Comparison: Type = Type.mkEnum(Enums.Comparison, Nil, SourceLocation.Unknown)
     lazy val Boxed: Type = Type.mkEnum(Enums.Boxed, Nil, SourceLocation.Unknown)
 
-    lazy val ChannelMpmcAdmin: Type = Type.mkEnum(Enums.ChannelMpmcAdmin, Nil, SourceLocation.Unknown)
+    lazy val ChannelMpmcAdmin: Type = Type.mkEnum(Enums.ChannelMpmcAdmin, Kind.Wild ->: Kind.Star, SourceLocation.Unknown)
 
     lazy val ConcurrentReentrantLock: Type = Type.mkEnum(Enums.ConcurrentReentrantLock, Nil, SourceLocation.Unknown)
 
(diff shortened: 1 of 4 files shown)


magnus-madsen:
(sent from fone)

I think we need to hear if Jonathan has Insights. I would worry that this is a famous problem reoccurring. We had this issue where we had channels with different elements and we wanted to select on them, but could not have them in the same lost. We eventually solved it by splitting channels onto two parts, one with the queue and one with the admin stuff.

paulbutcher:
I'm not suggesting that we resort to this (yet, at least). But something to consider:

Could we hold on to something *like* `Static` just for use within library code? User-level code won't be polluted by it.

Definitely not the first choice, but it might be good enough?

JonathanStarup:
its not easy to solve this atleast. it might be possible with manually subtyped effects as the list type. Ive not so far found a good solution for this in flix since you cant do the common trick of hiding the type parameter in a closure.

magnus-madsen:
I am ok with keeping static for internal use. We should minimize it's use, but we can keep it around. 

More important to get rid of scoped.

paulbutcher:
> I am ok with keeping static for internal use. We should minimize it's use, but we can keep it around.
> 
> More important to get rid of scoped.

OK, I'll leave this as-is for now.

magnus-madsen:
I think working on iterator or alternatively on stream would be a good next step, depending on your preferred line or attack.

PR #5179, opened 2022-12-25
chore: Further Static removal

paulbutcher:
This is the last "low hanging fruit" when it comes to `Static` removal (as per #5131). I've also taken the chance to delete the `NewChannel` tests as discussed in #5168


magnus-madsen:
Unfortunately I introduced conflicts with some of my changes :(

paulbutcher:
> Unfortunately I introduced conflicts with some of my changes :(

Ouch: that's twice with these changes now 😟. It would be great if we could get this merged to save me doing this for a third time.

magnus-madsen:
Done!

PR #5187, opened 2022-12-26
chore: Add region parameter to Iterable.iterator

paulbutcher:
As per #5131

There are rather more casts in this code than I would prefer (see recent conversation on Gitter), but I've used `upcast` wherever possible.


magnus-madsen:
The changes look surprisingly mild.

I will take a more detailed look tomorrow.

I suspect if we merge `done` and `next` into one function some of the casts will go away. I think that is also a better implementation anyway.

magnus-madsen:
Thanks!

Issue #5191, opened 2022-12-27
Swap type arguments of `Result` and `Validation`


magnus-madsen:
- [ ] Swap type arguments of `Result`
  - [ ] Add `Functor`, `Applicative`, and `Monad` instances for `Result`.
- [ ] Swap type arguments of `Validation`
  - [ ] Add `Functor` and `Applicative` instances for `Validation`.
- [ ] Update documentation (`doc.flix.dev`) and examples (flix.dev)

magnus-madsen:
@stephentetley Would you perhaps be interested in this?

stephentetley:
Yes - I'll take it on. I've plenty of time until the new year.

stephentetley:
Swapping type arguments for both Result and Validation done in PR #5193

magnus-madsen:
Thanks Stephen. I will merge this in a bit. Just waiting for arewefast.flix.dev to update related to another PR.

magnus-madsen:
I merged the first PR.

stephentetley:
> I merged the first PR.

Thanks - I'm making a PR for the FAM instances.

magnus-madsen:
I think we might be able to change: https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/Weeder.scala#L837 to use `Monad.flatMap`. Then you get the `let*` syntax for free without any uses.

stephentetley:
I've made a PR on flix book as well.

https://github.com/flix/book/pull/71

stephentetley:
> I think we might be able to change: https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/Weeder.scala#L837 to use `Monad.flatMap`. Then you get the `let*` syntax for free without any uses.

I'm neutral on this. At the moment I don't think there are restrictions on the type of `flatMap` so we can have `let*` syntax for regular monads and graded / indexed monads as well. Essentially we have an analog of Haskell's `RebindableSyntax`.

However, not having to use `use X.flatMap` would be convenient for the majority of the code currently being written.

I'm happy to go with what other people prefer.

magnus-madsen:
> I'm neutral on this. At the moment I don't think there are restrictions on the type of flatMap so we can have let* syntax for regular monads and graded / indexed monads as well. Essentially we have an analog of Haskell's RebindableSyntax.

Good point.

PR #5193, opened 2022-12-27
refactor: Swap type arguments of Result and Validation (issue #5191)

stephentetley:
This PR changes the order of the type arguments of `Result` and `Validation` so they are compatible with the `Functor` etc. classes. Work to add `Functor`, `Applicative` and `Monad` instances will be done in a follow up PR.

Mostly I've changed as little as possible, the exceptions are in the library suite tests - `TestResult.flix` and `TestValidation.flix`. Here I've changed type ascriptions to use fully specified type names rather than wildcards. I initially tried to just flip the existing type arguments but I  found it very confusing trying to track which tests I'd changed.

In other tests I've tried to keep type wildcards where they were used so that it doesn't reduce the number of tests that validate type wildcards are working.

### DIFF
--- a/examples/type-aliases.flix
+++ b/examples/type-aliases.flix
@@ -6,8 +6,8 @@ type alias M = Map[Int32, String]
 /// A function that returns a map of type M.
 def f(): M = Map#{ 1 => "Hello", 2 => "World"}
 
-// A polymorphic type alias for the type Map[a, Result[Int32, String]].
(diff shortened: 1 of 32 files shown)


magnus-madsen:
Fast service 😀

magnus-madsen:
@mlutze This obviously causes some breakages to your code.

magnus-madsen:
Thanks!

Issue #5080, opened 2022-12-11
`zipWithIndex` and `mapWithIndex` inconsistent


JonathanStarup:
`zip` uses `(a, Int32)` while `map` wants a `(Int32, a) -> ...` function. I think index first is the most practical.

flatMapWithIndex could also be nice since `zipWithIndex >> flatMap` is not too nice with its tuple argument.

#### `(Int32, a)` "Correct"
* `{Array, Chain, DelayList, Iterator, List, Nec, Nel, String}.zipWith` (if used `c |> C.zipWith(C.indices(c))`)
    * `MutList` is missing
* `Iterator.zipWithL` (if used `i |> Iterator.zipWithL(Iterator.indices(i))`)
* `{Array, Chain, DelayList, Iterator, List, Nec, Nel, Option, String}.zip` (if used `c |> C.zip(C.indices(c))`
    * `MutList` and `Result` is missing
* `{Array, Chain, DelayList, List, MutList, Nec, Nel, String}.mapWithIndex`
    * `Iterator` is missing
* `{Array, MutList}.forEachWithIndex`
    * `Chain`, `DelayList`, `Iterator`, `List`, `Nec`, `Nel`, `String`, `Option`, `Result` is missing. `MutDeque` could maybe be argued.
* `{Array, Mutlist}.transformWithIndex!`
    * `MutDeque` could be argued

#### `(a, Int32)` "Wrong"
* `{Chain, DelayList, Iterator, List, Nec, Nel}.zipWithIndex`
    * `Array`, `MutList` is missing

#### Other notes
* Only `Array` and `String` has `indices`. It should be all namespaces with `zipWithIndex`
* It would be really nice to have `Tuple.liftX` like `Result`/`Option`/etc. Then with `indices`, `zip`, and `Tuple.lift` you could easily achieve all the flatMapWithIndex variants you want and have to rely less on std lib.

stephentetley:
Some of these inconsistencies got fixed in the summer:

#4331 _WithIndex functions - functional arg subject last?

magnus-madsen:
Could either of you compile a list of what functions should be changed?

magnus-madsen:
> It would be really nice to have Tuple.liftX like Result/Option/etc. Then with indices, zip, and Tuple.lift you could easily achieve all the flatMapWithIndex variants you want and have to rely less on std lib.

What does this mean?

JonathanStarup:
[...]

PR #5201, opened 2022-12-29
refactor: make zipWithIndex and enumerator consistent with mapWithIndex (issue #5080)

stephentetley:
This PR changes the argument orders of `zipWithIndex` and `enumerator` so they match `mapWithIndex` - see issue #5080

I think `zipWithIndex` with index first is a good improvement over the status quo - a list of (index, value) pairs seems consistent with how we represent other data e.g. Map (key, value) pairs rather than (value, index) pairs.

Similarly a `foreach` comprehension with an `enumerator` seems more consistent with a traditional for-loop if the loop variables are in (index, value) order.

### DIFF
--- a/main/src/library/Chain.flix
+++ b/main/src/library/Chain.flix
@@ -908,7 +908,7 @@ namespace Chain {
         Foldable.joinWith(f, sep, c)
 
     ///
-    /// Returns a chain where each element `e` is mapped to `(e, i)` where `i`
+    /// Returns a chain where each element `e` is mapped to `(i, e)` where `i`
     /// is the index of `e`.
     ///
     pub def zipWithIndex(c: Chain[a]): Chain[(Int32, a)] =
--- a/main/src/library/Chain.flix
+++ b/main/src/library/Chain.flix
@@ -771,7 +771,7 @@ namespace Chain {
         case None    => [] @ r
         case Some(_) =>
             let arr = Array.new(r, length(c));
-            forEach(match (b, i) -> arr[i] = b, zipWithIndex(c));
+            forEach(match (i, b) -> arr[i] = b, zipWithIndex(c));
             arr
         }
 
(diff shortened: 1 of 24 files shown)


main/src/library/Chain.flix
```diff
@@ -911,10 +911,10 @@ namespace Chain {
     /// Returns a chain where each element `e` is mapped to `(e, i)` where `i`
```
JonathanStarup, line 911: The doc comment is outdated here

main/src/library/DelayList.flix
```diff
@@ -1162,11 +1162,11 @@ namespace DelayList {
     ///
```
JonathanStarup, line 1162: Heres also a comment

main/src/library/List.flix
```diff
@@ -930,10 +930,10 @@ namespace List {
     /// Returns a list where each element `e` is mapped to `(e, i)` where `i`
```
JonathanStarup, line 930: Here is also a comment

main/src/library/Nec.flix
```diff
@@ -674,10 +674,10 @@ namespace Nec {
     /// Returns a Nec where each element `e` is mapped to `(e, i)` where `i`
```
JonathanStarup, line 674: ^

main/src/library/Nel.flix
```diff
@@ -526,13 +526,13 @@ namespace Nel {
     /// Returns a new non-empty list where each element `e` is mapped to `(e, i)`
```
JonathanStarup, line 526: ^

JonathanStarup (changes requested):
I found some comments that reference the pair order

stephentetley:
Thanks for the review Jonathan - I'll get those comments sorted today.

main/src/library/DelayList.flix
stephentetley: fixed

main/src/library/List.flix
stephentetley: fixed

main/src/library/Nec.flix
stephentetley: fixed

main/src/library/Nel.flix
stephentetley: fixed

JonathanStarup (approved):

magnus-madsen:
Thanks guys!

PR #5202, opened 2022-12-30
feat: add implicit method `toSoftFailure`

jaschdoc:
This PR also rename SuccessWithFailure to SoftFailure. Thoughts?

Related to https://github.com/flix/flix/issues/5026

PR #5203, opened 2022-12-30
feat: update root AST on soft failure in `LanguageServer`

jaschdoc:
Depends on https://github.com/flix/flix/pull/5202


main/src/ca/uwaterloo/flix/api/lsp/LanguageServer.scala
```diff
@@ -338,15 +338,40 @@ class LanguageServer(port: Int, o: Options) extends WebSocketServer(new InetSock
             ("id" -> requestId) ~ ("status" -> "failure") ~ ("time" -> e) ~ ("result" -> results.map(_.toJSON))
           }
 
-        case failure =>
-          // Case 2: Compilation failed. Send back the error messages.
+        case SoftFailure(root, errors) =>
```
magnus-madsen, line 341: Gasp! Code duplication.. Good Henrik is not around.

What we use an or pattern `Success(...) | SoftFailure(...)` would that work?

jaschdoc: Haha yes! 🙈 

This was just a sketch to see if it was even the correct thing to do. I'll refactor it next year! 😄

jaschdoc: But in all seriousness I'll find a prettier solution :)

magnus-madsen: I think and OR pattern could work.

jaschdoc: I think an OR pattern has ambiguous variables that would be bound (`errors` could refer to both patterns) but I'll come up with some other way :)

magnus-madsen: I thought that was the point. Does Scala not allow it?

jaschdoc: Nope, that's also why we also just use `failure` as the error pattern

magnus-madsen:
> What do you think about this solution? The Or-pattern doesn't work because it binds `root` in both
> 
> ```scala
> case Success(root) | SoftFailure(root, errors) => ???
> ```
> 
> and as mentioned, Scala doesn't allow that type of shadowing / punning.

That's disappointing.

I think what you did is fine.

magnus-madsen:
Excited to see when we are able to survive a single error and still have VSCode support :)

Issue #5160, opened 2022-12-21
Ast-Printer


JonathanStarup:
### AST Printer

I think Flix should allow the programmer to see any intermediate representation of their code to
1. Understand and explore Flix
2. Debug code for compilation errors (for compiler devs)
3. Debug code and tests for runtime errors (for compiler devs)
4. Optimize code
5. Might allow "reformat code" in VSCode

For point 2, this process mustn't require the full compiler to run but only the phases leading up to the desired AST. For point 3, the identifiers in the printout must be consistent with runtime identifiers.

---

### API

I imagine that you can set an AST-print flag, `--Xprint-asts <phase>,<phase>,...,<phase>` where
- `<phase>`: is a name `parsedAST` of a phase

This will be output to `./build/asts/<phase>.flixir`

It must be a flag instead of a command for use case 3. If you want your AST and nothing else, then you can use `check` along with the `print-ast` flag.

### What do we print?

- ~Everything? (Could be quite big)~
- ~Only defs from a specific namespace?~
- Only defs from a specific file?
- ~A single def and all its dependencies?~

This is also a question of where to put the button in Vscode. Depending on the choice here, additional args to the flag might be needed.

---

### Task list

- [x] Pretty Printer library (https://homepages.inf.ed.ac.uk/wadler/papers/prettier/prettier.pdf)
    * #5161
- [x] Pretty Printer common AST & type formatting
   * #5161
- [ ] Pretty print each AST and type format
    - [X] Eraser
    - [X] Finalize
    - [X] LambdaLift
    - [X] MonoType
    - [ ] Type
    - [ ] ...
- [ ] Add compiler flag and customization
    - [X] flag
    - [ ] customization
- [x] #5898
    - [ ] vscode
    - [ ] CLI
- [ ] Merge LSP formatting 
- [ ] Add Vscode action to show ASTs
    * https://github.com/flix/vscode-flix/pull/255
    * https://github.com/flix/flix/pull/5565
    - [ ] filter to file/mod/def

magnus-madsen:
I will ask Magnus to add a "Show TypedAst" button to VSCode.

PR #5205, opened 2022-12-30
feat: add `Xprint-asts` option

JonathanStarup:
Related to #5160.

usage `--Xprint-asts Parser,Namer,Eraser` which means that the AST after the Parser is printed to `build/asts/Parser.txt`; the AST after the Namer phase is printed to `build/asts/Namer.txt`; and that the AST after the Eraser is printed to `build/asts/Eraser.txt`.

For now this will just be the complete AST


mlutze:
What's the purpose for having the numbered phases?

JonathanStarup:
Less knowledge required from the user. its easy to print AST 1 and then explore from there.

magnus-madsen:
May I also suggest we redesign this to work with `LanguageServer` because then we can get it on right click in VSCode?

In fact, I would suggest that we implement that feature simultaneously with this feature.

magnus-madsen:
I would a `PrintAstProvider` / `ShowAstProvider` to the language server and integrate it with some of the above options.

JonathanStarup:
> What about `--Xprint-ast Typer` => gives us the string `Typer` (call it `phase) then (somewhere) we can do `if (phase.getClass().getSimpleName == options.xprintphase)`

So there is no error if you misspell right? And you don't like the numbering? 

> In fact, I would suggest that we implement that feature simultaneously with this feature.

I wanted to flag so I can actually use it in the other PR without making that PR even larger. It can always be changed again. My idea was to add this flag and then slowly make it do something instead of doing it all at once

magnus-madsen:
I don't like the numbering. First, the numbers will change over time (!) and the code is ugly.

Other than that I support the idea.

magnus-madsen:
I also don't like all the stdout fiddling. May I just propose this:

```
--Xprint-asts Namer,Typer,Simplifier
```

that's all. Dumps to `build` which is our build directory (e.g. `Typer.txt`).

Good enough?

JonathanStarup:
> First, the numbers will change over time (!)

This is the point though. This means that you can print asts 1, 2, 3 for example without knowing about phase names, changes or whatever happens inside.

> and the code is ugly

Yeah. Requires some heavy abstraction/macro to derive it from the actual phases

---

I suppose the numbering thing will be apparent in the vscode UI at some point so maybe it doesn't matter

magnus-madsen:
Can we make it plural `printasts`?

magnus-madsen:
> Can we make it plural `printasts`?

Or is it actually more natural in singular?

JonathanStarup:
@magnus-madsen does this work? now phases can slowly add the conditional output to `build/phase.flix.ir` and the vscode can be built on the providers added (also lacking parsing)

magnus-madsen:
> @magnus-madsen does this work? now phases can slowly add the conditional output to `build/phase.flix.ir` and the vscode can be built on the providers added (also lacking parsing)
> 
> The vscode API is only a singular AST since there wont be naming consistency anyway.

Yes, overall. I will add some more comments.

 But I still think we would also like a command line flag.

JonathanStarup:
> > @magnus-madsen does this work? now phases can slowly add the conditional output to `build/phase.flix.ir` and the vscode can be built on the providers added (also lacking parsing)
> > The vscode API is only a singular AST since there wont be naming consistency anyway.
> 
> Yes, overall. I will add some more comments.
> 
> But I still think we would also like a command line flag.

Its still there `--Xprint-asts Namer,Typer,Eraser` works now (by doing nothing)

main/src/ca/uwaterloo/flix/api/lsp/LanguageServer.scala
```diff
@@ -302,6 +302,15 @@ class LanguageServer(port: Int, o: Options) extends WebSocketServer(new InetSock
       else
         ("id" -> id) ~ ("status" -> "success") ~ ("result" -> Nil)
 
+    case Request.PrintAst(id, phase) =>
```
magnus-madsen, line 305: PrintAst -> ShowAst

main/src/ca/uwaterloo/flix/api/lsp/LanguageServer.scala
```diff
@@ -302,6 +302,15 @@ class LanguageServer(port: Int, o: Options) extends WebSocketServer(new InetSock
...
+      if (current)
+        ("id" -> id) ~ ("status" -> "success") ~ ("result" -> PrintAstProvider.printAst(phase)(index, root, flix))
+      else
+        ("id" -> id) ~ ("status" -> "success") ~ ("result" -> Nil)
+
+    case Request.Phases(id) =>
```
magnus-madsen, line 311: Phases -> ListPhases

main/src/ca/uwaterloo/flix/api/lsp/Request.scala
```diff
@@ -139,6 +139,16 @@ object Request {
     */
   case class InlayHint(requestId: String, uri: String, range: Range) extends Request
 
+  /**
+    * A request to print the AST following a specific phase.
```
magnus-madsen, line 143: print -> show

main/src/ca/uwaterloo/flix/api/lsp/Request.scala
```diff
@@ -139,6 +139,16 @@ object Request {
...
+    */
+  case class PrintAst(requestId: String, phase: String) extends Request
```
magnus-madsen, line 145: -> ShowAst

main/src/ca/uwaterloo/flix/api/lsp/Request.scala
```diff
@@ -139,6 +139,16 @@ object Request {
...
+
+  /**
+    * A request to get the ordered list of phases.
+    */
+  case class Phases(requestId: String) extends Request
```
magnus-madsen, line 150: -> ListPhases

main/src/ca/uwaterloo/flix/api/lsp/provider/PhasesProvider.scala
```diff
@@ -0,0 +1,47 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ca.uwaterloo.flix.api.lsp.provider
+
+object PhasesProvider {
+  val phases: List[String] = List(
+    "Parser",
```
magnus-madsen, line 20: A smart trick is to do the following:

```
classOf[Typer].getSimpleName
```

since this is robust to refactorings and deletions.

magnus-madsen: It also works for search for uses.

main/src/ca/uwaterloo/flix/api/lsp/provider/PrintAstProvider.scala
```diff
@@ -0,0 +1,45 @@
...
+
+import ca.uwaterloo.flix.api.Flix
+import ca.uwaterloo.flix.api.lsp.Index
+import ca.uwaterloo.flix.language.ast.TypedAst.Root
+import org.json4s.JsonAST.JObject
+import org.json4s.JsonDSL._
+
+object PrintAstProvider {
+
+  private val IrFileExtension = "flix.ir"
```
magnus-madsen, line 26: May I propose `.flixir`? That way its _one_ extension.

main/src/ca/uwaterloo/flix/api/lsp/provider/PrintAstProvider.scala
```diff
@@ -0,0 +1,45 @@
...
+
+  /**
+    * Returns a JSON object with
+    *
+    * - `title` (a string like `Namer.flix.ir`)
+    *
+    * - `content` (a string with the ir representation).
```
magnus-madsen, line 33: content -> text?

main/src/ca/uwaterloo/flix/api/lsp/provider/PhasesProvider.scala
JonathanStarup: It does seem to work for `object` and javas `Typer.getClass.getSimpleName` appends a `$` at the end.

Additionally, it seems that Parser is not accessible for some reason?

magnus-madsen: Are these Scala objects? Maybe you need `classOf[Typer.type` or so?

magnus-madsen: I think objects just end with `$` in their name. If so, you could just strip that.

JonathanStarup:
RTM @magnus-madsen

PR #5206, opened 2022-12-31
ci: sleep in community build

mlutze:
To try to help with rate limiting


magnus-madsen:
Later we can remove it for a big performance boost :D

PR #5208, opened 2022-12-31
test: refactor compiler tests for namespace migration

mlutze:
last big refactor before migrating (I think) #4967


mlutze:
Lol rate limiting strategy still failing

PR #5209, opened 2022-12-31
feat: introduce default uses

mlutze:
related to #4967


magnus-madsen:
Is this scripted? :D

mlutze:
> Is this scripted? :D

No, but it's on my daily to-do list so it doesn't fall behind and turn into a huge merge conflict (and also so I can pretend I am programming again)

magnus-madsen:
> 

Soon won't have to pretend :)

main/src/library/Iterator.flix
```diff
@@ -42,8 +42,8 @@ namespace Iterator {
     ///
     pub def iterate(rc: Region[r], f: Unit -> Option[a] \ ef): Iterator[a, ef, r] =
         let f1 = () -> match f() {
-            case Some(a) => Ans(a)
-            case None    => Done
+            case Some(a) => Step.Ans(a)
```
magnus-madsen, line 45: Can we do a use instead? :)

mlutze:
@stephentetley @jaschdoc @paulbutcher
This is a major breaking change. From here on, enum cases will have to either be qualified (`Enum.Case`) or used (`use Enum.Case`), even in the scope in which the enum is declared.

Note that cases do _not_ have to be qualified if they are in the companion namespace of the enum. For example:

```
enum E {
    case C
}

mod E { // companion namespace
    def mkE(): E = C  // no qualification needed
}

mod F { // non-companion namespace
    def mkE(): E = E.C  // qualification needed
}
```

Some cases are automatically imported: `None`, `Some`, `Cons`, `Nil`, `Ok`, `Err`

Namespaces are still in an intermediate state as we more thoroughly test and re-introduce certain redundancy errors.

PR #5219, opened 2023-01-02
feat: add `Expression.Error` to front-end ASTs

jaschdoc:


jaschdoc:
> Introduce a new case class Expression.Error(m: CompilationMessage) into all frontend ASTs (WeededAst until LoweredAst). This should be last expression in every AST.

Does this include `LoweredAst`?

magnus-madsen:
> > Introduce a new case class Expression.Error(m: CompilationMessage) into all frontend ASTs (WeededAst until LoweredAst). This should be last expression in every AST.
> 
> Does this include `LoweredAst`?

No, not lowered ast. The frontend stops just before that.

main/src/ca/uwaterloo/flix/language/ast/TypedAst.scala
```diff
@@ -292,6 +293,16 @@ object TypedAst {
 
     case class FixpointProject(pred: Name.Pred, exp: TypedAst.Expression, tpe: Type, pur: Type, eff: Type, loc: SourceLocation) extends TypedAst.Expression
 
+    case class Error(m: CompilationMessage) extends TypedAst.Expression {
```
magnus-madsen, line 296: I think this is OK for now but may change in the future.

jaschdoc: I had originally defined it as
```scala
case class Error(m: CompilationMessage, tpe: Type, pur: Type, eff: Type, loc: SourceLocation) extends TypedAst.Expression
```
but I wasn't sure what types to pass to the constructor when mapping from `KindedAst` to `TypedAst` i.e. at lines 2478-2479 in the Typer.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -778,6 +778,10 @@ object Lowering {
       val defExp = LoweredAst.Expression.Def(sym, defTpe, loc)
       val argExps = mkPredSym(pred) :: visitExp(exp) :: Nil
       LoweredAst.Expression.Apply(defExp, argExps, tpe, pur, eff, loc)
+
+    case TypedAst.Expression.Error(m) =>
+      throw InternalCompilerException(s"Unexpected error expression near", m.loc)
```
magnus-madsen, line 783: Can we replace this by a Hole expression?

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1912,6 +1912,13 @@ object Typer {
           resultEff = Type.mkUnion(eff1, eff2, loc)
         } yield (constrs1 ++ constrs2, resultTyp, resultPur, resultEff)
 
+      case KindedAst.Expression.Error(m) =>
```
magnus-madsen, line 1915: I think it would be better if this had a type variable, purity variable, and effect variable and you unified with those.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -2467,6 +2474,10 @@ object Typer {
         val mergeExp = TypedAst.Expression.FixpointMerge(e1, e2, stf, e1.tpe, pur, eff, loc)
         val solveExp = TypedAst.Expression.FixpointSolve(mergeExp, stf, e1.tpe, pur, eff, loc)
         TypedAst.Expression.FixpointProject(pred, solveExp, tpe, pur, eff, loc)
+
+      case KindedAst.Expression.Error(m) =>
+        TypedAst.Expression.Error(m)
```
magnus-madsen, line 2479: then here you would pull out the type, purity and effect.

main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
```diff
@@ -812,6 +812,10 @@ object Kinder {
       mapN(exp1Val, exp2Val) {
         case (exp1, exp2) => KindedAst.Expression.FixpointProject(pred, exp1, exp2, Type.freshVar(Kind.Star, loc.asSynthetic), loc)
       }
+
+    case ResolvedAst.Expression.Error(m) =>
+      KindedAst.Expression.Error(m).toSuccess // TODO: Refactor to SoftFailure
```
magnus-madsen, line 817: I think we need a version of toSoftFailure that works with the empty list (since the error is already in the Validation).

magnus-madsen: Can we add that and use it everywhere?

jaschdoc: Yes but maybe that should be in a different PR because right now each phase has to be refactored to use `CompilationMessage` as the error type of the `Validation` instead of `KindError` or `WeederError`

magnus-madsen: Why?

magnus-madsen: If the List is empty it should be OK?

jaschdoc: Oh that's true of course but I figured we wanted `m` to be in the errors list

magnus-madsen: No, because its already there (!)

But indeed its a bit strange.

main/src/ca/uwaterloo/flix/language/ast/TypedAst.scala
magnus-madsen: Type.freshTypeVar and then of the right kind.

magnus-madsen: And replace Type,  Type, Type by Type.Var, Type.Var, Type.Var

jaschdoc:
Should we also add a `loc` field to `Error`? That way we can always just use `m.loc` but also have more fine grained control over where its `loc` is. Plus it's more consistent with all other AST nodes (of course the `loc` function is still implemented for `Error` currently as it's part of the AST interface)

jaschdoc:
Still unsure about the typer phase

magnus-madsen:
> Still unsure about the typer phase

I think its OK now.

magnus-madsen:
> Should we also add a `loc` field to `Error`? That way we can always just use `m.loc` but also have more fine grained control over where its `loc` is. Plus it's more consistent with all other AST nodes (of course the `loc` function is still implemented for `Error` currently as it's part of the AST interface)

Yes, lets do that instead. But we can do it in the next PR.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -767,6 +767,9 @@ object Lowering {
       val defExp = LoweredAst.Expression.Def(sym, defTpe, loc)
       val argExps = mkPredSym(pred) :: visitExp(exp) :: Nil
       LoweredAst.Expression.Apply(defExp, argExps, tpe, pur, eff, loc)
+
+    case TypedAst.Expression.Error(m, _, _, _) => ???
```
magnus-madsen, line 771: Lets make this into a HoleError or alternatively an InternalCompilerException.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1871,6 +1871,16 @@ object Typer {
...
+      case KindedAst.Expression.Error(m, tvar, pvar, evar) =>
+        val constrs = Nil
+        val expectedType = Type.freshVar(Kind.Star, m.loc)
+        val expectedPur = Type.freshVar(Kind.Bool, m.loc)
+        val expectedEff = Type.freshVar(Kind.Effect, m.loc)
+        for {
+          resultTyp <- unifyTypeM(tvar, expectedType, m.loc)
+          resultPur <- unifyTypeM(pvar, expectedPur, m.loc)
+          resultEff <- unifyTypeM(evar, expectedEff, m.loc)
+        } yield (constrs, resultTyp, resultPur, resultEff)
```
magnus-madsen, line 1883: Actually, you dont need `expectedType`, and friends. You can simply return `(Nil, tvar, pvar, evar)`.

magnus-madsen:
Two changes then we are good to go.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
jaschdoc: I'm not seeing any `HoleError`s for LoweredAst

magnus-madsen: Ah- its just called `Hole`. But actually I prefer the use of the exception.

magnus-madsen:
Some type errors...

jaschdoc:
Oops, I combined the statements "Don't use InferMonad.point" and "just return (Nil, tvar, pvar, evar)" but of course that doesn't make sense :D

magnus-madsen:
Next, I would like to see just a single error (e.g. from Weeder) turned into a soft failure :)

jaschdoc:
Roger

Issue #5190, opened 2022-12-27
Add `Collectable` type class


magnus-madsen:
- [ ] Add `Collectable` type class:

```
class Collectable[m: Type -> Type] {
    pub def collect(iter: Iterator[a, r]): m[a] \ r
}

instance Collectable[List] {
    pub def collect(iter: Iterator[a, r]): List[a] \ r = Iterator.toList(iter)
}
```

- [ ] Add instances for all immutable collections.
- [ ] Add instance for `Result`. (Add test cases.) -- if possible?
- [ ] Add instance for `Validation`. (Add test cases.) -- if possible?
- [ ] Add instance for `Option`. (Add test cases.) -- if possible?

Note: Collectable only supports collecting into immutable collections. We can lift that restriction with associated types later.

magnus-madsen:
@stephentetley Did you want to take a stab at this?

stephentetley:
I was going to, but I got stuck with trying to refactor `Iterator`. I could try with the current `Iterator`.

magnus-madsen:
Roger. Let me know if you get stuck again, and I will take a look.

Happy new years! 🎆

stephentetley:
Happy New Year to you too, Magnus!

I've taken a proper look at `Collectable` now and I've become doubtful that it will work out as a good abstraction. Ignoring mutable collections for the time being:

Firstly it works fine for `List` and `Chain`.

`Set` - instance needs `Order` constraint on the element type so can't match the class signature for `collect`.

`Nec` and `Nel` - the answer type is wrapped in an `Option` so doesn't match the class signature for `collect`.

`Map` - input stream needs to be a pair and first element needs an `Order` constraint so doesn't match the generic `collect`.

`Result` and `Option` - no single obvious idea of "collecting". `Option` can be written with the current definition of `collect` where it takes the "head" of the iterator (or `None`) but this seems a very particular idea of collecting. Result can't even do this because we need to produce some default `e` (error type) in the case of an empty iterator. 

Even in a future with associated types I think there may still be too much variance in the signature of `collect` to work well.

magnus-madsen:
Good points. I will have to think more about it.

> Set - instance needs Order constraint on the element type so can't match the class signature for collect.

Hmm. I wonder how they make it work in Rust, since surely they need either `Order` or `Hash`?

magnus-madsen:
Does any one read Rust? 

[...]

PR #5222, opened 2023-01-02
feat: add Collectable type class (issue #5190)

stephentetley:
This PR adds a `Collectable` type class with a single function `collect` and instances for `List`, `Chain` and `Set`.

`toChain` has been added to `Iterator` and tests have been added to `TestIterator`  for `toChain` and `toList` (previously missing).

`Collectable.collect` tests are in the respective collection modules.


magnus-madsen:
Thanks Stephen!

Issue #5215, opened 2023-01-02
Remove support for implicit regions


magnus-madsen:

paulbutcher:
Can I check what the intent of this is? Is it:

* To avoid things like `ref` picking up the "nearest" region automagically?
* To avoid `region r` magically introducing both a term-level and type-level variable?

(FWIW, I think that we should fix both of these, but I'm not sure which this issues is intended to address?)

magnus-madsen:
This one:

To avoid things like ref picking up the "nearest" region automagically?

paulbutcher:
Can we do the other one too? 😊

magnus-madsen:
> Can we do the other one too? 😊

I am not sure exactly what you mean? That you have to name the type level variable or that there should only be term/type level?

paulbutcher:
That there should be a clear indication in the syntax that *two* variables are being introduced. One way to do this might be:

```
region rh @ r { ... }
```

Although there are clearly plenty of other options.

stephentetley:
Hope this isn't treading on Matthew's toes - I've looked at the tests suites and changed them to use an explicit region where they matched this grep 

> find . -name "*.flix" | xargs grep -n " ref " | grep -v "@"

(Can't guarantee if finds everything.)

I'll make a PR once I've got one problem to compile.

mlutze:
Thanks Stephen, no toe-treading done :)

mlutze:
Remaining uses after Stephen's PR: (#5253)

```
   1725 main/test/ca/uwaterloo/flix/library/TestArray.flix
     20 main/test/ca/uwaterloo/flix/library/TestDebug.flix
      3 main/test/ca/uwaterloo/flix/library/TestFile.flix
      3 main/test/ca/uwaterloo/flix/library/TestFoldable.flix
     94 main/test/ca/uwaterloo/flix/library/TestMutList.flix
     42 main/test/ca/uwaterloo/flix/library/TestMutMap.flix
      1 main/test/ca/uwaterloo/flix/library/TestObject.flix
      3 main/test/ca/uwaterloo/flix/library/TestReducible.flix
      1 main/test/ca/uwaterloo/flix/library/TestToString.flix
      1 main/test/flix/Test.Eff.Advanced.flix
      2 main/test/flix/Test.Eff.Polymorphism.flix
     14 main/test/flix/Test.Exp.ArrayLength.flix
    171 main/test/flix/Test.Exp.ArrayLit.flix
[...]

PR #5223, opened 2023-01-03
refactor: remove use of implicit regions from standard library

mlutze:
related to #5215


magnus-madsen:
What about the compiler or will that come separately?

mlutze:
I'll do it separately if that's ok

PR #5224, opened 2023-01-03
feat: allow mod keyword

mlutze:
related to #4967


magnus-madsen:
What about dot separated names?

mlutze:
Should already be possible.

!!!jar

Issue #5226, opened 2023-01-03
Disable auto-completions on `???`


magnus-madsen:
When typing `???` VSCode should _not_ offer any completions.

For example, if I am writing:

```
let x = 123 + ???|
```

I don't want any suggestions. I believe the fix can be made around here:

https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala#L200

magnus-madsen:
![image](https://user-images.githubusercontent.com/2287953/210408093-fd5700c4-61dd-4e67-8877-19e4ccf27917.png)
Is what I don't want

JonathanStarup:
Thats why I believe the ""fill this hole" should be a quick action instead of autocomplete

PR #5228, opened 2023-01-03
feat: disable autocomplete for `???`  #5226

yisraelU:
related to #5226  
first-time PR , just learning the ropes here

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
@@ -184,6 +184,7 @@ object CompletionProvider {
     val letRegex = raw"\s*let\s+[^\s]*".r
     val letStarRegex = raw"\s*let[\*]\s+[^\s]*".r
     val namespaceRegex = raw"\s*namespace\s+.*".r
+    val tripleQuestionMarkRegex = raw"\s+\?\?\?".r
     val underscoreRegex = raw"(?:(?:.*\s+)|)_[^s]*".r
 
     // if any of the following matches we know the next must be an expression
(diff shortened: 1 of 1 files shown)


magnus-madsen:
Looks reasonable :-) Can I ask you to:

- Run `gradlew clean jar`. Take the jar from `build/libs`, place it into an empty directory, open that directory in VSCode (with the Flix plugin installed). It should start using your custom jar. Manually verify that `???` no longer triggers completion.
- Add your name to `AUTHORS.md` to confirm you agree to release your contributions under Apache 2.0

magnus-madsen:
> @magnus-madsen Thanks , it works . 

Thanks!

> however I am wondering, is there a valid use case for having the autocomplete run with only 1 question mark ?

Yes. We use that for auto-completion like on the dot in Java. For example, you can try:

```
def foo(): Int32 = 
let x: String = "Hello";
let y: Int32 = x?
y
```

and see what it suggests :)

magnus-madsen:
Merged. Thanks!

yisraelU:
> > @magnus-madsen Thanks , it works .
> 
> Thanks!
> 
> > however I am wondering, is there a valid use case for having the autocomplete run with only 1 question mark ?
> 
> Yes. We use that for auto-completion like on the dot in Java. For example, you can try:
> 
> ```
> def foo(): Int32 = 
> let x: String = "Hello";
> let y: Int32 = x?
> y
> ```
> 
> and see what it suggests :)

right but that has no preceding whitespace, I am only matching cases with at minimum 1 preceding whitespace char

magnus-madsen:
What do you mean? If one were to type ` ?`

yisraelU:
yes

magnus-madsen:
> yes

I sees. So rather we should give suggestions if we see `name?`, but not otherwise?

yisraelU:
Yes , from my understanding(albeit very minimal) I cannot see a utility in an autocomplete on a whitespace isolated question mark

magnus-madsen:
Roger. Feel free to take a stab at that :)

PR #5232, opened 2023-01-04
fix: open datalog arguments

### LINKED ISSUE #5068: Add `OpenSchema` expression
Given the program:

```
def isCyclic(db: #{Edge(Int32, Int32)}): Bool = 
    let pr = #{
        Path(x, y) :- Edge(x, y).
    };
    not (query db, pr select () from Cycle() |> List.isEmpty)
```

We get the type error:

```
❌ -- Type Error --------------------------------------------------

>> Unable to unify the types: '#(  )' and '#( Path(Int32, Int32) | s0 )'.

6 |     not (query db, pr select () from Cycle() |> List.isEmpty)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
             mismatched types.

Type One: #{ Edge(Int32, Int32) }
Type Two: #{ Edge(t0, t0), Path(t0, t0) | s1 }
```

because the argument `db` is a ***closed row***.

The type of `db` is `#{Edge = ... | {}}`. Notice that this type (which is like a list) ends in `{}`. The idea is for the compiler to change this to `#{Edge = ... | r}` for a fresh type variable `r`. This needs to happen around here:

https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/Typer.scala#L305

where `getSubstFromParams` should open the rows.

Note that the transformation must ***only*** take place if there is a closed row.

JonathanStarup:
Fixes #5068

### DIFF
--- a/main/test/flix/Test.Exp.Fixpoint.Compose.flix
+++ b/main/test/flix/Test.Exp.Fixpoint.Compose.flix
@@ -88,12 +88,42 @@ namespace Test/Exp/Fixpoint/Compose {
 
     @test
     def testFixpointCompose13(): Bool = {
-        let ans = query openRow(#{Coolio("quack!").}) select s from Duck(s);
+        let ans = query coolioQuack(#{Coolio("quack!").}) select s from Duck(s);
         ans == List#{"quack!"}
     }
 
-    def openRow(p: #{Coolio(String)}): #{Coolio(String), Duck(String) | r} = {
+    def coolioQuack(p: #{Coolio(String)}): #{Coolio(String), Duck(String) | r} = {
         p <+> #{Duck(s) :- Coolio(s).}
+    }
+
+    @test
+    def testFixpointCompose14(): Bool = {
+        let ans = query duckOutOfNothing(#{}) select s from Duck(s);
+        ans == List#{"quack!"}
+    }
+
+    def duckOutOfNothing(p: #{}): #{Duck(String) | r} = {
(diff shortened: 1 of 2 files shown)


JonathanStarup:
RFR @magnus-madsen

main/test/flix/Test.Exp.Fixpoint.Compose.flix
```diff
@@ -86,4 +86,14 @@ namespace Test/Exp/Fixpoint/Compose {
         let rb = solve c3 <+> c4;
         (ra, rb)
 
+    @test
+    def testFixpointCompose13(): Bool = {
+        let ans = query openRow(#{Coolio("quack!").}) select s from Duck(s);
+        ans == List#{"quack!"}
+    }
+
+    def openRow(p: #{Coolio(String)}): #{Coolio(String), Duck(String) | r} = {
```
magnus-madsen, line 95: I would like some more examples: Say: 

- 1x with 1 predicate
- 2x with 2 predicates (just trying different orders)
- 4x with 3 predicates (just trying different orders)

magnus-madsen: Something that checks the order is not important.

JonathanStarup: I'm not sure I follow, you want this test copied with 2 and 3 predicates?

Issue #5189, opened 2022-12-27
Refactor `Iterable` class


magnus-madsen:
As a step towards https://github.com/flix/flix/issues/5132, I would like to change `Iterable` from:

```
pub class Iterable[t: Type -> Region -> Type] {
    pub def iterator(r1: Region[r1], t: t[a, r2]): Iterator[a, r1 and r2] \ { Write(r1), Read(r2) }
    // ...
}
```

to

```
pub class Iterable[t: Type -> Type] {
    pub def iterator(r1: Region[r], t: t[a]): Iterator[a, r] \ r
    // ...
}
```

This means that we can implement `Iterable` for all immutable types (e.g. `List`), but we have to drop the instances for mutable collections. **This is not one step forward and one step back, because we are changing `Iterable` to a signature that is _compatible_ with associated types. Thus this _is future proofing_**.

paulbutcher:
That makes sense, but can I ask how far away associated types are? I.e. how long will this mean that we're without a good solution for mutable types?

magnus-madsen:
> 

I would say 2-6 months. 

In the interim the solution for mutable types is not bad, it is going to be:

```
region rh {
  let result = foreach(x <- immutableList, y <- MutList.iterator(rh, mutableList)) yield x + y
   // ...
}
```

I think that is very reasonable.

magnus-madsen:
But also, currently the situation is that it does not work for immutable collections!

stephentetley:
Has anyone started working on this? 

I can take a look at it and `Collectable` #5190 if no-one has started on them.

magnus-madsen:
> Has anyone started working on this?
> 
> I can take a look at it and `Collectable` #5190 if no-one has started on them.

I don't think anyone has started looking at the code, so you are welcome to take a first stab at it.

magnus-madsen:
Then Paul and Jakob and I can pitch in later with some of the other pieces of the puzzle.

stephentetley:
Thanks - I'll make a start.

magnus-madsen:
By the way, it would be interesting if `Collectable` can be implemented for `Result[List[Int32], ...]`, but I am not sure if that is possible in current Flix.

magnus-madsen:
> By the way, it would be interesting if `Collectable` can be implemented for `Result[List[Int32], ...]`, but I am not sure if that is possible in current Flix.

I guess this is not possible, sadly. I wonder if Haskell can do it.

mlutze:
[...]

PR #5234, opened 2023-01-04
refactor: refactor Iterable type class (issue #5189)

stephentetley:
This PR changes the `Iterable` type class to be kind `Type -> Type` and removes the `emumerator` method. It removes `Iterable` instances from mutable collections and adds `Iterable` instances for `List`, `Set` and `Chain`. Where foreach comprehensions could be swapped to use Lists (e.g. `Fixpoint/Ast/Datalog.flix`) I have done.

I added quite a few `forEachWithIndex` functions after all, they meant less changes were needed to `toString` functions the previously used `enumerator`.

`Test.Exp.ForEach.flix` changed tests to use lists on the foreach fragment "arrow right-hand side".

Examples - I've deleted:

examples/copying-characters-into-array-with-foreach.flix
examples/imperative-style-foreach-loops.flix

They no longer matched their titles and removing the enumerator comprehension left them not doing much at all.

`examples/iterating-over-arrays-with-foreach.flix` - I've swapped it to use lists in the comprehensions and renamed it to `examples/iterating-over-lists-with-foreach.flix`


main/test/ca/uwaterloo/flix/library/TestChain.flix
```diff
@@ -1410,6 +1410,22 @@ namespace TestChain {
         Chain.forEach(x -> r := x, List.toChain(1 :: 2 :: 3 :: Nil));
         3 == deref r
 
+    /////////////////////////////////////////////////////////////////////////////
+    // forEachWithIndex                                                        //
+    /////////////////////////////////////////////////////////////////////////////
+
+    @test
+    def forEachWithIndex01(): Bool \ IO =
+        let r = ref 21;
```
magnus-madsen, line 1419: This style of using a region implicit will soon disappear. So these tests should introduce an explicit region.

examples/copying-characters-into-array-with-foreach.flix
```diff
@@ -1,25 +0,0 @@
-// Easily iterate over arrays with foreach-loops.
```
magnus-madsen, line 1: I would like to keep these examples; if they cannot easily be reworked, perhaps we should just comment them for now?

magnus-madsen: I am just afraid of losing them to the sands of time :)

stephentetley: I could rewrite them with `forEachWithIndex` and put a comment saying to update them when we have Streams / mutable enumerators. Would that work?

magnus-madsen: That would be perfect :)

stephentetley: I've re-instated them. This included adding `forEachWithIndex` to Iterator which is a workaround because `enumerator` is broken.

main/test/ca/uwaterloo/flix/library/TestChain.flix
stephentetley: I've done all the `forEach` and `forEachWithIndex` tests for the stdlib including `forEach` ones for modules like `Option` that weren't in the original PR.

PR #5235, opened 2023-01-04
ci: use release downloader in community build

### LINKED ISSUE #5218: Rework community build to avoid 403 errors
The community build is suffering from 403 errors, presumably from rate limiting. We currently use the `flix install` command to grab the projects, but since this does associate a GitHub API token, we hit rate limiting faster.

I propose we switch to something like https://github.com/marketplace/actions/release-downloader . It takes an API key so we shouldn't hit rate limiting, and simply downloading the release matches the current behavior of the community build.

mlutze:
Fixes #5218


mlutze:
Seems to work well enough. The problem with some community build projects is that they are marked as pre-releases. @jaschdoc is this a dealbreaker for you or can you do a full release?

jaschdoc:
Sure :)

mlutze:
> Sure :)

Cool, let me know when you've done that. I'll make a release for each of my projects and then we should be able to merge this.

magnus-madsen:
I think we can also remove my hello world package from the matrix.

jaschdoc:
I've created a new release now :)

PR #5239, opened 2023-01-04
feat: disable autocomplete for ??? #5226 

yisraelU:
Followup from conversation in #5228
new rules
1. ? as a first character of a line will not prod autocomplete
2. ? as a non first character, will not prod autocomplete if directly preceded by a whitespace


magnus-madsen:
Thanks!

Issue #5229, opened 2023-01-04
Restrictable Variants


magnus-madsen:
Tasks:

Phase 1:
- [x] Rename the syntax `choose` to `relational_choose` and `choose*` to `relational_choose*`.
- [x] Rename `Test.Exp.Choose` to `Test.Exp.RelationalChoose` and `Test.Exp.ChooseStar` to `Test.Exp.RelationalChooseStar`.
- [x] Rename `Expression.Choose` to `Expression.RelationalChoose` and rename `ChoiceRule` to `RelationalChoiceRule`, and rename `ChoicePattern` to `RelationalChoicePattern`.

Phase 2:
- [x] Add `Expression.Choose` with syntax `choose e { ... }` and `choose* e {...}`
- [x] Add restrictable enum syntax: `restrictable enum Color[s][...] {...}`. Add `Declaration.RestrictableEnum` and `RestrictableCase`.
- [x] Add `Symbol.RestrictableEnum` and `TypeConstructor.RestrictableEnum`.
- [x] #5244
- [x] Post `Resolver` add a `Expression.RestrictableTag`.
- [x] Add Resolver check for valid `choose*` body
- [ ] Translate all these constructs to normal enums and matches in `Lowering` instead of monomorph.
- [x] Add parametric Kind for these special restriction types
- [x] Add `Cons of exp` annotation and allow it as the `choose*` body in Resolver`
- [ ] Add LSP support
- [x] Add type/kind annotation
- [ ] Improve unification performance #5336

Phase 3: 
- [ ] Kind inference.
- [x] Type inference
- [ ] remove `of`
- [ ] Open sets (???).
- [ ] Equality (etc).
- [ ] Ponder this #5398

Phase 4:
- [ ] Generalize and reuse code to avoid unnecessary duplication.
    - [ ] Reuse Symbols? 
- [ ] Tests. #5311
- [ ] Review code together.

Cleanup:
- [ ] Figure out why this PR  made the resolver performance drop https://github.com/flix/flix/pull/5265

Features:
* Must refer to tags by enum namespace (`Expr.tag`)

JonathanStarup:
I think we need to add a new parametric kind for the restriction type parameters

JonathanStarup:
Should we add `~` or is the existing `not` sufficient?

mlutze:
> Should we add `~` or is the existing `not` sufficient? We could also add `-`

There should already be a ~ for set effects.

mlutze:
[...]

PR #5241, opened 2023-01-05
feat: add `choose` parsing

JonathanStarup:
Related to #5229

A choose rule is exactly like a match rule, which is then to be sorted out in the weeder

This works
```
def f(): Int32 \ Impure = {
    println(choose Nil {
        case Nil => "a"
        case Cons(1, _) => Int32.toString(1)
    });
    println(choose* Cons(1, Nil) {
        case Nil => Nil
        case Cons(x, rest) => Cons(x+1, rest)
    })
}
```


main/src/ca/uwaterloo/flix/language/ast/ParsedAst.scala
```diff
@@ -755,6 +755,17 @@ object ParsedAst {
...
+    /**
+      * Choose Expression.
+      *
+      * @param sp1   the position of the first character in the expression.
+      * @param star  whether this is a choose* expression.
+      * @param exp   the match expressions.
+      * @param rules the rules of the pattern match.
+      * @param sp2   the position of the last character in the expression.
+      */
+    case class Choose(sp1: SourcePosition, star: Boolean, exp: ParsedAst.Expression, rules: Seq[MatchRule], sp2: SourcePosition) extends ParsedAst.Expression
```
magnus-madsen, line 767: Do we want to call it RestrictableChoose? Or perhaps move it above RelationalChoose?

magnus-madsen:
I can merge?

PR #5243, opened 2023-01-05
feat: add restrictable enum to weeder

JonathanStarup:
Related to #5229


JonathanStarup:
@mlutze Do you think it's fine just to copy the Ast nodes like this? even though `RestrictableCase` is identical to normal `Case`

mlutze:
Yeah, it will make experimental changes eaiser

JonathanStarup:
Magnus voted for reuse, so now I'm in a pickle 🥒

But its likely that we wont change anything in the enum declaration case right?

Do you want to make a decision @magnus-madsen?

magnus-madsen:
We keep the copy. @mlutze Can you review?

mlutze:
LGTM. It will make my life harder with the namespace migration, but that was inevitable.

Issue #5172, opened 2022-12-22
Java function interfaces seem to be hardcoded to use `Static`?


paulbutcher:
While working on #5131 I discovered that Java function interfaces seem to be hardcoded to use the `Static` region?

Consider, for example, [this test](https://github.com/flix/flix/blob/master/main/test/flix/Test.Java.Function.flix#L80-L88). It uses `Static` cryptically because `ref` uses `Static` if it's not given an explicit region. If I change this to use an explicit region as follows:

```
    @Test
    def testIntConsumer(): Bool = region r {
        import static java.util.stream.IntStream.range(Int32, Int32): IntStream \ r;
        import java.util.stream.IntStream.forEach(IntConsumer): Unit \ r;
        let st = ref 0 @ r;
        let stream = range(0, 9);
        let _ = forEach(stream, i -> st := i);
        deref st == 8
    }
```

Then I get the error:

```
-- Type Error -------------------------------------------------- main/test/flix/Test.Java.Function.flix

>> Mismatched Pure and Effectful Functions.

86 |         let _ = forEach(stream, i -> st := i);
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                     mismatched pure and effectful functions.

Type One: (Int32 -> Unit & Impure) -> Unit & r
Type Two: (Int32 -> Unit & r) -> t0 & b1 \ e2
```

I can fix this with a cast:

```
    @Test
    def testIntConsumer(): Bool = region r {
        import static java.util.stream.IntStream.range(Int32, Int32): IntStream \ r;
        import java.util.stream.IntStream.forEach(IntConsumer): Unit \ r;
        let st = ref 0 @ r;
        let stream = range(0, 9);
        let _ = forEach(stream, unsafe_cast i -> st := i as Int32 -> Unit \ IO);
        deref st == 8
    }
```

But presumably we should provide some way to achieve this without casts?

**Aside:**

It's not clear to me why I have to specify the full type in the cast. Why doesn't this work?

```
let _ = forEach(stream, unsafe_cast i -> st := i as _ \ IO);
```

paulbutcher:
@stephentetley Do I understand the above correctly?

stephentetley:
I think you are understanding correctly. 

In Resolver.scala I think making Java functions always impure rather than effectful is probably too strong (c. line 3180). I don't know if we have the means at the moment to make them effectful though.

magnus-madsen:
[...]

PR #5253, opened 2023-01-06
chore: change ref initializations in tests to use explicit regions (part of issue #5215)

stephentetley:
This PR changes tests and one example to use explicit regions for ref initializations where they had previously used an implicit region.

Ref initializations without explicit regions were identified with this grep (first a match grep then a not match grep). There might still be some instances the grep didn't find.

> find . -name "*.flix" | xargs grep -n " ref " | grep -v "@"

Fusion tests (`TestDelayList` / `TestDelayMap`) - I've added explicit region to all, if a test doesn't use a cast to IO I've removed IO from the type signature.

For the `__Consumer` tests in `Test.Java.Function.flix` I have unfortunately had to use the Static region to get them to compile - this seems a instance of this issue: "Java function interfaces seem to be hardcoded to use Static?" #5172

### DIFF
--- a/examples/imperative-style-foreach-loops.flix
+++ b/examples/imperative-style-foreach-loops.flix
@@ -54,7 +54,7 @@ def main(): Unit \ IO = region r {
     /// expression in the foreach loop must be
     /// of type unit.
     ///
-    let z = ref Nil;
+    let z = ref Nil @ r;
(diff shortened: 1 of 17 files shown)


magnus-madsen:
Thanks- sorry for the delay.

PR #5269, opened 2023-01-09
fix: Use interruptible locking in Channel code

paulbutcher:
Further towards #5123

This switches locks within the channel code to be interruptible, and adds a couple of tests which manufacture deadlock within channel code and confirm that it can be interrupted.


magnus-madsen:
Thanks!

PR #5273, opened 2023-01-10
feat: perform resolution on choose expressions

mlutze:
related to #5229. There is a TODO in there that is waiting on another PR but it would not hurt to merge as-is.


magnus-madsen:
failed

JonathanStarup (approved):

PR #5276, opened 2023-01-10
chore: Reinstate virtual threads support

paulbutcher:
As per #5123


magnus-madsen:
We have no tests, but I assume you gave it a spin.

paulbutcher:
> We have no tests, but I assume you gave it a spin.

I did, yes. 

I briefly looked at what would be necessary to persuade Gradle to use the right JVM version and configure it with virtual threads enabled, but quickly lost the will to live.

PR #5278, opened 2023-01-10
feat: add restrictable tag to documentor

JonathanStarup:
Related to #5229


mlutze (commented):
Looks fine, but just need to be sure that extra fields won't crash the documentation machine

magnus-madsen:
> Looks fine, but just need to be sure that extra fields won't crash the documentation machine

Extra fields should not.

PR #5279, opened 2023-01-10
feat: lower restrictable tag declaration

JonathanStarup:
related to #5229


main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -219,6 +220,34 @@ object Lowering {
       LoweredAst.Enum(doc, ann, mod, sym, tparams, derives, cases, tpe, loc)
   }
 
+  /**
+    * Lowers the given enum `enum0`.
```
mlutze, line 224: should probably mention that it does Restrictable to non-restrictable enum

PR #5281, opened 2023-01-10
fix: Switch to interruptible locking for lazy

paulbutcher:
This is, I believe, the last task required for #5123


JonathanStarup (approved):

paulbutcher:
I just realised that I had created an unnecessary (and unused) label, so have taken the liberty of sneaking in a deletion ahead of merge.

PR #5290, opened 2023-01-11
feat: lower restrictable choose

JonathanStarup:
Related to #5229


main/src/ca/uwaterloo/flix/language/ast/Type.scala
```diff
@@ -663,6 +663,18 @@ object Type {
     }
   }
 
+  /**
+    * Constructs the tuple type (A, B, ...) where the types are drawn from the list `ts`.
+    *
+    * Returns Unit if the list is empty.
+    * Returns the head of the list if it is a singleton list.
+    */
+  def mkTuplish(ts: List[Type], loc: SourceLocation): Type = ts match {
```
magnus-madsen, line 672: Tuplish? :)

magnus-madsen: Does this function not already exist?

JonathanStarup: It doesn't seem like it no, we didn't find anything

PR #5302, opened 2023-01-14
feat: perform type inference on 'of' expression

mlutze:
related to #5229


JonathanStarup (approved):
I think this code clearly shows that the `of`-construct is superfluous since these type constraints could just be moved to the rule for `choose*`

mlutze:
> I think this code clearly shows that the `of`-construct is superfluous since these type constraints could just be moved to the rule for `choose*`

I don't think so. The choose rule doesn't check that it matches the tag, only the enum.

JonathanStarup:
> I don't think so. The `choose` rule doesn't check that it matches the tag, only the enum.

Yeah, that was a rash statement; the removal of the annotated tag would require unifying the expression with an abstract singleton `caseset`.

But I don't mean that this rule is redundant in its current form, only that the `choose` rule could be expanded to include these constraints.

Issue #5225, opened 2023-01-03
Proposal for `Closeable`


paulbutcher:
This is a proposal for the "Exploit the above to support Closable/AutoClosable resources." task of #4850

I propose that we add a new namespace `Region` containing (initially at least) a single function `Region.onExit` as follows:

```
namespace Region {
    pub def onExit(f: Unit -> Unit \ ef, r: Region[r]): Unit \ Write(r)
}
```

This will add `f` to a queue of functions to be called when the region exits:

1. These functions will be called *after* any threads within the region have been `join`ed.
2. I'm not sure whether we should guarantee that they're called in any particular order (it would be easy enough to guarantee that they are, but I'm not sure whether we want to constrain the language semantics so that we can't change this at some point in the future?)

We can then modify types such as `Channel` so that they call this on creation to register themselves to be closed when the region exits.

Although the obvious way to do this would be some kind of type `Closeable` type class (rather than just a `Unit -> Unit` function), I don't think that this is possible because the `close` function wouldn't use the class's type parameter.

I propose that we handle errors (such as a channel being closed while it still contains values, or a channel being used after it's been closed) by throwing an exception. Which isn't very "Flix" I know, but I believe that any such errors represent bugs, not "real" errors (which should be handled through mechanisms like `Result`)? The other option would be to modify the return value of `region` to something that can be used to report an error (e.g. `Result`, `Validation`, ?) but this would be quite disruptive.

So I guess the questions are:

1. Are we OK with not having a type class?
2. Are we OK with errors being handled by exceptions?
3. Should we define an order in which functions are called on region exit?
4. More prosaically: Are we agreed that closing a channel when it still contains data should be an error?
5. Which other types should register themselves to be closed?

magnus-madsen:
What if we keep what you have, but add:

```
class Closeable[t] {
    pub def close(x: t): Unit \ r
}
```

Then to get the `Unit -> Unit` function (which I think we need), we can do `Region.onExit(() -> Closable.close(x))` in the right place.

mlutze:
```scala
namespace Region {
    pub def onExit(f: Unit -> Unit \ ef, r: Region[r]): Unit \ Write(r)
}
```

Where does the `ef` go?

paulbutcher:
Hmm. How about instead of `onExit`, we have:

```
namespace Region {
    pub def closeOnExit(x: a): Unit with Closable[a] = $ONEXIT$(() -> Closable.close(x)
}
```

[...]

PR #5319, opened 2023-01-16
feat: Towards Closeable

paulbutcher:
Draft because not working yet, but I think I need help.

This is the beginnings of code to implement #5225. I've defined the `Closeable` type class and added a `Region.onClose` method as discussed. And I've plumbed an `$ONCLOSE$` intrinsic all the way through to code generation (although there's no code being generated for it at the moment).

Unfortunately this results in the following error:

```
  -- Type Error -------------------------------------------------- Region.flix

  >> Expected type: '? -> ? & ? \ ?' but found type: '? -> ? & b0 \ ?'.

  26 |     $ONEXIT$(() -> Closeable.close(x), r)
                    ^^^^^^^^^^^^^^^^^^^^^^^^
                    expression has unexpected type. (FlixSuite.scala:70)
```

I assume that the problem is in [this line](https://github.com/paulbutcher/flix/blob/f95a42b50343ee40d6a452dda4a6df92f5ac380d/main/src/ca/uwaterloo/flix/language/phase/Typer.scala#L939)  in the typer. But I'm not sure what I'm doing wrong.

Assistance gratefully received!


main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -928,6 +928,21 @@ object Typer {
...
 
+      case KindedAst.Expression.OnExit(exp1, exp2, loc) =>
+        val regionVar = Type.freshVar(Kind.Bool, loc)
+        val regionType = Type.mkRegion(regionVar, loc)
+        val p = Type.freshVar(Kind.Bool, loc)
+        val ef = Type.freshVar(Kind.Effect, loc)
+        for {
+          (constrs1, tpe1, pur1, eff1) <- visitExp(exp1)
+          (constrs2, tpe2, pur2, eff2) <- visitExp(exp2)
+          _ <- expectTypeM(expected = Type.mkUncurriedArrowWithEffect(Nil, p, ef, Type.Unit, loc.asSynthetic), actual = tpe1, exp1.loc)
```
magnus-madsen, line 939: I think `Nil` here should be unit?

main/src/library/Closeable.flix
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+///
+/// A type class for types that can be closed.
+///
+pub class Closeable[a] {
+  pub def close(x: a): Unit \ ef
```
magnus-madsen, line 21: I think ef should be IO

paulbutcher: When we discussed this [here](https://github.com/flix/flix/issues/5225#issuecomment-1369681567) you said:

> I think it has to be effect polymorphic.

Does this comment mean that you now think differently?

magnus-madsen: Argh. Not sure. I think it depends... 😮 

I still don't fully understand what it means to close a resource.

magnus-madsen:
Not for this PR, but do you think - at some point- it would be possible/desirable to express Region in Flix itself?

paulbutcher:
> Not for this PR, but do you think - at some point- it would be possible/desirable to express Region in Flix itself?

I can see the logic behind doing so. Let me have a think.

magnus-madsen:
What's the full story here? This PR allows me to register a `Closable` in a region and upon exit have it closed. So far so good.

But how do we use it?

Let's say I have an Java `InputStream`. I can define a `Closable` instance that calls `InputStream.close()`. That's pretty clear and fine. But how do I register it? Presumably I can call some Java method:

```
let is = Files.newInputStream(... some string...)
```

but how do the two get connected?

(This problem is not really related to this PR and I will look at the PR in isolation, but I feel like I am missing something.)

magnus-madsen:
Followup: Do we need a `Openable` and a try-with-resources kind of thing that takes a value and a region and connects the two?

magnus-madsen:
Thought on this:

```
pub class Closeable[a] {
  pub def close(x: a): Unit \ ef
}

enum FlixInputStream[rh: Bool] {
    case IS(Region[rh], ##java.io.InputStream)
}

instance Closeable[FlixInputStream[rh]] {
    pub def close(x: FlixInputStream[rh]): Unit & ef = unsafe_cast ??? as _ \ ef
}
```

paulbutcher:
> What's the full story here?

So I think we need to disconnect the Java interop question here from the Flix case.

The Flix case is pretty straightforward: anything that's `Closeable` will also take a region. And that's the region that it's associated with.

So, for channels:

```
    region r1 {
        region r2 {
             Channel.unbuffered(r1);
             ...
         }
         ....
         // Channel is closed here because it's associated with r1
     }
```

For Java interop, we'll need some additional machinery. But the case that we should think about first is the Flix case.

main/src/ca/uwaterloo/flix/language/phase/Weeder.scala
```diff
@@ -721,6 +721,8 @@ object Weeder {
           case ("ARRAY_SLICE", e1 :: e2 :: e3 :: e4 :: Nil) => WeededAst.Expression.ArraySlice(e1, e2, e3, e4, loc).toSuccess
           case ("ARRAY_LENGTH", e1 :: Nil) => WeededAst.Expression.ArrayLength(e1, loc).toSuccess
 
+          case ("ONEXIT", e1 :: e2 :: Nil) => WeededAst.Expression.OnExit(e1, e2, loc).toSuccess
```
mlutze, line 724: I think for all the others we have `_` separated words

main/src/library/Closeable.flix
mlutze: I'm also skeptical of the `ef` effect.

mlutze: I'm inclined to think the ultimate solution is an associated type. But `IO` would be the overapproximation.

main/src/library/Region.flix
```diff
@@ -0,0 +1,27 @@
...
+
+namespace Region {
+
+  ///
+  /// Add x to a queue of `Closeable` to be closed when the region exits:
+  ///
+  /// * `Closeable`s are guaranteed to be closed after any threads within the region have been joined.
+  /// * `Closeable`s are not guaranteed to be closed in any particular order.
+  ///
+  pub def closeOnExit(x: a, r: Region[r]): Unit \ r with Closeable[a] =
```
mlutze, line 25: Per above, if Closable.close is IO then this would need to change too.

main/src/ca/uwaterloo/flix/api/Flix.scala
```diff
@@ -140,6 +140,7 @@ class Flix {
     "Chain.flix" -> LocalResource.get("/src/library/Chain.flix"),
     "Char.flix" -> LocalResource.get("/src/library/Char.flix"),
     "Choice.flix" -> LocalResource.get("/src/library/Choice.flix"),
+    "Closeable.flix" -> LocalResource.get("/src/library/Closeable.flix"),
```
mlutze, line 143: Since this will be a sugary feature, we might need to add it to the `MinLib` too.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -957,6 +957,21 @@ object Typer {
           resultEff = eff
         } yield (constrs, resultTyp, resultPur, resultEff)
 
+      case KindedAst.Expression.OnExit(exp1, exp2, loc) =>
```
mlutze, line 960: What I understand here is that we allow the first argument to have any effect. This effect will eventually be executed upon exiting something, but it doesn't show up in the effect system. This feels a little strange to me.

main/src/library/Closeable.flix
paulbutcher: OK, sounds like we should go with `IO` for now and perhaps get cuter in the future. I'll make the change.

main/src/library/Region.flix
paulbutcher: Is that true? All this is doing is registering the `Closable` to be closed when the region exits. `closeOnExit` has no effect other than to write to the region, surely?

main/src/ca/uwaterloo/flix/api/Flix.scala
paulbutcher: Good point.

paulbutcher: Hmm. Actually, I'm not so sure about that now I think about it. It's only used if something `Closeable` is ever created, and I don't imagine that anything `Closeable` will ever be in `MinLib`?

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
paulbutcher: I understand why it feels strange. *In practice* I think it'll work out because anything `Closeable` will have non-trivial (probably, but not necessarily, `IO`) effects within the region.

But I agree that the effect being "swallowed" here isn't idea. I guess the effect should be added to the effect of the `Scope` with which this `onClose` is associated? I'm not sure how to achieve that though?

paulbutcher: A thought:

The thing *calling* `onClose` will always be within a region. If we add the effect of the function passed to `onClose` to the `onClose` itself, it will end up in the region. At the "wrong" point (the point where `onClose` is called, not the point at which the region terminates) but perhaps that is OK?

mlutze: I see. I think the ideal would be to have an effect that carries the idea that there's no effect _here_, but it will occur later, like `Latent[IO]` or even `Latent[r, IO]` to associate a region. Then upon reaching the end of the region, the effect is translated into the proper type.

Of course, our system cannot handle that kind of complexity. 

For now, I think the safest approximation is to put the effect on the `onClose`.

main/src/library/Region.flix
mlutze: Ah right. The effect comes from the `onExit` call, I suppose.

Maybe I'll understand this all better upon seeing some example programs.

magnus-madsen:
I have been doodling:

```flix

pub class Closeable[a] {
    pub def close(x: a): Unit \ IO
}

instance Closeable[FileInputStream[r]] {
    pub def close(x: FileInputStream[r]): Unit \ IO = 
        // ... actually implementation would be here ...
        unsafe_cast ??? as _ \ IO
}


enum FileInputStream[r: Region]{
    case FileInputStream(Region[r], ##java.io.FileInputStream)
}

def open(rc: Region[r], file: String): FileInputStream[r] \ IO = 
    import new java.io.FileInputStream(String): ##java.io.FileInputStream \ IO as newStream;
    FileInputStream(rc, newStream(file))

def main(): Unit \ IO = 
    region rc {
        let _s = open(rc, "hello.txt");
        // how do we register that `_s` should be closed?
        println("Hello World!")
    }

// def main2(): Unit \ IO = 
//     region rc {
//         using(_s <- open(rc, "hello.txt")) {
//             println("Hello World!")
//         }
//     }

// // desugars to

// def main2(): Unit \ IO = 
//     region rc {
//         let s = open(rc, "hello.txt");
//         Region.closeOnExit(s, rc);
//         println("Hello World!");
//     }

// but why not then just to:

// def main3(): Unit \ IO = 
//     region rc {
//         let s = open(rc, "hello.txt");
//         println("Hello World!");
//         Closable.close(s)
//     }
```

magnus-madsen:
The question here is whether the closable resource needs to be associated with its region at runtime.

This is certainly the case for threads (because we have to propagate exceptions across them-- which is not standard).

But for resource, consider:

- If you have a region type variable in the resource then you automatically ensure the data is lexically scope (i.e. cannot escape).
- If we plan to introduce a `using` (or similar construct) then we can either associate the resource with runtime region (and close it when the region is closed) or we can simply insert a call to `close`.

mlutze:
Does the `using` construct you propose introduce its own region?

magnus-madsen:
> Does the `using` construct you propose introduce its own region?

It could or not. That's a choice we have to make.

paulbutcher:
I'm busy today, but I think this is much simpler than you're suggesting. I'll comment when I have time (not sure I will today).

paulbutcher:
Second, I really don't understand why we would need `using` or anything like it. I think the language as it currently stands gives us everything we need.

To take Magnus' example above:

```
def main(): Unit \ IO = 
    region rc {
        let _s = open(rc, "hello.txt");
        // how do we register that `_s` should be closed?
        println("Hello World!")
    }
```

The simplest, easiest, (and to my mind correct) solution is to modify `open` as follows:

```
def open(rc: Region[r], file: String): FileInputStream[r] \ IO = 
    import new java.io.FileInputStream(String): ##java.io.FileInputStream \ IO as newStream;
    let stream = FileInputStream(rc, newStream(file));
    Region.onClose(stream, rc);
    stream
```

This works as-is with the language as it currently stands, and has very obvious semantics: anything `Closeable` is closed when its region goes out of scope (what would it even mean for it *not* to be closed when its region goes out of scope?)

magnus-madsen:
Thanks for the comments Paul. I am back in thinking mode...

A few other comments:

>> If you have a region type variable in the resource then you automatically ensure the data is lexically scope (i.e. cannot escape).

> This isn't entirely true. Specifically, Channel has unsafeSend which can be used to send a value (most likely a channel handle, but potentially anything) to a thread which isn't within the same region and definitely might try to use that value outside its lexical scope.

While that may be true today, it is absolutely, 100%, our intention that the effect system should be sound and that memory belonging to a region cannot escape its scope.

paulbutcher:
> Thanks for the comments Paul. I am back in thinking mode...

Can you help me understand why this is a complicated decision? It seems very straightforward to me, which makes me worry that there's something I'm missing? Is there some use case that you're aware of that won't be handled by the approach I suggested above?

paulbutcher:
> You can forget to call onExit.

This is a problem for library authors, but not (in the absence of Java interop) for user code. So let's imagine that we have a `File` type, the library author has to remember to call `onClose` in `File.open`, but users of `File.open` get it automatically.

It's an issue for Java interop, certainly, but (as I've said before) I'm wary of complicating the language to make interop easier. And we can easily create utility functions which make this pattern straightforward?

magnus-madsen:
It seems to me that Java/C#/Kotlin/... all offer a construct like:

```
using (x = FileStream.open(...)) {
  // use the stream
}
```

I think this is the right abstraction, but what can go wrong is:

1. We could call `FileStream.open` outside of a `using` block (e.g. just somewhere).
2. We could have `x` escape its lexical scope.
3. We could have `x` closed inside the lexical scope, and have a use after its been closed, but before the scope ends.

I want to solve all three, but I think we only have the infrastructure for (2). 

Its possible the solution is `THIS PR + region-with-resources + something more`.

paulbutcher:
> It seems to me that Java/C#/Kotlin/... all offer a construct like:

They do, but they don't already have regions. I'm worried that we're reinventing regions here.

> 1. We could call `FileStream.open` outside of a `using` block (e.g. just somewhere).

Right. But we *can't* call `FileStream.open` without passing it a region. If `FileStream.open` calls `Region.onExit` then job done, yes?

magnus-madsen:
I have not forgotten about this. My current thinking is that we need both systems (i.e. this PR + the `region-with-resources` construct). Let me stew a bit more and return :)

paulbutcher:
Thanks Magnus, good to know.

If you're pretty certain that we need this PR, can we merge this and I can push ahead with the machinery to actually support the functionality? And we can then layer whatever syntax on top we decide we need?

magnus-madsen:
> Thanks Magnus, good to know.
> 
> If you're pretty certain that we need this PR, can we merge this and I can push ahead with the machinery to actually support the functionality? And we can then layer whatever syntax on top we decide we need?

Well, let me think :-) Give me a bit more time. (I typically need X showers before I am convinced :))

(I did not yet look at the code, so I may have low-level comments.)

main/src/ca/uwaterloo/flix/api/Flix.scala
paulbutcher: I'm going to mark this as resolved: we can add it to MinLib as/when it becomes necessary.

paulbutcher:
I've addressed @mlutze's comments, so this should be ready for you to take a look as/when you're ready @magnus-madsen

magnus-madsen:
I thought about it and I am OK with this approach. I don't think its the final version of "resource management for Flix", but at least I think its a step in the right direction. In addition to this, I think we need:

- We should add support for `region rc with ...`.
- We need to decide carefully what happens if you close a resource twice.
- In the future, I would like to explore how we can ensure that if you implement `Closable` then you cannot forget to use ON_CLOSE (or whatever it will be called).

magnus-madsen:
I did not review the code yet, so I will do that now.

main/src/ca/uwaterloo/flix/api/Flix.scala
magnus-madsen: Agreed.

main/src/library/Region.flix
```diff
@@ -0,0 +1,27 @@
...
+
+  ///
+  /// Add x to a queue of `Closeable` to be closed when the region exits:
+  ///
+  /// * `Closeable`s are guaranteed to be closed after any threads within the region have been joined.
+  /// * `Closeable`s are not guaranteed to be closed in any particular order.
+  ///
+  pub def closeOnExit(x: a, r: Region[r]): Unit \ {r, IO} with Closeable[a] = 
+    unsafe_cast $ON_EXIT$(() -> Closeable.close(x), r) as _ \ {r, IO}
+}
```
magnus-madsen, line 27: use 4 space indentation.

main/src/library/Region.flix
```diff
@@ -0,0 +1,27 @@
...
+
+namespace Region {
+
+  ///
+  /// Add x to a queue of `Closeable` to be closed when the region exits:
+  ///
+  /// * `Closeable`s are guaranteed to be closed after any threads within the region have been joined.
+  /// * `Closeable`s are not guaranteed to be closed in any particular order.
+  ///
+  pub def closeOnExit(x: a, r: Region[r]): Unit \ {r, IO} with Closeable[a] =
```
magnus-madsen, line 25: I agree with this type signature.

main/src/library/Closeable.flix
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+///
+/// A type class for types that can be closed.
+///
+pub class Closeable[a] {
+  pub def close(x: a): Unit \ IO
```
magnus-madsen, line 21: use 4 space indent

main/src/ca/uwaterloo/flix/api/lsp/Indexer.scala
```diff
@@ -207,6 +207,9 @@ object Indexer {
       val tpe = Type.mkRegion(sym.tvar, loc)
       Index.occurrenceOf(sym, tpe) ++ visitExp(exp) ++ Index.occurrenceOf(exp0)
 
+    case Expression.OnExit(exp1, exp2, _, _, _, _) =>
```
magnus-madsen, line 210: I would like to rename `OnExit`  to `ScopeExit` everywhere. 

This name is probably not that much better, but it reflects the name of a very similar construct we use in our paper + Isabelle proof, so to keep things consistent.

main/src/ca/uwaterloo/flix/language/phase/Stratifier.scala
```diff
@@ -152,6 +152,11 @@ object Stratifier {
         case e => Expression.Scope(sym, regionVar, e, tpe, pur, eff, loc)
       }
 
+    case Expression.OnExit(exp1, exp2, tpe, pur, eff, loc) =>
+      mapN(visitExp(exp1), visitExp(exp2)) {
+        case (r, e) => Expression.OnExit(r, e, tpe, pur, eff, loc)
```
magnus-madsen, line 157: could just use e1 and e2?

main/src/library/Region.flix
```diff
@@ -0,0 +1,27 @@
...
+namespace Region {
+
+  ///
+  /// Add x to a queue of `Closeable` to be closed when the region exits:
+  ///
+  /// * `Closeable`s are guaranteed to be closed after any threads within the region have been joined.
+  /// * `Closeable`s are not guaranteed to be closed in any particular order.
+  ///
+  pub def closeOnExit(x: a, r: Region[r]): Unit \ {r, IO} with Closeable[a] = 
+    unsafe_cast $ON_EXIT$(() -> Closeable.close(x), r) as _ \ {r, IO}
```
magnus-madsen, line 26: I think we need to fix the type inference for ON_EXIT so that this cast disappears.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
magnus-madsen: To chime in, but in a different way:

Its important that ON_EXIT is not pure because if it is then the optimizer can remove it.

Right now it can be pure if e1 and e2 are. Can they be? The answer is yes, because they are a lambda and a region. 

So we need to change the `resultPur` to be `Impure`. (Or if you want `r and Impure`).

main/src/library/Closeable.flix
paulbutcher: Oops, sorry for missing that (blush)

paulbutcher:
Thanks @magnus-madsen. I think I've addressed all your comments.

magnus-madsen:
Thanks!

PR #5325, opened 2023-01-17
feat: add literal case sets

JonathanStarup:
Now a case set constant can be `Constant(Cst + Var + Not)`

This doesn't in itself make the unification that much faster since it only helps when two constants "touch".

Right now, when trying to minimize by table, a constant like the one above is turned into three variables, one for each singleton, so the number of variables in regards to table lookup is unchanged. 

This PR doesn't introduce any simplifications involving knowledge of the universe, this is both from practical issues and lacking knowledge on my side about when actually materializing the complement is good (maybe always?).


JonathanStarup:
Based on by-hand testing it seems i broke typing somehow

magnus-madsen:
I am closing this because its too far behind master. But we do want constant sets in the types, but without anything fancy.

PR #5330, opened 2023-01-18
New iterator (WIP)

stephentetley:
This is a _work in progress_.

Iterator has been changed to use a region encapsulated in the constructor and just a `next` function rather than two functions `done` and `next`.

Unfortunately the effect types are difficult to work with, in some cases not the type they ought to be...

### DIFF
--- a/main/src/library/Foldable.flix
+++ b/main/src/library/Foldable.flix
@@ -233,7 +233,7 @@ pub class Foldable[t : Type -> Type] {
                 None
             }
         };
-        Iterator.new(r, next)
+        Iterator.iterate(r, next)
 
     ///
     /// Returns an iterator over `t` zipped with the indices of the elements.
--- a/main/src/library/Foldable.flix
+++ b/main/src/library/Foldable.flix
@@ -219,23 +219,26 @@ pub class Foldable[t : Type -> Type] {
     ///
     /// Returns an iterator over `t`.
     ///
-    pub def iterator(r: Region[r], t: t[a]): Iterator[a, ef, r] \ Write(r) =
+    pub def iterator(r: Region[r], t: t[a]): Iterator[a, r, r] \ Write(r) =
         let arr = Foldable.toArray(r, t);
         let len = Array.length(arr);
         let i = ref 0 @ r;
-        let done = () -> (deref i) >= len;
         let next = () -> {
-            let j = deref i;
-            let x = arr[j];
-            i := j + 1;
-            x
+            if ((deref i) < len) {
+                let j = deref i;
+                let x = arr[j];
+                i := j + 1;
+                Some(x)
+            } else {
+                None
+            }
         };
-        Iterator.Iterator(done, next)
+        Iterator.new(r, next)
 
     ///
     /// Returns an iterator over `t` zipped with the indices of the elements.
     ///
-    pub def enumerator(r: Region[r], t: t[a]): Iterator[(Int32, a), ef, r] \ Write(r) =
+    pub def enumerator(r: Region[r], t: t[a]): Iterator[(Int32, a), r, r] \ Write(r) =
         Foldable.iterator(r, t) |> Iterator.zipWithIndex
 
     ///
--- a/main/src/library/Foldable.flix
+++ b/main/src/library/Foldable.flix
@@ -219,7 +219,7 @@ pub class Foldable[t : Type -> Type] {
     ///
     /// Returns an iterator over `t`.
     ///
-    pub def iterator(r: Region[r], t: t[a]): Iterator[a, r] \ Write(r) =
+    pub def iterator(r: Region[r], t: t[a]): Iterator[a, ef, r] \ Write(r) =
         let arr = Foldable.toArray(r, t);
         let len = Array.length(arr);
         let i = ref 0 @ r;
@@ -235,7 +235,7 @@ pub class Foldable[t : Type -> Type] {
     ///
     /// Returns an iterator over `t` zipped with the indices of the elements.
     ///
-    pub def enumerator(r: Region[r], t: t[a]): Iterator[(Int32, a), r] \ Write(r) =
+    pub def enumerator(r: Region[r], t: t[a]): Iterator[(Int32, a), ef, r] \ Write(r) =
         Foldable.iterator(r, t) |> Iterator.zipWithIndex
 
     ///
--- a/main/src/library/Iterator.flix
+++ b/main/src/library/Iterator.flix
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 Jakob Schneider Villumsen
+ * Copyright 2021 Jakob Schneider Villumsen, Stephen Tetley
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,10 +14,11 @@
  * limitations under the License.
  */
 
-pub enum Iterator[a: Type, r: Region] {
-    case Iterator(Iterator.Done[r], Iterator.Next[a, r])
+pub enum Iterator[a: Type, ef: Bool, r: Region] {
+    case Iterator(Region[r], Unit -> Step[a] \ {r, ef})
 }
 
+/*
 instance Newable[Iterator[a]] {
     pub def new(r: Region[r]): Iterator[a, r] \ { Read(r), Write(r) } = Iterator.new(r)
 }
@@ -31,26 +32,33 @@ instance Scoped[Iterator[a]] {
 ///
 pub def iterator(r: Region[r], t: t[a]): Iterator[a, r] \ Write(r) with Foldable[t] =
     Foldable.iterator(r, t)
+*/
 
 namespace Iterator {
 
-    ///
-    /// The type of the done function.
-    ///
-    pub type alias Done[ef: Region] = Unit -> Bool \ ef
-
-    ///
-    /// The type of the next function.
-    ///
-    pub type alias Next[a: Type, ef: Region] = Unit -> a \ ef
+    pub enum Step[a] {
+        case Ans(a)
+        case Skip
+        case Done
+    }
 
     ///
     /// Returns an empty iterator.
     ///
-    pub def new(_: Region[r]): Iterator[a, r] \ { Read(r), Write(r) } =
-        let done = () -> unsafe_cast true as _ \ Read(r);
-        let next = () -> unsafe_cast bug!("Empty Iterator!") as _ \ Read(r);
-        unsafe_cast Iterator(done, next) as _ \ { Read(r), Write(r) }
+    pub def new(r: Region[r], f: Unit -> Option[a] \ r): Iterator[a, ef, r] =
+        let f1 = () -> match f() {
+            case Some(a) => Ans(a)
+            case None    => Done
+        };
+        let iterF = () -> unsafe_cast f1()  as _ \ { ef, r };
+        Iterator(r, iterF)
+
+    ///
+    /// Returns an empty iterator.
+    ///
+    pub def empty(r: Region[r]): Iterator[a, ef, r] =
+        let f = () -> unsafe_cast Done as _ \ ef;
+        Iterator(r, f)
 
     ///
     /// Returns an iterator containing only a single element, `x`.
@@ -74,70 +82,58 @@ namespace Iterator {
     ///
     /// Consumes the head element of `iter`.
     ///
-    pub def next(iter: Iterator[a, r]): Option[a] \ Read(r) =
-        let Iterator(done, next) = iter;
-        if (done())
-            None
-        else
-            Some(next())
-
-    ///
-    /// Returns true if the iterator is empty.
-    ///
-    /// Does *not* consume any elements of the iterator.
-    ///
-    pub def isEmpty(iter: Iterator[a, r]): Bool \ Read(r) =
-        let Iterator(done, _) = iter;
-        done()
+    pub def next(iter: Iterator[a, ef, r]): Option[a] \ {ef, Read(r) } =
+        let Iterator(_, iterF) = iter;
+        def loop() = match iterF() {
+            case Ans(a) => Some(a)
+            case Skip   => loop()
+            case Done   => None
+        };
+        loop()
 
     ///
     /// Returns an iterator of all integers between `b` (inclusive) and `e` (exclusive).
     ///
     /// Returns an empty iterator if `b >= e`.
     ///
-    @Lazy
-    pub def range(r: Region[r], b: Int32, e: Int32): Iterator[Int32, r] \ Write(r) =
-        let i = ref b @ r;
-        let done = () -> deref i >= e;
-        let next = () -> {
-                let res = deref i;
-                if (res < e) {
-                    i := res + 1;
-                    res
-                }
-                else {
-                    bug!("Empty Iterator!")
-                }
+    pub def range(r: Region[r], b: Int32, e: Int32): Iterator[Int32, ef, r] \ Write(r) =
+        let pos = ref b @ r;
+        if (e <= b)
+            empty(r)
+        else {
+            let f = () -> match (deref pos) {
+                case i if i < e => pos := i+1; Ans(i)
+                case _          => Done
             };
-        Iterator(done, next)
+            let iterF = () -> unsafe_cast f() as _ \ { ef, r };
+            Iterator(r, iterF)
+        }
 
     ///
     /// Returns an iterator over an iterable with the element `x` repeated `n` times.
     ///
     /// Returns an empty iterator if `n < 0`.
     ///
-    @Lazy
-    pub def repeat(r: Region[r], n: Int32, x: a): Iterator[a, r] \ Write(r) =
-        let i = ref 0 @ r;
-        let done = () -> deref i >= n;
-        let next = () -> {
-              let j = deref i;
-              if (j < n) {
-                i := j + 1;
-                x
-              }
-              else {
-                bug!("Empty Iterator!")
-              }
-            };
-        Iterator(done, next)
+    pub def repeat(r: Region[r], n: Int32, x: a): Iterator[a, ef, r] \ Write(r) =
+        let ix = ref n @ r;
+        let f = () -> {
+            let i = deref ix;
+            if (i < 1)
+                Done
+            else {
+                ix := i-1;
+                Ans(x)
+            }
+        };
+        let iterF = () -> unsafe_cast f() as _ \ { ef, r };
+        Iterator(r, iterF)
 
     ///
     /// Returns the sum of all elements in the iterator `iter`.
     ///
     /// Consumes the entire iterator.
     ///
-    pub def sum(iter: Iterator[Int32, r]): Int32 \ Read(r) =
+    pub def sum(iter: Iterator[Int32, ef, r]): Int32 \ { ef, Read(r) } =
         foldLeft((acc, x) -> acc + x, 0, iter)
 
     ///
@@ -145,7 +141,7 @@ namespace Iterator {
     ///
     /// Consumes the entire iterator.
     ///
-    pub def sumWith(f: a -> Int32 \ ef, iter: Iterator[a, r]): Int32 \ { ef, Read(r) } =
+    pub def sumWith(f: a -> Int32 \ ef1, iter: Iterator[a, ef, r]): Int32 \ { ef, ef1, Read(r) } =
         foldLeft((acc, x) -> acc + f(x), 0, iter)
 
     ///
@@ -153,29 +149,27 @@ namespace Iterator {
     ///
     /// Consumes the entire iterator.
     ///
-    pub def product(iter: Iterator[Int32, r]): Int32 \ Read(r) =
-        if (isEmpty(iter))
-            1
-        else
-            foldLeft((acc, x) -> acc * x, 1, iter)
+    pub def product(iter: Iterator[Int32, ef, r]): Int32 \ { ef, Read(r) } = match next(iter) {
+        case Some(i) => foldLeft((acc, x) -> acc * x, i, iter)
+        case None    => 1
+    }
 
     ///
     /// Returns the product of all elements in the iterator `iter` according to the function `f`.
     ///
     /// Consumes the entire iterator.
     ///
-    pub def productWith(f: a -> Int32 \ ef, iter: Iterator[a, r]): Int32 \ { ef, Read(r) } =
-        if (isEmpty(iter))
-            1
-        else
-            foldLeft((acc, x) -> acc * f(x), 1, iter)
+    pub def productWith(f: a -> Int32 \ ef1, iter: Iterator[a, ef, r]): Int32 \ { ef, ef1, Read(r) } = match next(iter) {
+        case Some(a) => foldLeft((acc, x) -> acc * f(x), f(a), iter)
+        case None    => 1
+    }
 
     ///
     /// Returns the contents of `iter` as an array.
     ///
     /// Consumes the entire iterator.
     ///
-    pub def toArray(r1: Region[r1], iter: Iterator[a, r2]): Array[a, r1] \ { Read(r2), Write(r1) } = region r0 {
+    pub def toArray(r1: Region[r1], iter: Iterator[a, ef, r2]): Array[a, r1] \ { ef, Read(r2), Write(r1) } = region r0 {
         let m = new MutList(r0);
         forEach(a -> MutList.push!(a, m), iter);
         MutList.toArray(r1, m)
@@ -186,7 +180,7 @@ namespace Iterator {
     ///
     /// Consumes the entire iterator.
     ///
-    pub def toList(iter: Iterator[a, r]): List[a] \ Read(r) =
+    pub def toList(iter: Iterator[a, ef, r]): List[a] \ { ef, Read(r) } =
         foldRight((a, acc) -> a :: acc, Nil, iter)
 
 
@@ -195,7 +189,7 @@ namespace Iterator {
     ///
     /// Consumes the entire iterator.
     ///
-    pub def toChain(iter: Iterator[a, r]): Chain[a] \ Read(r) =
+    pub def toChain(iter: Iterator[a, ef, r]): Chain[a] \ { ef, Read(r) } =
         foldLeft((acc, a) -> Chain.snoc(acc, a), Chain.empty(), iter)
 
     ///
@@ -203,37 +197,39 @@ namespace Iterator {
     ///
     /// Consumes the entire iterator.
     ///
-    pub def toMap(iter: Iterator[(a, b), r]): Map[a, b] \ Read(r) with Order[a] =
-        foldLeft((acc, ab) -> Map.insert(fst(ab), snd(ab), acc), Map.empty(), iter)
+    pub def toMap(iter: Iterator[(k, v), ef, r]): Map[k, v] \ { ef, Read(r) } with Order[k] =
+        foldLeft((acc, a) -> {let (k, v) = a; Map.insert(k, v, acc)}, Map.empty(), iter)
 
     ///
     /// Returns the contents of `iter` as a `Some(Nel)` if `iter` is not empty. Returns None otherwise.
     ///
     /// Consumes the entire iterator.
     ///
-    pub def toNel(iter: Iterator[a, r]): Option[Nel[a]] \ Read(r) = match toList(iter) {
-        case Nil     => None
-        case x :: xs => Some(Nel.Nel(x, xs))
-    }
+    pub def toNel(iter: Iterator[a, ef, r]): Option[Nel[a]] \ { ef, Read(r) } =
+        let swapCons = (a, acc) -> match acc {
+            case Nel(x, xs) => Nel(x, a :: xs)
+        };
+        match next(iter) {
+            case Some(a) => foldRight((x, acc) -> swapCons(x, acc), Nel.singleton(a), iter) |> Some
+            case None    => None
+        }
 
     ///
     /// Returns the contents of `iter` as a nec.
     ///
     /// Consumes the entire iterator.
     ///
-    pub def toNec(iter: Iterator[a, r]): Option[Nec[a]] \ Read(r) =
-        let f = (x, acc) -> match acc {
-            case None      => Some(Nec.singleton(x))
-            case Some(nec) => Some(Nec.cons(x, nec))
-        };
-        foldRight(f, None, iter)
+    pub def toNec(iter: Iterator[a, ef, r]): Option[Nec[a]] \ { ef, Read(r) } = match next(iter) {
+        case Some(a) => foldLeft((acc, x) -> Nec.snoc(acc, x), Nec.singleton(a), iter) |> Some
+        case None    => None
+    }
 
     ///
     /// Returns the contents of `iter` as a MutDeque.
     ///
     /// Consumes the entire iterator.
     ///
-    pub def toMutDeque(r1: Region[r1], iter: Iterator[a, r2]): MutDeque[a, r1] \ { Read(r2), Write(r1) }  =
+    pub def toMutDeque(r1: Region[r1], iter: Iterator[a, ef, r2]): MutDeque[a, r1] \ { ef, Read(r2), Write(r1) } =
         let d = new MutDeque(r1);
         forEach(x -> MutDeque.pushBack(x, d), iter);
         d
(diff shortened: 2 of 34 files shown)


magnus-madsen:
Thanks @stephentetley I will take a look, but might be a few days. 

As for the effects, I will look at it, but I think I understand the underlying issue (which would require some future changes on my side).

main/src/library/MutSet.flix
```diff
@@ -412,14 +412,14 @@ namespace MutSet {
     ///
     /// Returns an iterator over `s`.
     ///
-    pub def iterator(r1: Region[r1], s: MutSet[a, r2]): Iterator[a, r1 and r2] \ { Write(r1), Read(r2) } =
+    pub def iterator(r1: Region[r1], s: MutSet[a, r2]): Iterator[a, r1 and r2, r1] \ { Write(r1), Read(r2) } =
         let MutSet(ms) = s;
-        unsafe_cast Set.iterator(r1, deref ms) as Iterator[a, r1 and r2]
+        unsafe_cast Set.iterator(r1, deref ms) as Iterator[a, r1 and r2, r1]
```
magnus-madsen, line 417: This looks interesting. I think the effect system is "exposing us" for the fact that iteration through the set does not really touch `r2` because the underlying set is immutable!

Does the following help get rid of the cast?

```
Set.iterator(r1, deref ms) |> Iterator.map(x -> {discard deref ms; x})
```

(Yes, I know the expression is stupid).

magnus-madsen: If something like this can work, then I think should we apply it.

main/src/library/Map.flix
```diff
@@ -911,21 +911,21 @@ namespace Map {
...
         let Map(t) = m;
         RedBlackTree.iterator(r, t)
 
     ///
     /// Returns an iterator over keys in `m`.
     ///
-    pub def iteratorKeys(r: Region[r], m: Map[k, v]): Iterator[k, r] \ Write(r) =
-        iterator(r, m) |> Iterator.mapL(match (k, _) -> k)
+    pub def iteratorKeys(r: Region[r], m: Map[k, v]): Iterator[k, r, r] \ Write(r) =
+        iterator(r, m) |> Iterator.map(match (k, _) -> k)
```
magnus-madsen, line 922: Beautiful!

magnus-madsen: Or, even better, Iterator.map(fst) ?

main/src/library/Foldable.flix
```diff
@@ -219,23 +219,26 @@ pub class Foldable[t : Type -> Type] {
     ///
     /// Returns an iterator over `t`.
     ///
-    pub def iterator(r: Region[r], t: t[a]): Iterator[a, r] \ Write(r) =
+    pub def iterator(r: Region[r], t: t[a]): Iterator[a, r, r] \ Write(r) =
         let arr = Foldable.toArray(r, t);
```
magnus-madsen, line 223: Can this be avoided? Or is it not yet the time?

magnus-madsen (commented):
Thanks Stephen. This was a pleasure to read through and think about.

I looked at everything except `Iterator.flix` itself -- which I will look at next.

I added two major comments:

- An idea to get rid of some of the casts via a "clever" (redundant) map operation.
- A question of whether using Iterator.range followed by Iterator.map would not be more idiomatic in several cases.

main/src/library/Iterator.flix
```diff
@@ -14,179 +14,158 @@
  * limitations under the License.
  */
 
-pub enum Iterator[a: Type, r: Region] {
-    case Iterator(Iterator.Done[r], Iterator.Next[a, r])
+pub enum Iterator[a: Type, ef: Bool, r: Region] {
+    case Iterator(Region[r], Unit -> Iterator.Step[a] \ {r, ef})
 }
 
+/*
```
magnus-madsen, line 21: Everything in the comment can just be deleted. That's fine.

main/src/library/Iterator.flix
```diff
@@ -14,179 +14,158 @@
  * limitations under the License.
  */
 
-pub enum Iterator[a: Type, r: Region] {
-    case Iterator(Iterator.Done[r], Iterator.Next[a, r])
+pub enum Iterator[a: Type, ef: Bool, r: Region] {
```
magnus-madsen, line 17: I think this is a beautiful way to view iterators. There is the memory effects used by the iterator itself (r) and the user-effects (ef) collected by the user.

main/src/library/Iterator.flix
```diff
@@ -14,179 +14,158 @@
...
     Foldable.iterator(r, t)
+*/
 
 namespace Iterator {
 
-    ///
-    /// The type of the done function.
-    ///
-    pub type alias Done[ef: Region] = Unit -> Bool \ ef
+    pub enum Step[a] {
```
magnus-madsen, line 35: Perhaps a comment (to explain the obvious). Maybe focus on the idea that Skip is useful for filtering.

main/src/library/Iterator.flix
```diff
@@ -14,179 +14,158 @@
...
-    /// The type of the next function.
+    /// Returns an iterator built with the stepper function `f`.
     ///
-    pub type alias Next[a: Type, ef: Region] = Unit -> a \ ef
+    pub def iterate(rc: Region[r], f: Unit -> Option[a] \ ef): Iterator[a, ef, r] =
+        let f1 = () -> match f() {
+            case Some(a) => Ans(a)
+            case None    => Done
+        };
+        let iterF = () -> unsafe_cast f1()  as _ \ { ef, r };
```
magnus-madsen, line 49: I assume this cast is here because otherwise the type would be: `Iterator[a, ef, Pure]`. Is that right?

If so, my idea from earlier about doing a no-op effect in `rc` might be useful.

We could even consider adding e.g.:

```
def touch(_rc: Region[r]): Unit \ r = unsafe_cast () as Unit \ r
```

to the library-- which could then be used in `f1`.

magnus-madsen: Let me know if I should give more details.

main/src/library/Iterator.flix
```diff
@@ -14,179 +14,158 @@
...
     ///
     /// Returns an empty iterator.
     ///
-    pub def new(_: Region[r]): Iterator[a, r] \ { Read(r), Write(r) } =
-        let done = () -> unsafe_cast true as _ \ Read(r);
-        let next = () -> unsafe_cast bug!("Empty Iterator!") as _ \ Read(r);
-        unsafe_cast Iterator(done, next) as _ \ { Read(r), Write(r) }
+    pub def empty(rc: Region[r]): Iterator[a, ef, r] =
+        let f = () -> unsafe_cast Done as _ \ ef;
+        Iterator(rc, f)
```
magnus-madsen, line 57: Definitely same issue as above.

main/src/library/Iterator.flix
```diff
@@ -14,179 +14,158 @@
...
-                bug!("Empty Iterator!")
-            }
-            else {
-                d := true;
-                x
-            }
-        };
-        Iterator(done, next)
+    pub def singleton(rc: Region[r], x: a): Iterator[a, ef, r] \ Write(r) =
+        repeat(rc, 1, x)
```
magnus-madsen, line 63: Hah! 👍

main/src/library/Iterator.flix
```diff
@@ -14,179 +14,158 @@
...
-        let Iterator(done, _) = iter;
-        done()
+    pub def next(iter: Iterator[a, ef, r]): Option[a] \ { ef, Read(r) } =
+        let Iterator(_, iterF) = iter;
+        def loop() = match iterF() {
+            case Ans(a) => Some(a)
+            case Skip   => loop()
+            case Done   => None
+        };
+        loop()
```
magnus-madsen, line 77: I like it

main/src/library/Iterator.flix
```diff
@@ -14,179 +14,158 @@
...
+        if (e <= b)
+            empty(rc)
+        else {
+            let f = () -> match (deref pos) {
+                case i if i < e => pos := i+1; Ans(i)
+                case _          => Done
             };
-        Iterator(done, next)
+            let iterF = () -> unsafe_cast f() as _ \ { ef, r };
+            Iterator(rc, iterF)
```
magnus-madsen, line 94: Here the problem is seemingly different-- because you want `ef` in the signature, but of course there is no user-defined effect (yet!).

magnus-madsen: Thus the cast is required here, but once we have a proper safe cast thing that can be used instead.

main/src/library/Iterator.flix
```diff
@@ -14,179 +14,158 @@
...
+    pub def sum(iter: Iterator[Int32, ef, r]): Int32 \ { ef, Read(r) } =
         foldLeft((acc, x) -> acc + x, 0, iter)
 
     ///
     /// Returns the sum of all elements in the iterator `iter` according to the function `f`.
     ///
     /// Consumes the entire iterator.
     ///
-    pub def sumWith(f: a -> Int32 \ ef, iter: Iterator[a, r]): Int32 \ { ef, Read(r) } =
+    pub def sumWith(f: a -> Int32 \ ef1, iter: Iterator[a, ef, r]): Int32 \ { ef, ef1, Read(r) } =
```
magnus-madsen, line 129: I think in all other code we have been using `ef1` and `ef2` in these situations (instead of `ef` and `ef1`), but maybe you want to keep the associated of `ef` with the iterator? I don't have any opinion, just pointing it out.

main/src/library/Iterator.flix
```diff
@@ -14,179 +14,158 @@
...
     /// Returns the product of all elements in the iterator `iter`.
     ///
     /// Consumes the entire iterator.
     ///
-    pub def product(iter: Iterator[Int32, r]): Int32 \ Read(r) =
-        if (isEmpty(iter))
-            1
-        else
-            foldLeft((acc, x) -> acc * x, 1, iter)
+    pub def product(iter: Iterator[Int32, ef, r]): Int32 \ { ef, Read(r) } = match next(iter) {
```
magnus-madsen, line 137: The old code also had a match and a fold, but I guess a fold would be enough?

main/src/library/Iterator.flix
```diff
@@ -14,179 +14,158 @@
...
     /// Returns the product of all elements in the iterator `iter` according to the function `f`.
     ///
     /// Consumes the entire iterator.
     ///
-    pub def productWith(f: a -> Int32 \ ef, iter: Iterator[a, r]): Int32 \ { ef, Read(r) } =
-        if (isEmpty(iter))
-            1
-        else
-            foldLeft((acc, x) -> acc * f(x), 1, iter)
+    pub def productWith(f: a -> Int32 \ ef1, iter: Iterator[a, ef, r]): Int32 \ { ef, ef1, Read(r) } = match next(iter) {
```
magnus-madsen, line 147: Same comment about the match being redundant.

main/src/library/Iterator.flix
```diff
@@ -14,179 +14,158 @@
...
     ///
     /// Consumes the entire iterator.
     ///
-    pub def toArray(r1: Region[r1], iter: Iterator[a, r2]): Array[a, r1] \ { Read(r2), Write(r1) } = region r0 {
-        let m = new MutList(r0);
+    pub def toArray(rc: Region[r1], iter: Iterator[a, ef, r2]): Array[a, r1] \ { ef, Read(r2), Write(r1) } = region rc1 {
+        let m = new MutList(rc1);
         forEach(a -> MutList.push!(a, m), iter);
-        MutList.toArray(r1, m)
+        MutList.toArray(rc, m)
```
magnus-madsen, line 160: Very nice, should in theory be very efficient!

main/src/library/Iterator.flix
```diff
@@ -195,108 +174,84 @@ namespace Iterator {
...
 
     ///
     /// Returns the contents of `iter` as a map.
     ///
     /// Consumes the entire iterator.
     ///
-    pub def toMap(iter: Iterator[(a, b), r]): Map[a, b] \ Read(r) with Order[a] =
-        foldLeft((acc, ab) -> Map.insert(fst(ab), snd(ab), acc), Map.empty(), iter)
+    pub def toMap(iter: Iterator[(k, v), ef, r]): Map[k, v] \ { ef, Read(r) } with Order[k] =
+        foldLeft((acc, a) -> {let (k, v) = a; Map.insert(k, v, acc)}, Map.empty(), iter)
```
magnus-madsen, line 186: I believe a match lambda would work here.

main/src/library/Iterator.flix
```diff
@@ -195,108 +174,84 @@ namespace Iterator {
...
-    pub def toNel(iter: Iterator[a, r]): Option[Nel[a]] \ Read(r) = match toList(iter) {
-        case Nil     => None
-        case x :: xs => Some(Nel.Nel(x, xs))
-    }
+    pub def toNel(iter: Iterator[a, ef, r]): Option[Nel[a]] \ { ef, Read(r) } =
+        let swapCons = (a, acc) -> match acc {
+            case Nel(x, xs) => Nel(x, a :: xs)
+        };
+        match next(iter) {
+            case Some(a) => foldRight((x, acc) -> swapCons(x, acc), Nel.singleton(a), iter) |> Some
```
magnus-madsen, line 198: I think I understand, but is there any reason not to do:

```
match toList(iter) {
  case Nil => None
  case x :: xs => Nel(x, xs)
}
```

?

magnus-madsen: In fact, `... = toList(iter) |> List.toNel`.

main/src/library/Iterator.flix
```diff
@@ -195,108 +174,84 @@ namespace Iterator {
...
     ///
     /// Consumes the entire iterator.
     ///
-    pub def toNec(iter: Iterator[a, r]): Option[Nec[a]] \ Read(r) =
-        let f = (x, acc) -> match acc {
-            case None      => Some(Nec.singleton(x))
-            case Some(nec) => Some(Nec.cons(x, nec))
-        };
-        foldRight(f, None, iter)
+    pub def toNec(iter: Iterator[a, ef, r]): Option[Nec[a]] \ { ef, Read(r) } = match next(iter) {
```
magnus-madsen, line 207: What's the structure of the chain? I am trying to visualize it. I think it should like a list right? Which is probably exactly what you want?

magnus-madsen: i.e. completely right leaning.

main/src/library/Iterator.flix
```diff
@@ -195,108 +174,84 @@ namespace Iterator {
...
     ///
-    pub def forEachWithIndex(f: (Int32, a) -> Unit \ ef, iter: Iterator[a, r]): Unit \ { ef, Read(r) } = region r {
-        let ix = ref 0 @ r;
-        forEach(x -> {let i = deref ix; f(i, x); ix := i + 1}, iter)
-    }
+    pub def forEachWithIndex(f: (Int32, a) -> Unit \ ef1, iter: Iterator[a, ef, r]): Unit \ { ef, ef1, Read(r) } =
+        let Iterator(_, iterF) = iter;
+        def loop(i) = match iterF() {
+            case Skip   => loop(i)
+            case Ans(a) => f(i, a); loop(i+1)
```
magnus-madsen, line 251: Nifty.

main/src/library/Iterator.flix
```diff
@@ -468,10 +402,15 @@ namespace Iterator {
...
-        case Iterator(done, _) if done() => s
-        case Iterator(_, next)           => foldLeft(f, f(s, next()), iter)
-    }
+    pub def foldLeft(f: (b, a) -> b \ ef1, s: b, iter: Iterator[a, ef, r]): b \ { ef, ef1, Read(r) } =
+        let Iterator(_, iterF) = iter;
+        def loop(acc) = match iterF() {
+            case Skip   => loop(acc)
+            case Ans(a) => loop(f(acc, a))
+            case Done   => acc
+
```
magnus-madsen, line 411: extra linebreak.

main/src/library/Iterator.flix
```diff
@@ -500,16 +437,19 @@ namespace Iterator {
...
-            case Iterator(_, next)           => f(next(), _ -> foldRightWithCont(f, z, iter))
-        }
+    pub def foldRightWithCont(f: (a, Unit -> b \ { ef, Read(r) }) -> b \ ef, z: b, iter: Iterator[a, ef, r]): b \ { ef, Read(r) } =
+        let Iterator(_, iterF) = iter;
+        def loop() = match iterF() {
+            case Skip   => loop()
+            case Ans(a) => f(a, _ -> loop())
+            case Done   => z
+        };
+        loop()
```
magnus-madsen, line 447: Its pretty nice that `Iterator` can support `foldRightWithCont` with the right complexity. 

You would expect a corresponds between iterators and foldRightWithCont and here we see that in just a few lines it can be realized.

main/src/library/Iterator.flix
```diff
@@ -589,199 +517,169 @@ namespace Iterator {
...
-            let Iterator(done, next) = iter;
-            let i = ref 0 @ r;
-            let done1 = () -> done() or (deref i) >= n;
-            let next1 = () -> {
-                let x = next();
-                i := deref i + 1;
-                x
-            };
-            Iterator(done1, next1)
+    /// SPT TODO - check effect of `iter`
```
magnus-madsen, line 520: Something with the effect?

main/src/library/Iterator.flix
```diff
@@ -589,199 +517,169 @@ namespace Iterator {
...
     ///
-    pub def dropWhile(f: a -> Bool \ ef, iter: Iterator[a, r]): List[a] \ { ef, Read(r) } =
-        let Iterator(done, next) = iter;
-        match forward(x -> not f(x), done, next) {
-            case None    => Nil
-            case Some(i) => i :: toList(iter)
-        }
+    /// The original iterator `iter` should *not* be reused.
+    ///
+    pub def dropWhile(f: a -> Bool \ ef1, iter: Iterator[a, ef, r]): Iterator[a, r and ef and ef1, r] \ Write(r) =
```
magnus-madsen, line 562: Could `Skip` not be used here? if `f(a)` then skip?

main/src/library/Iterator.flix
```diff
@@ -589,199 +517,169 @@ namespace Iterator {
...
-        let cursor = ref None @ Scoped.regionOf(iter);
-        let done1 = () -> match deref cursor {
-            case None     => match flatMapForward(f, done, next) {
-                case None     => true
-                case Some(it) =>
-                    cursor := Some(it);
-                    false
+    /// Currently `f` has to generate an iterator with region `r`.
+    ///
+    pub def flatMap(f: a -> Iterator[b, ef1, r] \ ef2, ma: Iterator[a, ef, r]): Iterator[b, ef and ef1 and ef2, r] \ Write(r) =
```
magnus-madsen, line 585: I did not read this one yet. I remember trying for a simple implementation of `flatMap` with `append`, but I never got it to work because of the types.

main/src/library/Iterator.flix
```diff
@@ -589,199 +517,169 @@ namespace Iterator {
...
-                if (deref isReversed)
-                    ()
-                else {
-                    xs := List.reverse(deref xs);
-                    isReversed := true
-                };
-                true
+    /// The original iterators `sep` and `iter` should not be reused.
+    ///
+    pub def intercalate(sep: Iterator[a, ef2, r], iter: Iterator[Iterator[a, ef1, r], ef1, r]): Iterator[a, ef1 and ef2 and r, r] \ Write(r) =
```
magnus-madsen, line 640: What's the meaning of this? Will `sep` not be consumed after it has been intercalated once?

main/src/library/Iterator.flix
```diff
@@ -811,96 +708,67 @@ namespace Iterator {
...
     ///
-    def filterMapForward(f: a -> Option[b] \ ef, done: Done[r], next: Next[a, r]): Option[b] \ { ef, Read(r) } =
-        if (done())
-            None
-        else
-            match f(next()) {
-                case None    => filterMapForward(f, done, next)
-                case Some(b) => Some(b)
-            }
+    pub def toResultList(iter: Iterator[Result[e, a], ef, r]): Result[e, List[a]] \ { ef, Read(r) } =
```
magnus-madsen, line 732: Do we have these functions elsewhere? I guess its useful.

magnus-madsen:
Very nice. Sorry for the flood of comments. Its was nice reading through this. I think I learned a few things.

I think this is almost ready to be merged, just a few things to tidy up.

As for the unsafe_casts, I think they are fine/required. We may be able to remove some of them with a few tricks.

stephentetley:
Thanks Magnus, I'll work through your comments tomorrow.

main/src/library/Foldable.flix
```diff
@@ -219,21 +219,9 @@ pub class Foldable[t : Type -> Type] {
...
-                let j = deref i;
-                let x = arr[j];
-                i := j + 1;
-                Some(x)
-            } else {
-                None
-            }
-        };
-        Iterator.iterate(r, next)
+    pub def iterator(rc: Region[r], t: t[a]): Iterator[a, r, r] \ Write(r) =
```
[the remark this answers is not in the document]
stephentetley: Yes - updated to use `foldRight`.

main/src/library/Iterator.flix
stephentetley: I think `toResultList` was something Jakob must have needed. I haven't seen it else where.

For error handling in general we might end up needing another Iterator module where the `Step` date type includes an `Error` constructor.

main/src/library/Iterator.flix
stephentetley: `flatMap` is still a bit of a problem. 

I've now written it using the old version of `empty` (called `polymorphiEmpty` and defined internally as a helper function), it is tough to get to compile because it stores an iterator in a ref cell. I tried to write it casting the type of the stored iterator but it caused runtime crashes in the tests.

In the longer term I think the type signature of `flatMap` may be too complicated. I'm not sure the mapped function `f` needs an "ambient" effect - i.e. `ef2` floating at the right-hand side with the slash. The type for Iterator already contains an effect `ef1` so I don't think we lose any power using that - we will need something lift further effects into an Iterator which we don't have directly at the moment (we can use `map` currently which you have suggested to help eliminate casting).

~~~

pub def flatMap(f: a -> Iterator[b, ef1, r], ma: Iterator[a, ef, r]): Iterator[b, r and ef and ef1, r] \ Write(r) =

~~~

main/src/library/Iterator.flix
stephentetley: I didn't need `touch` in the end though it might still be useful.

I had in mind that one or other of these might be useful but didn't explore them at this time:

~~~

liftRegion(rc: Region[r1], iter: Iterator[a, ef, r]): Iterator[a, ef and r1, r]

liftRegion(rc: Region[r1], f: Unit -> Iterator[a, ef, r] \ r1): Iterator[a, ef and r1, r]

~~~

main/src/library/Iterator.flix
stephentetley: Have done.

main/src/library/Iterator.flix
stephentetley: Yes - effectful Iterators seem to form a (graded) monad - the set of effects possibly expanding as operations are chained together with `flatMap`.

I've changed the constructors `empty`, `singleton`, `range` and `repeat` to duplicate the region as the effect rather than them having a polymorphic effect through casting. I'm not 100% sure this is the right thing to do but it does make them simpler at the moment.

main/src/library/Iterator.flix
stephentetley: `empty` now casts to the Region `r`.

I've just noticed that `iterate` has become the odd one out of the "constructor" functions:

~~~

pub def iterate(rc: Region[r], f: Unit -> Option[a] \ ef): Iterator[a, ef, r]

~~~

For consistency it would have the following signature but that would contradict the type of `f` which doesn't necessarily access the region `rc` :

~~~

pub def iterate(rc: Region[r], f: Unit -> Option[a] \ ef): Iterator[a, r and ef, r]

~~~

main/src/library/Iterator.flix
stephentetley: Now updated to:

~~~

pub def singleton(rc: Region[r], x: a): Iterator[a, r, r] \ Write(r)

~~~

main/src/library/Iterator.flix
stephentetley: Yes this has worked exactly as hoped - the effect `ef` and region `r` that are encapsulated in the type of the Iterator parameter float to the right of of the type signature. No casting required.

main/src/library/Iterator.flix
stephentetley: I think this was either `range` or `repeat` that now return an iterator with "region access" as the effect rather than a polymorphic effect. With "region access" for the effect they don't need casting.

main/src/library/Iterator.flix
stephentetley: I missed this - I'll take another look at it.

main/src/library/Iterator.flix
stephentetley: I think this is cleaned up now if it was one of the `sum` / `product` functions.

main/src/library/Iterator.flix
stephentetley: Again, I think this is cleaned up now if it was one of the `sum` / `product` functions.

main/src/library/Iterator.flix
stephentetley: Thanks!

magnus-madsen:
Let me know when I should have another look.

main/src/library/Iterator.flix
stephentetley: I missed this - I'll take another look at it.

main/src/library/Iterator.flix
stephentetley: I missed this - I will add the simplification.

main/src/library/Iterator.flix
stephentetley: I think you could do `toNec` as a `foldRight` but then I think you would need to make the accumulator an `Option` and have to test it at each step unless `foldRightWithCont` helps...

main/src/library/Iterator.flix
stephentetley: Thanks - no harm in using accumulators even if you have a region accessible.

main/src/library/Iterator.flix
stephentetley: I missed this - I will fix.

main/src/library/Iterator.flix
stephentetley: I think that was all Jakob's work and I just removed the `done` thunk.  I still haven't really got an intuition about `foldRightWithCont` as I've not used it much.

main/src/library/Iterator.flix
stephentetley: I think this was `drop` which I changed the signature to "region access" is in the effect set of the return type.

main/src/library/Iterator.flix
stephentetley: `sep` is consumed (turned into a List) the first time `step` is run. As a list it is stored in the ref cell `sepList` so a new "sep" iterator can be rebuilt at each gap between the iterator-of-iterators.

stephentetley:
> 

I've a few corrections that I missed to done, should have them done before tomorrow.

main/src/library/String.flix
```diff
@@ -1212,21 +1212,25 @@ namespace String {
     ///
     /// Returns an iterator over `s`.
     ///
-    pub def iterator(r: Region[r], s: String): Iterator[Char, r] \ Write(r) =
-        let len = length(s);
+    pub def iterator(r: Region[r], s: String): Iterator[Char, r, r] \ Write(r) =
+        let len = String.length(s);
```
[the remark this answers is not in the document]
stephentetley: Updated - `range` works nicely as an index supplier. At first I thought we would need `mapWithIndex` so I added that, but didn't use it in the end.

main/src/library/MutSet.flix
stephentetley: Updated and fixed.

main/src/library/Foldable.flix
stephentetley: Updated - removed the use of an intermediate array.

main/src/library/Iterator.flix
stephentetley: I've changed to use ef1 and ef2 - will be in the next commit.

main/src/library/Iterator.flix
stephentetley: Changed in next commit.

main/src/library/Iterator.flix
stephentetley: Will be changed in next commit.

stephentetley:
The latest commit fixes `toNel` and `toMap` issues and changes effects type parameter names that were "ef and ef1" to "ef1 and ef2".

main/src/library/Iterator.flix
stephentetley: Now fixed

main/src/library/Iterator.flix
stephentetley: Now use match lambda.

main/src/library/Foldable.flix
```diff
@@ -219,23 +219,13 @@ pub class Foldable[t : Type -> Type] {
...
-        let done = () -> (deref i) >= len;
-        let next = () -> {
-            let j = deref i;
-            let x = arr[j];
-            i := j + 1;
-            x
-        };
-        Iterator.Iterator(done, next)
+    pub def iterator(rc: Region[r], t: t[a]): Iterator[a, r, r] \ Write(r) =
+        Foldable.foldRight((x, iter) -> Iterator.cons(x, iter), Iterator.empty(rc), t)
```
magnus-madsen, line 223: I am a bit concerned that if use `foldRight` we still have to traverse the entire collection. If we use `foldRightWithCont` that should be unnecessary since we can call the continuation when we need. However, I don't know if this is tricky to implement. 

Perhaps it would be better to omit `iterator` from Foldable otherwise.

main/src/library/Iterator.flix
magnus-madsen: I think my point was that the match seems redundant. We can just call foldLeft directly.

main/src/library/Iterator.flix
```diff
@@ -14,179 +14,155 @@
...
     /// Returns the product of all elements in the iterator `iter` according to the function `f`.
     ///
     /// Consumes the entire iterator.
     ///
-    pub def productWith(f: a -> Int32 \ ef, iter: Iterator[a, r]): Int32 \ { ef, Read(r) } =
-        if (isEmpty(iter))
-            1
-        else
-            foldLeft((acc, x) -> acc * f(x), 1, iter)
+    pub def productWith(f: a -> Int32 \ ef2, iter: Iterator[a, ef1, r]): Int32 \ { ef1, ef2, Read(r) } = match next(iter) {
```
magnus-madsen, line 144: And here we can also just call foldLeft directly (without the match).

main/src/library/Iterator.flix
magnus-madsen: I see. But then I find the function strange/unusual. 

Would it not be better if `sep` was foldable? Then you can iterate through it as often as you like.

Right now, if you have a list and an iterator, you have to get an iterator for the list to call intercalate.

main/src/library/Iterator.flix
```diff
@@ -589,199 +524,184 @@ namespace Iterator {
...
-                case Some(i) =>
-                    if (doneB())
-                        cursor := Some(append(i, new Iterator(r1))) // Has Boolean formula r2 since it is the last iterator of `iterB`.
-                    else                                                     // Has to be appended with an empty iterator to satisfy formula `r1 and formula of i` in `mkCursor`
-                        mkCursor(i);                                         // Same reason for appending with empty iterator below
-                    false
+        flatMap(step, iter)
+
+    ///
+    /// Helper for append.
```
magnus-madsen, line 683: I don't think append uses this?

main/src/library/Iterator.flix
magnus-madsen: I would delete it (if it is unused). It seems very irregular. 

If we want such functions we could add them back in a systematic way.

magnus-madsen (commented):
Looks good. I added just a few more comments that I think we should resolve and then I can push the button :)

main/src/library/Foldable.flix
stephentetley: The simplistic version using `foldRightWithCont` stack overflows for large foldables whereas the current `foldRight` version is stack safe.

~~~

    pub def iterator2(rc: Region[r], t: t[a]): Iterator[a, r, r] \ Write(r) with Foldable[t] =
        Foldable.foldRightWithCont((x, k) -> Iterator.cons(x, k()), Iterator.empty(rc), t)

~~~

I can try to write a version with `foldRightWithCont` that builds a cont (to be applied at the end to make an Iterator) rather than one that makes an iterator inside the fold.

main/src/library/Iterator.flix
stephentetley: `match` removed from both.

main/src/library/Iterator.flix
stephentetley: Used by `intercalate` not `append`` - fixed

main/src/library/Foldable.flix
magnus-madsen: How about we remove the `iterator` function for now then and revisit?

magnus-madsen:
I can merge now?

stephentetley:
I'll just work a bit more on Foldable.iterator.

magnus-madsen:
OK; otherwise we can remove it and leave it for later.

stephentetley:
Arrgh, I couldn't manage a good stack safe implementation with `foldRightWithCont`.

Is it so bad that the default `Foldable.iterator` is does a traversal? I think all cases in the stdlib overwrite it anyway.

magnus-madsen:
> Arrgh, I couldn't manage a good stack safe implementation with `foldRightWithCont`.
> 
> Is it so bad that the default `Foldable.iterator` is does a traversal? I think all cases in the stdlib overwrite it with an efficient traversal anyway.

Its okay, its just that you would expect an iterator to be lazy right?

I don't suppose anything else in Foldable can be used?

stephentetley:
> Its okay, its just that you would expect an iterator to be lazy right?

I tend to think of Iterators as steppable / streamable rather than lazy. I get myself bit confused if I think of them as lazy on a strict data structure like List or Chain. 

For DelayList the default implementation would be disappointing as I think it would force the whole list. But as it's overwritten that wouldn't happen.

> I don't suppose anything else in Foldable can be used?

I don't think so - I think anything except `foldRightWithCont` has to do a full traversal.

stephentetley:
My preference is we keep the "not so great" default iterator in Foldable, but I'm happy to just have it in the `Iterable` class instead. I think this would mean moving `enumerator` into `Iterable` though.

magnus-madsen:
I think we should remove it. We do have `Iterable` so better to use that anyway. Lets guide the programmer to the right tool.

If someone needs iterator on Foldable we can revisit.

magnus-madsen:
> Okay - I'll make an update at some point this evening but I've got to eat first so it could be an hour or two.

No worries :) Thanks for all your work!

stephentetley:
I've removed `iterator` from `Foldable` and also `Reducible` and added `enumerator` to `Iterable`.

There were a couple of data structures missing `Iterable` so I've added those and also merged to head.

magnus-madsen:
Thanks! -- Great to have this in.

PR #5332, opened 2023-01-19
feat: allow non-expr restrictable enums

mlutze:
related ot #5229


main/src/ca/uwaterloo/flix/language/ast/Scheme.scala
```diff
@@ -89,14 +90,14 @@ object Scheme {
     * Returns `true` if the given schemes are equivalent.
     */
   // TODO can optimize?
-  def equal(sc1: Scheme, sc2: Scheme, classEnv: Map[Symbol.ClassSym, Ast.ClassContext])(implicit flix: Flix): Boolean = {
+  def equal(sc1: Scheme, sc2: Scheme, classEnv: Map[Symbol.ClassSym, Ast.ClassContext])(implicit univ: Ast.Multiverse, flix: Flix): Boolean = {
```
magnus-madsen, line 93: Why not just pass `root`?

magnus-madsen: I would much prefer if we passed `root` as much as possible.

mlutze: I've now changed to do that inside of phases, but places like `Scheme` and `unification` are phase-ambivalent so we can't pass root there.

PR #5333, opened 2023-01-19
test: restrictable variants Expr example

JonathanStarup:


JonathanStarup:
@magnus-madsen Heres the test, it types, and should run after #5374

examples/restrictable-expression.flix
```diff
@@ -0,0 +1,62 @@
...
+def size(e: Expr[s]): Int32 = choose e {
+    case Expr.Cst(_) => 1
+    case Expr.Var(_) => 1
+    case Expr.Not(x) => size(x) + 1
+    case Expr.And(x, y) => size(x) + size(y) + 1
+    case Expr.Or(x, y) => size(x) + size(y) + 1
+    case Expr.Xor(x, y) => size(x) + size(y) + 1
+}
+
+def eq(e1: Expr[s1], e2: Expr[s2]): Bool = ??? // nested choose
```
magnus-madsen, line 19: What this for?

examples/restrictable-expression.flix
```diff
@@ -0,0 +1,62 @@
...
+
+def simplify(e: Expr[s]): Expr[(s -- <Expr.Xor>) ++ <Expr.Not>] =
+    choose* e {
+        case Expr.Var(x) => Expr.Var(x)
+        case Expr.Cst(b) => Expr.Cst(b)
+        case Expr.Not(x) => open Expr.Not(openExpr(simplify(x)))
+        case Expr.And(x, y) => Expr.And(???, ???)
```
magnus-madsen, line 26: This depends on the other open?

JonathanStarup: choose* still has the check that the body must be a tag, so i cant use openExpr

examples/restrictable-expression.flix
JonathanStarup: Just a cool function i wanted to write, I can remove it/ write it

examples/restrictable-expression.flix
magnus-madsen: OK, but that can be lifted, yes?

JonathanStarup: We've found that I can't, but i think that'll be discussed elsewhere

examples/restrictable-expression.flix
```diff
@@ -0,0 +1,124 @@
...
+    println("input is '${toString(input)}'");
+    println("input size ${size(input)}");
+    println("input evaluates to ${eval(input)}");
+    let fls = Expr.Cst(false);
+    println("input ${if (eq(input, fls)) "is" else "is not"} equal to '${toString(fls)}'")
+}
+
+
+/// Structural Equality
+def eq(e1: Expr[s1], e2: Expr[s2]): Bool = choose e1 {
```
magnus-madsen, line 66: What's the point of all this? I would remove it.

magnus-madsen:
I will merge this, but then we should iterate on it.

examples/restrictable-expression.flix
JonathanStarup: I think its important to have a index polymorphic equality function so I added it

PR #5337, opened 2023-01-19
feat: primitive set formula

JonathanStarup:
Related to #5229 

Add simple SetFormula objects based on BoolFormula. Vars are simple integers and constants are sets of integers as well using the same mechanism. This should make thing easy in the set formula world and keep the translation simple too.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/unification/SetFormula.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/unification/SetFormula.scala
@@ -65,7 +65,57 @@ object SetFormula {
 
   case class Or(f1: SetFormula, f2: SetFormula) extends SetFormula
 
+  // Smart Constructors
+
   val Empty: SetFormula = Cst(Set.empty)
+
+  def mkCst(s: Set[Int])(implicit universe: Set[Int]): SetFormula = {
+    if (s == universe) All else Cst(s)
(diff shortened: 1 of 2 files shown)


JonathanStarup:
@magnus-madsen is this what you had in mind?

main/src/ca/uwaterloo/flix/language/phase/unification/SetFormula.scala
```diff
@@ -0,0 +1,154 @@
...
+import scala.collection.immutable.SortedSet
+
+sealed trait SetFormula {
+
+  /**
+    * Returns the free variables in `this` expression.
+    */
+  final def freeVars: SortedSet[Int] = this match {
+    case SetFormula.All => SortedSet.empty
+    case SetFormula.Constant(_) => SortedSet.empty
```
magnus-madsen, line 16: Constant -> Cst (following similar naming).

magnus-madsen:
Yes, plus some smart constructors.

magnus-madsen:
We can refine those smart constructors later.

JonathanStarup:
I added some, Should we stick with boolean language? Change it to set terminology? or maybe both?

magnus-madsen:
I merged it, but I don't think we need the Alg part-- at least for now.

PR #5348, opened 2023-01-20
feat: allow parsing of custom kinds

mlutze:
related to #5229. untested


magnus-madsen:
What does this do?

mlutze:
We move kind resolution from the weeder to the Resolver.

This means that we'll be able to express kinds other than the built-in ones. For example, we can do this:

```scala
restrictable enum E[s]

pub def blah[s: E](a: E[s]): E[s] = ???
// type parameter `s` has kind `E`, AKA a set the cases of `E`
```

It's the easier half of allowing us to express set types.

PR #5353, opened 2023-01-22
feat: Implement ScopeExit

paulbutcher:
This is a follow-on to #5319 and the next step of #5225. Implements functionality to call functions passed to `$SCOPE_EXIT$` on region exit.


magnus-madsen:
A few things:

- Do we need to handle exceptions thrown by close?
- I understand the timeout test, but I don't like it :)
- Do we want to clear the list/remove its elements? Technically it should never be needed, I guess.

magnus-madsen:
Do we clear correctly, even if the thread containing the region crashes?

paulbutcher:
With the exception of the comment about the timeout (I also don't like it, but I can't think of another way to write the test: let me know if I've missed something) all your comments boil down to "what should Flix do about exceptions".

I think we are heading in a dangerous direction which will cause us many problems in the future. Even though I wrote it, I am increasingly of the opinion that our fix to https://github.com/flix/flix/issues/5123 is the wrong approach and has opened a Pandora's Box which is going to be a thorn in our side forever if we don't close it.

The situation today is that:

1. Flix's idiomatic approach to error handling is monadic (i.e. through `Option`, `Result`, `Validation`, etc.).
2. Flix doesn't really have an exception model: there is no way to throw an exception from Flix (apart from writing things like `String.concat("", unsafe_cast null as String)`) and no clearly thought out try/catch semantics (yes, we have `try` and `catch` but they're really only there for interop).
3. In particular, one of Flix's headline features is its effect model, and the effect model is currently entirely silent on the subject of exceptions. But throwing an exception is a much more powerful effect than writing to a file!
4. The fact that `1 / 0` doesn't throw an exception is a very powerful statement: it says that Flix goes out of its way to avoid exceptions, a very different position from just about every other language.

Through our fix to #5123, we've started to step away from this position: we now re-throw an exception that happens in a spawned thread in whichever thread contains the region with which the spawn is associated.

I don't think we've appreciated the magnitude of this decision: we are, de-facto, making exceptions first-class citizens within Flix. They are going to interact in strange ways with all kinds of things, the functionality in this PR being only one such.

The more I think about this, the more I think that a more sensible approach is to say that Flix's behaviour in the presence of exceptions is very simple and constrained:

1. Java functions called through interop which might throw should be wrapped with `try`/`catch` and the exception translated into something idiomatic such as `Result`.
2. If an unhandled exception escapes, the only thing that Flix guarantees is that it will be reported and the JVM will exit.

The only alternative to this is to bite the bullet and embrace exceptions fully, but that would be a major change to the semantics and philosophy of the language.

JonathanStarup:
> 1. Java functions called through interop which might throw should be wrapped with try/catch and the exception translated into something idiomatic such as Result.
> 2. If an unhandled exception escapes, the only thing that Flix guarantees is that it will be reported and the JVM will exit.

I agree with this. You mention 

> Through our fix to https://github.com/flix/flix/issues/5123, we've started to step away from this position: we now re-throw an exception that happens in a spawned thread in whichever thread contains the region with which the spawn is associated.

Isn't this just a method in which to enforce the behavior that a flix program really crashes on an exception?

JonathanStarup:
I think no matter the version of flix, StackOverflow exception will at least exist, so things should probably fail as gracefully as possible. what do other languages ensure for scoped resources?

paulbutcher:
> Isn't this just a method in which to enforce the behavior that a flix program really crashes on an exception?

No, for two reasons:

1. Firstly, it doesn't guarantee that a Flix program will crash on an exception. There are three ways in which it can fail:
    1. If the thread containing the region is blocked uninterruptibly
    2. If the thread containing the region never relinquishes the CPU
    3. If the thread containing the region catches the exception
        * This last might seem like a nitpick, but remember that it might be *expecting* to have to catch an exception (perhaps it's calling a Java function which might legitimately throw). Now that exception might come from two different places: the place that it was expecting the exception to come from or some spawned thread.
2. It does a lot more than just force the program to crash. If that's all we wanted to do, we could achieve it very easily by installing an `UncaughtExceptionHandler` and calling `System.exit`, but we're doing much more.

paulbutcher:
> I think no matter the version of flix, StackOverflow exception will at least exist, so things should probably fail as gracefully as possible. what do other languages ensure for scoped resources?

Java exceptions broadly divide into two different categories: Exceptions (also sometimes called checked exceptions) and Errors (also sometimes called unchecked exceptions). The general rule is that, under normal circumstances, you shouldn't catch errors (they indicate that something has gone wrong in the basic assumptions made by the program, and all bets are off: i.e. you're very unlikely to be able to recover from an error and the best you can do is exit). This is what Flix should do too, I think.

I don't think that scoped resources are relevant here: there are no resources that you could hold onto which wouldn't be released by exiting the JVM.

magnus-madsen:
Firstly, I share all your concerns (and more).

I think the approach in Rust is instructive:

- Rust has `Option` and `Result` and they are generally used for ordinary programming with error handling.
- Rust also has abort, which is used in exceptional circumstances, when execution is no longer possible.

I think both are necessary-- because there are a few errors you cannot recover from easily -- e.g. `ArrayIndexOutOfBounds` or `StackOverflow` and more.

As for Flix, I think the message should be the same as above. With the following twists:

- Flix has `Option` and `Result` and they should be used for ordinary programming.
- Flix has support for catching exceptions which should be used for interoperability with Java and as close to the Java call site as possible. 
- Flix code may, in exceptional circumstances, throw exceptions similar to abort in Rust.

Now, as for structured concurrency, Flix tries to:

- Prevent errors when possible (strong static type system, effect system, and lots of checks like unused variables).
- When not possible to prevent an error, Flix should surface the error as much as possible, so the system does not continue in a broken state. Here I think it is very sensible that exceptions are re-thrown in their spawning thread. In fact, this design choices seems so natural that the only reason it is not default in Java is probably because structured concurrency was not a thing then (at least in the language).

Separately there are two minor questions:

- Should we __force__ programmers to surround Java calls with try-catch? (Ditto for `null` checks)
- How should one propagate errors across threads in a _functional style_ with `Result` and friends?

magnus-madsen:
I think `ArrayIndexOutOfBounds` is instructive; is almost impossible to deal with. You don't want to have `Array.get` return an `Option` because this destroys performance. We cannot return a random element (because that is insane and also the array could be empty). I believe, no _practical_ static type system can deal with _all cases_ where `ArrayIndexOutOfBounds` may occur. Thus it becomes necessary to somehow be able to abort execution. That leaves `System.exit` or something slightly gentler.

magnus-madsen:
(Ill post a comment later on how we can design our way to a good place there will be OK)

paulbutcher:
> How should one propagate errors across threads in a functional style with Result and friends?

Given that right now the only mechanism available to a thread to report any kind of result (successful or otherwise) is to send a value on a channel, the obvious answer is that that value should be a `Result`. Are there circumstances where that wouldn't be sufficient?

paulbutcher:
It's good to know that we're broadly aligned @magnus-madsen. In particular I strongly agree with:

> When not possible to prevent an error, Flix should surface the error as much as possible, so the system does not continue in a broken state.

Where I think we disagree is:

> Here I think it is very sensible that exceptions are re-thrown in their spawning thread

I agree that this *would* be sensible *if* the JVM supported it well. But (as things currently stand at least) it doesn't. By implementing this re-throwing we have decided that we will *not* surface the error as much as possible: specifically we will fail to do so in the instances I enumerated [above](https://github.com/flix/flix/pull/5353#issuecomment-1400631099).

I guarantee that with things as they currently stand, some poor developer somewhere will find themselves trying to debug some mysterious failure in which their program either unexplainably hangs or throws some weird exception from some piece of code where that exception seems to make no sense. They will eventually, after hours or days of frustration, discover that the actual error was an exception in a spawned thread which got re-thrown elsewhere (possibly multiple times) and either swallowed or transmuted along the way.

I would prefer to be kind to that future developer and immediately report the error where it happens and then exit the JVM to avoid confounding them.

paulbutcher:
BTW, this discussion is interesting and important, but doesn't necessarily have any bearing on this PR specifically? Can we merge this to allow me to continue to make progress with the non-controversial aspects of the implementation?

main/test/flix/Test.Exp.Regions.flix
```diff
@@ -1384,11 +1384,20 @@ namespace Test/Exp/Regions {
...
-         let x  = ref false @ r1;
-         region r2 {
-             spawn { Thread.sleep(Time/Duration.fromSeconds(1)); x := true } @ r2
-         };
-         deref x
-     }
+    def testRegionExit01(): Bool \ IO = region r1 {
+        let x  = ref false @ r1;
+        region r2 {
+            spawn { Thread.sleep(Time/Duration.fromSeconds(1)); x := true } @ r2
```
magnus-madsen, line 1390: The sleep 1second test is not OK with me. Either reduce it to a much shorter time or drop the test.

magnus-madsen: You could maybe use a channel as sync. point.

paulbutcher: For reference this test isn't actually anything to do with this PR: it's only showing up as a change because I took the opportunity to fix some broken formatting.

As for how to fix this, the whole point of the test is to ensure that scope exit calls `join`. I don't think I can add any additional synchronisation without missing the point of the test. So I'll reduce the delay. There needs to be *some* delay though or the test could pass just because the thread executes "quickly".

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -336,8 +336,19 @@ object GenExpression {
       visitor.visitLabel(afterFinally)
 
     case Expression.ScopeExit(exp1, exp2, tpe, loc) =>
-      //!TODO: For now, just emit unit
-      compileConstant(visitor, Ast.Constant.Unit, MonoType.Unit, loc)
+      // Compile the expression representing the region
+      compileExpression(exp2, visitor, currentClass, lenv0, entryPoint)
```
JonathanStarup, line 340: I don't know whats assumed about `exp1` and `exp2` here, but its okay do run `exp2` before `exp1`? Shouldn't we adhere to left-to-right evaluation?

JonathanStarup: I think they should be evaluated exp1 then exp and use a swap to rearrange them.

JonathanStarup (commented):
Looks pretty good, just one question. Also, unrelated: the jvmcode setup we have now a wip, so if you have any feedback or complaints, feel free :)

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
magnus-madsen: We should always be left-to-right.

magnus-madsen:
I would like to ensure that if one of the close methods throws an exception we still continue and close the subsequent resource.

JonathanStarup:
> I would like to ensure that if one of the close methods throws an exception we still continue and close the subsequent resource.

What happens to the exception(s) then?

magnus-madsen:
> > I would like to ensure that if one of the close methods throws an exception we still continue and close the subsequent resource.
> 
> What happens to the exception(s) then?

Don't know.

paulbutcher:
> > > I would like to ensure that if one of the close methods throws an exception we still continue and close the subsequent resource.
> > 
> > 
> > What happens to the exception(s) then?
> 
> Don't know.

This "Don't know" is exactly the Pandora's Box I'm worried about. If we continue down this road, then we're de-facto promoting exceptions to first class members of Flix. But that has many implications of which we've barely scratched the surface.

There are only two way that a `close` method could possibly throw an exception:

1. It calls a Java method via interop without wrapping the method in `try/catch`
2. It does something intrinsically broken (exhausts the stack or similar)

The first is a programming error (the method should be wrapped in try/catch). The second is also a programming error. So whatever, if a `close` method throws an exception, the program is intrinsically broken.

Therefore, by the principles discussed above we should "surface the error as much as possible so that the system does not continue in a broken state".

So, if a `close` method throws, we should report the error and exit without doing anything more.

If we catch the error and try to clean up/continue all kinds of weird shit might happen (including that subsequent clean up throwing an error and masking the error which caused the problem in the first place, confusing the poor developer and making it much more difficult to debug than necessary).

I really (really!) think we're going down a dangerous road here. We're making something that *should* be very simple (report any unhandled exception and exit) into something complicated, difficult, and confusing.

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
paulbutcher: Ah, sorry, didn't realise that this was a requirement. Will fix.

magnus-madsen:
There is some more information here: https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

Notice that:

- Resources are de-allocated in reverse order of their allocation (i.e. following a stack discipline). 
- All resources are closed (which may lead to suppressed exceptions-- should a `close` throw).

magnus-madsen:
I think the higher-level questions remain, but we returning the purpose of this PR, I think the important thing is that:

- Resources registered in a region should be `closed` when the region is destroyed. 
- This should happen in the reverse order they were registered in (thus ensuring a stack discipline).
- We should introduce a `region rc with` construct (in a new PR).
- We must close _all_ resources (using finally), rethrowing the first exception and suppressing the rest. (Perhaps we can print them).

This is by no means the final solution, but it at least gets us to the place where exceptions are now propagated correctly from `close` and across threads. 

The higher-level questions of what we think of exceptions and how to even deal with the failure across threads remain.

But we can iterate on that.

magnus-madsen:
Here is a simple proposal. This is by no means the final solution, but rather the next steps I would take.

Recall that Java has:

- Checked exceptions (sub-classes of `Exception`, but not of `RuntimeException`).
  - e.g. `IOException`
- Unchecked exceptions (sub-classes of `RuntimeException`).
  - e.g. `NullPointerException`, `ArrayIndexOutOfBounds`.

What I propose is (as the next step, but not the last step): 

- An unchecked exception (e.g. `ArrayIndexOutOfBounds` immediately a `System.exit(1)`.
- A checked exception is propagated as today, i.e. down the call stack and back to the parent thread.

In the future we modify `import` such that if a method can throw a checked exception (recall that Java has this information) then you get the effect "Exception" (or whatever). This will encourage you to either: (1) Immediately handle the exception turning it into e.g. `Result` or (2) Carry the "Exception" effect around. If you call `spawn e` where `e` may throw then YOU also get the "Exception" effect, hence now you must deal with it. (Since this is exactly what the semantics does). 

Longer term, I want to study error handling in ZIO and Erlang in detail, but I think the above would be a good incremental step forwards.

paulbutcher:
I'm sorry @magnus-madsen, but that simple proposal still commits us to supporting exceptions by stealth, I think. Perhaps we should have the face to face conversation we agreed to on Gitter?

I think we agree on the following:

1. The primary error handling mechanism for Flix is `Result` and `Validation`.
1. Nevertheless, we have to support exceptions at *some* level because the underlying system will occasionally throw exceptions.
1. A correct Flix program should should convert (via `try`/`catch`) all [“checked”](https://docs.oracle.com/javase/tutorial/essential/exceptions/runtime.html) exceptions thrown by Java methods to `Result` or `Validation`

This implies that (I think that this is uncontroversial, but let me know if you disagree):

4. The only occasions in which a Flix program will throw an exception are therefore either
	* Panics such as `ArrayIndexOutOfBounds` or `StackOverflow`, or
	* Exceptions thrown by Java methods which were erroneously not converted to `Result` or `Validation`.
1. Neither of the above are recoverable; in both cases they indicate that the program contains a bug. There’s nothing that can be done to address the problem apart from changing the program to fix the bug.

This then raises the question of what should be done about these unrecoverable errors.

In our Gitter conversation you said “I would have thought such an "academic" approach would be unacceptable for a real-world app.” I *think* this may be based upon an outdated mental model of how commercial software is deployed these days (apologies if I’ve misinterpreted).

It used to be the case that web apps were typically deployed within big application servers like WebSphere, Tomcat and JBoss. These application servers were long-lived (running for months or years at a time) and served multiple applications. Nobody does this any more though.

These days people either use lightweight application servers like GlassFish or (more commonly) directly embed an HTTP server within the application via Jetty (this is what I did in [flix-http](https://github.com/paulbutcher/flix-http)), running multiple instances behind a load balancer.

So instead of a single large application server running multiple applications, each application runs multiple instances, typically across multiple servers, with a single application per JVM. These are typically short-lived (running for hours, or at most a few weeks) and are created and destroyed as the application scales up/down with load, or when new versions of the app are deployed.

All of the above means that exiting the JVM is not a big deal (in fact, it happens all the time). The better shops even run [Chaos Monkey](https://netflix.github.io/chaosmonkey/) or similar to destroy VMs (or whole server instances) at random.

Given that unhandled exceptions within Flix *always* indicate a bug (and speaking as a software engineer with several decades of commercial software development) the *only* thing I care about is that I know what that bug is. Anything that gets in the way of accurately reporting a bug is a problem.

Regarding the specifics of your proposal:

> An unchecked exception (e.g. ArrayIndexOutOfBounds immediately triggers a System.exit(1).

This is good.

> A checked exception is propagated as today, i.e. down the call stack and back to the parent thread.

I don't see what this gets us. Because (as per the discussion above) this will never happen in a correct Flix program (pure Flix can never throw a checked exception, and a checked exception should never be allowed to escape in a correctly written Flix program).

> In the future we modify import such that if a method can throw a checked exception

Agreed. But...

> or (2) Carry the "Exception" effect around.

I'm not sure I understand what this buys us? This is effectively making exceptions first-class within Flix. If that's what we want to do, then I'm OK with that I guess, but why add the complexity?

> Longer term, I want to study ZIO and Erlang in greater detail to understand how they handle failures.

I'm not sure how informative this will be:

* ZIO is written in Scala, a langauge that has first class exceptions.
* Erlang follows the "let it crash" philosophy, but that depends entirely on the ability to detect failures in and restart processes. We just don't have the same machinery available.

magnus-madsen:
We can have a chat? Do you want to DM me with a time?

I should point out that exceptions are just a special case of algebraic effects. Thus once we have AE adding the machinery I propose would come for free. (It would just be a special case of how to handle AE in general).

magnus-madsen:
1. An exception triggers (a) print stack trace and (b) system exit.
2. We call onExit in reverse order of registration (to maintain stack discipline).

paulbutcher:
@magnus-madsen I've updated this PR so that closeables are closed in reverse order of registration. I'll submit a separate PR to modify the exception behaviour.

Assuming you're OK with that, I think this PR can now be merged?

magnus-madsen:
OK, so I can merge?

paulbutcher:
> OK, so I can merge?

Yes please!

PR #5356, opened 2023-01-23
refactor: treat illegal intrinsic as soft failure

jaschdoc:
@magnus-madsen this change treats illegal intrinsics as soft failures so we can get started on the vscode end of the pipeline


main/src/ca/uwaterloo/flix/language/phase/Weeder.scala
```diff
@@ -748,7 +748,7 @@ object Weeder {
 
           case ("SCOPE_EXIT", e1 :: e2 :: Nil) => WeededAst.Expression.ScopeExit(e1, e2, loc).toSuccess
 
-          case _ => WeederError.IllegalIntrinsic(loc).toFailure
+          case _ => Validation.SoftFailure(WeededAst.Expression.Wild(loc), LazyList(IllegalIntrinsic(loc)))
```
magnus-madsen, line 751: Why does it become WeededAst.Expression.Wild? Should it not be Expression.Error?

jaschdoc: Oh yes of course! I took too much time off :D

magnus-madsen:
Seems it continues until the backend, but the Flix check should stop earlier.

magnus-madsen:
I am working on #5360

magnus-madsen:
In this call:

```
WeededAst.Expression.Error(IllegalIntrinsic(loc)).toSoftFailure
```

we are not actually putting the error into the list. So perhaps we need to change the signature of `toSoftFailure`.

magnus-madsen:
But we kind of need two versions: One that just soft fails and one that soft fails and puts the error into the applicative functor... Hm

magnus-madsen:
I fixed it in #5361, but please add the (missing) test case and continue with the pattern I have in Weeder:

```
            val err = WeederError.IllegalIntrinsic(loc)
            WeededAst.Expression.Error(err).toSoftFailure(err)
```

Note I call it `err` because we use `e` for expressions.

jaschdoc:
I'm not sure what test case you are referring to? I added https://github.com/flix/flix/blob/40a58e4d150f867d0a0a4509323a0490d3e7d9fe/main/test/ca/uwaterloo/flix/language/phase/TestWeeder.scala#L751

jaschdoc:
Should we perhaps do that in a separate PR?

magnus-madsen:
Yep

PR #5357, opened 2023-01-23
feat: minimize on subsitution

JonathanStarup:
related to #5229.

Temporary universe hack to be solved in later PR

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/unification/Substitution.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/unification/Substitution.scala
@@ -57,46 +57,53 @@ case class Substitution(m: Map[Symbol.KindedTypeVarSym, Type]) {
     */
   def apply(tpe0: Type): Type = {
     // NB: The order of cases has been determined by code coverage analysis.
-    // insideCaseSet means that we are inside a case set formula and should not minimize yet
-    def visit(insideCaseSet: Boolean, t: Type): Type =
-      t match {
-        case x: Type.Var => m.getOrElse(x.sym, x)
-        case Type.Cst(tc, _) => t
-        case Type.Apply(t1, t2, loc) =>
-          // only trigger minimization once for each formula
-          val y =
-            if (!insideCaseSet && isCaseSet(t2)) minimizeSetType(visit(insideCaseSet = true, t2))
-            else visit(insideCaseSet, t2)
-          val z =
-            if (!insideCaseSet && isCaseSet(t1)) minimizeSetType(visit(insideCaseSet = true, t1))
-            else visit(insideCaseSet, t1)
-          z match {
-            // Simplify boolean equations.
-            case Type.Cst(TypeConstructor.Not, _) => Type.mkNot(y, loc)
-            case Type.Apply(Type.Cst(TypeConstructor.And, _), x, _) => Type.mkAnd(x, y, loc)
-            case Type.Apply(Type.Cst(TypeConstructor.Or, _), x, _) => Type.mkOr(x, y, loc)
-
-            // Simplify set expressions
-            case Type.Cst(TypeConstructor.Complement, _) => SetUnification.mkComplement(y)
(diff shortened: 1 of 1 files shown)


magnus-madsen:
I think this hack is pretty ugly. Can we wait until #5359 and then hopefully it will be better? 

Perhaps it would also help with some more helper functions.

magnus-madsen:
!!!jar

JonathanStarup:
> I think this hack is pretty ugly

The concrete universe hack or the whole PR?

> Can we wait until https://github.com/flix/flix/pull/5359 and then hopefully it will be better?

I don't think we can avoid it. The type for `open Not(open Not(open Cst(true)))` is double exponential. We either need to do this (apply exhaustive evaluation after substitution) or improve the smart constructors of types by a lot. The types are reasonable coming out of SVE initially, but applying the substitution blows the formulas up.

I think we need n-ary caseset operations and constants with sets on the type level at least, in order to have a chance to avoid these crazy types

magnus-madsen:
I mean if we could at least move most code out of the main apply function.

mlutze:
The other clean alternative we discussed was just calling minimization at the end of apply, regardless of kind, but we're not sure of the costs

main/src/ca/uwaterloo/flix/language/phase/unification/Substitution.scala
```diff
@@ -57,13 +57,20 @@ case class Substitution(m: Map[Symbol.KindedTypeVarSym, Type]) {
     */
   def apply(tpe0: Type): Type = {
     // NB: The order of cases has been determined by code coverage analysis.
-    def visit(t: Type): Type =
+    // insideCaseSet means that we are inside a case set formula and should not minimize yet
+    def visit(insideCaseSet: Boolean, t: Type): Type =
```
magnus-madsen, line 61: Can we move this visit function out of apply?

magnus-madsen: Why are we not using the kinds to dispatch?

main/src/ca/uwaterloo/flix/language/phase/unification/Substitution.scala
```diff
@@ -82,13 +89,40 @@ case class Substitution(m: Map[Symbol.KindedTypeVarSym, Type]) {
...
+      val univ = List(
+        new Symbol.RestrictableCaseSym(enumSym, "Cst", SourceLocation.Unknown),
+        new Symbol.RestrictableCaseSym(enumSym, "Var", SourceLocation.Unknown),
+        new Symbol.RestrictableCaseSym(enumSym, "Not", SourceLocation.Unknown),
+        new Symbol.RestrictableCaseSym(enumSym, "And", SourceLocation.Unknown),
+        new Symbol.RestrictableCaseSym(enumSym, "Or", SourceLocation.Unknown),
+        new Symbol.RestrictableCaseSym(enumSym, "Xor", SourceLocation.Unknown)
+      )
+      val (env, univ2) = SetFormula.mkEnv(List(tpe), univ)
+      val something = SetFormula.fromCaseType(tpe, env, univ2)
```
magnus-madsen, line 118: please do more cleanup.

JonathanStarup:
> Do we have any understanding of why this blow-up happens inside subst?

When substitution inlines the formulas, it seems almost no minimization happens, so it inlines exponential formulas inside exponential formulas

main/src/ca/uwaterloo/flix/language/phase/unification/Substitution.scala
JonathanStarup: The kind of the partially applied type `Apply(And, All)` does not have kind `CaseSet`. Although the `minimize` call would also not do anything since it also checks for `CaseSet` kind.

Would you prefer if I first do the normal visit, and then call a recursive minimizeCase function. This way the functions will be clearer I think

JonathanStarup: @magnus-madsen ?

magnus-madsen:
We close this and open a new?

JonathanStarup:
Subsumed by #5364

Issue #5246, opened 2023-01-05
Add `foreach-yield` support


magnus-madsen:
I would like to add support for `foreach-yield` following #5132.

What we need is to desugar:

```
  foreach (x <- c1; y <- c2, z <- c3) 
    yield (x, y, z)
```

to

```
region rh {
    let c1 = List#{1, 2, 3};
    let c2 = List#{1, 2, 3};
    let iter = 
        Iterator.flatMap(x -> 
            Iterator.flatMap(y -> 
                Iterator.singleton(rh, (x, y)), 
            Iterable.iterator(rh, c2)), 
        Iterable.iterator(rh, c1));
    let result: Chain[(Int32, Int32)] = Collectable.collect(iter);
    println(result)
}
```

but of course without the variables, I just have them to make it easier to see the structure. 

This desugaring is similar to `for-yield` except for the implicit region, `Iterator.singleton` instead of `Iterator.point`, and the final call to `collect` which wraps everything.

@jaschdoc I am hoping you can do this and give it priority.

~Note there is a bug in iterator, so tests may not work just yet: https://github.com/flix/flix/issues/5245~ I was mistaken.

magnus-madsen:
Iterator is now in, so just need foreach-yield :)

PR #5358, opened 2023-01-23
feat: add `foreach-yield`

jaschdoc:
Closes https://github.com/flix/flix/issues/5246

### DIFF
--- a/main/test/flix/Test.Exp.ForEachYield.flix
+++ b/main/test/flix/Test.Exp.ForEachYield.flix
@@ -1,0 +1,8 @@
+namespace Test/Exp/Foreach {
+
+    @test
+    def testForEachYield01(): List[Int32] =
+        foreach (x <- 1 :: 2 :: Nil)
+            yield x
+
+}


jaschdoc:
- [ ] Add tests

magnus-madsen:
Add lots of tests :) :) :)

magnus-madsen:
In a systematic way, with different collections, etc.

magnus-madsen:
Ill have @mlutze do a quick look.

main/src/ca/uwaterloo/flix/language/phase/Weeder.scala
```diff
@@ -895,6 +895,71 @@ object Weeder {
...
+
+      // Declare functions
+      val fqnEmpty = "Iterator.empty"
+      val fqnSingleton = "Iterator.singleton"
+      val fqnFlatMap = "Iterator.flatMap"
+      val fqnIterator = "Iterable.iterator"
+      val fqnCollect = "Collectable.collect"
+
+      // Make region variable
+      val regionSym = Name.Ident(sp1, "ForEachYieldIterRegion" + flix.genSym.freshId(), sp2)
```
mlutze, line 912: Should be lowercase. Let's also add a `Flix.Delimiter` in the middle so it's clear that it is internal.

mlutze: Also I think there's something like `Name.SyntheticIdent`?

main/src/ca/uwaterloo/flix/language/phase/Weeder.scala
```diff
@@ -895,6 +895,71 @@ object Weeder {
           }
       }
 
+    case ParsedAst.Expression.ForEachYield(sp1, frags, exp, sp2) => {
```
mlutze, line 898: Some in-line example code  would make this easy to follow

main/test/ca/uwaterloo/flix/language/phase/TestResolver.scala
```diff
@@ -1397,4 +1397,37 @@ class TestResolver extends FunSuite with TestUtils {
     val result = compile(input, Options.TestWithLibMin)
     expectError[ResolutionError.IllegalWildType](result)
   }
+
+  test("IllegalForEachYield.01") {
```
mlutze, line 1401: Usually the naming convention is something like "XYZError.SpecificCase.Number".

So something like `UndefinedName.ForYield.01` would be nice here

jaschdoc:
This program prints `Nil` followed by `"a" :: Nil`
```flix
pub def main(): Unit \ IO =
    println("${freach(0)}");
    println("${freach(2)}")

def freach(x: Int32): List[String] =
    foreach (
        if x > 0
    ) yield "a"
```

Should we accept this as normal "dark" behavior?

magnus-madsen:
> 

Fun.

Thoughts @mlutze @JonathanStarup @paulbutcher

mlutze:
IMO no human would do this on purpose. Probably bug, so don't allow. If we find a realistic use-case later then we can re-allow it.

magnus-madsen:
OK, lets try to ban it then. 

But why does it even happen? Should we not be working on the empty iterator?

JonathanStarup:
> OK, lets try to ban it then.
> 
> But why does it even happen? Should we not be working on the empty iterator?

I read it like `{ | if x > 0}` which should be `{}`

magnus-madsen:
@jaschdoc Can you show the translation? None of us understand how you can even get an "a" out of thin air :)

jaschdoc:
Sure, will post an explanation later. I think we decided to allow this for `foreach` which was just desugared to `if (exp) body else ()`

jaschdoc:
> @jaschdoc Can you show the translation? None of us understand how you can even get an "a" out of thin air :)

It's desugared to (omitting regions for legibility)

```scala
Collectable.collect(
    if (x > 0) Iterator.singleton("a") else Iterator.empty()
)
```

magnus-madsen:
Yeah okay. Can we add a check that bans such forms?

jaschdoc:
Should be doable :)

main/src/ca/uwaterloo/flix/language/phase/Weeder.scala
```diff
@@ -895,6 +895,85 @@ object Weeder {
...
+      //
+      // Rewrites a foreach-yield loop into a series of iterators
```
magnus-madsen, line 900: Perhaps we need a full example here?

magnus-madsen:
> Should be doable :)

If this checks makes the code too unreadable, I don't want to do it. But maybe its simple?

jaschdoc:
I think it's just a matter of checking `fragments.headOption` for None and Some(guard) and let the default case be what we have now. Just haven't gotten around to it yet :)

jaschdoc:
Should we allow programs like the following which contains guards before iterating over some collection?
```scala
def f(x: Int32, ys: List[Int32]): List[String] =
    foreach (
        if x > 0;
        y <- xs
    ) yield (x, y)
```

magnus-madsen:
> Should we allow programs like the following which contains guards before iterating over some collection?
> 
> ```scala
> def f(x: Int32, ys: List[Int32]): List[(Int32, Int32)] =
>     foreach (
>         if x > 0;
>         y <- xs
>     ) yield (x, y)
> ```

No

jaschdoc:
Alright. Do we allow guards to refer to variables outside of the loop?

magnus-madsen:
> Alright. Do we allow guards to refer to variables outside of the loop at all (i.e. in **any** of the guard loop fragments regardless of position)?

I think so? You mean like:

```
let x = 2;
foreach (.... if y > x) 
```
should be OK

jaschdoc:
I think it's ready to be merged now

main/test/ca/uwaterloo/flix/language/phase/TestParser.scala
```diff
@@ -423,4 +423,13 @@ class TestParser extends FunSuite with TestUtils {
     val result = compile(input, Options.TestWithLibNix)
     expectError[ParseError](result)
   }
+
+  test("ParseError.ForEachYield.01") {
+    val input =
+      """
+        |def f(): List[Int32] = foreach () yield 1
```
mlutze, line 430: Any reason this has to be a Parse error rather than Weeder error?

main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
```diff
@@ -896,4 +896,61 @@ object WeederError {
          |""".stripMargin
     })
   }
+
+
+  /**
+    * An error raised to indicate that a loop does not iterate over any collection.
```
mlutze, line 902: We always describe each parameter.

main/test/flix/Test.Exp.ForEachYield.flix
```diff
@@ -0,0 +1,300 @@
...
+            yield (x, y, z)
+
+    @test
+    def testForEachYieldMix17(): Chain[(Int32, Int32, Bool)] =
+        let a = List.toSet((1, 2) :: (3, 4) :: Nil);
+        let b = Map#{"1" => true, "2" => true, "3" => true};
+        foreach ((x, y) <- a; z <- b)
+            yield (x, y, z)
+
+    // N.B. There is no instance of `Order` on `Chain` so we cannot have `m[Chain[a]] with Collectable[m]`
```
mlutze, line 238: Should we just add this?

main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
```diff
@@ -896,4 +896,61 @@ object WeederError {
...
+         |
+         |""".stripMargin
+    })
+  }
+
+  /**
+    * An error raised to indicate that a loop has guards before
+    * first iterable collection.
+    */
+  case class LoopGuardsBeforeCollection(loc: SourceLocation) extends WeederError {
```
mlutze, line 933: Do we need these as separate? Why not just "loop thingy must start with collection"?

main/test/flix/Test.Exp.ForEachYield.flix
jaschdoc: Separate PR?

main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
jaschdoc: I considered the same thing but decided maybe it would be "friendlier" to have more fine grained messages. Thoughts?

mlutze: I prefer just one but maybe @magnus-madsen has a preference.

main/test/ca/uwaterloo/flix/language/phase/TestParser.scala
magnus-madsen: I think its fine. But I would just remove the test then.

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
```diff
@@ -534,41 +534,34 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
       Expression ~ optional(optWS ~ ";" ~ optWS ~ Stm ~ SP ~> ParsedAst.Expression.Stm)
     }
 
-    def ForEach: Rule1[ParsedAst.Expression.ForEach] = {
+    object Loops {
```
magnus-madsen, line 537: Loops -> ForFragment

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
```diff
@@ -752,7 +745,8 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
         Upcast | Supercast | Mask | Intrinsic | New | ArrayLit | FList |
         FSet | FMap | ConstraintSet | FixpointLambda | FixpointProject | FixpointSolveWithProject |
         FixpointQueryWithSelect | ConstraintSingleton | Interpolation | Literal | Resume | Do |
-        Discard | Debug | ForYield | ForEach | NewObject | UnaryLambda | Open | OpenAs | HolyName | QName | Hole
+        Discard | Debug |  Loops.ForEachYield | Loops.ForYield | Loops.ForEach | NewObject |
```
magnus-madsen, line 748: ForEach and friends should be at the top-level, not inside some object.

main/src/ca/uwaterloo/flix/language/phase/Weeder.scala
```diff
@@ -881,6 +881,115 @@ object Weeder {
...
+      // wrapped in a Collectable.collect call:
+      //
+      //     foreach (x <- xs) yield x
+      //
+      // desugars to
+      //
+      //     region rh {
```
magnus-madsen, line 893: rh -> rc (this is the new convention)

magnus-madsen (commented):
Lots of small issues, but once resolved, I think we can get this merged!

main/test/ca/uwaterloo/flix/language/phase/TestParser.scala
jaschdoc: Why should the test be removed? I can convert it to a Weeder error in a separate PR :)

magnus-madsen: OK. Keep it :)

jaschdoc:
Aside from the question regarding the parse error test, it should be good to go now :)

magnus-madsen:
Thanks!

Issue #5338, opened 2023-01-19
Remove infix `rem` and `mod` expressions


JonathanStarup:
This program does not parse. It seems to be an issue whenever a `def` precedes a `mod` declaration (because its parsed as the `mod` operator).
```scala
def f(): Unit = ???
mod Inner {
    pub def f(): Unit = ???
}
```
Will this be the final drop that makes us require `{...}` for multiline defs?

magnus-madsen:
No, rather I think its the death of `mod` and `rem` as keywords. Why do they get special treatment? They are so rarely used. And they are just functions.

JonathanStarup:
Then they're also findable in Api and can have doc comments

magnus-madsen:
And auto complete and work on pipelines

magnus-madsen:
- [ ] Remove support for infix `rem` and `mod` expressions.
- [ ] Instead, add library functions for them.
- [ ] Refactor tests etc.

magnus-madsen:
@stephentetley Thanks Stephen

magnus-madsen:
@stephentetley I think this solved now?

Are there other expressions we should consider "demoting" to "just" functions?

stephentetley:
It's solved in the compiler - will the VsCode extension have them as keywords?

magnus-madsen:
> It's solved in the compiler - will the VsCode extension have them as keywords?

Apparently they were never keywords to begin with.

PR #5367, opened 2023-01-25
refactor: change rem and mod to functions (issue #5338)

stephentetley:
__Work in Progress__ 

I think I've done most of the work to change `rem` and `mod` to functions rather than operators, unfortunately this makes a mess of the `coreLibrary` list in Flix.scala, what should we do about bringing in extra modules into `coreLibrary` ?


magnus-madsen:
> I think I've done most of the work to change rem and mod to functions rather than operators, unfortunately this makes a mess of the coreLibrary list in Flix.scala, what should we do about bringing in extra modules into coreLibrary ?

Can we keep them out and just modify those tests that depend on them to use a larger version of the library?

stephentetley:
All tests compiled with `Options.TestWithLibMin` or `DefaultOptions` are failing so that 's quite a lot.

The only "outside" function that the current coreLibrary needs is Int64.mod used by `Thread.flix` - what about either defining a local version in that module or adding an "Internal/Boot.flix" module to `coreLibrary` with `int64Mod` and then both `Thread.flix` and `Int64.flix` can reference that?

By convention only standard lib code should use "Internal/Boot.flix" functions directly.

magnus-madsen:
> The only "outside" function that the current coreLibrary needs is Int64.mod used by Thread.flix - what about either defining a local version in that module or adding an "Internal/Boot.flix" module to coreLibrary with int64Mod and then both Thread.flix and Int64.flix can reference that?

I think we can just use the intrinsic there. $INT64_MOD` or whatever.

stephentetley:
I've changed the nanosecond calculation in `Thread.flix` to this - I hope its right due to the `>0` guard.

> let ns: Int32 = if (nsTotal > 0i64) intValue(valueOf($INT64_REM$(nsTotal, 1_000_000i64))) else 0;

I think the code should now be ready for review...

As far as I could tell, the `mod` and `rem` operators were pretty "shallow" within the compiler code and only appeared in `Parser.scala` and `Weeder.scala` where they have been eliminated.

The `mod` and `rem` operator tests originally in `Test.Exp.Binary.Arithmetic.flix` have been transplanted into the respective `TestIntX.flix` modules. Generally I've quotes to use the function versions of `mod` and `rem` as infix. Occasionally in functions like `logicalRightShift` I've used them with normal function application as they were in expressions with other infix operations.

magnus-madsen:
Thanks!

PR #5370, opened 2023-01-25
feat: bound case set analysis on hover 

JonathanStarup:


JonathanStarup:
!!!jar

JonathanStarup:
@magnus-madsen @mlutze You can try this out, hover will now give you some bounds analysis. Be aware that hover is now exponential time in the number of variables 🐌

magnus-madsen:
I like this idea, but I would like an implementation that is much cleaned up and documented. Right now its too hard to follow. 

I would also move everything into CaseSetFormula.

magnus-madsen:
I think documentation for each function + inline examples would help.

magnus-madsen:
I think we should also decide: is this for us or for users?

If for users then what information should we actually show?

JonathanStarup:
> I think we should also decide: is this for us or for users?
> 
> If for users then what information should we actually show?

I think its useful for both, but in a perfect world, it should probably be opt-in, such that this exp time computation is not done on any hover of restrictable enum type. Maybe it could also be a buwb kind of thing

magnus-madsen:
@JonathanStarup Can we resolve the conflict so I can play more with it?

JonathanStarup:
@magnus-madsen its merged now, should I trigger the hover thing on some condition on what do you think should happen?

magnus-madsen:
I would like:

- Move as much as possible into CaseSetFormula.
- Document each function with an example.
- Cleanup the code + only trigger on the right kind.

Then I will review again. 

Also I think we should only show upper and lower bounds.

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
```diff
@@ -78,10 +79,11 @@ object HoverProvider {
 
   private def hoverType(tpe: Type, loc: SourceLocation, current: Boolean)(implicit index: Index, root: Root, flix: Flix): JObject = {
     val minTpe = minimizeType(tpe)
+    val boundsRep = SetFormula.restrictableEnumBounds(minTpe).getOrElse("")
     val markup =
       s"""${mkCurrentMsg(current)}
          |```flix
-         |${FormatType.formatType(minTpe)}
+         |$boundsRep${FormatType.formatType(minTpe)}
```
magnus-madsen, line 86: I dont like that you call into some function and rely on it to return the same string for non-restrictable enums. 

Why not just insert an extra string after the call to FormatType?

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
```diff
@@ -94,10 +96,11 @@ object HoverProvider {
     val minPur = minimizeType(pur)
     val minEff = minimizeType(eff)
     val minTpe = minimizeType(tpe)
+    val boundsRep = SetFormula.restrictableEnumBounds(minTpe).getOrElse("")
     val markup =
       s"""${mkCurrentMsg(current)}
          |```flix
-         |${formatTypAndEff(minTpe, minPur, minEff)}
+         |$boundsRep${formatTypAndEff(minTpe, minPur, minEff)}
```
magnus-madsen, line 103: Same comment here.

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
JonathanStarup: It returns `None` for non-restrictable enums, so `$boundsRep` is just `""`. The String of `FormatType` is untouched. The awkwardness is just to avoid extra newlines on non-restrictable enums

main/src/ca/uwaterloo/flix/language/phase/unification/SetFormula.scala
```diff
@@ -221,6 +270,63 @@ object SetFormula {
...
+    * restrictableEnumBounds(Expr[s ++ <Expr.Cst>][Int32]) =
+    *     Must have  : {Cst}
+    *     Has at most: {Cst, Var, Not, And, Or, Xor}
+    *
+    * restrictableEnumBounds(Expr[s1 -- s2 -- <Expr.Xor, Expr.And> ++ <Expr.Cst, Expr.Not>]) =
+    *     Must have  : {Cst, Not}
+    *     Has at most: {Cst, Var, Not, Or}
+    * }}}
+    */
+  def restrictableEnumBounds(tpe: Type)(implicit root: TypedAst.Root): Option[String] = {
```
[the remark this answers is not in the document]
JonathanStarup: Its the entrypoint for `HoverProvider`

main/src/ca/uwaterloo/flix/language/phase/unification/SetFormula.scala
```diff
@@ -221,6 +299,63 @@ object SetFormula {
...
+    * Extracts the index type (kind CaseSet) of a restrictable enum along
+    * with its enumSym.
+    * {{{
+    * findIndexOfEnum(E[index: CaseSet(Color), Int32, String]) = Some((index, Color))
+    * findIndexOfEnum(Int32) = None
+    * findIndexOfEnum(Int32 -> E[index]) = None
+    * }}}
+    */
+  @tailrec
+  private def findIndexOfEnum(tpe: Type): Option[(Type, Symbol.RestrictableEnumSym)] = tpe match {
```
magnus-madsen, line 312: Question: Why not do a kind check and then `tpe.typeArguments.head`?

main/src/ca/uwaterloo/flix/language/phase/unification/SetFormula.scala
```diff
@@ -163,6 +164,83 @@ object SetFormula {
...
+    *
+    * boundsAnalysisType((s1 and  s2 and Cst{4, 6}) or Cst{1, 3}) =
+    *     ({1, 3}, {1, 2, 3, 5, 7, 8, ..., n})
+    * }}}
+    */
+  private def boundsAnalysis(f: SetFormula)(implicit univ: Set[Int]): (Set[Int], Set[Int]) = {
+    val bot = SetFormula.Cst(Set.empty): SetFormula
+    val top = SetFormula.Cst(univ): SetFormula
+
+    // create all possible assignments of the free variables (exponentially many)
```
magnus-madsen, line 221: Don't we already have a function that does this?

main/src/ca/uwaterloo/flix/language/phase/unification/SetFormula.scala
JonathanStarup: The kind of `tpe` is `Kind.Star`, but I could find `typeArguments.head` and then check if that's a case set?

main/src/ca/uwaterloo/flix/language/phase/unification/SetFormula.scala
JonathanStarup: `minimize` does this internally but not reusable. Ill try to refactor for reuse

JonathanStarup:
!!!jar

PR #5373, opened 2023-01-26
feat: add open_as expression

mlutze:
related to #5229

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/inference/RestrictableChooseInference.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/inference/RestrictableChooseInference.scala
@@ -319,7 +319,14 @@ object RestrictableChooseInference {
   }
 
   /**
-    * Performs type inference on the given OpenAs expression
+    * Performs type inference on the given OpenAs expression.
+    *
+    * `OpenAs X e` requires that `e` have the type X[s] for some s
+    * The result type of the expression is X[s + φ] for some free φ
+    *
+    *         Γ ⊢ e : X[s][α1 ... αn]
+    * -------------------------------------
+    * Γ ⊢ open_as X e : X[s + φ][α1 ... αn]
     */
   def inferOpenAs(exp0: KindedAst.Expression.OpenAs, root: KindedAst.Root)(implicit flix: Flix): InferMonad[(List[Ast.TypeConstraint], Type, Type, Type)] = exp0 match {
     case KindedAst.Expression.OpenAs(sym, exp, tvar, loc) =>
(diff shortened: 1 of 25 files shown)


magnus-madsen:
Can we merge `Open` and `OpenAs`?

main/src/ca/uwaterloo/flix/language/phase/inference/RestrictableChooseInference.scala
```diff
@@ -318,6 +318,41 @@ object RestrictableChooseInference {
       } yield (constrs, tpe, pur, eff)
   }
 
+  /**
+    * Performs type inference on the given OpenAs expression
+    */
+  def inferOpenAs(exp0: KindedAst.Expression.OpenAs, root: KindedAst.Root)(implicit flix: Flix): InferMonad[(List[Ast.TypeConstraint], Type, Type, Type)] = exp0 match {
```
magnus-madsen, line 324: I think we need some more documentation on all the return types.

JonathanStarup:
> Can we merge `Open` and `OpenAs`?

I think we need to remove the choose* body check before we remove the old open

mlutze:
> Can we merge `Open` and `OpenAs`?

It's a bit annoying to write `open_as Red Red` though

mlutze:
> > Can we merge `Open` and `OpenAs`?
> 
> It's a bit annoying to write `open_as Red Red` though

Maybe the syntax could be `open X [as Y]` where the Y part is required if X is not a tag

magnus-madsen:
> > > Can we merge `Open` and `OpenAs`?
> > 
> > 
> > It's a bit annoying to write `open_as Red Red` though
> 
> Maybe the syntax could be `open X [as Y]` where the Y part is required if X is not a tag

Lets do that and then I can merge.

JonathanStarup:
> > Can we merge `Open` and `OpenAs`?
> 
> It's a bit annoying to write `open_as Red Red` though

open_as Color Red

mlutze:
> > > Can we merge `Open` and `OpenAs`?
> > 
> > 
> > It's a bit annoying to write `open_as Red Red` though
> 
> `open_as Color Red`?

Good point. Still annoying though.

mlutze:
It looks like we need both of them. I tried removing `open` and replacing it with the `open_as` construct. This is the result:

```
-- Type Error -------------------------------------------------- main/test/flix/Test.Exp.Choose.Recursive.flix

>> Unable to unify the Boolean formulas: '{Cst} + c0' and '{Not}'.

67 |         let cstOrNot = if (true) open Expr.Cst(true) else open Expr.Not(open Expr.Cst(true));
                                                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                                                    mismatched boolean formulas.

Type One: {Cst} + c0
Type Two: {Not}
```

The issue is that previously, `open` applied specifically to the `Expr.Not` tag, opening its index both for input and output. With `open_as`, only the output is opened.

Issue #5141, opened 2022-12-18
Rewrite the `Safety` phase to use `Validation[Expression, ...]`


magnus-madsen:
Currently most functions have the signature:

```
private def visitExp(e0: Expression, renv: RigidityEnv)(implicit flix: Flix): List[CompilationMessage]
```

this should become:

```
private def visitExp(e0: Expression, renv: RigidityEnv)(implicit flix: Flix): Validation[Expression, CompilationMessage]
```

PR #5375, opened 2023-01-26
chore: basic change to Validation[Expression, CompilationMessage] in Safety

AStenbaek:
Fixes flix/flix#5141

Rough draft.
Interested in feedback on the monadic style I used.


main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -33,303 +33,518 @@ object Safety {
...
+          _ <- Validation.foldRight(exps)(visit(exp))((e, _) => visit(e))
+          x <- Success(exp0)
+        } yield x
 
       case Expression.Unary(_, exp, _, _, _, _) =>
-        visit(exp)
+        for {
+          _ <- visit(exp)
+          x <- Success(exp0)
+        } yield x
```
magnus-madsen, line 118: We will be inserting `TypedAst.Expression.Error` nodes. Hence we should not return `x` but actually rebuild the expression from its sub-expressions.

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -33,303 +33,518 @@ object Safety {
...
 
       case Expression.Binary(_, exp1, exp2, _, _, _, _) =>
-        visit(exp1) ::: visit(exp2)
+        for {
```
magnus-madsen, line 121: We should always avoid `for` because this treats `Validation` as a Monad, but actually its a lie. We should only treat it as an applicative functor.

magnus-madsen:
I will let @mlutze add some more comments :)

magnus-madsen:
@mlutze Could we prevent the accidental use of `for`?

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
magnus-madsen: Indeed, but if we have two sub-expressions we want to collect errors from both of them :)

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -33,303 +33,518 @@ object Safety {
     //
     // Collect all errors.
     //
-    val defErrors = root.defs.flatMap {
-      case (_, defn) => visitDef(defn, root)
-    }
 
-    val instanceErrors = visitSendable(root)
+    for {
+      _ <- Validation.foldRight(root.defs.toList)(root.toSuccess) {
```
mlutze, line 38: A few things here:

1. We never do multiline blocks inside `for`
2. Like Magnus said, we should avoid `for` when dealing with `Validation` since it stops at the first error.
3. `Validation.foldRight` does the same: stopping at the first error.
4. Yielding `errors` seems very suspicious to me. In general we yield the "good" part rather than the error.

What I would expect here is something like the following:
1. Use `Validation.traverse`  to go over the all the defs in the function. Assign the result to e.g., `defsVal`
2. Also use `Validation.traverse` to go over all the instances with `visitSendable`. Assign the result to e.g. `instsVal`
3. Map over `defsVal` and `instsVal` with `mapN`, returning a new root where we replace the `defs` and `insts` field.

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -33,303 +33,518 @@ object Safety {
...
 
     val sendableClass = new Symbol.ClassSym(Nil, "Sendable", SourceLocation.Unknown)
 
-    root.instances.getOrElse(sendableClass, Nil) flatMap {
-      case Instance(_, _, _, _, tpe, _, _, _, loc) =>
-        if (tpe.typeArguments.exists(_.kind == Kind.Bool))
-          List(SafetyError.SendableError(tpe, loc))
-        else
-          Nil
+    Validation.foldRight(root.instances.getOrElse(sendableClass, Nil))(root.toSuccess[Root, CompilationMessage]) {
```
mlutze, line 57: (Fold right bad :) )

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -33,303 +33,518 @@ object Safety {
...
   /**
     * Performs safety and well-formedness checks on the given expression `exp0`.
     */
-  private def visitExp(e0: Expression, renv: RigidityEnv, root: Root)(implicit flix: Flix): List[CompilationMessage] = {
+  private def visitExp(e0: Expression, renv: RigidityEnv, root: Root)(implicit flix: Flix): Validation[Expression, CompilationMessage] = {
 
-    def visit(exp0: Expression): List[CompilationMessage] = exp0 match {
-      case Expression.Cst(_, _, _) => Nil
+    def visit(exp0: Expression): Validation[Expression, CompilationMessage] = exp0 match {
+      case Expression.Cst(_, _, _) => Success(exp0)
```
mlutze, line 84: Stylistically we almost always use the implicit method `toSuccess`

mlutze (commented):
The overall takeaway is that we should be using the following pattern:
1. Individually visit each expression/node
2. Map over all the nodes together
3. Reconstruct the node

You can look [here](https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/Kinder.scala#L451) at a relatively new compiler phase that is fairly principled with this pattern.

mlutze:
> @mlutze Could we prevent the accidental use of `for`?

We can force authors to use `flatMapN` instead of `for` by removing the `flatMap` function from `Validation`.

This means in `Flix.scala` the big `for` thingy that goes through all the phases may get quite ugly, but maybe we can add a helper to keep it nice.

magnus-madsen:
> > @mlutze Could we prevent the accidental use of `for`?
> 
> We can force authors to use `flatMapN` instead of `for` by removing the `flatMap` function from `Validation`.
> 
> This means in `Flix.scala` the big `for` thingy that goes through all the phases may get quite ugly, but maybe we can add a helper to keep it nice.

I think we should try that. Will you create a ticket?

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
AStenbaek: Was also my original choice, but the compiler refused that without type annotation in a lot of cases.

AStenbaek:
Hope to fix it before next monday.

magnus-madsen:
@AStenbaek What's the status on this? I have someone else who will be making changes to these files soon and I would prefer if neither of you end up in conflict hell.

AStenbaek:
I will start tonight then. I expect a PR by tomorrow at the latest.

AStenbaek:
I have rewritten more like the style mentioned here. Might have some ugly things here and there, but the overall structure should be fixed.

magnus-madsen:
Looks better.

- I will ask Matt for another detailed review.
- I see the formatting is very non-standard. Try `ctrl+alt+l` to reformat the code.
- I also see some conflict.

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -33,303 +33,406 @@ object Safety {
...
+          Instance(doc, ann, mod, clazz, tpe, tconstrs, defs, ns, loc).toSuccess
+        }
     }
   }
 
   /**
     * Performs safety and well-formedness checks on the given definition `def0`.
     */
-  private def visitDef(def0: Def, root: Root)(implicit flix: Flix): List[CompilationMessage] = {
+  private def visitDef(def0: Def, root: Root)(implicit flix: Flix): Validation[Root, CompilationMessage] = {
```
mlutze, line 67: This should be returning a `Def`, not a `Root`

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -33,303 +33,406 @@ object Safety {
...
-          rules.flatMap { case MatchRule(_, g, e) => g.toList.flatMap(visit) ::: visit(e) }
+      case Expression.Match(exp, rules, tpe, pur, eff, loc) =>
+        val expVal = visit(exp)
+        val rulesVal = traverse(rules) { case MatchRule(p, g, e) =>
+          val eVal = visit(e)
+          g match {
+            case Some(gexp) =>
+              val gexpVal = visit(gexp)
+              mapN(eVal, gexpVal) { (e1, e2) => MatchRule(p, Some(e1), e2) }
+            case None => mapN(eVal) { e => MatchRule(p, None, e) }
```
mlutze, line 163: We can use `traverseOpt` here instead

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -33,303 +33,406 @@ object Safety {
...
+      case Expression.Sig(sym, tpe, loc) => Expression.Sig(sym, tpe, loc).toSuccess
 
-      case Expression.Hole(_, _, _) => Nil
+      case Expression.Hole(sym, tpe, loc) => Expression.Hole(sym, tpe, loc).toSuccess
 
-      case Expression.HoleWithExp(exp, _, _, _, _) =>
-        visit(exp)
+      case Expression.HoleWithExp(exp, tpe, pur, eff, loc) =>
+        val expVal = visit(exp)
+        mapN(expVal) { e => Expression.HoleWithExp(e, tpe, pur, eff, loc) }
```
mlutze, line 95: Style note: We almost never have lambdas on the same line as `map`. Usually it looks like

```scala
mapN(expVal) { 
    e => Expression.HoleWithExp(e, tpe, pur, eff, loc)
}
```

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -33,303 +33,406 @@ object Safety {
...
+        val expVal = visit(exp)
+        mapN(expVal) { e => Expression.Of(sym, e, tpe, pur, eff, loc) }
 
-      case e@Expression.Cast(exp, _, _, _, _, _, _, _) =>
-        val errors = checkCastSafety(e)
-        visit(exp) ::: errors
+      case e@Expression.Cast(exp, declaredType, declaredPur, declaredEff, tpe, pur, eff, loc) =>
+        val check = checkCastSafety(e)
+        val expVal = visit(exp)
+        mapN(expVal, check) { (e, _) => Expression.Cast(e, declaredType, declaredPur, declaredEff, tpe, pur, eff, loc) }
```
mlutze, line 280: Try to avoid shadowing `e`

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -33,303 +33,406 @@ object Safety {
...
-
-      case Expression.PutChannel(exp1, exp2, _, _, _, _) =>
-        visit(exp1) ::: visit(exp2)
-
-      case Expression.SelectChannel(rules, default, _, _, _, _) =>
-        rules.flatMap { case SelectChannelRule(_, chan, body) => visit(chan) :::
-          visit(body)
-        } :::
-          default.map(visit).getOrElse(Nil)
+        val objImpl = traverse(checkObjectImplementation(clazz, erasedType, methods, loc)) { x => x.toFailure }
```
mlutze, line 351: This looks strange to me. Perhaps `checkObjectImplementation` should just return a `Validation[ObjectImplementationOrWhatever, SafetyError]`?

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -33,303 +33,406 @@ object Safety {
...
+          case e: Expression.Tuple => mapN(visit(e)) { ex => Expression.Par(ex, loc) }
+          case _ => IllegalParExpression(exp, exp.loc).toFailure
         }
 
-      case Expression.ParYield(frags, exp, _, _, _, _) =>
-        frags.flatMap { case ParYieldFragment(_, e, _) => visit(e) } ::: visit(exp)
+      case Expression.ParYield(frags, exp, tpe, pur, eff, loc) =>
+        val fragsVal = traverse(frags) { case ParYieldFragment(pat, exp, loc) =>
+          mapN(visit(exp)) { e => ParYieldFragment(pat, e, loc) } }
+        val expVal = visit(exp) // Maybe?
```
mlutze, line 394: Why `Maybe?`? :)

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -636,16 +740,16 @@ object Safety {
         case Denotation.Relational => terms
         case Denotation.Latticenal => terms.dropRight(1)
       }
-      val err2 = relTerms.flatMap(freeVarsOf).filter(latVars.contains).map(
-        s => IllegalRelationalUseOfLatticeVariable(s, loc)
-      )
+      val err2 = mapN(traverse(relTerms.flatMap(freeVarsOf).filter(latVars.contains)) {
+        s => IllegalRelationalUseOfLatticeVariable(s, loc).toFailure
+      })(_ => den)
```
mlutze, line 745: Why do we map to `den` here?

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -717,53 +821,54 @@ object Safety {
...
-  private def checkTerms(terms: List[Expression], latVars: Set[Symbol.VarSym], loc: SourceLocation): List[CompilationMessage] = {
+  private def checkTerms(terms: List[Expression], latVars: Set[Symbol.VarSym], loc: SourceLocation): Validation[List[Expression], CompilationMessage] = {
     // Compute the free variables in all terms.
     val allVars = terms.foldLeft(Set.empty[Symbol.VarSym])({
       case (acc, term) => acc ++ freeVars(term).keys
     })
 
     // Compute the lattice variables that are illegally used in the terms.
-    allVars.intersect(latVars).toList.map(sym => IllegalRelationalUseOfLatticeVariable(sym, loc))
+    mapN(traverse(allVars.intersect(latVars).toList)(IllegalRelationalUseOfLatticeVariable(_, loc).toFailure)) { _ => terms }
```
mlutze, line 844: I think the structure of this function will need to change in order to go over the expressions individually. You can probably leave a TODO though.

mlutze (changes requested):
Mostly looks good. There are a few places where you map over a `Validation` like 
```scala 
    mapN(theVal) {
        case _ => xyz
    }
```
which seems suspicious to me since it means we're throwing away the transformed object. Probably the only time we should do this is if the return value is `Validation[Unit, ...]`

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
AStenbaek: I did that since `relTerms` is derived from `den`. This is mostly a result of me following the original structure somewhat, so I am not gonna argue it couldn't be cleaner.

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
AStenbaek: Not gonna disagree. Mostly a result of following the old structure.

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
AStenbaek: Forgot to delete it.

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
AStenbaek: Done

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
AStenbaek: Done

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
AStenbaek: Fixed

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
AStenbaek: Alright.

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
AStenbaek: Will keep that in mind going forward.

AStenbaek:
The "suspicious" pattern of throwing away a construction, is a bit of a hack to propagate errors, due to parts of the program being adapted quite directly. Ideally this should be changed, but I figured if someone else had to start working on the Safety file, it would be preferable to get the overall structure done to avoid merge conflicts.

magnus-madsen:
Also mark comments as resolved if they are done.

magnus-madsen:
Thanks!

magnus-madsen:
@AStenbaek For the next iteration, we should use `recoverOne`, but lets chat about that. Come find me?

PR #5382, opened 2023-01-27
feat: allow unit constants in restrictable tag patterns

JonathanStarup:
Related to #5229 

I've transformed a Unit pattern into a wild pattern to avoid expanding the AST's


mlutze:
Looks fine. Do we want to add a TODO since this is technically a hack?

mlutze (approved):

PR #5385, opened 2023-01-27
test: More testing of restrictable enum declarations

JonathanStarup:


magnus-madsen:
What about systematic tests for functions with signatures?

main/test/flix/Test.Dec.RestrictableTag.flix
```diff
@@ -0,0 +1,71 @@
+mod Test.Dec.RestrictableTag {
+
+/// Simple nullary cases
```
mlutze, line 3: No indent because we're waiting on the top-level namespace magic, right? :)

mlutze (approved):

JonathanStarup:
> What about systematic tests for functions with signatures?

I thought I'd just have this in its own PR since its isolated

magnus-madsen:
> > What about systematic tests for functions with signatures?
> 
> I thought I'd just have this in its own PR since its isolated

OK, so more tests to come.

PR #5386, opened 2023-01-27
chore: remove flatMap from Validation

### LINKED ISSUE #5377: Remove `flatMap` from `Validation`
In order to make `for` style less tempting.

mlutze:
fixes #5377. I added an implicit extension method so that Flix.scala does not become a big mess, but I am not sure whether it is worth it.


magnus-madsen:
Thanks!

PR #5392, opened 2023-01-29
test: add some type signature tests

mlutze:
related to #5229


magnus-madsen:
Perhaps also some simpler tests with Color?

magnus-madsen:
And with no recursion.

mlutze:
These tests have no recursion. Note that they use the simplified `Expr` enum which is all singletons

PR #5409, opened 2023-01-30
feat: add `Validation.softRecoverOne`

jaschdoc:
Related to https://github.com/flix/flix/issues/5026#issuecomment-1407690857

PR #5411, opened 2023-01-30
feat: improve error handling in Namer

jaschdoc:
Depends on https://github.com/flix/flix/pull/5409


magnus-madsen:
Why these cases? The general pattern, which you have in e.g. `Weeder` is that if there is call that may fail:

```
      mapN(weedLiteral(lit)) {
        case l => WeededAst.Expression.Cst(l, mkSL(sp1, sp2))
      }
      }.softRecoverOne(WeededAst.Expression.Error)
```

here `weedLiteral` may fail, then we do a soft recover. 

But in these cases I don't see what we are recovering from.

magnus-madsen:
I am dragging @mlutze into this. 

I think there is a general question/comment here about whether every expression should be equipped with a `softRecoverOne` function. This very much corresponds to inserting a try-catch handler around every recursive call and then returning a `Expression.Error`. 

What I originally envisioned was only installing this "try-catch" when there was an error that could occur from a non-expression returning function (e.g. `weedLiteral`) (since from a recursive call it would already be handled). But I now see that its somewhat difficult to see "when" these should be inserted.

magnus-madsen:
Blindly inserting them everywhere is not what we want, because then the whole AST ends up being replaced by error.

mlutze:
My idea would take a lot of rewriting, but I think it feels like the safest:

Add a case to `Validation`: We now have
```
enum Validation[T, E] {
    case Success(T)
    case RecoveredError(T, E)
    case RecoverableError(E)
    case FatalError(E)
}
```

My hypothesis is that at the error discovery site, we are in the best position to determine if an error is fatal or not,
rather than at the parent site. I also believe this will be more robust to refactoring, because there is one clear place the recoverability is identified, whereas with a try-catch style, it may be caught in any of several places in the call stack.

The only issue I see is that `flatMap` and `map` would need an additional `default` value for recoverable errors, which would require a lot of work.

magnus-madsen:
@mlutze I think we can look into this later.

jaschdoc:
> Blindly inserting them everywhere is not what we want, because then the whole AST ends up being replaced by error.

I'm pretty sure I've treated it like a try-catch in this PR (not just blind insertion). `visiType` could fail so the expressions that call this function ends with a recovery. Same thing goes for `visitJvmMethod`

magnus-madsen:
> > Blindly inserting them everywhere is not what we want, because then the whole AST ends up being replaced by error.
> 
> I'm pretty sure I've treated it like a try-catch in this PR (not just blind insertion). `visiType` could fail so the expressions that call this function ends with a recovery. Same thing goes for `visitJvmMethod`. In short I'd say it adheres to the same pattern as seen in the Weeder.

By using the new recover function this should become clearer.

PR #5417, opened 2023-02-02
feat: new restrictable tag type rule

JonathanStarup:
* `open Tag(...)` uses the schema instantiation `t -> Enum[s][targs...]` and then returns the type `Enum[s + {Tag}][targs...]`
    * this means that an open tag might still have an upper bound, compared to the previous rule
    * this means that for nested tags, the inner expression does not have to include the constructed label
* `Tag(...)` uses the schema instantiation `t -> Enum[{Tag}][targs...]` and returns `Enum[{Tag}][targs...]`
    * This rule exist to limit type variables for non-recurse tags
    * `open Tag` for non-recusive tags have no upperbound
    * For non-recursive tags, `open_as Enum Enum.Tag(...)` is equivalent to `open Enum.Tag(...)`.

### DIFF
--- a/main/test/ca/uwaterloo/flix/language/phase/TestTyper.scala
+++ b/main/test/ca/uwaterloo/flix/language/phase/TestTyper.scala
@@ -1916,21 +1916,4 @@ class TestTyper extends FunSuite with TestUtils {
         |""".stripMargin
     expectError[TypeError.GeneralizationError](compile(input, Options.TestWithLibNix))
   }
-
-  test("TestCaseSetAnnotation.05") {
-    val input =
-      """
-        |restrictable enum Binary[s] {
-        |    case Zero(Binary[s]), One(Binary[s]), End
-        |}
-        |
-        |// forgot open on recursive variable use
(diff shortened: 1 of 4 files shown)


main/test/ca/uwaterloo/flix/language/phase/TestTyper.scala
```diff
@@ -1916,21 +1916,4 @@ class TestTyper extends FunSuite with TestUtils {
         |""".stripMargin
     expectError[TypeError.GeneralizationError](compile(input, Options.TestWithLibNix))
   }
-
-  test("TestCaseSetAnnotation.05") {
```
magnus-madsen, line 1920: This is a negative test that was removed? Should we be scared?

JonathanStarup: Its much harder to observe this issue now since constructors "add on top"

magnus-madsen:
@mlutze Any comments?

mlutze:
Seems reasonable. Let's talk tomorrow

PR #5419, opened 2023-02-02
feat: improve error handling in Weeder pt. 2

jaschdoc:


magnus-madsen:
I am merging this, but I think we should move towards recovering from more specific errors (if possible).

Issue #5030, opened 2022-12-01
Add `Vector` data type to compiler and standard library


magnus-madsen:
Add a new `Vector` data type which is fundamentally like `Array` but is immutable and hence does not have a region.

We should finish #4760 before starting on this ticket.

Each bullet below should probably be its own PR.

### Phase 1: Basics

- [x] Add `Vector.flix` with:
```
type alias Vector[t] = Array[t, true]
```

### Phase 2: Library

Add a few functions to the `Vector` module:

- [x] Add `Vector.nth`
- [x] Add `Vector.append`
- [x] Add `Vector.length`
- [x] Add `Vector.repeat`
- [x] Add `Vector.slice`
- [x] Add `Vector.map`

Internally these can use `Array` functions or have their own implementation, but they must have unit tests. 
(Can be copy-pasted from `TestArray`).

### Phase 3: Syntax

- [x] Add the `Vector` type to the Flix compiler itself and drop the enum.
- [ ] Add pattern matching in `Vector`'s. (We can reuse old code from `Array` which is still in the compiler).

### Phase 4: Use Vector and Expand Library

- [ ] Use `Vector` in the Datalog engine.
- [ ] Use `Vector` in the channel implementation.
- [x] Expand the `Vector` module with functions from the `Array` module.

JonathanStarup:
I think maybe `Array[t, Static]` is better because then you can clearly see the casts instead of this magic pure array

magnus-madsen:
> I think maybe `Array[t, Static]` is better because then you can clearly see the casts instead of this magic pure array. The beginning of this implementation is in Fixpoint/Tuple

But no writes would ever be supported, so there would be no effect.

JonathanStarup:
I was just thinking about what to put into the vector `case Vector(...)`

magnus-madsen:
Its a good point; I think really what we want is to disallow `Array[t, true]` (and similar) in `Safety`?

JonathanStarup:
the counter point is that you need a cast to get this thing but yeah. it feels like the language assumes that its a region so true (and even false) is a bit weird

magnus-madsen:
> the counter point is that you need a cast to get this thing but yeah. it feels like the language assumes that its a region so true (and even false) is a bit weird

Ultimately this will be changed. Its more of an artifact of how we do things today.

JonathanStarup:
> `type alias Vector[t] = Array[t, true]`

[...]

PR #5420, opened 2023-02-02
feat: add immutable Vector datatype (issue #5030)

stephentetley:
This PR adds an immutable Vector type.

The type of `Vector` is currently a type alias to `Array`.

> type alias Vector[t] = Array[t, true]

Within the Vector module the Array should be effectively "interned" - provided a programmer uses only the Vector API to work on Vectors the internal Array won't escape. All "updating" functions will produce fresh vectors.

Generally functions that produce Vectors are written with `buildVector` which generates a fresh array, applies the vector initialization function and casts to the Vector type.

There are a couple of versions of `buildVector` in the API to produce two arrays (for `unzip`) and a MutList when the size is unknown (for `Iterator.toVector`). There is an internal version `buildVector0` that generates an array out of just a region - ideally we wouldn't need the=is function but it seems necessary for the vector copying function `slice`.


magnus-madsen:
Thanks Stephen. This looks great! :-)

I am trying to understand the "compiler API" such that `Vector` could become a native type. The purpose would be to make it safer to use and ensure that we did not miss anything. Later, we could also consider pattern matching.

In any case, the current array expressions are:

```
case class ArrayLit(...)
case class ArrayNew(...)
case class ArrayLoad(...)
case class ArrayLength(...)
case class ArrayStore(...)
case class ArraySlice(...)
```

I would replicate (copy) these, but since `Vector`s are immutable I would drop `ArrayNew` and `ArrayUpdate`.

But this is not sufficient, because we need a way to build `Vector`s incrementally. 

This suggests I need a few new primitives (following your code):

- `def buildVector(rc: Region[r], len: Int32, f: Array[a, r] -> Unit \ ef): Vector[a] \ ef`
- `def buildVector2(rc: Region[r], len1: Int32, len2: Int32, f: Array[a, r] -> Array[b, r] -> Unit \ ef): (Vector[a], Vector[b]) \ ef`
- `def buildVectorViaMutList(rc: Region[r], f: MutList[a, r] -> Unit \ ef): Vector[a] \ ef`
- `def buildVector0(rc: Region[r], f: Region[r] -> Array[a, r] \ ef): Vector[a] \ ef`

These have to use `unsafe_cast`, but that could be made safe if there was a primitive `Array[t, r] -> Vector[t]`. I think this would be safe as long as we copy the input array. I am not sure if this is sufficient once `Vector[t] != Array[t, _]`  because then we can no longer use `Array.X` functions. For example, `Vector.repeat` would not be able to call `Array.transform!`.

In any case, the above is pretty easy (on the compiler side).

Looking at `buildVector`, I see that it uses `f` to generate side-effects on the Vector. But could it instead look like:

```
def buildVector(len: Int32, f: Int32 -> a \ ef): Vector[a] \ ef
```

For e.g. `Vector.reverse` this look sufficient. For `Vector.intersperse` this requires some arithmetic, but should still work. For `Vector.flatten` should also be OK. For `Vector.filter` perhaps the right thing is just to use a `MutList` and go back to a `Vector`?

One issue is whether the compiler would also have to provide adapted copies of `buildVector2`, `buildVectorViaMutList`, and `buildVector0`.

stephentetley:
Thanks Magnus.

> def buildVector(len: Int32, f: Int32 -> a \ ef): Vector[a] \ ef

Good spot! I missed this - also the initializing traversal would be inside `buildVector` so that would insure every element is touched.

I'm on holiday today so I'll add that version as `buildVectorWithIndex` and see how many original uses of `buildVector` it can replace.

stephentetley:
> def buildVectorWithIndex(len: Int32, f: Int32 -> a \ ef): Vector[a] \ ef

I've just spotted this is the function `init` which we had already.

main/src/library/Vector.flix
```diff
@@ -510,12 +520,11 @@ namespace Vector {
...
     /// This is an explicit helper to avoid code duplication.
     ///
-    def rotateLeftHelper(n: Int32, v: Vector[a]): Vector[a] = region rc {
+    def rotateLeftHelper(n: Int32, v: Vector[a]): Vector[a] = 
         let len = length(v);
-        buildVector(rc, len, arr ->
-            forEachWithIndex((i, x) -> {let ix = Int32.mod(i - n, len); Array.put(x, ix, arr)}, v)
-        )
-    }
+        let f = ix -> {let readIx = Int32.mod(ix + n, len); get(readIx, v)};
```
magnus-madsen, line 525: Beautiful 👍

stephentetley: Thanks

stephentetley:
> One issue is whether the compiler would also have to provide adapted copies of `buildVector2` , `buildVectorViaMutList`, and `buildVector0` ?

These were a "McGuffin" that was stopping me see they weren't necessary - I've now eliminated them,

main/src/library/Vector.flix
```diff
@@ -0,0 +1,1389 @@
+/*
+ *  Copyright 2023 Magnus Madsen, Stephen Tetley
```
magnus-madsen, line 2: I don't think I deserve much credit for this one :)

main/src/library/Vector.flix
```diff
@@ -0,0 +1,1389 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+///
+/// A Vector data type which is fundamentally like Array but is immutable.
+///
+
+type alias Vector[a] = Array[a, true]
```
magnus-madsen, line 21: We would want instances here, but we cannot add those until `Vector` becomes a type in its own right.

main/src/library/Vector.flix
```diff
@@ -0,0 +1,1389 @@
...
+    ///
+    /// Retrieves the value at position `i` in the vector `v`.
+    ///
+    pub def get(i: Int32, v: Vector[a]): a =
+        Array.get(i, v)
+
+    ///
+    /// Optionally returns the element at position `i` in the vector `v`.
+    ///
+    pub def nth(i: Int32, v: Vector[a]): Option[a] =
```
magnus-madsen, line 88: I am not sure we want `nth` to return an `Option`. For other types, I think we don't because usually performance is important. Esp. for Array/Vector.

(Of course the program could then crash if out of bounds).

main/src/library/Vector.flix
```diff
@@ -0,0 +1,1389 @@
...
+    ///
+    pub def length(v: Vector[a]): Int32 = Array.length(v)
+
+    ///
+    /// Returns a fresh vector.
+    ///
+    /// The function `f` is applied to a uninitialized array, it should should set each element
+    /// to an actual value.
+    ///
+    pub def buildVector(rc: Region[r], len: Int32, f: Array[a, r] -> Unit \ ef): Vector[a] \ ef =
```
magnus-madsen, line 110: So if this becomes a primitive, we should be good?

main/src/library/Vector.flix
```diff
@@ -0,0 +1,1389 @@
...
+            let _ = f(arr);
+            arr
+        ) as Array[a, true] \ ef
+
+    ///
+    /// Returns a fresh array with the elements from the vector `v` from index `b` (inclusive) until index `e` (exclusive).
+    ///
+    pub def slice(start: {start = Int32}, end: {end = Int32}, v: Vector[a]): Vector[a] = region rc {
+        let arr = toArray(rc, v);
+        Array.slice(rc, start.start, end.end, arr) |> Array.toVector
```
magnus-madsen, line 122: And this also.

main/src/library/Vector.flix
```diff
@@ -0,0 +1,1389 @@
...
+    }
+
+    ///
+    /// Returns the vector `v` as an array.
+    ///
+    pub def toArray(rc: Region[r], v: Vector[a]): Array[a, r] \ Write(r) =
```
magnus-madsen, line 128: Indeed this does not need to be a primitive.

main/src/library/Vector.flix
```diff
@@ -0,0 +1,1389 @@
...
+    ///
+    pub def last(v: Vector[a]): Option[a] =
+        nth(length(v) - 1, v)
+
+    ///
+    /// Return a new vector, appending the elements `b` to elements of `a`.
+    ///
+    pub def append(a: Vector[a], b: Vector[a]): Vector[a] = region rc {
+        buildVector(rc, length(a) + length(b), arr -> {
+            Array.updateSequence!(0, a, arr);
```
magnus-madsen, line 245: This would need to become some more clever use of buildVector reading from a and then from b.

main/src/library/Vector.flix
```diff
@@ -0,0 +1,1389 @@
...
+
+    ///
+    /// Accumulates the result of applying `f` to `v` going left to right.
+    ///
+    /// That is, the result is of the form: `[s , f(s, x1), f(f(s, x1), x2),  ...]`.
+    ///
+    pub def scanLeft(f: (b, a) -> b \ ef, s: b, v: Vector[a]): Vector[b] \ ef = region rc {
+        let acc = ref s @ rc;
+        buildVector(rc, length(v) + 1, arr -> {
+            Array.put(s, 0, arr);
```
magnus-madsen, line 388: There uses of put are still problematic because we will not `Vector.put`.

One option could be to scanLeft into an array and then copy it to a Vector.

main/src/library/Vector.flix
```diff
@@ -0,0 +1,1389 @@
...
+    ///
+    /// Accumulates the result of applying `f` to `xs` going right to left.
+    ///
+    /// That is, the result is of the form: `[..., f(xn-1, f(xn, s)), f(xn, s), s]`.
+    ///
+    pub def scanRight(f: (a, b) -> b \ ef, s: b, v: Vector[a]): Vector[b] \ ef = region rc {
+        let len = length(v);
+        let acc = ref s @ rc;
+        buildVector(rc, len + 1, arr -> {
+            Array.put(s, len, arr);
```
magnus-madsen, line 407: Same comment here.

main/src/library/Vector.flix
```diff
@@ -0,0 +1,1389 @@
...
+        let sepArr = toArray(rc, sep);
+        buildVector(rc, len, arr -> {
+            forEachWithIndex((i, v) ->
+                if (i == 0) {
+                    let a1 = toArray(rc, v);
+                    Array.updateSequence!(0, a1, arr);
+                    pos := length(v)
+                } else {
+                    let ix = deref pos;
+                    Array.updateSequence!(ix, sepArr, arr);
```
magnus-madsen, line 556: Same comment about not using Array/updateSequence.

magnus-madsen:
I think if we get rid of the last usages of `Array.put` then we can probably easily add the native `Vector` type.

For some functions it might be simpler to just use the `Array.X` function and then convert it into a `Vector` (even if that entails a copy).

main/src/library/Vector.flix
stephentetley: I'm in the process of removing `buildVector` - it's doing nothing that a `region` block doesn't do. In a language without regions it would be useful to encapsulate the new array, but that's exactly what regions do so it can be eliminated in Flix.

main/src/library/Vector.flix
stephentetley: It was copied from the Array module, as its been written it's become more of its own.

main/src/library/Vector.flix
stephentetley: `get` is the faster version without out-of-bounds checking.

main/src/library/Vector.flix
stephentetley: There's now a helper function in Vector:

> def arrayUpdateSeqV!(i: Int32, sub: Vector[a], arr: Array[a, r]): Unit \ Write(r)  =

This is `Array.updateSequence` except `sub` is used directly as a vector and not converted into an array.

stephentetley:
I've just push an update to remove `buildVector`

Some of the `toX` functions could be simplified with folds so that is still todo...

main/src/library/Vector.flix
stephentetley: `scanLeft` and `scanRight` now initialize an interim array directly instead of using `buildVector` so they should be clearer as to how they work.

main/src/library/Vector.flix
stephentetley: As per new `scanLeft`

main/src/library/Vector.flix
stephentetley: `buildVector` is now eliminated. `arrayUpdateSeqV!` is used to update the interim array without needing to change `a` and `b` into arrays or cheating (i.e. relying on them being a type alias to array).

[`a` and `b` are now called `v1` and v2` to emphasize they are vectors]

stephentetley:
`toX` functions now updated.

All use of `Array.put` 's should now be identifiable as working on an interim array that is incrementally built and converted to a vector with `toVector`.

magnus-madsen:
Thanks Stephen.

I will merge this and then work can start on the compiler-side. Once that is implement, I / we can "connect" the two. 

Once that is done we can add the missing instances and do a final round of cleanup.

PR #5421, opened 2023-02-03
feat: add `recoverOne` variant

magnus-madsen:


magnus-madsen:
@jaschdoc @mlutze take a look at this PR.

I think this highlights a potential approach.

What I have done is made it explicit the type of error that `weedLiteral` can fail with (`IllegalLiteral`). 

Then I have added `recoverOne` which I *only* use to recover from `IllegalLiteral`.

This makes it explicit the type of error I want to replace with an `Expression.Error`.

magnus-madsen:
(Unfortunately nothing prevents you from recovering from an error that is not there..)

magnus-madsen:
@jaschdoc I think this the approach we should take for now. Does it make sense?

Issue #5408, opened 2023-01-30
Add `Order` instance to `Chain`


jaschdoc:
Related to https://github.com/flix/flix/pull/5358

Add tests for https://github.com/flix/flix/blob/7b3774f621dfdac625b215484c6ffbcc46f11160/main/test/flix/Test.Exp.ForEachYield.flix#L238

magnus-madsen:
Would it be an expensive instance?

stephentetley:
> Would it be an expensive instance?

No I don't think so - I've calculated "viewLeft-ing" on a "cons Chain" and a "snoc Chain". Snoc Chain would be the "bad" case and it works out nicer than I imagined.

There is a slightly more work for a "cons Chain" compared to a List as the head of the Chain is inside a C (Chain) constructor. For a "snoc Chain" there is an initial chase down the spine of the Chain but crucially the intermediate rebuilding within `viewLeft` generates a "cons" Chain out of the "snoc" Chain so proceeding further than the front has no additional "chase downs":

~~~

Left biased formed by consing:
Chain = C(O 1, C(O 2, C(O 3, O 4)) 
Trace of viewLeft:
step1: loop(O 1, acc = C(O 2, C(O 3, O 4))
step2: SomeLeft(1, C(O 2, C(O 3, O 4)))

Right biased - formed by snocing
Chain = C(C(C(O 1, O 2), O 3), O 4)
Trace of viewLeft:
step1: loop(C(C(O 1, O 2), O 3)), acc = O 4)
step2: loop(C(O 1, O 2), acc = C(O 3, O 4))
step3: loop(O 1,  acc = C(O 2, C(O 3, O 4))
step4: SomeLeft(1, C(O 2, C(O 3, O 4)))

~~~

stephentetley:
PR #5422 adds an Order instance

PR #5422, opened 2023-02-03
feat: add Order instance to Chain (issue #5408)

stephentetley:
This PR adds an `Order` instance to `Chain`.

`compare` is provided by a function inside the `Chain` module - `equals` already worked like this and directly implementing `compare` in the instance would be a bit ugly due to needing to use the view constructors with qualified names.


magnus-madsen:
Thanks!

Issue #5393, opened 2023-01-29
Should `slice` functions (Array.slice, String.slice, etc.) have named arguments?


stephentetley:
Should `slice` functions have named arguments? 

Currently Array.slice is:

~~~

pub def slice(r1: Region[r1], b: Int32, e: Int32, a: Array[a, r]): Array[a, r1] \ { Read(r), Write(r1) } =

~~~

However I forgot that the third parameter was "end" and not "length" so had a a bug until I checked the implementation.

~~~

pub def slice(r1: Region[r1], start: {start = Int32}, end: {end = Int32}, a: Array[a, r]): Array[a, r1] \ { Read(r), Write(r1) } =

~~~

magnus-madsen:
I think so yes.

stephentetley:
Thanks Magnus - I'll use named params for Vector and can look at changing the other versions after Vector is done.

stephentetley:
Now a PR... #5427

magnus-madsen:
Fixed in master.

PR #5427, opened 2023-02-04
refactor: slice functions given name parameters for start and end (issue #5393)

stephentetley:
This PR changes `slice` functions to use named params for "start" and "end" - without named params I'd sometimes thought the params were "start" and "length" and this has caused bugs.

The String variants `sliceLeft` and `sliceRight` are also changed.


magnus-madsen:
Thanks!

Issue #5350, opened 2023-01-22
Add region to `MutSet`, `MutList`, etc?


magnus-madsen:
I think maybe `MutSet` and friends should keep a reference to their region. This would make it easier to work with e.g. #5349.

stephentetley:
I'll take a look at this if no-one's started on it.

stephentetley:
PR #5428 

`MutList` and `MutDeque` already had a region in the enum so only `MutMap` and `MutSet` needed work doing.

magnus-madsen:
Fixed right?

PR #5428, opened 2023-02-05
refactor: add region to MutSet and MutMap (issue #5350)

stephentetley:
This PR adds a region into the enum for `MutSet` and `MutMap` (`MutDeque` and `MutList` carried a region already).

It also changes `TestMutSet` and `TestMutMap` to use `rc` as the region name in region blocks.


magnus-madsen:
Thanks again!

Issue #5434, opened 2023-02-05
Refactor `ErasedAst`  to use Intrinsics


magnus-madsen:
Refactor every expression except:

- Var
- Unary
- Binary
- Branch
- JumpTo
- Let
- LetRec
- Scope/ScopeExit
- TryCatch

into their respective `Intrinsic0`, `Intrinsic1`, and `Intrinsic2`. Try to keep the same order within the intrinsics.

magnus-madsen:
@jaschdoc I think I would like you to look into this next.

I know the ticket does not offer much description, but if you look at ErasedAst, maybe it all makes sense?

You can start with a PR where you have just refactored one expression.

magnus-madsen:
I recommend to start with:
- Expression.Is

Then I think we need `IntrinsicN` which uses a list of expressions. Then we can do:
- ApplyDef
- ApplyDefTail
- Tuple
- ArrayLit
- InvokeStaticMethod

magnus-madsen:
@jaschdoc I would like this together with #5026 to be the current priorities. Thanks! :)

PR #5454, opened 2023-02-09
feat: add intrinsics to `ErasedAst`

magnus-madsen:
Related to #5434


magnus-madsen:
@mlutze @JonathanStarup Thoughts? 

Its a bit hard to see in the diff, but I think this will make the infrastructure much easier to maintain.

Especially once we start on algebraic effects.

mlutze:
What's the rule for what becomes an intrinsic and what doesn't?

magnus-madsen:
As much as possible. Maybe not binding forms.

PR #5458, opened 2023-02-11
chore: Disable exception re-throwing

paulbutcher:
As dicussed in #5353, this change disables the exception re-throwing code.


magnus-madsen:
I cannot merge this just yet, we need to resolve #5459, but there I think the immediate solution is the proxy idea.

JonathanStarup (approved):
Looks good to me. The only footprint left in the code is the useless region in the UncaughtExceptionHandler which seems okay. This could drift out of sync long term but other than that

paulbutcher:
@magnus-madsen I think that this can be merged once [#5577](https://github.com/flix/flix/pull/5577) is merged.

paulbutcher:
Can this be merged, @magnus-madsen?

magnus-madsen:
(Just returned from vacation-- sorry for the long delay).

I had some more thoughts on this. 

Exiting the JVM the hard way is problematic because it (potentially) breaks a lot of tooling: (1) playground, (2) VSCode language server, (3) test framework (internal and external). 

I think what we have now is almost what we want: When an exception occurs, we backpropagate it across threads and ultimately it reaches main which shutdown the program. We may just want one tweak:

- We should consider marking threads spawned with `spawn` as "daemon". Thus when `main` exits, the whole program exits. No matter what does threads are doing. This ensures that if _any exception reaches main the whole program terminates_.

JonathanStarup:
> We should consider marking threads spawned with `spawn` as "daemon". Thus when `main` exits, the whole program exits. No matter what does threads are doing. This ensures that if _any exception reaches main the whole program terminates_.

Isn't this what breaks swing programs?

magnus-madsen:
> > We should consider marking threads spawned with `spawn` as "daemon". Thus when `main` exits, the whole program exits. No matter what does threads are doing. This ensures that if _any exception reaches main the whole program terminates_.
> 
> Isn't this what breaks swing programs?

:O

magnus-madsen:
I think re-throwing exceptions back along spawned threads is the right approach. 

If an exception occurs it will bubble all the way to main. Once can then install a handler there for logging etc. 

In any case, we starting work on AE soon, and Project Loom is arriving fall this year.

Once those things happen, we can revise.

Issue #5437, opened 2023-02-06
Refactor `CompletionProvider`


magnus-madsen:
- [x] Split `CompletionProvider` into several providers.
- [x] Refactor each new Provider into a class.

Add the following interface:

```
trait Completer {
  def matches(context: CompletionContext): Bool
  def excluded(context: CompletionContext): Bool
  def getCompletions(...): Iterable[Completion]
}
```

magnus-madsen:
@LukasRonn I think there are two more completers we can extract:

- `HoleCompleter` and 
- `FromErrorCompleter`.

magnus-madsen:
Once that is done, I would like to see these next steps:

Refactor the signature of `Completer` from:

```scala
def getCompletions(implicit context: CompletionContext, flix: Flix, index: Index, root: Option[TypedAst.Root], delta: DeltaContext): Iterable[Completion]
```

to

```
def getCompletions(ctx: CompletionContext)(implicit flix: Flix, index: Index, root: TypedAst.Root): Iterable[Completion]
```

(I have made the `root` non-Optional, hence inside `LanguageServer` we should only call `autoComplete` if there is a `root`.

magnus-madsen:
We should also change `Completion.toCompletionItem` to:

```scala
def toCompletionItem(ctx: CompletionContext)(implicit Flix: flix)
```

so we can get rid of these two objects from the different completion data types.

LukasRonn:
Yes,

I'll start from the top and work my way down. :)

LukasRonn:
When the last PR is merged, all tasks regarding refactoring of `CompletionProvider` should be done. :)

PR #5472, opened 2023-02-16
refactor: Add `WithCompleter`

LukasRonn:
Added `WithCompleter` for issue https://github.com/flix/flix/issues/5437 .

Also changed all other `Completer`s to return the specific type of `Completion`. `KeywordCompleter` now returns a `KeywordCompletion`, and so forth for the rest.


magnus-madsen:
Thanks!

Issue #5306, opened 2023-01-15
Rework `CompletionProvider` to use `.` dot namespaces


magnus-madsen:
The namespace mechanism has changed, so now we always use `.` as a separator instead of `/`

PR #5473, opened 2023-02-16
fix: rework `CompletionProvider` to use dot namespaces

mR4smussen:
Related to https://github.com/flix/flix/issues/5306


magnus-madsen:
Looks good.

magnus-madsen:
More to be done or should we mark this as non-draft?

mR4smussen:
The current status is that I haven’t tested this initial solution. I’m not completely sure I’m testing this in the right way, which is one of the things I wanted to discuss on **Friday**, @magnus-madsen

mR4smussen:
Testing that we are now suggesting `Fixpoint.AST` seems to work as intended. I think we should merge this @magnus-madsen

PR #5482, opened 2023-02-19
fix: change the default delimiter to work with Android's dex format

Sintrastes:
This patch changes the default delimiter used in Flix from `%` to `$`, as the Android `.dex` file format does not allow `%` in identifiers, and so this prevents Flix jars from being utilized on Android by default.


magnus-madsen:
@JonathanStarup Do you recall why we initially changed this?

JonathanStarup:
> @JonathanStarup Do you recall why we initially changed this?

@magnus-madsen I wrote it some other place, but it was something to do with inner classes begin represented with `$`

Sintrastes:
@JonathanStarup Following the BNF [here](http://www.dre.vanderbilt.edu/~schmidt/android/android-4.0/dalvik/docs/dex-format.html) (see SimpleName), there are a number of different unicode ranges supported.

Obviously a little bit silly (I'm not sure if there's a better options or not, I haven't looked hard enough yet), but I used "🏴" as well as a delimiter, which seemed to work.

Sintrastes:
Maybe there's some mathematical symbol we could use, or an "ascii % lookalike" in the given unicode ranges?

Sintrastes:
Plain ASCII `-` is also available, which I believe shouldn't cause issues so long as Flix does not allow kebab-case identifiers.

magnus-madsen:
Let me think a bit and I will get back to you :)

magnus-madsen:
@JonathanStarup I wonder if we use double $$ decompilers will get confused?

Sintrastes:
@magnus-madsen Any updates on this?

I tried dropping a few flix-generated class files into the decompiler [here](http://www.javadecompilers.com/), trying a few different decompiler options, and I couldn't see anything getting confused as an inner class.

Sintrastes:
Although if @JonathanStarup 's original comment (on the issue itself) more meant that the people reading the bytecode themselves could be confused by `$`, rather than decompilers, I suppose `$$` might be an option.

I could also test out `-` and see if that works.

magnus-madsen:
We know there was some issues with decompilers get confused by `$` (which is used for inner/anon classes by Java).

JonathanStarup:
Id say that being usable for adroid beats formatting in decompilers so if theres no better options perhaps we should just go back to $

magnus-madsen:
Roger. Let us do it.

PR #5485, opened 2023-02-20
refactor: fix jvm tuple string

JonathanStarup:
As found by @Sintrastes, our code relies on some weird coersion, so I've rewritten the toString of tuple, which should also be more null resistant


magnus-madsen:
What toString is this? We have `ToString` and `stringify` right?

magnus-madsen:
Paging @mlutze too: How many "toString"s do we have? :D

mlutze:
We have two general front-end stringifications: ToString and stringify.

But I recall we've already had to do more in the back-end for tuples

JonathanStarup:
> What toString is this? We have `ToString` and `stringify` right?

`stringify` doesn't match on composite things like tuples so it just calls `Objects.toString(x)`, at which point the generated tuple object needs to have a sensible toString implementation, which essentially requires a backend version of `stringify`. So we have a magic toString in the backend also.

This PR is not altering that state of affairs, just correctly using whats there already

magnus-madsen:
Yikes, okay.

PR #5488, opened 2023-02-20
refactor: use `Option[Root]` in `XProvider`

jaschdoc:
- [x] `HighlightProvider`
- [x] `HoverProvider`
- [x] `GotoProvider`

Related to https://github.com/flix/flix/issues/4934


magnus-madsen:
I will merge this, but after I think we can let it rest (if there is anything left), because Lukas will work on this part of the compiler and he might as well do this cleanup.

PR #5492, opened 2023-02-21
refactor: Add ImportCompleters

LukasRonn:
Working on the three different types of ImportCompleters related to https://github.com/flix/flix/issues/5437 .


LukasRonn:
RFR.
Added the three types of ImportCompleter, and fixed some errors in CompletionProvider due to type mismatch.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
```diff
@@ -118,4 +127,32 @@ object Completion {
...
+  case class ImportNewCompletion(name: String, priority: String, textEdit: TextEdit) extends Completion
+
+  /**
+    * Represents an importMethod completion (java methods)
+    *
+    * @param name     the name of the completion.
+    * @param priority the priority of the completion.
+    * @param textEdit the edit which is applied to a document when selecting this completion.
+    */
+  case class ImportMethodCompletion(name: String, priority: String, textEdit: TextEdit) extends Completion
```
magnus-madsen, line 148: Why not store the actual field and method?

magnus-madsen: java.lang.reflect.Field
java.lang.reflect.Method

and so forth

LukasRonn: And then just handle the completions in `Completion`, when it needs to be converted to a `CompletionItem`?

magnus-madsen: The point is to avoid strings and use richer types. If you are completing a field it would be better to have that field.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
```diff
@@ -118,4 +158,32 @@ object Completion {
...
+    */
+  case class ImportNewCompletion(constructor: Constructor[_], clazz: String, aliasSuggestion: Option[String], context: CompletionContext) extends Completion
+
+  /**
+    * Represents an importMethod completion (java methods)
+    *
+    * @param method  the method.
+    * @param clazz   clazz is the clazz in string form.
+    */
+  case class ImportMethodCompletion(method: Method, clazz: String, context: CompletionContext) extends Completion
```
magnus-madsen, line 178: What's the class? Could it not also be the object? e.g. java.lang.reflect.Class?

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -813,80 +805,9 @@ object CompletionProvider {
...
-    CompletionItem(
-      label = label,
-      // Prioritise executables with fewer parameters.
-      sortText = Priority.high(s"${exec.getParameterCount()}$label"),
-      textEdit = TextEdit(context.range, replace),
-      documentation = None,
-      insertTextFormat = InsertTextFormat.Snippet,
-      kind = CompletionItemKind.Method)
+  private def getImportCompletions()(implicit context: CompletionContext, index: Index, root: Option[TypedAst.Root], delta: DeltaContext): Iterable[CompletionItem] = {
+    if (root.isEmpty) Nil else (ImportNewCompleter.getCompletions ++ ImportMethodCompleter.getCompletions map (comp => comp.toCompletionItem)) ++
```
magnus-madsen, line 809: Instead of calling ++ and then toCompletionItem multiple times, you can first call ++ and then at the very end map and call toCompletionItem.

magnus-madsen: Also, this is just a online method, so why not inline it where it is used?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
LukasRonn: I don't know how to resolve this, since the clazz comes from the regex and I'm not quite sure how that works or what it does..

If we take the java.lang.reflect.Class object, and call its method `getName()`, I'm not sure that we get the desired string that we need as `label` in `CompletionItem`.

magnus-madsen: I see okay.

PR #5493, opened 2023-02-21
feat: disallow capture of region variables in spawn

magnus-madsen:


magnus-madsen:
We need to allow `Region[r]` and `Channel[t, r]` to escape.

PR #5496, opened 2023-02-22
refactor: return `Validation[Exp, ...]` in `Regions` phase

jaschdoc:
@magnus-madsen Is this the right direction?


magnus-madsen:
This could be the right approach, but I realize it may not be necessary... so stand by!

magnus-madsen:
This is not the right approach. My bad! We will something much simpler.

jaschdoc:
No worries! This was meant as an exploratory PR

Issue #5474, opened 2023-02-17
Rename `liftX` functions to `mapN` functions


magnus-madsen:

magnus-madsen:
@stephentetley Would you be interested in this? (Otherwise Ill ask Jakob).

stephentetley:
Yes - I'm happy to look at it.

magnus-madsen:
Fixed in master.

PR #5507, opened 2023-02-22
refactor: naming changes - liftN families of functions are now mapN (issue #5474)

stephentetley:
This PR changes the naming of "liftN" families of functions to be "mapN".

The lift families are changed in:
Applicative
Boxable
List
Option 
Result
Validation

The Boxable changes mandate a change to `Lowering.scala` in the compiler. The changes to the lines with `mkDefnSym` are necessary, the other changes are cosmetic so the scala functions aren't called `liftX` and `liftXb` when they generate code that now uses `Boxable.map`. 

To be honest I'm not sure if the cosmetic changes are worthwhile - I'm happy to revert them...

### DIFF
--- a/main/src/library/Applicative.flix
+++ b/main/src/library/Applicative.flix
@@ -28,11 +28,11 @@
 /// latter case a function is applied. A `List[a -> b]` is an applicative that contains a list of functions,
 /// which are all to be applied to all arguments contained in the arguments list.
 ///
-/// A minimal implementation must define `point` and at least one of `ap` and `liftA2` (if `liftA2` is implemented,
-/// `ap` can be defined based on `liftA2` as shown below). If both `ap` and `liftA2` are defined,
+/// A minimal implementation must define `point` and at least one of `ap` and `mapA2` (if `mapA2` is implemented,
+/// `ap` can be defined based on `mapA2` as shown below). If both `ap` and `mapA2` are defined,
 /// they must be equivalent to their default definitions:
-///     `ap(f: m[a -> b & e], x: m[a]): m[b] \ ef = liftA2(identity, f, x)`
-///     `liftA2(f: a -> b -> c & e, x: m[a], y: m[b]): m[c] \ ef = ap(Functor.map(f, x), y)`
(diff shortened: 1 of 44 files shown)


magnus-madsen:
Fast service!

I think we should leave the `Boxable` functions alone because their signatures are different from the applicative lifts. In particular, they don't take any arguments like the `mapN` functions.

stephentetley:
I'll revert all of `Boxable` this evening after work.

stephentetley:
`Boxable` should now be reverted if the tests pass.

main/src/library/Applicative.flix
```diff
@@ -50,28 +50,28 @@ pub lawful class Applicative[m : Type -> Type] with Functor[m] {
     /// Instances can define more efficient implementations than the default implementation
     /// (which is `Applicative.ap(Functor.map(f, x1), x2)`).
     ///
-    pub def liftA2(f: t1 -> t2 -> r \ ef, x1: m[t1], x2: m[t2]): m[r] \ ef = Applicative.ap(Functor.map(f, x1), x2)
+    pub def mapA2(f: t1 -> t2 -> r \ ef, x1: m[t1], x2: m[t2]): m[r] \ ef = Applicative.ap(Functor.map(f, x1), x2)
```
magnus-madsen, line 53: I think these should be `map2` without the `A` (at least to follow the Scala cats naming we are using).

stephentetley: Right - will fix.

stephentetley: There's also a `lift` function in `Functor` that is just `Functor.map`.

Do we want to remove it - or look at it another time?

Finally `Functor.as` shares its name with a keyword - would it be better called `Functor.replace`?

magnus-madsen: Yes to both i think

stephentetley: Both are done in the latest commit.

magnus-madsen:
Fixes #5474

magnus-madsen:
Thanks!

PR #5516, opened 2023-02-23
refactor: move exps to `ErasedAst.IntrinsicOperator0`

jaschdoc:
Related to https://github.com/flix/flix/issues/5434

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/ErasedAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/ErasedAst.scala
@@ -97,6 +97,8 @@ object ErasedAst {
 
     case class InvokeStaticMethod(method: Method, args: List[ErasedAst.Expression], tpe: MonoType, loc: SourceLocation) extends Expression
 
+    case class NewObject(name: String, clazz: java.lang.Class[_], tpe: MonoType, methods: List[ErasedAst.JvmMethod], loc: SourceLocation) extends Expression
+
     case class Intrinsic0(op: ErasedAst.IntrinsicOperator0, tpe: MonoType, loc: SourceLocation) extends Expression
 
     case class Intrinsic1(op: ErasedAst.IntrinsicOperator1, exp: ErasedAst.Expression, tpe: MonoType, loc: SourceLocation) extends Expression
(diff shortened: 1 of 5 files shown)


main/src/ca/uwaterloo/flix/language/ast/ErasedAst.scala
```diff
@@ -110,10 +106,14 @@ object ErasedAst {
 
   object IntrinsicOperator0 {
 
+    case object Region extends IntrinsicOperator0
+
     case object RecordEmpty extends IntrinsicOperator0
 
     case class GetStaticField(field: Field) extends IntrinsicOperator0
 
+    case class NewObject(name: String, clazz: java.lang.Class[_], methods: List[ErasedAst.JvmMethod]) extends IntrinsicOperator0
```
magnus-madsen, line 115: Do we know what name is? Is it the same as the name of the class?

jaschdoc: Yes it's used in JvmName: https://github.com/flix/flix/blob/25a47b7acb59c6bf3abf23e0e59566ec8c152c9c/main/src/ca/uwaterloo/flix/language/phase/jvm/JvmName.scala#L130

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmOps.scala
```diff
@@ -581,12 +576,26 @@ object JvmOps {
...
+
+        case IntrinsicOperator0.Cst(_) => Set.empty
+
+        case IntrinsicOperator0.Region => Set.empty
+
+        case IntrinsicOperator0.RecordEmpty => Set.empty
+
+        case IntrinsicOperator0.GetStaticField(_) => Set.empty
+
+        case IntrinsicOperator0.NewObject(_, _, methods) =>
```
magnus-madsen, line 589: I take it that this is necessary for the code to compile?

magnus-madsen: (which kind of breaks the idea that we should not have to match in each intrinsic)

magnus-madsen (commented):
I suggest to move `NewObject` back into `Expression` -- for now.

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmOps.scala
magnus-madsen: I might be able to pre-compute `ClosureInfo` thus solving this.

magnus-madsen: I was able to precompute closure info

magnus-madsen:
Yeah, we still have to keep NewObject in expression, at least for the moment.

jaschdoc:
RFR

PR #5523, opened 2023-02-25
fix: `run` arguments

### LINKED ISSUE #5521: `Environment.getArgs` doesn't work when running with `run` command

JonathanStarup:
Fixes #5521


JonathanStarup:
@mlutze can you think of any reason that `Console.withOut ...` wouldn't catch the output of the flix main?

mlutze:
Maybe if running Flix inside scala starts a new "process" of some sort where stdout is reset... but I don't know.

JonathanStarup:
Ready to merge, but as untested as the other packager things

magnus-madsen:
I don't think the bug has been fixed, but the PR has to be redone because we made changes.

PR #5524, opened 2023-02-25
refactor: move exps to `ErasedAst.IntrinsicOperatorN`

jaschdoc:
- [x] Add rest of expressions to `IntrinsicOperatorN`

Closes https://github.com/flix/flix/issues/5434


main/src/ca/uwaterloo/flix/language/ast/ErasedAst.scala
```diff
@@ -216,6 +206,24 @@ object ErasedAst {
 
   }
 
+  sealed trait IntrinsicOperatorN
+
+  object IntrinsicOperatorN {
+
+    case class Closure(sym: Symbol.DefnSym) extends IntrinsicOperatorN
+
+    case class ApplyClo(exp: ErasedAst.Expression) extends IntrinsicOperatorN
```
magnus-madsen, line 215: One thing that is important is that Intrinsics should not contain expression themselves. The point is to keep expression-related stuff in expression and have Intrinsics only contain some meta data.

magnus-madsen: If you look at ApplyClo I think it should really be represented as new intrinsic: `IntrinsicOperator1N` which has the expression:

```
    case class IntrinsicN(op: ErasedAst.IntrinsicOperatorN, exp: Expression, exps: List[ErasedAst.Expression], tpe: MonoType, loc: SourceLocation) extends Expression
```

magnus-madsen (commented):
I suggest to:

- Move `Closure` back into Expression (i.e. leave it as before-- for now).
- Introduce `IntrinsicOperator1N` and use it for ApplyClo and ApplyCloTail

magnus-madsen:
Thanks!

PR #5535, opened 2023-02-28
refactor: Add `DefCompleter`, `SignatureCompleter` and `OpCompleter`

LukasRonn:
Related to https://github.com/flix/flix/issues/5437 .

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
@@ -16,10 +16,11 @@
 package ca.uwaterloo.flix.api.lsp.provider.completion
 
 import ca.uwaterloo.flix.api.Flix
+import ca.uwaterloo.flix.api.lsp.provider.CompletionProvider
 import ca.uwaterloo.flix.api.lsp.provider.CompletionProvider.{Priority, convertJavaClassToFlixType}
 import ca.uwaterloo.flix.api.lsp.{CompletionItem, CompletionItemKind, InsertTextFormat, TextEdit}
-import ca.uwaterloo.flix.language.ast.{Symbol, Type}
-import ca.uwaterloo.flix.language.fmt.FormatType
+import ca.uwaterloo.flix.language.ast.{Symbol, Type, TypedAst}
+import ca.uwaterloo.flix.language.fmt.{FormatScheme, FormatType}
 import ca.uwaterloo.flix.language.ast.Symbol.{EnumSym, TypeAliasSym}
 
 import java.lang.reflect.{Constructor, Executable, Field, Method}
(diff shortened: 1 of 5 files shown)


main/src/ca/uwaterloo/flix/api/lsp/provider/completion/DefCompleter.scala
```diff
@@ -0,0 +1,54 @@
...
+    */
+  override def getCompletions(implicit context: CompletionContext, flix: Flix, index: Index, root: Option[TypedAst.Root], delta: DeltaContext): Iterable[DefCompletion] = {
+    if (root.isEmpty) {
+      return Nil
+    }
+
+    val word = context.word
+    val uri = context.uri
+
+    root.get.defs.values.filter(matchesDef(_, word, uri)).flatMap(decl => Option(Completion.DefCompletion(decl, context, flix)))
```
LukasRonn, line 35: Not sure if it is possible to just use a map instead of flatMap.. This way we remove the 'Option'.

LukasRonn: Yes, I can. 

But I'm unsure of the original purpose of having it as an option, so that's why I didn't change it.

magnus-madsen: I think its senseless, so just change it.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
```diff
@@ -80,6 +81,43 @@ sealed trait Completion {
...
+      CompletionProvider.getApplySnippet(name, decl.spec.fparams)(context).map(snippet => {
+        CompletionItem(label = CompletionProvider.getLabelForNameAndSpec(decl.sym.toString, decl.spec)(flix),
+          sortText = Priority.normal(name),
+          filterText = Some(CompletionProvider.getFilterTextForName(name)),
+          textEdit = TextEdit(context.range, snippet),
+          detail = Some(FormatScheme.formatScheme(decl.spec.declaredScheme)(flix)),
+          documentation = Some(decl.spec.doc.text),
+          insertTextFormat = InsertTextFormat.Snippet,
+          kind = CompletionItemKind.Function)
+      }).orNull
```
magnus-madsen, line 95: Wait. You can't return `null` here. You must never return `null`.

magnus-madsen: Maybe you have to refactor `getApplySnippet` into two functions: One that checks the number of arguments and one that does its computation.

LukasRonn: In that case, what if the required number of arguments isn't provided, what should be returned then?

magnus-madsen: Then the completion does not apply, right?

That's how the old code works, right?

LukasRonn: True.

Hmm, hopefully I'll come up with a neat way to do it and get rid of the `Option`.

LukasRonn:
I really don't know if this is the best way to do it.. Went back to use the `flatMap` to get rid of the all the None's, when the completion doesn't apply.

magnus-madsen:
> I really don't know if this is the best way to do it.. Went back to use the `flatMap` to get rid of the all the `None`'s, when the completion doesn't apply.

Seems reasonable.

magnus-madsen:
Thanks!

PR #5554, opened 2023-03-02
refactor: Add `UseCompleter`

LukasRonn:
Related to https://github.com/flix/flix/issues/5437 .


magnus-madsen:
Thanks!

(Almost there it looks like)

PR #5559, opened 2023-03-03
feat: handle show ast request

mR4smussen:
related to https://github.com/flix/vscode-flix/pull/255

The [ShowAstProvider.showAst](https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/api/lsp/provider/ShowAstProvider.scala#L35) method takes a `phase` argument, which I'm not completely sure what is, and where I can access it? I'm currently faking this argument when i parse the job.


magnus-madsen:
What's probably part of the confusion here is that the intend is to be able to show multiple asts: E.g. we could show the "ParsedAst" or the "TypedAst". This probably means that we need more than a button, e.g. either a list of options, or we need to ask the user to enter the name of the phase (the latter may be preferable).

magnus-madsen:
@JonathanStarup Can you chip in?

JonathanStarup:
The phase is just a string from this list https://github.com/flix/flix/blob/c00b6802f0d7b386fc2ca36edc3de45770b8ee08/main/src/ca/uwaterloo/flix/api/lsp/provider/ListPhasesProvider.scala

magnus-madsen:
This has been superseded by https://github.com/flix/flix/pull/5565

Issue #5410, opened 2023-01-30
Allow parsing 0 for-fragments


jaschdoc:
Any reason this has to be a Parse error rather than Weeder error?

_Originally posted by @mlutze in https://github.com/flix/flix/pull/5358#discussion_r1090040103_
            

The weeder should check that there is at least one valid fragment. This change affects all loop expressions (foreach, for, foreach yield)

PR #5562, opened 2023-03-04
refactor: handle zero for-fragments in Weeder

jaschdoc:
Closes https://github.com/flix/flix/issues/5410


magnus-madsen:
Thanks!

Issue #5217, opened 2023-01-02
Remove `new` and `Newable`


magnus-madsen:
- [ ] Remove the `Newable` type class.
- [ ] Remove `new` syntax.
- [ ] Remove `new` from LSP completion.
- [ ] Remove `new` from VSCode highlight.

mlutze:
What's the reasoning behind this. IMO it's more principled for a syntactic sugar to be tied to a type class.

magnus-madsen:
> What's the reasoning behind this. IMO it's more principled for a syntactic sugar to be tied to a type class.

Good point. Maybe we should just drop the syntax?

magnus-madsen:
Notice that right now in `new Foo(exp)` the `Foo` part has no meaning!!!

magnus-madsen:
This works:

```
def main(): Unit \ IO = 
    region rh {
        let fruits = new MutList(rh);
        MutSet.add!("Apple", fruits);
        MutSet.add!("Pear", fruits);
        MutSet.add!("Mango", fruits);
        MutSet.forEach(println, fruits)
    }
```
:D

mlutze:
Yeah I don't really see a purpose to the syntax. I guess it's nice to have a uniform way to create a data structure?

magnus-madsen:
> Yeah I don't really see a purpose to the syntax. I guess it's nice to have a uniform way to create a data structure?

Cheap parlor trick?

mlutze:
> > Yeah I don't really see a purpose to the syntax. I guess it's nice to have a uniform way to create a data structure?
> 
> Cheap parlor trick?

If we're prioritizing that, we could have a lot of fun during the hackathon...

magnus-madsen:
So the vote is to remove it?

mlutze:
That's my vote.

magnus-madsen:
OK, I updated the ticket.

mlutze:
Note that `new` is still a keyword in the interop syntax.

magnus-madsen:
True that

magnus-madsen:
@jaschdoc Ill give you this one too. Its not urgent, but if you get stuck on other things you can do it.

jaschdoc:
> This works:
> 
> ```
> def main(): Unit \ IO = 
>     region rh {
>         let fruits = new MutList(rh);
>         MutSet.add!("Apple", fruits);
>         MutSet.add!("Pear", fruits);
>         MutSet.add!("Mango", fruits);
[...]

PR #5570, opened 2023-03-04
refactor: remove `new` from LSP completer

jaschdoc:
I'm not sure if there are other lines that should be removed for the LSP part.

Related to https://github.com/flix/flix/issues/5217


magnus-madsen:
We keep it.

PR #5574, opened 2023-03-05
experiment: integrate user-defined effects into the Boolean effect language

magnus-madsen:


magnus-madsen:
@mlutze Thoughts? 

My plan is that this will enable us to:

- Work with the current Bool effect system, allowing us to: Work on the semantics in the backend.
- In parallel: Remove the compiler to have one effect system with a toggle to the set-based once, which should be the future, once we learn how to minimize its formulas.

mlutze:
How do we ensure that those vars are rigid in every situation?

magnus-madsen:
> How do we ensure that those vars are rigid in every situation?

Every time they enter the substitution?

magnus-madsen:
@mlutze This should be possible to port now.

mlutze:
I don't think using rigid vars is the right design. IMO we can achieve the same more cleanly with a dedicated type.

magnus-madsen:
Until the set system.

mlutze:
Yes, I am just not convinced that a constant masquerading as a rigid variable is the best option. I'm afraid it might break assumptions we make about variables. If we instead augment the Boolean translator to handle effect constants (which it did previously) then we should have the same behavior without that compromise.

magnus-madsen:
Okay

PR #5577, opened 2023-03-08
fix: Test output

### LINKED ISSUE #5459: No output from a test which throws an exception in a thread
I'm creating a new issue for this rather than re-opening old issues, as those issues have lots of discussion which is (I think) no longer relevant.

Once #5458 is merged, #5123 becomes a problem again. The situation isn't quite as bad as it used to be: a test which throws an exception in a thread does now fail, but the stack trace is silently swallowed by the Flix test framework. So the programmer knows that something's failed, but has no clue why.

We discussed different solutions to this at length in #5123. I still, very strongly, dislike the test framework redirecting stderr, and still can't think of a reason why this is desirable behaviour (I know that it's been asserted that "people requested this" but I still haven't seen an example of why it would be desirable).

We need a fix for this. The simplest approach would be to avoid redirecting stderr. But if we can't do that, can we please have a conversation about what the desired behaviour is (and why we chose to redirect stderr in the first place) so that we can design a solution that addresses the use case (whatever it is) without hiding critical information from the programmer.

paulbutcher:
Fixes #5459

This PR:

* Adds `TeeOutputStream` and uses it to both immediately output stdout and stderr, and collect the output
* Modifies `runTest` so that both stdout and stderr are printed in the event of a test failing

### DIFF
--- a/main/src/ca/uwaterloo/flix/tools/Tester.scala
+++ b/main/src/ca/uwaterloo/flix/tools/Tester.scala
@@ -190,7 +190,7 @@ object Tester {
           result match {
             case java.lang.Boolean.FALSE =>
               // Case 1: Assertion Error.
-              queue.add(TestEvent.Failure(sym, "Assertion Error" :: Nil, Duration(elapsed)))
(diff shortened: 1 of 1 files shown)


magnus-madsen:
Thanks!

PR #5578, opened 2023-03-08
chore: Return SoftFailure from Instances

### LINKED ISSUE #5502: Rewrite `Instances` to return `List[InstanceError]` and use `SoftFailure`

paulbutcher:
Fixes #5502

I think that this is all that's required here? I'm not sure whether there are any tests I could write that would be relevant?


magnus-madsen:
That's all. This phase was already of the right "shape".

PR #5579, opened 2023-03-08
feat: alpha rename output of `--Xprint-bool-unif`

magnus-madsen:


magnus-madsen:
!!!jar

magnus-madsen:
I think this should solve Q2 and Q3 (with a little bit of extra post-processing).

athas:
Looks good to me.  I'll update the tooling to count syntactically unique queries.

PR #5580, opened 2023-03-08
refactor: rework casts

magnus-madsen:


magnus-madsen:
@jaschdoc I would like to ask you to give a hand with this PR, as your next priority. You can fork this PR and then make a pull request that targets `issue-5142-rework-casts` instead of `master`. 

What I would like is for you to go through the test cases. Everywhere there is a `unsafe_cast foo as ...` we need to refactor it to one one of:

- `checked_cast(e)`
- `checked_ecast(e)`
- `unchecked_check(e as ...)`

We can always use the latter, since it corresponds to the old `unsafe_cast`, but it would be nice if we can make better use of the former two.

jaschdoc:
Roger

magnus-madsen:
@jaschdoc I would like this to be your next priority. I am almost done with the internals.

magnus-madsen:
@jaschdoc I finished internals.

W.r.t. I think we need to do two passes: one were you fix everything which is not specific tests for upcast/supercast/whatever. This is mostly just syntactic changes.

Then in the last pass, I think we have to redo the tests for checked_cast/unchecked_cast by deleting and reworking the existing tests.

magnus-madsen:
@jaschdoc The documentation here now has some more information:

https://doc.flix.dev/checked-casts.html
https://doc.flix.dev/unchecked-casts.html

magnus-madsen:
@stephentetley Just a heads up.

More information here: https://doc.flix.dev/checked-casts.html

stephentetley:
Thanks Magnus - I saw this was pending. 

It looks like it will be a significant improvement over always using `unsafe_cast`.

Issue #5194, opened 2022-12-28
Add `Ref` module


magnus-madsen:
Add a `Ref` module with these functions:

Milestone 1:
- [ ] `Ref.new(rh: Region[r], x: a): Ref[a, r] \ r`
- [ ] `Ref.get(rf: Ref[a]): a \ r`
- [ ] `Ref.put(x: a, rf: Ref[a]): Unit \ r`
- [ ] `Ref.transform(f: a -> a, rf: Ref[a]): Unit \ r`

Milestone 2:
- [ ] Refactor reference tests to use the above API

Milestone 3: (to be decided):
- [ ] Remove reference syntax (?)

(This has low priority, so can be worked on when there is nothing else.)

PR #5584, opened 2023-03-10
feat: add `Ref` module

jaschdoc:
Fixes https://github.com/flix/flix/issues/5194


jaschdoc:
Just community build that's failing.

PR #5586, opened 2023-03-10
draft: Channel close

paulbutcher:
One of the motivations behind implementing ScopeExit (#5353) was to allow channels to be closed. This PR implements channel close, but is still draft because of a number of issues. I'm increasingly unsure whether we want to close channels at all.

The code in this PR raises an error if:

* A channel is used after it's closed
* A channel is closed when it's non-empty

Regarding the first case, I *thought* that a channel could be used after it's closed by using `unsafeSend` to send it to another thread. This is the test I tried to write to demonstrate this:

```
    @test
    def testBufferedChannelUsedAfterClose01(): Bool \ IO = region rc1 {
        let (tx1, rx1) = Channel.buffered(rc1, 1);
        let (tx2, rx2) = Channel.buffered(rc1, 1);
        spawn {
            let tx3 = Channel.recv(rx1);
            let _ = Channel.recv(rx2);
            Channel.send((), tx3)
        } @ rc1;
        region rc2 {
            let (tx3, _) = Channel.buffered(rc2, 1);
            Channel.unsafeSend(tx3, tx1)
        };
        Channel.send((), tx2);
        Thread.sleep(Time/Duration.fromSeconds(1));
        true
    }
```

But this fails to compile with:

```
-- Type Error -------------------------------------------------- main/test/flix/Test.Exp.Concurrency.Buffered.flix

>> The region variable 'rc2' escapes its scope.

233 |         let (tx1, rx1) = Channel.buffered(rc1, 1);
                               ^^^^^^^^^^^^^^^^^^^^^^^^
                               region variable escapes.

The escaping expression has type:

  (Sender[Sender[Unit, rc2], rc1], Receiver[Sender[Unit, rc2], rc1])
```

Which is impressive! But it makes me wonder if there's ever any situation in which a channel could be used after it's closed? Can someone come up with a test that will demonstrate this?

As for the other error case (closing a channel when it's non-empty) that (unsurprisingly) causes a number of our existing tests to fail. They can be fixed, obviously, but I'm not sure whether we *want* this to be an error? If we do, then we probably need to have some kind of "clear channel" function.

The final issue is with the effect of `Closeable.close`. We discussed this [here](https://github.com/flix/flix/pull/5319#discussion_r1071480078) and decided to make it `IO`. But this causes issues for channels, and needs to be casted away in `Channel.buffered`. Of course, this may be moot if we decide that we don't want to do this for channels at all.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/Flix.scala
+++ b/main/src/ca/uwaterloo/flix/api/Flix.scala
@@ -119,6 +119,9 @@ class Flix {
     // Reflect
     "Reflect.flix" -> LocalResource.get("/src/library/Reflect.flix"),
 
+    // Closeable
+    "Closeable.flix" -> LocalResource.get("/src/library/Closeable.flix"),
+
     // Debug
     "Debug.flix" -> LocalResource.get("/src/library/Debug.flix"),
   )
--- a/main/src/ca/uwaterloo/flix/api/Flix.scala
+++ b/main/src/ca/uwaterloo/flix/api/Flix.scala
@@ -122,6 +122,9 @@ class Flix {
     // Closeable
     "Closeable.flix" -> LocalResource.get("/src/library/Closeable.flix"),
 
+    // Region
+    "Region.flix" -> LocalResource.get("/src/library/Region.flix"),
+
     // Debug
     "Debug.flix" -> LocalResource.get("/src/library/Debug.flix"),
   )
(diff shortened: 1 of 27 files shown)


magnus-madsen:
I think this outlines many of the issues with making channels closeable.

I don't think they should be; we don't "close" other mutable collections.

I also think its fine to leave some elements (unused) in a channel. I can imagine several scenarios where such a "lost" message would be acceptable or even intended.

paulbutcher:
I think I agree after trying to go down this route. But I guess you did think that we should at some point given that this was originally part of https://github.com/flix/flix/issues/4850? When did your opinion change?

magnus-madsen:
> I think I agree after trying to go down this route. But I guess you did think that we should at some point given that this was originally part of #4850? When did your opinion change?

It was more of an idea.

Issue #5465, opened 2023-02-12
Result.try - is matching java.lang.RuntimeException too restrictive?


stephentetley:
Should `Result.try`  match `##java.lang.Exception` rather than that the subclass `##java.lang.RuntimeException`?

As the main use case for `Result.try` is wrapping Java API functions I think we want to capture checked exceptions like `IOException` as well as runtime exceptions.

If we want to specifically catch  `##java.lang.RuntimeException` and subclasses I think it should be a second function.

...

A related issue is `try` is already a keyword for try ... catch and effect handlers, maybe we should favour a new name.

stephentetley:
It's probably time I looked at this.

Is there agreement to to match on `##java.lang.Exception` ?

As for renaming `Result.try`, I think my preference is `Result.guard`. Any opinions, agreement, disagreement...

stephentetley:
Thanks Matthew, I'll make a PR and we can discuss naming on that. I thought of and quite liked `tryCatch` as well although `guard` has the benefit of being a single word.

PR #5587, opened 2023-03-10
refactor: change Result.try to catch on Java.Lang.Exception, and rename to Result.guard (issue #5465)

stephentetley:
This PR changes `Result.try` to catch / match on `Java.Lang.Exception` rather than `Java.Lang.RuntimeException` as `Java.Lang.Exception` is the base class of all exceptions.

As `try` is a keyword the PR changes it to `guard`. I'm open to discussion and suggestion of alternative names.

### DIFF
--- a/main/src/library/Result.flix
+++ b/main/src/library/Result.flix
@@ -446,14 +446,14 @@ namespace Result {
     ///
     /// Returns `Ok(x)` if `f` was invoked without throwing an exception.
     ///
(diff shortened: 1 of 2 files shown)


magnus-madsen:
I think we should call it `tryCatch`. Its the functional version of `try { } catch { }`.

stephentetley:
I've renamed `guard` to `tryCatch`,.

magnus-madsen:
Thanks!

PR #5595, opened 2023-03-13
fix: toml example typo?

JonathanStarup:
It looks like a typo but im not sure

its also in the other examples


JonathanStarup:
@magnus-madsen done

PR #5600, opened 2023-03-14
use return type annotations to guide type inference

### LINKED ISSUE #5594: Insert `Expression.Ascribe` for declared return types

mlutze:
fixes #5594. I choose not to inject an expression and instead create a helper function. I feel that this better separates concerns, maintaining our low-sugar policy, and may allow for more specialized error messages in the future.


magnus-madsen:
Maybe we should skip it on the effect part?

mlutze:
> Maybe we should skip it on the effect part?

Why so?

magnus-madsen:
> > Maybe we should skip it on the effect part?
> 
> Why so?

Because of the tests failures? It prevents you from giving a less general type.

magnus-madsen:
Or perhaps something fishy is going on?

mlutze:
> Or perhaps something fishy is going on?

Yes, I think something weird is happening with regions.

mlutze:
> > > Or perhaps something fishy is going on?
> > 
> > 
> > Yes, I think something weird is happening with regions.
> 
> But only in a few cases right?

Yes, with StringBuilder. IIRC these are the same cases that didn't play nicely when we tried to make changes to regions previously.

mlutze:
After taking a closer look, I'm fairly confident this is the problem we have seen before: Non-idempotency of BU substitutions means that purified regions can be re-introduced when a substitution is re-applied.

As a temporary fix, I'm going to disable the pur/eff part of the helper. But I'll make a ticket to address this specifically, since I think it is important to have this for both types and effects.

magnus-madsen:
If you can come up with an example, then you should create a ticket and share it with @JonathanStarup

PR #5605, opened 2023-03-15
refactor: move ambiguity namer errors to redundancy

mlutze:
related to #4967


mlutze:
There are still some errors not yet caught. In particular, `uses` at file headers are not caught because they are thrown away before redundancy. we should maintain these both for Redundancy checks and for LSP support.

PR #5607, opened 2023-03-15
fix: typo in semantic operators

JonathanStarup:


magnus-madsen:
They are not used? How can you get away with renaming them without changes  to other files? 

image

JonathanStarup:
I think theyre partially not used, but where they are used, they are not used in a way that cares about their immediate super type. I think there's a todo from years back in the backend that says that we should move from op to sop

PR #5611, opened 2023-03-17
feat: allow uses in classes

mlutze:
related to #4967

Note that this is a breaking change: `use`s from outside a class are not carried to the inside.


magnus-madsen:
Too many conflicts. Lets take it up later.

PR #5612, opened 2023-03-17
chore: remove of expression

mlutze:


magnus-madsen:
We got rid of `of` in place of `open`, right? And later we may get rid of open too?

mlutze:
Yes, `open` and `open_as` we can remove once we get more efficient BU

PR #5613, opened 2023-03-17
refactor: change use dedicated AST node for laws instead of reusing Def

mlutze:


magnus-madsen:
What's the status? Can I close?

mlutze:
> What's the status? Can I close?

yep

Issue #5489, opened 2023-02-20
Remove `let*` and use `forM` instead


magnus-madsen:
- [ ] Remove use of `let*` in the standard library. Replace it by uses of `forM`.
- [ ] Remove `let*` syntax from the Parser.

See https://doc.flix.dev/monadic-for-yield.html

PR #5615, opened 2023-03-17
refactor: remove `let*`

jaschdoc:
Closes https://github.com/flix/flix/issues/5489


jaschdoc:
Community build fails.

jaschdoc:
@mlutze do you have the authority to rerun pipeline on this PR?

main/src/library/BigInt.flix
```diff
@@ -195,17 +195,15 @@ namespace BigInt {
...
-        let* f32max = Float32.tryToBigInt(Float32.maxValue());
-        if (x < f32min or x > f32max)
-            None
-        else
-            Some(d)
+        forM (
+            f32min <- Float32.tryToBigInt(Float32.minValue());
+            f32max <- Float32.tryToBigInt(Float32.maxValue());
+            if f32min <= x and x <= f32max
+        ) yield d
```
magnus-madsen, line 206: Neat

main/src/library/Console.flix
```diff
@@ -114,14 +114,14 @@ namespace Console {
...
-                let* g = String.slice(start = 2, end = 4, hex) |> stringToHex;
-                let* b = String.slice(start = 4, end = 6, hex) |> stringToHex;
-                Some(rgb((r, g, b), text, isFg = true))
-            } else Option.None
+            forM (
+                hex <- String.stripPrefix(substr = "#", hexCode);
+                r <- String.sliceLeft(end = 2, hex) |> stringToHex;
+                g <- String.slice(start = 2, end = 4, hex) |> stringToHex;
+                b <- String.slice(start = 4, end = 6, hex) |> stringToHex;
+                if String.length(hex) == 6
```
magnus-madsen, line 122: I think this rewrite is wrong. You know check `length == 6` too late-- which means that the slices may not work as intended? In either case, I think the control-flow has been obscured.

magnus-madsen: Possibly it cannot fail because Flix is pretty good with having functions that are total, but I still think the order is un-natural.

magnus-madsen:
@stephentetley Yet another breaking changes. Do you have a lot of `let*`? If so, I might considering leaving in support for a bit while just removing the last usages from the standard library.

main/src/library/Console.flix
magnus-madsen: Perhaps we want:

```
// in Option
def cond(b: Bool): Option[Unit] = if (b) Some(()) else None
```

then you can do:

```
forM(
  _ <- Option.cond(String.length(hex) == 6)
```

magnus-madsen: Thoughts?

stephentetley:
Thanks @magnus-madsen - I'm happy for `let*` to be removed immediately. I'd still have to change code later so I'd rather do it now.

main/src/library/Console.flix
jaschdoc: The `Option.cond` function seems redundant as we already have support for guards in `forM` and if you're not operating on monads you can easily wrap something in a specific monad by just writing `if (b) Some(x) else None`. I think this particular problem is just a matter of reordering where the guard appears.

main/src/library/Console.flix
```diff
@@ -114,14 +114,14 @@ namespace Console {
...
         let text = ToString.toString(s);
         let hexOpt: Option[String] = {
-            use Option.flatMap;
-            let* hex = String.stripPrefix(substr = "#", hexCode);
-            if (String.length(hex) == 6) {
-                let* r = String.sliceLeft(end = 2, hex) |> stringToHex;
-                let* g = String.slice(start = 2, end = 4, hex) |> stringToHex;
-                let* b = String.slice(start = 4, end = 6, hex) |> stringToHex;
-                Some(rgb((r, g, b), text, isFg = true))
-            } else Option.None
```
jaschdoc, line 124: @magnus-madsen

magnus-madsen:
@stephentetley @mlutze This PR removes `let*` thus breaking your code.

(If only you had a dollar for every time that happened... sorry)

jaschdoc:
> Do we have applicative for yet?

Yes

mlutze:
Allowing `=` in `for`s would also be helpful

mlutze:
community builds updated and passing

jaschdoc:
> Allowing `=` in `for`s would also be helpful

As a let-binding? I.e same as in scala

mlutze:
> > Allowing `=` in `for`s would also be helpful
> 
> As a let-binding? I.e same as in scala

Yes

PR #5616, opened 2023-03-17
feat: add ability to derive functor

mlutze:


magnus-madsen:
I am not strongly against this, but I don't want to offer DeriveFunctor without a more thorough discussion of what we should support automatic deriviation of.

Issue #5582, opened 2023-03-09
Refactor `Char.toInt32` and `Char.fromtInt32` to not use `unsafe_cast`


magnus-madsen:
If that is possible?

magnus-madsen:
@mlutze @JonathanStarup I tried to find a Java method that does the trick, but maybe you are better at Googleling than me?

mlutze:
I think there are problems with this since Java uses UTF-16, which means that conversion from an Int is a partial function. Java has `Character.toChars` which seems like a total function since it returns an array of characters. `Character.codePointAt` is the inverse.

magnus-madsen:
I decided to delete them (if possible).

PR #5619, opened 2023-03-18
refactor: remove type casts in `Char.toInt32/fromInt32`

jaschdoc:
Closes https://github.com/flix/flix/issues/5582


magnus-madsen:
What I would like is to:

- [ ] Delete `Char.fromInt32` and `Char.toInt32`. (Because they are dangerous.)
- [ ] For the functions that depend on them, e.g. `isAsciiDigit` re-implement them by either calling Java methods or implement them in Flix. If difficult then report those functions here and we can discuss.

magnus-madsen:
https://hackage.haskell.org/package/base-4.18.0.0/docs/src/GHC.Unicode.html#isAscii

magnus-madsen:
image

Issue #5614, opened 2023-03-17
Add basic support for regexs to the stdlib?


stephentetley:
I thought there already was a issue to add regexs to the standard library but I can't find it...

I've had a library for a couple of years now - `flix-regex` - that wraps Java's `java.util.regex` package but it has never really been nice enough for general use. The sticking point has always been that Java's `java.util.regex.Matcher` is quite imperative and should be wrapped as a monad (or maybe an iterator or effect...) but monad support in Flix is still pretty raw.

However I think we can get a very useful initial regex API for Flix by not exposing `Matcher` at all and providing just two modules: one wrapping `java.util.regex.Pattern` and the other a module of utility functions like Flix's `String` except using `Pattern` instead of `String` in functions like `replace`, `startsWith` etc. :

~~~

pub def split(pattern: {pattern = Pattern}, s: String): List[String]
pub def replaceFirst(from: {from = Pattern}, to: {to = String}, s: String): String
pub def replace(from: {from = Pattern}, to: {to = String}, s: String): String
pub def startsWith(prefix: {prefix = Pattern}, s: String): Bool

~~~

We can add something that wraps `java.util.regex.Matcher` when monads (or effects) are a bit more settled in Flix.

I've `Pattern` and the majority of the String API already written, the latter needs disentangling from `flix-regex` but I hope to have it done over the weekend.

magnus-madsen:
What's the reason that `java.util.regex.Matcher` is stateful?

If it is to cache the compiled state machine then that might be something that could benefit from language support.

I am thinking 
```
let pat = regex"abc*"
```

which instructs the compiler to build a cached `Matcher` for the regex while also checking -- at compile-time -- that the regex is valid. 

I would be happy to merge in a pure API (then later would could do something like the above). 

In terms of functionality, what functions should we offer?

stephentetley:
`java.util.regex.Matcher` is basically an iterator with some fancy additions (and it doesn't quite match Java's `iterator` interface). 

> let pat = regex"abc*"

... should produce a `Pattern` rather than a `Matcher`. `Matcher` needs a `Pattern` and an input string - its position in the input string is the `Matcher` state. `Pattern` is a compiled and presumably optimized regex - the `Pattern` object actually has the factory method for `Matcher` (`Matcher` itself doesn't have a constructor though it could easily have one). 

[...]

PR #5620, opened 2023-03-20
feat: basic regex support (issue #5614)

stephentetley:
This PR adds basic regex support to the stdlib. It provides a `Pattern` module wrapping `java.util.regex.Pattern` and twenty text manipulation functions mostly patterned after functions in Flix's existing `String` module.

Intentionally, it doesn't handle user defined iteration schemes over a matcher (Java's regex library provides a Matcher object which is a mutable iterator like object), or expose "match results" which would be the way to extract capture groups or start / end of the current match. For the time being, these are avoided being because they really need a monad (or an effect) to give them a nice functional interface.

Discussion points:

Maybe `Regex.Text` is not the best name for the String-like functions?

The String-like functions use named arguments (one element records) where the String library does. The Regex versions of the functions enjoy different types (Pattern and String) for the "needle" and the "haystack" that the String library doesn't (String and String). Perhaps the Regex functions don't need named arguments?

Regex.Text API: 

~~~

pub def isMatch(patt: Pattern, s: String): Bool
pub def isSubmatch(patt: Pattern, s: String): Bool
pub def indices(substr: {substr = Pattern}, s: String): List[Int32]
pub def submatches(substr: {substr = Pattern}, s: String): List[String]
pub def countSubmatches(substr: {substr = Pattern}, s: String): Int32
pub def split(pattern: {pattern = Pattern}, s: String): List[String]
pub def replace(from: {from = Pattern}, to: {to = String}, s: String): String
pub def replaceFirst(from: {from = Pattern}, to: {to = String}, s: String): String
pub def startsWith(prefix: {prefix = Pattern}, s: String): Bool
pub def endsWith(suffix: {suffix = Pattern}, s: String): Bool
pub def stripPrefix(substr: {substr = Pattern}, s: String): Option[String]
pub def stripSuffix(substr: {substr = Pattern}, s: String): Option[String]
pub def indexOfFirst(substr: {substr = Pattern}, s: String): Option[Int32]
pub def indexOfLast(substr: {substr = Pattern}, s: String): Option[Int32]
pub def indexOfFirstWithOffset(substr: {substr = Pattern}, offset: {offset = Int32}, s: String): Option[Int32]
pub def indexOfLastWithOffset(substr: {substr = Pattern}, offset: {offset = Int32}, s: String): Option[Int32]
pub def breakOnFirst(substr: {substr = Pattern}, s: String): (String, String)
pub def breakAfterFirst(substr: {substr = Pattern}, s: String): (String, String)
pub def breakOnLast(substr: {substr = Pattern}, s: String): (String, String)
pub def breakBeforeLast(substr: {substr = Pattern}, s: String): (String, String)

~~~

An implementation note - string scanning is different between Regex (subsequent matches, always left-to-right) and the String module (character by character, left-to-right or right-to-left), hence some functions are slightly different between the two modules:

~~~

String.indexOfLeft ~> Regex/Text.indexOfFirst
String.breakOnRight ~> Regex/Text.breakOnLast

etc.

~~~~


magnus-madsen:
I am wondering whether to reorganize the modules since some functionality that is "private" is currently public.

magnus-madsen:
For example, maybe we should throw everything into `Regex` and add compiler support for `Pattern`.

magnus-madsen:
> > For example, maybe we should throw everything into `Regex` and add compiler support for `Pattern`.
> 
> Yes that would make sense. `Matcher` could be a public module later when we decide whether it should be an interator, effect, monad...

Yes, for now I just want a simple lean interface.

magnus-madsen:
At some point there should be a discussion about what goes into the std lib and what goes into packages, but I think there is space for regex library in the std lib.

stephentetley:
I'm on holiday today, so I've updated the PR to put all the regex code in a single module.

stephentetley:
I've kept the `Pattern` type inside the `Regex` namespace. "Pattern" seemed a too general name to be be at the top level.

stephentetley:
Just pushed a new change to use `forM` rather than `let*`.

magnus-madsen:
Thanks!

PR #5622, opened 2023-03-20
ci: use head commits for community build

mlutze:
The community build runner now checks out the head commit of each repository, instead of getting the latest release.

This change makes it so that community build owners do not have to draft a new release every time a change is made to the Flix compiler. Instead, they can simply add a new commit to their repository.

I think this is nice as it lightens the load of maintaining a community build, and establishes a reasonable guideline for community build owners: the repo head should track Flix development head, and releases can be pinned to Flix releases (instead of halfway between two Flix releases)

### DIFF
--- a/.github/workflows/community-build.yaml
+++ b/.github/workflows/community-build.yaml
@@ -50,25 +50,14 @@ jobs:
         uses: actions/setup-java@v1
         with:
           java-version: 17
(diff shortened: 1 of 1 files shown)


magnus-madsen:
I think this is reasonable. Once things calm down a bit, we may consider switching back to pinned versions.

PR #5624, opened 2023-03-20
chore: add boilerplate for associated types

mlutze:
Boring tests to be written:
- [ ] NameError in assoc type def
- [ ] NameError in assoc type sig
- [ ] ResolutionError in assoc type def
- [ ] ResolutionError in assoc type sig

Interesting tests to be written:
- [ ] Kindchecking assoc type defs
- [ ] Ensuring termination of reduction
- [ ] MissingInstance
- [ ] Generalization/nonentailment/whatever error
- [ ] Positive tests

### DIFF
--- a/main/test/flix/experimental/Test.Dec.AssocType.flix
+++ b/main/test/flix/experimental/Test.Dec.AssocType.flix
@@ -3,6 +3,6 @@ mod Test/Dec/AssocType {
     class Coll[a] {
         type Elem[a]: Type
 
-        pub def contains(x: Elem[a], c: a): Bool = ???
+        pub def contains(x: Coll.Elem[a], c: a): Bool = ???
     }
 }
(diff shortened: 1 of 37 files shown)


magnus-madsen:
"What" should I review?

mlutze:
I don't really think there is anything controversial here. Maybe the AST nodes and their shapes

main/test/flix/experimental/Test.Dec.AssocType.flix
```diff
@@ -0,0 +1,8 @@
+mod Test/Dec/AssocType {
+
+    class Coll[a] {
+        type Elem[a]: Type
```
magnus-madsen, line 4: What's the reasoning for this specific form? I would have thought that it would just be `type Elem` or maybe `type Elem: Type`.

mlutze: This syntax is more precise IMO. Note that `Elem` is not a `Type`, but rather a type-level function `Type -> Type`, which will be used like `Elem[String]`. It also ensures (in case we expand to multi-param associated types) that we know which parameter to use for lookup.

mlutze:
related to #5647

Issue #5647, opened 2023-03-27
Associated types tasks


mlutze:
- [x] Modify unification to maintain deferred constraints
- [x] Replace types during monomorph
- [ ] NameError in assoc type def
- [ ] NameError in assoc type sig
- [ ] ResolutionError in assoc type def
- [ ] ResolutionError in assoc type sig
- [ ] Kindchecking assoc type defs
- [x] Ensuring termination of reduction
- [ ] MissingInstance
- [ ] Generalization/nonentailment/whatever error
- [ ] Positive tests
- [ ] LSP
- [ ] safety check for lowered types
- [ ] no missing assoc types in instances
- [ ] no extra assoc types in instances

PR #5651, opened 2023-03-28
feat: implement unification for associated types

mlutze:
related to #5647

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/unification/Unification.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/unification/Unification.scala
@@ -72,7 +72,7 @@ object Unification {
     * Unifies the two given types `tpe1` and `tpe2`.
     */
   // NB: The order of cases has been determined by code coverage analysis.
-  def unifyTypes(tpe1: Type, tpe2: Type, renv: RigidityEnv)(implicit flix: Flix): Result[Substitution, UnificationError] = (tpe1.kind, tpe2.kind) match {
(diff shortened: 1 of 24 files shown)


mlutze:
@magnus-madsen I'm putting the associated type environment in the unification functions. This will mean a lot of passing this map around. Any concerns about that?

magnus-madsen:
> 

Can you explain more?

mlutze:
All the `unifyM` functions will now take an additional parameter `atenv` which maps associated type symbol to a list of implementations of that associated type. So everywhere unification is done, we will need to pass around this environment.

magnus-madsen:
> All the `unifyM` functions will now take an additional parameter `atenv` which maps associated type symbol to a list of implementations of that associated type. So everywhere unification is done, we will need to pass around this environment.

Can we talk tomorrow offline?

mlutze:
@magnus-madsen RFR

This adds the deferred unifications to the infer monad. Remaining important steps are:
* add syntax for equality constraints
* implement the entailment check

main/src/ca/uwaterloo/flix/language/phase/unification/Unification.scala
```diff
@@ -27,16 +27,16 @@ object Unification {
   /**
     * Unify the two type variables `x` and `y`.
     */
-  private def unifyVars(x: Type.Var, y: Type.Var, renv: RigidityEnv)(implicit flix: Flix): Result[Substitution, UnificationError] = {
+  private def unifyVars(x: Type.Var, y: Type.Var, renv: RigidityEnv)(implicit flix: Flix): Result[(Substitution, List[(Type, Type)]), UnificationError] = {
```
magnus-madsen, line 30: Perhaps we want something richer than (Type, Type). I am almost positive want that!

magnus-madsen: I will let you do that in the next PR.

PR #5686, opened 2023-04-04
feat: check equality constraints in Typer

mlutze:
depends on #5695 

related to #5647

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/unification/AssocTypeSubstitution.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/unification/AssocTypeSubstitution.scala
@@ -1,0 +1,175 @@
+/*
+ * Copyright 2023 Matthew Lutze
+ *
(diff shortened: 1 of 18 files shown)


main/src/ca/uwaterloo/flix/language/phase/unification/AssocTypeSubstitution.scala
```diff
@@ -0,0 +1,175 @@
+/*
```
magnus-madsen, line 1: I would like an introduction to these classes at some point.

magnus-madsen: Is this going to replace Substitution?

mlutze: Maybe a more general form could be used, yes.

Issue #4800, opened 2022-10-31
Rename `namespace` to `mod`


magnus-madsen:
Proposal to rename the `namespace` keyword to `mod`:

Potential advantages:

- [ ] The `mod` keyword is 3 chars wide which is a design goal for keywords.
- [ ] The `mod` keyword would be familiar to Rust users now that we are moving towards a namespace mechanism close to theirs.

Disadvantages:

- [ ] Breaking change for no real reason (bike shedding).
- [ ] We use the word module for what is not "really" a modular in the academic PL sense (see e.g. ML modules).

We can make the change incrementally (i.e. allow both `namespace` and `mod` for a while).

In terms of terminology, we would write:

```
mod A/B/C {
  pub def f(): Int32=...
}
```

then the entire listing is a "module" whereas "A/B/C" is the "namespace of the module". I think such terminology would be useful.

paulbutcher:
If we just want to make it shorter, we could go with `ns`? (modulo similar comments about bikeshedding)

magnus-madsen:
> If we just want to make it shorter, we could go with `ns`? (modulo similar comments about bikeshedding)

Thought about it earlier: Two things:

- We like 3 letter keywords (because they equal 4 spaces with 1 space).
- I like the idea of being able to say `mod A/B/C { }` defines a module (the content) in the namespace `A/B/C`. If we use the keyword `ns` its less clear whether namespace refers to the name `A/B/C` or its content.

magnus-madsen:
@mlutze @JonathanStarup Like it or hate it or no opinion?

mlutze:
Neutral :)

JonathanStarup:
Note that we already have `12 mod 2` (not that's it's close conceptually or syntactically)

magnus-madsen:
We will change to `mod` as part of the other changes.

magnus-madsen:
We need to rename `namespace X/Y/Z` to `mod X.Y.Z` everywhere.
We should also update any uses.

magnus-madsen:
@jaschdoc The goal is to replace `namespace X/Y/Z` by `mod X.Y.Z` in the Flix Standard Library.

According to @mlutze this should just work (TM).

PR #5687, opened 2023-04-04
refactor: rename `namespace` to `mod`

jaschdoc:
This PR removes the `namespace` keyword and `/` namespace separator.

Closes https://github.com/flix/flix/issues/4800


magnus-madsen:
We should also be using `.` as separator instead of `/`. I suggest to do this in the same PR.

jaschdoc:
> We should also be using `.` as separator instead of `/`. I suggest to do this in the same PR.

Working on it! :smile:

jaschdoc:
Should these `/` separators also be `.` (in the key part of the association, of course, i.e., `Fixpoint._ -> LocalResource.get(...)`)?
https://github.com/flix/flix/blob/a9b17efdb3a9cb9b810d82361023f8ba89abc942/main/src/ca/uwaterloo/flix/api/Flix.scala#L213

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
```diff
@@ -162,7 +162,7 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
   object Declarations {
 
     def Namespace: Rule1[ParsedAst.Declaration.Namespace] = rule {
-      optWS ~ SP ~ (keyword("namespace") | keyword("mod")) ~ WS ~ Names.Namespace ~ optWS ~ '{' ~ UsesOrImports ~ Decls ~ optWS ~ '}' ~ SP ~> ParsedAst.Declaration.Namespace
```
magnus-madsen, line 165: Could we restore the Parser; just to give people a bit more time to upgrade themselves. Thanks!

jaschdoc:
@magnus-madsen reverted parser changes

magnus-madsen:
> Reverted parser changes

Thanks,- can you put them in a separate PR? :)

magnus-madsen:
Excellent. Thanks!

Issue #5671, opened 2023-03-31
Refactor `Bootstrap` class to properly use `Result`s


magnus-madsen:
E.g.

- `buildPkg` should have a return type of `Result[Unit, String]` where the string is the error message.

PR #5694, opened 2023-04-05
refactor: use `Validation` in Bootstrap

jaschdoc:
Closes https://github.com/flix/flix/issues/5671


main/src/ca/uwaterloo/flix/Main.scala
```diff
@@ -209,7 +209,7 @@ object Main {
           Bootstrap.bootstrap(cwd)(System.out) match {
             case Result.Ok(bootstrap) =>
               val result = bootstrap.test(o)
-              System.exit(getCode(result))
+              System.exit(toCode(result))
```
magnus-madsen, line 212: But if we are always under an OK branch, we might as well, just call exit(0)

jaschdoc: I think the Bootstrap is Ok but then the method called on the Bootstrap object can also fail (within the Ok branch)

main/src/ca/uwaterloo/flix/Main.scala
```diff
@@ -139,85 +143,97 @@ object Main {
           System.exit(getCode(result))
 
         case Command.Init =>
-          val result = Bootstrap.init(cwd, options)(System.out)
-          System.exit(getCode(result))
+          Bootstrap.init(cwd, options)(System.out) match {
+            case Result.Ok(_) =>
+              System.exit(EXIT_SUCCESS)
+            case Result.Err(e) =>
+              System.err.println(e)
```
jaschdoc, line 150: Is it correct to print these errors? (See `printErrors` below)

magnus-madsen:
I think we need to:

- Switch to `Validation` as the return type.
- Introduce a new error type so we are not returning strings.

@jaschdoc Before any programming, does this seem feasible?

jaschdoc:
Yeah that seems doable

magnus-madsen:
What's the status on this?

main/src/ca/uwaterloo/flix/Main.scala
```diff
@@ -32,6 +32,10 @@ import java.nio.file.Paths
   */
 object Main {
 
+  private val EXIT_SUCCESS = 0
+  private val EXIT_FAILURE = 1
```
magnus-madsen, line 36: I dont like these. Let us just inline them. Everyone should know that 0 is success :)

jaschdoc:
> What's the status on this?

I have to refactor this to Validation but I'll do that soon

jaschdoc:
Now uses Validation and has error types that are not just strings

main/src/ca/uwaterloo/flix/Main.scala
```diff
@@ -258,9 +272,9 @@ object Main {
   /**
     * Extracts the exit code from the given result.
     */
-  private def getCode(result: Result[_, Int]): Int = result match {
+  private def getCode[T, E](result: Result[T, E]): Int = result match {
```
jaschdoc, line 275: Maybe this function should be removed. I'll have to double check if it's still used somewhere

magnus-madsen: Yes, lets remove it.

PR #5700, opened 2023-04-06
chore: experiment with migrating to Scala 3

jaschdoc:
Just submitting this pull request to make the Gradle pipeline work for me, as it's quirky on my own machine.

This is just something I wanted to try out and experiment with for a bit.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -43,6 +43,7 @@ tasks.withType(ScalaCompile) {
     }
     scalaCompileOptions.additionalParameters = [
             "-source:3.0-migration",
+            "-Xignore-scala2-macros"
     ]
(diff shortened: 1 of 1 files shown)


magnus-madsen:
I guess this gets stuck with parboiled?

jaschdoc:
This PR has low priority compared to the other issues btw

magnus-madsen:
> This PR has low priority compared to the other issues btw

Definitely. Please don't spend any time on it. I was just curious where it got stuck. I suspected Parboiled is the issue and we will have to rewrite the Parser and that seems to be what you have just confirmed :)

magnus-madsen:
This will not happen before we have reimplemented the parser.

Issue #5634, opened 2023-03-25
Add Regular Expression Constants


magnus-madsen:
- [ ] Add a new `Ast.Constant.Regex(java.lang.String)` class (after `Str`).
- [ ] Add a new `ParsedAst.Constant.Regex` class.
- [ ] Extend the parser to allow the syntax `regex"string"`.
- [ ] In `Weeder` check that the string is a valid regex (by compiling it with Java).
- [ ] Add a new type constructor `Regex`.
- [ ] In the `Lowering` phase replace `Ast.Constant.Regex` with an `InvokeStatic` to `Pattern.compile`.

(Later we can extend the compiler to cache the result of `Pattern.compile` somewhere.)

magnus-madsen:
@stephentetley Would you be interested in this? (Otherwise I can ask Jakob.)

stephentetley:
Thanks Magnus - I'd be open to trying in about two months, but I really need to get my libraries working again with Flix head so that's my immediate priority.

magnus-madsen:
Of course. Sorry for all the churn. Hopefully the package manager, which arrives soon, will help a bit.

stephentetley:
I've started looking at this and have gone through the parser and weeder so far.

mlutze:
Thanks Stephen! Feel free to open a PR while it's in progress.

stephentetley:
Thanks Matthew, I've just made a WIP PR.

stephentetley:
The PR now compiles and passes tests: #5707

magnus-madsen:
Fixed in master.

PR #5707, opened 2023-04-07
feat: add regex literals (issue #5634)

stephentetley:
Work-in-progress...

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/dbg/FormatConstant.scala
+++ b/main/src/ca/uwaterloo/flix/language/dbg/FormatConstant.scala
@@ -40,5 +40,6 @@ object FormatConstant {
     case Ast.Constant.Int64(lit) => s"${lit}i64"
     case Ast.Constant.BigInt(lit) => s"${lit}ii"
     case Ast.Constant.Str(lit) => "\"" + lit + "\""
+    case Ast.Constant.Regex(lit) => "Regex#\"" + lit + "\""
   }
 }
--- a/main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
+++ b/main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
@@ -680,6 +680,30 @@ object WeederError {
   }
 
   /**
+    * An error raised to indicate that the case of an alias does not match the case of the original value.
+    *
+    * @param patt     the invalid regular expression
+    * @param loc      the location where the error occurred
+    */
+  case class InvalidRegularExpression(patt: String, loc: SourceLocation) extends IllegalLiteral {
+    def summary: String = s"The pattern literal '${patt}' is not a valid regular expression."
+
+    def message(formatter: Formatter): String = {
+      import formatter._
+      s"""${line(kind, source.name)}
+         |>> Invalid regular expression pattern literal.
+         |
+         |${code(loc, "The pattern literal is not a well-formed regular expression.")}
+         |
+         |""".stripMargin
+    }
+
+    def explain(formatter: Formatter): Option[String] = Some({
+      s"A pattern literals must be a valid regular expression."
+    })
+  }
+
+  /**
     * An error raised to indicate an empty interpolated expression (`"${}"`)
     *
     * @param loc the location where the error occurred.
--- a/main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
+++ b/main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
@@ -685,7 +685,7 @@ object WeederError {
     * @param patt     the invalid regular expression
     * @param loc      the location where the error occurred
     */
-  case class InvalidRegularExpression(patt: String, loc: SourceLocation) extends IllegalLiteral {
+  case class InvalidRegularExpression(patt: String, err: String, loc: SourceLocation) extends IllegalLiteral {
     def summary: String = s"The pattern literal '${patt}' is not a valid regular expression."
 
     def message(formatter: Formatter): String = {
@@ -695,6 +695,8 @@ object WeederError {
          |
          |${code(loc, "The pattern literal is not a well-formed regular expression.")}
          |
+         |Pattern compilation error:
+         |${err}
          |""".stripMargin
     }
 
--- a/main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
+++ b/main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
@@ -324,6 +324,33 @@ object WeederError {
   }
 
   /**
+    * An error raised to indicate an illegal regex pattern.
+    *
+    * @param loc the location where the illegal pattern occurs.
+    */
+  case class IllegalRegexPattern(loc: SourceLocation) extends WeederError {
+    def summary: String = "Illegal Regex pattern"
+
+    def message(formatter: Formatter): String = {
+      import formatter._
+      s"""${line(kind, source.name)}
+         |>> Illegal Regex pattern.
+         |
+         |${code(loc, "illegal Regex pattern.")}
+         |""".stripMargin
+    }
+
+    /**
+      * Returns a formatted string with helpful suggestions.
+      */
+    def explain(formatter: Formatter): Option[String] = Some({
+      import formatter._
+      s"${underline("Tip:")} A Regex cannot be used as a pattern. It can be used `if` guard with a predicate from the Regex module, e.g `isMatch` or `isSubmatch`."
+    })
+
+  }
+
+  /**
     * An error raised to indicate an illegal jvm field or method name.
     *
     * @param loc the location of the name.
@@ -701,7 +728,7 @@ object WeederError {
     }
 
     def explain(formatter: Formatter): Option[String] = Some({
-      s"A pattern literals must be a valid regular expression."
+      s"A pattern literal must be a valid regular expression."
     })
   }
 
--- a/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
@@ -1584,11 +1584,7 @@ object Lowering {
     val cls = Class.forName("java.util.regex.Pattern")
     val mtd = cls.getDeclaredMethod("compile", Class.forName("java.lang.String"))
     val exp = LoweredAst.Expression.Cst(Ast.Constant.Str(patt), Type.Str, loc)
-    println(mtd)
-    // val tpe = Type.mkNative(cls, loc)
-    // val tpe = Type.Str
-    val tpe = Type.mkPureArrow(Type.Str, Type.mkNative(cls, loc), loc)
-    // val tpe = Type.Unit
+    val tpe = Type.mkNative(cls, loc)
     val pur = Type.Pure
     val eff = Type.Empty
     LoweredAst.Expression.InvokeStaticMethod(mtd, exp :: Nil, tpe, pur, eff, loc)
--- a/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
@@ -1587,8 +1587,8 @@ object Lowering {
     println(mtd)
     // val tpe = Type.mkNative(cls, loc)
     // val tpe = Type.Str
-    // val tpe = Type.mkPureArrow(Type.Str, Type.mkNative(cls, loc), loc)
-    val tpe = Type.Unit
+    val tpe = Type.mkPureArrow(Type.Str, Type.mkNative(cls, loc), loc)
+    // val tpe = Type.Unit
     val pur = Type.Pure
     val eff = Type.Empty
     LoweredAst.Expression.InvokeStaticMethod(mtd, exp :: Nil, tpe, pur, eff, loc)
--- a/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
@@ -349,10 +349,6 @@ object Lowering {
     * Lowers the given expression `exp0`.
     */
   private def visitExp(exp0: TypedAst.Expression)(implicit root: TypedAst.Root, flix: Flix): LoweredAst.Expression = exp0 match {
-    case TypedAst.Expression.Cst(Ast.Constant.Regex(patt), tpe, loc) =>
-      val t = visitType(tpe)
-      mkRegex(patt, loc)
-
     case TypedAst.Expression.Cst(cst, tpe, loc) =>
       val t = visitType(tpe)
       LoweredAst.Expression.Cst(cst, t, loc)
@@ -1580,15 +1576,15 @@ object Lowering {
   /**
     * Returns an InvokeStaticMethod expression to compile the regex pattern `exp`.
     */
-  private def mkRegex(patt: String, loc: SourceLocation): LoweredAst.Expression = {
-    val cls = Class.forName("java.util.regex.Pattern")
-    val mtd = cls.getDeclaredMethod("compile", Class.forName("java.lang.String"))
-    val exp = LoweredAst.Expression.Cst(Ast.Constant.Str(patt), Type.Str, loc)
-    val tpe = Type.mkNative(cls, loc)
-    val pur = Type.Pure
-    val eff = Type.Empty
-    LoweredAst.Expression.InvokeStaticMethod(mtd, exp :: Nil, tpe, pur, eff, loc)
-  }
+//  private def mkRegex(patt: String, loc: SourceLocation): LoweredAst.Expression = {
+//    val cls = Class.forName("java.util.regex.Pattern")
+//    val mtd = cls.getDeclaredMethod("compile", Class.forName("java.lang.String"))
+//    val exp = LoweredAst.Expression.Cst(Ast.Constant.Str(patt), Type.Str, loc)
+//    val tpe = Type.mkNative(cls, loc)
+//    val pur = Type.Pure
+//    val eff = Type.Empty
+//    LoweredAst.Expression.InvokeStaticMethod(mtd, exp :: Nil, tpe, pur, eff, loc)
+//  }
 
   /**
     * Returns a new `VarSym` for use in a let-binding.
--- a/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
@@ -350,6 +350,10 @@ object Lowering {
     * Lowers the given expression `exp0`.
     */
   private def visitExp(exp0: TypedAst.Expression)(implicit root: TypedAst.Root, flix: Flix): LoweredAst.Expression = exp0 match {
+    case TypedAst.Expression.Cst(Ast.Constant.Regex(patt), tpe, loc) =>
+      val t = visitType(tpe)
+      mkRegex(patt, loc)
+
     case TypedAst.Expression.Cst(cst, tpe, loc) =>
       val t = visitType(tpe)
       LoweredAst.Expression.Cst(cst, t, loc)
@@ -1571,6 +1575,23 @@ object Lowering {
     val pur = Type.Pure
     val eff = Type.Empty
     LoweredAst.Expression.Tuple(exps, tpe, pur, eff, loc)
+  }
+
+  /**
+    * Returns an InvokeStaticMethod expression to compile the regex pattern `exp`.
+    */
+  private def mkRegex(patt: String, loc: SourceLocation): LoweredAst.Expression = {
+    val cls = Class.forName("java.util.regex.Pattern")
+    val mtd = cls.getDeclaredMethod("compile", Class.forName("java.lang.String"))
+    val exp = LoweredAst.Expression.Cst(Ast.Constant.Str(patt), Type.Str, loc)
+    println(mtd)
+    // val tpe = Type.mkNative(cls, loc)
+    // val tpe = Type.Str
+    // val tpe = Type.mkPureArrow(Type.Str, Type.mkNative(cls, loc), loc)
+    val tpe = Type.Unit
+    val pur = Type.Pure
+    val eff = Type.Empty
+    LoweredAst.Expression.InvokeStaticMethod(mtd, exp :: Nil, tpe, pur, eff, loc)
   }
 
   /**
--- a/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
@@ -1574,19 +1574,6 @@ object Lowering {
   }
 
   /**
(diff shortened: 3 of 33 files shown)


stephentetley:
Provided `mkRegex` is doing what I should be doing in `Weeder.flix` , what should the `tpe` argument be to the `InvokeStaticMethod` constructor?

Arrow type of `String -> java.util.regex.Pattern`

Just the answer `java.util.regex.Pattern`

Or, something else?

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1573,6 +1577,23 @@ object Lowering {
...
 
+  /**
+    * Returns an InvokeStaticMethod expression to compile the regex pattern `exp`.
+    */
+  private def mkRegex(patt: String, loc: SourceLocation): LoweredAst.Expression = {
+    val cls = Class.forName("java.util.regex.Pattern")
+    val mtd = cls.getDeclaredMethod("compile", Class.forName("java.lang.String"))
+    val exp = LoweredAst.Expression.Cst(Ast.Constant.Str(patt), Type.Str, loc)
+    println(mtd)
+    // val tpe = Type.mkNative(cls, loc)
```
magnus-madsen, line 1588: I think this should be the return type, i.e. `Type.mkNative` of `cls`.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1573,6 +1577,19 @@ object Lowering {
...
+  private def mkRegex(patt: String, loc: SourceLocation): LoweredAst.Expression = {
+    val cls = Class.forName("java.util.regex.Pattern")
+    val mtd = cls.getDeclaredMethod("compile", Class.forName("java.lang.String"))
+    val exp = LoweredAst.Expression.Cst(Ast.Constant.Str(patt), Type.Str, loc)
+    val tpe = Type.mkNative(cls, loc)
+    val pur = Type.Pure
+    val eff = Type.Empty
+    LoweredAst.Expression.InvokeStaticMethod(mtd, exp :: Nil, tpe, pur, eff, loc)
+  }
+
```
magnus-madsen, line 1592: Ah-- some code is also missing: You need to replace the Regex type with a Native type everywhere it may occur... There should be a `visitType` somewhere.

stephentetley: Thanks Magnus - that was probably my next sticking point as I couldn't define a `ToString` instance.

stephentetley:
Regex literals are now compiling and passing tests.

I've patterned the new `Regex` type after `BigDecimal` / `BigInteger` so the code follows them except there are no SemanticOperators on `Regex`.

I'm not sure that the code gen in GenExpression.scala is entirely correct. I just did the simplest thing I could to invoke the static method `java.util.regex.Pattern.compile`.

main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
```diff
@@ -679,6 +679,30 @@ object WeederError {
 
   }
 
+  /**
+    * An error raised to indicate that the case of an alias does not match the case of the original value.
+    *
+    * @param patt     the invalid regular expression
+    * @param loc      the location where the error occurred
+    */
+  case class InvalidRegularExpression(patt: String, loc: SourceLocation) extends IllegalLiteral {
```
magnus-madsen, line 688: We should add handful of tests to `TestWeeder`.

main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
```diff
@@ -679,6 +679,30 @@ object WeederError {
...
+    def summary: String = s"The pattern literal '${patt}' is not a valid regular expression."
```
magnus-madsen, line 689: Might it not be useful to include the underlying error message from Java? 

Otherwise it might be hard to understand what is broken.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1573,6 +1573,19 @@ object Lowering {
...
+    */
+//  private def mkRegex(patt: String, loc: SourceLocation): LoweredAst.Expression = {
+//    val cls = Class.forName("java.util.regex.Pattern")
+//    val mtd = cls.getDeclaredMethod("compile", Class.forName("java.lang.String"))
+//    val exp = LoweredAst.Expression.Cst(Ast.Constant.Str(patt), Type.Str, loc)
+//    val tpe = Type.mkNative(cls, loc)
+//    val pur = Type.Pure
+//    val eff = Type.Empty
+//    LoweredAst.Expression.InvokeStaticMethod(mtd, exp :: Nil, tpe, pur, eff, loc)
+//  }
```
magnus-madsen, line 1587: Remove commented code.

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
```diff
@@ -442,6 +442,10 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
       SP ~ "\"" ~ zeroOrMore(!("\"" | atomic("${") | atomic("%{")) ~ Chars.CharCode) ~ "\"" ~ SP ~> ParsedAst.Literal.Str
     }
 
+    def Regex: Rule1[ParsedAst.Literal.Regex] = rule {
+      SP ~ "Regex#\"" ~ capture(zeroOrMore(!("\"") ~ CharPredicate.All)) ~ "\"" ~ SP ~> ParsedAst.Literal.Regex
```
magnus-madsen, line 446: OK, I see that `Regex#` is consistent with our other literals. 

My initial inclination was something like `r"foo"`.

Ill explore what other languages do.

magnus-madsen: Does the above handle escape regexs correctly? E.g. backslashes etc?

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -1131,6 +1131,12 @@ object GenExpression {
       addSourceLine(visitor, loc)
       visitor.visitLdcInsn(s)
 
+    case Ast.Constant.Regex(patt) =>
```
magnus-madsen, line 1134: This looks ok.

magnus-madsen: The right thing for us to do is to eliminate Regex (and BigInt and BigDecimal) in `Lowering`.

We want to slim down the backend part of the compiler.

magnus-madsen: But, how about we merge it as is, and then in a subsequent PR you or some else can take up the `Lowering` part again?

I suppose there was some issue.

main/src/library/Regex.flix
```diff
@@ -121,91 +104,90 @@ mod Regex {
...
+    /// Compile the regular expression `regex` into a Regex object (internally
+    /// a Regex is `java.util.regex.Pattern`).
     ///
     /// Returns `Err` if the regex is ill-formed and cannot be compiled.
     ///
     /// The syntax of regular expressions is the same as Java, see the JDK Javadocs
     /// of the class `util.regex.Pattern` for reference.
     ///
-    pub def compile(regex: String): Result[String, Pattern] =
+    pub def compile(regex: String): Result[String, Regex] =
```
magnus-madsen, line 115: Two conflicting thoughts:

- If we drop `compile` then we have guarantee that regexs are always valid (because they must be made with `Regex#` which is checked at compile-time. This avoids all kinds of injection attacks. I know in JS-world this is a big problem.
- If we don't drop it, should we perhaps have `String.toRegex` or similar?

magnus-madsen: We could also name it `compileUnsafeRegex` (or whatever).

magnus-madsen: In any case, the documentation should probably strongly suggest not to use this method, but to write `Regex#` instead.

main/src/library/Regex.flix
```diff
@@ -121,91 +104,90 @@ mod Regex {
...
     }
 
     ///
-    /// Returns `true` if the string `input` is matched by the regular expression pattern `p`
+    /// Returns `true` if the string `input` is matched by the Regex `rx`
     /// at any position within the string `s`.
     ///
-    pub def isSubmatch(patt: Pattern, s: String): Bool = region rc {
-        newMatcher(rc, patt, s) |> find!
+    pub def isSubmatch(rx: Regex, s: String): Bool = region rc {
```
magnus-madsen, line 181: We also use `rx` for `Receiver` channels, but I guess its fine.

magnus-madsen: Certainly `Regex` and `rx` is better than `Pattern` which could be confusing.

main/test/flix/Test.Exp.Regex.flix
```diff
@@ -0,0 +1,9 @@
+mod Test.Exp.Regex {
+
+    @test
+    def testRegex01(): Regex = Regex#""
+
+    @test
+    def testRegex02(): Regex = Regex#".*"
+
```
magnus-madsen, line 8: I was expecting some more tests :)

main/test/flix/Test.Exp.Regex.flix
```diff
@@ -0,0 +1,9 @@
+mod Test.Exp.Regex {
```
magnus-madsen, line 1: Copyright header.

magnus-madsen (commented):
Thanks Stephen-- I added some comments.

main/test/flix/Test.Exp.Regex.flix
stephentetley: Am I free to use functions from Regex.flix at this point?

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
mlutze: Note that anything that messes with types in lowering causes problems with type classes. So moving `BigInt` to Lowering for example would mean that we cannot have an `Add` class for `BigInt`

main/test/flix/Test.Exp.Regex.flix
magnus-madsen: Yes, I think so

main/src/library/Regex.flix
stephentetley: I think we definitely want to a `compileWithFlags` whether or not we call it that.

I like the idea of `String.toRegex` as it fits nicely with how Flix's favours `toX` functions - maybe have both `String.toRegex` and String.toRegexWithFlags` ?

magnus-madsen: > I like the idea of String.toRegex as it fits nicely with how Flix's favours toX functions - maybe have both String.toRegex and String.toRegexWithFlags` ?

Yes

main/src/library/Regex.flix
stephentetley: I thought we might have `rx` elsewhere - I'll change to `rgx`.

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
stephentetley: > Does the above handle escape regexs correctly? E.g. backslashes etc?

I'd be surprised if it does. Now literals are generally working I'll add tests to check...

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
magnus-madsen: :(

We can do it post monomorph then?

stephentetley: I'll leave as is.

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
stephentetley: The above parser code doesn't handle escapes properly - I'm investigating.

main/src/library/Regex.flix
stephentetley: Moved into String.

main/test/flix/Test.Exp.Regex.flix
stephentetley: Added

main/src/library/Regex.flix
stephentetley: Changed to `rgx`

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
stephentetley: Done

main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
stephentetley: Done

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
stephentetley: Removed - Regex is now a type like BigDecimal.

main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
stephentetley: I've added four tests that rightfully fail due to bad literals.

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
stephentetley: The tests in `Test.Exp.Regex` basically check for escaping - all work.

main/test/flix/Test.Exp.Regex.flix
stephentetley: `Test.Exp.Regex` now has twelve tests.

`TestRegex.flix` in `library` has been updated to use regex literals except where we need to compile with flags or use the `unmatchable` expression.

stephentetley:
`regex-literals` now have lots of tests.

magnus-madsen:
@mlutze Will you take a quick look to see if I have forgotten anything.

I think what remains is to settle on the syntax.

magnus-madsen:
- Scala: `"abc".r`
- JavaScript: `/abc/`
- Python: `r'abc'`

Any others?

mlutze:
> - Scala: `"abc".r`
> - JavaScript: `/abc/`
> - Python: `r'abc'`
> - Ruby: `/abc/`
> 
> Any others? (I feel partial to just `r"abc"`.)

I think `r'abc` is just raw string for Python.

Still, I like something with `r` or `regex` in it since it implies extensibility. Maybe we will have compile-time validation of other strings in the future.

mlutze:
> > - Scala: `"abc".r`
> > - JavaScript: `/abc/`
> > - Python: `r'abc'`
> > - Ruby: `/abc/`
> > 
> > Any others? (I feel partial to just `r"abc"`.)
> 
> I think `r'abc` is just raw string for Python.
> 
> Still, I like something with `r` or `regex` in it since it implies extensibility. Maybe we will have compile-time validation of other strings in the future.

Another thing that might be nice would be to allow any delimiter (like latex `\verb#xyz#`) so that nothing has to be escaped.

main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
```diff
@@ -679,6 +679,32 @@ object WeederError {
...
+         |
+         |${code(loc, "The pattern literal is not a well-formed regular expression.")}
+         |
+         |Pattern compilation error:
+         |${err}
+         |""".stripMargin
+    }
+
+    def explain(formatter: Formatter): Option[String] = Some({
+      s"A pattern literals must be a valid regular expression."
```
mlutze, line 704: literals -> literal

mlutze:
!!!jar

main/src/library/String.flix
```diff
@@ -108,6 +110,41 @@ mod String {
...
+        })
+
+
+    ///
+    /// Compile the regular expression `regex` into a Regex with the supplied flags.
+    ///
+    /// Returns `Err` if the regex is ill-formed and cannot be compiled.
+    ///
+    /// Warning: It is strongly recommended to use a Regex literal instead of `toRegexWithFlags`.
+    /// Regex literals are checke at compile time. See the JDK documentation of
```
mlutze, line 138: checke -> checked

main/src/library/String.flix
```diff
@@ -108,6 +110,41 @@ mod String {
...
+    /// Compile the regular expression `regex` into a Regex object (internally
+    /// a Regex is `java.util.regex.Pattern`).
+    ///
+    /// Returns `Err` if the regex is ill-formed and cannot be compiled.
+    ///
+    /// The syntax of regular expressions is the same as Java, see the JDK Javadocs
+    /// of the class `util.regex.Pattern` for reference.
+    ///
+    /// Warning: It is strongly recommended to use a Regex literal instead of `toRegex`.
+    /// Regex literals are checke at compile time.
```
mlutze, line 123: checke -> checked

main/src/ca/uwaterloo/flix/language/phase/Weeder.scala
```diff
@@ -2328,6 +2330,13 @@ object Weeder {
       weedCharSequence(chars) map {
         string => Ast.Constant.Str(string)
       }
+
+    case ParsedAst.Literal.Regex(sp1, chars, sp2) =>
```
mlutze, line 2334: We need to add a check to make sure that we don't see a regex on the left side of a `match` case. See `Weeder.visitPattern`.

mlutze: (Though of course in the future a pattern match using regexes would be very nice)

main/src/ca/uwaterloo/flix/language/phase/PatternExhaustiveness.scala
```diff
@@ -678,6 +683,7 @@ object PatternExhaustiveness {
     case Pattern.Cst(Ast.Constant.Int64(_), _, _) => TyCon.Int64
     case Pattern.Cst(Ast.Constant.BigInt(_), _, _) => TyCon.BigInt
     case Pattern.Cst(Ast.Constant.Str(_), _, _) => TyCon.Str
+    case Pattern.Cst(Ast.Constant.Regex(_), _, _) => TyCon.Regex
```
mlutze, line 686: See other comment about weeding out Regex literals in match cases. This case should be unreachable (as with Null below) and we shouldn't need this TyCon.

mlutze (changes requested):

main/src/library/String.flix
stephentetley: Done

main/src/library/String.flix
stephentetley: Done

main/src/ca/uwaterloo/flix/language/phase/Weeder.scala
stephentetley: I've added another case to the `case ParsedAst.Pattern.Lit` match following the example of `Null` .

main/src/ca/uwaterloo/flix/language/phase/PatternExhaustiveness.scala
stephentetley: I've made the `Ast.Constant.Regex(_),..` case throw an error like the Null case below.

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
```diff
@@ -442,6 +442,10 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
       SP ~ "\"" ~ zeroOrMore(!("\"" | atomic("${") | atomic("%{")) ~ Chars.CharCode) ~ "\"" ~ SP ~> ParsedAst.Literal.Str
     }
 
+    def Regex: Rule1[ParsedAst.Literal.Regex] = rule {
+      SP ~ "Regex#\"" ~ zeroOrMore(!("\"") ~ Chars.CharCode) ~ "\"" ~ SP ~> ParsedAst.Literal.Regex
```
magnus-madsen, line 446: Let us change the syntax to `r"abc"`.

main/src/ca/uwaterloo/flix/language/dbg/FormatConstant.scala
```diff
@@ -40,5 +40,6 @@ object FormatConstant {
     case Ast.Constant.Int64(lit) => s"${lit}i64"
     case Ast.Constant.BigInt(lit) => s"${lit}ii"
     case Ast.Constant.Str(lit) => "\"" + lit + "\""
+    case Ast.Constant.Regex(lit) => "Regex#\"" + lit + "\""
```
magnus-madsen, line 43: -> r"..."

main/src/library/Regex.flix
```diff
@@ -419,12 +371,11 @@ mod Regex {
...
     ///
-    /// Create a Matcher for Pattern `patt` on the source String `input`.
+    /// Create a Matcher for Regex `rgx` on the source String `input`.
     ///
-    def newMatcher(_: Region[r], p: Pattern, s: String): Matcher[r] \ Write(r) =
+    def newMatcher(_: Region[r], rgx: Regex, s: String): Matcher[r] \ Write(r) =
         import java.util.regex.Pattern.matcher(##java.lang.CharSequence): ##java.util.regex.Matcher \ r;
-        let Pattern(p1) = p;
-        Matcher(matcher(p1, unchecked_cast(s as ##java.lang.CharSequence)))
+        Matcher(matcher(rgx, unchecked_cast(s as ##java.lang.CharSequence)))
```
magnus-madsen, line 378: This cannot be a checkedcast?

magnus-madsen (commented):
Added a few more comments, but otherwise looks ready to me.

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
magnus-madsen: Oh. I thought you agreed. What's your proposal?

mlutze: Yes, either one of those is OK for me.

mlutze: (Slight preference for the longer one :) )

magnus-madsen: OK, lets go for the longer one. `regex"foo"`.

stephentetley: Updated to `regex"foo"`.

stephentetley:
The latest push fixes the last two issues.

magnus-madsen:
Thanks Stephen

Issue #5713, opened 2023-04-11
Remove `compileBinaryExpr` in `GenExpression`


jaschdoc:
"Flatten" `compileBinaryExpr` by matching on the semantic operator and generating JVM instructions directly.

- [x] Refactor `compileArithmeticExpr`
- [x] Refactor `compileExponentiateExpr`
- [x] Refactor `compileComparisonExpr`
- [x] Refactor `compileLogicalExpr`
- [x] Refactor `compileBitwiseExpr`
- [ ] Investigate if type hierarchy in `SemanticOperator` can be refactored to have fewer branches. E.g. `Int8.Add` up to `Int32.Add` (inclusive) all share the same JVM instruction, so maybe they can have one immediate common super type.

@magnus-madsen could you add `student-programmer` label and assign this to me?

magnus-madsen:
I think we should able to:

- Inline `readVar`
- Inline `compileUnaryExpr`
- Combine `visitComparisonPrologue` and `visitComparisonEpilogue` into one function that takes two instructions (?) (or one instruction, I am confused).
- 


Can we comment the uses of e.g. `I2D`? What does it do?

jaschdoc:
> Combine `visitComparisonPrologue` and `visitComparisonEpilogue` into one function that takes two instructions (?) (or one instruction, I am confused).

Maybe just a function that takes a lambda that generates the correct instructions (both the method call on the visitor and the number of instructions vary).

magnus-madsen:
> > Combine `visitComparisonPrologue` and `visitComparisonEpilogue` into one function that takes two instructions (?) (or one instruction, I am confused).
> 
> Maybe just a function that takes a lambda that generates the correct instructions (both the method call on the visitor and the number of instructions vary).

I think that would make the code much more complicated to understand.

jaschdoc:
> > > Combine `visitComparisonPrologue` and `visitComparisonEpilogue` into one function that takes two instructions (?) (or one instruction, I am confused).
> > 
> > 
> > Maybe just a function that takes a lambda that generates the correct instructions (both the method call on the visitor and the number of instructions vary).
> 
> I think that would make the code much more complicated to understand.

Okay, so just leave it as is?

magnus-madsen:
Why can it not be a function that takes two arguments?

jaschdoc:
Because the number of instructions "in the middle" vary along with method calls. `BigInt` is quite different from `Char`:

[...]

PR #5715, opened 2023-04-11
refactor: remove `compileLogicalExpr`

jaschdoc:
Related to https://github.com/flix/flix/issues/5713


jaschdoc:
I've checked that it produces the same bytecode as a simple java program.

Issue #5710, opened 2023-04-11
Add GradedFunctor/GradedApplicative/GradedMonad


magnus-madsen:

stephentetley:
Thanks Magnus - I'll extract those classes from my `graded-lib`.

stephentetley:
I've started a draft PR #5710

stephentetley:
I've changed the status of the PR to ready to review.

magnus-madsen:
I suggest these go in a package.

PR #5716, opened 2023-04-11
WIP: add Graded classes for Functor, Applicative and Monad (issue #5710)

stephentetley:
Work in progress...

This is a WIP at the moment and I'm just making a PR so it is visible.

This adds graded versions of `Functor` , `Applicative` and `Monad`. Effects get "fully encapsulated" by the type constructor, e.g. for `GradedFunctor.map` the return type includes `ef1` from the effectful mapping function `f` (in regular Functor `ef1` "floats" to the right hand side of the return type and is not "captured"): 

> pub def map(f: a -> b \ ef1, x: m[ef, a]): m[ef and ef1, b]

[The "capturing" of the effect is the essence of the _grading_ of graded FAMs]

Currently graded FAM type constructors need to respect a fixed argument order - the "element" type is last type parameter (like regular Functor etc.) and the effect type must be next to last. With associated types we should be able to relax the position of the effect type.

My feeling is it would be unlikely to write a graded monad (or graded applicative, graded functor...) in the same module as a regular monad, hence the "library" code in the respective graded modules uses the same names as their regular counterparts.

I haven't written any tests yet...


stephentetley:
Now has tests, including ones using a region effect.

magnus-madsen:
Thanks for the PR :) We plan to discuss this in some detail in relation to our work on assoc types/effects. Ill keep you posted :)

stephentetley:
Thanks Magnus, I wonder if we will need just three graded monads once we have user defined effect handlers?

The three being - one simple CPS / codensity without errors, and two CPS + error monads each with different failure modes. One with backtracking failure where all the effect handlers must be restorable (like reader and pure state) and one with catchable failure where the effects like IO and mutable state cannot be reset.

Possibly async / resumptions will need another monad, the initial three would be in CPS for stack safety.

magnus-madsen:
@stephentetley @mlutze 
I think -- regardless of associated effects and effects and handlers -- that there is space for a graded library. I am not sure if it should go in the standard library, but I think there could be use cases for such a library.

The thing is that even when associated effects arrive there is a tension between eager and lazy evaluation of effects.

We _can_ probably unify e.g. `Functor` and `GradedFunctor` by using a lot of associated effects, but I do not think it is worth it.

Instead, I think `Functor` stays eager, e.g. its signature is:

```
pub def map(f: a -> b \ ef, x: m[a]): m[b] \ ef + Aef
```

where `Aef` is the associated effect of the functor. 

But, the point being, Flix is an eager language and its abstractions are eager. 

If you want to lazy / accumulated effects then you can use `GradedFunctor` -- which has a fundamentally different signature -- built up your effects and then fire them of.

Of course, instead of doing that, you can use effects and handlers, which will probably be the "preferred" idiomatic Flix style.

Thoughts?

magnus-madsen:
I think this could become its own package.

PR #5732, opened 2023-04-14
feat: use Vector in Datalog engine

dignissimus:


magnus-madsen:
Next step is to look at the signatures in `Lowering` and ensure they align.

magnus-madsen:
For example, one place is here:

https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala#L1035

main/src/library/Fixpoint/IndexSelection.flix
```diff
@@ -98,72 +98,72 @@ mod Fixpoint {
...
-        case RelOp.Loop(_) => ??? // TODO
+            (innerOp, Vector.append(test, innerGround))
+        case RelOp.Loop(_) => ??? // TODO;
     }
 
-    def longestPrefixQuery(varQuery: List[BoolExp[v]]): (List[BoolExp[v]], List[BoolExp[v]]) =
-        use Fixpoint.Ram.BoolExp.Eq;
-        use Fixpoint.Ram.RamTerm.{RowLoad, Lit};
+    def longestPrefixQuery(varQuery: Vector[BoolExp[v]]): (Vector[BoolExp[v]], Vector[BoolExp[v]]) =
+        use Fixpoint/Ram.BoolExp.Eq;
```
magnus-madsen, line 136: We should not use `/` as separator anymore. The new separator is always `.`

main/src/library/Fixpoint/IndexSelection.flix
```diff
@@ -98,72 +98,72 @@ mod Fixpoint {
...
-            (innerOp, test ::: innerGround)
-        case RelOp.Loop(_) => ??? // TODO
+            (innerOp, Vector.append(test, innerGround))
+        case RelOp.Loop(_) => ??? // TODO;
     }
 
-    def longestPrefixQuery(varQuery: List[BoolExp[v]]): (List[BoolExp[v]], List[BoolExp[v]]) =
-        use Fixpoint.Ram.BoolExp.Eq;
-        use Fixpoint.Ram.RamTerm.{RowLoad, Lit};
+    def longestPrefixQuery(varQuery: Vector[BoolExp[v]]): (Vector[BoolExp[v]], Vector[BoolExp[v]]) =
```
magnus-madsen, line 135: Perhaps a helper function like this (and others should be left to work on lists).

main/src/library/Fixpoint/Solver.flix
```diff
@@ -14,17 +14,17 @@
...
-    use Fixpoint.Shared.PredSym
-    use Fixpoint.Shared.PredSym.PredSym
+    use Fixpoint/Ast.{Datalog, Constraint, HeadTerm, Denotation}
+    use Fixpoint/Ast.Datalog.{Datalog, Model, Join}
+    use Fixpoint/Ast.Constraint.Constraint
+    use Fixpoint/Ast.HeadPredicate.HeadAtom
+    use Fixpoint/Ast.Denotation.{Relational, Latticenal}
+    use Fixpoint/Ram.RamSym
+    use Fixpoint/Shared.PredSym
+    use Fixpoint/Shared.PredSym.PredSym
```
magnus-madsen, line 28: The above should be reverted to use `mod` and with `.` as separator.

main/src/library/Set.flix
```diff
@@ -501,6 +501,15 @@ mod Set {
...
         foldLeft((acc, x) -> x :: acc, Nil, s)
 
+    ///
+    /// Returns the set `s` as a vector.
+    ///
+pub def toVector(s: Set[a]): Vector[a] = region rc {
+        let arr = Array.new(rc, size(s));
+        forEachWithIndex((i, x) -> Array.put(x, i, arr), s);
+        Array.toVector(arr)
+    }
```
magnus-madsen, line 511: indentation + test cases.

main/src/library/Vector.flix
```diff
@@ -1259,6 +1259,16 @@ mod Vector {
     pub def toSet(v: Vector[a]): Set[a] with Order[a] =
         foldRight(Set.insert, Set.empty(), v)
 
+    ///
+    /// Splits the vector `v` at position `n` into two parts
+    ///
+    pub def splitAt(n: Int32, v: Vector[a]): (Vector[a], Vector[a]) = region rc {
+        let arr = unchecked_cast(v as Array[a, rc]);
```
magnus-madsen, line 1266: We should never be using unchecked casts. It should be possible to implement the same functionality without :)

magnus-madsen (commented):
I added some more comments. Most are small things.

At at I high-level I do wonder if some parts of the Datalog engine are better left to use Lists, but I will get back to you on that.

magnus-madsen:
Proposal: I would split this PR into three smaller PRs:

- One that adds the necessary helper functions to Set.flix
- One that rewrites everything in `Ram` to use `Vector`.
- And finally one that touches everything in `Ast` and `Lowering`.

Does that sound good?

magnus-madsen:
(And sorry for the conflict and asking everything to be redone, but I think that is the best way forward.)

PR #5735, opened 2023-04-14
feat: parallelise operations on Vectors

dignissimus:
Automatically parallelises operations on vectors such as `map` and `count`
# TODO
- [ ] `minimum`
- [ ] `minimumBy`
- [ ] `maximum`
- [ ] `maximumBy`


magnus-madsen:
I think we should start by focusing on just one function, e.g. `parMapWithIndex`. Once we get that right, we can do more functions.

main/src/library/Vector.flix
```diff
@@ -472,17 +487,81 @@ mod Vector {
...
+            };
+            def parTransform(t, left, right) =
+                if (t <= 1) transform(left, right)
+                else {
+                    spawn parTransform((t - 2) / 2, left, (left + right) / 2) @ rc;
+                    spawn parTransform((t - 2) / 2, (left + right) / 2, right) @ rc
+                };
+            parTransform(n, 0, length(v));
+            unchecked_cast(result as Vector[b])
+        } as Vector[b] \ {})
```
magnus-madsen, line 562: I think what I have in mind is that, if the vector is sufficiently large, then we should start `n` threads (corresponding to the number of CPUs) and then each thread would operate on a segment of the vector in parallel. Probably using `par`:

https://doc.flix.dev/parallelism.html

then each thread would return a new vector corresponding to its block which we would concatenate together.

dignissimus: My current approach looks like this
```
pub def parMapWithIndex(n: Int32, f: (Int32, a) -> b, v: Vector[a]): Vector[b] =
    def transform(left, right) = region rc {
        let result = Array.new(rc, right - left);
        def loop(l) =
            if (l < right) {
                Array.put(f(l - left, get(l, v)), l - left, result);
                loop(l + 1)
            }
            else ();
        loop(left);
        Array.toVector(result)
    };
    def parTransform(t, left, right) =
        if (t <= 1) transform(left, right)
        else par (
            v1 <- parTransform((t - 2) / 2, left, (left + right) / 2);
            v2 <- parTransform((t - 2) / 2, (left + right) / 2, right)
        ) yield Vector.append(v1, v2);
    parTransform(n, 0, length(v))
```

magnus-madsen: This is much better, but it still suffers from recursion that constructs a tree.

What we would be doing is creating `n` blocks where if we have `n` threads and the vector has length `m` then each block is `m/n` long (where possibly the last block is shorter than the rest). This way we only end up with very few calls to `Vector.append`. Does that make sense?

dignissimus: Yes, this makes sense. I'll make this change

magnus-madsen:
@dignissimus What's the status on this? Should we pick it up again or let someone else complete it?

dignissimus:
@magnus-madsen I've been a bit busy this summer with internships and competitions - if someone wants to take this on then I'm happy for them to continue this and I'll happily take on something else when I become more available in the future, around September. I've just quickly merged upstream into this branch so that it's compatible with the current source code.

magnus-madsen:
> @magnus-madsen I've been a bit busy this summer with internships and competitions - if someone wants to take this on then I'm happy for them to continue this and I'll happily take on something else when I become more available in the future, around September. I've just quickly merged upstream into this branch so that it's compatible with the current source code.

OK- thanks! I will look around and see if anyone is interested. Or perhaps it will stay till you come back :)

magnus-madsen:
I am closing this. We can pick it up later if there is interest.

PR #5742, opened 2023-04-17
fix: ensure termination for assoc type resolution

mlutze:
related to #5647


magnus-madsen:
Termination seems like a useful property :)

PR #5753, opened 2023-04-20
fix: Only generating `VarCompletion` if it matches prefix

LukasRonn:


main/src/ca/uwaterloo/flix/api/lsp/provider/completion/VarCompleter.scala
```diff
@@ -32,6 +33,20 @@ object VarCompleter extends Completer {
       case Entity.FormalParam(fparam) => Completion.VarCompletion(fparam.sym, fparam.tpe)
     }
 
-    iter.toList
+    iter.toList.filter(comp => matchesVar(comp.sym, context.word))
+  }
+
+  /**
+    * Checks that the varSym matches the prefix
```
magnus-madsen, line 40: What prefix?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/VarCompleter.scala
```diff
@@ -32,6 +33,20 @@ object VarCompleter extends Completer {
...
+    *
+    * @param sym  the varSym.
+    * @param word the current prefix.
+    * @return     true, if the var matches prefix, false otherwise.
+    */
+  private def matchesVar(sym: Symbol.VarSym, word: String): Boolean = {
+    if (word.nonEmpty && word.head.isUpper)
```
magnus-madsen, line 47: Variables cannot be uppercase (??)

magnus-madsen:
What does this change do?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/VarCompleter.scala
LukasRonn: The prefix the user has typed in VSCode

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/VarCompleter.scala
LukasRonn: I was a bit too quick there. That is indeed true.

LukasRonn:
> What does this change do?

This ensures that `VarCompletion`s if the var match the word that the user is currently typing.

LukasRonn:
Example:

```
def bar(): String = 
   let oldString = "Hello";
   let newString = "Hello World!";
   n..
```

When you make a request to autocomplete with the prefix `n`, we are actually also creating a `VarCompletion` for `oldString`, even though it doesn't match. This is generally a problem for ranking `Completion`, if we create some `Completion` that doesn't match the prefix the user is typing. When that is the case, these `Completion` can overshadow the ones we actually need (those who matches prefix) and result in a wrongful boosting of a `Completion` in `CompletionRanker.scala`.

In this case, we won't get a "Best Pick", because `oldString` overshadows, because it appears before `newString` in the list of all possible completions.

PR #5754, opened 2023-04-20
fix: Fixes unintentional behavior in `DefRanker`

LukasRonn:


main/src/ca/uwaterloo/flix/api/lsp/provider/completion/ranker/DefRanker.scala
```diff
@@ -33,18 +33,22 @@ object DefRanker {
     * @return             Some(DefCompletion) if a better completion is possible, else none.
     */
   def findBest(completions: Iterable[Completion], deltaContext: DeltaContext): Option[DefCompletion] = {
-    getDefCompletions(completions).maxByOption(defComp => deltaContext.defs.get(defComp.decl.sym))
+    getDefCompletions(completions)
+      // If the map does not contain any of the keys, it unintentionally returns the first possible def completion.
+      // The filter makes sure, that this does not happen.
```
magnus-madsen, line 38: How can this be right given the description of maxByOption?

https://www.scala-lang.org/api/2.13.3/scala/collection/immutable/List.html#maxByOption[B](f:A=%3EB)(implicitcmp:scala.math.Ordering[B]):Option[A]

LukasRonn: In think it's because, when all possible defs are not contained in the map, it will return the first, since it's comparing `None` with `None`.

The function says, that it returns: "an option value containing the first element of this collection with the smallest value measured by function f with respect to the ordering cmp."

Since there is no "value" other than None, it will just take the first possible `DefCompletion`, as holding the "maximum value".

It will only return non if `getDefCompletions(completions)` is empty.

LukasRonn: I'm not sure if that's the correct explanation, however, that's what I experienced when I tested.

In the test, the function would return `Some(DefCompletion)`, even though no `DefCompletion` had a valid key. Furthermore in that case it would always return the first of all the possible `DefCompletion`s.

magnus-madsen: okay, but then it might be better to use `maxBy` and use ` deltaContext.defs.getOrElse(..., 0)`.

LukasRonn: But that would give the same result as `maxByOption` and return `Some(DefCompletion)` when it should return `None`

magnus-madsen: Is the problem that if there are no def completions **then** we should return `None` and this is not happening?

LukasRonn: No, if we have no def completions, then the function indeed does return `None` as the specification says.

However, if we some possible `DefCompletion`s but none of them are part of the `deltaContext.defs` map, then we should return `None`, and that is not what is currently happening.

PR #5751, opened 2023-04-20
chore: update dependencies

fabianhjr:
Tested locally via `./gradlew check`

Security/CVEs of direct dependencies:

- https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-8908 (Guava)
- https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-11050 (Java-WebSocket)

Scalatest upgrade required a a small rename/import change of

`import org.scalatest.FunSuite` to `import org.scalatest.funsuite.AnyFunSuite`

as per https://www.scalatest.org/release_notes/3.1.0

PR #5757, opened 2023-04-20
chore: update scalactic and scalatest

fabianhjr:
Split from https://github.com/flix/flix/pull/5751

Scalatest upgrade required a a small rename/import change of

`import org.scalatest.FunSuite` to `import org.scalatest.funsuite.AnyFunSuite`

as per https://www.scalatest.org/release_notes/3.1.0


magnus-madsen:
Merged. Thanks!

Issue #5238, opened 2023-01-04
Add `instanceof` binary expression?


stephentetley:
To interface with Java libraries that make heavy use of visitors having something equivalent to Java's `instanceof` would be very useful.

I don't think this would be covered by the new `typematch` operation, am I right? I'm anticipating `instanceof` would have to go all the way down to bytecode.

I've started a branch to investigate...

magnus-madsen:
We can, and yes it probably needs to go down to bytecode.

(And yes, type match is not the right tool.)

stephentetley:
Thanks for the confirmation on typematch, Magnus.

magnus-madsen:
We might want yet another match construct-- to avoid the `check then cast` anti-pattern from Java.

stephentetley:
I've made a draft PR #5760

magnus-madsen:
Fixed in master.

PR #5760, opened 2023-04-20
WIP feat: add instanceof expression (issue #5238)

stephentetley:
Once working, this PR will add an `instanceof` runtime test as per Java's `instanceof`.

I've done the plumbing work of adding an `instanceof` expression but I'm stuck at the code generation and need some help.

Firstly, should I be using something more appropriate than `className: String` throughout the abstract syntax?

Secondly, there is getting the code gen in `GenExpression.scala` to work. Should I be following the example of `IntrinsicOperator1.Is` ? I can see from printing className that it is not swapping dot with slash in the class name.

To test I use either of these to expressions at the Flix prompt:

> "Hello World" instanceof java.lang.String
> 100ii instanceof java.math.BigInteger

Todo after getting codegen to work: write tests and change syntax.


magnus-madsen:
> Once working, this PR will add an `instanceof` runtime test as per Java's `instanceof`.

OK

> I've done the plumbing work of adding an `instanceof` expression but I'm stuck at the code generation and need some help.
> 
> Firstly, should I be using something more appropriate than `className: String` throughout the abstract syntax?

Yes, I think a `Class[T]` object is better.

> Secondly, there is getting the code gen in `GenExpression.scala` to work. Should I be following the example of `IntrinsicOperator1.Is` ? I can see from printing className that it is not swapping dot with slash in the class name.

I think `instanceof ` takes on expression argument, so it should be an `IntrinsicOperator1`, yes.

> To test I use either of these to expressions at the Flix prompt:
> 
> > "Hello World" instanceof java.lang.String
> > 100ii instanceof java.math.BigInteger

Yes

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -261,7 +261,15 @@ object GenExpression {
         visitor.visitFieldInsn(PUTFIELD, className, s"clo$i", JvmOps.getClosureAbstractClassType(m.clo.tpe).toDescriptor)
       }
 
+    case Expr.Instanceof(exp, name, _, loc) =>
+      addSourceLine(visitor, loc)
+      val className = JvmName(ca.uwaterloo.flix.language.phase.jvm.JvmName.RootPackage, name).toInternalName
+      println(className);
+      compileExpression(exp, visitor, currentClass, lenv0, entryPoint)
+      visitor.visitTypeInsn(INSTANCEOF, className.toString)
```
magnus-madsen, line 269: @JonathanStarup Can you help with the right incantations?

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -345,6 +345,9 @@ object Safety {
       case Expression.FixpointProject(_, exp, _, _, _, _) =>
         visit(exp)
 
+      case Expression.Instanceof(exp, _, _, _, _, _) =>
```
magnus-madsen, line 348: We could check that you don't write impossible things like `123 instanceof String`, but maybe it is not worth it?

magnus-madsen:
Silly question: Would this work:

```
def instanceof(s: String, x: a): Bool = 
  import ... java.lang.Object.getClass(): ...;
  import ... java.lang.Class.forName(String): ...;
  import ... java.lang.Objects.equals(...) ...;
  equals(forName(s), getClass(x)
```

Perhaps its too ugly, unworkable, and maybe does not work with subtyping.

main/src/ca/uwaterloo/flix/language/ast/TypedAst.scala
```diff
@@ -313,6 +313,8 @@ object TypedAst {
 
     case class FixpointProject(pred: Name.Pred, exp: TypedAst.Expression, tpe: Type, pur: Type, eff: Type, loc: SourceLocation) extends TypedAst.Expression
 
+    case class Instanceof(exp: TypedAst.Expression, className: String, tpe: Type, pur: Type, eff: Type, loc: SourceLocation) extends TypedAst.Expression
```
mlutze, line 316: We can take a shortcut here:

tpe is just Bool, pur is just exp.pur, and eff is just exp.eff

main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
```diff
@@ -975,6 +975,12 @@ object Kinder {
         case (exp1, exp2) => KindedAst.Expression.FixpointProject(pred, exp1, exp2, Type.freshVar(Kind.Star, loc.asSynthetic), loc)
       }
 
+    case ResolvedAst.Expression.Instanceof(exp0, className, loc) =>
+      val expVal = visitExp(exp0, kenv0, senv, taenv, henv0, root)
+      mapN(expVal) {
+        exp => KindedAst.Expression.Instanceof(exp, className, Type.freshVar(Kind.Star, loc.asSynthetic), loc)
```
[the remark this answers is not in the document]
stephentetley: I've changed all the `Instanceof` constructors so tpe, pur and eff are implicit.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1983,6 +1983,15 @@ object Typer {
           resultEff = Type.mkUnion(eff1, eff2, loc)
         } yield (constrs1 ++ constrs2, resultTyp, resultPur, resultEff)
 
+      case KindedAst.Expression.Instanceof(exp, className, tvar, loc) =>
+        for {
+          (constrs, tpe, pur, eff) <- visitExp(exp)
+          // TODO - do we need to assert the type of `typ`?
+          resultTyp = Type.Bool
```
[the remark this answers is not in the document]
stephentetley: Thanks Matthew, updated.

main/src/ca/uwaterloo/flix/language/ast/TypedAst.scala
stephentetley: `Instanceof` constructors changed.

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
stephentetley: Thanks Jonathan - `instanceof` expressions are now working due to using `Class[_]` and `asm.Type.getInternalName`.

stephentetley:
`instanceof` is now working. Next step - I'll write some tests.

What about syntax? I'm not so keen on it being a named infix expression, but I'm short on ideas for what would be better.

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -1662,6 +1662,12 @@ object Resolver {
             case (e1, e2) => ResolvedAst.Expression.FixpointProject(pred, e1, e2, loc)
           }
 
+        case NamedAst.Expression.Instanceof(exp, className, loc) =>
+          val eVal = visitExp(exp, env0)
+          mapN(eVal) {
+            e => ResolvedAst.Expression.Instanceof(e, className, loc)
```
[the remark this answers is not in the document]
stephentetley: Yes, moved to `Class[_]`.

JonathanStarup:
> Thanks everyone, `instanceof` is now working. Next step - I'll write some tests.
> 
> What about syntax? I'm not so keen on it being a named infix expression, but I'm short on ideas for what would be better.

We could make it an ugly `$builtin$` thing, and then add `Object.instanceOf`

mlutze:
I wonder if it would be nice for the `instanceof` expression to act like a `match` expression rather than just a Boolean check. That way we have the `instanceof`-then-cast pattern rolled into one:

```
instanceof x {
    case y: ##foo.bar.Baz => ...
    case _ => ...
}
```

This way it's sort of a runtime `typematch`.

stephentetley:
Thanks Matthew, I like that.

I was also just about to suggest this with syntax similar to `unchecked_cast` , it would be simpler to implement but isn't as nice:

~~~

if (instanceof("Hello World" as ##java.lang.String)) "String" else "not String"


match o {
    case s if instanceof(s as ##java.lang.String) => println(s ++ "!")
}
              
~~~

magnus-madsen:
May I propose that we start with `exp instanceof Class`. Later, we can decide if we want prettier syntax. I am hoping that `instanceof` will be pretty rare.

magnus-madsen:
What remains to be done in this PR before I should take a detailed look?

stephentetley:
Thanks Magnus, I've changed status from draft as I think it's ready for a review.

I haven't derived a SemanticToken for the class name as per Matthew's suggestion above. I didn't find any similar code to crib from so I would need help with writing this.

I've changed the parser slightly so it should make it harder to write non-classnames on the right-hand side of `instanceof` . 

I haven't written "failing" tests - for Regex literals I added failing tests to `TestWeeder.scala` - would this be the right place for them?

magnus-madsen:
> I haven't written "failing" tests - for Regex literals I added failing tests to TestWeeder.scala - would this be the right place for them?

Depends on why they fail. Why do they fail?

main/src/ca/uwaterloo/flix/language/ast/ParsedAst.scala
```diff
@@ -1223,6 +1223,16 @@ object ParsedAst {
       */
     case class Debug(sp1: SourcePosition, kind: ParsedAst.DebugKind, exp: ParsedAst.Expression, sp2: SourcePosition) extends ParsedAst.Expression
 
+    /**
+      * Instanceof expression (currently a binary op - maybe change the syntax?).
```
magnus-madsen, line 1227: update comment.

stephentetley:
To check mis-typed or missing classnames.

main/src/ca/uwaterloo/flix/language/ast/ParsedAst.scala
```diff
@@ -1223,6 +1223,16 @@ object ParsedAst {
...
+      *
+      * @param exp the expression.
+      * @param name the Java class or interface name.
+      * @param sp2  the position of the last character in the expression.
+      */
+    case class Instanceof(exp: ParsedAst.Expression, name: Name.JavaName, sp2: SourcePosition) extends ParsedAst.Expression
```
magnus-madsen, line 1233: I would like to ask for one annoying change:

If we could keep similar AST nodes grouped together; so I think I would like `Instanceof` (named `InstanceOf`) to appear directly after Ascribe and directly before CheckedCast. This goes both for the ASTs and for all pattern matches.

Sorry! But it helps keep things neat :)

main/src/ca/uwaterloo/flix/language/phase/ClosureConv.scala
```diff
@@ -686,6 +692,10 @@ object ClosureConv {
         val e = visitExp(exp)
         Expression.Force(e, tpe, loc)
 
+      case Expression.Instanceof(exp, clsName, loc) =>
```
magnus-madsen, line 695: The informal convention is:

- If the type is string (i.e. early in the pipeline) it is `className`
- Once the `className` becomes a `Class[T]` object we call it `clazz`. 

(You can look at e.g. `NewObject`)

I believe this change happens between NamedAst and ResolvedAst (i.e. in the Resolver).

magnus-madsen: Thus after `ResolvedAst` it should be `clazz`.

main/src/ca/uwaterloo/flix/language/phase/Inliner.scala
```diff
@@ -744,6 +748,10 @@ object Inliner {
       val e = substituteExp(exp, env0)
       LiftedAst.Expression.Force(e, tpe, loc)
 
+    case OccurrenceAst.Expression.Instanceof(exp, className, loc) =>
```
magnus-madsen, line 751: clazz

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -2152,6 +2156,10 @@ object Lowering {
     case LoweredAst.Expression.Force(exp, tpe, pur, eff, loc) =>
       val e = substExp(exp, subst)
       LoweredAst.Expression.Force(e, tpe, pur, eff, loc)
+
+    case LoweredAst.Expression.Instanceof(exp, className, loc) =>
```
magnus-madsen, line 2160: clazz

main/src/ca/uwaterloo/flix/language/phase/Monomorph.scala
```diff
@@ -608,6 +608,10 @@ object Monomorph {
       case Expression.Force(exp, tpe, pur, eff, loc) =>
         val e = visitExp(exp, env0, subst)
         Expression.Force(e, subst(tpe), pur, eff, loc)
+
+      case Expression.Instanceof(exp, className, loc) =>
```
magnus-madsen, line 612: clazz

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -417,6 +417,11 @@ object OccurrenceAnalyzer {
       val (e, o1) = visitExp(sym0, exp)
       (OccurrenceAst.Expression.Force(e, tpe, loc), o1.increaseSizeByOne())
 
+
+    case Expression.Instanceof(exp, className, loc) =>
```
magnus-madsen, line 421: clazz

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
```diff
@@ -640,7 +640,11 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
     }
 
     def Assign: Rule1[ParsedAst.Expression] = rule {
-      LogicalOr ~ optional(optWS ~ operatorX(":=") ~ optWS ~ LogicalOr ~ SP ~> ParsedAst.Expression.Assign)
+      Instanceof ~ optional(optWS ~ operatorX(":=") ~ optWS ~ LogicalOr ~ SP ~> ParsedAst.Expression.Assign)
+    }
+
+    def Instanceof: Rule1[ParsedAst.Expression] = rule {
```
magnus-madsen, line 646: `InstanceOf`

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -1662,6 +1662,13 @@ object Resolver {
             case (e1, e2) => ResolvedAst.Expression.FixpointProject(pred, e1, e2, loc)
           }
 
+        case NamedAst.Expression.Instanceof(exp, className, loc) =>
+          val eVal = visitExp(exp, env0)
+          val clazzVal = lookupJvmClass(className, loc);
```
magnus-madsen, line 1667: Tests that fail this check would go in `TestResolver`, e.g. if you want a test that does `foo instanceof DoesNotExist`.

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -839,6 +843,10 @@ object Simplifier {
         val e = visitExp(exp)
         SimplifiedAst.Expression.Force(e, tpe, loc)
 
+      case SimplifiedAst.Expression.Instanceof(exp, clsname, loc) =>
```
magnus-madsen, line 846: clazz

main/src/ca/uwaterloo/flix/language/phase/Statistics.scala
```diff
@@ -156,6 +156,7 @@ object Statistics {
       case Expression.FixpointFilter(pred, exp, tpe, pur, eff, loc) => visitExp(exp)
       case Expression.FixpointInject(exp, pred, tpe, pur, eff, loc) => visitExp(exp)
       case Expression.FixpointProject(pred, exp, tpe, pur, eff, loc) => visitExp(exp)
+      case Expression.Instanceof(exp, cls, loc) => visitExp(exp)
```
magnus-madsen, line 159: cls -> _

(We should be better at using _ ourselves).

main/src/ca/uwaterloo/flix/language/phase/Stratifier.scala
```diff
@@ -505,6 +505,11 @@ object Stratifier {
         case e => Expression.FixpointProject(pred, e, tpe, pur, eff, loc)
       }
 
+    case Expression.Instanceof(exp, className, loc) =>
```
magnus-madsen, line 508: clazz

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1981,6 +1981,14 @@ object Typer {
           resultEff = Type.mkUnion(eff1, eff2, loc)
         } yield (constrs1 ++ constrs2, resultTyp, resultPur, resultEff)
 
+      case KindedAst.Expression.Instanceof(exp, className, loc) =>
```
magnus-madsen, line 1984: clazz

main/src/ca/uwaterloo/flix/language/ast/ErasedAst.scala
```diff
@@ -64,6 +64,10 @@ object ErasedAst {
 
     case class NewObject(name: String, clazz: java.lang.Class[_], tpe: MonoType, methods: List[JvmMethod], loc: SourceLocation) extends Expr
 
+    case class Instanceof(exp: Expr, clazz: java.lang.Class[_], loc: SourceLocation) extends Expr {
```
magnus-madsen, line 67: We should make this into an `IntrinsicOperator1` instead of an expression.

(We are reducing the "surface" of expressions in preparation for fine-grain call by value which we need for AE).

stephentetley:
Thanks Magnus, I've reordered the `InstanceOf` constructor and pattern matches in most of the files. 

I'll continue on the review issues in the week,

main/src/ca/uwaterloo/flix/language/ast/ParsedAst.scala
stephentetley: Cut down to just `InstanceOf expression.`

main/src/ca/uwaterloo/flix/language/ast/ParsedAst.scala
stephentetley: All constructors and match cases should be reordered.

main/src/ca/uwaterloo/flix/language/phase/ClosureConv.scala
stephentetley: All names should be changed, I've checked with grep as well.

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
stephentetley: Done in latest commit.

main/src/ca/uwaterloo/flix/language/phase/Statistics.scala
stephentetley: Done

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
stephentetley: Done

main/src/ca/uwaterloo/flix/language/ast/ErasedAst.scala
stephentetley: Done.

You might want to check the order of constructors.

main/src/ca/uwaterloo/flix/language/phase/Stratifier.scala
stephentetley: Done

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
stephentetley: Done

main/src/ca/uwaterloo/flix/language/phase/Inliner.scala
stephentetley: Done

main/src/ca/uwaterloo/flix/language/phase/Monomorph.scala
stephentetley: Done

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
stephentetley: Done

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
stephentetley: Done

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
stephentetley: I missed this, I will add some Resolver tests

stephentetley: I've added a Resolver test, but also a couple of Parser tests. I think there is really only one Resolver error - "does not exist", and the parser will pick up ill-formed names.

stephentetley:
I've think with the latest commit (that adds Resolver and Parser tests) I've addressed all the issues in the review.

PR #5775, opened 2023-04-24
test: add associate type tests

mlutze:
related to #5647. Also fixes a bug.


magnus-madsen:
> Also fixes a bug.

Ah ha!

magnus-madsen:
Do you have other test schemes in mind than this?

PR #5752, opened 2023-04-20
feat: Implemented `TypeEnumRanker`

LukasRonn:

PR #5776, opened 2023-04-24
fix: rework `EnumCompleter`

LukasRonn:
I was not able to filter out Enum's that are located in "other" unknown files? (think they might be part of library?). These do not give any problems regarding the ranking.

However, I believe it is sufficient to just check, if the enum matches the word the user has written, since the Enum's first letter is UpperCase.

This resolves problem with https://github.com/flix/flix/pull/5752 .


magnus-madsen:
Why do you need to filter anything? What is wrong with the current provider?

LukasRonn:
> Why do you need to filter anything? What is wrong with the current provider?

Same issues as with `VarCompletions` (https://github.com/flix/flix/pull/5753), where none-valid `TypeEnumCompletions` are generated. This filters all `TypeEnumCompletions` out, and only provides the ones that match the "word" the user has written. This ensures that the ranking works, and no none-valid `TypeEnumCompletions` can overshadow valid ones.

magnus-madsen:
Does it account for namespaces, e.g. if the user has written `Foo.Color`?

LukasRonn:
> Does it account for namespaces, e.g. if the user has written `Foo.Color`?

Was actually just testing that. Not sure yet.

magnus-madsen:
What's the status on this and what is it supposed to do?

LukasRonn:
> What's the status on this and what is it supposed to do?

I'll get to it in a bit. It is supposed to only suggest enums when the correct nameSpace is provided, and when the enum to user tries to auto-completion should only be the ones provided. Otherwise we have a problem with overshadowing, which results in wrongful boosting by `TypeEnumRanker`.

magnus-madsen:
Do you have a screenshot of an example?

How do this work with the module completion?

LukasRonn:
image

LukasRonn:
image

LukasRonn:
@magnus-madsen Works when no namespace is provided, and when an arbitrary amount of namespaces are nested.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
```diff
@@ -272,13 +272,13 @@ object Completion {
...
-    * @param priority      the priority of the EnumType.
+    * @param nameSuffix    the suffix for the name of the Enum.
+    * @param priority      the priority of the Enum.
     * @param textEdit      the edit which is applied to a document when selecting this completion.
     * @param documentation a human-readable string that represents a doc-comment.
     */
-  case class TypeEnumCompletion(enumSym: EnumSym, nameSuffix: String, priority: String, textEdit: TextEdit,
-                                documentation: Option[String]) extends Completion
+  case class EnumCompletion(enumSym: EnumSym, nameSuffix: String, priority: String, textEdit: TextEdit,
+                            documentation: Option[String]) extends Completion
```
LukasRonn, line 281: This could be cleaned up, however, I think that should be part of another PR, and IF I have time to do it.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
```diff
@@ -70,8 +70,8 @@ sealed trait Completion {
         textEdit = textEdit,
         insertTextFormat = insertTextFormat,
         kind = CompletionItemKind.Enum)
-    case Completion.TypeEnumCompletion(enumSym, nameSuffix, priority, textEdit, documentation) =>
-      CompletionItem(label = s"${enumSym.name}$nameSuffix",
+    case Completion.EnumCompletion(enumSym, nameSuffix, priority, textEdit, documentation) =>
```
magnus-madsen, line 73: What is `nameSuffix`? I don't think we should need that.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
magnus-madsen: Yes, I agree :)

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
```diff
@@ -192,4 +193,11 @@ object CompletionUtils {
     }
   }
 
+  /**
+    * Generates an moduleSym with the given nameSpace.
+    *
+    * @param fqn the fully qualified name as a List[String].
+    * @return    the module symbol for the given fully qualified name.
+    */
+  def mkModuleSym(fqn: List[String]): ModuleSym = new ModuleSym(fqn)
```
magnus-madsen, line 202: This belongs in Symbol.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/EnumCompleter.scala
```diff
@@ -0,0 +1,93 @@
...
+
+object EnumCompleter extends Completer {
+  /**
+    * Returns a List of Completion for enums.
+    */
+  override def getCompletions(context: CompletionContext)(implicit flix: Flix, index: Index, root: TypedAst.Root, delta: DeltaContext): Iterable[EnumCompletion] = {
+    // We use a fqn to lookup in the modules
+    // We therefore need to split the context.word at dots
+    val word = context.word.split('.').toList
+    val (fqn, subWord) = {
```
magnus-madsen, line 34: Why all this subword. Can we not let VSCode take care of it?

magnus-madsen: I would think all of these cases could become one.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/EnumCompleter.scala
```diff
@@ -0,0 +1,93 @@
...
+      }
+    }
+
+    getEnumCompletion(context, fqn, subWord)
+  }
+
+  /**
+    * Get all Enum completions
+    */
+  private def getEnumCompletion(context: CompletionContext, fqn: ModuleSym, subWord: String)(implicit root: TypedAst.Root): Iterable[EnumCompletion] = {
```
magnus-madsen, line 49: We always use symbols prefixed with `Symbol`, so `Symbol.ModuleSym`.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/EnumCompleter.scala
```diff
@@ -0,0 +1,93 @@
...
+    // Use fqn to lookup in modules
+    root.modules.get(fqn) match {
```
magnus-madsen, line 51: This is all very complicated.

Maybe you can do:

```
val enumsInModule = root.modules.getOrElse(fqn, Map.empty)
```

and then go from there.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/EnumCompleter.scala
```diff
@@ -0,0 +1,93 @@
...
+      case None =>
+        // Not a valid module, therefore no enums
+        Nil
+      case Some(syms) =>
+        syms.flatMap {
+          case enumSym: Symbol.EnumSym =>
+            // Collect all enums
+            if (matchesEnum(enumSym, subWord)) {
```
magnus-madsen, line 59: I don't think we want to care about subwords. Can we not just let VSCode do the filtering for us?

magnus-madsen (commented):
Screenshots look good, but I have questions about the complexity of the code.

magnus-madsen (commented):
Screenshots look good, but I have questions about the complexity of the code.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/EnumCompleter.scala
LukasRonn: No, we cannot let VSCode take care of it, because then the `Ranker` won't work as intended.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
LukasRonn: No clue honestly. It was just part of the original structure when creating a `CompletionItem` for enums.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/EnumCompleter.scala
LukasRonn: Sadly no. Would result in wrongful boosting by `Ranker`.

magnus-madsen: hmmmm

PR #5777, opened 2023-04-24
fix: Makes sure that the best-pick completion is always on top

LukasRonn:
Change in priority to ensure that the "bestPick" calculated by `CompletionRanker` is always on top.

This keeps the original order (priority) of all other completions and only affect the "bestPick".


magnus-madsen:
Is there a reason boost could simply not return `0`?

LukasRonn:
> Is there a reason findBest could simply not return `0`?

So the `sortText` for the `CompletionItem` (generated from the "best pick" `Completion`), should have the prefix value `0` instead of `1`?

magnus-madsen:
> > Is there a reason findBest could simply not return `0`?
> 
> So the `sortText` for the `CompletionItem` (generated from the "best pick" `Completion`), should have the prefix value `0` instead of `1`?

Seems reasonable.

LukasRonn:
> > > Is there a reason findBest could simply not return `0`?
> > 
> > 
> > So the `sortText` for the `CompletionItem` (generated from the "best pick" `Completion`), should have the prefix value `0` instead of `1`?
> 
> Seems reasonable.

I'll test it, and make sure it works.

LukasRonn:
> > > > Is there a reason findBest could simply not return `0`?
> > > 
> > > 
> > > So the `sortText` for the `CompletionItem` (generated from the "best pick" `Completion`), should have the prefix value `0` instead of `1`?
> > 
> > 
> > Seems reasonable.
> 
> I'll test it, and make sure it works.

It works with 0.

PR #5779, opened 2023-04-24
feat: Introduce `EnumTagCompleter`

LukasRonn:


LukasRonn:
> Might be a better way to do this.
> 
> Not working exactly like `UseCompleter` when we want to autocomplete Enums. Not able to do it the same way. That being said - this implementation does work.

I found the difference between how `UseCompleter` auto-completes an EnumTag compared to this implementation, and why it differs.

Example (`UseCompleter`):
```
use Color2.r

enum Color2 {
   case Red, Green, Yellow
}
```
Here it is provided with both `Color2.Red` and `Color2.Green`, since `r` is part of both cases.

The `UseCompleter` provides all 3 `UseCompletions` for the tags. `Color2.Red`, `Color2.Green` and `Color2.Yellow`, and VS-Code filters the suggestions (Through LSP and `CompletionItem` and the whole process).

However, we cannot allow the same behavior, because then we aren't able to rank (boost) the correct `Completion`, because (same with `VarCompletion`) the other tags can overshadow.

Therefore this implementation is the best suitable one to also ensure the prober boosting (`EnumTagRanker`).

This implementation is more restrictive, and requires the user to provide the correct "word" to get auto-completion suggestions.

Example (`EnumTagCompleter`):
```
enum Color2 {
   case Red, Green, Yellow
}

def foo(): Unit =
   let c1 = Color2.R
```

This will only provide a single completion: `Color2.Red`.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/EnumTagCompleter.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+    * Checks that the caseSym matches the prefix
+    *
+    * @param sym  the caseSym.
+    * @param word the current prefix.
+    * @return     true, if the var matches prefix, false otherwise.
+    */
+  private def matchesTag(sym: Symbol.CaseSym, word: String): Boolean =
+    word.nonEmpty && word.head.isUpper && sym.name.startsWith(word)
```
LukasRonn, line 68: We are not able to use `sym.name.contains(word)` in order to simulate the same behavior as `UseCompleter`, because VS-code will change the order of suggestions and the "best pick" will not appear as top.

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -221,7 +221,8 @@ object CompletionProvider {
       SignatureCompleter.getCompletions(context) ++
       FieldCompleter.getCompletions(context) ++
       OpCompleter.getCompletions(context) ++
-      MatchCompleter.getCompletions(context)
+      MatchCompleter.getCompletions(context) ++
+      EnumTagCompleter.getCompletions(context)
```
magnus-madsen, line 225: Instead of adding the EnumTagCompleter here, add it to the syntactic context for Pat.

LukasRonn:
> WIP - need to provide the whole namespace. Not ready for merge

I actually think this implementation does provide the correct completions.

I don't think we should provide the whole namespace, i.e.:
```
use A.B.Color3;

mod A {
    mod B {
        pub enum Color3 with ToString {
            case Red, Green, Blue
        }
    }
}

def main(): Unit \ IO = 
    let c1 = Color3.Blue;
    println(c1)
```
Is valid given line 1.

Please discuss.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/EnumTagCompleter.scala
```diff
@@ -0,0 +1,119 @@
...
+    * This checks if the provided nameSpace from the user matches the nameSpace for the Enum.
+    *
+    * It's allowed not to provide a namespace, hence the user could have a 'use A.B.Color2;'.
+    * The user should therefore not be required to provide the path.
+    *
+    * @param ns  the provided nameSpace
+    * @param sym the enumSym
+    * @return    true, if the ns.isEmpty or it matches sym.namespace, false otherwise.
+    */
+  private def providedNameSpaceIsValid(ns: List[String], sym: Symbol.EnumSym): Boolean = {
```
LukasRonn, line 106: This works if the following is invalid in Flix:

```
mod F {
    pub enum Color2 with ToString {
        case Red, Green, Blue
    }
}

enum Color2 with ToString {
    case Red, Blue, Green
}
```

LukasRonn:
We might also have to check if the Enum is public? Please discuss.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/EnumTagCompleter.scala
LukasRonn: Well, it also works if that's not the case, however tags will be generated for each enum.

In this specific case we would get duplicates (since the enums contain the same cases).

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/EnumTagCompleter.scala
```diff
@@ -0,0 +1,121 @@
...
+        getEnumTagCompletions(ns2, z, None) ++
+          // Case 3.2
+          getEnumTagCompletions(ns1, y, Some(z))
+      // Case 4:
+      case ns1 :: ns2 :: enum :: tag :: Nil =>
+        val ns = List(ns1, ns2)
+        getEnumTagCompletions(ns, enum, Some(tag))
+      case _ => Nil
+    }
+  }
```
magnus-madsen, line 71: The above code strikes me as incredibly complicated. How is one supposed to reason about it?

LukasRonn: The whole aspect with nameSpace makes it incredibly complicated. I may have come up with a better and easier solution. Have to test it first.

magnus-madsen:
!!!jar

LukasRonn:
@magnus-madsen 

I understand that the code might be very difficult to reason about.

The aspect with `nameSpace` makes it very complex and complicated. This is due to the fact that the user could have provided a `use A.B`, and that way to user does not have to provide the whole path.

This implementation ensures that the user can provide an arbitrary length of the `nameSpace`, although the `nameSpace` provided should match the same `nameSpace` for the `enum`. Here is an example:

```
mod A {
    mod B {
        mod C {
            pub enum Color3 with ToString {
                case Red, Green, Blue
            }
        }
    }
}

def main(): Unit \ IO = 
    let c1 = A.B.C.Color3 <- (Auto-complete)
    println(c1)
```
In the above example the user has to provide the whole exact path for the program to be valid.
However, if we have:
```
use A.B

mod A {
    mod B {
        mod C {
            pub enum Color3 with ToString {
                case Red, Green, Blue
            }
        }
    }
}

def main(): Unit \ IO = 
    let c1 = B.C.Color3 <- (Auto-complete)
    println(c1)
```
The user is not required to provide the whole path, since we have a `use A.B`. However, we should still be able to provide the possible auto-completions for this case.

magnus-madsen:
I understand, but I think this is trying to go too far. 
With this level of complexity and the way the code is written, it is very hard to reason about.

Perhaps what we should do is break the PR down into some simpler use cases and deal with those first.

magnus-madsen:
In general the use of `use` presents a challenge, right? That's not specific to enums, so perhaps that should be solved now nor as part of this PR?

LukasRonn:
> @magnus-madsen
> 
> I understand that the code might be very difficult to reason about.
> 
> The aspect with `nameSpace` makes it very complex and complicated. This is due to the fact that the user could have provided a `use A.B`, and that way to user does not have to provide the whole path.
> 
> This implementation ensures that the user can provide an arbitrary length of the `nameSpace`, although the `nameSpace` provided should match the same `nameSpace` for the `enum`. Here is an example:
> 
> ```
> mod A {
>     mod B {
>         mod C {
>             pub enum Color3 with ToString {
>                 case Red, Green, Blue
>             }
>         }
>     }
> }
> 
> def main(): Unit \ IO = 
>     let c1 = A.B.C.Color3 <- (Auto-complete)
>     println(c1)
> ```
> 
> In the above example the user has to provide the whole exact path for the program to be valid. However, if we have:
> 
> ```
> use A.B
> 
> mod A {
>     mod B {
>         mod C {
>             pub enum Color3 with ToString {
>                 case Red, Green, Blue
>             }
>         }
>     }
> }
> 
> def main(): Unit \ IO = 
>     let c1 = B.C.Color3 <- (Auto-complete)
>     println(c1)
> ```
> 
> The user is not required to provide the whole path, since we have a `use A.B`. However, we should still be able to provide the possible auto-completions for this case.

This does however also allow for invalid completions. If we don't have a `use A.B`, we would still get a possible completion for `let c1 = B.C.Color3 <- (Auto-complete)`. Here we would get a compile-error, saying: `name not found`. I don't think we would be able to work-around that. Please discuss.

magnus-madsen:
I think you are trying for a perfect solution, but you don't have enough infrastructure to solve it.

Specifically, you don't know what namespace you are currently auto-completing inside nor what `use`s are in scope.

Without this information its hard to solve the problem in all cases. 

What might be better is (a) work iteratively and to (b) focus on a simpler case first.

For example, we could start by focusing on fully-qualified names. 

Thus if you write `A.B.C` what we will do is lookup `A.B.C` and see if there is an enum `C` and offer its tags. 

Next, we will also see if there is an enum `A.B.` with a tag starting with `C`. 

Then we should actually complete the enum with proper placeholders, not just with its name. E.g. `Foo(x1, x2, ...)`.

magnus-madsen:
And at all times should the code we have remain clear and understandable.

magnus-madsen:
It is better to have something that works correctly and that we understand than something more general that has strange corner cases and we don't understand.

LukasRonn:
> Then we should actually complete the enum with proper placeholders, not just with its name. E.g. `Foo(x1, x2, ...)`.

You mean that we shouldn't have to type whole `enum`, before being provided with it's possible tags?

magnus-madsen:
Oh- I merged it. I thought it was working.

I guess you can make a new PR that fixes the issue and introduces the ranker.

LukasRonn:
> Oh- I merged it. I thought it was working.
> 
> 
> 
> I guess you can make a new PR that fixes the issue and introduces the ranker.

Thought I made it to a draft.. My bad.

LukasRonn:
> It seems to work? Just need to expand on the arity.

What part works? I don't get any suggestions?

PR #5788, opened 2023-04-25
feat: add `SyntacticContext.Pat`

magnus-madsen:


magnus-madsen:
@LukasRonn This makes it possible to know when one is inside a pattern match. This should be useful for EnumTagCompleter.

LukasRonn:
> @LukasRonn This makes it possible to know when one is inside a pattern match. This should be useful for EnumTagCompleter.

In what way?

I mean, we should be able to auto-complete a EnumTag in other cases than just in a pattern match, right?

For instance:
```
enum Car with ToString {
   case Fiat, Volvo, Mercedes
}

def foo(): Unit \ IO =
   let car1 = Car.Fiat;
   println(car1)
```

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -146,7 +146,7 @@ object CompletionProvider {
       case SyntacticContext.Decl.Class => return Nil
       case SyntacticContext.Expr.Constraint => return PredicateCompleter.getCompletions(context)
       case _: SyntacticContext.Type => return TypeCompleter.getCompletions(context)
-      // TODO: SYNTACTIC-CONTEXT
+      case _: SyntacticContext.Pat => return Nil
```
LukasRonn, line 149: Do you mean, that we should call the EnumTagCompleter here?

Issue #5772, opened 2023-04-24
Collapse all `IntrinsicOperator`s` into one


magnus-madsen:
Merge all of:

```scala
case class Intrinsic0(op: IntrinsicOperator0, tpe: MonoType, loc: SourceLocation) extends Expr

case class Intrinsic1(op: IntrinsicOperator1, exp: Expr, tpe: MonoType, loc: SourceLocation) extends Expr

case class Intrinsic2(op: IntrinsicOperator2, exp1: Expr, exp2: Expr, tpe: MonoType, loc: SourceLocation) extends Expr

case class Intrinsic3(op: IntrinsicOperator3, exp1: Expr, exp2: Expr, exp3: Expr, tpe: MonoType, loc: SourceLocation) extends Expr

case class IntrinsicN(op: IntrinsicOperatorN, exps: List[Expr], tpe: MonoType, loc: SourceLocation) extends Expr

case class Intrinsic1N(op: IntrinsicOperator1N, exp: Expr, exps: List[Expr], tpe: MonoType, loc: SourceLocation) extends Expr
```

into just:

```scala
case class Intrinsic(op: IntrinsicOperator, exps: List[Expr], tpe: MonoType, loc: SourceLocation) extends Expr
```

This also means to merge `IntrinsicOperatorXY` into one trait/object.

**When doing this, I would to have all operations in the same order as in `FinalAst.Expression` (i.e. where for example Array operations are grouped together).**

JonathanStarup:
So then the exps list is unsafely unpacked?

magnus-madsen:
@jaschdoc , I would like this done ASAP, if possible :)

(Otherwise let me know- and that is ok too)

magnus-madsen:
Fixed in master.

PR #5790, opened 2023-04-25
refactor: collapse `Expr.Intrinsic0` -> `Expr.Intrinsic`

jaschdoc:
Related to https://github.com/flix/flix/issues/5772


main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -314,35 +314,90 @@ object GenExpression {
...
+        case IntrinsicOperator.MatchError =>
+          addSourceLine(visitor, loc)
+          AsmOps.compileThrowFlixError(visitor, BackendObjType.MatchError.jvmName, loc)
+
+        case _ => throw InternalCompilerException("Unexpected Intrinsic Operator for 0 Expressions", loc)
+
+      }
+
+      case e1 :: Nil => ???
+      case e1 :: e2 :: Nil => op match {
```
magnus-madsen, line 354: The matching here seems a bit strange. I would have expected either:

```
case Expr.Intrinsic(InstrinsicOperator.Spawn, List(e1)) => ...
```

or 

```
case Expr.Intrinsic(InstrinsicOperator.Spawn, exps) => exps match
   case List(e1) => ...
   case _ => throw ...
```

but in any case, its important to match not on the order of arguments, but on the order they appear in InstrincOperator (which I would like to match the usual order of expression in e.g. FinalAst)

magnus-madsen: Does it make sense?

magnus-madsen:
OK. Turns out there are more changes needed to. I will work in it; you can continue on the compileBinary expr cleanup 👍

PR #5791, opened 2023-04-25
refactor: add phase names

JonathanStarup:
stored names for later prettyprinting interfacing


magnus-madsen:
But why? 200 lines just for this abstraction? And you won't be able to reflect the strings into `flix.cachedXYZ` anyway?

JonathanStarup:
> But why? 200 lines just for this abstraction? And you won't be able to reflect the strings into `flix.cachedXYZ` anyway?

Each phase name will show up in 4 places (showAstProvider, AstPrinterx3), so I thought it might be nice to have for ease of modification.  I can remove a lot of lines by removing the comments if you want.

magnus-madsen:
Why can't we just have a single function: `String -> TheAstAsString`?

JonathanStarup:
like 
```
enum PhaseString { case Parser, Eraser, .. }
def parse(s: String): PhaseString = match s {
    case "Parser" => Parser
    ..
}
```

magnus-madsen:
I don't know why that would be needed either. Perhaps you can start by adding support for 3 ASTs to be printed and then we can use that to see what is needed.

PR #5795, opened 2023-04-26
refactor: rework `EffSymCompleter`

magnus-madsen:


main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
```diff
@@ -32,6 +32,16 @@ sealed trait Completion {
     * Returns a LSP completion item for `this`.
     */
   def toCompletionItem(context: CompletionContext)(implicit flix: Flix): CompletionItem = this match {
+
+    case Completion.EffectCompletion(sym, doc) =>
+      val name = sym.toString
+      CompletionItem(label = name,
+        sortText = name,
```
LukasRonn, line 39: Shouldn't this have some Priority? Ex. `sortText = Priority.high(name)`

PR #5806, opened 2023-04-27
chore: create CBV AST

mlutze:
`--Xprint-asts Pre,Post` to see the transformation (in `build/asts/`)


magnus-madsen:
I think this can now be rebased onto master. 

The ControlSeparation should then happen after `Reducer` which now has a very simple AST.

Going from there it should be possible to add `Stmt` to `MonoTypedAst` and `ErasedAst`.

magnus-madsen:
Is the current belief that this works, but there is a bug in monomorph? Or is it that there is bug, but also pieces are missing here?

JonathanStarup:
there is a bug, but also pieces are missing here

magnus-madsen:
Rebasing onto #6000 might help (and/or using the new Verifier to find more bugs).

magnus-madsen:
@mlutze @JonathanStarup A few things to do before we can merge this:

- We need to restore benchmarks on arewefast.flix.dev so we can see the performance impact (Matt)
- We should drop the "Undo" phase and pull the `Expr/Stmt` forward through the backend until code generation. The path for this has already been laid, so it should be straightforward.
- Extensively review the code.

Once this feels solid, we can merge, and move on to the next part.

mlutze:
This can be closed, right_

JonathanStarup:
yea

Issue #5672, opened 2023-03-31
Remove syntax for exponentiation and remove `Exp` type class


magnus-madsen:
Objective:

- [ ] Remove `exp ** exp ` syntax used for exponentiation.
  - Start here: https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/Parser.scala#L693`
- [ ] Remove `Exp` type class.

jaschdoc:
Should we also remove the semantic operators for exponentiation then if there is no way to do this anymore? Or should we keep an intrinsic as implementation for a library call? E.g.
```scala
pub def exp(x: Int32, n: Int32): Int32 = $Int32Exp$(x, n)
```

magnus-madsen:
We still need the semantic operators. We are just removing the syntactic sugar.

magnus-madsen:
@stephentetley Would you be interested in this? It should be very similar to what you just did :)

stephentetley:
Yes - I can pick it up. Is the `Exp` class to be removed like the BitwiseX classes were?

magnus-madsen:
> Yes - I can pick it up. Is the `Exp` class to be removed like the BitwiseX classes were?

Yes. Thank you :)

stephentetley:
Should we call the binary function `pow` rather than `exp` and add a unary function `exp`?

> pub def pow(b: a, n: a): a

> pub def exp(n: a): a

magnus-madsen:
I think we were following Rust. What do they do?
Otherwise I have no opinion.

stephentetley:
I was looking at Java and Haskell which I believe follow my suggestion.

It looks like Rust does too:

https://docs.rs/num/latest/num/trait.Float.html#tymethod.exp
https://docs.rs/num/latest/num/fn.pow.html

`exp` should only be defined on `Float32`, `Float64` and `BigDecimal`.

magnus-madsen:
Ok. Let's do that.

stephentetley:
It turns out there is only a Java primitive for `exp` on Float64. 

To implement `exp` for  `Float32` I would have to (safely) convert to Float64 but risk an error converting back again.

`BigDecimal` has no primitive and there are precision problems trying to do-it-yourself:

https://stackoverflow.com/questions/11855775/bigdecimal-eulers-number-calculation-to-emulate-math-expdouble-a

I'm about to make a PR and have decided to just provide `exp` for `Float64`.

[...]

PR #5829, opened 2023-04-30
chore: remove exp operator and Exp class (issue #5672)

stephentetley:
This PR removes the exp operator `**` and Exp class.

`exp` is replaced by `pow` in the relevant number modules. `Float64` also has a function `exp` added in line with Haskell and Rust's `exp`.

The `pow` function in BigInt respects the type of `BigInteger.pow` in Java - the power value (second argument) is an `Int32` not a `BigInt`.

> pub def pow(x: BigInt, n: Int32): BigInt


magnus-madsen:
Thanks!

PR #5831, opened 2023-05-01
refactor: introduce `Stmt` to backend ASTs

magnus-madsen:


magnus-madsen:
@JonathanStarup @mlutze I think with this we can even start to move some expressions to statements.

PR #5833, opened 2023-05-01
chore: fix type formatting for new arrow types

mlutze:


magnus-madsen:
What's the new format?

mlutze:
This is only on the output, changing it to something reasonable instead of the something broken that it is now.

PR #5837, opened 2023-05-01
chore: remove effect set types

mlutze:
- removes everything to do with effect sets
- makes it so that the `&` part of the arrow type is ignored. Everything has to go in the set part.
- lots of tests are disabled, but I expect a quick followup to re-enable them after changing the parsing of effects


main/src/ca/uwaterloo/flix/language/ast/TypeConstructor.scala
```diff
@@ -314,46 +314,11 @@ object TypeConstructor {
     def kind: Kind = Kind.Bool ->: Kind.Bool ->: Kind.Bool
   }
 
-  /**
-    * A type constructor that represents the complement of an effect set.
-    */
-  case object Complement extends TypeConstructor {
```
magnus-madsen, line 320: I was actually thinking we would live the type constructors and types? (Why not?) Will we opt need them?

Even if they become deadcode now?

mlutze: IMO it is cleaner to get rid of it. We're basically introducing a "don't use these things" invariant in the compiler, which is best represented by "these things" not existing.

The type constructors are trivial to add back, and we may end up with a different representation anyway.

magnus-madsen: I see, but we keep some unification stuff around?

mlutze: I think we can get rid of it too. It's pretty much the same as what we have for case sets, and it's likely to be subject to big changes for optimization.

PR #5840, opened 2023-05-01
refactor: un-nest monomorph

JonathanStarup:
Instead of nesting visits inside specialize, they're their own methods with implicit mutable objects

The diff is a mess, but I've functionally changed the one weird occurrence of `subst0` to `subst` and added `Def2Def` and `DefQueue` to implicit arguments

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Monomorph.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Monomorph.scala
@@ -208,9 +208,9 @@ object Monomorph {
 
     implicit val r: Root = root
 
-    val defQueue: DefQueue = mutable.Set.empty
+    implicit val defQueue: DefQueue = mutable.Set.empty
 
-    val def2def: Def2Def = mutable.Map.empty
+    implicit val def2def: Def2Def = mutable.Map.empty
 
     /*
      * A map used to collect specialized definitions, etc.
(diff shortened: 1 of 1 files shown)


main/src/ca/uwaterloo/flix/language/phase/Monomorph.scala
```diff
@@ -292,358 +292,353 @@ object Monomorph {
...
 
-    visitExp(exp0, env0, subst0)
+  /**
+    * Specializes the given method `method` w.r.t. the current substitution.
+    *
+    * Returns the new method.
+    */
+  private def visitJvmMethod(method: JvmMethod, env0: Map[Symbol.VarSym, Symbol.VarSym], subst: StrictSubstitution)(implicit def2def: Def2Def, defQueue: DefQueue, root: Root, flix: Flix): JvmMethod = method match {
+    case JvmMethod(ident, fparams0, exp0, tpe, pur, loc) =>
+      val (fparams, env1) = specializeFormalParams(fparams0, subst)
```
JonathanStarup, line 633: This `subst` here was `subst0` before which I believe is a typo (`subst0` would be the argument to specialize instead of the concrete substitution for this function call)

magnus-madsen: Ok. Seems reasonable.

magnus-madsen:
I agree -- in general -- that is has been a mistake to next big functions.

We should only nest tiny helper functions.

magnus-madsen:
It might also be sensible to introduce a private case class `Context` and implicitly pass that around.

JonathanStarup:
> It might also be sensible to introduce a private case class `Context` and implicitly pass that around.

Its gonna grow with the enum change, should I collect the two maps into one object now?

magnus-madsen:
> > It might also be sensible to introduce a private case class `Context` and implicitly pass that around.
> 
> Its gonna grow with the enum change, should I collect the two maps into one object now?

Yeah, but let me merge this PR first.

Issue #5851, opened 2023-05-02
The test runner always returns 0, even when tests failed.


Release-Candidate:
Running `java -jar flix.jar test` always returns a value of 0, but should return a value not equal to 0 on errors (failed tests).

I'm going to have a look at the test runner and write a PR.

Release-Candidate:
See https://github.com/flix/flix/pull/5855

Release-Candidate:
Fixed in 0.36.0

PR #5855, opened 2023-05-02
fix: return 1 from flix if a test failed

Release-Candidate:
Return 1 from `java -jar flix.jar test` if one of the tests fails.

I've used 1 (not a named constant), as that is the same as returned by other `System.exit` calls.

See #5851


main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -605,7 +605,6 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
     build() flatMap {
       compilationResult =>
         Tester.run(Nil, compilationResult)
-        ().toOk
```
magnus-madsen, line 608: This seems ok, but I think we also need to do something in `SimplerRunner` (?)

main/src/ca/uwaterloo/flix/tools/Tester.scala
```diff
@@ -34,28 +34,36 @@ object Tester {
...
+  def run(filters: List[Regex], compilationResult: CompilationResult)(implicit flix: Flix): Result[Unit, Int] = {
     //
     // Find all test cases (both active and ignored).
     //
     val tests = getTestCases(filters, compilationResult)
 
     // Start the TestRunner and TestReporter.
     val queue = new ConcurrentLinkedQueue[TestEvent]()
-    val reporter = new TestReporter(queue, tests)
+    val failureQueue = new ConcurrentLinkedQueue[Int]()
```
magnus-madsen, line 45: I would prefer if we did something else. Can we not have TestReporter return a future with the result?

magnus-madsen (commented):
If I could ask you to add your name to `AUTHORS.md` to state you make your contributions open source.

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
Release-Candidate: Adding an `exit(1)` in case of a test failure is not a problem.

But:
The `--test` or `--benchmark` arguments should enter one of these branches
https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/tools/SimpleRunner.scala#L115-L121
but never reach that part because the program enters this branch
https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/tools/SimpleRunner.scala#L66
and exits afterwards.
So ` java -jar build/libs/flix.jar --test` and ` java -jar build/libs/flix.jar --benchmark` don't work at all now. 

Adding additional checks to
https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/tools/SimpleRunner.scala#L66
makes them working (again?).

Release-Candidate: See https://github.com/flix/flix/pull/5855/commits/ae31aad5881ca9f8b065aa381c821fb6118ec598

main/src/ca/uwaterloo/flix/tools/Tester.scala
Release-Candidate: Actually I don't see a reason why `TestReporter` runs in his own thread, if the main thread is blocking in the `join`s anyway. I propose just using the `TestReporter.run` function as a "normal" one.
See https://github.com/flix/flix/pull/5855/commits/d225d9fa9f9f409a11c470553ebfc4d8b16a9899

magnus-madsen:
Wooah. Hold your horses too many changes are happening at once.

I think the most important change is that we need `Tester.run` to return a `Result`. 

To accomplish that, I think the simplest change is to add a private mutable field to `TestReporter`:

```
private var success: Boolean = false
```

We set this field in the end of `run` in `TestReporter`.

We add a getter `def isSuccess: Boolean` which we can use in `Tester.run`.

Then we can do a online change in `Bootstrap` (which you have) and we are done.

magnus-madsen:
We should deal with `--benchmark` and `--test` separately. In fact, I suspect they should be removed.

Release-Candidate:
Ok, reverted `TestReporter` to a thread and changed the `TestReporter` to use `isSuccess`.

main/src/ca/uwaterloo/flix/tools/SimpleRunner.scala
```diff
@@ -63,7 +63,8 @@ object SimpleRunner {
     }
 
     // check if we should start a REPL
-    if (cmdOpts.command == Command.None && cmdOpts.files.isEmpty) {
```
magnus-madsen, line 66: Leave this unchanged. We will deal with `--test` separately.

main/src/ca/uwaterloo/flix/tools/SimpleRunner.scala
```diff
@@ -63,7 +63,8 @@ object SimpleRunner {
...
+    if (cmdOpts.command == Command.None && cmdOpts.files.isEmpty
```
magnus-madsen, line 66: same comment

main/src/ca/uwaterloo/flix/tools/SimpleRunner.scala
```diff
@@ -97,6 +98,7 @@ object SimpleRunner {
     timer.getResult match {
       case Validation.Success(compilationResult) =>
 
+        if (!cmdOpts.benchmark && !cmdOpts.test) {
```
magnus-madsen, line 101: same comment

main/src/ca/uwaterloo/flix/tools/SimpleRunner.scala
```diff
@@ -111,13 +113,17 @@ object SimpleRunner {
...
         }
+        }
 
         if (cmdOpts.benchmark) {
           Benchmarker.benchmark(compilationResult, new PrintWriter(System.out, true))(options)
         }
 
         if (cmdOpts.test) {
-          Tester.run(Nil, compilationResult)(flix)
+          Tester.run(Nil, compilationResult)(flix) match {
```
magnus-madsen, line 123: leave this unchanged too.

main/src/ca/uwaterloo/flix/tools/Tester.scala
```diff
@@ -50,13 +50,26 @@ object Tester {
...
+      Result.Err(1)
+    }
   }
 
   /**
     * A class that reports the results of test events as they come in.
     */
   private class TestReporter(queue: ConcurrentLinkedQueue[TestEvent], tests: Vector[TestCase])(implicit flix: Flix) extends Thread {
 
+    private var success = new java.util.concurrent.atomic.AtomicBoolean(true)
```
magnus-madsen, line 67: var -> val

magnus-madsen (commented):
Almost there :)

main/src/ca/uwaterloo/flix/tools/SimpleRunner.scala
Release-Candidate: Ah, sorry, I misunderstood ('dealing with --test/--benchmark').

magnus-madsen:
Merged. Congratulations!

Release-Candidate:
Great, thanks for your help.

I'm going to open an issue for the `--test` and `--benchmark` parameters. The `--test` argument is also used by the `Run Tests` code-lens, I'll add details to the issue.

I'd also add another issue to add the source location to the test failure output.

I've just got some "real" work to do now ...

PR #5863, opened 2023-05-04
refactor: monomorph visit type

JonathanStarup:


JonathanStarup:
@mlutze maybe you can help here. I want to touch every type of the program, but right now monomorph only looks at defs. What other places do i have to visit
```
  case class Root(classes: Map[Symbol.ClassSym, LoweredAst.Class],
                  instances: Map[Symbol.ClassSym, List[LoweredAst.Instance]],
                  sigs: Map[Symbol.SigSym, LoweredAst.Sig],
                  defs: Map[Symbol.DefnSym, LoweredAst.Def],
                  enums: Map[Symbol.EnumSym, LoweredAst.Enum],
                  effects: Map[Symbol.EffectSym, LoweredAst.Effect],
                  typeAliases: Map[Symbol.TypeAliasSym, LoweredAst.TypeAlias],
                  classEnv: Map[Symbol.ClassSym, Ast.ClassContext],
                  eqEnv: ListMap[Symbol.AssocTypeSym, Ast.AssocTypeDef])
```

And @magnus-madsen this replacement of `Pair[Int32, String]` into `Pair$42` should happen after all the substitution and specialization. We can either do this inline with the specialization, making sure that we're doing it in the correct order, or we could do it as its own pass after the specialization is done (a la typer where reconstruction is a second full pass)

magnus-madsen:
I don't think we should do it in a separate pass because then the intermediate AST would be in an invalid state.

magnus-madsen:
As for whether to replace in other places in the AST; I am not sure. @mlutze Thoughts?

Because most Monomorph do we even use e.g. type aliases?

JonathanStarup:
> As for whether to replace in other places in the AST; I am not sure. @mlutze Thoughts?
> 
> Because most Monomorph do we even use e.g. type aliases?

Matt argued that all typeclass stuff should conceptually be gone after the type checker, so defs is all that's needed (i will verify later)

> I don't think we should do it in a separate pass because then the intermediate AST would be in an invalid state.

I was just thinking a function `convertDef` that's called on all defs after the def specialization, so there is no intermediate AST. Another option, since i believe enum-monomorph does not affect functions, would be to have MonomorphEnums phase, which would leave the intermediate valid.

magnus-madsen:
Maybe it is easier to talk offline?

PR #5865, opened 2023-05-04
fix: missing type substitutions in monomorph

JonathanStarup:
Also:
* remove some unused variables
* remove some unused functions


magnus-madsen:
Is it "good" that the functions are not used (???)

PR #5867, opened 2023-05-04
feat: monomorph enums

JonathanStarup:
* Monomorphed Enums
* Consistently removed type aliases

Follow up PR: clean up JvmOps and Asts


magnus-madsen:
> Simplifier uses the enumSym Choice.Absent but that's now monomorphized... what to do?

We can drop this feature, if it no longer works.

> Simplifier translates Vector to Array ...?

Yes, because that is what they are. Immutable arrays.

magnus-madsen:
So exactly one test is failing for unknown reasons? And its a simple question of Cons not matching the interface?

magnus-madsen:
It would probably be reasonable to update the `Verifier` to type check `Is`, `Tag`, `Untag`. Might reveal the problem or might not.

main/src/ca/uwaterloo/flix/language/phase/MonomorphEnums.scala
```diff
@@ -0,0 +1,396 @@
...
+    case Expression.InvokeConstructor(constructor, args, tpe, pur, loc) =>
+      Expression.InvokeConstructor(constructor, args.map(visitExp), visitType(tpe), visitType(pur), loc)
+    case Expression.InvokeMethod(method, exp, args, tpe, pur, loc) =>
+      Expression.InvokeMethod(method, visitExp(exp), args.map(visitExp), visitType(tpe), visitType(pur), loc)
+    case Expression.InvokeStaticMethod(method, args, tpe, pur, loc) =>
+      Expression.InvokeStaticMethod(method, args.map(visitExp), visitType(tpe), visitType(pur), loc)
+    case Expression.GetField(field, exp, tpe, pur, loc) =>
+      Expression.GetField(field, visitExp(exp), visitType(tpe), visitType(pur), loc)
+    case Expression.PutField(field, exp1, exp2, tpe, pur, loc) =>
+      Expression.PutField(field, visitExp(exp1), visitExp(exp1), visitType(tpe), visitType(pur), loc)
```
magnus-madsen, line 214: Boom, mistake here. exp1 appears twice.

magnus-madsen: This is one reason I prefer that we write it out:

```
val e1 = visitExp(exp1) 
...
```

JonathanStarup:
No observable influence on throughput

magnus-madsen:
Good to see that all tests pass. I would like to see two cleanups before merging:

- [ ] Let's add some checks to `Verifier`. It should be very little code needed. (Remember to run it to actually see if there is an issue).
- [ ] "Reformat" MonomorphEnums such that expressions and types are named like we do in other phases (e.g. `e1`, `es`, and so forth).

Other cleanup can wait for later.

@mlutze should also take a detailed look to see that things look OK.

magnus-madsen:
But great work so far 👍 . Will be nice to have this in.

main/src/ca/uwaterloo/flix/language/phase/MonomorphEnums.scala
```diff
@@ -0,0 +1,449 @@
...
+        case LoweredAst.RelationalChoiceRule(pat, exp) =>
+          val newPat = pat.map {
+            case RelationalChoicePattern.Wild(loc) => RelationalChoicePattern.Wild(loc)
+            case RelationalChoicePattern.Absent(loc) => RelationalChoicePattern.Absent(loc)
+            case RelationalChoicePattern.Present(sym, tpe, loc) => RelationalChoicePattern.Present(sym, visitType(tpe), loc)
+          }
+          LoweredAst.RelationalChoiceRule(newPat, visitExp(exp))
+      }
+      Expression.RelationalChoose(exps.map(visitExp), newRules, visitType(tpe), visitType(pur), loc)
+    case Expression.Tag(sym, exp, tpe, pur, loc) =>
```
magnus-madsen, line 145: It might be good to add some comments to call out that this is where things happen :)

main/src/ca/uwaterloo/flix/language/phase/MonomorphEnums.scala
```diff
@@ -0,0 +1,449 @@
...
+    case Expression.Force(exp, tpe, pur, loc) =>
+      Expression.Force(visitExp(exp), visitType(tpe), visitType(pur), loc)
+  }
+
+  private def visitPat(pat: LoweredAst.Pattern)(implicit ctx: Context, root: LoweredAst.Root, flix: Flix): LoweredAst.Pattern = pat match {
+    case Pattern.Wild(tpe, loc) => Pattern.Wild(visitType(tpe), loc)
+    case Pattern.Var(sym, tpe, loc) => Pattern.Var(sym, visitType(tpe), loc)
+    case Pattern.Cst(cst, tpe, loc) => Pattern.Cst(cst, visitType(tpe), loc)
+    case Pattern.Tag(sym, tagPat, tpe, loc) =>
+      val newTagPat = visitPat(tagPat)
```
magnus-madsen, line 238: Ditto for calling out that here things happen.

main/src/ca/uwaterloo/flix/language/phase/MonomorphEnums.scala
```diff
@@ -0,0 +1,449 @@
...
+            case (Type.Pure, Type.EffUniv) => Type.Pure
+            case (Type.EffUniv, Type.Pure) => Type.Pure
+            case (Type.EffUniv, Type.EffUniv) => Type.EffUniv
+            case _ => throw InternalCompilerException(s"Unexpected non-simple effect $tpe", applyLoc)
+          }
+
+        // Simplify boolean equations.
+        case Type.Cst(TypeConstructor.Not, _) => ???
+        case Type.Apply(Type.Cst(TypeConstructor.And, _), x, _) => ???
+        case Type.Apply(Type.Cst(TypeConstructor.Or, _), x, _) => ???
```
magnus-madsen, line 368: @mlutze Probably want something here.

main/src/ca/uwaterloo/flix/language/phase/MonomorphEnums.scala
```diff
@@ -0,0 +1,449 @@
...
+            case _ => throw InternalCompilerException(s"Unexpected non-simple effect $tpe", applyLoc)
+          }
+        case Type.Apply(Type.Cst(TypeConstructor.Intersection, _), x, _) =>
+          (x, t2) match {
+            case (Type.Pure, Type.Pure) => Type.Pure
+            case (Type.Pure, Type.EffUniv) => Type.Pure
+            case (Type.EffUniv, Type.Pure) => Type.Pure
+            case (Type.EffUniv, Type.EffUniv) => Type.EffUniv
+            case _ => throw InternalCompilerException(s"Unexpected non-simple effect $tpe", applyLoc)
+          }
```
magnus-madsen, line 363: Is the above necessary? Smart constructors maybe? or?

main/src/ca/uwaterloo/flix/language/phase/MonomorphEnums.scala
```diff
@@ -0,0 +1,449 @@
...
+          LoweredAst.MatchRule(visitPat(pat), guard.map(visitExp), visitExp(exp))
+      }
+      Expression.Match(visitExp(exp), newRules, visitType(tpe), visitType(pur), loc)
+    case Expression.TypeMatch(exp, rules, tpe, pur, loc) =>
+      val newRules = rules.map {
+        case LoweredAst.TypeMatchRule(sym, tpe, exp) =>
+          LoweredAst.TypeMatchRule(sym, visitType(tpe), visitExp(exp))
+      }
+      Expression.TypeMatch(visitExp(exp), newRules, visitType(tpe), visitType(pur), loc)
+    case Expression.RelationalChoose(exps, rules, tpe, pur, loc) =>
```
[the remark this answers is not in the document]
JonathanStarup: I think its prettier to just let simplified be the one place to throw, more modular

main/src/ca/uwaterloo/flix/language/phase/MonomorphEnums.scala
JonathanStarup: Smart constructors do not guarantee that formulas are completely evaluated, even the type minimizer does not guarantee it

main/src/ca/uwaterloo/flix/language/phase/MonomorphEnums.scala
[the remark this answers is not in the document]
magnus-madsen: I understand- but it could also cause confusion. Better to say that this is not expected.

main/src/ca/uwaterloo/flix/language/phase/MonomorphEnums.scala
magnus-madsen: Hmm. @mlutze Thoughts?

It seems we do simplications in so many places :O

JonathanStarup:
@magnus-madsen Ive added tags/untags to verfier and it doesn't cause any more complaints than it did before

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -99,8 +99,31 @@ object Verifier {
...
 
         }
+        case AtomicOp.Tag(sym) =>
+          root.enums.get(sym.enumSym) match {
+            case None => throw InternalCompilerException(s"Unknown enum sym: '$sym'", loc)
+            case Some(e) => e.cases.get(sym) match {
+              case None => throw InternalCompilerException(s"Unknown enum case sym: '$sym' of '${e.sym}'", loc)
+              case Some(caze) =>
+                val List(t) = ts
+                check(expected = caze.tpeDeprecated)(t, loc)
```
JonathanStarup, line 109: note that these will be renamed tpe in the cleanup PR. The enum type is just `List$123` and the case type is the ground value of the tag `(Int32, String)` fx.

main/src/ca/uwaterloo/flix/language/phase/MonomorphEnums.scala
```diff
@@ -0,0 +1,698 @@
...
+    val LoweredAst.FormalParam(sym, mod, tpe, src, loc) = p
+    val t = visitType(tpe)
+    LoweredAst.FormalParam(sym, mod, t, src, loc)
+  }
+
+  /**
+    * Returns a scheme with specialized enums in its base and no aliases.
+    */
+  private def visitScheme(sc: Scheme)(implicit ctx: Context, root: LoweredAst.Root, flix: Flix): Scheme = {
+    val Scheme(quantifiers, tconstrs, econstrs, base) = sc
```
mlutze, line 497: Not visiting the other stuff? In theory it does not matter since tconstrs and econstrs shouldn't be used anymore. But maybe worth mentioning.

main/src/ca/uwaterloo/flix/language/phase/MonomorphEnums.scala
```diff
@@ -0,0 +1,698 @@
...
+
+  /**
+    * Specializes the given enum in respect to the given type arguments and the
+    * enums it itself uses.
+    *
+    * The type arguments:
+    * - must be ground
+    * - must not contain specialized enums
+    * - may have aliases
+    * - may be un-normalized
```
mlutze, line 510: Thank you for these clear requirements

main/src/ca/uwaterloo/flix/language/phase/MonomorphEnums.scala
mlutze: I like that this is separate from other type minimization due to the guarantees, as Jonathan said. Maybe later we can refactor if we find places that "lawful normalization" should be used.

mlutze (approved):

magnus-madsen:
Thanks Jonathan!

@JonathanStarup @mlutze @stephentetley This fixes the last major hack in the Flix compiler 🥳 Now we can all sleep better :)

There should not be any changes to reasonable programs, but more corner-case programs should now compile. (As they relate to polymorphic enums.)

Issue #5872, opened 2023-05-04
Add source location to test failure output


Release-Candidate:
Add a line containing the source file and start line of a failing test, to make finding the failing test easier for the user.

Output as of now (0.36.0):
```text
   FAIL  testContains01
    Assertion Error: Does Map contain 43?
      Expected: Map#{bar => 42, foo => 12} contains 43
      Actual:   Map#{bar => 42, foo => 12} does not contain 43
```
proposed change:
```text
   FAIL  testContainsEff01
         /Users/roland/Documents/code/flix-assertions/test/TestMain.flix:192
    Assertion Error: Does MutSet contain 43?
      Expected: MutSet#{2, 5} contains 43
      Actual:   MutSet#{2, 5} does not contain 43
```

This change is a single line in `Tester.scala`, all needed information is already there.
```flix
writer.println(s"         ${sym.loc.source.name}:${sym.loc.beginLine}")
```

magnus-madsen:
Seems reasonable.

Release-Candidate:
PR merged.

PR #5878, opened 2023-05-05
feat: add source location to failed test output

Release-Candidate:
Add the source file name and line of a failing test to the output.
Now:
```
   FAIL  testContains01
    Assertion Error: Does Map contain 43?
      Expected: Map#{bar => 42, foo => 12} contains 43
      Actual:   Map#{bar => 42, foo => 12} does not contain 43
```
With PR:
```
   FAIL  testContains01
         /Users/roland/Documents/code/flix-assertions/test/TestMain.flix:182
    Assertion Error: Does Map contain 43?
      Expected: Map#{bar => 42, foo => 12} contains 43
      Actual:   Map#{bar => 42, foo => 12} does not contain 43
```

See https://github.com/flix/flix/issues/5872


magnus-madsen:
Thanks!

PR #5892, opened 2023-05-07
chore: rename `region r`/` region rh` to ` region rc`

sockmaster27:
Fixes flix#5394


sockmaster27:
As requested, this only covers the standard library. There are still instances in other internal files.

magnus-madsen:
Thanks!

Issue #1239, opened 2020-07-06
Add String.stripMargin


magnus-madsen:
Given a string like: 

```
    | abc \n
    | def \n
```

returns something like":

```
abc \n
def \n
```

magnus-madsen:
@ZiyaoWei It is still relevant, please go ahead :)

magnus-madsen:
Fixed in master.

PR #5895, opened 2023-05-08
feat: add String.stripMarginWith

ZiyaoWei:
A first attempt at implementing #1239 - if it makes sense I'll add `stripMargin` as a wrapper.

Might not be idiomatic and any feedback is welcome! Also some of the tests might be a bit surprising but I am trying to follow the behavior of `stripMargin` in Scala.


main/src/library/String.flix
```diff
@@ -1228,6 +1228,43 @@ mod String {
         else
             loop(0, 0)
 
+    ///
+    /// For every line in string `s`, strip a leading prefix consisting of
+    /// blanks or control characters followed by `margin` from the line, if
+    /// such a prefix exists.
+    ///
+    pub def stripMarginWith(margin: {margin = String}, s: String): String =
+        let l = debug(linesWithNewLine(s));
```
magnus-madsen, line 1237: Remove debug

magnus-madsen:
If you could also add your name to `AUTHORS.md` to state that you agree to release your contributions under Apache 2.0 :)

PR #5897, opened 2023-05-08
fix: subst on effects in monomorph

JonathanStarup:
I'm not sure how this havent caused problems so far. Type variables were left behind in the purity because of these missing subst calls.

I've also explicitly removed classes/instances/sigs because that was implicit before.


magnus-madsen:
It is very few tests so could just be a simple bug somewhere?

JonathanStarup:
For this program
```
def main(): Int32 = region rrrrrrrrr {
    let d = MutDeque.new(rrrrrrrrr);
    MutDeque.pushBack(1, d);
    MutDeque.foldRightWithCont((e, k) -> (k() - e) * (e `Int32.rem` 2 + 1), 100, d)
}
```
Main has inferred scheme
```
Unit -> Int32 \ e117612 + e24347! + e24347!
```
In which the beginning loop assigns them both to Pure, i.e. the region is now pure instead of Impure

JonathanStarup:
This code in the simplifier converts the effects of the monomorpher. This seem like monomorph responsibility leaking into simplified IMO. there should almost be an error here if the effect is not pure or impure
```
private def simplifyPurity(eff: Type): Purity = {
    if (eff == Type.Pure)
      Pure
    else
      Impure
  }
```

magnus-madsen:
@mlutze Request review + expertise.

main/src/ca/uwaterloo/flix/language/phase/Monomorph.scala
```diff
@@ -313,61 +347,62 @@ object Monomorph {
...
     case Expression.Region(tpe, loc) =>
       Expression.Region(subst(tpe), loc)
 
     case Expression.Scope(sym, regionVar, exp, tpe, pur, loc) =>
       val freshSym = Symbol.freshVarSym(sym)
       val env1 = env0 + (sym -> freshSym)
-      // mark the region variable as Impure inside the region
-      val subst1 = subst + (regionVar.sym -> Type.Impure)
-      Expression.Scope(freshSym, regionVar, visitExp(exp, env1, subst1), subst(tpe), pur, loc)
+      // forcedly mark the region variable as Impure inside the region
```
mlutze, line 382: This seems reasonable

mlutze (approved):

Issue #5871, opened 2023-05-04
Inline `compileInt` into `compileExpression`


jaschdoc:
See https://github.com/flix/flix/issues/5847 for more information

PR #5901, opened 2023-05-09
refactor: simplify `compileInt` in `GenExpression`

jaschdoc:
Closes https://github.com/flix/flix/issues/5871


main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -989,7 +1065,9 @@ object GenExpression {
         // Adding source line number for debugging
         addSourceLine(visitor, loc)
         // We push the 'length' of the array on top of stack
-        compileInt(visitor, exps.length, isLong = false)
+        val arrLengthCst = Ast.Constant.Int32(exps.length)
+        val arrLengthExp = Expr.Cst(arrLengthCst, MonoType.Int32, loc.asSynthetic)
+        compileExpression(arrLengthExp, visitor, currentClass, lenv0, entryPoint)
```
magnus-madsen, line 1070: Hmm. I did not realize compileInt was used in places like here.

jaschdoc: Maybe we just introduce a simple `compileInt` helper function that handles these cases and the `Int32` case?

magnus-madsen: Could we maybe have compileInt and compileLong? Just to clear things up a bit?

jaschdoc: Sure thing

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -1801,6 +1829,59 @@ object GenExpression {
...
+
+      case 0 =>
+        visitor.visitInsn(LCONST_0)
+
+      case 1 =>
+        visitor.visitInsn(LCONST_1)
+
+      case 2 =>
+        visitor.visitInsn(ICONST_2)
+        visitor.visitInsn(I2L) // Sign extend to Long
```
magnus-madsen, line 1846: Why is this not needed for const0 and 1? Seems inconsistent.

jaschdoc: Because they are longs and these are ints. Apparently, only 0 and 1 are available as long constants

magnus-madsen: 😮

magnus-madsen:
I am not sure this refactoring is better than the original code? What do you think? 

My problem with the original code is that I did not like the isLong flag.

Maybe we should just split the original method into two and leave it at that?

jaschdoc:
> I am not sure this refactoring is better than the original code? What do you think?
> 
> My problem with the original code is that I did not like the isLong flag.
> 
> Maybe we should just split the original method into two and leave it at that?

Okay, yeah, I think that's a fine solution.

jaschdoc:
Ready

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -1843,26 +1842,58 @@ object GenExpression {
...
+    case 2 => visitor.visitInsn(ICONST_2)
+    case 3 => visitor.visitInsn(ICONST_3)
+    case 4 => visitor.visitInsn(ICONST_4)
+    case 5 => visitor.visitInsn(ICONST_5)
+    case _ if scala.Byte.MinValue <= i && i <= scala.Byte.MaxValue => visitor.visitIntInsn(BIPUSH, i)
+    case _ if scala.Short.MinValue <= i && i <= scala.Short.MaxValue => visitor.visitIntInsn(SIPUSH, i)
+    case _ => visitor.visitLdcInsn(i)
+  }
+
+  private def compileLong(visitor: MethodVisitor, i: Long): Unit = i match {
```
magnus-madsen, line 1860: Can we add some documentation to each method? Just minimal comments.

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -1843,26 +1842,58 @@ object GenExpression {
...
+    case -1 =>
+      visitor.visitInsn(ICONST_M1)
+      visitor.visitInsn(I2L) // Sign extend to long
+
+    case 0 => visitor.visitInsn(LCONST_0)
+    case 1 => visitor.visitInsn(LCONST_1)
```
magnus-madsen, line 1866: line break

jaschdoc:
Ready again :smile:

magnus-madsen:
Conflicted-- I thought it was better to get the big one in, and then do this one.

jaschdoc:
Sure, no worries :)

magnus-madsen:
Long history for such a small change, but I think the backend is shaping up nicely, don't you?

jaschdoc:
> Long history for such a small change, but I think the backend is shaping up nicely, don't you?

Agreed! Good thing, it'll be squashed

PR #5905, opened 2023-05-10
fix: `EnumTagCompleter` now provides cases

LukasRonn:
```
enum Color with ToString { case Red, Green, Blue }

def main(): Unit \ IO =
   let c1 = Color.G    <- Now suggests "Color.Green"
```

Also works with nameSpaces:
```
mod A.B.C { pub enum Car { case Fiat, Volvo } }

def main(): Unit \ IO =
   let c1 = A.B.C.Car   <- Now suggests all cases
```
```
mod A.B.C { pub enum Car { case Fiat, Volvo } }

def main(): Unit \ IO =
   let c1 = A.B.C.Car.F   <- Now suggests "A.B.C.Car.Fiat"
```


LukasRonn:
There is a lot of string-manipulation and I would love some comments on how to make it more clean.

magnus-madsen:
!!!jar

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/EnumTagCompleter.scala
```diff
@@ -18,29 +18,83 @@ package ca.uwaterloo.flix.api.lsp.provider.completion
...
-    val enumSym = Symbol.mkEnumSym(fqn)
+    // We have to try both options, since we don't know if the user has provided a possible tag
+    getEnumTagCompletion(enumSymWithoutTag, None) ++
+      getEnumTagCompletion(enumSymWithTag, Some(tag))
+  }
 
+  /**
+    * Gets completions for enum tags
+    */
+  private def getEnumTagCompletion(enumSym: EnumSym, tagOption: Option[String])(implicit root: TypedAst.Root): Iterable[EnumTagCompletion] = {
```
magnus-madsen, line 53: Split into two, and move some computation from getCompletions into each.

LukasRonn:
Not ready yet - Need to provide:
```
enum Color { case Red[Int32], Green, Blue }

def foo(): Unit = 
   let c1 = Color.R     <- Should suggest "Color.Red(?x)"
```

magnus-madsen:
OK, but would that not be better in a next PR?

LukasRonn:
> OK, but would that not be better in a next PR?

Honestly up to you. It's a relatively small fix.

magnus-madsen:
OK, if it is a small fix then get it in and I can review.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
```diff
@@ -203,13 +203,14 @@ sealed trait Completion {
         textEdit = TextEdit(context.range, name + " "),
         detail = None,
         kind = CompletionItemKind.Variable)
-    case Completion.EnumTagCompletion(enumSym, caseSym) =>
+    case Completion.EnumTagCompletion(enumSym, caseSym, snippet) =>
```
magnus-madsen, line 206: Why do you do it like this? If you look at PredicateCompletion you can see that I simply store the arity in the completion.

The formatting should be done elsewhere, not in the Completer itself.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
```diff
@@ -406,5 +407,5 @@ object Completion {
     * @param enumSym the sym of the enum.
     * @param caseSym the sym of the case (for that specific enum).
     */
-  case class EnumTagCompletion(enumSym: EnumSym, caseSym: CaseSym) extends Completion
+  case class EnumTagCompletion(enumSym: EnumSym, caseSym: CaseSym, snippet: String) extends Completion
```
magnus-madsen, line 410: snippet should be `arity: Int`

Issue #5881, opened 2023-05-05
Add `ModuleCompleter`


magnus-madsen:
Use both for Expr and Type contexts.

PR #5908, opened 2023-05-10
feat: Introduce `ModuleCompleter`

LukasRonn:
https://github.com/flix/flix/issues/5881


main/src/ca/uwaterloo/flix/api/lsp/provider/completion/ModuleCompleter.scala
```diff
@@ -0,0 +1,65 @@
...
+import ca.uwaterloo.flix.language.ast.TypedAst
+import ca.uwaterloo.flix.api.lsp.provider.completion.Completion.ModCompletion
+
+object ModuleCompleter extends Completer {
+  /**
+    * Returns a List of Completion for completer.
+    */
+  override def getCompletions(context: CompletionContext)(implicit flix: Flix, index: Index, root: TypedAst.Root, delta: DeltaContext): Iterable[ModCompletion] = {
+    val word = context.word.split('.').toList
+    root.modules.flatMap {
```
magnus-madsen, line 29: -> root.modules.keys

magnus-madsen:
Once #5910 is merged this should be cleaned up.

magnus-madsen:
A few screenshots? :)

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/ModuleCompleter.scala
```diff
@@ -0,0 +1,119 @@
...
+
+  /**
+    * Gets completions for modules when not having a fqn
+    */
+  private def getCurrentModuleCompletion(word: List[String])(implicit root: TypedAst.Root): Iterable[ModCompletion] = {
+    // The user has provided a subWord
+    // We know that there is at least one dot, so we split the word and
+    // make a fqn from the everything but the the last (hence it is the subWord).
+    val fqnWithOutLastWord = word.dropRight(1)
+    val subWord = word.takeRight(1)(0)
```
magnus-madsen, line 44: This is unsafe right, could give out of bounds.

LukasRonn: It's not unsafe, it would just produce an empty String.

LukasRonn: That's why i'm not using `.head`, which is unsafe.

magnus-madsen: What if takeright returns the empty string?

magnus-madsen:
I don't understand why this has to be so complicated.

Assume the programmer has written `Aaa.Bbb`, I would just _always_ find `Aaa` do a lookup, find all sub modules in `Aaa` and return.

Presumably `Aaa` could contain a submodule `Ccc`, so we might return `Bbb` and `Ccc` for a completion starting with `B`, but VSCode is smart enough to filter that, right?? Or am I completely off the rails?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/ModuleCompleter.scala
LukasRonn: If `takeRight(1)(0)` is the empty string?

LukasRonn:
> I don't understand why this has to be so complicated.
> 
> Assume the programmer has written `Aaa.Bbb`, I would just _always_ find `Aaa` do a lookup, find all sub modules in `Aaa` and return.
> 
> Presumably `Aaa` could contain a submodule `Ccc`, so we might return `Bbb` and `Ccc` for a completion starting with `B`, but VSCode is smart enough to filter that, right?? Or am I completely off the rails?

So you want VSCode to handle the filtering?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/ModuleCompleter.scala
magnus-madsen: What is the type of `takeRight(1)`?

LukasRonn:
> So you want VSCode to handle the filtering?

Because then we can just always give all modules:

```
root.modules.keys.map {
   case mod => ModCompletion(mod)
}
```
And VSCode would filter out invalid ones.

magnus-madsen:
> 

Well that seems like a reasonable first approximation.

Now, thinking about it there could be a million modues, so that is probably too much.

What's the middle road then?

magnus-madsen:
Does it actually work if you just return all modules?

magnus-madsen:
Here is an idea. Create a new PR with the changes to:

- Completion.scala
- ExprCompleter.scala
- TypeCompleter.scala

And add a ModuleCompleter that returns on all modules.

Then lets iterate in small steps from there.

LukasRonn:
> Does it actually work if you just return all modules?

image

magnus-madsen:
> > Does it actually work if you just return all modules?
> 
> image

I mean, that seems like a pretty good start then. Then from there we should have an argument why it needs improvements, and "take small steps" (in Henrik's voice).

LukasRonn:
If we want to have the neat implementation when only providing `Aaaa` when typing `Aa` and not all the other submodules, this current implementation is the only way, since we have to do the filtering.

LukasRonn:
> > Does it actually work if you just return all modules?
> 
> image

If you think this is good enough, then we can just roll with that :)

magnus-madsen:
Let us at least start with it!

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
```diff
@@ -213,6 +213,13 @@ sealed trait Completion {
         documentation = None,
         insertTextFormat = InsertTextFormat.Snippet,
         kind = CompletionItemKind.EnumMember)
+    case Completion.ModCompletion(modSym) =>
+      val name = modSym.ns.mkString(".")
```
magnus-madsen, line 217: Just do `modSym.toString`?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/ModuleCompleter.scala
```diff
@@ -0,0 +1,30 @@
...
+package ca.uwaterloo.flix.api.lsp.provider.completion
+
+import ca.uwaterloo.flix.api.Flix
+import ca.uwaterloo.flix.api.lsp.Index
+import ca.uwaterloo.flix.language.ast.TypedAst
+import ca.uwaterloo.flix.api.lsp.provider.completion.Completion.ModCompletion
+
+object ModuleCompleter extends Completer {
+  /**
+    * Returns a List of Completion for completer.
```
magnus-madsen, line 25: Remove comment or make it meaningful.

magnus-madsen:
I think the next step should be to do the simplest possible filtering.

If the user has written `A.B.C.` i.e. something _that ends in a dot_ then we only consider members of `A.B.C`.

That's it.

LukasRonn:
> I think Module completions should also come first, yeah?

Yes? Isn't that the case?

LukasRonn:
> > > I think Module completions should also come first, yeah?
> > 
> > 
> > Yes? Isn't that the case?
> 
> No because there are added last to ExprCompleter and TypeCompleter

Doesn't matter, because they are sorted through `Priority`.

LukasRonn:
!!!jar

PR #5915, opened 2023-05-11
fix: deprecated use of URL

magnus-madsen:


magnus-madsen:
@AnnaBlume99 One last thing, could you take a look at this problem? (See the failing test case in the CI).

The issue is that we are not supposed to use `new URL` because it is deprecated, so I switched to `URI`, but now a test fails. 

Maybe my fix combined with some fix from your side can solve it?

magnus-madsen:
I merged with the test marked as ignored.

AnnaBlume99:
Do you still want me to look at it? It's no problem for me to take a look :)

magnus-madsen:
Yes please :)

Issue #5919, opened 2023-05-12
Tag Completion Introduces Unnecessary `()`


magnus-madsen:
```
mod MyModule {
    pub enum Color with Eq {
        case Green,
        case Red
    }
}

@Test
pub def mytest(): Bool = 
    let c1 = MyModule.Color.Green(); // No need for ()
    let c2 = MyModule.Color.Red;
    c1 == c2
```

magnus-madsen:
@LukasRonn Ping

PR #5920, opened 2023-05-12
fix: Removes unnecessary `()` for enum tags

LukasRonn:
https://github.com/flix/flix/issues/5919


magnus-madsen:
Was this tested? I am not sure it works correctly.

LukasRonn:
> Was this tested? I am not sure it works correctly.

I did. Maybe I didn't have the correct jar.

LukasRonn:
> > Was this tested? I am not sure it works correctly.
> 
> I did. Maybe I didn't have the correct jar?

I somehow had a different jar, but one that produced the correct behavior.

I made a fix: https://github.com/flix/flix/pull/5924 .

PR #5924, opened 2023-05-12
fix: Remove unnecessary `()` for enum tags

LukasRonn:
fixes https://github.com/flix/flix/pull/5920 and https://github.com/flix/flix/issues/5919


LukasRonn:
image

PR #5926, opened 2023-05-13
fix: File.write inserts newline

### LINKED ISSUE #5688: File.write inserts an additional newline
Unless I've gone crazy, `File.write` seems to be wrong.  [Surely it should not insert an extra newline character here?](https://github.com/flix/flix/blob/a9b17efdb3a9cb9b810d82361023f8ba89abc942/main/src/library/File.flix#L620)

sockmaster27:
Fixes #5688


magnus-madsen:
Thanks!

PR #5933, opened 2023-05-13
chore: fix typos ` -> " in comments

ZiyaoWei:
There are a few typos in the comments of `String.flix` where `"`s are used instead of `` ` ``.


magnus-madsen:
Thanks!

PR #5934, opened 2023-05-13
feat: add String.stripMargin

ZiyaoWei:
This adds `String.stripMargin` which is a wrapper for `String.stripMarginWith`.


magnus-madsen:
Thanks!

PR #5945, opened 2023-05-15
feat: introduce separate Bool kind

mlutze:
depends on #5943. Followups are: use it for Choice type, fix type formatting


magnus-madsen:
Conflicted and probably needs to be rebased :/

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
```diff
@@ -1457,7 +1457,15 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
...
+      Or ~ zeroOrMore(WS ~ atomic("&") ~ WS ~ Type ~ SP ~> ParsedAst.Type.Intersection)
+    }
+
+    def Or: Rule1[ParsedAst.Type] = rule {
+      And ~ zeroOrMore(WS ~ atomic("or") ~ WS ~ Type ~ SP ~> ParsedAst.Type.Or)
+    }
+
+    def And: Rule1[ParsedAst.Type] = rule {
+      Ascribe ~ zeroOrMore(WS ~ atomic("and") ~ WS ~ Type ~ SP ~> ParsedAst.Type.And)
     }
```
magnus-madsen, line 1470: How is not parsed?

main/src/ca/uwaterloo/flix/language/phase/MonoTyper.scala
```diff
@@ -224,18 +224,17 @@ object MonoTyper {
 
             case TypeConstructor.Record => args.head
 
+            case TypeConstructor.False => MonoType.Unit
```
magnus-madsen, line 227: You mostly have True first, but sometimes False is first :)

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
mlutze: Good question. It's not.

The parser is likely going to need adjustment anyway.

mlutze:
There's more to be done here since there's no Boolean unification enabled for this Boolean kind. I'm looking at how best to parameterize our current stuff over different kinds.

PR #5946, opened 2023-05-15
fix: add support for `Vector` in `Debug.stringify`

### LINKED ISSUE #5719: Add `Vector` cases to `Debug.stringify`
Objective:
- [ ] Extend `Debug.stringify` (in `Debug.flix`) to also handle the `Vector` type.

sockmaster27:
Fixes #5719


magnus-madsen:
@mlutze I got a bit confused. Is what we are "asking for" reasonable/possible given that we simplify Vectors to Arrays?

I think it should be, but I maybe confused.

magnus-madsen:
I am marking this as blocked until we figure out what to do.

magnus-madsen:
@mlutze The elimination of Vector happens in Simplifier which is after Monomorph. Would we then not have generated the right code inside debug/typematch?

mlutze:
> @mlutze The elimination of Vector happens in Simplifier which is after Monomorph. Would we then not have generated the right code inside debug/typematch?

It works when `Vector` is the outermost type. But when it is inside an array, we don't get the special behavior because we call the java `Arrays.toString`.

magnus-madsen:
> > @mlutze The elimination of Vector happens in Simplifier which is after Monomorph. Would we then not have generated the right code inside debug/typematch?
> 
> It works when `Vector` is the outermost type. But when it is inside an array, we don't get the special behavior because we call the java `Arrays.toString`.

Because we don't want to recurse on the print function?

magnus-madsen:
@mlutze In this case:

```
case y: Array[_, _] =>
```

Could we just call `stringify` recursively? It feels a bit dangerous and I don't know if the element type is refined, but conceptually that is what we should be doing.

But maybe its best to leave it alone for now.

mlutze:
> @mlutze In this case:
> 
> ```
> case y: Array[_, _] =>
> ```
> 
> Could we just call `stringify` recursively? It feels a bit dangerous and I don't know if the element type is refined, but conceptually that is what we should be doing.
> 
> But maybe its best to leave it alone for now.

Ideally we could, but the issue is that we would want to call `Arrays.map` and maybe use a `StringBuilder`, but neither of these is in the MinLib.

sockmaster27:
This is the same problem as I had with Vector, though. I managed to get that (almost) working, by using what I assume is the compiler built in function, $VECTOR_GET$. It also just appends the strings to eachother instead of using StringBuilder, which I'm assuming is less than optimal. But if you just want to get it working without changing too much, couldn't we just do this for Array as well?

mlutze:
Yes, I think that would work. I guess in general one wouldn't want to print a huge array anyway.

magnus-madsen:
@sockmaster27 Thanks for the implementation. Ill need to talk to Matt about whether this is the direction we want.

magnus-madsen:
Unfortunately this cannot really work. We shall continue with what we have.

PR #5954, opened 2023-05-16
feat: Introduce `FieldRanker`

LukasRonn:
image

`FieldCompleter` needs a fix. Problem with "overshadowing" again, which means that we won't get a "best pick" in some cases.


main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
```diff
@@ -47,10 +47,11 @@ sealed trait Completion {
         sortText = Priority.normal(name),
         textEdit = TextEdit(context.range, s"$name "),
         kind = CompletionItemKind.Keyword)
-    case Completion.FieldCompletion(name) =>
+    case Completion.FieldCompletion(field, prefix) =>
```
magnus-madsen, line 50: What is prefix and why is it needed?

LukasRonn: The way `FieldCompleter` is constructed is quite confusing. I'll remove anything not needed in `FieldCompletion` when I'm fixing `FieldCompleter`. Will be in next PR.

magnus-madsen: Okay, but I don't like merging things that are halfway done, so I will hold you to your promise to fix it :)

LukasRonn: I Will, however, if you like, you Can just hold this PR until the other one is merged

LukasRonn: On the other hand, I just realized:

The prefix makes sure, that the "variable" defined as a record will be included, and not only the field.

```
def main(): Unit \ IO = 
    let p = { x = 1, y = 2 };
    let sum = p.x             <- The prefix is "p"
    println()
```

PR #5955, opened 2023-05-16
chore: invert Boolean algebra

mlutze:
Needed since the naive translation put "Empty" as "True", which is confusing

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/unification/BoolUnification.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/unification/BoolUnification.scala
@@ -67,38 +67,38 @@ object BoolUnification {
       // Case 2: Common unification instances.
       // Note: Order determined by code coverage.
       (tpe1, tpe2) match {
-        case (Type.Var(x, _), Type.Var(y, _)) =>
-          if (renv0.isFlexible(x)) {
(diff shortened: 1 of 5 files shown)


main/src/ca/uwaterloo/flix/language/phase/unification/BoolUnification.scala
```diff
@@ -78,26 +78,26 @@ object BoolUnification {
             return Ok((Substitution.empty, Nil)) // 1000 hits
           }
 
-        case (Type.Cst(TypeConstructor.Empty, _), Type.Cst(TypeConstructor.Empty, _)) =>
+        case (Type.Cst(TypeConstructor.All, _), Type.Cst(TypeConstructor.Empty, _)) =>
```
magnus-madsen, line 81: Ugh. You might have to redo this performance profiling? :-/

magnus-madsen:
Blocked on the performance specific aspects.

magnus-madsen:
I would potentially also wait with this refactor until Bool has been fully enabled and works for relational null (I think those are strong test cases).

mlutze:
> I would potentially also wait with this refactor until Bool has been fully enabled and works for relational null (I think those are strong test cases).

Note that relational null is currently working, just pretending to be a set for now.

magnus-madsen:
(We can continue at a later time.)

PR #5963, opened 2023-05-16
fix: check types in nested defs

### LINKED ISSUE #4838: Use type ascriptions in let rec
Currently they are abandoned in the Weeder

mlutze:
fixes #4838


magnus-madsen:
For unknown reasons, this seems to cause a slowdown?

PR #5964, opened 2023-05-17
fix: rework `ModuleCompleter`

LukasRonn:
image
image
image

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
@@ -60,16 +60,16 @@ object CompletionUtils {
   def getModuleAndFragment(context: CompletionContext): (Symbol.ModuleSym, String) = {
     // We use a fqn to generate a modSym
     // We therefore need to split the context.word at dots
-    val word = context.word.split('.').toList
+    val fqn = context.word.split('.').toList
 
-    if (word == Nil) {
+    if (fqn == Nil) {
       (Symbol.mkModuleSym(Nil), "")
     } else {
       // If the word provided ends with a dot, the subWord is the empty string
       if (context.word.takeRight(1) == ".") {
-        (Symbol.mkModuleSym(word), "")
+        (Symbol.mkModuleSym(fqn), "")
       } else {
-        (Symbol.mkModuleSym(word.dropRight(1)), word.takeRight(1)(0))
+        (Symbol.mkModuleSym(fqn.dropRight(1)), fqn.takeRight(1)(0))
       }
     }
   }
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
@@ -57,16 +57,20 @@ object CompletionUtils {
     * @param context the completionContext.
     * @return        a moduleSym from a fqn and a subWord.
     */
-  def generateModSymAndSubwordFromContext(context: CompletionContext): (Symbol.ModuleSym, String) = {
+  def getModuleAndFragment(context: CompletionContext): (Symbol.ModuleSym, String) = {
     // We use a fqn to generate a modSym
     // We therefore need to split the context.word at dots
     val word = context.word.split('.').toList
 
-    // If the word provided ends with a dot, the subWord is the empty string
-    if (context.word.takeRight(1) == ".") {
-      (Symbol.mkModuleSym(word), "")
+    if (word == Nil) {
+      (Symbol.mkModuleSym(Nil), "")
     } else {
-      (Symbol.mkModuleSym(word.dropRight(1)), word.takeRight(1)(0))
+      // If the word provided ends with a dot, the subWord is the empty string
+      if (context.word.takeRight(1) == ".") {
+        (Symbol.mkModuleSym(word), "")
+      } else {
+        (Symbol.mkModuleSym(word.dropRight(1)), word.takeRight(1)(0))
+      }
     }
   }
 
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
@@ -18,7 +18,7 @@ package ca.uwaterloo.flix.api.lsp.provider.completion
 import ca.uwaterloo.flix.api.Flix
 import ca.uwaterloo.flix.api.lsp.TextEdit
 import ca.uwaterloo.flix.api.lsp.provider.CompletionProvider.Priority
-import ca.uwaterloo.flix.language.ast.{Type, TypeConstructor, TypedAst}
+import ca.uwaterloo.flix.language.ast.{Symbol, Type, TypeConstructor, TypedAst}
 import ca.uwaterloo.flix.language.fmt.FormatType
 
 import java.lang.reflect.{Constructor, Executable, Method}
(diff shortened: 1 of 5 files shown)


main/src/ca/uwaterloo/flix/api/lsp/provider/completion/ModuleCompleter.scala
```diff
@@ -17,14 +17,46 @@ package ca.uwaterloo.flix.api.lsp.provider.completion
...
+
+  /**
+    * Checks if the last elem of the nameSpace matches the subWord (the word the user is currently typing).
+    *
+    * @param mod     the moduleSym.
+    * @param subWord the subWord provided by the user.
+    * @return        true, if the subWord matches the last part of nameSpace, false otherwise.
+    */
+  private def matchesMod(mod: ModuleSym, subWord: String): Boolean = {
+    val lastNs = mod.ns.takeRight(1)(0)
```
magnus-madsen, line 59: Can this not go index out of bounds? You need to check its not the empty list.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/EnumCompleter.scala
```diff
@@ -17,6 +17,7 @@ package ca.uwaterloo.flix.api.lsp.provider.completion
 
 import ca.uwaterloo.flix.api.Flix
 import ca.uwaterloo.flix.api.lsp.provider.completion.Completion.EnumCompletion
+import ca.uwaterloo.flix.api.lsp.provider.completion.CompletionUtils.generateModSymAndSubwordFromContext
```
magnus-madsen, line 20: We don't import functions, so it should be accessed as CompletionUtils.foo

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
```diff
@@ -48,6 +48,28 @@ object CompletionUtils {
...
+  def generateModSymAndSubwordFromContext(context: CompletionContext): (Symbol.ModuleSym, String) = {
+    // We use a fqn to generate a modSym
+    // We therefore need to split the context.word at dots
+    val word = context.word.split('.').toList
+
+    // If the word provided ends with a dot, the subWord is the empty string
+    if (context.word.takeRight(1) == ".") {
+      (Symbol.mkModuleSym(word), "")
+    } else {
+      (Symbol.mkModuleSym(word.dropRight(1)), word.takeRight(1)(0))
```
magnus-madsen, line 69: I also worry this can go index out of bounds.

LukasRonn: It won't go out of bounds. It will just return an empty string at worst.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
```diff
@@ -48,6 +48,28 @@ object CompletionUtils {
...
+  /**
+    * Used to make a moduleSym from a fqn and the "subWord".
+    *
+    * This function splits the context.word at dots to make the best fqn.
+    * If the word ends at a dot, the subWord is the empty string.
+    *
+    * @param context the completionContext.
+    * @return        a moduleSym from a fqn and a subWord.
+    */
+  def generateModSymAndSubwordFromContext(context: CompletionContext): (Symbol.ModuleSym, String) = {
```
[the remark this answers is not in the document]
LukasRonn: Couldn't come up with something better... Any suggestions?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
magnus-madsen: Word is a string right?

LukasRonn: Word is a `List[String]`

magnus-madsen: I believe the last `(0)` can go out of bounds, if the string is empty.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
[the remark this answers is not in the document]
magnus-madsen: Maybe `getModuleAndFragment` or something like that.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
magnus-madsen: Why on earth is "word" a `List[String]` -- thats not good naming.

In any case, if that is the case then `word.takeRight(1)` is `List[String]` and `word.takeRight(1)(0)` can go out of bounds if the list is empty, yes?

LukasRonn: But `context.word` will never be `Nil`

magnus-madsen: Defensive programming. We cannot know that, if it is a list type it can be `Nil`.

LukasRonn: > Why on earth is "word" a `List[String]` -- thats not good naming.
> 
> In any case, if that is the case then `word.takeRight(1)` is `List[String]` and `word.takeRight(1)(0)` can go out of bounds if the list is empty, yes?

Bad naming, I've added guards, to ensure that no matter what, we won't ever throw an exception. I'll push a new version with better naming.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/ModuleCompleter.scala
LukasRonn: I've added a guard

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
```diff
@@ -48,6 +48,32 @@ object CompletionUtils {
...
+    } else {
+      // If the word provided ends with a dot, the subWord is the empty string
+      if (context.word.takeRight(1) == ".") {
+        (Symbol.mkModuleSym(fqn), "")
+      } else {
+        (Symbol.mkModuleSym(fqn.dropRight(1)), fqn.takeRight(1)(0))
+      }
+    }
+  }
+
```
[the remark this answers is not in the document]
LukasRonn: Well i guess. It will just collect all subModules, since someString.startsWith(""), will always be true for a not-null someString.

magnus-madsen:
@LukasRonn Here is my take:

```
object Main {

  // TODO: Where I use List[String] it should be Symbol.ModuleSym

  sealed trait ModuleSymFragment

  object ModuleSymFragment {
    /**
     * Represents a partial module sym, i.e. `Aaa.Bbb.Ccc`.
     *
     * The partially comes from the fact that user could type more letters (Cs).
     */
    case class Partial(sym: List[String], suffix: String) extends ModuleSymFragment

    /**
     * Represents a complete module sym, i.e. `Aaa.Bbb.Ccc.`.
     *
     * The completeness comes from the fact that the user has typed ".".
     */
    case class Complete(sym: List[String]) extends ModuleSymFragment
  }

  /**
   * Parses a partial or complete module symbol.
   *
   * `Aaa.Bbb.Ccc.` => Complete(Aaa.Bbb.Ccc)
   * `Aaa.Bbb`      => Complete(Aaa.Bbb)
   * `Aaa`          => Partial(Nil, Aaa)
   * `Aaa.Bbb.Ccc`  => Partial(Aaa.Bbb, Ccc)
   * <empty>        => Partial(Nil, "")
   */
  def parseModuleSym(word: String): ModuleSymFragment =
    if (word.endsWith(".")) {
      // Case 1: Complete.
      ModuleSymFragment.Complete(word.split('.').toList)
    } else {
      // Case 2: Partial: Split into module part and suffix part.
      // Fortunately, splitting on '.' will just drop '.' if it is last.
      // Thus we simply need the init and the last element, if the list is non-empty.
      val parts = word.split('.').toList
      if (parts.isEmpty) {
        ModuleSymFragment.Partial(Nil, "")
      } else {
        ModuleSymFragment.Partial(parts.init, parts.last) // Safe because parts is non-empty.
      }
    }

  def main(args: Array[String]): Unit = {
    println(parseModuleSym("Aaa.Bbb.Ccc"))
    println(parseModuleSym("Aaa.Bbb"))
    println(parseModuleSym("Aaa."))
    println(parseModuleSym("Aaa.Bbb."))
    println(parseModuleSym(""))
  }

}
```

LukasRonn:
I suggest we discuss this in person.

magnus-madsen:
Can you paste the example program as text, so I can try it out?

LukasRonn:
> 

```
mod Aaaaa {
    mod Bbbbb {
        mod Ccccc { 
            pub enum Cpr with ToString { 
                case Fiat(Int32), Volvo 
            } 
            pub enum Car with ToString {case VW, Audi, Skoda}
        }
    }
    mod Bac {
        pub enum Test with ToString {case Red, Green, Blue}
    }
}

mod Ccc {
    pub enum Test24 with ToString {case Red, Green, Blue}
}

enum Color with ToString {
    case Red(Int32, Int32, String), Green, Blue(String)
}

def main(): Unit \ IO = 
    println()
```

PR #5971, opened 2023-05-17
feat: add reduced ast pretty printer

JonathanStarup:


main/src/ca/uwaterloo/flix/language/dbg/AstPrinter.scala
```diff
@@ -31,7 +31,11 @@ object AstPrinter {
     */
   def printAsts()(implicit flix: Flix): Unit = {
     val asts = flix.options.xprintphase
-    if (asts.nonEmpty) {
+    if (asts.isEmpty)
+      ()
+    else if (asts.contains("all") || asts.contains("All"))
```
magnus-madsen, line 36: Alternative suggestion: Just print all if no explicit phase is provided (?)

magnus-madsen: I guess you have to distinguish between "empty" and "all" which is cumbersome.

PR #5977, opened 2023-05-18
fix: Label shows argument types and detail shows enum type in `EnumTagCompleter`

LukasRonn:
image

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
@@ -18,7 +18,7 @@ package ca.uwaterloo.flix.api.lsp.provider.completion
 import ca.uwaterloo.flix.api.Flix
 import ca.uwaterloo.flix.api.lsp.TextEdit
 import ca.uwaterloo.flix.api.lsp.provider.CompletionProvider.Priority
-import ca.uwaterloo.flix.language.ast.{Type, TypeConstructor, TypedAst}
+import ca.uwaterloo.flix.language.ast.{Kind, Type, TypeConstructor, TypedAst}
 import ca.uwaterloo.flix.language.fmt.FormatType
 
 import java.lang.reflect.{Constructor, Executable, Method}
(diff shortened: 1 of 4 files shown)


main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
```diff
@@ -52,6 +52,18 @@ object CompletionUtils {
...
+    val regExpWithoutParen = """^[^()]*$""".r
+    val regExpWithParen = """^\(.*\)$""".r
+    val args = FormatType.formatType(cas.tpe)
+    args match {
+      case "Unit" => name
+      case regExpWithoutParen() => s"$name($args)"
+      case regExpWithParen() => s"$name$args"
+      case _ => name
+    }
+  }
```
LukasRonn, line 65: This is definitely not the right approach, but I was not able to come up with something better atm..

LukasRonn: However, it works perfectly.

magnus-madsen: This is crazy indeed. Can you not just use the FormatType class?

LukasRonn: Hmm, I'll try that :)

LukasRonn: Currently not able to find anything from Format that can help.. 

We want `Enum.Tag(Type)`.
If we use `FormatType.formatType(tagTyp)` it would produce:
```
   case Unit => "Unit"
   case (a single type) => "type"
   case Tuple(x,y) => "(x, y)"
```
The issue is, when the type is `Unit`, we do not want to add any description. If it's just a single type, we need to add the parens. When it's a `Tuple` we do not want to add any parens..

magnus-madsen: Maybe you can do a combination? Check I'd the type constructor is unit and otherwise use format type.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
```diff
@@ -52,6 +52,14 @@ object CompletionUtils {
 
   private def isUnitFunction(fparams: List[TypedAst.FormalParam]): Boolean = fparams.length == 1 && isUnitType(fparams(0).tpe)
 
+  def getLabelForEnumTags(name: String, cas: TypedAst.Case)(implicit flix: Flix): String = {
+    cas.tpe match {
+      case Type.Unit => name
+      case tpe: Type.Cst => s"$name(${FormatType.formatType(tpe)})"
+      case _ =>  s"$name${FormatType.formatType(cas.tpe)}"
+    }
+  }
```
LukasRonn, line 61: Hopefully this is better :)

PR #5981, opened 2023-05-18
feat: Introduced `MatchRanker`

LukasRonn:
image
image


LukasRonn:
When the user wants to use the `Match`, he now gets all possible enums, that he can match on. Even though they are part of other modules. However, the `MatchRanker` will only rank enums that is not in any namespace.

Issue #5709, opened 2023-04-11
Eliminate BigInt/BigDecimal/Regex in `Simplifier`


magnus-madsen:
In the `Simplifier` phase, we want to replace operations on `BigInt/BigDecimal/Regex` with direct calls to `InvokeMethod` and `InvokeStatic`.

- [x] BigDecimal
- [x] BigInt
- [ ] Regex

magnus-madsen:
I wonder if these are even needed in the AST at all...

magnus-madsen:
For example, can we implement `Add` for `BigInt` (`Add.flix`) directly with import?

jaschdoc:
*import* as in Java import?

magnus-madsen:
Compare:

```
instance Add[Int64] {
    pub def add(x: Int64, y: Int64): Int64 = $INT64_ADD$(x, y)
}

instance Add[BigInt] {
    pub def add(x: BigInt, y: BigInt): BigInt = $BIGINT_ADD$(x, y)
}

instance Add[String] {
    pub def add(x: String, y: String): String =
        import java.lang.String.concat(String): String \ {};
        (x `concat` y)
}

```

magnus-madsen:
I think we can get simply get rid of ` $BIGINT_ADD$(x, y)` no?

magnus-madsen:
We only need to keep the literals as part of the language/compiler.

jaschdoc:
Yeah, I think you are probably right.

magnus-madsen:
I think that would be very nice. Want to try it out for `BigInt.add` and see how it goes?

jaschdoc:
Yes, I'll do it after https://github.com/flix/flix/pull/5828

jaschdoc:
I'm not sure where the regex stuff in `Simplifier` is. I took a brief look at it, and there does not seem to be any regex related code there.

magnus-madsen:
I think there is some constant

PR #5984, opened 2023-05-18
refactor: remove `$BIGING_ADD$`

jaschdoc:
Related to https://github.com/flix/flix/issues/5709


magnus-madsen:
Yes, but now this can go too:

https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/ast/SemanticOperator.scala#L734

jaschdoc:
> Yes, but now this can go too:
> 
> https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/ast/SemanticOperator.scala#L734

Okay, I think so, but I was unsure :)

magnus-madsen:
Yes, so this is nice, because it saves code in the backend for only a little bit of code in the library 👍

Issue #5953, opened 2023-05-15
Add Codepoint module to the stdlib?


stephentetley:
Currently we just consider Char for "character" in the stdlib and as Flix's `Char` maps to Java's `Char` this means we can't represent supplementary characters so we only have limited support for Unicode (see the Javadoc for the class `java.lang.Character`).

I think we should add functions to Flix's stdlib so we can handle supplementary characters as well using the same representation as Java does - codepoints. Codepoints are an `int` in Java and I think we should keep them as an `Int32` in Flix - we could wrap them with an enum but the literal representation is almost always a hex number so an enum be inconvenient for that.

In the JDK, Char and codepoint operations are mixed in the same class `java.lang.Character`. With method overloading in Java this works out fine but it would mean using a naming convention in Flix to differentiate between the codepoint and Char versions of a function so I think a separate CodePoint module would be best.

I've a branch on the go that adds a Codepoint module (and fills out `Char.flix` with some more functions from `java.lang.Character` that we might as well have). 

The current API for Codepoint which covers most of what is provided by `java.lang.Character` is below. Note `Codepoint.toChars` gets us a safe version of "Int32 to Char(s)"

~~~

pub def isAscii(cp: Int32): Bool
pub def isLetter(cp: Int32): Bool
pub def isDigit(cp: Int32): Bool
pub def isLetterOrDigit(cp: Int32): Bool
pub def isAsciiDigit(cp: Int32): Bool
pub def isOctDigit(cp: Int32): Bool
pub def isHexDigit(cp: Int32): Bool
pub def isLowerCase(cp: Int32): Bool
pub def isUpperCase(cp: Int32): Bool
pub def isTitleCase(cp: Int32): Bool
pub def isWhiteSpace(cp: Int32): Bool
pub def isAlphabetic(cp: Int32): Bool
pub def isBmpCodePoint(cp: Int32): Bool
pub def isDefined(cp: Int32): Bool
pub def isIdeographic(cp: Int32): Bool
pub def isISOControl(cp: Int32): Bool
pub def isMirrored(cp: Int32): Bool
pub def isSupplementaryCodePoint(cp: Int32): Bool
pub def isValidCodePoint(cp: Int32): Bool
pub def toLowerCase(cp: Int32): Int32
pub def toUpperCase(cp: Int32): Int32
pub def getName(cp: Int32): Option[String]
pub def getNumericValue(cp: Int32): Option[Int32]
pub def charCount(cp: Int32): Int32
[...]

PR #5986, opened 2023-05-18
feat: add CodePoint module to improve Unicode character handling (issue #5953)

stephentetley:
This PR adds a CodePoint module that complements the existing Char module with functions to work on Code Points which fully represent Unicode characters (Char can only represent the subset of Unicode characters in the Basic Multilingual Plane). Code Points are represented as an `Int32`.

The PR also extends Char's API to include more functions from `java.lang.Character`.

StringBuilder gets a new function `appendCodePoint!` - as a Code Point is just an Int32 we can't use the overloaded `append!` typeclass method.

String gets variants of `charAt` and `nth` to retrieve Code Points.


magnus-madsen:
Thanks Stephen

Issue #5972, opened 2023-05-17
Tweak output of AstPrinter


magnus-madsen:
- [ ] Change "branching" to just "branch { }" (and drop the `with` imo).
- [ ] Change "goto" to "jumpto" (Goto might be better, but then we should rename internally, which we can do later).
- [ ] ~Change `label case$152847: ` to just `case$152847: `~

JonathanStarup:
Due to the nature of the pretty printing, anything might always be on one line, so for one line branches, if there is no `with`, I'm not sure how to delimit the head and the labels. Just removing them could turn into `{goto label42 label42: 42 default12: crash}`

PR #5990, opened 2023-05-19
feat: reformat pretty branch

JonathanStarup:
Part of #5972

* change `branching {...}` to `branch {...}`
* change `label something:` so `something:`


magnus-madsen:
Could we change `with` to some word that not uses? perhaps `cases` or whatever?

PR #5992, opened 2023-05-19
chore: remove Wild expression

mlutze:
Unused expression taking up space


magnus-madsen:
And this is not used in Datalog neither?

magnus-madsen:
```
def main(): Unit \ IO = 
    let pr = #{
        A(1, 2, 3).
        R(x) :- A(x, _, _).
    };
    query pr select x from R(x) |> println
```

This should print `1 :: Nil` if it does, we are probably OK.

mlutze:
1 :: Nil indeed.

Wild patterns still exist, unfortunately :)

PR #6003, opened 2023-05-21
refactor: remove `BigIntOp`

jaschdoc:
Related to https://github.com/flix/flix/issues/5709

### DIFF
--- a/main/src/library/Order.flix
+++ b/main/src/library/Order.flix
@@ -349,7 +349,9 @@ instance Order[BigInt] {
         import java.math.BigInteger.compareTo(BigInt): Int32 \ {};
         (x `compareTo` y) == 1
 
-    pub override def greaterEqual(x: BigInt, y: BigInt): Bool = $BIGINT_GE$(x, y)
+    pub override def greaterEqual(x: BigInt, y: BigInt): Bool =
+        import java.math.BigInteger.compareTo(BigInt): Int32 \ {};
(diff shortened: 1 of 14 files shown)


main/src/library/Order.flix
```diff
@@ -337,21 +337,29 @@ instance Order[Int64] {
...
     pub def compare(x: BigInt, y: BigInt): Comparison =
-        if ($BIGINT_LT$(x, y))
-            Comparison.LessThan
-        else if ($BIGINT_GT$(x, y))
-            Comparison.GreaterThan
-        else
-            Comparison.EqualTo
+        import java.math.BigInteger.compareTo(BigInt): Int32 \ {};
+        match x `compareTo` y {
+            case -1 => Comparison.LessThan
```
magnus-madsen, line 359: Are we sure it returns -1? Could it not return any negative number?

magnus-madsen:
Merged, but do check my question :)

main/src/library/Order.flix
jaschdoc: Yes, I checked the documentation, but we can change it to `... if x < 0 => Comparison.LessThan` if you think that's better (objectively it covers the entire input space if we also add the same check for `> 0`)

magnus-madsen: That at least "feels" safer.

jaschdoc: True :)

PR #6011, opened 2023-05-22
fix: remove redundancy in substitution

### LINKED ISSUE #6005: Drop `++` in `@@` in Substitution?
Can we do it?

mlutze:
fixes #6005


mlutze:
Note: this may have an effect on BU evaluation.

mlutze:
If the test runner performance is to be trusted, then this seems to have a substantial performance impact. 12:30 vs (13-14 mins)

magnus-madsen:
Could we check that the substitution is structurally equal?

magnus-madsen:
For all our test cases

mlutze:
Assertions show that the substitutions are always structurally the same.

PR #6020, opened 2023-05-25
refactor: destruct simplified def type

JonathanStarup:
After loweredAst, defs store their full type, i.e. `def f(x: Int32): Int32 \ Impure` stores the type `Int32 -> Int32 \ Impure`. This means that code that needs either purity or the return type needs to unsafely destruct the expected arrow type. 

This PR extracts the return type and the purity of the def, since the parameter types are already known form the formal parameters. 

I've asserted that, for all tests, this new implied type (reconstructing it from the formal parameters and the new fields) and the old type is identical. 

The purity should be `Purity` instead of `Type` but monomorph doesn't actually guarantee that at the moment

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/SimplifiedAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/SimplifiedAst.scala
@@ -30,7 +30,7 @@ object SimplifiedAst {
                   entryPoint: Option[Symbol.DefnSym],
                   sources: Map[Source, SourceLocation])
 
-  case class Def(ann: Ast.Annotations, mod: Ast.Modifiers, sym: Symbol.DefnSym, fparams: List[SimplifiedAst.FormalParam], exp: SimplifiedAst.Expression, tpe: Type, loc: SourceLocation)
+  case class Def(ann: Ast.Annotations, mod: Ast.Modifiers, sym: Symbol.DefnSym, fparams: List[SimplifiedAst.FormalParam], exp: SimplifiedAst.Expression, tpe: Type, purity: Type, loc: SourceLocation)
 
   case class Enum(ann: Ast.Annotations, mod: Ast.Modifiers, sym: Symbol.EnumSym, cases: Map[Symbol.CaseSym, SimplifiedAst.Case], tpeDeprecated: Type, loc: SourceLocation)
 
--- a/main/src/ca/uwaterloo/flix/language/dbg/printer/SimplifiedAstPrinter.scala
+++ b/main/src/ca/uwaterloo/flix/language/dbg/printer/SimplifiedAstPrinter.scala
@@ -34,13 +34,13 @@ object SimplifiedAstPrinter {
         DocAst.Enum(ann, mod, sym, cases)
     }.toList
     val defs = root.defs.values.map {
-      case SimplifiedAst.Def(ann, mod, sym, formals, exp, tpe, _) =>
+      case SimplifiedAst.Def(ann, mod, sym, formals, exp, tpe, _, _) =>
         DocAst.Def(
           ann,
           mod,
           sym,
           formals.map(printFormalParam),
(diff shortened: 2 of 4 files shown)


main/src/ca/uwaterloo/flix/language/dbg/printer/SimplifiedAstPrinter.scala
```diff
@@ -34,13 +34,13 @@ object SimplifiedAstPrinter {
...
     val defs = root.defs.values.map {
-      case SimplifiedAst.Def(ann, mod, sym, formals, exp, tpe, _) =>
+      case SimplifiedAst.Def(ann, mod, sym, formals, exp, tpe, _, _) =>
         DocAst.Def(
           ann,
           mod,
           sym,
           formals.map(printFormalParam),
-          TypePrinter.print(tpe),
+          DocAst.Type.Arrow(Nil, TypePrinter.print(tpe)),
```
JonathanStarup, line 43: This is a temporary solution since the DocAst expects the full function type and just uses the result of the arrow

magnus-madsen:
If @mlutze approves then you can merge (either of you).

main/src/ca/uwaterloo/flix/language/ast/SimplifiedAst.scala
```diff
@@ -30,7 +30,7 @@ object SimplifiedAst {
                   entryPoint: Option[Symbol.DefnSym],
                   sources: Map[Source, SourceLocation])
 
-  case class Def(ann: Ast.Annotations, mod: Ast.Modifiers, sym: Symbol.DefnSym, fparams: List[SimplifiedAst.FormalParam], exp: SimplifiedAst.Expression, tpe: Type, loc: SourceLocation)
+  case class Def(ann: Ast.Annotations, mod: Ast.Modifiers, sym: Symbol.DefnSym, fparams: List[SimplifiedAst.FormalParam], exp: SimplifiedAst.Expression, tpe: Type, purity: Type, loc: SourceLocation)
```
mlutze, line 33: purity -> pur

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -40,10 +40,14 @@ object Simplifier {
...
-    def visitDef(def0: LoweredAst.Def): SimplifiedAst.Def = {
-      val fs = def0.spec.fparams.map(visitFormalParam)
-      val exp = visitExp(def0.impl.exp)
-      SimplifiedAst.Def(def0.spec.ann, def0.spec.mod, def0.sym, fs, exp, def0.impl.inferredScheme.base, def0.sym.loc)
+    def visitDef(defn: LoweredAst.Def): SimplifiedAst.Def = defn match {
+      case LoweredAst.Def(sym, spec, impl) =>
+        val fs = spec.fparams.map(visitFormalParam)
+        val exp = visitExp(impl.exp)
+        val funType = impl.inferredScheme.base
+        val retType = funType.arrowResultType
```
mlutze, line 48: Remind me why we cannot get this from the spec. I thought spec and impl have the same schemes after monomorph?

mlutze (commented):
Are we doing this only for SimplifiedAst or all the way to the end?

mlutze (approved):

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
JonathanStarup: to maintain current types, this can be changed later

main/src/ca/uwaterloo/flix/language/ast/SimplifiedAst.scala
JonathanStarup: this file has purity but this cannot be Purity yet because of monomorph troubles, can also be fixed later

PR #6022, opened 2023-05-25
feat: remove relational choice from backend

### LINKED ISSUE #5888: Remove choice
Do you approve @magnus-madsen? Alternatively the required translation should be moved from the simplified into lowering (I don't know what difficulties that might cause)

JonathanStarup:
Fixes #5888


magnus-madsen:
I would like to keep the test cases for the type system. Like I would like to keep everything in the frontend, is that possible?

magnus-madsen:
For example, maybe we can just remove the `@test` annotations. That should still check that code compiles -- which is the interesting research part.

JonathanStarup:
> For example, maybe we can just remove the `@test` annotations. That should still check that code compiles -- which is the interesting research part.

Yes, i Just remembered. Should all be good now :)

magnus-madsen:
Merged.

PR #6026, opened 2023-05-25
fix: do not count guarded rules for exhaustiveness

### LINKED ISSUE #6015: Pattern match exhaustiveness guard bug
```scala
def main42(): Int32 = match None {
    case None if false => 42
    case Some(_) if false => 42
}
```

mlutze:
fixes #6015


JonathanStarup (commented):
Maybe add a negative test?

JonathanStarup (approved):

JonathanStarup:
@mlutze I don't know what our protocol is, but you can merge if you want to

magnus-madsen:
Still unhappy.

Issue #6030, opened 2023-05-29
Add `toVector` to Chain, Nel, etc.


stephentetley:
`Chain` and other data types are missing `toVector` functions.

All data types that have `toArray` should probably also have `toVector`.

I'll look at this...

stephentetley:
PR  #6031

PR #6031, opened 2023-05-29
feat: add toVector functions to more data types (issue #6030)

stephentetley:
This PR adds `toVector` functions and tests to data types that already have `toArray` but are missing `toVector`. 

The modules changed are:

Chain
Foldable
MutDeque
Nec
Nel
Reducible
String


main/src/library/String.flix
```diff
@@ -135,6 +135,14 @@ mod String {
     pub def toArray(r: Region[r], s: String): Array[Char, r] \ r =
         Array.init(r, i -> charAt(i, s), length(s))
 
+    ///
+    /// Returns the given string `s` as an array of characters.
```
mlutze, line 139: array -> vector

mlutze (commented):
I see we do a lot of `toArray` ... `toVector`. I wonder if in some cases we can avoid the double allocation that is done?

main/src/library/String.flix
stephentetley: Ffixed.

stephentetley:
> I see we do a lot of `toArray` ... `toVector`. I wonder if in some cases we can avoid the double allocation that is done?

Thanks Matthew. Potentially there is writing in terms of `init` though this might be difficult. Maybe `toVector` functions would be easier with an `unfold`-like function that has some pure state between iterations.

Also `init` doesn't quite have the right implementation at the moment - there is a call to `Array.toVector` at the end. The array we use for building is kept entirely internal to `init` which is what we would want. Because no references to the array escape I could imagine an implementation of `init` that cheaply "freezes" the array into a vector at the end rather copies it (we would need a `freeze` prim op).

magnus-madsen (approved):

magnus-madsen:
I suggest we merge this and then we can deal with performance considerations at a later time.

Copying an array is most likely cheap (compared to say traversing a `Chain` a few times).

PR #6032, opened 2023-05-30
refactor: specialize enum rep. in backend

JonathanStarup:
Now that enums are specialized we
* Remove args on `MonoType.Enum`
* Remove `TagInfo` since tag types are known just form their symbol
* Added enums traversal to `typedOf(root)`
    * As a result thereof, the `nestedTypesOf(enum)` now does not include the types of the cases since they're covered above

The performance seems unaffected


mlutze (approved):
I don't see any issues, but I also don't have a very good understanding of the back-end :)

PR #6033, opened 2023-05-30
refactor: un-deprecate enum/case types

JonathanStarup:
blocked by #6032


magnus-madsen:
@JonathanStarup @mlutze What is the "type" of an enum pre-monomorph? 

I would assume it would have a scheme.

Or perhaps it should not have a type at all-- only its cases would have types (schemes).


In the backend, post monomorphization, I can understand that enums have types.

mlutze:
The scheme is implicit: the type parameters are the quantifiers and the run type is the body.

Issue #5484, opened 2023-02-19
Use `Vector` in Datalog engine


magnus-madsen:
Each in their own PR:
- [ ] As a replacement for `List` in `Fixpoint/Ram`
- [ ] As the return type of `factsX` in `Solver.flix`.
- [ ] As a replacement for `List` in `Fixpoint/Ast`

JonathanStarup:
https://github.com/flix/flix/blob/master/main/src/library/Fixpoint/Tuple/Tuple.flix

This should also be replaced by Vector

sockmaster27:
I'm having some trouble with the `factsX` part. After replacing the type in the Flix source I naturally get the error:
```
#
# An unexpected error has been detected by the Flix compiler:
#
#   Unable to unify: 'PredSym -> (Datalog[Boxed] -> Vector[t111892])' and 'PredSym -> (Datalog[Boxed] -> List[Bool])'.
...
```
I find it very hard to tell where exactly in the compiler the `List` type is specified. I know it ends up [here](https://github.com/flix/flix/blob/19c1bafe2de9fd3e61394c5835c74785ee861361/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala#L815), but I'm struggling with finding out where this `tpe` originates from.

magnus-madsen:
@sockmaster27 I suggest we look at it together in person.

magnus-madsen:
Also have a look here:

https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/Typer.scala#L1865

PR #6042, opened 2023-06-04
refactor: use `Vector` in `Fixpoint/Ast`

sockmaster27:
Replaces all instances of `List` with `Vector` in `Fixpoint/Ast`, and edits places where these types are used, avoiding conversions wherever possible.

Fixes https://github.com/flix/flix/issues/5484


magnus-madsen:
Same comment about using `Vector.splitAt`.

magnus-madsen (approved):

Issue #5950, opened 2023-05-15
Update example source code for flix.dev


magnus-madsen:

magnus-madsen:
@sockmaster27 Unfortunately we have to do this again since we had several PRs that modified the examples.

PR #6043, opened 2023-06-04
refactor: return `Vector` in `Fixpoint.facts<X>`

### LINKED ISSUE #5484: Use `Vector` in Datalog engine

sockmaster27:
- [x] Change return types of functions
- [x] Change type in compiler

Fixes #5484


magnus-madsen:
I guess these changes mean we have to update the Flix website examples _again_ 🤣 

But I see there is already a ticket: https://github.com/flix/flix/issues/5950

magnus-madsen:
We should also update doc.flix.dev -- see #6079

magnus-madsen:
@stephentetley Heads-up: This PR changes the return type of `query` from `List` to `Vector` (which is better for performance).

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1904,9 +1904,6 @@ object Typer {
     } yield (tconstrs, tpe, eff)
   }
 
-  private def mkList(t: Type, loc: SourceLocation): Type =
```
magnus-madsen, line 1907: Thanks-- appreciate the cleanup effort 👍

magnus-madsen (approved):

PR #6052, opened 2023-06-05
feat: create separate unification path for Bool kind

mlutze:
The names are not great. I propose we follow up:
- rename `BoolUnification` to `EffUnification`
- rename `SimpleBoolUnification` to `BinaryUnification` or something like that. `Bool` is just too overloaded.

This copy-pastes a lot from `BoolUnification`. We should be able to generalize the implementation eventually.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/Type.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/Type.scala
@@ -882,19 +882,19 @@ object Type {
   }
 
   /**
-    * Returns the type `Or(tpe1, Or(tpe2, ...))`.
-    */
-  def mkOr(tpes: List[Type], loc: SourceLocation): Type = tpes match {
-    case Nil => Type.False
-    case x :: xs => mkOr(x, mkOr(xs, loc), loc)
-  }
-
-  /**
     * Returns the type `And(tpe1, And(tpe2, ...))`.
     */
   def mkAnd(tpes: List[Type], loc: SourceLocation): Type = tpes match {
     case Nil => Type.True
     case x :: xs => mkAnd(x, mkAnd(xs, loc), loc)
+  }
+
+  /**
+    * Returns the type `Or(tpe1, Or(tpe2, ...))`.
+    */
+  def mkOr(tpes: List[Type], loc: SourceLocation): Type = tpes match {
+    case Nil => Type.False
+    case x :: xs => mkOr(x, mkOr(xs, loc), loc)
   }
 
   /**
--- a/main/src/ca/uwaterloo/flix/language/ast/Type.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/Type.scala
@@ -807,8 +807,6 @@ object Type {
 
   /**
     * Returns the type `Union(tpe1, tpe2)`.
-    *
-    * Must not be used before kinding.
     */
   def mkUnion(tpe1: Type, tpe2: Type, loc: SourceLocation): Type = (tpe1, tpe2) match {
     case (Type.Cst(TypeConstructor.Pure, _), _) => tpe2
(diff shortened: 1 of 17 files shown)


main/src/ca/uwaterloo/flix/language/ast/Type.scala
```diff
@@ -354,6 +354,37 @@ object Type {
...
+
+  /**
+    * Represents the False Boolean algebra value.
+    */
+  val False: Type = Type.Cst(TypeConstructor.False, SourceLocation.Unknown)
+
+  /**
+    * Represents the Not type constructor.
+    *
+    * NB: This type has kind: * -> *.
```
magnus-madsen, line 370: That's not the right kind is it?

main/src/ca/uwaterloo/flix/language/ast/Type.scala
```diff
@@ -354,6 +354,37 @@ object Type {
...
+    */
+  val Not: Type = Type.Cst(TypeConstructor.Not, SourceLocation.Unknown)
+
+  /**
+    * Represents the And type constructor.
+    *
+    * NB: This type has kind: * -> (* -> *).
```
magnus-madsen, line 377: ditto

main/src/ca/uwaterloo/flix/language/ast/Type.scala
```diff
@@ -354,6 +354,37 @@ object Type {
...
+    */
+  val And: Type = Type.Cst(TypeConstructor.And, SourceLocation.Unknown)
+
+  /**
+    * Represents the Or type constructor.
+    *
+    * NB: This type has kind: * -> (* -> *).
```
magnus-madsen, line 384: ditto

main/src/ca/uwaterloo/flix/language/ast/Type.scala
```diff
@@ -765,6 +796,15 @@ object Type {
     case _ => Type.Apply(Type.Cst(TypeConstructor.Complement, loc), tpe0, loc)
   }
 
+  /**
+    * Returns the type `Not(tpe0)`.
```
magnus-madsen, line 800: Can this be used before kinding?

main/src/ca/uwaterloo/flix/language/ast/Type.scala
```diff
@@ -829,6 +869,54 @@ object Type {
...
+    case (Type.Var(sym1, _), Type.Var(sym2, _)) if sym1 == sym2 => tpe1
+    case _ => Type.Apply(Type.Apply(Type.Or, tpe1, loc), tpe2, loc)
+  }
+
+  /**
+    * Returns the type `Or(tpe1, Or(tpe2, ...))`.
+    *
+    * Must not be used before kinding.
+    */
+  def mkOr(tpes: List[Type], loc: SourceLocation): Type = tpes match {
```
magnus-madsen, line 905: Move after mkAnd to follow the same order.

main/src/ca/uwaterloo/flix/language/phase/unification/SimpleBoolFormulaAlgClassic.scala
```diff
@@ -0,0 +1,230 @@
+/*
```
magnus-madsen, line 1: What's the plan for this copy paste?

magnus-madsen:
> The names are not great. I propose we follow up:
> 
>     * rename `BoolUnification` to `EffUnification`

OK

> 
>     * rename `SimpleBoolUnification` to `BinaryUnification` or something like that. `Bool` is just too overloaded.
> 

It should probably ultimately be called `BoolUnification` since this is for the `Bool` kind.

> This copy-pastes a lot from `BoolUnification`. We should be able to generalize the implementation eventually.

OK. I might soon need a table.

magnus-madsen:
Presumably you do a few fixes and then I merge this and then you can a do a few easy follow up PRs to cleanup naming etcetera?

magnus-madsen:
Do we have a ticket for this work?

magnus-madsen:
Presumably I also get back my `Bool` kind?

main/src/ca/uwaterloo/flix/language/ast/Type.scala
mlutze: Actually the "don't use before kinding" stuff ends up being redundant, since we now have `UnkindedType`

main/src/ca/uwaterloo/flix/language/phase/unification/SimpleBoolFormulaAlgClassic.scala
mlutze: My idea is this:

- Remove `toType` and `fromType` from `BoolAlg`, instead replacing it with something like `convert`, which converts it to another `BoolAlg` representation.
- Make a `BoolAlg` implementation for `Type`/`Bool` and `Type`/`Eff`
- Use these conversions instead in what will be `EffUnification` and `BoolUnification`

We maintain the shortcuts in `EffUnification`

mlutze:
> Presumably I also get back my `Bool` kind?

Yes, `Choice` is now using `Bool`.

magnus-madsen (approved):

PR #6077, opened 2023-06-07
fix: improve typing for effect handlers

mlutze:


main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -1512,7 +1512,7 @@ object Typer {
           (tconstrss, _, effs) <- traverseM(rules)(visitHandlerRule).map(_.unzip3)
           resultTconstrs = (tconstrs :: tconstrss).flatten
           resultTpe <- unifyTypeM(tvar, tpe, loc)
-          resultEff = Type.mkUnion(eff :: effs, loc)
+          resultEff = Type.mkUnion(Type.mkDifference(eff, effType, loc) :: effs, loc)
```
magnus-madsen, line 1515: I don't understand. Does this actually work (??)

magnus-madsen (approved):

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
mlutze: It works in the tests (so far). ATM it requires the effect to be in the handled expression. And we don't have the "multiple handlers" syntax yet

PR #6094, opened 2023-06-11
chore: jline: 3.5.1 → 3.23.0

fabianhjr:
From 3.5.1 (2017-09-22) to 3.23.0 ( 2023-03-08 )


magnus-madsen:
We don't have test cases for this, so we would need someone to experiment with the REPL.

magnus-madsen:
See #6323

PR #6097, opened 2023-06-13
chore: rename effect-poly error to effectful error

### LINKED ISSUE #6083: Rework `Effect polymorphic function declared as pure`
This:

```
eff Print {
    pub def print(): Unit
}

eff Throw {
    pub def throw(): Unit
}

def f(): Unit = 
    do Print.print();
    do Throw.throw()
```

should give a better error message.

mlutze:
fixes #6083. I wanted to do something prettier, but a combination of us not having set effects anymore and annoyances I am blaming on the escaping region bug covinced me to take this lazy approach.


magnus-madsen:
I think we want to show the effects in the message.

magnus-madsen (approved):

Issue #6113, opened 2023-06-13
Def representation renovation


JonathanStarup:
- [x] Defs should store their return type instead of their full type
  - [X] SimplifiedAst #6020
  - [x] LiftedAst & OccurrenceAst #6116
  - [x] ReducedAst #6130
  - [x] MonoTypedAst #6144
  - [x] ErasedAst #6146
- [x] Defs should store their purity as `Purity` and not a `Type`
  - [x] Fix monomorpher to guarantee simple purities 
  - [x] SimplifiedAst
  - [x] LiftedAst
  - [x] ReducedAst
  - [x] MonoTypedAst
  - [x] ErasedAst
- [X] Defs should differentiate captured and explicit arguments #6034
- [x] Fix DocAst to use proper return type #6207
- [ ] make DocAst use def purity
- [x] Fix nullary lambdas #6115
  - [ ] make them Nel ? 
- [x] use closure information in backend #6117
  - [ ] avoid using nonempty cparams list
    - [ ] with bool flags? #6118
  - [ ] Optimize non-capture closures that just call a function

PR #6117, opened 2023-06-13
refactor: remove ClosureInfo

JonathanStarup:
Related to #6113

- Remove `ClosureInfo` (and their collection) and instead generate closures based on the `cparams` of defs
- Ensure that `cparams` for closures are always non-empty by adding a unit cparam if it is.
  - Another option is to actually separate closures/defs with a flag or a new case class, in which case it would be okay to have empty cparams

Metrics
- 8 % decrease in code size (from the clarification of closures vs defs)
- performance of eraser/backend seems slightly faster, hard to say
- Might slow down flix runtime a little with the added unit args

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
@@ -247,7 +247,7 @@ object Simplifier {
         val e1 = visitExp(exp1)
         val e2 = visitExp(exp2)
         val lambdaTyp = Type.mkArrowWithEffect(Type.Unit, eff, e1.tpe, loc)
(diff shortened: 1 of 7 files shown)


main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -247,15 +247,15 @@ object Simplifier {
         val e1 = visitExp(exp1)
         val e2 = visitExp(exp2)
         val lambdaTyp = Type.mkArrowWithEffect(Type.Unit, eff, e1.tpe, loc)
-        val fp = SimplifiedAst.FormalParam(Symbol.freshVarSym("spawn", BoundBy.FormalParam, loc), Ast.Modifiers.Empty, Type.mkUnit(loc), loc)
```
JonathanStarup, line 250: also a small fix here to reflect that these syms are unused

magnus-madsen (approved):

magnus-madsen:
Pretty nice! Will be interesting to see if it has performance benefits too.

PR #6118, opened 2023-06-13
fix: remove closure unit arg

JonathanStarup:
Remove the added unit arg in #6117

It seems this cost some runtime performance. This does, however, point to the fact that we should optimize this case. (closure with no captured arguments)


magnus-madsen:
I don't like this adhoc Boolean. What performance drop is it that you see?

JonathanStarup:
it was a fluke

PR #6120, opened 2023-06-14
test: add missing tests to TestFoldable.Flix

xdesou01:
Tests (which pass) added for count, sum, sumWith, exists, forEach


magnus-madsen:
I think the PR should be merged or rebased onto master so it does not include previous changes:

https://github.com/flix/flix/pull/6120/files

Issue #6109, opened 2023-06-13
Rework `OccurrenceAst` to use `AtomicOp`


magnus-madsen:
I suggest to break the refactoring into several PRs.

PR #6128, opened 2023-06-17
refactor: use `AtomicOp` in `OccurrenceAst`

jaschdoc:
Closes https://github.com/flix/flix/issues/6109


main/src/ca/uwaterloo/flix/language/phase/Inliner.scala
```diff
@@ -148,8 +148,20 @@ object Inliner {
         else
           LiftedAst.Expression.Is(sym, e, purity, loc)
 
-      case AtomicOp.Tag(sym) => ???
-      case AtomicOp.Untag(sym) => ???
+      case AtomicOp.Tag(sym) =>
+        val List(exp) = exps
```
mlutze, line 152: This match looks fishy. I know it's _right_, but it looks fishy :)

Do you think it would be nice to have some kind of `unpackAtomicOp(1|2|3)` function that explicitly throws an ICE when there is the wrong number of exps? Up to you.

mlutze (approved):

main/src/ca/uwaterloo/flix/language/phase/Inliner.scala
magnus-madsen: But why do we even need to match on tag?

jaschdoc: We do this in GenExpression I believe so this is consistent with other uses

magnus-madsen: If we must deal with some subset of ApplyAtomic then I think we have to match on those and use a wildcard for the rest.

jaschdoc:
Hmm... it's sort of weird to pack it into an atomic expr and then map it back to an Ast Expr in LiftedAst. Do you also want LiftedAst to use AtomicOp?

jaschdoc:
It seems awkward to map from `LiftedAst.Expr._` to `OccurrenceAst.Expr.ApplyAtomic` and then back to `LiftedAst.Expr._`. Do you also want `LiftedAst` to use `AtomicOp`?

magnus-madsen:
Yes,  we should. The goal is to use Apply Atomic everywhere

magnus-madsen:
Everywhere in the backend

jaschdoc:
Roger that. I will refactor `LiftedAst` first then :)

jaschdoc:
In a separate PR of course

jaschdoc:
Ready!

PR #6130, opened 2023-06-20
refactor: ReducedAst def return type

JonathanStarup:
Related to #6113

I've noted that the purity is unused on defs in SimplifiedAst, LiftedAst/OccurrenceAst, and ReducedAst. It is then erased in MonoTypedAst. The optimizer uses the def calls to look at purity, not looking at the definition itself.

Should we remove this unused purity information on defs or keep it around? In my opinion, it seems sensible to store it. It might be important for implementing handlers at some point.  Do you have input @mlutze?

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/MonoTyper.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/MonoTyper.scala
@@ -39,7 +39,8 @@ object MonoTyper {
     val cs = def0.cparams.map(visitFormalParam)
     val fs = def0.fparams.map(visitFormalParam)
     val s = visitStmt(def0.stmt)
-    val tpe = visitType(def0.tpe)
+    val tpe0 = visitType(def0.tpe)
+    val tpe = MonoType.Arrow(fs.map(_.tpe), tpe0)
     MonoTypedAst.Def(def0.ann, def0.mod, def0.sym, cs, fs, s, tpe, def0.loc)
   }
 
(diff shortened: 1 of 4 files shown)


mlutze:
Yes, I agree it's likely worth keeping for now. We can remove it later if ultimately unused.

main/src/ca/uwaterloo/flix/language/phase/MonoTyper.scala
```diff
@@ -39,7 +39,8 @@ object MonoTyper {
     val cs = def0.cparams.map(visitFormalParam)
     val fs = def0.fparams.map(visitFormalParam)
     val s = visitStmt(def0.stmt)
-    val tpe = visitType(def0.tpe)
+    val tpe0 = visitType(def0.tpe)
+    val tpe = MonoType.Arrow(fs.map(_.tpe), tpe0)
```
mlutze, line 43: So in `ReducedAst`, `tpe` means "return type" and in `MonoTypedAst` it means "function arrow type". Are there plans to make this consistent?

mlutze (approved):

main/src/ca/uwaterloo/flix/language/phase/MonoTyper.scala
magnus-madsen: We should make it consistent.

magnus-madsen (approved):

PR #6134, opened 2023-06-20
feat: support lsp code actions

sockmaster27:
Adds the framework for supporting code actions. The actions themselves can now be implemented in `CodeActionProvider.scala`.
I can see that the auto-formatting went a bit wild on some parts. I hope that isn't a problem.

Related to flix/vscode-flix#233

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/CodeActionContext.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/CodeActionContext.scala
@@ -21,10 +21,6 @@ import org.json4s.JsonAST._
 
 object CodeActionContext {
   def parse(json: JValue): Result[CodeActionContext, String] = {
-    val diagnosticsResult = json \ "diagnostics" match {
-      case JArray(l) => Result.traverse(l)(Diagnostic.parse)
-      case v => Err(s"Unexpected non-array diagnostics: '$v'.")
-    }
 
     val onlyResult: Result[List[CodeActionKind], String] = json \ "only" match {
       case JNothing => Ok(List())
@@ -34,18 +30,14 @@ object CodeActionContext {
 
     val triggerKindResult: Result[Option[CodeActionTriggerKind], String] = json \ "triggerKind" match {
       case JNothing => Ok(None)
-      case x => CodeActionTriggerKind.parse(x) match {
-        case Ok(v) => Ok(Some(v))
-        case Err(s) => Err(s)
-      }
+      case v => CodeActionTriggerKind.parse(v).map(Some(_))
     }
 
     for {
-      diagnostics <- diagnosticsResult
       only <- onlyResult
       triggerKind <- triggerKindResult
-    } yield CodeActionContext(diagnostics, only, triggerKind)
+    } yield CodeActionContext(only, triggerKind)
   }
 }
 
-case class CodeActionContext(diagnostics: List[Diagnostic], only: List[CodeActionKind], triggerKind: Option[CodeActionTriggerKind])
+case class CodeActionContext(only: List[CodeActionKind], triggerKind: Option[CodeActionTriggerKind])
--- a/main/src/ca/uwaterloo/flix/api/lsp/CodeActionContext.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/CodeActionContext.scala
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2023 Holger Dal Mogensen
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ca.uwaterloo.flix.api.lsp
+
+import ca.uwaterloo.flix.util.Result
+import ca.uwaterloo.flix.util.Result.{Err, Ok}
+import org.json4s.JsonAST._
+
+object CodeActionContext {
+  def parse(json: JValue): Result[CodeActionContext, String] = {
+    val diagnosticsResult = json \ "diagnostics" match {
+      case JArray(l) => Result.traverse(l)(Diagnostic.parse)
+      case v => Err(s"Unexpected non-array diagnostics: '$v'.")
+    }
+
+    val onlyResult: Result[List[CodeActionKind], String] = json \ "only" match {
+      case JNothing => Ok(List())
+      case JArray(l) => Result.traverse(l)(CodeActionKind.parse)
+      case v => Err(s"Unexpected non-array only-field: '$v'.")
+    }
+
+    val triggerKindResult: Result[Option[CodeActionTriggerKind], String] = json \ "triggerKind" match {
+      case JNothing => Ok(None)
+      case x => CodeActionTriggerKind.parse(x) match {
+        case Ok(v) => Ok(Some(v))
+        case Err(s) => Err(s)
+      }
+    }
+
+    for {
+      diagnostics <- diagnosticsResult
+      only <- onlyResult
+      triggerKind <- triggerKindResult
+    } yield CodeActionContext(diagnostics, only, triggerKind)
+  }
+}
+
+case class CodeActionContext(diagnostics: List[Diagnostic], only: List[CodeActionKind], triggerKind: Option[CodeActionTriggerKind]) {}
--- a/main/src/ca/uwaterloo/flix/api/lsp/CodeActionContext.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/CodeActionContext.scala
@@ -48,4 +48,4 @@ object CodeActionContext {
   }
 }
 
-case class CodeActionContext(diagnostics: List[Diagnostic], only: List[CodeActionKind], triggerKind: Option[CodeActionTriggerKind]) {}
+case class CodeActionContext(diagnostics: List[Diagnostic], only: List[CodeActionKind], triggerKind: Option[CodeActionTriggerKind])
--- a/main/src/ca/uwaterloo/flix/api/lsp/CodeActionKind.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/CodeActionKind.scala
@@ -35,16 +35,19 @@ sealed trait CodeActionKind {
 
 object CodeActionKind {
   def parse(json: JValue): Result[CodeActionKind, String] = json match {
-    case JString("") => Ok(CodeActionKind.Empty)
-    case JString("quickfix") => Ok(CodeActionKind.QuickFix)
-    case JString("refactor") => Ok(CodeActionKind.Refactor)
-    case JString("refactor.extract") => Ok(CodeActionKind.RefactorExtract)
-    case JString("refactor.inline") => Ok(CodeActionKind.RefactorInline)
-    case JString("refactor.rewrite") => Ok(CodeActionKind.RefactorRewrite)
-    case JString("source") => Ok(CodeActionKind.Source)
-    case JString("source.organizeImports") => Ok(CodeActionKind.SourceOrganizeImports)
-    case JString("source.fixAll") => Ok(CodeActionKind.SourceFixAll)
-    case v => Err(s"Code action kind not recognized as a valid string: '$v'.")
+    case JString(s) => s match {
+      case "" => Ok(Empty)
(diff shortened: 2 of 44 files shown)


magnus-madsen:
>  I can see that the auto-formatting went a bit wild on some parts. I hope that isn't a problem.

Unfortunately that is a bit of a problem because it makes code review difficult, and more importantly, it makes it harder to browse the git history, e.g. with git blame/bisect etc. I would better if you can go back and change the right IDE setting and then only have actual changes :)

magnus-madsen:
It may or may not not be preferable to have two PRs: one that simply introduced the new data types and one that contains changes to existing ones.

sockmaster27:
> > I can see that the auto-formatting went a bit wild on some parts. I hope that isn't a problem.
> 
> Unfortunately that is a bit of a problem because it makes code review difficult, and more importantly, it makes it harder to browse the git history, e.g. with git blame/bisect etc. I would better if you can go back and change the right IDE setting and then only have actual changes :)

This is caused by the fact that there is some inconsistent formatting in the doc comments. 
Some look like this:
```
/**
 * Doc comment
 */
```
and others look like this:
```
/**
  * Doc comment
  */
```

This is controlled by the "Add additional space for leading asterisk" option in IntelliJ, and as far as I can tell, there is no way to disable that (you have to choose whether they should be there or not). This seems to make it impossible to, for example, format-on-save without messing it up like I've done here.

Could I instead make a different PR to make this consistent across the code base?

JonathanStarup:
I believe this is the "correct" one.
```
/**
  * Doc comment
  */
```
I've thought about fixing it project-wide for a while so I support the idea 😄 but it looks like if you just change your setting then most of the noise should go away

magnus-madsen:
While there is (apparently) some whitespace issue, we should resolve that separately from this PR.

main/src/ca/uwaterloo/flix/api/lsp/CodeActionContext.scala
```diff
@@ -0,0 +1,51 @@
...
+
+    for {
+      diagnostics <- diagnosticsResult
+      only <- onlyResult
+      triggerKind <- triggerKindResult
+    } yield CodeActionContext(diagnostics, only, triggerKind)
+  }
+}
+
+case class CodeActionContext(diagnostics: List[Diagnostic], only: List[CodeActionKind], triggerKind: Option[CodeActionTriggerKind]) {}
```
mlutze, line 51: extra `{}`

main/src/ca/uwaterloo/flix/api/lsp/CodeActionContext.scala
```diff
@@ -0,0 +1,51 @@
...
+import org.json4s.JsonAST._
+
+object CodeActionContext {
+  def parse(json: JValue): Result[CodeActionContext, String] = {
+    val diagnosticsResult = json \ "diagnostics" match {
+      case JArray(l) => Result.traverse(l)(Diagnostic.parse)
+      case v => Err(s"Unexpected non-array diagnostics: '$v'.")
+    }
+
+    val onlyResult: Result[List[CodeActionKind], String] = json \ "only" match {
```
mlutze, line 29: 1. I'm surprised we have to roll our own parser here. Does `json4s` not provide conversion-to-case-class facilities?
2. If not, I think we can abstract away some of the common functionality here, like `getOptionalString` and `getOptionalList` seem to be generally useful.

main/src/ca/uwaterloo/flix/api/lsp/CodeActionTriggerKind.scala
```diff
@@ -0,0 +1,27 @@
+package ca.uwaterloo.flix.api.lsp
```
mlutze, line 1: License

main/src/ca/uwaterloo/flix/api/lsp/CodeActionTriggerKind.scala
```diff
@@ -0,0 +1,27 @@
...
+import org.json4s.{JInt, JValue}
+import ca.uwaterloo.flix.util.Result
+import ca.uwaterloo.flix.util.Result.{Err, Ok}
+
+sealed trait CodeActionTriggerKind
+
+object CodeActionTriggerKind {
+  def parse(json: JValue): Result[CodeActionTriggerKind, String] = {
+    def err(v: JValue) =
+      Err(s"Code action trigger kind not recognized as a valid integer: '$v'.")
```
mlutze, line 12: May be slightly misleading if we see e.g. `not recognized as a valid integer: '3'.`

main/src/ca/uwaterloo/flix/api/lsp/DiagnosticSeverity.scala
```diff
@@ -40,6 +43,22 @@ object DiagnosticSeverity {
     case Severity.Hint => Hint
   }
 
+  def parse(json: JValue): Result[DiagnosticSeverity, String] = {
+    def err(v: JValue) =
+      Err(s"Diagnostic severity not recognized as a valid integer: '$v'.")
```
mlutze, line 48: see above

main/src/ca/uwaterloo/flix/api/lsp/Diagnostic.scala
```diff
@@ -48,6 +49,58 @@ object Diagnostic {
     val summary = codeHint.summary
     Diagnostic(range, severity, None, None, summary, summary, Nil)
   }
+
+  def parse(json: JValue): Result[Diagnostic, String] = {
```
mlutze, line 53: See above regarding parsing

main/src/ca/uwaterloo/flix/api/lsp/DiagnosticTag.scala
```diff
@@ -26,6 +30,19 @@ sealed trait DiagnosticTag {
 }
 
 object DiagnosticTag {
+  def parse(json: JValue): Result[DiagnosticTag, String] = {
+    def err(v: JValue) =
+      Err(s"Diagnostic tag not recognized as a valid integer: '$v'.")
```
mlutze, line 35: see above regarding integer

main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
```diff
@@ -0,0 +1,30 @@
+package ca.uwaterloo.flix.api.lsp.provider
```
mlutze, line 1: license

mlutze (changes requested):

main/src/ca/uwaterloo/flix/api/lsp/CodeActionContext.scala
magnus-madsen: We usually do roll our own parsers. My surprise is that w eneed to parse diagnostics.

But I think @sockmaster27 is just faithfully implementing the API.

But I suspect we can get away with something simpler.

Q: CodeActionContext is something sent _to us_ right?

main/src/ca/uwaterloo/flix/api/lsp/CodeActionKind.scala
```diff
@@ -0,0 +1,67 @@
...
+ */
+package ca.uwaterloo.flix.api.lsp
+
+import ca.uwaterloo.flix.util.Result
+import ca.uwaterloo.flix.util.Result.{Ok, Err}
+import org.json4s.{JString, JValue}
+
+sealed trait CodeActionKind {
+  def toJSON: JValue = this match {
+    case CodeActionKind.Empty => JString("")
```
magnus-madsen, line 24: Empty string or undefined?

main/src/ca/uwaterloo/flix/api/lsp/CodeActionContext.scala
sockmaster27: Yes

magnus-madsen: If codeactioncontext is not sent to us, why do we need to parse it?

main/src/ca/uwaterloo/flix/api/lsp/CodeActionTriggerKind.scala
```diff
@@ -0,0 +1,27 @@
...
+
+    json match {
+      case JInt(i) => i.toInt match {
+        case 1 => Ok(Invoked)
+        case 2 => Ok(Automatic)
+        case _ => err(json)
+      }
+      case _ => err(json)
```
magnus-madsen, line 20: I think it would be better with two separate error messages.

magnus-madsen: But more to the point, according to the protocol these unknown values should not really occur. Hence it might be better for us to throw exceptions?

main/src/ca/uwaterloo/flix/api/lsp/LanguageServer.scala
```diff
@@ -70,17 +70,14 @@ class LanguageServer(port: Int, o: Options) extends WebSocketServer(new InetSock
...
-
   /**
     * A map from source URIs to source code.
     */
   val sources: mutable.Map[String, String] = mutable.Map.empty
-
+  /**
+    * The Flix instance (the same instance is used for incremental compilation).
+    */
+  private val flix: Flix = new Flix().setFormatter(NoFormatter).setOptions(o)
```
magnus-madsen, line 80: I still see some unnecessary changes to this file. I think it might be best to checkout the original file and then port your changes to it.

magnus-madsen: The reason I care is because I don't want to break "git blame" etc.

magnus-madsen:
This is looking really nice 👍 

Matt and I just have several comments with us trying to understand the code.

main/src/ca/uwaterloo/flix/api/lsp/CodeActionContext.scala
sockmaster27: It _is_ sent to us

main/src/ca/uwaterloo/flix/api/lsp/CodeActionKind.scala
sockmaster27: Empty string: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeActionKind

main/src/ca/uwaterloo/flix/api/lsp/CodeActionTriggerKind.scala
sockmaster27: I think it's more in line with the rest of the code to treat it as a `Result`

main/src/ca/uwaterloo/flix/api/lsp/LanguageServer.scala
sockmaster27: Sorry, I somehow missed these 🤔

main/src/ca/uwaterloo/flix/api/lsp/CodeActionTriggerKind.scala
magnus-madsen: In general you are totally write. But in this case its really should not fail because if it does something is very wrong.

In any case, I suspect that if we remove the list of diagnostics then we can remove this parsing code, and it becomes moot.

magnus-madsen: right*

main/src/ca/uwaterloo/flix/api/lsp/CodeActionContext.scala
sockmaster27: Removing parsing of `Diagnostic` pretty much eliminates this, so unless there are any objections I'll mark this as resolved

magnus-madsen: Sounds good 👍

magnus-madsen (approved):

magnus-madsen:
@mlutze Can you unblock this PR?

magnus-madsen:
@sockmaster27 I tried commenting in the code in `CodeActionProvider`, but I don't see any code actions?

sockmaster27:
@magnus-madsen 
Hmm that's weird. I tried that just yesterday. Are you sure you're using the right version of the plugin and loading the right version of the compiler?

magnus-madsen:
I thought so. I grabbed both masters after merging, build myself a jar, and am running the vscode extension.

On the other hand there are no errors either-- which would usually be the case.

magnus-madsen:
Wait! I see it. The cursor just has to be at the right position.

magnus-madsen:
I guess it only shows depending on the cursor or?

sockmaster27:
@magnus-madsen
It should show up as a little light bulb next to the line that the cursor is on (or sometimes just above that line). It might not show up if the line is empty? I don't remember, but if you're seeing it I guess it should be fine, right?

magnus-madsen:
Yes. At least we can figure it out once we start to add actual fixes.

Issue #6127, opened 2023-06-16
Set.toList is backwards


mlutze:

magnus-madsen:
Maybe a good ticket for Xavier?

magnus-madsen:
I wonder if other things are reversed.

magnus-madsen:
Fixed in master.

PR #6138, opened 2023-06-21
fix: unreverse Set.toList order (related to #6127)

xdesou01:
Also rearranged the few tests which relied on the reversed order


mlutze (approved):

magnus-madsen:
What about `toChain`, `toXYZ` and so forth?

xdesou01:
> What about `toChain`, `toXYZ` and so forth?
Yes, everything else appears to be correct.

magnus-madsen (approved):

PR #6152, opened 2023-06-24
feat: add a few code actions

magnus-madsen:


magnus-madsen:
@mlutze Can you review. This is mostly intended as a starting point for @sockmaster27. Hence the code is deliberately not 100% finished, but it should work without crashing.

magnus-madsen:
By review,  I mean approve ;)

mlutze (approved):

Issue #6048, opened 2023-06-05
Eliminate `StringOp.Eq/Neq`


magnus-madsen:
I think we can simply delegate to `String.equals`, right?

PR #6165, opened 2023-06-27
refactor: remove `StringOp.Eq/Neq` from `SemanticOperator`

jaschdoc:
Closes https://github.com/flix/flix/issues/6048


main/src/library/Eq.flix
```diff
@@ -117,8 +117,13 @@ instance Eq[BigInt] {
 }
 
 instance Eq[String] {
-    pub def eq(x: String, y: String): Bool = $STRING_EQ$(x, y)
-    pub override def neq(x: String, y: String): Bool = $STRING_NEQ$(x, y)
+    pub def eq(x: String, y: String): Bool =
+        import java.lang.String.equals(##java.lang.Object): Bool \ {};
+        equals(x, checked_cast(y))
+
+    pub override def neq(x: String, y: String): Bool =
```
magnus-madsen, line 124: No need to override then? I think we can just leave neq blank :)

magnus-madsen:
> I think `StringOp.Eq` is used for pattern matching somehow in the compiler. I'll investigate further

Hmm

magnus-madsen:
> The problem arises in `Simplifier` but I'm thinking we can generate a call to `Eq.eq` before that

Can you link the code? :)

jaschdoc:
> > The problem arises in `Simplifier` but I'm thinking we can generate a call to `Eq.eq` before that
> 
> Can you link the code? :)

https://github.com/flix/flix/blob/000ca53fec879bf971aebd8db9862064d16d66b7/main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala#L356

magnus-madsen:
That's interesting. We may have to replace that `mkEquals` with something that actually uses ` Eq.eq`.

jaschdoc:
Problem is, I don't see how to emit a function call from the Ast nodes in `SimplifiedAst` (`Eq.eq` has been assigned a symbol during compilation, right?)

magnus-madsen:
You may be right. Can we start by dropping the String.Neq then and see.

Let us ask @mlutze when he is home.

mlutze:
Can we just change it to `InvokeStaticMethod(Object.equals)`?

magnus-madsen (approved):

PR #6171, opened 2023-06-27
test: add type-level multiplication tests

xdesou01:


magnus-madsen (approved):

magnus-madsen:
@xdesou01 Looks good. I will merge once the CI completes. I think the next interesting step would be to express DeMorgans laws.

Issue #6155, opened 2023-06-25
Packager / Maven dependencies - version numbers should allow things that aren't numbers


stephentetley:
Currently the Packager disallows Maven dependencies that don't have proper semantic version numbers. Unfortunately this is too restrictive in practice as it is common to find Java packages that don't strictly use semantic version numbers, e.g:

https://mvnrepository.com/artifact/io.netty/netty-common/4.1.94.Final

magnus-madsen:
Yikes. Okay.

stephentetley:
Do we lose anything compelling if we change `MavenDependency` to have a String version tag rather than a `SemVar` one?

magnus-madsen:
> Do we lose anything compelling if we change `MavenDependency` to have a String version tag rather than a `SemVar` one?

No. I think you can do that :)

stephentetley:
Thanks - I'll make a PR tomorrow evening.

magnus-madsen:
Fixed?

stephentetley:
Yes

PR #6172, opened 2023-06-27
fix: packager - relax MavenDependency versionTag to just String (issue #6155)

stephentetley:
This PR relaxes the Packager so `MavenDependency` now parses the version number of a Maven package dependency a just as `String` and not a `SemVer`. Maven (the repository) does not enforce adherence to SemVars so the Packager has to be relaxed about what it parses. 

I've reordered the tests in `TestManifestParser.scala` so it looks like there are a lot of changes but really this is just changing the Maven dependency parsing examples to "Ok" tests as they no longer fail when the tag is not a strict `SemVer`.


magnus-madsen (approved):

magnus-madsen:
Thanks Stephen

PR #6178, opened 2023-06-29
docs: Add example for type-level programming with booleans

xdesou01:


examples/type-level-boolean-programming
```diff
@@ -0,0 +1,38 @@
+/// A list that tracks whether its length is even or odd.
```
magnus-madsen, line 1: We need a meaningful filename.

examples/type-level-boolean-programming
```diff
@@ -0,0 +1,38 @@
...
+enum EOList[a: Type, even: Bool] {
```
magnus-madsen, line 2: I think some more explanation is warranted.

examples/type-level-boolean-programming
```diff
@@ -0,0 +1,38 @@
...
+    case EONil
+    case EOCons(a, EOList[a, not even])
+}
+
+/// Returns an empty and even list.
```
magnus-madsen, line 7: empty what? empty list? which has an even light.

examples/type-level-boolean-programming
```diff
@@ -0,0 +1,38 @@
...
+def mkNilEO(): EOList[a, true] = EOList.EONil
```
magnus-madsen, line 8: Can we drop all the EO suffixes? I think we should only use them for the EOType.

examples/type-level-boolean-programming
```diff
@@ -0,0 +1,38 @@
...
+
+/// Attaches an element to the front of the list.
+def consEO(x: a, list: EOList[a, even]): EOList[a, not even] =
+    EOList.EOCons(x, list)
+
+/// Returns the head of the list.
+def headEO(list: EOList[a, even]): a = match list {
```
magnus-madsen, line 15: list -> l (everywhere)

magnus-madsen: I don't understand why you have "even" as the type argument, should it not be "false"?

examples/type-level-boolean-programming
```diff
@@ -0,0 +1,38 @@
...
+    case EOList.EOCons(x, _) => x
+    case EOList.EONil        => bug!("Empty input not allowed")
+}
+
+/// Returns the tail of the list.
+def tailEO(list: EOList[a, even]): EOList[a, not even] = match list {
+    case EOList.EOCons(_, smallerList) => smallerList
```
magnus-madsen, line 22: smallerList -> xs

examples/type-level-boolean-programming
```diff
@@ -0,0 +1,38 @@
...
+    case EOList.EONil                  => bug!("Empty input not allowed")
+}
+
+/// Applies a function `f` to each element of a list.
+def mapEO(f: a -> b, list: EOList[a, even]): EOList[b, even] = match list {
+    case EOList.EONil            => EOList.EONil
+    case EOList.EOCons(x, sList) => EOList.EOCons(f(x), mapEO(f, sList))
+}
+
+/// Prints both elements of a constructed list.
```
magnus-madsen, line 32: Drop main

examples/type-level-boolean-programming
```diff
@@ -0,0 +1,38 @@
...
+    case EOList.EOCons(_, smallerList) => smallerList
+    case EOList.EONil                  => bug!("Empty input not allowed")
+}
+
+/// Applies a function `f` to each element of a list.
+def mapEO(f: a -> b, list: EOList[a, even]): EOList[b, even] = match list {
+    case EOList.EONil            => EOList.EONil
+    case EOList.EOCons(x, sList) => EOList.EOCons(f(x), mapEO(f, sList))
+}
+
```
magnus-madsen, line 31: Add append.

examples/type-level-boolean-programming
xdesou01: I did that because the functionality shouldn't depend on whether the list is even. The intention is that `even` is just a variable used to determine evenness, whether it's `true` or `false`. Should I only allow for returning the heads of odd lists?

examples/type-level-boolean-programming
xdesou01: I am willing to, but I just wanted to check - don't most of the examples have main functions? And in that case, should I preface each function name with _ to avoid unusage errors?

examples/type-level-boolean-programming
xdesou01: added, but I get a type error with the current implementation (which I think is pretty reasonable, unless I'm missing something)

examples/even-odd-list-with-type-level-booleans
```diff
@@ -0,0 +1,45 @@
+/// A list of type `a` that uses a Boolean at the type level to track
```
magnus-madsen, line 1: Filename: `type-level-programming-odd-even-list.flix`

examples/even-odd-list-with-type-level-booleans
```diff
@@ -0,0 +1,45 @@
...
+enum EOList[a: Type, even: Bool] {
+    case Nil
+    case Cons(a, EOList[a, not even])
+}
+
+/// Returns an empty list of type `a` with even length (which is 0).
+def mkNil(): EOList[a, true] = EOList.Nil
+
+/// Attaches an element to the front of the list.
+def cons(x: a, l: EOList[a, e]): EOList[a, not e] =
```
magnus-madsen, line 12: You use `e` and `even` interchangeably. Let us just use `e` everywhere.

examples/even-odd-list-with-type-level-booleans
```diff
@@ -0,0 +1,45 @@
...
+    EOList.Cons(x, l)
+
+/// Returns the head of the list.
```
magnus-madsen, line 15: Returns the head element of an odd list (since such a list is guaranteed to be empty).

examples/even-odd-list-with-type-level-booleans
```diff
@@ -0,0 +1,45 @@
...
+def head(l: EOList[a, e]): a = match l {
+    case EOList.Cons(x, _) => x
+    case EOList.Nil        => bug!("Empty input not allowed")
```
magnus-madsen, line 18: In general, you pattern matches should have the same order as the enum, i.e. you should always have `Nil` first. (This is also the easiest case).

examples/even-odd-list-with-type-level-booleans
```diff
@@ -0,0 +1,45 @@
...
+}
+
+/// Returns the tail of the list.
+def tail(l: EOList[a, even]): EOList[a, not even] = match l {
+    case EOList.Cons(_, xs) => xs
+    case EOList.Nil         => bug!("Empty input not allowed")
```
magnus-madsen, line 24: Return `nil` instead.

examples/even-odd-list-with-type-level-booleans
```diff
@@ -0,0 +1,45 @@
...
+def cons(x: a, l: EOList[a, e]): EOList[a, not e] =
+    EOList.Cons(x, l)
+
+/// Returns the head of the list.
+def head(l: EOList[a, e]): a = match l {
+    case EOList.Cons(x, _) => x
+    case EOList.Nil        => bug!("Empty input not allowed")
+}
+
+/// Returns the tail of the list.
```
magnus-madsen, line 21: /// Returns the tail of the list. Note how this changes the evenness of the list.

examples/even-odd-list-with-type-level-booleans
```diff
@@ -0,0 +1,45 @@
...
+def tail(l: EOList[a, even]): EOList[a, not even] = match l {
+    case EOList.Cons(_, xs) => xs
+    case EOList.Nil         => bug!("Empty input not allowed")
+}
+
+/// Applies a function `f` to each element of a list.
```
magnus-madsen, line 27: /// Note how the evenness of the list is preserved.

examples/even-odd-list-with-type-level-booleans
```diff
@@ -0,0 +1,45 @@
...
+def map(f: a -> b, l: EOList[a, even]): EOList[b, even] = match l {
+    case EOList.Nil         => EOList.Nil
+    case EOList.Cons(x, xs) => EOList.Cons(f(x), map(f, xs))
+}
+
+/// Returns a concatenated list containing the elements of two lists.
+// def _append(l1: EOList[a, e1], l2: EOList[a, e2]): EOList[a, e1 and e2] = match l1 {
+//     case EOList.Nil         => l2
+//     case EOList.Cons(x, xs) => cons(x, _append(xs, l2))
+// }
```
magnus-madsen, line 37: Replace by:

```
/// Returns the concatenation of `l1` and `l2`.
/// Note that the result is even if both lists are even 
/// or both are odd (i.e. (e1 and e2) or (not e1 and not e2)
/// which is equivalent to not (e1 xor e2). 
def append(l1: EOList[a, e1], l2: EOList[a, e2]): EOList[a, not (e1 xor e2)] = match l1 {
     case EOList.Nil         => 
        // We have to insert unchecked casts because the Flix type system does not
        // refine the type of l1 when we pattern match on it, i.e. there no GADTs.
        unchecked_cast(l2 as EOList[a, not (e1 xor e2)])
     case EOList.Cons(x, xs) => 
        unchecked_cast(EOList.Cons(x, append(xs, l2)) as EOList[a, not (e1 xor e2)])
}
```

examples/even-odd-list-with-type-level-booleans
```diff
@@ -0,0 +1,45 @@
...
+
+/// Prints both elements of a constructed list.
```
magnus-madsen, line 39: Remove main, the example is already long enough, and it does not add anything.

examples/even-odd-list-with-type-level-booleans
xdesou01: I do not know what this means - an odd list is guaranteed to be empty? Wouldn't it be guaranteed not to be empty?

magnus-madsen: I meant non-empty.

xdesou01: So we don't support calling `head` for any even lists?

magnus-madsen: That's right

magnus-madsen: (one would have a different function returning an option, but we will not bother with that.)

examples/even-odd-list-with-type-level-booleans
xdesou01: smh I initially changed it to `type-level-programming-odd-even-list.filx`
Should be in accordance with all of your comments now.

magnus-madsen (approved):

PR #6179, opened 2023-06-29
test: add assoc type tests

magnus-madsen:


mlutze (commented):
We should try calling these to be sure we make it through Monomorph correctly.

magnus-madsen:
> We should try calling these to be sure we make it through Monomorph correctly.

We could call these, but in the other file there is something typing issue.

mlutze (approved):

PR #6181, opened 2023-06-30
feat: add an example of using Booleans for eager/lazy lists

magnus-madsen:


mlutze:
Why `HList`?

magnus-madsen:
> Why `HList`?

Temporary name. Suggestions most welcome :)

examples/type-level-programming-eager-lazy-list.flix
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+/// A hybrid zip function which is eager if both arguments are eager.
+/// If either argument (or both) is lazy then the list is lazy.
+def zip(l1: LEList[a, e1], l2: LEList[b, e2]): LEList[(a, b), e1 and e2] =
+    typematch (BoolProxy.BoolProxy : BoolProxy[e1 and e2]) {
+        case _: (BoolProxy[true]) =>
+            unchecked_cast(zipE(l1, l2) as LEList[(a, b), e1 and e2])
+        case _: _                 =>
+            unchecked_cast(zipL(l1, l2) as LEList[(a, b), e1 and e2])
```
mlutze, line 42: Really unfortunate that we have these casts :(

mlutze (approved):

examples/type-level-programming-eager-lazy-list.flix
magnus-madsen: Just add GADTs? :P

magnus-madsen: Actually that may not be enough. The casts are part of the specification no?

mlutze: If we get a `return SomeType` clause like in Coq GADTs then it should be enough.

PR #6187, opened 2023-06-30
feat: add example of empty-non empty lists

xdesou01:
WIP, type errors transpire whenever I try to implement anything (unless I make everything true or false, which isn't intuitive)


examples/type-level-boolean-programming
```diff
@@ -0,0 +1,38 @@
+/// A list that tracks whether its length is even or odd.
```
magnus-madsen, line 1: This should not be here.

magnus-madsen:
Try to avoid this:

image

You probably want to update/rebase your branch. (You can read online).

examples/type-level-boolean-programming
xdesou01: Good heavens I did it again

magnus-madsen:
What was the challenge here?

xdesou01:
> What was the challenge here?

In hindsight, I think I was just malding because it wouldn't be possible (given the current implementation) to construct a list with length > 1, because of how cons is implemented. I guess we can just overlook it as a side effect of the invariant?

xdesou01:
This is almost working, but I'm getting an error (removed extra functions for brevity):

```enum MyList[a: Type, _: Bool] {
    case Nil
    case Cons(a, MyList[a, true])
}
def _cons(x: a, l: MyList[a, true]): MyList[a, false] = MyList.Cons(x, l)

def _append(l1: MyList[a, true], l2: MyList[a, _]): MyList[a, false] = match l1 {
    case MyList.Nil         => l2
    case MyList.Cons(x, xs) => _cons(x, _append(xs, l2))
}```

![image](https://github.com/flix/flix/assets/103150835/80281f3c-e4aa-47d0-ae41-00b75f734cdb)

**Comment by magnus-madsen** [comment]
I would think that the type of `append` should be:

```
def _append(l1: MyList[a, e1], l2: MyList[a, e2]): MyList[a, e1 and e2]
```

**Comment by xdesou01** [comment]
> I would think that the type of `append` should be:
> 
> ```
> def _append(l1: MyList[a, e1], l2: MyList[a, e2]): MyList[a, e1 and e2]
> ```

I tried that first, but it's dubious because `_cons` has a specific boolean type, so things dont match up:

Type One: a -> (MyList[a, true] -> MyList[a, false])
Type Two: a -> (MyList[a, true and b2] -> MyList[a, false])

**Comment by magnus-madsen** [comment]
Cons should not require its argument to be empty. 

But also, we may need a unchecked_cast in append.

**Comment by xdesou01** [comment]
> Cons should not require its argument to be empty.
> 
> But also, we may need a unchecked_cast in append.

A type error is thrown when the cons argument isn't empty, which I believe is due to how the cons case for MyList is defined.

**Comment by magnus-madsen** [comment]
Here is my proposal:

```
enum ENList[a: Type, _e: Bool] {
    case Nil,
    case Cons(a, ENList[a, true])
}

type alias Empty    = true
type alias NonEmpty = false 

def nil(): ENList[a, Empty] = ENList.Nil

def cons(x: a, xs: ENList[a, e]): ENList[a, NonEmpty] = 
    ENList.Cons(x, unchecked_cast(xs as ENList[a, true]))

def head(l: ENList[a, NonEmpty]): a = match l {
    case ENList.Nil        => bug!("Impossible")
    case ENList.Cons(x, _) => x
}

def append(l1: ENList[a, e1], l2: ENList[a, e2]): ENList[a, e1 and e2] = 
    match l1 {
        case ENList.Nil         => 
            unchecked_cast(l2 as ENList[a, e1 and e2])
        case ENList.Cons(x, xs) => 
            ENList.Cons(x, unchecked_cast(append(xs, l2) as ENList[a, true]))
    }
```

but it needs some comments :)

magnus-madsen (approved):

Issue #6047, opened 2023-06-05
Eliminate `Ast.Constant.Regex` in `Simplifier`


magnus-madsen:
I don't mean to delete `Ast.Constant.Regex`, but rather in `Simplifier` it should be translated to `Expression.NewObject`.

jaschdoc:
What object should it be translated to? `Pattern`? Or should we compile it?

https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/util/regex/package-summary.html

https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/util/regex/Pattern.html

magnus-madsen:
What does the backend do? I think we should do what it does.

magnus-madsen:
Looks like a call to `Pattern.compile`. Nothing fancy.

jaschdoc:
`GenExpression` just calls `Pattern.compile` on the pattern, so I think we should translate it to an `InvokeStaticMethod` expression instead.

magnus-madsen:
Yes.

jaschdoc:
Perfect

PR #6192, opened 2023-07-01
refactor: eliminate `Ast.Constant.Regex` in `Simplifier`

jaschdoc:
Closes https://github.com/flix/flix/issues/6047

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
@@ -116,7 +116,8 @@ object GenExpression {
         addSourceLine(mv, loc)
         mv.visitLdcInsn(s)
 
-      case _ => throw InternalCompilerException(s"Unexpected Constant $cst", loc)
+      case Ast.Constant.Regex(lit) =>
+        throw InternalCompilerException(s"Unexpected Regex Constant $cst - this should have been eliminated in Simplifier", loc)
 
     }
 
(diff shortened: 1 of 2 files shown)


main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -116,11 +116,7 @@ object GenExpression {
         addSourceLine(mv, loc)
         mv.visitLdcInsn(s)
 
-      case Ast.Constant.Regex(patt) =>
-        addSourceLine(mv, loc)
-        mv.visitLdcInsn(patt.pattern)
-        mv.visitMethodInsn(INVOKESTATIC, JvmName.Regex.toInternalName, "compile",
-          AsmOps.getMethodDescriptor(List(JvmType.String), JvmType.Regex), false)
+      case _ => throw InternalCompilerException(s"Unexpected Constant $cst", loc)
```
magnus-madsen, line 119: Can we not remove it from `Ast.Constant`? or is that not possible?

At minimum, I think we should replace `_` by matching on the Regex.

jaschdoc: > Can we not remove it from `Ast.Constant`? or is that not possible?

Remove what exactly? :)

magnus-madsen: But can we drop `Ast.Constant.Regex` or is that used earlier in the compiler?

magnus-madsen: I see its used earlier.

magnus-madsen:
In that case I am not sure if this PR is worth it. Let me think.

magnus-madsen:
I decided its not worth it. Sorry for the trouble.

jaschdoc:
No worries :)

PR #6195, opened 2023-07-02
feat: add an example of using Booleans for string sanitization

magnus-madsen:


magnus-madsen:
Inspired by an example from @xdesou01 and cleaned up.

mlutze (approved):

PR #6197, opened 2023-07-03
feat: add an example of type-level 4-bit integers

magnus-madsen:


examples/type-level-programming-4bit-adder.flix
```diff
@@ -0,0 +1,35 @@
+///
+/// We can use type-level Booleans to represent 4-bit integers.
+///
+/// We represent a 4-bit integer as 4 bits: a, b, c, d.
+///
+enum I4[_a: Bool, _b: Bool, _c: Bool, _d: Bool] {
```
mlutze, line 6: Could we do `I4[b8, b4, b2, b1]` instead to show the value of each bit?

mlutze (approved):

examples/type-level-programming-4bit-adder.flix
magnus-madsen: That seems pretty non-standard??

mlutze: With the decimal number 4256 we would say "6 is in the ones place, 5 is in the tens place, 2 is in the hundreds place, 4 is in the thousands place".

Anyway not a blocker.

Issue #5824, opened 2023-04-30
Use `Ast.CallType` in `LiftedAst`


magnus-madsen:
- [ ] Merge `ApplyClo` and `ApplyCloTail` using `Ast.CallType` to distinguish them (i.e. `ApplyCloTail` is a tail call).
- [ ] Ditto for `ApplyDef` and `ApplyDefTail`.

You can see how it looks in `ErasedAst`.

jaschdoc:
Same of OccurrenceAst?

magnus-madsen:
Yeah why not

PR #6202, opened 2023-07-03
refactor: use `Ast.CallType` in `LiftedAst`

jaschdoc:
Closes https://github.com/flix/flix/issues/5824

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Inliner.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Inliner.scala
@@ -130,9 +130,9 @@ object Inliner {
         case _ => LiftedAst.Expression.ApplyAtomic(op, es, tpe, purity, loc)
       }
 
-    case OccurrenceAst.Expression.ApplyClo(exp, args, tpe, purity, loc) =>
+    case OccurrenceAst.Expression.ApplyClo(exp, exps, tpe, purity, loc) =>
       val e = visitExp(exp, subst0)
-      val as = args.map(visitExp(_, subst0))
+      val es = exps.map(visitExp(_, subst0))
       e match {
         case LiftedAst.Expression.ApplyAtomic(AtomicOp.Closure(sym), closureArgs, _, _, _) =>
           val def1 = root.defs.apply(sym)
@@ -142,13 +142,11 @@ object Inliner {
           if (canInlineDef(def1)) {
             val e1 = rewriteTailCalls(def1.exp)
             // Map for substituting formal parameters of a function with the closureArgs currently in scope
-            bindFormals(e1, (def1.cparams ++ def1.fparams).map(_.sym), closureArgs ++ as, Map.empty)
+            bindFormals(e1, (def1.cparams ++ def1.fparams).map(_.sym), closureArgs ++ es, Map.empty)
           } else {
-            LiftedAst.Expression.ApplyClo(e, as, tpe, purity, loc)
+            LiftedAst.Expression.ApplyClo(e, es, Ast.CallType.NonTailCall, tpe, purity, loc)
           }
-        case _ =>
-          val as = args.map(visitExp(_, subst0))
-          LiftedAst.Expression.ApplyClo(e, as, tpe, purity, loc)
+        case _ => LiftedAst.Expression.ApplyClo(e, es, Ast.CallType.NonTailCall, tpe, purity, loc)
       }
 
     case OccurrenceAst.Expression.ApplyDef(sym, args, tpe, purity, loc) =>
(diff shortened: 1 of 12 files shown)


main/src/ca/uwaterloo/flix/language/phase/Inliner.scala
```diff
@@ -130,70 +131,42 @@ object Inliner {
...
-          } else {
-            LiftedAst.Expression.ApplyClo(e, as, tpe, purity, loc)
-          }
-        case _ =>
-          val as = args.map(visitExp(_, subst0))
-          LiftedAst.Expression.ApplyClo(e, as, tpe, purity, loc)
-      }
-
-    case OccurrenceAst.Expression.ApplyDef(sym, args, tpe, purity, loc) =>
-      val as = args.map(visitExp(_, subst0))
```
magnus-madsen, line 155: Have you preserved the meaning of this code?

jaschdoc: Yes, the only difference is that I've renamed `args` to `exps` and merged `ApplyDef` and `ApplyDefTail` cases. The only difference those cases had were whether to call `rewriteTailCalls(def1.exp)` or to use `def1.exp` for `indFormals`. This difference is handled by

```scala
val e1 = ct match {
          case CallType.TailCall => def1.exp
          case CallType.NonTailCall => rewriteTailCalls(def1.exp)
        }
bindFormals(e1, (def1.cparams ++ def1.fparams).map(_.sym), es, Map.empty)
```

jaschdoc: Same thing for `ApplyClo`. I approached it very incrementally by first adding `ct` to `ApplyX` before merging `ApplyXTail` into anything and just preserving semantics with that addition. Then I rewrote each pattern match to match directly on the value of `ct` i.e., `ApplyDef(sym, exps, Ast.CallType.NonTailCall,...)` to ensure that semantics were stilled preserved. Finally, I looked at how to merge similar cases.

jaschdoc: In short, I'm fairly confident, it has the same meaning as before :)

magnus-madsen: OK. I will take a much closer looker later then.

magnus-madsen:
Oops. Conflicts, sorry :(

main/src/ca/uwaterloo/flix/language/phase/Inliner.scala
```diff
@@ -130,70 +131,42 @@ object Inliner {
...
-          // If `def1` is a single non-self call or
-          // it is trivial
-          // then inline the body of `def1`
-          if (canInlineDef(def1)) {
-            // Map for substituting formal parameters of a function with the freevars currently in scope
-            bindFormals(def1.exp, (def1.cparams ++ def1.fparams).map(_.sym), closureArgs ++ as, Map.empty)
+            bindFormals(e1, (def1.cparams ++ def1.fparams).map(_.sym), closureArgs ++ es, Map.empty)
           } else {
-            LiftedAst.Expression.ApplyCloTail(e, as, tpe, purity, loc)
+            LiftedAst.Expression.ApplyClo(e, es, ct, tpe, purity, loc)
```
magnus-madsen, line 151: I am still confused, are you not turning a tail call into a non-tail call?

magnus-madsen: Am I just stupid? :)

jaschdoc: Both non-tail and tail calls are contained in `ApplyClo` and the only difference in behavior for this case is what `e1` contains in the call to `bindFormals`. However, if we hit this else branch, we don't consider the `bindFormals` branch, so it just retains the call type it already has.

jaschdoc: Is there any way for us to measure if we lose any performance?

jaschdoc: Or to inspect the behavior of the call stack at runtime? JVM debugger / visualizer?

magnus-madsen: Ah, maybe I am stupid. Its because of the `ct` argument of course!

jaschdoc: Haha, exactly (modulo stupidity)!

main/src/ca/uwaterloo/flix/language/phase/Tailrec.scala
```diff
@@ -79,20 +80,20 @@ object Tailrec {
       /*
        * ApplyClo.
        */
-      case Expression.ApplyClo(exp, args, tpe, purity, loc) =>
-        Expression.ApplyCloTail(exp, args, tpe, purity, loc)
+      case Expression.ApplyClo(exp, exps, Ast.CallType.NonTailCall, tpe, purity, loc) =>
```
magnus-madsen, line 83: I think here instead of `Ast.CallType.NonTailCall` we just put wildcard `_`.

main/src/ca/uwaterloo/flix/language/phase/Tailrec.scala
```diff
@@ -79,20 +80,20 @@ object Tailrec {
...
+        Expression.ApplyClo(exp, exps, Ast.CallType.TailCall, tpe, purity, loc)
 
       /*
        * ApplyDef.
        */
-      case Expression.ApplyDef(sym, args, tpe, purity, loc) =>
+      case Expression.ApplyDef(sym, exps, Ast.CallType.NonTailCall, tpe, purity, loc) =>
```
magnus-madsen, line 89: Ditto

magnus-madsen (approved):

Issue #777, opened 2019-02-26
Add Support for Pattern Matching on Records


magnus-madsen:
We should allow records to be destructed with pattern matching, e.g.:

```
match r with {
  case { fstName, lstName } => 
}
```

magnus-madsen:
@nicdard This might actually be a good medium-sized project. It involves minor changes to the frontend of the compiler, a bit more work in typer, and some work in the Simplifier. In any case, it can be done by starting from some test cases and then working one phase at a time.

nicdard:
Very cool, thanks! I would really likes to take it feel free to assign this to me :)
Also it goes together with refactoring the pattern matching code that we discussed right?

magnus-madsen:
Not entirely, but that could be a natural thing to do after.

To get started, I would: (1) Write a few programs that use pattern matching on records (obviously they will not work yet). (2) Extende ParsedAst.Pattern with some record constructs (they should probably look closely like their Expression friends) and (3) Try to extend the Parser.

Once that works, you can work phase-by-phase.

I suggest to open a pull request as soon as you have some test cases. (See Text.Exp.Match.XYZ).

nicdard:
Perfect! Thanks for the guide :)

nicdard:
I am starting to work on this :)

magnus-madsen:
@jaschdoc I think you have the expertise to do this once the backend stuff clears the queue. What do you think?

jaschdoc:
I'd love to give it a try!

jaschdoc:
It seems OCaml allows these patterns for records
```scala
{ x }
{ x : Int32 }
{ x : Int32 = 1 }
{ x = 1 }
{ x, y : Int32 = 2 | _ }
```
https://v2.ocaml.org/releases/5.0/htmlman/patterns.html

magnus-madsen:
Does it not allow `{ l1 = x }`?

jaschdoc:
What are `l1` and `x` here?

magnus-madsen:
> What are `l1` and `x` here?

The l1 would be the field, and `x` would be a freely chosen variable.

For example, if you had two point records, you could do:

```
match (p1, p2) {
case ({x = x1, y = y1}, {x = x2, y = y2}) => 
```

I agree we will want other patterns, I just think this is a reasonable place to start.

jaschdoc:
> The l1 would be the field, and `x` would be a freely chosen variable.

Yes it does. I forgot to write that in the post above.

PR #6204, opened 2023-07-03
feat: add pattern matching on records

jaschdoc:
- [x] Do not allow `| pat` to be parsed without any other fields (handled in Weeder)
- [x] Consider parsing `| pat` as `| Pattern.Var` or ensure in Weeder that it is either wildcard or var (handled in Weeder)
- [x] Add more tests
- [x] NonLinearPattern
- [x] ShadowedVar
- [x] UnusedVar
- [x] TypeError (pattern)
- [x] Remove type ascription
- [x] Figure syntax design
- [x] Fix redundancy error so correct variable is unused
- [x] Try out this https://github.com/flix/flix/pull/6204#issuecomment-1672109389
- [x] Implement feedback / review
- [x] Fix types in Simplifier

Closes https://github.com/flix/flix/issues/777


main/test/flix/Test.Exp.Match.Record.flix
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @test
+    def testMatchEmptyRecord06(): Bool = match { x = 3 } {
+        case { x = 1 } => false
+        case { x = 2 } => false
+        case { x = 3 } => true
+        case { x = 4 } => false
+    }
+    */
```
magnus-madsen, line 57: Do you intend to allow:

```
case { x = foo, y = bar }
```

and what about:

```
case { x, y }
````

and what about

```
case { x = 1 | r }
```

?

Comment (jaschdoc) [reply main/test/flix/Test.Exp.Match.Record.flix:57]: Yes, but I see those as mixtures of the individual patterns outlined in `ParsedAst.RecordPattern`. However, once I dig further in, I may realize it doesn't quite work that way.

Comment (jaschdoc) [reply main/test/flix/Test.Exp.Match.Record.flix:57]: > What does other languages do?

I'll take a look at OCaml :)

**Comment by magnus-madsen** [comment]
It migh also be helpful to start with the `Typer`. This will guide you in what is possible.

**Comment by magnus-madsen** [comment]
So regardless of good intentions, I am going to suggest you start with the absolute simplest. Just allow pattern matching on `{ l1 = x, l2 = y }` where this binds the variables `x` and `y`. Do not try to support anything else in the first iteration :)

**Comment by jaschdoc** [comment]
Hmm I've got something working through `Weeder` but I'm considering starting over and adding simpler parser rules (similar to what OCaml does)

**Comment by magnus-madsen** [comment]
> Hmm I've got something working through `Weeder` but I'm considering starting over and adding simpler parser rules (similar to what OCaml does)

I think in any case a few iterations would be needed. You can start over, but you might also want to forge ahead to post Simplifier.

File: main/src/ca/uwaterloo/flix/language/phase/Namer.scala [hunk 2]
```diff
@@ -1349,6 +1365,10 @@ object Namer {
     case WeededAst.Pattern.Cst(Ast.Constant.Null, loc) => throw InternalCompilerException("unexpected null pattern", loc)
     case WeededAst.Pattern.Tag(qname, p, loc) => freeVars(p)
     case WeededAst.Pattern.Tuple(elms, loc) => elms flatMap freeVars
+    case WeededAst.Pattern.Record(pats, pat, _) =>
+      // pats.map(_.tpe.map(freeTypeVars)).filterNot(_.isEmpty).flatMap(_.get) ++ <---- include type vars?
```
jaschdoc, line 1369: Should type vars be included in `freeVars`?

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -1721,6 +1721,19 @@ object Resolver {
           mapN(esVal) {
             es => ResolvedAst.Pattern.Tuple(es, loc)
           }
+
+        case NamedAst.Pattern.Record(pats, pat, loc) =>
+          val psVal = traverse(pats) {
+            case NamedAst.Pattern.Record.RecordFieldPattern(field, tpe, pat1, loc1) =>
+              // Unsure about Wildness
+              mapN(traverseOpt(tpe)(resolveType(_, Wildness.ForbidWild, env, taenv, ns0, root)), traverseOpt(pat1)(visit)) {
```
jaschdoc, line 1729: What should the value of `Wildness` be in `resolveType` when resolving a type annotation in a pattern match on a field? e.g.
```scala
match { x = 1 } {
    case { x : Int32 = 1 } => ??? // Consider `Int32`
}
```

mlutze: It seems like wildness should be allowed, IMO.

jaschdoc: What does Wildness refer to? Kinds are in the next phase, correct? So what is Wildness here?

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -2540,6 +2540,41 @@ object Typer {
...
+          case KindedAst.Pattern.Record.RecordFieldPattern(field, tpe, pat1, loc1) => (tpe, pat1) match {
+            case (Some(t), Some(p)) =>
+              // { x : Type = pattern ... }
+              val freshRowVar = Type.freshVar(Kind.RecordRow, loc1)
+              val recordRow = Type.mkRecordRowExtend(field, t, freshRowVar, loc1)
+              for {
+                t1 <- visit(p)
+                _ <- unifyTypeM(recordRow, t1, loc1)
+                resultTyp = t
+              } yield resultTyp
```
jaschdoc, line 2555: @mlutze I think I might need some help with this.

I'm trying to come up with the type rules for the pattern `{ field : Type = Pattern }`. Note that there may be several of these, i.e.,
```ocaml
{ field1 : Type = Pattern, field2, field3 : Type }
```
hence the `...` in the comment in the code.

Is it correct to assume that the type of the field is a `recordRowExtend` type?
Furthermore, should it be unified with the constraints from the pattern (`t1 <- visit(p)`)?
Finally, what is the resulting type? I'm assuming that if all goes well, it should be the type of the provided type annotation? Or is it the `recordRowExtend` type?

main/src/ca/uwaterloo/flix/language/phase/Namer.scala
```diff
@@ -1087,20 +1086,36 @@ object Namer {
...
+    case WeededAst.Pattern.Tuple(elms, loc) =>
+      mapN(traverse(elms)(visitPattern)) {
+        case es => NamedAst.Pattern.Tuple(es, loc)
+      }
 
+    case WeededAst.Pattern.Record(pats, pat, loc) =>
+      val psVal = traverse(pats) {
+        case WeededAst.Pattern.Record.RecordFieldPattern(field, tpe, pat1, loc1) =>
+          mapN(traverseOpt(tpe)(visitType), traverseOpt(pat1)(visitPattern)) {
+            case (t, p) => NamedAst.Pattern.Record.RecordFieldPattern(field, t, p, loc1)
```
jaschdoc, line 1112: I think this is wrong. If `p` is `None` we should create a new pattern / symbol in the environment s.t. patterns like `{ field : Type }` or simply `{ field }` allow the following expression to use `field` as a variable, i.e.,
```scala
match { x = 1 } {
    case { x : Int32 } => println(x)
}
```

jaschdoc: Maybe the resolver also needs some rework to make this work

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
mlutze: First I will confuse you by answering your questions.

> Is it correct to assume that the type of the field is a `recordRowExtend` type?

No. Each field has the kind `Type`.

> Furthermore, should it be unified with the constraints from the pattern (`t1 <- visit(p)`)?

Yes. In particular, you should use the `expectTypeM` function because we expect the pattern to match that type.

> Finally, what is the resulting type? I'm assuming that if all goes well, it should be the type of the provided type annotation? Or is it the `recordRowExtend` type?

Yes, in this case you can use the provided type annotation as the result type.

---- 

I think there is some confusion about the structure of record types.

These are the relevant type constructors and their kinds.
- `Record: RecordRow -> Type`
- `RecordRowExtend(label): Type -> RecordRow -> Type`
- `RecordRowEmpty: RecordRow`

A record is a wrapper around a record row. A record row is essentially a list-like structure, where `RecordRowExtend(label)` is `Cons` and `RecordRowEmpty` is `Nil`.

The record type `{x: String, y: Bool}` has the following structure:

```
Record[
  RecordRowExtend(x)[
    String,
    RecordRowExtend(y)[
      Bool,
      RecordRowEmpty
    ]
  ]
]
```

For your purposes here, I propose that while traversing your list of patterns, you do not pay attention to the record structure above. Just infer the type of each field. That means you don't need to deal with anything involving `RecordRow` inside your traversal.

After you have your list of fields with their inferred types, you can use them to build the final record type.

jaschdoc: Ah, okay, that helps a lot, thanks!

jaschdoc: Am I correct in assuming that handling correspondence of the matched type and the pattern is handled by the `match-case` expression? This would mean that I only have to infer the type of the pattern, and the `match-case` type rules will ensure that the record types actually match (with correct fields etc.)?

mlutze: That is correct.

jaschdoc:
Hmm maybe the extension `{ ... | r }` just has to be a var instead of a pattern.

main/src/ca/uwaterloo/flix/language/phase/Namer.scala
jaschdoc: @mlutze what do you think about this?

mlutze: No, only term vars.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -2539,6 +2539,21 @@ object Typer {
           elementTypes <- traverseM(elms)(visit)
         } yield Type.mkTuple(elementTypes, loc)
 
+
+      case KindedAst.Pattern.Record(pats, pat, tvar, loc) =>
+        val ps = traverseM(pats)(visitRecordFieldPattern(_, root))
```
mlutze, line 2544: You can move this line into the `for` and avoid the `ps1 <- ps` later

mlutze (commented):
A few tests requests:
- NonLinearPattern
-  ShadowedVar
- UnusedVar
- TypeError (pattern)

main/src/ca/uwaterloo/flix/api/lsp/Indexer.scala
```diff
@@ -474,13 +474,22 @@ object Indexer {
...
   /**
     * Returns a reverse index for the given patterns `pats0`.
     */
   private def visitPats(pats0: List[Pattern]): Index = traverse(pats0)(visitPat)
 
+  /**
+    * Returns a reverse index for the given [[Pattern.Record.RecordFieldPattern]] `rfp`.
+    */
+  private def visitRecordFieldPattern(rfp: Pattern.Record.RecordFieldPattern): Index = {
+    Index.useOf(rfp.field) ++ visitPat(rfp.pat)
```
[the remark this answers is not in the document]
jaschdoc: Yes

main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
```diff
@@ -599,6 +599,13 @@ object SemanticTokensProvider {
 
     case Pattern.Tuple(pats, _, _) => pats.flatMap(visitPat).iterator
 
+    case Pattern.Record(pats, pat, tpe, loc) =>
+      val patsVal = pats.flatMap {
+        case Pattern.Record.RecordFieldPattern(field, tpe, pat, loc) =>
```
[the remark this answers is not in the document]
jaschdoc: What's more important? Fields or patterns? I'd think patterns are the most important factor in this case

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
```diff
@@ -1329,6 +1329,17 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
       Simple ~ optional(optWS ~ SP ~ operatorX("::") ~ SP ~ optWS ~ Pattern ~> ParsedAst.Pattern.FCons)
     }
 
+    def Record: Rule1[ParsedAst.Pattern] = {
+      def RecordFieldPattern: Rule1[ParsedAst.Pattern.RecordFieldPattern] = rule {
+        // TODO: Use TypeAndEffect instead of Type
```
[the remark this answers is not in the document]
jaschdoc: It might be the case that a field is a function with an effect

main/test/flix/Test.Exp.Match.Record.flix
```diff
@@ -0,0 +1,62 @@
...
+
+    @test
+    def testMatchRecord03(): Bool = match { x = 1 } {
+        case { x } => x == 1
+    }
+
+    @test
+    def testMatchRecord04(): Bool = match { x = 1 } {
+        case { x = 1 } => true
+        case { x = 2 } => false
```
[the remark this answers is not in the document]
jaschdoc: > So exhaustiveness is a later step?

What do you mean? :smile:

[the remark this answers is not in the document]
jaschdoc: > So exhaustiveness is a later step?

What do you mean? :smile:

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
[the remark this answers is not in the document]
mlutze: That should be covered with just `Type`

main/test/flix/Test.Exp.Match.Record.flix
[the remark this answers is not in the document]
mlutze: A separate PR?

main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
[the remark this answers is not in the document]
mlutze: I agree. That way you have `FindUsages` etc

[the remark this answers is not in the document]
jaschdoc: So do we not visit types here or how do we filter on patterns?

jaschdoc:
Assuming everything else is correct (which it is not), do we then accept for the time being that pattern exhaustiveness is slightly off, i.e., it always requires a wildcard match? I think this is something we can improve upon in later iterations.

jaschdoc:
Leaving this here for reference
https://www.cambridge.org/core/services/aop-cambridge-core/content/view/3165B75113781E2431E3856972940347/S0956796807006223a.pdf/warnings-for-pattern-matching.pdf

magnus-madsen:
> Assuming everything else is correct (which it is not), do we then accept for the time being that pattern exhaustiveness is slightly off, i.e., it always requires a wildcard match? I think this is something we can improve upon in later iterations.

No, I think such a requirement is useless.

I want to be able to write e.g:

```
let { fstName, lstName } = rec
```

If you require a wildcard this does not work.

magnus-madsen:
But I understand the issue, you could have literals:

```
let { fstName, lstName = 123 } = rec
```

which could be non-exhaustive. This suggest to me that the MVP should not allow literals.

magnus-madsen:
Later, we could introduce literals, but then we would need a complex check in `PatMatch` (which would need to check that if `x` has a literal then there must be some other pattern where it is a variable.)

jaschdoc:
Roger. Unrelated to pattern exhaustiveness; do we allow `{ | r }` as a pattern? I suppose it would have the same effect as `_`

magnus-madsen:
> Roger. Unrelated to pattern exhaustiveness; do we allow `{ | r }` as a pattern? I suppose it would have the same effect as `_`

I would disallow that specific form.

jaschdoc:
Ok, I've skipped Pattern Exhaustiveness, but I'm not quite sure about the code in `Simplifier`:

```scala
        
        /**
          * Matching a record may succeed or fail.
          *
          * We generate a fresh variable and let-binding for each component of the
          * record (field or extension) and then we recurse on the nested patterns
          * and freshly generated variables.
          */
        case (LoweredAst.Pattern.Record(pats, _, tpe, loc) :: ps, v :: vs) =>
          // We drop the optional record extension pattern `pat` since
          // the user should not have access to it as a variable
          val freshVars = pats.map(_ => Symbol.freshVarSym("innerField" + Flix.Delimiter, BoundBy.Let, loc))
          val fieldPats = pats.map(_.pat)
          val zero = patternMatchList(fieldPats ::: ps, freshVars ::: vs, guard, succ, fail)
          pats.zip(freshVars).foldRight(zero) {
            case ((LoweredAst.Pattern.Record.RecordFieldPattern(field, _, pat, loc1), name), exp) =>
              val varExp = SimplifiedAst.Expr.Var(v, tpe, loc)
              val recordSelectExp = SimplifiedAst.Expr.ApplyAtomic(AtomicOp.RecordSelect(field), List(varExp), pat.tpe, Pure, loc1)
              SimplifiedAst.Expr.Let(name, recordSelectExp, exp, succ.tpe, exp.purity, loc)
          }
```

magnus-madsen:
> class RecordExtend$Int32 cannot be cast to class dev.flix.runtime.Unit (RecordExtend$Int32 and dev.flix.runtime.Unit are in unnamed module of loader ca.uwaterloo.flix.language.phase.jvm.FlixClassLoader @126945f9)

Most likely the types are wrong.

jaschdoc:
It seems to be working. I'll add a bunch more tests, and then we should be good to go for the first iteration.

magnus-madsen:
@jaschdoc I suggest you do a pass through the changes to see if anything needs be tidied up. Then @mlutze and myself can take a look :)

main/test/flix/Test.Exp.Match.Record.flix
```diff
@@ -0,0 +1,73 @@
...
+    def testMatchRecord08(): Bool = match { x = 1 } {
+        case { x | r } => x == 1
+    }
+
+    @test
+    def testMatchRecord09(): Bool =
+        let rec = { c = "hello" };
+        let { c } = rec;
+        c == "hello"
+
```
magnus-madsen, line 72: I think tests should focus on destructing records and then testing that their components are correct.

E.g. more like `testMatchRecord03`

main/src/ca/uwaterloo/flix/language/ast/KindedAst.scala
```diff
@@ -240,6 +240,11 @@ object KindedAst {
 
     case class Tuple(elms: List[Pattern], loc: SourceLocation) extends Pattern
 
+    case class Record(pats: List[Record.RecordFieldPattern], pat: Option[Pattern], tvar: ast.Type.Var, loc: SourceLocation) extends KindedAst.Pattern
```
jaschdoc, line 243: Should `pat` be renamed to `extension` or `tail`? It denotes the `{ ... | r }` part.

main/src/ca/uwaterloo/flix/language/phase/Namer.scala
```diff
@@ -1357,6 +1381,9 @@ object Namer {
     case WeededAst.Pattern.Cst(Ast.Constant.Null, loc) => throw InternalCompilerException("unexpected null pattern", loc)
     case WeededAst.Pattern.Tag(qname, p, loc) => freeVars(p)
     case WeededAst.Pattern.Tuple(elms, loc) => elms flatMap freeVars
+    case WeededAst.Pattern.Record(pats, pat, _) =>
+      pats.map(_.pat.map(freeVars)).filterNot(_.isEmpty).flatMap(_.get) ++
```
jaschdoc, line 1385: This could be clarified, yeah?

jaschdoc:
I think maybe we can leave out the type ascription in the field pattern. If the field does not have the correct type, then it's a type error, otherwise it's a `typematch` expression.

What do we do with the following? Clearly, `t` is not the same type as `Int32`. Do we even allow parameters like that?
```scala
def f(r: { a = t | r }): Bool = match r {
    case { a : Int32 = _ } => true
    case _ => false
}
```

magnus-madsen:
Agree. Let use leave out type ascription. I don't see how it would be useful.

magnus-madsen:
Don't add a lot of test with literals, because I think they have to go.

magnus-madsen:
But tests with nested records and duplicate fields would be useful.

main/src/ca/uwaterloo/flix/language/ast/ParsedAst.scala
```diff
@@ -1264,6 +1265,27 @@ object ParsedAst {
...
 
+    /**
+      * Record Pattern
+      *
+      * @param sp1    the position of the first character in the pattern.
+      * @param fields list of [[RecordFieldPattern]].
+      * @param rest   optional record extension pattern `| r`.
+      * @param sp2    the position of the last character in the pattern.
+      */
+    case class Record(sp1: SourcePosition, fields: Seq[RecordFieldPattern], rest: Option[ParsedAst.Pattern], sp2: SourcePosition) extends ParsedAst.Pattern
```
jaschdoc, line 1276: Same question as above regarding naming

main/src/ca/uwaterloo/flix/language/ast/KindedAst.scala
magnus-madsen: I think it is fine

magnus-madsen:
> @mlutze @magnus-madsen I've looked thoroughly through the changes and added lots of tests. I think it's ready!

Will take a look later.

magnus-madsen:
!!!jar

magnus-madsen:
@jaschdoc I will be testing this extensively and then we can iterate and probably get it right in a single PR.

magnus-madsen:
I would have expected this to work:

```
def foo(): Bool = match { x = 1, y = 2, z = 3 } {
    case {x, y} => x + y == 0
}
```

which gives:

```
❌ -- Type Error -------------------------------------------------- file:///c%3A/tmp/flix/Records.flix

>> Missing field 'z' of type 'Int32'.

1 |> def foo(): Bool = match { x = 1, y = 2, z = 3 } {
2 |>     case {x, y} => x + y == 0
3 |> }

missing field.

The record type:

  (  )

does not contain the field 'z' of type Int32.
```

I think there must be a closed row somewhere that should be open (i.e. a free type variable).

jaschdoc:
> I would have expected this to work:
> 
> ```
> def foo(): Bool = match { x = 1, y = 2, z = 3 } {
>     case {x, y} => x + y == 0
> }
> ```
> 
> which gives:
> 
> ```
> ❌ -- Type Error -------------------------------------------------- file:///c%3A/tmp/flix/Records.flix
> 
> >> Missing field 'z' of type 'Int32'.
> 
> 1 |> def foo(): Bool = match { x = 1, y = 2, z = 3 } {
> 2 |>     case {x, y} => x + y == 0
> 3 |> }
> 
> missing field.
> 
> The record type:
> 
>   (  )
> 
> does not contain the field 'z' of type Int32.
> ```
> 
> I think there must be a closed row somewhere that should be open (i.e. a free type variable).

The way I designed is similar to how we write the types for records. Thus, if you match on a record that must have **exactly** the field `x, y` then you would write the pattern `{ x, y }`, but if you require that it just has **at least** `x, y` then you would write `{ x, y | r }`. Does that make sense?

magnus-madsen:
> The way I designed is similar to how we write the types for records. Thus, if you match on a record that must have exactly the field x, y then you would write the pattern { x, y }, but if you require that it just has at least x, y then you would write { x, y | r }. Does that make sense?

From a compiler writers perspective, yes :-)

But I think from a user's perspective, we would want that `{ x, y }` is internally treated as `{x, y | _ }`.

I think there is only upside?

jaschdoc:
I understand your point but my thought process was that consistency with the type level was important. I think that's more in line with the principle of least surprise and the principle of consistent syntax. Also I think OCaml does the same thing except they use ``{ x, y, _ }` but I'll have to double check on that. 😄

mlutze:
Note also that OCaml has subtyping on records, so specifying only a subset makes more sense for them.

magnus-madsen:
(Ok, Ill return to your point in a bit)

What's the intended behavior here:

```
def main(): Unit \ IO = 
    match {x = 1, y = 2} {
        case {x | r} => println(r.y)
    }
```

magnus-madsen:
Another issue (miscompilation) with duplicate labels:

```
def main(): Unit \ IO = 
    match {x = Some(123), x = 1} {
        case {x = Some(a), x = b} => println("${a} ${b}")
        case {x = a, x = b}       => println("${a} ${b}")
    }
```

jaschdoc:
Ah okay I assumed duplicate fields were prohibited so that's probably really buggy

magnus-madsen:
> > (Ok, Ill return to your point in a bit)
> > What's the intended behavior here:
> > ```
> > 
> > def main(): Unit \ IO = 
> > 
> >     match {x = 1, y = 2} {
> > 
> >         case {x | r} => println(r.y)
> > 
> >     }
> > ```
> 
> That you only have access to `x`. The extension just ensures that it is polymorphic

OK, but `r` should really just be the rest of the record :)

magnus-madsen:
> What are the rules for duplicate fields? How do I select the correct duplicate?

You use `RecordRestrict` to drop the prefix of labels you don't care about. Example: `let r = { x=1, x=2}; let r2 = {-x | r}, r2.x`.

magnus-madsen:
https://doc.flix.dev/records.html#record-restriction

jaschdoc:
> > > (Ok, Ill return to your point in a bit)
> 
> > > What's the intended behavior here:
> 
> > > ```
> 
> > > 
> 
> > > def main(): Unit \ IO = 
> 
> > > 
> 
> > >     match {x = 1, y = 2} {
> 
> > > 
> 
> > >         case {x | r} => println(r.y)
> 
> > > 
> 
> > >     }
> 
> > > ```
> 
> > 
> 
> > That you only have access to `x`. The extension just ensures that it is polymorphic
> 
> 
> 
> OK, but `r` should really just be the rest of the record :)

Isn't that unsafe?

Suppose you have a program:

```scala
def f(r: {x = Int32 | r}): Int32 = match r {
    case { x | r1 } => x + r1.y
}

f({ x = 1, z = () })
```

jaschdoc:
Ok, you can now access the record extension variable `{ ...| r }` as the whole record.

jaschdoc:
Now I just have to figure out duplicate fields (!)

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -457,6 +458,36 @@ object Simplifier {
...
+          val freshVars = pats.map(_ => Symbol.freshVarSym("innerField" + Flix.Delimiter, BoundBy.Let, loc))
+          val fieldPats = pats.map(_.pat)
+          val varExp = SimplifiedAst.Expr.Var(v, tpe, loc)
+          val zero = patternMatchList(fieldPats ::: ps, freshVars ::: vs, guard, succ, fail)
+          // Note that we reverse pats and freshVars because we still want to fold right
+          // but we want to restrict the record / matchVar from left to right
+          val (one, restrictedMatchVar) = pats.reverse.zip(freshVars.reverse).foldRight((zero, varExp): (SimplifiedAst.Expr, SimplifiedAst.Expr)) {
+            case ((LoweredAst.Pattern.Record.RecordFieldPattern(field, _, pat, loc1), name), (exp, matchVarExp)) =>
+              val recordSelectExp = SimplifiedAst.Expr.ApplyAtomic(AtomicOp.RecordSelect(field), List(matchVarExp), pat.tpe, Pure, loc1)
+              val restrictedMatchVarExp = SimplifiedAst.Expr.ApplyAtomic(AtomicOp.RecordRestrict(field), List(matchVarExp), tpe, matchVarExp.purity, loc1)
```
jaschdoc, line 478: I suspect the type `tpe` (and also the purity) is wrong in `restrictedMatchVarExp`. It seems rather cumbersome to update the type with a record restriction based on how type rules works for `RecordRestrict`. However, this works for now. Thoughts?

main/test/flix/Test.Exp.Match.Record.flix
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    @test
+    def testMatchRecord30(): Bool = match { a = false, a = true } {
+        case { a = x, a = y } => (not x) and y
+    }
+
+    @test
+    def testMatchRecord31(): Bool = match { a = false, a = true, a = 42} {
+        case { a = x, a = y | r } => ((not x) and y) and r.a == 42
```
jaschdoc, line 185: Is it correct that `r` should have access to the rightmost field `a` (i.e., `r` refers to the part that was **not** unpacked in the pattern)?

jaschdoc:
We now have support for duplicate fields (modulo the parts of the code where I've added comments / questions)

main/test/flix/Test.Exp.Match.Record.flix
magnus-madsen: Yes, this looks right. You peel of two a's to reveal the 3rd and last a.

jaschdoc:
Now we just have to figure out / agree on the design decision with the record extension.

magnus-madsen:
!!!jar

magnus-madsen:
Both @mlutze and I will review. (Matt is currently away, so may be a bit, but I will continue my explorations, but so far seems to work well 👍 )

main/test/flix/Test.Exp.Match.Record.flix
```diff
@@ -0,0 +1,193 @@
...
+    @test
+    def testMatchRecord29(): Bool = match { a = true, b = true } {
+        case { a | r } => a and r.b
+    }
+
+    @test
+    def testMatchRecord30(): Bool = match { a = false, a = true } {
+        case { a = x, a = y } => (not x) and y
+    }
+
```
magnus-madsen, line 182: It is good to see so many test cases, but I feel like some more systematic is needed (i.e. by naming each test case set better).

For example along the lines:

- matching on distinct fields
- matching with "rest row"
- matching in random order
- matching with duplicate labels
- matching with nested records

magnus-madsen:
Guess what this program does (or generates of error):

```
def main(): Unit \ IO = 
    match {x = 1, x = 2} {
        case {x, x} => println("${x} ${x}")
    }
```

Maybe this is OK. It might be tricky to do better.

magnus-madsen:
@mlutze @JonathanStarup We need a vote on the following. Should this be allowed or not?:

```
def main(): Unit \ IO = 
    match {x = 1, y = 2, z = 3} {
        case {x, y} => println(x + y)
    }
```

jaschdoc:
> Guess what this program does (or generates of error):
> 
> 
> 
> ```
> 
> def main(): Unit \ IO = 
> 
>     match {x = 1, x = 2} {
> 
>         case {x, x} => println("${x} ${x}")
> 
>     }
> 
> ```
> 
> 
> 
> Maybe this is OK. It might be tricky to do better.

Redundancy error?

jaschdoc:
> @mlutze @JonathanStarup We need a vote on the following. Should this be allowed or not?:
> 
> 
> 
> ```
> 
> def main(): Unit \ IO = 
> 
>     match {x = 1, y = 2, z = 3} {
> 
>         case {x, y} => println(x + y)
> 
>     }
> 
> ```

I'm starting to come around to this idea. It's just easier if that's what programmers are used to from other languages, e.g., JavaScript.

mlutze:
> @mlutze @JonathanStarup We need a vote on the following. Should this be allowed or not?:
> 
> ```
> def main(): Unit \ IO = 
>     match {x = 1, y = 2, z = 3} {
>         case {x, y} => println(x + y)
>     }
> ```

I still vote no.

magnus-madsen:
> > @mlutze @JonathanStarup We need a vote on the following. Should this be allowed or not?:
> > ```
> > def main(): Unit \ IO = 
> >     match {x = 1, y = 2, z = 3} {
> >         case {x, y} => println(x + y)
> >     }
> > ```
> 
> I still vote no.

What's the reasoning? I would think that: (a) we want things to be intuitive for the programmer and (b) we should be extensible by default. Forcing closed records seems to go against that. If it is inconsistent with type signatures maybe those can be fixed?
Didn't we do something like that for Datalog?

magnus-madsen:
I see some test failures.

mlutze:
> > > @mlutze @JonathanStarup We need a vote on the following. Should this be allowed or not?:
> > > ```
> > > def main(): Unit \ IO = 
> > >     match {x = 1, y = 2, z = 3} {
> > >         case {x, y} => println(x + y)
> > >     }
> > > ```
> > 
> > I still vote no.
> 
> What's the reasoning? I would think that: (a) we want things to be intuitive for the programmer and (b) we should be extensible by default. Forcing closed records seems to go against that. If it is inconsistent with type signatures maybe those can be fixed?
> Didn't we do something like that for Datalog?

If we make records extensible by default, then I vote yes.

Otherwise the syntax disparities are too strange IMO.

jaschdoc:
> > > > @mlutze @JonathanStarup We need a vote on the following. Should this be allowed or not?:
> > > > ```
> > > > def main(): Unit \ IO = 
> > > >     match {x = 1, y = 2, z = 3} {
> > > >         case {x, y} => println(x + y)
> > > >     }
> > > > ```
> > > 
> > > I still vote no.
> > 
> > What's the reasoning? I would think that: (a) we want things to be intuitive for the programmer and (b) we should be extensible by default. Forcing closed records seems to go against that. If it is inconsistent with type signatures maybe those can be fixed?
> > Didn't we do something like that for Datalog?
> 
> If we make records extensible by default, then I vote yes.
> 
> Otherwise the syntax disparities are too strange IMO.

I think the consistency is also worth more in the long run. It may be a bit odd in the beginning but once the programmer realizes the consistency with records it *should* be intuitive to use.

magnus-madsen:
The safe thing to do is to follow Matt and be conservative. We can always lift the restriction later.

So I suggest we "revert" to requiring the record to be closed and then thats what we merge in now.

main/src/ca/uwaterloo/flix/language/ast/ParsedAst.scala
```diff
@@ -1264,6 +1265,27 @@ object ParsedAst {
...
+
+    /**
+      * Represents a pattern for a field of a record.
+      * `field {: type} {= pattern}`
```
magnus-madsen, line 1280: no more type annotation.

main/test/flix/Test.Exp.Let.Match.flix
```diff
@@ -77,4 +77,39 @@ mod Test.Exp.Let.Match {
...
+    @test
+    def testLetMatch16(): Bool =
+        let { x | _ } = { x = 1, y = "" };
+        x == 1
+
+    @test
+    def testLetMatch17(): Bool =
+        let { x, y } = { x = 1, y = "" };
+        x == 1 and y == ""
+
```
magnus-madsen, line 99: Perhaps a few tests with reordering (see my comment about breaking tests into some categories).

main/src/ca/uwaterloo/flix/language/phase/Namer.scala
```diff
@@ -1107,8 +1108,23 @@ object Namer {
     case WeededAst.Pattern.Tag(qname, pat, loc) =>
       NamedAst.Pattern.Tag(qname, visitPattern(pat), loc)
 
-    case WeededAst.Pattern.Tuple(elms, loc) => NamedAst.Pattern.Tuple(elms map visitPattern, loc)
+    case WeededAst.Pattern.Tuple(elms, loc) =>
+      NamedAst.Pattern.Tuple(elms map visitPattern, loc)
```
magnus-madsen, line 1112: We stopped using the infix syntax, should be `elms.map(visitPattern)`.

main/src/ca/uwaterloo/flix/language/phase/Namer.scala
```diff
@@ -1107,8 +1108,23 @@ object Namer {
...
+    case WeededAst.Pattern.Record(pats, pat, loc) =>
+      val psVal = pats map {
+        case WeededAst.Pattern.Record.RecordFieldPattern(field, None, loc1) =>
+          // Introduce new symbols if there is no pattern
+          val sym = Symbol.freshVarSym(field.name, BoundBy.Pattern, field.loc)
+          val p = NamedAst.Pattern.Var(sym, field.loc)
+          NamedAst.Pattern.Record.RecordFieldPattern(field, p, loc1)
+
+        case WeededAst.Pattern.Record.RecordFieldPattern(field, Some(pat1), loc1) =>
+          NamedAst.Pattern.Record.RecordFieldPattern(field, visitPattern(pat1), loc1)
```
magnus-madsen, line 1123: Rewrite these two cases to avoid a deep match. You can just match ones and then again on `rest`

main/src/ca/uwaterloo/flix/language/phase/Namer.scala
magnus-madsen: This applies in a few other places too.

main/src/ca/uwaterloo/flix/language/phase/Namer.scala
```diff
@@ -1357,6 +1373,14 @@ object Namer {
...
     case WeededAst.Pattern.Tag(qname, p, loc) => freeVars(p)
     case WeededAst.Pattern.Tuple(elms, loc) => elms flatMap freeVars
+    case WeededAst.Pattern.Record(pats, pat, _) => recordPatternFreeVars(pats, pat)
+  }
+
+  /**
+    * Returns the free variables in the list of [[Record.RecordFieldPattern]] `pats` and `pat`.
+    */
+  private def recordPatternFreeVars(pats: List[Record.RecordFieldPattern], pat: Option[WeededAst.Pattern]): List[Name.Ident] = {
+    pats.map(_.pat.map(freeVars)).filter(_.isDefined).flatMap(_.get) ++ pat.map(freeVars).getOrElse(Nil)
```
magnus-madsen, line 1383: Oof. Can this not be made better? _.isDefined? :O

main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
```diff
@@ -1132,4 +1132,46 @@ object WeederError {
 
     override def explain(formatter: Formatter): Option[String] = None
   }
+
+  /**
+    * An error raised to indicate that a record pattern has shape `{ | r }`.
```
magnus-madsen, line 1137: -> has the illegal shape

main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
```diff
@@ -1132,4 +1132,46 @@ object WeederError {
...
+    *
+    * @param loc the location where the error occurred.
+    */
+  case class EmptyRecordExtensionPattern(loc: SourceLocation) extends WeederError {
+    override def summary: String = "A record pattern must define at least one field if it is open for extension."
```
magnus-madsen, line 1142: -> A record pattern must specify at least one field.

main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
```diff
@@ -1132,4 +1132,46 @@ object WeederError {
...
+
+    override def message(formatter: Formatter): String = {
+      import formatter._
+      s"""${line(kind, source.name)}
+         |>> Invalid record pattern.
```
magnus-madsen, line 1147: -> Unexpected record pattern.

main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
```diff
@@ -1132,4 +1132,46 @@ object WeederError {
...
+         |
+         |${code(loc, s"A record pattern must define at least one field if it is open for extension.")}
```
magnus-madsen, line 1149: -> A record pattern must specify at least one field.

main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
```diff
@@ -1132,4 +1132,46 @@ object WeederError {
...
+    override def explain(formatter: Formatter): Option[String] = None
+  }
+
+  /**
+    * An error raised to indicate that the extension of a record pattern is malformed.
+    *
+    * @param loc the location where the error occurred.
+    */
+  case class IllegalRecordExtensionPattern(loc: SourceLocation) extends WeederError {
+    override def summary: String = "An extension can only be a variable or a wildcard."
```
magnus-madsen, line 1163: -> A record extension must be either a variable or wildcard.

main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
```diff
@@ -1132,4 +1132,46 @@ object WeederError {
...
+
+    override def message(formatter: Formatter): String = {
+      import formatter._
+      s"""${line(kind, source.name)}
+         |>> Invalid record extension pattern.
```
magnus-madsen, line 1168: -> Unexpected record extension pattern.

main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
```diff
@@ -1132,4 +1132,46 @@ object WeederError {
...
+         |
+         |${code(loc, s"An extension can only be a variable or a wildcard.")}
```
magnus-madsen, line 1170: A record extension must be either a variable or wildcard.

magnus-madsen (commented):
I have reviewed all the easy code. Now I only have the hard code left :)

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -2548,6 +2548,23 @@ object Typer {
...
+          recordTail = optRecordTail.getOrElse(emptyRecord)
+          _recordExtension <- unifyTypeM(freshRecord, recordTail, loc.asSynthetic)
+          ps <- traverseM(pats)(visitRecordFieldPattern(_, root))
+          patTypes = ps.foldRight(freshRowVar: Type) {
+            case ((f, t, l), acc) => Type.mkRecordRowExtend(f, t, acc, l)
+          }
+          resultType = Type.mkRecord(patTypes, loc)
+          _ <- unifyTypeM(resultType, tvar, loc)
+        } yield resultType
+
```
magnus-madsen, line 2567: I think the above can be improved a bit in terms of readability:

- We can introduce several of the components with `val` bindings before entering the monadic for (i.e. avoid `=` in the for).
- Instead of a fold, a small helper function introduced locally would seem more appropriate.
- `pat` is treated as an option, but can we simply not determine what it is earlier? E.g. either its the empty record or its a variable, but both are types, right?

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
magnus-madsen: But conceptually you are right: it should be cumbersome. Because each RecordRestrict "peels of" one field.

But maybe a helper function would help?

magnus-madsen: (and fields may not occur exactly in the order expected).

main/src/ca/uwaterloo/flix/language/phase/PatternExhaustiveness.scala
```diff
@@ -716,6 +738,19 @@ object PatternExhaustiveness {
...
+      }.zip(all.take(fields.length))
+      val t = all.takeRight(1).head
+      TyCon.Record(fs, Some(t)) :: lst.drop(fields.length + 1)
+    case TyCon.Record(fields, None) =>
+      val all = lst.take(fields.length)
+      val fs = fields.map {
+        case (f, _) => f
+      }.zip(all.take(fields.length))
+      TyCon.Record(fs, None) :: lst.drop(fields.length)
     case a => a :: lst
```
magnus-madsen, line 754: This is wild, I have no idea how this works :)

magnus-madsen:
One final comment, instead of carrying an `Option` through everywhere (for the rest of the ASTs) maybe we should simply always have a pattern there? If there is nothing we simply put a synthetic wildcard pattern.

jaschdoc:
> One final comment, instead of carrying an `Option` through everywhere (for the rest of the ASTs) maybe we should simply always have a pattern there? If there is nothing we simply put a synthetic wildcard pattern.

I'm not sure that works with pattern exhaustiveness?

magnus-madsen:
> > One final comment, instead of carrying an `Option` through everywhere (for the rest of the ASTs) maybe we should simply always have a pattern there? If there is nothing we simply put a synthetic wildcard pattern.
> 
> I'm not sure that works with pattern exhaustiveness?

Option means closed right? But could we not have a RecordEmpty pattern?

jaschdoc:
> > > One final comment, instead of carrying an `Option` through everywhere (for the rest of the ASTs) maybe we should simply always have a pattern there? If there is nothing we simply put a synthetic wildcard pattern.
> 
> > 
> 
> > I'm not sure that works with pattern exhaustiveness?
> 
> 
> 
> Option means closed right? But could we not have a RecordEmpty pattern?

Hmm that might work. It's worth trying it out :)

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
jaschdoc: > (and fields may not occur exactly in the order expected).

What do you mean? `{a, b, c}` does not always appear in the order `a :: b :: c :: Nil`?

jaschdoc: > I cannot comment on the details, but its very important that types and effects are correct for later compiler phases.
> 
> As for effects, it should be straightforward: There is only the effect of the scrutinee.
> 
> As for the types, you can try to enable the `Verifier` and see if it complains about your code specifically (it currently complains of other code).

How do I enable the `Verifier`? Is it a flag?

main/src/ca/uwaterloo/flix/language/phase/Namer.scala
jaschdoc: What about this
```scala
pats.filter(_.pat.isDefined).flatMap(rfp => freeVars(rfp.pat.get)) ++ pat.map(freeVars).getOrElse(Nil)
```

main/src/ca/uwaterloo/flix/api/lsp/Indexer.scala
```diff
@@ -474,13 +474,22 @@ object Indexer {
...
   /**
     * Returns a reverse index for the given patterns `pats0`.
     */
   private def visitPats(pats0: List[Pattern]): Index = traverse(pats0)(visitPat)
 
+  /**
+    * Returns a reverse index for the given [[Pattern.Record.RecordFieldPattern]] `rfp`.
+    */
+  private def visitRecordFieldPattern(rfp: Pattern.Record.RecordFieldPattern): Index = {
+    Index.useOf(rfp.field) ++ visitType(rfp.tpe) ++ visitPat(rfp.pat)
```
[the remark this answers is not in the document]
jaschdoc: I'm not sure what this does, honestly :)

jaschdoc:
We must resolve https://github.com/flix/flix/pull/6204#discussion_r1286723459 and https://github.com/flix/flix/pull/6204#discussion_r1287482481 but apart from than that (and any new issues), I think we are in good shape now.

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
jaschdoc: I've created a helper function to make the record restriction type on the latest commit

magnus-madsen:
@mlutze I am close to pressing the button. Do you want to take one more pass? Perhaps to look at the more complicated bits?

mlutze:
> @mlutze I am close to pressing the button. Do you want to take one more pass? Perhaps to look at the more complicated bits?

Yes, I'll look it over again

main/src/ca/uwaterloo/flix/api/lsp/Indexer.scala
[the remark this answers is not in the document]
mlutze: This file is what says "a variable is used here" to the LSP so that when we click "find references" we can get it highlighted, for example.

That's why I think we need to note that a variable is declared here.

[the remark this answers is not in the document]
jaschdoc: Ok so what do I have to do?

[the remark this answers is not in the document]
magnus-madsen: @mlutze Type variables don't have "definition sites" though.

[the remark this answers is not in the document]
mlutze: What I am getting at is that when you have `case { x } => ...` then the `x` is playing two roles: a field and a variable déclaration site. My question is how this shows up in the IDE, highlighted as a field or as a variable, and whether the indexing conflicts; what happens when I do Find references on `x`?

main/test/ca/uwaterloo/flix/language/phase/TestWeeder.scala
```diff
@@ -377,6 +377,26 @@ class TestWeeder extends AnyFunSuite with TestUtils {
...
+
+  test("NonLinearPattern.05") {
+    val input =
+      """def f(): Bool = match { a = { b = 1 }, b = 1 } {
+        |  case { a = { b = x }, b = x } => true
+        |}
+      """.stripMargin
+    val result = compile(input, Options.TestWithLibNix)
+    expectError[WeederError.NonLinearPattern](result)
+  }
```
mlutze, line 398: One more test request: `{ x, x }`

main/src/ca/uwaterloo/flix/language/phase/Weeder.scala
```diff
@@ -2426,6 +2426,34 @@ object Weeder {
...
+            mapN(visitName(field), traverseOpt(pat)(visit)) {
+              case (_, p) =>
+                val f = Name.mkField(field)
+                val patLoc = mkSL(sp11, sp22)
+                WeededAst.Pattern.Record.RecordFieldPattern(f, p, patLoc)
+            }
+        }
+        val rsVal = traverseOpt(rest)(visit)
+        flatMapN(fsVal, rsVal) {
+          // Pattern { ... }
```
mlutze, line 2442: Thanks for these comments

main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
```diff
@@ -600,6 +600,16 @@ object SemanticTokensProvider {
 
     case Pattern.Tuple(pats, _, _) => pats.flatMap(visitPat).iterator
 
+    case Pattern.Record(pats, pat, tpe, loc) =>
+      val patsVal = pats.flatMap {
+        case Pattern.Record.RecordFieldPattern(_, tpe1, pat1, _) =>
+          visitType(tpe1) ++ visitPat(pat1)
```
mlutze, line 606: Is there no `field` semantic token?

main/src/ca/uwaterloo/flix/language/phase/PatternExhaustiveness.scala
```diff
@@ -716,6 +754,20 @@ object PatternExhaustiveness {
       } else {
         lst.take(numArgs)
       }) :: lst.drop(numArgs)
+    case TyCon.Record(fields, TyCon.RecordEmpty) =>
```
mlutze, line 757: Is there any comment or explanation that could demystify this part?

mlutze (commented):
Sorry for the hold-ups. I still need to get through the Simplifier and Typer.

main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
jaschdoc: Maybe this is it?
```
  /**
    * For identifiers that declare or reference a member property, member field, or member variable.
    */
  case object Property extends SemanticTokenType
```

main/src/ca/uwaterloo/flix/language/phase/PatternExhaustiveness.scala
jaschdoc: Was just ported over from when we had an `Option` type. I don't think it really affects anything.

main/src/ca/uwaterloo/flix/api/lsp/Indexer.scala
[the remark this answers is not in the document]
jaschdoc: Here is the current behavior (and also intended, I think): When using "go to references" on the field `x` where the record is created, it shows the pattern as a reference. When you hit "go to references" on the pattern `x`, the used variable in the match arm is shown as a reference.

Screenshot from 2023-08-17 10-25-37
Screenshot from 2023-08-17 10-25-50

mlutze (approved):

magnus-madsen (approved):

magnus-madsen:
Merged!!!! 🎉

PR #6207, opened 2023-07-04
refactor: `DocAst` contain ret type directly

JonathanStarup:
* changed the DocAst.Def to have the return type directly
* Removed the `DocAst.Expression.fromAtomic` function since its a duplicate function to `OpPrinter.print`, sorry for the confusing namespaces jakob (note: debug printers should not throw!)


magnus-madsen (approved):

jaschdoc:
Thanks for cleaning it up!

Issue #5852, opened 2023-05-02
Rework `addSourceLine` in `GenExpression`


magnus-madsen:
- [ ] (Conceptually) remove all calls to `addSourceLine`.
- [ ] Introduce a call to `addSourceLine` before any ___failable___ expression/bytecode instruction.

For example, the following bytecode instructions are failable:

- ArrayLoad
- ArrayStore
- ArrayLength

because the array could be null or the index could be out of the bounds.

On the other hand, `ArrayLit` and `ArrayNew` are not failable, since they cannot fail.

In general, arithemetic and control-flow cannot fail, but calls to functions (whether internal or external) are failable. 

Let me know if you need more information.

PR #6214, opened 2023-07-04
refactor: add `addSourceLine` in `GenExpression` before failable exprs

jaschdoc:
May require a thorough review. There are some expressions I'm unsure about. For instance, is `NewObject` failable? It should have been type checked, so no runtime should happen.

Closes https://github.com/flix/flix/issues/5852


magnus-madsen (approved):

magnus-madsen:
Lets see how it works.

PR #6217, opened 2023-07-05
feat: unexpected effect

mlutze:


magnus-madsen:
What, exactly, does this change do? :)

mlutze:
This just adds two new errors: unexpected effect and possible effect upcast. If we're expecting a supereffect of the inferred effect, it suggests to upcast. Note that there is no special casing for when we are looking at the top level, so there's no "add this to the effect annotation" prompt.

magnus-madsen:
> This just adds two new errors: unexpected effect and possible effect upcast. If we're expecting a supereffect of the inferred effect, it suggests to upcast. Note that there is no special casing for when we are looking at the top level, so there's no "add this to the effect annotation" prompt.

Can we use this to build a quickfix or not yet?

magnus-madsen:
Like what's the improvement?

mlutze:
Not the quickfix you want. The improvement is that before it was something like "mismatched effects: type one: ... type two: ..." so now it's at least a bit more specific and tells you that you can upcast.

magnus-madsen (approved):

PR #6219, opened 2023-07-05
chore: simplify `LiftedAst` and `LoweredAst`

magnus-madsen:


mlutze:
Is the new rule that back-end ASTs are not-fully qualified?

magnus-madsen:
> Is the new rule that back-end ASTs are not-fully qualified?

Yes, ultimately all ASTs. We are trying to slim down the visual space.

Obviously some phases will still be prefixed, e.g. when translating from one AST to another AST.

magnus-madsen:
I don't see the benefit (anymore) in having them prefix. You?

mlutze:
The main thing is that in the front end we have e.g. `NamedAst.Type` and `Type`. I _think_ there's no AST wherein we could have confusion, but I'm not totally sure.

magnus-madsen:
> The main thing is that in the front end we have e.g. `NamedAst.Type` and `Type`. I _think_ there's no AST wherein we could have confusion, but I'm not totally sure.

Yeah, but probably we should resolve such situations when we discover them.

PR #6220, opened 2023-07-05
chore: rename `LoweredAst.Expression` to `LoweredAst.Expr`

magnus-madsen:


magnus-madsen:
@mlutze I need permission :)

mlutze (approved):

PR #6221, opened 2023-07-05
refactor: check valid field name in records

jaschdoc:
I don't know if this was intentional, but this is currently allowed:
```scala
pub def main(): Unit \ IO =
    let p = { pub = 125, def = 1284, let = 22 };
    println(p.pub);
    println(p.def);
    println(p.let)
```

This PR makes the compiler throw a `WeederError` when using reserved words. Note that Scala disallows reserved words for fields in (case) classes, too.

### DIFF
--- a/main/test/flix/Test.Exp.Record.Polymorphism.flix
+++ b/main/test/flix/Test.Exp.Record.Polymorphism.flix
@@ -45,7 +45,7 @@ mod Test.Exp.Record.Polymorphism {
     def testRecordPolyExtend02(): Unit =
         let p1 = { fstName = "Luke", lstName = "Skywalker" };
         let p2 = { fstName = "Lucky", lstName = "Luke" };
-        let p3 = { fstName = "Bruce", lstName = "Wayne", alias = "Batman" };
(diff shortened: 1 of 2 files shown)


main/test/flix/Test.Exp.Record.Polymorphism.flix
```diff
@@ -45,7 +45,7 @@ mod Test.Exp.Record.Polymorphism {
     def testRecordPolyExtend02(): Unit =
         let p1 = { fstName = "Luke", lstName = "Skywalker" };
         let p2 = { fstName = "Lucky", lstName = "Luke" };
-        let p3 = { fstName = "Bruce", lstName = "Wayne", alias = "Batman" };
+        let p3 = { fstName = "Bruce", lstName = "Wayne", aka = "Batman" };
```
magnus-madsen, line 48: hehe

jaschdoc: :wink:

magnus-madsen (approved):

PR #6225, opened 2023-07-06
feat: add UseEffCompleter

xdesou01:
++ doesnt seem to work on multiple lines, so for now I am jamming the uses in one line.
I also wasn't sure what do use for `kind` in `case Completion.UseEffCompletion`'s expression, but `enum` seems to work.


magnus-madsen (approved):

magnus-madsen:
Merged! Next ones should not be too bad either.

PR #6237, opened 2023-07-10
feat: UseDefCompleter

xdesou01:
Seems to work as intended.
A decent bit of the code is reused from `UseEnumCompleter` (as opposed to `DefCompleter`), but I used a helper for string stripping with the regex.


magnus-madsen (approved):

magnus-madsen:
Thanks!

Issue #1237, opened 2020-07-06
Add String.wrap and String.unwrap?


magnus-madsen:
Inserts or removes `\n`'s.

magnus-madsen:
Before you write any code, give a quick description of how you intend to implement it. For inspiration try to take a look at other librarys (e.g. lodash, underscore from JavaScript) or what else you can find :)

The idea is to discuss an existing implementation before we try to implement it in Flix.

magnus-madsen:
This would still be nice to have... eventually :)

sockmaster27:
Wrap as in 
```
wrap(11, "Lorem ipsum dolor sit ametametametamet")

      ↓

 Lorem ipsum
 dolor sit
 ametametam-
 etamet
|___________|
     11
```
and unwrapping like
```
unwrap("Lorem\nipsum")

      ↓

 Lorem ipsum
```
?

magnus-madsen:
I think so, @mlutze ?

magnus-madsen:
What does Haskell or other languages do?

mlutze:
I'm not sure the hyphenation is possible though. Typically that happens around syllable boundaries.

sockmaster27:
@mlutze 
Yes it would probably be too hard to do 100% correctly, but I was just thinking in the case of an overflowing word which couldn't be shifted in any other way. To avoid
```
wrap(11, "Lorem ipsum dolor sit ametametametamet")

      ↓

 Lorem ipsum
 dolor sit
 ametametametamet
|___________|
     11
```

mlutze:
I think it's a fair design question.

Looking at a couple relevant libraries:

- [Apache WordUtils](https://commons.apache.org/proper/commons-text/javadocs/api-release/org/apache/commons/text/WordUtils.html#wrap(java.lang.String,int,java.lang.String,boolean)): does not wrap long words by default; but you can do it with an optional parameter
- [Text.Wrap](https://hackage.haskell.org/package/word-wrap-0.5/docs/Text-Wrap.html): takes a `WrapSettings` parameter; does not wrap long words by default


From this small sample it seems like wrapping long words is the less common default, but that it is desirable to have the option.

magnus-madsen:
What's the simplest to implement? Let us start with that.

magnus-madsen:
How difficult would it be to do the following:

[...]

PR #6245, opened 2023-07-11
feat: add `String.unwrap`

sockmaster27:
Related to #1237


main/src/library/String.flix
```diff
@@ -1346,6 +1346,12 @@ mod String {
         else
             Some(dropRight(length(substr.substr), s))
 
+    ///
+    /// Unwrap the string, replacing every newline with a space
+    ///
+    pub def unwrap(s: String): String =
+        replaceChar({from = '\n'}, {to = ' '}, s)
```
mlutze, line 1353: Should we take into account Windows-style and old Mac-style newlines? (CR LF and CR)

magnus-madsen (approved):

PR #6246, opened 2023-07-12
refactor: add skeleton for effect handlers

magnus-madsen:


magnus-madsen:
This is still a bit WIP, but can I get an approval @mlutze  :)

mlutze (approved):
I won't pretend that I understand it.

PR #6248, opened 2023-07-12
feat: add simplfied associated type syntax

### LINKED ISSUE #6196: Add new associated type syntax
Without explicit parameter

mlutze:
fixes #6196

### DIFF
--- a/main/test/flix/Test.Assoc.Type.flix
+++ b/main/test/flix/Test.Assoc.Type.flix
@@ -16,4 +16,20 @@ mod Test.Assoc.Type {
         pub def head(x: String): Option[Char] = if (String.isEmpty(x)) None else Some(String.charAt(0, x))
     }
 
+    class Collection2[t] {
+        type Elm: Type
+
+        pub def isEmpty(x: t): Bool
+        pub def contains(x: t, e: Collection2.Elm[t]): Bool
+        pub def head(x: t): Option[Collection2.Elm[t]]
+    }
+
+    instance Collection2[String] {
+        type Elm = Char
+
+        pub def isEmpty(x: String): Bool = ???
+        pub def contains(x: String, e: Char): Bool = ???
+        pub def head(x: String): Option[Char] = if (String.isEmpty(x)) None else Some(String.charAt(0, x))
+    }
+
 }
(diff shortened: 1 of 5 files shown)


main/test/flix/Test.Assoc.Type.flix
```diff
@@ -16,4 +16,20 @@ mod Test.Assoc.Type {
         pub def head(x: String): Option[Char] = if (String.isEmpty(x)) None else Some(String.charAt(0, x))
     }
 
+    class Collection2[t] {
+        type Elm: Type
```
magnus-madsen, line 20: Can I elide the kind also, and have it default to Type?

main/test/flix/Test.Assoc.Type.flix
```diff
@@ -16,4 +16,20 @@ mod Test.Assoc.Type {
         pub def head(x: String): Option[Char] = if (String.isEmpty(x)) None else Some(String.charAt(0, x))
     }
 
+    class Collection2[t] {
+        type Elm
```
magnus-madsen, line 20: Maybe we can have an intermediate test where the kind is present? :) So we have 3 tests.

main/test/flix/Test.Assoc.Type.flix
```diff
@@ -16,4 +16,20 @@ mod Test.Assoc.Type {
...
+
+        pub def isEmpty(x: t): Bool
+        pub def contains(x: t, e: Collection2.Elm[t]): Bool
```
magnus-madsen, line 23: Later I assume we will put `Elm` into the relevant scope?

mlutze: Yes. This is related to another issue where we have to qualify other signatures.

magnus-madsen (approved):

magnus-madsen:
Excellent. I think this syntax is much preferable (at least for simple cases).

main/test/flix/Test.Assoc.Type.flix
magnus-madsen: Is this fixable or requires a lot of work?

mlutze: Fixable but will be a breaking change

PR #6249, opened 2023-07-12
WIP: generalize Boolean algebras

mlutze:
The idea is that this will make it easy to convert among Boolean representations, so that we can test different minimization, satisfiability, etc. algorithms without having to rewrite everything for each combination of representations.


magnus-madsen:
(Closing for now. We can revisit later, but I don't think we want or need this.)

PR #6255, opened 2023-07-13
feat: allow reference to current scope in class signatures

mlutze:


mlutze:
The big problem here is that we create an inconsistency between instances and classes:

For a class, you enter a new namespace inside the class

```scala
mod M {
    def h(): String = ???

    class C[a] {
        type T
        def f(x: a): T[a] = g(x)
        def g(x: a): T[a] = M.h()
    }
}
```

whereas an instance lives in the parent namespace:

```scala
mod N {
    def h(): String = ???

    instance C[String] {
        type T = Char
        def f(x: String): C.T[String] = C.g(x)
        def g(x: String): T[String] = h()
    }
}
```

We could make the instance live in its own namespace too (the original plan for the namespace refactor):

```scala
mod N {
    def h(): String = ???

    instance C[String] {
        type T = Char
        def f(x: String): T[String] = g(x)
        def g(x: String): T[String] = N.h()
    }
}
```

But this leads to unfortunate things like:

```scala
mod Fixpoint.Ram {
    pub enum RowVar with Eq {
        case Named(String)
        case Index(Int32)
    }

    instance ToString[RowVar] {
        pub def toString(var: RowVar): String = match var {
            case RowVar.Named(name) => name
            case RowVar.Index(i) => "${i}$"
        }
    }
}
```

```
-- Resolution Error -------------------------------------------------- Fixpoint/Ram/RowVar.flix

>> Undefined type 'RowVar'.

26 |         pub def toString(var: RowVar): String = match var {
                                   ^^^^^^
                                   type not found.
```

So you have to do `use RowVar` at the top of your instance.

Maybe we could auto-use the instance's type, but this gets messy...

magnus-madsen:
I understand. Let me dwell on it.

magnus-madsen:
What do you think we should do? It seems strange that the members of the instance would not be in scope.

mlutze:
> What do you think we should do? It seems strange that the members of the instance would not be in scope.

Still thinking about this.

Rust puts traits and impls in their parent scope, and users need to write `Self::xyz` to access members.

magnus-madsen:
I look at `Self` and `self` in Rust. I don't think we want to use that approach or that name.

magnus-madsen:
Without thinking to deeply about it, what if we simply took the type members inside an instance and put them into scope?

mlutze:
> Without thinking to deeply about it, what if we simply took the type members inside an instance and put them into scope?

It would be inconsistent with def members, but might be OK, I guess.

magnus-madsen:
What was the conclusion here?

mlutze:
Afaik we were still discussing; no conclusion yet

magnus-madsen:
> Afaik we were still discussing; no conclusion yet

Does a "hack" work?

magnus-madsen:
I am closing this. I suggest to continue in a ticket.

PR #6260, opened 2023-07-15
feat: add `String.wrap`

sockmaster27:
Fighting for my life with all these different cases and off-by-one errors

Related to #1237

### DIFF
--- a/main/src/library/String.flix
+++ b/main/src/library/String.flix
@@ -1347,6 +1347,79 @@ mod String {
             Some(dropRight(length(substr.substr), s))
 
     ///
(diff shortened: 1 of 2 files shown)


magnus-madsen:
Maybe this is getting out of hande. The intention was never to spend a significant amount of effort or time on this specific ticket.

magnus-madsen:
How difficult would it be to do the following:

If we wrap at say 80 chars and there is a word there then we move it to the next line. If there is a word which is more than 80 chars itself then we hardwrap it (i.e. break it in the middle)? Can that be done in say 20 lines of code?

PR #6261, opened 2023-07-16
chore: support `AtomicOps` in `Verifier`

### LINKED ISSUE #6001: Extend `Verifier` to handle `AtomicOp`s

sockmaster27:
Related to #6001

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
@@ -83,31 +83,144 @@ object Verifier {
     case Expr.ApplyAtomic(op, exps, tpe, _, loc) =>
       val ts = exps.map(visitExpr)
 
-      def opType(sop: SemanticOp) = sop match {
-        case _: SemanticOp.BoolOp => MonoType.Bool
-        case _: SemanticOp.CharOp => MonoType.Char
-        case _: SemanticOp.Float32Op => MonoType.Float32
-        case _: SemanticOp.Float64Op => MonoType.Float64
-        case _: SemanticOp.Int8Op => MonoType.Int8
-        case _: SemanticOp.Int16Op => MonoType.Int16
-        case _: SemanticOp.Int32Op => MonoType.Int32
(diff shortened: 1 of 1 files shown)


sockmaster27:
Am I on the right track?

sockmaster27:
Should a test file be added?
Should the `Verifier` be inserted [here](https://github.com/flix/flix/blob/24f967a9ef00c6eb2d19a4e806d683f8808f0d0b/main/src/ca/uwaterloo/flix/api/Flix.scala#L523), and in that case in which position?

magnus-madsen:
It should be inserted around here:

https://github.com/flix/flix/blob/24f967a9ef00c6eb2d19a4e806d683f8808f0d0b/main/src/ca/uwaterloo/flix/api/Flix.scala#L598

magnus-madsen:
> Should a test file be added?

No test file. You insert it and if it does not fail, we are happy :)

magnus-madsen:
> Am I on the right track?

Not yet. You should follow the same style that was already there. Match on one operation at a time. Do not worry about code duplication for now. The goal is "obviously correct", not "smart".

sockmaster27:
Is the `Verifier` only used when debugging?

magnus-madsen:
> Is the `Verifier` only used when debugging?

More or less. We may decide to enable it by default later.

magnus-madsen:
> The verifier is spitting out a lot of errors that aren't related to this

Hmm. Like what?

magnus-madsen:
> Right now runtime errors thrown by the verifier are only printed to the terminal and do not fail the tests. Is this intentional?

Maybe all the errors was the reason it was made this way? (At least until they have been fixed).

magnus-madsen:
Ugh. I think @mlutze or @JonathanStarup needs to look deeper into this.

But how about we merge in your changes?

sockmaster27:
Sure, but this only covers the unary and binary operations

magnus-madsen:
@sockmaster27 Convert to non-draft.

magnus-madsen (approved):

PR #6279, opened 2023-07-22
fix: move shadowing errors to the shadowing variable

### LINKED ISSUE #4544: Create duplicate error for shadowed variable
Currently only the first variable is highlighted in VSCode.

ZiyaoWei:
I think this fixes #4544?

The VSCode extension now shows:

image


mlutze:
> This did not break any tests but arguably it should - would it be a good idea to refactor the tests in `TestRedundancy` to check the error locations? If so I would be happy to work on that.

In general, we don't check for error locations since it's somewhat brittle. However, I recommend adding a check for `ShadowingName` to all the tests that check for `ShadowedName`.

mlutze (approved):

mlutze:
This looks good. Maybe in a future PR we can find a way to emit only one error per variable usage, instead of duplicating the `ShadowedName` error.

magnus-madsen:
Fixes #4544

PR #6282, opened 2023-07-26
feat: add more `use` code actions

sockmaster27:
This proposes quickfix actions to add `use` statements for classes, effects and types.

It doesn't take into consideration whether they are public though.


magnus-madsen (approved):

magnus-madsen:
Thanks!

PR #6291, opened 2023-07-31
feat: MutPriorityQueue

xdesou01:


magnus-madsen:
Let me know when I should take a look.

main/src/library/MutHeap.flix
```diff
@@ -0,0 +1,240 @@
...
+        MutHeap(rc, ref Array.new(rc, 8) @ rc, ref 0 @ rc)
+
+    ///
+    /// Returns an empty MutHeap with a reserved amount of space
+    ///
+    pub def newWithSpace(rc: Region[r], size: Int32): MutHeap[a, r] \ r =
+        MutHeap(rc, ref Array.new(rc, size) @ rc, ref 0 @ rc)
+
+    ///
+    /// Reserves a specific amount of space for
```
mlutze, line 36: finish your sentence :)

main/src/library/MutHeap.flix
```diff
@@ -0,0 +1,240 @@
...
+    ///
+    pub def reserve(h: MutHeap[a, r], size: Int32): Unit \ r =
+        let MutHeap(rc, arr, s) = h;
+        if (size > deref s) {
```
mlutze, line 40: This should probably check capacity rather than size. You could also make a `capacity` helper function to make this clearer.

main/src/library/MutHeap.flix
```diff
@@ -0,0 +1,240 @@
...
+    ///
+    pub def pop(h: MutHeap[a, r]): Option[a] \ r with Order[a] =
+        let MutHeap(_, _, s) = h;
+        if (deref s > 0) {
+            match peek(h) {
+                case Some(elem) => {
+                    delete(h, elem);
+                    Some(elem)
+                }
+                case _          => None
```
mlutze, line 86: This should never happen, right?

main/src/library/MutHeap.flix
```diff
@@ -0,0 +1,240 @@
...
+            }
+        }
+        else None
+
+    ///
+    /// Inserts an element into a MutHeap
+    ///
+    pub def insert(h: MutHeap[a, r], elem: a): Unit \ r with Order[a] = {
+        let MutHeap(_, arr, s) = h;
+        if (deref s == 0) {
```
mlutze, line 96: Do we need the special behavior for 0 elements?

main/src/library/MutHeap.flix
```diff
@@ -0,0 +1,240 @@
...
+            heapifyUp(deref s, h);
+            s := deref s + 1
+        }
+    }
+
+    ///
+    /// Deletes an element from a MutHeap
+    ///
+    pub def delete(h: MutHeap[a, r], elem: a): Unit \ r with Order[a] = {
+        let MutHeap(_, arr, s) = h;
```
mlutze, line 112: Is there any way to be smarter about finding the element? E.g. if we traverse the like a tree then we can avoid branches if the node we are looking at is too big/small?

main/src/library/MutHeap.flix
```diff
@@ -0,0 +1,240 @@
...
+        }
+    }
+
+    ///
+    /// Returns a string representation of the MutHeap
+    ///
+    pub def toString(h: MutHeap[a, r]): String \ r with ToString[a] = region reg {
+        let MutHeap(_, arr, s) = h;
+        let sb = StringBuilder.new(reg);
+        StringBuilder.appendString!("MutHeap {", sb);
```
mlutze, line 129: We usually do `MutHeap#{...}`

main/src/library/MutHeap.flix
```diff
@@ -0,0 +1,240 @@
...
+        MutHeap(rc, ref Array.new(rc, size) @ rc, ref 0 @ rc)
+
+    ///
+    /// Reserves a specific amount of space for
+    ///
+    pub def reserve(h: MutHeap[a, r], size: Int32): Unit \ r =
+        let MutHeap(rc, arr, s) = h;
+        if (size > deref s) {
+            let newArr = Array.new(rc, size);
+            Array.forEachWithIndex((x, elem) -> {
```
mlutze, line 42: Can we avoid traversing the whole array?

main/src/library/MutHeap.flix
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    ///
+    /// Returns a string representation of the MutHeap
+    ///
+    pub def toString(h: MutHeap[a, r]): String \ r with ToString[a] = region reg {
+        let MutHeap(_, arr, s) = h;
+        let sb = StringBuilder.new(reg);
+        StringBuilder.appendString!("MutHeap {", sb);
+        Array.forEachWithIndex((i, x) -> {
```
mlutze, line 130: Can we avoid traversing the whole array?

main/src/library/MutHeap.flix
```diff
@@ -0,0 +1,240 @@
...
+
+    ///
+    /// Returns an iterator over MutHeap `h`
+    ///
+    /// Modifying `h` during iteration is undefined
+    ///
+    pub def iterator(rc: Region[r1], h: MutHeap[a, r2]): Iterator[a, r1 + r2, r1] \ {r1, r2} =
+        let MutHeap(_, arr, s) = h;
+        let numElems = deref s;
+        let idx = ref 0 @ rc;
```
mlutze, line 149: I tihnk this could be made simpler by just doing `Iterator.range` and then `Iterator.map`

main/src/library/MutHeap.flix
```diff
@@ -0,0 +1,240 @@
...
+                Some(elem)
+            }
+            else {
+                None
+            }
+        };
+        Iterator.iterate(rc, next)
+
+    ///
+    /// Reinforces the max heap invariant after an element is added
```
mlutze, line 163: Nitpicky: Re-enforce rather that reinforce IMO.

https://en.wiktionary.org/wiki/re-enforce

main/src/library/MutHeap.flix
```diff
@@ -0,0 +1,240 @@
...
+    def heapifyDown(idx: Int32, h: MutHeap[a, r]): Unit \ r with Order[a] =
+        let MutHeap(_, arr, s) = h;
+        let size = deref s;
+        let lChildIdx = idx * 2 + 1;
+        let rChildIdx = idx * 2 + 2;
+        let cur = Array.get(idx, deref arr);
+        if (size >= rChildIdx) {
+            if (size == rChildIdx) {
+                let child = Array.get(lChildIdx, deref arr);
+                if (Order.less(cur, child)) {
```
mlutze, line 192: Use `<`

main/src/library/MutHeap.flix
```diff
@@ -0,0 +1,240 @@
...
+                }
+                else ()
+            }
+        }
+        else ()
+
+    ///
+    /// Expands the internal array of MutHeap `h` if the array capacity is being approached
+    ///
+    /// Expansion transpires only when the number of elements in `h` is 1 less than the array capacity
```
mlutze, line 221: Does this mean we never use the last array slot? Why not when it is equal to the array capacity?

main/src/library/MutHeap.flix
```diff
@@ -0,0 +1,240 @@
...
+            let newArr  = Array.new(reg, newCapacity);
+            Array.forEachWithIndex((idx, x) -> Array.put(x, idx, newArr), deref arr);
+            arr := newArr
+        }
+        else ()
+
+    ///
+    /// Inserts each element of a list into the MutHeap `h`
+    ///
+    pub def readFromList(h: MutHeap[a, r], l: List[a]): Unit \ r with Order[a] =
```
mlutze, line 237: This seems like a testing utility, no?

main/test/ca/uwaterloo/flix/library/TestMutHeap.flix
```diff
@@ -0,0 +1,249 @@
...
+mod TestMutHeap {
+
+    /////////////////////////////////////////////////////////////////////////////
+    // size                                                                    //
+    /////////////////////////////////////////////////////////////////////////////
+
+    @Test
+    def testNew01(): Bool = region rc {
+        let MutHeap.MutHeap(_, arr, _) = MutHeap.new(rc);
+        Array.length(deref arr) == 8
```
mlutze, line 26: This is not part of the interface. I don't think we should test this.

main/test/ca/uwaterloo/flix/library/TestMutHeap.flix
```diff
@@ -0,0 +1,249 @@
...
+        MutHeap.delete(h, 9);
+        MutHeap.peek(h) == Some(8)
+    }
+
+    @Test
+    def testPeek04(): Bool = region rc {
+        let h = MutHeap.new(rc) : MutHeap[Int32, rc];
+        MutHeap.peek(h) == None
+    }
+
```
mlutze, line 119: Test request: Two max elements; delete one; check max.

main/src/library/MutHeap.flix
xdesou01: Though I could be mistaken, I don't believe there is. This is because the heap invariant merely asserts that each node's children are smaller than it, but the children need not be ordered against each other in any specific way.

xdesou01: Actually, I suppose it's possible to hasten a bit, but I think it would still be O(n)

main/src/library/MutHeap.flix
xdesou01: Good point

main/src/library/MutHeap.flix
xdesou01: Indeed.

main/src/library/MutHeap.flix
```diff
@@ -0,0 +1,240 @@
...
+    ///
+    /// Returns an empty MutHeap with a reserved amount of space
+    ///
+    pub def newWithSpace(rc: Region[r], size: Int32): MutHeap[a, r] \ r =
+        MutHeap(rc, ref Array.new(rc, size) @ rc, ref 0 @ rc)
+
+    ///
+    /// Reserves a specific amount of space for
+    ///
+    pub def reserve(h: MutHeap[a, r], size: Int32): Unit \ r =
```
magnus-madsen, line 38: I don't think this should be public.

main/src/library/MutHeap.flix
magnus-madsen: and sentences should end with a period :)

main/src/library/MutHeap.flix
```diff
@@ -0,0 +1,240 @@
...
+                }
+                case _          => None
+            }
+        }
+        else None
+
+    ///
+    /// Inserts an element into a MutHeap
+    ///
+    pub def insert(h: MutHeap[a, r], elem: a): Unit \ r with Order[a] = {
```
magnus-madsen, line 94: I think I would prefer the names `enqueue` and `enqueueAll`.

main/src/library/MutHeap.flix
```diff
@@ -0,0 +1,240 @@
...
+    ///
+    pub def peek(h: MutHeap[a, r]): Option[a] \ r =
+        let MutHeap(_, arr, s) = h;
+        if (deref s == 0) None
+        else Some(Array.get(0, deref arr))
+
+    ///
+    /// Removes and returns the top element of a MutHeap
+    ///
+    pub def pop(h: MutHeap[a, r]): Option[a] \ r with Order[a] =
```
magnus-madsen, line 78: -> dequeue.

main/src/library/MutHeap.flix
magnus-madsen: I don't think we should support this operation since it is expensive.

magnus-madsen:
Was all TODOs fixed? So I can take a fresh look?

xdesou01:
> Was all TODOs fixed? So I can take a fresh look?

I believe so. I added some other features after some discussion with Matt.

main/src/library/MutPriorityQueue.flix
```diff
@@ -0,0 +1,218 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+pub enum MutPriorityQueue[a: Type, r: Region] {
+    case MutPriorityQueue(Region[r], Ref[Array[a, r], r], Ref[Int32, r])
```
magnus-madsen, line 18: Document the three components.

main/src/library/MutPriorityQueue.flix
```diff
@@ -0,0 +1,218 @@
...
+}
+
+mod MutPriorityQueue {
+
+    ///
+    /// Returns a string representation of `pq`.
+    ///
+    pub def toString(pq: MutPriorityQueue[a, r]): String \ r with ToString[a] = region reg {
```
magnus-madsen, line 26: reg -> rc

main/src/library/MutPriorityQueue.flix
```diff
@@ -0,0 +1,218 @@
...
+        let MutPriorityQueue(_, arr, s) = pq;
+        let sb = StringBuilder.new(reg);
+        StringBuilder.appendString!("MutPriorityQueue {", sb);
+        Array.forEachWithIndex((i, x) -> {
```
magnus-madsen, line 30: Is there not a better way? What does the other data structures do?

main/src/library/MutPriorityQueue.flix
```diff
@@ -0,0 +1,218 @@
...
+        let MutPriorityQueue(_, _, s) = pq;
+        deref s == 0
+
+    ///
+    /// Optionally returns the top element of `pq`.
+    ///
+    pub def peek(pq: MutPriorityQueue[a, r]): Option[a] \ r =
+        let MutPriorityQueue(_, arr, s) = pq;
+        if (deref s == 0) None
+        else Some(Array.get(0, deref arr))
```
magnus-madsen, line 67: I would prefer

```
if (deref s == 0) 
    None
else 
    Some(Array.get(0, deref arr))
```

main/src/library/MutPriorityQueue.flix
```diff
@@ -0,0 +1,218 @@
...
+            heapifyDown(0, pq);
+            s := deref s - 1;
+            top
+        }
+        else None
+
+    ///
+    /// Enqueues each element in `l` into `pq`.
+    ///
+    pub def enqueueAll(pq: MutPriorityQueue[a, r], m: m[a]): Unit \ r with Order[a], Iterable[m] = region rc {
```
magnus-madsen, line 97: Could we use a foreach loop here?

main/src/library/MutPriorityQueue.flix
```diff
@@ -0,0 +1,218 @@
...
+        let it = Iterable.iterator(rc, m);
+        Iterator.forEach(x -> enqueue(pq, x), it)
+    }
+
+    ///
+    /// Returns an iterator over `pq`.
+    ///
+    /// Modifying `pq` during iteration is undefined and not recommended.
```
magnus-madsen, line 105: Drop comment. This is true for all iterators.

main/src/library/MutPriorityQueue.flix
xdesou01: I based this off of the toString functions in `MutSet`, `MutList`, `MutMap`, and `MutDeque`, which all follow a similar approach.

main/src/library/MutPriorityQueue.flix
xdesou01: *Correction: should be components 1 and 2, not 2 and 3

magnus-madsen:
Will you update the PR and then I can take another look? :)

xdesou01:
> Will you update the PR and then I can take another look? :)

yes, I'm just stuck on the `foreach` comment

main/src/library/MutPriorityQueue.flix
xdesou01: I was confused as to why this would be more efficient

xdesou01: And what the alternate option might entail

main/src/library/MutPriorityQueue.flix
```diff
@@ -0,0 +1,223 @@
...
+    case MutPriorityQueue(Region[r], Ref[Array[a, r], r], Ref[Int32, r])
+}
+
+mod MutPriorityQueue {
+
+    ///
+    /// Represents a mutable priority queue.
+    /// Explanation of component types (left to right):
+    /// Component 2: A reference to the backing array.
+    /// Component 3: A reference to the number of elements in the mutable priority queue.
```
magnus-madsen, line 27: Why are these comments here????

magnus-madsen: I think it should be attached to the enum just above.

magnus-madsen: Component 1: The region capability the queue is associated with.

main/src/library/MutPriorityQueue.flix
magnus-madsen: Roger. We should probably simplify those as well, but I guess that can be done later.

main/src/library/MutPriorityQueue.flix
```diff
@@ -0,0 +1,223 @@
...
+    ///
+    /// The maximum element (if it exists) can always be accessed in constant time.
```
magnus-madsen, line 29: Whats with this comment on toString??

main/src/library/MutPriorityQueue.flix
magnus-madsen: Seems not resolved?

magnus-madsen:
Ok, I think some merges happened the wrong way :) Maybe take another look?

main/src/library/MutPriorityQueue.flix
xdesou01: You asked if it could be done a different way. I said it was the same as the toString functions for other mutable structures.

magnus-madsen: But why does the doc comment for toString say "The maximum element (if it exists) can always be accessed in constant time."

xdesou01: Yes, that's my bad. I realize I replaced the comment in the wrong area. Fixing now.

xdesou01: The comment was meant for the enum

magnus-madsen: Ok, maybe do pass through the GitHub change to see that everything is in the right place.

xdesou01: Ok. It looks correct to me now.

magnus-madsen:
Drop `merge` and then we can "merge" :)

magnus-madsen (approved):

magnus-madsen:
Some tests failing though

magnus-madsen:
Merged. Thanks!

PR #6294, opened 2023-08-01
feat: improve error messages for associated types

mlutze:


magnus-madsen:
Example, just for fun?

mlutze:
```
>> Irreducible associated type: T[a]

218 | def g(x: a): String with C[a] = C.f(x)
          ^
          irreducible associated type.
```

```
>> Unsupported type equality: String ~ Int32

217 | def g(x: a): String with C[a] where C.T[a] ~ Int32 = C.f(x)
          ^
          unsupported type equality.
```
The location is not so nice. Further down the line I will try to pipe locations through the system. But this is at least a step in the right direction and should help during development.

magnus-madsen (approved):

PR #6303, opened 2023-08-02
refactor: introduce `Ast.Derivations`

### LINKED ISSUE #6302: Introduce `Ast.Derivations`
- [ ] Introduce a new case class (right after `Derivation`) in Ast.scala:

```scala
  /**
   * Represents a list of derivations with a source location.
   *
   * The source location spans the entire `with X, Y, z` clause.
   *
   * If there is no `with`-clause then the source location has zero
   * length and is positioned right after the enum type. For example,
   * if the enum is `enum Color {` then the source position would point
   * to the position right after `r` and have zero width.
   */
  case class Derivations(classes: List[Derivation], loc: SourceLocation)
```

- [ ] Refactor `TypedAst.Enum` to have an `Ast.Derivations` instead of a list of `Ast.Derivation`.

sockmaster27:
Fixes  #6302

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/NamedAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/NamedAst.scala
@@ -46,9 +46,9 @@ object NamedAst {
 
     case class Def(sym: Symbol.DefnSym, spec: Spec, exp: Expr) extends Declaration
 
-    case class Enum(doc: Ast.Doc, ann: Ast.Annotations, mod: Ast.Modifiers, sym: Symbol.EnumSym, tparams: TypeParams, derives: List[Name.QName], cases: List[Declaration.Case], loc: SourceLocation) extends Declaration
+    case class Enum(doc: Ast.Doc, ann: Ast.Annotations, mod: Ast.Modifiers, sym: Symbol.EnumSym, tparams: TypeParams, derives: List[Name.QName], derivesLoc: SourceLocation, cases: List[Declaration.Case], loc: SourceLocation) extends Declaration
 
-    case class RestrictableEnum(doc: Ast.Doc, ann: Ast.Annotations, mod: Ast.Modifiers, sym: Symbol.RestrictableEnumSym, index: TypeParam, tparams: TypeParams, derives: List[Name.QName], cases: List[Declaration.RestrictableCase], loc: SourceLocation) extends Declaration
+    case class RestrictableEnum(doc: Ast.Doc, ann: Ast.Annotations, mod: Ast.Modifiers, sym: Symbol.RestrictableEnumSym, index: TypeParam, tparams: TypeParams, derives: List[Name.QName], derivesLoc: SourceLocation, cases: List[Declaration.RestrictableCase], loc: SourceLocation) extends Declaration
 
     case class TypeAlias(doc: Ast.Doc, mod: Ast.Modifiers, sym: Symbol.TypeAliasSym, tparams: TypeParams, tpe: Type, loc: SourceLocation) extends Declaration
 
--- a/main/src/ca/uwaterloo/flix/language/phase/Namer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Namer.scala
@@ -141,7 +141,7 @@ object Namer {
     case NamedAst.Declaration.Def(sym, spec, exp) =>
       tryAddToTable(table0, sym.namespace, sym.name, decl)
 
-    case NamedAst.Declaration.Enum(doc, ann, mod, sym, tparams, derives, cases, loc) =>
+    case NamedAst.Declaration.Enum(doc, ann, mod, sym, tparams, derives, _, cases, loc) =>
       val table1Val = tryAddToTable(table0, sym.namespace, sym.name, decl)
       flatMapN(table1Val) {
         case table1 => fold(cases, table1) {
@@ -149,7 +149,7 @@ object Namer {
         }
       }
 
-    case NamedAst.Declaration.RestrictableEnum(doc, ann, mod, sym, index, tparams, derives, cases, loc) =>
+    case NamedAst.Declaration.RestrictableEnum(doc, ann, mod, sym, index, tparams, derives, _, cases, loc) =>
       val table1Val = tryAddToTable(table0, sym.namespace, sym.name, decl)
       flatMapN(table1Val) {
         case table1 => fold(cases, table1) {
@@ -312,7 +312,7 @@ object Namer {
     * Performs naming on the given enum `enum0`.
     */
   private def visitEnum(enum0: WeededAst.Declaration.Enum, ns0: Name.NName)(implicit flix: Flix): Validation[NamedAst.Declaration.Enum, NameError] = enum0 match {
-    case WeededAst.Declaration.Enum(doc, ann, mod0, ident, tparams0, derives, cases0, loc) =>
+    case WeededAst.Declaration.Enum(doc, ann, mod0, ident, tparams0, derives, derivesLoc, cases0, loc) =>
       val sym = Symbol.mkEnumSym(ns0, ident)
 
       // Compute the type parameters.
@@ -323,7 +323,7 @@ object Namer {
 
       mapN(casesVal) {
         case cases =>
-          NamedAst.Declaration.Enum(doc, ann, mod, sym, tparams, derives, cases, loc)
+          NamedAst.Declaration.Enum(doc, ann, mod, sym, tparams, derives, derivesLoc, cases, loc)
       }
   }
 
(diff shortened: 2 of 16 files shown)


main/src/ca/uwaterloo/flix/language/phase/Parser.scala
```diff
@@ -292,7 +293,7 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
       }
 
       rule {
-        Documentation ~ Annotations ~ Modifiers ~ SP ~ keyword("enum") ~ WS ~ Names.Type ~ TypeParams ~ optional(Types.Tuple) ~ Derivations ~ optWS ~ Body ~ SP ~> ParsedAst.Declaration.Enum
+        Documentation ~ Annotations ~ Modifiers ~ SP ~ keyword("enum") ~ WS ~ Names.Type ~ TypeParams ~ optional(Types.Tuple) ~ ((optWS ~ SP ~ Derivations ~ SP) | (SP ~ push(Seq.empty[Name.QName]) ~ SP)) ~ optWS ~ Body ~ SP ~> ParsedAst.Declaration.Enum
```
sockmaster27, line 296: I wanted to move this into a rule by itself, but I had trouble because of the fact that there is no `Rule3` for returning the two positions and the sequence. There is the `RuleN` but I couldn't get it to accept the mixing of types.

main/src/ca/uwaterloo/flix/language/ast/NamedAst.scala
```diff
@@ -46,9 +46,9 @@ object NamedAst {
 
     case class Def(sym: Symbol.DefnSym, spec: Spec, exp: Expr) extends Declaration
 
-    case class Enum(doc: Ast.Doc, ann: Ast.Annotations, mod: Ast.Modifiers, sym: Symbol.EnumSym, tparams: TypeParams, derives: List[Name.QName], cases: List[Declaration.Case], loc: SourceLocation) extends Declaration
+    case class Enum(doc: Ast.Doc, ann: Ast.Annotations, mod: Ast.Modifiers, sym: Symbol.EnumSym, tparams: TypeParams, derives: List[Name.QName], derivesLoc: SourceLocation, cases: List[Declaration.Case], loc: SourceLocation) extends Declaration
```
magnus-madsen, line 49: Here I would still want a `Derivations` class, but instead I would put it inside `NamedAst.scala` near the bottom of the file.

E.g.

```
case class Derivations(classes: List[Name.QName], loc: SourceLocation)
````

The same applies for ParsedAst, WeededAst, etc.

Comment (magnus-madsen) [reply main/src/ca/uwaterloo/flix/language/ast/NamedAst.scala:49]: Makes sense?

File: main/src/ca/uwaterloo/flix/language/phase/Parser.scala (hunk 1 above)
Comment (magnus-madsen) [reply main/src/ca/uwaterloo/flix/language/phase/Parser.scala:296]: We can simply add a `Derivations` rule which returns a `ParsedAst.Derivations` and then we are all good 👍

File: main/src/ca/uwaterloo/flix/language/ast/NamedAst.scala (hunk 2 above)
Comment (sockmaster27) [reply main/src/ca/uwaterloo/flix/language/ast/NamedAst.scala:49]: Yeah 👍

File: main/src/ca/uwaterloo/flix/language/phase/Parser.scala [hunk 3]
```diff
@@ -23,6 +23,7 @@ import ca.uwaterloo.flix.language.ast._
 import ca.uwaterloo.flix.util.Validation._
 import ca.uwaterloo.flix.util.{ParOps, Validation}
 import org.parboiled2._
+import org.parboiled2.support.hlist.HNil
```
magnus-madsen, line 26: Unused import or?

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
```diff
@@ -420,8 +421,14 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
       }
     }
 
-    def Derivations: Rule1[Seq[Name.QName]] = rule {
-      optWS ~ optional(keyword("with") ~ WS ~ oneOrMore(Names.QualifiedClass).separatedBy(optWS ~ "," ~ optWS)) ~> ((o: Option[Seq[Name.QName]]) => o.getOrElse(Seq.empty))
+    def Derivations: Rule1[ParsedAst.Derivations] = {
+      def WithStatement: Rule1[Seq[Name.QName]] = rule {
```
magnus-madsen, line 425: WithStatement -> WithClause

magnus-madsen:
@sockmaster27 Minor changes requested.

@mlutze Any comments?

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
sockmaster27: Yes - forgot to remove

magnus-madsen:
So once this is in, which should be soon, it will be easier to write a nice quickfix :)

main/src/ca/uwaterloo/flix/language/phase/Namer.scala
```diff
@@ -323,15 +323,15 @@ object Namer {
 
       mapN(casesVal) {
         case cases =>
-          NamedAst.Declaration.Enum(doc, ann, mod, sym, tparams, derives, cases, loc)
+          NamedAst.Declaration.Enum(doc, ann, mod, sym, tparams, NamedAst.Derivations(classes, loc1), cases, loc2)
```
mlutze, line 326: Could we get a slightly more semantic name than loc1 here?

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -541,16 +541,16 @@ object Resolver {
...
     */
   def resolveEnum(e0: NamedAst.Declaration.Enum, env0: ListMap[String, Resolution], taenv: Map[Symbol.TypeAliasSym, ResolvedAst.Declaration.TypeAlias], ns0: Name.NName, root: NamedAst.Root)(implicit flix: Flix): Validation[ResolvedAst.Declaration.Enum, ResolutionError] = e0 match {
-    case NamedAst.Declaration.Enum(doc, ann, mod, sym, tparams0, derives0, cases0, loc) =>
+    case NamedAst.Declaration.Enum(doc, ann, mod, sym, tparams0, NamedAst.Derivations(classes, loc1), cases0, loc2) =>
       val tparamsVal = resolveTypeParams(tparams0, env0, ns0, root)
       flatMapN(tparamsVal) {
         case tparams =>
           val env = env0 ++ mkTypeParamEnv(tparams.tparams)
-          val derivesVal = resolveDerivations(derives0, env, ns0, root)
+          val derivesVal = resolveDerivations(classes, env, ns0, root)
```
mlutze, line 549: Since we already have a dedicated function, can we just take the derives as input?

mlutze (approved):

magnus-madsen:
@sockmaster27 If you fix Matt's comments then I can merge :)

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
magnus-madsen: I think its because we don't like the nested pattern match.

Instead just change the signature of `resolveDerivations` then it takes a `derivations` value. That seems reasonable, no?

sockmaster27: We could, but again, then we'd just have to pull out both of the fields at the end

magnus-madsen: I don't understand. The point is to avoid a deep pattern match and have resolveDerivations take the derivation type.

magnus-madsen: This way it is structurally recursive: it means that the control-flow follows the structure of the data.

magnus-madsen: Also, lets say we extend Derivations with more information then that would immediately be available inside resolveDerivations

sockmaster27: I just mean that we can either do
```
case NamedAst.Declaration.Enum(... NamedAst.Derivations(classes, dloc), ...) =>
  ...
  resolveDerivations(classes, ...)
  ...
  ResolvedAst.Declaration.Enum(... Ast.Derivations(derives, dloc), ...)
```
or
```
case NamedAst.Declaration.Enum(... derives0, ...) =>
  ...
  resolveDerivations(derives0, ...)
  ...
  ResolvedAst.Declaration.Enum(... Ast.Derivations(derives, derives0.loc), ...)
```

We can certainly do it like that, but I'm just trying to say that no matter what it's not just going to be
```
case NamedAst.Declaration.Enum(... derives0, ...) =>
  ...
  resolveDerivations(derives0, ...)
  ...
  ResolvedAst.Declaration.Enum(... derives0, ...)
```

magnus-madsen: Take and return?

sockmaster27: Yes, I get it now

sockmaster27: I think

magnus-madsen:
The same refactoring can be done for the earlier phases.

magnus-madsen:
Excellent. Now we can work on the quickfixes.

PR #6316, opened 2023-08-09
feat: generate HTML documentation

### LINKED ISSUE #6309: Add `HtmlDocumentor`
Add a new object `HtmlDocumentor` (next to `Documentor.scala`) which takes a `TypedAst.Root` and generates static html files.

The phase should output HTML files to `build/doc` and should generate one HTML per module.

For example:

- `List.html`
- `Fixpoint.Ast.html`

and so forth.

The generated HTML should mirror the __right hand side__ of https://api.flix.dev/ (source code here: https://github.com/flix/api.flix.dev).

The stylesheet should be re-usable, as is.

Once we can generate the right-hand side, we can focus on how to generate a menu on the left and how to support search.

It is important to structure the class as a collection of small private functions that each generate HTML snipplets for their input. Following the structure of the AST would be useful.

Later we will have to decide on how to include the CSS and favicons, etc.

The goal is to generate fully static HTML pages, style them with CSS, and only use minimal JavaScript for search.

The goal is also to slightly restructure how some information is presented, but we will get to that.

sockmaster27:
Fixes #6309


sockmaster27:
May i change the name of the `Documentor` phase to `JsonDocumentor`/`DocumentorJson`?

magnus-madsen:
> May i change the name of the `Documentor` phase to `JsonDocumentor`/`DocumentorJson`?

Yes, but it should be done in a separate PR. Then we can have HtmlDocumentor eller JSONDocumentor.

sockmaster27:
TODO: Figure out what's wrong with markdown formatter

billede

magnus-madsen:
!!!jar

sockmaster27:
@magnus-madsen 
Right now it's hard coded to link to the Flix GitHub repo. I'm wondering if this is fine for now, and in that case how we could integrate it without exposing something broken. 
Maybe it could be hid behind a feature flag, or just let the command be marked as experimental?

magnus-madsen:
> @magnus-madsen Right now it's hard coded to link to the Flix GitHub repo. I'm wondering if this is fine for now, and in that case how we could integrate it without exposing something broken. Maybe it could be hid behind a feature flag, or just let the command be marked as experimental?

In iteration 1 that is fine. We will revisit later.

magnus-madsen:
My plan is for us to do several iterations, but first we need to replicate more or less what we have :)

sockmaster27:
Ok. I'll do some last things on Monday and after then I think it's ready to be merged.

build.gradle
```diff
@@ -56,6 +57,10 @@ sourceSets {
         scala {
             srcDirs = ['main/src']
         }
+
+        resources {
+            srcDirs = ['main/src/resources']
+        }
```
sockmaster27, line 63: IntelliJ seems to be struggling with this, but Gradle handles it fine.

magnus-madsen:
Is there a way to run it yet? I tried the jar but could not get it work.

sockmaster27:
@magnus-madsen 
It should work if you run
```
java -jar '...\\flix.jar' --doc-html
```

sockmaster27:
Wothout the `...` of course

magnus-madsen:
!!!jar

magnus-madsen:
We need to call mkdirs at some point:

```
Exception in thread "main" java.nio.file.NoSuchFileException: .\build\doc
        at java.base/sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:85)
        at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:103)
```

magnus-madsen:
But it looks like a reasonable starting point 👍 . A few things I noticed:

- Need to add some sorting order.
- Generated HTML is not on one line, but a few lines (?).
- Missing annotations (I think (??)).
- For classes we probably want to show all their members (expanded). Ditto for effects.

But overall this seems like good progress :)

sockmaster27:
> We need to call mkdirs at some point:

Hmm weird... I thought it was already doing that.

sockmaster27:
> * Generated HTML is not on one line, but a few lines (?).

This seems to be because the markup library (txtmark) puts some newlines in between tags. I don't believe it can be changed, but we could either strip them, or just let them be.

magnus-madsen:
> > * Generated HTML is not on one line, but a few lines (?).
> 
> This seems to be because the markup library (txtmark) puts some newlines in between tags. I don't believe it can be changed, but we could either strip them, or just let them be.

Perhaps we just let them be or deal with it later.

magnus-madsen:
Do you want some feedback on the code now or do you want to finish the draft?

sockmaster27:
> TODO: Figure out what's wrong with markdown formatter
> 
> billede

Escaping the string before it is formatted solves this, but introduces this problem instead: 

image

sockmaster27:
The remaining problems are:
- Code between a `<` and a `>` is not formatted properly in markdown
- The resource folder is treated weirdly when compiling with IntelliJ
- `benchmarks` should probably not be in the resource folder
- The source links are hardcoded to the Flix GitHub repo

sockmaster27:
Oh yeah, and the fact that `Files.createDirectories` doesn't appear to work

magnus-madsen:
I will do two reviews; one of the look and one of the code.

As for the look: Great job so far! A few minor comments:

- [ ] I think the title should just be "Flix | List" and not "Flix Doc | List"
- [ ] The color `--doc-code-color` seems off compared to `api.flix.dev`. Here it is red, but there it is more purple.
- [ ] Missing favicons (I think they can just be copied as-is).

Next I will take a look at the code.

build.gradle
magnus-madsen: OK- that is acceptable.

build.gradle
```diff
@@ -22,6 +22,7 @@ dependencies {
     implementation('org.parboiled:parboiled_2.13:2.4.1')
     implementation('org.scalactic:scalactic_2.13:3.2.15')
     implementation("org.scala-lang.modules:scala-parallel-collections_2.13:0.2.0")
+    implementation('com.github.rjeschke:txtmark:0.13')
```
magnus-madsen, line 25: We are generally very, very, very careful about including new libraries. In this case I guess it is almost required, but can I ask you to explore:

- How much does this increase the Flix har filesize by (it does not seem much).
- Can this library be trusted and is it the best/most popular/most tested library out there?
- What would we lose if we drop this library? (I guess some formatting?)

main/src/ca/uwaterloo/flix/Main.scala
```diff
@@ -85,7 +85,8 @@ object Main {
     var options = Options(
       lib = cmdOpts.xlib,
       debug = cmdOpts.xdebug,
-      documentor = cmdOpts.documentor,
+      documentorJson = cmdOpts.documentorJson,
```
magnus-madsen, line 88: The JsonDocumenter is out-of-date. I don't want to delete it-- since I may restore it later.

But instead of having two cmd options let us just keep our single doc command.

We can then use it to trigger *both* documentors. Makes sense?

main/src/ca/uwaterloo/flix/Main.scala
```diff
@@ -278,7 +279,8 @@ object Main {
     */
   case class CmdOpts(command: Command = Command.None,
                      args: Option[String] = None,
-                     documentor: Boolean = false,
+                     documentorJson: Boolean = false,
+                     documentorHtml: Boolean = false,
```
magnus-madsen, line 283: so these are merged into one / reverted.

main/src/ca/uwaterloo/flix/api/Flix.scala
```diff
@@ -89,6 +89,7 @@ class Flix {
     * A cache of ASTs for debugging.
     */
   private var cachedDocumentorAst: TypedAst.Root = TypedAst.empty
+  private var cachedHtmlDocumentorAst: TypedAst.Root = TypedAst.empty
```
magnus-madsen, line 92: The original code is actually strange. I suggest to drop these two fields.

main/src/ca/uwaterloo/flix/api/Flix.scala
```diff
@@ -585,7 +586,8 @@ class Flix {
     initForkJoin()
 
     cachedDocumentorAst = Documentor.run(typedAst)
-    cachedLoweringAst = Lowering.run(cachedDocumentorAst)
+    cachedHtmlDocumentorAst = HtmlDocumentor.run(cachedDocumentorAst)
```
magnus-madsen, line 589: And down here you run the phases, but don't store anything.

main/src/ca/uwaterloo/flix/language/fmt/FormatType.scala
```diff
@@ -86,6 +86,18 @@ object FormatType {
...
+    */
+  def formatSimpleType(tpe: SimpleType)(implicit flix: Flix): String =
+    formatSimpleTypeWithOptions(tpe, flix.getFormatOptions)
+
+  /**
+    * Transforms the given simple type into a string, using the given format options.
+    */
+  def formatSimpleTypeWithOptions(tpe: SimpleType, fmt: FormatOptions): String =
+    format(tpe)(fmt)
+
```
magnus-madsen, line 100: This looks a bit boiler-platy. What's happening here? One helper that forwards to another? Perhaps at least formatSimplyTypeWithOptions can be moved into formatSimpleType?

@mlutze Any thoughts?

main/src/ca/uwaterloo/flix/language/phase/Documentor.scala
```diff
@@ -47,7 +47,7 @@ object Documentor {
     //
     // Determine whether to generate documentation.
     //
-    if (!flix.options.documentor) {
+    if (!flix.options.documentorJson) {
```
magnus-madsen, line 50: This is reverted.

main/src/ca/uwaterloo/flix/language/phase/Documentor.scala
```diff
@@ -47,7 +47,7 @@ object Documentor {
...
       return root
```
magnus-madsen, line 51: I think we should just return Unit here (see my earlier comments on the phases).

main/src/ca/uwaterloo/flix/util/Options.scala
```diff
@@ -114,7 +116,8 @@ object Options {
   */
 case class Options(lib: LibLevel,
                    debug: Boolean,
-                   documentor: Boolean,
+                   documentorJson: Boolean,
+                   documentorHtml: Boolean,
```
magnus-madsen, line 120: This is reverted, as discussed.

main/src/resources/doc/styles.css
```diff
@@ -0,0 +1,181 @@
...
+}
+code .kind {
+    color: var(--kind-color);
+}
+code .effect {
+    color: var(--effect-color);
+}
+code .tpe-constraint {
+    text-decoration: underline;
+}
```
magnus-madsen, line 181: Looks reasonable, later we can tweak this file :)

main/test/ca/uwaterloo/flix/TestMain.scala
```diff
@@ -69,10 +69,16 @@ class TestMain extends AnyFunSuite {
...
+    val args = Array("--doc-json", "p.flix")
     val opts = Main.parseCmdOpts(args).get
-    assert(opts.documentor)
+    assert(opts.documentorJson)
+  }
+
+  test("--doc-html") {
+    val args = Array("--doc-html", "p.flix")
+    val opts = Main.parseCmdOpts(args).get
+    assert(opts.documentorHtml)
```
magnus-madsen, line 81: This goes as before.

main/src/ca/uwaterloo/flix/language/phase/HtmlDocumentor.scala
```diff
@@ -0,0 +1,640 @@
...
+    * The directory where to write the ouput.
+    */
+  val OutputDirectory: Path = Paths.get("./build/doc")
+
+  /**
+    * The path to the the stylesheet, relative to the resources folder.
+    */
+  val Stylesheet: String = "/doc/styles.css"
+
+  def run(root: TypedAst.Root)(implicit flix: Flix): TypedAst.Root = flix.phase("HtmlDocumentor") {
```
magnus-madsen, line 49: Call flix.phase, but lets change return type to Unit.

main/src/ca/uwaterloo/flix/language/phase/HtmlDocumentor.scala
```diff
@@ -0,0 +1,640 @@
...
+
+  /**
+    * Escape any HTML in the string.
+    */
+  private def esc(s: String): String = xml.Utility.escape(s)
+
+  /**
+    * A represention of a module that's easier to work with while generating documention.
+    */
+  private case class Module(namespace: List[String],
```
magnus-madsen, line 616: I am a bit worried about this approach and below with class.

Because this means there is now duplicate information and more importantly, we have to maintain these data structures.

Why do you need them?

main/src/ca/uwaterloo/flix/language/phase/HtmlDocumentor.scala
```diff
@@ -0,0 +1,640 @@
...
+      case ex: IOException => throw new RuntimeException(s"Unable to write to path '$path'.", ex)
+    }
+  }
+
+  /**
+    * Create a raw link to the given `SourceLocation`.
+    */
+  private def createLink(loc: SourceLocation): String = {
+    // TODO make it also work for local user code
+    s"https://github.com/flix/flix/blob/master/main/src/library/${loc.source.name}#L${loc.beginLine}-L${loc.beginLine}"
```
magnus-madsen, line 605: Lets move the URL to a constant as the others above.

main/src/ca/uwaterloo/flix/language/phase/HtmlDocumentor.scala
```diff
@@ -0,0 +1,640 @@
...
+  private def docList[T](list: List[T])(docElt: T => Unit)(implicit flix: Flix, sb: StringBuilder): Unit = {
+    for ((e, i) <- list.zipWithIndex) {
+      docElt(e)
+      if (i < list.length - 1) {
+        sb.append(", ")
+      }
+    }
+  }
+
+  private def clearOutputDirectory(): Unit = {
```
magnus-madsen, line 563: Drop this-- in general Flix will never clear directories. (And if it will, it will be the job of the package manager and an eventual clean command.)

main/src/ca/uwaterloo/flix/language/phase/HtmlDocumentor.scala
```diff
@@ -0,0 +1,640 @@
...
+    // Determine whether to generate documentation.
+    //
+    if (!flix.options.documentorHtml) {
+      return root
+    }
+
+    clearOutputDirectory()
+    writeStyles()
+    val modules = splitModules(root)
+    modules.par.foreach {
```
magnus-madsen, line 60: Lets not write to the disk in parallel. Better to just do it one file at a time.

main/src/ca/uwaterloo/flix/language/phase/HtmlDocumentor.scala
```diff
@@ -0,0 +1,640 @@
...
+
+    sb.append("</body>")
+
+    sb.toString()
+  }
+
+  /**
+    * Generates the string representing the head of the HTML document.
+    */
+  private def mkHead(name: String): String = {
```
magnus-madsen, line 196: Use raw interpolated strings:

```scala
s"""fooo
    |bar
    |baz
"""
```

main/src/ca/uwaterloo/flix/language/phase/HtmlDocumentor.scala
```diff
@@ -0,0 +1,640 @@
...
+  }
+
+  /**
+    * Documents the given `Class`, `clazz`.
+    *
+    * The result will be appended to the given `StringBuilder`, `sb`.
+    */
+  private def docClass(clazz: Class)(implicit flix: Flix, sb: StringBuilder): Unit = {
+    docAnnotations(clazz.ann)
+    sb.append("<code>")
```
magnus-madsen, line 266: We could use raw interpolated strings here (and still append the "big string" to the string builder). But I am not sure if it is better. Thoughts?

magnus-madsen:
@sockmaster27 I think this looks very promising. I added a lot of comments and then we can do one more round once they have been fixed 👍

main/src/ca/uwaterloo/flix/language/fmt/FormatType.scala
sockmaster27: It's just an attempt to somewhat mirror the rest of the API

main/src/ca/uwaterloo/flix/language/phase/HtmlDocumentor.scala
sockmaster27: It's just nicer to have all the relevant data in one structure mirroring the way that you would think about it, instead of having each `docX` function contain both the logic for extracting it from the `root` and generating the code.

main/src/ca/uwaterloo/flix/language/phase/HtmlDocumentor.scala
magnus-madsen: Yes, I understand.

main/src/ca/uwaterloo/flix/language/phase/HtmlDocumentor.scala
sockmaster27: How come?

magnus-madsen: It is slower.

main/src/ca/uwaterloo/flix/language/phase/HtmlDocumentor.scala
sockmaster27: I started out doing something similar, but it became pretty hard to read with the different indentation levels, so I changed it to this.

sockmaster27:
> * Missing favicons (I think they can just be copied as-is).

Shouldn't we be using the logo for this?

build.gradle
sockmaster27: - How much does this increase the Flix har filesize by (it does not seem much).
  - 62,259 bytes
- Can this library be trusted and is it the best/most popular/most tested library out there?
  - I believe it can be trusted, even though it hasn't been updated since 2015. It's apparently known to be pretty fast and minimal, and I get the impression that it is generally accepted. If we want something that's a bit more actively maintained I could look into Atlassian's [commonmark-java](https://github.com/commonmark/commonmark-java). This might be worth doing at some point anyway, to resolve the first point of this: https://github.com/flix/flix/pull/6316#issuecomment-1677031100.
- What would we lose if we drop this library? (I guess some formatting?)
  - We would lose the markdown support in the documentation. If we were willing to spend time on it, we could look into deciding on a minimal set of features we would like to support (like code and unordered lists) and implement that ourselves, but to me that sounds like a lot of effort to halfway replicate what we could get with a library.

magnus-madsen:
> > * Missing favicons (I think they can just be copied as-is).
> 
> Shouldn't we be using the logo for this?

What I mean is that there are some favicons on api.flix.dev right?

main/src/ca/uwaterloo/flix/language/phase/HtmlDocumentor.scala
magnus-madsen: Ok

main/src/ca/uwaterloo/flix/language/phase/HtmlDocumentor.scala
magnus-madsen: Okay. We can start this way and see what happens.

magnus-madsen:
!!!jar

magnus-madsen:
If you include the icons and fix the conflict then I think we can merge this.

sockmaster27:
This one, right?
image
I was just thinking that maybe it would make more sense to use the usual logo:
image

magnus-madsen:
The F one, the one we currently use.

(We need to resolve the logo situation, but that is a separate task.)

sockmaster27:
Okay. Does it exist in an SVG version? In some contexts it's very blurry.

magnus-madsen:
> Okay. Does it exist in an SVG version? In some contexts it's very blurry.

I thought there was a collection of images? 

But no, no svg

sockmaster27:
For now I'm just using the 32x32

magnus-madsen:
!!!jar

magnus-madsen:
Running the build jar now gives:

```
No `flix.toml'. Will load source files from `*.flix`, `src/**`, and `test/**`.
     __   _   _
    / _| | | (_)            Welcome to Flix 0.39.0
   | |_  | |  _  __  __
   |  _| | | | | \ \/ /     Enter an expression to have it evaluated.
   | |   | | | |  >  <      Type ':help' for more information.
   |_|   |_| |_| /_/\_\     Type ':quit' or press 'ctrl + d' to exit.

java.nio.file.FileSystemNotFoundException
        at jdk.zipfs/jdk.nio.zipfs.ZipFileSystemProvider.getFileSystem(ZipFileSystemProvider.java:156)
        at jdk.zipfs/jdk.nio.zipfs.ZipFileSystemProvider.getPath(ZipFileSystemProvider.java:142)
        at java.base/java.nio.file.Path.of(Path.java:209)
```

magnus-madsen:
I think this class might be useful: https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/util/LocalResource.scala#L33

magnus-madsen:
Here is an example of a use: 

https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/api/Flix.scala#L132

sockmaster27:
Thank you! Istg I've tried like 7 different ways of doing that.

magnus-madsen:
!!!jar

magnus-madsen (approved):

magnus-madsen:
I will merge this and then we can move on to phase 2 🥳 

image

PR #6337, opened 2023-08-14
refactor: introduce `MonoType` in `Simplifier`

### LINKED ISSUE #6242: Introduce `MonoType` in `Simplifier`
... refactor all subsequent ASTs and phases.

The specific translation from `Type` to `MonoType` can be found in `Reducer` (and stolen/moved from there).

Instead of starting from SimplifiedAst, it would also be okay to "work backwards" from `Reducer`.

sockmaster27:
Fixes #6242


main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -45,30 +45,42 @@ object Simplifier {
...
-      case LoweredAst.Expr.Cst(cst, tpe, loc) => SimplifiedAst.Expr.Cst(cst, tpe, loc)
+      case LoweredAst.Expr.Cst(cst, tpe, loc) =>
+        val t = visitType(tpe)
+        SimplifiedAst.Expr.Cst(cst, t, loc)
 
       case LoweredAst.Expr.Lambda(fparam, exp, tpe, loc) =>
         val p = visitFormalParam(fparam)
         val e = visitExp(exp)
-        SimplifiedAst.Expr.Lambda(List(p), e, tpe, loc)
+        visitType(tpe) match {
```
magnus-madsen, line 72: I understand the desire to do this, but we simply don't.

If we were to do it, we would do it consistently (e.g. everywhere for all expressions), but that turns out to be too much work.

Instead we simply use Type everywhere (MonoType) and that has worked out fine.

magnus-madsen: So I request to drop these special cases for Lambda and just use MonoType everywhere in SimplifiedAst.

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -78,60 +90,74 @@ object Simplifier {
...
+            val t = visitType(tpe)
+            SimplifiedAst.Expr.ApplyAtomic(op, es1, t, purity, loc)
 
           case AtomicOp.Spawn =>
-            // Wrap the expression in a closure: () -> tpe \ eff
+            // Wrap the expression in a closure: () -> tpe \ ef
             val List(e1, e2) = es
-            val lambdaTyp = Type.mkArrowWithEffect(Type.Unit, eff, e1.tpe, loc)
-            val fp = SimplifiedAst.FormalParam(Symbol.freshVarSym("_spawn", BoundBy.FormalParam, loc), Ast.Modifiers.Empty, Type.mkUnit(loc), loc)
+            val lambdaTyp = MonoType.Arrow(Nil, e1.tpe)
```
magnus-madsen, line 111: I think here `Nil` should be `List(Unit)`.

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -78,60 +90,74 @@ object Simplifier {
...
-            SimplifiedAst.Expr.ApplyAtomic(AtomicOp.Spawn, List(lambdaExp, e2), tpe, Purity.Impure, loc)
+            val t = visitType(tpe)
+            SimplifiedAst.Expr.ApplyAtomic(AtomicOp.Spawn, List(lambdaExp, e2), t, Purity.Impure, loc)
 
           case AtomicOp.Lazy =>
             // Wrap the expression in a closure: () -> tpe \ Pure
             val e = es.head
-            val lambdaTyp = Type.mkArrowWithEffect(Type.Unit, Type.Pure, e.tpe, loc)
-            val fp = SimplifiedAst.FormalParam(Symbol.freshVarSym("_lazy", BoundBy.FormalParam, loc), Ast.Modifiers.Empty, Type.mkUnit(loc), loc)
+            val lambdaTyp = MonoType.Arrow(Nil, e.tpe)
```
magnus-madsen, line 120: I think here `Nil` should be `List(Unit)`.

magnus-madsen (changes requested):

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
sockmaster27: Then we just need to do it in a later phase

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
sockmaster27: Both pass the tests

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
magnus-madsen: We do? Where do we require Type.Lambda? I think it should just be `Type`?

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
magnus-madsen: Yes, but List(Unit) is more correct.

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
sockmaster27: In `LambdaLift.scala` we need to access `tpe.result`. Would it make more sense to handle it there?

magnus-madsen: Keep it local

magnus-madsen (approved):

Issue #6332, opened 2023-08-13
Add `toXYZ` functions to `Adaptor`


magnus-madsen:
Add the following functions:

```scala
mod Adaptor {
    ///
    /// Returns a comparator for elements of type `a`.
    ///
    /// The comparator is backed by the ordering on `a`.
    ///
    /// Note: Currently requires a Proxy (until Java types support generics).
    ///
    pub def comparator(_: Proxy[a]): ##java.util.Comparator with Order[a] = ???

    ///
    /// Returns a fresh (unmodifiable) Java `Iterator` over the elements of `ma`.
    ///
    pub def toIterator(ma: m[a]): ##java.util.Iterator with Iterable[m] = ???

    ///
    /// Alias for `toArrayList`.
    ///
    pub def toList(ma: m[a]): ##java.util.List with Foldable[m] = checked_cast(toArrayList(ma))

    ///
    /// Returns the elements of the given foldable `ma` as a new Java `ArrayList`.
    ///
    /// Creates a fresh `ArrayList` and copies all elements in `ma` into it.
    ///
    pub def toArrayList(ma: m[a]): ##java.util.ArrayList with Foldable[m] = ???

    ///
    /// Returns the elements of the given foldable `ma` as a new Java `LinkedList`.
    ///
    /// Creates a fresh `LinkedList` and copies all elements in `ma` into it.
    ///
    pub def toLinkedList(ma: m[a]): ##java.util.LinkedList with Foldable[m] = ???

    ///
    /// Alias for `toTreeSet`.
    ///
    pub def toSet(ma: m[a]): ##java.util.Set with Order[a], Foldable[m] = ???

    ///
    /// Returns the elements of the given foldable `ma` as a new `TreeSet`.
    ///
    pub def toTreeSet(ma: m[a]): ##java.util.TreeSet with Order[a], Foldable[m] = ???

    ///
    /// Alias for `toTreeMap`.
    ///
    pub def toMap(m: Map[k, v]): ##java.util.Map with Order[k] = ???

    ///
    /// Returns all key-value pairs of the given map `m` as a new `TreeMap`.
    ///
    /// The `TreeMap` uses a `Comparator` constructed from the `Order` on `k`.
    ///
    pub def toTreeMap(m: Map[k, v]): ##java.util.TreeMap with Order[k] = ???
}
```

The functions probably require the `\ IO` effect.

Issue #6331, opened 2023-08-13
Add `Adaptor` module


magnus-madsen:
- [ ] Add `toXYZ` functions to `Adaptor`
- [ ] Add `fromXYZ` functions to `Adaptor`
- [ ] Add `asXYZ` functions to `Adaptor`

stephentetley:
What are the conversions for? Marshaling between Flix and Java? Changing representation between collections like list and array, vector, etc?

magnus-madsen:
> What are the conversions for? Marshaling between Flix and Java? Changing representation between collections like list and array, vector, etc?

Yes, both marshalling and converting back/forth. 

(Sorry I did not finish writing the APIs yet).

stephentetley:
I've been working on marshaling quite a bit so I'm happy to take a look at when you have an API written.

stephentetley:
I was on holiday today, so I made a PR: 

#6339 - feat: add Adaptor module (issue #6331)

magnus-madsen:
@stephentetley @mlutze Any thoughts on this:

```
class ToJava[t: Type] {
    type Out
    pub def toJava(t: t): ToJava.Out[t] \ IO
}

instance ToJava[List[t]] {
    type Out = ##java.util.List
    pub def toJava(t: List[t]): ##java.util.List \ IO = Adaptor.toList(t)
}
```

We can have it both directions.

I don't think it replaces `Adaptor` because you may need a specific type of linked list, but I can it can make things a bit easier.

stephentetley:
https://github.com/stephentetley/interop-base

At the moment I just have a `GoesToJava` class. `ComesFromJava` is a set of helper functions.

magnus-madsen:
Fixed in master-- soon to be replaced with one that is generics aware.

PR #6339, opened 2023-08-14
feat: add Adaptor module (issue #6331)

stephentetley:
This PR adds a module `Adaptor` for marshaling between Flix and Java.

Currently it just covers collection marshaling as per sub issues #6332 and #6333. 

Iterators - I've actually implemented `fromIterator` rather than `toIterator`, i.e. make a Flix Iterator from a Java one. Java Iterators have a (`hasNext`, `next`) API whereas Flix iterators have just a `next` API (which the return type being Option - None shows the iterator is finished). I think its possible to make a Java iterator from a Flix one with a one element buffer but it would be rather involved so I haven't look at it so far.

Marshaling collection elements (not the collection itself) is currently done with casting this means we can't have numbers and chars in collections (Java overcomes this with object types for prim types `java.lang.Character` for `char`, `java.lang.Integer` for `int` etc.). At some point we might need to use e.g a type class to make marshaling elements more controllable.

Because of this restriction on element types, the tests just test collections with `String` elements at the moment.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/Flix.scala
+++ b/main/src/ca/uwaterloo/flix/api/Flix.scala
@@ -293,6 +293,7 @@ class Flix {
     "Graph.flix" -> LocalResource.get("/src/library/Graph.flix"),
     "Vector.flix" -> LocalResource.get("/src/library/Vector.flix"),
     "Regex.flix" -> LocalResource.get("/src/library/Regex.flix"),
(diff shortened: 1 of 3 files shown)


magnus-madsen (approved):

magnus-madsen:
Excellent! Thank you Stephen.

PR #6345, opened 2023-08-15
chore: add symbol to 'Run' lense

sockmaster27:
Like that of Rust-Analyzer:

image

image


magnus-madsen:
Cool idea. Do we want it for run tests too?

sockmaster27:
I think it looks a bit too noisy if they always appear together

image

magnus-madsen:
I think its fine, I would prefer the consistency.

magnus-madsen (approved):

PR #6350, opened 2023-08-15
feat: add new `Lexer`

herluf-ba:
A new lexer in preparation of the coming resilient parser project. This lexer is also resilient in that it will produce `Err` tokens instead of halting, although it is unclear how the rest of the pipeline should continue if lexing fails. 

The lexer is able to tokenize the Flix standard library without producing error tokens. That means that every sub-string gets a TokenKind other than `Err`. However it's not obvious that no misidentifications happen.

**For Reviewers**
1. There are some design-decisions and some behaviors that are unclear to me. These linger as TODO's in the code. Please answer them if possible
2. Through manual inspection I have tried to convince myself that no misidentifications happen. ie. `)` gets tokenized as `TokenKind.RBrace`.  What's a better way to test this?
3. As with misidentifications I am also fairly certain that line and column positions are precise. How can I better test this?


herluf-ba:
@magnus-madsen

herluf-ba:
Running with `--XParser` and a stub `foo.flix` file produces:
```
ok 		Fixpoint/PredSymsOf.flix
ok 		Object.flix
ok 		CommutativeGroup.flix
ok 		CommutativeSemiGroup.flix
ok 		Fixpoint/Ram/RamStmt.flix
ok 		PartialOrder.flix
ok 		Fixpoint/Interpreter.flix
ok 		Ref.flix
ok 		Int64.flix
ok 		Result.flix
ok 		Int32.flix
ok 		Int8.flix
ok 		MonadZero.flix
ok 		Sub.flix
ok 		Collectable.flix
ok 		Fixpoint/Debugging.flix
ok 		Benchmark.flix
ok 		Witherable.flix
ok 		Concurrent/ReentrantLock.flix
ok 		Reflect.flix
ok 		Console.flix
ok 		Time/Instant.flix
ok 		Fixpoint/Ast/Constraint.flix
ok 		Vector.flix
ok 		Neg.flix
ok 		Fixpoint/SubstitutePredSym.flix
ok 		Add.flix
ok 		BigDecimal.flix
ok 		Float32.flix
ok 		Thread.flix
ok 		MutDeque.flix
ok 		Time/Duration.flix
ok 		Fixpoint/Ast/Denotation.flix
ok 		Mul.flix
ok 		Fixpoint/Ram/RamSym.flix
ok 		Fixpoint/Solver.flix
ok 		Bool.flix
ok 		MonadZip.flix
ok 		Fixpoint/Shared/PredSym.flix
ok 		BigInt.flix
ok 		MultiMap.flix
ok 		Region.flix
ok 		FromString.flix
ok 		Concurrent/Channel.flix
ok 		Down.flix
ok 		Group.flix
ok 		Fixpoint/Ast/BodyPredicate.flix
ok 		Regex.flix
ok 		Nel.flix
ok 		List.flix
ok 		Comparison.flix
ok 		Array.flix
ok 		Fixpoint/Ast/HeadTerm.flix
ok 		Sendable.flix
ok 		Applicative.flix
ok 		LowerBound.flix
ok 		Fixpoint/Options.flix
ok 		Iterator.flix
ok 		MutList.flix
ok 		Fixpoint/Ast/PrecedenceGraph.flix
ok 		Set.flix
ok 		Fixpoint/Compiler.flix
ok 		Boxable.flix
ok 		CodePoint.flix
ok 		Fixpoint/Ram/BoolExp.flix
ok 		Fixpoint/Simplifier.flix
ok 		System.flix
ok 		Channel.flix
ok 		Monad.flix
ok 		Fixpoint/Ram/RowVar.flix
ok 		Concurrent/Condition.flix
ok 		Fixpoint/Stratifier.flix
ok 		Fixpoint/Ram/RelOp.flix
ok 		SemiGroup.flix
ok 		Functor.flix
ok 		Prelude.flix
ok 		Fixpoint/Ast/Polarity.flix
ok 		Traversable.flix
ok 		Eq.flix
ok 		Fixpoint/IndexSelection.flix
ok 		Foldable.flix
ok 		JoinLattice.flix
ok 		DelayMap.flix
ok 		Reducible.flix
ok 		StringBuilder.flix
ok 		Closeable.flix
ok 		Fixpoint/Ast/VarSym.flix
ok 		UnorderedFoldable.flix
ok 		CommutativeMonoid.flix
ok 		foo.flix
ok 		MutMap.flix
ok 		File.flix
ok 		Monoid.flix
ok 		Assert.flix
ok 		Identity.flix
ok 		Char.flix
ok 		Concurrent/CyclicBarrier.flix
ok 		MeetLattice.flix
ok 		Nec.flix
ok 		GetOpt.flix
ok 		Iterable.flix
ok 		Float64.flix
ok 		Fixpoint/Ram/RamTerm.flix
ok 		ToString.flix
ok 		Fixpoint/VarsToIndices.flix
ok 		MutSet.flix
ok 		Option.flix
ok 		Validation.flix
ok 		Fixpoint/Ast/Fixity.flix
ok 		Debug.flix
ok 		String.flix
ok 		Chain.flix
ok 		DelayList.flix
ok 		Order.flix
ok 		RedBlackTree.flix
ok 		Fixpoint/Ast/HeadPredicate.flix
ok 		Int16.flix
ok 		Random.flix
ok 		Fixpoint/Ast/Datalog.flix
ok 		Hash.flix
ok 		Filterable.flix
ok 		Map.flix
ok 		Div.flix
ok 		Boxed.flix
ok 		Fixpoint/Ast/BodyTerm.flix
ok 		Time.flix
ok 		UpperBound.flix
ok 		Environment.flix
ok 		Time/Epoch.flix
ok 		Graph.flix
ok 		Choice.flix
```

magnus-madsen:
Excellent progress!

magnus-madsen:
> Running with --XParser and a stub foo.flix file produces: ...

How about printing the number of tokens produced in each file? Just see that it looks sensible?

magnus-madsen:
In addition to outputting tokens, you could count that there is an equal number of `(` and `)` and so forth.

magnus-madsen:
Even better, you could print:

```
file.flix numberOfTokens numLeftBraces numRightBraces numLeftParen numRightParen numDefKeyword numClassKeyword
```

you get the idea :)

magnus-madsen:
I will give feedback in three rounds:

1. I will try to answer your TODOs/questions.
2. Then I will suggest how we can break this PR into a few smaller PRs
3. Then I will keep some feedback for file (which we can pick up in the follow-up PRs).

magnus-madsen:
!!!jar

main/src/ca/uwaterloo/flix/language/ast/TokenKind.scala
```diff
@@ -17,16 +17,332 @@ package ca.uwaterloo.flix.language.ast
...
 sealed trait TokenKind
 
-object TokenKind {
+// NOTE: Tokens are named for 'what they are' rather than 'what they represent'.
+// So '::' is not named 'Cons' but instead 'ColonColon' as the lexer should be oblivious to the concept of cons
+
+// NOTE: Builtin type keywords like 'Float32' are lexed into 'Float32Keyword' whereas Float32 *literals* are lexed to 'Float32'
+// This aligns naming of literals with other types such as 'String' and 'Bool'
+
+// TODO: double ## signifies something to do with java?
```
magnus-madsen, line 26: Yes, all Java names start with `##`. For example `##java.lang.String`.

main/src/ca/uwaterloo/flix/language/ast/TokenKind.scala
```diff
@@ -17,16 +17,332 @@ package ca.uwaterloo.flix.language.ast
...
+
+  case object BigInt extends TokenKind
+
+  case object BigDecimal extends TokenKind
+
+  case object String extends TokenKind
+
+  case object Char extends TokenKind
+
+  case object Decorator extends TokenKind // TODO: this refers to stuff like '@Internal'. Is this the right name for it?
```
magnus-madsen, line 313: Annotation.

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -17,12 +17,14 @@ package ca.uwaterloo.flix.language.phase
...
 import ca.uwaterloo.flix.language.CompilationMessage
-import ca.uwaterloo.flix.language.ast.{Ast, ReadAst, Token}
+import ca.uwaterloo.flix.language.ast.{Ast, LexerErr, ReadAst, Token, TokenKind}
 import ca.uwaterloo.flix.util.{ParOps, Validation}
 import ca.uwaterloo.flix.util.Validation._
-
 import scala.collection.mutable
 
+
+// TODO: internals? something to handle $DEFAULT$
```
magnus-madsen, line 26: We have built-in "functions". They are always written `$FOO$`.

magnus-madsen: Random example: https://github.com/flix/flix/blob/master/main/src/library/Int32.flix#L307

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -38,31 +40,505 @@ object Lexer {
...
+          }
+        }
+      }
+    }
+
+    TokenKind.Err(LexerErr.MalformedNumber)
+  }
+
+  // Advances state past a decorator by looking for the next non-letter character
+  // TODO: can decorator names have numbers, underscores, bang in them?
```
magnus-madsen, line 469: I think we should just allow a-zA-Z

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -38,31 +40,505 @@ object Lexer {
...
+  }
+
+  // Advances state past a block-comment. Supports nested block comments by maintaining a level counter.
+  private def blockComment()(implicit s: State): TokenKind = {
+    var l = 1
+    while (!isAtEnd()) {
+      (peek(), peekpeek()) match {
+        case ('/', Some('*')) => {
+          l += 1
+          if (l >= 256) { // TODO: Should this nesting limit be removed?
```
magnus-madsen, line 502: I think its reasonable. Probably can lower it to 32 or even less.

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -38,31 +40,505 @@ object Lexer {
...
+          advance()
+          if (l == 0) {
+            return TokenKind.BlockComment
+          }
+        }
+        case _ => advance()
+      }
+    }
+
+    TokenKind.BlockComment // TODO: Is an unclosed block comment at eof an error?
```
magnus-madsen, line 519: Yes

magnus-madsen:
In terms of merging this, I would like to split it into a few PRs:

- PR1: The minor change to `Token.scala` (essentially a one-liner, but might as well also add yourself to `AUTHORS.md` at the same time).
- PR2: Extract `LexerErr` from `TokenKind` and place it into `language/errors/LexerError.scala`. Refactor it to implement `CompilationMessage` and comment it like you see in e.g. `WeederError`. Try to make good errors and paste some screenshots.
- PR3: Update `TokenKind.scala` with all the new tokens.

(The order of the above is not important).

Once merged, we can rebase this PR onto master, and then we can move on to `Lexer.scala`.

main/src/ca/uwaterloo/flix/language/ast/Token.scala
```diff
@@ -15,4 +15,4 @@
  */
 package ca.uwaterloo.flix.language.ast
 
-case class Token(kind: TokenKind, text: String)
+case class Token(kind: TokenKind, text: String, line: Int, column: Int)
```
magnus-madsen, line 18: column -> col

main/src/ca/uwaterloo/flix/language/ast/TokenKind.scala
```diff
@@ -17,16 +17,332 @@ package ca.uwaterloo.flix.language.ast
...
+  case object BlockComment extends TokenKind
+
+  case class Err(kind: LexerErr) extends TokenKind
 
   case object Eof extends TokenKind
 
-  case object Err extends TokenKind
+}
+
+sealed trait LexerErr
```
magnus-madsen, line 331: Rework this to look like e.g. `WeederError` or `TypeError`.

Comment each error in detail, ideally with an example.

main/src/ca/uwaterloo/flix/language/ast/TokenKind.scala
```diff
@@ -17,16 +17,332 @@ package ca.uwaterloo.flix.language.ast
...
+
+object LexerErr {
+
+  case object UnexpectedSymbol extends LexerErr
```
magnus-madsen, line 335: Symbol has a specific meaning in Flix. Perhaps UnexpectedChar?

magnus-madsen:
A quick performance experiment:

```
  MonomorphEnums                  92,8 ms (02,6%)
  Weeder                         103,1 ms (02,8%)
  Namer                          104,2 ms (02,9%)
->Parser                       ->171,9 ms (04,7%)
  Optimizer                      254,9 ms (07,0%)
  Resolver                       256,5 ms (07,1%)
->Lexer                        ->284,5 ms (07,9%)
  Monomorph                      412,3 ms (11,4%)
  JvmBackend                     495,2 ms (13,7%)
  Typer                          781,9 ms (21,6%)
```

Not bad for first draft. We can probably recover that performance with a few tweaks 👍

main/src/ca/uwaterloo/flix/language/ast/TokenKind.scala
magnus-madsen: OK, I realize that you probably need to keep these, because you use them in TokenKind.Err-- but we also need to connect them to some LexerError in the compiler.

herluf-ba:
> Even better, you could print:
> 
> ```
> file.flix numberOfTokens numLeftBraces numRightBraces numLeftParen numRightParen numDefKeyword numClassKeyword
> ```
> 
> you get the idea :)

I implemented this and got the following results:
```
                          filename    (    )   /%   //    [    ]class  def  err    {    }
                       Object.flix    8    5    1    6    1    1    0    2    0    3    3
          Fixpoint/PredSymsOf.flix    1    1    1    3    2    2    1    1    0    2    2
             CommutativeGroup.flix    9    9    1    6   75   75    1    0    0    1    1
         Fixpoint/Ram/RamStmt.flix   19   19    1    0   15   15    0    1    0    5    5
                      Prelude.flix   73   71    0   99    4    4    0   19    0    8    8
        Fixpoint/Ast/Polarity.flix    0    0    1    0    0    0    0    0    0    2    2
         CommutativeSemiGroup.flix   21   21    1    9   77   77    1    2    0    1    1
                  Traversable.flix   22   22    1   31   34   34    1    4    0    4    4
         Fixpoint/Interpreter.flix  238  238    1    2   45   45    0    9    0   25   25
                        Int32.flix  225  198    1  225    6    6    0   55    0   53   53
                         Int8.flix  199  181    1  200    3    3    0   50    0   36   36
                    MonadZero.flix    1    1    1    6    3    3    1    1    0    1    1
                          Sub.flix   17   17    1    6    9    9    1    9    0   11   11
                  Collectable.flix    1    1    1    6    4    4    1    1    0    2    2
                 PartialOrder.flix  145  145    1   13   69   69    1   15    0   26   26
           Fixpoint/Debugging.flix   39   39    1    9    7    7    0    3    0    7    7
                          Ref.flix    6    6    1   14    5    5    0    4    0    2    2
                       Result.flix  240  240    1  182  156  156    0   55    0   47   47
                        Int64.flix  243  228    1  233    7    7    0   56    0   57   57
                  JoinLattice.flix  166  166    1   14   70   70    1   15    0   25   25
                    Benchmark.flix   89   86    1   63    5    5    0   18    0   18   18
                   Witherable.flix    5    5    1   11   15   15    1    2    0    2    2
                      Float32.flix  174  174    1  162    9    9    0   35    0   41   41
      Fixpoint/Ast/Denotation.flix   18   18    1   22    9    9    0    3    0    4    4
                          Mul.flix   17   17    1    6    9    9    1    9    0   11   11
     Concurrent/ReentrantLock.flix   41   39    1   29    3    3    0    8    0   10   10
            CommutativeMonoid.flix   27   27    1    9   78   78    1    2    0    1    1
                          foo.flix    0    0    0    5    0    0    0    0    0    0    0
          Fixpoint/Ram/RamSym.flix   52   52    1    0   27   27    0    7    0   17   17
                      Reflect.flix   15   15    0    5   12   12    0    2    0    2    2
                         Bool.flix   18   18    1   30    4    4    0   12    0    3    3
                 Time/Instant.flix   29   26    1   21    1    1    0    7    0    8    8
                      Console.flix  251  218    1  211   41   40    0   54    0   24   24
      Fixpoint/Ast/Constraint.flix   16   15    1    0   16   16    0    3    0    8    8
                     Identity.flix   87   87    1    3   83   83    0   20    0   16   16
     Concurrent/CyclicBarrier.flix    9    8    1   10    0    0    0    2    0    2    2
           Concurrent/Channel.flix  141  139    1  247   18   18    0   17    0   16   16
         Fixpoint/Ram/BoolExp.flix   21   21    1    0   27   27    0    1    0    4    4
                         Down.flix   43   43    1    4   46   46    0    9    0   17   17
                        Monad.flix   37   37    1   54   49   49    1    8    0    6    6
                     MutDeque.flix  238  230    1  172   60   60    0   43    0   45   44
   Fixpoint/Ast/BodyPredicate.flix   36   36    1    0   21   21    0    5    0   11   11
          Fixpoint/Stratifier.flix   66   65    1   23    8    8    0    5    0   15   15
           Fixpoint/Ram/RelOp.flix   19   19    1    0   22   22    0    1    0    9    9
              Fixpoint/Solver.flix  536  535    1  215  362  362    0   42    0   46   46
        Fixpoint/Ast/HeadTerm.flix   17   17    1    0    4    4    0    1    0    4    4
                     MonadZip.flix   15   15    1   21   17   17    1    4    0    1    1
      Fixpoint/Shared/PredSym.flix    4    4    1    0    1    1    0    1    0    4    4
                         List.flix  799  789    1  559  404  404    0  195    0  123  123
                   Comparison.flix    4    2    1   14    0    0    0    2    0    3    3
                     MultiMap.flix  389  388    1  227  198  198    0   68    0   25   24
                     Sendable.flix   14   14    1    3  146  146    1    0    0    0    0
                   FromString.flix   30   29    1    6   26   26    1   13    0   15   15
                          Nel.flix  439  436    1  263  251  251    0  111    0   85   85
                       BigInt.flix  155  155    1  168    7    7    0   37    0   55   55
                       Region.flix    4    4    1    8    2    2    0    1    0    2    2
                        Group.flix  142  142    1   33   76   76    1   30    0   28   28
                     Foldable.flix  226  226    1  192  106  106    1   49    0   10   10
                      Boxable.flix  504  503    1   54  320  320    0   37    0   37   37
                     Iterator.flix  386  381    1  313  115  115    0   64    0   92   92
                StringBuilder.flix   57   55    1   48   27   27    0   15    0   10   10
                    Closeable.flix    1    1    1    3    1    1    1    1    0    1    1
          Fixpoint/Ast/VarSym.flix    1    1    1    0    0    0    0    0    0    2    2
                      MutList.flix  478  434    1  324  144  144    0   96    0   85   85
         Concurrent/Condition.flix   21   20    1   18    2    2    0    4    0    6    6
                        Array.flix  937  841    1  564  398  398    0  170    0  241  241
 Fixpoint/Ast/PrecedenceGraph.flix   13   13    1    0    3    3    0    2    0    5    5
                         File.flix  671  643    1  235   93   92    2   48    0  190  190
            Fixpoint/Compiler.flix  300  298    1  101   32   32    0    9    0   48   48
                       Assert.flix    7    7    1    5    2    2    0    1    0    4    4
             Fixpoint/Options.flix    1    1    1    3    0    0    0    1    0    1    1
                       System.flix   51   49    1   51    2    2    0   11    0    7    7
                           Eq.flix  149  148    1   21  146  146    1   38    0   33   33
                       GetOpt.flix  294  293    1  175   95   95    0   26    0   43   43
                   LowerBound.flix   93   93    1    6   65   65    1   11    0   11   11
                       String.flix  675  620    1  531   44   44    0  133    0  166  166
                     DelayMap.flix  570  566    1  314  195  195    0   87    0   56   55
                    Reducible.flix  174  174    1  137   64   64    1   41    0    6    6
            UnorderedFoldable.flix   23   23    1   34   12   12    1    8    0    1    1
                       MutSet.flix  173  172    1  202  105  105    0   48    0   32   31
                       MutMap.flix  286  285    1  317  152  152    0   72    0   52   51
                  MeetLattice.flix  166  166    1   16   70   70    1   15    0   25   25
                       Monoid.flix  117  117    1   13   82   82    1   22    0   21   21
                         Char.flix   76   76    1  109    5    5    0   29    0   32   32
                     ToString.flix   90   90    1    6  147  147    1   28    0   53   53
                      Float64.flix  172  172    1  177   10   10    0   37    0   39   39
                       Option.flix  310  310    1  194  263  263    0   83    0   72   72
                        Order.flix  410  410    1  108  148  148    1   72    0   50   50
         Fixpoint/Ram/RamTerm.flix   21   21    1    9   21   21    0    1    0    4    4
                          Nec.flix  901  893    1  365  355  355    0  157    0  123  123
                        Int16.flix  215  197    1  211    4    4    0   52    0   42   42
                   Validation.flix  186  186    1  119  141  141    0   38    0   26   26
                  Applicative.flix   87   87    1  109  106  106    1   16    0    2    2
         Fixpoint/Ast/Datalog.flix   63   63    1    0   23   23    0    4    0   15   15
                       Vector.flix  885  813    1  508  386  386    0  179    0  118  118
       Fixpoint/VarsToIndices.flix  123  119    1    3   11   11    0    4    0    7    7
          Fixpoint/Ast/Fixity.flix    0    0    1    0    0    0    0    0    0    2    2
                          Neg.flix   19   19    1    6    9    9    1    9    0   11   11
   Fixpoint/SubstitutePredSym.flix    1    1    1    3    2    2    1    1    0    2    2
                        Chain.flix  628  614    1  352  278  278    0  138    0   96   96
                Time/Duration.flix   12   12    1   12    0    0    0    4    0    1    1
                        Debug.flix   80   75    0   12   35   35    0    4    0   35   35
                     Iterable.flix    4    4    1    9    7    7    1    2    0    1    1
                   BigDecimal.flix   85   82    1  111    7    7    0   22    0   24   24
                          Add.flix   20   20    1    6   10   10    1   10    0   13   13
                       Thread.flix   10   10    1    3    0    0    0    1    0    3    3
                         Hash.flix  126  126    1   13   76   76    1   24    0   44   44
                      Channel.flix   19   17    1   27   12   12    0    6    0    3    3
          Fixpoint/Ram/RowVar.flix   13   13    1    0    2    2    0    2    0    6    6
                   Filterable.flix    6    6    1    9    7    7    1    2    0    1    1
                          Set.flix  408  407    1  240  204  204    0   84    0   35   35
                    SemiGroup.flix  185  183    1   12   82   82    1   26    0   36   36
                      Functor.flix   19   19    1   33   28   28    1    7    0    2    2
                    CodePoint.flix  110  102    1  142   11   11    0   33    0   42   42
                        Regex.flix  327  323    1  269   54   54    0   55    0  152  152
          Fixpoint/Simplifier.flix  125  125    1   27   14   14    0    4    0   21   21
      Fixpoint/IndexSelection.flix  180  180    1   56   17   17    0    5    0   16   16
                 RedBlackTree.flix  730  709    1  370  184  184    0   90    0   85   85
   Fixpoint/Ast/HeadPredicate.flix   17   17    1    0   16   16    0    3    0    9    9
                    DelayList.flix 1007  993    1  521  270  270    0  137    0  105  104
                          Map.flix  619  618    1  404  347  347    0  125    0   45   45
                       Random.flix   44   42    0   35    2    2    0   10    0    4    4
                          Div.flix   26   26    1    6   10   10    1   10    0   13   13
                        Boxed.flix   91   90    1    3    3    3    0    3    0    7    7
        Fixpoint/Ast/BodyTerm.flix    7    7    1    0    4    4    0    1    0    4    4
                         Time.flix    1    1    1    3    0    0    0    0    0    1    1
                   UpperBound.flix   93   93    1    6   65   65    1   11    0   11   11
                  Environment.flix   60   59    1   49   14   14    0   16    0   21   21
                   Time/Epoch.flix    9    9    1   12    0    0    0    4    0    1    1
                        Graph.flix  392  378    1  191  102  102    0   33    0   80   81
                       Choice.flix   35   35    2   42   28   28    0   16    0   15   15
```

Some things to note:
- There are no error tokens a before
- The amount of def, class, line comments and block comments seem sensible (to me atleast)
- The amount of '[' matches ']'. Same with '{' and '}'.
- There is a mismatched amount of '(' and ')' in many files. I will investigate this now!

magnus-madsen:
What's your thought on the discrepancy between ( and )? The rest looks reasonable right?

herluf-ba:
> What's your thought on the discrepancy between ( and )? The rest looks reasonable right?

I figured out that it was due to two errors:
- Java names (starting with `##`) was eating up a lot more chars than it should
- The `number` function was consuming an extra char whenever there was no number type at the end of a literal. ie. `(1)` would become `(` and `1)`. 
I fixed both of these in my most recent commit, and now the discrepancy is gone 👍

magnus-madsen:
!!!jar

JonathanStarup:
Maybe i couldn't find it, but isnt this PR lacking handling of `"hello ${42} world"`

herluf-ba:
> Maybe i couldn't find it, but isnt this PR lacking handling of `"hello ${42} world"`.
> 
> strings like `"hello ${" // "} world"` and `"hello ${ 23 + {32} } world"` are especially fun

It is missing that! Thanks for pointing that out 🙂  I guess it would be nice to have a `TokenKind.FormatString` for the rest of the pipeline to know if its a plain string or something that needs further evaluation.

JonathanStarup:
That sounds good yeah. There is an issue that the lexer needs more complexity than constant lookahead in order to know where the interpolation part stops.

magnus-madsen:
> > Maybe i couldn't find it, but isnt this PR lacking handling of `"hello ${42} world"`.
> > strings like `"hello ${" // "} world"` and `"hello ${ 23 + {32} } world"` are especially fun
> 
> It is missing that! Thanks for pointing that out 🙂 I guess it would be nice to have a `TokenKind.FormatString` for the rest of the pipeline to know if its a plain string or something that needs further evaluation.

I actually suggest not dealing with this until we get to the parser.

JonathanStarup:
That makes sense, practically. I just wanted to point out that the lexer cannot be generally correct with constant lookahead. I tried to make a flix lexer in flix and got stuck at this point. I think it can be solved with lexer "modes" or something like that

magnus-madsen:
> 

The lexer has infinite lookahead. This is necessary just to parse a number, string, or comment :)

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -44,31 +50,575 @@ object Lexer {
...
+
+  // Checks whether the following substring matches a keyword. Note that *comparison includes current*
+  private def keyword(k: String)(implicit s: State): Boolean = {
+    // check if the keyword can appear before eof
+    if (s.current.offset + k.length > s.src.data.length) {
+      return false
+    }
+
+    val start = s.current.offset - 1
+    val matches = s.src.data.slice(start, start + k.length).sameElements(k.toCharArray)
```
magnus-madsen, line 313: This looks slow and on a hot path. Assume that keywords can only be ASCII (not emojis) then just write an ordinary loop. Thus avoiding costly slice, toCharArray, and sameElements.

herluf-ba: I changed this to a while loop to avoid allocations. Now the lexer phase runs at roughly 1/4 of the current parser!
```
  Lexer                          114,3 ms (01,8%)
  Parser                         410,8 ms (06,4%)
```

magnus-madsen:
A bit of early feedback:

- Every def/trait/etc should have a block comment. Comments should be entire sentences ending with a period.
- classes should either be at the very top or the very bottom depending on how important they are.
- private constants should be at the top and start with an uppercase letter.
- Ranges should of the form `if (x <= y && y <= z)`.
- Functions like `retreat` and `number` share the same form, but are not conceptually the same. Maybe it would be better to have `acceptNumber`?
- magic constants should have names (e.g. 32)

main/src/ca/uwaterloo/flix/language/ast/TokenKind.scala
```diff
@@ -15,14 +15,13 @@
  */
 package ca.uwaterloo.flix.language.ast
 
-sealed trait TokenKind
-
 /* Tokens are named for 'what they are' rather than 'what they represent'.
```
magnus-madsen, line 18: I just noticed this is not a block comment.

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -33,6 +32,13 @@ object Lexer {
     }
 
     flix.phase("Lexer") {
+
+      // TODO: Remove this debug printing
+      //      val state = new State(root.sources.head._1)
+      //      val stats = tokenStats()(state)
+      //      val s = stats.toSeq.sortBy(_._1).map(k => "%5s".format(k._1)).mkString("")
+      //      println(f"${"%34s".format("filename")}${s}")
```
magnus-madsen: Obviously all the debugging goes away, but you know that :)

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -43,29 +49,559 @@ object Lexer {
...
+    // Add a virtual eof token at the last position
+    s.tokens += Token(TokenKind.Eof, "<eof>", s.current.line, s.current.column)
+
+    // TODO: Remove this debug printing
+    //    val stats = tokenStats()
+    //    val debug = stats.toSeq.sortBy(_._1).map(v => "%5d".format(v._2)).mkString("")
+    //    println(f"${"%34s".format(src.name)}${debug}")
+
+
+    val hasErrors = s.tokens.exists(t => t.kind.isInstanceOf[TokenKind.Err])
```
magnus-madsen, line 58: hasErrors -> errorTokens?

magnus-madsen: Then you can also just have one collect.

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -43,29 +49,559 @@ object Lexer {
...
+      Validation.SoftFailure(s.tokens.toArray, LazyList.from(s.tokens).collect {
+        case Token(TokenKind.Err(e), t, l, c) => tokenErrToCompilationMessage(e, t, l, c)
+      })
+    } else {
+      s.tokens.toArray.toSuccess
+    }
+  }
+
+  /**
+   * Advances state one char forward returning the char it was previously sitting on.
```
magnus-madsen, line 70: instead of state maybe we should say "current position" everywhere?

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -43,29 +49,559 @@ object Lexer {
...
+      case '`' => infixFunction()
+      case _ => TokenKind.Err(TokenErrorKind.UnexpectedChar)
+    }
+
+    addToken(kind)
+  }
+
+  // Adds a token by consuming the characters between start and current
+  private def addToken(k: TokenKind)(implicit s: State): Unit = {
+    val t = s.src.data.slice(s.start.offset, s.current.offset).mkString("")
```
magnus-madsen, line 287: This means that we create a new string for every token. This is probably a bit expensive. For example, we will now have X objects representing "def" floating around. Ideas for solutions?

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -43,29 +49,559 @@ object Lexer {
...
+    s.tokens += Token(k, t, s.start.line, s.start.column)
+    s.start = new Position(s.current.line, s.current.column, s.current.offset)
+  }
+
+  // Checks whether the following substring matches a keyword. Note that *comparison includes current*
+  private def keyword(k: String)(implicit s: State): Boolean = {
```
magnus-madsen, line 295: I think this function could make use of some small helper functions.

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -43,29 +49,559 @@ object Lexer {
...
+    // check if the keyword can appear before eof
+    if (s.current.offset + k.length > s.src.data.length) {
+      return false
+    }
+
+    val start = s.current.offset - 1
+    var matches = true
+    var o = 0
```
magnus-madsen, line 303: you mostly use long variable names, but here you use `o`.

magnus-madsen:
!!!jar

magnus-madsen:
When you feel the lexer is shaping up, you should remove this check:

```
    if (!flix.options.xparser) {
      // New lexer and parser disabled. Return immediately.
      return Map.empty[Ast.Source, Array[Token]].toSuccess
    }
```

so the lexer is always run.

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
herluf-ba: A very simplistic solution seems to be to just make `Token.text` optional. Then we could retain the text only where it is not derivable from `TokenKind`. What I don't like about that solution is that we don't encode what `TokenKind`s one should expect text on in the type system, so the programmer should "just know".

One way tackle that would be to make `Token.text` private optional and then define a `Token.get_text()` that matches on its own kind to produce its text. Then the details would be more neatly hidden.

What do you think?

magnus-madsen: What about just storing the offset and a pointer to the entire string? Then we can compute the string by demand.

magnus-madsen:
Will you enable the lexer by default?

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
herluf-ba: Yep, I will try and do it like that

herluf-ba:
I went ahead and removed the `--xparser` option too, since it is unused now

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
herluf-ba: I'm not sure I follow. Which parts would you extract in to helper functions?

magnus-madsen:
@herluf-ba Good, and now you see that some tests fail :)

herluf-ba:
That might also be because I found out that there is a format string like this one `"#{"${expr}"}"` in the standard library. It's in `Fixpoint/Ram/RamStmt.flix`

Before the lexer was just failing silently to tokenize that properly, but now its at least producing an error 👍

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -21,51 +21,670 @@ import ca.uwaterloo.flix.language.ast.{Ast, TokenErrorKind, ReadAst, SourceKind,
...
+ * the lexer will simply produce a token of kind `TokenKind.Err` an move on instead of halting.
+ * There are some unrecoverable errors though, for example unterminated block-comments or unclosed string literals.
+ * In these cases a `TokenKind.Err` will still be produced but it will contain the rest of the source text.
+ * See `LexerError` for all error states.
+ */
 object Lexer {
+  /**
+   * The maximal allowed nesting level of block-comments.
+   */
+  private val MAX_BLOCK_COMMENT_NESTING_LEVEL = 32
```
magnus-madsen, line 40: The naming convention is actually `MaxBlockCommentNestingLevel`.

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
magnus-madsen: Perhaps the while loop which computes matches?

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -21,51 +21,670 @@ import ca.uwaterloo.flix.language.ast.{Ast, TokenErrorKind, ReadAst, SourceKind,
...
+      advance()
+    }
+  }
+
+  /**
+   * Moves current position past a name (both upper- and lower-case).
+   * There are edge cases of variable holes, IE. "x?", and java names like "Map$Entry",
+   * which is the reason this function will return a `TokenKind`.
+   */
+  private def name(isUpper: Boolean)(implicit s: State): TokenKind = {
```
magnus-madsen, line 404: I would call these `lexName` or `acceptName` because they return TokenKind and to distinguish them from `peek` which is a lower-level primitive.

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -21,51 +21,670 @@ import ca.uwaterloo.flix.language.ast.{Ast, TokenErrorKind, ReadAst, SourceKind,
...
+    }
+    TokenKind.NameGreek
+  }
+
+  /**
+   * Checks whether `c` lies in unicode range U+0370 to U+03FF
+   */
+  private def isGreekNameChar(c: Char): Boolean = {
+    val i = c.toInt
+    i >= 880 && i <= 1023
```
magnus-madsen, line 462: I would prefer the style `880 <= i && i <= 1023`.

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -21,51 +21,670 @@ import ca.uwaterloo.flix.language.ast.{Ast, TokenErrorKind, ReadAst, SourceKind,
...
+    }
+    TokenKind.NameMath
+  }
+
+  /**
+   * Checks whether `c` lies in unicode range U+2190 to U+22FF
+   */
+  private def isMathNameChar(c: Char): Boolean = {
+    val i = c.toInt
+    i >= 8592 && i <= 8959
```
magnus-madsen, line 485: Ditto

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
magnus-madsen: Any reason not to use hex decimal literals?

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -21,51 +21,670 @@ import ca.uwaterloo.flix.language.ast.{Ast, TokenErrorKind, ReadAst, SourceKind,
...
+      }
+    }
+    TokenKind.UserDefinedOperator
   }
 
-  private def advance()(implicit s: State): Char = ???
+  /**
+   * The characters allowed in a user defined operator mapped to their `TokenKind`s
+   */
+  private val validUserOpTokens = Map(
```
magnus-madsen, line 535: This should be next to MAX_BLOCK_COMMENT_NESTING_LEVEL since it is a "static field".

magnus-madsen:
> That might also be because I found out that there is a format string like this one `"${"${expr}"}"` in the standard library. It's in `Fixpoint/Ram/RamStmt.flix`
> 
> Before the lexer was just failing silently to tokenize that properly, but now its at least producing an error 👍

Yes, its called string interpolation internally.

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
herluf-ba: Only that I hadn't thought about it! Will do :)

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
herluf-ba: I have done so but honestly I feel like this reads worse than before. Did I misunderstand you, or do our opinions differ here? 😅 

```scala
  /**
   * Checks whether the following substring matches a keyword. Note that __comparison includes current__.
   */
  private def isMatch(keyword: String)(implicit s: State): Boolean = {
    // Check if the keyword can appear before eof.
    if (s.current.offset + keyword.length > s.src.data.length) {
      return false
    }

    // Check if the next n characters in source matches those of keyword one at a time.
    val start = s.current.offset - 1
    var matches = true
    var offset = 0
    while (matches && offset < keyword.length) {
      if (s.src.data(start + offset) != keyword(offset)) {
        matches = false
      } else {
        offset += 1
      }
    }
    matches
  }

  /**
   * Checks whether the following substring matches a keyword. Note that __comparison includes current__.
   * Also note that this will advance the current position past the keyword if there is a match
   */
  private def isKeyword(keyword: String)(implicit s: State): Boolean = {
    val matches = isMatch(keyword)
    if (matches) {
      for (_ <- 1 until k.length) {
        advance()
      }
    }
    matches
  }
```

magnus-madsen: Hmm okay. I leave it to you then.

main/src/ca/uwaterloo/flix/language/ast/Token.scala
```diff
@@ -15,4 +15,9 @@
  */
 package ca.uwaterloo.flix.language.ast
 
-case class Token(kind: TokenKind, text: String, line: Int, col: Int)
+case class Token(kind: TokenKind, src: Array[Char], startOffset: Int, endOffset: Int, line: Int, col: Int) {
+  /**
+   * Computes the lexeme that the token refers to by slicing it from `src`.
+   */
+  def text(): String = src.slice(startOffset, endOffset).mkString("")
```
magnus-madsen, line 22: Perhaps this should be a lazy val and instead of slice and mkstring we can use substring?

herluf-ba: I made it into a `lazy val`. Didn't know about those, that's neat! 

`src` is an `Array[Char]` so I think we need to do slice + mkString.

magnus-madsen:
Looks better-- but still some issues apparently.

magnus-madsen:
!!!jar

herluf-ba:
Yea, I am working on getting all tests to pass 👍 

Currently we are failing 40 tests due to multiple issues:
- Missing unicode handling in `acceptChar`. Fails at `'☃'`
- There's bugs with peek and retreat usage causing OOB-errors
- Escaped characters in strings together with string interpolation is tricky. `"\${a}"` is just a regular string literal!
- I think $ needs to be thought off as a separator. It seems to be used like that when importing java objects.

magnus-madsen:
What's the status on this?

herluf-ba:
!!jar

herluf-ba:
!!!jar

herluf-ba:
@magnus-madsen A bit slow progress in this, but now it seems to finally be working. We are now able to lex the standard library, all tests and all community builds in the github checks pipeline, so I think this is ready for a 'final lookover' 👍

magnus-madsen:
!!!jar

magnus-madsen (changes requested):
A few things:

- I would like to restore --Xparser, but with the opposite semantics. If the flag is set then the new lexer/parser is disabled. (This means that Flix will run with the new lexer by default, but if there is a bug, users can disable it with --Xparser as a temporary measure.)
- I would like a separate PR for the updates to Token.scala, TokenErrorKind, TokenKind.
- I would also like a separate PR for the updates to LexerError.scala.

This should not be too hard, just copy paste the files into a new branch and do a new PR.

Once those simpler PRs are in then hopefully we can get the main Lexer PR in quickly. So we can move on.

magnus-madsen:
@herluf-ba This can now be merged/rebased s.t. changes are mostly isolated to Lexer.scala

magnus-madsen:
> @magnus-madsen Now there is only changes in `Lexer.scala` and some minor changes related to `--Xparser`

You reformatted the code :( (or maybe IDEA did), but it breaks diffing anyway. We are trying to keep a clean history.

herluf-ba:
Yep that is IDEA doing that. I will be careful with auto-formatting from now on. 
There is still something with the end of `Options.scala` and `Main.scala` but I can't figure out what it is exactly. Do you know?

magnus-madsen:
> Yep that is IDEA doing that. I will be careful with auto-formatting from now on. There is still something with the end of `Options.scala` and `Main.scala` but I can't figure out what it is exactly. Do you know?

You could restore the previous file and just make your changes?

magnus-madsen:
@stephentetley This PR enables a new Lexer. It _should_ not cause any new parsing issues. But if you start to see such issues, let us know :)

magnus-madsen (approved):

stephentetley:
Thanks for the heads up Magnus, will do.

PR #6351, opened 2023-08-15
chore: enable `Lexer` in `Flix.scala`

magnus-madsen:


magnus-madsen:
@mlutze Can you approve? This is just a temporary fix because the current setup does not allow correct performance benchmarking.

mlutze (approved):

Issue #6327, opened 2023-08-13
Add HtmlDocumentor (v2)


magnus-madsen:
- [x] Milestone 1: #6309 is to replicate everything from api.flix.dev as static html pages. 
- [x] Milestone 2: Add a menu on the left. Unlike the current menu, it should not show the entire project, but instead it should just be a summary (bullet point list) of the defs (etc) in the current module. Thus it would mirror the structure of Rustdoc which also just shows "children" on the left.

Milestone 3:
- [x] Don't show type parameters on defs.
- [x] Don't show `\ Pure`.
- [x] Annotations on one line. (Placement of source button is also not aligned with them?)
- [x] Add Flix version somewhere.
- [x] Add a link to the top of the menu which allows navigation back to Prelude.html (e.g. could be "Flix").
- [x] Show "Signatures" by default (if possible).
- [x] I think using a monospaced font for the menu is a bit too aggressive. Perhaps we could try 'Roboto' or something?
- [x] Emit HTML before resources (so if a crash happens due to broken JAR at least the HTML is there.)

Milestone 4:
- [x] Output `Prelude` as the filename `index.html`
- [x] Effects should come right after classes (and then enums and type aliases) both on the page and in the menu.
- [x] A "<- back" link after "Flix" in the menu might be useful when in a sub module.
- [x] A "link" button next to source would be useful, e.g. this would point to e.g. `Adaptor.html#def-toLinkedList` and allow one to link directly to an item.
- [x] Rename "ops" to `Operations` and show by default.

Milestone 5: 
- [x] Merge classes/effects/enums with their module (i.e. List.html should show the definition of list, Foldable.html should show the Foldable class and so forth).

sockmaster27:
> * Don't show type parameters.

Does this only apply to definitions (and signatures) or classes and enums too?

magnus-madsen:
Lets only go with definitions:

![image](https://github.com/flix/flix/assets/2287953/cf2744da-82a1-4ed2-b60f-7ec4b04318fe)

magnus-madsen:
(I am updating the lists a bit)

magnus-madsen:
@sockmaster27 I updated the list. Still looking for more smaller improvements.

magnus-madsen:
@sockmaster27 Let me know if you need help understanding milestone 4. Its a bit difficult to understand, but the idea is that instead of showing e.g. "List" enum has in Prelude, we want to show it in "List.html" above everything else. 

[...]

PR #6355, opened 2023-08-16
feat: improve HTML documentation

sockmaster27:
Milestone 2

Related to #6327

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/HtmlDocumentor.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/HtmlDocumentor.scala
@@ -74,12 +74,13 @@ object HtmlDocumentor {
     }
   }
 
+  private def moduleName(mod: Module): String = if (mod.sym.isRoot) RootNS else mod.sym.toString
+
   /**
     * Splits the modules present in the root into a set of `HtmlDocumentor.Module`s, making them easier to work with.
     */
   private def splitModules(root: TypedAst.Root): Map[Symbol.ModuleSym, Module] = root.modules.map {
     case (sym, mod) =>
-      val namespace = sym.ns
       val uses = root.uses.getOrElse(sym, Nil)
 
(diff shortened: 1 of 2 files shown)


magnus-madsen:
Feel free to post a screen shot when there is something ready

sockmaster27:
image

sockmaster27:
Personally, I really like this design:
image
image

magnus-madsen:
I think it looks good. We should probably have these categories:

- Modules
- Classes
- Effects
- Types
- Type Aliases
- Definitions

Did I forget anything? 

I would try to port the colors from: api.flix.dev (i.e. the red, green, etc symbols).

magnus-madsen:
Looks great so far!

A few suggestions for improvements (I understand this is still a WIP):

- [ ] Use small icons/glyps of different colors for different entities (see api.flix.dev).
- [ ] Bold the "Modules", "Definitions" etc. headers.
- [ ] Decrease the font size of the list elements. 
- [ ] Use a code font for the list elements (not headings).
- [ ] Clicking on "Definitions" could jump to the top of where definitions start on the page.
- [ ] Sort everything.

sockmaster27:
We should probably think about whether or not we actually want ligatures in the documentation -- On one hand they make it easier to read, but on the other, they might genuinely hurt communication. 

For example:
image
vs.
image

There are compromises of course, like only disabling them in names.

magnus-madsen:
!!!jar

magnus-madsen:
Looks like I broke something with the `doc` command.

sockmaster27:
Works fine for me?

magnus-madsen:
Its my bad :)

magnus-madsen:
TLDR: If the input program had an error nothing would be printed.

magnus-madsen (approved):

magnus-madsen:
Excellent. Upwards and onwards 🚀

PR #6359, opened 2023-08-17
fix: broken `doc` command

magnus-madsen:


magnus-madsen:
@mlutze Now I think it is OK

magnus-madsen:
TLDR: If the input program had an error nothing would be printed.

mlutze (approved):

PR #6367, opened 2023-08-17
chore: use system font in sidebar items in HTML documentation

sockmaster27:
I think the system font is the best font to use here to stay consistent.

Related to #6327

### DIFF
--- a/main/src/resources/doc/styles.css
+++ b/main/src/resources/doc/styles.css
@@ -85,7 +85,6 @@ nav {
 nav ul {
     margin-top: 0.2rem;
     padding-left: 0.5rem;
-    font-family: Fira Code, monospace;
     font-size: 0.85rem;
     list-style: none;
 }


magnus-madsen:
What is the system font? I would prefer we actually set the font.

sockmaster27:
The system font is the one that matches the headings in the sidebar

magnus-madsen:
> The system font is the one that matches the headings in the sidebar

Right, but it would be different depending on Windows/Linux/Mac. I would like us to fix a font. If you don't have a suggestion, I can look once I am home. But maybe 'Roboto'?

sockmaster27:
I think [Inter](https://fonts.google.com/specimen/Inter/about?category=Sans+Serif&vfonly=true&subset=latin) would be the safe choice.

magnus-madsen:
Let's try it

sockmaster27:
May I change the font in the doc strings too?

magnus-madsen:
> May I change the font in the doc strings too?

Just the menu for now. 

But later there will be an opportunity to play :)

sockmaster27:
I'll just drop some options here...

Roboto:
image

Open Sans:
image

Geologica:
image

Noto Sans:
image

Inter:
image


I think I'm somewhere between the last two, slightly leaning towards Noto Sans. Any thoughts?

sockmaster27:
We have to remember that picking the right font is the most important part of this whole project.

magnus-madsen:
How about this: If you give me 'Roboto' but only in the menu then later I will allow you to experiment with a different palette :)

PR #6372, opened 2023-08-17
feat: add dark mode to HTML documentation

sockmaster27:
Respects the system preference - A JavaScript toggle could be added.

image

Text color may need tweaking?

### DIFF
--- a/main/src/resources/doc/styles.css
+++ b/main/src/resources/doc/styles.css
@@ -3,11 +3,11 @@
     --bg-color-2: #fcfcfd;
     --bg-color-3: rgb(248, 249, 250);
 
+    --border-color: rgba(0, 0, 0, 0.07);
+
     --heading-color: rgb(33, 37, 41);
     --text-color: rgb(17, 17, 17);
     --faded-text-color: #8a8a8a;
-
-    --section-arrow-color: rgb(85, 91, 96);
 
     --annotation-color: #b15422;
     --keyword-color: #030303;
@@ -27,6 +27,37 @@
     --eff-color: #ffaefa;
     --type-alias-color: #f9cdad;
     --def-color: #83af9b;
+
+    --shadow-color: rgba(0, 0, 0, 0.05);
+    --highlight-color: rgba(0, 0, 0, 0.3);
+}
+@media (prefers-color-scheme: dark) {
+    :root {
+        --bg-color-1: #151719;
+        --bg-color-2: #101214;
+        --bg-color-3: #0e0f11;
+
+        --border-color: rgba(255, 255, 255, 0.15);
+    
+        --heading-color: #e6e6e6;
+        --text-color: #f7f7f7;
+        --faded-text-color: #8a8a8a;
+    
+        --annotation-color: #e7a37e;
+        --keyword-color: #fcfcfc;
+        --name-color: #ff6696;
+        --type-color: #6699ff;
+        --kind-color: #999999;
+        --effect-color: #ff944d;
+    
+        --doc-code-color: #ff66b3;
+    
+        --show-hide-color: #68a4fd;
+        --show-hide-hover-color: #3c86f6;
+
+        --shadow-color: rgba(0, 0, 0, 0.5);
+        --highlight-color: hsla(50, 0%, 100%, 0.3);
+    }
 }
 
 html, body {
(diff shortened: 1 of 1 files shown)


magnus-madsen:
Can we have a toggle button?

magnus-madsen:
We can grab something from here: https://codepen.io/Umer_Farooq/pen/eYJgKGN

There are also many others on codepen. Something simple should suffice.

sockmaster27:
I just did something as simple as possible:
image
image

When JavaScript is disabled this appears when it is hovered:
image

sockmaster27:
Right now I'm using [`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) to remember the setting which should work if the pages are all hosted on the same domain. 
When they are just loaded as files, though, each page has its own setting, which is a bit annoying. If we want to fix this, we could use some JavaScript to override links, SPA-style, so that instead of actually navigating to the new page, the script would just load the contents of the file, preserving the state of the script.

magnus-madsen:
> Right now I'm using [`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) to remember the setting which should work if the pages are all hosted on the same domain. When they are just loaded as files, though, each page has its own setting, which is a bit annoying. If we want to fix this, we could use some JavaScript to override links, SPA-style, so that instead of actually navigating to the new page, the script would just load the contents of the file, preserving the state of the script.

OK. We will host the pages on the same domain (api.flix.dev) so that should be OK.

main/src/ca/uwaterloo/flix/language/phase/HtmlDocumentor.scala
```diff
@@ -220,7 +225,12 @@ object HtmlDocumentor {
     val sortedDefs = mod.defs.sortBy(_.sym.name)
 
     sb.append(mkHead(moduleName(mod)))
-    sb.append("<body>")
+    sb.append("<body class='no-script'>")
+
+    sb.append("<button id='theme-toggle' disabled aria-describedby='no-script'>")
+    sb.append("<span>Toggle theme.</span>")
+    sb.append("<div role='tooltip' id='no-script'>Enable JavaScript to use this toggle</div>")
```
magnus-madsen, line 232: I find this text a bit long. Perhaps we can just skip it? So what if there is a "sun" you cannot click on?

main/src/resources/doc/index.js
```diff
@@ -0,0 +1,31 @@
+const storeKey = "flix-html-docs:use-dark-theme";
```
magnus-madsen, line 1: Perhaps STORE_KEY?

main/src/resources/doc/index.js
```diff
@@ -0,0 +1,31 @@
...
+    if (dark) {
+        body.classList.remove("light");
+        body.classList.add("dark");
+    } else {
+        body.classList.remove("dark");
+        body.classList.add("light");
+    }
+}
+
+let dark;
```
magnus-madsen, line 16: You have both a local variable called dark and you use it as a parameter to setTheme.

magnus-madsen: Obviously that "works" but it is not pretty :)

main/src/resources/doc/index.js
```diff
@@ -0,0 +1,31 @@
...
+const stored = localStorage.getItem(storeKey);
+if (stored === null) {
+    dark = window.matchMedia("(prefers-color-scheme: dark)").matches;
```
magnus-madsen, line 19: // User has not stored any preference. Check if the browser prefers a dark theme.

main/src/resources/doc/index.js
```diff
@@ -0,0 +1,31 @@
...
+} else {
+    dark = stored === "true";
```
magnus-madsen, line 21: // User has set a preference. Use it.

main/src/ca/uwaterloo/flix/language/phase/HtmlDocumentor.scala
sockmaster27: I don't see how it hurts, but I guess we could shorten it down to "Requires JavaScript"?

magnus-madsen: Yes, lets do that 👍

main/src/resources/doc/index.js
magnus-madsen: 🤷

magnus-madsen (approved):

PR #6373, opened 2023-08-17
feat: link back to prelude in HTML documentation

sockmaster27:
Related to #6327


magnus-madsen (approved):

magnus-madsen:
Semi-related: We don't actually have an `index.html`. But we probably should?

sockmaster27:
Would that be prelude, or something else?

magnus-madsen:
> Would that be prelude, or something else?

I think that would make sense.

Issue #6348, opened 2023-08-15
Add `ToJava` class


magnus-madsen:

stephentetley:
I've taken a look at this, PR to follow...

stephentetley:
PR #6378 - feat: add ToJava class (issue #6348)

magnus-madsen:
Was this fixed or do we have more plans?

stephentetley:
Fixed for now. We probably want to consider `ToJava` for element types (i.e not collections) at some point.

magnus-madsen:
> Fixed for now. We probably want to consider `ToJava` for element types (i.e not collections) at some point.

Any specifics? Maybe Optional?

stephentetley:
`Optional` we will certainly need to look at.

I was initially thinking about some of the the basic types (`Bool`, `Char` and the numeric types):

~~~

instance ToJava[Bool] {
    type O = ##java.lang.Boolean
    pub def toJava(x: Bool): ##java.lang.Boolean \ IO =
        checked_ecast(ToJava.boolToJava(x))
}

....

mod ToJava {


    pub def boolToJava(b: Bool): ##java.lang.Boolean =
        import static java.lang.Boolean.valueOf(Bool): ##java.lang.Boolean \ {};
        valueOf(b)

   ...

}

~~~

magnus-madsen:
Fixed in master.

PR #6378, opened 2023-08-17
feat: add ToJava class (issue #6348)

stephentetley:
This PR adds the `ToJava` class from issue #6348 and several collection instances.

I think the only contentious instance is for (Flix) Vector which goes to ##java.util.List. Java's ##java.util.Vector would have been an alternative but Vector seems disfavoured in Java now and is not used so much as an argument or a result in API calls.


magnus-madsen:
> 

Yes, lets do that for now.

magnus-madsen:
Ill ask @mlutze for a quick review, since he is leading on assoc types.

stephentetley:
I've reverted Prelude so `toJava` is removed.

mlutze (approved):
Looks good. Thanks, Stephen!

Issue #6290, opened 2023-07-30
Remove `numArgs` from `TyCon.Enum` in `PatternExhaustiveness`


jaschdoc:
It is a variable derived from `TyCon.Enum.args.length` and doesn't really serve a purpose

jaschdoc:
I believe so

PR #6379, opened 2023-08-18
chore: cleanup documentation and unused variable in PatternExhaustiveness

ZiyaoWei:
I was looking into #6290 but @jaschdoc scooped me 😅 which is totally fine, since I was mostly trying to learn how it works anyway. I did find a few smaller things that seems worth fixing so opening a pull request here.

(Also if people don't mind me asking - what's a good reference for how that phase works? I am reading through [Warnings for pattern matching](https://www.cambridge.org/core/services/aop-cambridge-core/content/view/3165B75113781E2431E3856972940347/S0956796807006223a.pdf/warnings-for-pattern-matching.pdf) which is referenced in the comments, and it still seems like fairly accurate?

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/PatternExhaustiveness.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/PatternExhaustiveness.scala
@@ -529,7 +529,7 @@ object PatternExhaustiveness {
     * If the constructors are a complete signature, then they are exhaustive for the type, and we just have to
     * check that their arguments are also exhaustive
     *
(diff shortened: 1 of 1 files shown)


magnus-madsen:
> (Also if people don't mind me asking - what's a good reference for how that phase works? I am reading through [Warnings for pattern matching](https://www.cambridge.org/core/services/aop-cambridge-core/content/view/3165B75113781E2431E3856972940347/S0956796807006223a.pdf/warnings-for-pattern-matching.pdf) which is referenced in the comments, and it still seems like fairly accurate?

Exactly!

magnus-madsen (approved):

PR #6387, opened 2023-08-18
feat: tweak css in `HtmlDocumentor`

magnus-madsen:


magnus-madsen:
@mlutze If you could just smash that button :)

mlutze (approved):

Issue #5917, opened 2023-05-11
Refactor code generation for Arrays


magnus-madsen:
This is a hard one.

I would like to refactor the code generation for arrays (i.e. `Array*` in the AST).

I would like two things:

1. To get rid of all of this nonsense:

```
        if (elmType == MonoType.Str) {
          visitor.visitTypeInsn(ANEWARRAY, "java/lang/String")
        } else if (elmType.isInstanceOf[MonoType.Native]) {
          val native = elmType.asInstanceOf[MonoType.Native]
          val name = native.clazz.getName.replace('.', '/')
          visitor.visitTypeInsn(ANEWARRAY, name)
        } else if (jvmType == JvmType.Object) { // Happens if the inner type is an object type
          visitor.visitTypeInsn(ANEWARRAY, "java/lang/Object")
        } else { // Happens if the inner type is a primitive type
          visitor.visitIntInsn(NEWARRAY, AsmOps.getArrayTypeCode(jvmType))
        }
```

We need a more principled approach.

2. To properly support multidimensional arrays: https://docs.oracle.com/javase/specs/jvms/se11/html/jvms-6.html

The key is to ensure that we always create arrays of the right type. 

This should be possible since we have all the information we need.

However, one simplification that might be useful would be to introduce a new `MonoType.ArrayMultiDim`:

```
  case class ArrayMultiDim(tpe: MonoType, dim: Int32) extends MonoType
```

Let me know if you have questions.

magnus-madsen:
You might want to come up with a (written) plan before starting on the implementation.

magnus-madsen:
@jaschdoc Are you up for this next? :)

magnus-madsen:
Or would you rather work on another feature?

JonathanStarup:
Ive looked at this a bit so if you need extra hands or opinions im here :)

jaschdoc:
> @jaschdoc Are you up for this next? :)

Yes. I had delayed it until all the record stuff was wrapped up but I will start working on it soon :)

magnus-madsen:
https://docs.oracle.com/javase/specs/jls/se7/html/jls-10.html
https://docs.oracle.com/javase/specs/jvms/se11/html/jvms-6.html#jvms-6.5.multianewarray

jaschdoc:
> ```scala
> case class ArrayMultiDim(tpe: MonoType, dim: Int32) extends MonoType
> ```
> 
[...]

PR #6393, opened 2023-08-21
refactor: remove special cases in codegen for `AtomicOp.New`

jaschdoc:

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
@@ -742,7 +742,6 @@ object GenExpression {
         // Evaluating the 'length' of the array
         compileExpr(exp2)
         // Instantiating a new array of type jvmType
-        mv.visitTypeInsn(ANEWARRAY, jvmTpe.toDescriptor)
         jvmTpe match {
           case JvmType.Reference(name) => mv.visitTypeInsn(ANEWARRAY, name.toInternalName)
           case _ => mv.visitIntInsn(NEWARRAY, AsmOps.getArrayTypeCode(jvmTpe))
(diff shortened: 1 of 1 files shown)


magnus-madsen:
@jaschdoc I am not sure this approach can work. Note that all of these are different:

`Object[]`
`int[]`
`Integer[]`
`int[][]`
`Object[][]`

and may be required in different contexts. Maybe you can find Jonathan and have him help explain?

jaschdoc:
> @jaschdoc I am not sure this approach can work. Note that all of these are different:
> 
> `Object[]` `int[]` `Integer[]` `int[][]` `Object[][]`
> 
> and may be required in different contexts. Maybe you can find Jonathan and have him help explain?

Roger that, thanks :)

jaschdoc:
By the way, this was just meant as a small PR that changed the original `if-else` to a simpler pattern match

magnus-madsen:
> By the way, this was just meant as a small PR that changed the original `if-else` to a simpler pattern match

Oh I see, but then some cases are missing or?

jaschdoc:
It *should* be equivalent to what we had, because it matched on the mono type and the erased jvm type to figure out if it were a string, native, object, or a primitive. I looked into the code, and it seems that if we just match on the jvm type instead of the erased jvm type, the string, native, and object cases are captured by the `Reference` case and the primitive cases are handled by the wild case.

jaschdoc:
However, currently it generates byte code for the specific reference type instead of just `Object` for anything that's not string, native, or primitive. I'm thinking, though, since it generated arrays for specific types of native classes (which would be anything imported by the user) then it should also handle the complex types the compiler generates the same way. Thus, this PR *should not* result in any observable change in behavior.

magnus-madsen (approved):

PR #6394, opened 2023-08-21
feat: Lexer error

herluf-ba:
This PR adds `LexerErr` as a new `CompilationError` as well as some glue code to produce `LexerErr`s in the lexing phase.


herluf-ba:
I accidentally based this PR on `lexer` rather than `master`. Ill fix and reopen

magnus-madsen:
You can rebase, but OK

PR #6395, opened 2023-08-21
feat: Add `LexerError` as a new type of `CompilationMessage`

herluf-ba:
Write a `LexerError` type for compilation messages when lexing fails.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/TokenErrorKind.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/TokenErrorKind.scala
@@ -15,20 +15,23 @@
  */
 package ca.uwaterloo.flix.language.ast
 
-sealed trait TokenKind
+sealed trait TokenErrorKind
 
-object TokenKind {
+object TokenErrorKind {
 
-  case object LParen extends TokenKind
+  case object UnexpectedChar extends TokenErrorKind
 
-  case object RParen extends TokenKind
+  case object UnterminatedString extends TokenErrorKind
 
-  // TODO: LEXER
+  case object UnterminatedChar extends TokenErrorKind
 
-  case object Eof extends TokenKind
+  case object UnterminatedInfixFunction extends TokenErrorKind
 
-  case class Err(kind: TokenErrorKind) extends TokenKind
+  case object DoubleDottedNumber extends TokenErrorKind
 
+  case object BlockCommentTooDeep extends TokenErrorKind
+
+  case object UnterminatedBlockComment extends TokenErrorKind
+
+  case object UnterminatedBuiltIn extends TokenErrorKind
 }
-
-
--- a/main/src/ca/uwaterloo/flix/language/ast/TokenErrorKind.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/TokenErrorKind.scala
@@ -19,19 +19,19 @@ sealed trait TokenErrorKind
 
 object TokenErrorKind {
 
+  case object BlockCommentTooDeep extends TokenErrorKind
+
+  case object DoubleDottedNumber extends TokenErrorKind
+
   case object UnexpectedChar extends TokenErrorKind
 
-  case object UnterminatedString extends TokenErrorKind
+  case object UnterminatedBlockComment extends TokenErrorKind
+
+  case object UnterminatedBuiltIn extends TokenErrorKind
 
   case object UnterminatedChar extends TokenErrorKind
 
   case object UnterminatedInfixFunction extends TokenErrorKind
 
-  case object DoubleDottedNumber extends TokenErrorKind
-
-  case object BlockCommentTooDeep extends TokenErrorKind
-
-  case object UnterminatedBlockComment extends TokenErrorKind
-
-  case object UnterminatedBuiltIn extends TokenErrorKind
+  case object UnterminatedString extends TokenErrorKind
 }
--- a/main/src/ca/uwaterloo/flix/language/ast/TokenKind.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/TokenKind.scala
@@ -27,6 +27,27 @@ object TokenKind {
 
   case object Eof extends TokenKind
 
-  case object Err extends TokenKind
+  case class Err(kind: ErrKind) extends TokenKind
 
+}
+
+sealed trait ErrKind
+
+object ErrKind {
+
+  case object UnexpectedChar extends ErrKind
+
+  case object UnterminatedString extends ErrKind
+
+  case object UnterminatedChar extends ErrKind
+
+  case object UnterminatedInfixFunction extends ErrKind
+
+  case object DoubleDottedNumber extends ErrKind
+
+  case object BlockCommentTooDeep extends ErrKind
+
+  case object UnterminatedBlockComment extends ErrKind
+
+  case object UnterminatedBuiltIn extends ErrKind
 }
--- a/main/src/ca/uwaterloo/flix/language/errors/LexerError.scala
+++ b/main/src/ca/uwaterloo/flix/language/errors/LexerError.scala
@@ -32,14 +32,14 @@ object LexerError {
    * @param loc the location of char.
    */
   case class UnexpectedChar(s: String, loc: SourceLocation) extends LexerError {
-    override def summary: String = s"Unexpected character '$char'"
+    override def summary: String = s"Unexpected character '$s'."
 
     override def message(formatter: Formatter): String = {
       import formatter._
       s"""${line(kind, source.name)}
          |>> Unexpected character '${red(s)}'.
          |
-         |${code(loc, "found here")}
+         |${code(loc, "Unexpected character.")}
          |
          |""".stripMargin
     }
@@ -53,14 +53,14 @@ object LexerError {
    * @param loc The location of the opening `"`.
    */
   case class UnterminatedString(loc: SourceLocation) extends LexerError {
-    override def summary: String = s"Unterminated string"
+    override def summary: String = s"Unterminated string."
 
     override def message(formatter: Formatter): String = {
       import formatter._
       s"""${line(kind, source.name)}
          |>> Unterminated string.
          |
-         |${code(loc, "starts here")}
+         |${code(loc, "String starts here.")}
          |
          |""".stripMargin
     }
@@ -74,14 +74,14 @@ object LexerError {
    * @param loc The location of the opening `'`.
    */
   case class UnterminatedChar(loc: SourceLocation) extends LexerError {
-    override def summary: String = s"Unterminated char"
+    override def summary: String = s"Unterminated char."
 
     override def message(formatter: Formatter): String = {
       import formatter._
       s"""${line(kind, source.name)}
          |>> Unterminated char.
          |
-         |${code(loc, "starts here")}
+         |${code(loc, "Char starts here")}
          |
          |""".stripMargin
     }
@@ -95,14 +95,14 @@ object LexerError {
    * @param loc The location of the opening '`'.
    */
   case class UnterminatedInfixFunction(loc: SourceLocation) extends LexerError {
-    override def summary: String = s"Unterminated infix function"
+    override def summary: String = s"Unterminated infix function."
 
     override def message(formatter: Formatter): String = {
       import formatter._
       s"""${line(kind, source.name)}
          |>> Unterminated infix function.
          |
-         |${code(loc, "starts here")}
+         |${code(loc, "Infix function starts here.")}
          |
          |""".stripMargin
     }
@@ -116,14 +116,14 @@ object LexerError {
    * @param loc The location of the opening "$".
    */
   case class UnterminatedBuiltin(loc: SourceLocation) extends LexerError {
-    override def summary: String = s"Unterminated built-in"
+    override def summary: String = s"Unterminated built-in."
 
     override def message(formatter: Formatter): String = {
       import formatter._
       s"""${line(kind, source.name)}
          |>> Unterminated built-in.
          |
-         |${code(loc, "starts here")}
+         |${code(loc, "Built-in starts here.")}
          |
          |""".stripMargin
     }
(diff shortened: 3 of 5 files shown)


main/src/ca/uwaterloo/flix/language/ast/TokenKind.scala
```diff
@@ -27,6 +27,27 @@ object TokenKind {
 
   case object Eof extends TokenKind
 
-  case object Err extends TokenKind
+  case class Err(kind: ErrKind) extends TokenKind
 
 }
+
+sealed trait ErrKind
```
magnus-madsen, line 34: This should go in its own file together with the object below.

main/src/ca/uwaterloo/flix/language/errors/LexerError.scala
```diff
@@ -16,5 +16,186 @@
 package ca.uwaterloo.flix.language.errors
 
 import ca.uwaterloo.flix.language.CompilationMessage
+import ca.uwaterloo.flix.language.ast.SourceLocation
```
magnus-madsen, line 19: Put your own name in the copyright :)

main/src/ca/uwaterloo/flix/language/errors/LexerError.scala
```diff
@@ -16,5 +16,186 @@
...
+import ca.uwaterloo.flix.util.Formatter
+
+sealed trait LexerError extends CompilationMessage {
```
magnus-madsen, line 22: Remember to reformat file with intellij idea.

main/src/ca/uwaterloo/flix/language/errors/LexerError.scala
```diff
@@ -16,5 +16,186 @@
...
+  val kind = "Lexer Error"
+}
+
+object LexerError {
+
+  /**
+   * An error raised when an unexpected character, such as €, is encountered
```
magnus-madsen, line 29: We write full sentences as documentation, so everything should end with a period.

main/src/ca/uwaterloo/flix/language/errors/LexerError.scala
```diff
@@ -16,5 +16,186 @@
...
+   *
+   * @param s the problematic character. This is a string to handle escaped characters
+   * @param loc  the location of `char`
```
magnus-madsen, line 32: char not in `

main/src/ca/uwaterloo/flix/language/errors/LexerError.scala
```diff
@@ -16,5 +16,186 @@
...
+   */
+  case class UnexpectedChar(s: String, loc: SourceLocation) extends LexerError {
+    override def summary: String = s"Unexpected character '$char'"
+
+    override def message(formatter: Formatter): String = {
+      import formatter._
+      s"""${line(kind, source.name)}
+         |>> Unexpected character '${red(s)}'.
+         |
+         |${code(loc, "found here")}
```
magnus-madsen, line 42: We usually repeat the message e.g. unexpected character

main/src/ca/uwaterloo/flix/language/errors/LexerError.scala
```diff
@@ -16,5 +16,186 @@
...
+   */
+  case class UnterminatedString(loc: SourceLocation) extends LexerError {
+    override def summary: String = s"Unterminated string"
+
+    override def message(formatter: Formatter): String = {
+      import formatter._
+      s"""${line(kind, source.name)}
+         |>> Unterminated string.
+         |
+         |${code(loc, "starts here")}
```
magnus-madsen, line 63: Lets write "string starts here" or some such.

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -17,15 +17,20 @@ package ca.uwaterloo.flix.language.phase
...
 import scala.collection.mutable
 
 object Lexer {
 
-  def run(root: ReadAst.Root)(implicit flix: Flix): Validation[Map[Ast.Source, Array[Token]], CompilationMessage] = {
+  // TODO: A single source can have more than one error that we would like to report.
+  // For instance there might be an unexpected char multiple spots in the source file.
+  // Therefore run needs to return something like `Validation[Map[Ast.Source, Array[Token]], Array[CompilationMessage]]`
+  // but that seems to clash with the rest of the pipeline. How do we resolve this?
+  def run(root: ReadAst.Root)(implicit flix: Flix): Validation[Map[Ast.Source, Array[Token]], Array[CompilationMessage]] = {
```
magnus-madsen, line 33: This should not be Array. A validation (e.g. failure and softFailure) already supports multiple failures.

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -64,11 +69,40 @@ object Lexer {
 
   private def addToken(token: Token)(implicit s: State): Unit = ???
 
-  private class State {
+  // Converts a `Token` of kind `TokenKind.Err` into a CompilationMessage.
```
magnus-madsen, line 72: Doc comments use /**

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -64,11 +69,40 @@ object Lexer {
...
+  // NOTE: Why is this necessary?
+  // We would like the lexer to capture as many errors as possible before terminating.
+  // To do this, the lexer will produce error tokens instead of halting,
+  // each holding a kind of the simple type `ErrKind`.
+  // So we need this mapping to produce a `CompilationMessage`, which is a case class, if there were any errors.
+  private def tokenErrToCompilationMessage(t: Token)(implicit s: State): Option[CompilationMessage] = {
```
magnus-madsen, line 78: Suggestion: Change signature to `Token.Err` then change return type to CompilationMessage.

Then where you use this, you can write `tokens.collect { case Token.Err => }`, from this you can construct a `Failure` object.

herluf-ba: I don't think this is possible. `Token` holds a `TokenKind` which can then be an `TokenKind.Err`. But we need the rest of the data on `Token` as well (`text`, `line` and `col`). Does scala have a way to express the type "Any token where its `kind` field is an instance of `TokenKind.Err`" ?

magnus-madsen: Yes, like a deep pattern match?

```
case Token(TokenKind.Err(...), ..., ...) => 
```

main/src/ca/uwaterloo/flix/language/ast/TokenErrorKind.scala
```diff
@@ -0,0 +1,37 @@
...
+
+  case object UnterminatedInfixFunction extends TokenErrorKind
+
+  case object DoubleDottedNumber extends TokenErrorKind
+
+  case object BlockCommentTooDeep extends TokenErrorKind
+
+  case object UnterminatedBlockComment extends TokenErrorKind
+
+  case object UnterminatedBuiltIn extends TokenErrorKind
```
magnus-madsen, line 36: Sort these and keep them in alphabetical order :)

main/src/ca/uwaterloo/flix/language/errors/LexerError.scala
```diff
@@ -26,20 +26,20 @@ sealed trait LexerError extends CompilationMessage {
 object LexerError {
```
magnus-madsen, line 27: sort the case classes inside this object (to keep them in the same order as TokenErrorKind).

main/src/ca/uwaterloo/flix/language/errors/LexerError.scala
```diff
@@ -26,20 +26,20 @@ sealed trait LexerError extends CompilationMessage {
...
 
   /**
-   * An error raised when an unexpected character, such as €, is encountered
+   * An error raised when an unexpected character, such as €, is encountered.
    *
-   * @param s the problematic character. This is a string to handle escaped characters
-   * @param loc  the location of `char`
+   * @param s   the problematic character. This is a string to handle escaped characters.
```
magnus-madsen, line 31: What does it mean "This is a string to handle"? Maybe simply:

```@param s the unexpected character```?

File: main/src/ca/uwaterloo/flix/language/errors/LexerError.scala [hunk 14]
```diff
@@ -132,19 +132,19 @@ object LexerError {
   }
 
   /**
-   * An error raised when an unterminated block comment is encountered
+   * An error raised when an unterminated block comment is encountered.
    *
-   * @param loc The location of the opening "\*"
+   * @param loc The location of the opening "\*".
```
magnus-madsen, line 137: Do you mean /* :)

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -47,14 +43,16 @@ object Lexer {
...
 
-  private def lex(src: Ast.Source): Validation[Array[Token], Array[CompilationMessage]] = {
+  private def lex(src: Ast.Source): Validation[Array[Token], CompilationMessage] = {
     implicit val s: State = new State(src)
     // TODO: LEXER
 
     val hasErrors = s.tokens.exists(t => t.kind.isInstanceOf[TokenKind.Err])
     if (hasErrors) {
+      // TODO: Fix type error in return type
       s.tokens.flatMap(tokenErrToCompilationMessage).toArray.toFailure
```
magnus-madsen, line 56: I think we want soft failure here. To allow parsing with error tokens.

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -69,52 +67,54 @@ object Lexer {
...
+   * So we need this mapping to produce a `CompilationMessage`, which is a case class, if there were any errors.
+   */
   private def tokenErrToCompilationMessage(t: Token)(implicit s: State): Option[CompilationMessage] = {
     t.kind match {
       case TokenKind.Err(e) => e match {
-        case ErrKind.UnexpectedChar => Some(
+        case TokenErrorKind.UnexpectedChar => Some(
           LexerError.UnexpectedChar(
             t.text,
             SourceLocation(None, s.src, SourceKind.Real, t.line, t.col, t.line, t.col + t.text.length)
```
magnus-madsen, line 84: Lots of duplication here.

magnus-madsen (commented):
Great. I think one or two more rounds of feedback and then this is ready to go :)

main/src/ca/uwaterloo/flix/language/errors/LexerError.scala
herluf-ba: Good suggestion. I was trying to justify using the type string instead of char for `s` since I thought that might confuse a reader. I think its unnecessary though.

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
herluf-ba: I'm having trouble getting this to work. It seems to me like I need to use `Validation.sequence` and maybe `Validation.mapN`?

magnus-madsen: In that case you probably need to construct `SoftFailure` manually.

herluf-ba: Yep that did the trick! I ended up with 
```scala
    Validation.SoftFailure(s.tokens.toArray, LazyList.from(s.tokens).collect {
      case Token(TokenKind.Err(e), t, l, c) => tokenErrToCompilationMessage(e, t, l, c)
    })
```

magnus-madsen:
I will do one final review. Do you have a screenshot of a lexer error?

magnus-madsen (approved):

magnus-madsen:
I will merge once tests complete. Next up I think is the definition of `TokenKind`.

herluf-ba:
> I will do one final review. Do you have a screenshot of a lexer error?

For a quick test I just prepended a `€` to `foo.flix` added an error token manually like so:
```scala
s.tokens += Token(TokenKind.Err(TokenErrorKind.UnexpectedChar), "€", 0, 0)
```

That gives:
```
-- Lexer Error -------------------------------------------------- foo.flix

>> Unexpected character '€'.

0 | 
   ^
   Unexpected character.
```

The source code isn't showing up. Any ideas why that might be?

magnus-madsen:
Merged!

magnus-madsen:
> The source code isn't showing up. Any ideas why that might be?

Let me take a look. Probably the positions are off or the way you construct the location is wrong.

herluf-ba:
Yep, the input field is set to `None`

magnus-madsen:
> Yep, the input field is set to `None`

Note `ParserInput` comes from the current parser. Not sure you can or should construct it.

magnus-madsen:
Possibly `SourceLocation` needs to be slightly refactored.

herluf-ba:
Yea, it seems like a quick and dirty solution is to replace `ParserInput` with a string in `SourceLocation` for now. That might hurt performance though. I can see that `ParserInput` works on a byte array, presumably for faster substring access. 

I can make a PR's with the modification to `SourceLocation` if that's something we want?

magnus-madsen:
> Yea, it seems like a quick and dirty solution is to replace `ParserInput` with a string in `SourceLocation` for now. That might hurt performance though. I can see that `ParserInput` works on a byte array, presumably for faster substring access.
> 
> I can make a PR with the modification to `SourceLocation` if that's something we want?

I am OK with a refactor, but this is production code, so it has to be right. And it has to be tested so we don't cause breakages.

It seems what we want is really a map from `LineNo -> String`.

herluf-ba:
Makes sense. I will dip my toes into a refactor soon 🙂

PR #6400, opened 2023-08-22
feat: add `Reader.flix`

### LINKED ISSUE #6297: Add `Reader` class
```scala
///
/// A type class for types which are resources that can be accessed as bytes.
///
class Reader[t] {
    ///
    /// Reads `k` bytes from the underlying resource and writes them into `b`.
    ///
    /// Returns `Ok(k)` to signify that `k` bytes was successfully read and written to `b`.
    ///
    /// Guarantees that `0 <= k < length(b)`.
    ///
    /// Returns `Err(e)` if some underlying I/O error occurs.
    ///
    pub def read(b: Array[Int8, r], r: t): Result[IOError, Int32] \ {r, IO}
}
```

magnus-madsen:
Fixes #6297

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/Flix.scala
+++ b/main/src/ca/uwaterloo/flix/api/Flix.scala
@@ -216,6 +216,8 @@ class Flix {
     "MutMap.flix" -> LocalResource.get("/src/library/MutMap.flix"),
 
     "File.flix" -> LocalResource.get("/src/library/File.flix"),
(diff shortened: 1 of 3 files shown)


magnus-madsen:
(We deliberately start with a `Reader` that does not have assoc effects)

magnus-madsen:
@JonathanStarup Maybe you can approve this one?

JonathanStarup (approved):

PR #6404, opened 2023-08-22
refactor: add `MonoType.ArrayMultiDim`

jaschdoc:
This PR changes `MonoType.Array` to `MonoType.ArrayMultiDim` (we should consider just adding `dim` to `MonoType.Array` instead of renaming the type, in my opinion), but does not yet change any behavior.


magnus-madsen:
I think we don't yet fully know how to proceed, so I would suggest to continue with this PR adding support for `multianewarray` and see how it works. Then we can decide if the new type is worth it or not later. But at least for now it should make it easier to understand what is happening.

jaschdoc:
> 

Roger. I'll explore the issue a bit more then

PR #6408, opened 2023-08-23
feat: generate correct jvm array types

jaschdoc:

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/jvm/AsmOps.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/jvm/AsmOps.scala
@@ -137,25 +137,6 @@ object AsmOps {
     case BackendType.Array(_) => AASTORE
   }
 
-  def getArrayType(tpe: BackendType): String = {
-    @tailrec
-    def visit(tt: BackendType, acc: String): String = tt match {
-      case BackendType.Bool => acc + "Z"
-      case BackendType.Char => acc + "C"
-      case BackendType.Int8 => acc + "B"
-      case BackendType.Int16 => acc + "S"
-      case BackendType.Int32 => acc + "I"
-      case BackendType.Int64 => acc + "J"
-      case BackendType.Float32 => acc + "F"
-      case BackendType.Float64 => acc + "D"
-      case BackendType.Reference(ref) if acc.isEmpty => ref.jvmName.toInternalName
-      case BackendType.Reference(ref) => acc + "L" + ref.jvmName.toInternalName + ";"
-      case BackendType.Array(t) => visit(t, acc + "[")
-    }
-
-    visit(tpe, "")
-  }
-
   /**
     * Returns the CheckCast type for the value of the type specified by `tpe`
     */
--- a/main/src/ca/uwaterloo/flix/language/phase/jvm/AsmOps.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/jvm/AsmOps.scala
@@ -122,6 +122,22 @@ object AsmOps {
   }
 
   /**
+    * Returns the array store instruction for arrays of the given JvmType tpe
+    */
+  def getArrayStoreInstruction(tpe: BackendType): Int = tpe match {
+    case BackendType.Bool => BASTORE
+    case BackendType.Char => CASTORE
+    case BackendType.Int8 => BASTORE
+    case BackendType.Int16 => SASTORE
+    case BackendType.Int32 => IASTORE
+    case BackendType.Int64 => LASTORE
+    case BackendType.Float32 => FASTORE
+    case BackendType.Float64 => DASTORE
+    case BackendType.Reference(_) => AASTORE
+    case BackendType.Array(_) => AASTORE
+  }
+
+  /**
     * Returns the array type code for the value of the type specified by `tpe`
     */
   def getArrayTypeCode(tpe: JvmType): Int = tpe match {
--- a/main/src/ca/uwaterloo/flix/language/phase/jvm/AsmOps.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/jvm/AsmOps.scala
@@ -188,6 +188,23 @@ object AsmOps {
   /**
     * Returns the Array fill type for the value of the type specified by `tpe`
     */
+  def getArrayFillType(tpe: BackendType): String = tpe match {
+    case BackendType.Bool => "([ZZ)V"
+    case BackendType.Char => "([CC)V"
+    case BackendType.Int8 => "([BB)V"
+    case BackendType.Int16 => "([SS)V"
+    case BackendType.Int32 => "([II)V"
+    case BackendType.Int64 => "([JJ)V"
+    case BackendType.Float32 => "([FF)V"
+    case BackendType.Float64 => "([DD)V"
+    case BackendType.Reference(_) => "([Ljava/lang/Object;Ljava/lang/Object;)V"
+    case BackendType.Array(_) => "([Ljava/lang/Object;Ljava/lang/Object;)V"
+  }
+
+
+  /**
+    * Returns the Array fill type for the value of the type specified by `tpe`
+    */
   def getArrayFillType(tpe: JvmType): String = tpe match {
     case JvmType.Void => throw InternalCompilerException(s"Unexpected type $tpe", SourceLocation.Unknown)
     case JvmType.PrimBool => "([ZZ)V"
--- a/main/src/ca/uwaterloo/flix/language/phase/jvm/AsmOps.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/jvm/AsmOps.scala
@@ -24,6 +24,8 @@ import ca.uwaterloo.flix.util.{InternalCompilerException, JvmTarget}
 import org.objectweb.asm.Opcodes._
 import org.objectweb.asm.{ClassWriter, Label, MethodVisitor}
 
+import scala.annotation.tailrec
+
 object AsmOps {
 
   /**
@@ -145,6 +147,25 @@ object AsmOps {
     case BackendType.Float32 => Some(T_FLOAT)
     case BackendType.Float64 => Some(T_DOUBLE)
     case BackendType.Array(_) | BackendType.Reference(_) => None
+  }
+
+  def getArrayType(tpe: BackendType): String = {
+    @tailrec
+    def visit(tt: BackendType, acc: String): String = tt match {
+      case BackendType.Bool => acc + "Z"
+      case BackendType.Char => acc + "C"
+      case BackendType.Int8 => acc + "B"
+      case BackendType.Int16 => acc + "S"
+      case BackendType.Int32 => acc + "I"
+      case BackendType.Int64 => acc + "J"
+      case BackendType.Float32 => acc + "F"
+      case BackendType.Float64 => acc + "D"
+      case BackendType.Reference(ref) if acc.isEmpty => ref.jvmName.toInternalName
+      case BackendType.Reference(ref) => acc + "L" + ref.jvmName.toInternalName + ";"
+      case BackendType.Array(t) => visit(t, acc + "[")
+    }
+
+    visit(tpe, "")
   }
 
   /**
@@ -444,9 +465,9 @@ object AsmOps {
     * If the field is a primitive then it is boxed using the appropriate java type, if it is not a primitive
     * then we just return the field
     *
-    * @param method     MethodVisitor used to emit the code to a method
-    * @param fieldType  type of the field to be boxed
-    * @param classType  class that the field is defined on
+    * @param method    MethodVisitor used to emit the code to a method
+    * @param fieldType type of the field to be boxed
+    * @param classType class that the field is defined on
     * @param fieldName name of the field to be boxed
     */
   def boxField(method: MethodVisitor, fieldType: JvmType, classType: JvmType.Reference, fieldName: String): Unit = {
--- a/main/src/ca/uwaterloo/flix/language/phase/jvm/AsmOps.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/jvm/AsmOps.scala
@@ -135,6 +135,18 @@ object AsmOps {
     case JvmType.Reference(_) => throw InternalCompilerException(s"Expected primitive type. Actual type: $tpe", SourceLocation.Unknown)
   }
 
+  def getArrayTypeCode(tpe: BackendType): Option[Int] = tpe match {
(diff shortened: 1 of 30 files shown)


JonathanStarup:
hmmm that is a very good question. i think we should look at what surface programs we expect to turn into a multi init

jaschdoc:
> hmmm that is a very good question. i think we should look at what surface programs we expect to turn into a multi init.

Maybe these:

(We can just generate length 0 arrays)
```scala
pub def main(): Unit \ IO =
    let _a : Array[Array[Array[Int32, Static], Static], Static] = Array.new(Static, 10);
    println("Hello")
```
```scala
pub def main(): Unit \ IO =
    let _a = Array#{Array#{1, 2, 3, 4, 5, 6} @ Static, Array#{7, 8, 9}  @ Static } @ Static;
    println("Hello")
```

JonathanStarup:
isnt the first example filled with null?

almost think we should give up on multidim init unless we add it to the surface. it seems ad hoc to reprganise these singular instantiations into multi ones

JonathanStarup:
I think your link says that we can do it. also: multidim is used for regular matrices like 45x45 but an array of an array doesn't have to be homogenous in its inner length do nested anewarray must be fine. 

Your test above actually cant use multidim because the inner lengths are different.

keep in mind even though the type if 5 dimensional, all aspects of it can be null so initialization isnt an issue.

jaschdoc:
@JonathanStarup what do you think of this version of the code? It instantiates the correct (specific) types for arrays but for some reason the Flix jvm classes cannot be found on runtime, so they are just erased in `BackendType.toBackendType` helper function. Do you have a clue why this might be the case?

main/src/ca/uwaterloo/flix/language/phase/jvm/AsmOps.scala
```diff
@@ -120,19 +122,50 @@ object AsmOps {
...
-    case JvmType.Void => throw InternalCompilerException(s"Unexpected type $tpe", SourceLocation.Unknown)
-    case JvmType.PrimBool => T_BOOLEAN
-    case JvmType.PrimChar => T_CHAR
-    case JvmType.PrimFloat => T_FLOAT
-    case JvmType.PrimDouble => T_DOUBLE
-    case JvmType.PrimByte => T_BYTE
-    case JvmType.PrimShort => T_SHORT
-    case JvmType.PrimInt => T_INT
-    case JvmType.PrimLong => T_LONG
-    case JvmType.Reference(_) => throw InternalCompilerException(s"Expected primitive type. Actual type: $tpe", SourceLocation.Unknown)
```
JonathanStarup, line 135: Hmm maybe we should add an intermediate interface `Primitive` between `BackendType` and all the primitives.

main/src/ca/uwaterloo/flix/language/phase/jvm/AsmOps.scala
```diff
@@ -120,19 +122,50 @@ object AsmOps {
...
+    def visit(tt: BackendType, acc: String): String = tt match {
+      case BackendType.Bool => acc + "Z"
+      case BackendType.Char => acc + "C"
+      case BackendType.Int8 => acc + "B"
+      case BackendType.Int16 => acc + "S"
+      case BackendType.Int32 => acc + "I"
+      case BackendType.Int64 => acc + "J"
+      case BackendType.Float32 => acc + "F"
+      case BackendType.Float64 => acc + "D"
+      case BackendType.Reference(ref) if acc.isEmpty => ref.jvmName.toInternalName
```
JonathanStarup, line 163: Other than this case, this function is the same as calling `tt.toDescriptor`, why is that?

main/src/ca/uwaterloo/flix/language/phase/jvm/BackendType.scala
```diff
@@ -134,4 +135,32 @@ object BackendType {
          MonoType.SchemaEmpty | MonoType.SchemaExtend(_, _, _) | MonoType.Native(_) |
          MonoType.Region => BackendObjType.JavaObject.toTpe
   }
+
+  /**
+    * Computes the `BackendType` based on the given `MonoType`.
+    */
+  def toBackendType(tpe: MonoType): BackendType = tpe match {
```
JonathanStarup, line 142: This performs some erasure but doesn't say so in its name or docs. Maybe call it `toFlixErasedBackendType` and then explain what that means.

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -1485,6 +1477,11 @@ object GenExpression {
 
   }
 
+  private def visitArrayInstantiate(mv: MethodVisitor, tpe: BackendType): Unit = tpe match {
+    case BackendType.Array(_) | BackendType.Reference(_) => mv.visitTypeInsn(ANEWARRAY, AsmOps.getArrayType(tpe))
+    case _ => mv.visitIntInsn(NEWARRAY, AsmOps.getArrayTypeCode(tpe).head)
```
JonathanStarup, line 1482: If `getArrayTypeCode` is only used here, then I think it should be inlined. Then you avoid the option.

JonathanStarup (changes requested):

main/src/ca/uwaterloo/flix/language/phase/jvm/AsmOps.scala
jaschdoc: I see. For some reason the jvm is quite inconsistent when it comes to arrays of objects; if the component type is a non-array object then it uses the internal name, otherwise the object must use its descriptor with a `[` prepended.

main/src/ca/uwaterloo/flix/language/phase/jvm/AsmOps.scala
jaschdoc: Do you think we should add that in `BackendType` or what role do you imagine `Primitive` having?

jaschdoc: I assume it should be a `case object` that extends `BackendType` if we just want it to wrap all the usages as is.

main/src/ca/uwaterloo/flix/language/phase/jvm/AsmOps.scala
jaschdoc: I refactor `BackendType.Array` to override the `toDescriptor` method

main/src/ca/uwaterloo/flix/language/phase/jvm/AsmOps.scala
jaschdoc: The primitive type could actually implement most of the functions in `AsmOps`.

JonathanStarup: yeah, but leave that to another PR

jaschdoc:
@JonathanStarup it may bloat `BackendType` but I've moved some functions from `AsmOps` to `BackendType` to encapsulate the functionality we had spread over several functions and files. In short, it improves the readability of `GenExpression`. We can revert it, though, it's your call.

main/src/ca/uwaterloo/flix/language/phase/jvm/BackendType.scala
```diff
@@ -44,63 +46,60 @@ object VoidableType {
...
+    case BackendType.Primitive(PrimitiveType.Float64) => true
+    case BackendType.Primitive(_) | BackendType.Array(_) |
+         BackendType.Reference(_) => false
+  }
 
-  case object Int8 extends BackendType
+  /**
+    * Returns the Array fill type for the value of the type specified by `tpe`
+    */
+  def toArrayFillType: String = s"([${this.toErased.toDescriptor}${this.toErased.toDescriptor})V"
```
JonathanStarup, line 83: You can remove more strings if you want. `mkDescriptor(Array[this.toErased], Array[this.toErased])(VoidableType.Void).toDescriptor` or something like that

main/src/ca/uwaterloo/flix/language/phase/jvm/BackendType.scala
```diff
@@ -44,63 +46,60 @@ object VoidableType {
...
 
-  case object Int16 extends BackendType
+}
 
-  case object Int32 extends BackendType
+object BackendType {
 
-  case object Int64 extends BackendType
+  case class Primitive(tpe: PrimitiveType) extends BackendType
```
JonathanStarup, line 89: This is very Flix like but maybe `Primitive` should just be a marker interface, i.e. a trait that is implemented by the primitive types. Those primitive types would look exactly like they did before (so still accessed by `BackendType.Int32`). This would all code "ignore" the fact that this trait is there but you can match on it if you want to. Hope that makes sense

main/src/ca/uwaterloo/flix/language/phase/jvm/BackendType.scala
```diff
@@ -44,63 +46,60 @@ object VoidableType {
...
 
-  case object Float32 extends BackendType
+  case class Array(tpe: BackendType) extends BackendType {
+    override def toDescriptor: String = {
```
JonathanStarup, line 92: Can you move this code into the trait implementation? I think its unexpected in our code style that the method is overridden.

main/src/ca/uwaterloo/flix/language/phase/jvm/BackendType.scala
```diff
@@ -114,24 +113,52 @@ object BackendType {
...
+    case MonoType.Int64 => Primitive(PrimitiveType.Int64)
     case MonoType.Unit | MonoType.BigDecimal | MonoType.BigInt | MonoType.String | MonoType.Regex |
          MonoType.Array(_) | MonoType.Lazy(_) | MonoType.Ref(_) | MonoType.Tuple(_) |
          MonoType.Enum(_) | MonoType.Arrow(_, _) | MonoType.RecordEmpty | MonoType.RecordExtend(_, _, _) |
          MonoType.SchemaEmpty | MonoType.SchemaExtend(_, _, _) | MonoType.Native(_) |
          MonoType.Region => BackendObjType.JavaObject.toTpe
   }
+
+  /**
+    * Computes the `BackendType` based on the given `MonoType`.
```
JonathanStarup, line 138: explain what FlixErased means (just 1-2 sentences)

main/src/ca/uwaterloo/flix/language/phase/jvm/BytecodeInstructions.scala
```diff
@@ -467,24 +468,24 @@ object BytecodeInstructions {
...
 
-    case BackendType.Array(BackendType.Bool) => INVOKESTATIC(BackendObjType.Arrays.BoolArrToString)
-    case BackendType.Array(BackendType.Char) => INVOKESTATIC(BackendObjType.Arrays.CharArrToString)
-    case BackendType.Array(BackendType.Int8) => INVOKESTATIC(BackendObjType.Arrays.Int8ArrToString)
-    case BackendType.Array(BackendType.Int16) => INVOKESTATIC(BackendObjType.Arrays.Int16ArrToString)
-    case BackendType.Array(BackendType.Int32) => INVOKESTATIC(BackendObjType.Arrays.Int32ArrToString)
-    case BackendType.Array(BackendType.Int64) => INVOKESTATIC(BackendObjType.Arrays.Int64ArrToString)
-    case BackendType.Array(BackendType.Float32) => INVOKESTATIC(BackendObjType.Arrays.Float32ArrToString)
-    case BackendType.Array(BackendType.Float64) => INVOKESTATIC(BackendObjType.Arrays.Float64ArrToString)
+    case BackendType.Array(BackendType.Primitive(Bool)) => INVOKESTATIC(BackendObjType.Arrays.BoolArrToString)
```
JonathanStarup, line 481: with the `Primitive` trait change then all these lines would be unchanged

main/src/ca/uwaterloo/flix/language/phase/jvm/PrimitiveType.scala
```diff
@@ -0,0 +1,61 @@
...
+    case PrimitiveType.Char => "C"
+    case PrimitiveType.Int8 => "B"
+    case PrimitiveType.Int16 => "S"
+    case PrimitiveType.Int32 => "I"
+    case PrimitiveType.Int64 => "J"
+    case PrimitiveType.Float32 => "F"
+    case PrimitiveType.Float64 => "D"
+  }
+
+  def toErasedString: String = this match {
```
JonathanStarup, line 20: There is no erasure here so probably just `toString`

main/src/ca/uwaterloo/flix/language/phase/jvm/PrimitiveType.scala
```diff
@@ -0,0 +1,61 @@
...
+    case PrimitiveType.Int8 => T_BYTE
+    case PrimitiveType.Int16 => T_SHORT
+    case PrimitiveType.Int32 => T_INT
+    case PrimitiveType.Int64 => T_LONG
+    case PrimitiveType.Float32 => T_FLOAT
+    case PrimitiveType.Float64 => T_DOUBLE
+  }
+}
+
+object PrimitiveType {
```
JonathanStarup, line 43: As eluded to in other comments, i think these types should still live in `BackendType`

JonathanStarup (changes requested):

jaschdoc:
> I think the `Primitive` type should be its own PR but I'll let it be up to you if you want to git that

Agreed, I just wanted to experiment with it first :)

jaschdoc:
@JonathanStarup I have reset this branch to before the primitive type refactor, so now it's just like on `main` except it generates the correct array types. I have created https://github.com/flix/flix/pull/6420 where we can continue with the refactor

main/src/ca/uwaterloo/flix/language/phase/jvm/BackendType.scala
jaschdoc: I'm not sure, I follow. Could you explain that a bit more? What do you want to remove?

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -1485,6 +1477,21 @@ object GenExpression {
 
   }
 
+  private def visitArrayInstantiate(mv: MethodVisitor, tpe: BackendType): Unit = {
```
magnus-madsen, line 1480: Comment?

JonathanStarup (approved):
Nothing except magnuses doc request

jaschdoc:
Ready

Issue #6374, opened 2023-08-17
HtmlDocumentor Checklist


magnus-madsen:
Checklist:
- [x] All pages should produce valid HTML.
- [ ] CSS should be valid
- [x] All pages should render without JavaScript.
- [x] All pages should work locally (i.e. without internet).
- [x] All links should work.
- [x] Mobile should work.

magnus-madsen:
https://validator.w3.org/nu/?doc=https%3A%2F%2Fapi.flix.dev%2F

magnus-madsen:
@sockmaster27 Looks good so far. A few things (other than the menu thing which is still a problem):

- I get a CSS warning.
- I think links to GitHub should have the "nofollow" tag-- this would also help link checkers.

Other than fixing these three things, are there any small tweaks or refactorings you want to do before we move on to the next thing?

(I plan to revisit the HtmlDocumenter-- but not for a few months while we work on other things and get ideas for how  to improve it).

magnus-madsen:
@sockmaster27 Uhm.. Are the `source` links now broken??

sockmaster27:
> * I get a CSS warning.

Where? I don't see anything?

magnus-madsen:
It was just some online checker

magnus-madsen:
https://jigsaw.w3.org/css-validator/validator?uri=api.flix.dev&profile=css3svg&usermedium=all&warning=1&vextwarning=&lang=en

PR #6415, opened 2023-08-25
feat: save fonts locally in HTML documentation

sockmaster27:
Should finalize #6374


sockmaster27:
I'm not sure I'm a big fan of using both Inter and Noto Sans btw. They're very similar.

magnus-madsen:
I am not sure about adding these fonts to the repo. What is the problem with depending on Google? (I guess it does not work offline).

sockmaster27:
@magnus-madsen 
Yes, exactly

sockmaster27:
Plus, I believe it allows us to set a content security policy that's just like `default-src: 'self'`, since we don't need to load anything from any other domains.

magnus-madsen:
(Converted to draft while I decide what to do.)

sockmaster27:
@magnus-madsen 
What's the holdup? 🤔

magnus-madsen:
> @magnus-madsen What's the holdup? 🤔

My indecision about whether this is a good idea and whether to add 1MB of binary data to the repo.

magnus-madsen:
I have decided against including fonts-- for now at least.

Issue #6402, opened 2023-08-22
Add instances for `Reader`


magnus-madsen:

stephentetley:
We don't have a (binary) File type currently in the standard library to be test case for `Reader`. 

Should I make a wrapper over the JDK's `FileInputStream` as standard example / implementation?

magnus-madsen:
> We don't have a (binary) File type currently in the standard library to be test case for `Reader`.
> 
> Should I make a wrapper over the JDK's `FileInputStream` as standard example / implementation?

I think so yes. I am not sure what we should do about the `File` module. It was intended as something quick and dirty, but I still think such a high-level API can be useful: sometimes you just want to dump strings into files other times you want precise control.

magnus-madsen:
Perhaps `File` module should have been called "`Files`", i.e. its a collection of utility functions for working on files, not a file itself.

stephentetley:
> Perhaps the `File` module should have been called "`Files`", i.e. its a collection of utility functions for working on files, not a file itself.

Yes, that seems to be the JDK way of doing things e.g. `java.nio.file.Files`

magnus-madsen:
Note that Java's `File` / `Path` seems to be like a value object that just holds the name of a file whereas a Rust `File` is actually an open resource that you can read from:

https://doc.rust-lang.org/std/fs/struct.File.html

I understand you might want both (because paths are tricky), but I think it would be nice to have a Rust-style `File` that you can obtain from just a string (if you know you are in the "easy case") and then just read directly from.

magnus-madsen:
Since Flix is not following Java terminology, perhaps we could have both `File`, `Path`, and `Files` where a `File` can be obtained from a string or via detour to `Path` (if one wants to be careful output legal file system paths/names).

magnus-madsen:
For now, I think we should introduce a `File` which wraps a Java `FileInputStream` (or something to that effect).

stephentetley:
I favour keeping the name `FileInputStream` because it's read only so only it only has half the capabilities what you might expect a "File" to have. When we get round to adding `FileOutputStream` that would have equal claim on the name `File` as it provides the write only view of things.

Should I pair `FileInputStream` with a region to emphasize it is an imperative "resource" like `StringBuilder` or is enough that all the operations working on it will use the `IO` effect?

i.e. 
~~~

enum FileInputStream[_: Region](##java.io.FileInputStream)

~~~

Or:

~~~

[...]

PR #6422, opened 2023-08-29
feat: add File a wrapper over the JDK's RandomAccessFile (see #6402)

stephentetley:
This PR adds a `File` module which is a wrapper over Java's `RandomAccessFile`. It changes the name of the existing `File` module to `Files` - this is a convention that Java uses: collect utility functions in a module with a pluralized name.

`File` is made an instance of `Reader`.

I've added a `ToString` instance to `IOError` (used in debugging but not needed in the final code) and an `IOError` version of `Result.tryCatch` which is very useful for wrapping IO functions that can fail.

Tests use a set of prepared files and also write to new files in the `testdata` directory.

Downsides: 
`RandomAccessFile` is not a modern Java class. The set of methods it provides seem biased towards serializing and deserializing Java data between Java programs. It is missing proper Unicode support and and reading and writing IEEE 754 floats and doubles (the format `RandomAccessFile` uses for Float and Double is a Java specific one). Also with Java not having unsigned types, the widening / truncation is not very pleasant - i.e. to work with unsigned "bytes" `Int32` is the chosen number type... 

My suspicion is that `File` (wrapping RandomAccessFile) isn't really going to be the general file type we would want as the canonical `File` type in the standard library in the long term. That said I thing it would be useful in the short term as the current `Files` module doesn't really support incremental reading and writing.


magnus-madsen:
Thanks Stephen-- Looking forward to review it.

One quick thing: Could we do the rename of `File` to `Files` in a separate PR?

stephentetley:
> One quick thing: Could we do the rename of File to Files in a separate PR?

Yes I'll do that.

stephentetley:
PR  #6423

stephentetley:
I'll make a new PR for this, the conflicts are a bit messy for me.

PR #6424, opened 2023-08-29
feat: add new File module wrapping Java's RandomAccessFile (see issue #6402)

stephentetley:
This PR adds a `File` module which is a wrapper over Java's `RandomAccessFile`. 

`File` is made an instance of `Reader`.

I've added a `ToString` instance to `IOError` (used in debugging but not needed in the final code, still seems genrally useful) and an `IOError` version of `Result.tryCatch` which is very useful for wrapping IO functions that can fail.

Tests use a set of prepared files and also write to new files in the `testdata` directory.

Problems:
`RandomAccessFile` is not a modern Java class. The set of methods it provides seem biased towards serializing and deserializing Java data between Java programs. It is missing proper Unicode support and reading and writing IEEE 754 floats and doubles (the format `RandomAccessFile` uses for `Float` and `Double` is a Java specific one). Also with Java not having unsigned types, the widening / truncation is not very pleasant - i.e. to work with unsigned "bytes" Int32 is the chosen number type...

My suspicion is that this `File` module (wrapping RandomAccessFile) isn't really going to be the general file type we would want as the canonical `File` type in the standard library in the long term. That said I thing it would be useful in the short term as the current `Files` module doesn't really support incremental reading and writing.


magnus-madsen:
My baby was just born :) -- so I will be a bit slower on the PRs/issues. 

(Also both Matt and Jonathan are away, so no-one is manning the fort until mid next week.)

stephentetley:
> My baby was just born :) -- so I will be a bit slower on the PRs/issues.

Oh wow, congratulations Magnus !!!

magnus-madsen (approved):

magnus-madsen:
Thanks Stephen and thanks for the comments.

I merging this with the understanding that probably- at some point in the future- we will want to build on a better foundation that `RandomAccessFile`. Still, I think this is a good starting point, and it is a step towards a better API.

magnus-madsen:
Do you want to continue with `Writer`? and their buffered counter parts?

stephentetley:
Thanks Magnus. 

Yes I'll look at Writer and the buffered versions.

PR #6427, opened 2023-09-03
refactor: rename `Field` -> `Label` for Records

### LINKED ISSUE #6399: Rename "record field" to "record label" everywhere
The idea is that records will have "labels" and in the future structs will have "fields".

This means everywhere in the compiler (but also on doc.flix.dev and flix.dev), we need "record field" to become "record label".

jaschdoc:
Closes #6399


jaschdoc:
I have not yet renamed anything in the backend as I'm not sure what to rename. For instance, we have a `PUTFIELD` function but I'm not certain it's only used for records. Also, I'm unsure about what to do with `Name.Field`.

magnus-madsen:
Awaits `Name.Field -> Name.Label`

main/src/ca/uwaterloo/flix/language/ast/Name.scala
```diff
@@ -278,6 +283,32 @@ object Name {
...
 
+  /**
+    * The name of a label.
+    *
+    * @param name the name of the label.
+    * @param loc  the specific occurrence of the name.
+    */
+  case class Label(name: String, loc: SourceLocation) {
+    /**
+      * Two predicate names are equal if their names are the same.
```
magnus-madsen, line 294: predicate -> label

main/src/ca/uwaterloo/flix/language/ast/Name.scala
```diff
@@ -278,6 +283,32 @@ object Name {
...
+      */
+    override def hashCode(): Int = name.hashCode
+
+    /**
+      * Two predicate names are equal if their names are the same.
```
magnus-madsen, line 299: predicate -> label

main/src/ca/uwaterloo/flix/api/lsp/Entity.scala
```diff
@@ -108,6 +108,12 @@ object Entity {
     def precision: Precision = Precision.High
   }
 
+  case class Label(e: Name.Label) extends Entity {
```
magnus-madsen, line 111: Any reason not to take out Field? I would worry that with it present that could be things that have not been renamed correctly?

It should be easy to bring back later.

magnus-madsen:
Once there are PRs for flix.dev and doc.flix.dev I will merge this :)

magnus-madsen (approved):

magnus-madsen:
Thanks-- big one.

I would like the next priority to be to update the docs and website.

PR #6428, opened 2023-09-03
feat: add 'copy link' button to HTML documentation

sockmaster27:
Related to #6327
Should finalize milestone 4


sockmaster27:
Animation

sockmaster27:
When JavaScript is disabled it will act as a normal link

magnus-madsen:
I like it, but not its placement. Can we simply put it next to the "source" link on the right-- we can reuse the icon.

sockmaster27:
Here's a comparison
image
image

Personally, I think the second one looks a bit weird, since the link refers to the entire box, and it goes against the convention.

If it's the spacing you're objecting to we have the option to do something like this
image
or this 
image

magnus-madsen:
I like this one:

image

but I would put it into a box like source with the text "link".

I think its important to understand this feature will very rarely be used (like looking at the source), hence it makes sense to "hide it on the right". If we put it on the left, then eye always sees it before you get to the real content.

magnus-madsen:
Also I think it is consistent that the main content is on the left and "meta content" is on the right.

sockmaster27:
looks like this now
image

magnus-madsen:
!!!jar

magnus-madsen (approved):

magnus-madsen:
Looks good!

Issue #6426, opened 2023-09-03
Add conversions between Java primitive types and their object versions to the stdlib


stephentetley:
For marshaling to and from Java collections we should be able to put and get Java primitives like `char` (Flix `Char`) and `int` (Flix `Int32`) using their corresponding object types `java.lang.Character` and `java.lang.Integer`. We should add these conversion functions to the standard library as they are canonical.

stephentetley:
This one is fairly self explanatory so I've already made a PR - #6429

PR #6429, opened 2023-09-03
feat: add conversion functions for Char and the number types to/from their Java object versions (issue #6426)

stephentetley:
See issue #6426 

This PR add conversion functions to Char and the primitive number types to convert them to and from their Java corresponding Java objects. Converting to Java objects is necessary to store them in generic collections from `java.util` that can only store objects and not primitive types. 

`BigInt` and `BigDecimal` are not primitive types so don't need these conversion functions.

Possible design issues: 
1. Maybe these functions should go in `Adaptor`, though they would need name changes to distinguish the types.
2. If they stay in the current modules, maybe they should diverge from the Java naming for for `charValue`, `byteValue` etc.

Potential work todo - I haven't used the new `valueOf` functions at other places in their respective modules that are already importing and using Java's `valueOf`. Unless it's likely to be a performance hit, it would be nicer to use the new function rather than import the Java one.

### DIFF
--- a/main/src/library/Char.flix
+++ b/main/src/library/Char.flix
@@ -240,4 +240,24 @@ mod Char {
             case c                  => Some(c)
         }
 
+    ///
+    /// Get the primitive Char value from its object representation (i.e. ##java.lang.Character).
+    ///
+    /// This function is expected to be used when marshaling Chars from Java. Generally in Flix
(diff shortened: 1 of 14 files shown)


magnus-madsen:
Thanks!

PR #6437, opened 2023-09-09
feat: merge classes and enums with their companion module in HTML documentation

sockmaster27:
Related to #6327
(Should finalize milestone 5)


sockmaster27:
I have yet to include effects in this, since there are no examples

sockmaster27:
Looks like this 
image

magnus-madsen:
You could add a dummy effect and pass it in:

```
eff Foo {}

mod Foo {
  def baz()
}
```

main/src/ca/uwaterloo/flix/language/phase/HtmlDocumentor.scala
```diff
@@ -174,11 +176,20 @@ object HtmlDocumentor {
     * but with all items that shouldn't appear in the documentation removed.
     */
   private def filterItems(mods: Map[Symbol.ModuleSym, Module]): Map[Symbol.ModuleSym, Module] = mods.map {
-    case (sym, Module(_, parent, uses, submodules, classes, enums, effects, typeAliases, defs)) =>
+    case (sym, Module(_, parent, uses, mainItem, submodules, classes, enums, effects, typeAliases, defs)) =>
       sym -> Module(
         sym,
         parent,
         uses,
+        mainItem match {
```
magnus-madsen, line 184: is mainItem more like what we call the Prelude or RootModule?

magnus-madsen:
!!!jar

magnus-madsen:
Looks like a good start! I am not yet entirely sure where to end up, but lets consider `Foldable`:

image

This is pretty nice, but `Foldable` has lots of functions which can be access by typing `Foldable.XYZ`:

image

so I have two thoughts:

- For type classes, we should probably expand "Defintions" by default, and
- For type classes, we should probably show their member definitions on the left. 

But maybe we want to use two different icons, because `Foldable.exists` and `Foldable.fold` differ in that the latter is in the companion module whereas the former is in the type class itself.

magnus-madsen:
Maybe the left menu should have:

```
Class Signatures
- foldLeft
- foldRight
Class Definitions
- exists
- filter
- ...
Module Definitions
- fold

main/src/ca/uwaterloo/flix/language/phase/HtmlDocumentor.scala
sockmaster27: I'm not sure I understand. 
`mainItem` is the enum/class that's displayed at the top of the given module.

magnus-madsen:
It is possible that when generating a page, we should think about whether we are generating it for:

- A pure module
- A type class and its companion module
- An effect and its companion module
- A type and its companion module

main/src/ca/uwaterloo/flix/language/phase/HtmlDocumentor.scala
magnus-madsen: Ah okay. Its just that the word "main" to me means the "main function".

sockmaster27: I see. Do you propose an alternative?

sockmaster27:
@magnus-madsen 
To avoid special-cases, wouldn't it be better to do something more like

* Foldable
  * Signatures
    - foldLeft
    - foldRight
    ...
  * Definitions
    - count
    - dropWhile
    ...
* Definitions
  * fold

magnus-madsen:
> @magnus-madsen To avoid special-cases, wouldn't it be better to do something more like
> 
>     * Foldable
>       
>       * Signatures
>         
>         * foldLeft
>         * foldRight
>           ...
>       * Definitions
>         
>         * count
>         * dropWhile
>           ...
> 
>     * Definitions
>       
>       * fold

We could start there.  The goal is really to unify the content of the `Foldable` type class with its companion module.

magnus-madsen:
I think in time we will want separate "templates" / "layouts" for the "kind" of page, e.g. one for types, one for classes, one for effects, and so forth. At the same time, a page like `Prelude` would not show all its constituent type classes on the right. To see a type class you would have to navigate to its page, e.g. `Foldable.html`. Does that make sense?

magnus-madsen:
Perhaps we do a simple refactor first (like we are doing now) and then we iterate in subsequent PRs.

sockmaster27:
Yes, that sounds like a major project. I think we should just start with this PR as it is, since it's not too big of a change, and then do a PR for each type of page we want to add.

magnus-madsen:
Do you want to do the menu in this PR or the next?

sockmaster27:
I probably won't be working on it until next weekend, so I think it's best to just merge this one if possible

magnus-madsen (approved):

PR #6438, opened 2023-09-10
feat: add spacing in subsections in HTML documentation

sockmaster27:
This adds lines like this
image


magnus-madsen:
Conflicted. Where is the extra space?

sockmaster27:
I guess it's not spacing, but the extra lines separating each signature, definition and instance.

magnus-madsen (approved):

PR #6439, opened 2023-09-10
feat: add `ParOps.mapValues`

### LINKED ISSUE #2974: Add `ParOps.mapValues` for `Map[K, V]` and use it broadly
The idea is to have a function `mapValues` which applies a function `V1 => V2` to all values in a Map in parallel and rebuilds the map.

We can then use it in several places in the compiler.

sockmaster27:
Related #2974


magnus-madsen:
We also need to figure out where to use it. A few tests might be nice too.

magnus-madsen:
Going to ask @mlutze to comment.

main/src/ca/uwaterloo/flix/util/ParOps.scala
```diff
@@ -43,6 +43,15 @@ object ParOps {
     result.seq
   }
 
+  /**
+    * Apply the given function `f` to each value in the map `m` in parallel.
+    */
+  @inline
+  def mapValues[K, A, B](m: Map[K, A])(f: A => B)(implicit flix: Flix): Map[K, B] =
+    parMap(m) {
```
magnus-madsen, line 51: A question is whether using the parMap function here creates too many intermediate data structures.

mlutze: Maybe there is a nice general interface that all 3 functions can use.

mlutze (approved):

mlutze:
Can this be merged in so I can use it? :)

mlutze:
Would also be nice to have some helpers for dealing with parallel processing things that return Validations

magnus-madsen:
But I would like parValues to be used where parMap is used today (on maps), and I would like it to be more efficient.

mlutze:
> But I would like parValues to be used where parMap is used today (on maps), and I would like it to be more efficient.

Can't that be done as a next step?

magnus-madsen:
> > But I would like parValues to be used where parMap is used today (on maps), and I would like it to be more efficient.
> 
> Can't that be done as a next step?

Okay

magnus-madsen (approved):

PR #6444, opened 2023-09-10
fix: single argument enum tag display bug in LSP completion

### LINKED ISSUE #6025: Formatting of enum cases in VSCode Completion hints
image

`Ansa` should be `Ans`

ZiyaoWei:
Fixes #6025 

...which is caused by incorrectly adding surrounding parentheses for Enum tags. From the commit history it seems that the intended logic correct but the implementation isn't ([comment](https://github.com/flix/flix/pull/5977/commits/dfe95cdfc76a5e1c0aa92f91e7d4ded1a429a397#r1197611912)).

This changes the logic so that instead of looking at the type of the arguments, we just look at the arity and would only add parentheses when arity is 1.

Screenshot -
image

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
@@ -257,7 +257,7 @@ sealed trait Completion {
       val args = (1 until arity + 1).map(i => s"?elem$i").mkString(", ")
       val snippet = if (args.isEmpty) name else s"$name($args)"
       CompletionItem(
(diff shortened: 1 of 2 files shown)


magnus-madsen:
LGTM. @mlutze ?

magnus-madsen:
> Another interesting thing is that I cannot seem to trigger the case `case tpe: Type.Cst` - is that even possible?

Yes, e.g. `Bool` or `String` should be of Type.Cst.

mlutze (approved):

magnus-madsen:
Thanks!

Issue #6095, opened 2023-06-12
Add support for (diff-based) mutation testing


magnus-madsen:

Ibragimov11:
Hello!
I am interested in this issue, but could you please add some description about what is expected?

magnus-madsen:
Hi Ibragimov. Can you swing by our gitter channel and then we can talk? In in the city currently, but I can write when back.

magnus-madsen:
We are here: https://app.gitter.im/#/room/#flix_Lobby:gitter.im

magnus-madsen:
@Ibragimov11 I added some details to the ticket.

magnus-madsen:
@Ibragimov11 just FYI, I'm on vacation. I'll be back in a week.

Ibragimov11:
> @Ibragimov11 I added some details to the ticket.

Thanks a lot, it's much clearer now. I will write you a more detailed answer in gitter now

Ibragimov11:
>What mutations should we start with?

I think it's better to start with mutations for booleans and numbers.

Booleans:
true  -> false
false -> true
&&    -> ||
||       -> &&
and   -> or
or     -> and

Numbers:
increment mutator: n -> n + 1
decrement mutator: n -> n - 1

Ibragimov11:
>What AST should we mutate? E.g. do we do it early (WeededAst) or late (TypedAst)?

I've been thinking about this question for a long time, but I couldn't come up with any ideas. Intuitively, I would prefer typecast, but I would like to hear your opinion about it.

Ibragimov11:
#6446

magnus-madsen:
I agree that is very sensible to start with Booleans and numbers.

As for the AST, in some sense it matters less, but I think the TypedAst is the right AST-- because that is when you have all information. For example, in the typed ast you know that an expression has type Bool or type Int32-- but in the WeededAst such information is not readily available.

PR #6446, opened 2023-09-11
feat: add mutation testing

Ibragimov11:
related to #6095


main/src/ca/uwaterloo/flix/language/phase/Mutator.scala
```diff
@@ -0,0 +1,149 @@
...
+      case Constant.Unit => exp
+      case Constant.Null => ???
+      case Constant.Bool(lit) => ???
+      case Constant.Char(lit) => ???
+      case Constant.Float32(lit) => ???
+      case Constant.Float64(lit) => ???
+      case Constant.BigDecimal(lit) => ???
+      case Constant.Int8(lit) => ???
+      case Constant.Int16(lit) => ???
+      case Constant.Int32(lit) =>
```
Ibragimov11, line 51: Do we want to be able to create more than 1 mutations if necessary? I think that would be a good solution. For example, for Int n : `n + 1` and ` n - 1`.

magnus-madsen: Usually one makes exactly one change to the program.

main/src/ca/uwaterloo/flix/language/phase/Mutator.scala
```diff
@@ -0,0 +1,149 @@
...
+import ca.uwaterloo.flix.language.ast.TypedAst._
+import ca.uwaterloo.flix.util.Validation
+import ca.uwaterloo.flix.util.Validation._
+
+object Mutator {
+
+  // TODO: The purpose of this phase is to:
+  // 1. Pick a random def (perhaps rather a def from a specific file ?)
+  // 2. Pick a random expr (how?)
+  // 3. Slightly modify the expr (e.g. changing (x + 1) to (x - 1).
```
Ibragimov11, line 16: I don't quite understand why we need random.

Don't we want to implement mutations for various Nodes, which we will use when parsing AST wherever possible?

main/src/ca/uwaterloo/flix/language/phase/Mutator.scala
Ibragimov11: Is it right in this case to create separate mutators for different changes? For example, to increase or decrease a number by 1, it can be MutatorIncrementer and MutatorDecrementer.

main/src/ca/uwaterloo/flix/language/phase/Mutator.scala
magnus-madsen: Not sure i understand the question

Ibragimov11: It says here to mutate a random one def/expr. But the rest are all not needed? Or if needed, how to get this with a random selection?

magnus-madsen: What I mean is that when you do mutation testing, you should only mutate one thing at a time.

For example, it would not be sensible to change _every_ `+` to a `-`.

So you have to pick a random `def` and change one if its expressions.

Ibragimov11:
@magnus-madsen I have added a basic version of the mutator for constants and booleans. Similarly, it is possible to implement other mutations, but first I want to get feedback on the already written code. Is this a good solution?

magnus-madsen:
This seems pretty reasonable to me 👍 

Next up is defining a strategy that applies _ONE_ mutator to _ONE_ random place in the Ast.

magnus-madsen:
One question is whether:

```
 private val binaryFloat32OpMutants: Map[Float32Op, Float32Op] = Map(
    Float32Op.Add -> Float32Op.Sub,
    Float32Op.Sub -> Float32Op.Add,
    Float32Op.Mul -> Float32Op.Div,
    Float32Op.Div -> Float32Op.Mul,
  )
```

is "enough". This seems to only replace Add with Sub, but don't you also want to try replacing Add with e.g. "Mul"?

magnus-madsen:
Stylistically, I would put all mutators in one file, but we can worry about that later.

Ibragimov11:
> Next up is defining a strategy that applies _ONE_ mutator to _ONE_ random place in the Ast.

Yes, I can implement this and I understand your idea that only _ONE_ expr needs to be mutated at a time.

But it seems to me that at the moment my decision does not contradict this rule.

I seem to have confused you a bit, and that's what I meant:
- We have a source file in which there are N expressions
- Iterate over expressions and mutate only the selected expr at each step
- As a result, we get N new files, in each of which only one mutation was committed

For example, having a source file:
```
2 + 4
8 - 5
6 * 3
```
and mutators _Add -> Sub_, _Sub -> Add_, _Mul -> Div_, I will get next 3 files:

1)
```
2 - 4
8 - 5
6 * 3
```
2)
```
2 + 4
8 + 5
6 * 3
```
3)
```
2 + 4
8 - 5
6 / 3
```

And in this case, we have full coverage of our source code with mutations, when the use of random would be less effective and would not guarantee it.

Is this a valid path or should I use random anyway??

Ibragimov11:
> One question is whether:
> 
> ```
>  private val binaryFloat32OpMutants: Map[Float32Op, Float32Op] = Map(
>     Float32Op.Add -> Float32Op.Sub,
>     Float32Op.Sub -> Float32Op.Add,
>     Float32Op.Mul -> Float32Op.Div,
>     Float32Op.Div -> Float32Op.Mul,
>   )
> ```
> 
> is "enough". This seems to only replace Add with Sub, but don't you also want to try replacing Add with e.g. "Mul"?

I can add this mutation too, no problem. But I don't quite understand why we need to have an additional mutant `Add->Mul`, which checks no more cases than `Add->Sub`

These are not all the mutations that I planned to add, but of course it would be better to discuss this first. I'm sorry I didn't think of that.

I was guided by these three solutions: 
1) https://github.com/avito-tech/go-mutesting#arithmeticbase
2) https://stryker-mutator.io/docs/mutation-testing-elements/supported-mutators/#arithmetic-operator
3) https://infection.github.io/guide/mutators.html#Binary-Arithmetic

And in each of them, the implemented mutations for binary arithmetic expressions looks like this

```
Add -> Sub
Sub -> Add
Mul -> Div
Div -> Mul
Rem -> Mul
```

magnus-madsen:
Note that if you enumerate all possible mutations you "set of files" (set of ASTs) might be huge!

magnus-madsen:
How do you envision the overall loop?

- Generate a million ASTs
- Test them all

or

- Generate one AST
- Test it
- Then loop

Ibragimov11:
> How do you envision the overall loop?
> 
> * Generate a million ASTs
> * Test them all
> 
> or
> 
> * Generate one AST
> * Test it
> * Then loop

Thanks for the hint 
The second way is much better, and I'm going to implement it

Ibragimov11:
@magnus-madsen hello!

I have refactored my solution by following the points from #6095

- Collected the basic logic in the MutationTester class
- Added the command :test-with-mutator
- Implemented the following set of mutators to start (will be expanded):

  1. Incrementer
  ```
    x -> x + 1
  ```
  2. Decrementer
  ```
    x -> x + 1
  ```
  3. BooleanMutator
  ```
    True -> False
  ```

Ibragimov11:
@magnus-madsen hello!

### I have refactored my solution by following the points from #6095

- Collected the basic logic in the MutationTester class
- Added the command :test-with-mutator
- Implemented the following set of mutators to start (will be expanded):

  1. Incrementer
  ```
    x -> x + 1
  ```
  2. Decrementer
  ```
    x -> x - 1
  ```
  3. BooleanMutator
  ```
    True   ->   False
    False  ->   True
    Not(e) ->   e
    &&     ->   ||
    ||     ->   &&
  ```
  4. ArithmeticUnaryMutator
  ```
    -n   ->   n
    !n   ->   n
  ```
  5. ArithmeticBinaryMutator
  ```
    +  ->  -
    -  ->  +
    *  ->  /
    /  ->  *
    %  ->  *
    
    // bitwise
    &   ->   |
    |   ->   &
    ^   ->   &
    >>  ->   <<
    <<  ->   >>
  ```
  6. ConditionalNegateMutator
  ```
    ==  ->  !=
    !=  ->  ==
    <   ->  >=
    <=  ->  >
    >   ->  <=
    >=  ->  <
  ```
  7. ConditionalBoundaryMutator
  ```
    <   ->  <=
    <=  ->  <
    >   ->  >=
    >=  ->  >
  ```
  
- Also, in my opinion, it is valid to filter library sources and tests, since we do not want to mutate them.

  
### Some next planned steps
- Implement a Reporter class that collects and returns information about the execution. Think about how to display the surviving mutants in the output
- Add new mutators. Also research this for datalog
- Think about parallelization so that the program runs faster
- Add Tests


I would greatly appreciate it if you could take a look at this and leave me some comments.

Ibragimov11:
Added mutators

```hs
--------------------------------------------
// Constants:
	
  // Booleans:
    False -> True
    True  -> False

  // Numbers:
    n -> n + 1
    n -> n - 1

  // Strings:
    "" -> "Mutant"
    s  -> ""

--------------------------------------------
// Unary:

  // Booleans:
    not expr -> expr
    expr -> not expr

  // Arithmetic:
    - expr -> expr

  // Bitwise:
    ~ expr -> expr

--------------------------------------------
// Binary:

  // Booleans:
    && -> ||
    || -> &&

  // Arithmetic:
    +  ->  -
    -  ->  +
    *  ->  /
    /  ->  *
    %  ->  *

  // Bitwise:
    &   ->  |
    |   ->  &
    ^   ->  &
    >>  ->  <<
    <<  ->  >>

  // Conditional Negate:
    ==  ->  !=
    !=  ->  ==
    >   ->  <=
    <   ->  >=
    >=  ->  < 
    <=  ->  >

  // Conditional Boundary:
    >   ->  >=
    <   ->  <=
    >=  ->  >
    <=  ->  <

--------------------------------------------
ConstraintSet:
  1. remove each constraint
  2. mutate each constraint
      Head
        1. mutate each argument
      Body
        1. remove each item
        2. mutate each item
            Atom
              1. mutate each argument
              2. inverse polarity
            Functional
              1. mutate expr
            Guard
              1. mutate expr

where

      Atom: (not) Rule(X1, ..., Xn)
Functional: let var = fun(X1, ..., Xn)
     Guard: if expr
--------------------------------------------
```

Ibragimov11:
Reporting looks like this

image

Ibragimov11:
Now I think that the main part of the logic is implemented and I'm going to write unit tests.

magnus-madsen:
A few comments that could be helpful:

- Did you try to run your tool on List and TestList? That should be a useful test case. How many mutants do you find?
- You will need to deal with non-termination. For example, by maintaining a step counter. (See other PR).
- Do you mutate pattern matches? (E.g. by swapping cases etc.)
- If you replace `Add.add` by `Sub.sub` do you follow the types? For example `a + b` cannot become `a - b` if a and b are strings.

Issue #6448, opened 2023-09-11
Introduce Constraint-based Typer


magnus-madsen:
- [x] Add `ca.uwaterloo.util.collection.Chain` with minimal infrastructure (fast concat `++`)
- [x] Separate reconstruction into separate phase `TypeReconstruction` (probably requires `KindedAst.Def` to hold a subst.)
- [ ] Add a new data type `TypeConstraint` (rename/rework existing, if needed).
- [ ] Add a new phase `TypeConstraints` which mimics `Typer.inferExp` but returns a chain of `TypeConstraint`s.
- [ ] Add a new phase `TypeSolver` which, for each function, uses unification to produce a substitution.

PR #6452, opened 2023-09-13
feat: add internal Chain data structure

mlutze:
related to #6448

### DIFF
--- a/main/src/ca/uwaterloo/flix/util/collection/Chain.scala
+++ b/main/src/ca/uwaterloo/flix/util/collection/Chain.scala
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2023 Matthew Lutze
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ca.uwaterloo.flix.util.collection
+
+/**
+  * A linear data structure that allows fast concatenation.
+  */
+sealed trait Chain[+A] extends Iterable[A] {
+
+  /**
+    * Returns an iterator over the chain, from left to right.
+    */
+  override def iterator: Iterator[A] = this match {
+    case Chain.Empty => Iterator.empty
+    case Chain.One(x) => Iterator.single(x)
+    case Chain.Many(l, r) => Iterator.concat(l, r)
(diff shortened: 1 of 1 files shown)


magnus-madsen:
Tests are for suckers? :D

magnus-madsen:
What about an overloaded apply, so one can write `Chain(x, y, z, w)`. That'd be nice.

main/src/ca/uwaterloo/flix/util/collection/Chain.scala
```diff
@@ -0,0 +1,52 @@
...
+  /**
+    * The empty chain.
+    */
+  override val empty: Chain[A] = Chain.Empty
+}
+
+object Chain {
+  case object Empty extends Chain[Nothing]
+  case class One[A](x: A) extends Chain[A]
+
```
magnus-madsen, line 50: Extra line break?

magnus-madsen:
Thoughts on chain of chains? e.g. another `ManyMany(xs: Seq[Chain[t]])`?

magnus-madsen:
Also, I think we want a fast `toList` or something like that.

main/src/ca/uwaterloo/flix/util/collection/Chain.scala
```diff
@@ -0,0 +1,52 @@
...
+  override def iterator: Iterator[A] = this match {
+    case Chain.Empty => Iterator.empty
+    case Chain.One(x) => Iterator.single(x)
+    case Chain.Many(l, r) => Iterator.concat(l, r)
+  }
+
+  /**
+    * Concatenates `this` chain and `that` chain.
+    */
+  def ++[B >: A](that: Chain[B]) : Chain[B] = (this, that) match {
```
magnus-madsen, line 35: This is actually slow for such a vital operation. Better to match on this and then go from there.

mlutze:
> Tests are for suckers? :D

I was only following convention. :P We have no tests for Bimap, ListMap, Multimap. But I guess I will be better

> Also, I think we want a fast `toList` or something like that.

We get it for free from implementing `Iterable`

> Thoughts on chain of chains? e.g. another `ManyMany(xs: Seq[Chain[t]])`?

It shall be done.

> What about an overloaded apply, so one can write `Chain(x, y, z, w)`. That'd be nice.

It shall be done.

magnus-madsen:
> We get it for free from implementing Iterable

Yes, but is it efficient? You will construct N iterators. Can we not do better?

mlutze:
> > We get it for free from implementing Iterable
> 
> Yes, but is it efficient? You will construct N iterators. Can we not do better?

Not sure. If we do CPS then we're constructing N closures. I propose we go with this and see if performance is an issue.

magnus-madsen:
> > > We get it for free from implementing Iterable
> > 
> > 
> > Yes, but is it efficient? You will construct N iterators. Can we not do better?
> 
> Not sure. If we do CPS then we're constructing N closures. I propose we go with this and see if performance is an issue.

Perhaps the depth is always shallow and you can do a straightforward right recursion?

magnus-madsen (approved):

Issue #905, opened 2019-11-08
Add Tailrec Annotation


magnus-madsen:
Add a `@tailrec` annotation like in Scala and verify that the function is tail recursive.

mlutze:
Would the validation go in the TailRec compilation phase?

magnus-madsen:
No- we have no checks after the Simplifier. It would probably go in phase after typing "close" to the redundancy phase. (It could even go in that phase, I guess?)

magnus-madsen:
I guess it could go in the TailRec phase- but that would violate the principle that after the Simplifier we start to erase information from the AST.

magnus-madsen:
OK; I guess putting it in TailRec makes sense-- I think the algorithm that is there pretty much matches what is required for the tailrec check.

mlutze:
What do we think about the idea of inverting the annotation semantics, so that we need to annotate a non-tail-recursive function instead? That way, the user has to acknowledge a "dangerous" construction

magnus-madsen:
> 
> 
> What do we think about the idea of inverting the annotation semantics, so that we need to annotate a non-tail-recursive function instead? That way, the user has to acknowledge a "dangerous" construction

I think this is a very interesting idea. What I wonder is how many functions will need to be marked as non-tail recursive. If it is the majority then its probably not a good idea. 

Sometime in the fall @racerand is planning to look at implementing tail recursion modulo cons. As part of that work, it might be interesting for him to investigate the distribution of tail recursive vs. non-tail recursive functions.

magnus-madsen:
Wrong button

magnus-madsen:
Something that I just realized: A function can be tail recursive and still cause a stackoverflow (i.e. by having a call before the tail call that overflows the stack). 

This means that the "inverse" of tail recursive is NOT "being stack unsafe". You can be stack unsafe even if you are tail recursive. 

Proving that a function is tail recursive or tail recursive modulo cons is easy, but proving that a call cannot overflow the stack is difficult because of higher-order functions and mutual recursion.

mlutze:
Could stack safety be modeled as part of the type of a function, in a way similar to effects? That could address the higher-order function issue. Doesn't help the mutual recursion though, which I think may be in Halting Problem territory?

magnus-madsen:
> 
> 
[...]

PR #6454, opened 2023-09-13
feat: add `@Tailrec` annotation

jaschdoc:
Closes https://github.com/flix/flix/issues/905


jaschdoc:
I'm unsure if these are in `NonTailPosition` or should just carry the `tailrec` flag `RestrictableChoose`, `RelationalChoose` but I've set it as `NonTailPosition` to be conservative.

jaschdoc:
@magnus-madsen what do you think of this as an initial draft? I still have to refactor it a bit and add documentation, but maybe you could check that I haven't missed anything crucial? :)

jaschdoc:
Note that this only checks function calls are tail calls and not that the function is actually tail recursive

main/src/ca/uwaterloo/flix/language/errors/SafetyError.scala
```diff
@@ -597,4 +597,44 @@ object SafetyError {
...
+      import formatter._
+      s"""${line(kind, source.name)}
+         |>> Function annotated with @Tailrec must be tail recursive.
+         |
+         |${code(loc, "A recursive call in non-tail position occurs here.")}
+         |
+         |""".stripMargin
+    }
+
+    override def explain(formatter: Formatter): Option[String] = Some({
```
jaschdoc, line 620: Not sure if we should have this explanation, but I included it, so we can discuss it.

main/src/ca/uwaterloo/flix/language/errors/SafetyError.scala
```diff
@@ -597,4 +597,44 @@ object SafetyError {
...
     )
   }
+
+  /**
+    * An error raised to indicate that a function marked with the `@Tailrec` annotation
+    * has at least non-tail-recursive function call.
+    *
+    * @param loc the location of the non-tail-recursive call.
+    */
+  case class NonTailRecursiveFunction(loc: SourceLocation) extends SafetyError {
```
jaschdoc, line 607: Maybe also include the name of the function marked annotated with `@Tailrec` in the parameter list.

jaschdoc:
@magnus-madsen the current implementation works, but I think a review is in order.

magnus-madsen:
To demonstrate that it works, could you annotate List.flix?

main/src/ca/uwaterloo/flix/language/errors/SafetyError.scala
magnus-madsen: In both the explanation and error message I think it is important to understand that tail recursion (despite the name) has nothing to do with tail calls!

In fact, the point is simply that if a function calls another function in tail position then the stack does not grow.

jaschdoc:
> I don't suppose we can annotate inner defs/letrec? or?

I can look into it

magnus-madsen:
I just realized that there is terminology confusion about tail recursion vs. calls being in tail position. 

In other words, we have to begin by defining what we mean.

For example, can we annotate this program or not?

/// Flix, despite being a JVM-language,
/// supports full tail call elimination.

/// We can demonstrate this with a naive implementation
/// of a program that computes if a number is odd.

```
/// Returns true if n is odd.
def isOdd(n: Int32): Bool =
    if (n == 0) false else isEvn(n - 1)

/// Returns true if n is even.
def isEvn(n: Int32): Bool =
    if (n == 0) true else isOdd(n - 1)

/// We can now compute if 12345 is odd.
/// In a language without TCE this would
/// quickly consume all stack space.
def main(): Unit \ IO =
    isOdd(12345) |> println
```

jaschdoc:
> I just realized that there is terminology confusion about tail recursion vs. calls being in tail position.
> 
> In other words, we have to begin by defining what we mean.
> 
> For example, can we annotate this program or not?
> 
> /// Flix, despite being a JVM-language, /// supports full tail call elimination.
> 
> /// We can demonstrate this with a naive implementation /// of a program that computes if a number is odd.
> 
> ```
> /// Returns true if n is odd.
> def isOdd(n: Int32): Bool =
>     if (n == 0) false else isEvn(n - 1)
> 
> /// Returns true if n is even.
> def isEvn(n: Int32): Bool =
>     if (n == 0) true else isOdd(n - 1)
> 
> /// We can now compute if 12345 is odd.
> /// In a language without TCE this would
> /// quickly consume all stack space.
> def main(): Unit \ IO =
>     isOdd(12345) |> println
> ```

Currently, no, since the implementation specifically disallows non-recursive functions, i.e., there has to be at least one recursive call in the function, and each recursive call must be in tail position for compilation to be successful with the annotation.

jaschdoc:
It can easily be disabled, though. It just figured I'd implement what Scala has and then we could discuss from there.

magnus-madsen:
Maybe two annotations? Names?

magnus-madsen:
Maybe:

```
@StackSafe
@TailRecursive
```

jaschdoc:
I guess the current annotation is nice to have as Scala devs will be familiar with it. However, there could potentially be two other interesting annotations:

1. A lax form of tail recursion: this only has a restriction that all directly recursive calls are in tail position. It doesn't guarantee that there are any recursive calls at all.

2. A strict form of tail recursion which takes into account the fact that we have full tail call elimination: this annotation would require that all function calls (also to other functions) are also checked for tail recursion. This would completely guarantee that this function would not grow the stack more than a constant amount. Does that make sense?

magnus-madsen:
Yes, but 2. is difficult (impossible) in the presence of higher-order functions, mutable memory, etc. etc.

jaschdoc:
> Maybe:
> 
> 
> 
> ```
> 
> @StackSafe
> 
> @TailRecursive
> 
> ```

So `@StackSafe` would be the relaxed version of `@Tailrec` which just guarantees that if there is a recursive call, it is in tail position?

magnus-madsen:
I am not sure. Still thinking about it.

Whats your though on this program:

```
def stupidFib() = 
  if (blah) stupidFib() else {
    foo();
    0
  }
```

The function is tail recursive, but what about the fact that it calls foo? That could blow the stack.

jaschdoc:
> I am not sure. Still thinking about it.
> 
> Whats your though on this program:
> 
> ```
> def stupidFib() = 
>   if (blah) stupidFib() else {
>     foo();
>     0
>   }
> ```
> 
> The function is tail recursive, but what about the fact that it calls foo? That could blow the stack.

Hmm Scala allows this one
```scala
import scala.annotation.tailrec

object Main {
  def main(args: Array[String]): Unit = {
    println(stupidFib(10, 0))
  }

  def other(): Unit = println("ay")

  @tailrec
  def stupidFib(n: Int, acc: Int): Int =
    if (n >= 0) stupidFib(n - 1, acc + n) else {
      other();
      acc
    }
}
```

I guess Scala's assumption is that other functions are allowed to do anything, but at least the annotated function won't blow the stack any more than a constant factor, i.e., the memory used by the annotated function is determined by the “foreign” functions it calls.

In your example, `foo` would have an effect (otherwise it's a pure unused result), and can we give any guarantees on effectful functions and stack safety? Maybe both `@Tailrec` and `@StackSafe` require the functions to be pure? Then we could have another annotation `@EffectfulTailRec` which is like Scala's annotation?

jaschdoc:
This fails with

```flix
    @Tailrec
    pub def foldRightWithCont(f: (a, Unit -> b \ ef) -> b \ ef, z: b, l: List[a]): b \ ef = match l {
        case Nil     => z
        case x :: xs => f(x, _ -> foldRightWithCont(f, z, xs))
    }
```

```
-- Safety Error -------------------------------------------------- List.flix

>> Function 'List.foldRightWithCont' annotated with @Tailrec can only have recursive calls in tail position.

665 |         case x :: xs => f(x, _ -> foldRightWithCont(f, z, xs))
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                        A recursive call in non-tail position occurs here.
```
What do we think of this? Is it ok? Should it be rejected at all? Should it fail with `Does not contain a recursive call`? If we don't reject it, you could get around the tailrec annotation with
`def f(...) = ... let g = _ -> f(...); g()`

magnus-madsen:
So, what does the Scala annotation do? I suspect it:

- Requires the function to call itself recursively at least once.
- Ensures that all calls to the function itself occurs in tail position.

Does that sound right?

Next question, what if we did the "opposite":

- For any function, if it calls itself recursively, then it must be in a tail position.

If a function wants to call itself recursively, in a non-tail position, then it must be marked with StackUnsafe.

Could we implement that "opposite" approach and count how many functions in the standard library break its?

Alternatively to implementing it, could we do the count manually for e.g. List?

magnus-madsen:
I think we are right to reject foldRIghtWithCont. It does not call itself tail recursively.

jaschdoc:
> Yes, but 2. is difficult (impossible) in the presence of higher-order functions, mutable memory, etc. etc.

What if we require that all related functions are also annotated with the same strict annotation?

jaschdoc:
> So, what does the Scala annotation do? I suspect it:
> 
>     * Requires the function to call itself recursively at least once.
> 
>     * Ensures that all calls to the function itself occurs in tail position.
> 
> 
> Does that sound right?

Correct (from what I know).

> 
> Next question, what if we did the "opposite":
> 
>     * For any function, if it calls itself recursively, then it must be in a tail position.
> 
> 
> If a function wants to call itself recursively, in a non-tail position, then it must be marked with StackUnsafe.
> 
> Could we implement that "opposite" approach and count how many functions in the standard library break its?
> 
> Alternatively to implementing it, could we do the count manually for e.g. List?

I like this idea. Safe is the default, and you have to opt out of safety.

Another possibility is that we are even stricter with what `@Tailrec` allows: it is **only** allowed to call itself (no other function) and all calls must be in tail position.

magnus-madsen:
The safe is default is too restrictive (thanks Troels). I will ponder this some more.

magnus-madsen:
OK. Forget everything I said so far.

I think the most important thing is that for the feature to be useful we need to be able to annotate and check local defs. 

That way, we can annotate most of `List` and have it checked.

magnus-madsen:
> Now that we have annotations for letrec, should I annotate the `List` functions (and letrec) and see what we get?

I guess doing it to a few at least would be useful.

magnus-madsen:
The priority for now is on Desugaring, but it would be reasonable to get this PR to a state that passes all tests.

Do we know what the problem is? (Maybe it is related to currying (?))

jaschdoc:
> The priority for now is on Desugaring, but it would be reasonable to get this PR to a state that passes all tests.
> 
> Do we know what the problem is? (Maybe it is related to currying (?))

Still investigating the problem, but I'll let it sit for now.

PR #6453, opened 2023-09-13
refactor: split Typer into Typer and TypeReconstruction

mlutze:
related to #6448
depends on #6469

PR #6457, opened 2023-09-14
refactor: remove inferred scheme

mlutze:
should help with #6453


magnus-madsen:
@stephentetley We are making several changes to the compiler infrastructure over the next few weeks. None of these should impact end-user programs, but let us know if you observe any regressions. (Hopefully there will be none with all our tests).

magnus-madsen (approved):

stephentetley:
Thanks for the heads up Magnus.

PR #6461, opened 2023-09-15
feat: add links between types in HTML documentation

### LINKED ISSUE #6442: HtmlDocumenter: Add links between entities
If List.head returns an Option then it should link to Option.

sockmaster27:
Fixes #6442

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/fmt/FormatType.scala
+++ b/main/src/ca/uwaterloo/flix/language/fmt/FormatType.scala
@@ -37,6 +37,15 @@ object FormatType {
     formatTypeWithOptions(renamed, flix.getFormatOptions)
   }
 
+  def formatTypeMappingNames(tpe: Type)(mapName: SimpleType.Name => String)(implicit flix: Flix): String = {
+    try {
+      val fmt = flix.getFormatOptions
+      format(SimpleType.fromWellKindedType(tpe)(fmt))(mapName)(fmt)
+    } catch {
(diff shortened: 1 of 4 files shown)


sockmaster27:
I'm not quite sure how this would be done the most cleanly. Right now I've modified `FormatTypes` to introduce a new function called `formatTypeMappingNames`.

magnus-madsen:
We cannot make changes to formatType-- so indeed we are stuck. 

Let me think about it...

magnus-madsen:
(meanwhile, I would like the refactoring of "modules" to be the top priority.)

sockmaster27:
> We cannot make changes to formatType-- so indeed we are stuck.
> 
> Let me think about it...

Have you looked at the code? I don't think it's the best solution ideologically, but I don't see how we are ever going to get anything better, other than just effectively having a copy of the functionality.

magnus-madsen:
> > We cannot make changes to formatType-- so indeed we are stuck.
> > Let me think about it...
> 
> Have you looked at the code? I don't think it's the best solution ideologically, but I don't see how we are ever going to get anything better, other than just effectively having a copy of the functionality.

Yes, but I am not sure we can pollute compiler code with tooling code. Imagine we did this for every tool that needs something.

magnus-madsen:
But as I said, let me think about it. Maybe I will end up where you are :)

sockmaster27:
@magnus-madsen 
Would it help if I called it a variation of the strategy pattern? 😅

magnus-madsen:
Will revisit in the future.

PR #6469, opened 2023-09-18
fix: give unique IDs to instance definitions

mlutze:
related to #6453 and #6448


mlutze:
> This makes me nervous. Why is this needed?

We are making a map `def -> subst`. The instance defs collide. We can't differentiate them since they don't have namespace information the way that other defs do.

The real solution is to make a namespace for each instance and do magic to make name resolution work as expected. But this is a minimal solution that unblocks the Typer refactoring.

magnus-madsen (approved):

PR #6474, opened 2023-09-21
feat: effects in ReducedAst

JonathanStarup:
Add `Do` `TryWith` and `Resume`


magnus-madsen (approved):

magnus-madsen:
Thanks!

magnus-madsen:
Next, we need effects in ReducedAst, so we can generate interfaces for them?

PR #6475, opened 2023-09-21
feat: pretty print def effects

JonathanStarup:


main/src/ca/uwaterloo/flix/language/dbg/DocAst.scala
```diff
@@ -358,6 +358,18 @@ object DocAst {
     def Var(id: Int): Type = AsIs(s"var$id")
   }
 
+  sealed trait Effect
```
magnus-madsen, line 361: This should probably be called Eff

magnus-madsen: or maybe even better Purity.

JonathanStarup: It will later also capture things like `r1 + r2` so I pick Eff

magnus-madsen (approved):

Issue #6459, opened 2023-09-14
Remove lowercase annotations


jaschdoc:
Discussed in https://github.com/flix/flix/discussions/6449

jaschdoc:
I think so. @magnus-madsen  never really decided on one or the other

magnus-madsen:
Not really. I need to make a decision and then we need to implement it.

magnus-madsen:
Fixed in master.

PR #6458, opened 2023-09-14
style: remove lowercase annotations

jaschdoc:
Closes https://github.com/flix/flix/issues/6459

PR #6477, opened 2023-09-21
style: remove uppercase annotations

jaschdoc:
See https://github.com/flix/flix/pull/6458

Closes https://github.com/flix/flix/issues/6459


magnus-madsen:
Will revisit in the future.

PR #6486, opened 2023-09-24
feat: display associated types in HTML documentation

### LINKED ISSUE #6363: Add support for associated types to `HtmlDocumentor`
The Flix Standard Library does not yet use associated effects, but here is an example:

```
class Associative[m: Type] {
    type Key: Type
    type Value: Type
    pub def entryAt(coll: m, k: Associative.Key[m]): Associative.Value[m]
}
```

What I would like is for the HtmlDocumenter to also have a group of "associated types and effects" (before signatures) which shows these (i.e. Key and Value here along with their kinds). 

~~We should also check that the `entryAt` function is printed reasonably.~~ (EDIT: They are printed fine already).

sockmaster27:
Fixes #6363


magnus-madsen:
screenshot? :)

sockmaster27:
@magnus-madsen 
image

magnus-madsen (approved):

PR #6489, opened 2023-09-24
feat: add link to class in type constraint in HTML documentation

sockmaster27:
Related to #6442


magnus-madsen (approved):

magnus-madsen:
I notice that with clauses for enums are not hyperlinked.

See for example Files.FileType.

PR #6506, opened 2023-09-27
refactor: change Files.flix to use IOError on failure (issue ##6501)

stephentetley:
This PR changes `Files.flix` to use `IOError` for failure and adds more class derivings to the `IOError` type. `TestFiles.flix` did not need updating.

In `Files`, functions with simple control flow (no internal reference to `Ok` / `Err`) I've changed to use IOError.tryCatch otherwise I've kept the implementations the same.


mlutze (approved):

magnus-madsen:
Thanks Stephen

PR #6511, opened 2023-09-30
feat: show companion modules under modules in menu in HTML documentation

### LINKED ISSUE #6484: HtmlDocumenter: Menu should include classes, effects, types, etc. as modules
image

sockmaster27:
Still not happy about it but here you go

Fixes #6484


magnus-madsen:
!!!jar

magnus-madsen (approved):

magnus-madsen:
Thanks!

magnus-madsen:
> Still not happy about it but here you go

We can revisit at some point in the future.

PR #6512, opened 2023-09-30
feat: use `ParOps.mapValues` in `Lowering`

sockmaster27:
Related to #2974

PR #6514, opened 2023-09-30
chore: improve performance of `ParOps.mapValues`

sockmaster27:
Related to #2974

### DIFF
--- a/main/src/ca/uwaterloo/flix/util/ParOps.scala
+++ b/main/src/ca/uwaterloo/flix/util/ParOps.scala
@@ -18,9 +18,9 @@ package ca.uwaterloo.flix.util
 
 import java.util
 import java.util.concurrent.{Callable, Executors}
-
 import scala.jdk.CollectionConverters._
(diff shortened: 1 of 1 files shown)


sockmaster27:
This is just here to document this failed experiment. 

With the changes introduced to the Lowering phase in #6512, this change increases time spent in this phase from ~100ms to ~230ms :(

I'm guessing this is caused by the many conversions which are done internally anyway: immutable map -> iterator -> parallel map -> mutable map -> iterator -> immutable map.

magnus-madsen:
Indeed. Also this does not respect `--threads` nor the threadpool we use.

sockmaster27:
> Indeed. Also this does not respect `--threads` nor the threadpool we use.

Really? What makes it different from `parMap`?

sockmaster27:
> > Indeed. Also this does not respect `--threads` nor the threadpool we use.
> 
> Really? What makes it different from `parMap`?

@magnus-madsen ?

PR #6516, opened 2023-10-01
fix: ignore enums prefixed with `_` in Redundancy

sockmaster27:


mlutze:
Can we get a test for this?

magnus-madsen:
I would like a description of how this changes the language.

sockmaster27:
> Can we get a test for this?

I don't see any tests for the other cases like tags, defs and effects. There are the `HiddenVarSym` ones that ensure that formal parameters and local variables can't be used if they are prefixed, but not that they aren't marked as redundant (at least directly).

Should these tests be added for all cases?

sockmaster27:
> I would like a description of how this changes the language.

This makes enum names consistent with other symbols checked by the Redundancy phase, in that they can be prefixed with an underscore to ignore an 'Unused enum' error.

Doing this will also ignore all the enums unused tags.

In the same way that it was already the case with effects, trying to use a prefixed enum as a type fails with a parsing error: 
image

Trying to instantiate it currently gives a resolution error:
image

magnus-madsen:
We need tests. At least for new code. Ideally for all code :)

sockmaster27:
Should I add tests for the other cases in a new PR?

magnus-madsen:
> Should I add tests for the other cases in a new PR?

Yes, but let us finish with HtmlDocumentor first.

magnus-madsen (approved):

PR #6520, opened 2023-10-01
feat(Intern Assesment): extend the Flix compiler to support auto-completion of the "new object" construct

dghosef:
suggest body-less methods when autocompleting new objects.


magnus-madsen:
!!!jar

PR #6521, opened 2023-10-02
fix: add all type parameters to rigidity environment

mlutze:


magnus-madsen:
TLDR?

mlutze:
> TLDR?

With a def like

```
def foo(): a  \ ef = ???
```

The `a` and `ef` are not added to the rigidity environment. That makes them act as free variables. In some cases this is caught later by generalization check, but it does not seem always to be the case for some reason (see the schema test that had to change).

magnus-madsen (approved):

PR #6522, opened 2023-10-03
feat: tweak `Token`, `TokenKind`, and `TokenErrorKind`

herluf-ba:
Small updates to Token, TokenKind and TokenErrorKind as per https://github.com/flix/flix/pull/6350?notification_referrer_id=NT_kwDOAPFTIbM3Mzk0MzA3ODM1OjE1ODE1NDU3#pullrequestreview-1654486186

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/Token.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/Token.scala
@@ -15,4 +15,23 @@
  */
 package ca.uwaterloo.flix.language.ast
 
-case class Token(kind: TokenKind, text: String, line: Int, col: Int)
+/**
+ * A Token holding a kind, line and column.
+ * Token does not hold its lexeme directly to avoid duplication of common keywords like "def".
+ * Instead it holds a pointer to its source along with start and end offsets.
+ *
+ * @param kind  The kind of token this instance represents
+ * @param src   A pointer to the source that this lexeme stems from
+ * @param start The absolute character offset into `src` of the beginning of the lexeme
+ * @param end   The absolute character offset into `src` of the end of the lexeme
+ * @param line  The line that the lexeme __starts__ on
+ * @param col   The column that the lexeme __starts__ on
+ */
+case class Token(kind: TokenKind, src: Array[Char], start: Int, end: Int, line: Int, col: Int) {
+  /**
+   * Computes the lexeme that the token refers to by slicing it from `src`.
+   */
+  lazy val text: String = src.slice(start, end).mkString("")
+
+  override def toString: String = s"Token($kind, $text, $line, $col)"
+}
--- a/main/src/ca/uwaterloo/flix/language/ast/Token.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/Token.scala
@@ -17,21 +17,27 @@ package ca.uwaterloo.flix.language.ast
 
 /**
  * A Token holding a kind, line and column.
+ *
  * Token does not hold its lexeme directly to avoid duplication of common keywords like "def".
  * Instead it holds a pointer to its source along with start and end offsets.
  *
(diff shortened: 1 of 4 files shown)


magnus-madsen:
I see some compiler errors. You may have to slightly tweak things.

main/src/ca/uwaterloo/flix/language/ast/Token.scala
```diff
@@ -15,4 +15,23 @@
  */
 package ca.uwaterloo.flix.language.ast
 
-case class Token(kind: TokenKind, text: String, line: Int, col: Int)
+/**
+ * A Token holding a kind, line and column.
```
magnus-madsen, line 19: Add an extra line break.

main/src/ca/uwaterloo/flix/language/ast/Token.scala
```diff
@@ -15,4 +15,23 @@
...
+ * Token does not hold its lexeme directly to avoid duplication of common keywords like "def".
+ * Instead it holds a pointer to its source along with start and end offsets.
+ *
+ * @param kind  The kind of token this instance represents
```
magnus-madsen, line 23: Each comment should be a sentence, so end with a period.

main/src/ca/uwaterloo/flix/language/ast/Token.scala
```diff
@@ -15,4 +15,23 @@
...
+ * @param src   A pointer to the source that this lexeme stems from
+ * @param start The absolute character offset into `src` of the beginning of the lexeme
+ * @param end   The absolute character offset into `src` of the end of the lexeme
+ * @param line  The line that the lexeme __starts__ on
+ * @param col   The column that the lexeme __starts__ on
+ */
+case class Token(kind: TokenKind, src: Array[Char], start: Int, end: Int, line: Int, col: Int) {
+  /**
+   * Computes the lexeme that the token refers to by slicing it from `src`.
```
magnus-madsen, line 32: Add comment:

Note: This is explicitly lazy because we do not want to compute strings that are never used.

main/src/ca/uwaterloo/flix/language/ast/Token.scala
```diff
@@ -15,4 +15,23 @@
...
+   */
+  lazy val text: String = src.slice(start, end).mkString("")
+
+  override def toString: String = s"Token($kind, $text, $line, $col)"
```
magnus-madsen, line 36:
```
/**
 * Returns a string representation of this token. Should only be used for debugging.
 */
```

main/src/ca/uwaterloo/flix/language/ast/TokenKind.scala
```diff
@@ -95,6 +97,8 @@ object TokenKind {
 
   case object InfixFunction extends TokenKind
 
+  case object KeywordAbsent extends TokenKind
```
magnus-madsen, line 100: I believe we have removed Absent, Present, and relational_choose and relational_choose_star.

magnus-madsen:
Minor tweaks then ready.

main/src/ca/uwaterloo/flix/language/ast/Token.scala
herluf-ba: That's not necessarily true? The compiler needs to look at text content of tokens like literals and names to know their values later in the pipeline?

herluf-ba: I wrote this now:
```scala
  /**
   * Computes the lexeme that the token refers to by slicing it from `src`.
   * Note: This is explicitly lazy since we do not want to compute strings that are never used,
   * such as keywords for instance. Take care to only access text when necessary.
   */
```

magnus-madsen (approved):

PR #6523, opened 2023-10-03
feat: add  LexerError.StringInterpolationTooDeep

herluf-ba:
Small updates to LexerError.scala  as per https://github.com/flix/flix/pull/6350?notification_referrer_id=NT_kwDOAPFTIbM3Mzk0MzA3ODM1OjE1ODE1NDU3#pullrequestreview-1654486186

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/errors/LexerError.scala
+++ b/main/src/ca/uwaterloo/flix/language/errors/LexerError.scala
@@ -18,7 +18,6 @@ package ca.uwaterloo.flix.language.errors
 import ca.uwaterloo.flix.language.CompilationMessage
 import ca.uwaterloo.flix.language.ast.SourceLocation
 import ca.uwaterloo.flix.util.Formatter
-import ca.uwaterloo.flix.language.phase.Lexer.{InterpolatedStringMaxNestingLevel, BlockCommentMaxNestingLevel}
 
 sealed trait LexerError extends CompilationMessage {
   val kind = "Lexer Error"
@@ -43,10 +42,7 @@ object LexerError {
          |""".stripMargin
     }
 
-    override def explain(formatter: Formatter): Option[String] = Some({
-      import formatter._
-      s"${underline("Tip:")} Ensure that block-comments are not nested more than $BlockCommentMaxNestingLevel levels deep."
-    })
+    override def explain(formatter: Formatter): Option[String] = None
   }
 
   /**
@@ -104,7 +100,7 @@ object LexerError {
     override def message(formatter: Formatter): String = {
       import formatter._
       s"""${line(kind, source.name)}
-         |>> Unterminated block-comment.
+         |>> Missing '*/' in block-comment.
          |
          |${code(loc, "Block-comment starts here.")}
          |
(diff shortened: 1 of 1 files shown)


magnus-madsen:
I see some compiler errors. You may have to slightly tweak things.

main/src/ca/uwaterloo/flix/language/errors/LexerError.scala
```diff
@@ -18,6 +18,7 @@ package ca.uwaterloo.flix.language.errors
 import ca.uwaterloo.flix.language.CompilationMessage
 import ca.uwaterloo.flix.language.ast.SourceLocation
 import ca.uwaterloo.flix.util.Formatter
+import ca.uwaterloo.flix.language.phase.Lexer.{InterpolatedStringMaxNestingLevel, BlockCommentMaxNestingLevel}
```
magnus-madsen, line 21: Perhaps its better to break this dependency. And simply say its too deep.

We don't want programmers to necessarily know such internal limits.

main/src/ca/uwaterloo/flix/language/errors/LexerError.scala
```diff
@@ -44,7 +45,7 @@ object LexerError {
 
     override def explain(formatter: Formatter): Option[String] = Some({
       import formatter._
-      s"${underline("Tip:")} Ensure that block-comments are not nested more than 32 levels deep."
+      s"${underline("Tip:")} Ensure that block-comments are not nested more than $BlockCommentMaxNestingLevel levels deep."
```
magnus-madsen, line 48: So drop the tip.

main/src/ca/uwaterloo/flix/language/errors/LexerError.scala
```diff
@@ -196,5 +197,49 @@ object LexerError {
...
+   *
+   * @param loc The location of the opening `{`.
+   */
+  case class UnterminatedStringInterpolation(loc: SourceLocation) extends LexerError {
+    override def summary: String = s"Unterminated string interpolation."
+
+    override def message(formatter: Formatter): String = {
+      import formatter._
+      s"""${line(kind, source.name)}
+         |>> Unterminated string interpolation.
```
magnus-madsen, line 212: Something about missing end brance?

main/src/ca/uwaterloo/flix/language/errors/LexerError.scala
herluf-ba: Hmm, If I am a user of Flix and I get this error my first thought would be "well how deep can it go then?".
My thought is that the allowed nesting level is necessary information to fix the error.

Should we really remove it from the message?

magnus-madsen: If you need more than 32  nesting depth something is deeply wrong.

magnus-madsen: Like then you should talk to the compiler developers because you are doing something very strange :)

herluf-ba: Okay, that's fair. I will remove it 👍

main/src/ca/uwaterloo/flix/language/errors/LexerError.scala
herluf-ba: I think all the `Unterminated*` errors would be better off with a `Missing '_' in ____`, so they are all similar.
I already made that change.

magnus-madsen (approved):

PR #6526, opened 2023-10-04
Bug: Out-Of-Bounds errors in `Lexer`

herluf-ba:
This is a fix for out of bounds errors occurring in the new `Lexer`. 
Two things were done:
1. The bug was fixed by being more careful in the `advance`  function
2. A `tryLex` wrapper function was introduced to prevent crashing the compiler if/when something similar is found.


herluf-ba:
Closing and reopening with a clean branch for cleaner git history

PR #6527, opened 2023-10-04
bug: fix OOB errors in `Lexer` and added `tryLex`

herluf-ba:
This is a fix for out of bounds errors occurring in the new Lexer.
Two things were done:
1. The bug was fixed by being more careful in the `advance` function
2. A `tryLex` wrapper function was introduced to prevent crashing the compiler if/when something similar is found.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
@@ -35,12 +35,12 @@ object Lexer {
   /**
    * The maximal allowed nesting level of block-comments.
    */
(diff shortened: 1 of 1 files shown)


herluf-ba:
I am not sure I understand what is actually checked in the `semantic commit message` check, even though I did read through the semantic commits specification.

What is wrong with the title in this PR?

magnus-madsen:
Thanks-- but in the future let us try to keep each PR to one unit of work (i.e. no bug fix together with other stuff) :)

magnus-madsen:
Time for some more testing :)

PR #6529, opened 2023-10-04
chore: simpler error handling in `Lexer`

herluf-ba:
Simplifies error handling in the new `Lexer` by removing `TokenErrorKind`. Instead `LexerError`s are now contained directly within a `TokenKind.Err`.


herluf-ba:
This has the great benefit of removing the function `tokenErrToCompilationMessage` which even felt like a mistake when writing it.

herluf-ba:
@magnus-madsen Do you have time to look this over?

magnus-madsen (approved):

magnus-madsen:
I think this is a good simplification. Top marks 👍

PR #6534, opened 2023-10-04
feat: Add Resumption classes

JonathanStarup:
* rename `unwindThunk` -> `unwindThunkToType`
  * Add new `unwindThunk` that just unwinds, and use it in `Thunk.run`
* Add invokedynamic and related mechanisms to the framework in order to support lambdas
  * Introduce `LambdaFactory` types and other dynamic call related classes
* Add `Resumption`, `ResumptionCons`, and `ResumptionNil`]
  * Add temp `Handler` for use in the `rewind` function
* Corrected example java files to reflect current design

None of the new classes are used in generated code


magnus-madsen:
Does this change any semantics? Is it safe to merge?

magnus-madsen (approved):

PR #6535, opened 2023-10-04
feat: add annotations to LetRec

jaschdoc:
The plan is to also check `@Tailrec` for LetRec when we figure out the direction for that annotation.

Related to https://github.com/flix/flix/pull/6454


magnus-madsen:
Maybe we should break these PRs into smaller pieces that can be merged.

jaschdoc:
Continued in other PR

PR #6542, opened 2023-10-08
feat: show all items as submodules in HTML documentation

### LINKED ISSUE #6484: HtmlDocumenter: Menu should include classes, effects, types, etc. as modules

sockmaster27:
Fixes #6484


magnus-madsen (approved):

magnus-madsen:
@sockmaster27 Does this conclude everything we were planning to do with HtmlDocumenter?

So we can leave it on the shelf for a while and move onto other things?

sockmaster27:
> @sockmaster27 Does this conclude everything we were planning to do with HtmlDocumenter?
> 
> So we can leave it on the shelf for a while and move onto other things?

Yes, I believe so

PR #6543, opened 2023-10-08
fix: differentiate uses of item's name and qualified name in HTML documentation

### LINKED ISSUE #6484: HtmlDocumenter: Menu should include classes, effects, types, etc. as modules

sockmaster27:
Fixes #6484


magnus-madsen:
Screenshot of e.g. `Chain` module?

sockmaster27:
image

sockmaster27:
Ah - One thing I need to fix first:
image

magnus-madsen (approved):

Issue #6525, opened 2023-10-03
Improvements to Lexer


magnus-madsen:
In order of importance:

- [x] Index out of bounds (e.g. while typing something with block comments) -- Very high priority.
- [x] Test cases for lexer errors.
- [x] Error messages are off-by-one? E.g. unterminated string.
- [x] Incremental lexing with changeSet.
- [x] https://github.com/flix/flix/issues/6524
- [x] Fix Intellij IDEA warnings.

herluf-ba:
example of index out of bounds error:
```
enum Foo {
    case F { /*
}
```

Solution for now: catch exceptions in the lexer and print them to stdout

magnus-madsen:
And fix the bugs:)

magnus-madsen:
@herluf-ba I think we also did ourselves a disservice by not writing explicit negative test cases for the corner cases in lexing block comments etc.

herluf-ba:
> And fix the bugs:)

Yes of course 👍 

> @herluf-ba I think we also did ourselves a disservice by not writing explicit negative test cases for the corner cases in lexing block comments etc.

Agreed. I was thinking about how we could have avoided this situation, and I also landed on negative testing. Resiliency is an extra new feature and by "just" testing with well-formed code from the standard library the resiliency part never gets tested properly.

magnus-madsen:
Exactly right! We learned something. Put it in the report :)

magnus-madsen:
@herluf-ba A few minor bugs (?):

- The file consisting of just a single " does not produce a lexer error.
- In this program: `def f(): String = "` the source location is right, but no source code is shown.
- This program `def f(): String = "/**/"` (on a single line, does not lex, but it should).

herluf-ba:
Thanks for these cases, Ill look in to them now 👍

herluf-ba:
Hmm something is weird here. I ran through all of these examples but I am getting wildly different results:
- `"` -> Gives an `UnterminatedString`. I added a test case for this too (LexerError.UnterminatedString.03)
- `def f(): String = "` -> The source code is now shown for any `LexerError`. This is a known bug that I am working on.
- `def f(): String = "/**/"` does lex for me.
- `*/` contains valid tokens as far as the lexer is concerned. This is a parser error.
- `def f(): String = "\${"` should and does indeed lex. The dollar sign is escaped so this is actually not a string interpolation.
[...]

PR #6548, opened 2023-10-10
feat: use `ChangeSet` in `Lexer`

herluf-ba:
This PR makes the `Lexer` able to reuse tokens from sources that are still fresh when the compiler is run incrementally.


herluf-ba:
@magnus-madsen This should be the last actionable point from #6525 🙂

magnus-madsen (approved):

PR #6550, opened 2023-10-10
test: Additional test-case for `Lexer`

herluf-ba:


main/test/ca/uwaterloo/flix/language/phase/TestLexer.scala
```diff
@@ -175,6 +175,12 @@ class TestLexer extends AnyFunSuite with TestUtils {
     expectError[LexerError.UnterminatedString](result)
   }
 
+  test("LexerError.UnterminatedString.03") {
+    val input = """""""
```
magnus-madsen, line 179: Jesus christ. Maybe it would be more clear if we did not use a raw string?

magnus-madsen (approved):

PR #6551, opened 2023-10-10
fix: make TestFile.flix use the temp directory for creating outfile01.bin

stephentetley:
This PR updates TestFile.flix to write `outfile01.bin` to the temp directory identified by `Environment.getTemporaryDirectory` rather than the `{tests-path}/library/testdata` folder as was reported on Flightdeck.

As far as I can tell the answer from `Environment.getTemporaryDirectory` is already suffixed with a path separator so we don't need to add one directly.

Not done / should I do? - I don't delete `outfile01.bin` at the end of the tests. Should I write a test at the end of the file that does this?

### DIFF
--- a/main/test/ca/uwaterloo/flix/library/TestFile.flix
+++ b/main/test/ca/uwaterloo/flix/library/TestFile.flix
@@ -23,13 +23,28 @@ mod TestFile {
 
     ///
     /// Helper - make a filepath within the temp directory.
-    /// Assumes getTemporaryDirectory() returns a path suffixed with the separator.
+    /// Use `java.nio.file.Path` functions for platform independence.
     ///
-    def makeTempPath(fileName: String): Result[IOError, String] =
-        match Environment.getTemporaryDirectory() {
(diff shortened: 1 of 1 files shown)


magnus-madsen:
No need to delete anything from the temp directory. That should happen automatically (on process exit?)

stephentetley:
> No need to delete anything from the temp directory. That should happen automatically (on process exit?)

That doesn't seem to happen when running inside IntelliJ on Windows (outfile01.bin remains after closing IntelliJ)

stephentetley:
Uh oh, I didn't see failing test `setLength01` failing on Windows - I have to see what's going on.

stephentetley:
> Uh oh, I didn't see failing test `setLength01` failing on Windows - I'll have to see what's going on.

I think the problem was not due to not making a proper path on Linux. The root directory on Windows had a separator at the end so it worked with string concatenation but Linux needs the path forming properly.

magnus-madsen (approved):

magnus-madsen:
Thanks Stephen

PR #6552, opened 2023-10-11
fix: Show source code in `LexerError`s

herluf-ba:
Creates a `ParserInput` to be used in a `SourceLocation` when spawning `LexerErrors`.

The program `def f(): String = "${"` gives these errors for instance:
```
-- Lexer Error -------------------------------------------------- foo.flix

>> Missing '}' in string interpolation.

1 | def f(): String = "${"
                        ^
                        Interpolation starts here.


-- Lexer Error -------------------------------------------------- foo.flix

>> missing '"' in string.

1 | def f(): String = "${"
                         ^
                         String starts here.
```


herluf-ba:
This closes #6525

herluf-ba:
Oh, and there is no significant change to performance. Checking with `--Xbenchmark-phases` I get the exact same results as before the changes with the `Lexer` clocking in at `~1.3%`

magnus-madsen (approved):

PR #6553, opened 2023-10-11
fix: handle nested block-comments termniation within strings in `Lexer` 

herluf-ba:
From the discussion here https://github.com/flix/flix/discussions/6531 some tricky cases were found:
```
/* def example(): String = "*//" */
/* def example(): String = "//*" */
```
These should both lex to block comments.

This PR makes sure this is the case.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
@@ -875,20 +875,30 @@ object Lexer {
    */
   private def acceptBlockComment()(implicit s: State): TokenKind = {
     var level = 1
+    var isInString = false
     while (!eof()) {
       (peek(), peekPeek()) match {
+        case ('"', _) =>
+          isInString = !isInString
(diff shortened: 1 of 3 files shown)


magnus-madsen:
(1) Test cases?
(2) Do we really care about these corner cases?

herluf-ba:
I can add a couple of test cases for these. I think we do care, since I could see someone bump into them and be quite confused. Something like:
1. write some function `def explainBlockComments(): String = "Block comments start with /* "`
2. decide to comment out that code using a block comment.

In any case I think it is more consistent for the Lexer to behave this way, since it now has the invariant "block comment opening or termination means nothing within strings". Before this change this was not the case.

JonathanStarup:
I've added a counterargument in the discussion thread

herluf-ba:
It seems like, we won't be able to handle this case due to the counter-argument Jonathan mentions. 
This PR should probably not make it into `master`.

magnus-madsen:
Indeed.

PR #6554, opened 2023-10-12
feat: use levels to detect escaping regions

magnus-madsen:


magnus-madsen:
@mlutze There is one test that fails: `Test.Exp.ChooseStar.Polymorphic.flix` but that must almost be because of a bug in the level generation. Can you take a look? (I assume some wrong level is used somewhere).

magnus-madsen:
@mlutze Specifically, this line is incorrect:

https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/inference/RestrictableChooseInference.scala#L138

It calls `inferExp` directly. Any call to `inferExp` directly is wrong (except inside e.g. visitDef) because it implicitly assumes that the level is zero/top.

magnus-madsen:
What I learned from this PR:

- We can use levels to reject escaping regions.
- Updating levels is non-commutative.
- BU + Levels is tricky.

PR #6555, opened 2023-10-13
fix: explicitly pass initial level to `inferExp`

magnus-madsen:


mlutze (approved):
Could be implicit, no?

magnus-madsen:
> Could be implicit, no?

Maybe. But with visit there would be two implicit variables. I did not want to deal with that.

PR #6556, opened 2023-10-13
feat: use levels for variable ordering in SVE

magnus-madsen:


magnus-madsen:
Okay... Lots of unexpected test failures.

magnus-madsen:
I suspect purity information is wrong which causes the optimizer to break.

magnus-madsen:
This PR should probably be split into two.

PR #6557, opened 2023-10-13
feat: purify letrec

magnus-madsen:


magnus-madsen:
@mlutze This one I would like to merge. I am less sure about the variable ordering one.

magnus-madsen:
@mlutze I am pretty sure this avoids the `alpha -> alpha + rc` problem... For the current system...

main/src/ca/uwaterloo/flix/language/phase/unification/Unification.scala
```diff
@@ -444,6 +444,43 @@ object Unification {
     visit(liftM(fs.head), fs.tail)
   }
 
+  /**
+    * Purifies all effect variables in `tpe` (which should be the type of a letrec-bound function).
```
mlutze, line 448: But only the ones at the right level?

magnus-madsen: Good point. I thought about it too, but was unsure what to do. It's is probably correct to consider higher levels, but then again there would have been a call to this function for that level. So maybe not necessary. Only doing it for the same level seemed the safest.

magnus-madsen:
@mlutze request approval

mlutze (approved):

PR #6560, opened 2023-10-15
feat: add annotations to LetRec

jaschdoc:
Supersedes #6535

Related to https://github.com/flix/flix/pull/6454

Related to https://github.com/flix/flix/issues/905


jaschdoc:
Note to self:
- add code hint with `Tailrec` (later PR)
- add check in safety (see #6454)

main/src/ca/uwaterloo/flix/language/phase/Weeder.scala
```diff
@@ -1134,9 +1134,20 @@ object Weeder {
         case err: WeederError => WeededAst.Expr.Error(err)
       }
 
-    case ParsedAst.Expression.LetRecDef(sp1, ident, fparams, tpeAndEff, exp1, exp2, sp2) =>
+    case ParsedAst.Expression.LetRecDef(sp1, ann, ident, fparams, tpeAndEff, exp1, exp2, sp2) =>
       val mod = Ast.Modifiers.Empty
       val loc = mkSL(sp1, sp2)
+      val annVal = flatMapN(visitAnnotations(ann)) {
+        case Ast.Annotations(annotations) if annotations.nonEmpty =>
```
magnus-madsen, line 1141: I think we actually also check for duplicate annotations? Dont we?

jaschdoc: Yes, but that's handled by `visitAnnotations`

jaschdoc: I have a test for duplicate annotations

magnus-madsen:
Conflict, sorry.

jaschdoc:
> I see some inconsistency:
> 
>     * We add annotations to ASTs post Safety/Tailrec, where the annotations are not used. That would be OK,
> 
>     * But then we don't add modifiers to LetRec.
> 
> 
> So I think either drop annotations from backend ASTs or add modifiers too.
> 
> (I would suggest to drop the annotations from the backend Asts).

Okay great! I was unsure about this, but it didn't make sense to me to have annotations in the backend.

main/test/ca/uwaterloo/flix/language/phase/TestWeeder.scala
```diff
@@ -1207,4 +1207,43 @@ class TestWeeder extends AnyFunSuite with TestUtils {
     expectError[WeederError.IllegalRecordExtensionPattern](result)
   }
 
+  test("IllegalInnerFunctionAnnotation.01") {
+    val input =
+      """
+        |def f(): Int32 = {
+        | @Test
```
magnus-madsen, line 1214: Why only 1 space?

magnus-madsen (approved):

PR #6574, opened 2023-10-19
chore: remove `--Xno-set-effects` and `--Xno-bool-effects`

magnus-madsen:


mlutze (approved):

magnus-madsen:
@mlutze I did not touch the printing. I assume that is OK?

mlutze:
> @mlutze I did not touch the printing. I assume that is OK?

Yes, setting them to false is fine for now. We can remove those flags later.

magnus-madsen:
> > @mlutze I did not touch the printing. I assume that is OK?
> 
> Yes, setting them to false is fine for now. We can remove those flags later.

Can you create a ticket? I am not exactly sure what to state.

PR #6578, opened 2023-10-19
chore: remove `--Xvirtual-threads`

magnus-madsen:
(We don't want people to start depending on this until we are actually ready.)


mlutze:
I thought `X` means that you cannot depend on it?

magnus-madsen:
> I thought `X` means that you cannot depend on it?

The point is that the flag goes away.

magnus-madsen:
The flag existed so we could test something. Not as a feature in itself.

mlutze (approved):

PR #6580, opened 2023-10-19
chore: remove `--Xdebug`

magnus-madsen:


mlutze:
Is this one not handy still?

magnus-madsen:
> Is this one not handy still?

I don't see why. Its pretty arbitrary. When does it ever provide enough information?

Issue #6583, opened 2023-10-19
Replace dependence on `scala-parallel-collections` with ordinary Java


magnus-madsen:
The idea is to drop the use of `scala.collection.parallel._` and instead use ordinary Java threads/thread pools/fork join.

I recommend to start by changing the `build.gradle` file to remove the dependency so its easy to see what has to be changed.

magnus-madsen:
@sockmaster27 Once the PR with `parMapSeq` is in, you can look at this.

magnus-madsen:
@sockmaster27 Let me know if you need more information.

magnus-madsen:
Inspiration may be found in: https://github.com/flix/flix/pull/6584

sockmaster27:
> Inspiration may be found in: #6584

What's the reason for not just merging this?

magnus-madsen:
> > Inspiration may be found in: #6584
> 
> What's the reason for not just merging this?

We are not switching to Java 21 just yet.

magnus-madsen:
But in any case- that PR alone is not sufficient.

Also its not clear if VirtualThreads are even better than ordinary threads, so it would be nice to have an ordinary thread implementation.

magnus-madsen:
I think we are stuck on this. Let me look into it and report back.

magnus-madsen:
Fixed in master.

PR #6589, opened 2023-10-21
refactor: introduce `ParOps.parMapValuesSeq`

### LINKED ISSUE #2974: Add `ParOps.mapValues` for `Map[K, V]` and use it broadly

sockmaster27:
Fixes #2974

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
@@ -72,25 +72,17 @@ object Kinder {
     flatMapN(visitTypeAliases(root.taOrder, root)) {
       taenv =>
 
-        val enumsVal = ParOps.mapValuesFallible(root.enums) {
-          enum => visitEnum(enum, taenv, root)
-        }
+        val enumsVal = ParOps.mapValuesFallible(root.enums)(visitEnum(_, taenv, root))
 
-        val restrictableEnumsVal = ParOps.mapValuesFallible(root.restrictableEnums) {
-          enum => visitRestrictableEnum(enum, taenv, root)
-        }
+        val restrictableEnumsVal = ParOps.mapValuesFallible(root.restrictableEnums)(visitRestrictableEnum(_, taenv, root))
 
         val classesVal = visitClasses(root, taenv, oldRoot, changeSet)
 
         val defsVal = visitDefs(root, taenv, oldRoot, changeSet)
 
-        val instancesVal = ParOps.mapValuesFallible(root.instances) {
-          insts => traverse(insts)(visitInstance(_, taenv, root))
-        }
+        val instancesVal = ParOps.mapValuesFallible(root.instances)(traverse(_)(i => visitInstance(i, taenv, root)))
 
-        val effectsVal = ParOps.mapValuesFallible(root.effects) {
-          eff => visitEffect(eff, taenv, root)
-        }
+        val effectsVal = ParOps.mapValuesFallible(root.effects)(visitEffect(_, taenv, root))
 
         mapN(enumsVal, restrictableEnumsVal, classesVal, defsVal, instancesVal, effectsVal) {
           case (enums, restrictableEnums, classes, defs, instances, effects) =>
--- a/main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
@@ -72,39 +72,29 @@ object Kinder {
     flatMapN(visitTypeAliases(root.taOrder, root)) {
       taenv =>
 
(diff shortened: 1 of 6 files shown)


main/src/ca/uwaterloo/flix/util/ParOps.scala
```diff
@@ -52,6 +52,18 @@ object ParOps {
       case (k, v) => (k, f(v))
     }.toMap
 
+  /**
+    * Apply the given fallible function `f` to each value in the map `m` in parallel,
+    * returning the resulting map if all calls are successful.
+    */
+  @inline
+  def mapValuesFallible[K, A, B, E](m: Map[K, A])(f: A => Validation[B, E])(implicit flix: Flix): Validation[Map[K, B], E] = {
```
magnus-madsen, line 60: rename to `parMapSeq`

magnus-madsen:
Has this been applied to every phase where possible?

main/src/ca/uwaterloo/flix/util/ParOps.scala
sockmaster27: Wouldn't that make it sound like it corresponded to `parMap` instead of `mapValues`?

sockmaster27:
> Has this been applied to every phase where possible?

Yes, I believe so.

main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
```diff
@@ -72,39 +72,29 @@ object Kinder {
...
-            visitEnum(enum, taenv, root).map(sym -> _)
-        }))
-
-        // Extra type annotations are required due to limitations in Scala's type inference.
-        val restrictableEnumsVal = Validation.sequence(ParOps.parMap(root.restrictableEnums)({
-          pair: (Symbol.RestrictableEnumSym, ResolvedAst.Declaration.RestrictableEnum) =>
-            val (sym, enum) = pair
-            visitRestrictableEnum(enum, taenv, root).map(sym -> _)
-        }))
+        val enumsVal = ParOps.mapValuesFallible(root.enums) {
```
magnus-madsen, line 75: this  can become `....mapValues(root.enums)(visitEnum(_, taenv, root)`.

Same pattern can be used everywhere else.

main/src/ca/uwaterloo/flix/language/phase/Stratifier.scala
```diff
@@ -59,20 +59,16 @@ object Stratifier {
 
     // Compute the stratification at every datalog expression in the ast.
     val newDefs = flix.subphase("Stratify Defs") {
-      val result = ParOps.parMap(root.defs)(kv => visitDef(kv._2)(root, g, flix).map(d => kv._1 -> d))
-      Validation.sequence(result)
+      ParOps.mapValuesFallible(root.defs)(v => visitDef(v)(root, g, flix))
```
magnus-madsen, line 62: implicit can be omitted? (comment applies everywhere)

sockmaster27: If you mean `(root, g, flix)` then `root` and `g` aren't implicit.

magnus-madsen: Oh okay.

main/src/ca/uwaterloo/flix/util/ParOps.scala
magnus-madsen: Maybe parMapValuesSeq

sockmaster27: Should `mapValues` then also be renamed to `parMapValues`?

magnus-madsen: Ah, maybe without the par

sockmaster27: Actually, looking at the other functions, it looks like they should all just be called `par`-, right?

magnus-madsen (approved):

magnus-madsen:
Next up https://github.com/flix/flix/issues/6583

magnus-madsen:
I think there are still a few uses of `parMap` which should become `parMapSeq` (if you search for uses of the former). Will you check? Thanks!

sockmaster27:
> I think there are still a few uses of `parMap` which should become `parMapSeq` (if you search for uses of the former). Will you check? Thanks!

Not any that can utilize `parMapValuesSeq`, but should we introduce the more general `parMapSeq`?

magnus-madsen:
Yes, let us do it.

PR #6594, opened 2023-10-23
refactor: remove simple gen objects

JonathanStarup:
As code generation is moved into `BackendObjectType`, the old `GenXClass` just becomes a call into that new code. This PR removes the `GenXClass`es that just forward the call into `BackendObjectType`.

In the future, we can just have a list of `Generatable` things that we can generate in parallel.


magnus-madsen:
Can you give a short description of what this does?

magnus-madsen (approved):

PR #6595, opened 2023-10-23
refactor: avoid string concat in `toDescriptor`

JonathanStarup:
Also clarifies the method perhaps


magnus-madsen:
Build issues.

magnus-madsen (approved):

PR #6596, opened 2023-10-23
refactor: explicit erasure

### LINKED ISSUE #6572: Add new internal types `AnyType` and `AnyBool`
... which are used in `Monomorph` when free (unbound) type variables of their kind are encountered.

Functionally `AnyType` is equivalent to `Unit`, but we should keep its type distinct.

I don't think we need `AnyEff` or `AnyRec` etc, because there we will substitute in `Pure`, `EmptyRecord`, and so forth.

JonathanStarup:
Fixes #6572


main/src/ca/uwaterloo/flix/language/ast/MonoType.scala
```diff
@@ -56,6 +56,8 @@ object MonoType {
 
   case object Region extends MonoType
 
+  case class ErasedType(kind: Kind) extends MonoType
```
magnus-madsen, line 59: I would much prefer if this was called AnyType or UndeterminedType or really anything that does not use the word erased.

magnus-madsen:
I see little code and a lot of conflicts, so probably easier to redo this from scratch?

PR #6608, opened 2023-10-27
refactor: introduce `ParOps.parMapSeq`

sockmaster27:


mlutze:
I recommend calling this `parTraverse` since the signature matches the signature of `Validation.traverse`.

magnus-madsen:
@mlutze I suggest to create a ticket with a list of the operations we want in ParOps and name suggestions :)

magnus-madsen (approved):

Issue #6591, opened 2023-10-22
Rename `class` to `trait`


magnus-madsen:
- [x] Add `trait` keyword.
- [x] Remove `class` keyword.
- [x] Rework all library documentation to talk about traits.
- [x] Rework all errors to talk about traits.
- [ ] Rename "compiler" to talk about traits.
- [x] Update book
- [x] Update VSCode
- [x] Update highlight.js fork (used on doc.flix.dev)
- [x] Other highlighters? (Can probably wait as we periodically update those).

mlutze:
🚂

stephentetley:
I imagine this is fairly "mechanical" and won't involve deep changes in the compiler?

If so, I'm happy to look at at.

magnus-madsen:
> I imagine this is fairly "mechanical" and won't involve deep changes in the compiler?
> 
> If so, I'm happy to look at at.

Yes. This would be pretty mechnical. 

Renaming the keyword is only a one line change, but we would probably want everything named something class to become trait internally.

In addition, we would want to update the documentation and the website.

magnus-madsen:
A checklist (could be separate PRs):

- Rename keyword.
- Rework all library documentation to talk about traits.
- Rework all errors to talk about traits.
- Rename data structures in the compiler to talk about traits.

stephentetley:
I've started looking at it and it has already become a huge set of changes.

What about just changing the name of `Symbol.ClassSym` to Symbol.TraitSym` plus changing variable names where it is used for the first PR?

This class is used "everywhere" so the first PR would still be a big one but it would be very limited in scope?

Another matter that comes up straight away - we often use `clazz` as a cute variable name instead `class` which is obviously a reserved word. Is there a similar cute name we want / could use for `trait`?

magnus-madsen:
> What about just changing the name of Symbol.ClassSym to Symbol.TraitSym plus changing variable names where it is used for the first PR?

I am very happy with small PRs, so I am totally onboard with this.

> Another matter that comes up straight away - we often use clazz as a cute variable name instead class which is obviously a reserved word. Is there a similar cute name we want / could use for trait?

Not sure. We sometime used three letter variables.

magnus-madsen:
I just occurs to me that Matt has a pretty large PR -- with the new typer -- open. Changing the compiler right now would probably give him lots of conflicts. Much better to wait until that is done (say 3-6 weeks). 

Perhaps instead we should start with the Flix standard library, examples, and documentation?

[...]

PR #6612, opened 2023-10-27
refactor: allow `trait` and `class` as keywords. Update stdlib, examples and main.test.flix tests to use `trait`.

stephentetley:
This PR is the first part of renaming `class` to `trait` (issue #6591), it adds `trait` as a keyword in the parser, `class` remains a keyword as well.

I've changed the stdlib and the examples to use `trait` rather than `class`. Because "real code" is now using `trait` I've changed the `main.test.flix` tests to use `trait` as well. In cases where there were specific test files to test type classes (Test.Dec.Class.flix, Test.Integ.Class.Schema.flix, Test.Kind.Class.flix) I have duplicated the files to test them using `trait`. At the end of the transition period from `class` to `trait` the Class versions should be removed.

Hopefully these changes to the `main.test.flix` tests will not be particularly onerous for Matthew when he merges in the new type checker. I have left the `LanguageSuite` tests as-is as they are likely to more conflicting and not as easy to resolve given they use embedded Flix code inside Scala code.

### DIFF
--- a/examples/larger-examples/restrictable-variants/colors.flix
+++ b/examples/larger-examples/restrictable-variants/colors.flix
@@ -8,7 +8,7 @@ mod Colors {
 
     // To run this function press the floating "Run" button right below these
     // three lines. This can be done for any function that is `pub` and does not
-    // take arguments. (It doesnt not appear until the code is sucessfully compiled)
+    // take arguments. (It doesnt not appear until the code is successfully compiled)
     pub def main(): Unit \ IO = {
         // create the color red
         let red = Color.Red;
@@ -93,7 +93,7 @@ mod Colors {
 
 }
 
-// Type Class Instances
+// Trait Instances
 
 instance ToString[Colors.Color[i]] {
     pub def toString(x: Colors.Color[i]): String = choose x {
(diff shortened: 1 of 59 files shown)


examples/larger-examples/restrictable-variants/colors.flix
```diff
@@ -8,7 +8,7 @@ mod Colors {
 
     // To run this function press the floating "Run" button right below these
     // three lines. This can be done for any function that is `pub` and does not
-    // take arguments. (It doesnt not appear until the code is sucessfully compiled)
+    // take arguments. (It doesnt not appear until the code is successfully compiled)
```
magnus-madsen, line 11: 👍

main/test/flix/Test.Dec.Class.flix
```diff
@@ -1,3 +1,5 @@
+// TODO this module is replicated by `Test.Dec.Trait.flix` for the `trait` keyword.
+// Remove this module at the end of the transition period from `class` to `trait`.
```
magnus-madsen, line 2: 👍

magnus-madsen (approved):

magnus-madsen:
Thank you!

PR #6615, opened 2023-10-29
refactor: replace `scala-parallel-collections` with native Java threads

### LINKED ISSUE #6583: Replace dependence on `scala-parallel-collections` with ordinary Java

sockmaster27:
Fixes #6583

### DIFF
--- a/main/src/ca/uwaterloo/flix/util/ParOps.scala
+++ b/main/src/ca/uwaterloo/flix/util/ParOps.scala
@@ -31,11 +31,10 @@ object ParOps {
     */
   @inline
   def parMap[A, B](xs: Iterable[A])(f: A => B)(implicit flix: Flix): Iterable[B] = {
-    val in = xs.toVector
-    val out = ArrayBuffer.fill(in.size)(null.asInstanceOf[B])
+    val out = ArrayBuffer.fill(xs.size)(null.asInstanceOf[B])
 
     val futuresBuilder = ArrayBuffer.newBuilder[Future[?]]
-    futuresBuilder.sizeHint(in.size)
+    futuresBuilder.sizeHint(xs.size)
     val futures = futuresBuilder.result()
 
     val threadPool = flix.threadPool
@@ -98,6 +97,7 @@ object ParOps {
       override def call(): Set[T] = next(t)
     }
 
+    // Use global thread pool.
     val threadPool = flix.threadPool
 
     // A mutable variable that holds the currently reachable Ts.
--- a/main/src/ca/uwaterloo/flix/util/ParOps.scala
+++ b/main/src/ca/uwaterloo/flix/util/ParOps.scala
@@ -34,10 +34,10 @@ object ParOps {
     val in = xs.toVector
     val out = ArrayBuffer.fill(in.size)(null.asInstanceOf[B])
 
-    val executorService: ExecutorService = Executors.newFixedThreadPool(Runtime.getRuntime.availableProcessors())
+    val threadPool = flix.threadPool
     val futures = ArrayBuffer.empty[Future[?]]
     for ((elm, idx) <- xs.zipWithIndex) {
-      futures += executorService.submit(new Runnable {
+      futures += threadPool.submit(new Runnable {
         override def run(): Unit = {
           out(idx) = f(elm)
         }
--- a/main/src/ca/uwaterloo/flix/util/ParOps.scala
+++ b/main/src/ca/uwaterloo/flix/util/ParOps.scala
@@ -22,7 +22,6 @@ import scala.jdk.CollectionConverters._
 import ca.uwaterloo.flix.api.Flix
 
 import scala.collection.mutable.ArrayBuffer
-import scala.reflect.ClassTag
 
 object ParOps {
 
--- a/main/src/ca/uwaterloo/flix/util/ParOps.scala
+++ b/main/src/ca/uwaterloo/flix/util/ParOps.scala
@@ -17,7 +17,7 @@
 package ca.uwaterloo.flix.util
 
 import java.util
-import java.util.concurrent.{Callable, ExecutorService, Executors, Future}
+import java.util.concurrent.{Callable, Future}
 import scala.jdk.CollectionConverters._
 import scala.collection.parallel._
 import ca.uwaterloo.flix.api.Flix
@@ -98,8 +98,7 @@ object ParOps {
       override def call(): Set[T] = next(t)
     }
 
-    // Initialize a new executor service.
-    val executorService = Executors.newFixedThreadPool(flix.options.threads)
+    val threadPool = flix.threadPool
 
     // A mutable variable that holds the currently reachable Ts.
     var reach = init
@@ -117,7 +116,7 @@ object ParOps {
       }
 
       // Invoke all callables in parallel.
-      val futures = executorService.invokeAll(callables)
+      val futures = threadPool.invokeAll(callables)
 
       // Compute the set of all inferred Ts in this iteration.
       // May include Ts discovered in previous iterations.
@@ -129,9 +128,6 @@ object ParOps {
       delta = newReach -- reach
       reach = reach ++ delta
     }
-
-    // Shutdown the executor service.
-    executorService.shutdown()
 
     // Return the set of reachable Ts.
     reach
--- a/main/src/ca/uwaterloo/flix/util/ParOps.scala
+++ b/main/src/ca/uwaterloo/flix/util/ParOps.scala
@@ -17,11 +17,12 @@
 package ca.uwaterloo.flix.util
 
 import java.util
-import java.util.concurrent.{Callable, Executors}
-
+import java.util.concurrent.{Callable, ExecutorService, Executors, Future}
 import scala.jdk.CollectionConverters._
 import scala.collection.parallel._
 import ca.uwaterloo.flix.api.Flix
+
+import scala.collection.mutable.ArrayBuffer
 
 object ParOps {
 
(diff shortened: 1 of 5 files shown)


sockmaster27:
Also a nice performance boost

Before:
```
Throughput (best): 20.725 lines/sec (with 16 threads.)

  min:  4.680, max: 20.725, avg: 17.124, median: 19.219

  The highest throughput was in iteration: 5 (out of 10).
  The ratio between the best and worst iteration was: 4,4x.

Finished 10 iterations on 66.136 lines of code in 47 seconds.
```

After:
```
Throughput (best): 23.072 lines/sec (with 16 threads.)

  min:  4.814, max: 23.072, avg: 18.763, median: 21.414

  The highest throughput was in iteration: 7 (out of 10).
  The ratio between the best and worst iteration was: 4,8x.

Finished 10 iterations on 66.136 lines of code in 43 seconds.
```

sockmaster27:
Should I also look at making `parMapValues` more efficient?

magnus-madsen:
Is the performance boost real? That would be surprising! (but a welcome surprise, to be sure).

main/src/ca/uwaterloo/flix/util/ParOps.scala
```diff
@@ -17,30 +17,38 @@
...
   def parMap[A, B](xs: Iterable[A])(f: A => B)(implicit flix: Flix): Iterable[B] = {
-    // Build the parallel array.
-    val parArray = xs.toParArray
-
-    // Configure the task support.
-    parArray.tasksupport = flix.forkJoinTaskSupport
+    val out = ArrayBuffer.fill(xs.size)(null.asInstanceOf[B])
+
+    val futuresBuilder = ArrayBuffer.newBuilder[Future[?]]
+    futuresBuilder.sizeHint(xs.size)
```
magnus-madsen, line 36: The argument `xs` is `Iterable` which mean it could be a `List`. You have several calls to `xs.size` which could be `O(n)` for each.

main/src/ca/uwaterloo/flix/util/ParOps.scala
```diff
@@ -17,30 +17,38 @@
...
     * Apply the given function `f` to each element in the list `xs` in parallel.
     */
   @inline
   def parMap[A, B](xs: Iterable[A])(f: A => B)(implicit flix: Flix): Iterable[B] = {
-    // Build the parallel array.
-    val parArray = xs.toParArray
-
-    // Configure the task support.
-    parArray.tasksupport = flix.forkJoinTaskSupport
+    val out = ArrayBuffer.fill(xs.size)(null.asInstanceOf[B])
```
magnus-madsen, line 33: Any reason to use an `ArrayBuffer` over a raw array? 

The array will never change its size anyway.

main/src/ca/uwaterloo/flix/util/ParOps.scala
```diff
@@ -17,30 +17,38 @@
...
+
+    val futuresBuilder = ArrayBuffer.newBuilder[Future[?]]
+    futuresBuilder.sizeHint(xs.size)
+    val futures = futuresBuilder.result()
```
magnus-madsen, line 37: OK; so I understand why this is done. I just wonder if we could do better.

In particular, if we can avoid allocation of `n` `Future`s.

It seems one option would be to use https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html#awaitTermination-long-java.util.concurrent.TimeUnit-

But the problem is that it shutsdown the threadpool. Then we would need to create a new thread pool, but that is very expensive.

So I am not sure if it can be done better than what you have?

magnus-madsen: Maybe it would be more efficient to use a CountDownLatch:

https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/util/concurrent/CountDownLatch.html

magnus-madsen: Does it make sense? I am trying to reduce allocation of `Future`s.

main/src/ca/uwaterloo/flix/util/ParOps.scala
```diff
@@ -69,14 +77,45 @@ object ParOps {
     */
   @inline
   def parAgg[A, S](xs: Iterable[A], z: => S)(seq: (S, A) => S, comb: (S, S) => S)(implicit flix: Flix): S = {
```
magnus-madsen, line 79: Impressive that you got this to work.

magnus-madsen: I actually wanted something very simple, e.g. based on https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/RecursiveTask.html

(I should have stated that of course).

magnus-madsen: Once you have tried `RecursiveTask`, we can compare your hand-rolled implementation to Java state-of-the-art :) (pre VTs).

magnus-madsen (changes requested):

magnus-madsen:
I added lots of comments; its mostly my bad, I should have pointed out the concurrency primitives I wanted us to use.

sockmaster27:
> Is the performance boost real? That would be surprising! (but a welcome surprise, to be sure).

I'm seeing it pretty consistently on my machine.

magnus-madsen:
> > Is the performance boost real? That would be surprising! (but a welcome surprise, to be sure).
> 
> I'm seeing it pretty consistently on my machine.

Where is this from do you think? parMap or aggregate?

magnus-madsen:
!!!jar

magnus-madsen:
I don't see much, unfortunately:

```
master

Throughput (best): 34.813 lines/sec (with 24 threads.)

  min:  7.221, max: 34.813, avg: 28.973, median: 33.515

  The highest throughput was in iteration: 7 (out of 10).
  The ratio between the best and worst iteration was: 4,8x.

Finished 10 iterations on 66.136 lines of code in 28 seconds.

PR

Throughput (best): 34.859 lines/sec (with 24 threads.)

  min:  7.280, max: 34.859, avg: 28.837, median: 32.994

  The highest throughput was in iteration: 5 (out of 10).
  The ratio between the best and worst iteration was: 4,8x.

Finished 10 iterations on 66.136 lines of code in 28 seconds.
```

sockmaster27:
> > > Is the performance boost real? That would be surprising! (but a welcome surprise, to be sure).
> > 
> > 
> > I'm seeing it pretty consistently on my machine.
> 
> Where is this from do you think? parMap or aggregate?

From my testing, I found the new aggregate to be slightly slower than the old one, but it has almost no impact on the overall performance. Even when I tried running it completely sequentially, it was barely noticable.
So it's all `parMap` (if it actually is faster).

sockmaster27:
> I don't see much, unfortunately:

Did you run them each more than once?

magnus-madsen:
> > I don't see much, unfortunately:
> 
> Did you run them each more than once?

Yes

main/src/ca/uwaterloo/flix/util/ParOps.scala
sockmaster27: It's possible in `parAgg`, but in this particular case it requires a `B: ClassTag` bound which seems unnecessarily noisy.

main/src/ca/uwaterloo/flix/util/ParOps.scala
sockmaster27: I'm having trouble imagining a recursive version that would be efficient.

sockmaster27: What do you have in mind?

main/src/ca/uwaterloo/flix/util/ParOps.scala
```diff
@@ -17,30 +17,37 @@
...
     */
   @inline
   def parMap[A, B](xs: Iterable[A])(f: A => B)(implicit flix: Flix): Iterable[B] = {
-    // Build the parallel array.
-    val parArray = xs.toParArray
-
-    // Configure the task support.
-    parArray.tasksupport = flix.forkJoinTaskSupport
+    val size = xs.size
+    val out = ArrayBuffer.fill(size)(null.asInstanceOf[B])
```
mlutze, line 34: Could this just be Array.new(size)?

magnus-madsen: 👍

main/src/ca/uwaterloo/flix/util/ParOps.scala
```diff
@@ -17,30 +17,37 @@
...
+    val threadPool = flix.threadPool
+    val latch = new CountDownLatch(size)
+    for ((elm, idx) <- xs.zipWithIndex) {
+      threadPool.submit(new Runnable {
+        override def run(): Unit = {
+          out(idx) = f(elm)
+          latch.countDown()
+        }
+      })
+    }
```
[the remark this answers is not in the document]
sockmaster27: Doesn't appear to do anything to the performance even on my machine unfortunately.

main/src/ca/uwaterloo/flix/util/ParOps.scala
sockmaster27: Same problem as [this](https://github.com/flix/flix/pull/6615#discussion_r1382553948) unfortunately

main/src/ca/uwaterloo/flix/util/ParOps.scala
magnus-madsen: @mlutze @JonathanStarup Any idea how to rewrite this function using plain Java primitives?

main/src/ca/uwaterloo/flix/util/ParOps.scala
magnus-madsen: Using a class tag is fine, if that is the only problem.

magnus-madsen: Alternatively use an array of objects and then cast

magnus-madsen:
Let us split this PR into two: We can merge in the new parMap now.

magnus-madsen:
For parAgg, if I look at all the uses, I think it can simply be dropped and replaced by something with a simpler signature.

Something which would be easier to implement.

magnus-madsen:
I think we can go with something like:

```
def parAgg[A, S](xs: Iterable[A], z: => S)(f: A => S, comb: (S, S) => S)(implicit flix: Flix): S
```

sockmaster27:
> I think we can go with something like:
> 
> ```
> def parAgg[A, S](xs: Iterable[A], z: => S)(trans: A => S, comb: (S, S) => S)(implicit flix: Flix): S
> ```

What's the point of having both `z` and `trans`?

magnus-madsen:
You need z if the iterable is empty.

magnus-madsen:
@sockmaster27 I would really like if we could get the first part (parMap) merged in, via a separate PR.

PR #6616, opened 2023-10-29
refactor: update KeywordOtherCompleter and Weeder to recognize `trait` keyword

stephentetley:
This PR updates the completions in `KeywordOtherCompleter` and the set of reserved words in `Weeder` to recognize the new `trait` keyword.

Part of the work on rename `class` to `trait` issue #6591 .


stephentetley:
Should I be looking at `SemanticTokenType` as well?

magnus-madsen (approved):

magnus-madsen:
> Should I be looking at `SemanticTokenType` as well?

Yes

PR #6619, opened 2023-10-31
chore: update GitHub actions to the latest major release for security

KengoTODA:
I found several actions still depend on old major releases, so I suggest updating them in batch.

I've confirmed that all workflows work fine with these fixes:

- [Compiler Tests](https://github.com/KengoTODA/flix/actions/runs/6705975929)
- [Community Build](https://github.com/KengoTODA/flix/actions/runs/6705975928)
- [Semantic Commit Message](https://github.com/KengoTODA/flix/actions/runs/6705975941)
- [JAR on Demand](https://github.com/KengoTODA/flix/actions/runs/6706049488)

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -15,9 +15,9 @@ jobs:
 
     steps:
       - name: Check out master
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


magnus-madsen:
Thanks!

PR #6631, opened 2023-11-05
chore: rename `PatternExhaustiveness` to `PatMatch`

magnus-madsen:
(This will make it easier to print graphs with phase names).


mlutze (approved):
I don't love the name (one might think it's simplification of pattern matches)

magnus-madsen:
> I don't love the name (one might think it's simplification of pattern matches)

Suggestions are welcome (but keep them short).

PR #6636, opened 2023-11-05
feat: improve error handling in Weeder

jaschdoc:
Adds `recoverOne` to more cases.


magnus-madsen:
The intent was  good; but this is not the right path. This would be like adding try-catch everywhere to a program.

Instead, we should only use recoverOne directly around a place that could "throw".

magnus-madsen:
Unfortunately, the types are not strong enough to tell us where this can be.

magnus-madsen:
Make sense?

jaschdoc:
Roger :)

PR #6645, opened 2023-11-06
feat: use parallel evaluation in `Namer`

magnus-madsen:


main/src/ca/uwaterloo/flix/language/phase/Namer.scala
```diff
@@ -99,10 +99,10 @@ object Namer {
     case DesugaredAst.Declaration.Namespace(ident, usesAndImports0, decls0, loc) =>
       val ns = Name.NName(ident.sp1, ns0.idents :+ ident, ident.sp2)
       val usesAndImports = usesAndImports0.map(visitUseOrImport)
-      val declsVal = traverse(decls0)(visitDecl(_, ns))
+      val declsVal = ParOps.parMapSeq(decls0)(visitDecl(_, ns))
```
mlutze, line 102: I'm not sure that this buys us anything if we're already parallel over compilation units, especially since namespaces can be nested.

magnus-madsen: But we are not? Maybe we  should be?

mlutze: Ah right, I misread the above. IMO we should just move the parallelism over the compilation units rather than namespaces. I'd guess the performance impact is the same, but not sure.

magnus-madsen: I create a ticket.

magnus-madsen:
I need the approval :)

mlutze (approved):

PR #6649, opened 2023-11-07
feat: use parallel evaluation in `PatMatch`

magnus-madsen:


mlutze (approved):
We need parFlatMap :)

PR #6656, opened 2023-11-07
feat: use parallel evaluation in `MonoDefs`

magnus-madsen:


magnus-madsen:
Damn it, I feel into the reformat trap too :(

mlutze (approved):

PR #6658, opened 2023-11-07
feat: concurrency example

JonathanStarup:
Using channels to have a dynamically updated worklist


magnus-madsen:
Needs comments. Also not really sure this is an example we want, even though its fun.

magnus-madsen:
I suggest to move this to its own repository.

PR #6668, opened 2023-11-08
feat: use parallel evaluation in `MonoTypes`

magnus-madsen:


magnus-madsen:
@JonathanStarup I think I would like to merge this in, but then we need another approach.

Possibly something about collecting the new enums before creating them.

magnus-madsen:
@mlutze Can you approve? @JonathanStarup Can comment later on what to do.

mlutze (approved):

Issue #6421, opened 2023-08-28
Split `Weeder` into `Weeder` and `Desugar`


magnus-madsen:
The idea is to split the weeder into two: one that performs all the syntactic checks and one that rewrites constructs (e.g. `form`).

Crucially, the `Desugar` phase should not emit any errors at all. It should simply translate expressions into other expressions.

I suggest to do this incrementally: 

- [x] Introduce a new ast `DesugaredAst` based on `WeededAst`.
- [x] Introduce a phase `Desugar` between `Weeder` and `Namer` (at first this would be more or less the identity function).
- [ ] Move one translation at a time from `Weeder` into `Desugar`.

magnus-madsen:
@jaschdoc I would like this to be the next PR because we will be building on it soonish.

Does the task and outline seem sensible?

I prefer small PRs.

jaschdoc:
Yes :)

jaschdoc:
- [ ] Introduce a new ast `DesugaredAst` based on `WeededAst`.
- [ ] Introduce a phase `Desugar` between `Weeder` and `Namer` (at first, this would be more or less the identity function).
- [ ] Move one translation at a time from `Weeder` into `Desugar`.

jaschdoc:
- [x] Infix
- [ ] LambdaMatch
- [x] ApplicativeFor
- [x] ForEach
- [x] MonadicFor
- [x] ForEachYield
- [ ] LetMatch
- [ ] LetImport (?)
- [ ] Static (?)
- [ ] Tuple
- [ ] RecordLit (?)
- [ ] RecordOperation (?)
- [ ] FCons
- [ ] FAppend
- [ ] ListLit
- [ ] SetLit
- [ ] MapLit
- [ ] Interpolation
- [ ] FixpointSolveWithProject
- [ ] FixpointQueryWithSelect
- [ ] Debug (?)

magnus-madsen:
- Cleanup and reorder.

magnus-madsen:
- Liftharded symbols/names/strings to top of `Desugar` class.

jaschdoc:
>  Liftharded

What does this mean?

magnus-madsen:
It means to move the constant strings to the top of the file (e.g. "Functor")

magnus-madsen:
Are we getting close to done? What is missing?

jaschdoc:
[...]

PR #6672, opened 2023-11-08
refactor: move LetMatch to Desugar

jaschdoc:
Related to https://github.com/flix/flix/issues/6421

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/WeededAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/WeededAst.scala
@@ -133,6 +133,8 @@ object WeededAst {
 
     case class ForEachYield(frags: List[ForFragment], exp: Expr, loc: SourceLocation) extends Expr
 
+    case class LetMatch(pat: Pattern, mod: Ast.Modifiers, tpe: Option[Type], exp1: Expr, exp2: Expr, loc: SourceLocation) extends Expr
+
     case class Tuple(exps: List[Expr], loc: SourceLocation) extends Expr
 
     case class RecordEmpty(loc: SourceLocation) extends Expr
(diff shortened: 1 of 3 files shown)


main/src/ca/uwaterloo/flix/language/ast/WeededAst.scala
```diff
@@ -133,6 +133,8 @@ object WeededAst {
 
     case class ForEachYield(frags: List[ForFragment], exp: Expr, loc: SourceLocation) extends Expr
 
+    case class LetMatch(pat: Pattern, mod: Ast.Modifiers, tpe: Option[Type], exp1: Expr, exp2: Expr, loc: SourceLocation) extends Expr
```
magnus-madsen, line 136: What kind of ordering are we using here?

magnus-madsen: Can we try to keep everything sorted in the same way, following WeededAst? :)

magnus-madsen:
Ill merge as-is and then we can have a reordering PR :)

magnus-madsen (approved):

main/src/ca/uwaterloo/flix/language/ast/WeededAst.scala
jaschdoc: It already does, right?
I think Weeder was wrong before I began this (LetMatch comes after ForEachYield in Weeder and Ast, but Tuple is next in Ast whereas LetRecDef is next in Weeder)

jaschdoc: Anyway, we can change this after :)

PR #6681, opened 2023-11-09
feat: optimize monoTypes

JonathanStarup:
* before
  * discover enums in defs, when a new enum is found, immediately specialize
* Trick 1
  * first discover enums in defs in parallel, then recursively specialized enums with frontier synchronization.
  * Performance gain: 1.2x (1 vs 8 cores)
* Trick 2
  *  don't synchronize `get`, only put
  * (I don't know if this is actually safe in the scala hashmap)
  * Performance gain (with trick 1) 2.6x (1 vs 8 cores)

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/MonoTypes.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/MonoTypes.scala
@@ -62,7 +62,7 @@ object MonoTypes {
       * 2b. Pure + Pure and Pure should not both be present (formulas should be fully evaluated)
       * 3. Contain no specialized enums, i.e. should use List[Int32] instead of List$32
       */
-    private val enum2enum: mutable.Map[(Symbol.EnumSym, Type), Symbol.EnumSym] = mutable.Map.empty
+    private val enum2enum: mutable.Map[(Symbol.EnumSym, Type), (Symbol.EnumSym, List[Type])] = mutable.Map.empty
 
     /**
       * A map used to collect specialized definitions.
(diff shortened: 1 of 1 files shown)


JonathanStarup:
@magnus-madsen this is not polished but i think the takeaway is that we should take advantage of the fact that enum2enum has no updates, only insertions. Seems everybody is slamming the map lock

magnus-madsen:
I think you have to synchronize on `get` otherwise the map could be outdated. It is a race condition (like in MonoDefs-- which should be fixed).

magnus-madsen:
Try to swap the Scala map for https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html

magnus-madsen:
I think you have to use [computeIfAbsent](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#computeIfAbsent-K-java.util.function.Function-)

JonathanStarup:
> I think you have to synchronize on `get` otherwise the map could be outdated. It is a race condition (like in MonoDefs-- which should be fixed).

how would it be outdated? either its there, which will always have the same mapping; or it's not there, in which case it goes to the synchronized getOrPut

JonathanStarup:
> Try to swap the Scala map for https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html

but yeah this also doesn't always block on get so it should be perfect

magnus-madsen:
You are both accessing and updating a mutable data structure concurrently-- that is a not generally allowed.

magnus-madsen:
What was the conclusion here?

JonathanStarup:
Ill write it up in tickets

PR #6686, opened 2023-11-10
refactor: use native Java threads in `parMap`

sockmaster27:


main/test/ca/uwaterloo/flix/util/TestParOps.scala
```diff
@@ -0,0 +1,41 @@
...
+package ca.uwaterloo.flix.util
+
+import ca.uwaterloo.flix.api.Flix
+import org.scalatest.funsuite.AnyFunSuite
+
+import java.util.concurrent.Executors
+
+class TestParOps extends AnyFunSuite {
+
+  implicit var flix: Flix = new Flix()
```
magnus-madsen, line 25: val?

magnus-madsen:
Can you try `Xperf` and `--Xbenchmark-throughput` to check that performance is about the same?

sockmaster27:
====== master ======
```
Throughput (best): 30.001 lines/sec (with 16 threads.)

  min:  4.749, max: 30.001, avg: 22.573, median: 25.631

  The highest throughput was in iteration: 6 (out of 10).
  The ratio between the best and worst iteration was: 6,3x.

Finished 10 iterations on 66.136 lines of code in 39 seconds.
```


====== new ======
```
Throughput (best): 27.977 lines/sec (with 16 threads.)

  min:  4.921, max: 27.977, avg: 22.856, median: 27.517

  The highest throughput was in iteration: 8 (out of 10).
  The ratio between the best and worst iteration was: 5,7x.

Finished 10 iterations on 66.136 lines of code in 38 seconds.
```

sockmaster27:
====== master ======
```
Throughput (best): 34.832 lines/sec (with 16 threads.)

  min: 28.304, max: 34.832, avg: 31.480, median: 32.166

Finished 7 iterations on 66.136 lines of code in 14 seconds.
```


====== new ======
```
Throughput (best): 32.967 lines/sec (with 16 threads.)

  min: 30.725, max: 32.967, avg: 31.807, median: 32.055

Finished 7 iterations on 66.136 lines of code in 14 seconds.
```

magnus-madsen:
!!!jar

magnus-madsen:
I also get lower performance. But no idea why?

magnus-madsen:
Can we try to use a https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinPool.html and then use the `execute` method on it? This should safe some allocations of futures.

magnus-madsen:
I would use this constructor: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinPool.html#ForkJoinPool-int- inside Flix

sockmaster27:
I've now tried using `.execute()` both on the threadpool and the forkjoinpool, but no matter what, running Xperf seems to just land on a best somewhere random between 31.000 and 37.000 lines per second.

sockmaster27:
So for now I'm still just using the threadpool.

magnus-madsen:
So-- for unknown reasons -- Scala ParArray remains faster??

sockmaster27:
> So-- for unknown reasons -- Scala ParArray remains faster??

It seems to be varying wildly, but maybe(???) ParArray is a little more consistent towards the middle. It's very noisy and I think at this point the differences from run to run are completely overshadowing the differences between the implementations.

magnus-madsen:
> > So-- for unknown reasons -- Scala ParArray remains faster??
> 
> It seems to be varying wildly, but maybe(???) ParArray is a little more consistent towards the middle. It's very noisy and I think at this point the differences from run to run are completely overshadowing the differences between the implementations.

You can bump the N in CompilerPerf. I can also try when I get home.

I wonder if the countdown latch is the bottleneck.

magnus-madsen:
But is the code using ForkJoinPool??

sockmaster27:
> But is the code using ForkJoinPool??

I tried both, they were identical, so I stayed with the simpler thread pool.

magnus-madsen:
!!!jar

magnus-madsen:
See https://github.com/flix/flix/pull/6690

PR #6690, opened 2023-11-12
feat: use `ForkJoin` for `ParOps`

magnus-madsen:


main/src/ca/uwaterloo/flix/util/ParOps.scala
```diff
@@ -17,30 +17,41 @@
...
+  def parMap[A, B: ClassTag](xs: Iterable[A])(f: A => B)(implicit flix: Flix): Iterable[B] = {
+    // Compute the size of the input and construct a new empty array to hold the result.
+    val size = xs.size
+    val out: Array[B] = new Array(size)
+
+    // Construct a new count down latch to track the number of threads.
+    val latch = new CountDownLatch(size)
+
+    // Iterate through the elements of `xs`. Use a local variable to track the index.
+    var idx = 0
```
mlutze, line 42: zipWithIndex could avoid some mutation

mlutze (approved):

main/src/ca/uwaterloo/flix/util/ParOps.scala
magnus-madsen: I actually switched from zipWithIndex __because it improved performance!!__

PR #6693, opened 2023-11-12
feat: add `n` flag to `Xperf`

magnus-madsen:


main/src/ca/uwaterloo/flix/Main.scala
```diff
@@ -376,7 +378,11 @@ object Main {
             .required()
         )
 
-      cmd("Xperf").action((_, c) => c.copy(command = Command.CompilerPerf)).hidden()
+      cmd("Xperf").action((_, c) => c.copy(command = Command.CompilerPerf)).children(
+        opt[Int]("n")
```
mlutze, line 382: Does the command line parser have a "default" option that would automatically display in the usage text?

mlutze (approved):

main/src/ca/uwaterloo/flix/Main.scala
magnus-madsen: Huh?

mlutze: E.g. we write `cmd(....).default(CompilerPerf.Default)`

And then when the user does `flix --help` it displays

```
usage: blah blah
    -Xperf n       default: 7
```

magnus-madsen: Not sure, but in any case, the plan is to hide all `X` options and this one is already hidden.

PR #6694, opened 2023-11-12
feat: add `--frontend` flag to `Xperf`

magnus-madsen:


main/src/ca/uwaterloo/flix/util/Options.scala
```diff
@@ -49,7 +49,8 @@ object Options {
     xnoqmc = false,
     xsummary = false,
     xparser = false,
-    xperfn = None
+    XPerfN = None,
```
mlutze, line 52: Why capital letters?

mlutze (approved):

main/src/ca/uwaterloo/flix/util/Options.scala
magnus-madsen: Yeah, I don't know. In fact what should we name these child options?

I was thinking of dropping type safety and just dumping all internal flags into a map?

mlutze: > Yeah, I don't know. In fact what should we name these child options?
> 
> I was thinking of dropping type safety and just dumping all internal flags into a map?

Maybe the child options go in their own `PerfOptions` case class?
I don't like dropping type safety because then I really have to remember what everything is called when hacking around

PR #6721, opened 2023-11-17
refactor: introduce `LocalContext` and `SharedContext` in `Redundancy`

magnus-madsen:
- Introduce `LocalContext`
- Introduce `SharedContex`
- Drop `IllegalSingleVariable` (but see #6724)
- Drop support for restrictable variants.


magnus-madsen:
I would like to get rid of `Used`, but this is a bit of a pain in the neck!

mlutze (approved):

PR #6725, opened 2023-11-20
feat: collect def local args

JonathanStarup:
* Collect local variables in defs
* Clarifying comments in VarOffsets
* Changed TryCatch in VarOffsets not to reuse indices (variable re-initialization cannot handle that)

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
@@ -20,6 +20,7 @@ import ca.uwaterloo.flix.language.ast._
 import ca.uwaterloo.flix.util.ParOps
 
 import java.util.concurrent.ConcurrentLinkedQueue
+import scala.collection.mutable
 import scala.jdk.CollectionConverters._
 
 object Reducer {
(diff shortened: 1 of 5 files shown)


JonathanStarup:
@magnus-madsen I used a mutable list to collect local params. Is that allowed? It is asymptotically faster (avoids list concatenation) if that matters

main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
```diff
@@ -66,99 +69,105 @@ object Reducer {
...
-      val es = methods.map(m => visitExpr(m.clo))
-      val specs = methods.map {
-        case LiftedAst.JvmMethod(ident, fparams, _, retTpe, purity, loc) =>
-          val f = fparams.map(visitFormalParam)
-          ReducedAst.JvmMethod(ident, f, retTpe, purity, loc)
-      }
-      ctx.anonClasses.add(ReducedAst.AnonClass(name, clazz, tpe, specs, loc))
-
-      ReducedAst.Expr.NewObject(name, clazz, tpe, purity, specs, es, loc)
+  private def visitExpr(lparams: mutable.ArrayBuffer[ReducedAst.LocalParam], exp0: LiftedAst.Expr)(implicit ctx: SharedContext): ReducedAst.Expr = {
```
magnus-madsen, line 72: This also most right, but should follow the same pattern as in e.g. `Redundancy`. 

So introduce a `case class LocalContext(lparams: mutable.ArrayBuffer[ReducedAst.LocalParam])` and __implictly__ pass it around.

magnus-madsen:
Merged!

PR #6727, opened 2023-11-20
refactor: make `nestedTypesOf` tail recursive

JonathanStarup:
Should be faster, because of the removed flatMap

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
@@ -18,8 +18,11 @@ package ca.uwaterloo.flix.language.phase
 import ca.uwaterloo.flix.api.Flix
 import ca.uwaterloo.flix.language.ast._
 import ca.uwaterloo.flix.util.ParOps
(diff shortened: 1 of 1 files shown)


main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
```diff
@@ -265,52 +267,41 @@ object Reducer {
         // the enum type itself
         val eType = e.tpe
         // the types inside the cases
-        val caseTypes = e.cases.values.flatMap(c => nestedTypesOf(c.tpe)(root, flix))
```
JonathanStarup, line 268: This is double work so I removed the call

magnus-madsen:
But the real solution, of course, is to remove the visit entirely right? Instead we should do it as part of the normal traversal.

magnus-madsen:
Any impact on performance with `Xperf --n 17`?

JonathanStarup:
PR Reducer: 465.9578
Master Reducer: 564.29455

magnus-madsen:
!!!jar

magnus-madsen (approved):

Issue #6698, opened 2023-11-13
Effect Handlers


JonathanStarup:
What is needed for effect handlers to work (in the backend)? The filled dots indicate a mix between complexity and how cumbersome it is -- trying to estimate the relative hours it will take.
- [x] Bind resume as a variable (🔴🔴🔴⭕⭕) #6642
  - To use an effect class per effect, resume must be a regular variable.
- [x] Lambda lift handler bodies (🔴⭕⭕⭕⭕) #6708
- [x] Effect classes (🔴⭕⭕⭕⭕)  #6711
  - A Flix effect is a named set of defs, which will be a backend class with a field for each def, e.g., a `print` field that holds a `Fn1$Obj` def.
- [x] Add frame information to defs (🔴🔴⭕⭕⭕) #6738
  - find local variables #6725
  - add pc field
  - The frame holds local variable information (needed for resumption).
- [x] Add `pc` resumption points that load the resumption argument (🔴🔴🔴🔴⭕)
- [x] implement `TryWith` (🔴🔴⭕⭕⭕)
  - remember to Value wrap the result of the handler op closure 
- [x] implement `Do` (🔴⭕⭕⭕⭕)
- [x] implement `resume` (🔴⭕⭕⭕⭕)
  - materialize resume into a normal function (with Value boxing/unboxing)

PR #6738, opened 2023-11-21
feat: save def locals and pc

JonathanStarup:
Related to #6698

* Merge `GenFunctionClass` and `GenClosureClasses` since they are 95% identical
* save all local variables of functions/closures as fields which are restored upon entry
* have a `pc` field which is currently unused

Unexpected problem: if the `copy` function is inlined, methods become too big to be valid. the `copy` function is the one responsible for creating a new frame, i.e. duplicating the current def/closure. These things have tens of local variables and its done on each non-tail call.

An overview of the function/closure class is in the documentation


magnus-madsen:
I see some build errors.

> Unexpected problem: if the copy function is inlined then methods become too big to be valid.

Is this solved? What "copy" function is this?

JonathanStarup:
> I see some build errors.
> 
> > Unexpected problem: if the copy function is inlined then methods become too big to be valid.
> 
> Is this solved? What "copy" function is this?

Ah, never rename manually. It is resolved. I've expanded the explanation a bit in the main text.

JonathanStarup:
@magnus-madsen After running some metrics, the reason is that some functions have up to 90 local variables. Just estimating based on random programs I think the median number is maybe 10

magnus-madsen:
> @magnus-madsen After running some metrics, the reason for potentially large methods, is that some functions have up to 90 local variables. Just estimating based on random programs The median use 0 local variables but if you don't count those, then the median is probably around 10

What functions?

Will this cause breakages? (i.e. I don't want to merge something if it breaks code).

Does disabling the optimizer change it?

magnus-madsen:
!!!jar

magnus-madsen:
I see functions like:

```
Vector.join$5365225
RedBlackTree.rotate$5339518
Concurrent.Channel.put$5339935
Concurrent.Channel.getHelper$5341144
```

with more than 50 local vars-- which admittedly is on the high end. Nevertheless, the JVM supports up to 65535.

So I assume the problem is with the code this PR is generating?

magnus-madsen:
Ah-- but I guess you fix is to use a helper function -- copy -- to avoid the code blowup and then everything should be fine?

JonathanStarup:
> Ah-- but I guess you fix is to use a helper function -- copy -- to avoid the code blowup and then everything should be fine?

yes exactly, otherwise the methods become too large

magnus-madsen:
Do you believe this to be stable enough to merge in?

main/src/ca/uwaterloo/flix/language/phase/jvm/GenFunAndClosureClasses.scala
```diff
@@ -0,0 +1,306 @@
...
+        macc + (functionname -> JvmClass(functionname, code))
+
+      case (macc, _) =>
+        macc
+    }, _ ++ _)
+  }
+
+  private def isClosure(defn: Def): Boolean = defn.cparams.nonEmpty
+
+  private def isFunction(defn: Def): Boolean = defn.cparams.isEmpty
```
magnus-madsen, line 59: Something about excluded middle, could define isClosure in terms of isFunction

main/src/ca/uwaterloo/flix/language/phase/jvm/GenFunAndClosureClasses.scala
```diff
@@ -0,0 +1,306 @@
...
+
+  private sealed trait FunctionKind
+
+  private case object Function extends FunctionKind
+
+  private case object Closure extends FunctionKind
+
+  /**
+    * `(a|b)` is used to represent the function (left) or closure version (right)
```
magnus-madsen, line 68: Thanks for example 👍

magnus-madsen (approved):

main/src/ca/uwaterloo/flix/language/phase/jvm/GenFunAndClosureClasses.scala
JonathanStarup: Yeah but i like it for readability. I can delete one of them if you want

JonathanStarup:
> Do you believe this to be stable enough to merge in?

@magnus-madsen It should be stable, I'm mostly questioning performance

magnus-madsen:
Merged! Will you update the ticket :)

PR #6751, opened 2023-11-25
refactor: use `toHardFailure` from `Validation`

magnus-madsen:


magnus-madsen:
@mlutze If I could get another approval :) Danke sehr!

mlutze (approved):

PR #6755, opened 2023-11-27
chore: cleanup `PatMatch`

magnus-madsen:


magnus-madsen:
@mlutze This phase is a good candidate for incrementalism because there is only a single dependency: `enumSym <- defSym`.

mlutze (approved):

PR #6757, opened 2023-11-28
feat: add `Pattern.Error`

magnus-madsen:


magnus-madsen:
CC @Herbstein This will allow errors inside patterns-- so now we can have resilient expressions and patterns.

mlutze (approved):

PR #6759, opened 2023-11-28
feat: recoverable `DuplicateAnnotation`

magnus-madsen:


magnus-madsen:
@mlutze So much work for so little. But I think this new "programming pattern" is better.

magnus-madsen:
I think the use here of imperative programming + traverse is reasonable. Its seems to separate the two concerns pretty well, while collecting all errors.

mlutze (approved):

PR #6762, opened 2023-11-29
feat: Op-stack-less Effect Operations

JonathanStarup:
fposter,small,wall_texture,square_product,600x600


magnus-madsen:
What was the reason we did not end up needing this?

JonathanStarup:
This is one out of three implementations. This one does on-demand let-binding instead of just doing it for everything. I think we should still do something like this when we get to effect optimization

PR #6767, opened 2023-11-29
refactor: move Interpolation to Desugar

jaschdoc:
Related to https://github.com/flix/flix/issues/6421


jaschdoc:
There is a lot of weeding going on / special cases for `Interpolation` in Weeder. I can split it up into desugaring and weeding but I'm not sure it's worth the extra case in Desugar? Thoughts?

magnus-madsen:
I don't think this one is worth it. Lets keep it as-is for now and move on.

magnus-madsen:
I would wait with this. Once the new parser arrives it will be more clear what to do where.

PR #6799, opened 2023-12-01
feat: recoverable `DuplicateTag`

magnus-madsen:


magnus-madsen:
@mlutze For this one, please double check the code :)

mlutze (commented):
I'm not sure this is the right way. I prefer recovery by eliminating the duplicate tag.

mlutze:
Confusing name errors are better than confusing type errors

magnus-madsen:
> I'm not sure this is the right way. I prefer recovery by eliminating the duplicate tag.

What do you mean?

magnus-madsen:
But in any case this is not my last iteration on this.

mlutze (approved):

PR #6800, opened 2023-12-01
feat: recoverable `NonUnaryAssocType`

magnus-madsen:


mlutze (approved):
I'd really prefer an error type, but OK

PR #6806, opened 2023-12-01
feat: add `TypeConstructor.Error`

magnus-madsen:


magnus-madsen:
This does not yet deal with unification-- so that remains to be sorted out.

main/src/ca/uwaterloo/flix/language/fmt/FormatType.scala
```diff
@@ -351,6 +352,9 @@ object FormatType {
 
       case SimpleType.Tuple(elms) =>
         elms.map(visit(_, Mode.Type)).mkString("(", ", ", ")")
+
+      case SimpleType.Error => "Error"
```
mlutze, line 356: Should this be something more distinctive?

mlutze (approved):

main/src/ca/uwaterloo/flix/language/fmt/FormatType.scala
magnus-madsen: I think once `Type.Error` is actually used/shows up somewhere it would be sensible to do something better here.

PR #6810, opened 2023-12-02
feat: make effects stack-less

JonathanStarup:
A compromise between a simple let-binder and a clever solution. We let bind everything, but complexity is added since, when looking at the body of a top-level let-binding (`let x = e1; e2`) we don't want to let-bind the outermost structure of `e1`.

This last small optimization is about a 2x decrease in let-binders and avoid the type issues with regions and the bug with if-statements generated from pattern matches.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/EffectBinder.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/EffectBinder.scala
@@ -17,8 +17,12 @@
 package ca.uwaterloo.flix.language.phase
 
 import ca.uwaterloo.flix.api.Flix
+import ca.uwaterloo.flix.language.ast.Ast.BoundBy
 import ca.uwaterloo.flix.language.ast.ReducedAst._
-import ca.uwaterloo.flix.util.ParOps
+import ca.uwaterloo.flix.language.ast.{Level, Purity, Symbol}
+import ca.uwaterloo.flix.util.{InternalCompilerException, ParOps}
+
+import scala.collection.mutable
 
 
 object EffectBinder {
@@ -27,7 +31,6 @@ object EffectBinder {
     * Identity Function.
     */
   def run(root: Root)(implicit flix: Flix): Root = flix.phase("EffectBinder") {
-    implicit val lctx: LocalContext = LocalContext.mk()
     val newDefs = ParOps.parMapValues(root.defs)(letBindEffectsDef)
     root.copy(defs = newDefs)
   }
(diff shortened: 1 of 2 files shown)


magnus-madsen:
!!!jar

main/src/ca/uwaterloo/flix/language/phase/EffectBinder.scala
```diff
@@ -123,69 +129,160 @@ object EffectBinder {
...
+      case Expr.NewObject(_, _, _, _, _, _, _) => letBindExpr(binders)(e)
+    }
+    bind(expTransformed)
+  }
+
+  private def letBindExpr(binders: mutable.ArrayBuffer[Expr])(e: Expr)(implicit lctx: LocalContext, flix: Flix): Expr.Var = {
+    val loc = e.loc.asSynthetic
+    val sym = Symbol.freshVarSym("anf", BoundBy.Let, loc)(Level.Default, flix)
+    lctx.lparams.addOne(LocalParam(sym, e.tpe))
+    binders.addOne(Expr.Let(sym, e, null, null, null, loc))
```
magnus-madsen, line 284: What are these null values? Isn't that a disaster waiting to happen?

magnus-madsen: I assume they only exist temporarily, but could we have some documentation + examples + a helper function that creates them and explains?

magnus-madsen: It seems to me  you could create your own type:

```
Binder = LetBinder | LetRecBinder | LocalParamBinder
```

to avoid this?

main/src/ca/uwaterloo/flix/language/phase/EffectBinder.scala
```diff
@@ -123,69 +129,160 @@ object EffectBinder {
...
+      val e = letBindEffects(binders)(exp)
+      Expr.Resume(e, tpe, loc)
+  }
+
+  /**
+    * Let-binds all sub-expressions including the top-most one.
+    * Always returns a variable or a constant.
+    */
+  private def letBindEffects(binders: mutable.ArrayBuffer[Expr])(exp: Expr)(implicit lctx: LocalContext, flix: Flix): Expr = {
+    val expTransformed = letBindInnerEffects(binders)(exp)
```
magnus-madsen, line 251: This variable is used in the call on line 277, so it would be meaningful to move it there to make clear both what 277 does and also that thebind function does not close over it.

magnus-madsen:
Oh it would be nice to have "EffectBinder" in `ShowAstProvider`.

main/src/ca/uwaterloo/flix/language/phase/EffectBinder.scala
JonathanStarup: done

magnus-madsen:
RFM?

magnus-madsen:
!!!jar

PR #6820, opened 2023-12-03
fix: correct bool-op shortcircuit handling

### LINKED ISSUE #6813: Issue in codegen
https://github.com/mlutze/flix-json/actions/runs/7073758873/job/19254066724#step:4:1

JonathanStarup:
fixes  #6813

And and Or use shortcircuiting which was not represented in the translation. luckily, in the code gen for these operations, the right operand is not on the stack when the left operand is evaluated, so translation is like an if-statement.

I've verified that no tests crash, but some still fail, probably due to https://github.com/mlutze/flix-json/issues/75


magnus-madsen:
Let's add at least one or two tests

magnus-madsen (approved):

PR #6834, opened 2023-12-04
feat: lex `CommentDoc`

herluf-ba:
Add support for lexing doc comments. Any comment starting with `///` is considered a doc-comment.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
@@ -331,7 +331,11 @@ object Lexer {
         TokenKind.Hash
       }
       case '/' => if (peek() == '/') {
-        acceptLineComment()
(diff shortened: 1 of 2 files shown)


main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -331,7 +331,11 @@ object Lexer {
         TokenKind.Hash
       }
       case '/' => if (peek() == '/') {
-        acceptLineComment()
+        if (peekPeek().contains('/')) {
```
magnus-madsen, line 334: contains or equals?

herluf-ba: `peekPeek()` returns an `Option[Char]` so I think `contains` is right.

magnus-madsen (approved):

PR #6842, opened 2023-12-04
feat: drop unused `UndefinedVar`

magnus-madsen:


magnus-madsen:
The last remaining case cannot happen because we collect cparams in an earlier phase.

mlutze (approved):

PR #6843, opened 2023-12-04
feat: recoverable `UnderAppliedTypeAlias`

magnus-madsen:


main/src/ca/uwaterloo/flix/language/phase/unification/BoolUnification.scala
```diff
@@ -51,6 +51,13 @@ object BoolUnification {
       return Ok((Substitution.empty, List(Ast.BroadEqualityConstraint(tpe1, tpe2))))
     }
 
+    // Check for Type.Error
```
mlutze, line 54: I think we need to do this recursively in the type. Otherwise we get crashes if there's something nested and we can't translate it.

main/src/ca/uwaterloo/flix/language/phase/unification/Unification.scala
```diff
@@ -135,6 +135,10 @@ object Unification {
 
     case (_, _: Type.AssocType) => Result.Ok(Substitution.empty, List(Ast.BroadEqualityConstraint(tpe1, tpe2)))
 
+    case (Type.Cst(TypeConstructor.Error(k1), _), t2) if k1 == t2.kind => Result.Ok(Substitution.empty, Nil)
```
mlutze, line 138: This might be too strict. `KindUnification` is probably better here.

magnus-madsen: Why? Now we infer the kinds.

main/src/ca/uwaterloo/flix/language/phase/unification/BoolUnification.scala
magnus-madsen: Yikes. That makes this even harder / costly.

magnus-madsen: This seems hard, so I created a ticket: https://github.com/flix/flix/issues/6844

magnus-madsen:
Can you approve? (This is still an improvement over master).

main/src/ca/uwaterloo/flix/language/phase/unification/Unification.scala
mlutze: Because kind inference is not complete. I believe there are a few cases where Kind.Wild gets by.

mlutze (approved):

main/src/ca/uwaterloo/flix/language/phase/unification/Unification.scala
magnus-madsen: Hmm

PR #6862, opened 2023-12-08
fix: safe quicksortPartition

### LINKED ISSUE #6861: java.lang.ArrayIndexOutOfBoundsException when calling List.sort on custom enum with Order
I ran into this issue while solving [advent of code 2023 day 7](adventofcode.com/2023/day/7) using Flix version 0.42.0. I was still working on the code so the `Order` was not well defined, but I would still expect it to run and return *some* kind of list (even if it's in a weird order) instead of running into index out of bounds exception.

Here's a sort of minimized reproduction:

```flix
enum Hand with ToString, Eq {
    case FiveOfAKind(String)
    case FourOfAKind(String)
    case FullHouse(String)
    case ThreeOfAKind(String)
    case TwoPair(String)
    case OnePair(String)
    case HighCard(String)
}

instance Order[Hand] {
    pub def compare(x: Hand, y: Hand): Comparison =
        match (x, y) {
            case (Hand.FiveOfAKind(_), Hand.FiveOfAKind(_)) => Comparison.GreaterThan
            case _ => Comparison.LessThan
        }
}

def main(): Unit \ IO =
    let testHands = Hand.FiveOfAKind("JJJJJ") :: Hand.HighCard("29QA4") :: Hand.TwoPair("6A9A9") :: Nil;
    testHands |> List.sort |> ToString.toString |> println
```

It crashes with the following error:

```
java.lang.ArrayIndexOutOfBoundsException: Index 3 out of bounds for length 3                
	at Array.Def$swap$exclamation$405956.applyFrame(Unknown Source)
	at Array.Def$swap$exclamation$405956.invoke(Unknown Source)
	at Array.Def$sortWithin$exclamation$405879.applyFrame(Unknown Source)
	at Array.Def$sortWithin$exclamation$405879.invoke(Unknown Source)
	at List.Def$sort$405690.applyFrame(Unknown Source)
	at List.Def$sort$405690.invoke(Unknown Source)
	at Def$main$.applyFrame(Unknown Source)
	at Def$main$.invoke(Unknown Source)
	at Ns.m_main$(Unknown Source)
	at Main.main(Main)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at ca.uwaterloo.flix.language.phase.jvm.Loader$.mainFunction$1(Loader.scala:73)
	at ca.uwaterloo.flix.language.phase.jvm.Loader$.$anonfun$load$10(Loader.scala:82)
	at ca.uwaterloo.flix.language.phase.jvm.Loader$.$anonfun$load$10$adapted(Loader.scala:82)
	at ca.uwaterloo.flix.runtime.shell.Shell.run(Shell.scala:377)
	at ca.uwaterloo.flix.runtime.shell.Shell.execEval(Shell.scala:305)
	at ca.uwaterloo.flix.runtime.shell.Shell.execReloadAndEval(Shell.scala:320)
	at ca.uwaterloo.flix.runtime.shell.Shell.execute(Shell.scala:162)
	at ca.uwaterloo.flix.runtime.shell.Shell.loop(Shell.scala:113)
	at ca.uwaterloo.flix.Main$.main(Main.scala:240)
	at ca.uwaterloo.flix.Main.main(Main.scala)
```

JonathanStarup:
Fixes #6861

functionally equivalent for correct definitions of `cmp` but will not crash on functions that have `s <=> s == LessThan`


magnus-madsen:
I am not sure I like this. An order implementation could be broken in several ways.

We have now added a call to `min` which replaces an exception by silent wrong behavior for _one_ function for _one_ type of error. What about all the other functions that rely on Order? Why is silent wrong behavior better than a crash?

JonathanStarup:
@magnus-madsen I think it's a bit better than what you describe; this makes it so that no order will ever crash the sort function. surely that's good, right? I get the exception vs silent point, but this crash is so specific that it doesn't seem useful, especially since the crash is so unclear (array out of bounds)
- "one function" all sort functions on most data structures pretty much
- "one type of error" technically, yes, but now all errors in an ordering give non-crash, arbitrary permutation - making error behavior more consistent, not less

magnus-madsen:
Why is silent corruption better?

JonathanStarup:
95 % silent corruption with 5 % out of bounds seems worse than 100% silent corruption, especially since those 5% are not easily defined.

I guess we could replace the `min` with an if that throws a better exception

magnus-madsen:
But why these 5%? The are lots of other functions you can change too? And why pay a performance cost?

PR #6864, opened 2023-12-08
refactor: `Resolver.getJVMType` to return `Result`

magnus-madsen:


main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -3195,7 +3195,7 @@ object Resolver {
     * Performs name resolution on the given `signature`.
     */
   private def lookupSignature(signature: List[UnkindedType], loc: SourceLocation)(implicit flix: Flix): Validation[List[Class[_]], ResolutionError] = {
-    traverse(signature)(getJVMType(_, loc))
+    Result.traverse(signature)(getJVMType(_, loc)).toValidation
```
mlutze, line 3198: This should traverse as validation to collect all the errors

mlutze (approved):

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
magnus-madsen: I will be replacing lookupSignature as something that returns `Result` in my next PR.

Issue #6866, opened 2023-12-10
Typing "use x;y" in a function crashes the extension


Andriamanitra:
If you type `use x;y` in a function the VSCode extension will crash and you will need to restart the editor to get it working again.

Version info:
* OS: 64-bit Linux (OpenSUSE Tumbleweed)
* Flix 0.42.0
* vscode-flix v1.11.0

To reproduce:
1. Open a fresh folder in VSCode
2. Create a new main.flix file
3. Type the following code in it:
```
def main(): Unit =
    use x;y
```
4. As soon as you type the `y`, the extension crashes with error message like this:

```
❌ -- Parse Error -------------------------------------------------- file:///tmp/tmp.bGqyyytxDw/main.flix

>> Parse Error: Unexpected end of input, expected UnaryOp1, UnaryOp2 or Ref (line 2, column 11):
    use x;
          ^

Syntactic Context: OtherExpr.

Exception in thread "ForkJoinPool-7-worker-24" java.util.NoSuchElementException: head of empty list
	at scala.collection.immutable.Nil$.head(List.scala:662)
	at scala.collection.immutable.Nil$.head(List.scala:661)
	at ca.uwaterloo.flix.language.phase.Resolver$.tryLookupQualifiedName(Resolver.scala:2751)
	at ca.uwaterloo.flix.language.phase.Resolver$.ca$uwaterloo$flix$language$phase$Resolver$$lookupQualifiedName(Resolver.scala:2797)
	at ca.uwaterloo.flix.language.phase.Resolver$Expressions$.visitExp$1(Resolver.scala:1020)
	at ca.uwaterloo.flix.language.phase.Resolver$Expressions$.resolve(Resolver.scala:1685)
	at ca.uwaterloo.flix.language.phase.Resolver$.$anonfun$resolveDef$1(Resolver.scala:518)
	at ca.uwaterloo.flix.util.Validation$.flatMapN(Validation.scala:502)
	at ca.uwaterloo.flix.language.phase.Resolver$.resolveDef(Resolver.scala:515)
	at ca.uwaterloo.flix.language.phase.Resolver$.visitDecl(Resolver.scala:365)
	at ca.uwaterloo.flix.language.phase.Resolver$.$anonfun$visitUnit$3(Resolver.scala:335)
	at ca.uwaterloo.flix.util.Validation$.$anonfun$fastTraverse$1(Validation.scala:263)
	at scala.collection.immutable.List.foreach(List.scala:333)
[...]

PR #6869, opened 2023-12-11
fix: create unqualified use error

mlutze:
fixes https://github.com/flix/flix/issues/6866


magnus-madsen (approved):

magnus-madsen:
@mlutze Is the community build thing OK?

PR #6873, opened 2023-12-12
feat: Effect Handlers

JonathanStarup:


JonathanStarup:
!!!jar

JonathanStarup:
Okay it seems i was presumptous

JonathanStarup:
!!!jar

JonathanStarup:
@magnus-madsen effects now work! (75% of tests pass, the rest fail because of unresolved erasure issues)

magnus-madsen:
> @magnus-madsen effects now work! (90% of tests pass, the rest fail because of unresolved erasure issues)

Nice! You mean as exceptions or with resumption?

What about the unresolved erasure issues?

main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
```diff
@@ -61,7 +61,7 @@ sealed trait BackendObjType {
     case BackendObjType.Arrays => JvmName(JavaUtil, "Arrays")
     case BackendObjType.StringBuilder => JvmName(JavaLang, "StringBuilder")
     case BackendObjType.Objects => JvmName(JavaLang, "Objects")
-    case BackendObjType.LambdaMetaFactory => JvmName(JavaLangInvoke, "LambdaMetaFactory")
+    case BackendObjType.LambdaMetaFactory => JvmName(JavaLangInvoke, "LambdaMetafactory")
```
magnus-madsen, line 64: 🤦  😆

magnus-madsen:
Not really sure where the errors are now.

JonathanStarup:
> > @magnus-madsen effects now work! (90% of tests pass, the rest fail because of unresolved erasure issues)
> 
> Nice! You mean as exceptions or with resumption?
> 
> What about the unresolved erasure issues?

All effect functionality should work now!

The issue is the conversion where all functions return a Boxed value. Its pretty ad hoc and i think ive done it in a wrong way. in this pr everything is type a -> Obj, but really i should just keep the a -> b type and make that mean "a function that takes a and returns a value which you can extract b from"

magnus-madsen:
Do you have an example? I tried this:

```
eff Throw {
    pub def throw(): Unit
}

def main(): Unit \ IO =
    let r = f();
    println("hello: ${r}")


def f(): String  =
    try {
        f()
    } with Throw {
        def throw(_k) = "a"
    }

def g(): Unit \ Throw = do Throw.throw()

```

magnus-madsen:
I guess we need to lift
```
-- Syntax Error -------------------------------------------------- foo.flix

>> Non-Unit return type. All effect operations must return Unit.

2 |     pub def throw(): Int32
                ^^^^^
                non-Unit return type
```

to be able to write useful programs.

magnus-madsen:
```

**Comment by JonathanStarup** [comment]
Yeah, I've discussed it with Matt. We should draft some type rules and implement them

**Comment by JonathanStarup** [comment]
@magnus-madsen The only thing broken now should be effect operations that return non-object return types (will fix that now in the eraser)

**Comment by magnus-madsen** [comment]
> @magnus-madsen The only thing broken now should be effect operations that return non-object return types (will fix that now in the eraser)

And a conflict.

**Comment by magnus-madsen** [comment]
Are you marking this as RFR?

File: main/src/ca/uwaterloo/flix/language/phase/jvm/GenFunAndClosureClasses.scala [hunk 2]
```diff
@@ -294,5 +327,68 @@ object GenFunAndClosureClasses {
     m.visitEnd()
   }
 
+  private def compileOnCall(v: ClassWriter, classType: JvmType.Reference, defn: Def): Unit = {
```
magnus-madsen, line 330: Is this entire method for debugging?

JonathanStarup:
@magnus-madsen not polished for merging yet, but ready for testing :) no known bugs remain

PR #6877, opened 2023-12-14
feat: backend effect functions

JonathanStarup:
* `ResumptionWrapper` is a helper calls to make a `Resumption` a callable function in flix protocol
* `Effect.op` functions are unused, but will be used to generate code in try-with
* Reducer now also looks for types in effects (this is needed now since the above functions assume the existence of some function interfaces)

From the docs:
```java
eff SomeEffect {
    pub def flip(unit: Unit, cont: Bool -> Result): Value
    pub def add(x: Int32, y: Int32, cont: Int32 -> Result): Value
}
```
is generated like so:
```java
public final class Eff$SomeEffect implements Handler {
    public Fn2$Obj$Obj$Obj flip;
    public Fn3$Int32$Int32&Obj$Obj add;
    public static EffectCall flip(Object var0, Handler h, Resumption r) {
        Fn2$Obj$Obj$Obj f = ((Eff$SomeEffect) h).flip;
        f.arg0 = var0;
        f.arg1 = new ResumptionWrapper(r);
        return f.invoke();
    }
    public static EffectCall add(Int var0, Int var1, Handler h, Resumption r) {
        Fn2$Obj$Obj$Obj f = ((Eff$SomeEffect) h).flip;
        f.arg0 = var0;
        f.arg1 = var1;
        f.arg2 = new ResumptionWrapper(r);
        return f.invoke();
    }
}
```

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
@@ -291,7 +291,20 @@ object Reducer {
         sacc + eType ++ caseTypes
     }
 
-    val result = defTypes ++ enumTypes
+    val effectTypes = root.effects.foldLeft(Set.empty[MonoType]) {
+      case (sacc, (_, e)) =>
+        val opTypes = e.ops.map{
+          case op =>
+            val paramTypes = op.fparams.map(_.tpe)
+            val resType = op.tpe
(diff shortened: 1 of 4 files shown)


magnus-madsen (approved):

JonathanStarup:
@magnus-madsen just fixed a final issue, now all test on my in and it should be ready

PR #6882, opened 2023-12-18
fix: bug in `Lowering` and `Inliner`

magnus-madsen:


magnus-madsen:
@JonathanStarup Hopefully this will help with #6873, but I cannot be sure.

mlutze (approved):

PR #6883, opened 2023-12-18
feat: enable `Verifier`

magnus-madsen:


mlutze (approved):

magnus-madsen:
@JonathanStarup There is still some problems with unsorted records. Maybe you can take a look at that-- as part of your phase re-ordering.

PR #6897, opened 2023-12-19
feat: drop `NonUnitOperationType`

magnus-madsen:


JonathanStarup (approved):
The typing rules are hardcoded with unit multiple places so there's more work

magnus-madsen:
> The typing rules are hardcoded with unit multiple places so there's more work

I don't see anything in the Typer, but maybe it will be come clear later, when we try to write some programs.

PR #6898, opened 2023-12-19
test: add `Test.Handler.ZeroShot.flix`

magnus-madsen:


JonathanStarup (approved):
looks good. There should probably also be tests with try and Do not in tail position

PR #6902, opened 2023-12-20
fix: selective entrypoint generation

### LINKED ISSUE #6901: Filter NS methods
All reachable functions have a `Def$someFunction` class based on the tree shaker. These classes have an `invoke` function that uses our calling convention. For some subset of those functions, we generate a matching `NS.m_someFunction` that calls `Def$someFunction` but unwinds the trampoline. This PR limits those `NS.m_` methods to those that might be used.

JonathanStarup:
Fixes #6901 

Now only tests, benchmarks, and the main function has a `NS.m_<function>` function

The actual change is the two conditionals, but genMain also used the `m_<function>` so now its just inlined


main/src/ca/uwaterloo/flix/language/phase/jvm/Loader.scala
```diff
@@ -34,7 +34,7 @@ object Loader {
       //
       // Decorate each defn in the ast with its method object unless its a closure.
       //
-      for ((sym, defn) <- root.defs if defn.cparams.isEmpty) {
+      for ((sym, defn) <- root.defs if defn.cparams.isEmpty && (defn.ann.isTest || flix.options.entryPoint.contains(defn.sym))) {
```
magnus-madsen, line 37: We also need to check if it has the benchmark annotation.

magnus-madsen: But I think by now we need to introduce a helper function somewhere.

JonathanStarup:
This is needed for handlers btw, since the entrypoint function generation for `pub example(): Unit \ Print` will generate non-valid bytecode (verifier error)

magnus-madsen (approved):

Issue #6746, opened 2023-11-22
Rework Validation


magnus-madsen:
Phase 1:

- [x] Drop all `sequenceT` functions -- they are unused.
- [x] Add a smart constructor `Validation.success(t)`
- [x] Add a smart constructor `Validation.softFailure(t, e)`
- [x] Add a smart constructor `Validation.softFailures(t, es)` -- takes a list of errors
- [x] Add a smart constructor `Validation.hardFailure(e)`
- [x] Add `toResult: Result[(t, List[e]), List[e])` to Validation -- `Success` and `SoftFailure` are both treated as `Result.Ok`
- [x] Drop implicit class `ToSuccess`.
- [x] Drop implicit class `ToSoftFailure`.
- [x] Drop implicit class `ToFailure`.
- [ ] Make `Success` case class private.
- [ ] Make `SoftFailure` case class private.
- [ ] Make `Failure` case class private.
- [ ] Rename `Failure` case class to `HardFailure`.
- [x] Introduce `Recoverable` trait next to `CompilationMessage`
- [x] Introduce `Unrecoverable` trait next to `CompilationMessage`.
- [x] Modify the signature of `softFailure/softFailures` to take a `Recoverable`. Adjust errors as required.
- [x] Modify the signature of `hardFailure/hardFailures` to take an `Unrecoverable`. Adjust errors as required.
- [x] `ap` and `flatten` should be private, if possible.
- [x] `SuccessNil`, `SuccessNone`, and `SuccessUnit` should be private (or removed if unused).
- [ ] `withSoftFailure` should require `Recoverable`
- [ ] `withSoftFailures` should require `Recoverable`
- [x] Rename `get` to `unsafeGet`
- [x] Drop `map` directly on `Validation` and instead introduce `mapN` for a single argument (for consistency).
- [ ] Refactor `Validation.fold` to `Validation.foldLeft`. It should be curried and take a Validation as zero arg. 
- [ ] Ensure every function is tested.
- [ ] Move all private functions last in Validation (public, private) should be the order.
- [ ] Drop `traverseX`. Instead use `traverse` or better `withSoftFailure`...
- [ ] Drop `sequenceX` use `traverse`.
- [ ] Drop `sequence` use `traverse`. 

Multiple meaningful PRs please.

magnus-madsen:
@jaschdoc I think starting to use `Validation.success(foo)` one phase at a time is the next reasonable thing to do.

magnus-madsen:
[...]

PR #6911, opened 2023-12-23
feat: remove `Validation.map`

jaschdoc:


jaschdoc:
Blocked until we resolve https://github.com/flix/flix/issues/6746#issuecomment-1868320988

magnus-madsen:
Suggestion: Can we add `mapN` for a single element and use it everywhere except for `Flix.scala`? Then we can come up with a solution for that place.

magnus-madsen:
Looks good, just needs a conflict resolved.

jaschdoc:
Conflict resolved :)

Also, re-run `semantic commit message` GitHub action... apparently it had an internal server error :/

magnus-madsen (approved):

PR #6912, opened 2023-12-23
refactor: make Validation.Success private

jaschdoc:
- [ ] Refactor `Result.Ok((_, Chain.empty)) => ...` to use one of the newly introduced functions in `Validation`.

Depends on https://github.com/flix/flix/pull/6909


main/src/ca/uwaterloo/flix/Main.scala
```diff
@@ -125,11 +125,14 @@ object Main {
...
 
         case Command.Init =>
-          Bootstrap.init(cwd)(System.err) match {
-            case Validation.Success(_) =>
+          Bootstrap.init(cwd)(System.err).toResult match {
+            case Result.Ok((_, Nil)) =>
               System.exit(0)
-            case failure =>
-              failure.errors.map(_.message(formatter)).foreach(println)
+            case Result.Ok((_, failures)) =>
```
jaschdoc, line 131: Matching on `Result.Ok((_, Nil))` and `Result.Ok((_, failures))`, i.e., list emptiness, is slightly awkward here. Should we just ignore the errors if it is `Ok` or should we try to recover from them somehow? Or is the plan that the compiler automatically recovers during each phase when we are done refactoring?

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -467,9 +467,10 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
...
     reconfigureFlix(flix)
 
-    flix.check() match {
-      case Validation.Success(_) => Validation.success(())
-      case failure => Validation.toHardFailure(BootstrapError.GeneralError(flix.mkMessages(failure.errors)))
+    flix.check().toResult match {
+      case Result.Ok((_, Nil)) => Validation.success(())
+      case Result.Ok((_, errors)) => Validation.toHardFailure(BootstrapError.GeneralError(flix.mkMessages(errors)))
+      case Result.Err(errors) => Validation.toHardFailure(BootstrapError.GeneralError(flix.mkMessages(errors)))
     }
```
jaschdoc, line 474: This should be possible without converting to `Result` first. Maybe something like a `mapErr` or `mapFailure` could do the trick?

jaschdoc:
@magnus-madsen I would like some input on this :)

magnus-madsen:
> @magnus-madsen I would like some input on this :)

What's the question? Just general feedback?

jaschdoc:
> > @magnus-madsen I would like some input on this :)
> 
> What's the question? Just general feedback?

Yes, I asked some questions above in the code.

magnus-madsen:
I forgot to submit my comments 😶

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
jaschdoc: Alright, I'll see if there are any more uses

main/src/ca/uwaterloo/flix/Main.scala
```diff
@@ -124,11 +124,16 @@ object Main {
           }
 
         case Command.Init =>
-          Bootstrap.init(cwd)(System.err) match {
-            case Validation.Success(_) =>
+          Bootstrap.init(cwd)(System.err).toResult match {
+            case Result.Ok((_, Nil)) =>
```
magnus-madsen, line 128: I think in these matches, I would prefer a non-deep match, i.e. match on Ok and Err. Then in the Ok case further split with an if-then-else on if(errors.isEmpty) ...

magnus-madsen: Scratch that. I change my mind.

magnus-madsen:
Any questions that remain?

jaschdoc:
Nope

jaschdoc:
I definitely think we should have a `mapErr` to avoid pattern matching on `Result.Ok((_, Nil))` a lot of places and having 2 error branches:
```scala
v.toResult match {
    case Result.Ok((_, Nil)) => // Handle success
    case Result.Ok((_, failures)) => // Handle failures <--- this is misleading / confusing
    case Result.Err(failures) =>       // Handle failures
}
```

jaschdoc:
Added the following functions to `Validation`:
- `mapFailures`
- `flatMapFailures`
- `onSuccess`
- `onFailure`

jaschdoc:
Starting over from new PR

Issue #6758, opened 2023-11-28
Replace Scala's `LazyList` in `Validation` with our own `Chain`


magnus-madsen:
I am tired of Scalas non-sense. We just need a `Chain` with fast concat.

magnus-madsen:
@jaschdoc This is a nice simple refactoring which probably does not clash with anything else going on.

You might want to add a constructor to `Chain` that takes a `Sequence[A]`

PR #6915, opened 2023-12-24
refactor: replace LazyList with Chain in Validation

jaschdoc:
Closes https://github.com/flix/flix/issues/6758


jaschdoc:
Okay, getting stack overflow on simple equality checks... I have no idea what's going on :thinking:

main/src/ca/uwaterloo/flix/util/collection/Chain.scala
```diff
@@ -47,6 +47,25 @@ sealed trait Chain[+A] extends Iterable[A] {
...
+    * Returns the amount of elements in the chain.
+    */
+  def length: Int = this match {
+    case Chain.Empty => 0
+    case Chain.Link(l, r) => l.length + r.length
+    case Chain.Many(cs) => cs.map(_.length).sum
+    case Chain.Proxy(xs) => xs.length
+  }
+
+  override def equals(obj: Any): Boolean = obj match {
```
magnus-madsen, line 61: I think this risks being too slow. Instead do toList == toList

Also, add a hashcode

main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
```diff
@@ -968,7 +969,7 @@ object Kinder {
       val pvar = Type.freshVar(Kind.Eff, m.loc)
       // Note: We must NOT use [[Validation.toSoftFailure]] because
       // that would duplicate the error inside the Validation.
-      Validation.SoftFailure(KindedAst.Expr.Error(m, tvar, pvar), LazyList.empty)
+      Validation.SoftFailure(KindedAst.Expr.Error(m, tvar, pvar), Chain.empty)
```
magnus-madsen, line 972: This is using the constructor directly, but I guess we will deal with that later and separately.

main/test/ca/uwaterloo/flix/util/TestChain.scala
```diff
@@ -25,6 +25,15 @@ class TestChain extends AnyFunSuite {
...
 
+  test("TestChainReflexive.01") {
+    assertResult(Chain.empty)(Chain.empty)
+  }
+
+  test("TestChainReflexive.02") {
+    val result = Chain.empty == Chain.empty
+    assert(result)
+  }
+
```
magnus-madsen, line 36: A few more tests might be useful.

I guess they should be called TestEq

main/src/ca/uwaterloo/flix/util/collection/Chain.scala
```diff
@@ -47,6 +47,25 @@ sealed trait Chain[+A] extends Iterable[A] {
     * The empty chain.
     */
   override val empty: Chain[A] = Chain.Empty
+
+  /**
+    * Returns the amount of elements in the chain.
+    */
+  def length: Int = this match {
```
magnus-madsen, line 54: Where do we use this?

main/src/ca/uwaterloo/flix/util/collection/Chain.scala
magnus-madsen: Because I think calling ViewLeft repeatedly has cost O(n ** 2)

main/src/ca/uwaterloo/flix/util/collection/Chain.scala
```diff
@@ -74,7 +93,17 @@ object Chain {
   /**
     * Returns a chain containing the given elements.
     */
-  def apply[A](xs: A*): Chain[A] = Chain.Proxy(xs)
+  def apply[A](xs: A*): Chain[A] = if (xs.isEmpty) Chain.empty else Chain.Proxy(xs)
+
+  /**
+    * Returns a chain containing the given elements.
+    */
+  def from[A](xs: Seq[A]): Chain[A] = if (xs.isEmpty) Chain.empty else Chain.Proxy(xs)
```
magnus-madsen, line 101: Nice

main/src/ca/uwaterloo/flix/util/collection/Chain.scala
```diff
@@ -85,4 +114,30 @@ object Chain {
     * Concatenates the given sequence of chains, in order.
     */
   def concat[A](cs: Seq[Chain[A]]): Chain[A] = Chain.Many(cs)
+
+  private sealed trait ViewLeft[+A] {
```
magnus-madsen, line 118: I think this can be dropped, because we should just use toList in equals

main/src/ca/uwaterloo/flix/util/collection/Chain.scala
magnus-madsen: I would add a toList (written in CPS) if there is not already one, and then we can use that.

jaschdoc: Alright roger that. We could also just concatenate all of the inner `Seq`s and compare those?

jaschdoc: The current problem is that `Iterable` is **not reflexive**.

main/src/ca/uwaterloo/flix/util/collection/Chain.scala
jaschdoc: Line 87 in `Validation`:

```scala
case Validation.HardFailure(errors) if errors.length == 1 =>
```

main/src/ca/uwaterloo/flix/util/collection/Chain.scala
magnus-madsen: If converted to a list that problem goes away.

jaschdoc:
RFR

magnus-madsen:
I see some test failures.

jaschdoc:
Fixed it :+1:

magnus-madsen:
@mlutze extra pair of eyes.

main/src/ca/uwaterloo/flix/util/collection/Chain.scala
```diff
@@ -47,34 +47,97 @@ sealed trait Chain[+A] extends Iterable[A] {
...
-  private case object Empty extends Chain[Nothing]
+  private case object Empty extends Chain[Nothing] {
+
+    /**
+      * Returns `this` as a [[List]].
+      *
+      * NB.: This method must be overridden in each case class,
+      * otherwise it results in a [[StackOverflowError]].
+      */
+    override def toList: List[Nothing] = Nil
```
mlutze, line 82: Why would a match cause an overflow?

main/src/ca/uwaterloo/flix/util/collection/Chain.scala
```diff
@@ -47,34 +47,97 @@ sealed trait Chain[+A] extends Iterable[A] {
...
+      * NB.: This method must be overridden in each case class,
+      * otherwise it results in a [[StackOverflowError]].
+      */
+    override def toList: List[A] = xs.toList
+  }
+
+  /**
+    * Returns a chain containing the given elements.
+    */
+  def apply[A](xs: A*): Chain[A] = if (xs.isEmpty) Chain.empty else Chain.Proxy(xs)
```
mlutze, line 130: Should we just call `from`?

main/src/ca/uwaterloo/flix/util/collection/Chain.scala
jaschdoc: No clue but there seems to be some kind of circular pattern going on in `Iterable` when not overriding `toList` (also when overriding it in the trait instead of each case class) that causes it

main/test/ca/uwaterloo/flix/util/TestChain.scala
```diff
@@ -54,4 +54,49 @@ class TestChain extends AnyFunSuite {
...
+    assert(c1 != c2)
+  }
+
+  test("TestEq.07") {
+    val c1 = Chain(1) ++ Chain(3)
+    val c2 = Chain(1, 2)
+    assert(c1 != c2)
+  }
+
+  test("TestEq.08") {
```
magnus-madsen, line 97: I notice that none of these tests check for equality between chains that are of very different shape.

magnus-madsen: Like e.g. a purely left-leaning versus purely right-leaning chain with the same elements.

main/src/ca/uwaterloo/flix/util/collection/Chain.scala
magnus-madsen, line 82: `toList` should be defined on the `Chain` trait.

We never use this OO-style because it splits a method all over the place.

magnus-madsen:
@jaschdoc I added some minor comments which should be easy to fix.

Ill let Matt have a look over it too and then I think we are good to go.

main/src/ca/uwaterloo/flix/util/collection/Chain.scala
jaschdoc: Yes but that causes a stack overflow for some reason. I think it's related to dynamic method lookup and to prevent the stack overflow the method must be defined directly on the case class.

main/test/ca/uwaterloo/flix/util/TestChain.scala
jaschdoc: Ah yes, I'll look into it.

main/src/ca/uwaterloo/flix/util/collection/Chain.scala
jaschdoc: See https://github.com/flix/flix/pull/6915#discussion_r1443985120

magnus-madsen: No, thats not possible. There must a bug in the implementation then.

magnus-madsen: Just define `toList` on the trait and pattern match on `this`.

jaschdoc: Don't kill the messenger, but that's what happens :)

```
Exception in thread "ForkJoinPool-2-worker-2" java.lang.StackOverflowError
	at ca.uwaterloo.flix.util.collection.Chain.equals$(Chain.scala:73)
	at ca.uwaterloo.flix.util.collection.Chain$Empty$.equals(Chain.scala:84)
	at ca.uwaterloo.flix.util.collection.Chain.toList(Chain.scala:50)
	at ca.uwaterloo.flix.util.collection.Chain.toList$(Chain.scala:49)
	at ca.uwaterloo.flix.util.collection.Chain$Empty$.toList(Chain.scala:84)
	at ca.uwaterloo.flix.util.collection.Chain.equals(Chain.scala:74)
	at ca.uwaterloo.flix.util.collection.Chain.equals$(Chain.scala:73)
	at ca.uwaterloo.flix.util.collection.Chain$Empty$.equals(Chain.scala:84)
	at ca.uwaterloo.flix.util.collection.Chain.toList(Chain.scala:50)
	at ca.uwaterloo.flix.util.collection.Chain.toList$(Chain.scala:49)
	at ca.uwaterloo.flix.util.collection.Chain$Empty$.toList(Chain.scala:84)
	at ca.uwaterloo.flix.util.collection.Chain.equals(Chain.scala:74)
	at ca.uwaterloo.flix.util.collection.Chain.equals$(Chain.scala:73)
	at ca.uwaterloo.flix.util.collection.Chain$Empty$.equals(Chain.scala:84)
	at ca.uwaterloo.flix.util.collection.Chain.toList(Chain.scala:50)
	at ca.uwaterloo.flix.util.collection.Chain.toList$(Chain.scala:49)
	at ca.uwaterloo.flix.util.collection.Chain$Empty$.toList(Chain.scala:84)
	at ca.uwaterloo.flix.util.collection.Chain.equals(Chain.scala:74)
	at ca.uwaterloo.flix.util.collection.Chain.equals$(Chain.scala:73)
	at ca.uwaterloo.flix.util.collection.Chain$Empty$.equals(Chain.scala:84)
	at ca.uwaterloo.flix.util.collection.Chain.toList(Chain.scala:50)
	at ca.uwaterloo.flix.util.collection.Chain.toList$(Chain.scala:49)
	at ca.uwaterloo.flix.util.collection.Chain$Empty$.toList(Chain.scala:84)
	at ca.uwaterloo.flix.util.collection.Chain.equals(Chain.scala:74)
	at ca.uwaterloo.flix.util.collection.Chain.equals$(Chain.scala:73)
	at ca.uwaterloo.flix.util.collection.Chain$Empty$.equals(Chain.scala:84)
	at ca.uwaterloo.flix.util.collection.Chain.toList(Chain.scala:50)
	at ca.uwaterloo.flix.util.collection.Chain.toList$(Chain.scala:49)
	at ca.uwaterloo.flix.util.collection.Chain$Empty$.toList(Chain.scala:84)
	at ca.uwaterloo.flix.util.collection.Chain.equals(Chain.scala:74)
	at ca.uwaterloo.flix.util.collection.Chain.equals$(Chain.scala:73)
	at ca.uwaterloo.flix.util.collection.Chain$Empty$.equals(Chain.scala:84)
	at ca.uwaterloo.flix.util.collection.Chain.toList(Chain.scala:50)
	at ca.uwaterloo.flix.util.collection.Chain.toList$(Chain.scala:49)
[...]
```

jaschdoc: Ok well, tests seem to be working on GH so it must be my machine again hehe

jaschdoc: I have tried debugging it and it seems something weird happens in `Iterable` and I'm not sure why. It's definitely a possibility that my code is faulty, but it seems simple enough to understand by reading it.

magnus-madsen: Ah. I did not realize that Chain extends Iterable. I think we should drop that-- Scala collections are notoriously brittle.

jaschdoc:
Removed the `Iterable` implementation, added custom functions for the stuff that depended on `Iterable` on the rest of the compiler. However, it still throws a stack overflow error (same issue as before).
Even though the OO-pattern is slightly ugly and goes against the style, it solves the issue with little extra code and keeps the code spill small too. We have a working solution at commit b081e45.

magnus-madsen:
> Removed the Iterable implementation, added custom functions for the stuff that depended on Iterable on the rest of the compiler. However, it still throws a stack overflow error (same issue as before).

Plot thickens. We are missing something obvious. Maybe post on flightdeck?

main/src/ca/uwaterloo/flix/util/collection/Chain.scala
```diff
@@ -43,10 +43,81 @@ sealed trait Chain[+A] extends Iterable[A] {
...
+  def length: Int = this match {
+    case Chain.Empty => 0
+    case Chain.Link(l, r) => l.length + r.length
+    case Chain.Many(cs) => cs.map(_.length).sum
+    case Chain.Proxy(xs) => xs.length
+  }
+
+  override def hashCode(): Int = this.toList.hashCode()
+
+  override def equals(obj: Any): Boolean = obj match {
```
magnus-madsen, line 117: Try to declare this final.

magnus-madsen:
Can you declare equals final? I am guessing the answer is no which solves the riddle :)

jaschdoc:
Still does not work... wild.

magnus-madsen:
> Still does not work... wild.

Hm. What if you make the case classes into non-case classes?

magnus-madsen:
Ill ask around tomorrow about the mystery. Meanwhile, add a few tests for each of the new functions?

main/src/ca/uwaterloo/flix/util/collection/Chain.scala
```diff
@@ -43,10 +43,81 @@ sealed trait Chain[+A] extends Iterable[A] {
...
+    case Chain.Proxy(xs) => xs.headOption
+  }
+
+  /**
+    * Returns `this` as a [[List]].
+    */
+  def toList: List[A] = this match {
+    case Chain.Empty => List.empty
+    case Chain.Link(l, r) => l.toList ++ r.toList
+    case Chain.Many(cs) => cs.flatMap(_.toList).toList
```
magnus-madsen, line 66: I think this flatMap call is sus. 

But even if it is not, its not great performance. 

I suggest, lets create a ListBuffer, call foreach on this and add each element to the buffer. Then turn the buffer into a list.

main/src/ca/uwaterloo/flix/util/collection/Chain.scala
```diff
@@ -43,10 +43,81 @@ sealed trait Chain[+A] extends Iterable[A] {
...
+    case Chain.Link(l, _) => l.head
+    case Chain.Many(cs) => cs.find(_.head.isDefined).flatMap(_.head)
+    case Chain.Proxy(xs) => xs.headOption
+  }
+
+  /**
+    * Returns `this` as a [[List]].
+    */
+  final def toList: List[A] = this match {
+    case Chain.Empty => List.empty
```
mlutze, line 64: My untested theory is that this match checks whether `this == Chain.Empty`, which causes an infinite loop due to the `toList` call in `equals`.

Can't we sidestep all of this by inheriting Iterable and then just doing `Iterable.iterate.toList`?

magnus-madsen: I would prefer not to depend on Scalas broken collections.

jaschdoc: > My untested theory is that this match checks whether `this == Chain.Empty`, which causes an infinite loop due to the `toList` call in `equals`.

Forgive me, but why is it necessarily a problem? Or rather, why does it cause an infinite loop?

mlutze: `this == Chain.Empty` calls `this.toList` which calls `this == Chain.Empty` which calls...

magnus-madsen: Amazing.

jaschdoc: Wow...

jaschdoc: That also explains why the OO implementation works.

magnus-madsen: I think we owe Matt a beer.

jaschdoc: > Can't we sidestep all of this by inheriting `Iterable` and then just doing `Iterable.iterate.toList`?

Hmm... I tried experimenting with this but it doesn't seem like we can do that without using `equals` implicitly somewhere.

mlutze: > > Can't we sidestep all of this by inheriting `Iterable` and then just doing `Iterable.iterate.toList`?
> 
> Hmm... I tried experimenting with this but it doesn't seem like we can do that without using `equals` implicitly somewhere.

Can you hack the `unapply` function to do `instanceof` instead?

jaschdoc: I'll look into it

magnus-madsen:
Does the mutable non-oo solution not work?

main/src/ca/uwaterloo/flix/util/collection/Chain.scala
magnus-madsen: At the risk of sounding like old man yells at cloud: I dont want to extend Iterable and I don't want to have OO-style.

magnus-madsen: CleanShot 2024-01-08 at 18 58 52@2x

magnus-madsen: Why do we need equals on Chain? Maybe thats what we should drop?

mlutze: If we don't redefine it then equals will be structural equality which may be weird.

I disagree with Iterable being bad because it's a very simple interface but I won't fight anyone on it.

IMO just overriding `unapply` will fix everything.

magnus-madsen: We could also drop case classes?

jaschdoc:
> Does the mutable non-oo solution not work?

Which one was that again?

jaschdoc:
I have a reflection-based impl of `toList` on the `Chain` trait working now. I will clean it up but aad13ca demonstrates that it works.

jaschdoc:
I think 77f362c is a good and simple solution. It's also possible to make this work with extending `Iterable[A]` if we still want that. See c7af3d4

mlutze:
> I think 77f362c is a good and simple solution. It's also possible to make this work with extending `Iterable[A]` if we still want that. See c7af3d4

Did the custom `unapply` not work? Otherwise I'm ok with this solution

mlutze:
Actually can you match on the type only for the`Empty` case?

jaschdoc:
> > I think [77f362c](https://github.com/flix/flix/commit/77f362c755912784c76695bb8e99ef211dfaf114) is a good and simple solution. It's also possible to make this work with extending `Iterable[A]` if we still want that. See [c7af3d4](https://github.com/flix/flix/commit/c7af3d4d36b35b80320690940ffa14cd3ab4a148)
> 
> Did the custom `unapply` not work? Otherwise I'm ok with this solution

Nope

magnus-madsen (approved):

PR #6929, opened 2023-12-27
feat: upgrade Gradle to 8.5

### LINKED ISSUE #6925: Upgrade Gradle to 8.5

magnus-madsen:
Fixes #6925


mlutze (approved):
Why is the jar in the repo? Is that standard?

magnus-madsen:
> Why is the jar in the repo? Is that standard?

Yes! Its Gradle... The design choices are ... Interesting.

PR #6932, opened 2023-12-27
feat: emit Java 21 bytecode

### LINKED ISSUE #6927: Update `JvmTarget` to support Java 21
... and no lower versions.

magnus-madsen:
Fixes #6927


main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
```diff
@@ -772,7 +772,7 @@ object BackendObjType {
       StaticField(this.jvmName, IsPrivate, IsFinal, NotVolatile, "counter", JvmName.AtomicLong.toTpe)
 
     def ArgsField: StaticField =
-      StaticField(this.jvmName, IsPrivate, IsFinal, NotVolatile, "args", BackendType.Array(String.toTpe))
+      StaticField(this.jvmName, IsPrivate, NotFinal, NotVolatile, "args", BackendType.Array(String.toTpe))
```
mlutze, line 775: What's this about?

mlutze (approved):

main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
magnus-madsen: Turns out Java broke backward compat. When assigning a final field, you used to be able to do it from a helper function _in the constructor_. But now it must be directly in the constructor, and not in a helper. Something like this.

PR #6936, opened 2023-12-28
refactor: change module structure of the fixpoint engine, part 1 Datalog datatypes

stephentetley:
See Discussion #6935

This PR puts all the datatypes of the Datalog Ast into a single module. The module has the namespace `Fixpoint.Ast.Datalog`.

`Lowering` and `Flix.scala` needed changing as well as some `use` declarations inside the fixpoint engine (`Compiler`, `Solver`, etc.).

Next step is to put all the Ram Ast datatypes into a single module.


magnus-madsen (approved):

magnus-madsen:
Thanks!

stephentetley:
Oh, `Constraint.flix` should have been deleted. I'll remove it as part of the merging the Ram datatypes.

PR #6937, opened 2023-12-29
refactor: change module structure of the fixpoint engine, part 2 Ram datatypes

stephentetley:
Part 2 of reorganizing the Fixpoint engine modules.

This PR merges all the Ram Ast datatypes into a single module - `Fixpoint.Ast.Ram`.

It also deletes `Constraint.flix` that was missed in the previous PR that merged all the Datalog Ast datatypes.

Next step in the Fixpoint engine reorganization after this - move the Fixpoint "pipeline" modules into the `Fixpoint.Phase` namespace.


magnus-madsen (approved):

magnus-madsen:
Thanks Stephen!

PR #6938, opened 2023-12-29
refactor: change module structure of the fixpoint engine, part 3 Phase namespace

stephentetley:
Fixpoint Engine refactor part 3.

This PR moves the Datalog compiler/transformation passes into the `Fixedpoint.Phase` namespace. It also adds gives the remaining top level Fixpoint modules qualified namespaces e.g. the file `Fixpoint/Solver.flix` previous had the namespace `Fixpoint`, now its namespace is `Fixpoint.Solver`.


magnus-madsen:
Could you run `java -jar flix.jar doc` on an empty project? 

Ideally nothing in `Fixpoint` should some up in the API, assuming everything has been correctly marked `@Internal`.

main/src/library/Fixpoint/Interpreter.flix
```diff
@@ -14,19 +14,19 @@
...
+    pub type alias Database[v: Type, r: Region] = MutMap[RamSym[v], MutMap[Vector[v], v, r], r]
+    pub type alias SearchEnv[v: Type, r: Region] = (Array[Vector[v], r], Array[v, r])
 
-    def interpret(rc: Region[r], stmt: RamStmt[v]): Database[v, r] \ r with Order[v] =
+    pub def interpret(rc: Region[r], stmt: RamStmt[v]): Database[v, r] \ r with Order[v] =
         interpretWithDatabase(rc, MutMap.new(rc), stmt)
 
-    def interpretWithDatabase(rc: Region[r], db: Database[v, r], stmt: RamStmt[v]): Database[v, r] \ r with Order[v] =
-        masked_cast(notifyPreInterpret(stmt));
+    pub def interpretWithDatabase(rc: Region[r], db: Database[v, r], stmt: RamStmt[v]): Database[v, r] \ r with Order[v] =
```
magnus-madsen, line 28: For example, this may need to be marked `@Internal`

stephentetley: I will do, I'm making a jar now...

magnus-madsen: It seems we did not manage on master either: https://api.flix.dev/Fixpoint.html

magnus-madsen: It is minor point, in any case.

stephentetley:
Quite a few things needed `@Internal` so I'll update the PR.

Is there anything we can do type aliases, it looks like they can't be internal?

magnus-madsen:
> Is there anything we can do type aliases, it looks like they can't be internal?

Sounds like an oversight. If you create a ticket we can look at that separately.

stephentetley:
Right, I've updated the PR with `@Internal` annotations. 

The two public type aliases in `Fixpoint.Interpreter` can't be made `@Internal` - on a type alias `@Internal` causes a parse error.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -45,14 +45,14 @@ object Lowering {
   private object Defs {
     lazy val Box: Symbol.DefnSym = Symbol.mkDefnSym("Boxable.box")
 
-    lazy val Solve: Symbol.DefnSym = Symbol.mkDefnSym("Fixpoint.solve")
-    lazy val Merge: Symbol.DefnSym = Symbol.mkDefnSym("Fixpoint.union")
-    lazy val Filter: Symbol.DefnSym = Symbol.mkDefnSym("Fixpoint.project")
-    lazy val Rename: Symbol.DefnSym = Symbol.mkDefnSym("Fixpoint.rename")
+    lazy val Solve: Symbol.DefnSym = Symbol.mkDefnSym("Fixpoint/Solver.solve")
```
magnus-madsen, line 48: I probably missed this early. But technically these paths should be `Fixpoint.Solver.solve` ,right?

magnus-madsen: I am surprised the old-style still works

magnus-madsen: I created a ticket. This seems to be a separate issue: https://github.com/flix/flix/issues/6940

magnus-madsen (approved):

magnus-madsen:
Thanks!

PR #6942, opened 2023-12-30
refactor: disable safe mode in documentation parsing

sockmaster27:
Should be redundant when using panic mode

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/HtmlDocumentor.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/HtmlDocumentor.scala
@@ -1183,7 +1183,6 @@ object HtmlDocumentor {
     // Panic mode will escape all < and > characters
     val config =
       txtmark.Configuration.builder()
-        .enableSafeMode()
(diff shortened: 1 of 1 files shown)


magnus-madsen:
Does this resolve the escaping issue? Whats the potential consequence/harm?

sockmaster27:
No, it was just something I uncovered while looking at it. As far as I can tell, this should not have any consequences, since you shouldn't be able to inject anything without `<` and `>`.

magnus-madsen:
OK, and now the HTML render correctly?

sockmaster27:
No, it renders identically.

magnus-madsen (approved):

magnus-madsen:
OK, so more work is needed to fix the problem?

sockmaster27:
> OK, so more work is needed to fix the problem?

Yeah, I propose we switch to JetBrains' version as it seems significantly better maintained, hopefully saving us from future headaches.

magnus-madsen:
Lets give it a try at least.

PR #6946, opened 2023-12-31
refactor: swap out markdown parser

sockmaster27:
Swapped out the markdown parser in an attempt to fix the escaping problems, which did not work.
Leaving this here if we ever decided to switch for other reasons.


magnus-madsen:
Could you explain why it did not work? :)

sockmaster27:
Because it has no option to do any HTML-escaping, so that has to be done manually, except for code blocks where it always escapes both `<`, `>`, `&` and I believe also `"` and `'`.
So it's easier just to use the more minimal txtmark and unescape the `&`.

PR #6947, opened 2023-12-31
fix: escaping of doc comments

### LINKED ISSUE #6918: Escape for flixdoc
Issue with html escaping of some chars, see e.g.
https://api.flix.dev/Monad.html

sockmaster27:
Fixes #6918

This just escapes all `&amp;` instances once, but it should actually just be done inside code blocks.
This means that outside of code blocks, writing `&` and `&amp;` will both just result in a `&`, but I hope that's acceptable for now.


magnus-madsen:
So e.g. the Monad page looks correct now?

sockmaster27:
Yep
image

magnus-madsen (approved):

PR #6950, opened 2024-01-02
fix: bug in `IDE.flix`

magnus-madsen:


olhotak (approved):
Yes, top is correct. It was top in previous versions of this IDE example.

Also, it's correct that x is -9 in n2, n3, and emain, and everything
else is Top.

mlutze:
Oops, hit the merge button of review button.

PR #6955, opened 2024-01-03
feat: cast local variables once

JonathanStarup:
Instead of having erased vars in functions, we will cast the erased def fields once at the beginning and use them directly from the stack. Right now this requires casts on store but later that wont be neccesary

code on main
```
def example {
  Object var0 = this.arg0;
  ((String) var0).concat((String) var0)
}
```
This PR
```
def example {
  String var0 = (String) this.arg0;
  var0.concat(var0)
}
```


magnus-madsen:
How comfortable are you that this is correct? :)

JonathanStarup:
95%. we control access to local variables so casting stores instead of loads should be equivalent

magnus-madsen (approved):

PR #6961, opened 2024-01-03
feat: remove boxing

JonathanStarup:


magnus-madsen:
Conflicted.

magnus-madsen:
How can you get away with removing boxing? What should happen when we have an Int32 and call a Java function that wants an Integer?

JonathanStarup:
Its unused so this is a non-change. I think you'd have to box in surface syntax

magnus-madsen (approved):

PR #6967, opened 2024-01-04
refactor: explicit unit tag optimization

JonathanStarup:
This boolean (which in current is just `tpe == MonoType.Unit` will allow this optimization to continue even with erasure.

It would be nicer to have this information in the structure of the case but the wide span of this AST makes it cumbersome. This solution is in #6968


magnus-madsen:
What is the problem with checking if `tpe == MonoType.Unit`?

JonathanStarup:
@magnus-madsen the erasure phase will make `MonoType.Unit` become `MonoType.Object`, making it false always.

JonathanStarup:
Im beginning to be a bit unsure why we do erasure "so soon"

magnus-madsen (approved):

magnus-madsen:
This is broken now. Perhaps it can be done as part of the reordering of backend phases? 

(Closing since this is less than 15 lines).

PR #6968, opened 2024-01-04
refactor: explicit unit tag optimization

JonathanStarup:
Alternative to #6967


magnus-madsen:
I was going to go with the other one.

PR #6979, opened 2024-01-08
refactor: dont flip boolean lattice

JonathanStarup:
The internal booleans use a flipped lattice but this only adds confusion


magnus-madsen:
I don't understand how this bug - if present - would not trigger several test cases?

magnus-madsen:
I want Matt to review this, so I can have https://en.wikipedia.org/wiki/Guilt_by_Association

magnus-madsen:
If you flip the lattice will this not have performance implications here:

https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/unification/EffUnification.scala#L83-L128

You need to run the profiler and see what it says.

PR #6981, opened 2024-01-08
feat: add `Root.reachable` set

### LINKED ISSUE #6978: Figure out a proper way to deal with entry points
E.g. dont have all these scattered isBenchmark, isTest, etc.

magnus-madsen:
Fixes #6978

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/EntryPoint.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/EntryPoint.scala
@@ -23,6 +23,8 @@ import ca.uwaterloo.flix.util.Validation.{flatMapN, mapN}
 import ca.uwaterloo.flix.util.collection.ListMap
 import ca.uwaterloo.flix.util.{InternalCompilerException, Validation}
 
+import scala.collection.mutable
+
 /**
   * Processes the entry point of the program.
   *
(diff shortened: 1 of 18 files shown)


main/src/ca/uwaterloo/flix/language/phase/EntryPoint.scala
```diff
@@ -70,12 +71,26 @@ object EntryPoint {
...
       // Case 2: No entry point. Don't touch anything.
-      case None => Validation.success(root)
+      case None => Validation.success(root.copy(reachable = getReachable(root)))
+    }
+  }
+
+  /**
+   * Returns all reachable definitions.
+   */
+  private def getReachable(root: TypedAst.Root): Set[Symbol.DefnSym] = {
```
mlutze, line 86: Seems like this could easily be done functionally, but OK.

mlutze (approved):

main/src/ca/uwaterloo/flix/language/phase/EntryPoint.scala
magnus-madsen: I HAVE BECOME A FAN OF REGIONS

magnus-madsen: Argh, caps.

magnus-madsen: Yeah, its a bit lazy. I guess a filter could have done the job. But I have a map, so I would need a fold I guess.

PR #6982, opened 2024-01-08
fix: try-with typing

JonathanStarup:
This commit fixes the typing of try and its continuation. It simplifies the typing by forcing the try block to have effect `E` (if `E` is the handled effect)


magnus-madsen:
Failing tests.

magnus-madsen (approved):

PR #6996, opened 2024-01-09
feat: allow storing GitHub token in environment variable

### LINKED ISSUE #6971: Use GitHub token if available in environment `GITHUB_TOKEN`
If no GitHub token is specified on the command line, we should check if one is available in the environment variable `GITHUB_TOKEN`.

sockmaster27:
Fixes #6971


magnus-madsen (approved):

magnus-madsen:
Thanks!

Issue #6699, opened 2023-11-13
Add quickfix for `UndefinedName`and misspellings


magnus-madsen:
If the programmer writes:

```
def main(): Unit \ IO = 
    let mississippi = 1;
    println(mississsippi + 1)
```

I would like a quick fix for `UndefinedName` which proposes to fix the spelling on line 3.

Ditto, if the programmer writes:

```
def main(): Unit \ IO = 
    let l = List.range(1, 100);
    println(List.lengt(l))
```

But these cases probably have to be handled separately.

The starting point for this work is in `CodeActionProvider`

PR #6999, opened 2024-01-10
feat: add quickfix for misspellings (Issue #6699)

bathtub-01:
Please check if I am doing things right:
- add a `mkFixMisspelling` function to handle `UndefinedName`
- for now it can only suggest hard-coded changes :)
- the remaining part is to check the namespace of the undefined name and figure out a list of *probably correct* identifiers

I would like to know how you define a "misspelling". Some rules I can think about are:
- only apply this quickfix to names that are longer than 3-4 characters
- a "misspelled" name differs by only 1-2 characters from an existing identifier

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
@@ -322,6 +322,7 @@ object CodeActionProvider {
 
   /**
     * Returns a list of quickfix code action to suggest possibly correct spellings.
+    * Uses Levenshtein Distance to find closed spellings.
     */
   private def mkFixMisspelling(qn: Name.QName, loc: SourceLocation, env: Map[String, Symbol.VarSym], uri: String): List[CodeAction] = {
     if (qn.ident.name.length > 3) {
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
@@ -325,10 +325,13 @@ object CodeActionProvider {
     * Uses Levenshtein Distance to find closed spellings.
     */
   private def mkFixMisspelling(qn: Name.QName, loc: SourceLocation, env: Map[String, Symbol.VarSym], uri: String): List[CodeAction] = {
-    if (qn.ident.name.length > 3) {
-      val possibleNames = env.toList.map(_._1).filter(n => (n.length - qn.ident.name.length).abs < 3)
-      possibleNames.filter(n => Similarity.levenshtein(qn.ident.name, n) < 3).map(mkCorrectSpelling(_, loc, uri))
-    } else
+    val minLength = 3
+    val maxDistance = 3
+    val possibleNames: List[String] = env.toList.map(_._1).filter(n => (n.length - qn.ident.name.length).abs < maxDistance)
+                                                          .filter(n => Similarity.levenshtein(n, qn.ident.name) < maxDistance)
+    if (qn.ident.name.length > minLength)
+      possibleNames.map(n => mkCorrectSpelling(n, loc, uri))
+    else
       Nil
   }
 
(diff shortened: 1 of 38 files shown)


magnus-madsen:
This looks like a good start. The UndefinedName class contains a set of local variables in scope, so you could try to find one that is close to the spelling there.

bathtub-01:
Utilize the `Similarity.levenshtein` function to calculate the Levenshtein Distance between names to find close spellings.

Currently the quickfix only works for non-qualified names. Considering supporting qualified names (like `List.map`).

magnus-madsen:
We could polish this up (maybe its already ready) and merge it in?

main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
```diff
@@ -319,6 +320,35 @@ object CodeActionProvider {
     command = None
   )
 
+  /**
+    * Returns a list of quickfix code action to suggest possibly correct spellings.
+    * Uses Levenshtein Distance to find closed spellings.
+    */
+  private def mkFixMisspelling(qn: Name.QName, loc: SourceLocation, env: Map[String, Symbol.VarSym], uri: String): List[CodeAction] = {
+    if (qn.ident.name.length > 3) {
```
magnus-madsen, line 328: I see the constant 3 repeated a lot. Let make it into a local variable. E.g. could be `MinLength`.

main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
```diff
@@ -319,6 +320,35 @@ object CodeActionProvider {
...
+      val possibleNames = env.toList.map(_._1).filter(n => (n.length - qn.ident.name.length).abs < 3)
```
magnus-madsen, line 329: Does Scala not have a fillterKeys or something? Maybe not?

magnus-madsen: Otherwise a fold might be better? or even an imperative loop buidning up a ListBuffer of CodeActions.

main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
```diff
@@ -319,6 +320,35 @@ object CodeActionProvider {
...
+      Nil
+  }
+
+  /**
+    * Internal helper function for `mkFixMisspelling`.
+    * Returns a quickfix code action for a possibly correct name.
+    */
+  private def mkCorrectSpelling(correct_name: String, loc: SourceLocation, uri: String): CodeAction =
+    CodeAction(
+      title = s"Spelling fix: do you mean `$correct_name`?",
```
magnus-madsen, line 341: Maybe keep the message shorter, e.g. "Did you mean: `$correct....

main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
```diff
@@ -319,6 +320,35 @@ object CodeActionProvider {
...
+      possibleNames.filter(n => Similarity.levenshtein(qn.ident.name, n) < 3).map(mkCorrectSpelling(_, loc, uri))
+    } else
+      Nil
+  }
+
+  /**
+    * Internal helper function for `mkFixMisspelling`.
+    * Returns a quickfix code action for a possibly correct name.
+    */
+  private def mkCorrectSpelling(correct_name: String, loc: SourceLocation, uri: String): CodeAction =
```
magnus-madsen, line 339: correct_name -> correctName

magnus-madsen (changes requested):

main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
bathtub-01: Do you mean something like this to avoid intermediate List construction?

```scala
if (qn.ident.name.length > minLength) 
      env.toList.foldLeft(List.empty[CodeAction]) { (acc, kv) =>
        val name = kv._1
        if ((name.length - qn.ident.name.length).abs < minDistance && Similarity.levenshtein(qn.ident.name, name) < minDistance)
          mkCorrectSpelling(name, loc, uri) :: acc
        else 
          acc
      }
    else 
      Nil
```

magnus-madsen: Hmm. It was not so much about performance. It was about the clarify of the code, but perhaps what you have is clearer than a fold...

magnus-madsen: I will let you choose.

bathtub-01: I would prefer the older one. I have modified it a bit to make it better looking :)

magnus-madsen (approved):

magnus-madsen:
Merged. Thanks!

PR #6987, opened 2024-01-08
feat: add `SafetyError.IllegalEntryPointSignature`

magnus-madsen:
Fixes #6980

EDIT: I nuked some tests that were beyond repair.

PR #7002, opened 2024-01-10
test: remove invalid effect tests

JonathanStarup:
Entrypoints are not allowed to have unhandled effects


magnus-madsen:
I have a PR that deletes these. Just approve it instead: https://github.com/flix/flix/pull/6987

PR #7005, opened 2024-01-10
refactor: make Validation.Success constructor private

jaschdoc:
Depends on https://github.com/flix/flix/pull/7004
Supersedes https://github.com/flix/flix/pull/6912

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/Bootstrap.scala
+++ b/main/src/ca/uwaterloo/flix/api/Bootstrap.scala
@@ -23,7 +23,8 @@ import ca.uwaterloo.flix.tools.pkg.{FlixPackageManager, JarPackageManager, Manif
 import ca.uwaterloo.flix.tools.{Benchmarker, Tester}
 import ca.uwaterloo.flix.util.Result.{Err, Ok}
 import ca.uwaterloo.flix.util.Validation.flatMapN
-import ca.uwaterloo.flix.util.Validation
+import ca.uwaterloo.flix.util.collection.Chain
+import ca.uwaterloo.flix.util.{Result, Validation}
 
 import java.io.{PrintStream, PrintWriter}
 import java.nio.file._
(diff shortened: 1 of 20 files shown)


main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -668,10 +668,10 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
     // Build artifacts
     println("Building project...")
     val buildResult = buildPkg()
-    buildResult match {
-      case Validation.Success(_) => // Continue
-      case Validation.SoftFailure(_, e) => return Validation.SoftFailure((), e)
-      case Validation.HardFailure(e) => return Validation.HardFailure(e)
+    buildResult.toSoftResult match {
```
magnus-madsen, line 671: I think the original code was in error. We want a hard failure here too.

main/src/ca/uwaterloo/flix/api/lsp/LanguageServer.scala
```diff
@@ -337,16 +337,16 @@ class LanguageServer(port: Int, o: Options) extends WebSocketServer(new InetSock
     val t = System.nanoTime()
     try {
       // Run the compiler up to the type checking phase.
-      flix.check() match {
-        case Success(root) =>
+      flix.check().toSoftResult match {
```
magnus-madsen, line 340: Exactly. Here we want the soft result.

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
magnus-madsen: Sorry. What I mean is, we can just use toHardFailure and get the same result.

magnus-madsen: Its late. What I mean is, lets use toHardResult and just return HardFailure always. Just like all the other cases.

I will leave it for a subsequent PR.

magnus-madsen (approved):

Issue #6972, opened 2024-01-07
Use GitHub token if available in project root `.GITHUB_TOKEN` file


magnus-madsen:

magnus-madsen:
I would like the order to be that: 
- Use --github-token argument if available.
- Next, use .GITHUB_TOKEN file if available.
- Next, use GITHUB_TOKEN environment variable if available.

magnus-madsen:
@sockmaster27 After release this should be priority.

PR #7020, opened 2024-01-11
feat: allow storing the GitHub token in `.GITHUB_TOKEN` file

sockmaster27:
Fixes: #6972

### DIFF
--- a/main/src/ca/uwaterloo/flix/Main.scala
+++ b/main/src/ca/uwaterloo/flix/Main.scala
@@ -74,19 +74,8 @@ object Main {
     // get GitHub token
     val githubToken =
       cmdOpts.githubToken
-        .orElse(FileOps.readString(cwd.resolve("./.GITHUB_TOKEN")))
+        .orElse(FileOps.readLine(cwd.resolve("./.GITHUB_TOKEN")))
         .orElse(sys.env.get("GITHUB_TOKEN"))
-        .map { t =>
-          val trimmed = t.trim
-
-          // if the token contains newlines, any requests using it in the header will be malformed
-          if (trimmed.contains("\n")) {
-            Console.err.println("GitHub token must not contain any newline characters. Will now exit.")
-            System.exit(1)
-          }
-
-          trimmed
-        }
 
     // construct flix options.
     var options = Options(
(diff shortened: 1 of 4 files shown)


main/src/ca/uwaterloo/flix/Main.scala
```diff
@@ -69,10 +71,22 @@ object Main {
...
     }
 
-    // Get GitHub token
+    // get GitHub token
     val githubToken =
       cmdOpts.githubToken
+        .orElse(FileOps.readString(cwd.resolve("./.GITHUB_TOKEN")))
         .orElse(sys.env.get("GITHUB_TOKEN"))
+        .map { t =>
+          val trimmed = t.trim
```
magnus-madsen, line 80: This code here breaks the flow and seem unnatural. May I instead suggest to have FileOps.readLine which reads only the first line?

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -91,6 +91,7 @@ object Bootstrap {
          |artifact/
          |build/
          |lib/
+         |.GITHUB_TOKEN
```
magnus-madsen, line 94: move up before the folders.

main/src/ca/uwaterloo/flix/util/FileOps.scala
```diff
@@ -23,6 +23,22 @@ import java.nio.file.{Files, LinkOption, Path}
...
 
+  /**
+    * Reads the contents of the file at the given path `p` if it is possible.
+    *
+    * Returns None if:
+    * - The path doesn't exist.
+    * - The path isn't a file.
+    * - The current process doesn't have the necessary permissions.
+    */
+  def readString(p: Path): Option[String] = {
```
magnus-madsen, line 34: Maybe one line? Maybe also check its: ordinary file?

magnus-madsen: "regular file"

main/src/ca/uwaterloo/flix/Main.scala
sockmaster27: I agree, but secretly only reading the first line sounds a bit unpredictable. What about moving this part to a helper function called something like `sanitizeGithubToken`?

magnus-madsen: I think simple-stupid is best. What's wrong with reading just the first line?

magnus-madsen: I really dont think we want to start doing input validation on the token.

magnus-madsen (approved):

PR #7024, opened 2024-01-11
feat: use terminal colors in the build system

### LINKED ISSUE #6974: Use terminal colors for all interesting paths/options/properties in the build system

sockmaster27:
Fixes #6974


sockmaster27:
Release:

image

Running in folder mode:

image

Name of output file being occupied by something else (in this case some directories):

image

image

sockmaster27:
Slightly related - It seems that the terminal is quickly cluttered by having all the `Building project...` type steps sticking around forever. Perhaps they should disappear again when the step is completed without errors?

magnus-madsen:
I think `flix.toml` should also be highlighted.

magnus-madsen:
> Slightly related - It seems that the terminal is quickly cluttered by having all the `Building project...` type steps sticking around forever. Perhaps they should disappear again when the step is completed without errors?

In the past we tried clearing things, but it _always_ ended up being confusing in edge cases.

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -501,7 +500,9 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
   /**
     * Builds a jar package for the project.
     */
-  def buildJar(): Validation[Unit, BootstrapError] = {
+  def buildJar(flix: Flix): Validation[Unit, BootstrapError] = {
```
magnus-madsen, line 503: I think there was a point in these not requiring an Flix object.

In other words, they are independent of the specific flix compiler instance.

I think we should just pass the formatter implicitly.

magnus-madsen:
If there are flix/maven dependencies which are downloaded, should we color them too?

magnus-madsen:
Could you show a screenshot of a package that depends on the museum package? 

https://doc.flix.dev/packages.html#understanding-dependency-resolution

sockmaster27:
billede

magnus-madsen:
Lets keep the extra chars.

magnus-madsen:
Conflicted, sorry

sockmaster27:
RFM

magnus-madsen (approved):

magnus-madsen:
Thanks!

PR #7028, opened 2024-01-12
docs: effect examples

JonathanStarup:


magnus-madsen:
Note: There is an issue where subfolders in the example folder are not compiled nor tested.

magnus-madsen:
@JonathanStarup Do you want to wrap these up and we can merge them? Then I can ask @jaschdoc to port more examples.

magnus-madsen:
@JonathanStarup The scheduler example gives me ClassCastException.

magnus-madsen:
> removed the failing example

But why? It is a nice example. Can it not be made to work?

JonathanStarup:
I just wanted to debug it while these could be merged. also i cant reference the source

magnus-madsen:
> I just wanted to debug it while these could be merged. also i cant reference the source

OK, so you are looking into it?

magnus-madsen:
Did you want to extend `ExampleSuite.scala` so these are actually compiled?

magnus-madsen (approved):

PR #7029, opened 2024-01-13
feat: only generate HTML documentation for modules declared in flix.toml

### LINKED ISSUE #6377: Integrate `flix.toml` with `HtmlDocumenter`
We need to come up with a way for a package to describe what modules it provides.

sockmaster27:
Fixes #6377


sockmaster27:
This is kind of a proof-of-concept. It doesn't work with companion modules yet.

main/src/ca/uwaterloo/flix/language/phase/HtmlDocumentor.scala
```diff
@@ -71,9 +71,9 @@ object HtmlDocumentor {
     */
   val LibraryGitHub: String = "https://github.com/flix/flix/blob/master/main/src/library/"
 
-  def run(root: TypedAst.Root)(implicit flix: Flix): Unit = {
+  def run(root: TypedAst.Root, includedModules: Option[List[String]])(implicit flix: Flix): Unit = {
```
magnus-madsen, line 74: Instead of the vague type, `Option[List[String]`, I think what we want is `Set[Symbol.ModuleSym]`.

This will ensure that we are actually taking about modules of the right format.

I think it should also help simplify some of the later code.

magnus-madsen:
We might want to split in two: one PR for flix.toml extension and one that uses it in HtmlDocumentor.

sockmaster27:
I've created #7081 to introduce the the changes to the manifest which should be merged before this one.

sockmaster27:
But I think we need to specify some things. This is my interpretation:
1. If `MyMod` is included, all its items and submodules should also be included
1. If `MyMod.MySubMod` is included, then `MyMod` will be included, but with all other items than `MySubMod` removed
1. If e.g. `MyEnum` is included, then both the enum and its companion module is included
1. If `MyEnum.MySubMod` is included, then the `MyEnum` enum should be removed and only the companion module should be included as per 2.

sockmaster27:
> But I think we need to specify some things. This is my interpretation:
> 
>     1. If `MyMod` is included, all its items and submodules should also be included
> 
>     2. If `MyMod.MySubMod` is included, then `MyMod` will be included, but with all other items than `MySubMod` removed
> 
>     3. If e.g. `MyEnum` is included, then both the enum and its companion module is included
> 
>     4. If `MyEnum.MySubMod` is included, then the `MyEnum` enum should be removed and only the companion module should be included as per 2.

@magnus-madsen
Do you agree?

magnus-madsen:
I agree with 1. and 2. I don't understand the special cases for enums in 3 and 4. 

I think if `X.Y.Z` is included then we include `X.Y.Z` and everything under it. 

In addition we need the spine of the tree, so `X.Y` with only `Z` mentioned. Similarly for `X`.

In any case, try to keep it straightforward.

sockmaster27:
> I agree with 1. and 2. I don't understand the special cases for enums in 3 and 4. 

Not specifically enums, just anything that has a companion module.

sockmaster27:
I believe it now functions properly but with these rules:

1. If `MyMod` is included, all its items and submodules should also be included.

2. If `MyMod.MySubMod` is included, then `MyMod` will be included, but with all other items than `MySubMod` removed.

3. If an item with a companion module, e.g. `MyMod.MyEnum` is included, then only the module is included. If you want to include the item too, you will have to include the entirety of `MyMod`.

4. If `MyEnum.MySubMod` is included, then the `MyEnum` item should be removed and only the companion module should be included as per 2.

sockmaster27:
A related problem which I would like to fix in a different PR is that with the current mechanism, writing something like
```
mod Four {
    pub def fourMod(): Unit \ IO =
        println("four")
}

trait Four[a] {
    pub def fourTrait(_four: a): Unit \ IO =
        println("trait four")
}
```
where the trait is _not_ `pub` results in `fourMod()` being excluded from the documentation. We would need to defer pairing items with their companion modules until after filtering has been done.

magnus-madsen:
!!!jar

magnus-madsen (approved):

Issue #7042, opened 2024-01-15
Add a special typing rule for `do` for `Void`


magnus-madsen:
Add a special type rule for `do` where if the operation returns `Void` then we replace that type by a fresh type variable.

magnus-madsen:
Fixed in master.

PR #7047, opened 2024-01-16
refactor: add `Void` type constructor

### LINKED ISSUE #7041: Refactor `Void` to be a built-in type instead of being declared in `Prelude`

magnus-madsen:
Fixes #7041 and #7042


main/src/ca/uwaterloo/flix/language/phase/TypeInference.scala
```diff
@@ -986,6 +986,16 @@ object TypeInference {
           } yield (tconstrs, tpe, eff)
         }
 
+        // We special case the result type of the operation.
+        val operationType = operation.spec.tpe.typeConstructor match {
```
mlutze, line 990: What happens if we make the `Void` type as `fresh a` everywhere, instead of just here?

mlutze: Not sure exactly how we would do it, but it seems possible...

mlutze (approved):

main/src/ca/uwaterloo/flix/language/phase/TypeInference.scala
magnus-madsen: Unsound

magnus-madsen: What I have added is essentially sugar for `match do Foo.foo() {}`, i.e. case elimination on the empty type, but only for an effect operation.

PR #7048, opened 2024-01-16
feat: make ParOps deterministic for single-threaded compilation

mlutze:


magnus-madsen:
Does it work?

mlutze:
Yep, all tests pass with threads=1

magnus-madsen (approved):

Issue #7018, opened 2024-01-11
Add `Test.Handler.ZeroShot.Recursive.flix`


magnus-madsen:
We would like to add the following family of tests. 

They do not need to be duplicated for types. We can stick with `Int32`.

```
    def testRecursiveTryWith01(): Bool =
        def r(x) = try {
            if (x == 0) 0 else { r(x - 1); do Throw.throw(); unreachable!() }
        } with Throw {
            def throw(_k) = 42
        };
        Assert.eq(42, r(10)) // Add tests for X = 0, 1, 2, 3, and 10
```

PR #7055, opened 2024-01-16
test: add `Test.Handler.ZeroShot.Recursive.flix`

jaschdoc:
Closes https://github.com/flix/flix/issues/7018

### DIFF
--- a/main/test/flix/Test.Handler.ZeroShot.Recursive.flix
+++ b/main/test/flix/Test.Handler.ZeroShot.Recursive.flix
@@ -1,0 +1,48 @@
+mod Test.Handler.ZeroShot.Recursive {
+
+    @Test
+    def testRecursiveTryWith01(): Bool =
+        def r(x) = try {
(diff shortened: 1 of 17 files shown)


main/test/flix/Test.Handler.ZeroShot.Recursive.flix
```diff
@@ -0,0 +1,52 @@
+mod Test.Handler.ZeroShot.Recursive {
+
+    eff Throw {
+        pub def throw(): Unit
```
magnus-madsen, line 4: Once https://github.com/flix/flix/pull/7047 is merged then `Unit` should become `Void`.

magnus-madsen:
The wrong place was changed to Void.

magnus-madsen:
But actually. Lets go back to Unit in both cases. We can have separate tests for Void later.

magnus-madsen (approved):

Issue #7019, opened 2024-01-11
Add `Test.Handler.Ask.flix`


magnus-madsen:
Following this pattern:

```
eff Ask {
    pub def ask(x: Int32): Unit
}

def generator(x: Int32): Unit \ Ask =
    do Ask.ask(x); generator(x + 1)

def sample(limit: Int32): List[Int32] =
    try {
        generator(0); Nil
    } with Ask {
        def ask(x, k) = if (x == limit) Nil else x :: k()
    }

@Test
def testSample01(): Bool =
    Assert.eq(Nil, sample(0))
```

with several testSamples

magnus-madsen:
@jaschdoc 

I have reopen this because what I would like in the next step is the above pattern repeated for each type (as outlined in the other ticket).

For example, I think what we have is `Test.Handler.Gen.Int32.flix`

jaschdoc:
Roger

magnus-madsen:
See https://github.com/flix/flix/issues/7059 for an update.

jaschdoc:
So can we close this now that you have made issues for the remaining tasks?

jaschdoc:
Do you also want tests for each primitive type and `AtomicInteger` for this?

magnus-madsen:
> Do you also want tests for each primitive type and `AtomicInteger` for this?

Yes please. Thank you sir.

PR #7056, opened 2024-01-16
test: add `Test.Handler.Gen.flix`

jaschdoc:
Closes https://github.com/flix/flix/issues/7019

### DIFF
--- a/main/test/flix/Test.Handler.Gen.flix
+++ b/main/test/flix/Test.Handler.Gen.flix
@@ -1,7 +1,7 @@
 mod Test.Handler.Gen {
 
     eff Gen {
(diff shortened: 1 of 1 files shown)


magnus-madsen (approved):

main/test/flix/Test.Handler.Gen.flix
```diff
@@ -0,0 +1,53 @@
+mod Test.Handler.Gen {
+
+    eff Gen {
+        pub def gen(x: Int32): Void
```
magnus-madsen, line 4: This should have Unit. Most def :)

Because its resumable (see the call to k)

magnus-madsen (approved):

Issue #7016, opened 2024-01-11
Introduce `Test.Handler.ZeroShot.X.flix` for every primitive type.


magnus-madsen:
Given that #7015 has been done, the goal is now to _duplicate_ the whole into multiple files for each of the primitive types and one Java type:

- Bool
- Char
- Int8-Int64
- Float32-Float64
- String
- A Java type, e.g. `java.util.concurrent.atomic.AtomicInteger`.

I would suggest that we copy into `Test.Handler.ZeroShot.Int32.flix` and so forth.

Here is an example of how the first test would now look like:

```
    @Test
    def testLinear01(): Bool =
        def f1() = do Throw.throw();
        let result = try {
            f1()
        } with Throw {
            def throw(_k) = 42
        };
        Assert.eq(42, result)
```

PR #7057, opened 2024-01-16
test: add `Test.Handler.ZeroShot.X.flix` for every primitive type

jaschdoc:
Closes https://github.com/flix/flix/issues/7016

### DIFF
--- a/main/test/flix/Test.Handler.ZeroShot.Int32.flix
+++ b/main/test/flix/Test.Handler.ZeroShot.Int32.flix
@@ -1,0 +1,128 @@
+mod Test.Handler.ZeroShot.Int32 {
+
+    eff Throw {
+        pub def throw(): Int32
+    }
+
+    @Test
+    def testLinear01(): Bool =
+        def f1() = do Throw.throw();
+        let result = try {
+            f1()
+        } with Throw {
+            def throw(_k) = 42
+        };
+        Assert.eq(42, result)
+
+    @Test
+    def testLinear02(): Bool =
+        def f1() = do Throw.throw();
+        def f2() = { f1(); unreachable!() };
+        let result = try {
+            f2()
+        } with Throw {
+            def throw(_k) = 42
+        };
+        Assert.eq(42, result)
+
+    @Test
+    def testLinear03(): Bool =
+        def f1() = do Throw.throw();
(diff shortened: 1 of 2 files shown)


main/test/flix/Test.Handler.ZeroShot.AtomicInteger.flix
```diff
@@ -0,0 +1,169 @@
+mod Test.Handler.ZeroShot.AtomicInteger {
+
+    enum AtomicInt(##java.util.concurrent.atomic.AtomicInteger)
```
magnus-madsen, line 3: haha! Smart.

But, we actually need the concrete type to test that it works with java types.

You can use a type alias though, `type alias AtomicInt = ##...`.

magnus-madsen: But, on a second thought, this made me realize that we will want a `Test.Handler.ZeroShot.Option.flix` which uses the Flix `Option` type (let us say with an Int32 inside).

magnus-madsen: Lets call the file Test.Handler.ZeroShot.Enum.flix, thanks

jaschdoc: What about tuples and records?

magnus-madsen: No that's OK. They are handled like enums more or less.

jaschdoc:
RFR

main/test/flix/Test.Handler.ZeroShot.AtomicInteger.flix
```diff
@@ -0,0 +1,163 @@
+mod Test.Handler.ZeroShot.AtomicInteger {
+
+    type alias AtomicInt = ##java.util.concurrent.atomic.AtomicInteger
```
magnus-madsen, line 3: I have 3 OCD changes, I would request:

- Rename the type alias from AtomicInt to AtomicInteger.
- Add a helper function `newAtomicInteger` right after the type alias and use it below.
- Add a helper function `toInt` right after the above function and use it below.

magnus-madsen (approved):

PR #7062, opened 2024-01-17
refactor: fix MonoDefs

JonathanStarup:


magnus-madsen:
What's the plan here? This seems to make some design changes.

JonathanStarup:
Just to do the same thing with clearer design

magnus-madsen:
@mlutze @JonathanStarup Let's have a chat about this at somepoint. Then if the this is the way to go, the PR needs to be cleaned.

magnus-madsen:
Can I close?

PR #7070, opened 2024-01-18
refactor: convert tuple to BackendObjectType

JonathanStarup:


main/src/ca/uwaterloo/flix/language/phase/jvm/JvmBackend.scala
```diff
@@ -65,6 +65,7 @@ object JvmBackend {
       val erasedLazyTypes = JvmOps.getErasedLazyTypesOf(types)
       val erasedExtendTypes = JvmOps.getErasedRecordExtendsOf(types)
       val erasedFunctionTypes = JvmOps.getErasedArrowsOf(types)
+      val erasedTuplesTypes = JvmOps.getErasedTupleTypesOf(types)
```
magnus-madsen, line 68: All of these might end up being expensive. Perhaps we need to do just one iteration at some point.

magnus-madsen (approved):

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmBackend.scala
JonathanStarup: These are filtered and then genClass is called on all of them. Soon (or now?) it should be `types.map(genClass)` where genClass figures out what to do depending on the type

Issue #7060, opened 2024-01-17
Add `Test.Handler.Gen.X.flix`


magnus-madsen:
Add tests, for each type, based on the following skeleton:

```scala
eff Gen {
    pub def gen(): Int32
}

def generator(): Unit \ Gen =
    do Gen.gen(); generator()

def sample(limit: Int32): List[Int32] =
    region rc {
        let counter = ref 0 @ rc;
        try {
            generator(); Nil
        } with Gen {
            def gen(k) = 
                let i = getAndInc(counter);
                if (i == limit) Nil else i :: k(i)
        }
    }

def getAndInc(r: Ref[Int32, r]): Int32 \ r = 
    let i = deref r;
    r := (i + 1);
    i
```

This is pretty much the dual of `Test.Handler.Ask`.

jaschdoc:
I assume this is only for numeric types, i.e., `float`, `int`, `AtomicInteger`?

magnus-madsen:
> I assume this is only for numeric types, i.e., `float`, `int`, `AtomicInteger`?

All types. But for non addable types we have do something else.

PR #7075, opened 2024-01-19
test: add `Test.Handler.Gen.X.flix`

jaschdoc:
Closes https://github.com/flix/flix/issues/7060

### DIFF
--- a/main/test/flix/Test.Handler.Gen.AtomicInteger.flix
+++ b/main/test/flix/Test.Handler.Gen.AtomicInteger.flix
@@ -1,0 +1,76 @@
+mod Test.Handler.Gen.AtomicInteger {
+
+    type alias AtomicInteger = ##java.util.concurrent.atomic.AtomicInteger
+
+    def newAtomicInteger(n: Int32): AtomicInteger =
+        import new java.util.concurrent.atomic.AtomicInteger(Int32): AtomicInteger \ {} as mk;
+        mk(n)
+
+    def toInt(n: AtomicInteger): Int32 =
+        import java.util.concurrent.atomic.AtomicInteger.intValue(): Int32 \ {} as intValue;
+        intValue(n)
+
+    eff Gen {
+        pub def gen(): AtomicInteger
+    }
+
+    def generator(): Unit \ Gen =
+        do Gen.gen(); generator()
+
+    def sample(limit: Int32): List[Int32] =
+        region rc {
+            let counter = ref newAtomicInteger(0) @ rc;
+            try {
+                generator(); Nil
+            } with Gen {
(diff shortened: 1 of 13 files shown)


jaschdoc:
Should we drop `AtomicInteger` or is it okay? I think it's awkward with all the conversion going on between `AtomicInteger` and `Int32`.

jaschdoc:
@magnus-madsen I think I have everything covered now.

- `Unit`
- `Bool`
- `Char`
- `String`
- `Int8-64`
- `Float32-64`
- `Option`
- `AtomicInteger`

main/test/flix/Test.Handler.Gen.AtomicInteger.flix
```diff
@@ -0,0 +1,76 @@
...
+        intValue(n)
+
+    eff Gen {
+        pub def gen(): AtomicInteger
+    }
+
+    def generator(): Unit \ Gen =
+        do Gen.gen(); generator()
+
+    def sample(limit: Int32): List[Int32] =
```
magnus-madsen, line 20: I would be better for the test case if the return type was List[AtomicInteger].

main/test/flix/Test.Handler.Gen.Int8.flix
```diff
@@ -0,0 +1,66 @@
...
+                    if (i == limit) Nil else i :: k(i)
+            }
+        }
+
+    def getAndInc(r: Ref[Int8, r]): Int8 \ r =
+        let i = deref r;
+        r := (i + 1i8);
+        i
+
+    def range(a: Int8, b: Int8): List[Int8] =
```
magnus-madsen, line 27: I guess the problem is that List.range is only for Int32s.

magnus-madsen: So you need your own.

main/test/flix/Test.Handler.Gen.Char.flix
```diff
@@ -0,0 +1,96 @@
...
+            try {
+                generator(); Nil
+            } with Gen {
+                def gen(k) =
+                    let i = getAndInc(counter);
+                    if (i == limit) Nil else i :: k(i)
+            }
+        }
+
+    def next(c: Char): Char = match c {
```
magnus-madsen, line 22: Seems reasonable.

magnus-madsen (approved):

main/test/flix/Test.Handler.Gen.AtomicInteger.flix
magnus-madsen: Lets fix this in a subsequent PR.

PR #7081, opened 2024-01-21
feat: introduce `modules` property to flix.toml

sockmaster27:
Related to #6377


main/src/ca/uwaterloo/flix/tools/pkg/IncludedModules.scala
```diff
@@ -0,0 +1,36 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ca.uwaterloo.flix.tools.pkg
+
+import ca.uwaterloo.flix.language.ast.Symbol
+
+sealed trait IncludedModules {
+  def contains(sym: Symbol.ModuleSym): Boolean
```
magnus-madsen, line 21: The implementation should be here, by pattern matching on this.

We dont use the OO style with override :)

main/src/ca/uwaterloo/flix/tools/pkg/IncludedModules.scala
```diff
@@ -0,0 +1,36 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ca.uwaterloo.flix.tools.pkg
+
+import ca.uwaterloo.flix.language.ast.Symbol
+
+sealed trait IncludedModules {
```
magnus-madsen, line 20: @mlutze Can we come up with a better name? Perhaps just PackageModules?

main/src/ca/uwaterloo/flix/tools/pkg/ManifestParser.scala
```diff
@@ -469,6 +487,31 @@ object ManifestParser {
...
+  /**
+    * Creates the `IncludedModules` object from `optArray`.
+    */
+  private def toIncludedModules(optArray: Option[TomlArray], p: Path): Result[IncludedModules, ManifestError] = {
+    optArray match {
+      case None =>
+        Ok(IncludedModules.All)
+      case Some(array) =>
+        var moduleSet = Set.empty[Symbol.ModuleSym]
+        for (i <- 0 until array.size()) {
```
magnus-madsen, line 499: Is it not possible to iterate directly through the array?

magnus-madsen:
Ill let @mlutze have a look too.

main/src/ca/uwaterloo/flix/tools/pkg/IncludedModules.scala
mlutze: ModuleSet?

main/src/ca/uwaterloo/flix/tools/pkg/ManifestParser.scala
```diff
@@ -469,6 +487,31 @@ object ManifestParser {
...
+          try {
+            val string = array.getString(i)
```
mlutze, line 501: What happens if the array contains a non-string? Maybe we should do this validation in the parsing part?

main/src/ca/uwaterloo/flix/tools/pkg/IncludedModules.scala
magnus-madsen: Let's call it `PackageModules`. That makes the most sense to me.

main/src/ca/uwaterloo/flix/tools/pkg/Manifest.scala
```diff
@@ -21,6 +21,7 @@ case class Manifest(name: String,
                     description: String,
                     version: SemVer,
                     repository: Option[GitHub.Project],
+                    modules: IncludedModules,
```
magnus-madsen, line 24: this field should still be called just `modules` that is fine.

magnus-madsen (approved):

PR #7083, opened 2024-01-21
refactor: output bytecode to `build/class` dir

### LINKED ISSUE #6712: Code/Meta seperation in `build/`
Right now auxiliary information is put along side actual generated code, e.g. `build/asts/...` is right next to `build/Clo$main$146782`. We should have `build/code/` and `build/meta/` or something like that.

It also looks prettier when looking for meta stuff instead of immediately seeing 100s of compiled java classes.

sockmaster27:
Fixes #6712


magnus-madsen:
Can you confirm that build-jar still works as expected?

sockmaster27:
Yes, running `build-jar` still succeeds

magnus-madsen:
> Yes, running `build-jar` still succeeds

But is the jar non-empty? Does the jar actually work?

magnus-madsen:
I suspect its broken-- take a look at `Bootstrap.scala`

magnus-madsen:
Did you check that the built JAR works?

magnus-madsen:
> > Did you check that the built JAR works?
> 
> The `Main.class` is now in the root of the JAR

Does it run?

sockmaster27:
> > > Did you check that the built JAR works?
> > 
> > 
> > The `Main.class` is now in the root of the JAR
> 
> Does it run?

Yes

magnus-madsen (approved):

PR #7097, opened 2024-01-22
fix: tests always pass

### LINKED ISSUE #7086: Return value of Bool function is wrong
This program:

```
@Test 
def fail(): Bool = false
```

Succeeds. If we look in Tester near line 197, the problem is that the result seems to be `null`.

JonathanStarup:
Fixes #7086

This is not a pretty fix. Maybe the proper solution is to actually add these functions pre-eraser and just generate in the backend. Something like `Root.Api` which is the list of functions that GenNamespaceClasses currently just makes up. These functions would call their corresponding def and unwrap the Value

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Eraser.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Eraser.scala
@@ -5,7 +5,7 @@ import ca.uwaterloo.flix.language.ast.Ast.CallType
 import ca.uwaterloo.flix.language.ast.ReducedAst.Expr._
 import ca.uwaterloo.flix.language.ast.ReducedAst._
 import ca.uwaterloo.flix.language.ast.{AtomicOp, MonoType, Purity, SourceLocation, Symbol}
-import ca.uwaterloo.flix.util.ParOps
+import ca.uwaterloo.flix.util.{InternalCompilerException, ParOps}
 import ca.uwaterloo.flix.util.collection.MapOps
 
 /**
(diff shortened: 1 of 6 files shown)


magnus-madsen:
This does not look nice. Cluttering the ASTs with extra information. 

What's the problem? You seem to suggest some functions are missing.

JonathanStarup:
`def fail(): Bool = false`

the type of fail is erased from `Unit -> Bool` to `Unit -> Object`. This is correct internally because when anyone calls `fail` the eraser inserts and explicit unwrapping of the underlying boolean value. When `Ns.m_fail` is generated however, we just know the erased function type (`Unit -> Object`) so when we try to unwrap the Value we unwrap it to an object, thereby getting `null`.

JonathanStarup:
Maybe a `ErasedFrom(MonoType)` such that the type itself knows? I don't know how big this addition would be

JonathanStarup:
I think solution 2 (materialize the functions) makes sense long term, then that phase would decide and build the functions that are accessible from the outside

magnus-madsen:
How 

> I think solution 2 (materialize the functions) makes sense long term, then that phase would decide and build the functions that are accessible from the outside

Right. This is where we need a facade pattern or `@Export` functionality.

JonathanStarup:
Yet another idea - main works despite superficially being under the same bug, because if main has a non-unit return type, then another main function is generated that calls the original one. We could do something like this for tests. insert asserts for bool return type tests

JonathanStarup:
These solutions are pretty big though, maybe this PR is a hotfix and we think of a proper solution asap afterwards?

main/src/ca/uwaterloo/flix/language/phase/Eraser.scala
```diff
@@ -42,10 +42,11 @@ object Eraser {
   }
 
   private def visitDef(defn: Def): Def = defn match {
-    case Def(ann, mod, sym, cparams, fparams, lparams, pcPoints, exp, tpe, purity, loc) =>
+    case Def(ann, mod, sym, cparams, fparams, lparams, pcPoints, exp, tpe, originalTpe, purity, loc) =>
+      if (tpe != originalTpe.tpe) throw InternalCompilerException(s"Types in def do not align. tpe = '$tpe', originalTpe = '$originalTpe'", loc)
```
magnus-madsen, line 46: Remove this. We don't have explicit assertions.

main/src/ca/uwaterloo/flix/language/phase/Eraser.scala
```diff
@@ -5,7 +5,7 @@ import ca.uwaterloo.flix.language.ast.Ast.CallType
 import ca.uwaterloo.flix.language.ast.ReducedAst.Expr._
 import ca.uwaterloo.flix.language.ast.ReducedAst._
 import ca.uwaterloo.flix.language.ast.{AtomicOp, MonoType, Purity, SourceLocation, Symbol}
-import ca.uwaterloo.flix.util.ParOps
+import ca.uwaterloo.flix.util.{InternalCompilerException, ParOps}
```
magnus-madsen, line 8: Remove :)

main/test/ca/uwaterloo/flix/library/TestDebug.flix
```diff
@@ -172,23 +172,23 @@ mod TestDebug {
     def testFunctionStringify01(): Bool = {
         def f(x: Int32): Int32 = x;
         let s = stringify(f);
-        s == "Int32 -> Int32"
+        s == "Int32 -> Obj"
```
magnus-madsen, line 175: I really wonder if this debugging stuff is worth it. But I guess that is a separate discussion.

magnus-madsen (approved):

PR #7102, opened 2024-01-22
feat: remove `equals` and `hashcode` from backend objects

### LINKED ISSUE #7101: Remove Equals and HashCode from FlixErrors
Does't seem to me like they are used

JonathanStarup:
Fixes #7101

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
@@ -710,8 +710,6 @@ object BackendObjType {
       cm.mkField(EndLineField)
       cm.mkField(EndColField)
 
-      cm.mkMethod(EqualsMethod)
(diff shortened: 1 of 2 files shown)


magnus-madsen:
I remembered the reason these are here. I can expand, but the simple reason was that if you want to write automated testing tools you may need to know when two errors are the same.

JonathanStarup:
So you want to leave them in / keep supporting it for UnhandledEffectError? On the other hand I don't think its too hard to add back in

magnus-madsen:
> So you want to leave them in / keep supporting it for UnhandledEffectError? On the other hand I don't think its too hard to add back in

Give me some time to think on it.

magnus-madsen (approved):

magnus-madsen:
We can still manually inspect the fields to check for equality, so I guess its OK.

JonathanStarup:
The fields of matcherror is also deleted in this PR, so you'd have to compare the strings of error message i think

PR #7103, opened 2024-01-22
docs: scheduler effect example

JonathanStarup:


magnus-madsen:
You mentioned some issue with typing?

magnus-madsen:
@JonathanStarup Does not yet typecheck. What should happen?

PR #7104, opened 2024-01-22
feat: add unhandled effect error

### LINKED ISSUE #7090: Introduce `UnhandledEffectError`
This program lacks a handler:

```
eff Ask {
    pub def ask(): String
}

def main(): Unit \ IO = 
    region rc {
        spawn do Ask.ask() @ rc
    }
```

and crashes with 

```
java.lang.ClassCastException: class dev.flix.runtime.Suspension$ 
cannot be cast to class dev.flix.runtime.Value$ 
```

I would like to have introduced an `UnhandledEffectError` (extends FlixError) This would be similar to our use of e.g. `MatchError`

JonathanStarup:
Fixes #7090


JonathanStarup:
Verified to work manually, although most location are `Unknown`

magnus-madsen (approved):

PR #7105, opened 2024-01-23
refactor: deleted unused code from LoweredAst

### LINKED ISSUE #7077: Delete constraint stuff from LoweredAst

JonathanStarup:
Fixes #7077

It's surprising how much code is unused.

Questions before merging: 
- some things like `loc` could maybe be kept for each class just for consistency?
- Instances could keep their names and type for sanity even though its not used?
- Should the "transitive" unused values also be removed (deleted in Lowered, Simplified, Lifted, Reduced)

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/LoweredAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/LoweredAst.scala
@@ -16,48 +16,42 @@
 
 package ca.uwaterloo.flix.language.ast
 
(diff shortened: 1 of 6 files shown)


magnus-madsen:
LoweredAst is before MonoDefs and MonoTypes, but has to be moved after right?

Because otherwise various type class instances are not resolved correctly.

JonathanStarup:
closed due to future move of lowering

PR #7106, opened 2024-01-23
refactor: remove effect translation example

JonathanStarup:


magnus-madsen:
You want to remove it now or when all bugs have been worked out?

JonathanStarup:
we can just look at them in the generated code

magnus-madsen (approved):

Issue #7099, opened 2024-01-22
Add `Test.Handler.NewObject.flix`


magnus-madsen:
The issue here is very similar to https://github.com/flix/flix/issues/7098

But here we must test that we can handle effects inside new object expressions.

For example:

```flix
import java.lang.Runnable

eff Ask {
    pub def ask(): String
}

def newRunnable(): Runnable \ IO = new Runnable {
    def run(_this: Runnable): Unit \ IO = 
        do Ask.ask(); () // TODO: Insert handler here.
}

def main(): Unit \ IO = 
    import java.lang.Runnable.run(): Unit \ IO;
    let r = newRunnable();
    run(r)
```

So if https://github.com/flix/flix/issues/7098 is solved first then it should be possible to use similar ideas here. 

Note that we must call `run` through the Java interface. Otherwise it does not "count".

PR #7112, opened 2024-01-23
test: add `Test.Handler.OneShot.X.flix`

jaschdoc:
Closes https://github.com/flix/flix/issues/7063

PR #7113, opened 2024-01-23
test: add `Test.Handler.NewObject.X.flix`

jaschdoc:
Closes https://github.com/flix/flix/issues/7099


magnus-madsen:
I had something much smaller in mind. Like 100 lines because in this case we do not need to test different types. 

Simple suggestion: rename `Test.Handler.NewObject.Int32.flix` to just `Test.Handler.NewObject.flix` and nuke all the other files.

jaschdoc:
This also has the same bug as #7112

jaschdoc:
I copied the style that works which we also had in the other PR but the tests should use the other style :)

magnus-madsen:
For both spawn and new object, I had something much simpler in mind.

Really here we just need to check that you can handle an effect.

The code path for the rest is pretty much the same.

So I would like to keep this "PR-style", drop all tests except those for Int32, and merge it in. Ditto for spawn.

Later, if necessary, we can revisit.

magnus-madsen:
The typing rule will not be fixed now, anyway. So it makes sense to merge in something that works until that day arrives.

jaschdoc:
Roger

jaschdoc:
RFR

magnus-madsen (approved):

Issue #7089, opened 2024-01-21
Reword `Impure` to `IO` in specific places in the compiler


magnus-madsen:
We still seem to use the word `Impure` in some places where we mean `IO`.

Note: Its fine to use to continue to use impure (as in having any effect), but we should not use it when we explicitly mean `IO`.

This is probably best solved after the type inference refactor.

Issue #7096, opened 2024-01-22
Add `Eff.Logger` effect


magnus-madsen:
Blocked by #7699 

Expand the following to a real implementation:

```scala
pub eff Logger {
    // TODO: DOC:
    // Note: The msg is lazy. We could also use Lazy[String] ? 
    pub def log(s: Logger.Severity, msg: Unit -> String): Unit
}

mod Logger {

    pub enum Severity {
        case Fatal
        case Warn
        case Info
        case Debug
        case Trace
    }

    // TODO: Add convenience methods for each severity, e.g.: 
    pub def fatal(msg: a): Unit \ Logger with ToString[a] = do Logger.log(Severity.Fatal, () -> "${msg}")

    // TODO: Add convenience methods for each severity, e.g.: 
    pub def lazyFatal(msg: Unit -> a): Unit \ Logger with ToString[a] = do Logger.log(Severity.Fatal, () -> "${msg()}")

    // TODO: DOC
    pub def ioHandler(f: Unit -> a \ ef + Logger): a \ ef + IO = 
        Logger.ioHandlerWithSeverity(Severity.Trace, f)

    // TODO: DOC
    pub def toListHandler(f: Unit -> a \ ef + Logger): (a, List[String]) \ ef + IO = 
        // TODO: Here we would use a MutList to collect the formatted messages and then return them.
        // A question is whether instead of String, we should have some LogMessage data structure?
        checked_ecast(???)

    // TODO: pub def toCollectable() similiar to toListHandler, but would use the Collectable trait.

    // TODO: An IO handler which prints all messages at least `s` severity to stderr.
    // We could start with output like:
    // [2014-07-02 20:52:39] [DEBUG] - This is a debug message
    pub def ioHandlerWithSeverity(s: Severity, f: Unit -> a \ ef + Logger): a \ ef + IO = 
        checked_ecast(???)

}
```

We shall probably have to do a few iterations. The goal is not to have a "once since fits all", but to have a `Logger` that works 90% of the time. We can always add an `AdvancedLogger` later. 

The key ingredients of the above design are:

- Use of a `Severity` which can be filtered on.
- `Logger` only has a single operation, but the companion module offers lots of helpers. Programmers are intended to write `Logger.info` for example.
- Use of `ToString`.
- Use of `laziness` via `Unit -> a`. We may also consider whether `Lazy[a]` is better.
- Various standard handlers, e.g. to print to stderr or to collect into a list.

Things missing or deferred: 

[...]

PR #7114, opened 2024-01-24
feat: distinguish IO and EffUniv

### LINKED ISSUE #7091: The `IO` effect should behave like any other effect
We currently model `IO` as the top element of the effect lattice. But instead, we should simply treat `IO` as any other effect, except it cannot be handled. This refactoring will fix some issues with the effect system and allow code to mix user-defined effects with `IO` effects.

Currently this program type checks:

```
eff Ask {
    pub def ask(): String
}

def main(): Unit \ IO = 
    do Ask.ask();
    println("Hello World!")

```

but it should not because it has unhandled effects.

In any case, this refactoring should probably not happen until the new type system is in place.

mlutze:
fixes #7091


magnus-madsen:
Did you get to the dragon part?

mlutze:
> Did you get to the dragon part?

Yes :D

mlutze:
@magnus-madsen RFR
A few notes:
1. I disabled Closeable tests. The interface seems broken as IO is no longer top, and I think it should probably have an Assoc Effect instead.
2. There's still a bit of weirdness around erasure of effects. I don't think it's made worse by this PR though. We need a refactoring like in #7062 to clarify invariants

magnus-madsen:
Could you outline how you monomorph effects now? In particular, how many functions do you generate for uses of List.map with different effects?

mlutze:
> Could you outline how you monomorph effects now? In particular, how many functions do you generate for uses of List.map with different effects?

Probably far too many :)

I'll look into it

magnus-madsen:
Maybe do the experiment and list them?

magnus-madsen:
Can you also do a performance test? with XPerf

mlutze:
```scala
@test
def map08(): Bool \ IO = List.map(i -> unchecked_cast(i > 2 as _ \ IO), 1 :: 2 :: Nil) == false :: false :: Nil

@test
def map09(): Bool \ IO = List.map(i -> unchecked_cast(i > 2 as _ \ IO + IO), 1 :: 2 :: Nil) == false :: false :: Nil

pub eff E

@test
def map10(): Bool \ IO = unchecked_cast(List.map(i -> unchecked_cast(i > 2 as _ \ IO + E), 1 :: 2 :: Nil) == false :: false :: Nil as _ \ IO)

@test
def map11(): Bool \ IO = unchecked_cast(List.map(i -> unchecked_cast(i > 2 as _ \ E + IO), 1 :: 2 :: Nil) == false :: false :: Nil as _ \ IO)
```

yields

```
(List.map$156802,(Int32 -> Bool \ IO + E) -> (List[Int32] -> List[Bool] \ IO + E))
(List.map$156774,(Int32 -> Bool \ IO) -> (List[Int32] -> List[Bool] \ IO))
```

So the duplication is still modulo Boolean equivalence, which is good.

mlutze:
### Feature
```

Throughput (best): 18,539 lines/sec (with 8 threads.)

  min: 15,198, max: 18,539, avg: 16,807, median: 17,114

Finished 7 iterations on 66,001 lines of code in 27 seconds.

Process finished with exit code 0
```

### Master
```

Throughput (best): 19,951 lines/sec (with 8 threads.)

  min: 15,778, max: 19,951, avg: 17,773, median: 18,055

Finished 7 iterations on 66,002 lines of code in 26 seconds.

Process finished with exit code 0
```

mlutze:
> @mlutze Does this help Stephen's example?

link?

magnus-madsen:
Does this also fix https://github.com/flix/flix/issues/7089 ?

main/src/ca/uwaterloo/flix/language/ast/Type.scala
```diff
@@ -325,6 +326,11 @@ object Type {
     */
   val Pure: Type = Type.Cst(TypeConstructor.Pure, SourceLocation.Unknown)
 
+  /**
+    * Represents the IO effect.
+    */
+  val IO: Type = Type.Cst(TypeConstructor.Effect(Symbol.mkEffectSym(Name.RootNS, Name.Ident(SourcePosition.Unknown, "IO", SourcePosition.Unknown))), SourceLocation.Unknown)
```
magnus-madsen, line 332: I was thinking of hardcoding the source position of IO. (i.e. we know where it is in Prelude), but I did not get to actually try it. I think such an approach could be generally useful. Perhaps it will not be feasible until we replace SourcePosition by something else when we get to the new Parser.

main/src/ca/uwaterloo/flix/language/ast/UnkindedType.scala
```diff
@@ -359,8 +359,8 @@ object UnkindedType {
   /**
     * Constructs the type a -> b \ IO
     */
-  def mkImpureArrow(a: UnkindedType, b: UnkindedType, loc: SourceLocation): UnkindedType = {
-    val eff = Some(UnkindedType.Cst(TypeConstructor.EffUniv, loc))
+  def mkIoArrow(a: UnkindedType, b: UnkindedType, loc: SourceLocation): UnkindedType = {
+    val eff = Some(UnkindedType.Cst(TypeConstructor.Effect(new Symbol.EffectSym(Nil, "IO", SourceLocation.Unknown)), loc))
```
magnus-madsen, line 363: Perhaps we should move the hardcoded IO symbol to the Symbol.scala class? I think that would be good.

main/src/ca/uwaterloo/flix/language/ast/UnkindedType.scala
```diff
@@ -575,7 +575,7 @@ object UnkindedType {
       val elmType = getFlixType(comp)
       UnkindedType.mkApply(
         UnkindedType.Cst(TypeConstructor.Array, SourceLocation.Unknown),
-        List(elmType, UnkindedType.Cst(TypeConstructor.EffUniv, SourceLocation.Unknown)),
+        List(elmType, UnkindedType.Cst(TypeConstructor.Effect(new Symbol.EffectSym(Nil, "IO", SourceLocation.Unknown)), SourceLocation.Unknown)),
```
magnus-madsen, line 578: Ditto

main/src/ca/uwaterloo/flix/language/ast/Type.scala
magnus-madsen: see later comment about the symbol.

main/src/ca/uwaterloo/flix/language/dbg/DocAstFormatter.scala
```diff
@@ -433,7 +433,7 @@ object DocAstFormatter {
 
   private def formatEffect(effect: Eff, paren: Boolean = true)(implicit i: Indent): Doc = effect match {
     case Eff.Pure => empty
-    case Eff.Impure => text(" ") :: text("\\") +: text("IO")
+    case Eff.Impure => text(" ") :: text("\\") +: text("Impure")
```
magnus-madsen, line 436: When is something now Impure? :)

main/src/ca/uwaterloo/flix/language/fmt/FormatEff.scala
```diff
@@ -22,7 +22,7 @@ object FormatEff {
 
   def formatEff(eff: Type)(implicit flix: Flix): String = eff match {
     case Type.Cst(TypeConstructor.Pure, _) => "Pure"
-    case Type.Cst(TypeConstructor.EffUniv, _) => "IO"
+    case Type.Cst(TypeConstructor.EffUniv, _) => "Impure"
```
magnus-madsen, line 25: I think we have to call top for "Univ"or something like that.

magnus-madsen: Calling it impure, while perhaps correct, would be terribly confusing. Especially given early documentation. So I think a new word is better and I would hope that in practice we would never see "Univ" actually printed.

main/src/ca/uwaterloo/flix/language/phase/MonoDefs.scala
```diff
@@ -444,7 +442,7 @@ object MonoDefs {
               val subst1 = caseSubst @@ subst.nonStrict
               // visit the body under the extended environment
               val body = visitExp(body0, env1, StrictSubstitution(subst1, root.eqEnv))
-              val eff = Type.mkUnion(exp.eff, body0.eff, loc.asSynthetic)
+              val eff = Type.mkUnion(e.eff, body.eff, loc.asSynthetic)
```
magnus-madsen, line 445: I see here the old code was using the wrong (non-monomorphed) sub expressions.

main/src/ca/uwaterloo/flix/language/phase/MonoDefs.scala
```diff
@@ -89,8 +89,6 @@ object MonoDefs {
             case Some(tpe) => tpe.map(default)
             case None => default(t)
           }
-          // Erase concrete effects like Print.
-          case Type.Cst(TypeConstructor.Effect(_), _) => Type.EffUniv
```
magnus-madsen, line 93: So presumably now we create specific versions for every demanded effectset-- which is probably okay. Later we can simplify that to Pure, Impure, ControlImpure.

main/src/ca/uwaterloo/flix/language/phase/MonoTypes.scala
```diff
@@ -437,7 +437,7 @@ object MonoTypes {
     */
   private def specializeEnum(sym: Symbol.EnumSym, args0: List[Type], loc: SourceLocation)(implicit ctx: SharedContext, root: LoweredAst.Root, flix: Flix): Symbol.EnumSym = {
     // type arguments
-    val args = args0.map(eraseAliases).map(TypeNormalization.normalizeType)
+    val args = args0.map(eraseAliases).map(eraseEffects).map(TypeNormalization.normalizeType)
```
magnus-madsen, line 440: Only here? Nowhere else?

What happens when this phase is deleted?

main/src/ca/uwaterloo/flix/language/phase/MonoTypes.scala
```diff
@@ -475,4 +475,20 @@ object MonoTypes {
     Ast.CaseSymUse(new Symbol.CaseSym(freshEnumSym, sym.sym.name, sym.sym.loc), sym.loc)
   }
 
+  /**
+    * Replaces all effects in the type with Impure.
+    */
+  private def eraseEffects(tpe: Type): Type = tpe match {
+    case Type.Cst(TypeConstructor.Effect(_), _) => Type.Impure
```
magnus-madsen, line 482: We still have Type.Impure? We should probably delete it and replace it by Univ.

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -732,8 +732,7 @@ object Simplifier {
     */
   private def simplifyEffect(eff: Type): Purity = eff match {
     case Type.Cst(TypeConstructor.Pure, _) => Purity.Pure
-    case Type.Cst(TypeConstructor.EffUniv, _) => Purity.Impure
-    case _ => throw InternalCompilerException(s"Unexpected purity '$eff'", eff.loc)
+    case _ => Purity.Impure
   }
```
magnus-madsen, line 736: This is pretty simple :)

main/src/ca/uwaterloo/flix/language/phase/TypeInference.scala
```diff
@@ -156,14 +156,11 @@ object TypeInference {
...
                     val declaredEff = declaredScheme.base.arrowEffectType
 
-                    if (declaredEff == Type.Pure && inferredEff == Type.Impure) {
-                      // Case 1: Declared as pure, but impure.
-                      return Validation.toHardFailure(TypeError.ImpureDeclaredAsPure(loc))
-                    } else if (declaredEff == Type.Pure && inferredEff != Type.Pure) {
-                      // Case 2: Declared as pure, but effectful.
+                    if (declaredEff == Type.Pure && inferredEff != Type.Pure) {
+                      // Case 1: Declared as pure, but effectful.
                       return Validation.toHardFailure(TypeError.EffectfulDeclaredAsPure(inferredEff, loc))
```
magnus-madsen, line 161: Pray why is this a hardFailure? Hm

magnus-madsen: Perhaps that is for another ticket.

main/src/library/Debug.flix
```diff
@@ -77,48 +76,48 @@ mod Debug {
...
-                import static java.util.Arrays.toString(Array[##java.lang.Object, rc]): String \ {};
-                toString(unchecked_cast(y as Array[##java.lang.Object, rc]))
-            case y: Array[BigDecimal, rc] =>
-                import static java.util.Arrays.toString(Array[##java.lang.Object, rc]): String \ {};
-                toString(unchecked_cast(y as Array[##java.lang.Object, rc]))
-            case y: Array[String, rc] =>
-                import static java.util.Arrays.toString(Array[##java.lang.Object, rc]): String \ {};
-                toString(unchecked_cast(y as Array[##java.lang.Object, rc]))
+            case y: Array[Bool, _] =>
+                import static java.util.Arrays.toString(Array[Bool, Static]): String \ {};
```
magnus-madsen, line 80: I wonder whether to drop Static?

main/src/library/Debug.flix
```diff
@@ -77,48 +76,48 @@ mod Debug {
...
+                toString(unchecked_cast(y as Array[Bool, Static]))
```
magnus-madsen, line 81: What's the inferred type of y compared to the type you cast to?

magnus-madsen (commented):
Questions:

- Programmers cannot explicitly write top, correct? (which is good, I think)

magnus-madsen:
I feel pretty good about this. What about you?

I guess I wonder:

- Presumably the performance profile in EffUnification changes?
- Why did nothing in unification change? I guess because we have just "added a new effect". We still have top/Univ, it is just less used now.

magnus-madsen:
These changes are 100% necessary, but I wonder if they will help or hurt the switch to the new typer. Or maybe it will just be neutral.

magnus-madsen:
!!!jar

main/src/ca/uwaterloo/flix/language/ast/Type.scala
mlutze: I was thinking that IO might be considered a built-in type rather than part of the standard library, since it is so integral to the language.

main/src/ca/uwaterloo/flix/language/dbg/DocAstFormatter.scala
mlutze: Impure is top. I'll rename to Univ.

main/src/ca/uwaterloo/flix/language/phase/MonoDefs.scala
mlutze: Yes, I will revisit the old type match ticket after this to see if it's magically fixed.

main/src/ca/uwaterloo/flix/language/phase/MonoTypes.scala
mlutze: Indeed, the rules for erasure and normalization are not totally clear. It's a problem that pervades these two phases, hence Jonathan's PR with `NormalType` etc.

main/src/library/Debug.flix
mlutze: IMO it's useful, especially for interop

main/src/library/Debug.flix
mlutze: `y` effectively has the type `exists r. Array[Bool, r]` So there's no type we can give to the import that would not require a cast.

mlutze:
> Questions:
> 
> - Programmers cannot explicitly write top, correct? (which is good, I think)

Incorrect. `Impure` is still a keyword.

mlutze:
> I feel pretty good about this. What about you?

I think it is reasonable. It solves region unsoundness issues and seems to work well.

> - Presumably the performance profile in EffUnification changes?

Yes. I think there are still shortcuts that apply though.

> - Why did nothing in unification change? I guess because we have just "added a new effect". We still have top/Univ, it is just less used now.

Yes.

magnus-madsen:
> > Questions:
> > 
> > * Programmers cannot explicitly write top, correct? (which is good, I think)
> 
> Incorrect. `Impure` is still a keyword.

Should we remove it?

mlutze:
> > > Questions:
> > > 
> > > * Programmers cannot explicitly write top, correct? (which is good, I think)
> > 
> > 
> > Incorrect. `Impure` is still a keyword.
> 
> Should we remove it?

Since `Impure`/`Univ` is going to pop up occasionally in error messages, it would be strange that a user could not use it themselves.

magnus-madsen:
> > > > Questions:
> > > > 
> > > > * Programmers cannot explicitly write top, correct? (which is good, I think)
> > > 
> > > 
> > > Incorrect. `Impure` is still a keyword.
> > 
> > 
> > Should we remove it?
> 
> Since `Impure`/`Univ` is going to pop up occasionally in error messages, it would be strange that a user could not use it themselves.

OK. But what should we call it then? Univ?

mlutze:
> > > > > Questions:
> > > > > 
> > > > > * Programmers cannot explicitly write top, correct? (which is good, I think)
> > > > 
> > > > 
> > > > Incorrect. `Impure` is still a keyword.
> > > 
> > > 
> > > Should we remove it?
> > 
> > 
> > Since `Impure`/`Univ` is going to pop up occasionally in error messages, it would be strange that a user could not use it themselves.
> 
> OK. But what should we call it then? Univ?

Sounds good to me.

mlutze:
Can we merge this and then have the "aesthetic" changes in a followup?

magnus-madsen:
If you make a todo list :)

magnus-madsen:
But I still need to look at performance.  So bot before tonight

mlutze:
Followup to-dos:
- [ ] Check whether https://github.com/flix/flix/issues/7096 is fixed
- [ ] Rename top type to `Univ`
- [ ] Check on whether old typematch mysteries are resolved
- [ ] Non-hard failure for `EffectfulDeclaredAsPure`
- [ ] Remove `ImpureDeclaredAsPure`
- [ ] Hardcode `IO` effect symbol in `Symbol.scala`

magnus-madsen:
!!!jar

magnus-madsen:
- Add a test that https://github.com/flix/flix/discussions/7143 does not compile.

magnus-madsen (approved):

magnus-madsen:
- Reprofile special cases in EffUnification

magnus-madsen:
Will you do these, and if not, make tickets for them?

magnus-madsen:
@stephentetley This could cause breakages; but only good ones: Properly distinguishing regions from IO.

magnus-madsen:
@mlutze I did not observe any slowdown. 

I wonder if this will help or hurt the new type system. Maybe it has no impact... But at least it clarifies things.

PR #7121, opened 2024-01-25
feat: rework enums in backend (MonoTypes)

JonathanStarup:
Upsides: simplified code, simpler types, ...
Downsides:
- Enums terms are not statically types (`Some(42)` has type `Option`)
  - The `Verifier` cannot do as much checking
  - The optimizer has less knowledge.
- This transformation is a form of erasure, but `MonoTypes` is phase 4 out of 15. This might be a bit early.
  - Can't we do it later? The issue is that we cannot represent enums without variables unless we copy and specialize. This is exactly what we want to avoid. If we keep it around, enums will have non-ground types and phases that want to use enum information have to do some substitution / normalization / type normalization work.
- This disables unit tag optimization (where one instance is reused)

Perhaps an idea to preserve the enum structure is to transform
```
enum Option[t] { case None, Some(t) }
```
into
```
enum Option { case None, Some(Erased) }
```
Then you can still check for unit tags and to optimizations


magnus-madsen:
> The optimizer has less knowledge. In this PR root.enums is not set to empty, because Optimizer wants to check the number of cases in a given enum.

What is this used for? Is it for singleton optimization?

JonathanStarup:
If its an enum with a single case, then `Is` is replaced by `true`

magnus-madsen:
> If its an enum with a single case, then `Is` is replaced by `true`. Concretely probably not important but it highlights the loss of information

What a simple, boring, useless optimization :)

main/src/ca/uwaterloo/flix/language/ast/Symbol.scala
```diff
@@ -36,14 +36,6 @@ object Symbol {
     new DefnSym(id, sym.namespace, sym.text, sym.loc)
   }
 
-  /**
-    * Returns a fresh enum symbol based on the given symbol.
-    */
-  def freshEnumSym(sym: EnumSym)(implicit flix: Flix): EnumSym = {
```
magnus-madsen, line 42: I like that we can get rid of this. Thank heavens.

main/src/ca/uwaterloo/flix/language/ast/ReducedAst.scala
```diff
@@ -23,10 +23,9 @@ import java.lang.reflect.Method
 
 object ReducedAst {
 
-  val empty: Root = Root(Map.empty, Map.empty, Map.empty, Set.empty, List.empty, None, Set.empty, Map.empty)
+  val empty: Root = Root(Map.empty, Map.empty, Set.empty, List.empty, None, Set.empty, Map.empty)
 
   case class Root(defs: Map[Symbol.DefnSym, Def],
-                  enums: Map[Symbol.EnumSym, Enum],
```
magnus-madsen, line 29: Good. Its gone. So no-one can mistakenly use it.

main/src/ca/uwaterloo/flix/language/phase/MonoTypes.scala
```diff
@@ -20,84 +20,21 @@ import ca.uwaterloo.flix.api.Flix
...
+import ca.uwaterloo.flix.util.collection.MapOps
 import ca.uwaterloo.flix.util.{InternalCompilerException, ParOps}
 
+import scala.annotation.tailrec
 import scala.collection.mutable
 
 /**
   * This phase does two things:
-  * - Specializes polymorphic enums with the specific types that it is used as.
+  * - Erase enums, such that `Option[t]` becomes `Option`
```
magnus-madsen, line 33: I don't think this has to be a phase anymore. We can just roll it into MonoDefs. 

(Then later, separately, we can rename that back).

magnus-madsen: This would be very nice. Then monomorph can be understood as just one phase.

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -53,7 +53,7 @@ object Simplifier {
   private def visitEnum(decl: LoweredAst.Enum)(implicit flix: Flix): SimplifiedAst.Enum = decl match {
     case LoweredAst.Enum(_, ann, mod, sym, _, _, cases0, enumType, loc) =>
       val cases = cases0.map {
-        case (tag, LoweredAst.Case(caseSym, tagType, _, tagLoc)) => tag -> SimplifiedAst.Case(caseSym, visitType(tagType), tagLoc)
+        case (tag, LoweredAst.Case(caseSym, _, _, tagLoc)) => tag -> SimplifiedAst.Case(caseSym, null /* unused */, tagLoc)
```
magnus-madsen, line 56: Say what? Null in my compiler! What is this.

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -222,30 +222,12 @@ object Verifier {
...
-            case Some(e) => e.cases.get(sym) match {
-              case None => throw InternalCompilerException(s"Unknown enum case sym: '$sym' of '${e.sym}'", loc)
-              case Some(caze) =>
-                val List(t) = ts
-                check(expected = e.tpe)(t, loc)
-                check(expected = caze.tpe)(tpe, loc)
-                tpe
-            }
-          }
+        case AtomicOp.Tag(_) =>
```
magnus-madsen, line 225: Same for AtomicOp.Is too?

main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
```diff
@@ -41,6 +41,8 @@ sealed trait BackendObjType {
     case BackendObjType.Lazy(tpe) => JvmName(RootPackage, mkClassName("Lazy", tpe))
     case BackendObjType.Ref(tpe) => JvmName(RootPackage, mkClassName("Ref", tpe))
     case BackendObjType.Tuple(elms) => JvmName(RootPackage, mkClassName("Tuple", elms))
+    case BackendObjType.Tagged => JvmName(RootPackage, mkClassName("Tagged"))
+    case BackendObjType.Tag(tpe) => JvmName(RootPackage, mkClassName("Tag", tpe))
```
magnus-madsen, line 45: What is the difference between Tagged and Tag?

main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
```diff
@@ -302,6 +304,57 @@ object BackendObjType {
...
+      val cm = ClassMaker.mkAbstractClass(this.jvmName)
+
+      cm.mkConstructor(Constructor)
+
+      cm.mkField(NameField)
+
+      cm.closeClassMaker()
+    }
+
+    def NameField: InstanceField = InstanceField(this.jvmName, IsPublic, NotFinal, NotVolatile, "tag", String.toTpe)
```
magnus-madsen, line 318: Perhaps final?

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -561,45 +562,40 @@ object GenExpression {
...
-        // First we compile the `exp`
+        val taggedType = BackendObjType.Tagged
+
         compileExpr(exp)
-        // We check if the enum is `instanceof` the class
-        mv.visitTypeInsn(INSTANCEOF, classType.name.toInternalName)
+        val ins = {
+          import BytecodeInstructions._
+          CHECKCAST(taggedType.jvmName) ~ GETFIELD(taggedType.NameField) ~
+            pushString(JvmOps.getTagName(sym)) ~ INVOKEVIRTUAL(JavaObject.EqualsMethod)
```
magnus-madsen, line 571: Not equals! That is slow. It should be ==, i.e. if_acmp

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -561,45 +562,40 @@ object GenExpression {
...
-        // Calling the constructor of the class
-        mv.visitMethodInsn(INVOKESPECIAL, classType.name.toInternalName, "<init>", constructorDescriptor, false)
 
-      case AtomicOp.Untag(sym) =>
+        val tagType = BackendObjType.Tag(BackendType.toErasedBackendType(exp.tpe))
+
+        val ins = {
+          import BytecodeInstructions._
+          NEW(tagType.jvmName) ~ DUP() ~ INVOKESPECIAL(tagType.Constructor) ~
+            DUP() ~ pushString(JvmOps.getTagName(sym)) ~ PUTFIELD(tagType.NameField) ~
```
magnus-madsen, line 583: Neat to call String.intern

magnus-madsen:
I think we should merge this in, i.e. it is a good strategy. Do you agree? 

Just a few things to fix.

main/src/ca/uwaterloo/flix/language/phase/MonoTypes.scala
JonathanStarup: I agree, but can we make it a separate PR? I don't think its trivial where to 1) erase alias and 2) transform the types given that monoDefs doesn't even visit every type as it is

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
JonathanStarup: I figured a hard signal like `null` would be better than `Unit` :P

main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
JonathanStarup: Tagged is something that has a `tag: String`. `Tag[t]` is indexed, and has a value of type `t`

main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
JonathanStarup: It's created like
```
let t = new Tag()
t.name = something
t.value = somethingElse
```
So it can't be final

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
JonathanStarup: But how do we guarantee that they are in fact equal? then we need to have one instance of each of the tag names right?

JonathanStarup:
Yes. I do however think that though `Optimizer` doesn't do meaningful enum optimization right now, it feels bad to take enum information away from it. Where would we do singleton enum optimizations or optimize enums with a unit value?

main/src/ca/uwaterloo/flix/language/phase/MonoTypes.scala
magnus-madsen: OK

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
magnus-madsen: We never use null. Why is there no real value here?

main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
magnus-madsen: I don't get it.

main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
magnus-madsen: OK

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
magnus-madsen: Check what intern does.

magnus-madsen: Real string comparison is way too slow.

magnus-madsen:
> Yes. I do however think that though Optimizer doesn't do meaningful enum optimization right now, it feels bad to take enum information away from it. Where would we do singleton enum optimizations or optimize enums with a unit value?

I don't care about these two optimizations. They have very little actual performance impact compared to real optimizations such as inlining. Better to have a good architecture and then later figure out where to place them.

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
JonathanStarup: For Option.Some the type is `t` - but MonoDefs/Types promises to ground the types

main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
JonathanStarup:
```
abstract class Tagged { public String name; }
class Tag$Int32 extends Tagged { public int value }
class Tag$Float32 extends Tagged { public float value }
```

So when you want to compile `e Is `enum.case`, we don't know yet the exact type (it might return false), so we need a common interface on which to ask what its case name is. It becomes `e.name == enum.case`

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
JonathanStarup: Ahh, need to call String.intern, i see 👍

magnus-madsen:
What's the status on this?

JonathanStarup:
I just need to implement your feedback

JonathanStarup:
`ACMP` without intern works fine, i just flipped the boolean by accident :L

JonathanStarup:
RFR @magnus-madsen

main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
magnus-madsen: Right ok

main/src/ca/uwaterloo/flix/language/phase/MonoTypes.scala
```diff
@@ -382,34 +290,31 @@ object MonoTypes {
...
-    // It is important that eraseAliases happens BEFORE enum specialization
-    visitInner(eraseAliases(tpe))
+  private def visitType(tpe: Type): Type = tpe match {
+    case Type.Cst(TypeConstructor.Enum(sym, _), loc) =>
+      // Remove type arguments from enums.
+      Type.Cst(TypeConstructor.Enum(sym, Kind.Star), loc)
+    case Type.Cst(tc, loc) =>
+      Type.Cst(tc, loc)
+    case Type.Apply(tpe1, tpe2, loc) =>
+      Type.Apply(visitType(tpe1), visitType(tpe2), loc)
```
magnus-madsen, line 302: @JonathanStarup I will probably merge this now, but you might want to check if type arguments are actually erased, because at least here you rebuild the Type.Apply which is how we represent type arguments.

magnus-madsen (approved):

magnus-madsen:
Feels good to have this merged in 👍

PR #7129, opened 2024-01-26
refactor: get rid of indexed indices in Indexer

mlutze:
The existing pattern in Indexer:

```
idx1 = ...
idx2 = ...
idx1 ++ idx2 ++ ...
```

is clunky and error-prone.
It is easy to forget to `++` something at the end, and any time we add or remove a field, we have to change the indices.
This new system solves both of those.

If this is deemed reasonable, then I'd like to do the same thing in SemanticTokensProvider.


magnus-madsen (approved):

magnus-madsen:
I think this is reasonable.

Issue #7093, opened 2024-01-22
Port effect handler examples


magnus-madsen:
We already have some effect handler examples (see `examples/effects`). The goal of this ticket is to port more examples.

Here are some places to look for interesting examples:

- https://effekt-lang.org/docs/casestudies
- https://effekt-lang.org/examples
- https://www.eff-lang.org/handlers-tutorial.pdf
- https://koka-lang.github.io/koka/doc/book.html#sec-handlers
- https://v2.ocaml.org/manual/effects.html
- https://www.eff-lang.org/try/

- The goal is to have interesting, useful examples. Not to just port blindly :)
- Focus on short examples, do not spend hours on 100+ line examples.
- Don't port an example if we already have it.
- Be sure to explain/document what is happening.
- Do provide attribute (e.g. the following example from XYZ).
- I would do one example per PR.

magnus-madsen:
@jaschdoc This is something to look at, once the effect handler tests are complete.

stephentetley:
Here's the `yield` effect from the Koka book if you want to add it. I ported yesterday - yielding uses a condition so it can exit early if required.

https://koka-lang.github.io/koka/doc/book.html#why-handlers

```
// Flix version of the Yield effect examaple from the Koka book: 
// https://koka-lang.github.io/koka/doc/book.html#why-handlers

mod Handler.YieldE {

    eff YieldE {
        pub def yieldE(i: Int32): Bool
    }

    pub def traverse(xs: List[Int32]): Unit \ ef + YieldE = match xs {
        case x :: rs => 
            let more = do YieldE.yieldE(x); 
            if (more) traverse(rs) else ()
        case Nil => ()
    }

    pub def printElems(): Unit \ IO = 
        try {
            traverse(1 :: 2 :: 3 :: 4 :: Nil)
        } with YieldE {
            def yieldE(i, k) = {
                println(i);
                k(i <= 2)
            }
        }

}

pub def main(): Unit \ IO = 
    Handler.YieldE.printElems()

```

magnus-madsen:
@stephentetley Yes, that would be nice.

jaschdoc:
Not that it's a port of another example, but I could try to come up with a small example that emulates the supervisors from Erlang / Elixir if that is interesting.

magnus-madsen:
> Not that it's a port of another example, but I could try to come up with a small example that emulates the supervisors from Erlang / Elixir if that is interesting.

I think they have to be concurrent, right?

[...]

PR #7133, opened 2024-01-26
docs: add `PrettyPrinter` effect example

jaschdoc:
Ports the pretty printing example from Effekt: https://effekt-lang.org/docs/casestudies/prettyprinter

Related to https://github.com/flix/flix/issues/7093


magnus-madsen:
This is a nice example, but perhaps its too ambitious as the _first example_. We barely know if the feature works :) :)

jaschdoc:
> This is a nice example, but perhaps its too ambitious as the _first example_. We barely know if the feature works :) :)

True, but I'm nearly done with it. Just need to figure out what the `show` function does and then add the rest along with documentation and examples / tests.

I picked this because many of the examples are either complicated or not really useful or trivial, so I figured this may be a good place as any to start :)

jaschdoc:
Not sure this is possible due to https://effekt-lang.org/docs/concepts/effect-safety#higher-order-functions-and-relative-purity

> Here, the block passed to each does use another effect, namely the Console effect. That is, only from the point of view of each, the block f is pure.
>
> This shift in perspective, provides guarantees for the caller of each: the implementation cannot observe (besides global side channeling) any effects used by f.
>
> This comes at the cost of some guarantees for the implementor of each: Calling f can have side-effects observable to the user, losing referential transparency to some extent. We cannot simply replace a call to f by its results. Also, calling it twice is different from calling it once.

This seems to be the opposite of Flix, i.e., a higher-order function can restrict its parameters to exactly the declared effects. It seems that the examples that use the pretty printing library call the functions with more effects than declared.

jaschdoc:
I think we should close this. We can revisit, but it requires a rewrite entirely. At the time I did not know the difference between Effekt's effect system and ours.

jaschdoc:
It seems, though, that this example can work in Flix, we just have to think a bit more about the effects. Maybe there is a hidden requirement of shallow handlers here too so that could be a roadblock.

PR #7136, opened 2024-01-26
refactor: use IteratorOps.all in SemanticTokensProvider

mlutze:


magnus-madsen:
Compile issues.

magnus-madsen (approved):

magnus-madsen:
Is this now done consistently everywhere?

mlutze:
> Is this now done consistently everywhere?

Those were the two places of which I am aware where the `x1` `x2` etc. pattern was

Issue #7036, opened 2024-01-14
Change `Parser.FormalParamList` to produce `ParsedAst.FormalParamList`


magnus-madsen:
Add `ParsedAst.FormalParamList` with sp1 and sp2.

The idea is that we need to know the position before and after the '(' and ')' so we can give right positions.

magnus-madsen:
Related to https://github.com/flix/flix/issues/7032

magnus-madsen:
@jaschdoc Please consider this as the same priority of the effect and handler tests.

magnus-madsen:
Keeping this open because I need to use this information.

PR #7147, opened 2024-01-29
refactor: add `ParsedAst.FormalParamList`

jaschdoc:
Closes https://github.com/flix/flix/issues/7036


jaschdoc:
@magnus-madsen I spent some time making the constructor private, but I'm not sure it's worth the extra complexity. Commit edc9889 has a simpler implementation which we should probably just use.

magnus-madsen:
Definitely not worth it. Just make it a case class like everything else.

main/src/ca/uwaterloo/flix/language/ast/ParsedAst.scala
```diff
@@ -1771,6 +1771,12 @@ object ParsedAst {
     */
   case class FormalParam(sp1: SourcePosition, mod: Seq[ParsedAst.Modifier], ident: Name.Ident, tpe: Option[ParsedAst.Type], sp2: SourcePosition)
 
+  /**
```
magnus-madsen, line 1774: Comment

magnus-madsen (approved):

Issue #7145, opened 2024-01-28
Add negative test cases to TestTyper for IO + user-def effects


magnus-madsen:
Add a few tests like:

```
eff Ask {
    pub def ask(): String
}

def main(): Unit \ IO = 
    do Ask.ask();
    println("Hello World!")
```

which should fail because `Ask` is unhandled.

PR #7154, opened 2024-01-29
test: add negative test cases for `IO` and effects

jaschdoc:
Closes https://github.com/flix/flix/issues/7145

### DIFF
--- a/main/test/ca/uwaterloo/flix/language/phase/TestTyper.scala
+++ b/main/test/ca/uwaterloo/flix/language/phase/TestTyper.scala
@@ -1310,4 +1310,30 @@ class TestTyper extends AnyFunSuite with TestUtils {
     expectError[TypeError.UnexpectedEffect](result)
   }
 
+  test("TestIOAndCustomEffect.06") {
+    val input =
+      """
+        |eff Gen {
+        |    pub def gen(): String
+        |}
+        |
+        |eff AskTell {
+        |    pub def askTell(x: Int32): Int32
+        |}
+        |
+        |pub def f(): Unit \ IO =
+        |    let _ = try {
+        |        do Gen.gen();
+        |        do AskTell.askTell(42)
+        |    } with Gen {
+        |        def gen(k) = k("a")
+        |    };
+        |    do Gen.gen();
+        |    do AskTell.asktell(42);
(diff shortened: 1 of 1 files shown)


jaschdoc:
`TestIOAndCustomEffect.03` produces `PossibleCheckedEffectCast` even when wrapping in `checked_ecast`:
```
>> Expected effect: 'IO' but found effect: 'e2 + e0 + e1'.

12 |> pub def f(): Unit \ IO = checked_ecast({
13 |>     let _ = try {
14 |>         do Gen.gen();
15 |>         do AskTell.askTell(42)
16 |>     } with Gen {
17 |>         def gen(k) = k("a")
18 |>     };
19 |>     checked_ecast(())
20 |> })

expression has unexpected effect.

'IO' appears to be a superset of 'e2 + e0 + e1'.
Consider using 'checked_ecast'?
```

magnus-madsen:
Can you drop the checked casts? The goal is for the program not to type check.

magnus-madsen:
If it gives PossibleCheckCast that is OK for now. It is still an error at least. (Just the wrong one).

jaschdoc:
> Can you drop the checked casts? The goal is for the program not to type check.

The effect casts are only around the `()` (unit values) to avoid using a library function so tests are faster. The errors messages posted above are checked in VSCode both with and without the effect casts around the `try-with` expression.

magnus-madsen:
What errors? I don't want a checked cast. I just want that if you have a user-defined effect then you cannot write IO.

jaschdoc:
> What errors? I don't want a checked cast. I just want that if you have a user-defined effect then you cannot write IO.

This error: https://github.com/flix/flix/pull/7154#issuecomment-1914342895

No checked casts at all?

jaschdoc:
Ah wait I understand now

magnus-madsen:
Yeah, I should explain better.

Old days: IO was the biggest effect. So if you had the Ask and Print effect they were subsumed by IO.
New days: IO just an effect like any other. It should _NOT_ subsume any effects.

jaschdoc:
Error is pipeline-related, not this PR itself

magnus-madsen (approved):

Issue #7079, opened 2024-01-20
Add syntactic sugar to allow multiple `with` blocks


magnus-madsen:
For example, instead of:

```scala
def main(): Unit \ IO = 
    try {
        try greeting() with Ask {
            def ask(k) = k("Bond, James Bond")
        }
    } with Say {
        def say(s, k) = { println(s); k() }
    }
```

We want to just write one `try` and then attach multiple `with` blocks. Note that the desugaring must ensure that the first handler is innermost.

jaschdoc:
For reference:

```scala
def main(): Unit \ IO = 
    try {
        greeting()
    } with Ask {
        def ask(k) = k("Bond, James Bond")
    } with Say {
        def say(s, k) = { println(s); k() }
    }
```
becomes
```scala
def main(): Unit \ IO = 
    try {
        try greeting() with Ask {
            def ask(k) = k("Bond, James Bond")
        }
    } with Say {
        def say(s, k) = { println(s); k() }
    }
```

magnus-madsen:
Yes

jaschdoc:
I'm not sure this is possible currently.

This program fails:
```scala

    eff A {
        pub def a(): Unit
    }
    eff B {
        pub def b(): Unit
    }

    def greeting(): Unit \ A + B =
        do A.a();
        do B.b()

    @Test
    def testGreeting(): Bool =
        let result = try {
            try {
                greeting()
            } with A {
                def a(k) = k()
            }
        } with B {
            def b(_) = ()
        };
        Assert.eq((), result)
```

```
Unable to compile. Failed with: 1 errors.


>> Expected effect: 'Pure' but found effect: 'e0'.

139 |>         let result = try {
140 |>             try {
141 |>                 greeting()
142 |>             } with A {
143 |>                 def a(k) = k()
144 |>             }
145 |>         } with B {
146 |>             def b(_) = ()
147 |>         };
148 |>         Assert.eq((), result)

expression has unexpected effect.

'Pure' appears to be a superset of 'e0'.
Consider using 'checked_ecast'?
```

I think @JonathanStarup explained that it's hard to remove one of the effects since a type variable may be bound in a substitution. Maybe he can clarify. We could potentially desugar to functions where the types are ascribed as that solves the problem but then we need to know the type of `greeting` which we do not know during desugaring.

magnus-madsen:
Just focus on the desugaring. We are dealing with the types and effects elsewhere.

magnus-madsen:
[...]

PR #7156, opened 2024-01-29
feat: allow multiple handlers for single `try`-block

jaschdoc:
Closes https://github.com/flix/flix/issues/7079


main/src/ca/uwaterloo/flix/language/phase/Parser.scala
```diff
@@ -1027,17 +1027,27 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
...
 
-      rule {
+      def NormalTry: Rule1[ParsedAst.Expression.Try] = rule {
         SP ~ keyword("try") ~ WS ~ Expression ~ optWS ~ Body ~ SP ~> ParsedAst.Expression.Try
       }
+
+      def ChainedTry: Rule1[ParsedAst.Expression.TryChainedHandlers] = rule {
+        SP ~ keyword("try") ~ WS ~ Expression ~ optWS ~ HandlerBody ~ optWS ~ oneOrMore(HandlerBody).separatedBy(optWS) ~ SP ~>
+          ((sp1: SourcePosition, exp: ParsedAst.Expression, handler1: ParsedAst.CatchOrHandler.Handler, otherHandlers: Seq[ParsedAst.CatchOrHandler.Handler], sp2: SourcePosition) =>
+            ParsedAst.Expression.TryChainedHandlers(sp1, exp, Seq.concat(Seq(handler1), otherHandlers), sp2))
```
jaschdoc, line 1045: Maybe this could be simplified by just adding `handler1` to the case class

mlutze:
Why do we need separate chained and unchained tries?

jaschdoc:
> Why do we need separate chained and unchained tries?

Do you want to desugar during parsing or what do you have in mind? :)

jaschdoc:
> > > Why do we need separate chained and unchained tries?
> > 
> > 
> > Do you want to desugar during parsing or what do you have in mind? :)
> 
> It seems like `Try` with only one `With` is just a special case of `Try` with `n` `With`s.

True, but I think @magnus-madsen  just wanted to desugar it and have simpler typing rules? Could this partially solve the problem with the substitution we talked about last Friday?

mlutze:
> > > > Why do we need separate chained and unchained tries?
> > > 
> > > 
> > > Do you want to desugar during parsing or what do you have in mind? :)
> > 
> > 
> > It seems like `Try` with only one `With` is just a special case of `Try` with `n` `With`s.
> 
> True, but I think @magnus-madsen just wanted to desugar it and have simpler typing rules? Could this partially solve the problem with the substitution we talked about last Friday?

Desugaring is fine. I just think that we don't need the "unchained" version in the parser.

jaschdoc:
Ah okay

jaschdoc:
Should we allow combining `catch` and `with`?
```scala
try { ... } catch { ... } with { ... }
```

mlutze:
> Should we allow combining `catch` and `with`?
> ```scala
> try { ... } catch { ... } with { ... }
> ```

Probably not :D

jaschdoc:
Ok I'm currently simplifying the parser but it seems weeding just becomes much more complicated. Lots of edge cases to handle the rules from the main branch. Is that worth it?

jaschdoc:
RFR

jaschdoc:
Weeding becomes significantly more complicated if we collapse the parsing into one rule

magnus-madsen:
Can we not parse: `Try ... (CatchHandlerList | WithHandlerList)`?

jaschdoc:
> Is it because of some idea that we must desugar later?

Yes but also because we handle all parsing with 1 rules so everything has to be enforced during weeding

magnus-madsen:
What if we add ParsedAst.CatchList and ParsedAst.HandlerList?

magnus-madsen:
A try block can have multiple catch blocks with the exact same desugaring 👍

magnus-madsen:
Those two tricks should help alot?

jaschdoc:
> Can we not parse: `Try ... (CatchHandlerList | WithHandlerList)`?

That could improve the solution. There is also a much simpler solution in 125d217

jaschdoc:
I will try the `CatchHandlerList | WithHandlerList` solution now :)

magnus-madsen:
@mlutze Ill let you take a stab and then I can come back.

main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
```diff
@@ -689,10 +689,13 @@ object Desugar {
...
 
-    case WeededAst.Expr.TryWith(exp, eff, rules, loc) =>
+    case WeededAst.Expr.TryWith(exp, handlers, loc) =>
       val e = visitExp(exp)
-      val rs = rules.map(visitHandlerRule)
-      Expr.TryWith(e, eff, rs, loc)
+      handlers.foldLeft(e) {
+        case (acc, handler) =>
+          val rs = handler.rules.map(visitHandlerRule)
+          Expr.TryWith(acc, handler.eff, rs, loc.asSynthetic)
```
mlutze, line 697: I think this can be a "real" location.

mlutze (approved):

main/src/ca/uwaterloo/flix/language/ast/ParsedAst.scala
```diff
@@ -1040,12 +1040,12 @@ object ParsedAst {
     /**
       * Try Expression.
       *
-      * @param sp1            the position of the first character in the expression.
-      * @param exp            the guarded expression.
-      * @param catchOrHandler the handler (catch/with) of the try expression.
-      * @param sp2            the position of the last character in the expression.
+      * @param sp1      the position of the first character in the expression.
+      * @param exp      the guarded expression.
+      * @param handlers list of the handlers (catch/with) of the try expression.
```
magnus-madsen, line 1045: the list of catch or with blocks of the try expression.

main/src/ca/uwaterloo/flix/language/ast/ParsedAst.scala
```diff
@@ -1873,26 +1873,47 @@ object ParsedAst {
...
+    case class Catch(rules: Seq[ParsedAst.CatchRule])
 
     /**
       * A `with` block for handling Flix effects.
       *
       * @param eff   the effect to be handled.
       * @param rules the handler rules.
       */
-    case class Handler(eff: Name.QName, rules: Option[Seq[ParsedAst.HandlerRule]]) extends CatchOrHandler
+    case class WithHandler(eff: Name.QName, rules: Option[Seq[ParsedAst.HandlerRule]])
```
magnus-madsen, line 1915: Why are rules here optional?

jaschdoc: Not sure, it was like this before. Maybe because you can have an empty handler?

mlutze: IMO don't complicate the parser to make types nice in ParsedAst.

magnus-madsen: But why is it optional in the parser?

mlutze: Because you can do `with E` (no braces) when `E` has no ops

mlutze: I guess it's a design decision. Is an empty effect like an empty enum?

magnus-madsen: I would drop it. We can reintroduce if the need arise.

With empty enums there was actually a use case.

jaschdoc: Apparently we allowed this
```scala
// Test.Exp.Effect.flix
    // @test
    pub def handleEff03(): Bool = {
        try true with Empty
    }
```

magnus-madsen: Yeah, I don't think we should do that :)

magnus-madsen:
Test failires

jaschdoc:
Yes this is unsolvable right now as discussed earlier (type inference problem) so the feature is here but does not work :/

magnus-madsen:
> Yes this is unsolvable right now as discussed earlier (type inference problem) so the feature is here but does not work :/

Change the tests? Catch the same effect?

main/test/flix/Test.Exp.TryWith.flix
```diff
@@ -90,23 +90,114 @@ mod Test.Exp.TryWith {
...
+        try {
+            greetAll("Cheshire Cat" :: "Queen of Hearts" :: "White Rabbit" :: "Dormouse" :: Nil)
+        } with Greet {
+            def greet(_, _cont) = ()
+        } with Greet {
+            def greet(_, _cont) = ()
+        }
+    }
+
+    // TODO disabled until better type system is in
```
magnus-madsen, line 182: I prefer not to have TODOs and comment code. Just delete what does not work or adjust it :)

magnus-madsen: It also messes with our linecount etc.

jaschdoc: Roger

magnus-madsen (approved):

magnus-madsen:
Thanks!

PR #7171, opened 2024-01-30
feat: introduce Stmt to backend

### LINKED ISSUE #7169: Introduce `Expr.Stm` to reduce let-bindings with unused variables
- Introduce in `Inliner` for lets with dead symbols, but impure bodies. Near line 213.

JonathanStarup:
Fixes #7169

building without treeshaker1 and 2 reduces letbinders from 2380 to 2261 (5% reduction)

There is still plenty unused letbinder (189 places of `let _something: Unit = ();`)

optimizer runs | master | PR | reduction % (`(master - PR) / master`)
--- | --- | --- | --- |
2 | 2380 | 2261 | 5%
3 | 1974 | 1855 | 6%
4 | 1672 | 1553 | 7%
5 | 1572 | 1453 | 8%

You can see that this PR is more effective the more times you loop optimizer with 5 loop iterations

Also a fun fact, around 50% of the letbinder at the last AST is from the EffectBinder phase


main/src/ca/uwaterloo/flix/language/phase/Inliner.scala
```diff
@@ -212,8 +212,12 @@ object Inliner {
       /// Both code size and runtime are reduced
       if (isDeadAndPure(occur, exp1.purity)) {
         visitExp(exp2, subst0)
-      } else {
+      } else if (occur == Dead && occur != DontInline) {
```
magnus-madsen, line 215: I would prefer a helper method named isDeadAndImpure right next to isDeadAndPure.

main/src/ca/uwaterloo/flix/language/phase/Inliner.scala
```diff
@@ -246,6 +250,18 @@ object Inliner {
       val e2 = visitExp(exp2, subst0)
       LiftedAst.Expr.LetRec(varSym, index, defSym, e1, e2, tpe, purity, loc)
 
+    case OccurrenceAst.Expression.Stmt(exp1, exp2, tpe, purity, loc) =>
+      /// Case 1:
```
magnus-madsen, line 254: Can this case actually happen or is it dead code?

magnus-madsen:
> There is still plenty unused letbinder (189 places of let _something: Unit = ();)

Keep in mind that the optimizer computes a fixpoint-- except we stop before the fixpoint.

In other words, each iteration will remove code, and introduce new opportunities for code removal.

But at some point we stop and we may be left with code that would have been removed in the next round.

JonathanStarup:
> > There is still plenty unused letbinder (189 places of let _something: Unit = ();)
> 
> Keep in mind that the optimizer computes a fixpoint-- except we stop before the fixpoint.
> 
> In other words, each iteration will remove code, and introduce new opportunities for code removal.
> 
> But at some point we stop and we may be left with code that would have been removed in the next round.

Yeah, I found some data, it looks like we should consider running it another more rounds

main/src/ca/uwaterloo/flix/language/phase/Inliner.scala
JonathanStarup: The optimizer loops so it will give itself Stmts.

PR #7175, opened 2024-01-31
feat: formal argument ocurrence information

### LINKED ISSUE #7172: make Optimizer handle unused parameters
In the AST there is a lot of unused parameters (looking specifically at `_lift$xyz` variables. After a run of the optimizer, there variables are inlined somehow, resulting in a lot of binders like `let _lift$150000: Unit = ();`. They are not easily removed as they keep appearing while closures are inlined so we should avoid them being produced at all.

I think this can be fixed in `Inliner.bindFormals` line 342.

There needs to be a check for purity too since arguments to unused parameters might be impure

JonathanStarup:
Fixes #7172 

Stores formal parameter occur information and uses it to eliminate let-binders and dead pure expressions upon inlining.


main/src/ca/uwaterloo/flix/language/phase/Inliner.scala
```diff
@@ -318,13 +318,23 @@ object Inliner {
...
+
   /**
     * Recursively bind each argument in `args` to a let-expression with a fresh symbol
     * Add corresponding symbol from `symbols` to substitution map `env0`, mapping old symbols to fresh symbols.
     * Substitute variables in `exp0` via the filled substitution map `env0`
     */
   private def bindFormals(exp0: OccurrenceAst.Expression, symbols: List[Symbol.VarSym], args: List[LiftedAst.Expr], env0: Map[Symbol.VarSym, Symbol.VarSym])(implicit root: OccurrenceAst.Root, flix: Flix): LiftedAst.Expr = {
     (symbols, args) match {
+      case (sym :: nextSymbols, e1 :: nextExpressions) if isWildAndPure(sym, e1) =>
+        // if the parameter unused and the arguemnt is pure, then throw it away.
```
magnus-madsen, line 336: argument

main/src/ca/uwaterloo/flix/language/ast/OccurrenceAst.scala
```diff
@@ -27,7 +27,7 @@ object OccurrenceAst {
                   reachable: Set[Symbol.DefnSym],
                   sources: Map[Source, SourceLocation])
 
-  case class Def(ann: Ast.Annotations, mod: Ast.Modifiers, sym: Symbol.DefnSym, cparams: List[OccurrenceAst.FormalParam], fparams: List[OccurrenceAst.FormalParam], exp: OccurrenceAst.Expression, context: DefContext, tpe: MonoType, purity: Purity, loc: SourceLocation)
+  case class Def(ann: Ast.Annotations, mod: Ast.Modifiers, sym: Symbol.DefnSym, cparams: List[(OccurrenceAst.FormalParam, Occur)], fparams: List[(OccurrenceAst.FormalParam, Occur)], exp: OccurrenceAst.Expression, context: DefContext, tpe: MonoType, purity: Purity, loc: SourceLocation)
```
JonathanStarup, line 30: Ocurence information is in a tuple because not all formal parameters have occurrence information. effect operations don't have implementations so the field would be meaningless there.

magnus-madsen:
(I promise I will get to it :))

JonathanStarup:
yeah yeah, im just keeping the soup warm 😄

main/src/ca/uwaterloo/flix/language/ast/OccurrenceAst.scala
magnus-madsen: I see. Would it make sense to make it optional on FormalParam then, or would that make everything worse?

JonathanStarup: Yeah i tried it out but formal param is used in effects and trycatch handlers etc - makes too much of a mess in my opinion

magnus-madsen (approved):

PR #7176, opened 2024-01-31
feat: add Verifier Ast node name on error

JonathanStarup:


magnus-madsen:
I am not merging this because it is not the direction I have in mind.

I want to turn the prints into exceptions. Once that happens then there is no reason to carry the AST node. Instead, simply place a breakpoint on the throw statement and start debugging from there.

PR #7177, opened 2024-01-31
fix: Make eraser have correct types

JonathanStarup:
I'm not sure this should be merged - let me explain:
If you call a function like `f: String -> String`, then the function is actually erased into `fe: String -> Obj` (`Obj` represents a `Value` wrapper) which means that `f(42)` becomes `cast (untag f(42): Object) as String`. This transformation is type correct and the Verifier likes it

When a tail-recursive call is made, the cast and untag expression around the calls are omited because the tail call will stop computation anyway. But now it looks like `f(42)` returns Object and the surrounding context wants a String. 

This PR adds back in these cast and untag operations that will never be run.Maybe we should instead fix it by making the verifier be tail-call aware - instead of checking a tail call like any other call it will just verify that tailcalls have the same type as the surrounding function?

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/Flix.scala
+++ b/main/src/ca/uwaterloo/flix/api/Flix.scala
@@ -616,6 +616,7 @@ class Flix {
     cachedEraserAst = Eraser.run(cachedEffectBinderAst)
     cachedReducerAst = Reducer.run(cachedEraserAst)
     cachedVarOffsetsAst = VarOffsets.run(cachedReducerAst)
+    Verifier.run(cachedVarOffsetsAst)
     val result = JvmBackend.run(cachedVarOffsetsAst)
 
     // Write formatted asts to disk based on options.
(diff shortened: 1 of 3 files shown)


magnus-madsen:
Not sure either. Let us talk offline. I don't think we should make changes just to accommodate the verifier. Its a safety net, but it does not have to be perfect.

JonathanStarup:
@magnus-madsen how do you think we shoud solve it? is it too much to have verifier be able to return anytype (different from the monodefs anytime). The visitExpr function could return an Option[MonoType] where None means the expression doesn't return

magnus-madsen:
Can we special case the verifier or just leave that case omitted?

JonathanStarup:
The issue is that its the `tpe` of applyDef/Clo that is wrong, so any call to visitExp might return the wrong type and the inside apply doesn't know that the outer expression expects

JonathanStarup:
Alternative solution was made #7564

JonathanStarup:
Oh, actually the same solution. In conclusion, this changes make the verifier run for very little code size (conjectured)

Issue #7185, opened 2024-02-04
Update other files in `examples` to be valid


magnus-madsen:
It seems there are other files, e.g. lambda-calculus.flix that needs a make over.

magnus-madsen:
And to become subject to automated testing.

PR #7186, opened 2024-02-05
test: add old examples to test suites

jaschdoc:
Closes https://github.com/flix/flix/issues/7185


magnus-madsen (approved):

magnus-madsen:
Excellent. Thanks! Good to have these back under automated test. I guess later we can consider some revisions.

Issue #6585, opened 2023-10-20
Allow meta-constructor in `typematch`?


magnus-madsen:
E.g. allow:

```scala
     typematch(f) {
          case Pure(g) => parMapWithKey(g, t)
          case _       => seqMapWithKey(f, t)
     }

```

magnus-madsen:
@mlutze What do you think about this syntactic sugar?

mlutze:
Can we define that in the standard lib?

```scala
enum EffMatch[a: Type, b: Type, ef: Eff] {
    case Pure(a -> b)
    case Impure(a -> b \ ef)
}

def effmatch(x: f: a -> b \ ef): EffMatch[a, b, ef] = typematch f {
    case g: a -> b \ Pure => Pure(g)
    case _: _ => Impure(f)
}

**Comment by magnus-madsen** [comment]
Interesting idea

**Comment by magnus-madsen** [comment]
I propose to add this to the Prelude (What Matt had, but renamed)

```
enum Purity[a: Type, b: Type, ef: Eff] {
    case Pure(a -> b)
    case Impure(a -> b \ ef)
}

def purityOf(f: a -> b \ ef): Purity[a, b, ef] = typematch f {
    case g: a -> b \ Pure => Pure(g)
    case _: _ => Impure(f)
}
```

Although: I think the name `Pure` clashes with the effect of the same name?

If the name is taken, perhaps `PureFn`and `ImpureFn` can be used.

While this little trick makes the code cleaner, it still does not solve the problem for curried functions.

magnus-madsen:
@jashdoc Could you take a stab? Add the above enum and function, and then use it in two different places so we can see how it looks/feels.

jaschdoc:
Sure :)

PR #7187, opened 2024-02-05
feat: add meta constructor for `typematch`

jaschdoc:
Closes https://github.com/flix/flix/issues/6585

### DIFF
--- a/main/src/library/Prelude.flix
+++ b/main/src/library/Prelude.flix
@@ -164,3 +164,13 @@ pub def printUnlessUnit(x: a): Unit \ IO with ToString[a] = {
         case _: _ => println(x)
     }
 }
+
+enum Purity[a: Type, b: Type, ef: Eff] {
+    case Pure(a -> b)
+    case Impure(a -> b \ ef)
+}
+
+def purityOf(f: a -> b \ ef): Purity[a, b, ef] = typematch f {
+    case g: a -> b \ Pure => Purity.Pure(g)
+    case _: _             => Purity.Impure(f)
+}
--- a/main/src/library/DelayList.flix
+++ b/main/src/library/DelayList.flix
@@ -449,9 +449,10 @@ mod DelayList {
     ///
     @Experimental @LazyWhenPure
     pub def mapWithIndex(f: (Int32, a) -> b \ ef, l: DelayList[a]): DelayList[b] \ ef =
-        match purityOf(f) {
-            case Purity.Pure(g)   => mapWithIndexL(g, l)
-            case Purity.Impure(g) => mapWithIndexE(g, l)
+        typematch f {
+            case g: (Int32, a) -> b \ {} => mapWithIndexL(g, l)
+            case g: (Int32, a) -> b \ ef => mapWithIndexE(g, l)
+            case _: _                    => unreachable!()
         }
 
     ///
--- a/main/src/library/DelayList.flix
+++ b/main/src/library/DelayList.flix
@@ -497,10 +497,9 @@ mod DelayList {
     ///
     @Experimental @LazyWhenPure
     pub def flatMap(f: a -> DelayList[b] \ ef, l: DelayList[a]): DelayList[b] \ ef =
-        typematch f {
-            case g: a -> DelayList[b] \ {} => flatMapL(g, l)
-            case g: a -> DelayList[b] \ ef => flatMapE(g, l)
-            case _: _                      => unreachable!()
+        match purityOf(f) {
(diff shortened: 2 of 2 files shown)


jaschdoc:
It seems to improve readability a lot in my opinion

magnus-madsen:
> It seems to improve readability a lot in my opinion

I think so too. @mlutze Should we do it everywhere?

mlutze:
> > It seems to improve readability a lot in my opinion
> 
> I think so too. @mlutze Should we do it everywhere?

Seems OK.

IMO we should hide this under `Reflect.flix` rather than putting it in the prelude.

I also slightly prefer just making `purityOf2` or something like that instead of using `curry`.

jaschdoc:
> I also slightly prefer just making `purityOf2` or something like that instead of using `curry`.

Hmm... yeah I get that but I also thought that adding `curry` / `uncurry` would be more general and would be a handy helper anyway (we already have stuff like `id`)

magnus-madsen:
I actually prefer purityOf2. Its less likely to go wrong. Please use that :)

jaschdoc:
Should I move it to `Reflect.flix`?

magnus-madsen:
> Should I move it to `Reflect.flix`?

I think it is OK for now. (Unless Matt fights me?)

But, could you add some tests for purityOf and purityOf2. We just need some evidence that they actually work.

(Not that your code is wrong, but monomorph could be broken.)

main/src/library/Prelude.flix
```diff
@@ -164,3 +164,23 @@ pub def printUnlessUnit(x: a): Unit \ IO with ToString[a] = {
...
+
+pub def purityOf2(f: a -> b -> c \ ef): Purity[(a, b), c, ef] = typematch f {
+    case g: (a, b) -> c \ Pure => Purity.Pure(match (x, y) -> g(x, y))
+    case _: _                  => Purity.Impure(match (x, y) -> f(x, y))
+}
+
+pub def purityOf3(f: a -> b -> c -> d \ ef): Purity[(a, b, c), d, ef] = typematch f {
+    case g: (a, b, c) -> d \ Pure => Purity.Pure(match (x, y, z) -> g(x, y, z))
+    case _: _                  => Purity.Impure(match (x, y, z) -> f(x, y, z))
+}
```
mlutze, line 186: These need documentation.

Could we do the currying thing inside these functions so that the user of the functions doesn't have to handle any currying stuff?

magnus-madsen: Ah yes. Brilliant idea!

jaschdoc: > These need documentation.
> 
> Could we do the currying thing inside these functions so that the user of the functions doesn't have to handle any currying stuff?

I tried experimenting with currying inside `purityOf` but I couldn't make it work. It infers `a -> (b -> c) \ ef` where `a -> (b -c \ ef) is expected`

jaschdoc:
Added new types for other function arities.

jaschdoc:
It's pretty hard to get it working without either new types or curry / uncurry at call site. Sort of like a whack-a-mole.

magnus-madsen:
@mlutze I am not sure what we proposed can work.

mlutze:
What about just making the Purity thing like an `Either` type? 

```scala
enum Purity[p, i] {
    case Pure(p)
    case Impure(i)
}
```

So we no longer enforce that it's really pure or impure in the type, but we have the flexibility in what goes in and out.

magnus-madsen:
> What about just making the Purity thing like an `Either` type?
> 
> ```scala
> enum Purity[p, i] {
>     case Pure(p)
>     case Impure(i)
> }
> ```
> 
> So we no longer enforce that it's really pure or impure in the type, but we have the flexibility in what goes in and out.

Or enum Purity2, enum Purity3. What shall we do?

magnus-madsen:
What's the safe, sane choice?

mlutze:
> What's the safe, sane choice?

I think we necessarily have to trade a bit of safety for elegance:

- `Purity[p, i]`
  - less safe because `p` and `i` could be anything
  - more elegant because we don't have to introduce a new type for each arity
- `PurityN`
  - (the opposite of the above)

I have a slight preference for the first of these because this is an "advanced" feature. We don't expect people to construct their own `Purity` values anyway.

jaschdoc:
I prefer the latter since it has consistent name (least surprises) and is also the safest

magnus-madsen:
I think this is nice and a big step forward, so I will merge it.

If we want to fight over details, we can do it later, in subsequent PR. For now I think this is great 👍

PR #7189, opened 2024-02-05
fix: make record code gen unreliant on type ordering

JonathanStarup:


mlutze (commented):
Did you add any tests?

JonathanStarup:
> Did you add any tests?

I cannot - it is unobservable with current inference

magnus-madsen (approved):

Issue #7188, opened 2024-02-05
Datalog Ford-Fulkerson


jaschdoc:
Here are the functions that I think may be relevant so far. I think I will adjust it when I explore the implementation a bit more. I will update this issue when I have an idea of how to approach implementing it.

```scala
mod Graph {

    // Consider returning just Int32
    // Maybe generalize Constraint and Flow to typeclass so it can be both floats and ints
    type alias Constraint = Int32
    type alias Flow = Int32

    pub def fordFulkerson(g: m[(t, t, Constraint)]): Flow with Foldable[m], Order[t] = ???

    pub def fordFulkersonNetwork(g: m[(t, t, Constraint)]): Set[(t, t, Flow)] with Foldable[m], Order[t] = ???

    pub def flowValue(g: m[(t, t)]): Int32 with Foldable[m], Order[t] = ???

    pub def source(g: m[(t, t)]): Option[t] with Foldable[m], Order[t] = ???

    pub def hasSource(g: m[(t, t)]): Bool with Foldable[m], Order[t] = ???

    pub def hasSink(g: m[(t, t)]): Bool with Foldable[m], Order[t] = ???

    pub def sink(g: m[(t, t)]): Option[t] with Foldable[m], Order[t] = ???

    pub def isSTNetwork(g: m[(t, t)]): Bool with Foldable[m], Order[t] = ???

    pub def paths(src: {src = t}, dst: {dst = t}, g: m[(t, t)]): Set[Set[(t, t)]] with Foldable[m], Order[t] = ???

}
```

jaschdoc:
@magnus-madsen could you assign me to this?

PR #7191, opened 2024-02-05
feat: add Datalog-based Ford-Fulkerson algorithm

jaschdoc:
Closes https://github.com/flix/flix/issues/7188


jaschdoc:
Just a dump of functions... (note to self)


```scala
mod Graph {

    use Path.Path;

    enum Path {
        case Path(List[Int32])
    }

    instance Eq[Path] {
        pub def eq(x: Path, y: Path): Bool =
            let Path(xs) = x;
            let Path(ys) = y;
            List.length(xs) == List.length(ys)
    }

    instance Order[Path] {
        pub def compare(x: Path, y: Path): Comparison =
            let Path(xs) = x;
            let Path(ys) = y;
            if (List.length(xs) ==  List.length(ys))
                Comparison.EqualTo
            else if (List.length(xs) <= List.length(ys))
                Comparison.GreaterThan
            else
                Comparison.LessThan
    }

    instance LowerBound[Path] {
        // The longest list
        pub def minValue(): Path = Path(List.range(1, 100))
    }

    instance PartialOrder[Path] {
        pub def lessEqual(x: Path, y: Path): Bool =
            let Path(xs) = x;
            let Path(ys) = y;
            List.length(xs) >= List.length(ys)
    }

    instance JoinLattice[Path] {
        pub def leastUpperBound(x: Path, y: Path): Path =
            let Path(xs) = x;
            let Path(ys) = y;
            if (List.length(xs) <= List.length(ys)) x else y
    }

    instance MeetLattice[Path] {
        pub def greatestLowerBound(x: Path, y: Path): Path = ???
    }


    pub def init(y: Int32, x: Int32): Path =
        Path(y :: x :: Nil)

    pub def cons(z: Int32, p: Path): Path = match p {
        case Path(xs) => Path(z :: xs)
    }


    // Consider returning just Int32
    // Maybe generalize Constraint and Flow to typeclass so it can be both floats and ints
    type alias Constraint = Int32
    type alias Flow = Int32

    pub def maxFlow(g: m[(t, t, Constraint)]): Option[Flow] with Foldable[m], Order[t] = ???

    def nodesWeighted(): #{ Edge(t, t, Constraint), Node(t) | r } with Order[t] = #{
        Node(x) :- Edge(x, _, _).
        Node(x) :- Edge(_, x, _).
    }

    pub def paths(src: { src = t }, dst: { dst = t }, g: m[(t, t)]): List[t] with Foldable[m], Order[t] =
        let edges = inject g into Edge;
        let rules = #{
            Reach(x, y; init(y, x)) :- Edge(x, y).
            Reach(x, z; cons(z, p)) :- Reach(x, y; p), Edge(y, z).
        };
        let result = query edges, rules select fn from Reach(src, dst; fn);
        Vector.get(0, result)

    pub def fordFulkerson(): #{ Edge(t, t, Constraint), Node(t), Flow(Flow) | r } with Order[t] = ???

    pub def fordFulkersonNetwork(g: m[(t, t, Constraint)]): Set[(t, t, Flow)] with Foldable[m], Order[t] = ???

    pub def flowValue(g: m[(t, t)]): Int32 with Foldable[m], Order[t] = ???

    pub def source(g: m[(t, t)]): Option[t] with Foldable[m], Order[t] = ???

    pub def hasSource(g: m[(t, t)]): Bool with Foldable[m], Order[t] = ???

    pub def hasSink(g: m[(t, t)]): Bool with Foldable[m], Order[t] = ???

    pub def sink(g: m[(t, t)]): Option[t] with Foldable[m], Order[t] = ???

    pub def isSTNetwork(g: m[(t, t)]): Bool with Foldable[m], Order[t] = ???

}

```

jaschdoc:
@magnus-madsen maybe you should take a look at what I have so far. I have a feeling that the main loop `fordFulkerson` can be prettier by using datalog. Other than that I have a todo to handle the residual network properly.

main/src/library/Graph.flix
```diff
@@ -560,4 +560,153 @@ mod Graph {
         a
     }
 
+
```
magnus-madsen, line 563: A few comments to begin with.

main/src/library/Graph.flix
```diff
@@ -560,4 +560,153 @@ mod Graph {
...
+    //////////////////////////////////////
+    // WIP Ford-Fulkerson               //
```
magnus-madsen, line 565: I would like this in the examples Datalog folder.

main/src/library/Graph.flix
```diff
@@ -560,4 +560,153 @@ mod Graph {
...
+    //////////////////////////////////////
+
+    use Path.Path;
+
+    enum Path {
+        case Path(List[Int32])
```
magnus-madsen, line 571: Later we may want an explicit Bot element.

main/src/library/Graph.flix
```diff
@@ -560,4 +560,153 @@ mod Graph {
...
+
+    type alias From = Int32
+    type alias To = Int32
+    type alias UpperCapacity = Int32
+    type alias Flow = Int32
+    type alias FlowNetwork = (From, To, UpperCapacity, Flow)
+
+    // TODO: Handle residual network, i.e., reversed edges with negative flow and allow finding paths in that combined with normal network.
+
+    pub def maxFlow(src: { src = From }, dst: { dst = To }, g: m[(From, To, UpperCapacity)]): Flow with Foldable[m] =
```
magnus-madsen, line 636: This has a decent structure.

main/src/library/Graph.flix
```diff
@@ -560,4 +560,153 @@ mod Graph {
...
+        def fordFulkerson(flowNetwork) = {
+            let path = augmentingPath(src, dst, flowNetwork);
```
magnus-madsen, line 638: Perhaps it would be better for augmentingPath to return an Option.

main/src/library/Graph.flix
```diff
@@ -560,4 +560,153 @@ mod Graph {
...
+
+    def getMaxFlow(dst: To, g: m[FlowNetwork]): Flow with Foldable[m] =
+        let edges = inject g into Edge;
+        let rules = #{
+            Flow(f) :- Edge(_, dst, _, f).
+        };
+        let result = query edges, rules select f from Flow(f);
+        Vector.sum(result)
+
+    def augmentingPath(src: { src = From }, dst: { dst = To }, g: m[FlowNetwork]): Path with Foldable[m] =
```
magnus-madsen, line 658: This seems reasonable, except perhaps for returning Option.

main/src/library/Graph.flix
```diff
@@ -560,4 +560,153 @@ mod Graph {
...
+        let rules = #{
+            Reach(x, y; init(y, x)) :- Edge(x, y, u, f), if (u - f) > 0.
+            Reach(x, z; cons(z, p)) :- Reach(x, y; p), Edge(y, z, u, f), if (u - f) > 0.
+        };
+        let s = src.src;
+        let t = dst.dst;
+        let result = query edges, rules select fn from Reach(s, t; fn);
+        Vector.get(0, result)
+
+    def pathMinCapacity(p: Path, g: m[FlowNetwork]): Int32 with Foldable[m] =
```
magnus-madsen, line 669: This function I am not so happy with. I wonder if we can do better somehow. Either by different implementation (could be functional) or by extending the aug path computation.

(It is okay to be functional too, the goal is to show that a mix is useful).

main/src/library/Graph.flix
```diff
@@ -560,4 +560,153 @@ mod Graph {
...
+        let edges = inject g into Edge;
+        let rules = #{
+            PathCapacity(u - f)   :- Edge(x, y, u, f), Path(x), Path(y), if f <= u and f >= 0.
+            NonMinCapacity(u)     :- PathCapacity(u), PathCapacity(v), if v < u.
+            MinCapacity(u)        :- PathCapacity(u), not NonMinCapacity(u).
+        };
+        let result = query ps, edges, rules select u from MinCapacity(u);
+        Vector.get(0, result)
+
+    def increaseFlow(p: Path, incr: Int32, g: m[FlowNetwork]): Vector[FlowNetwork] with Foldable[m] =
```
magnus-madsen, line 681: Ok sure, but I think a map on g could also suffice? :)

main/src/library/Graph.flix
magnus-madsen: If we have the path and we have the current flow, can we not simple recurse through the list, looking up the flow and capacity to find the minimum?

main/src/library/Graph.flix
jaschdoc: Yeah I agree but checking that the edge is on the given path is easier in datalog

jaschdoc: But I guess it's just a call to filter first

magnus-madsen: memberOf

jaschdoc:
Apart from restructuring and bug fixing, what do you think of this? I'm not sure it terminates currently.

jaschdoc:
The algorithm works now. I will move it to an example file, but feedback on style etc. would be nice.

examples/larger-examples/datalog/ford-fulkerson.flix
```diff
@@ -0,0 +1,584 @@
...
+    println("Expected maximum flow 20, got ${g2}");
+    println("Expected maximum flow 2000, got ${g3}");
+    println("Expected maximum flow 20, got ${g4}")
+
+mod FordFulkerson {
+
+    use Path.Path;
+
+    enum Path with ToString {
+        case Path(List[Int32])
```
magnus-madsen: Lets add an explicit Bot element which corresponds to an infinitely long path.

examples/larger-examples/datalog/ford-fulkerson.flix
```diff
@@ -0,0 +1,584 @@
...
+            let Path(xs) = x;
+            let Path(ys) = y;
+            List.length(xs) == List.length(ys)
+    }
+
+    instance Order[Path] {
+        pub def compare(x: Path, y: Path): Comparison =
+            let Path(xs) = x;
+            let Path(ys) = y;
+            if (List.length(xs) ==  List.length(ys))
```
magnus-madsen: You can use the spaceship! <=>

examples/larger-examples/datalog/ford-fulkerson.flix
```diff
@@ -0,0 +1,584 @@
...
+
+    instance JoinLattice[Path] {
+        pub def leastUpperBound(x: Path, y: Path): Path =
+            let Path(xs) = x;
+            let Path(ys) = y;
+            if (List.length(xs) <= List.length(ys)) x else y
+    }
+
+    instance MeetLattice[Path] {
+        pub def greatestLowerBound(x: Path, y: Path): Path = ???
```
magnus-madsen: Lets implement it. Just the reverse of JoinLattice.

examples/larger-examples/datalog/ford-fulkerson.flix
```diff
@@ -0,0 +1,584 @@
...
+
+    pub def cons(z: Int32, p: Path): Path = match p {
+        case Path(xs) => Path(z :: xs)
+    }
+
+    type alias From = Int32
+    type alias To = Int32
+    type alias UpperCapacity = Int32
+    type alias Flow = Int32
+    type alias FlowNetwork = (From, To, UpperCapacity, Flow)
```
magnus-madsen: Not sure how I feel about some of these.

The purpose is to be an example, not a real implementation for the library.

magnus-madsen: Let me think about it.

jaschdoc: Okay, I think it significantly improves readability as everything is just `Int32` otherwise. :)

magnus-madsen: But could also be polymorphic. Ill think on it.

magnus-madsen: Keep the capacities as Int32 makes sense, but the nodes could be polymorphic.

magnus-madsen:
My comments from earlier are still relevant they are just not in the rightplace anymore.

magnus-madsen:
I just want to say this looks like a really nice example 👍 

I am very happy with the attention to detail and the graphs :)

magnus-madsen:
Looking forward to reviewing it in detail.

examples/larger-examples/datalog/ford-fulkerson.flix
```diff
@@ -0,0 +1,649 @@
...
+    pub enum Path[a] with ToString {
+        case Path(List[a])
+        case Bot // Infinitely long path
+    }
+
+    instance Eq[Path[a]] {
+        pub def eq(x: Path[a], y: Path[a]): Bool = match (x, y) {
+            case (Path(xs), Path(ys)) => List.length(xs) == List.length(ys)
+            case (Bot, _)             => false
+            case (_, Bot)             => false
```
magnus-madsen, line 132: Bot should be equal to bot.

By convention cases with Bot also goes first. But how could you know that :)

examples/larger-examples/datalog/ford-fulkerson.flix
```diff
@@ -0,0 +1,649 @@
...
+        }
+    }
+
+    instance Order[Path[a]] {
+        pub def compare(x: Path[a], y: Path[a]): Comparison = match (x, y) {
+            case (Path(xs), Path(ys)) => List.length(xs) <=> List.length(ys)
+            case (Bot, _)             => Comparison.GreaterThan
+            case (_, Bot)             => Comparison.GreaterThan
```
magnus-madsen, line 140: This should be smaller than right? It cannot be symmetric.

examples/larger-examples/datalog/ford-fulkerson.flix
```diff
@@ -0,0 +1,649 @@
...
+
+    instance LowerBound[Path[a]] {
+        // The longest list
+        pub def minValue(): Path[a] = Bot
+    }
+
+    instance PartialOrder[Path[a]] {
+        pub def lessEqual(x: Path[a], y: Path[a]): Bool = match (x, y) {
+            case (Path(xs), Path(ys)) => List.length(xs) >= List.length(ys)
+            case (Bot, _)             => false
```
magnus-madsen, line 152: Bot is inf, so its is smaller than anything. So this should be true.

examples/larger-examples/datalog/ford-fulkerson.flix
```diff
@@ -0,0 +1,649 @@
...
+            case (_, Bot)             => false
```
magnus-madsen, line 153: This case goes away.

examples/larger-examples/datalog/ford-fulkerson.flix
```diff
@@ -0,0 +1,649 @@
...
+        }
+    }
+
+    instance JoinLattice[Path[a]] {
+        pub def leastUpperBound(x: Path[a], y: Path[a]): Path[a] = match (x, y) {
+            case (Path(xs), Path(ys)) => if (List.length(xs) <= List.length(ys)) x else y
+            case (Bot, _)             => Bot
```
magnus-madsen, line 160: If either is bot then you take the other.

examples/larger-examples/datalog/ford-fulkerson.flix
```diff
@@ -0,0 +1,649 @@
...
+            case (_, Bot)             => Bot
+        }
+    }
+
+    instance MeetLattice[Path[a]] {
+        pub def greatestLowerBound(x: Path[a], y: Path[a]): Path[a] = match (x, y) {
+            case (Path(xs), Path(ys)) => if (List.length(xs) > List.length(ys)) x else y
+            case (Bot, _)             => Bot
+            case (_, Bot)             => Bot
```
magnus-madsen, line 169: This is correct in both cases (but again this should go first).

examples/larger-examples/datalog/ford-fulkerson.flix
```diff
@@ -0,0 +1,649 @@
...
+        case Path(xs) => List.indexOf(x, xs)
+        case Bot      => None
+    }
+
+    // Type aliases for readability
+    type alias Source = Int32
+    type alias Destination = Int32
+    type alias UpperCapacity = Int32
+    type alias Flow = Int32
+    type alias FlowNetwork[a] = (a, a, UpperCapacity, Flow)
```
magnus-madsen, line 210: Lets try to get rid of these. Some can be replaced by polymorphism others can just be Int32.

examples/larger-examples/datalog/ford-fulkerson.flix
```diff
@@ -0,0 +1,649 @@
...
+    /// No pre-assigned flow is allowed.
+    ///
+    /// The following assumptions also apply:
+    /// - `src` and `dst` is connected in `g`
+    /// - `g` contains no negative cycles
+    /// - `g` is labeled in ascending order from `src` to `sink`
+    /// - The label of `src` has the lowest value in the graph
+    /// - The label of `dst` has the highest value in the graph
+    ///
+    pub def maxFlow(src: { src = a }, dst: { dst = a }, g: m[(a, a, UpperCapacity)]): Flow with Foldable[m], Order[a] =
```
magnus-madsen, line 32: Lets drop the records.

magnus-madsen: I would also drop the type alias, and I would like edges to be `t, Int32, t` (sorry!)

examples/larger-examples/datalog/ford-fulkerson.flix
```diff
@@ -0,0 +1,649 @@
...
+                fordFulkerson(updatedNetwork)
+        };
+        // Init with 0 flow
+        fordFulkerson(zeroFlow(g))
+
+    ///
+    /// Returns a flow network with zero flow.
+    ///
+    def zeroFlow(g: m[(a, a, UpperCapacity)]): Vector[FlowNetwork[a]] with Foldable[m], Order[a] =
+        Foldable.toVector(g) |> Vector.map(match (x, y, z) -> (x, y, z, 0))
```
magnus-madsen, line 47: Neat

examples/larger-examples/datalog/ford-fulkerson.flix
```diff
@@ -0,0 +1,649 @@
...
+
+    ///
+    /// Returns the sum of the flows directly connected to `dst`.
+    ///
+    def getMaxFlow(dst: a, g: m[FlowNetwork[a]]): Flow with Foldable[m], Order[a] =
```
magnus-madsen, line 52: It is not exactly clear to me what this computes. I think its the sum of the flow on all ingoing edges to dst?

examples/larger-examples/datalog/ford-fulkerson.flix
```diff
@@ -0,0 +1,649 @@
...
+        let rules = #{
+            Reach(x, y; init(y, x)) :- Edge(x, y, u, f),                 if (u - f) > 0. // Forward edge
+            Reach(x, z; cons(z, p)) :- Reach(x, y; p), Edge(y, z, u, f), if (u - f) > 0. // Forward edge
+            Reach(x, y; init(y, x)) :- Edge(y, x, u, f),                 if f > 0.       // Back edge
+            Reach(x, z; cons(z, p)) :- Reach(x, y; p), Edge(z, y, u, f), if f > 0.       // Back edge
+        };
+        let s = src.src;
+        let t = dst.dst;
+        let result = query edges, rules select fn from Reach(s, t; fn);
+        Vector.head(result) |> Option.flatMap(toOption)
```
magnus-madsen, line 75: Ey!? :) What kind of sorcery is `Option.flatMap(toOption)`

magnus-madsen (commented):
I am still thinking about the forward / backward edges stuff.

examples/larger-examples/datalog/ford-fulkerson.flix
```diff
@@ -0,0 +1,649 @@
...
+    ///
+    /// Returns an augmenting path if one exists.
+    ///
+    /// An edge is in an augmenting path if its flow can be increased, i.e., the flow is strictly less than the capacity,
+    /// or if it has non-zero flow.
+    ///
+    def augmentingPath(src: { src = a }, dst: { dst = a }, g: m[FlowNetwork[a]]): Option[Path[a]] with Foldable[m], Order[a] =
+        let edges = inject g into Edge;
+        let rules = #{
+            Reach(x, y; init(y, x)) :- Edge(x, y, u, f),                 if (u - f) > 0. // Forward edge
```
magnus-madsen, line 67: Ok, maybe we should keep the arguments in the order they are. Hmm.

magnus-madsen: What is u and what is f?

jaschdoc: `U`pper bounds on capacities and `f`low

examples/larger-examples/datalog/ford-fulkerson.flix
jaschdoc: Yes

examples/larger-examples/datalog/ford-fulkerson.flix
jaschdoc: I think I was confused by the comment `case Bot // Infinitely long path`

examples/larger-examples/datalog/ford-fulkerson.flix
jaschdoc: Hmm okay I'm confused as to why infinity is smaller than any constant, but I'll accept it :)

examples/larger-examples/datalog/ford-fulkerson.flix
jaschdoc: How?

jaschdoc:
```scala
case (Bot, _)             => true
            case (_, Bot)             => true
            case (Path(xs), Path(ys)) => List.length(xs) >= List.length(ys)
```

jaschdoc:
It seems the changes to `Path` broke everything

examples/larger-examples/datalog/ford-fulkerson.flix
jaschdoc: Do you also want `Path[t]`?

examples/larger-examples/datalog/ford-fulkerson.flix
magnus-madsen: Yeees, but I think you can just drop it :)

examples/larger-examples/datalog/ford-fulkerson.flix
```diff
@@ -0,0 +1,642 @@
...
+    pub enum Path[a] with ToString {
+        case Path(List[a])
+        case Bot // Infinitely long path
+    }
+
+    instance Eq[Path[a]] {
+        pub def eq(x: Path[a], y: Path[a]): Bool = match (x, y) {
+            case (Bot, Bot)           => true
+            case (Bot, _)             => false
+            case (_, Bot)             => false
```
magnus-madsen, line 130: Drop the two false cases and just make a default case at the bottom which is false.

examples/larger-examples/datalog/ford-fulkerson.flix
```diff
@@ -0,0 +1,642 @@
...
+            case (Path(xs), Path(ys)) => List.length(xs) == List.length(ys)
+        }
+    }
+
+    instance Order[Path[a]] {
+        pub def compare(x: Path[a], y: Path[a]): Comparison = match (x, y) {
+            case (Bot, Bot)           => Comparison.EqualTo
+            case (Bot, _)             => Comparison.LessThan
+            case (_, Bot)             => Comparison.LessThan
```
magnus-madsen, line 139: Uh oh. Bot cannot be both smaller and greater than everything. I think we need:

```
            case (Bot, Bot)           => Comparison.EqualTo
            case (Bot, _)             => Comparison.LessThan
            case (_, Bot)             => Comparison.GreaterThan
```

examples/larger-examples/datalog/ford-fulkerson.flix
```diff
@@ -0,0 +1,642 @@
...
+
+    instance LowerBound[Path[a]] {
+        // The longest list
+        pub def minValue(): Path[a] = Bot
+    }
+
+    instance PartialOrder[Path[a]] {
+        pub def lessEqual(x: Path[a], y: Path[a]): Bool = match (x, y) {
+            case (Bot, _)             => true
+            case (_, Bot)             => true
```
magnus-madsen, line 152: Drop this.
Make a default case at the end which is false.

jaschdoc:
Updating tests, then we are ready (forgot to update from `(t, t, Int32, Int32)` to `(t, Int32, Int32, t)`).

magnus-madsen (approved):

magnus-madsen:
Thanks!

PR #7212, opened 2024-02-12
wip: Improve Verifier

JonathanStarup:


main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -229,32 +236,92 @@ object Verifier {
...
         case AtomicOp.Untag(sym) =>
           val List(t1) = ts
           check(expected = MonoType.Enum(sym.enumSym))(actual = t1, loc)
+          // We do not know the case terms so tpe cannot be verified.
+          tpe
+
+        case AtomicOp.ArrayLength =>
+          val List(t1) = ts
+          check(expected = MonoType.Int32)(actual = tpe, loc)
+          checkWith(t1, { case arr: MonoType.Array => arr }, "Array", loc)
```
JonathanStarup, line 251: @magnus-madsen Is this too much? 

Here we want t1 to be some array type but the element type is unconstrained here, so I can't construct an array to compare to. You could write
```
val MonoType.Array(_) = t1
```
But that wouldn't make a nice exception. That made me make this function. 

I also like it for other places - if t1 has to be `Array[t]` and t2 has to be `t` then the array part is really expected, but the `t` part is just that they have to agree. So its nice to split the destruction of the array and the checking of the element type

JonathanStarup: Just wanted to skip writing one for ref/lazy/arrow/array etc. but i can do that

magnus-madsen: How many? Its like 7 no?

JonathanStarup: Its also a bit weird to write `getArrayElement` and ignore the result

magnus-madsen: I see what you mean.

magnus-madsen: assertArrayTypeAndGetElm

magnus-madsen: IDK: Just not strings please :)

magnus-madsen:
@JonathanStarup Remember: The verifier does not have to be perfect. It is just intended for us (and students) to quick catch stupid errors.

JonathanStarup:
> @JonathanStarup Remember: The verifier does not have to be perfect. It is just intended for us (and students) to quick catch stupid errors.

But you also want to make its errors hard crashes right?

magnus-madsen:
> > @JonathanStarup Remember: The verifier does not have to be perfect. It is just intended for us (and students) to quick catch stupid errors.
> 
> But you also want to make its errors hard crashes right?

If it is a bug then for sure a crash.

magnus-madsen:
But of course bug should mean "definitely wrong", not "uh oh we have not finished some refactor yet".

PR #7216, opened 2024-02-14
feat: add `Coerce` type class

### LINKED ISSUE #7213: Add `Coerce` type class
```
trait Coerce[t: Type -> Type] {
    pub def coerce(x: t[a]): a
}
```

magnus-madsen:
Fixes #7213


magnus-madsen:
@mlutze @JonathanStarup If either of you can approve. I need it for a paper. In the future this can be improved by using an associated type, but I did not want to rely on that for now.

This reason this is neat is because it allows you to unwrap newtype-style enums. (No automatic derivation yet).

So you can write:

```
query db, pr select (x, coerce(d)) from Dist(x; d)
```

where `d` has type `Down[t]`.

mlutze (approved):

PR #7217, opened 2024-02-14
feat: improve ShowAst

JonathanStarup:
Fixes https://github.com/flix/vscode-flix/issues/315
* ShowAstProvider now choses the phases with the closest name to the given string
* The compiler now only links strings to phases in one place

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/dbg/AstPrinter.scala
+++ b/main/src/ca/uwaterloo/flix/language/dbg/AstPrinter.scala
@@ -30,91 +30,72 @@ object AstPrinter {
     * Writes all the formatted asts, requested by the flix options, to disk.
     */
   def printAsts()(implicit flix: Flix): Unit = {
-    val asts = flix.options.xprintphase
-    if (asts.isEmpty)
-      ()
-    else if (asts.contains("all") || asts.contains("All"))
-      printAllAsts()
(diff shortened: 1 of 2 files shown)


main/src/ca/uwaterloo/flix/language/dbg/AstPrinter.scala
```diff
@@ -30,91 +30,72 @@ object AstPrinter {
...
+    for ((phase, printer) <- phaseMap)
+      writeToDisk(phase, printer())
+  }
+
+  /**
+    * Returns a list of map the phases of flix along with a thunked pretty printed AST.
+    *
+    * Returns only the phases that can be pretty printed if `includeUnfinished` is false.
+    */
+  def allPhases(includeUnfinished: Boolean = true)(implicit flix: Flix): Map[String, () => String] = {
```
JonathanStarup, line 60: This bool flag  is not the prettiest, but when crashing or printing all asts, we don't want a bunch of "wip" files, but we do want them here when ShowAstProvider is asking us to print "Parser" (so we don't give "Lowering" or whatever the closest string match is)

magnus-madsen (approved):

magnus-madsen:
I will merge and deploy the updated VSCode extension later.

Issue #7214, opened 2024-02-13
Tab key crashes the repl


stetimi:
This is on 98beb9659 (0.44+)
Enter the repl, type Str, hit tab.

flix> Str                                                                       
java.lang.StringIndexOutOfBoundsException: Range [0, -1) out of bounds for length 0
	at java.base/jdk.internal.util.Preconditions$1.apply(Preconditions.java:55)
	at java.base/jdk.internal.util.Preconditions$1.apply(Preconditions.java:52)
	at java.base/jdk.internal.util.Preconditions$4.apply(Preconditions.java:213)
	at java.base/jdk.internal.util.Preconditions$4.apply(Preconditions.java:210)
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:98)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckFromToIndex(Preconditions.java:112)
	at java.base/jdk.internal.util.Preconditions.checkFromToIndex(Preconditions.java:349)
	at java.base/java.lang.String.checkBoundsBeginEnd(String.java:4865)
	at java.base/java.lang.String.substring(String.java:2834)
	at org.jline.reader.impl.CompletionMatcherImpl.defaultMatchers(CompletionMatcherImpl.java:107)
	at org.jline.reader.impl.CompletionMatcherImpl.compile(CompletionMatcherImpl.java:47)
	at org.jline.reader.impl.LineReaderImpl.doComplete(LineReaderImpl.java:4497)
	at org.jline.reader.impl.LineReaderImpl.doComplete(LineReaderImpl.java:4431)
	at org.jline.reader.impl.LineReaderImpl.expandOrComplete(LineReaderImpl.java:4370)
	at org.jline.reader.impl.LineReaderImpl$1.apply(LineReaderImpl.java:3856)
	at org.jline.reader.impl.LineReaderImpl.readLine(LineReaderImpl.java:721)
	at org.jline.reader.impl.LineReaderImpl.readLine(LineReaderImpl.java:512)
	at ca.uwaterloo.flix.runtime.shell.Shell.loop(Shell.scala:107)
	at ca.uwaterloo.flix.Main$.main(Main.scala:261)
	at ca.uwaterloo.flix.Main.main(Main.scala)

magnus-madsen:
This happens inside the library or? The Flix code on line 107 looks reasonable.

stetimi:
Yes, from the stack trace it looks like what you say. It might be how flix
[...]

PR #7223, opened 2024-02-15
fix: update repl's Parsed class so tab won't crash repl (#7214)

stetimi:
Parsed is now more conformant with what JLine expects.
Tested \\\\ for multi-line statements still works as before.


magnus-madsen (approved):

magnus-madsen:
Thanks @stetimi

PR #7225, opened 2024-02-16
chore: Make Std. lib.  inline doc-omments into line comments.

herluf-ba:
Some comments in the standard library that should be line comments rather than doc comments.


herluf-ba:
@magnus-madsen Prep for new parser

mlutze (approved):

magnus-madsen (approved):

Issue #7163, opened 2024-01-30
List of reserved keywords used in the standard library


magnus-madsen:
TBD

magnus-madsen:
@herluf-ba If you can generate a list of defs and local variables whose name should change, I can ask one of the student programmers to fix it.

herluf-ba:
@magnus-madsen 
```
word     file
static   Prelude.flix:4:16
static   Concurrent/Channel.flix:84:21
static   Concurrent/Channel.flix:126:20
static   Concurrent/Channel.flix:127:45
mod      BigInt.flix:45:13
not      BigInt.flix:145:9
new      Fixpoint/Compiler.flix:91:17
new      Fixpoint/Compiler.flix:98:17
query    Fixpoint/Interpreter.flix:75:55
query    Fixpoint/Interpreter.flix:119:15
query    Fixpoint/Ram/RelOp.flix:35:25
debug    Prelude.flix:131:5
for      Traversable.flix:58:13
mod      Int32.flix:66:13
mod      Int64.flix:66:13
open     File.flix:74:13
mod      Int8.flix:66:13
not      Bool.flix:55:13
and      Bool.flix:60:13
or       Bool.flix:65:13
mod      Int16.flix:66:13
solve    Fixpoint/Solver.flix:35:13
project  Fixpoint/Solver.flix:95:13
project  Fixpoint/Solver.flix:109:36
project  Fixpoint/Solver.flix:109:52
query    Fixpoint/Solver.flix:720:17
choose   Random.flix:28:13
query    Fixpoint/Simplifier.flix:116:42
query    Fixpoint/Simplifier.flix:120:58
query    Fixpoint/Simplifier.flix:123:60
++       SemiGroup.flix:40:9
query    Fixpoint/IndexSelection.flix:116:21
query    Fixpoint/IndexSelection.flix:122:59
query    Fixpoint/IndexSelection.flix:125:59
```

magnus-madsen:
@jaschdoc I would like you to get on this quick. It should not take too long.

We are trying to ensure that keywords are not used as names for functions (or local variables).

In the future, the Parser will support writing e.g. `List.forAll(and, ...)`. So its OK to remove `Bool.and` for now.

Only, my suggestions should not require huge changes. If they do, let me know.

With that said, here is my proposal for iteration 1:

- [ ] `Map.query` becomes `Map.rangeQuery`.
- [ ] `Map.queryWith` becomes `Map.rangeQueryWith`.
[...]

PR #7242, opened 2024-02-19
refactor: remove `IntX.{mod, rem}`

jaschdoc:
Related to https://github.com/flix/flix/issues/7163

### DIFF
--- a/main/src/library/BigInt.flix
+++ b/main/src/library/BigInt.flix
@@ -39,30 +39,6 @@ mod BigInt {
         pow(x, n)
 
     ///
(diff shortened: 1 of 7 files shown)


JonathanStarup:
If the intention is too keep them (which I root for) couldn't they just be renamed `modulo` and `remainder`

magnus-madsen:
> If the intention is too keep them (which I root for) couldn't they just be renamed `modulo` and `remainder`

I think so.

magnus-madsen:
@jaschdoc I think we are going to rename them both: `remainder` and `modulo`.

Issue #7218, opened 2024-02-14
Port Effekt Benchmarks


magnus-madsen:
- https://github.com/effekt-lang/effekt/tree/master/examples/benchmarks
- https://github.com/effect-handlers/effect-handlers-bench/blob/main/benchmarks/effekt/

magnus-madsen:
(I am pausing this until set effects arrive).

PR #7246, opened 2024-02-19
docs: add `Range` example from Effekt

jaschdoc:
I don't know if this is too simple, but maybe it serves as a good introduction.

Related to https://github.com/flix/flix/issues/7093
Related to https://github.com/flix/flix/issues/7218


jaschdoc:
This one is from benchmarks https://github.com/effect-handlers/effect-handlers-bench/blob/main/benchmarks/effekt/iterator/main.effekt

examples/effects/Range.flix
```diff
@@ -0,0 +1,64 @@
...
+/// sums that range.
+///
+/// There are three things to pay attention to:
+/// - How to use effects to signal the caller
+/// - Handling the effect at the call site
+/// - Using the continuation to transfer control back to the callee
+///
+/// Inspired by the Iterator benchmark from Effekt Lang
+///
+pub def main(): Unit \ IO =
```
JonathanStarup, line 15: imo we should have `{ }` around function bodies longer than one line

examples/effects/Range.flix
```diff
@@ -0,0 +1,64 @@
...
+    pub def sumTo(n: Int32): Int32 = region rc {
+        let s = Ref.new(rc, 0);
+        try {
+            // Call an effectful function.
+            // Note that `sumTo` does not
+            // inherit the `Emission` effect type
+            // that `range` has, because we
+            // handle the effect below.
+            range(0, n);
+            Ref.get(s)
```
JonathanStarup, line 52: Do we prefer `Ref.{get, put}` instead of `!` and `:=`?

magnus-madsen: We prefer get and put because we are dropping those symbols later.

JonathanStarup:
Its a bit of a weird use of effects but I think its actually a nice example to look at to learn about effect semantics and how it interacts with mutable state

magnus-madsen:
I was going to suggest we take another look at these examples once Jonathans PR lands.

magnus-madsen:
@jaschdoc Test failures

jaschdoc:
Yeah just wanted to run tests to check what broke :)

magnus-madsen:
> Yeah just wanted to run tests to check what broke :)

Probably removal of `do` everywhere is at least one thing.

magnus-madsen (approved):

PR #7254, opened 2024-02-19
docs: add effect NonDeterminism example from Eff lang

jaschdoc:
- [ ] Add documentation

Depends on https://github.com/flix/flix/issues/7277
Related https://github.com/flix/flix/issues/7093


jaschdoc:
@JonathanStarup it is this example which requires shallow handlers.

The `backtrack` function adds a handler for `Failure` within the handler for `Decision`. However, `backtrack` does not remove the `Failure` effect so we still need a handler at the call site. However, this top-level handler will now also handle `Failure` within the handler for `Decision` in `backtrack`. Is this correct?
The test `testEightQueens01` demonstrates this problem.

magnus-madsen:
I suppose we cannot do it then.

PR #7257, opened 2024-02-20
feat: Parser2Error

herluf-ba:
Adds an error type for `Parser2`.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/errors/Parser2Error.scala
+++ b/main/src/ca/uwaterloo/flix/language/errors/Parser2Error.scala
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2023 Herluf Baggesen
+ *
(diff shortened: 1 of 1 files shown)


magnus-madsen:
I would prefer not to merge in the DevError. We want to keep master clean.

main/src/ca/uwaterloo/flix/language/errors/Parser2Error.scala
```diff
@@ -0,0 +1,72 @@
+/*
+ * Copyright 2023 Herluf Baggesen
```
magnus-madsen, line 2: 2024

herluf-ba: How time flies!

magnus-madsen (approved):

PR #7259, opened 2024-02-20
feat: Add `SyntaxTree` type

herluf-ba:
Adds `SyntaxTree` (formerly `UntructuredTree`) which is the data type that the new parser targets.


herluf-ba:
@magnus-madsen This is ready for review too. 
@mlutze Can you take a look too? 🙂

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
```diff
@@ -0,0 +1,504 @@
...
+   * @param kind     The kind of the node.
+   * @param loc      The location that the node spans in the source file.
+   * @param children The children of the node.
+   */
+  case class Tree(kind: TreeKind, var loc: SourceLocation, var children: Array[Child]) {
+    /**
+     * Utility function that computes a textual representation of a [[SyntaxTree.Tree]].
+     * Meant for debugging use.
+     */
+    def toDebugString(nesting: Int = 1): String = {
```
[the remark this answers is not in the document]
magnus-madsen: Perhaps it would be better to move this _functionality_ out of the definition of the tree and into the phase that uses it. Thats typically what we do.

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
```diff
@@ -0,0 +1,504 @@
...
+
+    case object Parameters extends TreeKind
+
+    case object QName extends TreeKind
+
+    case object Source extends TreeKind
+
+    case object TypeParameter extends TreeKind
+
+    case object TypeParameters extends TreeKind
```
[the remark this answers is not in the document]
magnus-madsen: @herluf-ba I think we have come to the decision that we prefer "TypeParameterList" over "TypeParams". 

The same would go for AnnotationList etc.

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
```diff
@@ -0,0 +1,504 @@
...
+
+  /**
+   * A child in a [[SyntaxTree]].
+   * It holds a [[Token]] or another [[SyntaxTree.Tree]].
+   */
+  object Child {
+    /**
+     * A [[SyntaxTree]] child holding a [[Token]].
+     */
+    case class Token(token: ca.uwaterloo.flix.language.ast.Token) extends Child
```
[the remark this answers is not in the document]
magnus-madsen: Or do both renamings.

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
```diff
@@ -0,0 +1,504 @@
...
+     */
+    case class ErrorTree(error: Parser2Error) extends TreeKind
+
+    case object Annotations extends TreeKind
+
+    case object Argument extends TreeKind
+
+    case object ArgumentNamed extends TreeKind
+
+    case object Arguments extends TreeKind
```
magnus-madsen, line 97: Is this an expression list or what is it?

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
```diff
@@ -0,0 +1,504 @@
...
+
+    case object Case extends TreeKind
+
+    case object Comments extends TreeKind
```
magnus-madsen, line 101: Plural vs singular? How to understand? Is it more than one comment.

magnus-madsen: Ah, maybe its LineComments?

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
```diff
@@ -0,0 +1,504 @@
...
+
+      case object CheckedTypeCast extends Expr
+
+      case object Do extends Expr
+
+      case object FixpointConstraint extends Expr
+
+      case object FixpointConstraintSet extends Expr
+
+      case object FixpointFrom extends Expr
```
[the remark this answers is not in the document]
magnus-madsen: I notice that for a few other constructs. I have no problem with it, I am just asking how it works.

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
```diff
@@ -0,0 +1,504 @@
...
+
+      case object IfThenElse extends Expr
+
+      case object InstanceOf extends Expr
+
+      case object Intrinsic extends Expr
+
+      case object JvmMethod extends Expr
+
+      case object KeyValue extends Expr
```
magnus-madsen, line 232: This is for Map literals?

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
```diff
@@ -0,0 +1,504 @@
...
+
+      case object TryWithBody extends Expr
+
+      case object TryWithRule extends Expr
+
+      case object Tuple extends Expr
+
+      case object TypeMatch extends Expr
+
+      case object TypeMatchRule extends Expr
```
magnus-madsen, line 314: Is this an expression? Ditto for MatchRule

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
```diff
@@ -0,0 +1,504 @@
...
+
+      case object Constraint extends Type
+
+      case object Constraints extends Type
+
+      case object Derivations extends Type
+
+      case object EffectSet extends Type
+
+      case object Function extends Type
```
magnus-madsen, line 368: Usually called Arrow, but perhaps this is one case where we should keep your name 👍

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
```diff
@@ -0,0 +1,504 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ca.uwaterloo.flix.language.ast
+
+import ca.uwaterloo.flix.language.errors.Parser2Error
+
+
+/**
+ * A Syntax tree representing the __syntax__ in a source file.
```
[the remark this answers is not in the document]
herluf-ba: Actually I think it should be "textual content of a source file" 🤔

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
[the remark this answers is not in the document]
herluf-ba: I will move it into the parser then 👍
I is quite handy for debugging, so I was not planning on removing it. If some day someone else is going to work on the parser, I think they would need to write something similar.

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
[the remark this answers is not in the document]
herluf-ba: I am actually more inclined to renaming both cases of `Child` to `Child.TreeNode` and `Child.TokenNode`. That way, there is no clashes and we get to keep the nice and short `Token`.

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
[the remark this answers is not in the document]
herluf-ba: The reasoning is that it is convenient to have a node in the syntax tree for each `FixpointFrom` or similar, to be able to `pickAll(TreeKind.Expr.FixpointFrom, ...` in the weeder and do further transformations/validation from there.

I can see why it might be a little misleading to extend `Expr` directly. The catch is that `Expr` really is just "a group of tree kinds that relate to expressions" rather than "the group of tree kinds that make up valid expressions".

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
herluf-ba: In an application `foo(a, b)` the kind `Arguments` is used to group all the arguments passed.
```
Call
  Ident 
    'foo'
  Arguments
    Argument
      Expr
        Ident
          'a'
    Argument
      Expr
        Ident
          'b'
```

So yes, this is a list of expressions 👍

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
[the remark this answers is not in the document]
magnus-madsen: No its fine. Expr is expressions and things related to expressions.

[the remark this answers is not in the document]
magnus-madsen: We can leave it or have ExprFragments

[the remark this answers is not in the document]
magnus-madsen: Or rename the fragments to FixpointFromFragment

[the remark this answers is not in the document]
magnus-madsen: Thoughts?

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
herluf-ba: So it's both line and block comments. The intention is to consume as many subsequent comments into the same tree node of kind `Comments`.

Note that it would also be completely valid to just have a single tree node per comment. I think it wouldn't change much really, since the Weeder just ignores comments.

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
herluf-ba: Exacly. I intentionally named it very generally because I though I would be able to reuse it, but that didn't turn out.

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
herluf-ba: I think `Function` makes sense here. It better communicates what the construct is in my opinion.

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
herluf-ba: I renamed it `LiteralMapKeyValueFragment`

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
magnus-madsen: Yeah, its just that in 500 other places we call it arrow.

magnus-madsen:
@herluf-ba Let me know when I should do another review.

herluf-ba:
@magnus-madsen I think this is ready for another round 👍

magnus-madsen:
@herluf-ba I have some changes, but I also got a headache :(

I will try to get to it soon.

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
```diff
@@ -0,0 +1,489 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ca.uwaterloo.flix.language.ast
+
+import ca.uwaterloo.flix.language.errors.Parser2Error
+
+/**
+ * A Concrete syntax tree representing the textual content of a source file.
```
magnus-madsen, line 21: ->

```

A [[SyntaxTree]] is unstructured: it allows much more flexibility than later
abstract syntax trees. This flexibility is used to capture source code that may
contain faulty syntax. The tree has nodes that hold a [[TreeKind]] and zero or
more children. Each child is either a [[TokenChild]] or a [[TreeChild]]. 

Note that [[SyntaxTree]] offers few guarantees. In particular:

    - There is no guarantee that a specific node is present or absent as a child.
    - There is no guarantee that a specific node has a specific number of children.
```

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
```diff
@@ -0,0 +1,489 @@
...
+   * @param loc      The location that the node spans in the source file.
+   * @param children The children of the node.
+   */
+  case class Tree(kind: TreeKind, var loc: SourceLocation, var children: Array[Child])
+
+  sealed trait Child
+
+  /**
+   * A child in a [[SyntaxTree]].
+   * It holds a [[Token]] or another [[SyntaxTree.Tree]].
```
magnus-madsen, line 46: Drop this sentence.

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
```diff
@@ -0,0 +1,489 @@
...
+object SyntaxTree {
+
+  /**
+   * A node in a [[SyntaxTree]]
+   *
+   * @param kind     The kind of the node.
+   * @param loc      The location that the node spans in the source file.
+   * @param children The children of the node.
+   */
+  case class Tree(kind: TreeKind, var loc: SourceLocation, var children: Array[Child])
```
magnus-madsen, line 40: Does var and children need to be mutable? (`var`)?

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
```diff
@@ -0,0 +1,489 @@
...
+
+  /**
+   * A child in a [[SyntaxTree]].
+   * It holds a [[Token]] or another [[SyntaxTree.Tree]].
+   */
+  object Child {
+    /**
+     * A [[SyntaxTree]] child holding a [[Token]].
+     */
+    case class NodeToken(token: Token) extends Child
```
magnus-madsen, line 52: -> TokenChild

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
```diff
@@ -0,0 +1,489 @@
...
+
+    /**
+     * A [[SyntaxTree]] child holding a nested [[SyntaxTree.Tree]]
+     */
+    case class NodeTree(tree: SyntaxTree.Tree) extends Child
```
magnus-madsen, line 57: -> TreeChild

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
```diff
@@ -0,0 +1,489 @@
...
+  }
+
+
+  /**
+   * A common super-type for [[TreeKind]]s
+   */
+  sealed trait TreeKind
+
+  /**
+   * Different kinds of syntax nodes in a [[SyntaxTree]].
```
magnus-madsen, line 67: line break between the two sentences.

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
```diff
@@ -0,0 +1,489 @@
...
+   * The only error kind that holds data is the special [[TreeKind.ErrorTree]].
+   */
+  object TreeKind {
+    /**
+     * A special error kind wrapping a [[Parser2Error]].
+     */
+    case class ErrorTree(error: Parser2Error) extends TreeKind
+
+    case object AnnotationList extends TreeKind
```
magnus-madsen, line 76: Intuitively, why is there no Annotation kind? Is it because it is just a token?

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
```diff
@@ -0,0 +1,489 @@
...
+
+    case object Operator extends TreeKind
+
+    case object Parameter extends TreeKind
+
+    case object ParameterList extends TreeKind
+
+    case object QName extends TreeKind
+
+    case object Source extends TreeKind
```
magnus-madsen, line 104: What is this?

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
```diff
@@ -0,0 +1,489 @@
...
+      /**
+       * A marker kind used to wrap nested expressions.
+       * For instance on a binary expression "1 + 2" you would do
+       * Expr
+       * Binary
+       * Expr
+       * LiteralNumber
+       * Operator
+       * Expr
+       * LiteralNumber
```
magnus-madsen, line 159: Formatting lost? :)

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
```diff
@@ -0,0 +1,489 @@
...
+    object Type {
+      /**
+       * A marker kind used to wrap nested types.
+       * For instance on a tuple type "(Int32, Bool)" you would do
+       * Type
+       * Tuple
+       * Type
+       * Ident
+       * Type
+       * Ident
```
magnus-madsen, line 327: formatting lost?

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
```diff
@@ -0,0 +1,489 @@
...
+      /**
+       * A marker kind used to wrap nested patterns.
+       * For instance on cons pattern "0 :: xs" you would do
+       * Pattern
+       * FCons
+       * Pattern
+       * Literal
+       * Pattern
+       * Ident
+       */
```
magnus-madsen, line 396: formatting lost?

magnus-madsen (commented):
I added more comments. Sorry for being so pedantic, but data type definitions are the most important :)

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
magnus-madsen: (I know reordering is probably a bit refactor, but consistency is key)

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
```diff
@@ -0,0 +1,489 @@
...
+
+      case object RecordOpUpdate extends Expr
+
+      case object RecordSelect extends Expr
+
+      case object Ref extends Expr
+
+      case object Scope extends Expr
+
+      case object ScopeName extends Expr
```
magnus-madsen, line 275: What is this?

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
```diff
@@ -0,0 +1,489 @@
...
+
+      case object CaseSet extends Type
+
+      case object Constant extends Type
+
+      case object Constraint extends Type
+
+      case object ConstraintList extends Type
+
+      case object DerivationList extends Type
```
magnus-madsen, line 349: I am not sure if this belongs under type. It is really only for enums, right? So maybe it goes at the top in the unsorted bucket?

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
magnus-madsen: I agree it should have been called Function, but for consistency, let us rename it to Arrow. Thanks!

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
herluf-ba: Yep, the `TokenKind` is enough to recognize an annotation. If there is a child with a kind that is not an annotation, that's a parser error!

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
herluf-ba: dang

herluf-ba: Might have to loose those comments, I wasn't too sure about them either

magnus-madsen: I like the comments, when they had formatting. I  think if you use // the formatter will not destroy them.

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
herluf-ba: Reordering so `loc` is last 👍 I think children and loc became `var`s since it was very convenient when building the tree. I will try to make the `val` s and just see if I get into major trouble when merging back with my parser branch.

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
herluf-ba: Yea that's only for enums (and restrictable enums). I will move it to the top-most namespace

main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
herluf-ba: That refers to the "scope" or region part following some expressions
```
ref e @ rc
        ^^ This part here
```
I use it to mark what expression is the scope name (so the weeded doesn't confuse  `e` with `rc`)

herluf-ba:
@magnus-madsen I went over your comments again. Sorry for the long turn-around! we have been taking turns being sick at home 🕺

magnus-madsen (approved):

PR #7264, opened 2024-02-20
feat: introduce reference equality

### LINKED ISSUE #6330: Reference equality
Could likely be used as a base case to speed up a lot of equality implementations

mlutze:
fixes #6330. Not sure if Reflect is the right place for it.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/SemanticOp.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/SemanticOp.scala
@@ -661,4 +661,17 @@ object SemanticOp {
 
   }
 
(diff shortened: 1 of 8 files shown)


mlutze:
maybe it should be in `Object` instead?

magnus-madsen:
I am not even sure I want this feature. Does it not raise all kinds of problems? In particular, it seems it would fix our bytecode-level representation of objects?

mlutze:
It's something that has to be used carefully. But I think it is at the same level of danger as the `default` function

magnus-madsen:
I am closing this until we have had a longer discussion.

PR #6499, opened 2023-09-26
feat: introduce constraint reduction

mlutze:

PR #7275, opened 2024-02-23
feat: add shortcut in Boolean unification

mlutze:
extracted from #6499


magnus-madsen (approved):

magnus-madsen:
Test failures? Uh oh... but why

mlutze:
They all have to do with effect handlers, it seems.

PR #7276, opened 2024-02-23
chore: relax TypeError requirements in TestTyper

mlutze:
extracted from #6499


mlutze:
> @mlutze Do we have a ticket to restore this functionality?

#7278

Issue #6989, opened 2024-01-08
Improvements to `Chain`


magnus-madsen:
We go again:

- [x] `Chain` should not extend `Iterable` instead we should implement the required methods directly (e.g. `map`).
- [x] Drop `concat` and `iterator` methods.
- [x] Drop `Many` case class.
- [ ] `Empty`, `Link`, and `Proxy` should not be case classes. Just normal classes with vals. They should still be private.
- [ ] Why do we have both `apply` and `from` methods. Can we settle on one? I am OK with from.
- [x] Implement `foreach`.
- [ ] Rewrite `toList` to use a buffer and call `foreach`.

magnus-madsen:
@jaschdoc Lets do this next-- together with the open PRs.

magnus-madsen:
> IMO we should have both `apply` and `from`, but not two of each.

Why?

magnus-madsen:
> They serve different purposes. `apply` is for a chain "literal", and `from` is for conversion from another Iterable.

Argument accepted.

jaschdoc:
- [ ] override `toString`

jaschdoc:
> `Empty`, `Link`, and `Proxy` should not be case classes. Just normal classes with vals. They should still be private

Why? We can't do pattern matching in the trait then :/ I guess that forces us to write things in object oriented style?

PR #7282, opened 2024-02-26
feat: add improvements to `Chain`

jaschdoc:
Closes https://github.com/flix/flix/issues/6989


main/src/ca/uwaterloo/flix/util/collection/Chain.scala
```diff
@@ -117,23 +113,19 @@ sealed trait Chain[+A] {
...
   /**
     * Displays all elements of this collection in a string using a separator string.
     */
-  final def mkString(sep: String): String = this match {
-    case Chain.Empty => ""
-    case Chain.Link(l, r) => l.mkString(sep) ++ sep ++ r.mkString(sep)
-    case Chain.Proxy(xs) => xs.mkString(sep)
+  final def mkString(sep: String): String = {
+    val sb = new StringBuilder()
+    this.toList.addString(sb, sep)
```
magnus-madsen, line 127: This looks strange to me. I would have expected a for loop. What is addString on a list? Is that some kind of implicit coercion? I would avoid that.

magnus-madsen:
I think the ticket also has some comments about making case classes into ordinary classes.

main/src/ca/uwaterloo/flix/util/collection/Chain.scala
jaschdoc: I looked at the docs and just converts each element in the list to a stringbuilder using `sep` to separate the strings. Nothing fancy honestly.
https://www.scala-lang.org/api/current/scala/collection/immutable/List.html#addString-ff4

jaschdoc:
> I think the ticket also has some comments about making case classes into ordinary classes.

The only thing you wrote is

> `Empty`, `Link`, and `Proxy` should not be case classes. Just normal classes with vals. They should still be private.

I tried that but naturally that ruins pattern matching. I can add unapply methods?

magnus-madsen:
Does it ruin pattern matching if you match on the type and make the fields public with val?

main/src/ca/uwaterloo/flix/util/collection/Chain.scala
mlutze: What a weird function name

main/src/ca/uwaterloo/flix/util/collection/Chain.scala
```diff
@@ -100,9 +92,13 @@ sealed trait Chain[+A] {
     * Applies `f` this every element in `this`.
     */
   final def foreach(f: A => Unit): Unit = this match {
-    case Chain.Empty => ()
-    case Chain.Link(l, r) => l.foreach(f); r.foreach(f)
-    case Chain.Proxy(xs) => xs.foreach(f)
+    // N.B.: We have to use reflection to avoid
```
mlutze, line 95: Does this quality as reflection?

magnus-madsen: No, its not reflection. Just instanceof.

magnus-madsen: - Update comment (actually just remove, this is standard scala code)

main/src/ca/uwaterloo/flix/util/collection/Chain.scala
magnus-madsen: lets not use this function, its very confusing

main/src/ca/uwaterloo/flix/util/collection/Chain.scala
jaschdoc: Why not update it to
```
    // N.B.: We must use type annotations
    // infinite recursion when pattern matching
    // since it calls the equals method which
    // depends on toList.
```

magnus-madsen: We must "match on the type" -
same text

magnus-madsen (approved):

Issue #7288, opened 2024-02-26
Add Datalog shortest path program for arbitrary path lengths


jaschdoc:
> > > Once this is done, lets do one with paths where the edges have distances. Then you need paths to be `Path(List[a], Int32)`.
> > 
> > 
> > Should the shortest Dist be updated with this new `Path` definition or what is your point exactly? :)
> 
> A new program that computes shortests paths, but where each path does not have length 1, but some arbitrary length stored on the path.

From https://github.com/flix/flix/pull/7236

jaschdoc:
@magnus-madsen can you assign this to me, please? :)

PR #7289, opened 2024-02-26
feat: add datalog-based single source shortest path for arbitrary length

jaschdoc:
Closes https://github.com/flix/flix/issues/7288


magnus-madsen (approved):

magnus-madsen:
Thanks!

PR #7290, opened 2024-02-26
refactor: make `Validation.SoftFailure` private

jaschdoc:


magnus-madsen:
Conflict, sorry!

main/src/ca/uwaterloo/flix/language/phase/Regions.scala
```diff
@@ -35,13 +35,16 @@ import scala.collection.immutable.SortedSet
 object Regions {
 
   def run(root: Root)(implicit flix: Flix): Validation[Unit, CompilationMessage] = flix.phase("Regions") {
-    val errs = ParOps.parMap(root.defs)(kv => visitDef(kv._2)).flatten
+    val (recoverable, rest) = ParOps.parMap(root.defs)(kv => visitDef(kv._2)).flatten.partition(_.isInstanceOf[Recoverable])
+    val (unrecoverable, errs) = rest.partition(_.isInstanceOf[Unrecoverable])
```
magnus-madsen, line 39: Regions phase can ONLY return RegionVarEscapes which can be soft failure. So I think all this code can be simplified.

main/src/ca/uwaterloo/flix/language/phase/Regions.scala
```diff
@@ -35,14 +35,11 @@ import scala.collection.immutable.SortedSet
 object Regions {
 
   def run(root: Root)(implicit flix: Flix): Validation[Unit, CompilationMessage] = flix.phase("Regions") {
-    val errs = ParOps.parMap(root.defs)(kv => visitDef(kv._2)).flatten
+    val recoverable = ParOps.parMap(root.defs)(kv => visitDef(kv._2)).flatten.filter(_.isInstanceOf[Recoverable])
```
magnus-madsen, line 38: If you change the return type of visitDef to the specific error, I think everything works out (?) (i.e. without any instance of casts)?

main/src/ca/uwaterloo/flix/language/phase/Regions.scala
```diff
@@ -35,14 +35,11 @@ import scala.collection.immutable.SortedSet
...
 
     // TODO: Instances
-
-    errs match {
-      case Nil => Validation.success(())
-      case es => Validation.SoftFailure((), Chain.from(es))
-    }
+    Validation.toSuccessOrSoftFailure((), recoverable.asInstanceOf[Iterable[Recoverable]])
```
magnus-madsen, line 41: I also think this can work out.

main/src/ca/uwaterloo/flix/language/phase/Regions.scala
jaschdoc: Nope still requires type casts

main/src/ca/uwaterloo/flix/language/phase/Regions.scala
```diff
@@ -35,20 +35,17 @@ import scala.collection.immutable.SortedSet
...
+    val errors = ParOps.parMap(root.defs)(kv => visitDef(kv._2)).flatten
 
     // TODO: Instances
-
-    errs match {
-      case Nil => Validation.success(())
-      case es => Validation.SoftFailure((), Chain.from(es))
-    }
+    Validation.toSuccessOrSoftFailure((), errors.asInstanceOf[Iterable[Recoverable]])
+      .asInstanceOf[Validation[Unit, CompilationMessage]]
```
magnus-madsen, line 42: But now-- can we remove these instanceof? :)

jaschdoc: Nope :)

jaschdoc: Scala can't infer the types :/

magnus-madsen: :(

magnus-madsen: Can we just annotate `errors: List[TypeError.RegionVarEscapes]`?

jaschdoc: Does not work either

magnus-madsen: Hmm. But in the other phases we do not need casts. Take Safety for example:

https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/Safety.scala#L45

jaschdoc: Ah but that's because `SafetyError` extends `Recoverable` directly in the trait. `TypeError` can be both `Unrecoverable` and `Recoverable` so each case class extends one of them. We can either introduce a new `RegionErorr` type that extends `Recoverable` or lose some safety by extending both traits directly on `TypeError`. I prefer the former solution.

jaschdoc: Okay forget everything I said. We forgot to extends either `Recoverable` on `RegionVarEscapes`.

magnus-madsen (approved):

PR #7293, opened 2024-02-26
refactor: polish `Tailrec`

JonathanStarup:


JonathanStarup:
Somehow its wrong to say that the body of letrec (`e2` in `letrec x = e1; e2`) is in tail position 🤔

magnus-madsen:
> Somehow its wrong to say that the body of letrec (`e2` in `letrec x = e1; e2`) is in tail position 🤔

What was the solution to this?

JonathanStarup:
The inliner also didn't look at letrec when rewriting recursive calls

magnus-madsen (approved):

magnus-madsen:
Thanks!

PR #7294, opened 2024-02-26
refactor: precompute `typesOf` in `Reducer`

### LINKED ISSUE #6677: Precompute `JvmOps.typesOf` in `Reducer`
I would do it in two PRs:

- [x] Move the computation of `typesOf` into `Reducer`. Store the types as part of the AST-- like done for AnonClasses.
- [ ] Get rid of `typesOf` and instead collect types while traversing the AST.

sockmaster27:
Fixes #6677


main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
```diff
@@ -33,131 +33,152 @@ import scala.jdk.CollectionConverters._
...
 
     val newDefs = ParOps.parMapValues(root.defs)(visitDef)
+    val defTypes = ctx.defTypes.asScala.toSet
 
-    val newRoot = root.copy(defs = newDefs, anonClasses = ctx.anonClasses.asScala.toList)
-    val types = typesOf(newRoot)
-    newRoot.copy(types = types)
+    val effectTypes = root.effects.values.toSet.flatMap(typesOfEffect)
+
+    val types = nestedTypesOf(Set.empty, Queue.from(defTypes ++ effectTypes))
```
JonathanStarup, line 47: Since this is single threaded, can you try and make the SharedContext hold a concurrent set of types instead? just so there's no duplicates to go over. I think it should be faster but you can run benchmarks for both

main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
```diff
@@ -33,131 +33,152 @@ import scala.jdk.CollectionConverters._
...
-        case JvmMethod(ident, fparams, clo, retTpe, purity, loc) =>
-          val c = visitExpr(clo)
-          JvmMethod(ident, fparams, c, retTpe, purity, loc)
-      }
-      ctx.anonClasses.add(AnonClass(name, clazz, tpe, specs, loc))
+      Def(ann, mod, sym, cparams, fparams, ls, pcPoints, e, tpe, unboxedType, purity, loc)
+  }
+
+  private def visitExpr(exp0: Expr)(implicit lctx: LocalContext, ctx: SharedContext): Expr = {
+    ctx.defTypes.add(exp0.tpe)
```
JonathanStarup, line 74: All the cases that have `ctx.defTypes.add(tpe)` seems redundant because of this line right? You do like the code did previously but i think this should be the only call to ctx.detTypes.add

JonathanStarup:
I think this should be merged. This PR perfectly eliminates the double ast traversal while maintaining current behavior. We can always make additional improvements in other PRs.

JonathanStarup (approved):

main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
sockmaster27: A little over 30ms with this

main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
magnus-madsen: OK, can we do that? Or was it already done?

main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
```diff
@@ -178,103 +193,18 @@ object Reducer {
     *
     * We use a concurrent (non-blocking) linked queue to ensure thread-safety.
     */
-  private case class SharedContext(anonClasses: ConcurrentLinkedQueue[AnonClass])
+  private case class SharedContext(anonClasses: ConcurrentLinkedQueue[AnonClass], defTypes: ConcurrentHashMap.KeySetView[MonoType, java.lang.Boolean])
```
magnus-madsen, line 196: I am probably getting old and cranky, but what the hell is KeySetView and why do I need it? Can we not just operate on the concurrent hash map? :)

magnus-madsen:
I added some comments before merge.

@JonathanStarup Will you open a ticket with the next steps to be done after this?

main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
```diff
@@ -33,131 +33,145 @@ import scala.jdk.CollectionConverters._
...
-    implicit val ctx: SharedContext = SharedContext(new ConcurrentLinkedQueue)
+    implicit val ctx: SharedContext = SharedContext(new ConcurrentLinkedQueue, ConcurrentHashMap.newKeySet())
 
     val newDefs = ParOps.parMapValues(root.defs)(visitDef)
+    val defTypes = ctx.defTypes.asScala.toSet
 
-    val newRoot = root.copy(defs = newDefs, anonClasses = ctx.anonClasses.asScala.toList)
-    val types = typesOf(newRoot)
-    newRoot.copy(types = types)
+    val effectTypes = root.effects.values.toSet.flatMap(typesOfEffect)
```
magnus-madsen, line 45: .values.toSet and flatMap can be expensive here (?)

main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
sockmaster27: Yes, but it would just require you to replace all `x.add(y)` with `x.insert(y, true)` or whatever.

main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
sockmaster27: It's done

main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
sockmaster27: This takes the Reducer phase from ~30ms to ~50ms on my machine.
Do you still want it?

magnus-madsen: Is this with XPerf and n=21?

sockmaster27: This is with `--Xbenchmark-phases`.

main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
JonathanStarup: Its on effects so i guess its small, but we could traverse it in parallel like with defs. Ill add it to the followup

JonathanStarup:
> @JonathanStarup Will you open a ticket with the next steps to be done after this?

#7347

main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
magnus-madsen: I want to keep the compiler consistent. So we should do everything the same.

Maybe KeySetView is the right thing to do, but then let us explore it in a separate PR and use it everywhere.

Meanwhile for this PR let us keep it simple stupid, i.e. use a ConcurrentMap with Unit

main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
JonathanStarup: yeah it can wait i think

magnus-madsen (approved):

magnus-madsen:
@JonathanStarup Merged. You can proceed with the cleanup :)

PR #7297, opened 2024-02-27
docs: MonoDefs new documentation

JonathanStarup:
Update documentation to reflect upcoming changes. This is a way to describe and discuss changes before I go and change the code.


main/src/ca/uwaterloo/flix/language/phase/MonoDefs.scala
```diff
@@ -686,6 +731,10 @@ object MonoDefs {
   /**
     * Performs type erasure on the given type `tpe`.
     *
+    * Returns the input type as-is if the input type is normalized.
+    *
+    * TODO should be deleted
+    *
     * Flix does not erase normal types, but it does erase Boolean and caseset formulas.
     */
   private def eraseType(tpe: Type)(implicit root: LoweredAst.Root, flix: Flix): Type = tpe match {
```
JonathanStarup, line 740: Based on our discussion of typematch - if we want typematch to work always, we cannot have this function - so it seems we should delete this. It also doesn't erase things properly because it only looks at variables. It seems like a no-op to me

magnus-madsen: Agreed it. We should not erase here.

magnus-madsen: Rebase and its gone.

main/src/ca/uwaterloo/flix/language/phase/MonoDefs.scala
```diff
@@ -195,15 +232,16 @@ object MonoDefs {
...
+      * of that function.
       *
       * For example, if the function:
       *
       * -   def fst[a, b](x: a, y: b): a = ...
       *
       * has been specialized w.r.t. to `Int` and `Str` then this map will contain an entry:
       *
-      * -   (fst, (Int, Str) -> Int) -> fst$1
+      * -   (fst, Int32 -> Str -> Int) -> fst$1 (TODO what is function normalization)
```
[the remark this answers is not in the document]
mlutze: I think we agreed that n-ary functions are different from m-ary functions

magnus-madsen (approved):

magnus-madsen:
Thanks -- good comments!

PR #7298, opened 2024-02-27
refactor: remove `MonoDefs.erase`

JonathanStarup:
Conceptually it should never do anything and I've verified that it never does anything on our tests

Reasoning:
- Erase only does something for variables and associated types
- variables and associated types have been eliminated by the strict substitution


main/src/ca/uwaterloo/flix/language/phase/Eraser.scala
```diff
@@ -26,7 +26,6 @@ import ca.uwaterloo.flix.util.ParOps
   * - Lazy
   *   - component type erasure
   *   - force casting
-  * - Enum (todo)
```
JonathanStarup, line 29: just old leftover todo that's not actually a todo

magnus-madsen (approved):

PR #7306, opened 2024-02-27
feat: introduce constraint generation

mlutze:
related to #6499

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
@@ -695,11 +695,8 @@ object ConstraintGeneration {
         (resTpe, resEff)
 
       case Expr.TryCatch(exp, rules, loc) =>
-        val (tpes, effs) = rules.map {
-          case KindedAst.CatchRule(_, _, body) =>
-            visitExp(body)
-        }.unzip
         val (tpe, eff) = visitExp(exp)
+        val (tpes, effs) = rules.map(visitCatchRule).unzip
         val ruleTpe = c.unifyAllTypesM(tpes, Kind.Star, loc)
         c.unifyTypeM(tpe, ruleTpe, loc)
         val resTpe = tpe
@@ -1089,9 +1086,9 @@ object ConstraintGeneration {
   }
 
   /**
-    * Generates constraints for the given match rule.
+    * Generates constraints for the given typematch rule.
     *
-    * Returns the pattern type, the body's type, and the body's effect
+    * Returns the the body's type and the body's effect
     */
   private def visitTypeMatchRule(rule: KindedAst.TypeMatchRule)(implicit c: TypeContext, root: KindedAst.Root, flix: Flix): (Type, Type) = rule match {
     case KindedAst.TypeMatchRule(sym, declTpe, exp) =>
@@ -1106,5 +1103,14 @@ object ConstraintGeneration {
       c.unifyTypeM(sym.tvar, declTpe, sym.loc)
 
       visitExp(exp)
+  }
+
+  /**
+    * Generates constraints for the given catch rule.
+    *
+    * Returns the the body's type and the body's effect
+    */
+  private def visitCatchRule(rule: KindedAst.CatchRule)(implicit c: TypeContext, root: KindedAst.Root, flix: Flix): Unit = rule match {
+    case KindedAst.CatchRule(_, _, exp) => visitExp(exp)
   }
 }
--- a/main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
@@ -687,9 +687,8 @@ object ConstraintGeneration {
         (resTpe, resEff)
 
       case Expr.Without(exp, _, _) =>
-        //        val effType = Type.Cst(TypeConstructor.Effect(effUse.sym), effUse.loc)
-        //        val expected = Type.mkDifference(Type.freshVar(Kind.Bool, loc), effType, loc)
-        // TODO EFF-MIGRATION use expected
+        // We ignore the `without` here.
+        // TODO EFF-MIGRATION Use set subtraction when we have set effects.
         val (tpe, eff) = visitExp(exp)
         val resTpe = tpe
         val resEff = eff
--- a/main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
@@ -27,7 +27,7 @@ object ConstraintGeneration {
     * Generates constraints for the given expression `exp0`, adding them to the type context `c`.
     *
     * Returns the type of the expression and its effect.
-    * The type and effect make include variables that must be resolved.
+    * The type and effect may include variables that must be resolved.
     */
   def visitExp(exp0: KindedAst.Expr)(implicit c: TypeContext, root: KindedAst.Root, flix: Flix): (Type, Type) = {
     // Make the context's level available
--- a/main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
@@ -731,12 +731,11 @@ object ConstraintGeneration {
         val op = lookupOp(opUse.sym, opUse.loc)
         val effTpe = Type.Cst(TypeConstructor.Effect(opUse.sym.eff), loc)
 
-        val opTpe = getDoType(op)
-
         // length check done in Resolver
-        val effs = (exps zip op.spec.fparams) map {
-          case (arg, fparam) => visitArg(arg, fparam)
-        }
+        val effs = visitOpArgs(op, exps)
+
+        // specialize the return type of the op if needed
+        val opTpe = getDoType(op)
 
         c.unifyTypeM(opTpe, tvar, loc)
         val resTpe = tvar
@@ -1105,11 +1104,22 @@ object ConstraintGeneration {
   }
 
   /**
+    * Generates constraints unifying each argument's type with the corresponding parameter of the operation.
+    *
+    * The number of arguments must match the number of parameters (this check is done in Resolver).
+    */
+  private def visitOpArgs(op: KindedAst.Op, args: List[KindedAst.Expr])(implicit c: TypeContext, root: KindedAst.Root, flix: Flix): List[Type] = {
+    (args zip op.spec.fparams) map {
+      case (arg, fparam) => visitOpArg(arg, fparam)
+    }
+  }
+
+  /**
     * Generates constraints unifying the given argument's type with the formal parameter's type.
     *
     * Returns the effect of the argument.
     */
-  def visitArg(arg: KindedAst.Expr, fparam: KindedAst.FormalParam)(implicit c: TypeContext, root: KindedAst.Root, flix: Flix): Type = {
+  private def visitOpArg(arg: KindedAst.Expr, fparam: KindedAst.FormalParam)(implicit c: TypeContext, root: KindedAst.Root, flix: Flix): Type = {
     val (tpe, eff) = visitExp(arg)
     c.expectTypeM(expected = fparam.tpe, actual = tpe, arg.loc)
     eff
--- a/main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
@@ -905,7 +905,7 @@ object ConstraintGeneration {
     *
     * Returns the pattern's type. The type may be a variable which must later be resolved.
     */
-  def visitPattern(pat0: KindedAst.Pattern)(implicit c: TypeContext, root: KindedAst.Root, flix: Flix): Type = {
+  private def visitPattern(pat0: KindedAst.Pattern)(implicit c: TypeContext, root: KindedAst.Root, flix: Flix): Type = {
     // Make the context's level available
     // This is a def rather than a val because c is mutable.
     implicit def level: Level = c.getLevel
--- a/main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
@@ -957,18 +957,10 @@ object ConstraintGeneration {
         val freshRowVar = Type.freshVar(Kind.RecordRow, loc.asSynthetic)
         val freshRecord = Type.mkRecord(freshRowVar, loc.asSynthetic)
 
-        def mkRecordType(patTypes: List[(Name.Label, Type, SourceLocation)]): Type = {
-          val ps = patTypes.foldRight(freshRowVar: Type) {
-            case ((lbl, t, l), acc) => Type.mkRecordRowExtend(
-              lbl, t, acc, l)
-          }
-          Type.mkRecord(ps, loc)
-        }
-
         val tailTpe = visitPattern(pat)
         c.unifyTypeM(freshRecord, tailTpe, loc.asSynthetic)
         val patTpes = pats.map(visitRecordLabelPattern)
-        val resTpe = mkRecordType(patTpes)
+        val resTpe = mkRecordType(patTpes, freshRowVar, loc)
         c.unifyTypeM(resTpe, tvar, loc)
         resTpe
 
@@ -1121,9 +1113,24 @@ object ConstraintGeneration {
     *
     * Returns the type and effect of the rule body.
     */
-  private def visitDefaultRule(exp0: Option[KindedAst.Expr], loc: SourceLocation)(implicit l: Level, c: TypeContext, root: KindedAst.Root, flix: Flix): (Type, Type) =
+  private def visitDefaultRule(exp0: Option[KindedAst.Expr], loc: SourceLocation)(implicit l: Level, c: TypeContext, root: KindedAst.Root, flix: Flix): (Type, Type) = {
     exp0 match {
       case None => (Type.freshVar(Kind.Star, loc), Type.Pure)
       case Some(exp) => visitExp(exp)
     }
+  }
+
+
+  /**
+    * Builds a record type from the given fields and rest type.
+    *
+    * The rest type represents the "tail" of the record.
+    */
+  private def mkRecordType(patTypes: List[(Name.Label, Type, SourceLocation)], rest: Type, loc: SourceLocation): Type = {
+    val ps = patTypes.foldRight(rest) {
+      case ((lbl, t, l), acc) => Type.mkRecordRowExtend(
+        lbl, t, acc, l)
+    }
+    Type.mkRecord(ps, loc)
+  }
 }
--- a/main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
@@ -671,7 +671,7 @@ object ConstraintGeneration {
         }
 
       case Expr.UncheckedCast(exp, declaredTpe, declaredEff, tvar, loc) =>
-        // A cast expression is unsound; the type system assumes the declared type and effect are correct.
+        // An unchecked cast expression is unsound; the type system assumes the declared type and effect are correct.
         val (actualTyp, actualEff) = visitExp(exp)
         c.unifyTypeM(tvar, declaredTpe.getOrElse(actualTyp), loc)
         val resTpe = tvar
@@ -679,6 +679,8 @@ object ConstraintGeneration {
         (resTpe, resEff)
 
       case Expr.UncheckedMaskingCast(exp, _) =>
+        // A masking cast expression is unsound; the type system assumes the declared type and effect are correct.
+        // The expression is treated as impure later in the compiler to prevent erasure in optimizations.
         val (tpe, _) = visitExp(exp)
         val resTpe = tpe
         val resEff = Type.Pure
--- a/main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
@@ -992,6 +992,8 @@ object ConstraintGeneration {
 
   /**
     * Generates constraints for the pattern.
+    *
+    * Returns the pattern's type. The type may be a variable which must later be resolved.
     */
   def visitPattern(pat0: KindedAst.Pattern)(implicit c: TypeContext, root: KindedAst.Root, flix: Flix): Type = {
     // Make the context's level available
--- a/main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
@@ -32,7 +32,7 @@ object ConstraintGeneration {
     implicit def level: Level = c.getLevel
 
     exp0 match {
-      case Expr.Var(sym, loc) => (sym.tvar, Type.Pure)
+      case Expr.Var(sym, _) => (sym.tvar, Type.Pure)
       case Expr.Def(sym, tvar, loc) =>
         val defn = root.defs(sym)
         val (tconstrs, defTpe) = Scheme.instantiate(defn.spec.sc, loc.asSynthetic)
@@ -51,11 +51,11 @@ object ConstraintGeneration {
         val resEff = Type.Pure
         (resTpe, resEff)
 
-      case Expr.Hole(sym, tpe, loc) =>
+      case Expr.Hole(_, tpe, _) =>
         (tpe, Type.Pure)
 
       case Expr.HoleWithExp(exp, tvar, evar, loc) =>
-        val (tpe, eff) = visitExp(exp)
+        val (_, eff) = visitExp(exp)
         // effect type is AT LEAST the inner expression's effect
         val atLeastEff = Type.mkUnion(eff, Type.freshVar(Kind.Eff, loc.asSynthetic), loc.asSynthetic)
         c.unifyTypeM(atLeastEff, evar, loc)
@@ -66,10 +66,10 @@ object ConstraintGeneration {
 
       case e: Expr.OpenAs => RestrictableChooseConstraintGeneration.visitOpenAs(e)
 
-      case Expr.Use(sym, alias, exp, loc) =>
+      case Expr.Use(_, _, exp, _) =>
         visitExp(exp)
 
-      case Expr.Cst(cst, loc) =>
+      case Expr.Cst(cst, _) =>
         val resTpe = Type.constantType(cst)
         val resEff = Type.Pure
         (resTpe, resEff)
@@ -349,19 +349,19 @@ object ConstraintGeneration {
         (resTpe, resEff)
 
       case Expr.Stm(exp1, exp2, loc) =>
-        val (tpe1, eff1) = visitExp(exp1)
+        val (_, eff1) = visitExp(exp1)
         val (tpe2, eff2) = visitExp(exp2)
         val resTpe = tpe2
         val resEff = Type.mkUnion(eff1, eff2, loc)
         (resTpe, resEff)
 
-      case Expr.Discard(exp, loc) =>
-        val (tpe, eff) = visitExp(exp)
+      case Expr.Discard(exp, _) =>
+        val (_, eff) = visitExp(exp)
         val resTpe = Type.Unit
         val resEff = eff
         (resTpe, resEff)
 
-      case Expr.Let(sym, mod, exp1, exp2, loc) =>
+      case Expr.Let(sym, _, exp1, exp2, loc) =>
         val (tpe1, eff1) = visitExp(exp1)
         c.unifyTypeM(sym.tvar, tpe1, loc)
         val (tpe2, eff2) = visitExp(exp2)
@@ -369,7 +369,7 @@ object ConstraintGeneration {
         val resEff = Type.mkUnion(eff1, eff2, loc)
         (resTpe, resEff)
 
-      case Expr.LetRec(sym, ann, mod, exp1, exp2, loc) =>
+      case Expr.LetRec(sym, _, _, exp1, exp2, loc) =>
         // exp1 is known to be a lambda syntactically
         val (tpe1, eff1) = visitExp(exp1)
         c.unifyTypeM(sym.tvar, tpe1, exp1.loc)
@@ -378,7 +378,7 @@ object ConstraintGeneration {
         val resEff = Type.mkUnion(eff1, eff2, loc)
         (resTpe, resEff)
 
-      case Expr.Region(tpe, loc) =>
+      case Expr.Region(tpe, _) =>
         val resTpe = tpe
         val resEff = Type.Pure
         (resTpe, resEff)
@@ -409,7 +409,7 @@ object ConstraintGeneration {
 
       case Expr.TypeMatch(exp, rules, loc) =>
         val bodies = rules.map(_.exp)
-        val (tpe, eff) = visitExp(exp)
+        val (_, eff) = visitExp(exp)
         // rigidify all the type vars in the rules
         rules.flatMap(_.tpe.typeVars.toList).map(_.sym).foreach(c.rigidifyM)
         // unify each rule's variable with its type
@@ -634,8 +634,8 @@ object ConstraintGeneration {
         val resEff = actualEff
         (resTpe, resEff)
 
-      case Expr.InstanceOf(exp, clazz, loc) =>
-        val (tpe, eff) = visitExp(exp)
+      case Expr.InstanceOf(exp, _, _) =>
+        val (_, eff) = visitExp(exp)
         c.expectTypeM(expected = Type.Pure, actual = eff, exp.loc)
         val resTpe = Type.Bool
         val resEff = Type.Pure
@@ -668,14 +668,14 @@ object ConstraintGeneration {
         val resEff = declaredEff.getOrElse(actualEff)
         (resTpe, resEff)
 
-      case Expr.UncheckedMaskingCast(exp, loc) =>
-        val (tpe, eff) = visitExp(exp)
+      case Expr.UncheckedMaskingCast(exp, _) =>
+        val (tpe, _) = visitExp(exp)
         val resTpe = tpe
         val resEff = Type.Pure
         (resTpe, resEff)
 
-      case Expr.Without(exp, effUse, loc) =>
-        val effType = Type.Cst(TypeConstructor.Effect(effUse.sym), effUse.loc)
+      case Expr.Without(exp, _, _) =>
+        //        val effType = Type.Cst(TypeConstructor.Effect(effUse.sym), effUse.loc)
         //        val expected = Type.mkDifference(Type.freshVar(Kind.Bool, loc), effType, loc)
         // TODO EFF-MIGRATION use expected
         val (tpe, eff) = visitExp(exp)
@@ -685,7 +685,7 @@ object ConstraintGeneration {
 
       case Expr.TryCatch(exp, rules, loc) =>
         val (tpes, effs) = rules.map {
-          case KindedAst.CatchRule(sym, clazz, body) =>
+          case KindedAst.CatchRule(_, _, body) =>
             visitExp(body)
         }.unzip
         val (tpe, eff) = visitExp(exp)
@@ -759,7 +759,7 @@ object ConstraintGeneration {
             // The operation type is `Void`. Flix does not have subtyping, but here we want something close to it.
             // Hence we treat `Void` as a fresh type variable.
             // An alternative would be to allow empty pattern matches, but that is cumbersome.
-            Type.freshVar(Kind.Star, operation.spec.tpe.loc, false, Ast.VarText.Absent)
+            Type.freshVar(Kind.Star, operation.spec.tpe.loc, isRegion = false, Ast.VarText.Absent)
           case _ => operation.spec.tpe
         }
 
@@ -774,7 +774,7 @@ object ConstraintGeneration {
 
         (resTpe, resEff)
 
-      case Expr.InvokeConstructor(constructor, args, loc) =>
+      case Expr.InvokeConstructor(constructor, args, _) =>
         val classTpe = Type.getFlixType(constructor.getDeclaringClass)
         val (_, _) = args.map(visitExp).unzip
         val resTpe = classTpe
@@ -790,14 +790,13 @@ object ConstraintGeneration {
         val resEff = Type.IO
         (resTpe, resEff)
 
-      case Expr.InvokeStaticMethod(method, args, loc) =>
-        val returnTpe = Type.getFlixType(method.getReturnType)
+      case Expr.InvokeStaticMethod(method, args, _) =>
         val (_, _) = args.map(visitExp).unzip
         val resTpe = Type.getFlixType(method.getReturnType)
         val resEff = Type.IO
         (resTpe, resEff)
 
-      case Expr.GetField(field, clazz, exp, loc) =>
+      case Expr.GetField(field, clazz, exp, _) =>
         val classType = Type.getFlixType(clazz)
         val fieldType = Type.getFlixType(field.getType)
         val (tpe, _) = visitExp(exp)
@@ -806,7 +805,7 @@ object ConstraintGeneration {
         val resEff = Type.IO
         (resTpe, resEff)
 
-      case Expr.PutField(field, clazz, exp1, exp2, loc) =>
+      case Expr.PutField(field, clazz, exp1, exp2, _) =>
         val fieldType = Type.getFlixType(field.getType)
         val classType = Type.getFlixType(clazz)
         val (tpe1, _) = visitExp(exp1)
@@ -817,26 +816,26 @@ object ConstraintGeneration {
         val resEff = Type.IO
         (resTpe, resEff)
 
-      case Expr.GetStaticField(field, loc) =>
+      case Expr.GetStaticField(field, _) =>
         val fieldType = Type.getFlixType(field.getType)
         val resTpe = fieldType
         val resEff = Type.IO
         (resTpe, resEff)
 
-      case Expr.PutStaticField(field, exp, loc) =>
+      case Expr.PutStaticField(field, exp, _) =>
         val (valueTyp, _) = visitExp(exp)
         c.expectTypeM(expected = Type.getFlixType(field.getType), actual = valueTyp, exp.loc)
         val resTpe = Type.Unit
         val resEff = Type.IO
         (resTpe, resEff)
 
-      case Expr.NewObject(name, clazz, methods, loc) =>
+      case Expr.NewObject(_, clazz, methods, _) =>
 
         /**
           * Generates constraints for the JVM method.
           */
         def visitJvmMethod(method: KindedAst.JvmMethod): Unit = method match {
-          case KindedAst.JvmMethod(ident, fparams, exp, returnTpe, eff, loc) =>
+          case KindedAst.JvmMethod(_, fparams, exp, returnTpe, _, _) =>
 
             /**
               * Constrains the given formal parameter to its declared type.
@@ -847,7 +846,7 @@ object ConstraintGeneration {
             }
 
             fparams.foreach(visitFormalParam)
-            val (bodyTpe, bodyEff) = visitExp(exp)
+            val (bodyTpe, _) = visitExp(exp)
             c.expectTypeM(expected = returnTpe, actual = bodyTpe, exp.loc)
           // TODO ASSOC-TYPES check eff matches declared eff ?
         }
@@ -929,14 +928,14 @@ object ConstraintGeneration {
       case Expr.Spawn(exp1, exp2, loc) =>
         val regionVar = Type.freshVar(Kind.Eff, loc)
         val regionType = Type.mkRegion(regionVar, loc)
-        val (tpe1, _) = visitExp(exp1)
+        val (_, _) = visitExp(exp1)
         val (tpe2, _) = visitExp(exp2)
         c.expectTypeM(expected = regionType, actual = tpe2, exp2.loc)
         val resTpe = Type.Unit
         val resEff = Type.mkUnion(Type.IO, regionVar, loc)
         (resTpe, resEff)
 
-      case Expr.ParYield(frags, exp, loc) =>
+      case Expr.ParYield(frags, exp, _) =>
         val patterns = frags.map(_.pat)
         val parExps = frags.map(_.exp)
         val patLocs = frags.map(_.loc)
@@ -971,7 +970,7 @@ object ConstraintGeneration {
       case e: Expr.FixpointInject => SchemaConstraintGeneration.visitFixpointInject(e)
       case e: Expr.FixpointProject => SchemaConstraintGeneration.visitFixpointProject(e)
 
-      case Expr.Error(m, tvar, evar) =>
+      case Expr.Error(_, tvar, evar) =>
         val resTpe = tvar
         val resEff = evar
         (resTpe, resEff)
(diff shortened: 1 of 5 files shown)


magnus-madsen:
- Check public/private.
- Check that every def has documentation.

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+
+object ConstraintGeneration {
+
+  /**
+    * Generates constraints for the given expression.
+    */
+  def visitExp(exp0: KindedAst.Expr)(implicit c: TypeContext, root: KindedAst.Root, flix: Flix): (Type, Type) = {
+    // Make the context's level available
+    // This is a def rather than a val because c is mutable.
+    implicit def level: Level = c.getLevel
```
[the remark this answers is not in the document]
mlutze: If we call c.getLevel then we have to call it every time we create a var, which is tedious.

An implicit def is the same as an implicit val, it's just not evaluated eagerly. That means that if `c.level` changes, a `def` will reflect that change, whereas a `val` will not.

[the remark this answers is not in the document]
mlutze: I'd like to remove levels anyway after this.

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+
+    exp0 match {
+      case Expr.Var(sym, loc) => (sym.tvar, Type.Pure)
+      case Expr.Def(sym, tvar, loc) =>
+        val defn = root.defs(sym)
+        val (tconstrs, defTpe) = Scheme.instantiate(defn.spec.sc, loc.asSynthetic)
```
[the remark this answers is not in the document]
mlutze: `cconstrs`?

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+        c.unifyTypeM(tvar, defTpe, loc)
+        c.addClassConstraintsM(tconstrs, loc)
```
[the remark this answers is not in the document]
mlutze: Yes

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+          case Ast.CheckedCastType.TypeCast =>
+            // Ignore the inferred type of exp.
+            val (_, eff) = visitExp(exp)
+            c.unifyTypeM(evar, eff, loc)
+            val resTpe = tvar
+            val resEff = evar
+            (resTpe, resEff)
+
+          case Ast.CheckedCastType.EffectCast =>
+            // We simply union the purity and effect with a fresh variable.
```
magnus-madsen, line 668: Outdated

magnus-madsen: Can we update both comments to make them use the same wording?

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+        (resTpe, resEff)
+
+      case Expr.InstanceOf(exp, clazz, loc) =>
+        val (tpe, eff) = visitExp(exp)
+        c.expectTypeM(expected = Type.Pure, actual = eff, exp.loc)
+        val resTpe = Type.Bool
+        val resEff = Type.Pure
+        (resTpe, resEff)
+
+      case Expr.CheckedCast(cast, exp, tvar, evar, loc) =>
```
magnus-madsen, line 657: Add a comment that this is sound (see the other expressions).

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+            // We simply union the purity and effect with a fresh variable.
+            val (tpe, eff) = visitExp(exp)
+            c.unifyTypeM(tvar, tpe, loc)
+            val resTpe = tvar
+            val resEff = Type.mkUnion(eff, evar, loc)
+            (resTpe, resEff)
+        }
+
+      case Expr.UncheckedCast(exp, declaredTpe, declaredEff, tvar, loc) =>
+        // A cast expression is unsound; the type system assumes the declared type is correct.
```
magnus-madsen, line 677: declared type and effect?

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+        val (actualTyp, actualEff) = visitExp(exp)
+        c.unifyTypeM(tvar, declaredTpe.getOrElse(actualTyp), loc)
+        val resTpe = tvar
+        val resEff = declaredEff.getOrElse(actualEff)
+        (resTpe, resEff)
+
+      case Expr.UncheckedMaskingCast(exp, loc) =>
```
magnus-madsen, line 684: I dont remember what this does. Can we get a comment like the others.

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+        val (tpe, eff) = visitExp(exp)
+        val resTpe = tpe
+        val resEff = Type.Pure
+        (resTpe, resEff)
+
+      case Expr.Without(exp, effUse, loc) =>
+        val effType = Type.Cst(TypeConstructor.Effect(effUse.sym), effUse.loc)
+        //        val expected = Type.mkDifference(Type.freshVar(Kind.Bool, loc), effType, loc)
+        // TODO EFF-MIGRATION use expected
```
magnus-madsen, line 693: Drop this stuff, but add a short comment about what to do once sets arrive.

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+        val (tpe, eff) = visitExp(exp)
+        val resTpe = tpe
+        val resEff = eff
+        (resTpe, resEff)
+
+      case Expr.TryCatch(exp, rules, loc) =>
```
magnus-madsen, line 699: Lets add a visitTryCatchRule

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+        val resTpe = tpe
+        val resEff = Type.mkUnion(eff :: effs, loc)
+        (resTpe, resEff)
+
+      case Expr.TryWith(exp, effUse, rules, tvar, loc) =>
+        val effect = root.effects(effUse.sym)
+        val ops = effect.ops.map(op => op.sym -> op).toMap
+        val continuationEffect = Type.freshVar(Kind.Eff, loc)
+
+        def unifyFormalParams(op: Symbol.OpSym, expected: List[KindedAst.FormalParam], actual: List[KindedAst.FormalParam]): Unit = {
```
magnus-madsen, line 716: move out

magnus-madsen: and document

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+          // length check done in Resolver
+          c.expectTypeArguments(op, expectedTypes = expected.map(_.tpe), actualTypes = actual.map(_.tpe), actual.map(_.loc), loc)
+        }
+
+        def visitHandlerRule(rule: KindedAst.HandlerRule, tryBlockTpe: Type): (Type, Type) = rule match {
```
magnus-madsen, line 721: move out and document

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+            visitExp(body)
+        }.unzip
+        val (tpe, eff) = visitExp(exp)
+        val ruleTpe = c.unifyAllTypesM(tpes, Kind.Star, loc)
+        c.unifyTypeM(tpe, ruleTpe, loc)
+        val resTpe = tpe
+        val resEff = Type.mkUnion(eff :: effs, loc)
+        (resTpe, resEff)
+
+      case Expr.TryWith(exp, effUse, rules, tvar, loc) =>
```
magnus-madsen, line 711: Can we clean this code up? At least a bit and add some useful comments? :)

magnus-madsen:
Important: Lets immediately start calling this code from somewhere, even if we throw away the result!

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+        val resEff = eff
+        (resTpe, resEff)
+
+      case Expr.ArrayLit(exps, exp, tvar, evar, loc) =>
+        val regionVar = Type.freshVar(Kind.Eff, loc)
+        val regionType = Type.mkRegion(regionVar, loc)
+        val (tpes, effs) = exps.map(visitExp).unzip
+        val (tpe, eff) = visitExp(exp)
+        c.expectTypeM(expected = regionType, actual = tpe, exp.loc)
+        val elmTpe = c.unifyAllTypesM(tpes, Kind.Star, loc)
```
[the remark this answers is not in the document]
mlutze: We can introduce a variable and add it to the unified types in the case where it's needed.

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+        val correctedBodyEff = c.purifyEff(effUse.sym, bodyEff) // Note: Does not work for polymorphic effects.
+        val (_, effs) = rules.map(visitHandlerRule(_, tpe)).unzip
+        c.unifyTypeM(continuationEffect, Type.mkUnion(correctedBodyEff :: effs, loc), loc)
+        c.unifyTypeM(tvar, tpe, loc)
+        val resultTpe = tpe
+        //        val resultEff = continuationEffect
+        val resultEff = Type.freshVar(Kind.Eff, loc) // TODO ASSOC-TYPES should be continuationEffect
+        (resultTpe, resultEff)
+
+      case Expr.Do(op, args, tvar, loc) =>
```
magnus-madsen, line 757: args -> exps

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+        val effect = root.effects(op.sym.eff)
```
magnus-madsen, line 758: effect -> eff or perhaps decl?

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+        val operation = effect.ops.find(_.sym == op.sym)
```
magnus-madsen, line 759: operation -> op

magnus-madsen: extract this line into a function.

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+            val defn = root.defs(sym)
+            val (tconstrs1, declaredType) = Scheme.instantiate(defn.spec.sc, loc2.asSynthetic)
+            val constrs1 = tconstrs1.map(_.copy(loc = loc))
+            Some((sym, tvar2, constrs1, declaredType))
+
+          case KindedAst.Expr.Sig(sym, tvar2, loc2) =>
+            // Case 2: Lookup the sym and instantiate its scheme.
+            val sig = root.classes(sym.clazz).sigs(sym)
+            val (tconstrs1, declaredType) = Scheme.instantiate(sig.spec.sc, loc2.asSynthetic)
+            val constrs1 = tconstrs1.map(_.copy(loc = loc))
```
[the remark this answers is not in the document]
mlutze: likely bug

But let's handle it in a followup. I'd like to avoid semantic changes if possible.

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+        val resEff = Type.mkUnion(effTpe :: operation.spec.eff :: effs, loc)
+
+        (resTpe, resEff)
+
+      case Expr.InvokeConstructor(constructor, args, loc) =>
+        val classTpe = getFlixType(constructor.getDeclaringClass)
+        val (_, _) = args.map(visitExp).unzip
+        val resTpe = classTpe
+        val resEff = Type.IO
+        (resTpe, resEff)
```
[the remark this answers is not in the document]
magnus-madsen: BIG TEXT: I would think that every case would include calls to c.unify unless there is a good reason (e.g. constant). 

Could we double check that? If there is no such call it should be documented imho! "This expression does not generate constraints because ...".

Presumably getFlixType does not generate constraints!

Do we not need to match the arguments against the types of the constructor? Am I stupid?

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+          exp0 match {
+            case None => (Type.freshVar(Kind.Star, loc), Type.Pure)
+            case Some(exp) => visitExp(exp)
+          }
+
+        val (ruleTypes, ruleEffs) = rules.map(visitSelectRule).unzip
+        val (defaultType, eff2) = visitDefaultRule(default)
+        c.unifyAllTypesM(tvar :: defaultType :: ruleTypes, Kind.Star, loc)
+        val resTpe = tvar
+        val resEff = Type.mkUnion(regionVar :: eff2 :: ruleEffs, loc)
```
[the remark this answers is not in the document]
magnus-madsen: Maybe at the end of the day, you can do a search for `::`.

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+        (resTpe, resEff)
+
+      case Expr.Spawn(exp1, exp2, loc) =>
```
magnus-madsen, line 942: We need a comment here that we don't fully understand yet what the type of spawn should be.

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+        val resEff = eff
+        (resTpe, resEff)
+
+      case e@Expr.FixpointConstraintSet(cs, tvar, loc) => SchemaConstraintGeneration.visitFixpointConstraintSet(e)
+      case e@Expr.FixpointLambda(pparams, exp, tvar, loc) => SchemaConstraintGeneration.visitFixpointLambda(e)
+      case e@Expr.FixpointMerge(exp1, exp2, loc) => SchemaConstraintGeneration.visitFixpointMerge(e)
+      case e@Expr.FixpointSolve(exp, loc) => SchemaConstraintGeneration.visitFixpointSolve(e)
+      case e@Expr.FixpointFilter(pred, exp, tvar, loc) => SchemaConstraintGeneration.visitFixpointFilter(e)
+      case e@Expr.FixpointInject(exp, pred, tvar, loc) => SchemaConstraintGeneration.visitFixpointInject(e)
+      case e@Expr.FixpointProject(pred, exp1, exp2, tvar, loc) => SchemaConstraintGeneration.visitFixpointProject(e)
```
magnus-madsen, line 985: e: , for all

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1069 @@
...
+  /**
+    * Generates constraints for the pattern.
+    */
+  def visitPattern(pat0: KindedAst.Pattern)(implicit c: TypeContext, root: KindedAst.Root, flix: Flix): Type = {
+    // Make the context's level available
+    // This is a def rather than a val because c is mutable.
+    implicit def level: Level = c.getLevel
+
+    pat0 match {
+
```
magnus-madsen, line 1001: linebreak

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1069 @@
...
+        (resTpe, resEff)
+    }
+  }
+
+  /**
+    * Generates constraints for the pattern.
+    */
+  def visitPattern(pat0: KindedAst.Pattern)(implicit c: TypeContext, root: KindedAst.Root, flix: Flix): Type = {
+    // Make the context's level available
+    // This is a def rather than a val because c is mutable.
```
magnus-madsen, line 997: update comment same as above.

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1069 @@
...
+        val caze = decl.cases(symUse.sym)
+
+        // Instantiate the type scheme of the case.
+        val (_, tagType) = Scheme.instantiate(caze.sc, loc.asSynthetic)
+
+        //
+        // The tag type is a function from the type of variant to the type of the enum.
+        //
+        val tpe = visitPattern(pat)
+        c.unifyTypeM(tagType, Type.mkPureArrow(tpe, tvar, loc), loc)
```
magnus-madsen, line 1024: simplify as above.

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1069 @@
...
+
+      case KindedAst.Pattern.Tuple(elms, loc) =>
+        val tpes = elms.map(visitPattern)
+        Type.mkTuple(tpes, loc)
+
+      case KindedAst.Pattern.Record(pats, pat, tvar, loc) =>
+        val freshRowVar = Type.freshVar(Kind.RecordRow, loc.asSynthetic)
+        val freshRecord = Type.mkRecord(freshRowVar, loc.asSynthetic)
+
+        def mkRecordType(patTypes: List[(Name.Label, Type, SourceLocation)]): Type = {
```
magnus-madsen, line 1036: extract

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1069 @@
...
+
+      case KindedAst.Pattern.RecordEmpty(loc) => Type.mkRecord(Type.RecordRowEmpty, loc)
+
+      case KindedAst.Pattern.Error(tvar, _) => tvar
+
+    }
+  }
+
+  /**
+    * Generates constraints for the record label pattern.
```
magnus-madsen, line 1059: What does this return?

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+
+      case Expr.Hole(sym, tpe, loc) =>
+        (tpe, Type.Pure)
+
+      case Expr.HoleWithExp(exp, tvar, evar, loc) =>
+        val (tpe, eff) = visitExp(exp)
+        // effect type is AT LEAST the inner expression's effect
+        val atLeastEff = Type.mkUnion(eff, Type.freshVar(Kind.Eff, loc.asSynthetic), loc.asSynthetic)
+        c.unifyTypeM(atLeastEff, evar, loc)
+        // result type is whatever is needed for the hole
```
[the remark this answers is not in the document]
mlutze: This is for hole-with-expression, which is different, I guess?

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+          */
+        def visitDefaultRule(exp0: Option[KindedAst.Expr]): (Type, Type) =
+          exp0 match {
+            case None => (Type.freshVar(Kind.Star, loc), Type.Pure)
+            case Some(exp) => visitExp(exp)
+          }
+
+        val (ruleTypes, ruleEffs) = rules.map(visitSelectRule).unzip
+        val (defaultType, eff2) = visitDefaultRule(default)
+        c.unifyAllTypesM(tvar :: defaultType :: ruleTypes, Kind.Star, loc)
```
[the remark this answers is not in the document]
mlutze: `ruleTypes` is a list so it's N things unified, not just 3

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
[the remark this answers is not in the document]
mlutze: Ditto for N things

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1069 @@
...
+
+      case Expr.Lambda(fparam, exp, loc) =>
+        c.unifyTypeM(fparam.sym.tvar, fparam.tpe, loc)
+        val (tpe, eff) = visitExp(exp)
+        val resTpe = Type.mkArrowWithEffect(fparam.tpe, eff, tpe, loc)
+        val resEff = Type.Pure
+        (resTpe, resEff)
+
+      case KindedAst.Expr.Unary(sop, exp, tvar, loc) => sop match {
+        case SemanticOp.BoolOp.Not =>
```
JonathanStarup, line 156: It seems all of these unary operations could use the same `helper(exp, tvar, type)` function

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+              */
+            def visitFormalParam(fparam: KindedAst.FormalParam): Unit = fparam match {
+              case KindedAst.FormalParam(sym, _, tpe, _, loc) =>
+                c.unifyTypeM(sym.tvar, tpe, loc)
+            }
+
+            fparams.foreach(visitFormalParam)
+            val (bodyTpe, bodyEff) = visitExp(exp)
+            c.expectTypeM(expected = returnTpe, actual = bodyTpe, exp.loc)
+          // TODO ASSOC-TYPES check eff matches declared eff ?
```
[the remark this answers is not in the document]
mlutze: A probable bug. To be addressed properly in followup.

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1069 @@
...
+          val resTpe = tvar
+          val resEff = eff
+          (resTpe, resEff)
+
+        case _ => throw InternalCompilerException(s"Unexpected unary operator: '$sop'.", loc)
+      }
+
+      case KindedAst.Expr.Binary(sop, exp1, exp2, tvar, loc) => sop match {
+
+        case SemanticOp.BoolOp.And | SemanticOp.BoolOp.Or =>
```
JonathanStarup, line 217: These could also all share a helper function, then itd be easier to read IMO, the reader just have to verify the hardcoded types of each operation, not all the copied code.

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1069 @@
...
+
+      case Expr.Discard(exp, _) =>
+        val (_, eff) = visitExp(exp)
+        val resTpe = Type.Unit
+        val resEff = eff
+        (resTpe, resEff)
+
+      case Expr.Let(sym, _, exp1, exp2, loc) =>
+        val (tpe1, eff1) = visitExp(exp1)
+        c.unifyTypeM(sym.tvar, tpe1, loc)
```
JonathanStarup, line 378: let uses `loc` and letrec uses `exp1.loc`. which is best?

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+        val elmVar = Type.freshVar(Kind.Star, loc)
+        val regionVar = Type.freshVar(Kind.Eff, loc)
+        val arrayType = Type.mkArray(elmVar, regionVar, loc)
+        val (tpe1, eff1) = visitExp(exp1)
+        val (tpe2, eff2) = visitExp(exp2)
+        val (tpe3, eff3) = visitExp(exp3)
+        c.expectTypeM(expected = arrayType, actual = tpe1, exp1.loc)
+        c.expectTypeM(expected = Type.Int32, actual = tpe2, exp2.loc)
+        c.expectTypeM(expected = elmVar, actual = tpe3, exp3.loc)
+        c.unifyTypeM(evar, Type.mkUnion(List(regionVar, eff1, eff2, eff3), loc), loc)
```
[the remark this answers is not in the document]
mlutze: Adding to followup list

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1069 @@
...
+      case Expr.Let(sym, _, exp1, exp2, loc) =>
+        val (tpe1, eff1) = visitExp(exp1)
+        c.unifyTypeM(sym.tvar, tpe1, loc)
+        val (tpe2, eff2) = visitExp(exp2)
+        val resTpe = tpe2
+        val resEff = Type.mkUnion(eff1, eff2, loc)
+        (resTpe, resEff)
+
+      case Expr.LetRec(sym, _, _, exp1, exp2, loc) =>
+        // exp1 is known to be a lambda syntactically
```
JonathanStarup, line 385: I don't think that matters to the rule? if not then we should just remove the comment.

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+        val resTpe = tpe2
+        val resEff = Type.mkUnion(eff1, eff2, loc)
+        (resTpe, resEff)
+
+      case Expr.Region(tpe, loc) =>
+        val resTpe = tpe
+        val resEff = Type.Pure
+        (resTpe, resEff)
+
+      case Expr.Scope(sym, regionVar, exp, evar, loc) =>
```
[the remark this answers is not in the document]
JonathanStarup: Maybe `enterRegion` should be a higher order function instead of separate enter/exit calls?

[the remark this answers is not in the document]
mlutze: It must be inside because it must be resolved before purification happens.

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+        c.enterRegionM(regionVar.sym)
+        c.unifyTypeM(sym.tvar, Type.mkRegion(regionVar, loc), loc)
+        val (tpe, eff) = visitExp(exp)
+        c.exitRegionM(evar, eff, loc)
+        val resTpe = tpe
+        val resEff = evar
+        (resTpe, resEff)
+
+      case Expr.Match(exp, rules, loc) =>
```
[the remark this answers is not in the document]
JonathanStarup: I think you could make a helper function for guards fx. since it justs checks that everything is bool and returns a list of effects. That would help the outer readability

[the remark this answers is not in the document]
JonathanStarup: similar for visitPatterns, and visitBodies.
```scala
visitPatterns(rules.map(_.pat), tpe, ...)
val guardEffs = visitGuards(rules.flatMap(_.guard), ...)
val (bodyTpe, bodyEffs) = visitBodies(rules.map(_.exp), ...)
val resTpe = bodyTpe
val resEff = Type.mkUnion(eff :: guardEffs ::: bodyEffs, loc)
```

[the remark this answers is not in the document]
JonathanStarup: perhaps there can be a helper function for branching expressions. Can be used for if branches, match branches, typematch branches

JonathanStarup (changes requested):
I got too hungry at match, will do more later

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
mlutze: @magnus-madsen Fight Jonathan

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1081 @@
...
+      case e: Expr.RestrictableTag => RestrictableChooseConstraintGeneration.visitRestrictableTag(e)
+
+      case Expr.Tuple(elms, loc) =>
+        val (elmTpes, elmEffs) = elms.map(visitExp).unzip
+        val resTpe = Type.mkTuple(elmTpes, loc)
+        val resEff = Type.mkUnion(elmEffs, loc)
+        (resTpe, resEff)
+
+      case Expr.RecordEmpty(loc) =>
+        val resTpe = Type.mkRecord(Type.RecordRowEmpty, loc)
```
JonathanStarup, line 472: `Type.RecordRowEmpty` should be `Type.mkRecordRowEmpty(loc)` right? same thing in the next case

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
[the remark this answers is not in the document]
JonathanStarup: could also be varargs no? `mkUnion(tpe1: Type, tpes: Type*): Type`

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1081 @@
...
+        val resEff = evar
+        (resTpe, resEff)
+
+      case Expr.ArrayLength(exp, loc) =>
+        val elmVar = Type.freshVar(Kind.Star, loc)
+        val regionVar = Type.freshVar(Kind.Eff, loc)
+        val (tpe, eff) = visitExp(exp)
+        c.expectTypeM(Type.mkArray(elmVar, regionVar, loc), tpe, exp.loc)
+        val resTpe = Type.Int32
+        val resEff = eff
```
JonathanStarup, line 580: I almost fell into the trap of suggesting a union was missing, maybe a comment is warranted?

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1081 @@
...
+        (resTpe, resEff)
+
+      case Expr.Assign(exp1, exp2, evar, loc) =>
+        val elmVar = Type.freshVar(Kind.Star, loc)
+        val regionVar = Type.freshVar(Kind.Eff, loc)
+        val refType = Type.mkRef(elmVar, regionVar, loc)
+        val (tpe1, eff1) = visitExp(exp1)
+        val (tpe2, eff2) = visitExp(exp2)
+        c.expectTypeM(expected = refType, actual = tpe1, exp1.loc)
+        c.expectTypeM(expected = elmVar, actual = tpe2, exp2.loc)
```
JonathanStarup, line 641: I dont think this one should use expected since elmVar is just a variable

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1081 @@
...
+        c.expectTypeM(expected = expectedTpe.getOrElse(Type.freshVar(Kind.Star, loc)), actual = actualTpe, loc)
+        c.unifyTypeM(actualTpe, tvar, loc)
+        c.expectTypeM(expected = expectedEff.getOrElse(Type.freshVar(Kind.Eff, loc)), actual = actualEff, loc)
+        val resTpe = tvar
+        val resEff = actualEff
+        (resTpe, resEff)
+
+      case Expr.InstanceOf(exp, _, _) =>
+        val (_, eff) = visitExp(exp)
+        c.expectTypeM(expected = Type.Pure, actual = eff, exp.loc)
```
JonathanStarup, line 659: I don't think this should be there, why would it matter that the expression being checked is pure? if it is something impure you could just letbind it so I'm not sure what it achieves. @magnus-madsen ?

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
JonathanStarup: Ast.CheckedCastType.TypeCast looks unsound to me, is it checked post-typechecking?

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1081 @@
...
+            val resEff = Type.mkUnion(eff, evar, loc)
+            (resTpe, resEff)
+        }
+
+      case Expr.UncheckedCast(exp, declaredTpe, declaredEff, tvar, loc) =>
+        // A cast expression is unsound; the type system assumes the declared type is correct.
+        val (actualTyp, actualEff) = visitExp(exp)
+        c.unifyTypeM(tvar, declaredTpe.getOrElse(actualTyp), loc)
+        val resTpe = tvar
+        val resEff = declaredEff.getOrElse(actualEff)
```
JonathanStarup, line 688: This could be nominated to have an `evar` so this getOrElse logic isn't copied in reconstruction.

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
JonathanStarup: could also add `visitRules` a la my suggestion in match so the main snippet here is more succinct

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1081 @@
...
+        }
+
+        val tailTpe = visitPattern(pat)
+        c.unifyTypeM(freshRecord, tailTpe, loc.asSynthetic)
+        val patTpes = pats.map(visitRecordLabelPattern)
+        val resTpe = mkRecordType(patTpes)
+        c.unifyTypeM(resTpe, tvar, loc)
+        resTpe
+
+      case KindedAst.Pattern.RecordEmpty(loc) => Type.mkRecord(Type.RecordRowEmpty, loc)
```
JonathanStarup, line 1061: Type.mkRecordRowEmpty(loc)

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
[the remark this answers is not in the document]
mlutze: needs a location though

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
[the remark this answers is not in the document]
mlutze: I opted for `visitMatchRule` since it's consistent with what we do in other phases

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
mlutze: Yes, in the Safety phase

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1069 @@
...
+          case Some((sym, tvar2, constrs1, declaredType)) =>
+            //
+            // Special Case: We are applying a Def or Sig and we break apart its declared type.
+            //
+            val declaredEff = declaredType.arrowEffectType
+            val declaredArgumentTypes = declaredType.arrowArgTypes
+            val declaredResultType = declaredType.arrowResultType
+
+            val (tpes, effs) = exps.map(visitExp).unzip
+            c.expectTypeArguments(sym, declaredArgumentTypes, tpes, exps.map(_.loc), loc)
```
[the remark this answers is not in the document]
mlutze: per discussion, this problem seems to be addressed in an earlier phase where things are desugared to match the number of args

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
mlutze: eh, it creates an extra object for no real gain (type locations for inferred types are evil)

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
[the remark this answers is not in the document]
JonathanStarup: could be `mkUnion(loc, tpe, tpes)` but not as cool no

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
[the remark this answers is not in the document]
magnus-madsen: Jonathans suggestion or just leave this as-is, until a refactor. (Update ticket).

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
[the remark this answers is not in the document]
magnus-madsen: Thanks for the info. I suggest we make this just like hole. Then, once we have sets and subeffecting, we get the right solution.

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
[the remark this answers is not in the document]
magnus-madsen: Ooookay? :)

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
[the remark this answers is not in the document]
magnus-madsen: I don't know. I am just observing an inconsistency.

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
magnus-madsen: I think the comment is there to remind us not to check that it is a lambda?

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
magnus-madsen: I agree with Matt

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1081 @@
...
+        c.expectTypeM(expected = elmVar, actual = tpe2, exp2.loc)
+        c.unifyTypeM(evar, Type.mkUnion(eff1, eff2, regionVar, loc), loc)
+        val resTpe = Type.Unit
+        val resEff = evar
+        (resTpe, resEff)
+
+      case Expr.Ascribe(exp, expectedTpe, expectedEff, tvar, loc) =>
+        // An ascribe expression is sound; the type system checks that the declared type matches the inferred type.
+        val (actualTpe, actualEff) = visitExp(exp)
+        c.expectTypeM(expected = expectedTpe.getOrElse(Type.freshVar(Kind.Star, loc)), actual = actualTpe, loc)
```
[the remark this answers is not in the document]
magnus-madsen: I thought about that too. But it seems it would be less elegant.

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
magnus-madsen: Agree. It should not require purity. Good catch!

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
magnus-madsen: Yes

magnus-madsen:
The reason to be so picky is that we really don't want bugs in the generation. We want all our bugs in the solving, so we can focus our attention there.

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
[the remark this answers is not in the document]
JonathanStarup: I think the current code is the correct type rule. if you think of `???(???, someExpression: t \ ef, ???)` (where `???` has unknown effect) then it will have any type and effect `ef + alpha`

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
JonathanStarup: there is no wellformedness checking other places in constraint gen. It also doesn't matter to the type rule whether or not its a lambda.

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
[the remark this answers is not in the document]
JonathanStarup: I think its more elegant?
```
c.expectTypeM(expected = expectedTpe.getOrElse(Type.freshVar(Kind.Star, loc)), actual = actualTpe, loc)
// vs
expectedTpe.foreach(t => c.expectTypeM(expected = t, actual = actualTpe, loc))
```

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+        // Lookup the case declaration.
+        val caze = decl.cases(symUse.sym)
+
+        // Instantiate the type scheme of the case.
+        val (_, tagType) = Scheme.instantiate(caze.sc, loc.asSynthetic)
+
+        //
+        // The tag type is a function from the type of variant to the type of the enum.
+        //
+        val (tpe, eff) = visitExp(exp)
```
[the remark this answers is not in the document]
mlutze: I prefer to avoid them, as they can get lost over the edge of the screen

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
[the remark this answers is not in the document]
mlutze: I also have a few changes on top of this PR and I don't want to get buried in conflicts.

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1071 @@
...
+        c.addClassConstraintsM(tconstrs, loc)
+        val resTpe = defTpe
+        val resEff = Type.Pure
+        (resTpe, resEff)
+
+      case Expr.Sig(sym, tvar, loc) =>
+        val sig = root.classes(sym.clazz).sigs(sym)
+        val (tconstrs, sigTpe) = Scheme.instantiate(sig.spec.sc, loc.asSynthetic)
+        c.unifyTypeM(tvar, sigTpe, loc)
+        c.addClassConstraintsM(tconstrs, loc)
```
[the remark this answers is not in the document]
mlutze: Maybe. I'll follow up.

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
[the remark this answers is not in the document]
magnus-madsen: CleanShot 2024-02-29 at 21 04 12@2x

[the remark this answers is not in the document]
JonathanStarup: See, struggling to get 80 width of code

image

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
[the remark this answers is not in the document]
mlutze: listed

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
mlutze: listed

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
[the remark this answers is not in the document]
mlutze: listed

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
mlutze: listed

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
mlutze: listed

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
mlutze: exp1.loc is more precise. fixed

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
mlutze: IIRC we wanted to avoid introducing another boolean variable at all costs

mlutze: listed

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
```diff
@@ -0,0 +1,1069 @@
...
+
+      case Expr.Apply(exp, exps, tvar, evar, loc) =>
+        //
+        // Determine if there is a direct call to a Def or Sig.
+        // By treating these as special cases, we can:
+        // - have better error messages (knowing the precise types of the arguments, etc)
+        // - have better performance (we don't generate unnecessary type variables)
+        //
+        val knownTarget = exp match {
+          case KindedAst.Expr.Def(sym, tvar2, loc2) =>
```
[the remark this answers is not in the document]
mlutze: this area mixes with some corrections I made on top of this PR

I'll fix this in the next PR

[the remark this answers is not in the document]
mlutze: listed

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
mlutze: opting for no further action

open a PR if you disagree

main/src/ca/uwaterloo/flix/language/phase/ConstraintGeneration.scala
mlutze: I think if I add a comment then Magnus will tell me to remove it :P

mlutze:
I've addressed all the comments. I have deferred a lot of them because when I rebase the constraint resolution PR on top of this one, I am certain I will find errors, and I want to be sure that those are problems with my translation, rather than problems with my trying to fix minor bugs.

I also am already aware of some problems here wrt associated types and effects. But I want to isolate the fixes to those problems.

magnus-madsen:
@mlutze Ready for me to do another pass then?

When will we do some of the follow up things?

mlutze:
> @mlutze Ready for me to do another pass then?
> 
> When will we do some of the follow up things? Before merging in the solver, after the rebase?

Yes.

My plan is: merge, rebase, fix new bugs if present, rebase, fix associated type bugs, rebase, fix bugs from list, rebase, merge solver, fix style stuff from the list.

Only the first rebase should be painful.

magnus-madsen (approved):

magnus-madsen:
@mlutze Merged!

PR #7326, opened 2024-03-01
fix: consider econstrs in scheme instantiation

mlutze:
depends on #7325


magnus-madsen:
@mlutze Test failures

magnus-madsen (approved):

magnus-madsen:
(transient failure)

mlutze:
can always call rerun to be sure :)

Issue #7331, opened 2024-03-03
Rename java keywords


magnus-madsen:
Rename `get/set/new` to:

```
import java_new java.io.File(String): ##java.io.File \ IO as newFile;
import java_get_field flix.test.TestClass.boolField: Bool \ IO as getField;
import java_set_field flix.test.TestClass.boolField: Unit \ IO as setField;
```

Update documentation!

PR #7338, opened 2024-03-04
refactor: change java import keywords

jaschdoc:
Closes https://github.com/flix/flix/issues/7331


jaschdoc:
I can update the parser to still support the old syntax?

magnus-madsen:
> I can update the parser to still support the old syntax?

Yes, let us allow both for now.

Can you also take a look at `ImportCompleter` and all its subcompleters? They need to be adjusted.

Ditto for the book.

magnus-madsen:
- [ ] Book
- [ ] Highlight.js
- [ ] VSCode
- [ ] Completion providers for everything related to import

magnus-madsen:
Also let us _NOT_ allow both syntaxes. My goal is to get the "pain" over this week, so by next week Herluf has a clean slate.

jaschdoc:
- [ ] Book
- [ ] Highlight.js
- [ ] VSCode
- [ ] Completion Providers

jaschdoc:
What must be updated in the vscode extension?

magnus-madsen:
> What must be updated in the vscode extension?

`Syntax.tm`

magnus-madsen:
@jaschdoc This:

https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/ImportFieldCompleter.scala#L28-L31

must also be updated. Ditto for the other Import* classes.

magnus-madsen (approved):

PR #7340, opened 2024-03-04
refactor: change module Concurrent to use regions

wine99:
`testChannel.flix` passes, but I am not sure the change is correct since I am still learning both the language itself and the compiler, and there are actually quite a few tests that fail, e.g.
image
I guess there are more things to change accordingly, but I don't know how to continue.

If someone would like to finish this refactor or help me finish, I would love to follow along so that I can learn more about the flix compiler

### DIFF
--- a/main/src/library/Concurrent/Channel.flix
+++ b/main/src/library/Concurrent/Channel.flix
@@ -79,15 +79,16 @@ mod Concurrent.Channel {
     /// structure on which select can operate.
     ///
     @Internal
-    pub enum Mpmc[a](
-        MpmcAdmin, // admin
-        MutDeque[a, Static] // elementDeque
+    pub enum Mpmc[r: Region, a: Type](
+        Region[r],
+        MpmcAdmin[r], // admin
+        MutDeque[a, r] // elementDeque
     )
 
     /// Returns the `MpmcAdmin` of `c`.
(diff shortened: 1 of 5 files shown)


magnus-madsen:
@JonathanStarup Could you take a look to help finish this? Its probably just something in Lowering

JonathanStarup:
I haven't looked too far into it yet, but I want to note that this change looks like it breaks select. From the doc comment:
```
/// -- Mpmc / MpmcAdmin split --
/// The select function must work on heterogeneous channel collections,
/// which cannot happen without existential types. This means that the
/// non-polymorphic aspects relevant to select is split into its own
/// structure on which select can operate.
```

Concretely, when you call `selectFrom` all channels have to have the same region. This is not true for all use cases, so either `MpmcAdmin` has to be non-polymorphic or we have to cheat in lowering and cast all the channels to some common region

JonathanStarup:
This is unrelated to this PR but is related to regioned channels in general #7348

magnus-madsen:
@wine99 I think the suggestion is that you will have to make your own regioned ReentrantLock.

We can easily make this PR go through due to technical issues :/

magnus-madsen:
As it turns out, we cannot yet do this refactoring.

Issue #5064, opened 2022-12-07
`MutList.flatMap`


JonathanStarup:

stetimi:
I can take a look at this.

PR #7346, opened 2024-03-06
feat: add MutList.flatMap (#5064)

stetimi:


main/src/library/MutList.flix
```diff
@@ -359,6 +359,37 @@ mod MutList {
...
+    /// Apply `f` to every element in `v` and concatenate the results.
+    ///
+    /// The result is a new mutable list.
+    ///
+    pub def flatMap(rc1: Region[r1], f: a -> MutList[b, r1] \ ef, v: MutList[a, r]): MutList[b, r1] \ { ef, r, r1 } =
+        let MutList(_, ra, rl) = v;
+        let a = deref ra;
+        let l = deref rl;
+        let bss = Array.new(rc1, l);
+        def loop(i, size)  = {
```
stetimi, line 372: An alternative way to init the array here, but I'm not sure whether using init with a side-effect is good style or not:
```
    let rsize = ref 0 @ rc1;
    def get(i) = {
        let bs = f(Array.get(i, a));
        rsize := deref rsize + MutList.length(bs);
        bs        
    };
    let bss = Array.init(rc1, get, l);
    let size = deref rsize;
```

stetimi:
Actually, maybe using the underlying Array.flatMap would be better.

JonathanStarup (approved):
Looks good to me. I think this is better than Array.flatMap since there is unused indices in MutList

magnus-madsen (approved):

magnus-madsen:
Merged. Thanks!

Issue #7329, opened 2024-03-03
Rename `new` functions to `empty` everywhere


magnus-madsen:
For example,

```
pub def new(rc: Region[r]): MutDeque[a, r] \ r =
```

becomes

```
pub def empty(rc: Region[r]): MutDeque[a, r] \ r =
```

Update documentation!

PR #7349, opened 2024-03-06
refactor: rename new to empty

jaschdoc:
- `Collection.new` -> `Collection.empty`
- `Ref.new` -> `Ref.to`
- `Random.new` -> `Random.fresh`

Closes https://github.com/flix/flix/issues/7329


jaschdoc:
Lots failures on the community build obviously

magnus-madsen:
@jaschdoc A compiler test is also failing.

main/src/library/Array.flix
```diff
@@ -408,7 +408,7 @@ mod Array {
     ///
     /// Returns the result of applying `f` to every element in `a`.
     ///
-    /// The result is a new array.
+    /// The result is an empty array.
```
magnus-madsen, line 411: @jaschdoc The search and replace here is problematic. In most cases its NOT an empty array. The documentation is trying to indicate that what is returned is new array (but it still contains elements).

magnus-madsen: I will go a head and merge this, but can you restore Array.flix and manually restore the necessary comments in a new PR?

magnus-madsen (approved):

main/src/library/Array.flix
jaschdoc: > @jaschdoc The search and replace here is problematic. In most cases its NOT an empty array. The documentation is trying to indicate that what is returned is new array (but it still contains elements).

This was deliberate hehe oops

PR #7351, opened 2024-03-07
feat: avoid redundant `IO` cast on import

### LINKED ISSUE #7333: `desugarJvmX` should not insert an unchecked_cast if the effect is already IO

JonathanStarup:
Fixes #7333

Note that there is still a cast, but its only for the type when the effect is IO


magnus-madsen (approved):

magnus-madsen:
Thanks!

PR #7359, opened 2024-03-08
feat: update `Lexer` and `TokenKind`

herluf-ba:
During finalization of the new parser some `TokenKind`s needed to be removed and others added.
Furthermore some corrections were made to the `Lexer`.


magnus-madsen:
Any impact on performance? (Since the lexer is enabled by default I need to know)

magnus-madsen:
By impact, I mean, I don't want any 10x or 100x slowdowns. I don't care about small changes.

herluf-ba:
Actually, it seems like there is! I will investigate, good thing you asked.

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -420,59 +429,79 @@ object Lexer {
...
       case _ if isKeyword("Array#") => TokenKind.ArrayHash
       case _ if isKeyword("Map#") => TokenKind.MapHash
       case _ if isKeyword("List#") => TokenKind.ListHash
       case _ if isKeyword("Vector#") => TokenKind.VectorHash
-      case _ if isKeyword("regex\"") => acceptRegex()
+      case _ if isMatch("regex\"") => acceptRegex()
       case _ if isMathNameChar(c) => acceptMathName()
       case _ if isGreekNameChar(c) => acceptGreekName()
+      case c if c.isLetter => acceptName(c.isUpper)
+      case c if DIGITS.contains(c) => acceptNumber()
```
magnus-madsen, line 477: This is very slow

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -59,8 +59,14 @@ object Lexer {
     '&' -> TokenKind.Ampersand,
     '|' -> TokenKind.Bar,
     '^' -> TokenKind.Caret,
+    '$' -> TokenKind.Dollar,
   )
 
+  /**
+   * Since Flix support hex decimals, a 'digit' can also be some select characters.
+   */
+  private val DIGITS = List('0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E', 'F', 'a', 'b', 'c', 'd', 'e', 'f')
```
magnus-madsen, line 68: Better to define this as a function and compare if the char is between two numbers.

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -420,59 +429,79 @@ object Lexer {
...
 
+  /**
+   * Check that the potential keyword is sufficiently separated, taking care not to go out-of-bounds.
+   * A keyword is separated if it is surrounded by whitespace, parenthesis, brackets or curlies.
+   * Note that __comparison includes current__.
+   */
+  private def isSeparated(keyword: String)(implicit s: State): Boolean = {
+    // TODO: We can be smarter!
+    val VALID_SEPARATORS =  List('(', ')', '[', ']', '{', '}', ',', ';', '~', '+', '-')
+    def isSep(c: Char) = c.isWhitespace || VALID_SEPARATORS.contains(c)
```
magnus-madsen, line 498: The call to contains is also likely to be slow.

magnus-madsen: I would just check if `c == ...`

herluf-ba: I think past me was trying to say that maybe the whole `VALID_SEPARATORS` idea is not the right approach.
> `// TODO: We can be smarter!`
I think I will investigate that a little, otherwise I can go for the `c == ... || c == ...` style.

magnus-madsen:
What's the current performance impact (if any?)

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
herluf-ba: We are still about 2x slower, so I want to try and squeeze a bit more!
I didn't copy the doc-comment into the snippet. The full thing is
```scala
/**
   * Check that the potential keyword is sufficiently separated, taking care not to go out-of-bounds.
   * A keyword is separated if it is surrounded by anything __but__ a character, digit or underscore.
   * Note that __comparison includes current__.
   */
  private def isSeparated(keyword: String)(implicit s: State): Boolean = {
    def isSep(c: Char) = !(c.isWhitespace || c.isLetter || c.isDigit || c == '_')
    s.src.data.lift(s.current.offset - 2).forall(isSep) && s.src.data.lift(s.current.offset + keyword.length - 1).forall(isSep)
  }
```

magnus-madsen: Ok. Is that significant for the lexer?

herluf-ba: We are talking 200ms vs 400ms on my computer. That's about 4% vs. 2%

magnus-madsen: Does it mean you want me to hold of merging?

magnus-madsen: Hmm. Yeah thats a bit in the high-end.

magnus-madsen (approved):

magnus-madsen:
!!!jar

PR #7360, opened 2024-03-08
feat: update `SyntaxTree`

herluf-ba:
During finalization of the new parser some missing `TreeKind`s were added.


main/src/ca/uwaterloo/flix/language/ast/SyntaxTree.scala
```diff
@@ -38,7 +38,7 @@ object SyntaxTree {
    * @param loc      The location that the node spans in the source file.
    * @param children The children of the node.
    */
-  case class Tree(kind: TreeKind, children: Array[Child], loc: SourceLocation)
+  case class Tree(kind: TreeKind, var children: Array[Child], var loc: SourceLocation)
```
magnus-madsen, line 41: Why are these now vars?

herluf-ba: That's for when trees are being built in the `buildTree` function. There is a lot of appending children and updating the source locations. 
I agree that they should not be `var`s. I wonder if this is a good use-case for the builder pattern?
To exemplify I am talking about using a `SyntaxTree.Builder` class like so:
```scala
object SyntaxTree {

  case class Builder private (kind: TreeKind = TreeKind.ErrorTree, children: Array[Child] = Array.empty, loc: SourceLocation = SourceLocation.Unknown) {
    def withKind(kind: TreeKind): Builder = copy(kind = kind)

    def withLoc(loc: SourceLocation): Builder = copy(loc = loc)

    def withChild(child: Child): Builder = copy(children = children :+ child)

    def build(): SyntaxTree.Tree = SyntaxTree.Tree(kind, children, loc)
  }
```
What do you think?

magnus-madsen: No builders. This is not OO. 

Can you show the code that uses mutation?

herluf-ba: It's in this function: https://github.com/flix/flix/blob/33fd427d33b8a5c103b7f74e52601cf6480cdff1/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala#L180

The mutations are just simple appends (or reassignment in the case of `loc`):
```scala
// line 220
stack.last.children = stack.last.children :+ child
// line 225
stack.last.children = stack.last.children :+ SyntaxTree.Child.TokenChild(token)
```

magnus-madsen: I suggest to leave as-is for now.

magnus-madsen (approved):

PR #7362, opened 2024-03-08
refactor: avoid using Scheme in ClassEnvironment

mlutze:


mlutze:
@magnus-madsen The only tests failing are associated types tests. Are we ok with letting associated types be broken during transition?

magnus-madsen:
> @magnus-madsen The only tests failing are associated types tests. Are we ok with letting associated types be broken during transition?

Yes, I think so. Where are they used? We could mark those tests are ignore.

magnus-madsen:
Can we (temporarily) remove associated types from the relevant TCs?

mlutze:
It's only the tests specifically targeting associated types, so disabling them for now should be sufficient.

magnus-madsen (approved):

PR #7369, opened 2024-03-10
feat: resolve associated effects in EffUnification

mlutze:
extracted from #6499

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/unification/EffUnification.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/unification/EffUnification.scala
@@ -124,11 +124,11 @@ object EffUnification {
         case (_, Type.Cst(TypeConstructor.Error(_), _)) =>
           return Ok((Substitution.empty, Nil))
 
-        case (t1@Type.Var(x, _), t2) if renv0.isFlexible(x) && !t2.typeVars.contains(t1) =>
-          return Ok(Substitution.singleton(x, t2), Nil)
-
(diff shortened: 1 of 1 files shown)


magnus-madsen:
!!!jar

main/src/ca/uwaterloo/flix/language/phase/unification/EffUnification.scala
```diff
@@ -127,26 +129,100 @@ object EffUnification {
...
+          lookupOrSolve(t1, t2, renv)
+        }
+
+        substRes.map {
+          case subst0 =>
+            // replace the temporary variables with the assoc types they represent
+            val subst1 = Substitution(backSubst) @@ subst0
+            // remove the temporary variables from the substitution
+            val subst = backSubst.keys.foldLeft(subst1)(_.unbind(_))
+            (subst, Nil)
```
magnus-madsen, line 178: Would we want to remove some bindings from subst? (e.g. the bindings for variables we no longer care about).

magnus-madsen (approved):

main/src/ca/uwaterloo/flix/language/phase/unification/EffUnification.scala
mlutze: Isn't that what line 177 does?

magnus-madsen: Aaah. Something about forests and trees.

Issue #7368, opened 2024-03-10
Remove usages of `:=` for references


magnus-madsen:
- [ ] Replace usages of `:=` with `Ref.put`.
- [ ] Introduce `$REF_ASSIGN$` and use it to implement `Ref.put`.

The `ref` and `deref` keywords can stay for now. I am just trying to get rid of `:=`.

magnus-madsen:
I have decided to wait with this. Essentially everyone was nice to upgrade their code, so I don't want to break it again. Moreover, it seems people have several comments about how to implement structs, so let us let this rest for a bit.

PR #7371, opened 2024-03-10
refactor: remove `:=` stynax

jaschdoc:
Closes https://github.com/flix/flix/issues/7368

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser.scala
@@ -632,7 +632,7 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
   // Expressions                                                             //
   /////////////////////////////////////////////////////////////////////////////
   def Expression: Rule1[ParsedAst.Expression] = rule {
-    Expressions.Assign
+    Expressions.InstanceOf
   }
 
   def ExpressionEOI: Rule1[ParsedAst.Expression] = rule {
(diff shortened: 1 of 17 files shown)


jaschdoc:
Only community build failing

magnus-madsen:
> Also removed a lot of `deref` and `ref` keywords (but not all)
> 
> Closes #7368

Ah gosh. I wish you had not because ref and deref are not leaving. The problem is that we need `:=` for structs etc. But with ref and deref there is no issue.

jaschdoc:
> > Also removed a lot of `deref` and `ref` keywords (but not all)
> > Closes #7368
> 
> Ah gosh. I wish you had not because ref and deref are not leaving. The problem is that we need `:=` for structs etc. But with ref and deref there is no issue.

Oh oh...

jaschdoc:
> > Also removed a lot of `deref` and `ref` keywords (but not all)
> > Closes #7368
> 
> Ah gosh. I wish you had not because ref and deref are not leaving. The problem is that we need `:=` for structs etc. But with ref and deref there is no issue.

I guess they are easy to search/replace. I could do it for the entire library and tests?

jaschdoc:
RFR

magnus-madsen:
@jaschdoc Suggestions:

- We can still merge this on, but let us leave the "old syntax" so as not to cause breakages.
- Can we update https://doc.flix.dev/references.html so that it uses fresh/get/put etc.

Thanks!

magnus-madsen (approved):

magnus-madsen:
Thanks-- now life will be easier in the future.

PR #7376, opened 2024-03-11
refactor: rename `CallType` -> `ExpPosition`

JonathanStarup:
Related to# 7183

Also reames
- `TailCall` -> `Tail`
- `NonTailCall` -> `NonTail`


magnus-madsen:
Why is this name better? Does it not communicate less information?

JonathanStarup:
I want to add it do `Do` and `TryWith` which are not function calls

magnus-madsen (approved):

PR #7379, opened 2024-03-11
style: remove dead code

jaschdoc:


magnus-madsen:
I think this code is trying to communicate some intent, even if it is not too clear.

PR #7380, opened 2024-03-11
chore: synchronize BDDFormulaAlg

mlutze:
extracted from #6499. This will certainly cause a slowdown. But maybe adding it separately lets us quantify that slowdown?


mlutze:
!!!rerun

mlutze:
Just looking at the run time for tests in CI, the performance difference does not appear significant.

magnus-madsen (approved):

PR #7381, opened 2024-03-11
refactor: make `Validation.HardFailure` private

jaschdoc:

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/Bootstrap.scala
+++ b/main/src/ca/uwaterloo/flix/api/Bootstrap.scala
@@ -17,12 +17,13 @@ package ca.uwaterloo.flix.api
 
 import ca.uwaterloo.flix.api.Bootstrap.{getArtifactDirectory, getManifestFile, getPkgFile}
 import ca.uwaterloo.flix.language.CompilationMessage
+import ca.uwaterloo.flix.language.errors.Unrecoverable
 import ca.uwaterloo.flix.language.phase.HtmlDocumentor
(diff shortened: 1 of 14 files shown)


jaschdoc:
It seems like the design is too rigid (and maybe a bit clunky / restrictive). Take `Validation.toHardResult`. It collapses all errors into `Result.Err`, but the only way we are able to construct a `Validation` is by using the smart constructors which require the marker traits `Recoverable` or `Unrecoverable` but then the `toHardResult` function collapses both `Recoverable` and `Unrecoverable` into the same collection so now it is hard to convert back.
Why do we do this? Well, we want to pattern match on the outcome of `Validation` throughout the compiler. One way to avoid this is to allow pattern matching on `Validation` but this is exactly the opposite goal of what we are trying to achieve here. One way to do this is to avoid calling `toResult` before the compiler has to output something to the user and then only compose `Validation`s internally. But then the problem arises that we may want to convert `SoftFailure` to `HardFailure` or `Success` to `HardFailure`.
I'm not sure but I think we can use `PartialFunction` in some of the functions for `Validation` to handle some of the weird cases that arise from pattern matching at the call site. Currently, it feels like the caller has to handle some of the edge cases that really only matter to `Validation`.
I'm not sure about all of this but I will keep chipping away at the problem, but my hunch is that the design has some flaws that need to be fixed before this PR can be merged.

magnus-madsen:
Lets discuss offline.

magnus-madsen:
Where does it happen that we need to go "in and out of" Validation?

PR #7389, opened 2024-03-12
chore: rename uses of keywords `from`, `query`, `not` and `yield` from std. lib.

herluf-ba:
This PR cleans up missed refactors from the standard library and its tests.
- rename uses of keywords `from`, `query`, `not` and `yield`
- remove test files for `class` that are duplicates of `trait` tests.
- change uses if double unary negation to include parenthesis. `--1` -> `-(-1)`

**Note** that there is one controversial change! In `Prelude.flix` debug printing functions are defined. But since `debug`, `debug!` and `debug!!` are keywords this now poses a problem. Removing the definitions for `Prelude.flix` does not seem to cause any issues so I suggest we do just that. But I would like some feedback, since I want to be careful not to break anything.


magnus-madsen:
> **Note** that there is one controversial change! In `Prelude.flix` debug printing functions are defined. But since `debug`, `debug!` and `debug!!` are keywords this now poses a problem. Removing the definitions for `Prelude.flix` does not seem to cause any issues so I suggest we do just that. But I would like some feedback, since I want to be careful not to break anything.

This was an ugly hack anyway.

magnus-madsen (approved):

PR #7391, opened 2024-03-13
chore: remove `debug` functions

magnus-madsen:


JonathanStarup:
😮

JonathanStarup:
I guess not having these means that `debug` is just a regular keyword like `ref` - you also cannot use `ref` as a function

JonathanStarup (approved):

magnus-madsen:
Yes

magnus-madsen:
It was an ugly hack anyway

PR #7400, opened 2024-03-14
chore: rename defs with keyword names in examples

magnus-madsen:


JonathanStarup (approved):
Are they not used anywhere?

magnus-madsen:
> Are they not used anywhere?

Nope, its an example.

PR #7401, opened 2024-03-14
refactor: more rename `class` to `trait`

magnus-madsen:


magnus-madsen:
I know there is more class to trait renaming to do, but this is just for the parser.

mlutze (approved):

PR #7408, opened 2024-03-17
docs: add datalog-based subtyping example

jaschdoc:


jaschdoc:
@magnus-madsen do you want the return types of the functions to be datalog fragments? I.e., `methodsOf` returns `#{...}` instead of `Vector[String]`?

examples/larger-examples/datalog/subtyping.flix
```diff
@@ -0,0 +1,143 @@
+mod Subtyping {
+
+    type alias Types[r : SchemaRow] = #{ Class(String), Interface(String), ClassExtends(String, String), ClassImplements(String, String), InterfaceExtends(String, String), Method(String, String) | r }
+
+    def open(types: #{ Class(String), Interface(String), ClassExtends(String, String), ClassImplements(String, String), InterfaceExtends(String, String), Method(String, String) }): Types[r] = types
+
+    def subtypeRules(): #{ IsSubtype(String, String), Class(String), Interface(String), ClassExtends(String, String), ClassImplements(String, String), InterfaceExtends(String, String) | r } = #{
```
magnus-madsen, line 7: I think it would be nice if we could have a function `subType` or whatever that takes some input as lists and returns a Datalog program which contains a relation SubType where other relations, e.g. Class, ClassExtends, etc. are hidden using local predicates / rho abstraction. Then the types become much smaller. Does it make sense?

magnus-madsen: So instead of all this type aliasing and use of open, I want to use https://doc.flix.dev/local-predicates.html where one function does the heavy computation of SubType and then the rest of the functions simply use it.

magnus-madsen:
```
def subType(classes: List[...], classExtends: List[...]): #{SubType(...)} = #(SubType) -> #{ ... }
```

jaschdoc: Something like this?
```
    pub def subtypes(classes: List[String], interfaces: List[String], classExtensions: List[(String, String)], implementations: List[(String, String)], interfaceExtensions: List[(String, String)], methods: List[String]): #{ IsSubtype(String, String)| r } =
        let cs    = inject classes             into Class;
        let is    = inject interfaces          into Interface;
        let ce    = inject classExtensions     into ClassExtends;
        let impls = inject implementations     into ClassImplements;
        let ies   = inject interfaceExtensions into InterfaceExtends;
        let ms    = inject methods             into Method;
        let rules = #(IsSubtype) -> #{
            IsSubtype(sub, super) :- Class(sub), Class(super), if sub == super.
            IsSubtype(sub, super) :- Interface(sub), Interface(super), if sub == super.
            IsSubtype(sub, super) :- Class(sub), Class(super), ClassExtends(sub, super).
            IsSubtype(sub, super) :- Class(sub), Interface(super), ClassImplements(sub, super).
            IsSubtype(sub, super) :- Interface(sub), Interface(super), InterfaceExtends(sub, super).
            IsSubtype(sub, super) :- IsSubtype(sub, a), IsSubtype(a, super).
        };
        solve cs, is, ce, impls, ies, ms, rules project IsSubtype

```

magnus-madsen: Yes, but the rules look strange. You can just do 

ISSubtype(x, x) :- Class(x)

magnus-madsen: In addition to subtype, I would also like something like HasMethod(className, methodName).

magnus-madsen: But yes, this looks like the right direction. So we have one method that computes some general information and then we have smaller methods that call this big method and answer specify queries.

jaschdoc:
@magnus-madsen it works with rho abstraction now

examples/larger-examples/datalog/subtyping.flix
```diff
@@ -0,0 +1,176 @@
...
+            Subtype(sub, super) :- Interface(sub), Interface(super), InterfaceExtends(sub, super).
+            Subtype(sub, super) :- Subtype(sub, a), Subtype(a, super).
+
+            AvailableMethod(class, method) :- Method(t, method), Subtype(class, t).
+            NonSupertype(class)            :- Class(class), Class(otherClass), not Subtype(otherClass, class), if class != otherClass.
+            ExtendsFinalClass(sub, super)  :- FinalClass(super), Subtype(sub, super), if sub != super.
+        };
+        #(Subtype, AvailableMethod, NonSupertype, ExtendsFinalClass) -> rules <+> toPredicates(program)
+
+    def toPredicates(program: Program): #{ Class(String), FinalClass(String), Interface(String), ClassExtends(String, String),  InterfaceExtends(String, String), Implementation(String, String), Method(String, String) | r } =
```
magnus-madsen, line 28: I would like this to be inlined into the above function.

examples/larger-examples/datalog/subtyping.flix
```diff
@@ -0,0 +1,176 @@
...
+        classes = Vector[String],
+        finalClasses = Vector[String],
+        interfaces = Vector[String],
+        implementations = Vector[(String, String)],
+        classExtends = Vector[(String, String)],
+        interfaceExtends = Vector[(String, String)],
+        methods = Vector[(String, String)]
+    }
+
+    def subtypes(program: Program): #{ Subtype(String, String), AvailableMethod(String, String), NonSupertype(String), ExtendsFinalClass(String, String) | r } =
```
magnus-madsen, line 13: We could rename this function to `classHierarchy` or `classHierarchyAnalysis`

magnus-madsen, line 13: program -> p (single variable name).

examples/larger-examples/datalog/subtyping.flix
```diff
@@ -0,0 +1,176 @@
+mod Subtyping {
+
+    type alias Program = {
+        classes = Vector[String],
+        finalClasses = Vector[String],
+        interfaces = Vector[String],
+        implementations = Vector[(String, String)],
```
magnus-madsen, line 7: What is implementations? Maybe its classImplements?

examples/larger-examples/datalog/subtyping.flix
```diff
@@ -0,0 +1,176 @@
...
+            Subtype(x, x)       :- Class(x).
+            Subtype(x, x)       :- Interface(x).
+            Subtype(sub, super) :- Class(sub), Interface(super), Implementation(sub, super).
+            Subtype(sub, super) :- Class(sub), Class(super), ClassExtends(sub, super).
+            Subtype(sub, super) :- Interface(sub), Interface(super), InterfaceExtends(sub, super).
+            Subtype(sub, super) :- Subtype(sub, a), Subtype(a, super).
+
+            AvailableMethod(class, method) :- Method(t, method), Subtype(class, t).
+            NonSupertype(class)            :- Class(class), Class(otherClass), not Subtype(otherClass, class), if class != otherClass.
+            ExtendsFinalClass(sub, super)  :- FinalClass(super), Subtype(sub, super), if sub != super.
```
magnus-madsen, line 24: I want to move this rule into a separate method. Then that method can be given finalClasses as an argument.

examples/larger-examples/datalog/subtyping.flix
```diff
@@ -0,0 +1,176 @@
...
+        let ce    = inject program.classExtends     into ClassExtends;
+        let ies   = inject program.interfaceExtends into InterfaceExtends;
+        let ms    = inject program.methods          into Method;
+        cs <+> fcs <+> is <+> ce <+> impls <+> ies <+> ms
+
+    pub def subtypesDistinct(types: Program): Vector[(String, String)] =
+        let v = query subtypes(types) select (x, y) from Subtype(x, y);
+        Vector.filter(match (x, y) -> x != y, v)
+
+    pub def methodsOf(tpe: String, types: Program): Vector[String] =
```
magnus-madsen, line 42: I don't want this double layering. Lets just have /one/ function like hasMethod.

examples/larger-examples/datalog/subtyping.flix
```diff
@@ -0,0 +1,176 @@
...
+    pub def hasMethod(tpe: String, method: String, types: Program): Bool =
+        not (methodsOf(tpe, types) |> Vector.filter(Eq.eq(method)) |> Vector.isEmpty)
+
+    pub def isNonSupertype(tpe: String, types: Program): Bool =
+        not (nonSupertypes(types) |> Vector.filter(Eq.eq(tpe)) |> Vector.isEmpty)
+
+    pub def extendsFinalClass(tpe: String, types: Program): Bool =
+        not (finalClassExtensions(types) |> Vector.filter(match (sub, _) -> sub == tpe) |> Vector.isEmpty)
+
+
```
magnus-madsen, line 64: Dropping all of the above queries, I think the following functions would be useful:

- `def hasField(p: Program, field: String): Bool`
- `def hasMethod(p: Program, method: String): Bool`
- `def illegalNonThrowable(p: Program, throws: Vector[...]): Vector[...]` // can only throw subtypes of Throwable.
- `def illegalExtendsFinalClass(p: Program, finalClasses: Vector[...]): Vector[...]` // cannot extend final classes.

magnus-madsen:
I expected to take some iterations on this, so don't add comments yet. You can just do some more programming.

examples/larger-examples/datalog/subtyping.flix
```diff
@@ -0,0 +1,301 @@
+mod Subtyping {
```
magnus-madsen, line 1: Good progress so far. I like where this is going.

examples/larger-examples/datalog/subtyping.flix
```diff
@@ -0,0 +1,301 @@
...
+        let rules = #{
+            Subtype(x, x)       :- Class(x).
+            Subtype(x, x)       :- Interface(x).
+            Subtype(sub, super) :- Class(sub), Interface(super), ClassImplements(sub, super).
+            Subtype(sub, super) :- Class(sub), Class(super), ClassExtends(sub, super).
+            Subtype(sub, super) :- Interface(sub), Interface(super), InterfaceExtends(sub, super).
+            Subtype(sub, super) :- Subtype(sub, a), Subtype(a, super).
+
+            AvailableField(class, field)        :- Field(t, field), Subtype(class, t).
+            AvailableMethod(class, method)      :- ClassMethod(t, method), Subtype(class, t).
```
magnus-madsen, line 32: The (new) idea is that `AvailableField` and `AvailableMethod` captures fields/methods which must be available on a type. But since a type can be abstract (i.e. interface or class) then the programmer may have forgot to specify a method. We capture the methods _actually_ available in the ConcreteMethod relation:

```
ConcreteMethod(class, method) :- Class(class), ClassMethod(class, method).
ConcreteMethod(sub, method) :- Class(sub), Subtype(sub, super), ConcreteMethod(super, method).
```

We expose `(Subtype, AvailableField, AvailableMethod, ConcreteMethod)` as the return type. Drop  `NonImplementedMethod`.

Later, before `illegalFinalClasses`, we can add something like:

```
def missingMethods(...): Vector[...] = 
  inject p.interfaces into Interface;
 ... #{  MissingMethod(class, method) :- AvailableMethod(class, method), not ConcreteMethod(class, method), not Interface(class).
     }
// Note: Here you have to inject "Interface" _again_, but that is acceptable. 
```

examples/larger-examples/datalog/subtyping.flix
```diff
@@ -0,0 +1,301 @@
...
+        query classHierarchyAnalysis(p) select (tpe, method) from NonImplementedMethod(tpe, method)
+
+    pub def isSubtype(p: Program, subtype: String, possibleSupertype: String): Bool =
+        let res = query classHierarchyAnalysis(p) select subtype from Subtype(subtype, possibleSupertype);
+        not Vector.isEmpty(res)
+
+    pub def illegalFinalClasses(p: Program, finalClasses: Vector[String]): Vector[String] =
+        let fcs  = inject finalClasses into FinalClass;
+        let rule = #{
+            ExtendsFinalClass(sub, super)  :- FinalClass(super), Subtype(sub, super), if sub != super.
```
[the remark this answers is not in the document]
magnus-madsen: Technically its not the final class thats the problem. So lets call it something like `illegalFinalClassExtension`

examples/larger-examples/datalog/subtyping.flix
magnus-madsen: Lets class the module/file something like ClassHierarchyAnalysis

examples/larger-examples/datalog/subtyping.flix
```diff
@@ -0,0 +1,301 @@
...
+
+    type alias Program = {
+        classes = Vector[String],
+        interfaces = Vector[String],
+        classImplements = Vector[(String, String)],
+        classExtends = Vector[(String, String)],
+        interfaceExtends = Vector[(String, String)],
+        fields = Vector[(String, String)],
+        classMethods = Vector[(String, String)],
+        interfaceSignatures = Vector[(String, String)]
```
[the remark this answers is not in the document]
magnus-madsen: We can actually just call it `Final` since only classes can be final anyway.

examples/larger-examples/datalog/subtyping.flix
```diff
@@ -0,0 +1,301 @@
...
+        query fcs, rule, classHierarchyAnalysis(p) select tpe from ExtendsFinalClass(tpe, _)
+
+    pub def illegalNonThrowable(p: Program, throws: Vector[String]): Vector[String] =
+        let ths = inject throws into Throws;
+        let rule = #{
+            IllegalThrowable(t) :- Throws(t), not Subtype(t, "java.lang.Throwable").
+        };
+        query ths, rule, classHierarchyAnalysis(p) select tpe from IllegalThrowable(tpe)
+
+
```
magnus-madsen, line 64: I have an idea for another extension, but maybe its best to wait until after the above refactorings.

But if we extended ConcreteMethod to ConcreteMethod(class, method, source), then we could check that every class that specifies an equals method also remembers to implement a hashCode method.

examples/larger-examples/datalog/subtyping.flix
```diff
@@ -0,0 +1,301 @@
...
+        };
+        query fcs, rule, classHierarchyAnalysis(p) select tpe from ExtendsFinalClass(tpe, _)
+
+    pub def illegalNonThrowable(p: Program, throws: Vector[String]): Vector[String] =
+        let ths = inject throws into Throws;
+        let rule = #{
+            IllegalThrowable(t) :- Throws(t), not Subtype(t, "java.lang.Throwable").
+        };
+        query ths, rule, classHierarchyAnalysis(p) select tpe from IllegalThrowable(tpe)
+
```
magnus-madsen, line 63: Add another method to compute essentially:

IllegalMultipleInheritance(class) :- Class(super1), Class(super2), Subtype(class, super1), Subtype(class, super2), if super1 != super2.

jaschdoc: I'm not sure this one makes much sense outside of `classHierarchyAnalysis` since what we intuitively want is
```
IllegalMultipleInheritance(class, super1, super2) :- Class(class), Class(super1), Class(super2), ClassExtends(class, super1), ClassExtends(class, super2), if class != super1 and class != super2 and super1 != super2.
```
Then we will still get duplicates since if `A` extends `B, C` then `A` extends `C, B` but this way we exclude all combinations of transitively inheriting from `Object` if that makes sense.

Here is the output of a test:
```
Expected: Vector#{(au.content.MultipleInheritance, au.content.Super1, au.content.Super2)}
Actual:   Vector#{(au.content.MultipleInheritance, au.content.Super1, au.content.Super2), (au.content.MultipleInheritance, au.content.Super1, java.lang.Object), (au.content.MultipleInheritance, au.content.Super2, au.content.Super1), (au.content.MultipleInheritance, au.content.Super2, java.lang.Object), (au.content.MultipleInheritance, java.lang.Object, au.content.Super1), (au.content.MultipleInheritance, java.lang.Object, au.content.Super2), (au.content.SpecialArrayList, java.lang.Object, java.util.ArrayList), (au.content.SpecialArrayList, java.util.ArrayList, java.lang.Object), (java.lang.Exception, java.lang.Object, java.lang.Throwable), (java.lang.Exception, java.lang.Throwable, java.lang.Object)}
```

examples/larger-examples/datalog/subtyping.flix
jaschdoc: The class on which `method` is implemented?

examples/larger-examples/datalog/subtyping.flix
magnus-madsen: OK. Agreed. We do not need this.

examples/larger-examples/datalog/class-hierarchy-analysis.flix
```diff
@@ -0,0 +1,388 @@
...
+        let ths = inject p.throws into Throws;
+        let rule = #{
+            IllegalThrowable(t) :- Throws(t), not Subtype(t, "java.lang.Throwable").
+        };
+        query ths, rule, classHierarchyAnalysis(p) select tpe from IllegalThrowable(tpe)
+
+    ///
+    /// Finds and returns a vector of tuples representing illegal multiple inheritance in the program `p`.
+    ///
+    pub def illegalMultipleInheritance(p: Program): Vector[(String, String, String)] =
```
magnus-madsen, line 127: Let us leave it as-is.

examples/larger-examples/datalog/class-hierarchy-analysis.flix
```diff
@@ -0,0 +1,388 @@
...
+            Subtype(sub, super) :- Interface(sub), Interface(super), InterfaceExtends(sub, super).
+            Subtype(sub, super) :- Subtype(sub, a), Subtype(a, super).
+
+            AvailableField(class, field) :- Field(super, field), Subtype(class, super).
+
+            AvailableMethod(class, method) :- Class(class), ClassMethod(super, method), Subtype(class, super).
+            AvailableMethod(class, method) :- Class(class), AbstractMethod(super, method), Subtype(class, super).
+
+            ConcreteMethod(class, method, class) :- Class(class), ClassMethod(class, method).
+            ConcreteMethod(sub, method, source)  :- Class(sub), Subtype(sub, super), ConcreteMethod(super, method, source).
```
magnus-madsen, line 67: Proposal:

ConcreteMethod -> InheritedMethod

And we change the last rule:

```
InheritedMethod(sub, method, source)  :- Class(sub), Subtype(sub, super), InheritedMethod(super, method, source), not ClassMethod(sub, method, sub).
```

Will that work? Such that we account for methods that are overriden?

examples/larger-examples/datalog/class-hierarchy-analysis.flix
```diff
@@ -0,0 +1,388 @@
...
+
+    ///
+    /// Returns a vector of pairs where each pair `(class, method)`
+    /// represents that an implementation of `method` is missing from
+    /// `class` in the program `p`.
+    ///
+    pub def missingMethods(p: Program): Vector[(String, String)] =
+        let is   = inject p.interfaces into Interface;
+        let rule = #{
+            SimpleConcreteMethod(class, method) :- ConcreteMethod(class, method, _).
```
magnus-madsen, line 99: SimpleConcreteMethod -> ConcreteMethod (now that this name is freed up)

examples/larger-examples/datalog/class-hierarchy-analysis.flix
```diff
@@ -0,0 +1,388 @@
...
+            MissingMethod(class, method) :- AvailableMethod(class, method), not SimpleConcreteMethod(class, method), not Interface(class).
```
magnus-madsen, line 100: I would swap `not Interface(class)` for `Class(class`). Seems more reasonable.

examples/larger-examples/datalog/subtyping.flix
magnus-madsen: I think you got it right.

magnus-madsen (approved):

PR #7410, opened 2024-03-18
fix: add missing source lines to errors

JonathanStarup:
Maybe we should abbreviate parts of the expression to avoid this issue


magnus-madsen (approved):

magnus-madsen:
> Maybe we should abbreviate parts of the expression to avoid this issue

There is a ticket somewhere to make this stuff consistent. Its assigned to Jakob (eventually).

PR #7411, opened 2024-03-18
fix: Try-Catch typing

JonathanStarup:


JonathanStarup:
@mlutze I hope this doesn't conflict with anything, but I guess it should be fixed in the generator?

mlutze:
> @mlutze I hope this doesn't conflict with anything, but I guess it should be fixed in the generator?

It's likely also broken on my branch. But the test will catch it.

main/src/ca/uwaterloo/flix/language/phase/TypeInference.scala
```diff
@@ -889,7 +889,10 @@ object TypeInference {
       case KindedAst.Expr.TryCatch(exp, rules, loc) =>
         val rulesType = rules map {
           case KindedAst.CatchRule(sym, clazz, body) =>
-            visitExp(body)
+            for {
+              _ <- unifyTypeM(sym.tvar, Type.mkNative(clazz, sym.loc), sym.loc)
```
magnus-madsen, line 893: What about being an instance of Throwable?

magnus-madsen: Does this also need to be fixed in ConstraintGeneration?

JonathanStarup: > What about being an instance of Throwable?

I don't know how to do that. This PR still improves things though

magnus-madsen: CC @mlutze

magnus-madsen: I guess your test will help Matt make the required refactor.

JonathanStarup: Ive corrected the constraint generator

PR #7412, opened 2024-03-18
chore: move debugging facilities to Debug.scala

mlutze:
extracted from #6499


magnus-madsen:
Any reason not to make it slightly nicer? E.g. threadsafe by passing the relevant symbol and using that as the name?

mlutze:
> Any reason not to make it slightly nicer? E.g. threadsafe by passing the relevant symbol and using that as the name?

I think to do that properly I need threadlocal stuff and it becomes a pain for just a debugging helper.

The intended usage of this is to record constraints of _one_ definition throughout resolution.

magnus-madsen (approved):

PR #7421, opened 2024-03-19
feat: Normalize records in monomorpher

JonathanStarup:
Unique type representation of records


JonathanStarup:
#7425 merged instead

PR #7422, opened 2024-03-19
feat: add AtomicOp cases to verifier

felix-berg:


main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -235,6 +242,68 @@ object Verifier {
           check(expected = MonoType.Enum(sym.enumSym))(actual = t1, loc)
           tpe
 
+        case AtomicOp.ArrayLength =>
+          val List(arrt) = ts
```
JonathanStarup, line 246: the variable name makes sense, but for consistency's sake, these variables should just be called t or t1 t2 t3 ...

This goes for all  the cases

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -53,6 +53,13 @@ object Verifier {
         println(s"  tpe1 = $tpe1")
         println(s"  tpe2 = $tpe2")
         println()
+
+      case MismatchedShape(tpe, exp, loc) =>
+        println(s"Mismatched shape near ${loc.format}")
+        println()
+        println(s"  tpe      = $tpe")
```
JonathanStarup, line 60: Switch these two lines to be consistent with UnexpectedType

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -235,6 +242,68 @@ object Verifier {
           check(expected = MonoType.Enum(sym.enumSym))(actual = t1, loc)
           tpe
 
+        case AtomicOp.ArrayLength =>
+          val List(arrt) = ts
+          arrt match {
+            case MonoType.Array(_) => check(expected = MonoType.Int32)(actual = tpe, loc)
+            case _ => throw MismatchedShape(tpe, s"An Array", loc)
```
JonathanStarup, line 249: I think its more aligned with the other errors just to write `"Array"`

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -235,6 +242,68 @@ object Verifier {
...
+          check(expected = MonoType.Int32)(actual = lent, loc)
+          check(expected = MonoType.Array(elmt))(actual = tpe, loc)
+
+        case AtomicOp.ArrayLit =>
+          tpe match {
+            case MonoType.Array(elmt) =>
+              ts.foreach(t => check(expected = elmt)(actual = t, loc))
+            case _ => throw MismatchedShape(tpe, s"An Array", loc)
+          }
+          tpe
```
JonathanStarup, line 263: I think it looks better with tpe inside the array case, then there's no exception "in the middle"

This also goes for some following cases

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -235,6 +242,68 @@ object Verifier {
...
+
+        case AtomicOp.ArrayLoad =>
+          val List(arrt, idxt) = ts
+          check(expected = MonoType.Int32)(actual = idxt, loc)
```
JonathanStarup, line 267: Even though its not strictly necessary, it's nice to check types from left-to-right to aid debugging

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -235,6 +242,68 @@ object Verifier {
...
+            case MonoType.Array(elmt) => check(expected = elmt)(actual = tpe, loc)
+            case _ => throw MismatchedShape(tpe, s"An Array", loc)
+          }
+          tpe
+
+        case AtomicOp.ArrayStore =>
+          val List(arrt, idxt, expt) = ts
+          check(expected = MonoType.Int32)(actual = idxt, loc)
+          arrt match {
+            case MonoType.Array(elmt) => check(expected = elmt)(actual = expt, loc)
```
JonathanStarup, line 278: you use `check` everywhere, but for nice errors its better to use `checkEq` in cases like this where both types are unknown - we don't know what the correct type is, just that they have to agree

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -235,6 +242,68 @@ object Verifier {
...
+          check(expected = MonoType.Lazy(elm))(actual = tpe, loc)
+
+        case AtomicOp.Force =>
+          val List(laz) = ts
+          laz match {
+            case MonoType.Lazy(elm) => check(expected = elm)(actual = tpe, loc)
+            case _ => throw MismatchedShape(tpe, s"Lazy", loc)
+          }
+
+        case AtomicOp.Tuple => check(expected = MonoType.Tuple(ts))(actual = tpe, loc)
```
JonathanStarup, line 305: `AtomicOp.Index(idx: Int)` is the inverse of tuple (extract the idx'th value from the tuple), so I think you should add that. It's a bit different from the rest because you have to look into the tuple type of the exp but give it a shot :)

JonathanStarup (commented):
The typing rules looks good, there's just some style and consistency things :)

magnus-madsen:
@felix-berg You can use "Resolve conversation" to mark the tasks you have cleared.

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -53,6 +53,13 @@ object Verifier {
         println(s"  tpe1 = $tpe1")
         println(s"  tpe2 = $tpe2")
         println()
+
+      case MismatchedShape(tpe, exp, loc) =>
```
JonathanStarup, line 57: `exp` is a bit confusing in a compiler context, `expected` is better

JonathanStarup (approved):

magnus-madsen:
@felix-berg Merged. Congratulations on landing your first PR 🚀

PR #7423, opened 2024-03-20
feat: add missing superclass code action

stetimi:
Adds a quickfix for a missing superclass in a derivation.
E.g. 
before
```
enum Direction with Order {case N, E, S, W}

Missing super class instance 'Eq'.(Instance Error)
```
after
```enum Direction with Order, Eq {case N, E, S, W}```


magnus-madsen (approved):

magnus-madsen:
Thanks!

PR #7425, opened 2024-03-20
feat: Normalize records in monomorpher

JonathanStarup:
Alternative to #7421

This keeps  `apply` simpler but has n^2 time complexity on the length of records


magnus-madsen:
I like it better 👍

magnus-madsen (approved):

PR #7437, opened 2024-03-21
feat: change `Xverify` to `xnoverify`

JonathanStarup:


magnus-madsen:
It should still be Xnoverify. Its a non-official option.

JonathanStarup:
i was conceptually confused since we are pre 1.0, couldnt official options also just change?

magnus-madsen:
> i was conceptually confused since we are pre 1.0, couldnt official options also just change?

Well, ideally not. For example, why why would e.g `build` change?

magnus-madsen (approved):

PR #7438, opened 2024-03-21
feat: replace IFDS example with a polymorphic modular version

olhotak:
Replace IFDS example with a polymorphic modular version modelled on new IDE example.


magnus-madsen (approved):

magnus-madsen:
Thanks @olhotak

PR #7441, opened 2024-03-22
feat: lazy arg iteration

JonathanStarup:
very minor optimization, @magnus-madsen you can just close this if its too small


magnus-madsen:
To me this would just cause more confusion because data is duplicated (even if because of immutability it is the same).

JonathanStarup:
could also be called argsIterator to be clearer, but i didn't want to produce an actual iterator

Issue #7440, opened 2024-03-22
Improve Verifier performance


magnus-madsen:
... it is current having a major effect on performance.

I think there most be (a) some unintended inefficiencies and (b) lack of parallelism.

magnus-madsen:
@JonathanStarup CC

JonathanStarup:
I don't think there's any inefficiency except sequential defs checking. The slower time from the other traversal phases is probably from all the repeated equality checking on types.

JonathanStarup:
There are some unused variables, which shouldn't be there, but none of them are bugs

magnus-madsen:
![image](https://github.com/flix/flix/assets/2287953/3ae92016-a736-4c8e-8d01-0122d6cf7a76)

If cost cannot be improved, we may have to make it optional again. Still its more expensive than Monomorpher?

JonathanStarup:
I'm hoping ParOps will help enough but otherwise yeah, its just a traversal and equality checks

magnus-madsen:
Traversal is very fast-- look at the other phases. Maybe you can attach a debugger and see?

You can just run the verifier phase 1000 times to measure.

JonathanStarup:
good enough now?

magnus-madsen:
Hmm. I clock it at 100ms-- higher than most frontend and backend phases.

magnus-madsen:
I think we should still investigate. Someone needs to connect a profiler.

magnus-madsen:
We now use RunVerifiers.scala

PR #7442, opened 2024-03-22
feat: Verifier in parallel

JonathanStarup:
Relates to  #7440

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
@@ -31,6 +31,7 @@ object Verifier {
       root
     } else {
       ParOps.parMap(root.defs.values)(visitDef(_)(root))
+      root
     }
   }
 
--- a/main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
@@ -27,11 +27,11 @@ import ca.uwaterloo.flix.util.{InternalCompilerException, ParOps}
 object Verifier {
 
   def run(root: Root)(implicit flix: Flix): Root = flix.phase("Verifier") {
-    if (!flix.options.xnoverify) {
+    if (flix.options.xnoverify) {
+      root
+    } else {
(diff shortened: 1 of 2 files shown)


main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -28,7 +28,7 @@ object Verifier {
 
   def run(root: Root)(implicit flix: Flix): Root = flix.phase("Verifier") {
     if (!flix.options.xnoverify) {
```
magnus-madsen, line 30: Could we change this to early return?

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -28,7 +28,7 @@ object Verifier {
...
-      root.defs.values.foreach(d => visitDef(d)(root))
+      ParOps.parMap(root.defs.values)(visitDef(_)(root))
```
magnus-madsen, line 31: Is there not a better parMap thing or  do we have to construct the .values thing?

JonathanStarup: Its not super clear from the docs, but values is `Iterable` and is based on `valuesIterator` so I'm pretty sure its not materialized.

There is no wellsuited ParOps things - it doesn't have foreach style methods

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
JonathanStarup: I wrote it like this to be consistent with Optimizer - I can change both? They can also both just use if-else

magnus-madsen: Yes, please

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
magnus-madsen: Ok

magnus-madsen (approved):

magnus-madsen:
Woops. Build issue.

PR #7446, opened 2024-03-22
refactor: rework channel and spawn effects

magnus-madsen:


main/src/library/Parallel.flix
```diff
@@ -0,0 +1,29 @@
...
+mod Parallel {
+
+    @Experimental
+    pub def race(f: a -> b \ ef, l: Nel[a]): b \ {ef, NonDet} with Sendable[b] = try {
+            region rc {
+                let (tx, rx) = Channel.buffered(rc, Nel.length(l));
+                foreach (x <- l) {
+                    spawn (Channel.send(f(x), tx)) @ rc
+                };
+                Channel.recv(rx)
```
magnus-madsen, line 25: This does not actually race because of the region.

Issue #7414, opened 2024-03-18
Add `TryCatch` exception type check to `Safety`


JonathanStarup:
Currently you can write

```
try 42 catch {
    case ex: #java.lang.String => ex
}
```

Safety should check that the exception type (`java.lang.String` here) is a subtype of `java.lang.Throwable`

stetimi:
Doesn't look too tricky.. I could take a look

stetimi:
Should this be closed now?

magnus-madsen:
Yes

PR #7448, opened 2024-03-22
feat: check type of value caught in try-catch (#7414)

stetimi:
The error below might be improved I think: the highlighting could be better on ##java.lang.String rather than _s

But I didn't find any SourceLocation in the parsed expression for the ## bit.

```
>> Catch type is not a subclass of Throwable.

5 |         case _s: ##java.lang.String => "x"
                 ^^
                 Type should be java.lang.Throwable or a subclass.
```

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/errors/SafetyError.scala
+++ b/main/src/ca/uwaterloo/flix/language/errors/SafetyError.scala
@@ -165,6 +165,24 @@ object SafetyError {
   }
 
   /**
+   * An error raised to indicate that the Java class in a catch clause is not a Throwable.
+   *
+   * @param loc the location of the catch parameter.
+   */
+  case class IllegalCatchType(loc: SourceLocation) extends SafetyError with Recoverable {
+    def summary: String = s"Exception type is not a subclass of Throwable."
+
+    override def message(formatter: Formatter): String = {
+      import formatter._
+      s"""${line(kind, source.name)}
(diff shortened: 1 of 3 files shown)


magnus-madsen:
- Add documentation for each method.
- Add a few more tests.

stetimi:
I added comments, and an extra test showing two catch clauses with one having the error.

main/src/ca/uwaterloo/flix/language/errors/SafetyError.scala
```diff
@@ -164,6 +164,24 @@ object SafetyError {
     override def explain(formatter: Formatter): Option[String] = None
   }
 
+  /**
+   * An error raised to indicate that the Java class in a catch clause is not a Throwable.
+   *
+   * @param loc the location of the catch parameter.
+   */
+  case class IllegalCatchType(loc: SourceLocation) extends SafetyError with Recoverable {
+    def summary: String = s"Catch type is not a subclass of Throwable."
```
JonathanStarup, line 173: Maybe `"Exception type is not a subclass of Throwable."` is more expect language?

stetimi: Agreed, and updated.

magnus-madsen (approved):

magnus-madsen:
Thanks @stetimi

PR #7454, opened 2024-03-23
feat: add closure to `Verifier`

felix-berg:


main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -302,6 +302,21 @@ object Verifier {
             case _ => failMismatchedShape(t1, "Tuple", loc)
           }
 
+        case AtomicOp.Closure(sym) =>
+          val defn = root.defs(sym)
+          if (defn.cparams.length != ts.length)
+            failMismatchedShape(ts, s"List of types matching closure parameters of $sym", loc)
```
JonathanStarup, line 308: This error can be better but I'm not sure exactly how. Right now the error says `expected = 'List of types matching closure parameters of clo$2761', found = List(Int32, Int32)`. it doesn't say what length or types it expected.

I think more work can be put into creating some nice error - but a simpler decent solution could be to take the approach of ApplyDef where a big arrow type is built, the `(ts..) -> tpe` should be the same as `(defn.cparams..) -> (defn.fparams..) -> defn.tpe`.

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -302,6 +302,21 @@ object Verifier {
...
+
+          val ctps = defn.cparams.map(cp => cp.tpe)
```
JonathanStarup, line 310: `_.tpe` is shorthand for `cp => cp.tpe`

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
felix-berg: yea thats way better

JonathanStarup (approved):

magnus-madsen (approved):

magnus-madsen:
Thanks @felix-berg

PR #7455, opened 2024-03-23
feat: add `FuzzyPrefixes`

magnus-madsen:


magnus-madsen:
@herluf-ba These test cases should be very useful for you. They essentially compile a program streaming it in small chuncks from left-to-right, top-to-bottom.

main/test/ca/uwaterloo/flix/TestStreaming.scala
```diff
@@ -0,0 +1,64 @@
...
+    */
+  private def compile(input: String): Unit = {
+    val length = input.length
+    val step = length / Chunks
+
+    val flix = new Flix()
+    flix.compile()
+    for (i <- 0 until Chunks) {
+      val e = Math.min(i * step, length)
+      val s = input.substring(0, e)
```
mlutze, line 58: This seems to add code from the beginning over and over.

It might be simpler just to used `StringOps.grouped`?

magnus-madsen: What's the benefit

magnus-madsen: Oh I see. It's a helper.

magnus-madsen: Cannot use grouped. I compute prefixes.

mlutze: Ah, didn't realize that was intentional.

mlutze (approved):

PR #7459, opened 2024-03-25
fix: off by 1 error in lexer

stetimi:


magnus-madsen:
@herluf-ba CC

herluf-ba:
@magnus-madsen Looks good. There is always one off-by-one error no matter what  😅

magnus-madsen (approved):

PR #7461, opened 2024-03-25
feat: handle leading and trailing `_` in numbers in lexer.

herluf-ba:
This makes `Lexer` produce errors on number literals that lead or end on `_`. For instance `123_` or `0x_123`.

Note that `TestLexer` is also added to `PhaseSuite`, as it should have been a while ago!


herluf-ba:
@magnus-madsen I think #7459 should land before this one.

main/src/ca/uwaterloo/flix/language/errors/LexerError.scala
```diff
@@ -259,17 +322,16 @@ object LexerError {
...
+   *
+   * @param loc The location of the opening `{`.
+   */
   case class UnterminatedStringInterpolation(loc: SourceLocation) extends LexerError with Recoverable {
     override def summary: String = s"Unterminated string interpolation."
 
     override def message(formatter: Formatter): String = {
       import formatter._
       s"""${line(kind, source.name)}
-         |>> Missing '}' in string interpolation.
```
magnus-madsen, line 272: @herluf-ba Was this intended?

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -960,6 +963,59 @@ object Lexer {
...
+        case _ => return advance() match {
+          case '_' => TokenKind.Err(LexerError.TrailingUnderscoreInNumber(sourceLocationAtCurrent()))
+          case _ if isMatch("f32") => error.getOrElse(TokenKind.LiteralFloat32)
+          case _ if isMatch("f64") => error.getOrElse(TokenKind.LiteralFloat64)
+          case _ if isMatch("i8") => error.getOrElse(TokenKind.LiteralInt8)
+          case _ if isMatch("i16") => error.getOrElse(TokenKind.LiteralInt16)
+          case _ if isMatch("i32") => error.getOrElse(TokenKind.LiteralInt32)
+          case _ if isMatch("i64") => error.getOrElse(TokenKind.LiteralInt64)
+          case _ if isMatch("ii") => error.getOrElse(TokenKind.LiteralBigInt)
+          case _ if isMatch("ff") => error.getOrElse(TokenKind.LiteralBigDecimal)
```
magnus-madsen, line 1008: There is some duplication in suffixes. But we can handle that later.

main/src/ca/uwaterloo/flix/language/errors/LexerError.scala
herluf-ba: No that seems weird, I must have yanked that line by accident.

herluf-ba:
@magnus-madsen I am having some trouble with the `DoubleDottedNumber.{01, 02, 03}` tests. When I run `TestLexer` they fail because only a `ParseError` is produced. But when I then run the exact same test in a `foo.flix` file with --Xlib=nix the lexer error shows up! Any ideas?

magnus-madsen (approved):

magnus-madsen:
> @magnus-madsen I am having some trouble with the `DoubleDottedNumber.{01, 02, 03}` tests. When I run `TestLexer` they fail because only a `ParseError` is produced. But when I then run the exact same test in a `foo.flix` file with --Xlib=nix the lexer error shows up! Any ideas?

Huh? It sounds like something in your code. Tests are run pretty similar to how programs are run.

herluf-ba:
The `ParseError` is coming from the old parser. The new one isn't even on this branch! I found the issue, The test passes when I add a newline to the test. So `val input = "1.2.3\n"` rather than `val input = "1.2.3"`. That indicates an issue with the Lexer (So my code indeed ;-) )

herluf-ba:
@magnus-madsen This is ready now

PR #7475, opened 2024-03-26
feat: remove redundant unchecked casts

### LINKED ISSUE #7465: A method import with an IO effect should not introduce an unchecked cast

magnus-madsen:
Fixes #7465


JonathanStarup (commented):
In one place subst is applied to declaredEff in another its not, which should it be?

you use == so it doesnt take boolean equivalence into account. same thing with associated types and aliases

magnus-madsen:
@JonathanStarup Good points. I added some comments.

JonathanStarup (approved):
I think you could try to unify them, but comment is good

magnus-madsen:
> I think you could try to unify them, but comment is good

Yes, but not in there are assoc types as you wrote. Then it _may be_ that the unchecked cast is useless, but you cannot know until monomorphization.

PR #7484, opened 2024-03-27
feat: recoverable `UndefinedName`

magnus-madsen:
We need to support recoverable undefined names for proper auto-completion support.

This PR solves the issue for lookupQName, but there still remains some issues for other names.

I am not really sure why we have such a broad type of error. For example, why is it used for associated types too?


magnus-madsen:
@mlutze Ping :)

mlutze (approved):

Issue #7286, opened 2024-02-26
Post New Typer


mlutze:
- [x] https://github.com/flix/flix/issues/7427 -- I think this is important before work on the std lib.
- [ ] https://github.com/flix/flix/issues/7428 -- I think this is important before work on the std lib.
- [x] #7503
- [ ] https://github.com/flix/flix/issues/7433
- [ ] https://github.com/flix/flix/issues/7477
- [ ] https://github.com/flix/flix/issues/7476
- [ ] #7278 
- [x] #7499
- [x] #7500
- [ ] #7507
- [x] #7501
- [x] Check on source location of constraints arising from sigs that are not directly applied
- [ ] try out merging `Scheme.instantiate` and `addClassConstraints`
- [x] look at location copying for class constraints https://github.com/flix/flix/pull/7306#discussion_r1505429717
- [ ] consider helper for unary and binary ops https://github.com/flix/flix/pull/7306#discussion_r1505768372
- [x] naming in Expr.Apply(Expr.Def https://github.com/flix/flix/pull/7306#discussion_r1505760601
- [ ] Add MismatchedKinds error
- [x] Make Scheme.lessThanEqual return Bool
- [x] Change `args` fields in AST nodes to `exps`
- [x] check effect of JVM method matches declared
- [x] unifyAllTypes don't return anything https://github.com/flix/flix/pull/7306#discussion_r1505474714
- [x] instanceof -- allow non-pure argument
- [x] Require guards to be pure
- [x] Rename `TypeContext` methods
- [x] Remove levels
- [x] Remove old typer and associated functionality.

JonathanStarup:
@mlutze is this still being used?

mlutze:
I think it's stale wrt the new new typer stuff.

PR #7485, opened 2024-03-27
chore: remove old typing code

mlutze:
related to #7286 

There's still some stuff in unification that is used. We should migrate it to ConstraintGeneration so we have a SSOT.


magnus-madsen (approved):

magnus-madsen:
Goodbye old type inference, rest well.

PR #7497, opened 2024-03-28
fix: Parser2 should return an AST so that Weeder2 runs

magnus-madsen:


magnus-madsen:
@herluf-ba This ensures that the Weeder2 is run, but of course now if Parser2 fails the whole thing fails. 
So run silent is not really run silent anymore.

magnus-madsen:
So I am not entirely sure what to do. What we have now _IS_ running the Lexer and Parser-- which is good. It is just not running the Weeder yet.

herluf-ba:
I agree that we should try to run the Weeder too if possible. My worry is that before we have recover sets, users could get a ton of errors from the chatty parser. But I think `toHardResult`  gets around that? It cast `SoftFailure` to `Result.Err` too right?

herluf-ba:
Yep, that seems reasonable. With a bad file the parser gets silenced, but with an ok file we catch a potential chrash.

magnus-madsen:
I think I have a reasonable solution: If Parser2 succeeds then we pass the AST on to the Weeder2. If Parser2 fails then we do not know what to do and we simply return the empty AST (successfully). Thus Weeder2 will be tested when Parser2 succeeds.

magnus-madsen:
@herluf-ba If you agree, I need you to click review, and approve the PR :)

Now the rubber-band is on the other glove!

herluf-ba (approved):

herluf-ba:
Being on this side of things is new to me 😅

magnus-madsen:
Ah, you must also submit a review. Even if it is empty.

magnus-madsen:
Files Changed -> Review Changes -> Approve

herluf-ba (approved):

herluf-ba:
I did that. It seems it's not enough since I don't have write access to flix

magnus-madsen:
Hmm

magnus-madsen:
I cannot figure it out. I think we need to sit down and look at persmissions at some point.

Meanwhile, @mlutze can you approve?

mlutze (approved):

PR #7504, opened 2024-03-28
refactor: cleanup `TypeContext`

magnus-madsen:


main/src/ca/uwaterloo/flix/language/phase/constraintgeneration/TypeContext.scala
```diff
@@ -122,15 +123,17 @@ class TypeContext {
     *   tpe1 ~ tpe3
     * }}}
     */
-  def unifyType3(tpe1: Type, tpe2: Type, tpe3: Type, loc: SourceLocation): Unit = {
```
mlutze, line 125: Overloads make me sad, but ok.

mlutze (approved):

main/src/ca/uwaterloo/flix/language/phase/constraintgeneration/TypeContext.scala
magnus-madsen: I'm just being consistent. We could get rid of them but then it must be everywhere

PR #7505, opened 2024-03-28
feat: add Xprint-typer option

### LINKED ISSUE #7499: Typer: Add `--Xprint-typer` to emit type constraints to files

mlutze:
fixes #7499

### DIFF
--- a/main/src/ca/uwaterloo/flix/Main.scala
+++ b/main/src/ca/uwaterloo/flix/Main.scala
@@ -103,6 +103,7 @@ object Main {
       xprintphase = cmdOpts.xprintphase,
       xsummary = cmdOpts.xsummary,
       xparser = cmdOpts.xparser,
+      xprinttyper = cmdOpts.xprinttyper,
       XPerfFrontend = cmdOpts.XPerfFrontend,
       XPerfN = cmdOpts.XPerfN
     )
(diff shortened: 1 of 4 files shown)


main/src/ca/uwaterloo/flix/Main.scala
```diff
@@ -361,6 +362,7 @@ object Main {
                      xprintphase: Set[String] = Set.empty,
                      xsummary: Boolean = false,
                      xparser: Boolean = false,
+                     xprinttyper: Option[String] = None,
```
magnus-madsen, line 365: What's the reason not to dump _ALL_ type constraints?

magnus-madsen (approved):

main/src/ca/uwaterloo/flix/Main.scala
mlutze: It would take forever :)

The current model outputs a file for _each iteration_ of constraint resolution, which means up to hundreds of files per definition.

magnus-madsen: Uh oh. I wonder if that is not something best left tuning in code.

magnus-madsen: The AST debugging is just dumping the ASTs.

magnus-madsen: which is nice to get a general overview

PR #7510, opened 2024-03-29
chore: rename class to trait in Resolver

mlutze:
related to https://github.com/flix/flix/issues/6591

Note: maybe this should not be merged if instead we want to be more systematic.


magnus-madsen (approved):

magnus-madsen:
We have to start somewhere, so we might as well start here.

PR #7515, opened 2024-03-31
fix: show trait instances in HTML documentation

### LINKED ISSUE #7435: Flixdoc: show instances
When viewing a trait: https://api.flix.dev/Eq.html

We should have a list of instances that are implemented by that trait at the bottom.

For example, Eq has many instances, e.g. for most collections.

sockmaster27:
Fixes  #7435


magnus-madsen:
Screenshots?

sockmaster27:
billede

It simply restores how it was before.
The section is hidden by default.

magnus-madsen:
Ok. Any improvements we could make? 

Perhaps we will want show how associated types/effects are defined by each instance. 

https://doc.flix.dev/associated-types.html

magnus-madsen (approved):

PR #7522, opened 2024-04-01
chore: rename class to trait in HTMLDocumentor (issue #6591)

stephentetley:
This PR renames class to trait within HTMLDocumentor. 

There are some field names that continue to use "class" at the moment as the are types defined outside this module.


magnus-madsen (approved):

magnus-madsen:
Thanks!

PR #7524, opened 2024-04-01
chore: rename class to trait in Typer and Instances (issue #6591)

stephentetley:
This PR renames class to trait within Type and Instances. 

Some constructor and field names continue to use "class" at the moment as they are types defined outside these modules.


magnus-madsen:
Thanks!

PR #7525, opened 2024-04-01
feat: add --albatross flag

mlutze:
related to #1337


magnus-madsen:
I think we need a full-blown date-time parsing library, so I can write for example `--albatross 1.5fortnights`.

mlutze:
> I think we need a full-blown date-time parsing library, so I can write for example `--albatross 1.5fortnights`.

I think php has that built-in, so we could just add it as a dependency.

magnus-madsen:
Lets look at this in about a years time :)

PR #7526, opened 2024-04-01
refactor: non optional nth on Array and Vector

stetimi:
Adding nth + IndexOutOfBounds effect, just on Array and Vector (I prefer to do it incrementally if that's acceptable).

### DIFF
--- a/main/src/library/Array.flix
+++ b/main/src/library/Array.flix
@@ -111,13 +111,15 @@ mod Array {
         $ARRAY_STORE$(a, i, x)
 
     ///
-    /// Optionally returns the element at position `i` in the array `a`.
+    /// Returns the element at position `n` in `a`.
     ///
-    pub def nth(i: Int32, a: Array[a, r]): Option[a] \ r =
-        if (0 <= i and i < length(a))
-            Some(get(i, a))
+    /// Throws `IndexOutOfBounds` if `n` is negative or `a` has length less than `n`.
+    ///
+    pub def nth(n: Int32, a: Array[a, r]): a \ {r, IndexOutOfBounds} =
+        if (0 <= n and n < length(a))
+            get(n, a)
         else
-            None
+            do IndexOutOfBounds.throw("Index ${n} out of bounds")
 
     ///
     /// Returns `true` if the given array `a` is empty.
--- a/main/src/library/MutDeque.flix
+++ b/main/src/library/MutDeque.flix
@@ -418,10 +418,7 @@ mod MutDeque {
             if (i == d->back) {
                 ()
             } else {
-                match Array.nth(i, d->values) {
-                    case Some(x) => f(x)
-                    case None    => bug!("An error occurred in MutDeque.forEach!")
-                };
+                f(Array.get(i, Ref.get(a)));
                 loop(Int32.bitwiseAnd(i + 1, c))
             }
         };
--- a/main/src/library/MutList.flix
+++ b/main/src/library/MutList.flix
@@ -108,14 +108,16 @@ mod MutList {
             do OutOfBounds.outOfBounds("index ${i} is out of bounds for MutList of length ${MutList.length(v)}")
 
     ///
-    /// Optionally returns the element at position `i` in the mutable list `v`.
+    /// Returns the element at position `n` in `v`.
     ///
-    pub def nth(i: Int32, v: MutList[a, r]): Option[a] \ r =
+    /// Throws `IndexOutOfBounds` if `n` is negative or `v` has length less than `n`.
+    ///
+    pub def nth(n: Int32, v: MutList[a, r]): a \ {r, IndexOutOfBounds} =
         let MutList(_, a, l) = v;
         if (0 <= i and i < Ref.get(l))
-            Array.nth(i, Ref.get(a))
+            Array.get(i, Ref.get(a))
         else
-            None
+            do IndexOutOfBounds.throw("Index ${n} out of bounds")
 
     ///
     /// Returns the number of elements in the given mutable list `v`.
--- a/main/src/library/Prelude.flix
+++ b/main/src/library/Prelude.flix
@@ -109,6 +109,13 @@ pub eff NonDet
 pub eff Sys
 
 ///
+/// An effect raised to indicate that an index is out of range.
+///
+pub eff IndexOutOfBounds {
+    pub def throw(msg: String): Void
+}
+
+///
 /// The pure identity function.
 ///
 pub def identity(x: a): a = x
--- a/main/src/library/Vector.flix
+++ b/main/src/library/Vector.flix
@@ -216,13 +216,15 @@ mod Vector {
     pub def get(i: Int32, v: Vector[a]): a = $VECTOR_GET$(v, i)
 
     ///
-    /// Optionally returns the element at position `i` in the vector `v`.
+    /// Returns the element at position `n` in `v`.
     ///
-    pub def nth(i: Int32, v: Vector[a]): Option[a] =
-        if (0 <= i and i < length(v))
-            Some(get(i, v))
+    /// Throws `IndexOutOfBounds` if `n` is negative or `v` has length less than `n`.
+    ///
+    pub def nth(n: Int32, v: Vector[a]): a \ IndexOutOfBounds =
+        if (0 <= n and n < length(v))
+            get(n, v)
         else
-            None
+            do IndexOutOfBounds.throw("Index ${n} out of bounds")
 
     ///
     /// Returns `true` if the given vector `v` is empty.
@@ -317,7 +319,11 @@ mod Vector {
     /// Returns `None` if `v` is empty.
     ///
     pub def head(v: Vector[a]): Option[a] =
-        nth(0, v)
+        if (length(v) > 0) {
+            Some(get(0, v))
+        } else {
+            None
+        }
 
     ///
     /// Returns `Some(x)` if `x` is the last element of `v`.
@@ -325,7 +331,12 @@ mod Vector {
     /// Returns `None` if `v` is empty.
     ///
     pub def last(v: Vector[a]): Option[a] =
-        nth(length(v) - 1, v)
+        let l = length(v);
+        if (l > 0) {
+            Some(get(l - 1, v))
+        } else {
+            None
+        }
 
     ///
     /// Return a new vector, appending the elements `v2` after elements of `v1`.
--- a/main/test/ca/uwaterloo/flix/library/TestArray.flix
+++ b/main/test/ca/uwaterloo/flix/library/TestArray.flix
@@ -82,27 +82,49 @@ mod TestArray {
 
     @test
     def testNth01(): Bool = region rc {
-        Array.nth(0, Array#{1, 2, 3} @ rc) == Some(1)
+        try {
+            Assert.eq(1, Array.nth(0, Array#{1, 2, 3} @ rc))
+        } with IndexOutOfBounds {
+            def throw(_, _) = false
+        }
     }
 
     @test
     def testNth02(): Bool = region rc {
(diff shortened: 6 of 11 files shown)


main/src/library/Array.flix
```diff
@@ -76,13 +76,15 @@ mod Array {
         $ARRAY_STORE$(a, i, x)
 
     ///
-    /// Optionally returns the element at position `i` in the array `a`.
+    /// Returns the element at position `n` in `a`.
     ///
-    pub def nth(i: Int32, a: Array[a, r]): Option[a] \ r =
-        if (0 <= i and i < length(a))
-            Some(get(i, a))
+    /// If `n` is negative or `a` has length less than n, an IndexOutOfBounds effect is thrown.
```
stetimi, line 81: "an IndexOutOfBounds effect is thrown": is "thrown" the right word for an effect?

main/test/ca/uwaterloo/flix/library/TestArray.flix
```diff
@@ -82,27 +82,49 @@ mod TestArray {
...
+            Assert.eq(3, Array.nth(2, Array#{1, 2, 3} @ rc))
+        } with IndexOutOfBounds {
+            def throw(_, _) = false
+        }
     }
 
     @test
     def testNth04(): Bool = region rc {
-        Array.nth(3, Array#{1, 2, 3} @ rc) == None
+        try {
```
stetimi, line 112: If I had test code like this in java, I'd replace it with assertThrows() - which checks that particular exceptions are thrown. Is it possible in Flix too?

main/src/library/Array.flix
```diff
@@ -76,13 +76,15 @@ mod Array {
...
+    ///
+    pub def nth(n: Int32, a: Array[a, r]): a \ {r, IndexOutOfBounds} =
```
stetimi, line 83: Note this is a breaking API change, which is causing one of the community builds to fail.

stetimi:
@mlutze ready to review I think

main/test/ca/uwaterloo/flix/library/TestArray.flix
mlutze: We don't have that now, and I don't think it can be done generically, but we can add one for IndexOutOfBounds specifically.

magnus-madsen:
Hi. I am a bit busy at the moment, but I will try to get around to it. Also just FYI: We may not be able to merge this in until there is a bit more progress on the effect side.

main/src/library/Vector.flix
```diff
@@ -189,13 +189,15 @@ mod Vector {
     pub def get(i: Int32, v: Vector[a]): a = $VECTOR_GET$(v, i)
 
     ///
-    /// Optionally returns the element at position `i` in the vector `v`.
+    /// Returns the element at position `n` in `v`.
     ///
-    pub def nth(i: Int32, v: Vector[a]): Option[a] =
-        if (0 <= i and i < length(v))
-            Some(get(i, v))
+    /// If `n` is negative or `v` has length less than n, an IndexOutOfBounds effect is thrown.
```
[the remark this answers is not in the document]
stetimi: Updated

main/src/library/Array.flix
magnus-madsen: Ok

magnus-madsen:
Why is there a file called `^1` included? :)

main/src/library/Array.flix
magnus-madsen: I think we should use the phrase: `Throws `IndexOutOfBounds` if ....`

main/src/library/Prelude.flix
```diff
@@ -193,3 +193,10 @@ pub def purityOf3(f: a -> b -> c -> d \ ef): Purity3[a, b, c, d, ef] = typematch
     case g: a -> b -> c -> d \ Pure => Purity3.Pure(g)
     case _: _                       => Purity3.Impure(f)
 }
+
+///
+/// An effect raised to indicate that an index is out of range.
+///
+pub eff IndexOutOfBounds {
+    pub def throw(msg: String): Void
+}
```
magnus-madsen, line 202: Looks good, but if you could move it right after the IO effect near the top.

main/src/library/MutDeque.flix
```diff
@@ -442,10 +442,12 @@ mod MutDeque {
...
             } else {
-                match Array.nth(i, deref a) {
-                    case Some(x) => f(x)
-                    case None    => bug!("An error occurred in MutDeque.forEach!")
+                let elt = try {
+                    Array.nth(i, deref a)
+                } with IndexOutOfBounds {
+                    def throw(_, _) = bug!("An error occurred in MutDeque.forEach!")
                 };
+                f(elt);
```
magnus-madsen, line 450: I think it woud be more natural to have 

```
try {
   f(Array.nth(i, deref a))
} catch ...
```

main/src/library/MutList.flix
```diff
@@ -68,12 +68,16 @@ mod MutList {
...
     ///
-    /// Optionally returns the element at position `i` in the mutable list `v`.
+    /// Optionally returns the element at position `n` in the mutable list `v`.
     ///
-    pub def nth(i: Int32, v: MutList[a, r]): Option[a] \ r =
+    pub def nth(n: Int32, v: MutList[a, r]): Option[a] \ r =
         let MutList(_, a, l) = v;
-        if (0 <= i and i < deref l)
-            Array.nth(i, deref a)
+        if (0 <= n and n < deref l)
```
magnus-madsen, line 75: It seems like we are checking twice. Both here and now inside Array.nth. 

I also strikes me as wrong that `MutList.nth` should return an Option since we are just in the process of having all `nth` functions return the element directly + exception. Can we change it do that instead?

main/src/library/Vector.flix
```diff
@@ -285,15 +287,23 @@ mod Vector {
     /// Returns `None` if `v` is empty.
     ///
     pub def head(v: Vector[a]): Option[a] =
```
magnus-madsen, line 289: Performance for Vector is important, so here I would do the bounds check directly (i.e. that length != 0) and then return None/Some based on that.

main/src/library/Vector.flix
```diff
@@ -285,15 +287,23 @@ mod Vector {
...
+        } with IndexOutOfBounds {
+            def throw(_, _) = None
+        }
 
     ///
     /// Returns `Some(x)` if `x` is the last element of `v`.
     ///
     /// Returns `None` if `v` is empty.
     ///
     pub def last(v: Vector[a]): Option[a] =
```
magnus-madsen, line 301: Same comment as for head.

main/test/ca/uwaterloo/flix/library/TestArray.flix
magnus-madsen: Yeah, unfortunately we have to use "manual" try-with for now.

magnus-madsen: So I think the test cases are fine.

main/test/ca/uwaterloo/flix/library/TestArray.flix
```diff
@@ -82,27 +82,49 @@ mod TestArray {
...
+            Array.nth(3, Array#{1, 2, 3} @ rc);
+            false
+        } with IndexOutOfBounds {
+            def throw(msg, _) = Assert.eq("Index 3 out of bounds", msg)
```
magnus-madsen, line 116: Just return true, let us not depend on the specific message.

main/test/ca/uwaterloo/flix/library/TestArray.flix
```diff
@@ -82,27 +82,49 @@ mod TestArray {
...
     }
 
     @test
     def testNth05(): Bool = region rc {
-        Array.nth(-2, Array#{1, 2, 3} @ rc) == None
+        try {
+            Array.nth(-2, Array#{1, 2, 3} @ rc);
+            false
+        } with IndexOutOfBounds {
+            def throw(msg, _) = Assert.eq("Index -2 out of bounds", msg)
```
magnus-madsen, line 126: Just return true, let us not depend on the specific message.

main/test/ca/uwaterloo/flix/library/TestVector.flix
```diff
@@ -108,36 +108,58 @@ mod TestVector {
...
+        }
 
     @test
     def nth05(): Bool =
-        Vector.nth(-2, Vector#{1, 2, 3}) == None
+        try {
+            Vector.nth(-2, Vector#{1, 2, 3});
+            false
+        } with IndexOutOfBounds {
+            def throw(msg, _) = Assert.eq("Index -2 out of bounds", msg)
```
magnus-madsen, line 161: Just return true, let us not depend on the specific message.

magnus-madsen (commented):
OK. I did a detailed review. If you fix these things I think its good to go.

main/src/library/MutDeque.flix
stetimi: My thinking was f might throw IndexOutOfBounds, and then we'd report it as a bug in forEach.

magnus-madsen: Yes, we should still do that. Its just there is no reason for the local variable elt.

stetimi: Would using `Array.get` be ok here too?

main/src/library/MutList.flix
stetimi: MutList has to check, because the Array can be longer than the length of the MutList (the minimum capacity is 8).
Would Array.get would be acceptable here too?

magnus-madsen: Lets do one check in MutList and then use Array.get internally.

magnus-madsen:
@stetimi Could you mark the comments as resolved, if they are indeed resolved :)

stetimi:
Sure, I wasn't too sure of the protocol.
I think they're all resolved, but I'll double check and close them off when
I'm free.

stetimi:
> @stetimi Could you mark the comments as resolved, if they are indeed resolved :)

Done

main/test/flix/Test.Exp.Regions.flix
```diff
@@ -454,7 +454,7 @@ mod Test.Exp.Regions {
                 MutList.push!(1, l);
                 MutMap.put!(1, l, Array.get(0, a));
                 match MutMap.get(1, Array.get(0, a)) {
-                    case Some(k) => MutList.nth(0, k) == Some(1)
+                    case Some(k) => optionalNth(0, k) == Some(1)
```
magnus-madsen, line 457: These are just test cases, so I think we should just use `get` and avoid all the optionalNth stuff!

magnus-madsen:
I added one comment about simplifying the test cases. Once that is done, I will let @mlutze have a look and we can probably merge.

He also needs to fix one of his projects before the merge.

Thanks!

main/test/flix/Test.Exp.Regions.flix
stetimi: There is no `get` in `MutList`. Should I: 
1. add it and delegate down to `Array.get`
2. use `nth` and put try..with statements around the relevant tests?

magnus-madsen:
@stetimi Interested in resurrecting this and getting it merged?

stetimi:
> @stetimi Interested in resurrecting this and getting it merged?

Sure, I'd be happy to help if I can.
Some work's already been done to add an Indexable trait with implementations on Array, Vector and others, so I think this PR is obsolete.
I could perhaps add more implementations as in https://github.com/flix/flix/issues/7983 if that is useful?

magnus-madsen:
I think this PR has been overtaken by recent changes. 

We should discuss exactly what should be the plan for `nth` in some ticket and then from there figure out next steps.

PR #7534, opened 2024-04-01
chore: rename class to trait in various LSP modules (issue #6591)

stephentetley:
This PR renames class to trait in Indexer and various other LSP modules. I think all the changes are internal and don't involve the LSPs API (I don't really know how it works).

I'm back at work tomorrow, so the rate of PRs will slow down.


magnus-madsen (approved):

magnus-madsen:
Thanks!

PR #7542, opened 2024-04-03
feat: collect effect constraints to be unified together

### LINKED ISSUE #7503: Typer: Use `FastBoolUnification`

mlutze:
fixes #7503

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolver.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolver.scala
@@ -108,13 +108,14 @@ object ConstraintSolver {
       // We resolve the constraints under the environments we created. //
       ///////////////////////////////////////////////////////////////////
       resolve(constrs, initialSubst, renv)(cenv, eenv, flix) match {
-        case Result.Ok(ResolutionResult(subst, deferred, _)) =>
+        case Result.Ok(ResolutionResult(subst, deferred)) =>
           Debug.stopRecording()
-
           // If there are any constraints we could not resolve, then we report an error.
           // TODO ASSOC-TYPES here we only consider the first error
           getFirstError(deferred, renv) match {
+            // Case 1: Type constraints are resolved.
             case None => Validation.success(subst)
+            // Case 2: Type constraints are not resolved.
             case Some(err) => Validation.toSoftFailure(subst, err)
           }
 
@@ -212,6 +213,12 @@ object ConstraintSolver {
     // We give up if this exceeds MaxIterations.
     var count = 0
 
+    // Effect constraints that we are deferring until the end
+    var effs: List[TypeConstraint.Equality] = Nil
+
+    // Trait constraints that we are deferring until the end
+    var traits: List[TypeConstraint.Trait] = Nil
+
     // Whether the last resolution attempt made progress.
     // We give up if a resolution attempt does not make progress.
     var progress = true
@@ -223,22 +230,30 @@ object ConstraintSolver {
 
       count += 1
 
-      Debug.recordGraph(curr, subst)
+      Debug.recordGraph(curr, effs, traits, subst)
 
       resolveOneOf(curr, subst, renv) match {
         // Case 1: Success. Update the tracking variables.
-        case Result.Ok(ResolutionResult(newSubst, newConstrs, newProgress)) =>
+        case Result.Ok(IntermediateResult(newSubst, newConstrs, moreEffs, moreTraits, newProgress)) =>
           curr = newConstrs
           subst = newSubst
+          effs = moreEffs ::: effs
+          traits = moreTraits ::: traits
           progress = newProgress
 
         // Case 2: Error. Break out of the loop and return the error.
-        case res@Result.Err(_) =>
+        case Result.Err(err) =>
           Debug.stopRecording()
-          return res
+          return Result.Err(err)
       }
     }
-    Result.Ok(ResolutionResult(subst, curr, progress = true))
+
+    // After we reduced the normal constraints as far as possible, we try the effect constraints and the trait constraints.
+    for {
+      s <- resolveEffConstraints(effs, renv, subst)
+      _ <- resolveFinalTraitConstraints(traits, renv, s)
+    } yield ResolutionResult(s, curr)
+
   }
 
   /**
@@ -246,9 +261,9 @@ object ConstraintSolver {
     *
     * Applies the initial substitution `subst0` before attempting to resolve.
     */
-  private def resolveOneOf(constrs: List[TypeConstraint], subst0: Substitution, renv: RigidityEnv)(implicit tenv: Map[Symbol.TraitSym, Ast.TraitContext], eenv: ListMap[Symbol.AssocTypeSym, Ast.AssocTypeDef], flix: Flix): Result[ResolutionResult, TypeError] = {
-    def tryResolve(cs: List[TypeConstraint]): Result[ResolutionResult, TypeError] = cs match {
-      case Nil => Result.Ok(ResolutionResult(subst0, cs, progress = false))
+  private def resolveOneOf(constrs: List[TypeConstraint], subst0: Substitution, renv: RigidityEnv)(implicit tenv: Map[Symbol.TraitSym, Ast.TraitContext], eenv: ListMap[Symbol.AssocTypeSym, Ast.AssocTypeDef], flix: Flix): Result[IntermediateResult, TypeError] = {
+    def tryResolve(cs: List[TypeConstraint]): Result[IntermediateResult, TypeError] = cs match {
+      case Nil => Result.Ok(IntermediateResult(subst0, Nil, Nil, Nil, progress = false))
       case hd :: tl => resolveOne(hd, renv, subst0).flatMap {
         // if we're just returning the same constraint, then have made no progress and we need to find something else to reduce
         case hdRes if !hdRes.progress => tryResolve(tl).map {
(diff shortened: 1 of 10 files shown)


mlutze:
> @mlutze This should stay open, right?

Yes

magnus-madsen:
@mlutze This seems very close to working. Do you need help with debugging? It seems Datalog related. So maybe lowering or something with free variables? I am kind of unsure how you can get those errors we are getting though.

magnus-madsen:
Ah- its not calling the new solver yet, right?

magnus-madsen:
Bug is non-deterministic, but this reproduces it always with threads = 1:

```
@test
def main(): Bool =
    let p = #{
        A(1, 2, 3).
        R(x, y, z) :- A(x, y, z).
    };
    let r = query p select (x, y, z) from R(x, y, z);
    Vector.memberOf((1, 2, 3), r)
```

magnus-madsen:
`--Xno-optimizer` suggests that the optimizer does something (probably based on wrong effect information).

magnus-madsen:
@mlutze @JonathanStarup I have isolated the bug. It has to do with UncheckedMaskingCast.

If I remove:

```
Interpreter.flix:32:    masked_cast(Fixpoint.Debugging.notifyPreInterpret(stmt));
Solver.flix:47:   masked_cast(Fixpoint.Debugging.notifyPreSolve(d, stf));
Solver.flix:64:    masked_cast(Fixpoint.Debugging.notifyPostSolve(model))
```

All tests (except one unrelated) run.

magnus-madsen:
I suspect it has to do with `UncheckedMaskingCast` being erased and then later the inliner making changes which gives rise to inconsistent effects.

@JonathanStarup Does the verifier not check for effect consistency? E.g. if a sub-expression is control-impure presumable the expression itself must also be control-impure.

magnus-madsen:
I have a workaround here https://github.com/flix/flix/pull/7659

It solves nothing, but at least it makes the problem go-away.

magnus-madsen:
I think now there is just Type.Error left to deal with.

mlutze:
> I think now there is just Type.Error left to deal with.

I think ideally we could treat every Type.Error as a free var. Then any type errors we get out will only be ones we would have no matter what the real value of Type.Error is. What do you think?

magnus-madsen:
> > I think now there is just Type.Error left to deal with.
> 
> I think ideally we could treat every Type.Error as a free var. Then any type errors we get out will only be ones we would have no matter what the real value of Type.Error is. What do you think?

I was thinking constant?

magnus-madsen:
> > > I think now there is just Type.Error left to deal with.
> > 
> > 
> > I think ideally we could treat every Type.Error as a free var. Then any type errors we get out will only be ones we would have no matter what the real value of Type.Error is. What do you think?
> 
> I was thinking constant?

Ill let you experiment/decide. I don't think the choice right now is important.

magnus-madsen:
I see lots of non-determinism indeed.

mlutze:
> Disabling the optimizer seems to make the tests pass. But what causes the non-determinism which sometimes allows everything to go through?

One source is the mapping of types to Boolean terms.

magnus-madsen:
Another source is the naming of variables which affects the solutions computed by the solver (?)

magnus-madsen:
I think we need the Verifier to reject illegal programs w.r.t effects.

magnus-madsen:
What's the working thesis: That we produce expressions with non-deterministic effects that are only sometimes wrong and hence only sometimes screwed up by the optimizer?

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolver.scala
```diff
@@ -226,28 +233,36 @@ object ConstraintSolver {
...
+        case Result.Err(err) =>
           Debug.stopRecording()
-          return res
+          return Result.Err(err)
       }
     }
-    Result.Ok(ResolutionResult(subst, curr, progress = true))
+
+    // After we reduced the normal constraints as far as possible, we try the effect constraints and the trait constraints.
+    for {
```
magnus-madsen, line 251: I don't like that this is outside the while-loop. While this may not be the source of our current issues, this cannot in general work.

magnus-madsen: We should really try to follow a model of always working on a constraint set. We are only done when it is empty or there is a conflict. There should be no other criteria/control-flow.

magnus-madsen:
I have created #7672 and #7673 sadly neither seems to address the issue, but nevertheless I believe they make things safer.

magnus-madsen:
Interestingly running `Reachable.flix` __always__ fails deterministically. So its the presence of _other tests_ that some times make it pass.

magnus-madsen:
In monomorpher, if I compare what source locations are made pure by the strict subst:

```
      case Kind.Eff =>
        // If an effect variable is free, we may assume its Pure due to the subst. lemma.
        println(tpe0.loc)
        Type.Pure
```

it seems that this PR has more source locations made pure, including in e.g. 

```
MutList.flix:854:9
MutList.flix:851:17
(and several other places)
```

which at least suggests that more variables are free in this PR...

main/src/ca/uwaterloo/flix/language/phase/unification/EffUnification2.scala
```diff
@@ -16,26 +16,24 @@
...
-   * @param loc the source location of the entire equation system, e.g. the entire function body.
+    *
+    * @param l    the list of unification equations.
+    * @param renv the rigidity environment.
+    * @param loc  the source location of the entire equation system, e.g. the entire function body.
     */
   def unifyAll(l: List[(Type, Type, SourceLocation)], renv: RigidityEnv, loc: SourceLocation)(implicit flix: Flix): Result[Substitution, UnificationError] = {
     // Compute a bi-directional map from type variables to ints.
-    implicit val bimap: Bimap[Type.Var, Int] = mkBidirectionalVarMap(l)
+    implicit val bimap: Bimap[Atom, Int] = mkBidirectionalVarMap(l)
```
magnus-madsen, line 36: We could do an experiment where we call the current Boolean solver here. 

Assuming it does not run out of resources that could tell us whether the two solvers agree?

magnus-madsen:
@JonathanStarup We believe the bug is in the typer, but we could really benefit from having a verifier that checks effects s.t. the backend does not generate incorrect code.

magnus-madsen:
Did you have some success or not yet?

mlutze:
> Did you have some success or not yet?

Not yet :(

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolver.scala
mlutze: Can I do this in a next revision?

I'd like to try to refactor by making a TypeConstraintSet which separates type constraints, effect constraints, and class constraints. I think that will help simplify things and make it easier to keep it all in the same loop.

magnus-madsen: Yes. Sounds good on both accounts.

main/src/ca/uwaterloo/flix/language/ast/TypeConstructor.scala
```diff
@@ -414,6 +414,6 @@ object TypeConstructor {
   /**
     * A type constructor which represents an erroneous type of the given `kind`.
     */
-  case class Error(kind: Kind) extends TypeConstructor
+  case class Error(id: Int, kind: Kind) extends TypeConstructor
```
magnus-madsen, line 417: What's the id for?

mlutze: In `EffUnification2` I map error effects to solver variables. I need different errors to have different IDs, otherwise they all get mapped to the same variable, which causes confusing errors.

magnus-madsen:
RFM?

mlutze:
> test failures?

please no

magnus-madsen:
What a pain. Same issue or new issue?

magnus-madsen:
> > What a pain. Same issue or new issue?
> 
> Seems new. There's no runtime crash, but apparently a nondeterministic miscompilation. I cannot reliably reproduce.

Try to compile one of the tests in isolation. That worked last time.

mlutze:
> > > What a pain. Same issue or new issue?
> > 
> > 
> > Seems new. There's no runtime crash, but apparently a nondeterministic miscompilation. I cannot reliably reproduce.
> 
> Try to compile one of the tests in isolation. That worked last time.



> > > What a pain. Same issue or new issue?
> > 
> > 
> > Seems new. There's no runtime crash, but apparently a nondeterministic miscompilation. I cannot reliably reproduce.
> 
> Try to compile one of the tests in isolation. That worked last time.

not working for me :(

magnus-madsen:
One in about ten run produces false:

```
mod Reachable.Imperative {
    pub def reachable(origin: n, edges: es): Set[n] \ Iterable.Aef[es] with Iterable[es], Order[n] where Iterable.Elm[es] ~ (n, n) =
        region rc {
            // collect edge lists
            let edgeMap = MutMap.empty(rc);
            foreach((start, end) <- edges) {
                let startEdges = MutMap.getOrElsePut!(start, MutList.empty(rc), edgeMap);
                MutList.push!(end, startEdges)
            };
            // define the reachable set
            let reachable = MutSet.empty(rc);
            // explore graph depth first by task list
            let taskList = MutDeque.empty(rc);
            taskList |> MutDeque.pushFront(origin);
            def whileLoop() = {
                MutDeque.popFront(taskList) |> Option.forEach(node -> {
                    // this node has now been reached
                    reachable |> MutSet.add!(node);
                    // add all non-reached end points to tasklist
                    let endPoints = MutMap.getWithDefault(node, MutList.empty(rc), edgeMap);
                    MutList.forEach(nextNode -> {
                        let alreadyReached = MutSet.memberOf(nextNode, reachable);
                        if (not alreadyReached)
                            taskList |> MutDeque.pushFront(nextNode)
                        else ()
                    }, endPoints);
                    whileLoop()
                })
            };
            whileLoop();
            MutSet.toSet(reachable)
        }

    @test
    pub def test(): Bool = validate(reachable)
}

pub def main(): Unit \ IO =
    validate(Reachable.Imperative.reachable) |> println

pub def graph1(): Set[(Int32, Int32)] = Set#{}

pub def graph2(): Set[(Int32, Int32)] = Set#{
    (1, 2), (2, 3), (3, 4), (4, 1)
}

pub def graph3(): Set[(Int32, Int32)] = Set#{
    (1, 2), (1, 3), (4, 5)
}

pub def graph4(): Set[(Int32, Int32)] = Set#{
    (2, 3), (3, 4)
}

pub def graph5(): Set[(Int32, Int32)] = Set#{
    (4, 5), (5, 6), (4, 3), (4, 2), (12, 13), (29, 4)
}

pub def validate(reachable: Int32 -> Set[(Int32, Int32)] -> Set[Int32]): Bool = {
    let test1 = reachable(42, graph1()) == Set#{42};
    let test2 = reachable(1, graph2()) == Set#{1, 2, 3, 4};
    let test3 = reachable(1, graph3()) == Set#{1, 2, 3};
    let test4 = reachable(1, graph4()) == Set#{1};
    let test5 = reachable(4, graph5()) == Set#{2, 3, 4, 5, 6};
    test1 and test2 and test3 and test4 and test5
}
```

magnus-madsen:
Add this to `TypeReconstruction.scala`:

```
    case KindedAst.Expr.Apply(exp, exps, tvar, evar, loc) =>
      val e = visitExp(exp)
      val es = exps.map(visitExp(_))

      val applyEff = subst(evar)
      val arrowEff = e.tpe.typeArguments.headOption

      if (Some(applyEff) != arrowEff) {
        println(arrowEff.getOrElse("<None>"))
        println(applyEff)
        println(loc.format)
        println()
      }

      TypedAst.Expr.Apply(e, es, subst(tvar), applyEff, loc)
```

and it shows lots of:

```
Pure
e104501 & Aef['26648]
Foldable.flix:100:46

(e104501 & Aef['26648]) + Aef['26648]
Aef['26648]
Foldable.flix:100:9

Aef['26648]
Aef['26648]
Foldable.flix:163:9

e26689 + Aef['26648]
e26689 + Aef['26648]
Foldable.flix:56:9

Aef['26648]
Aef['26648]
Foldable.flix:258:9

Aef['26648]
Aef['26648]
Foldable.flix:128:7

e26831 + Aef['26648]
Aef['26648] + e26831
Foldable.flix:179:9

```

magnus-madsen:
Observation `Aef['26648]` and `Aef['26648]` are not equal if they have different source locations.

This is because Type.AssocType is a case class, but notably Type.Apply and Type.Cst ignores source locations.

Not that

magnus-madsen:
The equality thing is bad, but not the cause.

main/src/ca/uwaterloo/flix/language/ast/TypeConstructor.scala
JonathanStarup: Could we do this in its own PR to make this easier to digest?

PR #7543, opened 2024-04-03
feat: use `tryPick` over `pick` in `tryPickTypeNoWild`

herluf-ba:
Found a small error while testing


magnus-madsen (approved):

magnus-madsen:
Test case or?

herluf-ba:
I caught this via `CompilerSuite` with `ToJava.flix` so I don't think it needs it's own test case.

PR #7551, opened 2024-04-07
fix: include public companion mods of private items in HTML documentation

sockmaster27:


sockmaster27:
Defers pairing of companion modules until after filtering

sockmaster27:
Fixes https://github.com/flix/flix/pull/7029#issuecomment-1951214123

magnus-madsen (approved):

PR #7562, opened 2024-04-10
feat: add `Spawn` and `Region` to verifier

felix-berg:
Currently fails on `par (...) yield`-expressions.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
@@ -369,6 +369,24 @@ object Verifier {
         case AtomicOp.Cast =>
           tpe
 
+        case AtomicOp.Region => tpe match {
+          case MonoType.Unit | MonoType.Region => tpe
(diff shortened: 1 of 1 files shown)


magnus-madsen:
> Currently fails on `par (...) yield`-expressions.

Interesting. We can have @JonathanStarup take a look :)

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -369,6 +369,19 @@ object Verifier {
         case AtomicOp.Cast =>
           tpe
 
+        case AtomicOp.Region =>
+          check(expected = MonoType.Region)(actual = tpe, loc)
+
+        case AtomicOp.Spawn =>
+          val List(t1, t2) = ts
+          t1 match {
+            case MonoType.Arrow(List(MonoType.Unit), _) => ()
```
JonathanStarup, line 378: In the other PR you also check for a `Unit -> _` functions, you should choose one method and do it the same way both places.

JonathanStarup (approved):

magnus-madsen (approved):

PR #7568, opened 2024-04-11
feat: add FuzzDuplicateLines and FuzzSwapLines

herluf-ba:
Adds fuzz testing for line-duplication and line swaps.
Line swaps are disabled by default since they take a long time to run. See comment in source.


magnus-madsen:
>  Compiler Tests / build (pull_request) Successful in 14m 

This already explodes test time. Can you add some heuristics - like I had - to bring the time down to 8m or so. (Lets say at most 1min for the fuzz tests).

main/test/flix/fuzzers/FuzzDeleteLines.scala
```diff
@@ -22,21 +22,16 @@ import org.scalatest.funsuite.AnyFunSuite
...
-  test("simple-card-game") {
-    val input = Files.lines(Paths.get("examples/simple-card-game.flix"))
-    compileAllLinesLessOne(input)
-  }
-
-  test("the-ast-typing-problem-with-polymorphic-records") {
-    val input = Files.lines(Paths.get("examples/the-ast-typing-problem-with-polymorphic-records.flix"))
-    compileAllLinesLessOne(input)
-  }
+  private val testFiles = List(
```
magnus-madsen, line 27: testFiles -> TestFiles (everywhere)

magnus-madsen: Nvm. See below.

main/test/flix/fuzzers/FuzzDeleteLines.scala
```diff
@@ -22,21 +22,16 @@ import org.scalatest.funsuite.AnyFunSuite
...
+    "simple-card-game" -> Files.lines(Paths.get("examples/simple-card-game.flix")),
+    "the-ast-typing-problem-with-polymorphic-records" -> Files.lines(Paths.get("examples/the-ast-typing-problem-with-polymorphic-records.flix")),
+    "using-channels-and-select" -> Files.lines(Paths.get("examples/using-channels-and-select.flix")),
+  )
 
-  test("using-channels-and-select") {
-    val input = Files.lines(Paths.get("examples/using-channels-and-select.flix"))
-    compileAllLinesLessOne(input)
+  testFiles.foreach {
+    case (name, input) => test(s"$name-delete-lines")(compileAllLinesLessOne(name, input))
```
magnus-madsen, line 34: We try hard to avoid programmatic construction of test cases because it sometimes breaks in IDEA etc.

So, I would really prefer to go back to the old style.

main/test/flix/fuzzers/FuzzerSuite.scala
```diff
@@ -19,5 +19,10 @@ import org.scalatest.Suites
...
 class FuzzerSuite extends Suites(
   new FuzzPrefixes,
-  new FuzzLines
+  new FuzzDeleteLines,
+  new FuzzDuplicateLines,
+  // Swap lines takes a _long_ time to run.
+  // In a file with just 100 lines, there is (100 * 99) / 2 unique non-equal swaps.
+  // That's 4950 compiles. Assuming each take 1sec to run that ends up at 1.3 hours.
+  //  new FuzzSwapLines,
 )
```
magnus-madsen, line 28: I think you should implement two FuzzerSuites:

- One that can run in one minute (and does limited, but deterministic fuzzing).
- And one that aims to take 1 hour to run (also deterministic).

The latter would not be enabled by default.

Perhaps we can have a gradle command `testFuzzer` or similar.

magnus-madsen:
Did you try to run this for extended hours on the Big Machine (TM)?

herluf-ba:
Yep. I ran duplicate line, delete line and prefixes on BigMachine™ on the whole standard library. I skipped swap lines since I ballparked that it would finish in a couple of months.

main/test/flix/fuzzers/FuzzerSuite.scala
herluf-ba: I added heuristics to all fuzz tests now. I did not add the testFuzzer command, but it's really easy just to increase the `N` values for each fuzz test suite if we want to hack that together sometime 👍

herluf-ba:
@magnus-madsen This is ready for another look 👍

main/test/flix/fuzzers/FuzzSwapLines.scala
```diff
@@ -0,0 +1,74 @@
...
+    */
+  private def compileWithSwappedLines(name: String, stream: java.util.stream.Stream[String]): Unit = {
+    val lines = stream.iterator().asScala.toList
+    val numLines = lines.length
+    val numSwapLinesFixed = numLines.min(numSwapLines)
+    val step = numLines / numSwapLinesFixed
+
+    val flix = new Flix()
+    flix.compile()
+    for (i_ <- 0 until numSwapLinesFixed) {
```
magnus-madsen, line 64: Why i and j with underscore?

main/test/flix/fuzzers/FuzzSwapLines.scala
```diff
@@ -0,0 +1,74 @@
...
+    compileWithSwappedLines("simple-card-game", lines)
+  }
+
+  test("using-channels-and-select") {
+    val lines = Files.lines(Paths.get("examples/using-channels-and-select.flix"))
+    compileWithSwappedLines("using-channels-and-select", lines)
+  }
+
+  test("the-ast-typing-problem-with-polymorphic-records") {
+    val lines = Files.lines(Paths.get("examples/the-ast-typing-problem-with-polymorphic-records.flix"))
```
magnus-madsen, line 46: Could we store the path once and then use path.getFileName().toString?

main/test/flix/fuzzers/FuzzDeleteLines.scala
```diff
@@ -22,40 +22,45 @@ import org.scalatest.funsuite.AnyFunSuite
...
 
   /**
     * We compile all variants of the given program where we omit a single line.
-    *
     * For example, we omit line 1 and compile the program. Then we omit line 2 and compile the program. And so forth.
-    *
     * The program may not be valid: We just care that it does not crash the compiler.
     */
-  private def compileAllLinesLessOne(stream: java.util.stream.Stream[String]): Unit = {
+  private def compileAllLinesExceptOne(name: String, stream: java.util.stream.Stream[String]): Unit = {
```
magnus-madsen, line 51: Is the comment still up to date?

main/test/flix/fuzzers/FuzzDuplicateLines.scala
```diff
@@ -0,0 +1,68 @@
...
+import ca.uwaterloo.flix.api.Flix
+import org.scalatest.funsuite.AnyFunSuite
+
+import java.nio.file.{Files, Paths}
+import scala.jdk.CollectionConverters._
+
+class FuzzDuplicateLines extends AnyFunSuite with TestUtils {
+
+  /**
+    * Number of lines to duplicate from each file.
```
magnus-madsen, line 28: Is that right? The comment seems misleading.

magnus-madsen (commented):
I think some more polish would be good :)

main/test/flix/fuzzers/FuzzSwapLines.scala
herluf-ba: Like so?
```scala
test("simple-card-game") {
    val filepath = Paths.get("examples/simple-card-game.flix")
    val lines = Files.lines(filepath)
    compileWithDuplicateLine(filepath.getFileName.toString, lines)
  }
```

main/test/flix/fuzzers/FuzzDuplicateLines.scala
herluf-ba: I agree, it is now
```scala
  /**
    * Number of variants to make of each file. Each variant has a single line duplicated.
    */
  private val N = 10
```

main/test/flix/fuzzers/FuzzSwapLines.scala
herluf-ba: That's so further down it can be multiplied with the step size to get the "real" `i`.
I changed it to some names that better reflect this: `i_ -> i` and `i -> iStepped`.

main/test/flix/fuzzers/FuzzDeleteLines.scala
herluf-ba: Updated it 👍

main/test/flix/fuzzers/FuzzSwapLines.scala
magnus-madsen: Yeah

magnus-madsen:
Ping me when I should look again.

magnus-madsen:
How much time do the tests take?

herluf-ba:
The pipeline finished in 9m. Did you want it to be 8 or under? I can tweak the parameters then

magnus-madsen:
Yeah, maybe make it a but smaller

magnus-madsen (approved):

PR #7569, opened 2024-04-11
feat: use `WeededAst` from new parser/weeder

herluf-ba:
Uses `WeededAst` produced by the new pipeline rather than the old. 
The new pipeline is no longer run silently, but is still after the old one.
That means it only ever sees code the old pipeline deems fine.


magnus-madsen:
@herluf-ba some conflicts and test cases to look at :)

magnus-madsen:
@herluf-ba I have released a new stable version, so this can be merged in, once it passes tests, to begin experiments.

herluf-ba:
@magnus-madsen 
I have made PR's #7596  and #7597 to fix the two bugs that make the pipeline fail here. Once those land we should be ready to merge.

herluf-ba:
@mlutze The community build of fcwg is failing due to use of `for` in `TestTransform.flix`. Could you change it?

mlutze:
> @mlutze The community build of fcwg is failing due to use of `for` in `TestTransform.flix`. Could you change it?

done

herluf-ba:
See #7601 👍

magnus-madsen (approved):

magnus-madsen:
@JonathanStarup @mlutze @jaschdoc @sockmaster27 We have moved onto stage 2 of merging the new parser.

We are now running the old parser first. If the file parses, then we give it it to the new parser and use its result.

@jaschdoc @sockmaster27 Next time you are working, I would like you to take 15min to build a jar from master and play around in VSCode and see if you can make the new parser misbehave. Take a program and try to add/remove things from it and ensure that the new parser is well-behaved.

jaschdoc:
Can't seem to narrow it down anymore than this but
```scala
mod Test.Dec.Assoc.Type.Eff.Multiple {

    ////////////////////////////////////////////////////////
    // Type Definitions                                   //
    ////////////////////////////////////////////////////////

    eff _Observe {
        pub def captureInt32(x: Int32): Unit
        pub def captureString(x: String): Unit
    }



    ////////////////////////////////////////////////////////
    // Polymorphic functions using Transformable          //
    ////////////////////////////////////////////////////////
//
//    pub def transformAndCheckEq(eq: Transformable.B[a], x: a): Bool \ Transformable.E[a] with Transformable[a], Eq[Transformable.B[a]] =
//        Transformable.transform(x) == eq

}

```

produces
```
❌ -- Parse Error -------------------------------------------------- file:///.../TestEffMultiple.flix

>> Parse Error: Expected declaration but found CurlyR

21 | }
     ^
     Here
Syntactic Context: Unknown.

❌ -- Parse Error -------------------------------------------------- file:///.../TestEffMultiple.flix

>> Parse Error: Expected CurlyR, but found Eof

21 | }
      ^
      Here
Syntactic Context: Unknown.

```

magnus-madsen:
@jaschdoc Can you open a new ticket? Thanks!

PR #7570, opened 2024-04-14
feat: add `XCompilerMemory`

magnus-madsen:


mlutze (approved):
Would be nice to get JSON output so we can dashboard this

magnus-madsen:
> Would be nice to get JSON output so we can dashboard this

Yes. We need a CI/dashboard tzar.

PR #7574, opened 2024-04-14
feat: reduce memory usage of columns in `Token` and `SourceLocation`

magnus-madsen:


magnus-madsen:
Note: I must use `toShort` because the addition of shorts returns ints!

mlutze (approved):

PR #7589, opened 2024-04-16
feat: reduce memory usage of `Type.Apply` in `MonoTypes`

magnus-madsen:


main/src/ca/uwaterloo/flix/language/phase/MonoTypes.scala
```diff
@@ -308,7 +310,14 @@ object MonoTypes {
...
           tpe
         case Type.Apply(tpe1, tpe2, loc) =>
-          Type.Apply(visitType(tpe1), visitType(tpe2), loc)
+          val t1 = visitType(tpe1)
+          val t2 = visitType(tpe2)
+          // Performance: Reuse tpe0, if possible.
+          if ((t1 eq tpe1) && (t2 eq tpe2)) {
+            tpe
+          } else {
+            Type.Apply(t1, t2, loc)
```
mlutze, line 319: Should we have a helper for this?

mlutze (approved):

main/src/ca/uwaterloo/flix/language/phase/MonoTypes.scala
magnus-madsen: Its kind of hard, because its specific. I would wait a bit and see.

magnus-madsen: We could funnel everything through Type.mkApply-- but I would wait just a bit before doing that,

PR #7595, opened 2024-04-17
feat: remove backend ast caching / asap pretty printing

### LINKED ISSUE #7576: Reduce backend ast caching
AST's in the backend are cached to enable ast printing on crash. This requires memory ofc. can this be reduced?

JonathanStarup:
Fixes #7576

Changes:
- `--Xprint-phases` is now on/off
- Show Ast LSP command doesn't take a phase argument and returns an object with the `build/asts/` path
- printing is put into the `Flix.phase` command, since it already has the phase name
- phases supply their specific printer, in their body
- ast are pretty printed asap if set in options, instead of after the fact
- crashing no longer prints the asts


magnus-madsen:
What's the status on this PR?

main/src/ca/uwaterloo/flix/api/lsp/provider/ShowAstProvider.scala
```diff
@@ -32,17 +33,14 @@ object ShowAstProvider {
...
-      val (closestPhase, closestPrettyPrinter) = Similarity.closestMatch(phase, phasesWithNames)
-      astObject(closestPhase, closestPrettyPrinter())
+  def showAst(phase: String)(implicit index: Index, root: Option[Root], flix: Flix): JObject = {
+    val names = AstPrinter.phaseNames().m1.keys.map(s => (s, s)).toMap
+    val closestName = Similarity.closestMatch(phase, names)
+    val oldOpts = flix.options
+    flix.setOptions(oldOpts.copy(xprintphase = Set(closestName)))
+    flix.compile()
+    flix.setOptions(oldOpts)
+    astObject(closestName, s"Check in build/asts/${closestName}.flixir")
```
JonathanStarup, line 43: It works, but it breaks `ShowAst`. Its not easy to hold onto the string so it can be given to lsp, so maybe the easier choice is just that the compiler should give vscode the path to the ast file, which vscode can then open? instead of giving the string back directly

magnus-madsen: Interesting approach. How about you refactor towards that.

@sockmaster27 Can we make it so that ShowAst get passed back a path which it then opens?

sockmaster27: I don't see why not

magnus-madsen: @JonathanStarup Can you make the changes to the compiler? Then once that is in place, @sockmaster27 can make the changes to VSCode.

main/src/ca/uwaterloo/flix/api/lsp/provider/ShowAstProvider.scala
```diff
@@ -32,20 +35,21 @@ object ShowAstProvider {
...
-      astObject(closestPhase, closestPrettyPrinter())
+  private def closestPhaseName(phase: String) = {
+    val phaseNames = AstPrinter.phaseNames().m1.keys
+    Similarity.closestMatch(phase, phaseNames.map(s => (s, s)).toMap)
   }
 
-  private def astObject(phase: String, text: String): JObject = {
-    ("title" -> s"$phase.$IrFileExtension") ~ ("text" -> text)
+  private def pathObject(path: Path): JObject = {
+    ("path" -> path.toAbsolutePath.toString)
```
JonathanStarup, line 53: Now i return an object with a path value. I'm not fully sure how to make this path into a string that the receiver will understand but here is something. To note for implementing the other side is that
- show ast always triggers a full recompile, that might fail
- the returned path *should* exist but it might not if we mess up in the list of phase names. it would be nice if that doesn't crash vscode.

main/src/ca/uwaterloo/flix/api/lsp/provider/ShowAstProvider.scala
JonathanStarup: New plan: show ast provider just returns a path where they all live

JonathanStarup:
@magnus-madsen I think this PR is in a good spot now, the code is pretty clean compared to before

magnus-madsen (approved):

magnus-madsen:
Thanks!

PR #7597, opened 2024-04-17
fix: Call expression followed by Record Select

herluf-ba:
This fixes a bug where a record select after a call was not handled correctly, resulting in unequal WeededAsts.
Example: `myFunctionThatReturnsARecord().field`.

The fix resurfaced a dear old friend of mine. The issue of using `.` as a terminator for fixpoint constraints in a set.
This is now also fixed (again), mainly by removing stuff!


magnus-madsen:
> The fix resurfaced a dear old friend of mine.

Haha!

magnus-madsen (approved):

PR #7603, opened 2024-04-19
feat: Better parser recovery with recover sets

herluf-ba:
This PR adds recover sets to the parser. This makes the parser produce far fewer and more local errors. Local in the sense that the error is close to the actual syntactic issue. A recover set is a set of `TokenKind`s that is used to make decisions about error-recovery. Here is a short motivation:

Generally, the parser can be thought off as a series of nested loops:
```rust 
loop {
  // Parse declarations
  loop {
    // Parse block of expressions
    loop {
      // Parse tuple expression
    }
  }
}
```
When an _unexpected_ token is found within a loop the parser needs to decide whether to 
1. Wrap the token in an error an continue parsing whatever rule it was parsing.
2. Break the loop and delegate error handling to the calling rule.

An example of when we want 1. could be:
```scala
def main(a: Int32, , c: Int32): Int32 = a + b + c
//                 ^ We still want to parse 'c' as a parameter so just wrap this ',' in an error and carry on.
```

An example of when we want 2. could be:
```scala
// We still want to parse main as a function, so it makes sense to stop parsing enum variants once we see 'def'.
enum Fruits {
  case Apple,

def main(): Unit = ()
```

Recover sets help the parser decide between 1. and 2. when it encounters an unexpected token. The rule is:
> When an unexpected token, `t` is found, if `t` is in the recover set of the current context, use 2. otherwise use 1.

So with the first example, since Comma is not in the recover set of the parameterlist rule, we carry on parsing parameters.
And in the second example, since KeywordDef _is in_ the recover set of the enum variant rule, we break out of the enum loop into the declaration loop, where the 'def' can get picked up.

## Testing
To test recoveries a new suite `TestParserRecovery` has been added. It tests recovery by running a compiler check on malformed code and checking that:
1. A ParseError is produced.
2. `main` is still defined in the `WeededAst`, ie. the parser recovered and understood the main function properly.
The test suite is *disabled* for now, because the previous parser (which still runs) does not pass the tests.

## Performance
Right now recover-sets are implemented with `List`, which is a performance concern. A solution using bit-sets is probably best. I would like to merge this first though, and improve performance in a later PR.

A `--XBenchmark` run reveals that performance is comparable to before recoversets. ie. we are still wanting for a ~2x speedup to be as fast as before.
```
﻿﻿  Weeder                          97,2 ms (02,1%)
  Parser                         285,4 ms (06,0%)
  Lexer                          149,2 ms (03,1%)
  Weeder2                        175,7 ms (03,7%)
  Parser2                        470,7 ms (09,9%)
```


magnus-madsen:
Looking forward to reviewing 👍

magnus-madsen:
A crash:

```
enum Legumes[] { Chickpea, Beans }

def main(): Unit = ()
```

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -529,6 +541,32 @@ object Parser2 {
   private val NAME_TAG = List(TokenKind.NameUpperCase, TokenKind.KeywordPure, TokenKind.KeywordImpure)
   private val NAME_PREDICATE = List(TokenKind.NameUpperCase)
 
+  private val MODIFIERS: List[TokenKind] = List(TokenKind.KeywordSealed, TokenKind.KeywordLawful, TokenKind.KeywordPub, TokenKind.KeywordInline, TokenKind.KeywordOverride)
+
+  /**
+    * Sets of the [[TokenKind]] that can appear as the first token of different concepts.
+    * For instance, FIRST_DECL are all the tokens that can be the first of a declaration ('enum', 'def', 'mod' and so on).
+    */
+  private val FIRST_DECL: List[TokenKind] = MODIFIERS ::: List(TokenKind.CommentDoc, TokenKind.Annotation, TokenKind.KeywordMod, TokenKind.KeywordDef, TokenKind.KeywordEnum, TokenKind.KeywordTrait, TokenKind.KeywordInstance, TokenKind.KeywordType, TokenKind.KeywordEff, TokenKind.KeywordRestrictable)
```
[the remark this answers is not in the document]
magnus-madsen: And if they have no order, why not make them into a `Set[TokenKind]`?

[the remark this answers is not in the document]
magnus-madsen: Also: https://en.wikipedia.org/wiki/Optimal_binary_search_tree

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -529,6 +541,32 @@ object Parser2 {
...
+  private val FIRST_RECORD_OP: List[TokenKind] = List(TokenKind.Plus, TokenKind.Minus, TokenKind.NameLowerCase)
+
+  /**
+    * Sets of [[TokenKind]]s used for recovery.
+    * For instance, when parsing sequences of expressions in a loop and an unexpected token is found, a recovery means breaking the loop.
+    * An example would be:
+    * def foo(): Unit = bar(1,
+    * enum Legumes { case Beans, Chickpeas }
+    * Here we recover from parsing argument expressions of the unfinished call to `bar` and still discover `Legumes`, because [[TokenKind.KeywordEnum]] is part of [[RECOVERY_EXPR]].
+    */
```
magnus-madsen, line 564: I would like a bit more commentary and also two examples.

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -616,9 +654,13 @@ object Parser2 {
...
     if (at(TokenKind.DotCurlyL)) {
       val mark = open()
-      separated(() => aliasedName(NAME_USE), checkForItem = () => atAny(NAME_USE))
-        .within(TokenKind.DotCurlyL, TokenKind.CurlyR)
-        .zeroOrMore()
+      new Separated(
+        itemName = "name",
+        getItem = () => aliasedName(NAME_USE),
+        checkForItem = () => atAny(NAME_USE),
+        recoverOn = RECOVER_DECL ::: (FIRST_TOP_LEVEL_USE :+ TokenKind.Semi),
```
magnus-madsen, line 661: Repeatedly performing this concatenation seems expensive.

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -632,9 +674,13 @@ object Parser2 {
...
     if (at(TokenKind.DotCurlyL)) {
       val mark = open()
-      separated(() => aliasedName(NAME_JAVA))
-        .within(TokenKind.DotCurlyL, TokenKind.CurlyR)
-        .zeroOrMore()
+      new Separated(
+        itemName = "name",
+        getItem = () => aliasedName(NAME_JAVA),
+        checkForItem = () => atAny(NAME_JAVA),
+        recoverOn = RECOVER_DECL ::: (FIRST_TOP_LEVEL_USE :+ TokenKind.Semi),
```
magnus-madsen, line 681: I see the same set is also reused.

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -2710,20 +2840,28 @@ object Parser2 {
...
-        .optionallyWith(TokenKind.Bar, () => name(NAME_VARIABLE))
-        .zeroOrMore()
+      new Separated(
+        itemName = "schema term",
+        getItem = schemaTerm,
+        checkForItem = () => atAny(NAME_PREDICATE),
+        delimiters = (TokenKind.HashCurlyL, TokenKind.CurlyR),
+        recoverOn = RECOVER_TYPE,
+        optionallyWith = Some(TokenKind.Bar, () => name(NAME_VARIABLE))
+      ).zeroOrMore()
```
magnus-madsen, line 2850: Do you also construct a Separated and then _immediately_ call zeroOrMore or oneOrMore?

magnus-madsen: If so, perhaps zeroOrMore and oneOrMore should be helper functions.

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -129,7 +129,7 @@ object Weeder2 {
         val qname = Name.QName(tree.loc.sp1, namespace, ident, tree.loc.sp2)
         val res = Validation.success(UseOrImport.Use(qname, alias, tree.loc))
         // Check for illegal alias
-        if (ident.name.charAt(0).isUpper != alias.name.charAt(0).isUpper) {
+        if ((ident.name.nonEmpty && alias.name.nonEmpty) && ident.name.charAt(0).isUpper != alias.name.charAt(0).isUpper) {
```
magnus-madsen, line 132: Do we need to check for this pattern elsewhere? Can you search for ident.name usages?

main/test/ca/uwaterloo/flix/TestUtils.scala
```diff
@@ -35,14 +35,20 @@ trait TestUtils {
   def compile(s: String, o: Options): Validation[CompilationResult, CompilationMessage] =
     new Flix().setOptions(o).addSourceCode("<test>", s).compile()
 
+  /**
+    * Checks the given input string `s` with the given compilation options `o`.
+    */
+  def check(s: String, o: Options): Validation[TypedAst.Root, CompilationMessage] =
```
magnus-madsen, line 41: Move above compile.

main/test/ca/uwaterloo/flix/TestUtils.scala
```diff
@@ -35,14 +35,20 @@ trait TestUtils {
...
     errors.map(_.message(Formatter.NoFormatter)).mkString("\n\n")
   }
 
   /**
-   * Asserts that the validation is a failure with a value of the parametric type `T`.
-   */
-  def expectError[T](result: Validation[CompilationResult, CompilationMessage])(implicit classTag: ClassTag[T]): Unit = result.toHardResult match {
+    * Private generic version. Asserts that the validation is a failure with a value of the parametric type `T`.
+    */
+  private def expectErrorGen[R, T](result: Validation[R, CompilationMessage])(implicit classTag: ClassTag[T]): Unit = result.toHardResult match {
```
magnus-madsen, line 51: private functions should go at the bottom.

main/test/ca/uwaterloo/flix/TestUtils.scala
```diff
@@ -56,10 +62,31 @@ trait TestUtils {
...
 
+  /**
+   * Asserts that the compilation result is a failure with a value of the parametric type `T`.
+   */
+  def expectError[T](result: Validation[CompilationResult, CompilationMessage])(implicit classTag: ClassTag[T]): Unit = expectErrorGen[CompilationResult, T](result)
 
+  /**
+    * Asserts that the result of a compiler check is a failure with a value of the parametric type `T`.
+    */
+  def expectErrorOnCheck[T](result: Validation[TypedAst.Root, CompilationMessage])(implicit classTag: ClassTag[T]): Unit = expectErrorGen[TypedAst.Root, T](result)
```
magnus-madsen, line 75: Move up

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -213,15 +213,31 @@ object Parser2 {
     stack.head
   }
 
+  /**
+    * Get first non-comment previous position of the parser as a [[SourceLocation]].
+    * TODO: It might make sense to seek the first non-comment position here.
+    */
+  private def previousSourceLocation()(implicit s: State): SourceLocation = {
```
magnus-madsen, line 220: I am confused. The token is the previous one, but the source location is similar to currentSourceLocation. What gives?

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -333,13 +349,15 @@ object Parser2 {
 
   /**
     * Checks if the parser is at a token of a specific `kind`.
+    * TODO: This is done __alot__. Maybe we can use bitsets to be faster?
```
magnus-madsen, line 352: For checking a single token nothing is faster here.

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -378,22 +396,33 @@ object Parser2 {
...
-      // Better still, if it is a DocComment we can say "DocComments are allowed above xyz".
-      val error = ParseError(s"Expected $kind, but found ${nth(0)}", SyntacticContext.Unknown, currentSourceLocation())
-      advanceWithError(error)
+    if (eat(kind)) {
+      return
     }
+
+    val mark = open()
+    val error = nth(0) match {
+      case TokenKind.CommentLine | TokenKind.CommentBlock => ParseError(s"Invalid comment", SyntacticContext.Unknown, previousSourceLocation())
```
magnus-madsen, line 405: We try to avoid or patterns because they can be deceptive to read/scan.

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
magnus-madsen, line 550: I am debating if we should move these constants into the TokenKind class or their own class, instead of having them in the middle of the Parser.

magnus-madsen:
Another crash (but where the fault is may be less clear):

```
def makeSquare(square: Square, g: ##java.awt.Graphics2D): Unit \ IO = {
    import java_new java.awt.Color(Int32, Int32, Int32): ##java.awt.Color \ IO as newColor;
    import java.awt.Graphics2D.setColor(##java.awt.Color): Unit \ IO;
    import 
    setColor(g, newColor(54, 54, 54));
    import java.awt.Graphics2D.fillRect(Int32, Int32, Int32, Int32): Unit \ IO;
    fillRect(g, square.minCorner.x, square.minCorner.y, square.dim.w, square.dim.h)
}
```

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -529,6 +541,32 @@ object Parser2 {
   private val NAME_TAG = List(TokenKind.NameUpperCase, TokenKind.KeywordPure, TokenKind.KeywordImpure)
   private val NAME_PREDICATE = List(TokenKind.NameUpperCase)
 
+  private val MODIFIERS: List[TokenKind] = List(TokenKind.KeywordSealed, TokenKind.KeywordLawful, TokenKind.KeywordPub, TokenKind.KeywordInline, TokenKind.KeywordOverride)
```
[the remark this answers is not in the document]
herluf-ba: Yet order does not matter.

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -529,6 +541,32 @@ object Parser2 {
...
+  private val FIRST_TYPE: List[TokenKind] = List(TokenKind.NameUpperCase, TokenKind.NameMath, TokenKind.NameGreek, TokenKind.Underscore, TokenKind.NameLowerCase, TokenKind.KeywordUniv, TokenKind.KeywordPure, TokenKind.KeywordFalse, TokenKind.KeywordTrue, TokenKind.ParenL, TokenKind.CurlyL, TokenKind.HashCurlyL, TokenKind.HashParenL, TokenKind.NameJava, TokenKind.AngleL, TokenKind.KeywordNot, TokenKind.Tilde, TokenKind.KeywordRvnot, TokenKind.KeywordStaticUppercase)
+  private val FIRST_PATTERN: List[TokenKind] = List(TokenKind.NameLowerCase, TokenKind.NameGreek, TokenKind.NameMath, TokenKind.Underscore, TokenKind.KeywordQuery, TokenKind.LiteralString, TokenKind.LiteralChar, TokenKind.LiteralFloat32, TokenKind.LiteralFloat64, TokenKind.LiteralBigDecimal, TokenKind.LiteralInt8, TokenKind.LiteralInt16, TokenKind.LiteralInt32, TokenKind.LiteralInt64, TokenKind.LiteralBigInt, TokenKind.KeywordTrue, TokenKind.KeywordFalse, TokenKind.LiteralRegex, TokenKind.KeywordNull, TokenKind.NameUpperCase, TokenKind.ParenL, TokenKind.CurlyL, TokenKind.Minus)
+  private val FIRST_TOP_LEVEL_USE: List[TokenKind] = List(TokenKind.KeywordUse, TokenKind.KeywordImport)
```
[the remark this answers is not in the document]
herluf-ba: Before any `mod`ules there can be a number of uses and imports according to the grammar. These are the tokens in the FIRST set of that rule. They are used in the recover sets of a top-level use for instance.

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -529,6 +541,32 @@ object Parser2 {
...
+  private val FIRST_RECORD_OP: List[TokenKind] = List(TokenKind.Plus, TokenKind.Minus, TokenKind.NameLowerCase)
```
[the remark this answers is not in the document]
herluf-ba: The FIRST set of tokens that can occur in a record-op. So `+`, `-` or a NameLower for extend, restrict and update respectively.  They are used when parsing a record operation, to check if there are more items to parse or the record literal has ended.
```scala
{ x = 1, -y, +z | w }
```

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
herluf-ba: I'm not sure I understand this. Did you mean that 
1. `zeroOrMore` should be a function that internally constructs a `Separated` and calls it or that
2. `zeroOrMore` should be a function the takes the same parameters as used for construction of `Separated` and just does the computation directly. This refactor would make us able to delete `Separated` I believe.
3. Something other than that?

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
herluf-ba: I did such a search when I found that issue, with no results 👍 I think I searched both for `charAt` and also just `0` to see if I was indexing in a bad way anywhere.

herluf-ba: I did such a search when I found that issue, with no results 👍 I think I searched both for `charAt` and also just `0` to see if I was indexing in a bad way anywhere.

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
magnus-madsen: Does Separated have to be a class/object? It has no state, right?

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
herluf-ba: That's definitely another approach to encoding the information, just having a number of functions ala `isFirstType: TokenKind -> Boolean`. I think it makes sense to move to that actually...

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
herluf-ba: Would you rather I added a branch for `CommentLine` and `CommentBlock` with identical bodies?

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -423,91 +452,74 @@ object Parser2 {
     * ParieldFragments "par (x <- e1; y <- e2; z <- e3) yield ...".
     * and many many more...
     *
-    * Delimiters and separators differ, and can be configured using [[within]] and [[by]]. Afterwards call [[zeroOrMore]] to actually consume any tokens.
+    * Call [[zeroOrMore]] to actually consume any tokens.
     */
   private class Separated(
```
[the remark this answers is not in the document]
herluf-ba: I agree.

herluf-ba:
@magnus-madsen Thanks for the comments. There are some small fixes about comments and ordering that I am just going to add. But there are also two largerish refactors that I am going to make based on your feedback:
1. I will remove `Separated` in favor of functions `oneOrMore` and `zeroOrMore` to avoid unnecessary OOP.
2. I will refactor FIRST and RECOVER sets into functions ala `isFirstType` or `isRecoverExpr`. Each will take only a `TokenKind` and return a bool.

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
herluf-ba: I forgot to answer your question
> Another question: Could e.g. FIRST_TYPE be a Boolean function on each token kind?

Yes. The FIRST set of a grammar rule is static data. The tokens that are possible as the first token in an expression for instance, does not change. I think this data belongs in the parser as `isFirst*` and `isRecover*` functions though. Let me try and argue that! 

To me the concerns we have are:
1. **Where is it practical to store FIRST and RECOVER sets?** I think in the parser, because if it ever needs to change that means the grammar is changing, and then you are already working in the parser. This is also the spot where I think you most often need to read FIRST and RECOVER sets.
2. **How can we have fast member checks on FIRST and RECOVER sets?** This is crucial for performance. By capturing the data as functions in the parser, we can change those functions in pursuit of performance, with no refactoring needed on the call-site.

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -716,8 +716,9 @@ object Weeder2 {
       ) {
         case Some(expr) => Validation.success(expr)
         case None =>
+          // Fall silently back on Expr.Error. Parser should produce error here.
```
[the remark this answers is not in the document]
herluf-ba: We are returning a `success` because a `softFailure` would add another error message. I don't think we want that, because in the case of an syntactically incorrect program, I think it is best to output a syntax error only. A type error or resolution error is very likely to be of low quality, because the program is in a state that's very hard to reason about.

I agree that the `Validation.success(Expr.Error(...))` doesn't have a nice ring to it! But I think it is the only way to perpetuate the error, while not outputting anything to the programmer (to "fail silently", in a lack of better wording). I didn't want to make changes to `Validation` to facilitate "a failure with nothing to output".

> We don't have to do this anywhere else, why do we have to do it here?

I think we do have to do something similar elsewhere. In the Typer for instance, we should make sure not to produce an error ala "Expected Int32 but found Error". In general, any phase after the Parser should ignore error nodes as much as possible. They can't be reasoned with.

magnus-madsen:
@herluf-ba Seems like you introduced a parse error :) Not in the parser, but in the compiler itself :P

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
magnus-madsen: Yes

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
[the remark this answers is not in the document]
magnus-madsen: I think something like "Fall back on Expr.Error. The Parser has reported an error here." is ok.

herluf-ba:
> Seems like you introduced a parse error :) Not in the parser, but in the compiler itself :P

I must have gotten into the habit of writing faulty code from those test cases! That's bad muscle memory...

herluf-ba:
@magnus-madsen I think this is ready for another look. I have addressed all your comments. In particular:
- Fixed both crashes (thanks for those)
- Refactored into `zeroOrMore` and `oneOrMore` functions
- I __did not__ end up refactoring FIRST and RECOVER sets into functions. In the midst of doing that, I changed my mind, and now I believe its best to keep them as sets.
- I did a tiny bit of optimization because I discovered that `buildTree` was very slow. It's just 4 changed lines but gives a boost of ~2% on my machine. The reasoning is documented in comments.

magnus-madsen (approved):

magnus-madsen:
Merged. Thanks!

Issue #6326, opened 2023-08-13
Add `NewObject` completer


magnus-madsen:
I would like to write:

```
new ##java.io.InputStream ...
```

and then be offered two types of completions:

- [ ] Completion of `java.io.InputStream`
- [ ] Completion of all abstract methods/signatures in `InputStream` (i.e. a template to fill out).

magnus-madsen:
@xdesou01 If you have time, this would also be a fun one. Especially part B where you fill out the entire body of the class.

magnus-madsen:
@sockmaster27 You can continue from here: #6520

magnus-madsen:
Ill let @jaschdoc take a stab

PR #7605, opened 2024-04-19
feat: add `NewObject` completion

jaschdoc:
Closes https://github.com/flix/flix/issues/6326


jaschdoc:
Was testing out the completers and got an error from the program below (which is not a valid program but I was midway through typing it).

```scala
pub def main(): Unit \ IO =
    println("The sum from 0..42 is")


pub def f(x: Option[a]): Option[a] = match
```


```
#
#
#   head of empty list
#
# This is a bug in the Flix compiler. Please report it here:
#
# https://github.com/flix/flix/issues
#
# -- Flix Compiler --
#
# Flix Version : 0.46.0
#   incremental: All
#
# -- Java Virtual Machine --
#
# JVM Version  : 21.0.1 (2023-10-17)
# JVM Vendor   : GraalVM Community
# JAVA_HOME    : /home/_/.sdkman/candidates/java/21.0.1-graalce
# System       : Linux
#
# -- Stack Trace --
java.util.NoSuchElementException: head of empty list
	at scala.collection.immutable.Nil$.head(List.scala:662)
	at scala.collection.immutable.Nil$.head(List.scala:661)
	at ca.uwaterloo.flix.language.phase.TypeReconstruction$.visitExp(TypeReconstruction.scala:195)
	at ca.uwaterloo.flix.language.phase.TypeReconstruction$.visitDef(TypeReconstruction.scala:35)
	at ca.uwaterloo.flix.language.phase.Typer$.$anonfun$visitDef$1(Typer.scala:179)
	at ca.uwaterloo.flix.util.Validation$.mapN(Validation.scala:429)
	at ca.uwaterloo.flix.language.phase.Typer$.visitDef(Typer.scala:178)
	at ca.uwaterloo.flix.language.phase.Typer$.$anonfun$visitDefs$2(Typer.scala:162)
	at ca.uwaterloo.flix.util.ParOps$.$anonfun$parTraverseValues$1(ParOps.scala:106)
	at ca.uwaterloo.flix.util.ParOps$.$anonfun$parMap$2(ParOps.scala:64)
	at java.base/java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1423)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:387)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1312)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1843)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1808)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:188)
```

magnus-madsen:
Thanks-- can you open a ticket for the bug?

magnus-madsen:
For the completer, it should fill in all methods you have to implement.

magnus-madsen:
For example, if you write `new java.util.Comparator` then it should fill in the `compare` method you have to implement.

jaschdoc:
> For example, if you write `new java.util.Comparator` then it should fill in the `compare` method you have to implement.

Thanks! Working on it :smile: :runner: :hammer:
Just had to figure out what I was doing and what the goal was hehe

magnus-madsen:
Goal is like Intellij IDEA when you get it to fill out everything you have to implement. So a bit like `InstanceCompleter`.

jaschdoc:
Hmm it's kind of awkward to do completions as the new parser fails when writing `new ##java.lang.Runnable` or just `new java.lang.Runnable` so it seems the completion is not even run? I'm not sure if the parser has anything to do with the LSP side of things, though.

magnus-madsen:
> Hmm it's kind of awkward to do completions as the new parser fails when writing `new ##java.lang.Runnable` or just `new java.lang.Runnable` so it seems the completion is not even run? I'm not sure if the parser has anything to do with the LSP side of things, though.

You can disable the new parser if you need to then we can look into when you have something.

jaschdoc:
Ok I think this is blocked...
When writing `import java.lang.Comparable` the compiler succeeds with parsing so the `ImportCompleter` is fired. However, when writing `new java.lang.Comparable` it fails parsing after `new` until a word followed by `{}` is encountered. The weird thing is that it follows the same syntax / structure as `match exp { }` but I can't reproduce the behavior of the `MatchCompleter` either (i.e., matching the regex).
I'm clueless at this point. Maybe @sockmaster27 knows what's up?

magnus-madsen:
What if you write `new new java.lang.R` then you should be able to select runnable and have everything filled out

jaschdoc:
> What if you write `new new java.lang.R` then you should be able to select runnable and have everything filled out

Doesn't work

magnus-madsen:
Are you trying in an expression context?

magnus-madsen:
I guess we should remember it is now called `java_new`

jaschdoc:
> I guess we should remember it is now called `java_new`

Good point.

jaschdoc:
> Are you trying in an expression context?

The term that comes after `new` is a `Type` not `Exp`.

jaschdoc:
I think I'm starting to narrow down how it works. So on the design side of things, do we want it to work like `ImportNewCompleter` where it fetches next package prefix and finally lists the classes / types in the package and then finally fills out the required functions (this is difficult)? Or do we only want to look at imported classes / interfaces (which is much easier to do at the moment)?

jaschdoc:
Ok, the core sort of works now. It finds the abstract classes and interfaces in the package we typed. However, I have no idea why no vscode suggestions appear...

magnus-madsen:
A few screenshots of different completions?

mlutze:
> Ok this is close to done. The only thing I'm not sure how to approach is that `formatType` doesn't prepend `##` to native types. This means that `NewObjectCompleter` has to do it manually. However, Arrays pose a bit of a challenge, since we want `formatType` to format these but also add `##` if the type argument is a native class. E.g., `Array[Array[java.lang.Object]]` should be `Array[Array[##java.lang.Object]]` (not adding regions). Maybe @mlutze has a suggestion for a fix?

I think our best option is probably to make a new formatter for generated code. We might be able to use the SimpleType interface.

jaschdoc:
> > Ok this is close to done. The only thing I'm not sure how to approach is that `formatType` doesn't prepend `##` to native types. This means that `NewObjectCompleter` has to do it manually. However, Arrays pose a bit of a challenge, since we want `formatType` to format these but also add `##` if the type argument is a native class. E.g., `Array[Array[java.lang.Object]]` should be `Array[Array[##java.lang.Object]]` (not adding regions). Maybe @mlutze has a suggestion for a fix?
> 
> I think our best option is probably to make a new formatter for generated code. We might be able to use the SimpleType interface.

Roger that. I was afraid you were going to say that hehe :D

magnus-madsen:
Before writing a lot of code, just reuse FormatType. Perfect is the enemy of good here. We don't want yet another formatter to main.

jaschdoc:
Screenshot from 2024-05-01 11-56-50
Screenshot from 2024-05-01 11-56-55
Screenshot from 2024-05-01 11-57-09
Screenshot from 2024-05-01 11-57-34

magnus-madsen (commented):
General comments:

- We should also add support to Parser2. Maybe @herluf-ba can help?
- Add some doc comments for the helper functions.
- I don't think you need System.lineSeparator. I think \n is sufficient because it is interpreted by VSCode.

jaschdoc:
@herluf-ba I'm not sure how to add the `NewObject` syntactic context to `Parser2`. Is this something you can help me with?

magnus-madsen:
I think you want to mark `IllegalQualifiedNameInInvokeConstructor` in `CompletionProvider.getSyntacticContext` as `Expr.NewObject`

jaschdoc:
Should we postpone this until the new java interop is ready?

magnus-madsen:
> Should we postpone this until the new java interop is ready?

Its low priority, but the new expression stuff will not change much.

jaschdoc:
Seems like this is quite hard with the new parser. I think I might need some help.
It tries to parse an `ArgumentList` after `new` but I can't figure out what the syntactic context is nor make it print any debug information at that point.

magnus-madsen:
> Seems like this is quite hard with the new parser. I think I might need some help. It tries to parse an `ArgumentList` after `new` but I can't figure out what the syntactic context is nor make it print any debug information at that point.

Yeah, it seems tricky. We probably have to look at it together.

jaschdoc:
- [ ] Merge
- [ ] Add syntactic context to `pick` in `Weeder2`

magnus-madsen:
We are stuck because the parser cannot give us the information we need. We will have to revisit later.

PR #7606, opened 2024-04-20
feat: rank syntactic contexts in `CompletionProvider`

magnus-madsen:
This is a minor tweak which is needed if the program contains both parse errors and other errors (which becomes possible with the new Parser).

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
@@ -285,11 +285,16 @@ object CompletionProvider {
   private def getSyntacticContext(uri: String, pos: Position, errors: List[CompilationMessage]): SyntacticContext =
     errors.filter({
       case err => pos.line <= err.loc.beginLine
-    }).collectFirst({
-      case ParseError(_, ctx, _) => ctx
-      case WeederError.MalformedIdentifier(_, _) => SyntacticContext.Import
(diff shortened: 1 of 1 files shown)


magnus-madsen:
(Alot of the completion code will have to be rewritten, but this is just a stop gap).

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -285,11 +285,16 @@ object CompletionProvider {
   private def getSyntacticContext(uri: String, pos: Position, errors: List[CompilationMessage]): SyntacticContext =
     errors.filter({
       case err => pos.line <= err.loc.beginLine
-    }).collectFirst({
-      case ParseError(_, ctx, _) => ctx
-      case WeederError.MalformedIdentifier(_, _) => SyntacticContext.Import
-      case ResolutionError.UndefinedType(_, _, _) => SyntacticContext.Type.OtherType
-      case ResolutionError.UndefinedName(_, _, _, isUse, _) => if (isUse) SyntacticContext.Use else SyntacticContext.Expr.OtherExpr
-    }).getOrElse(SyntacticContext.Unknown)
+    }).map({
```
mlutze, line 288: Can't this whole thing just be a minByOption?

mlutze (approved):

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
magnus-madsen: Assumes non-emptiness.

Issue #7433, opened 2024-03-21
Add associated types and effects to the Flix Standard Library


magnus-madsen:
Post-merge of the new typer:

Ordered from simple to more complex:

- [ ] Add an associated type to `Add` to allow `Set#{1, 2, 3} + 1`.
- [ ] Add an associated type to `Sub` to allow `Set#{1, 2, 3} - 1`.
- [ ] Add an associated type to `Coerce`.
- [ ] Add associated types and effects to `ToJava` and `FromJava`.
- [ ] Add associated types and effects to `Iterable`. Implement new instances for mutable collections.
- [ ] Add an associated effect to `ToString`. Implement new instances for mutable collections.
- [ ] Add an associated effect to `Foldable`. Implement new instances for mutable collections.
- [ ] Add an associated effect to `Reducible`. Implement new instances for mutable collections.

To be discussed:

- Add associated effects to `Add`, `Sub`, ... `Div`...?
- Add associated effects to `Functor` and friends...?
- Add associated effects to `Filterable`, `Traversable`, and friends...?

magnus-madsen:
@stephentetley Would you be interested in leading this? (There is no rush)

We could use someone with a steady hand to ensure that we keep the standard library consistent.

stephentetley:
Yes I can do, I'm a bit busy for a week or so as I've some work related (not Flix) code that I want to move forward plus finishing off the `class` to `trait` and `toVector`. But after that I'll be on it.

stephentetley:
`UnorderedFoldable` is also to do.

PR #7615, opened 2024-04-23
refactor: add assoc type and effect to Iterable (part of issue #7433)

stephentetley:
This PR adds an associated type and effect to `Iterable` and makes `Iterable`'s kind `Type` rather than `Type -> Type`.

The collections that already had `Iterable` instances have been updated along with two examples.

### DIFF
--- a/examples/larger-examples/restrictable-variants/sequences.flix
+++ b/examples/larger-examples/restrictable-variants/sequences.flix
@@ -905,6 +905,8 @@ instance Foldable[Seq.Seq[s]] {
     pub def foldRightWithCont(f: (a, Unit -> b \ ef) -> b \ ef, s: b, l: Seq.Seq[s][a]): b \ ef = Seq.foldRightWithCont(f, s, l)
 }
 
-instance Iterable[Seq.Seq[s]] {
+instance Iterable[Seq.Seq[s][a]] {
+    type Elm[Seq.Seq[s][a]] = a
+    type Aef[Seq.Seq[s][a]] = Pure
     pub def iterator(rc: Region[r], l: Seq.Seq[s][a]): Iterator[a, r, r] \ r = Seq.iterator(rc, l)
 }
--- a/examples/larger-examples/restrictable-variants/sequences.flix
+++ b/examples/larger-examples/restrictable-variants/sequences.flix
@@ -906,7 +906,6 @@ instance Foldable[Seq.Seq[s]] {
 }
 
 instance Iterable[Seq.Seq[s][a]] {
-    type Elm[Seq.Seq[s][a]] = a
-    type Aef[Seq.Seq[s][a]] = Pure
+    type Elm = a
     pub def iterator(rc: Region[r], l: Seq.Seq[s][a]): Iterator[a, r, r] \ r = Seq.iterator(rc, l)
 }
--- a/main/src/library/Chain.flix
+++ b/main/src/library/Chain.flix
@@ -116,7 +116,6 @@ instance Collectable[Chain] {
 
 instance Iterable[Chain[a]] {
     type Elm[Chain[a]] = a
-    type Aef[Chain[a]] = Pure
     pub def iterator(rc: Region[r], c: Chain[a]): Iterator[a, r, r] \ r = Chain.iterator(rc, c)
 }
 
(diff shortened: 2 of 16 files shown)


main/src/library/MutDisjointSets.flix
```diff
@@ -61,7 +61,9 @@ mod MutDisjointSets {
     ///
     /// Updates `s` with the elements of the iterable `m`.
     ///
-    pub def makeSets(m: m[t], s: MutDisjointSets[t, r]): Unit \ r with Iterable[m], Order[t] = region rc {
+    /// TODO: with Order[Iterable.Elm[m]] missing
```
magnus-madsen, line 64: I think we can remove the TODO, keep the ticket, and then if we make a fix we can search for `Elm[m] ~ elt`.

main/src/library/Iterable.flix
```diff
@@ -17,17 +17,26 @@
 ///
 /// A trait for immutable data structures that can be iterated.
 ///
-pub trait Iterable[t: Type -> Type] {
+pub trait Iterable[t] {
+    ///
+    /// The element type of the Iterable.
+    ///
+    type Elm[t]: Type
```
magnus-madsen, line 24: I think its good to put kind annotations on both Elm and Aef 👍

main/src/library/Iterable.flix
```diff
@@ -17,17 +17,26 @@
...
+
+    ///
+    /// The associated effect of the Iterable.
+    ///
+    type Aef[t]: Eff
```
magnus-madsen, line 29: We have the ability to specify a default effect:

```
    type Aef[t]: Eff = Pure
```

then instances, if they are pure, can simply omit a definition of `Aef`.

main/src/library/MutDisjointSets.flix
stephentetley: Fixed - the comment was a stray from before Matthew supplied his fix that I forgot to delete.

main/src/library/Iterable.flix
stephentetley: Thanks

main/src/library/Iterable.flix
stephentetley: I've added the default and removed `Aef` definitions when it was Pure.

examples/larger-examples/restrictable-variants/sequences.flix
```diff
@@ -905,6 +905,8 @@ instance Foldable[Seq.Seq[s]] {
     pub def foldRightWithCont(f: (a, Unit -> b \ ef) -> b \ ef, s: b, l: Seq.Seq[s][a]): b \ ef = Seq.foldRightWithCont(f, s, l)
 }
 
-instance Iterable[Seq.Seq[s]] {
+instance Iterable[Seq.Seq[s][a]] {
+    type Elm[Seq.Seq[s][a]] = a
+    type Aef[Seq.Seq[s][a]] = Pure
```
magnus-madsen, line 910: Here we can also drop = Pure

main/src/library/Chain.flix
```diff
@@ -114,7 +114,8 @@ instance Collectable[Chain] {
     pub def collect(iter: Iterator[a, ef, r]): Chain[a] \ { ef, r } with Order[a] = Iterator.toChain(iter)
 }
 
-instance Iterable[Chain] {
+instance Iterable[Chain[a]] {
+    type Elm[Chain[a]] = a
```
magnus-madsen, line 118: I think instead of writing out the function, we should just write:

```
type Elm = a
```

Does that make sense/work?

magnus-madsen:
Thanks-- I added two more comments for things I just spotted.

examples/larger-examples/restrictable-variants/sequences.flix
stephentetley: Fixed

main/src/library/Chain.flix
stephentetley: Done all.

magnus-madsen (approved):

stephentetley:
Thanks Magnus. Yes I think we should change Map and RedBlackTree so Iterable iterates on the key-value pairs. This seems like the behaviour a user would most often want.

PR #7619, opened 2024-04-25
fix: Parse error when last token of module is a DocComment

### LINKED ISSUE #7604: Parser error on modules
Related to https://github.com/flix/flix/pull/7569

I can't  narrow it down anymore than this but
```scala
mod A {

    eff _B {}

    ///

}

```

produces
```
❌ -- Parse Error -------------------------------------------------- file:///.../TestEffMultiple.flix

>> Parse Error: Expected declaration but found CurlyR

7 | }
     ^
     Here
Syntactic Context: Unknown.

❌ -- Parse Error -------------------------------------------------- file:///.../TestEffMultiple.flix

>> Parse Error: Expected CurlyR, but found Eof

7 | }
      ^
      Here
Syntactic Context: Unknown.

```

herluf-ba:
Closes #7604

This issue was this program:
```scala
mod A {

    eff _B {}

    ///

}
```
Which gave a parser error. The reason was that the doc-comment indicates that a declaration is about to follow, but it does not. The fix is to treat the doc-comment as a regular comment, which will behave nicely in an IDE setting, and be harmless otherwise.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -653,8 +653,8 @@ object Parser2 {
     def declaration()(implicit s: State): Mark.Closed = {
       val mark = open(consumeDocComments = false)
       docComment()
-      // Handle case where the last thing in a file is a doc-comment
-      if (eof()) {
+      // Handle case where the last thing in a file or module is a doc-comment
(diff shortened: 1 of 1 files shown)


magnus-madsen:
Uh oh

herluf-ba:
@magnus-madsen Should be fixed now. We are really getting some value out of the test suites 👍

PR #7622, opened 2024-04-25
feat: add jvm-expressions to `Verifier`

felix-berg:


main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -558,6 +580,27 @@ object Verifier {
     case _ => failMismatchedShape(rec, "Record", loc)
   }
 
+  /**
+    * Convert from the java class `klazz` to the corresponding `MonoType`
+    */
+  private def javaClassToMonoType(klazz: Class[_]): MonoType = klazz.getName match {
```
JonathanStarup, line 586: I'm not fully sure it is a good solution, but could you try going the other way? We know the whole set of expected Monotypes, so match the expected type, and then assert what the class should be (by name, or something else)

JonathanStarup: maybe call it `checkJavaEq`

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -384,7 +384,29 @@ object Verifier {
           check(expected = MonoType.Region)(actual = t2, loc)
           check(expected = MonoType.Unit)(actual = tpe, loc)
 
-        case _ => tpe // TODO: VERIFIER: Add rest
+        case AtomicOp.GetField(field) =>
+          checkEq(tpe, javaClassToMonoType(field.getType), loc)
```
JonathanStarup, line 388: I think, now that you've gone into checking the java types, that you should try verifying the other types too, like checking that `List(t) = ts; checkJavaEq(field.getDeclaringClass, t, ...)`

felix-berg: But in this case, `tpe` may be a subclass of the declaring class, and only the field itself is given.
Should I walk the subtree of `getDeclaringClass` looking for a match or just ignore it? Or maybe I missed something.
Specifically fails here: main/test/flix/Test.Exp.Jvm.PutField.flix:87:9

felix-berg: new commit addresses the problem, but not super elegantly

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -384,7 +384,44 @@ object Verifier {
           check(expected = MonoType.Region)(actual = t2, loc)
           check(expected = MonoType.Unit)(actual = tpe, loc)
 
-        case _ => tpe // TODO: VERIFIER: Add rest
+        case AtomicOp.GetField(field) =>
```
JonathanStarup, line 387: these cases are very pretty

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -528,6 +565,78 @@ object Verifier {
...
+
+      case _ => failMismatchedTypes(tpe, klazz, loc)
+    }
+  }
+
+  /**
+    * Returns `true` if and only if `c1` is a subclass of `c2`
+    */
+  @tailrec
+  private def isSubclass(c1: Class[_], c2: Class[_]): Boolean =
```
JonathanStarup, line 637: I think Class supports this inherently with `isAssignableFrom` or something like that, that also allows `int[] <: Object` and other weird cases

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -528,6 +565,78 @@ object Verifier {
...
+    )
+
+    tpe match {
+      case MonoType.Array(elmt) if klazz.isArray =>
+        checkJavaSubtype(elmt, klazz.getComponentType, loc)
+
+      case MonoType.Native(k) if isSubclass(k, klazz) =>
+        tpe
+
+      case _ if monoToPrimitiveName.contains(tpe) && klazz.getName == monoToPrimitiveName(tpe) =>
```
JonathanStarup, line 623: I appreciate the short cases here, but I think it might be simpler to read if these 7 cases are inlined.

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -528,6 +565,78 @@ object Verifier {
...
+        tpe
+
+      case _ if monoToClassName.contains(tpe) && isSubclass(Class.forName(monoToClassName(tpe)), klazz) =>
```
JonathanStarup, line 626: same with these

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -528,6 +565,78 @@ object Verifier {
...
+      MonoType.Int64 -> "long",
+      MonoType.Float32 -> "float",
+      MonoType.Float64 -> "double",
+      MonoType.Bool -> "boolean",
+      MonoType.Char -> "char",
+      MonoType.Unit -> "void"
+    )
+
+    val monoToClassName = Map[MonoType, String](
+      MonoType.String -> "java.lang.String",
```
JonathanStarup, line 595: I stedet for Class.forName(string) kan du så ikke bruge `classOf[String]` eller noget i den stil? undgå de her typo-sensitive strenge

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -384,7 +384,44 @@ object Verifier {
...
+          check(expected = MonoType.Unit)(actual = tpe, loc)
+
+        case AtomicOp.PutStaticField(field) =>
+          val List(t) = ts
+          checkJavaSubtype(t, field.getType, loc)
+          check(expected = MonoType.Unit)(actual = tpe, loc)
+
+        case AtomicOp.InstanceOf(_) =>
+          val List(t) = ts
+          checkJavaSubtype(t, new Object().getClass, loc) // must not be primitive type
```
JonathanStarup, line 408: Is this a requirement in typechecking? if so just close this comment, just wanted to make sure you've checked.

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -384,7 +384,44 @@ object Verifier {
...
+          check(expected = MonoType.Bool)(actual = tpe, loc)
+
+        case AtomicOp.InvokeConstructor(constructor) =>
+          checkJavaParameters(ts, constructor.getParameterTypes.toList, loc)
+          checkJavaSubtype(tpe, constructor.getDeclaringClass, loc)
+
+        case AtomicOp.InvokeMethod(method) =>
+          val t = ts.head
```
JonathanStarup, line 416: I think `val t :: pts = ts` would be more in line with the other cases

JonathanStarup (approved):

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
felix-berg: Yea, in Java, `e instanceof ...` requires that `e` is a reference (from my testing...)

PR #7627, opened 2024-04-26
feat: resilience in `PredDeps`

magnus-madsen:


main/src/ca/uwaterloo/flix/language/phase/PredDeps.scala
```diff
@@ -62,10 +62,10 @@ object PredDeps {
         case Type.Cst(TypeConstructor.Unit, _) => (Nil, den)
         case _ => (List(t), den) // Unary
       }
-    case _: Type.Var =>
-      // This could occur when querying or projecting a non-existent predicate
+    case _ =>
```
mlutze, line 65: This could be a bad type, but not Type.Error?

Should we instead try to ensure that bad types are always Type.Error?

magnus-madsen: What do you mean? I don't understand.

mlutze: This default case covers the event where type inference failed. When that happens, I would hope that the type is `Type.Error` rather than some random type.

magnus-madsen: How are you going to ensure that? It will be difficult because no only do we not know where to replace a type there may be multiple solutions.

magnus-madsen: In any case thats beyond the scope of this PR

mlutze (approved):

PR #7632, opened 2024-04-26
feat: Implement `SyntaxTree.Child` directly in `Token` and `Tree`

### LINKED ISSUE #7590: Parser: Possibly replace `trait Child` with `trait TreeOrToken` to reduce memory usage
i.e. avoid the wrappers `TokenChild` and `TreeChild` by implementation of the trait directly.

herluf-ba:
Closes #7590


main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -3412,8 +3412,8 @@ object Parser2 {
   def syntaxTreeToDebugString(tree: SyntaxTree.Tree, nesting: Int = 1): String = {
     s"${tree.kind}${
       tree.children.map {
-        case SyntaxTree.Child.TokenChild(token) => s"\n${"  " * nesting}'${token.text}'"
-        case SyntaxTree.Child.TreeChild(tree) => s"\n${"  " * nesting}${syntaxTreeToDebugString(tree, nesting + 1)}"
+        case token: Token => s"\n${"  " * nesting}'${token.text}'"
```
magnus-madsen, line 3415: We try to avoid type based matches. You can still just match on the constructor, right?

magnus-madsen:
You've refactor, but also replaced all matches with type-based matches. Thats not great because exhaustive checking can be more brittle. We can still just match on the constructors right?

herluf-ba:
> You've refactor, but also replaced all matches with type-based matches. Thats not great because exhaustive checking can be more brittle. We can still just match on the constructors right?

I will replace with constructor matches. We can't have exhaustiveness checks anymore because `Child` is no longer `sealed`. https://docs.scala-lang.org/tour/pattern-matching.html#sealed-types

magnus-madsen:
> > You've refactor, but also replaced all matches with type-based matches. Thats not great because exhaustive checking can be more brittle. We can still just match on the constructors right?
> 
> I will replace with constructor matches. We can't have exhaustiveness checks anymore because `Child` is no longer `sealed`. https://docs.scala-lang.org/tour/pattern-matching.html#sealed-types
> 
> `sealed` requires all derivers to be in the same file. Meaning that `Token` and `SyntaxTree.Tree` should share a file, which wouldn't be ideal.

Hmm. I see.

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -925,7 +925,7 @@ object Weeder2 {
       // Note: This visitor is used by both expression literals and pattern literals.
       expectAny(tree, List(TreeKind.Expr.Literal, TreeKind.Pattern.Literal))
       tree.children(0) match {
-        case Child.TokenChild(token) => token.kind match {
+        case token@Token(_, _, _, _, _, _ ,_ ,_) => token.kind match {
```
magnus-madsen, line 928: But why are you binding it? Why not take out the kind directly as a variable and match on it :)

herluf-ba:
@magnus-madsen I have removed as many `@` style matches as possible. Some cases where it isn't possible:
- The item is being passed into a function call  `case token@Token(...) => ... func(token)`
- A def is called on the item `case token@Token(...) => token.mkSourcePosition(...)`

magnus-madsen (approved):

magnus-madsen:
Merged. Thanks!

PR #7634, opened 2024-04-26
feat: add `TokenKind.display` for better errors

herluf-ba:
This PR add a way to get a display string for a `TokenKind`.

This allows `Parser2` to emit more easily read errors like these:
```scala
eff
def main(): Int32 = 123
```
```text
>> Parse Error: Expected <Name> before 'def'

1 | eff
    ^^^
    Here
Syntactic Context: Unknown.
```
or here:
```scala
def foo(): Int32 = 123;
enum Legumes { case BlackBeans }
```
```text
>> Parse Error: Expected <expression> before 'enum'

1 | def foo(): Int32 = 123;
                          ^
                          Here
Syntactic Context: OtherExpr.
```

This PR also aligns all errors that `Parser2` produces to these principles:
- A concrete TokenKind is displayed with ticks, ie. `'and'`
- A runtime specific TokenKind is displayed with angles, ie. `<name>`
- A large group of token kinds with a sensible name is also displayed with angles, ie. `<expression>` or `<pattern>`
- Use phrasing "Expected xyz before wqk" where possible.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/Flix.scala
+++ b/main/src/ca/uwaterloo/flix/api/Flix.scala
@@ -546,8 +546,8 @@ class Flix {
     val result = for {
       afterReader <- Reader.run(getInputs)
       afterLexer <- Lexer.run(afterReader, cachedLexerTokens, changeSet)
-//      afterParser <- Parser.run(afterReader, entryPoint, cachedParserAst, changeSet)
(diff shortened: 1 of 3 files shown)


herluf-ba:
While going through all TokenKinds I found out that `TokenKind.KeywordImpure` was still in there even though the lexer cannot produce it anymore, so I removed it

magnus-madsen:
Merged. Thanks!

PR #7635, opened 2024-04-26
refactor: add assoc effect to ToJava (part of issue #7433)

stephentetley:
This PR adds an associated effect to `ToJava`.

Adding a `FromJava` class will follow in another PR.

### DIFF
--- a/main/src/library/Chain.flix
+++ b/main/src/library/Chain.flix
@@ -121,6 +121,7 @@ instance Iterable[Chain[a]] {
 
 instance ToJava[Chain[a]] {
     type O = ##java.util.List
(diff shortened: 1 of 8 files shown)


main/src/library/Chain.flix
```diff
@@ -121,6 +121,7 @@ instance Iterable[Chain[a]] {
 
 instance ToJava[Chain[a]] {
     type O = ##java.util.List
```
magnus-madsen, line 123: I think we should consider renaming this to `Out`-- but that can happen later.

magnus-madsen (approved):

PR #7651, opened 2024-04-29
feat: add FromJava trait (part of issue #7433)

stephentetley:
This PR adds the trait `FromJava` and instances for types that have `ToJava` except `Nec` and `Nel` as we cannot assert the input to them is non-empty.
        
The tests were derived from `TestToJava.flix` and I spotted some inconsistent names that I corrected.


magnus-madsen (approved):

magnus-madsen:
Thanks Stephen!

PR #7654, opened 2024-04-30
feat: add `SyntacticContext` to `expect`, `expectAny` and `name`

### LINKED ISSUE #7630: Ensure autocompletions work with new Parser
I tried to enable the new parser through and through. It works pretty well, but some auto-completions are broken for various reasons. 

In particular, I noted that these are broken:

- `use ...`
- `import ...`

But also many others. I think we need to do a pass over all them.

The fixes seem simple:

- If a new error has been introduced then ensure it is used correctly in `CompletionProvider.getSyntacticContext`.
- If a parse error occurs then ensure its syntactic context is set correctly.

herluf-ba:
This PR add syntactic context to all calls to expect, expectAny and name to enable IDE completions. 

Closes #7630


magnus-madsen (approved):

magnus-madsen:
Merged! Thanks-- we can tweak things in follow-up PRs.

magnus-madsen:
I saw a significant memory reduction in the profiler 👍

PR #7656, opened 2024-04-30
refactor: add associated element type and effect to Collectable (part of issue #7433)

stephentetley:
This PR changes the kind of `Collectable` to Type and adds an associated element type and effect. It also removes the `Order` constraint on `collect` that I think was unnecessary (the `Set` instance uses an `Order` constraint but it isn't needed at the trait level).

I've added a handful of instance tests.


magnus-madsen (approved):

magnus-madsen:
Thanks Stephen, very nice 👍

PR #7659, opened 2024-05-01
chore: rework Datalog debugging

magnus-madsen:
I believe (unchecked) effects casts are fundamentally broken, including `masked_cast`.

This does not solve the problem at all, but merely makes it go away.

### DIFF
--- a/main/src/library/Fixpoint/Debugging.flix
+++ b/main/src/library/Fixpoint/Debugging.flix
@@ -23,46 +23,61 @@ mod Fixpoint.Debugging {
     ///
     /// Invoked before the fixpoint computation.
     ///
+    /// Note: Returns `d` to ensure it is not erased.
+    ///
     @Internal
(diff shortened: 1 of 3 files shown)


mlutze (approved):

JonathanStarup:
In my perspective `masked_cast` isn't broken, its meaning and use is just a bit odd.

JonathanStarup:
did you figure out how specifically? it still seems weird to me

magnus-madsen:
> did you figure out how specifically? it still seems weird to me

No exactly. I suspect that inlining leads to effects that are strange which causes miscompilation.

E.g. things like a pure expression having subexpressions with control-effects!

JonathanStarup:
> pure expression having subexpressions with control-effects

Yeah, but as long as it isn't control effects then it should be okay. Here its just impure not control impure. but yeah i do think effects might not all be updated in the backend transformations. some of the effect recomputation seems a little dubious as far as i remember

magnus-madsen:
It was typed as control impure.

PR #7660, opened 2024-05-01
refactor: add associated element type and effect to Foldable (part of issue #7433)

stephentetley:
Adding associated element and effect to `Foldable`.

Currently has parser failure - Parser: is stuck #7646

### DIFF
--- a/examples/larger-examples/Reachable.flix
+++ b/examples/larger-examples/Reachable.flix
@@ -52,7 +52,7 @@ mod Reachable.Imperative {
 }
 
 mod Reachable.Functional {
-    pub def reachable(origin: n, edges: f[(n, n)]): Set[n] with Foldable[f], Order[n] =
+    pub def reachable(origin: n, edges: f[(n, n)]): Set[n] \ Foldable.Aef[f] with Foldable[f], Order[n] =
         // collect edge lists
         let edgeMap = (Map.empty(), edges) ||> Foldable.foldLeft(map -> match (start, end) -> {
             let startEdges = Map.getWithDefault(start, Nil, map);
@@ -76,7 +76,7 @@ mod Reachable.Functional {
 }
 
 mod Reachable.Logic {
-    pub def reachable(origin: n, edges: f[(n, n)]): Set[n] with Foldable[f], Order[n] =
+    pub def reachable(origin: n, edges: f[(n, n)]): Set[n] \ Foldable.Aef[f] with Foldable[f], Order[n] =
         let edgeFacts = inject edges into Edge;
         // origin is reachable and this set is expanded by the edge facts
         let reachable = #{
--- a/examples/larger-examples/datalog/ford-fulkerson.flix
+++ b/examples/larger-examples/datalog/ford-fulkerson.flix
@@ -29,7 +29,7 @@ mod FordFulkerson {
     /// - The label of `src` has the lowest value in the graph
     /// - The label of `dst` has the highest value in the graph
     ///
-    pub def maxFlow(src: t, dst: t, g: m[(t, Int32, t)]): Int32 with Foldable[m], Order[t] =
+    pub def maxFlow(src: t, dst: t, g: m[(t, Int32, t)]): Int32 \ Foldable.Aef[m] with Foldable[m], Order[t] =
         def fordFulkerson(flowNetwork) = match augmentingPath(src, dst, flowNetwork) {
             case None       => getMaxFlow(dst, flowNetwork)
             case Some(path) =>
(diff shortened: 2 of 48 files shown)


magnus-madsen:
@stephentetley Do you know which file is stuck? 

Because in the linked issue there is actually a parser error (an errnernous _ aftera case).

stephentetley:
`Foldable.flix` according to the stack trace.

If I take it out of the source tree and make it an independent project seems to be parseable by Flix 0.46 but then it has a free variable in type signature error (I have to rename the trait to get this so it doesn't have a name clash with the existing `Foldable`).

magnus-madsen:
@stephentetley I think the parser error has been fixed now.

stephentetley:
Thanks Magnus, I'm now getting two types of error.

A redundancy error if a type parameter is only used in trait constraints and not in the function signature per se:

~~~

-- Redundancy Error -------------------------------------------------- Foldable.flix

>> Unused type parameter 'a'. The parameter is not referenced anywhere.

369 |     pub def join(sep: String, t: t): String \ Foldable.Aef[t] with ToString[a] where Foldable.Elm[t] ~ a =
                                                                                                             ^
                                                                                                             unused type parameter.
                                                                                                             
~~~                                                                                                             

A type error trying to inject foldables into Datalog. Does `inject ... into ...` have a builtin type of kind `Type -> Type` that needs changing?

~~~

-- Type Error -------------------------------------------------- Graph.flix

>> Unable to unify the types: 'm' and ''0[t1]'.

30 |         let edges = inject g into Edge;
                         ^^^^^^^^^^^^^^^^^^
                         mismatched types.

Type One: m
Type Two: '0[t1]

~~~


> pub def toVector(iter: Iterator[a, ef, r]): Vector[a] \ { ef, r }
        
~~~

magnus-madsen:
Thanks Stephen,

@mlutze Can you take a look at the redundancy error? I think here: https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/Redundancy.scala#L241 we should simply also look at constraints.

As for `inject`, we need to make a change here: https://github.com/flix/flix/blob/2818d8f74dc2812cbefd9fe388202df7ed797296/main/src/ca/uwaterloo/flix/language/phase/typer/SchemaConstraintGen.scala#L132 @mlutze Can you help sketch the change?

mlutze:
> As for `inject`, we need to make a change here:
> 
> https://github.com/flix/flix/blob/2818d8f74dc2812cbefd9fe388202df7ed797296/main/src/ca/uwaterloo/flix/language/phase/typer/SchemaConstraintGen.scala#L132
> 
> @mlutze Can you help sketch the change?

We need to change a few things. First, we must change this line so the kind of the foldable should just be `Kind.Star`
```
        val freshTypeConstructorVar = Type.freshVar(Kind.Star ->: Kind.Star, loc)
```

We can get rid of the element type variable:
```diff
-           val freshElmTypeVar = Type.freshVar(Kind.Star, loc)
```

Then the result type is just the foldable type, not applied to anything.
```
        c.unifyType(tpe, Type.mkApply(freshTypeConstructorVar, List(freshElmTypeVar), loc), loc)
```
(replace `Type.mkApply(...)` with `freshTypeConstructorVar`)

And finally we need to associate the Foldable's associated type with the relation.
We need to construct the associated type:
```diff
+         val elmSym = new Symbol.AssocTypeSym(foldableSym, "Elm", loc)

+         val elmTpe = Type.AssocType(Ast.AssocTypeConstructor(elmSym, loc), freshTypeConstructorVar, Kind.Star, loc)
```

and to use that in the type constraint:
```
        c.unifyType(tvar, Type.mkSchema(Type.mkSchemaRowExtend(pred, Type.mkRelation(List(freshElmTypeVar), loc), freshRestSchemaTypeVar, loc), loc), loc)
```
(replace `freshElmTypeVar` with with `elmTpe`)

Feel free to rename/reorganize as you please.

mlutze:
What is the plan for Traversable? I see that for now we have removed the Foldable trait from it.

stephentetley:
Thanks Matthew, I'm not sure what will happen to `Traversable`'s `Foldable` constraint now that Traversable and Foldable's kinds are different.

I was planning to look at `Traversable` when the `Foldable` changes compile. It might be that `Foldable` should return to kind `Type -> Type` and just get an associated effect (which would allow it to be used on mutable collections).

magnus-madsen:
We also discussed that e.g. `Functor` should most likely stay higher-kinded and only have an associated effect (down the road). Because part of being a functor means to be higher-kinded.

stephentetley:
I've decided to revert `Foldable` back to kind `Type -> Type` at least for the time being. 

The upside is it works with `Traversable` which doesn't look like a good candidate to be made kind `Type`.

The downside is we lose some possible instances, however I can only think of `String` and treating `String` as a container of `Chars` is often seen as an antipattern for Unicode handling.

stephentetley:
I'm stuck trying to change `SchemaConstraintGen.scala` so that `visitFixpointInject` accounts for `Aef[m]` rather than needs to be `Pure`.

@mlutze - would you be able to take another look. Thank you.

mlutze:
> I'm stuck trying to change `SchemaConstraintGen.scala` so that `visitFixpointInject` accounts for `Aef[m]` rather than needs to be `Pure`.
> 
> @mlutze - would you be able to take another look. Thank you.

Right, I forgot about the effect!

Similar to before, we need to construct the effect:
```diff
+         val aefSym = new Symbol.AssocTypeSym(foldableSym, "Aef", loc)

+         val aefTpe = Type.AssocType(Ast.AssocTypeConstructor(aefSym, loc), freshTypeConstructorVar, Kind.Eff, loc)
```

And then the resulting effect should be the union of the effect of the subexpression and the associated effect.

```diff
-        val resEff = eff
+        val resEff = Type.mkUnion(eff, aefTpe, loc)
```

I think that's all that needs to be done here.

stephentetley:
Thanks Matthew! 

Nearly there...

stephentetley:
Tests are passing so it should now be ready for review.

magnus-madsen (approved):

magnus-madsen:
@jaschdoc You need to add an associated effect somewhere in your code. Thanks :)

magnus-madsen:
Excellent! Thanks Stephen.

PR #7661, opened 2024-05-02
feat: Fix parser crash on bad match case

### LINKED ISSUE #7646: Parser: is stuck
```
instance ToString[Card] {
    pub def toString(x: Card): String = match x {
        case Card.Card(r, s) _ => "${r} of ${s}"
    }
}
```

herluf-ba:
Closes #7646 

The example now gives 3 parser errors:
```
-- Parse Error -------------------------------------------------- main/foo.flix

>> Parse Error: Expected '=>' before '_'

3 |         case Card.Card(r, s) _ => "${r} of ${s}"
                               ^
                               Here
Syntactic Context: OtherExpr.

-- Parse Error -------------------------------------------------- main/foo.flix

>> Parse Error: Expected '}' before '=>'

3 |         case Card.Card(r, s) _ => "${r} of ${s}"
                                 ^
                                 Here
Syntactic Context: OtherExpr.

-- Parse Error -------------------------------------------------- main/foo.flix

>> Parse Error: Expected 'type' or 'def' before '=>'

3 |         case Card.Card(r, s) _ => "${r} of ${s}"
                                   ^^
                                   Here
Syntactic Context: Instance.
```
**Reasoning**: 

First error is because parser expects '=>' but finds '_'. This stops parsing of a match-case and returns into the match-expression rule. 

Next error is because the match-expression rule, seeing that it is not at the start of a new match-case, expects a closing '}'. Then control is returned to the instance-body loop. 

The final error is because the instance-body sees something that does not look like a type def or a new definition, namely "=>".  It then skips ahead until a stable spot is found, which ends up being EOF.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/Flix.scala
+++ b/main/src/ca/uwaterloo/flix/api/Flix.scala
@@ -547,8 +547,8 @@ class Flix {
     val result = for {
       afterReader <- Reader.run(getInputs)
       afterLexer <- Lexer.run(afterReader, cachedLexerTokens, changeSet)
-      afterParser <- Parser.run(afterReader, entryPoint, cachedParserAst, changeSet)
-      afterWeeder <- Weeder.run(afterParser, cachedWeederAst, changeSet)
+//      afterParser <- Parser.run(afterReader, entryPoint, cachedParserAst, changeSet)
+//      afterWeeder <- Weeder.run(afterParser, cachedWeederAst, changeSet)
 
       // Plan for migrating to new parser + weeder:
       // Stage 1 [ACTIVE]
@@ -582,9 +582,9 @@ class Flix {
       // Update caches for incremental compilation.
       if (options.incremental) {
         this.cachedLexerTokens = afterLexer
-        this.cachedParserAst = afterParser
+//        this.cachedParserAst = afterParser
         this.cachedParserCst = afterParser2
-        this.cachedWeederAst = afterWeeder
+//        this.cachedWeederAst = afterWeeder
         this.cachedWeederAst2 = afterWeeder2
         this.cachedDesugarAst = afterDesugar
         this.cachedKinderAst = afterKinder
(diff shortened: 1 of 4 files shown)


herluf-ba:
I think I need to reconsider the `previousSourcePosition`. It's not consistent in this example. I feel like the old way, which is used on the last of the 3 errors, is more intuitive.

magnus-madsen:
Lets add test cases: Just grab the ticket id, copy the source program, and slam that bad boy into TestParser:

```
test("Regression.TICKETID") {
...
}
```

magnus-madsen:
Indeed the program is very broken, so it is sensible that it would trigger multiple errors before recovery hits.

main/src/ca/uwaterloo/flix/api/Flix.scala
```diff
@@ -547,8 +547,8 @@ class Flix {
     val result = for {
       afterReader <- Reader.run(getInputs)
       afterLexer <- Lexer.run(afterReader, cachedLexerTokens, changeSet)
-      afterParser <- Parser.run(afterReader, entryPoint, cachedParserAst, changeSet)
-      afterWeeder <- Weeder.run(afterParser, cachedWeederAst, changeSet)
+//      afterParser <- Parser.run(afterReader, entryPoint, cachedParserAst, changeSet)
```
magnus-madsen, line 550: This should be undone. We are not making these changes just yet :)

magnus-madsen:
Note that the bug in https://github.com/flix/flix/pull/7660 is separate from this.

Note also that some test cases fail :)

main/src/ca/uwaterloo/flix/api/Flix.scala
herluf-ba: Yea haha, Will do!

magnus-madsen (approved):

PR #7665, opened 2024-05-02
fix: Handle bad enum case crash.

### LINKED ISSUE #7645: Parser: crash on mangled enum
```
// A Suit type deriving an Eq and ToString instance
enum Suit with Eq, ToString {
    case 
    //case Clubs
    //case Hearts
    //case Spades
    //case Diamonds
}
```

herluf-ba:
Closes #7645 

Added a test case with the example. Two error messages are produced instead of a crash now: 
```
-- Parse Error -------------------------------------------------- main/foo.flix

>> Parse Error: Misplaced comments

4 |>     //case Clubs
5 |>     //case Hearts
6 |>     //case Spades
7 |>     //case Diamonds

Here
Syntactic Context: Unknown.

-- Parse Error -------------------------------------------------- main/foo.flix

>> Parse Error: Expected <Name> or 'Pure' before '}'

7 |     //case Diamonds
        ^^^^^^^^^^^^^^^
        Here
Syntactic Context: Enum.
```

**Reasoning**: 

The enum case rule sees `case` and expects a name next. To the parser, comments are invisible, which is achieved by consuming them immediately when opening a new tree node. So when the parser looks for a tag-name it sees the `}` next and produces the bottom-most error. 

The first error is produced by the weeder. It looks for a tree node of kind Ident which should contain the tag-name of the enum-case. Within the ident however is only a CommentList. This represents a general case of disallowed comment placement, ala `let /* why even would we want this ?*/ x = ???` so the weeder makes an error out of it.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -430,7 +430,7 @@ object Parser2 {
       case TokenKind.CommentDoc => ParseError(s"Doc-comments can only decorate declarations.", context, previousSourceLocation())
       case at =>
         val kindsDisplayed = prettyJoin(kinds.toList.map(k => s"${k.display}"))
-        ParseError(s"Expected $kindsDisplayed before ${at.display}", SyntacticContext.Unknown, previousSourceLocation())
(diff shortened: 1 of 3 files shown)


magnus-madsen (approved):

magnus-madsen:
Thanks!

PR #7666, opened 2024-05-02
feat: open lambda effect

JonathanStarup:
Allows lambdas to have larger than necessary effects

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
@@ -157,7 +157,8 @@ object ConstraintGen {
       case Expr.Lambda(fparam, exp, loc) =>
         c.unifyType(fparam.sym.tvar, fparam.tpe, loc)
         val (tpe, eff) = visitExp(exp)
-        val resTpe = Type.mkArrowWithEffect(fparam.tpe, eff, tpe, loc)
+        val arrowEff = Type.mkUnion(eff :: Type.freshVar(Kind.Eff, loc) :: Nil, loc)
+        val resTpe = Type.mkArrowWithEffect(fparam.tpe, arrowEff, tpe, loc)
         val resEff = Type.Pure
         (resTpe, resEff)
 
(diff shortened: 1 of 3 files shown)


magnus-madsen:
We should be able to remove the checked_ecasts in the std lib now, right?

magnus-madsen:
Will you investigate the failing tests?

JonathanStarup:
> > downside is also that we taint the definition instead of checking at application so information is less precise
> 
> What do you mean?

fx that `x -> x` will have type `a -> a \ Crash` if needed, this means that we might instrument pure code with effect handler things because of the larger effect. It would be more exact to enlarge the effect use-site (a la proper subtyping) because the definition would still say pure

magnus-madsen:
> > > downside is also that we taint the definition instead of checking at application so information is less precise
> > 
> > 
> > What do you mean?
> 
> fx that `x -> x` will have type `a -> a \ Crash` if needed, this means that we might instrument pure code with effect handler things because of the larger effect. It would be more exact to enlarge the effect use-site (a la proper subtyping) because the definition would still say pure

I see. But we were struggling to make that work? What do you propose?

JonathanStarup:
It was just an observation, its nice to have this as a solution

JonathanStarup:
@magnus-madsen it solves all of our `checked_ecast` uses

main/src/library/Nel.flix
```diff
@@ -408,7 +408,7 @@ mod Nel {
             case _        => unreachable!()
         };
         let Nel(x, xs) = l;
-        loop(x :: xs, x -> x)
+        loop(x :: xs, id -> id)
```
magnus-madsen, line 411: We have an identity function :) and even an eidentity :)

magnus-madsen: Oh. You removed OK 👍 

But lets please not call an argument id. We can call it z or w.

magnus-madsen:
It seems the base system works then? (Other than a horrible slowdown)

What about integration with the rest of the compiler? Monomorphization? Purity reflection? Traits?

JonathanStarup:
New version: #7847

PR #7668, opened 2024-05-02
fix: capture doc comments again

### LINKED ISSUE #7608: Parser: Doc comments are not included
The `node.doc` attribute of all AST nodes is empty when it reaches the `HTMLDocumentor` phase.

This means that running the `:doc` command produces this
image
when it should be producing this
image

Could it be a problem with the new parser?

herluf-ba:
Closes #7608 

The issue was that `CommentList` was introduced to group sequential comments into a single syntax node, but the weeder was not updated to select into `CommentList`s and capture text. `pickDocumentation` was updated to remedy that.

Another related issue was that the very first doc comment in a file would be placed under the root node rather than the first declaration. That was also fixed.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -638,7 +638,7 @@ object Parser2 {
   /// GRAMMAR                                                                             //
   //////////////////////////////////////////////////////////////////////////////////////////
   private def root()(implicit s: State): Unit = {
-    val mark = open()
+    val mark = open(consumeDocComments = false)
     usesOrImports()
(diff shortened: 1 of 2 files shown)


magnus-madsen (approved):

magnus-madsen:
@sockmaster27 Can you confirm that everything works correct now?

sockmaster27:
@magnus-madsen 
Yes

magnus-madsen:
@sockmaster27 Are you sure? Enums seem to be missing their documentation?

magnus-madsen:
Same for traits

sockmaster27:
@magnus-madsen 
No, you're right, it's still missing for enums and traits.

PR #7674, opened 2024-05-03
fix: Capture doc-comments on top-level traits.

### LINKED ISSUE #7608: Parser: Doc comments are not included

herluf-ba:
Closes #7608 

Fixes issue where a doc-comment immediately after the license block-comment would get misplaced into the wrong CommentList.


magnus-madsen:
@herluf-ba Test failures

magnus-madsen (approved):

PR #7684, opened 2024-05-06
refactor: add associated effect to Traversable, Filterable, Witherable (part of issue #7433)

stephentetley:
I'm getting a type / kind error trying to inherit `Foldable`'s associated effect in subtraits. I seems like I can inherit an associated effect one level down (`Foldable` -> `Traversable` but not two levels: `Foldable` -> `Traversable` -> `Witherable`)

Side note - I've added `Foldable` as a supertrait to `Filterable` so it could have `Foldable`'s assoc effect and it seems like anything `Filterable` would be `Foldable`. `Witherable` has the "grandchild" problem whether or not `Filterable` has `Foldable` as a supertrait.

~~~

-- Kind Error -------------------------------------------------- Witherable.flix

>> No constraint of the 'Foldable' trait for the type 't'.

27 |     pub def wither(f: a -> m[Option[b]] \ ef, t: t[a]): m[t[b]] \ (ef + aef) with Applicative[m] where Foldable.Aef[t] ~ aef =
                                                                                                                         ^
                                                                                                                         missing constraint

-- Kind Error -------------------------------------------------- Witherable.flix

>> No constraint of the 'Foldable' trait for the type 't'.

35 |     pub def sequenceWither(t: t[m[Option[a]]]): m[t[a]] \ aef with Applicative[m] where Foldable.Aef[t] ~ aef =
                                                                                                          ^
                                                                                                          missing constraint


~~~


main/src/library/Filterable.flix
```diff
@@ -17,16 +17,17 @@
...
 /// A trait for filtering container functors.
 ///
-pub trait Filterable[m : Type -> Type] with Functor[m] {
+pub trait Filterable[m : Type -> Type] with Functor[m], Foldable[m] {
 
     ///
     /// Applies the partial function `f` to every element in `x` collecting the results.
     ///
-    pub def filterMap(f: a -> Option[b] \ ef, t: m[a]): m[b] \ ef
+    pub def filterMap(f: a -> Option[b] \ ef, t: m[a]): m[b] \ (ef + aef) where Foldable.Aef[m] ~ aef
```
mlutze, line 25: Is there a problem with `(ef + Foldable.Aef[m])`?

mlutze (approved):

main/src/library/Filterable.flix
stephentetley: Same error for both ways - I tried with `(ef + Foldable.Aef[m])` first then wondered if making a variable would alleviate it.

mlutze: I see.

This PR will hopefully fix the issue for you: https://github.com/flix/flix/pull/7686

stephentetley: Thanks Matthew

magnus-madsen: PR with fix merged.

stephentetley: Answering the top question I've reverted back to `(ef + Foldable.Aef[m])`

stephentetley:
Thanks for the PR #7686, Matthew. This fixes things in the "normal world". There 's now a failure in the "laws world":

~~~
Unable to compile. Failed with: 2 errors.

-- Type Error -------------------------------------------------- Traversable.flix

>> Unable to unify the effect formulas: 'Pure' and 'Aef[t]'.

38 |     law identity: forall(t: t[a], f: a -> m[a]) with Eq[m[t[a]]], Applicative[m] . Traversable.traverse(f, t) == Applicative.point(t)
                                                                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                                                                        mismatched effect formulas.

Type One: Pure
Type Two: Aef[t]

-- Type Error -------------------------------------------------- Traversable.flix

>> Unable to unify the effect formulas: 'Pure' and 'Aef[t]'.

43 |>     law identity: forall(t: t[a], f: a -> m[a]) with Eq[m[t[a]]], Applicative[m] . Traversable.sequence(Functor.map(f, t)) == Applicative.point(t)
44 |> 
45 |>     // Missing laws: naturality and composition.

mismatched effect formulas.

Type One: Pure
Type Two: Aef[t]

~~~

Is there syntax to add associated type / effects to laws?

magnus-madsen:
I think laws are kind of broken. They inhabit the world of "not fully implemented yet" -- which is a dangerous world to be in. 

I have been trying to eliminate features that are in that state. Perhaps we should do it with laws too. Then we can bring them back once we have something fully working.

mlutze:
> I think laws are kind of broken. They inhabit the world of "not fully implemented yet" -- which is a dangerous world to be in.
> 
> I have been trying to eliminate features that are in that state. Perhaps we should do it with laws too. Then we can bring them back once we have something fully working.

There's a question about what laws even mean in the context of effects. But I guess here we would add the constraint `Aef[m] ~ Pure` so that concern is less important.

Laws are nice in their current state in that they are clear, type-checked documentation.

It should be fairly simple to add support for a constraint here. Should I do so?

magnus-madsen:
> > I think laws are kind of broken. They inhabit the world of "not fully implemented yet" -- which is a dangerous world to be in.
> > I have been trying to eliminate features that are in that state. Perhaps we should do it with laws too. Then we can bring them back once we have something fully working.
> 
> There's a question about what laws even mean in the context of effects. But I guess here we would add the constraint `Aef[m] ~ Pure` so that concern is less important.
> 
> Laws are nice in their current state in that they are clear, type-checked documentation.
> 
> It should be fairly simple to add support for a constraint here. Should I do so?

OK

stephentetley:
Thanks to Herluf's recent parser changes, `Traversable` now compiles with the laws commented out. Is commenting out the laws acceptable for a temporary fix?

magnus-madsen:
@mlutze Did you make changes to support associated effects in laws or not yet?

mlutze:
> @mlutze Did you make changes to support associated effects in laws or not yet?

Yes: https://github.com/flix/flix/pull/7690

magnus-madsen:
> Thanks to Herluf's recent parser changes, `Traversable` now compiles with the laws commented out. Is commenting out the laws acceptable for a temporary fix?

Do you mean that something does not parse correctly?

stephentetley:
> > Thanks to Herluf's recent parser changes, `Traversable` now compiles with the laws commented out. Is commenting out the laws acceptable for a temporary fix?
> 
> Do you mean that something does not parse correctly?

The parser failed when the last lines of traits were comments not code, as per issue #7711 this got fixed in Herluf's changes.

stephentetley:
I've now re-instated the `Traversable` identity laws - they need a constraint that the Foldable associated effect of accessing elements must be pure.

mlutze (approved):

magnus-madsen (approved):

Issue #7639, opened 2024-04-27
Add `Iterable` instances for mutable data structures


magnus-madsen:
e.g. `MutSet`, `MutMap`, etc.

JonathanStarup:
Completed by #7687

PR #7687, opened 2024-05-06
feat: add Iterable instances for mutable data structures (issue #7639)

stephentetley:
This PR adds `Iterable` instances for mutable data structures and improves the documentation of `Iterable.Aef`.

See issue #7639

### DIFF
--- a/main/src/library/Array.flix
+++ b/main/src/library/Array.flix
@@ -14,6 +14,13 @@
  * limitations under the License.
  */
 
+instance Iterable[Array[a, r]] {
+    type Elm = a
+    type Aef = r
+    pub def iterator(rc: Region[r1], a: Array[a, r]): Iterator[a, r + r1, r1] \ (r + r1) =
+        checked_ecast(Array.iterator(rc, a))
+}
+
 mod Array {
 
     ///
--- a/main/src/library/Iterable.flix
+++ b/main/src/library/Iterable.flix
@@ -24,19 +24,22 @@ pub trait Iterable[t] {
     type Elm[t]: Type
 
     ///
-    /// The associated effect of the Iterable.
+    /// The associated effect of creating the Iterable and accessing the elements
+    /// in the underlying data structure or generator.
+    ///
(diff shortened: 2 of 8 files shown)


main/src/library/Array.flix
```diff
@@ -14,6 +14,13 @@
  * limitations under the License.
  */
 
+instance Iterable[Array[a, r]] {
+    type Elm = a
+    type Aef = r
+    pub def iterator(rc: Region[r1], a: Array[a, r]): Iterator[a, r + r1, r1] \ (r + r1) =
+        checked_ecast(Array.iterator(rc, a))
```
mlutze, line 21: Philosophical musing: It seems that creating an iterator over something shouldn't _actually_ touch that thing; it's only when iterating over it that we touch it. So I would expect this to have only effect `r1` and not `r`.

When I was prototyping associated effects before, that's what I did for the sample collections, though it required some tedious rewriting of the implementations since many of them use `r + r1` for the effect.

mlutze (approved):
Thanks, Stephen!

main/src/library/Array.flix
stephentetley: Ideally `iterator` would probably be "dont touch input in building the iterator" i.e. 

> pub def iterator(rc: Region[r1], a: Array[a, r]): Iterator[a, r + r1, r1] \ r1

But most iterators from mutable data structures are imperative in the "setup" phase as well as the "next/iterate" phase.

magnus-madsen: What works is done in the setup phase?

magnus-madsen: I also know we had some discussion about whether the "middle" argument of the iterator should include the region itself or not.

magnus-madsen (approved):

magnus-madsen:
Thanks! 

Imagine- now I can iterate over a collection!

main/src/library/Array.flix
stephentetley: Initializing ref cells that are updated by `next`.

PR #7689, opened 2024-05-06
feat: add ToJava and FromJava instances for builtin types String, Char and the number types

stephentetley:
This PR added `ToJava` and `FromJava` instances for the builtin types `String`, `Char` and the number types `Int32` etc.

For Flix types that are primitive types in Java, the conversion is to the respective Java object rather than the primitive type (the Flix type is already the primitive) type. ie. Flix `Char` goes to Java's `java.lang.Character`. `String`, `BigInt` and `BigDecimal` are just identity. Converting to the Java object means they can be stored in Java collections.

The PR also adds `toJavaObject(a: t): ##java.lang.Object with ToJava[t]` to convert safely to the Java representation and upcast to `java.lang.Object`.

### DIFF
--- a/main/src/library/ToJava.flix
+++ b/main/src/library/ToJava.flix
@@ -15,10 +15,71 @@
  */
 
 ///
-/// A trait for marshaling values to Java.
+/// A trait for marshaling values to Java objects.
 ///
 pub trait ToJava[t: Type] {
     type Out[t]: Type
     type Aef[t]: Eff = Pure
     pub def toJava(t: t): ToJava.Out[t] \ ToJava.Aef[t]
+}
+
+mod ToJava {
+
+    ///
+    /// Marshal a type that supports `ToJava` to its respective to `ToJava.Out` type
+    /// and upcast it to a `java.lang.Object`
+    ///
+    pub def toJavaObject(t: t): ##java.lang.Object \ ToJava.Aef[t] with ToJava[t] =
+        unchecked_cast(ToJava.toJava(t) as ##java.lang.Object)
+
+}
+
+instance ToJava[Int8] {
+    type Out = ##java.lang.Byte
+    pub def toJava(i: Int8): ##java.lang.Byte = Int8.valueOf(i)
+}
+
+instance ToJava[Int16] {
+    type Out = ##java.lang.Short
+    pub def toJava(i: Int16): ##java.lang.Short = Int16.valueOf(i)
+}
+
+instance ToJava[Int32] {
+    type Out = ##java.lang.Integer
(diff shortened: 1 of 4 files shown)


mlutze (approved):

stephentetley:
Thanks Matthew, yes the collection instances should have ToJava/FromJava constraints on the element types. We are already casting in the underlying implementation in `Adaptor` like `toArrayList` so we should use `ToJava.toJavaObject` instead. This could do with being a new issue so I don't forget.

magnus-madsen (approved):

magnus-madsen:
Just so I understand. Remind me, the "point" of  `ToJava/FromJava` is mostly for convenience? 

E.g. you have a Java value and you want a Flix value, so instead of hunting for the right function, you just call `FromJava`?

Second, what about something like `List[Integer]` can I get that as a `List[Int32]`? It seems I would currently get `List[Integer]`.

magnus-madsen:
> Second, what about something like List[Integer] can I get that as a List[Int32]? It seems I would currently get List[Integer].

I guess we first need to deal with Java generics before this question becomes relevant.

main/src/library/ToJava.flix
```diff
@@ -15,10 +15,71 @@
  */
 
 ///
-/// A trait for marshaling values to Java.
+/// A trait for marshaling values to Java objects.
 ///
 pub trait ToJava[t: Type] {
     type Out[t]: Type
```
magnus-madsen, line 21: I just noticed that this can just be Out: Type

stephentetley:
> Just so I understand. Remind me, the "point" of ToJava/FromJava is mostly for convenience?

Yes, mostly for convenience and some safety.

The current `toArrayList` we have isn't very safe we will have a cast failure if we supply it with `List[Int32]` as `Int32` needs boxing as `java.lang.Integer`: 

`pub def toArrayList(ma: m[a]): ##java.util.ArrayList \ (IO + Foldable.Aef[m]) with Foldable[m]`

This would be better, provide this PR is accepted as `Int32`'s `ToJava` instance takes case of boxing so a `List[Int32]` would be valid:

`pub def toArrayList(ma: m[a]): ##java.util.ArrayList \ (IO + Foldable.Aef[m]) with Foldable[m], ToJava[a]`

I think we are going to always need versions where pass in the conversion function rather than use the trait as we won't always want to use the default conversion:

`pub def toArrayListWith(ma: m[a], f: a -> ##java.lang.Object \ ef): ##java.util.ArrayList \ (IO + ef + Foldable.Aef[m]) with Foldable[m]`

PR #7693, opened 2024-05-07
fix: fix datalog crash

### LINKED ISSUE #7640: Parser: crash with Datalog stuff
```
def baz(): #{ Path(Int32, Int32) } = #{
    Edge(1, 2).
    Path(x, y) :- 
}
```

I think this one can actually be found by fuzzing one of the Datalog examples. I propose that the fix also adds such a test. 

See examples/larger-examples/datalog

herluf-ba:
Closes #7640 

Fixes a crash on malformed datalog program and adds the program to parser test suite. 
Also adds a datalog example (`ford-fulkerson.flix`) to the fuzzer suite and fixes 3 more crashes found by doing that.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/TokenKind.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/TokenKind.scala
@@ -240,6 +240,31 @@ sealed trait TokenKind {
   })
 
   /**
+    * Checks if this token is one of the [[TokenKind]]s that can validly appear as the first token in a declaration within an instance.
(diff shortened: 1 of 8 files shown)


magnus-madsen:
> Closes #7640
> 
> Fixes a crash on malformed datalog program and adds the program to parser test suite. Also adds a datalog example (`ford-fulkerson.flix`) to the fuzzer suite and fixes 3 more crashes found by doing that.

If this continues, soon we will run out of crashes!

magnus-madsen (approved):

PR #7697, opened 2024-05-07
feat: demote `Pure` from being a keyword

### LINKED ISSUE #7638: Parser: strange error message
```
enum Color {
    case Red, 
}
```

```
>> Parse Error: Expected <Name> or 'Pure' before '}'

12 |     case Red, 
                 ^
                 Here
```

How can Pure be expected here?

herluf-ba:
Closes #7638

This PR demotes `Pure` from being a keyword to just a regular `UppercaseName`.
Instead `Pure` is now declared as a type-alias in `Prelude.flix` like so:
```scala
///
/// Pure denotes the empty effect set.
///
pub type alias Pure = \ { }
```
Notice we are using `\ { }` rather than just `{}`. This notion of an _empty effect set literal_ is also implemented in this PR. It is needed to disambiguate between the empty record type `{}` and the empty effect set `{}`. Otherwise the type-alias would be read as an empty record type.

## Considerations of the empty effect set literal
- `\ { }` effectively replaces `Pure`. This means that we don't need special handling when `Pure` us mentioned in a type, simplifying error messages and the parser/weeder implementation.
-`\ { }` leads with a `\` which is syntactically consistent with other uses of effects like `\ IO`.
- Refactoring existing code that uses `Pure` is *rarely* needed because of the type-alias in `Prelude.flix`. The exception is instances, because they may not refer to type-aliases. In these cases `Pure` can be replaced with `\ { }` directly. 
- `\ { }` disambiguates effect sets from record types completely. The empty record type is now `{}` free of context.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/Flix.scala
+++ b/main/src/ca/uwaterloo/flix/api/Flix.scala
@@ -547,8 +547,8 @@ class Flix {
     val result = for {
       afterReader <- Reader.run(getInputs)
       afterLexer <- Lexer.run(afterReader, cachedLexerTokens, changeSet)
-      afterParser <- Parser.run(afterReader, entryPoint, cachedParserAst, changeSet)
-      afterWeeder <- Weeder.run(afterParser, cachedWeederAst, changeSet)
+//      afterParser <- Parser.run(afterReader, entryPoint, cachedParserAst, changeSet)
+//      afterWeeder <- Weeder.run(afterParser, cachedWeederAst, changeSet)
 
       // Plan for migrating to new parser + weeder:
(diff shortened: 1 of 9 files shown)


herluf-ba:
@magnus-madsen This is a draft, while I implement `/ { }` in the old parser too!

herluf-ba:
Oh, turns out that wasn't too bad 👍  Let me know what you think!

magnus-madsen:
But we use backslashes for effects, right? :)

herluf-ba:
We are doing #7701 instead. That's should be another PR so closing this one.

PR #7703, opened 2024-05-07
refactor: cleanup `Flix.phase`

magnus-madsen:


magnus-madsen:
I am not super happy with this. But I think it is one step in the right direction. At least it makes `phase` sane again.

magnus-madsen:
I plan to do some kind of follow up.

JonathanStarup:
Oh right, I didn't think of typeclasses in scala

JonathanStarup (approved):

PR #7706, opened 2024-05-08
fix: Emit error on faulty program

### LINKED ISSUE #7698: Parser: incorrect program with brace parses without error
```
} def f(): Unit = ()
```

herluf-ba:
Closes #7698 

Adds the example as a test-case. 

The fix was to delete some code that handled the case of a trailing doc-comment
```scala
mod Foo {
  def bar(): Int32 = 123
  /// Like this one
}
def main(): Unit = ()
/// Or this one
```

Turns out, that special handling was no longer needed due to other updates!

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -621,6 +621,7 @@ object Parser2 {
     // Note: In case of a misplaced CommentDoc, we would just like to consume it into the comment list.
     // This is forgiving in the common case of accidentally inserting an extra '/'.
     def atComment() = if (consumeDocComments) nth(0).isComment else nth(0).isCommentNonDoc
+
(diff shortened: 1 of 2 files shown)


magnus-madsen:
I feel like there was some issues with bugs/crashes when encountering eof too early. Are we sure that is not the case anymore? :)

herluf-ba:
I have been adding these cases as test cases so I'm fairly certain this is not an issue. 

The only thing I am aware of right now is ungraceful behavior when the only thing in a module is a doc-comment.

```scala
mod Bar {
   /// Like so
}
def main(): Unit = ()
```

`main` is missed here. But it was so before also I believe, and I think that's a different PR.

herluf-ba:
@magnus-madsen I made this issue https://github.com/flix/flix/issues/7708

Could you assign me and add label:bug?

magnus-madsen (approved):

PR #7707, opened 2024-05-08
feat: add `Separation` enum for better error-handling

herluf-ba:
In recent issues regarding parser crashes the fix has been to use `zeroOrMore`/`oneOrMore` on multiple occasions. 
This prompted me to run through the parser to see if there where more `while`-loops that were replaceable. This is desirable for a couple of reasons:
1. It centralizes a spot for debugging/optimization: `zeroOrMore`.
2. It makes sure looping is done in a resilient manner. 
3. It improves readability at the call-site.

I saw that if we abstract separation just a little, allowing `Required`, `Optional` and `None` we could capture a lot of use-cases.

## Removing unnecessary `if`s
Going through the parser I also found a number of unnecessary `if`s. This is a remnant from before recover-sets, and is no longer needed!

This should admittedly have been a separate PR.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -455,6 +455,34 @@ object Parser2 {
   }
 
   /**
+    * Enumeration of possible separation modes between a list of items.
+    */
+  sealed trait Separation
+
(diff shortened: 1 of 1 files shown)


magnus-madsen (approved):

magnus-madsen:
Thanks!

PR #7710, opened 2024-05-09
fix: capture `main` after bad module

### LINKED ISSUE #7708: Parser: `main` is missed with misplaced doc-comment
`main` is missed in this example:
```scala
mod Bar {
   /// Like so
}
def main(): Unit = ()
```

The program is a parse error, doc-comments must decorate a declaration, but recovery should be able to pick up `main`.

herluf-ba:
Closes #7708 

Example added as a test case, along with two similar examples.


magnus-madsen:
@herluf-ba test failure

magnus-madsen (approved):

PR #7712, opened 2024-05-09
feat: Add colors to parser errors

### LINKED ISSUE #7700: Parser: Add colors to `ParseError`
image

To make it pretty :)

herluf-ba:
Closes #7700

This PR adds colors to parser errors. It also makes sure all errors are similarly worded, has context and the right source location and is reasonably concise.
Here's a preview of how the colors look for this faulty program:
```scala
def one() Int16 = 1i16

def main(): Bool =
    let l = List#{1, 2, 3,};
    List.length(l) == one()
```
Screenshot 2024-05-09 at 12 04 40

Although completely unrelated please take a moment to enjoy the type error despite having two syntax errors.


magnus-madsen:
@herluf-ba Conflicted again, sorry.

magnus-madsen:
I like the changes 👍

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -292,7 +292,8 @@ object CompletionProvider {
       case ResolutionError.UndefinedName(_, _, _, isUse, _) => if (isUse) (1, SyntacticContext.Use) else (2, SyntacticContext.Expr.OtherExpr)
       case ResolutionError.UndefinedType(_, _, _) => (1, SyntacticContext.Type.OtherType)
       case WeederError.MalformedIdentifier(_, _) => (2, SyntacticContext.Import)
-      case ParseError(_, ctx, _) => (5, ctx)
+      case UnexpectedToken(_, _, ctx, _, _) => (5, ctx)
+      case ParseError(_, ctx, _, _) => (5, ctx)
```
magnus-madsen, line 296: Whats the reason for these two additions? How do they impact auto completion?

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -724,7 +727,7 @@ object Weeder2 {
         case Some(expr) => Validation.success(expr)
         case None =>
           // Fall back on Expr.Error. Parser has reported an error here.
-          val err = ParseError("Expected expression", SyntacticContext.Expr.OtherExpr, tree.loc)
+          val err = ParseError(fmt => s"Expected ${fmt.cyan("<expression>")}.", SyntacticContext.Expr.OtherExpr, tree.loc)
```
magnus-madsen, line 730: We usually do all the formatting inside the error. 

Would it be possible to break ParseError into several kinds of errors?

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
herluf-ba: Before we had only `ParseError`, but I split out the most common use-case into the new `UnexpectedToken`.
Adding `UnexpectedToken` with the same priority(5) as `ParseError` just replicates the behavior we had before.

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
herluf-ba: I agree that it's not nice to have to do formatting in `Parser2`. Let me see if I can come up with a nicer solution!

main/src/ca/uwaterloo/flix/language/errors/ParseError.scala
```diff
@@ -18,28 +18,64 @@ package ca.uwaterloo.flix.language.errors
...
+
+/**
+  * An generic parse error. We should prefer [[UnexpectedToken]] when possible.
+  *
+  * @param message Error message to be displayed.
+  * @param ctx     The syntactic context.
+  * @param loc     The source location.
+  * @param hint    Optional hint with more details about the error
+  */
+case class ParseError(message: String, ctx: SyntacticContext, loc: SourceLocation, hint: Option[String] = None) extends CompilationMessage with Recoverable {
```
magnus-madsen, line 68: The "message" here is an anti-pattern. It means that we cannot use the type to determine which specific error is in play.

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
herluf-ba: I see, I'll do another revision then.

main/src/ca/uwaterloo/flix/language/errors/ParseError.scala
herluf-ba: I see, I'll do another revision then.

magnus-madsen: It helps structure the code. Consider WeederError, TypeError etc.

All could have a `message: String`, but it provides no structure.

herluf-ba: Yep, I completely agree. I made something much more akin to `WeederError` now. Could you have another look?

herluf-ba:
@magnus-madsen This is ready for another round of review 👍

magnus-madsen (commented):
I have added a lot of comments. My review boils down to:

- We should think of the error classes themselves as data types in their own right. This means they must be "well designed" and limit how "stringly" typed they are. The reason is that this make the errors reusable, e.g. for VSCode when we want to provide quick fixes, auto-complete etc.

main/src/ca/uwaterloo/flix/language/errors/ParseError.scala
```diff
@@ -18,28 +18,234 @@ package ca.uwaterloo.flix.language.errors
...
+      s"""${line(kind, source.name)}
+         |>> Malformed ${red(name)}.
+         |
+         |${code(loc, s"Here")}$hintStr
+         |""".stripMargin
+    }
+  }
+
+  /**
+    * An __internal__ error indicating that a Mark.Opened did not have a corresponding Mark.Closed when parsing ended.
```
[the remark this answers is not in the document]
herluf-ba: We want to guarantee that every `Mark.Opened` has a corresponding `Mark.Closed`. This is property is asserted in `buildTree`. But we also want each `Mark.Opened` to have a `TreeKind`. Before the mark is closed we _cannot_ know the kind, so instead we use `ErrorTree(UnclosedParserMark(...))`. This is safe because of the assert in `buildTree`. It's mostly a workaround to make types match, so I not particularly invested that things must be this way.

Another design could be to introduce a synthetic `TreeKind.UnclosedMark` with similar doc-comments. This would be another means to the same end, and be just as safe (because of the assert), but maybe we like the flavor of it better?

main/src/ca/uwaterloo/flix/language/errors/ParseError.scala
```diff
@@ -18,28 +18,234 @@ package ca.uwaterloo.flix.language.errors
...
+    override val kind = s"Parse Error ($ctx)"
+    override val context: SyntacticContext = ctx
+
+    def summary: String = {
+      val expectedStr = s"Expected ${prettyJoin(expected)}"
+      val actualStr = actual.map(a => s" before $a").getOrElse("")
+      s"$expectedStr$actualStr."
+    }
+
+    def message(formatter: Formatter): String = {
```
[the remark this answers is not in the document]
herluf-ba: I see your point. The "errors as data-types" makes a lot of sense when we consider IDE-integration also 👍

To me it seems like in this particular case we are pretty much doing that? The string is being built in two-stages, but it's still being built from data that should be generally applicable.

Originally I passed in `TokenKind`s rather than strings, but it doesn't cover all the use-cases. For instance, we want to report that we `Expected <expression>` instead of a list of all the tokens that can lead an expression.

main/src/ca/uwaterloo/flix/language/errors/ParseError.scala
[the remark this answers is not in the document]
magnus-madsen: Let me put it differently: Is this an error the programmer can encounter and fix?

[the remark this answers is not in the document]
herluf-ba: No.

main/src/ca/uwaterloo/flix/language/errors/ParseError.scala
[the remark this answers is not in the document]
magnus-madsen: Yes. This is the good kind of stuff I am looking for! 🚀

main/src/ca/uwaterloo/flix/language/errors/ParseError.scala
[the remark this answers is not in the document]
magnus-madsen: How/where do you discover this problem?

I think TreeKind.UnclosedMark would be better (?)

But maybe thats for another PR.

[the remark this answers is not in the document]
herluf-ba: On second thought, I'll include it here, since `ParseError.UnclosedParserMark` only exists on this branch.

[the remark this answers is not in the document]
magnus-madsen: Why can't that assert throw an exception?

herluf-ba:
@magnus-madsen I addressed all of your comments 👍

magnus-madsen (approved):

magnus-madsen:
Merged! Thanks

PR #7720, opened 2024-05-15
refactor: remove `reduceIfThenElse` from `Inliner`

magnus-madsen:


magnus-madsen:
I am removing this because I have doubts about its soundness. Moreover, I suspect that a later phase will still have to undo it in the presence of control effects.

mlutze (approved):

PR #7722, opened 2024-05-15
refactor: simplify `Purity.fromType`

magnus-madsen:


magnus-madsen:
@JonathanStarup I am making this simplification until proper sets arrive.

JonathanStarup:
I guess its fine but doesnt i just lose performance? whats the upside?

magnus-madsen:
> I guess its fine but doesnt it just lose performance? whats the upside?

The upside is that its obviously correct-- which is important now that we have big issues with some other parts of the compiler :(

JonathanStarup:
This would just mask away effect errors id think (a lot of things just become Control Impure)

I feel like evaluate formula is also obviously correct.

JonathanStarup (approved):

mlutze (approved):

magnus-madsen:
> This would just mask away effect errors id think (a lot of things just become Control Impure)
> 
> I feel like evaluate formula is also obviously correct.

If we had the Verifier then I would be more inclined to keep it.

magnus-madsen:
Also this change is only meant to be temporary.

JonathanStarup:
This code is just 1 to 1 evaluating the formula so I don't think this has any errors - but you can do it if you want, it just seems counterproductive to mask errors if we're currently weeding out effect bugs

PR #7729, opened 2024-05-15
feat: update exception handling in `LanguageServer`

magnus-madsen:


mlutze (approved):
Seems hacky. Wouldn't it be better to ensure stderr is shown?

magnus-madsen:
> Seems hacky. Wouldn't it be better to ensure stderr is shown?

Yeah. I just can never seem to get VSCode to show anything so this is a desperation move.

PR #7737, opened 2024-05-16
fix: match trywith effect in inference and reconstruction

mlutze:


magnus-madsen:
What does this do?

mlutze:
Not every expression has a type or effect variable, because they can be reconstructed simply from their subexpressions. The try-with reconstruction looks like this:
```scala
    case KindedAst.Expr.TryWith(exp, effUse, rules, tvar, loc) =>
      val e = visitExp(exp)
      val rs = rules map {
        case KindedAst.HandlerRule(op, fparams, hexp, _) =>
          val fps = fparams.map(visitFormalParam(_, subst))
          val he = visitExp(hexp)
          TypedAst.HandlerRule(op, fps, he)
      }
      val tpe = subst(tvar)
      val eff = Type.mkUnion(rs.map(_.exp.eff), loc) // TODO temp simplification
      TypedAst.Expr.TryWith(e, effUse, rs, tpe, eff, loc)
```

As I understand it, the `temp simplification` is important for the back-end. So I have changed the inference to correspond to it.

This will allow the effect verifier to behave correctly.

magnus-madsen (approved):

PR #7740, opened 2024-05-17
feat: nice error when keyword is used as name.

herluf-ba:
For instance:
```scala
def new(): Int32 = 123
```
gives 
```
>> Expected <Name>, <greek name>, <name>, <user-defined operator> or <math name> before 'new'.

1 | def new(): Int32 = 123
        ^^^
        Here
Hint: 'new' is a keyword.
```


magnus-madsen (approved):

magnus-madsen:
Thanks!

PR #7748, opened 2024-05-20
feat: add type constructors for method and static method return types

### LINKED ISSUE #7745: Add two new type constructors: `MethodReturnType` and `StaticMethodReturnType`
After `Native` add: 

```scala
  /**
    * A type constructor that represents the _return type_ of a Java method.
    *
    * A method return type can be resolved when the receiver object and argument types are known.
    *
    * A few examples:
    *
    * - The type: `Apply(InvokeMethod("length", 0), String)` is equivalent to `Int32`.
    * - The type: `Apply(Apply(InvokeMethod("startsWith", 1), String), String)` is equivalent to `Bool`.
    * - The type: `Apply(Apply(Apply(InvokeMethod("substring", 2), String), Int32), Int32)` is equivalent to `String`.
    *
    * The type constructor requires at least one type argument: the type of the receiver object.
    */
  case class MethodReturnType(name: String, arity: Int) extends TypeConstructor {
    def kind: Kind = Kind.mkArrow(arity + 1)
  } 

  /**
    * A type constructor that represents the _return type_ of a Java static method.
    *
    * A static method return type can be resolved when the argument types are known.
    *
    * A few examples:
    *
    * - The type: `Apply(InvokeStaticMethod(String.class, "valueOf", 1), Bool)` is equivalent to `String`.
    * - The type: `Apply(InvokeStaticMethod(String.class, "valueOf", 1), Char)` is equivalent to `String`.
    *
    * The type constructor does not require any arguments.
    */
  case class StaticMethodReturnType(clazz: Class[_], name: String, arity: Int) extends TypeConstructor { 
    def kind: Kind = Kind.mkArrow(arity)
  } 
```

chanattan:
Fixes #7745

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/fmt/SimpleType.scala
+++ b/main/src/ca/uwaterloo/flix/language/fmt/SimpleType.scala
@@ -424,6 +424,8 @@ object SimpleType {
         case TypeConstructor.Enum(sym, _) => mkApply(Name(sym.name), t.typeArguments.map(visit))
         case TypeConstructor.RestrictableEnum(sym, _) => mkApply(Name(sym.name), t.typeArguments.map(visit))
         case TypeConstructor.Native(clazz) => Name(clazz.getName)
+        case TypeConstructor.MethodReturnType(name, arity) => throw InternalCompilerException("MethodReturnType shouldn't exist.", t0.loc)
+        case TypeConstructor.StaticMethodReturnType(clazz, name, arity) => throw InternalCompilerException("StaticMethodReturnType shouldn't exist.", t0.loc)
         case TypeConstructor.Ref => mkApply(Ref, t.typeArguments.map(visit))
         case TypeConstructor.Tuple(l) =>
           val tpes = t.typeArguments.map(visit).padTo(l, Hole)
--- a/main/src/ca/uwaterloo/flix/language/fmt/SimpleType.scala
+++ b/main/src/ca/uwaterloo/flix/language/fmt/SimpleType.scala
@@ -424,8 +424,8 @@ object SimpleType {
         case TypeConstructor.Enum(sym, _) => mkApply(Name(sym.name), t.typeArguments.map(visit))
         case TypeConstructor.RestrictableEnum(sym, _) => mkApply(Name(sym.name), t.typeArguments.map(visit))
         case TypeConstructor.Native(clazz) => Name(clazz.getName)
-        case TypeConstructor.MethodReturnType(name, arity) => throw InternalCompilerException("MethodReturnType shouldn't exist.", t0.loc)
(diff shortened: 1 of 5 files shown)


main/src/ca/uwaterloo/flix/language/fmt/SimpleType.scala
```diff
@@ -424,6 +424,8 @@ object SimpleType {
         case TypeConstructor.Enum(sym, _) => mkApply(Name(sym.name), t.typeArguments.map(visit))
         case TypeConstructor.RestrictableEnum(sym, _) => mkApply(Name(sym.name), t.typeArguments.map(visit))
         case TypeConstructor.Native(clazz) => Name(clazz.getName)
+        case TypeConstructor.MethodReturnType(name, arity) => throw InternalCompilerException("MethodReturnType shouldn't exist.", t0.loc)
```
magnus-madsen, line 427: We would probably want some pretty printing here. Ask @JonathanStarup

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -280,6 +280,10 @@ object Simplifier {
 
           case TypeConstructor.Native(clazz) => MonoType.Native(clazz)
 
+          case TypeConstructor.MethodReturnType(name, arity) => throw InternalCompilerException("MethodReturnType shouldn't exist.", tpe.loc)
+
+          case TypeConstructor.StaticMethodReturnType(clazz, name, arity) => throw InternalCompilerException("StaticMethodReturnType shouldn't exist.", tpe.loc)
```
magnus-madsen, line 285: By convention, "error cases" go towards the bottom in the pattern match. See below for some of the other pattern matches.

magnus-madsen: I think we should just say "Unexpected type" like the rest.

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -3314,6 +3314,10 @@ object Resolver {
 
         case TypeConstructor.Native(clazz) => Result.Ok(clazz)
 
+        case TypeConstructor.MethodReturnType(name, arity) => throw InternalCompilerException("MethodReturnType shouldn't exist.", loc)
+
+        case TypeConstructor.StaticMethodReturnType(clazz, name, arity) => throw InternalCompilerException("StaticMethodReturnType shouldn't exist.", loc)
+
```
magnus-madsen, line 3320: I suggest to return `Result.Err(ResolutionError.IllegalType(tpe, loc))` in these two cases for now.

magnus-madsen (commented):
@chanattan @JonathanStarup Added some comments

magnus-madsen:
The PR text should be "Fixes #7745"

magnus-madsen (approved):

PR #7750, opened 2024-05-21
feat: add `InvokeMethod2` to `Parser`

### LINKED ISSUE #7749: Add `InvokeMethod2` to front-end ASTs
Add a new node:

```scala
case class InvokeMethod2(ident: Name.Ident, exp: Expr, exps: List[Expr], loc: SourceLocation) extends Expr
```

Just below the existing `InvokeMethod`.

To begin with, we should just support the syntax:

```
.name1.name2(exps..) // Note the prefix dot
```

Here `name2` is a `Name.Ident`. We should also parse `name1` as a `Name.Ident`, but wrap it in an `Expr.Var`.

That way, we can later extend the parser without making changes to the rest.

chanattan:
Fixes #7749

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser.scala
@@ -1003,7 +1003,7 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
     }
 
     def InvokeMethod2: Rule1[ParsedAst.Expression] = rule {
-      SP ~ "." ~ LowerCaseVariable ~ "." ~ Names.JavaMethod ~ ArgumentList ~ SP ~> ParsedAst.Expression.InvokeMethod2
+      SP ~ "." ~ Names.LowerCaseName ~ "." ~ Names.JavaMethod ~ ArgumentList ~ SP ~> ParsedAst.Expression.InvokeMethod2
     }
 
     def Debug: Rule1[ParsedAst.Expression.Debug] = {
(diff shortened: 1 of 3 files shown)


magnus-madsen (commented):
Great to see some quick progress. Some high-level comments:

- [x] Rename `JavaApply` to `InvokeMethod2` everywhere
- [x] Move `InvokeMethod2` to appear after `Expr.Do` everywhere
- [ ] Add support to `Parser2` and `Weeder2`

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
```diff
@@ -1089,6 +1089,10 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
       SP ~ keyword("force") ~ WS ~ RecordSelect ~ SP ~> ParsedAst.Expression.Force
     }
 
+    def JavaApply: Rule1[ParsedAst.Expression] = rule {
+      SP ~ "." ~ LowerCaseVariable ~ "." ~ Names.JavaMethod ~ ArgumentList ~ SP ~> ParsedAst.Expression.JavaApply
+    }
```
magnus-madsen, line 1094: @herluf-ba Can you help Chanattan by sketching how this should look in Parser2/Weeder2?

magnus-madsen: Strictly speaking, we should not be using ArgumentList here.

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
```diff
@@ -1175,6 +1179,16 @@ class Parser(val source: Source) extends org.parboiled2.Parser {
       SP ~ Names.QName ~ SP ~> ParsedAst.Expression.QName
     }
 
+    def LowerCaseVariable: Rule1[ParsedAst.Expression] = {
```
magnus-madsen, line 1182: Do we need this? I think we can just use `Names.LowerCaseName`.

JonathanStarup: That returns a Name.Ident i believe, thus the fuzz here

magnus-madsen: Lets use Name.Ident then.

JonathanStarup: but that just postpones the trouble no? then its more work later to evaluate this `Ident`

main/src/ca/uwaterloo/flix/language/phase/Parser.scala
magnus-madsen: Ok that is fine

magnus-madsen (approved):

magnus-madsen:
Merged. Thanks!

PR #7751, opened 2024-05-21
feat: invoke method implementation in first front end ast

### LINKED ISSUE #7749: Add `InvokeMethod2` to front-end ASTs

chanattan:
Part of the fixes #7749

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/Indexer.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/Indexer.scala
@@ -225,6 +225,10 @@ object Indexer {
     case Expr.Apply(exp, exps, _, _, _) =>
       visitExp(exp) ++ visitExps(exps) ++ Index.occurrenceOf(exp0)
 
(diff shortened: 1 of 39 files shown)


magnus-madsen:
A quick comment to save you some work: the last AST that will need to be extended is TypedAst. So no changes will be needed in LoweredAst, SimplifiedAst, or later :)

magnus-madsen:
- [ ] Rename `JavaApply` to `InvokeMethod2` everywhere
- [ ] Ensure `InvokeMethod2` appears right after `InvokeMethod`.
- [ ] No changes should be required to ASTs after `TypedAst`.

PR #7759, opened 2024-05-23
fix: PredicateCompleter

herluf-ba:
The following example now invokes `PredicateCompleter`:
```scala
﻿def isConnected(s: Set[(Int32, Int32)], src: Int32, dst: Int32): Bool =
    let rules = #{
        Path(x, y) :- Edge(x, y).
        P // <- Suggests Path/2
    };
    let edges = inject s into Edge;
    let paths = query edges, rules select true from Path(src, dst);
    not (paths |> Vector.isEmpty)
```


magnus-madsen (approved):

magnus-madsen:
Thanks!

PR #7762, opened 2024-05-23
feat: add support for `InvokeMethod2` to `Parser2` and `Weeder2`

### LINKED ISSUE #7760: Add support for `InvokeMethod2` to `Parser2` and `Weeder2`

chanattan:
Work in progress.
This should fix #7760.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/Ast.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/Ast.scala
@@ -847,10 +847,6 @@ object Ast {
 
       case object Do extends Expr
 
-      case object Dot extends Expr
-
-      case object InvokeMethod2 extends Expr
-
       case object OtherExpr extends Expr
     }
 
--- a/main/src/ca/uwaterloo/flix/language/ast/Ast.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/Ast.scala
@@ -847,6 +847,10 @@ object Ast {
 
       case object Do extends Expr
 
+      case object Dot extends Expr
+
+      case object InvokeMethod2 extends Expr
+
       case object OtherExpr extends Expr
     }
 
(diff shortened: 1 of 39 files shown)


main/src/ca/uwaterloo/flix/language/ast/Ast.scala
```diff
@@ -847,6 +847,10 @@ object Ast {
 
       case object Do extends Expr
 
+      case object Dot extends Expr
```
herluf-ba, line 850: This is not needed 🙂

main/src/ca/uwaterloo/flix/language/ast/TokenKind.scala
```diff
@@ -690,6 +693,8 @@ object TokenKind {
 
   case object KeywordDo extends TokenKind
 
+  case object KeywordDot extends TokenKind
```
herluf-ba, line 696: Remove this 👍 We should be able to get by without new `TokenKind`s

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -2336,6 +2337,16 @@ object Parser2 {
       close(mark, TreeKind.Expr.Do)
     }
 
+    private def doInvokeMethod2Expr()(implicit s: State): Mark.Closed = {
+      assert(at(TokenKind.Dot))
+      val mark = open()
+      expect(TokenKind.Dot, SyntacticContext.Expr.Dot)
+      name(NAME_QNAME, context = SyntacticContext.Expr.InvokeMethod2)
```
herluf-ba, line 2344: I think you can just remove this line and the line below it.

herluf-ba: Lines 2344 and 2345 to be specific

chanattan: Thank you for all your solutions. I can see clearer now.
However, it seems like that with the current code it only parses one dot, so I added another expect TokenKind.Dot statement. In addition, expression instead of name seemed to not parse `x` only. I will keep trying some things before pushing, otherwise it seems to work now with above changes.

magnus-madsen:
@herluf-ba Let me know when it looks good and I should take a look.

Also, do double check that things occur "in the right order".

PR #7766, opened 2024-05-25
fix: `HoleCompleter`

herluf-ba:
This example now triggers the `HoleCompletor`
```scala
def foo(): Bool = 
    let l = List.range(1, 1000);
    l? // <- Completions triggers here
```


magnus-madsen (approved):

magnus-madsen:
Thanks! -- I think this completes the completers?

PR #7772, opened 2024-05-27
fix: source location on built-in binary ops

### LINKED ISSUE #7637: Parser: issue with semantic tokens / source locations
See here:

image

It seems that the parsing/weeding/desugaring of `+` makes it token span too long. I think this behavior is new. 

Note how the literal of `x` is colored compared to `y`.

herluf-ba:
Closes #7637 

The example from the issue now has proper highlighting. The issue was a wrong source location on buitin operators.

Screenshot 2024-05-27 at 09 18 53


magnus-madsen (approved):

magnus-madsen:
Thanks!

PR #7774, opened 2024-05-27
fast set unification experimentation

JonathanStarup:
- rename ops to set language
- make solver symmetrical
- add set elem
- flip lattice

#7542 has been merged into this - real change is just `FastSetUnification` (and `EffUnification2`)


magnus-madsen:
Once Edward and you are happy with the impl, I would propose:

- Create a new PR with only FastSetUnif and TestFastUnif
- Then we can iterate on that and get it merge in

Separately, try to tweak a few type rules.

Lastly, when Matt is home, and the last bugs are fixed, then he can build on top of FastSetUnif which is ready for him.

PR #7777, opened 2024-05-27
fix: avoid double reporting lexer errors

herluf-ba:
A simple change that avoids mentioning `TokenKind.Err` in parser errors.


magnus-madsen:
Fix appreciated, but seems a helper function would be useful no?

herluf-ba:
> Fix appreciated, but seems a helper function would be useful no?

Apparently past-me thought so too, since I found that the functionality is already implemented in `closeWithError` 🤷‍♂️ Just needed a small addition to cover this use-case too.

magnus-madsen (approved):

PR #7779, opened 2024-05-27
feat: LexerError for free dots

### LINKED ISSUE #7727: Parser: disallow whitespace in qualified names
This:

```
println(area(Shape    .    Rectangle(2, 4)))
```

should not allowed. If we allow it, it could cause problems for semi colon inference later, I think.

herluf-ba:
Closes #7727 

With the example from the issue we now get error:
```
-- Lexer Error -------------------------------------------------- main/foo.flix

>> '.' has whitespace on both sides.

18 |     area(Shape    .    Rectangle(2, 4)) |> println
                       ^
                       here
```
Note that the rule is that no `.` can have whitespace _on both sides_. That's because we need to be able to parse laws and fixpoint constraints:
```scala
law reflexivity: forall(x: a). x == x
//                           ^ Like this

// But this means that this is also fine:
println(area(Shape.    Rectangle(2, 4)))
```
All whitespace information is lost after the lexer, so the check needs to be in this phase. We can't do special handling of laws and fixpoint constraints in the lexer though, because it's context-free. Hence this compromise.

### DIFF
--- a/examples/larger-examples/datalog/sequence.flix
+++ b/examples/larger-examples/datalog/sequence.flix
@@ -2,7 +2,7 @@
 
 def main(): Unit \ IO =
 
-    let numLetters = #{NumLetters(20).};
+    let numLetters = #{ NumLetters(20). };
 
     let lp = #{
 
--- a/main/src/ca/uwaterloo/flix/language/ast/TokenKind.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/TokenKind.scala
@@ -52,6 +52,7 @@ sealed trait TokenKind {
       case TokenKind.CurlyR => "'}'"
       case TokenKind.Dollar => "'$'"
       case TokenKind.Dot => "'.'"
+      case TokenKind.DotWhiteSpace => "'. '"
       case TokenKind.DotCurlyL => "'.{'"
       case TokenKind.Equal => "'='"
       case TokenKind.EqualEqual => "'=='"
@@ -637,6 +638,8 @@ object TokenKind {
   case object Dollar extends TokenKind
 
   case object Dot extends TokenKind
+
+  case object DotWhiteSpace extends TokenKind
 
   case object DotCurlyL extends TokenKind
 
--- a/main/src/ca/uwaterloo/flix/language/errors/LexerError.scala
+++ b/main/src/ca/uwaterloo/flix/language/errors/LexerError.scala
@@ -111,18 +111,18 @@ object LexerError {
   }
 
   /**
-    * An error raised when a period is surrounded by whitespace on both sides.
-    * This is problematic because we want to disallow tokens like: "Rectangle   .   Shape".
+    * An error raised when a period has whitespace before it.
+    * This is problematic because we want to disallow tokens like: "Rectangle   .Shape".
     *
     * @param loc The location of the '.'.
     */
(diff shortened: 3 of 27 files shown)


herluf-ba:
@magnus-madsen I can see that there are many examples of laws ending on a dot with surrounding whitespace:
```scala
law reflexivity: forall(x: a) . x == x
//                            ^ Like this
```
I have updated most cases for this PR by I wonder if this something that we would like to still support? We could do something entirely different from what I have implemented here. Instead we could have a `TokenKind.DotWS` that's different from a `TokenKind.Dot`. Then we could expect the different kinds in the parser.

magnus-madsen:
Laws are not yet support or used. How about we simply change their syntax. We can drop the dot. Do you have a suggestion?

mlutze:
> @JonathanStarup @mlutze want to suggest some new syntax? What would this look like in Coq or Isabelle?

Coq uses a comma instead of a dot. It would be a little weird here, but I could live with it.

magnus-madsen:
> > @JonathanStarup @mlutze want to suggest some new syntax? What would this look like in Coq or Isabelle?
> 
> Coq uses a comma instead of a dot. It would be a little weird here, but I could live with it.

Do we even need a separator?

magnus-madsen:
Why not equals etc?

magnus-madsen:
@JonathanStarup Can you show us how you would write a property in Coq/Isabelle?

mlutze:
> Why not equals etc?

Equals doesn't make sense to me. I'm okay with no punctuation as well.

magnus-madsen:
What about thin or thick arrow?

JonathanStarup:
would be
```isabelle
lemma reflexivity: "forall (x :: a) . x = x"
```

JonathanStarup:
We could also require the parenthesis, thus using it as the delimiter
```
law reflexivity: forall(x: a) x == x
```

magnus-madsen:
I agree. Let us simply drop the dot.

herluf-ba:
@magnus-madsen @JonathanStarup That's great, I'll update the parser and all laws in tests and examples with the new syntax.

We still need to consider fixpoint constraints though, as they terminate on `.` and have a newline after them. Could we allow newline on the right-hand side of dots?

JonathanStarup:
sounds sensible

magnus-madsen:
Would it be sensible to have two tokens? Dot and DotWhiteSpace?

magnus-madsen:
>  Could we allow newline on the right-hand side of dots?

I guess we can. I don't think this would break future semi colon inference because an expression cannot end with a dot.

herluf-ba:
New syntax for laws: The dot is out!

New syntax for fixpoint constraints: There must be a newline after the `.` ending the constraint.

New syntax for qualified names: There can be no whitespace around the `.` between names.

magnus-madsen (approved):

magnus-madsen:
@herluf-ba Looks good, but some compile error.

magnus-madsen (approved):

herluf-ba:
@mlutze Syntax for laws changed 🙂 Now there is no dot!
Would you mind deleting the `.` in your community repositories at these spots:
```
  flix-json > Jsonable.flix > line 26
  flixball > State.flix > line 9
```

PR #7789, opened 2024-05-29
feat: Explicit Export

JonathanStarup:
## Example
```
mod Something.Cool {
    @Export
    pub def plusOne(x: Int32): Int32 = x + 1
}
```
This ensures that `Something.Cool` has method
```
public static final int plusOne(int x) {...}
```
exists. If you do something you're not allowed to (polymorphism, bad name, flix types, control effects) then it'll give you a safety error.

## Whats done

This PR has no actual work in it, it just exposes the amount of export that we already support (simple types and Object without control effects).

- Renamed `A.B.C.Ns$` classes to `A.B.C` (or `Root$` for the empty namespace)
  - Renamed exported methods inside from `m_name` to `name`
- Added `@Export`
  - These are entrypoints
  - They have specific checking in Safety, which can result in the new `IllegalExportXYZ` errors


magnus-madsen:
Cool!

Is plusOne allowed to use all the good stuff inside, e.g. control effects?

magnus-madsen:
I am surprised the diff is so small. This is great. I will add some comments and then we can merge and document.

magnus-madsen:
But first I need to think a bit. This only works for "Java like signatures", right?

JonathanStarup:
inside everything can be used yes, and yeah its only a small set of types (which we can improve over time)

JonathanStarup:
I have clarified and made doc comments now

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmOps.scala
```diff
@@ -246,8 +246,7 @@ object JvmOps {
   }
 
   private def getNamespaceName(ns: List[String]): JvmName = {
-    val name = JvmName.mkClassName("Ns")
-    JvmName(ns, name)
+    JvmName(ns, "Export")
```
magnus-madsen, line 249: Do we want it to reside in Export? Why can it not just be where we expect?

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmOps.scala
```diff
@@ -258,7 +257,10 @@ object JvmOps {
     * find      =>  m_find
     * length    =>  m_length
     */
-  def getDefMethodNameInNamespaceClass(sym: Symbol.DefnSym): String = "m_" + mangle(sym.name)
+  def getDefMethodNameInNamespaceClass(defn: ReducedAst.Def): String = {
+    if (defn.ann.isExport) defn.sym.name
```
magnus-madsen, line 261: Why is it not necessary to mangle the exported name? 

I assume because of restriction on it, but if so we should document it here.

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -95,14 +98,40 @@ object Safety {
...
       } else {
         SafetyError.IllegalEntryPointSignature(defn.sym.loc) :: Nil
       }
     }
   }
 
+  /**
+    * Returns `true` if the given `defn` is annotated to be exported.
+    */
+  private def isExported(defn: Def): Boolean = {
```
magnus-madsen, line 113: I think this helper belongs on Ast.Annotations or around there.

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -117,6 +146,69 @@ object Safety {
...
+    *
+    * Currently this is only types that are immune to erasure, i.e. the
+    * primitive java types and Object.
+    *
+    * Note that signatures should be exportable as-is, this means that even
+    * though the signature contains `MyAlias[Bool]` where
+    * `type alias MyAlias[t] = t`, it uses flix-specific type information which
+    * is not allowed.
+    */
+  private def isExportType(tpe: Type): Boolean = tpe.typeConstructor match {
```
magnus-madsen, line 186: -> isExportableType

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -117,6 +146,69 @@ object Safety {
...
+    case None => false
+    case Some(value) => value match {
```
magnus-madsen, line 188: value -> cst

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -117,6 +146,69 @@ object Safety {
...
+      case TypeConstructor.Bool => true
+      case TypeConstructor.Char => true
+      case TypeConstructor.Float32 => true
+      case TypeConstructor.Float64 => true
+      case TypeConstructor.Int8 => true
+      case TypeConstructor.Int16 => true
+      case TypeConstructor.Int32 => true
+      case TypeConstructor.Int64 => true
+      case TypeConstructor.BigInt => true
```
magnus-madsen, line 197: String? :)

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -117,6 +146,69 @@ object Safety {
...
+      // Error is accepted to avoid cascading errors
+      case TypeConstructor.Error(_) => true
+      case _ => false
+    }
+  }
+
+  /**
+    * Returns `true` if given `defn` has a name that is directly valid in Java.
+    */
+  private def validJavaName(defn: Def): Boolean = {
```
magnus-madsen, line 208: Define on Symbol.DefnSym, not on Def

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -95,14 +98,40 @@ object Safety {
...
+  }
+
+  /**
+    * Checks that the function is able to be exported.
+    *
+    * The function should be public, use a Java-valid name, be non-polymoprhic,
+    * use exportable types, and use non-algebraic effects
+    */
+  private def visitExportDef(defn: Def): List[SafetyError] = {
+    val pub = if (isPub(defn)) Nil else List(SafetyError.IllegalExportSignature(defn.sym.loc, "Exported functions must be `pub`."))
```
magnus-madsen, line 124: We usually introduce specific errors for each error type.

magnus-madsen:
This seems like a reasonable start... Assume we merge this once the things are fixed.

What would be the next steps towards more realistic export functionality?

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmOps.scala
JonathanStarup: I didn't know what to do about the Root namespace edge case

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmOps.scala
JonathanStarup: The name is checked in safety, so mangling is not required - I can add a comment

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
JonathanStarup: String is erased to Object, so the signature is not unchanged. This would require actual infrastructure

JonathanStarup:
# Support java objects
- We need to avoid type erasure on exported signatures, should be simple changes in `Eraser`
- Allow it in safety checking
# Other things
- Allow nullary functions (just need to auto-generate a wrapper, maybe just last second in the backend)
- Allow `Unit` as a return type (just need to auto-generate a wrapper, maybe just last second in the backend)
- Allow polymorphic types (should be easy, but requires annotations to bytecode)

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmOps.scala
magnus-madsen: Hmm. True. You could disallow it the same way you disallow polymorphism etc.

JonathanStarup:
@magnus-madsen now theres negative tests and the `Export` class part has been removed

magnus-madsen (approved):

magnus-madsen:
Thanks!

LionelMeli:
Hello Jonathan,

Thanks for this feature.
Could we change the classes generated like Namespaces classes, Fn1${...} classes, Clo1${...} from default java package to dev.flix.runtime for example ?
Because the only way to use the classes generated with @Export annotation in Java as the classes are in default package is through introspection...

JonathanStarup:
The classes like `fn$1` are not meant to be used externally - they use our custom calling convention that doesn't use the java stack. The classes that are meant for external use are in `Ns.something` and I hope that they can be used easily?

LionelMeli:
Hi,

I understand. My point here is more about the default package where are located the classes with Export annotation

Suppose we've this code:
```
mod BasicFct{
    import java.lang.Object

    @Export
    pub def add2Int32(x: Int32, y: Int32): Int32 = x + y

    @Export
    pub def add2String(x: Object, y: Object): Object =
        let s1 = unchecked_cast(x as String);
        let s2 = unchecked_cast(y as String);
        checked_cast(s1+s2)

}
```
The Flix compiler (0.48.0) generates into the default package the class BasicFct with 2 methods add2Int32 and add2String...
and generates into the package BasicFct 2 classes Def$add2Int32 and Def$add2String.
If I want use the 2 methods exported I must call it through intropection. It should be great if the methods exported are in one(or more) classe(s) not in the default package.

JonathanStarup:
oh so we should put the basicfct class into the basicfct package? would that help? or put basicfct into the flix package?

LionelMeli:
BasicFct class could be everywhere except the default package. However into basicfct package it's ok.

PR #7814, opened 2024-06-01
refactor: use `SourceLocation` in `NName`

magnus-madsen:


main/src/ca/uwaterloo/flix/language/phase/Namer.scala
```diff
@@ -99,7 +99,7 @@ object Namer {
     */
   private def visitNamespace(decl: DesugaredAst.Declaration.Namespace, ns0: Name.NName)(implicit flix: Flix): Validation[NamedAst.Declaration.Namespace, NameError] = decl match {
     case DesugaredAst.Declaration.Namespace(ident, usesAndImports0, decls0, loc) =>
-      val ns = Name.NName(ident.sp1, ns0.idents :+ ident, ident.sp2)
+      val ns = Name.NName(ns0.idents :+ ident, ident.loc)
```
magnus-madsen, line 102: Unclear if the current code and new code is correct.

mlutze: Depends on what the location of an NName means. I don't know if we even use it anywhere.

mlutze (approved):

PR #7817, opened 2024-06-01
refactor: use `SourceLocation` in `QName`

magnus-madsen:
Towards only using source locations in Name


magnus-madsen:
A careful review to see that I did not screw up would be useful.

main/src/ca/uwaterloo/flix/language/ast/Name.scala
```diff
@@ -26,31 +26,25 @@ object Name {
...
     * Creates a qualified name from the given namespace `ns` and name `name`.
     */
-  def mkQName(ns: List[String], name: String, sp1: SourcePosition, sp2: SourcePosition): QName = {
-    val loc = SourceLocation(isReal = true, sp1, sp2)
-    val nname = NName(ns.map(t => Name.Ident(sp1, t, sp2)), loc)
-    val ident = Ident(sp1, name, sp2)
-    QName(sp1, nname, ident, sp2)
+  def mkQName(ns: List[String], name: String, loc: SourceLocation): QName = {
+    val nname = NName(ns.map(t => Name.Ident(loc.sp1, t, loc.sp2)), loc)
+    val ident = Ident(loc.sp1, name, loc.sp2)
```
mlutze, line 44: Will we change Ident to use `SourceLocation` too?

main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
```diff
@@ -1135,7 +1135,7 @@ object Desugar {
     // ignore redundant effect casts to IO
     val correctedEff = eff match {
       case None => None
-      case Some(DesugaredAst.Type.Ambiguous(Name.QName(_, Name.NName(Nil, _), Name.Ident(_, "IO", _), _), _)) => None
+      case Some(DesugaredAst.Type.Ambiguous(Name.QName(Name.NName(Nil, _), Name.Ident(_, "IO", _), _), _)) => None
```
mlutze, line 1138: Can we just check that the name is unqualified and matches IO? This is hard to read.

main/src/ca/uwaterloo/flix/language/ast/Name.scala
```diff
@@ -26,31 +26,25 @@ object Name {
...
+    QName(nname, ident, loc)
   }
 
-  /**
-    * Returns the given identifier `ident` as qualified name in the root namespace.
-    */
-  def mkQName(ident: Ident): QName = QName(ident.sp1, RootNS, ident, ident.sp2)
```
mlutze, line 52: Why did we get rid of this helper? Seems like it's used a lot.

mlutze (approved):

main/src/ca/uwaterloo/flix/language/ast/Name.scala
magnus-madsen: Yes-- thats my ultimate plan. I am taking small steps towards that.

main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
magnus-madsen: I suppose I can refactor.

main/src/ca/uwaterloo/flix/language/ast/Name.scala
magnus-madsen: My plan is to get rid of these smart constructors and focus _on the data_.

magnus-madsen: We could an expression to deal with that.

mlutze: IMO the root of the problem is that our standard library is unqualified. We should just have a `Flix` namespace and put the who stdlib in it.

mlutze: Then we can fully separate qualified and unqualified names

magnus-madsen: This is orthogonal though

mlutze: Ok maybe that's actually orthogonal, but we should do it anyway :)

PR #7819, opened 2024-06-02
chore: Updated SemVer --- patch is now required and removed maven version artifacts

### LINKED ISSUE #7795: SemVer specification
At the moment [SemVer.scala](https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/tools/pkg/SemVer.scala) does not follow the official specification of [SemVer](https://semver.org/).

Consider the definition we currently use:
https://github.com/flix/flix/blob/acb4e30fd09a242577853a1f38236e935e14bbd2/main/src/ca/uwaterloo/flix/tools/pkg/SemVer.scala#L33

This has multiple issues if you compare it directly to the SemVer specification.
1. _Patch is not optional._ This may a result of some confusion with how Rust allows dependencies to be specified --- Package version in Rust _must_ follow SemVer, but they support some shorthand for dependencies (https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html). If we want to support this, we should be clear about it.
2. `buildDot` parameter, seems out of place, which is also apparent when looking at the `toString` method. This seems to be a result of mixing Maven versioning and SemVer versioning.

I suggest moving to something a long the lines of:
```scala
case class SemVer(major: Int, minor: Int, patch: Int, preRelease: Option[String], build: Option[String])
```

I would also suggest we add a parse method to `SemVer.scala` as there seems to be some code duplication for parsing:
- https://github.com/flix/flix/blob/acb4e30fd09a242577853a1f38236e935e14bbd2/main/src/ca/uwaterloo/flix/tools/pkg/github/GitHub.scala#L328
- https://github.com/flix/flix/blob/acb4e30fd09a242577853a1f38236e935e14bbd2/main/src/ca/uwaterloo/flix/tools/pkg/ManifestParser.scala#L218

AStenbaek:
fix #7795 

This is probably there bare minimum.
I still think a parsing method should be added to `SemVer.scala`.


magnus-madsen (approved):

magnus-madsen:
Thanks!

PR #7822, opened 2024-06-03
fix: handle empty singleton enum

### LINKED ISSUE #7804: Parser: Handle empty singleton enums
```
enum Foo()
```

This creates a malformed enum where the argument is an empty tuple, which is an invalid type. I'm guessing this is a parser issue.

herluf-ba:
Closes #7804 

With this example
```scala
enum Foo()
```
We now get error:
```
-- Parse Error (OtherType) -------------------------------------------------- main/foo.flix

>> Expected at least one <type>.

13 | enum Foo()
             ^
             Here
```

And crucially no cases are passed on in the resulting `WeededAst.Enum` meaning that we no longer spawn an invalid empty tuple type.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -1014,7 +1014,17 @@ object Parser2 {
       val isShorthand = at(TokenKind.ParenL)
       if (isShorthand) {
         val markType = open()
-        Type.tuple()
+        val mark = open()
+        oneOrMore(
+          namedTokenSet = NamedTokenSet.Type,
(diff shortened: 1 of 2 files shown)


herluf-ba:
@mlutze Could you verify that the behavior with empty singleton enums is what we want here?

magnus-madsen:
Could we have a test case? :)

herluf-ba:
> Could we have a test case? :)

Sure 👍

magnus-madsen (approved):

magnus-madsen:
Conflicted, sorry!

mlutze:
> @mlutze Could you verify that the behavior with empty singleton enums is what we want here?

This looks fine to me.

Issue #7116, opened 2024-01-24
Unconfident wording in effect error


mlutze:
```
'r + IO' appears to be a superset of 'IO'.
Consider using 'checked_ecast'?
```

Surely we _know_ that it's a subset. We should just say "is a subset"

JonathanStarup:
It seems that it is wrong sometimes, so maybe not 😄 (from #7118)
```
'Pure' appears to be a superset of 'e1 + e0'.
```

mlutze:
Maybe instead all our type errors should be tentative. "We're not totally sure, but we think there might be a problem with this type"

JonathanStarup:
Gone?

PR #7825, opened 2024-06-03
feat: `FastSetUnification`

### LINKED ISSUE #7118: Typing issue with handlers
Why does the following not type check?:

```scala
eff Ask {
    pub def ask(x: Int32): Unit
}

eff Gen {
    pub def gen(): Int32
}

def runGenAsk(): Int32 =
    try {
        try {
            do Ask.ask(do Gen.gen()); 84
        } with Gen {
            def gen(k) = k(42)
        }
    } with Ask {
        def ask(x, k) = { k() }
    }
```

### LINKED ISSUE #7209: Typing of multiple handlers
```
eff Ask {
    pub def ask(): String
}

eff Say {
    pub def say(s: String): Unit
}

def greeting(): Unit \ {Ask, Say} = 
    let name = do Ask.ask();
    do Say.say("Hello Mr. ${name}")

def main(): Unit \ IO = 
    try {
        try greeting() with Ask {
            def ask(k) = k("Bond, James Bond")
        }
    } with Say {
        def say(s, k) = { println(s); k() }
    }
```

### LINKED ISSUE #7699: Add `FastSetUnification`
- [ ] Use `FastBoolUnification` as baseline.
- [ ] Copy `TestFastBoolUnification` to `TestFastSetUnification`
- [ ] Replace `Term.True` and `Term.False` by a new `Term.Val(s: Set[Int])` -- this represents a concrete set. 
- [ ] Add an alias `Term.Bot` which is just `Term.Val(empty)`.
- [ ] Add a new term `Term.Top` -- this symbolically represents the universe.
- [ ] Ensure that `propagateAnd` is _only_ used if a term does not contain any concrete sets (e.g. Term.Val). ***Otherwise it is incorrect***.

JonathanStarup:
Fixes #7699
Fixes #7209
Fixes #7118
Maybe fixes part of #7116


magnus-madsen:
What's the plan for polish? I think it makes sense to do a thorough pass through the code before I look at it in detail.

JonathanStarup:
Yeah ill give it a full look over and ping you

magnus-madsen:
> Yeah ill give it a full look over and ping you

Ok. Since this is a such a central piece of what we do, I have a high standard :)

It would be good to ensure all comments are up to date and that the code is as clean as possible.

magnus-madsen (commented):
I will add general comments before code-level comments:

- Test cases do not seem to contain elements.
- I am not sure I like SetEval it seems like an extra level of abstraction. At least it should not be written in OO-style.
- Why does Term suddenly have both & and inter? How many ways do we know have to construct terms? :)
- What is freeUnknowns?
- toRawString ugh? Maybe some of this can be moved to a debugging class.
- I still don't understand why Inter is defined the way it is with an option. It seems very asymmetric.

JonathanStarup:
* Test cases do not seem to contain elements.
  * The fuzzed tests do. Do you want me to extract examples for std lib? 
* I am not sure I like SetEval it seems like an extra level of abstraction. At least it should not be written in OO-style.
  * I've changed it to func style.
  * I like that we don't have to know the universe. and its just 20 lines (excluding comments)
* Why does Term suddenly have both & and inter? How many ways do we know have to construct terms? :)
  * Oversigt, its simplified now 
* What is freeUnknowns?
  * Constants and variables. Before we called flexify that converted csts to vars. instead evaluate now just views constants like that to begin with, saving a formula conversion
* toRawString ugh? Maybe some of this can be moved to a debugging class.
  * Moved to the fuzzer. We can also delete it, it was just useful for debugging.
* I still don't understand why Inter is defined the way it is with an option. It seems very asymmetric.
  * `{e_12} ∩ {e_42} = empty` so structurally we reflected that. union doesn't behave like that so it is asymmetric. I can also make it a zero/singleton set if you want.

JonathanStarup:
@magnus-madsen ive given it all a read-through and rewrite but it probably requires multiple, being so long.

magnus-madsen:
> * Test cases do not seem to contain elements.
>       
>       * The fuzzed tests do. Do you want me to extract examples for std lib?

Yes, definitely. This would be very useful.

>       * Constants and variables. Before we called flexify that converted csts to vars. instead evaluate now just views constants like that to begin with, saving a formula conversion

Do we assume the domains are disjoint?

>       * `{e_12} ∩ {e_42} = empty` so structurally we reflected that. union doesn't behave like that so it is asymmetric. I can also make it a zero/singleton set if you want.

But what about `{throw, print} ∩ ef`. Why is that not allowed to have a compact representation?

magnus-madsen:
> It is less cool since we don't use minus yet, but I can find some examples

Regions should be their own effects. Then there is plenty to take from.

JonathanStarup:
Compact concrete sets:
1. We could change the element part of an intersection to represent a union
  a. This means that the most compact element rep is an intersection and that `{a} u {b}` should simplify to `Inter({a, b}, empty...)`
  b. This allows simplifying fx `({a} u {b}) inter ({b} u {c})` since it becomes a nested intersection
2. We could change `Elem(Int)` to `ElemSet(Set[Int])`
  a. both union and intersection will just have zero/one element set
  b. The most compact element rep is an element set

I think the second is the most comprehensible, do you have a comment @magnus-madsen?

magnus-madsen:
@JonathanStarup Can we do both?

JonathanStarup:
Efficient really is the enemy of elegant

magnus-madsen:
We should definitely do (2) then we can fight over (1).

JonathanStarup:
!!!jar

magnus-madsen:
If you add support for try-with I will give it a spin :)

JonathanStarup:
@magnus-madsen the formulas of the two are very similar in size

The have slightly different behaviour for associated effects on free variables - the old solver might say that `x1 ~ Aef[x2]` gives `x1 -> Aef[x2]` whereas the new solver says "no, i cant solve until you figure out what x2 is". both are correct

they disagree! the old solver will say that `Eff1 - Eff2` is not equivalent to `Eff1`, but the new solver knows

magnus-madsen:
> @magnus-madsen the formulas of the two are very similar in size
> 
> The have slightly different behaviour for associated effects on free variables - the old solver might say that `x1 ~ Aef[x2]` gives `x1 -> Aef[x2]` whereas the new solver says "no, i cant solve until you figure out what x2 is". both are correct

Can I get access to the raw data?

magnus-madsen:
!!!jar

magnus-madsen:
Maybe I am having a long day, but I tried:

```flix

eff Throw {
    pub def throwIt(): Void
}

def handler(f: Unit -> Unit \ ef): Unit -> Unit \ ef - Throw = ???


def main(): Unit \ IO = 
    let f = () -> do Throw.throwIt();
    let g = handler(f);
    println("Hello World!")
    
```

And the type of `g` is: `Unit -> Unit \ Throw & (~Throw)`

JonathanStarup:
That's correct, but its equivalent to pure
```
eff Throw {
    pub def throwIt(): Void
}

def handler(f: Unit -> Unit \ ef): Unit -> Unit \ ef - Throw = ???

def mustBePure(f: a -> b \ {}): a -> b \ {} = f

def main(): Unit \ IO =
    let f = () -> do Throw.throwIt();
    let g: _ -> _ \ {} = handler(f);
    let _ = mustBePure(g);
    println("Hello World!");
    ???
```

magnus-madsen:
Ping me when I should take another look.

JonathanStarup:
!!!jar

JonathanStarup:
Try it out @magnus-madsen - now with `without` and `try-with` using minus

magnus-madsen:
Super excited!

(Just need to find some time, lots of teaching)

magnus-madsen:
How should @mlutze and I approach the review of this? What's the parts we should be most careful about.

JonathanStarup:
Good question - `FastSetUnification` is at the heart of it but I still think I can polish more. maybe look at the big picture stuff first

magnus-madsen:
Master:

```
Throughput (best): 123,074 lines/sec (with 10 threads.)

  min: 86,808, max: 123,074, avg: 110,489, median: 112,445

Finished 51 iterations on 66,877 lines of code in 30 seconds.
```

Branch:

```
Throughput (best): 110,639 lines/sec (with 10 threads.)

  min: 82,137, max: 110,639, avg: 103,958, median: 105,314

Finished 51 iterations on 66,877 lines of code in 32 seconds.
```

JonathanStarup:
is that good enough or do you want it to be as fast? I haven't done much optimizing, and not specifically against the actual compiler

magnus-madsen:
> is that good enough or do you want it to be as fast? I haven't done much optimizing, and not specifically against the actual compiler

I think its good enough 👍 Lets leave performance as-is, and focus on correctness and polish.

magnus-madsen:
I see you have a lot of debugging infrastructure:

- Do you know for sure this does not print at some inconvenient time?
- Can we move it into debugging classes?

magnus-madsen (commented):
Overall the code has clear structure.

I want to start with some very high-level questions:

- The code for operating on terms (constructing and evaluating them) consists of simple function. In contrast, the rule and phase abstractions have multiple layers of indirection and data types with complicated control-flow. Why is this necessary? And if there is an argument for it, could we not use traits/interfaces/type classes in a more principled way?
- I sometimes see return types with Option[ConflictException]. This seems strange.
- Would it make sense to push the set representation further and unify Inter and Union? Then a Boolean/two-valued enum would tell us which one is which. Or does that go too far?

magnus-madsen (commented):
The structure is not that bad, actually.

I think it identified some good abstraction, but then it has an over-reliance on type aliases and the Option. Lastly, it got confused trying to both (1) track debug information and (2) becoming its own meta framework.

magnus-madsen:
!!!jar

JonathanStarup:
theres no handler? how does that even parse

mlutze:
>     * Do you know for sure this does not print at some inconvenient time?

Sounds like something an effect system would help with!

magnus-madsen:
Low-level comment: We need some proofing in the presence of Type.Error.

PR #7838, opened 2024-06-06
fix: add termination to module recover loop

herluf-ba:
This fixes a bug I caught while fuzzing the parser.

Here's a minimal example the produces the issue:
```scala
mod DelayMap {
    @Experimental
    pub de
```
This example has been added as a test case. 
3 tests were unignored as well, since they all pass.


magnus-madsen (approved):

magnus-madsen:
Thanks!

(Fuzzing is great!)

Issue #7769, opened 2024-05-26
Extend Method Resolution


magnus-madsen:
Each can be its own PR:

- [x] Add support for the types: `BigInt`, `BigDecimal`, and `String`
- [x] Add `TypeError.MethodNotFound` and `TypeError.AmbiguousMethod` and use these when the result set is empty or > 2.
- [x] Add support for the type `Native` (which has a reference to a class)

magnus-madsen:
CC @chanattan @mlutze

chanattan:
#7844 should take care of the three above tasks.

PR #7844, opened 2024-06-06
feat: extend method resolution with `TypeError.AmbiguousMethod`

chanattan:
Fixes one of #7769 tasks.
WIP:
- more tests should be done with an ambiguous method, e.g., with overloading.
- cleaning code
- finish subTyping

`valueOf` method of `String` doesn't work as it is a static method, which gives upon call:

` Exception in thread "main" java.lang.IncompatibleClassChangeError: Expecting non-static method 'java.lang.String java.lang.String.valueOf(int)'`

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/errors/TypeError.scala
+++ b/main/src/ca/uwaterloo/flix/language/errors/TypeError.scala
@@ -102,7 +102,7 @@ object TypeError {
       s"""${line(kind, source.name)}
          |>> Java method '$methodName' from type '${red(formatType(tpe0, Some(renv)))}' with arguments types (${tpes.mkString(", ")}) is ambiguous.
          | Possible candidate methods:
-         |  ${methods.map(m => m.getName).mkString(", ")}
+         |  ${methods.map(m => s"${m.getName}(${m.getParameterTypes.map(t => t.getName).mkString(", ")})").mkString(", ")}
          |
          |${code(loc, s"java method '${methodName} ambiguous")}
          |""".stripMargin
--- a/main/src/ca/uwaterloo/flix/language/errors/TypeError.scala
+++ b/main/src/ca/uwaterloo/flix/language/errors/TypeError.scala
@@ -24,6 +24,7 @@ import ca.uwaterloo.flix.language.ast._
 import ca.uwaterloo.flix.language.fmt.FormatEqualityConstraint.formatEqualityConstraint
 import ca.uwaterloo.flix.language.fmt.FormatType.formatType
 import ca.uwaterloo.flix.util.{Formatter, Grammar}
+import java.lang.reflect.Method;
 
 /**
   * A common super-type for type errors.
@@ -79,6 +80,31 @@ object TypeError {
          |>> Java method '$methodName' from type '${red(formatType(tpe0, Some(renv)))}' with arguments types (${tpes.mkString(", ")}) not found.
          |
          |${code(loc, s"java method '${methodName} not found")}
+         |""".stripMargin
+    }
+  }
+
+  /**
+   * Java ambiguous method type error.
+   *
+   * @param tpe0    the type of the receiver object.
+   * @param tpes    the types of the arguments.
+   * @param methods a list of possible candidate methods on the type of the receiver object.
+   * @param renv    the rigidity environment.
+   * @param loc     the location where the error occured.
+   */
+  case class AmbiguousMethod(methodName: String, tpe0: Type, tpes: List[Type], methods: List[Method], renv: RigidityEnv, loc: SourceLocation)(implicit flix: Flix) extends TypeError {
+    // TODO INTEROP better comment with candidate methods formatting
+    def summary: String = s"Ambiguous Java method '$methodName' in type '$tpe0' with arguments types (${tpes.mkString(", ")})."
+
+    def message(formatter: Formatter): String = {
+      import formatter._
+      s"""${line(kind, source.name)}
+         |>> Java method '$methodName' from type '${red(formatType(tpe0, Some(renv)))}' with arguments types (${tpes.mkString(", ")}) is ambiguous.
+         | Possible candidate methods:
+         |  ${methods.map(m => m.getName).mkString(", ")}
+         |
+         |${code(loc, s"java method '${methodName} ambiguous")}
          |""".stripMargin
     }
   }
--- a/main/src/ca/uwaterloo/flix/language/errors/TypeError.scala
+++ b/main/src/ca/uwaterloo/flix/language/errors/TypeError.scala
@@ -99,12 +99,15 @@ object TypeError {
 
     def message(formatter: Formatter): String = {
       import formatter._
+      def methodToStr(m: Method) = {
+         s"${m.getName}(${m.getParameterTypes.map(t => t.getName).mkString(", ")})"
+      }
       s"""${line(kind, source.name)}
          |>> Java method '$methodName' from type '${red(formatType(tpe0, Some(renv)))}' with arguments types (${tpes.mkString(", ")}) is ambiguous.
          | Possible candidate methods:
-         |  ${methods.map(m => s"${m.getName}(${m.getParameterTypes.map(t => t.getName).mkString(", ")})").mkString(", ")}
+         |  ${methods.map(m => methodToStr(m)).mkString(", ")}
          |
-         |${code(loc, s"java method '${methodName} ambiguous")}
+         |${code(loc, s"Ambiguous Java method '${methodName}'")}
          |""".stripMargin
     }
   }
--- a/main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
@@ -21,6 +21,7 @@ import ca.uwaterloo.flix.language.errors.TypeError
 import ca.uwaterloo.flix.language.phase.unification.Unification
 import ca.uwaterloo.flix.util.Result
 import ca.uwaterloo.flix.util.collection.{ListMap, ListOps}
+import java.lang.reflect.Method;
 
 object TypeReduction {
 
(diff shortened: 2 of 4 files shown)


magnus-madsen:
We will want some tests in TestTyper class.

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
```diff
@@ -152,24 +153,27 @@ object TypeReduction {
       )
 
     candidateMethods.length match {
+      case 0 => JavaResolutionResult.MethodNotFound
       case 1 =>
         val tpe = Type.Cst(TypeConstructor.JvmMethod(candidateMethods.head), loc)
         JavaResolutionResult.Resolved(tpe)
-      case _ => JavaResolutionResult.MethodNotFound // For now, method not found either if there is an ambiguity or no method found
+      case _ => JavaResolutionResult.AmbiguousMethod(candidateMethods.toList) // For now, method not found either if there is an ambiguity or no method found
```
mlutze, line 160: Looks like this comment is no longer up-to-date.

main/src/ca/uwaterloo/flix/language/errors/TypeError.scala
```diff
@@ -83,6 +84,31 @@ object TypeError {
...
+  case class AmbiguousMethod(methodName: String, tpe0: Type, tpes: List[Type], methods: List[Method], renv: RigidityEnv, loc: SourceLocation)(implicit flix: Flix) extends TypeError {
+    // TODO INTEROP better comment with candidate methods formatting
+    def summary: String = s"Ambiguous Java method '$methodName' in type '$tpe0' with arguments types (${tpes.mkString(", ")})."
+
+    def message(formatter: Formatter): String = {
+      import formatter._
+      s"""${line(kind, source.name)}
+         |>> Java method '$methodName' from type '${red(formatType(tpe0, Some(renv)))}' with arguments types (${tpes.mkString(", ")}) is ambiguous.
+         | Possible candidate methods:
+         |  ${methods.map(m => m.getName).mkString(", ")}
```
mlutze, line 105: Is this just going to print the same name N times?

chanattan: Yes probably, this is temporary as I didn’t have tests for non-static ambiguous methods.

chanattan:
WIP: TypeReduction is still a bit messy, the two added tests pass but the code could be cleaner and the reduction may not be exhaustive.

main/src/ca/uwaterloo/flix/language/errors/TypeError.scala
```diff
@@ -83,6 +84,31 @@ object TypeError {
...
+  case class AmbiguousMethod(methodName: String, tpe0: Type, tpes: List[Type], methods: List[Method], renv: RigidityEnv, loc: SourceLocation)(implicit flix: Flix) extends TypeError {
+    // TODO INTEROP better comment with candidate methods formatting
+    def summary: String = s"Ambiguous Java method '$methodName' in type '$tpe0' with arguments types (${tpes.mkString(", ")})."
+
+    def message(formatter: Formatter): String = {
+      import formatter._
+      s"""${line(kind, source.name)}
+         |>> Java method '$methodName' from type '${red(formatType(tpe0, Some(renv)))}' with arguments types (${tpes.mkString(", ")}) is ambiguous.
+         | Possible candidate methods:
+         |  ${methods.map(m => s"${m.getName}(${m.getParameterTypes.map(t => t.getName).mkString(", ")})").mkString(", ")}
```
mlutze, line 105: This is a lot to do inline. Can we move this into a helper?

main/src/ca/uwaterloo/flix/language/errors/TypeError.scala
```diff
@@ -83,6 +84,31 @@ object TypeError {
...
+         |
+         |${code(loc, s"java method '${methodName} ambiguous")}
```
mlutze, line 107: capital Java

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
```diff
@@ -156,29 +159,64 @@ object TypeReduction {
     val candidateMethods = clazz.getMethods
       .filter(m => m.getName == method)
       .filter(m => m.getParameterCount == ts.length)
-      .filter(m =>
-        (m.getParameterTypes zip ts).foldLeft(true){ case (acc, (clazz, tpe)) => acc && (Type.getFlixType(clazz) == tpe) } // Parameter types correspondance
-      )
+      .filter(m => // Parameter types correspondance with subtyping
+        (m.getParameterTypes zip ts).foldLeft(true) {
```
mlutze, line 163: you can use `forall` instead of a fold.

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
```diff
@@ -156,29 +159,64 @@ object TypeReduction {
...
       case 1 =>
         val tpe = Type.Cst(TypeConstructor.JvmMethod(candidateMethods.head), loc)
         JavaResolutionResult.Resolved(tpe)
-      case _ => JavaResolutionResult.MethodNotFound // For now, method not found either if there is an ambiguity or no method found
+      case _ =>
+        // Among candidate methods if there already exists the method with the exact signature, we should ignore the rest. E.g.: append(String), append(Object) in SB for the call append("a") should already know to use append(String)
+        // TODO: there is surely a better way to repeat the operation, pre-factorization should be possible
+        val exactSignature = candidateMethods
+          .filter(m => // Parameter types correspondance with subtyping
+            (m.getParameterTypes zip ts).foldLeft(true) {
```
mlutze, line 179: idem

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
```diff
@@ -156,29 +159,64 @@ object TypeReduction {
...
+          case _ => JavaResolutionResult.AmbiguousMethod(candidateMethods.toList)
+        }
+    }
+  }
+
+  /**
+   * Helper method to define a sub-typing relation between two given Flix types.
+   * Returns true if tpe2 is a sub-type of type tpe1, false otherwise.
+   */
+  private def isSubtype(tpe1: Type, tpe2: Type)(implicit flix: Flix): Boolean = {
```
mlutze, line 195: The order is backward for my intuition. In Scala, I would expect to read this as "tpe1 is a subtype of tpe2".

mlutze: You should search the codebase for an `isSubtype function. I think we already have one somewhere.

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
```diff
@@ -156,29 +159,64 @@ object TypeReduction {
     val candidateMethods = clazz.getMethods
       .filter(m => m.getName == method)
```
mlutze, line 160: Would it be cleaner to have an `isCandidate` function?

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
```diff
@@ -156,29 +159,64 @@ object TypeReduction {
...
+        val exactSignature = candidateMethods
+          .filter(m => // Parameter types correspondance with subtyping
+            (m.getParameterTypes zip ts).foldLeft(true) {
+              case (acc, (clazz, tpe)) => acc && (Type.getFlixType(clazz) == tpe)
+            })
+        exactSignature.length match {
+          
+          
+          case 1 => JavaResolutionResult.Resolved(Type.Cst(TypeConstructor.JvmMethod(exactSignature.head), loc))
+          case _ => JavaResolutionResult.AmbiguousMethod(candidateMethods.toList)
```
mlutze, line 186: What happens if it is 0?

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
chanattan: I did search a bit and didn’t find anything for Type, I’ll look again later. Magnus expected me to do a new helper method of subtyping.

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
chanattan: Inside the ambiguous case we have at least two candidate methods. If there is one among them that has the exact signature then we pick it. Otherwise there remain the candidate methods that cause ambiguity => ambiguous error.
In particular this case is indeed binary: 0 then it means the user has to do a proper uppercast to select the method he wants, 1 then the exact method exists, 2 or more cannot happen as it would be already prohibited by Java since there would be a fixed ambiguity.

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
chanattan: I’ll do this, thank you.

main/src/ca/uwaterloo/flix/language/errors/TypeError.scala
chanattan: I’ll do this, thank you. Do I do it in as a nested function or outside but in the same file?

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
magnus-madsen: We may have such a function, but I think we will need a new function here anyway.

main/src/ca/uwaterloo/flix/language/errors/TypeError.scala
mlutze: A local function is probably fine. We can move it if we need it elsewhere.

magnus-madsen (approved):

magnus-madsen:
Merged!

PR #7846, opened 2024-06-07
feat: track future slack effect variables

JonathanStarup:
track `isSlack` similarly to `isRegion`. Currently `isSlack` is always `false`


JonathanStarup:
@magnus-madsen you want this merged in or should it just live on that summary branch?

magnus-madsen:
> @magnus-madsen you want this merged in or should it just live on that summary branch?

Not sure yet. Can we defer. @mlutze Thoughts?

JonathanStarup:
for a paper we want to count the variables that are "subtyping slack", so if some rule says that `a <: b` then it becomes `a+s = b` where `s` will be flagged as a slack variable

PR #7847, opened 2024-06-07
wip: just open lambda typing

JonathanStarup:
TODO
- Check `discard` and `let _ = ...` (see `MutSet.iterator`)
- Check other casts


magnus-madsen:
Outdated. We can resurrect once we have set effects.

PR #7850, opened 2024-06-07
feat: remove underlying collection effect form imediate iterator effect

### LINKED ISSUE #7849: Remove `aef` from immediate effect of `Iterator.iterator`
The question is: should building an iterator immediately read the underlying collection (it is mostly reading length fx)

JonathanStarup:
Fixes #7849

#### Pro: (smaller) Effect reflects behaviour
This simplifies the effect footprint of creating an iterator (it can be thought of as returning a function that is pure, from the viewpoint of the underlying data)

#### Con: probably less efficient
This costs performance because we're not allowed to peak inside the collection before actually iterating. Fx length can't be precomputed

#### Effectful Lazy use case
Note that many of the iterators reimplements a lazy value with refs of options

### DIFF
--- a/main/src/library/Array.flix
+++ b/main/src/library/Array.flix
@@ -17,8 +17,8 @@
 instance Iterable[Array[a, r]] {
     type Elm = a
     type Aef = r
-    pub def iterator(rc: Region[r1], a: Array[a, r]): Iterator[a, r + r1, r1] \ (r + r1) =
-        checked_ecast(Array.iterator(rc, a))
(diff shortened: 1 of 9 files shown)


magnus-madsen:
I think this should be discussed at the next workshop.

mlutze:
> I think this should be discussed at the next workshop.

When's the next workshop?

magnus-madsen:
I think we can have one the week you come home.

magnus-madsen:
We can discuss this, but meanwhile the PR has become outdated.

PR #7851, opened 2024-06-09
refactor: add `SharedContext` to `Namer`

magnus-madsen:
This is in preparation for removing the use of `Validation` in `Namer`


main/src/ca/uwaterloo/flix/language/phase/Namer.scala
```diff
@@ -1629,4 +1637,22 @@ object Namer {
...
   }
+
+  /**
+    * Companion object for [[SharedContext]]
+    */
+  private object SharedContext {
+    /**
+      * Returns a fresh shared context.
+      */
+    def mk(): SharedContext = new SharedContext(new ConcurrentLinkedQueue())
```
mlutze, line 1648: We use a queue here, but we're not expecting any popping, right?

mlutze (approved):

main/src/ca/uwaterloo/flix/language/phase/Namer.scala
magnus-madsen: No, but there is no append-only list thing.

PR #7852, opened 2024-06-10
feat: add `Kind.Error`

magnus-madsen:


magnus-madsen:
It is hard to know all the consequences, but I suppose we will learn them over time.

The most likely consequence is simply that in addition to a kind error you may get subsequent unification errors (that are not relevant or confusing.)

main/src/ca/uwaterloo/flix/language/phase/unification/KindUnification.scala
```diff
@@ -48,6 +48,10 @@ object KindUnification {
     case (Kind.CaseSet(sym), Kind.WildCaseSet) =>
       Some(Kind.CaseSet(sym))
 
+    case (Kind.Error, k) => Some(k)
+
```
mlutze, line 52: extra linebreak

mlutze (approved):

main/src/ca/uwaterloo/flix/language/phase/unification/KindUnification.scala
magnus-madsen: It was on purpose. The other cases are similarly separated :)

PR #7857, opened 2024-06-10
chore: defer resolution of `MethodReturnType`

magnus-madsen:


magnus-madsen:
@mlutze Can you approve and merge.

@chanattan This helps, but it does not solve your problem. You still need type annotations for the result of a constructor call.

PR #7861, opened 2024-06-10
feat: add support for escaped Java names

magnus-madsen:
```
def foo(): Bool =
    import java.lang.String.%startsWith%(String): Bool \ {};
    startsWith("Hello World", "Hello")
```


magnus-madsen:
@herluf-ba I could use some help. A challenge here is that I am leftover with the `%` symbols. I could weed them out later, but then I cannot use the same token.

magnus-madsen:
We solved it in a different way.

PR #7868, opened 2024-06-11
refactor: move `Fixity` to `shared`

magnus-madsen:


mlutze (approved):
Are we removing AST entirely?

magnus-madsen:
> Are we removing AST entirely?

We are removing `Ast.scala`.

PR #7870, opened 2024-06-11
feat: add syntax for `InvokeStaticMethod2`

magnus-madsen:


magnus-madsen:
@chanattan This adds (limited) syntax for static method calls, e.g.:

```
import java.lang.{String => JString}

def f(): String = JString¤valueOf(123)
```

This should allow you to progress from `ConstraintGen`.

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -1442,8 +1442,14 @@ object Resolver {
...
         case NamedAst.Expr.InvokeStaticMethod2(clazzName, methodName, exps, loc) =>
           val esVal = traverse(exps)(visitExp(_, env0))
-          mapN(esVal) {
-            es => ResolvedAst.Expr.InvokeStaticMethod2(Class.forName(clazzName.name), methodName, es, loc)
+          flatMapN(esVal) {
+            es => env0.get(clazzName.name) match {
+              case Some(List(Resolution.JavaClass(clazz))) =>
+                Validation.success(ResolvedAst.Expr.InvokeStaticMethod2(clazz, methodName, es, loc))
+              case _ =>
+                val m = ResolutionError.UndefinedJvmClass(clazzName.name, "", loc)
```
mlutze, line 1450: What's the ""? Should there be a TODO here?

mlutze (approved):

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
magnus-madsen: Its a mistake that it even takes a string, but we will deal with that later.

PR #7879, opened 2024-06-12
feat: add `unsafe` expression

magnus-madsen:
This is a shorthand for an unchecked effect cast to pure. 

We will want it for our OO-refactoring, as shown with one example in String.flix

We can discuss later if it should have richer semantics.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
@@ -1599,7 +1599,7 @@ object Weeder2 {
       expect(tree, TreeKind.Expr.Unsafe)
       mapN(pickExpr(tree)) {
         expr =>
-          // We desugar an unsafe block to an unchecked cast to pure.
+          // We desugar an unsafe expression to an unchecked cast to pure.
           val declaredType = None
           val declaredEff = Some(Type.Pure(tree.loc))
           Expr.UncheckedCast(expr, declaredType, declaredEff, tree.loc)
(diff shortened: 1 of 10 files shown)


JonathanStarup:
I know you said later, but it seems likely (when wrapping mutable things in java fx) that would want the cast to some region `r` instead of `{}`. what about `unsafe <effect> { <exp> }`

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1594,6 +1595,17 @@ object Weeder2 {
       }
     }
 
+    private def visitUnsafeExpr(tree: Tree): Validation[Expr, CompilationMessage] = {
```
mlutze, line 1598: Does this belong in Weeder or Desugarer?

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1594,6 +1595,17 @@ object Weeder2 {
...
+      expect(tree, TreeKind.Expr.Unsafe)
+      mapN(pickExpr(tree)) {
+        expr =>
+          // We desugar an unsafe expression to an unchecked cast to pure.
+          val declaredType = None
+          val declaredEff = Some(Type.Pure(tree.loc))
```
mlutze, line 1604: Synthetic location

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
magnus-madsen: Fixed

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
magnus-madsen: Fixed.

magnus-madsen:
> I know you said later, but it seems likely (when wrapping mutable things in java fx) that would want the cast to some region `r` instead of `{}`. what about `unsafe <effect> { <exp> }`

I think these are all good points, but I am just adding it now - as is - so Chanattan can start refactoring the library.

magnus-madsen:
@mlutze RFR

mlutze (approved):

PR #7880, opened 2024-06-12
refactor: support for `InvokeMethod2` in `Files.flix`

### LINKED ISSUE #7869: Refactor `Files.flix` to use `InvokeConstructor2`, `InvokeMethod2`, and so forth
We can start to refactor `Files.flix` to use the new Java constructs.

This will increase our confidence that they are usable.

For example, we can already use `InvokeMethod2` to simplify a function like `copyOver`.

At the very end, we will replace the currency symbol by the chosen syntax and we can merge it.

chanattan:
Should fix #7869.


chanattan:
Oh. It seems some tests are failing indirectly, looks like a parsing issue since it takes in types such as `exists`.
WIP.

magnus-madsen:
> Oh. It seems some tests are failing indirectly, looks like a parsing issue since it takes in types such as `exists`. WIP.

OK, I suggest to undo those specific refactorings until we have an escape mechanism.

magnus-madsen:
No longer relevant (already fixed).

PR #7882, opened 2024-06-13
feat: `Iterator.upcast`

JonathanStarup:
Support upcasting the iterator effect without having to do
```
iter |> Iterator.map(eidentity)
```
which is probably reflected in runtime


magnus-madsen:
I don't think we should hide it further. The right thing should just be to have a lambda there.

JonathanStarup:
The difference was also that the current code uses `map` so its a traversal at runtime, whereas this code just casts the inside function, so no traversal

Issue #7876, opened 2024-06-12
`Adaptor.flix` bug - unsafe use of `unchecked_cast`


JonathanStarup:
## Bug
Example
```
def main(): Unit \ IO = {
    import java.lang.Object.toString(): String \ IO;
    let opt = Some(42i64);
    let javaOpt = Adaptor.toOptional(opt);
    println(toString(checked_cast(javaOpt)))
}
```
Error
```
#
# An unexpected error has been detected by the Flix compiler:
#
#   Inconsistent stackmap frames at branch target 161
...
```
Offending code (https://github.com/flix/flix/blob/master/main/src/library/Adaptor.flix#L159)
```
pub def toOptional(o: Option[a]): ##java.util.Optional \ IO =
    import static java.util.Optional.empty(): ##java.util.Optional \ IO;
    import static java.util.Optional.of(##java.lang.Object): ##java.util.Optional \ IO;
    match o {
        case None    => empty()
        case Some(a) => of(unchecked_cast(a as ##java.lang.Object))
```
## Solution
If you change the unchecked cast to a checked cast, you get the error that you cannot cast a type variable to a type - no further explanation. If think we should change the error to explain that this will crash(!) on primitive types, its not just a missing case in the compiler

JonathanStarup:
@magnus-madsen this also clears things up about boxing - we do not do boxing

mlutze:
Java has `OptionalInt` etc. Maybe an associated type could fix this?

JonathanStarup:
OptionalInt doesnt inherit/implement anything. I don't this function can exist 🤔 At least not in an injective way since both `Int32` and `java.lang.integer` would be mapped to the same type

JonathanStarup:
@magnus-madsen @mlutze actually most things in adapter has this bug. maybe we should do something with boxing and a helper function or something

magnus-madsen:
Not sure what to do. It seems broken.

magnus-madsen:
Typematch?

JonathanStarup:
That would fix it (until we have generics on java types like `ArrayList[t]`)

JonathanStarup:
Another fix could be to add `t1 !~ t2` constraints where either t1 or t2 must be ground, but that sounds more like a paper than a quick fix

magnus-madsen:
I think recent changes to generics have resolved this.

PR #7886, opened 2024-06-14
feat: clarify `Adaptor.toOptional`

JonathanStarup:
Not fixing the bug, just making the bump softer to hit your head on.

Related to #7876

### DIFF
--- a/main/src/library/Adaptor.flix
+++ b/main/src/library/Adaptor.flix
@@ -156,12 +156,24 @@ mod Adaptor {
     ///
     /// Returns the `o` as a new Java `Optional`.
     ///
+    /// OBS: Will crash on primitive types, i.e. integers, booleans, floats, and chars.
+    ///
     pub def toOptional(o: Option[a]): ##java.util.Optional \ IO =
         import static java.util.Optional.empty(): ##java.util.Optional \ IO;
         import static java.util.Optional.of(##java.lang.Object): ##java.util.Optional \ IO;
-        match o {
-            case None    => empty()
-            case Some(a) => of(unchecked_cast(a as ##java.lang.Object))
+        typematch o {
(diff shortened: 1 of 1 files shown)


mlutze:
> OptionalInt doesnt inherit/implement anything. I don't this function can exist 🤔 At least not in an injective way since both `Int32` and `java.lang.integer` would be mapped to the same type if we do boxing.

I think it's OK if it's not injective. We have to specify the type on the way back anyway.

JonathanStarup:
I just think its very error prone, since you manually have to remember it. But I'm also not sure since this is not used anywhere, so its hard to know what the good design is

mlutze:
What do you have to remember?

If we just box the Int in the type match, then we have a total function, right?

JonathanStarup:
If was imagining that i want to use some `Option[t] -> Option[t]` java function, så `Optional -> Optional`.

If I then want to wrap it in a flix option then I can call `toOptional` on my `Int32` just fine but when I call `fromOptional` to `Int32` then I crash. Kind of a niche situation perhaps

JonathanStarup:
I think we could explore checking casts after monomorphization

magnus-madsen:
We could also box?

magnus-madsen:
I guess this is complicated. Lets discuss at the workshop.

JonathanStarup:
In a world with java generics (i.e. `java.lang.Optional[t]`) the typematch+box version of the function can't be typed

mlutze:
> In a world with java generics (i.e. `java.lang.Optional[t]`) the typematch+box version of the function can't be typed

Good point. Let's rediscuss at the next Flix workshop :P

JonathanStarup:
@mlutze could a constraint like `_ !~ Int32` be okay because the rhs is simple and ground? Maybe we could allow `a not primitive` as a constraint

magnus-madsen:
We should discuss more before PR.

PR #7889, opened 2024-06-15
refactor: enable warmup of `SocketServer`

magnus-madsen:
This will enable us to restart the server every night and still have decent performance.


magnus-madsen:
I did some cleanup while I was there.

mlutze (approved):

Issue #6735, opened 2023-11-21
The call to `${line(kind, source.name)}` in every error message should be implicit


magnus-madsen:
Right now we have to write:

```
    def message(formatter: Formatter): String = {
      import formatter._
      s"""${line(kind, source.name)} // THIS LINE SEEMS TO APPEAR EVERYWHERE
         |>> Cannot derive ${magenta(classSym.name)} for empty enum '${red(sym.name)}'.
         |
         |${code(loc, "illegal derivation")}
         |
         |Flix cannot derive any trait instances for empty enumerations.
         |""".stripMargin
    }
```

magnus-madsen:
@jaschdoc Something to look at after Validation cleanup.

PR #7893, opened 2024-06-15
refactor: add `CompilationMessage.messageWithLoc`

jaschdoc:
Closes https://github.com/flix/flix/issues/6735

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/Flix.scala
+++ b/main/src/ca/uwaterloo/flix/api/Flix.scala
@@ -465,9 +465,9 @@ class Flix {
     */
   def mkMessages(errors: Chain[CompilationMessage]): List[String] = {
     if (options.explain)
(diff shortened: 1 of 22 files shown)


magnus-madsen:
@jaschdoc Did you run it? Won't it print the header twice now because all the error messages contain the header?

jaschdoc:
Oops yes haha, forgot to remove it from all `message` definitions!

magnus-madsen:
> Oops yes haha, forgot to remove it from all `message` definitions!

Righto. And perhaps do a few tests to see that its look correct.

magnus-madsen:
It should like now, i.e. header, then one blank line, and then the message beginning with >

jaschdoc:
I went through the diffs and `line` calls and wrote a test. Everything should be fine now.

magnus-madsen (approved):

Issue #7872, opened 2024-06-11
Refactor the Standard Library to always use `import` for Java types


magnus-madsen:
For example:

```
mod Adaptor {

    /// ...
    pub def comparator(_: Proxy[a]): ##java.util.Comparator \ IO with Order[a] =
```

should become

```
mod Adaptor {
    import java.util.Comparator

    /// ...
    pub def comparator(_: Proxy[a]): Comparator \ IO with Order[a] =
```

- Ensure that all imports occur just after the module declaration (i.e. inside it).
- Ensure that imports are alphabetically sorted.

EDIT: Updated example based on discussion.

JonathanStarup:
~Should they not be inside the `mod`? ~

Ah I was just about to complain that we don't enforce it - but they do have to be inside the mod

magnus-madsen:
Hmm. I guess they have to be inside the module.

jaschdoc:
We still can't import in instances :/

magnus-madsen:
> We still can't import in type class instances :/ I can try to fix it but maybe someone already has that assigned to them?

What does @mlutze

jaschdoc:
Use regex `def .* ##`

jaschdoc:
Should be done now

magnus-madsen:
Thanks!

PR #7899, opened 2024-06-17
refactor: update java imports in `Char.flix`

jaschdoc:


jaschdoc:
Should we rename `Character` to `JCharacter` to avoid confusion between `Char` and `Character`?

jaschdoc:
> > Should we rename `Character` to `JCharacter` (or `JChar`) to avoid confusion between `Char` and `Character`?
> 
> What do you think is the best principle? My hunch is that one should only rename if there is a clash.

Seems like a fine principle but I think in this case it could be confusing since it clashes conceptually with `Char`

jaschdoc:
Good point

PR #7901, opened 2024-06-17
refactor: update java imports in `Environment.flix`

jaschdoc:

### DIFF
--- a/main/src/library/Environment.flix
+++ b/main/src/library/Environment.flix
@@ -15,6 +15,7 @@
  */
 mod Environment {
 
+    // TODO: Sort alphabetically
     import java.util.{Map => JMap, Set => JSet, Iterator => JIterator}
     import java.lang.Exception
 
--- a/main/src/library/Environment.flix
+++ b/main/src/library/Environment.flix
@@ -15,9 +15,8 @@
  */
 mod Environment {
 
-    // TODO: Sort alphabetically
-    import java.util.{Map => JMap, Set => JSet, Iterator => JIterator}
-    import java.lang.{Runtime, Object}
+    import java.util.{Iterator => JIterator, Map => JMap, Set => JSet}
+    import java.lang.{Object, Runtime}
 
     ///
     /// Returns the arguments passed to the program via the command line.
--- a/main/src/library/Environment.flix
+++ b/main/src/library/Environment.flix
@@ -17,7 +17,6 @@ mod Environment {
 
     // TODO: Sort alphabetically
     import java.util.{Map => JMap, Set => JSet, Iterator => JIterator}
-    import java.lang.Exception
 
     ///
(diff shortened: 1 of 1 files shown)


main/src/library/Environment.flix
```diff
@@ -28,15 +31,15 @@ mod Environment {
...
+        import static java.lang.System.getenv(): JMap \ rc;
+        import java.util.Map.entrySet(): JSet \ { rc };
+        import java.util.Set.iterator(): JIterator \ { rc };
         let _ = rc; // Avoids redundancy error.
         try {
             let iter = getenv() |> entrySet |> iterator;
             getEnvHelper(iter, Map.empty())
         } catch {
-            case _:##java.lang.Exception => Map.empty()
+            case _: ##java.lang.Exception => Map.empty()
```
jaschdoc, line 42: The try-catch rule just looks up the java class in `Resolver` and does not consider imports. Should I open a ticket, or is this going to be fixed?

jaschdoc: The solution is just to first do a lookup in the imports to get the actual java class and if that fails then fall back to trying to find a class for the literal (which is the current impl). Maybe the other way around?

jaschdoc: It seems this should be fixed in namer or maybe weeder even?

magnus-madsen: @mlutze Any insights?

@jaschdoc If you can fix it, probably in Resolver, you are welcome.

jaschdoc: I'll take a stab at it in a new pr

main/src/library/Environment.flix
```diff
@@ -126,15 +130,15 @@ mod Environment {
...
     def getEnvHelper(iter: ##java.util.Iterator, m: Map[String, String]): Map[String, String] =
         import java.util.Iterator.hasNext(): Bool \ {};
-        import java.util.Iterator.next(): ##java.lang.Object \ {};
-        import java.util.Map$Entry.getKey(): ##java.lang.Object \ {};
-        import java.util.Map$Entry.getValue(): ##java.lang.Object \ {};
+        import java.util.Iterator.next(): Object \ {};
+        import java.util.Map$Entry.getKey(): Object \ {};
+        import java.util.Map$Entry.getValue(): Object \ {};
         if (hasNext(iter))
             let e = unchecked_cast(next(iter) as ##java.util.Map$Entry);
```
jaschdoc, line 143: Same thing applies to `##java.util.Map$Entry`

main/src/library/Environment.flix
mlutze: I would guess that this is parsed specifically as a Java type, so you'll might need changes from parser to resolver.

magnus-madsen (approved):

Issue #7897, opened 2024-06-17
Rework records to use hash instead of dot


magnus-madsen:
- [x] Refactor the compiler to use `#` instead of `.` for `RecordSelect`
  - The key change needs to happen in `Parser2.scala` near line 1298.
  - In addition lines 840-845 in `Weeder2.scala` should be commented and later repurposed for Java fields/methods. Printing the source location here may help identify all the places in the standard library that has to be updated.
- [x] Refactor the entire standard library and tests to use new syntax `#`-based syntax.
- [x] Update `RecordSelect` in `DocAst.scala`
- [x] Update `LabelCompleter.scala` to work correctly with `#`.
- [x] Ensure VSCode formatting looks OK
- [x] Update syntax on https://doc.flix.dev
- [x] Update samples on https://flix.dev
- [x] Update samples on https://play.flix.dev

jaschdoc:
- [ ]  Refactor the compiler to use `#` instead of `.` for `RecordSelect`
    - The key change needs to happen in `Parser2.scala` near line 1298.
    - In addition lines 840-845 in `Weeder2.scala` should be commented and later repurposed for Java fields/methods. Printing the source location here may help identify all the places in the standard library that has to be updated.
- [ ]  Refactor the entire standard library and tests to use new syntax `#`-based syntax.
- [ ]  Update `RecordSelect` in `DocAst.scala`
- [ ]  Update `LabelCompleter.scala` to work correctly with `#`.
- [ ]  Ensure VSCode formatting looks OK
- [ ]  Update syntax on https://doc.flix.dev
- [ ]  Update samples on https://flix.dev
- [ ]  Update samples on https://play.flix.dev

jaschdoc:
What do you mean by `Ensure VSCode formatting looks OK`?

magnus-madsen:
> What do you mean by `Ensure VSCode formatting looks OK`?

Just that syntax highlighting is not broken for some unknown reason.

magnus-madsen:
Note: I am the only one who can update the playground and website, but you can do a cumbersome pass through the documentation. Thanks!

jaschdoc:
I'm still unsure how to fix parsing of function application after record select, e.g.,

```flix
(ide#eshCallStart)(n, d, p)
```
which should just be
```flix
ide#eshCallStart(n, d, p)
```
[...]

PR #7902, opened 2024-06-17
feat: resolve java types against namespace imports in `try-catch`

jaschdoc:
Related to https://github.com/flix/flix/pull/7901


jaschdoc:
@magnus-madsen quick review? Still need to add documentation

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -3235,8 +3255,8 @@ object Resolver {
   /**
     * Returns the class and field reflection objects for the given `className` and `fieldName`.
     */
-  private def lookupJvmField(className: String, fieldName: String, static: Boolean, loc: SourceLocation)(implicit flix: Flix): Result[(Class[_], Field), ResolutionError with Recoverable] = {
-    lookupJvmClass(className, loc).flatMap {
+  private def lookupJvmField(className: String, fieldName: String, static: Boolean, loc: SourceLocation, ns0: Name.NName, root: NamedAst.Root)(implicit flix: Flix): Result[(Class[_], Field), ResolutionError with Recoverable] = {
```
magnus-madsen, line 3258: root should be implicit. Flix and root are always implicit because they are so common.

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -3165,13 +3167,31 @@ object Resolver {
...
     val initialize = false
     Result.Ok(Class.forName(className, initialize, flix.jarLoader))
   } catch {
-    case ex: ClassNotFoundException => Result.Err(ResolutionError.UndefinedJvmClass(className, ex.getMessage, loc))
-    case ex: NoClassDefFoundError => Result.Err(ResolutionError.UndefinedJvmClass(className, ex.getMessage, loc))
+    case ex: ClassNotFoundException => lookUpJvmClassInImport(className, loc, ns0, root)
+    case ex: NoClassDefFoundError => lookUpJvmClassInImport(className, loc, ns0, root)
+  }
+
+  private def lookUpJvmClassInImport(className: String, loc: SourceLocation, ns0: Name.NName, root: NamedAst.Root)(implicit flix: Flix): Result[Class[_], ResolutionError with Recoverable] = {
```
magnus-madsen, line 3179: I am surprised this method is necessary. I would expect something like it to already exist?

magnus-madsen: Actually, instead of doing this, can you just do what InvokeStaticMethod2 does? Thats going to be the future anyway.

jaschdoc: true maybe I just haven't looked around enough :D

magnus-madsen:
@jaschdoc The future plan is that _all Java types must be imported_. So its totally fine if you just do a look up in the environment like in InvokeStaticMethod2.

jaschdoc:
Hmm `instanceof` tests are still parser error... not sure how to expect a non `##` symbol. @herluf-ba do you know what's going on? I can explain more on gitter or here if necessary?

magnus-madsen:
> Hmm `instanceof` tests are still parser error... not sure how to expect a non `##` symbol. @herluf-ba do you know what's going on? I can explain more on gitter or here if necessary?

Can we do the exceptions first and then later the instanceof?

jaschdoc:
I will update tests and std lib in smaller PRs related to https://github.com/flix/flix/issues/7872

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -3177,6 +3177,19 @@ object Resolver {
     case ex: NoClassDefFoundError => Result.Err(ResolutionError.UndefinedJvmClass(className, ex.getMessage, loc))
   }
 
+  /**
+    * Returns the class reflection object for the given `className`.
+    */
+  private def lookupJvmClass2(className: String, env0: ListMap[String, Resolution], loc: SourceLocation)(implicit flix: Flix): Result[Class[_], ResolutionError with Recoverable] = {
+    lookupJvmClass(className, loc) match {
```
magnus-madsen, line 3184: So, in the future this extra lookup should go away... But. that of course requires refactoring more code.

magnus-madsen: The point being that all Java classes and interfaces must be imported.

magnus-madsen (approved):

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
jaschdoc: Yes, but wanted to do this in a soft way so I can progress on the Java import issue before breaking everything :)

PR #7904, opened 2024-06-17
refactor: undo exclusion of tests

magnus-madsen:
We need them for benchmarking.


mlutze (approved):
Maybe there's a better fix somewhere out there

PR #7914, opened 2024-06-20
feat: rework record syntax to `#`

jaschdoc:


magnus-madsen (approved):

magnus-madsen:
@stephentetley This is a breaking change which replaces `.` with `#` for record access (see diff for examples). This is a bit painful I know, but it will allow us to use the dot Java interop.

@chanattan With this merged in, you can start work on using the actual dot for the syntax.

PR #7916, opened 2024-06-20
feat: update `LabelCompleter` with `#` for `RecordSelect`

jaschdoc:
- [x] Perform manual tests in VSCode using updated Parser from https://github.com/flix/flix/pull/7914


magnus-madsen:
This may be enough. You can rebase, build a JAR and try it out.

jaschdoc:
I did that locally with the updated parser but it didn't seem to work :/

magnus-madsen:
> I did that locally with the updated parser but it didn't seem to work :/

Can you paste the program you used? Also double check if # has to be escape in regex world.

magnus-madsen:
Also, in the previous regex, what did [] do?

jaschdoc:
> > I did that locally with the updated parser but it didn't seem to work :/
> 
> Can you paste the program you used? Also double check if # has to be escape in regex world.

The `#` symbol is not used for anything according to the java documentation https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html so I think it's fine to use as a literal. I think `[]` was used to use `.` as a literal since it can mean *any character* if it's not escaped or used in `[]`.
The original regex was ` (.*)[.].*`

jaschdoc:
Here is my test program:
```flix
def main(): Unit \ IO =
    let r = { abcde = { bcdef = { cdefg = 2 } }, x = 3 };
    let s = r.abcde.bcdef.cdefg;
    println(s)
```

magnus-madsen:
> Also the current impl is wrong since it suggests labels that do not exist. It should only suggest `r.abcde.bcdef`.

That is a known limitation, so its an orthogonal issue. 

Perhaps # does not trigger autocomplete? I seem to remember there is something called autocomplete chars somewhere. Maybe in the VSCode repo? But I guess you can trigger it manually and still not see anything?

jaschdoc:
Yeah, hitting `ctrl+space` to load suggestions just gives all functions in the std lib

jaschdoc:
Okay here's a weird bug: I'm using the updated compiler that I built locally where the updated parser is merged but it still runs the above program just fine when using `.` instead of `#`... how?

magnus-madsen:
You need to make a change in Weeder2 too... See the original ticket.

jaschdoc:
Note (to self): There may be something with `triggerCharacters`. Try searching for it on https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion

jaschdoc:
I will do some manual testing once https://github.com/flix/vscode-flix/pull/419 is merged

magnus-madsen:
> I will do some manual testing once [flix/vscode-flix#419](https://github.com/flix/vscode-flix/pull/419) is merged

I pushed a new version of the extension.

jaschdoc:
It works now!

Screenshot from 2024-06-23 15-34-56

magnus-madsen (approved):

PR #7919, opened 2024-06-20
feat: add `Null` subtyping support

chanattan:


chanattan:
Well, it works on my branch, I wonder why it does that.

chanattan:
Ah no, it indeed doesn't work. This is probably related to null values that are not yet fully supported.

magnus-madsen:
Can you give some more information?

chanattan:
There is a mismatched type error from unsubstituted return types. I believe there are some tricky things with null values, but I think I have to add full support for null subtyping for Flix values to see if it works.

PR #7928, opened 2024-06-21
feat: bundle class list with Flix

magnus-madsen:
As it turns out, we cannot reliable use JAVA_HOME __nor__ can we even assume the `classlist` file exists!

It is much better to just ship the 500 lines.


mlutze (approved):
How did you generate the file?

magnus-madsen:
> How did you generate the file?

Blood, sweat and tears

PR #7930, opened 2024-06-22
fix: add error for dot syntax in record selection

jaschdoc:
Updates the weeder with a helpful error message when using dot syntax with record selection.

Related to https://github.com/flix/flix/pull/7916


jaschdoc:
Hmm it seems the `#` symbol clashes with `##` for java types when used in an expression. However, we do not resolve all uses of java types against imports. For instance, a catch-rule works fine with `##java.lang...` since a record selection cannot occur in that position, however, `1ii instanceof ##java.math.BigInt` is a syntax error since it is parsed as record selection.
The way I see it, we have two options:
1. We do not merge this weeder change just yet and allow parsing `record.label`, i.e., the dot syntax for records and simultaneously update the resolver to resolve imports everywhere a java type is used. Hopefully, #7902 completely fixes this. We then merge this change.
3. We merge this change and #7902 in one big PR... I think that becomes too big, though.

In summary, the first option silently allows the old record syntax and then we fix things in the background and finally disallow it, or we do *everything* in one pr.

magnus-madsen:
> In summary, the first option silently allows the old record syntax and then we fix things in the background and finally disallow it, or we do everything in one pr.

Lets do it incrementally.

jaschdoc:
@magnus-madsen compiler works now. We "just" have breaking changes in the community build.

jaschdoc:
Somehow, the `IllegalRecordSelectSyntax` error makes it all the way to `Lowering` which crashes the compiler. Where should this be handled?

magnus-madsen:
> Somehow, the `IllegalRecordSelectSyntax` error makes it all the way to `Lowering` which crashes the compiler. Where should this be handled?
> 
> EDIT: When I try this in VSCode everything works, i.e., compilation of a program with `record.label` fails but it does not crash the compiler. So apparently it only crashes in some tests... weird.

Uh. Hmm. It should not if you return a soft failure.

magnus-madsen:
Where do you see it crashing? I don't see it.

magnus-madsen:
Could you split this PR into two: one with all the fixes to the syntax and one with the actual error. (because the latter PR is kind of blocked until community builds are updated).

magnus-madsen:
What's the "status" on this?

jaschdoc:
This is done. Only fails because of the community build where libraries still use the dot notation for records

PR #7932, opened 2024-06-22
refactor: require imported name in `instanceof`

magnus-madsen:
This PR requires you to import a Java class or interface before using it in an `instanceof`.


mlutze (approved):

jaschdoc (commented):
Thanks!

PR #7948, opened 2024-06-24
style: update broken record select with function application in examples

jaschdoc:
Depends on https://github.com/flix/flix/pull/7978
Related to https://github.com/flix/flix/issues/7897#issuecomment-2186231503


jaschdoc:
Note to self: I think I need a separate `labelName` function instead of parsing a label as a normal name. The reason is that it allows `#` as a qualifier:
```scala
def main(): Unit \ IO =
    let l = List#map(x -> x + 1, 1 :: 2 :: 3 :: Nil);
    println(l)
```

magnus-madsen (approved):

PR #7949, opened 2024-06-24
Added test for structs

dghosef:
This cannot be merged until struct support is complete, but this file adds some testing for structs.

This verifies, for example, that the current implementation of the parser works to some degree at least


magnus-madsen:
> This verifies, for example, that the current implementation of the parser works to some degree at least

Exactly 👍

PR #7953, opened 2024-06-24
refactor: update java imports in `Files.flix`

jaschdoc:


magnus-madsen:
@jaschdoc Conflicted sorry.

I should clarify, since we are getting rid of `import` for constructors, methods, etc. it would actually be okay to leave those as-is, since Chanattan will be deleting them.

magnus-madsen (approved):

Issue #7934, opened 2024-06-23
Structs: Add `struct` declarations to `TypedAst` and all subsequent ASTs


magnus-madsen:
Add a struct declaration to `TypedAst` and all subsequent ASTs (`LoweredAst`, `SimplifiedAst`, and so forth).

I think something like

```scala
case class Struct(doc: Ast.Doc, ann: Ast.Annotations, mod: Ast.Modifiers, sym: Symbol.StructSym, tparams: List[TypeParam], fields: Map[Name.Field, Field], tpe: Type, loc: SourceLocation)
```

Right after `case class Enum`

In addition, all `Root`s should carry a new map (right after enums):

```scala
..., structs: Map[Symbol.StructSym, Struct],...
```

magnus-madsen:
Fixed in master.

PR #7967, opened 2024-06-25
feat: Added structs to the various ASTs

dghosef:
This addresses #7934
This is a big pull request of mostly repetitive work. Placeholder values(like `Map.empty`) are used to silence compile errors when necessary.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/DesugaredAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/DesugaredAst.scala
@@ -21,6 +21,8 @@ import ca.uwaterloo.flix.language.ast.Ast.Denotation
 import ca.uwaterloo.flix.language.ast.shared.Fixity
 import ca.uwaterloo.flix.util.collection.MultiMap
 
+import java.lang.reflect.Field
+
 object DesugaredAst {
 
   val empty: Root = Root(Map.empty, None, MultiMap.empty)
@@ -47,6 +49,8 @@ object DesugaredAst {
     case class Law(doc: Ast.Doc, ann: Ast.Annotations, mod: Ast.Modifiers, ident: Name.Ident, tparams: KindedTypeParams, fparams: List[FormalParam], exp: Expr, tpe: Type, eff: Type, tconstrs: List[TypeConstraint], loc: SourceLocation) extends Declaration
 
     case class Enum(doc: Ast.Doc, ann: Ast.Annotations, mod: Ast.Modifiers, ident: Name.Ident, tparams: TypeParams, derives: Derivations, cases: List[Case], loc: SourceLocation) extends Declaration
+
+    case class Struct(doc: Ast.Doc, ann: Ast.Annotations, mod: Ast.Modifiers, sym: Symbol.StructSym, tparams: List[TypeParam], fields: Map[Name.StructField, Field], tpe: Type, loc: SourceLocation)
 
     case class RestrictableEnum(doc: Ast.Doc, ann: Ast.Annotations, mod: Ast.Modifiers, ident: Name.Ident, index: TypeParam, tparams: TypeParams, derives: Derivations, cases: List[RestrictableCase], loc: SourceLocation) extends Declaration
 
--- a/main/src/ca/uwaterloo/flix/language/ast/DesugaredAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/DesugaredAst.scala
@@ -21,8 +21,6 @@ import ca.uwaterloo.flix.language.ast.Ast.Denotation
 import ca.uwaterloo.flix.language.ast.shared.Fixity
 import ca.uwaterloo.flix.util.collection.MultiMap
 
-import java.lang.reflect.Field
-
 object DesugaredAst {
 
   val empty: Root = Root(Map.empty, None, MultiMap.empty)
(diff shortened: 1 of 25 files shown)


main/src/ca/uwaterloo/flix/language/ast/DesugaredAst.scala
```diff
@@ -21,6 +21,8 @@ import ca.uwaterloo.flix.language.ast.Ast.Denotation
 import ca.uwaterloo.flix.language.ast.shared.Fixity
 import ca.uwaterloo.flix.util.collection.MultiMap
 
+import java.lang.reflect.Field
```
magnus-madsen, line 24: Unused import?

magnus-madsen: Looking at the diff, this seems to have happened in every AST :)

main/src/ca/uwaterloo/flix/language/ast/MonoAst.scala
```diff
@@ -20,15 +20,18 @@ import ca.uwaterloo.flix.language.ast.Ast.{Denotation, EliminatedBy, Source}
...
-  val empty: Root = Root(Map.empty, Map.empty, None, Set.empty, Map.empty)
+  val empty: Root = Root(Map.empty, Map.empty, None, Set.empty, Map.empty, Map.empty)
 
   case class Root(defs: Map[Symbol.DefnSym, Def],
                   effects: Map[Symbol.EffectSym, Effect],
                   entryPoint: Option[Symbol.DefnSym],
                   reachable: Set[Symbol.DefnSym],
-                  sources: Map[Source, SourceLocation])
+                  sources: Map[Source, SourceLocation],
+                  structs: Map[Symbol.StructSym, Struct])
```
magnus-madsen, line 34: In every AST after typed, I think structs should come just after defs-- for consistency.

main/src/ca/uwaterloo/flix/language/ast/DesugaredAst.scala
dghosef: It is not unused because the `fields` field of the `Struct` case has a `Map` of fields whose value type is `Field` as suggested in #7934

main/src/ca/uwaterloo/flix/language/ast/MonoAst.scala
dghosef: Got it. We don't want to preserve the alphabetical ordering then?

main/src/ca/uwaterloo/flix/language/ast/DesugaredAst.scala
magnus-madsen: Java reflection fields? Maybe it should be Name dot Field?

main/src/ca/uwaterloo/flix/language/ast/MonoAst.scala
magnus-madsen: No. That was accidental here.

magnus-madsen: It goes defs, enums, structs, and other stuff

main/src/ca/uwaterloo/flix/language/ast/DesugaredAst.scala
dghosef: `Name.Field` is the key type. I can make a new `StructField` type?

dghosef: This is a map from field names to field values. `Name.Field`(I called it `Name.StructField` but that can be changed) is the key type of the map but not the value type. The value type I think should contain information about the struct field(probably a type and a location). You're right, the java reflection field doesn't make sense. I think I should add a `StructField` case to the AST and use that type.

main/src/ca/uwaterloo/flix/language/ast/KindedAst.scala
```diff
@@ -298,6 +301,8 @@ object KindedAst {
 
   case class Case(sym: Symbol.CaseSym, tpe: Type, sc: Scheme, loc: SourceLocation)
 
+  case class StructField(sym: Symbol.StructFieldSym, tpe: Type, sc: Scheme, loc: SourceLocation)
```
magnus-madsen, line 304: Fields should not have type schemes, but that will probably become clear later :)

main/src/ca/uwaterloo/flix/language/ast/Symbol.scala
```diff
@@ -546,6 +546,34 @@ object Symbol {
     def namespace: List[String] = enumSym.namespace :+ enumSym.name
   }
 
+  /**
+   * Struct Field Symbol.
+   */
+  final class StructFieldSym(val structSym: Symbol.StructSym, val name: String, val loc: SourceLocation) extends Symbol {
```
magnus-madsen, line 552: This symbol may not be necessary. The Name.Field might be enough, but we can discuss later.

magnus-madsen (approved):

magnus-madsen:
@dghosef I added two comments - which we can deal with later. I will merge now, so you can more easily progress 👍

main/src/ca/uwaterloo/flix/language/ast/KindedAst.scala
magnus-madsen: (A struct can be polymorphic, but a field itself cannot be: its type will be determined by the type of the struct).

Issue #7944, opened 2024-06-24
Java: Refactor Standard Library to use `InvokeConstructor2`


magnus-madsen:
We have the right syntax for `InvokeConstructor2` so we can start to refactor the library. See below for an example.

We need to update the following files. I suggest one PR per file:

- [x] Adaptor
- [x] BigDecimal
- [x] BigInt
- [x] File
- [ ] Files
- [x] Float64
- [x] Int16
- [x] Int32
- [x] Int64
- [x] Int8
- [x] Random
- [x] StringBuilder
- [x] CyclicBarrier
- [x] ReetrantLock
- [x] System

For example:

```
pub def toBigDecimal(x: Int32): BigDecimal =
    import java_new java.math.BigDecimal(Int32): BigDecimal \ {} as fromInt32;
    fromInt32(x)
```

can become:

```
import java.math.BigDecimal

pub def toBigDecimal(x: Int32): BigDecimal =
    unsafe new BigDecimal(x)
```

chanattan:
We have to add a parser functionality support to make it work with `|>`, however this seems like higher-order java interoperability stuff that we have not talked about yet.

magnus-madsen:
Ah yes. We will have to discuss.

magnus-madsen:
@chanattan Once the modules are refactored, we should have the old InvokeConstructor throw an exception in Weeder

magnus-madsen:
@chanattan Just two modules to go :)

magnus-madsen:
After those there test cases to fix.

magnus-madsen:
@chanattan Can you get rid of the last usages of `import java_new` -- Thanks!

chanattan:
@magnus-madsen could you check the array PR before I can fix the last java_new please?

magnus-madsen:
Fixed in master.

PR #7969, opened 2024-06-25
refactor: support for `InvokeConstructor2` in `File.flix`

chanattan:
Part of #7944.

### DIFF
--- a/examples/the-ast-typing-problem-with-polymorphic-records.flix
+++ b/examples/the-ast-typing-problem-with-polymorphic-records.flix
@@ -59,8 +59,8 @@ def typeCheck2(e: Expr[r]): Expr[(tpe = Type | r)] = match e {
     case Expr.False  => Expr.False
     case Expr.Cst(i) => Expr.Cst({ +tpe = Type.TInt | { value = i#value | i}})
     case Expr.Add(i) =>
-        let e1 = typeCheck2(i.exp1);
-        let e2 = typeCheck2(i.exp2);
+        let e1 = typeCheck2(i#exp1);
+        let e2 = typeCheck2(i#exp2);
             Expr.Add({ +tpe = Type.TInt | {exp1 = e1, exp2 = e2 | i} })
     case Expr.Ite(i) =>
       let e1 = typeCheck2(i#exp1);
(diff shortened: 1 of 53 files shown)


chanattan:
Three things:
- Removed `File` enum as it had conflict with `java.io.File`, substitued File type with `##java.io.RandomAccess`
- The problem with import order is still present here, as in other parts of the STL all code was inside a module here imports had to be above doc because of outside-module code
- RandomAccessFile was imported both outside the module and inside, it seems like RandomAccessFile is not found in `open` function getting an `Undefined class` otherwise. I suspect there are some issues with import from point 1 too

magnus-madsen:
It seems there are two things in play:

1. A different design which could possibly be better.
2. A refactor with just using the new `new` syntax.

Would it be possible to do just (2) and keep (1) for later?

chanattan:
I tried to keep the File enum but renaming it, the tag wasn't found unfortunately.

I don't think doing (2) before (1) is possible, at least without considering partial refactor.
The conflicts appear when function arguments use the same enum name as a Java class, e.g.:
```scala
def foo(r: File) : Unit \ IO =
    // r has the type of the enum
    let File(r) = f;
    ...
```
So we can't really keep the enum that is used to be later unpacked, unless we go back to strong Java types, i.e., `##java.io.File` and no longer `File` inferred.

magnus-madsen:
I think we should import the Java File as JFile:

```
import java.io.{File => JFile}
```

and work from there.

chanattan:
Thank you. The renaming was used in Random.flix and not here, I probably tried to over-refactor it.

magnus-madsen:
What's the plan here? Import File as JFile and use that everywhere?

chanattan:
> What's the plan here? Import File as JFile and use that everywhere?

That's basically it, the changes are minimal since the renaming removes any conflict with File enum.

magnus-madsen:
> > What's the plan here? Import File as JFile and use that everywhere?
> 
> That's basically it, the changes are minimal since the renaming removes any conflict with File enum.

Just replacing all the import java_new

chanattan:
Yes and updating the instantiation. There is only one java_new in the file.

magnus-madsen (approved):

PR #7982, opened 2024-06-25
feat: Parse indexing

JonathanStarup:


main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1029,6 +1031,28 @@ object Weeder2 {
...
+    }
+
+    private def visitIndexWriteExpr(tree: Tree): Validation[Expr, CompilationMessage] = {
+      expect(tree, TreeKind.Expr.IndexWrite)
+      pickAll(TreeKind.Expr.Expr, tree) match {
+        case e1 :: e2 :: e3 :: Nil => mapN(visitExpr(e1), visitExpr(e2), visitExpr(e3)){
+          case (exp1, exp2, exp3) =>
+            Expr.Apply(Expr.Ambiguous(Name.mkQName("Writable.put", exp1.loc), exp1.loc), List(exp1, exp2, exp3), tree.loc)
+        }
+        case _ => ???
```
JonathanStarup, line 1052: I am not sure what to write here, it seems unreachable in some sense?

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1029,6 +1031,28 @@ object Weeder2 {
       }
     }
 
+    private def visitIndexExpr(tree: Tree): Validation[Expr, CompilationMessage] = {
```
magnus-madsen, line 1034: The traits should be called Index and IndexMut with `get` and `put` methods.

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
magnus-madsen: If it cannot happen then internal compiler exception. If it can happen then Expr.Error.

You can experiment.

What happens if you write `e[]` or `e[`etc. Building a jar and typing one char at a time should make it clear.

PR #7988, opened 2024-06-26
feat: add `--Xdeprecated` flag

magnus-madsen:


magnus-madsen:
@mlutze or @JonathanStarup if you could approve. @chanattan will need this feature to gradually disable/deprecate old import syntax.

magnus-madsen:
This flag -- as a feature in itself -- is not intended to stay. It is just for a transition period.

mlutze:
> This flag -- as a feature in itself -- is not intended to stay. It is just for a transition period.

Though maybe it should? (Flix Workshop)

mlutze (approved):

PR #7997, opened 2024-06-26
refactor: unsupport `InvokeConstructor` call

chanattan:
Part of #7944.


chanattan:
Something like this? @magnus-madsen

Issue #8000, opened 2024-06-27
Structs: Add structs expressions to `Lexer`, `Parser`, and `Weeder`


magnus-madsen:
We should add expressions for working with structs.

Here is approximately what we want in `WeededAst`:

```scala
case class StructNew(qname: Name.QName, exps: List[(Name.Ident, Expr)], exp: Expr, loc: SourceLocation) extends Expr
case class StructGetField(exp: Expr, field: Name.Ident, loc: SourceLocation) extends Expr
case class StructPutField(exp1: Expr, field: Name.Ident, exp2: Expr, loc: SourceLocation) extends Expr
```

(Maybe `Name.Ident` should be `Name.StructField`?)

I would put these right before `ArrayNew`.

Ultimately, we will want syntax like:

```
let s = new A.B.C.Node { prev = exp1, next = exp2 } @ exp // the last exp is the region.
s.prev.next = s.next.next
```

But the above syntax conflicts with our current syntax, so until we resolve that, I propose the following __temporary__ syntax: 

```
let s = struct new A.B.C.Node { prev = exp1, next = exp2 } @ exp // the last exp is the region.
s€prev€next = s€next€next
```

If you have better suggestions, feel free to propose them :)

dghosef:
Is `exp` in `StructNew` for the region?

dghosef:
Also should `qname` be instead of type `Type` since there may be type variables(I think - I might be wrong about this)

For example: 
let s = new A.B.C.Node[Int32] { prev = exp1, next = exp2 } @ exp

magnus-madsen:
> Is `exp` in `StructNew` for the region?

Yes

magnus-madsen:
> Also should `qname` be instead of type `Type` since there may be type variables(I think - I might be wrong about this)
> 
> For example: let s = new A.B.C.Node[Int32] { prev = exp1, next = exp2 } @ exp

No, the type variables would be inferred. I guess there is a question of the name being imported (via `use`). 
But I think we can start with just a QName.

JonathanStarup:
@dghosef this is outdated right?

dghosef:
Yea. Thanks

PR #8003, opened 2024-06-27
feat: Add support for expressions of the form `new Struct {a = expr, b = expr ...} @ region` to the parser and weeder

dghosef:
Addresses #8000

Tests will come in a new draft PR


main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -2468,8 +2468,24 @@ object Parser2 {
       expect(TokenKind.KeywordNew, SyntacticContext.Expr.OtherExpr)
       Type.ttype()
 
-      // NewObject or InvokeConstructor?
-      if (at(TokenKind.CurlyL)) {
+      // NewStruct, NewObject, or InvokeConstructor?
+      if (at(TokenKind.CurlyL) && (nth(1) == TokenKind.NameLowerCase || (nth(1) == TokenKind.CurlyR && nth(2) == TokenKind.At))) {
```
magnus-madsen, line 2472: I am surprised this works. Will it not shadow the next case, so to speak?

dghosef: The cases are
```
new StructName { } @ r // case 1
new StructName {field1 = val, ...} @ r // case 2
new Type { } // case 3
new Type { def fn1 ... } // case 4
new Type( ... ) // case 5
```
My cases don't shadow the others because
- case 1 checks for the `@` in the empty struct case to avoid conflict with case 3
- case 2 checks for a lowercase name to avoid with case 4(which must have a `def` instead
- case 1 and 2 check for `{` to prevent conflict with case 5

magnus-madsen: OK- that helps!

magnus-madsen (approved):

PR #8004, opened 2024-06-27
refactor: add `InvokeConstructor` deprecated

### LINKED ISSUE #7989: Java: Deprecate InvokeConstructor
Once the refactor of the standard library and tests, we should disable the old `InvokeConstructor` by doing something like:

```
if (!options.xdeprecated) {
  val m = WeederError.Deprecated(...)
  Validation.toSoftFailure(...)
} else {
  // as before
}
```

In `Weeder` at the relevant expression.

chanattan:
Fixes #7989.


magnus-madsen:
@chanattan The Boolean was correct. But you need to rewrite the test cases to get rid of the old invoke constructor :)

magnus-madsen:
Not relevant anymore.

PR #8005, opened 2024-06-27
feat: add support for arrays with subtyping

chanattan:
@mlutze

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolver.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolver.scala
@@ -277,7 +277,7 @@ object ConstraintSolver {
       // Apply substitution now
       val tpes = tpes0.map(subst0.apply)
       // Ensure that simplification for constructor parameters is done
-      val allKnown = tpes.forall(isKnown)
+      val allKnown = tpes.forall(t => isKnown(t) && t.typeVars.forall(isKnown))
 
       if (allKnown) {
         TypeReduction.lookupConstructor(clazz, tpes, cvar.loc) match {
@@ -296,7 +296,7 @@ object ConstraintSolver {
       val tpe = subst0(tpe0)
       val tpes = tpes0.map(subst0.apply)
       // Ensure that simplification for method parameters is done
-      val allKnown = isKnown(tpe) && tpes.forall(isKnown)
+      val allKnown = isKnown(tpe) && tpes.forall(t => isKnown(t) && t.typeVars.forall(isKnown))
 
       if (allKnown) {
         TypeReduction.lookupMethod(tpe, methodName.name, tpes, mvar.loc) match {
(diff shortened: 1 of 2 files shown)


main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolver.scala
```diff
@@ -277,7 +277,7 @@ object ConstraintSolver {
       // Apply substitution now
       val tpes = tpes0.map(subst0.apply)
       // Ensure that simplification for constructor parameters is done
-      val allKnown = tpes.forall(isKnown)
+      val allKnown = tpes.forall(t => isKnown(t) && t.typeVars.forall(isKnown))
```
mlutze, line 280: What's the contract of the `isKnown` function? It seems like we ought to change it so it implements essentially what you have here.

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
```diff
@@ -265,6 +267,10 @@ object TypeReduction {
       case (Type.Cst(TypeConstructor.Native(clazz), _), Type.Cst(TypeConstructor.Str, _)) => clazz.isAssignableFrom(classOf[String])
       case (Type.Cst(TypeConstructor.Native(clazz), _), Type.Cst(TypeConstructor.BigInt, _)) => clazz.isAssignableFrom(classOf[BigInteger])
       case (Type.Cst(TypeConstructor.Native(clazz), _), Type.Cst(TypeConstructor.BigDecimal, _)) => clazz.isAssignableFrom(classOf[java.math.BigDecimal])
+      case (Type.Cst(TypeConstructor.Array, _), Type.Cst(TypeConstructor.Array, _)) =>
+        val (elmType2, elmType1) = (tpe2.typeVars.head, tpe1.typeVars.head)
```
mlutze, line 271: Why do you swap 2 and 1 here?

We don't want `typeVars` here, but rather `typeArguments`.

Can we just unpack the type arguments instead? Like `List(elmType1, rc1) = tpe1.typeArgs` That way we avoid the `.head.tail` stuff

chanattan: I am checking if the type of array1 (tpe1) is a subtype of array2 (tpe2), I just did the tuple like this since this is what was used in the match case. I initially had in mind something like: is tpe2 a supertype of tpe1 that is why, later refactoring can be envisioned.

Ah I see, thank you I will change that.

Thank you for the tip, I am not so used with Scala indeed this seems better.

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolver.scala
chanattan: isKnown is a higher-level constraint that allows variables to be substituted as long as the constraint solver is running, so that we can go back later, stronger, to this solving. Since we apply substitution beforehand, isKnown was returning false if the type remained a variable.

magnus-madsen:
@chanattan @mlutze What's the status on this?

chanattan:
> @chanattan @mlutze What's the status on this?

I’m off this week but I am working on it. Then subtyping tests, then refactoring STL, then dot parsing. Everything is supposed to be done by next week.

magnus-madsen (approved):

PR #8007, opened 2024-06-27
feat: Add checking that the last type parameter in a `struct` declaration is of kind `Effect`

dghosef:
The reasoning for this is that the last type parameter of a struct is always its region. Thus, it must have the effect kind


magnus-madsen (approved):

magnus-madsen:
Looks good 🚀

PR #8008, opened 2024-06-27
ci: run IDE end-to-end tests in CI

sockmaster27:


magnus-madsen:
What does this do?

mlutze:
Nice work, I've been wanting this for a while.

The logs show a bunch of errors even though the tests pass. Are the errors benign?

sockmaster27:
It seems to be very stable. The only thing  that I can think of is that since it's running on VS Code 'stable', an update could theoretically cause failures, but I think that's very unlikely. If we really wanted to we could pin it and update it manually.

sockmaster27:
> The logs show a bunch of errors even though the tests pass. Are the errors benign?

Yes. You can see something similar in the rust-analyzer repo: https://github.com/rust-lang/rust-analyzer/actions/runs/9578105353/job/26407648010#step:9:1

magnus-madsen:
We should probably merge in the huge test pr before this pr. We can discuss more on Monday.

magnus-madsen (approved):

PR #8013, opened 2024-06-28
feat: add parser tests for struct expressions

dghosef:
This adds parser tests for creating structs, modifying struct fields, and reading struct fields. Heavily based on their `Record` counterparts. 

Marked as draft because can't push until struct support is implemented


magnus-madsen:
This is still relevant right? We are just waiting for more compiler support?

dghosef:
Yes

PR #8014, opened 2024-06-28
chore: use consistent icons in HTML documentation

sockmaster27:
This replaces the hacky unicode character icons with the equivalent from Google's Material Symbols icon set. This makes the generated pages look subtly more professional, and ensures consistency between devices.

Before:
image

After:
image

Note that the light/dark-mode icons have now been swapped, as this seems to be more conventional.
This PR also introduces use of the `:has()` CSS selector, which has recently gained [wider support](https://caniuse.com/css-has), which happens to eliminate the need for some of the JS.


magnus-madsen:
Thanks-- this is more professional indeed. A few questions:

1. Do we need to have a license file somewhere or how does it work?
2. Do the icons work well on mobile too?

sockmaster27:
> 1. Do we need to have a license file somewhere or how does it work?

It's Apache 2.0, but since it's kind of a font, I'm not entirely sure. For the generated website, it shouldn't be necessary, but we might want to include it for the compiler itself to be safe. Do we have some established way of doing this beyond just including it in the source code?

> 2. Do the icons work well on mobile too?

Yes.

magnus-madsen:
How about putting the license file in the icons folder?

sockmaster27:
Can this be merged or should it be discussed further?

magnus-madsen:
We can merge this, but make the following changes:

- Drop LICENSE
- Add LICENSE.md with the following text:

The following files in this directory:

- foo
- bar

are part of the XYZ project and licensed under the Apache 2.0 license. See:

url to project

magnus-madsen (approved):

PR #8017, opened 2024-06-29
fix: downgrade antlr to 4.11.1

### LINKED ISSUE #8016: ANTLR Tool version 4.11.1 used for code generation does not match the current runtime version 4.13.1
We have to downgrade until the toml lib is upgraded.

magnus-madsen:
Fixes #8016


mlutze (approved):
Why did tests not catch this?

magnus-madsen:
> Why did tests not catch this?

because it just prints a warning with `System.out` which is super annoying.

PR #8018, opened 2024-06-29
feat: improve Java class list

magnus-madsen:
The file is only 100KB which is probably OK.
Its extracted from openjdk so its the ground truth.
It may include private classes, conversely it may omit inner classes.

Practical experience shows its much better. Probably this will do the job 99% and getting the last 1% will be difficult.


main/src/ca/uwaterloo/flix/util/ClassList.scala
```diff
@@ -23,477 +23,14 @@ object ClassList {
   /**
     * The class list for Java 21.
     *
-    * Excludes: (a) all special classes (e.g. lambdas), and (b) all internal classes (e.g. sun classes).
+    * Computed as follows:
+    *
+    * {{{
+    * $ git clone git@github.com:openjdk/jdk.git
```
mlutze, line 29: Thanks

mlutze (approved):

PR #8023, opened 2024-07-01
feat: Add name resolution for struct expressions

dghosef:
The only significant transformation is with expressions of the form
```
new StructName { } @ r
```

I don't think anything can be done to the get and put struct expressions since with those the struct name is not listed


magnus-madsen (approved):

magnus-madsen:
Looks good. Resolver will be more involved, but not too much :)

PR #8028, opened 2024-07-02
feat: added struct type to monotype enum

dghosef:


JonathanStarup:
I think -- similar to the tuple type -- that the struct should store its arity so it is easier to assemble later

magnus-madsen (approved):

magnus-madsen:
> I think -- similar to the tuple type -- that the struct should store its arity so it is easier to assemble later

Does the arity matter? We also need the types right?

PR #8039, opened 2024-07-03
feat: Make traits for struct fields

dghosef:
The easiest way to do type inference for structs is to use `traits` and `instances` under the hood. For example
```
struct S [r] {
    a: Int32
}

```
will generate something like this:
```
trait Dot_a[st] {
    type FieldType[st]: Type
    type Aef[st]: Eff
    pub def get(s: st): Dot_a.FieldType[st] \ Dot_a.Aef[st]
    pub def put(s: st, v: Dot_a.FieldType[st]): Unit \ Dot_a.Aef[st]
}
instance Dot_a[S[r]] {
    type FieldType = Int32
    type Aef = r
    pub def get(st: S[r]): Int32 \ r = {TBD}
    pub def put(st: S[r], v: Int32): Unit \ r = {TBD}
}
```

This PR does the `trait` part of this process.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
@@ -228,13 +228,14 @@ object Weeder2 {
 
 
       // STRUCT TODO: Give this a name that cannot be accidentally overshadowed by the user?
-      val ident = Name.Ident("Dot" ++ fieldName, loc)
+      val traitName = "Dot_" ++ fieldName
+      val ident = Name.Ident(traitName, loc)
       val tparam = WeededAst.TypeParam.Kinded(Name.Ident("st_typ", loc), Kind.Ambiguous(Name.mkQName("Type", loc), loc))
       val fieldAssocType = WeededAst.Declaration.AssocTypeSig(doc, mod, Name.Ident("FieldType", loc), TypeParam.Unkinded(Name.Ident("st_typ", loc)), Kind.Ambiguous(Name.mkQName("Type"), loc), None, loc)
       val fieldAssocEff = WeededAst.Declaration.AssocTypeSig(doc, mod, Name.Ident("Eff", loc), TypeParam.Unkinded(Name.Ident("st_typ", loc)), Kind.Ambiguous(Name.mkQName("Eff"), loc), None, loc)
       val tvar = WeededAst.Type.Var(Name.Ident("st_typ", loc), loc)
-      val fieldType = Type.Apply(Type.Ambiguous(Name.mkQName("DotField.FieldType", loc), loc), tvar, loc)
-      val effType = Type.Apply(Type.Ambiguous(Name.mkQName("DotField.Aef", loc), loc), tvar, loc)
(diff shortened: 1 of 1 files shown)


dghosef:
The biggest potential downside to this approach is that it would require some work to make the error messages pretty. However, this would be in keeping with the philosophy of doing the simplest thing first.

magnus-madsen:
Very cool that this idea seems to work! I would like to explore this direction further -- I have many comments and questions, but probably best discussed online. 

Meanwhile, I think we should also push on a direct implementation similar to the one we use for Java interop. I would probably be most comfortable that we merge such a solution in first and then only later refactor one based on traits.

magnus-madsen:
A list of things to think about:

- Does this work correctly with structs with multiple parameters?
- Does `Dot` need to be split into two: one for get and one for put?
- How would we add field visibility on top of this?
- Probably the instances should be generated in the `Deriver`
- How does e.g. Rust deal with this same problem?
- What are the error messages like?

dghosef:
> How does e.g. Rust deal with this same problem?
Rust requires type annotations so it doesn't have this problem I think. For example:
```Rust
struct Person {
    name: String,
    age: u8,
}

fn main() {
    |x| x.age;
    ()
}
```
gives the following error:
```
error[E0282]: type annotations needed
 --> src/main.rs:8:6
  |
8 |     |x| x;
  |      ^
  |
help: consider giving this closure parameter an explicit type
  |
8 |     |x: /* Type */| x;
  |       ++++++++++++

For more information about this error, try `rustc --explain E0282`.
error: could not compile `playground` (bin "playground") due to 1 previous error
```

dghosef:
Change of plans from the original comment. This PR compiles:
```
Struct[a, r] {
    x: a
}
```
to
```
trait DotGet_x[s] {
    type FieldType: Star
    type Aef: Eff
    def get(struct: s): DotGet_x.FieldType[s] \ DotGet_x.Aef[s]
}

trait DotPut_x[s] {
    type FieldType: Star
    type Aef: Eff
    def get(struct: s, value: DotPut_x.FieldType[s]): Unit \ DotPut_x.Aef[s]
}

instance DotGet_x[Struct[a, r]] {
    type FieldType = a,
    type Aef = r,
    def get(struct: Struct[a, r]): a \ r = $StructGet(struct, x)$
}

instance DotPut_x[Struct[a, r]] {
    type FieldType = a,
    type Aef = r,
    def put(struct: Struct[a, r], v: a): Unit \ r = $StructPut(struct, x, v)$
}
```

It currently fails because of issue #8364. It also probably shouldn't be merged until the `Ref` and `MutDeque` migration to structs is complete so that we can be confident of its correctness

dghosef:
I manually tested with `MutDeque` and `Ref` and nothing seems to break, which is a good sign.

If/when this gets merged the next thing would probably be to work on error messages

magnus-madsen:
!!!jar

magnus-madsen:
!!!jar

dghosef:
!!jar

dghosef:
!!!jar

dghosef:
!!!jar

dghosef:
!!!jar

dghosef:
!!!jar

dghosef:
!!!jar

magnus-madsen:
We got the paper and artifact out. Actually merging this will have to wait for the future :)

PR #8046, opened 2024-07-08
feat: incremental compilation with `changeset`

JonathanStarup:


magnus-madsen:
Unfortunately this is not so simply because of dependencies.

JonathanStarup:
What kind of dependencies? the only thing I could think of type class/duplicate things that would only get one source location

I think we should fix it though, its the only thing that makes the vscode plugin usable (not crashing and killing performance) for me with a large project

magnus-madsen:
This requires a lot more infrastructure. Right now we are focused on resilience.

PR #8050, opened 2024-07-13
feat: add search to HTML documentation

sockmaster27:
Builds on #8014


sockmaster27:
New layout on
- Desktop:
image
- Mobile:
image
image

sockmaster27:
Search view on

- Desktop:
image

- Mobile:
image

magnus-madsen:
Nice. Lets look at it together at the next meeting.

Do you know how much the page size increase by? And does it degrade gracefully without JavaScript?

sockmaster27:
For Prelude, the page itself only goes from ~160kB to ~170kB, but because each page needs to load all the other ones it ends up loading ~7MB. That happens asynchronously though, so it doesn't impact the load-speed, and it isn't downloaded when JS is disabled.

Hopefully the browser's cache will take care of this when navigating between pages, but I believe that would depend on the settings of whatever server they are deployed on. Alternatively, we could manually cache the HTML files using the localstorage API or something like that, if we thought it would be worth it.

We could also defer the loading until the search box is actually clicked, but I suspect that would lead to a horrible experience on slow networks.

magnus-madsen:
I think we should have a discussion about using something like http://elasticlunr.com/ with a precomputed index.

magnus-madsen:
Using a library gives a lot of upside: Minimal maintenance burden, excellent search quality, and compact index.

The only question is how to precompute the index.

sockmaster27:
I worry about over-complicating it, as it hurts the compiler. I would suggest trying out the simple solution first, and then looking at what actually needs improving before making it into a major project.

PR #8053, opened 2024-07-15
fix: system newlines

JonathanStarup:
I'm not sure about all of the `\n` instances


magnus-madsen:
We should discuss in greater detail: some of the changes (e.g. output) is fine, but making the input depend on the line separator would make the compiler dependent on the platform, right?

magnus-madsen:
In other words, I think we should _only_ use lineSeparator for output-- if at all.

magnus-madsen:
Outdated.

PR #8054, opened 2024-07-17
feat: Working structs

dghosef:
This PR should probably be split into multiple at some point. It includes:`struct` declarations and expressions(new, field get, field put) in all compiler phases.

The flaw that makes structs not quite practically usable is that currently struct get/put expressions assume the struct name is the same as the module name


JonathanStarup:
If you want feedback on anything just let me know :)

magnus-madsen:
!!!jar

magnus-madsen:
Very nice! I played a bit with the implementation.

I suggest we discuss next steps on Monday. 

We have some design decisions to make and then we should also discuss how to break the PR into smaller parts.

dghosef:
Everything from here is now in `master` or is a separate PR in progress

PR #8055, opened 2024-07-19
feat: HashSet

JonathanStarup:
@magnus-madsen we could remove all order-dependent functions to make it more "safe", e.g. only fold, and requiring Order to call toList fx. This would also make it commutative since the ordering is unobservable

Not completely finished, some tricky functions missing - but all present code is finished


magnus-madsen:
> @magnus-madsen we could remove all order-dependent functions to make it more "safe", e.g. only fold, and requiring Order to call toList fx. This would also make it commutative since the ordering is unobservable

I think that would make sense. Note that we have UnorderedFoldable.

What hash table implementation/algorithm are you using?

JonathanStarup:
Just the trait Hash. If you mean for the structure, its just a RBT with hash keys, using a list-set as the last resort

I think that we should generalize it one step though, the set could itself carry the `Hash.hash` function, so you could have any set of derived keys that might collide, instead of tying it to hashing

main/src/library/HashSet.flix
```diff
@@ -0,0 +1,546 @@
...
+    pub def toString(x: HashSet[t]): String = HashSet.toString(x)
+}
+
+instance Collectable[HashSet[a]] with Eq[a], Hash[a] {
+    type Elm = a
+    pub def collect(iter: Iterator[a, ef, r]): HashSet[a] \ { ef, r } =
+        HashSet.collectI(iter)
+}
+
+instance Hash[HashSet[t]] with Hash[t] {
```
mlutze, line 37: Can we just derive this?

main/src/library/HashSet.flix
```diff
@@ -0,0 +1,546 @@
...
+    use ListSet.lift
+
+    /// Invariant: the list has no duplicates
+    pub enum ListSet[t](List[t])
+
+    instance Eq[ListSet[t]] with Eq[t] {
+        pub def eq(x: ListSet[t], y: ListSet[t]): Bool = ListSet.eq(x, y)
+    }
+
+    mod ListSet {
```
mlutze, line 85: Should ListSet be its own thing?

main/src/library/HashSet.flix
```diff
@@ -0,0 +1,546 @@
...
+        pub def empty(): ListSet[t] = ListSet(List.empty())
+
+        pub def singleton(x: t): ListSet[t] = ListSet(List#{x})
+
+        pub def isEmpty(s: ListSet[t]): Bool = match list(s) {
+            case Nil => true
+            case _ => false
+        }
+
+        pub def list(s: ListSet[t]): List[t] = {
```
mlutze, line 95: toList?

main/src/library/HashSet.flix
```diff
@@ -0,0 +1,546 @@
...
+    pub def toString(s: HashSet[a]): String with ToString[a] = region local {
+        let inner = iterator(local, s) |> Iterator.join(", ");
+        "HashSet(${inner})"
+    }
+
+
+    ///
+    /// Returns the empty set.
+    ///
+    /// `HashSet#{}` is syntactic sugar for `empty` i.e. `HashSet#{} == empty()`.
```
mlutze, line 177: Is this true?

main/src/library/HashSet.flix
JonathanStarup: oh maybe, didn't know that hash is derivable

main/src/library/HashSet.flix
JonathanStarup: It could, but its not very useful outside of this i'd think

HashMap needs ListMap too

main/src/library/HashSet.flix
JonathanStarup: Search replace 🤡

magnus-madsen:
I suggest package first.

Issue #8019, opened 2024-06-29
Refactor `Stratifier`


magnus-madsen:
- [ ] Introduce `SharedContext` for errors
- [ ] Remove usage of `Validation`

PR #8057, opened 2024-07-22
refactor: add `SharedContext` to `Stratifier`

jaschdoc:
- [x] Add new separate `SharedContext[T]`
- [x] Introduce in `Stratifier`
- [x] Refactor error cases to add to `SharedContext`
- [x] Update those functions to **not** return `Validation`

Related to https://github.com/flix/flix/issues/8019


magnus-madsen:
We should not have a generic `SharedContext` -- the shared part of the name refers to it being shared across threads, not across phases, but I see how that can be confusing :)

Instead, we should have `SharedContext` be declared inside `Stratifier` itself, privately and at the very bottom.

jaschdoc:
Ahh I only did it to reduce code duplication. The name makes sense hehe :)

main/src/ca/uwaterloo/flix/language/phase/Stratifier.scala
```diff
@@ -86,13 +91,13 @@ object Stratifier {
...
     */
-  private def visitInstance(i0: TypedAst.Instance)(implicit root: Root, g: LabelledPrecedenceGraph, flix: Flix): Validation[TypedAst.Instance, StratificationError] =
+  private def visitInstance(i0: TypedAst.Instance)(implicit root: Root, g: LabelledPrecedenceGraph, flix: Flix, sctx: SharedContext): Validation[TypedAst.Instance, StratificationError] =
     mapN(traverse(i0.defs)(d => visitDef(d)))(ds => i0.copy(defs = ds))
 
   /**
     * Performs stratification of the given definition `def0`.
     */
-  private def visitDef(def0: Def)(implicit root: Root, g: LabelledPrecedenceGraph, flix: Flix): Validation[Def, StratificationError] =
+  private def visitDef(def0: Def)(implicit root: Root, g: LabelledPrecedenceGraph, flix: Flix, sctx: SharedContext): Validation[Def, StratificationError] =
```
magnus-madsen, line 100: Sorry, one more nitty details. We always try to pass implicits in this order:

SpecificThing, LocalContext, SharedContext, root, flix

magnus-madsen: It goes from most specific to least specific

jaschdoc: Roger

jaschdoc:
Yes :)

magnus-madsen (approved):

PR #8098, opened 2024-07-29
feat: refactor import in `Thread.flix`

magnus-madsen:


mlutze (approved):
Weird that we're using an intrinsic here

magnus-madsen:
> Weird that we're using an intrinsic here

Its something with lib min.

PR #8100, opened 2024-07-29
fix: allow `Unit` argument for `InvokeStaticMethod`

magnus-madsen:


mlutze (commented):
I don't see this as a Resolver problem. Isn't this a parsing/weeding issue?

magnus-madsen:
> I don't see this as a Resolver problem. Isn't this a parsing/weeding issue?

I agree its a mess, but I cannot know what to do before Resolver.

magnus-madsen:
We will have to rework the solution later anyway, but for now this is sensible: When we discover its a InvokeStaticMethod we also patch the Unit argument. Later can decide whether to do this elsewhere.

mlutze (approved):

PR #8102, opened 2024-07-29
feat: Complete kinder support for structs

dghosef:
This PR
- fixes a small typo in namer
- constructs a scheme for each struct
- Simplifies the kinder struct implementation

All changes have been manually tested


magnus-madsen:
@mlutze Can you do an initial review?

mlutze (commented):
The high-level problem I see here is that I don't think this is the right place for these errors. Missing fields, etc. does not seem to be related to kinding, so these checks ought to be in the Resolver.

magnus-madsen:
Right-- thats an issue.

Also, we should not use HardErrors any more.

dghosef:
I see. In the case that we get one of these errors, would I just return an `Error` ast node?

mlutze:
> I see. In the case that we get one of these errors, would I just return an `Error` ast node?

Correct

mlutze:
Note also that while manual testing is a good start, we should have at the _very least_ one proper test for each error.

magnus-madsen:
At least 3 tests for each user facing error :)

dghosef:
> At least 3 tests for each user facing error :)

Oh right, I had written tests for the errors(the manual testing referred to the functionality which can't be tested right now without breaking the build) but forgot to push.

Anyway, I'll remove these errors from the kinder and put them elsewhere and then move my tests there and make a new PR for those

dghosef:
Ok I moved the errors to another PR that will come tomorrow. The scope of this one is now smaller

main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
```diff
@@ -114,35 +114,48 @@ object Kinder {
    */
   private def visitStruct(struct0: ResolvedAst.Declaration.Struct, taenv: Map[Symbol.TypeAliasSym, KindedAst.TypeAlias], root: ResolvedAst.Root)(implicit flix: Flix): Validation[KindedAst.Struct, KindError] = struct0 match {
     case ResolvedAst.Declaration.Struct(doc, ann, mod, sym, tparams0, fields0, loc) =>
-      // if not annotated tparams are assumed to be Star except the last one which is Eff
-      val tparams1 = tparams0 match {
+      // In the case in which the user doesn't supply any type params,
+      // the parser will have already notified the user of this error
+      // The recovery step here is to simply add a single type param that is never used
+      val tparams1 = if (tparams0.tparams.isEmpty) {
+        val regionTparam = ResolvedAst.TypeParam.Unkinded(Name.Ident("", loc), Symbol.freshUnkindedTypeVarSym(Ast.VarText.Absent, true, loc), loc)
```
magnus-madsen, line 121: We probably want to give it some name, not just the empty string, at least for debugging purposes.

How about `_rc`.

mlutze: Counterintuitively, the `isRegion` parameter here should actually be `false`. It's because we're not defining a literal region here, but rather a region variable.

mlutze: Btw we should almost always use named parameters for Booleans (`isRegion = false`)

main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
```diff
@@ -114,35 +114,48 @@ object Kinder {
...
+      // if not annotated, tparams are assumed to be Star except the last one which is Eff
+      val tparams2 = tparams1 match {
         case ResolvedAst.TypeParams.Kinded(t) => ResolvedAst.TypeParams.Kinded(t)
         case ResolvedAst.TypeParams.Unkinded(t) =>
           val tparams = t.init.map(tparam => ResolvedAst.TypeParam.Kinded(tparam.name, tparam.sym, Kind.Star, tparam.loc)) :+ ResolvedAst.TypeParam.Kinded(t.last.name, t.last.sym, Kind.Eff, t.last.loc)
           ResolvedAst.TypeParams.Kinded(tparams)
       }
-      val kenv0 = getKindEnvFromTypeParamsDefaultStar(tparams1)
-      val kenvVal = kenv0 + (tparams1.tparams.last.sym -> Kind.Eff)
+      val kenv0 = getKindEnvFromTypeParamsDefaultStar(tparams2)
```
mlutze, line 133: Since the last param (which is `Eff`) is included here, I'm afraid this might cause a kind error since this line will say it is `Star`

main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
dghosef: What if `_rc` shadows an actual name?

dghosef: Maybe `$rc`? since variables can't have `$`

main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
```diff
@@ -114,35 +114,48 @@ object Kinder {
...
       flatMapN(kenvVal) {
         case kenv =>
-          val tparamsVal = traverse(tparams1.tparams)(visitTypeParam(_, kenv))
+          val tparamsVal = traverse(tparams2.tparams)(visitTypeParam(_, kenv))
 
           flatMapN(tparamsVal) {
-            case tparams =>
+            case tparams0 =>
+              val KindedAst.TypeParam(name, ksym, tloc) = tparams0.last
+              val lastTparam = KindedAst.TypeParam(name, ksym.withKind(Kind.Eff), tloc)
```
mlutze, line 143: It seems like we're kind of doing the same thing twice here.

If we've established in the kenv above that the last param is Eff, then we shouldn't have to force that in this line.

mlutze:
Give a ping when this is ready :)

dghosef:
> Give a ping when this is ready :)

@mlutze I think it's ready

main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
```diff
@@ -114,35 +114,36 @@ object Kinder {
...
+      // the parser will have already notified the user of this error
+      // The recovery step here is to simply add a single type param that is never used
+      val tparams1 = if (tparams0.tparams.isEmpty) {
+        val regionTparam = ResolvedAst.TypeParam.Unkinded(Name.Ident("$rc", loc), Symbol.freshUnkindedTypeVarSym(Ast.VarText.Absent, isRegion = false, loc), loc)
+        ResolvedAst.TypeParams.Unkinded(List(regionTparam))
+      } else {
+        tparams0
+      }
+      // if not annotated, tparams are assumed to be Star except the last one which is Eff
+      val tparams2 = tparams1 match {
```
mlutze, line 127: What happens if I write `struct S[a: Type, r: Type]`?

main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
```diff
@@ -114,35 +114,36 @@ object Kinder {
...
+      val fieldsVal = traverse(fields0) {
+        case field0 => mapN(visitStructField(field0, kindedTparams, kenv, taenv, root)) {
+          field => field.sym -> field
+        }
       }
-    }
+      mapN(fieldsVal) {
+        case fields =>
+          val targs = kindedTparams.map(tparam => Type.Var(tparam.sym, tparam.loc.asSynthetic))
+          val sc = Scheme(tparams2.tparams.map(t => t.sym.withKind(t.kind)), List(), List(), Type.mkStruct(sym, targs, loc))
```
mlutze, line 143: You should be able to use your `kindedTparams` here.

main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
mlutze, line 127: This technique still seems a little off. But we're in uncharted waters with the restriction on the last parameter.

I propose the following: We ignore whether the tparams are kinded or not. I.e., we just do `tparamsN = tparams0.tparams` so we just have a list of TypeParams. Then you can add the last one if needed.

Then we map over them with `getKindEnvFromTypeParamDefaultStar` and `KindEnv.merge` the result.

Finally we can map over the params with `visitTypeParam` and then deal with fields.

dghosef: We get a kind error, which I think is the correct behavior

dghosef: This won't work because then the last tparam will initially be given kind `Star` and then with the merge we try to assign it kind `Eff`, the merge will fail with a `Validation.hardFailure`(see the `+` operator of `KindEnv`). I initially tried this, but it didn't work, hence my current method of kinding before inserting I merged a while ago.

mlutze:
I'm working on a PR that should facilitate your work here a little better

dghosef:
> I'm working on a PR that should facilitate your work here a little better

Yea I saw that issue - thanks!

Will be working on throw in the meantime

mlutze:
The changes I made have created some merge conflicts, but I think ultimately it should make things easier.
The main difference is that we no longer have `TypeParams`, but rather just a list of `TypeParam`. This way we can break apart the list with less trouble, traverse the first N-1 parameters with the "default star" function, and check the last parameter separately, requiring it to have kind `Eff`.

magnus-madsen:
Is this PR still relevant or can we close it?

dghosef:
It is, just need to get the resolver changes in first

magnus-madsen:
@mlutze Can you do first review? Thanks!

main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
```diff
@@ -114,30 +114,37 @@ object Kinder {
...
+      // the parser will have already notified the user of this error
+      // The recovery step here is to simply add a single type param that is never used
+      val tparams1 = if (tparams0.isEmpty) {
+        val regionTparam = ResolvedAst.TypeParam.Unkinded(Name.Ident("$rc", loc), Symbol.freshUnkindedTypeVarSym(Ast.VarText.Absent, isRegion = false, loc), loc)
+        List(regionTparam)
+      } else {
+        tparams0
+      }
+      // if not annotated, tparams are assumed to be Star except the last one which is Eff
+      val tparams2 = tparams1.init.map(makeKinded(_, Kind.Star)) :+ makeKinded(tparams1.last, Kind.Eff)
```
mlutze, line 127: This is closer, but not quite what I had in mind.

To be consistent with other declarations, the pattern should be:
1. Infer kinds (creating a kind environment)
2. Use that kind environment to update the declaration

So here we should start by building a kind environment from all except the last parameter by using `getKindEnvFromTypeParams`. Then we visit the last parameter with a new function `getKindEnvFromRegion` or something, which defaults to `Eff` rather than `Star`.

So now we have 2 Kind environments. We merge the two and carry on by visiting the type params (rather than "manually" transforming them) and calling the appropriate `visitX` on all the other stuff.

mlutze: This logic should somewhat resemble what goes on with `RestrictableEnum` since that also has a parameter (`index`) which must have a particular kind.

dghosef: Oh, I think I understand a bit better. Does what I have now make more sense?

mlutze: Yes, this looks good.

main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
```diff
@@ -1794,7 +1814,8 @@ object Kinder {
   private def getStructKind(struct0: ResolvedAst.Declaration.Struct)(implicit flix: Flix): Kind = struct0 match {
     case ResolvedAst.Declaration.Struct(_, _, _, _, tparams0, _, _) =>
       // tparams default to zero except for the region param
-      val tparams = tparams0
+      val tparamsStart = tparams0.init.map(makeKinded(_, Kind.Star))
+      val tparams = tparamsStart :+ makeKinded(tparams0.last, Kind.Eff)
```
mlutze, line 1818: Just need to do the same thing here and then we're golden

magnus-madsen:
Hit me up when Matt is happy

mlutze (approved):

magnus-madsen (approved):

PR #8112, opened 2024-07-30
feat: return instantiated variable mapping in `Scheme.instantiate`

dghosef:
This PR updates `Scheme.instantiate` to return a type variable mapping from the original generalized variable to the instantiated variable. It also updates all current uses of `Scheme.instantiate` to ignore it.

This is useful for structs.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/Scheme.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/Scheme.scala
@@ -57,7 +57,7 @@ object Scheme {
         // Determine the rigidity of the fresh type variable.
         macc + (tvar -> Type.freshVar(tvar.kind, loc, tvar.isRegion, Ast.VarText.Absent))
     }
-    val freshVars = substMap.map(s => s._1.id -> s._2)
+    val freshVars = substMap.map { case (k, v) => k.id -> v }
 
     /**
       * Replaces every variable occurrence in the given type using `freeVars`.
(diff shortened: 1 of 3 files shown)


magnus-madsen:
Question: Should we return the map or a Substitution?

dghosef:
> Question: Should we return the map or a Substitution?

The map because
1) It is easy to construct the substitution when we need it
2) Having access to the raw map is useful in instantiate because sometimes I want to look up individual type variables. e.g. to find what the instantiated region variable is. `Substitution.m` gives a `kindedTypeVarSym -> Type` map but it is usefule to lookup a `kindedTypeVarSym -> Type.Var` map

magnus-madsen:
@mlutze If you agree, you can merge.

main/src/ca/uwaterloo/flix/language/ast/Scheme.scala
```diff
@@ -45,18 +45,19 @@ object Scheme {
...
     // Compute the fresh variables taking the instantiation mode into account.
     //
-    val freshVars = sc.quantifiers.foldLeft(Map.empty[Int, Type.Var]) {
+    val substMap = sc.quantifiers.foldLeft(Map.empty[Symbol.KindedTypeVarSym, Type.Var]) {
       case (macc, tvar) =>
         // Determine the rigidity of the fresh type variable.
-        macc + (tvar.id -> Type.freshVar(tvar.kind, loc, tvar.isRegion, Ast.VarText.Absent))
+        macc + (tvar -> Type.freshVar(tvar.kind, loc, tvar.isRegion, Ast.VarText.Absent))
     }
+    val freshVars = substMap.map(s => s._1.id -> s._2)
```
mlutze, line 60: The `_1` stuff always makes me sad.

Can we do `substMap.map { (k, v) => k.id -> v }` instead?

mlutze (approved):

PR #8115, opened 2024-07-30
feat: add `StructNew`, `StructGet`, and `StructPut` to all ASTs and also update `Struct` and `StructField` definitions

dghosef:
This is a big PR that has a lot of repetitive work that adds relevant data structures for structs to all phases. Merging this would be helpful in allowing me to parallelize all my other struct PRs and reduce dependencies between them.

Struct expressions come after array expressions in the ASTs that have arrays and they come after `NewObject` expressions in the ASTs that don't have arrays.


main/src/ca/uwaterloo/flix/language/ast/LiftedAst.scala
```diff
@@ -86,6 +86,12 @@ object LiftedAst {
 
     case class NewObject(name: String, clazz: java.lang.Class[_], tpe: MonoType, purity: Purity, methods: List[JvmMethod], loc: SourceLocation) extends Expr
 
+    case class StructNew(sym: Symbol.StructSym, exps: List[(Symbol.StructFieldSym, Expr)], exp: Expr, tpe: MonoType, purity: Purity, loc: SourceLocation) extends Expr
+
+    case class StructGet(sym: Symbol.StructSym, exp1: Expr, field: Name.Label, tpe: MonoType, purity: Purity, loc: SourceLocation) extends Expr
```
magnus-madsen, line 91: You should consider if at least Get and Put could be atomics. That would save you alot of work.

magnus-madsen:
@dghosef I added one comment which is an easy fix. Once this is done, I can merge.

You should consider whether you can use Atomic for the backend.

PR #7926, opened 2024-06-21
feat: add Flix non-primitives and Java objects subtyping with `Null` type

chanattan:
Takes back work from #7919.

PR #8118, opened 2024-07-30
chore: update import syntax in `Files`

chanattan:
Related to #7944,
should be merged in after array support, i.e., #7926.
@magnus-madsen


magnus-madsen:
@chanattan I see some test failures.

magnus-madsen:
I think you should import as:

```
import java.lang.{String => JString}
```

and use JString to avoid clash with the Flix string.

chanattan:
Indeed it seems to be more careful this way, then I also did it for the File import.
Also it should be considered after merging arrays support.

magnus-madsen (approved):

PR #8124, opened 2024-07-30
feat: check for illegal nested try catch blocks

### LINKED ISSUE #2477: Disallow nested try-catch
Nested try-catch'es with overlapping catch types does not nest properly (issue discussed #2255) so we should disallow them syntactically. This doesn't seem like a great loss since they should probably be limited in small conversion functions where nesting is not needed.
This can be done in two ways:
* In the parser, by adding try-catch the the top of the recursion (a bit like stm) and have every recursive expression rule point to the rule below try-catch
* Add a check to an AST traversal phase. maybe the weeder

JonathanStarup:
Fixes  #2477

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Safety.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Safety.scala
@@ -70,7 +70,7 @@ object Safety {
       case (acc, e) => acc.markRigid(e)
     }
     sig.exp match {
-      case Some(exp) => visitExp(exp, renv)(flix, inTryCatch = false)
+      case Some(exp) => visitExp(exp, renv)(inTryCatch = false, flix)
       case None => Nil
     }
   }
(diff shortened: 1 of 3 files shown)


main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -226,12 +226,10 @@ object Safety {
   /**
     * Performs safety and well-formedness checks on the given expression `exp0`.
     */
-  private def visitExp(e0: Expr, renv: RigidityEnv)(implicit flix: Flix): List[SafetyError] = {
+  private def visitExp(e0: Expr, renv: RigidityEnv)(implicit flix: Flix, inTryCatch: Boolean): List[SafetyError] = {
```
magnus-madsen, line 229: inTryCatch should be fore flix :)

magnus-madsen (approved):

PR #8126, opened 2024-07-30
feat: add support for other Flix primitives in subtyping

chanattan:


main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
```diff
@@ -264,12 +266,20 @@ object TypeReduction {
...
+      case (Type.Cst(TypeConstructor.Native(clazz), _), Type.Cst(TypeConstructor.Float32, _)) => clazz.isAssignableFrom(classOf[java.lang.Float])
+      case (Type.Cst(TypeConstructor.Native(clazz), _), Type.Cst(TypeConstructor.Float64, _)) => clazz.isAssignableFrom(classOf[java.lang.Double])
+      case (Type.Cst(TypeConstructor.Native(clazz), _), Type.Cst(TypeConstructor.Int8, _)) => clazz.isAssignableFrom(classOf[java.lang.Byte])
+      case (Type.Cst(TypeConstructor.Native(clazz), _), Type.Cst(TypeConstructor.Int16, _)) => clazz.isAssignableFrom(classOf[java.lang.Short])
+      case (Type.Cst(TypeConstructor.Native(clazz), _), Type.Cst(TypeConstructor.Int32, _)) => clazz.isAssignableFrom(classOf[java.lang.Integer])
+      case (Type.Cst(TypeConstructor.Native(clazz), _), Type.Cst(TypeConstructor.Int64, _)) => clazz.isAssignableFrom(classOf[java.lang.Long])
+      case (Type.Cst(TypeConstructor.Native(clazz), _), Type.Cst(TypeConstructor.Str, _)) => clazz.isAssignableFrom(classOf[java.lang.String])
+      case (Type.Cst(TypeConstructor.Native(clazz), _), Type.Cst(TypeConstructor.BigInt, _)) => clazz.isAssignableFrom(classOf[java.math.BigInteger])
```
mlutze, line 278: This looks fine for String and BigInt (though we should add tests)

However, for `Bool`, `Char,` `Float*`, `Int*`, subtyping is _not_ possible. In particular, there's a difference between `java.lang.Short` (the Java class) and `short` (the Java primitive). So we can get rid of these cases.

chanattan: I removed the cases, thank you.

For int comparisons, are Int32 in Flix and Integer in Java the same so that isSubtype passes in the equality check?

Also, I still don't understand how these cases are not possible?
Don't we consider something like,
```java
public void test(Integer i) {
   ...
}

...
test(5)
```
?

magnus-madsen: Java has `int` and `Integer`. They are not the same, `Integer` is boxed.

Flix has `Int32` which is `int`. 

If you want to call a method that requires `Integer` you must manually box it first.

magnus-madsen:
Yes, lets get some systematic tests.

magnus-madsen:
@chanattan Ping us when this is ready for another look 👍

chanattan:
@magnus-madsen @mlutze since I removed the cases mentioned above the tests should already be present for most of them, also some tests are present in the new PR #8153 that should be merged in afterwards.

magnus-madsen:
@mlutze If you are good with this, you can merge.

mlutze (approved):

PR #8127, opened 2024-07-30
Argument parsing demo

JonathanStarup:
@magnus-madsen its almost a feature
```
do E.op // means the same as do E.op()
new Class // means the same as new Class()
object.fun // means the same as object.fun()
```


JonathanStarup:
The fix is to peek and expect the opening paren before the zeroOrMore thing. ill fix it at some point

PR #8130, opened 2024-07-30
feat: add support for static field access on imported Java classes

magnus-madsen:


mlutze (approved):
Would be nice to have more tests to see this in context. E.g. as a parameter

magnus-madsen:
> Would be nice to have more tests to see this in context. E.g. as a parameter

As a parameter???

mlutze:
> > Would be nice to have more tests to see this in context. E.g. as a parameter
> 
> As a parameter???

E.g. `foo(Math.PI)`

PR #8137, opened 2024-07-31
chore: update import syntax for `Vector.flix`

magnus-madsen:


magnus-madsen:
@mlutze What's your thought on doing it this way? I just mark the iterator creation as pure...

mlutze:
It seems like the whole adapter interface is flawed. If calling Java stuff is IO, then surely the adapter must inherently have the IO effect.

magnus-madsen:
> It seems like the whole adapter interface is flawed. If calling Java stuff is IO, then surely the adapter must inherently have the IO effect.

One can make that argument. But we should probably discuss that separately.

mlutze (approved):

Issue #8096, opened 2024-07-29
Clean up old imports in standard library


mlutze:
```bash
grep -rI "import" | cut -f1 -d: | grep "flix$" | sort -u | awk '{ print ("- [ ] " $0) }'
```

- [x] examples/using-laziness-for-logging.flix
- [x] main/src/library/Adaptor.flix
- [x] main/src/library/Add.flix
- [x] main/src/library/Benchmark.flix
- [x] main/src/library/BigDecimal.flix
- [x] main/src/library/BigInt.flix
- [x] main/src/library/Chain.flix
- [x] main/src/library/Char.flix
- [x] main/src/library/CodePoint.flix
- [x] main/src/library/Concurrent/Channel.flix
- [x] main/src/library/Concurrent/Condition.flix
- [x] main/src/library/Concurrent/CyclicBarrier.flix
- [x] main/src/library/Concurrent/ReentrantLock.flix
- [x] main/src/library/Console.flix
- [x] main/src/library/Debug.flix
- [x] main/src/library/Div.flix
- [x] main/src/library/Environment.flix
- [x] main/src/library/Eq.flix
- [x] main/src/library/File.flix
- [x] main/src/library/Files.flix
- [x] main/src/library/Fixpoint/Solver.flix
- [x] main/src/library/Float32.flix
- [x] main/src/library/Float64.flix
- [x] main/src/library/FromJava.flix
- [x] main/src/library/Hash.flix
- [x] main/src/library/Int16.flix
- [x] main/src/library/Int32.flix
- [x] main/src/library/Int64.flix
- [x] main/src/library/Int8.flix
- [x] main/src/library/IOError.flix
- [x] main/src/library/List.flix
- [x] main/src/library/Map.flix
- [x] main/src/library/Mul.flix
- [x] main/src/library/Nec.flix
- [x] main/src/library/Neg.flix
- [x] main/src/library/Nel.flix
- [x] main/src/library/Object.flix
- [x] main/src/library/Order.flix
- [x] main/src/library/PartialOrder.flix
- [x] main/src/library/Prelude.flix
- [x] main/src/library/Random.flix
- [x] main/src/library/Regex.flix
- [x] main/src/library/Result.flix
- [x] main/src/library/Set.flix
- [x] main/src/library/StringBuilder.flix
- [x] main/src/library/String.flix
- [x] main/src/library/Sub.flix
- [x] main/src/library/System.flix
- [x] main/src/library/Thread.flix
[...]

PR #8152, opened 2024-07-31
chore: new method invocation syntax in `Test.Exp.Jvm.NewObject.flix`

dghosef:
I chose not to refactor `checked_cast` expressions since they don't seem to be compatible java style method invocation.

This addresses https://github.com/flix/flix/issues/8096


magnus-madsen (approved):

magnus-madsen:
@dghosef Yes, we should be able to remove checked_cast with the new style.

PR #8164, opened 2024-07-31
refactor: refactor and short fix for arrays in subtyping

chanattan:
@magnus-madsen @mlutze
there was a mistake in subtyping for arrays, as subtyping was made the other way.
Fixed and refactored to be more intuitive. If you can merge it fast for the other tests to be checked.
Thank you.


magnus-madsen:
@chanattan Are you sure this is right? You swap both the arguments you match on and the cases, so there is no difference?

chanattan:
> @chanattan Are you sure this is right? You swap both the arguments you match on and the cases, so there is no difference?

Yes for most the cases there is no difference as it is only pure refactoring to make it more intuitive. But I didn't swap the arguments for the Arrays which allows to check subtyping in the right way.

magnus-madsen (approved):

PR #8166, opened 2024-07-31
chore: update imports in `Adaptor.flix`

JonathanStarup:
Part of #8096


magnus-madsen (approved):

magnus-madsen:
@JonathanStarup I see some failures. Not sure what is happening here...

JonathanStarup:
@chanattan do you know what these types are about? There is `TypeConstructor.JvmMethod(_)` leftover to simplified so it looks like some of them are not reduced in the solver

chanattan:
> @chanattan do you know what these types are about? There is `TypeConstructor.JvmMethod(_)` leftover to simplified so it looks like some of them are not reduced in the solver. Its the same in #8167

They should be simplified when the type reduction/resolution found the corresponding Java method.
I am not on desk right now, can you debug to see if it remains unsimplified somewhere? I didn’t see it back when I tried to debug.

PR #8168, opened 2024-07-31
fix: check for reducible types in `ConstraintSolver`

magnus-madsen:


magnus-madsen:
@mlutze @JonathanStarup This is a hack, no doubt. But I don't think it causes any issues with anything else and its very short. Can either of you approve and merge? Thanks!

mlutze (approved):

PR #8170, opened 2024-08-01
chore: update import syntax in StringBuilder.flix

magnus-madsen:


magnus-madsen:
@mlutze Works, need approval.

mlutze (approved):
Can't these all be checked casts?

magnus-madsen:
> Can't these all be checked casts?

Possibly. I used unchecked cast because I wanted to be exact about the effect. 

I think this will get cleaned up later when we get Jonathan's subeffecting and he goes over all casts.

Issue #8154, opened 2024-07-31
Add `throw` expression


magnus-madsen:
Add a `Throw` expression just before `TryCatch` in the expressions. In TypedAst it should have the form:

```
case class Throw(exp: Expr, tpe: Type, eff: Type, loc: SourceLocation) extends Expr
```

Its effect should be IO.

I think it can start out as an AtomicOp in the backend.

Its syntax should simply be `throw e`.

I suggest to make several PRs starting from the lexer.

JonathanStarup:
Why do we need throw?

magnus-madsen:
> Why do we need throw?

Purely for interop and completeness. You should not use it, unless you absolutely have to.

magnus-madsen:
@dghosef Would you say this is done?

dghosef:
Yes

PR #8184, opened 2024-08-01
feat: add `throw` to lexer

dghosef:
related to #8154


magnus-madsen:
Unfortunately some code already uses `throw` as an effect name. I suggest to rename those defs to use `raise` as the name.

magnus-madsen:
Sorry for being a major pain, but I had a different refactoring in mind. I have done it myself here https://github.com/flix/flix/pull/8199 so just need Matt to approve and then you should be good to again with the lexer.

magnus-madsen (approved):

PR #8189, opened 2024-08-01
chore: Fix regex imports

dghosef:
#8096 

Unfortunately, `Regex.flix` at one point calls a method entitled `region`. To fix this, temporarily accept `₹region` as a non-keyword "region" token. Should work for any lowercase name.


magnus-madsen (approved):

JonathanStarup:
@magnus-madsen im a bit concerned about the checked_cast and unsafe combo. i think it could go wrong with the optimizer?

magnus-madsen:
> @magnus-madsen im a bit concerned about the checked_cast and unsafe combo. i think it could go wrong with the optimizer? One of the subexpressions here have type Unit and effect {}. sounds dangerous

Can you open a ticket with more details? Thanks!

PR #8194, opened 2024-08-01
fix: parse missing argument list

JonathanStarup:
treats `a.f` as different than `a.f()` in select missing cases. illustrated in #8127


main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -2510,6 +2511,7 @@ object Parser2 {
       } else {
         // Case 3: new Type(exps...)
         arguments()
+        arguments()
```
magnus-madsen, line 2514: What? Explain... You call arguments twice???

JonathanStarup: ah, my bad. this also shows why `arguments` behave a bit funny. it fails very silently

magnus-madsen (approved):

PR #8201, opened 2024-08-02
chore: remove dead code

mlutze:
Remember when every effect variable was actually two effect variables?


magnus-madsen (approved):

magnus-madsen:
> Remember when every effect variable was actually two effect variables?

Pepperidge farm remembers!

PR #8202, opened 2024-08-02
chore: remove 2 from Parser2 and Weeder2

mlutze:


magnus-madsen:
I don't want to do this just yet. I was hoping to use the renaming as part of a reviewing process.

PR #8196, opened 2024-08-01
chore: disable deprecated features for tests

magnus-madsen:

PR #8209, opened 2024-08-02
refactor: fix failing tests

chanattan:
Related to #8196.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/DesugaredAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/DesugaredAst.scala
@@ -40,19 +40,19 @@ object DesugaredAst {
 
     case class Instance(doc: Ast.Doc, ann: Ast.Annotations, mod: Ast.Modifiers, trt: Name.QName, tpe: Type, tconstrs: List[TypeConstraint], assocs: List[Declaration.AssocTypeDef], defs: List[Declaration.Def], loc: SourceLocation) extends Declaration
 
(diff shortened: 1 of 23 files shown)


chanattan:
For `IllegalType.01` and `IllegalWildType.05` the errors are different, should I change the test names?
Also for `MismatchingReturnType` I changed them to `MismatchingType` as we no longer explicit the signature of an imported function but inference directly looks at types of the inferred function. @magnus-madsen

magnus-madsen:
> For `IllegalType.01` and `IllegalWildType.05` the errors are different, should I change the test names? Also for `MismatchingReturnType` I changed them to `MismatchingType` as we no longer explicit the signature of an imported function but inference directly looks at types of the inferred function. @magnus-madsen

Sure, yes.

Just apply comment sense and add some comments on GitHub (You can comment on code here) and then I will take a look.

If a test is no longer relevant, feel free to remove it or rework it.

magnus-madsen:
Could we deal with TestResolver in a PR separate from this one?

chanattan:
> > For `IllegalType.01` and `IllegalWildType.05` the errors are different, should I change the test names? Also for `MismatchingReturnType` I changed them to `MismatchingType` as we no longer explicit the signature of an imported function but inference directly looks at types of the inferred function. @magnus-madsen
> 
> Sure, yes.
> 
> Just apply comment sense and add some comments on GitHub (You can comment on code here) and then I will take a look.
> 
> If a test is no longer relevant, feel free to remove it or rework it.

I did what seemed intuitive to me.
Now about Test.Java.Function.flix we should be working on another PR first that deals with IntPredicate and higher-order Flix functions, if I'm not wrong. It should be a specific case in TypeReduction.

magnus-madsen:
What's the problem with Test.Java.Function?

magnus-madsen:
Fixed in master.

PR #8213, opened 2024-08-03
refactor: Test.Java.Function.flix

chanattan:
Related to #8196.
The problem is that there is no yet support for lambda functions as parameters in Java methods, that is the Java equivalent would be an IntPredicate.

Something like

```scala
stream.filter(obj -> obj.func())
```
isn't normally working.

From #8209.
@magnus-madsen

### DIFF
--- a/main/test/flix/Test.Java.Function.flix
+++ b/main/test/flix/Test.Java.Function.flix
@@ -19,165 +19,128 @@ mod Test.Java.Function {
     import java.util.function.DoubleUnaryOperator
     import java.util.stream.DoubleStream
     import java.util.stream.Stream
+    import java.util.Optional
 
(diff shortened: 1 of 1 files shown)


magnus-madsen:
We already implement the right interfaces, so I think what you need is to extend what subtyping allows.

magnus-madsen:
i.e. a flix function type should be treated as a subtype of the appropriate interface.

magnus-madsen:
This code may be useful for inspiration:

https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/Resolver.scala#L3559-L3594

magnus-madsen:
Fixed in master.

PR #8222, opened 2024-08-05
feat: support non-generic flix functions interfaces in subtyping

chanattan:
Related to #8213.
WIP.


chanattan:
I added some cases from https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/Resolver.scala#L3559-L3594 @magnus-madsen @mlutze,
could you review them, please?

In order of progression:

- The unresolved types for anonymous functions were, in some cases, solved (i.e., known) as I was before running with no lib. I then proceeded to add some cases from Resolver.
- It seems like the type args are not known for this `e -> e.equals("a")` whereas they are known for `e -> e == "a"`.
- I tried to make the following program work with the type-known predicate.

The following program crashes at the Verifier:

```scala
def main(): Unit \ IO = region rc {
    let arr = Array#{"a", "b", "c"} @ rc;
    let stream0 = Stream.of(arr);
    let filtered = stream0.filter(e -> e == "a");
    let elm = filtered.findFirst().get().toString();
    println(elm)
}
```
but does pass the TypeReduction. It is because the Flix function type `Str -> Bool` is not handled in the match case as a `Predicate`. I can work on it if necessary as it doesn't seem too hard.

Here is the error:

```scala
ca.uwaterloo.flix.util.InternalCompilerException: Mismatched types near /home/csok/IdeaProjects/flix/test.flix:9:20: tpe1 = Arrow(List(String),Bool), class = interface java.util.function.Predicate
	at ca.uwaterloo.flix.language.phase.Verifier$.failMismatchedTypes(Verifier.scala:682)
	at ca.uwaterloo.flix.language.phase.Verifier$.checkJavaSubtype(Verifier.scala:619)
	at ca.uwaterloo.flix.language.phase.Verifier$.$anonfun$checkJavaParameters$1(Verifier.scala:574)
	...
```

Then here is a working program with IntConsumer:

```scala
def main() : Unit \ IO = region rc {
    let arr = Array#{1, 2, 3} @ rc;
    let stream0 = IntStream.of(arr);
    stream0.forEach(x -> println(x + 1i32))
}
```
Output:
```
2
3
4
```

magnus-madsen:
I will let @JonathanStarup comment. He is the expert :)

magnus-madsen:
(Note: The goal here is only to support non-generic uses cases-- for now)

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
```diff
@@ -293,13 +293,46 @@ object TypeReduction {
       case (Type.Apply(Type.Apply(Type.Cst(TypeConstructor.Array, _), elmType1, _), rcVar1, _),
       Type.Apply(Type.Apply(Type.Cst(TypeConstructor.Array, _), elmType2, _), rcVar2, _)) =>
         isSubtype(elmType1, elmType2)
+      // Arrow to Java function interface
+      case (Type.Apply(Type.Apply(Type.Apply(Type.Cst(TypeConstructor.Arrow(n), _), eff, _), var_arg, _), var_ret, _), Type.Cst(TypeConstructor.Native(clazz), _)) =>
+        val targsVal = tpe1.typeArguments.map(purifyType)
+        targsVal match { // TODO INTEROP: generics support
+          // Pattern-matching Flix functions of the form: eff -> arg -> ret
+          case eff :: TypeConstructor.Native(_) :: TypeConstructor.Native(_) :: Nil =>
```
JonathanStarup, line 301: So, if the goal is just to support non-polymorphic functions, then i'd expect `t -> Unit \ eff` to be `Consumer[t]`. @magnus-madsen does any function in java actually use non-polymorphic function interfaces?

JonathanStarup: Maybe we should start with just IntConsumer and the other non-generic function interfaces?

magnus-madsen: I just want to most basic things to work, .e.g IntConsumer

chanattan:
I'm trying to support basic things, i.e., IntConsumer, IntPredicate and IntUnaryOperator. The problem is that the types for anonymous functions are not always known even with IntStream, e.g.:

```scala
stream0.forEach(e -> println(e))
```

this program does no longer generate the additionnal constraint of addition: `e + 1i32` that allows to infer e is of type `Int32`.
Do you think we should add a stronger constraint in the Constraintsolver?

chanattan:
This should work for now, we may force annotation for current refactoring. If you have any insight, I would be happy to have some information @JonathanStarup.

magnus-madsen:
@JonathanStarup Could you take a look? Thanks!

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
```diff
@@ -293,13 +293,91 @@ object TypeReduction {
       case (Type.Apply(Type.Apply(Type.Cst(TypeConstructor.Array, _), elmType1, _), rcVar1, _),
       Type.Apply(Type.Apply(Type.Cst(TypeConstructor.Array, _), elmType2, _), rcVar2, _)) =>
         isSubtype(elmType1, elmType2)
+      // Arrow to Java function interface
+      // TODO INTEROP: generics support
+      // Pattern-matching Flix functions of the form: eff -> arg -> ret
+      // IntStream
+      case (Type.Apply(Type.Apply(Type.Apply(Type.Cst(TypeConstructor.Arrow(n), _), eff, _), var_arg, _), var_ret, _), Type.Cst(TypeConstructor.Native(clazz), _))
```
JonathanStarup, line 300: I think the `n` in `TypeConstructor.Arrow(n)` should be `2`

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
```diff
@@ -293,13 +293,91 @@ object TypeReduction {
...
+      if clazz == classOf[java.util.function.IntConsumer] =>
+        val targsVal = tpe1.typeArguments.map(purifyType)
+        targsVal match {
```
JonathanStarup, line 303: and then here it should be `(var_arg, var_ret) match {case (Type.Cst(TypeConstructor.Int32, _), Type.Cst(TypeConstructor.Unit, _)) => ...` }

JonathanStarup: oh and also not snake_case

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
chanattan: I see, thank you.

chanattan:
I did some better refactoring.

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
```diff
@@ -293,13 +293,49 @@ object TypeReduction {
       case (Type.Apply(Type.Apply(Type.Cst(TypeConstructor.Array, _), elmType1, _), rcVar1, _),
       Type.Apply(Type.Apply(Type.Cst(TypeConstructor.Array, _), elmType2, _), rcVar2, _)) =>
         isSubtype(elmType1, elmType2)
+      // Arrow to Java function interface
+      // TODO INTEROP: generics support
+      case (Type.Apply(Type.Apply(Type.Apply(Type.Cst(TypeConstructor.Arrow(2), _), eff, _), var_arg, _), var_ret, _), Type.Cst(TypeConstructor.Native(clazz), _)) =>
+        (purifyType(var_arg), purifyType(var_ret)) match {
```
JonathanStarup, line 299: it looks good now, but instead of purifyType, maybe just match on (var_arg, var_ret) and have case (Type.Cst(a, _), Type.Cst(b, _) => and then inside that case have the current match.

chanattan: Ah I see, something like this?
Thank you.

magnus-madsen:
@JonathanStarup RFM?

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
```diff
@@ -293,6 +293,37 @@ object TypeReduction {
       case (Type.Apply(Type.Apply(Type.Cst(TypeConstructor.Array, _), elmType1, _), rcVar1, _),
       Type.Apply(Type.Apply(Type.Cst(TypeConstructor.Array, _), elmType2, _), rcVar2, _)) =>
         isSubtype(elmType1, elmType2)
+      // Arrow to Java function interface
+      // TODO INTEROP: generics support
+      case (Type.Apply(Type.Apply(Type.Apply(Type.Cst(TypeConstructor.Arrow(2), _), eff, _), var_arg, _), var_ret, _), Type.Cst(TypeConstructor.Native(clazz), _)) =>
```
magnus-madsen, line 298: no underscores in variable names, thanks :)

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
```diff
@@ -293,6 +293,37 @@ object TypeReduction {
...
+                clazz == classOf[java.util.function.DoublePredicate]
+              case (TypeConstructor.Float64, TypeConstructor.Float64) =>
+                clazz == classOf[java.util.function.DoubleUnaryOperator]
+              // LongStream
+              case (TypeConstructor.Int64, TypeConstructor.Unit) =>
+                clazz == classOf[java.util.function.LongConsumer]
+              case (TypeConstructor.Int64, TypeConstructor.Bool) =>
+                clazz == classOf[java.util.function.LongPredicate]
+              case (TypeConstructor.Int64, TypeConstructor.Int64) =>
+                clazz == classOf[java.util.function.LongUnaryOperator]
```
magnus-madsen, line 322: Are these all the relevant functions?

JonathanStarup:
looks good other than remaining comments yeah

magnus-madsen:
OK, I will @chanattan have another go.

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
chanattan: This should be. As they were supported in Resolver.scala.

chanattan: (for non-generic things)

magnus-madsen:
Can we add some tests? Thanks!

magnus-madsen:
Fixed in master.

PR #8224, opened 2024-08-05
feat: add `throw` to `AtomicOp`

dghosef:
#8154


main/src/ca/uwaterloo/flix/language/ast/AtomicOp.scala
```diff
@@ -98,4 +98,6 @@ object AtomicOp {
 
   case object MatchError extends AtomicOp
 
+  case object Throw extends AtomicOp
```
magnus-madsen, line 101: Could we move Throw after PutStaticField? That way its group with other JVM operations.

The same goes for pattern matches etc.

magnus-madsen (approved):

Issue #8158, opened 2024-07-31
Add varargs support


mlutze:
# What is "VarArgs?"

Java supports "variable arity" methods (also called varargs). These allow one to write a function like

```java
public void printf(String format, Object... args) {
  // ...
}
```

and then call it with a variable number of arguments:
```java
printf("Age: %d, Name: "%s", 8, "Alice")  // 3 args
printf("Blah")                       // 1 arg
printf("Hello, %d", "Mr. Mouse") // 2 args
```

This is just syntactic sugar for a function taking an array as an argument:

```java
public void printf(String format, Object[] args) {
  // ...
}
```

and then constructing the appropriate array at the call site.

# How can we support it?

We'll need to make changes in a few places.

## Type checking

Internally, Java reports the `println` function above as taking exactly two arguments: a String and an Object array. But there is also a `Method.isVarArgs` function that tells us that `args` _actually_ means a list of arguments instead of just an array.

In type inference, we need to separate _varargs_ methods from non-_varargs_ methods and consider the non-_varargs_ methods first. If none of those match, we then consider the _varargs_ methods. In the _varargs_ case, we look for a method whose last parameter is an array and check that all the _varargs_ parameters are subtypes of the element type of that array.

For example, if we have the Flix code
```scala
System.in.printf("hello", "test", "blah")
```

Then we:
1. Consider all non-_varargs_ methods matching `printf` (there are none)
2. Consider all _varargs_ methods matching `printf`:
  - `printf(Locale l, String format, Object... args)` does not fit due to the first parameter `Locale`
  - `printf(String format, Object... args)` does fit: `test` and `blah` are both `String`s which are subtypes of `Object`
  
## Code gen

We still need to desugar the _varargs_ into Arrays. We can do this in code generation: Add a new array literal construction for the _varargs_ at the call site.

JonathanStarup:
sounds slick

magnus-madsen:
@mlutze Fixed?

mlutze:
Still have to remove the old syntax, I think.

magnus-madsen:
> Still have to remove the old syntax, I think.

What's the old syntax?

mlutze:
```
    @Test
    def testInvokeVarArgs01(): Unit \ Assert + IO =
        let p = Path.of("hello", (...{}: Vector[String]));
        assertEq(expected = "hello", p.toString())
```

magnus-madsen:
We can remove that

PR #8225, opened 2024-08-05
feat: add support for varargs

chanattan:
WIP.
Related to #8158.


mlutze:
Since it's marked as a draft, I will hold off on reviewing, but let me know if I should take a look.

magnus-madsen:
What's needed to push this over the finish line? :)

chanattan:
I am still testing some cases on my own, it should be half-way done.

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -222,6 +239,35 @@ object Simplifier {
       throw InternalCompilerException(s"Unexpected expression: $exp0.", loc)
   }
 
+  /**
+   * Helper method to get a corresponding monotype to the given type.
+   */
+  private def getMonotype(tpe: Type): MonoType =
```
mlutze, line 245: Can we do `visitType` for this one?

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
```diff
@@ -258,22 +279,53 @@ object TypeReduction {
    * @param methodName the potential candidate method's name
    * @param ts         the list of parameter types of the potential candidate method
    */
-  private def isCandidateMethod(cand: Method, methodName: String, isStatic: Boolean = false, ts: List[Type])(implicit flix: Flix): Boolean = {
+  private def isCandidateMethod(cand: Method, methodName: String, isStatic: Boolean = false, isVariadic: Boolean = false, ts: List[Type])(implicit flix: Flix): Boolean = {
     val static = java.lang.reflect.Modifier.isStatic(cand.getModifiers)
     (if (isStatic) static else !static) &&
       (cand.getName == methodName) &&
-      (cand.getParameterCount == ts.length) &&
+      // Either check candidate method if it is not variadic
```
mlutze, line 286: This is very hard to read. Can we refactor it into a series of guards?

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
```diff
@@ -258,22 +279,53 @@ object TypeReduction {
...
+      if (argsDiff <= 1) { // We remove the vararg array and check for required arguments
+        // Check subtyping for required args
+        // if (argsDiff <= 0) TODO INTEROP: support where argsDiff <= 0
+        (cand.getParameterTypes zip ts.slice(0, argsDiff + 1)).forall {
+          case (clazz, tpe) => isSubtype(tpe, Type.getFlixType(clazz))
+        } &&
+        // Check subtyping for varargs
+        ts.slice(argsDiff + 1, ts.length).forall(t => isSubtype(t, Type.getFlixType(tpe)))
+      } else false
+    } else false
```
mlutze, line 326: I would reverse these to be early-return guards. E.g. 
```
(if !cand.isVarArgs) {
  return false
}
```

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -616,6 +616,49 @@ object Verifier {
     }
   }
 
+  /**
+   * Returns a corresponding Java class for the given monotype.
+   */
+  def getJavaType(tpe: MonoType, loc: SourceLocation): Class[_] = {
```
mlutze, line 622: Jonathan says this is fine

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
```diff
@@ -220,10 +221,30 @@ object TypeReduction {
...
-      case 0 => JavaMethodResolutionResult.MethodNotFound
+      case 0 =>
+        // If no method was found, check if it is potentially a variadic method
+        val varMethods = clazz.getMethods.filter(m => isCandidateMethod(m, methodName, isStatic, isVariadic = true, ts))
+        varMethods.length match {
+          case 0 => JavaMethodResolutionResult.MethodNotFound
+          case 1 =>
+            val tpe = Type.Cst(TypeConstructor.JvmMethod(varMethods.head), loc)
+            JavaMethodResolutionResult.Resolved(tpe)
+          case 2 =>
```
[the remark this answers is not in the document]
chanattan: Yes thank you

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
chanattan: Oh it seems right. I removed the function and tried with visitType instead.

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
chanattan: Ok thank you. It isn't complete for now as we want to have a working version.

chanattan:
The base should work, now `checkJavaSubtype` instead of `checkEq` seems indeed weaker but stronger in that it requires a Java (class) comparison. That is why we have so many errors, we should probably have two cases to match the two functions depending on the array element type. There we consider again a certain comparison between MonoTypes and Java types, what is your opinion @mlutze?

mlutze:
> The base should work, now `checkJavaSubtype` instead of `checkEq` seems indeed weaker but stronger in that it requires a Java (class) comparison. That is why we have so many errors, we should probably have two cases to match the two functions depending on the array element type. There we consider again a certain comparison between MonoTypes and Java types, what is your opinion @mlutze?

Yeah that's tricky. Ugh.

Maybe instead of converting to a Java class, we can just create a new `isSubType` function that mostly does the same as `isJavaSubtype` but also allows Flix stuff. Essentially the same as the subtyping check you had in another phase.

chanattan:
I see, I will think about this and implement it a bit later. Thank you!

magnus-madsen:
We can skip the verifier check if it is varargs.

chanattan:
> We can skip the verifier check if it is varargs.

May we add a flag to ArrayLit or create an equivalent operation so that it gets ignored? I believe I should still build something in the Simplifier but it should only get ignored in Verifier.

mlutze:
> > We can skip the verifier check if it is varargs.
> 
> May we add a flag to ArrayLit or create an equivalent operation so that it gets ignored? I believe I should still build something in the Simplifier but it should only get ignored in Verifier.

Right. We can still do the Simplifier translation, but in Verifier, we can just do something on the `invoke` expression where we check if the method is varargs, and if it is, then we skip checking it.

Recall that Verifier is only there for our (the compiler developers) debugging purposes. In general, if it becomes a burden, we shouldn't bother unless we know there is a bug.

magnus-madsen:
> Recall that Verifier is only there for our (the compiler developers) debugging purposes. In general, if it becomes a burden, we shouldn't bother unless we know there is a bug.

Exactly. It is only to help ourselves.

chanattan:
> > > We can skip the verifier check if it is varargs.
> > 
> > 
> > May we add a flag to ArrayLit or create an equivalent operation so that it gets ignored? I believe I should still build something in the Simplifier but it should only get ignored in Verifier.
> 
> Right. We can still do the Simplifier translation, but in Verifier, we can just do something on the `invoke` expression where we check if the method is varargs, and if it is, then we skip checking it.
> 
> Recall that Verifier is only there for our (the compiler developers) debugging purposes. In general, if it becomes a burden, we shouldn't bother unless we know there is a bug.

I see thank you.
However the Simplifier transforms Invoke into an ApplyAtomic where we lose the method, as the arguments are evaluated before the handler expression. The only thing I see as a workaround is either a new AtomicOp (VarArray) or a flag in ArrayLit.

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
```diff
@@ -258,21 +278,54 @@ object TypeReduction {
...
+  private def isCandidateVariadicMethod(cand: Method, ts: List[Type])(implicit flix: Flix): Boolean = {
+    if (!cand.isVarArgs)
+      return false
+    val paramCount = cand.getParameterCount
+    val varArr = cand.getParameterTypes.last
+    val tpe = varArr.getComponentType
+    // Check minimum required arity
+    // e.g., for a defined function f(T arg1, T... args) arg1 is required.
+    val argsDiff = paramCount - ts.length
+    if (argsDiff > 1) // We remove the vararg array and check for required arguments
```
JonathanStarup, line 319: you write that you now check something, but it just returns false, which makes the other method return false. is the comment outdated?

JonathanStarup (commented):
Could you add some tests also? otherwise it seems good :)

magnus-madsen:
Outdated. Will be fixed in separate PR.

PR #8226, opened 2024-08-05
refactor: remove SemanticOp.StringOp.Concat and cases instances of it

saisoma123:
Removed StringOp and Concat from SemanticOp and removed case instances across multiple fies.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Deriver.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Deriver.scala
@@ -901,7 +901,7 @@ object Deriver {
     * Builds a string concatenation expression from the given expressions.
     */
   private def concat(exp1: KindedAst.Expr, exp2: KindedAst.Expr, loc: SourceLocation)(implicit flix: Flix): KindedAst.Expr = {
-    KindedAst.Expr.InvokeMethod2(exp1, Name.Ident("concat", loc), List(exp2), Type.freshVar(Kind.JvmConstructorOrMethod, loc), Type.freshVar(Kind.Star, loc), Type.freshVar(Kind.Eff, loc), loc)
+    KindedAst.Expr.UncheckedMaskingCast(KindedAst.Expr.InvokeMethod2(exp1, Name.Ident("concat", loc), List(exp2), Type.freshVar(Kind.JvmConstructorOrMethod, loc), Type.freshVar(Kind.Star, loc), Type.freshVar(Kind.Eff, loc), loc), loc)
   }
 
   /**
--- a/main/src/ca/uwaterloo/flix/language/phase/Deriver.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Deriver.scala
@@ -901,7 +901,7 @@ object Deriver {
     * Builds a string concatenation expression from the given expressions.
     */
   private def concat(exp1: KindedAst.Expr, exp2: KindedAst.Expr, loc: SourceLocation)(implicit flix: Flix): KindedAst.Expr = {
-    KindedAst.Expr.Binary(SemanticOp.StringOp.Concat, exp1, exp2, Type.freshVar(Kind.Star, loc), loc)
+    KindedAst.Expr.InvokeMethod2(exp1, Name.Ident("concat", loc), List(exp2), Type.freshVar(Kind.JvmConstructorOrMethod, loc), Type.freshVar(Kind.Star, loc), Type.freshVar(Kind.Eff, loc), loc)
   }
 
   /**
--- a/main/src/ca/uwaterloo/flix/language/ast/SemanticOp.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/SemanticOp.scala
@@ -651,18 +651,4 @@ object SemanticOp {
 
   }
 
-  /**
-    * String Operators.
-    */
-  sealed trait StringOp extends SemanticOp
-
-  object StringOp {
-
-    /**
-      * Concatenate.
-      */
-    case object Concat extends StringOp with BinaryOp
-
-  }
-
 }
--- a/main/src/ca/uwaterloo/flix/language/dbg/printer/OpPrinter.scala
+++ b/main/src/ca/uwaterloo/flix/language/dbg/printer/OpPrinter.scala
@@ -112,8 +112,7 @@ object OpPrinter {
          Int16Op.Add |
          Int32Op.Add |
          Int64Op.Add |
-         Int64Op.And |
-         StringOp.Concat => plus
+         Int64Op.And => plus
     case Float32Op.Sub |
          Float64Op.Sub |
          Int8Op.Sub |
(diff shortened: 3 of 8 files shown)


mlutze:
This looks good so far. For the Weeder, you would also just be able to remove the case that creates the StringOp.

In the Deriver, there is a function called `concat`, which generates a string concatenation expression. This should be replaced with an InvokeMethod2 expression, where the methodName is `"concat"`, from the Java class `String`.

This means that the generated code will be `exp1.concat(exp2)` instead of `exp1 + exp2`.

Let me know if you need any clarification.

saisoma123:
`    private def visitStringInterpolationExpr(tree: Tree): Validation[Expr, CompilationMessage] = {
      expect(tree, TreeKind.Expr.StringInterpolation)
      val init = WeededAst.Expr.Cst(Ast.Constant.Str(""), tree.loc)
      var isDebug = false
      // Check for empty interpolation
      if (tryPick(TreeKind.Expr.Expr, tree).isEmpty) {
        val error = EmptyInterpolatedExpression(tree.loc)
        return Validation.toSoftFailure(Expr.Error(error), error)
      }

      Validation.fold(tree.children, init: WeededAst.Expr) {
        // A string part: Concat it onto the result
        case (acc, token@Token(_, _, _, _, _, _)) =>
          isDebug = token.kind == TokenKind.LiteralDebugStringL
          val loc = token.mkSourceLocation()
          val lit0 = token.text.stripPrefix("\"").stripSuffix("\"").stripPrefix("}")
          val lit = if (isDebug) lit0.stripSuffix("%{") else lit0.stripSuffix("${")
          if (lit == "") {
            Validation.success(acc)
          } else {
            val (processed, errors) = Constants.visitChars(lit, loc)
            val cst = Validation.success(Expr.Cst(Ast.Constant.Str(processed), loc)).withSoftFailures(errors)
            mapN(cst) {
              cst => Expr.Binary(SemanticOp.StringOp.Concat, acc, cst, tree.loc.asSynthetic)
            }
          }
        // An expression part: Apply 'toString' to it and concat the result
        case (acc, tree: Tree) if tree.kind == TreeKind.Expr.Expr =>
          mapN(visitExpr(tree))(expr => {
            val loc = tree.loc.asSynthetic
            val funcName = if (isDebug) {
              isDebug = false
              "Debug.stringify"
            } else "ToString.toString"
            val str = Expr.Apply(Expr.Ambiguous(Name.mkQName(funcName), loc), List(expr), loc)
            Expr.Binary(SemanticOp.StringOp.Concat, acc, str, loc)
          })
        // Skip anything else (Parser will have produced an error.)
        case (acc, _) => Validation.success(acc)
      }
    }
`
Just to confirm, the first 2 cases in this method from Weeder2.scala can be removed with no issue?

magnus-madsen:
@mlutze Suggested the cases can be removed, but I would think rather the two uses of StringOp.Concat has to be refactored to use InvokeMethod

mlutze:
> @mlutze Suggested the cases can be removed, but I would think rather the two uses of StringOp.Concat has to be refactored to use InvokeMethod

Yes, agreed. I forgot about String interpolation

saisoma123:
Just refactored with InvokeMethod2

magnus-madsen:
I will take a more detailed look. Could you add your name to `AUTHORS.md` as confirmation that you agree to make your contributions open source. Thanks!

main/src/ca/uwaterloo/flix/language/phase/Deriver.scala
```diff
@@ -901,7 +901,7 @@ object Deriver {
     * Builds a string concatenation expression from the given expressions.
     */
   private def concat(exp1: KindedAst.Expr, exp2: KindedAst.Expr, loc: SourceLocation)(implicit flix: Flix): KindedAst.Expr = {
-    KindedAst.Expr.Binary(SemanticOp.StringOp.Concat, exp1, exp2, Type.freshVar(Kind.Star, loc), loc)
+    KindedAst.Expr.InvokeMethod2(exp1, Name.Ident("concat", loc), List(exp2), Type.freshVar(Kind.JvmConstructorOrMethod, loc), Type.freshVar(Kind.Star, loc), Type.freshVar(Kind.Eff, loc), loc)
```
magnus-madsen, line 904: @mlutze The deriver is post type inference right, so we should have the specific types and effects here, right?

mlutze: Wrong. Deriver is pre-inference.

magnus-madsen:
I added one comment. I suppose this is the reason the tests are failing.

saisoma123:
> I will take a more detailed look. Could you add your name to `AUTHORS.md` as confirmation that you agree to make your contributions open source. Thanks!

Just opened a PR with my name on AUTHORS.md

main/src/ca/uwaterloo/flix/language/phase/Deriver.scala
saisoma123: This is the line causing the tests to fail?

magnus-madsen: @mlutze Any idea why the tests are failing them? I thought it was because of specifying the wrong types.

saisoma123: Just for reference, I was looking at this KindedAST.Expr method     `case class InvokeMethod2(exp: Expr, methodName: Name.Ident, exps: List[Expr], mvar: Type.Var, tvar: Type.Var, evar: Type.Var, loc: SourceLocation) extends Expr
`

mlutze: I believe the tests are failing due to the impurity of Java operations. To solve this, we need the expression to be `unsafe e1.concat(e2)`. The `unsafe` expression is internally called `MaskingCast`, I think, so we just need to wrap that around our concat expression here.

magnus-madsen: Sounds good.

saisoma123: I wrapped the concat expr with `KindedAST.Expr.UncheckedMaskingCast`

mlutze:
The last errors are trickier to deal with. Right now we still have a purity issue with string interpolation. But this time, we cannot wrap it with `unsafe` because that would ignore all the effects of the parameters. E.g. for `e1.concat(e2)` where `e1` has effect `Foo`, that `Foo` should not be erased. 

@magnus-madsen any thoughts? I see these options: 

1. Make a lambda and cast the body. 
2. Add a secret helper to the min lib.
3. Disallow effects in string interpolation.

magnus-madsen:
Ugh. The intention was never for this issue to be so hard... 

Perhaps its better to leave this issue be for now, and to take a stab at something else.

magnus-madsen:
This turned out to be much more difficult than we expected. Thus its probably not even worth it.

PR #8239, opened 2024-08-06
feat: add `struct` support to verifier

dghosef:


main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -231,11 +231,39 @@ object Verifier {
             case _ => failMismatchedShape(t1, "Array", loc)
           }
 
-        case AtomicOp.StructNew(_, _) => throw new RuntimeException("JOE TBD")
+        case AtomicOp.StructNew(sym0, fields0) =>
+          val region :: fieldTpes = ts
```
magnus-madsen, line 235: This pattern match could fail. I suppose thats still okay, but a neater structure might be:

ts match {
  case MonoType.Struct :: _ => // OK
  case _ => Error expected ...

magnus-madsen: I would see if you can cleanup the code a bit :)

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -45,6 +45,15 @@ object Verifier {
     checkEq(decl.tpe, ret, decl.loc)
   }
 
+  private def checkStructType(tpe: MonoType, sym0: Symbol.StructSym, loc: SourceLocation): Unit = {
```
magnus-madsen, line 48: Add a comment and move it down with all the other check functions near the bottom of the file :)

magnus-madsen (approved):

PR #8244, opened 2024-08-06
feat: add `throw` to `DocAst`

dghosef:
#8154

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/dbg/DocAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/dbg/DocAst.scala
@@ -101,8 +101,6 @@ object DocAst {
 
     case class TryCatch(d: Expr, rules: List[(Symbol.VarSym, Class[_], Expr)]) extends Atom
 
-    case class Throw(d: Expr) extends Atom
-
     case class TryWith(d1: Expr, eff: Symbol.EffectSym, rules: List[(Symbol.OpSym, List[Ascription], Expr)]) extends Atom
 
     case class Stm(d1: Expr, d2: Expr) extends LetBinder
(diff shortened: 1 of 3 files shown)


magnus-madsen:
@JonathanStarup You are the expert. If its OK you can approve and merge.

main/src/ca/uwaterloo/flix/language/dbg/DocAstFormatter.scala
```diff
@@ -190,6 +190,10 @@ object DocAstFormatter {
           text("try") +: curly(bodyf) +:
             text("catch") +: curly(rs)
         )
+      case Throw(exp) =>
+        group(
+          text("throw") +: aux(exp, paren = false, inBlock = true)
```
JonathanStarup, line 195: inBlock means that let fx doesn't get {} around it, but it should have (e.g. `throw {let x = 42; x + 1}`) so inBlock should be false.

The group(..) around it all doesn't do anything because text, +:, and, aux doesn't have "open" newlines, so that can also be removed.

Its a bit obscure

JonathanStarup: Oh, but also this will just be deleted by my other comment, so nevermind the details

main/src/ca/uwaterloo/flix/language/dbg/DocAst.scala
```diff
@@ -101,6 +101,8 @@ object DocAst {
 
     case class TryCatch(d: Expr, rules: List[(Symbol.VarSym, Class[_], Expr)]) extends Atom
 
+    case class Throw(d: Expr) extends Atom
```
JonathanStarup, line 104: `throw exp` is just a keyword, so instead of a full case class, just make a `def` like with `Discard` or `Untag`

magnus-madsen (approved):

PR #8249, opened 2024-08-06
feat: add `Throw` support to `PatMatch`, `PredDeps`, `Redundancy`, `Regions`, `Stratifier`, and `CodeHinter`

dghosef:
#8154 

These are all straightforward modifications. I didn't split them up so as to avoid the PR spamming I've been doing, but I would be happy to split them up if that's preferable

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/PatMatch.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/PatMatch.scala
@@ -198,7 +198,7 @@ object PatMatch {
         val ruleExps = rules.map(_.exp)
         (exp :: ruleExps).flatMap(visitExp)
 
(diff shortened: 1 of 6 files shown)


magnus-madsen:
> #8154
> 
> These are all straightforward modifications. I didn't split them up so as to avoid the PR spamming I've been doing, but I would be happy to split them up if that's preferable

For such small changes its fine.

magnus-madsen:
@dghosef We refactored stratifier recently (we are moving away from Validation), so a small change is needed here.

dghosef:
> @dghosef We refactored stratifier recently (we are moving away from Validation), so a small change is needed here.

Got it!

magnus-madsen (approved):

PR #8256, opened 2024-08-07
refactor: refine input sources

magnus-madsen:


magnus-madsen:
@mlutze I did not fully update all comments because I am still exploring exactly how to model this...

main/src/ca/uwaterloo/flix/language/phase/Reader.scala
```diff
@@ -59,7 +63,9 @@ object Reader {
               result += (src -> ())
             }
 
-          case Input.Unknown => throw InternalCompilerException("Impossible to read source code from unknown input.", SourceLocation.Unknown)
+          case Input.FileInPackage(_, _, _) => throw InternalCompilerException("Impossible.", SourceLocation.Unknown)
+
+          case Input.Unknown => throw InternalCompilerException("Impossible.", SourceLocation.Unknown)
```
mlutze, line 68: Not so helpful error messages :)

mlutze (approved):

main/src/ca/uwaterloo/flix/language/phase/Reader.scala
magnus-madsen: I think its because our modelling is wrong and these things should go away.

PR #8257, opened 2024-08-07
feat: configurable max iterations

mlutze:
just for Jonathan


magnus-madsen:
I am very skeptical of this. Why do we need it? If the number is too low either there is a bug or it can be increased.

JonathanStarup:
But we will never guess the correct number? now the user has a way out when they make weird 300 case enums

JonathanStarup:
Its also nice for us for testing typing changes etc. It seems like a nice flexibility that doesn't really cost anything. its also an X flag

magnus-madsen (approved):

magnus-madsen:
I will allow it because its an X flag, so it can be removed at any time.

PR #8259, opened 2024-08-07
refactor: introduce implicit `SecurityContext` parameter

magnus-madsen:


magnus-madsen:
I think I have found the right model. This is just the foundation.

mlutze (approved):

PR #8265, opened 2024-08-08
WIP n-ary enums

mlutze:


magnus-madsen:
What's important to me is that we push this through to the backend such that we have Tag0, Tag1, ...,Tagn savings memory on our representation of Cons.

magnus-madsen:
I am not sure it is worth dealing with restrictable variants... I would only do it if it is easier.

mlutze:
> I am not sure it is worth dealing with restrictable variants... I would only do it if it is easier.

The main complexity is in the back end. I won't be able to deal with that, so I'm closing this PR.

magnus-madsen:
> > I am not sure it is worth dealing with restrictable variants... I would only do it if it is easier.
> 
> The main complexity is in the back end. I won't be able to deal with that, so I'm closing this PR.

Why is that so complicated?

mlutze:
> > > I am not sure it is worth dealing with restrictable variants... I would only do it if it is easier.
> > 
> > 
> > The main complexity is in the back end. I won't be able to deal with that, so I'm closing this PR.
> 
> Why is that so complicated?

Something about monomorph and erasing... Jonathan explained it to me and it sounded like it fell into the "let's redesign monomorph instead" bucket.

JonathanStarup:
> Why is that so complicated?

Our current monomorph tricks for enums are very particular and do not work if you remove the tuple inside the enum

JonathanStarup:
Should be easy if we follow the ticket i just made

JonathanStarup:
@magnus-madsen @mlutze The type of `None` is still `Option[t]` right (and not `Unit -> Option[t]`)?

mlutze:
> @magnus-madsen @mlutze The type of `None` is still `Option[t]` right (and not `Unit -> Option[t]`)? 

Right

PR #8266, opened 2024-08-08
feat: better Datalog safety checking

JonathanStarup:
We have some errors in our safety checking of datalog constraints. It is not so easy to get right


magnus-madsen:
I suppose we should discuss this in detail, so I will mark it as a draft now.

How does your change compare to what is in the paper?

JonathanStarup:
Its almost the same as the paper, but a little more permissive. we can discuss it irl (also because I'm not totally sure my current thinking is correct)

PR #8272, opened 2024-08-09
feat: add structs to simplifier

dghosef:


main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -222,7 +237,7 @@ object Simplifier {
       throw InternalCompilerException(s"Unexpected expression: $exp0.", loc)
   }
 
-  private def visitType(tpe: Type)(implicit flix: Flix): MonoType = {
+  private def visitType(tpe: Type)(implicit flix: Flix, root: MonoAst.Root): MonoType = {
```
magnus-madsen, line 240: We should pass the root as an implicit parameter (everywhere you use it here)

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -21,6 +21,7 @@ import ca.uwaterloo.flix.language.ast.Ast.BoundBy
 import ca.uwaterloo.flix.language.ast.{Purity, Symbol, _}
 import ca.uwaterloo.flix.language.dbg.AstPrinter._
 import ca.uwaterloo.flix.util.{InternalCompilerException, ParOps}
+import ca.uwaterloo.flix.language.phase.unification.Substitution
```
magnus-madsen, line 24: Is this used?

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
dghosef: > We should pass the root as an implicit parameter (everywhere you use it here)

Is that not already what I do?

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
dghosef: > Is this used?

Yea - line 195

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
magnus-madsen: Doh, my bad. I was confused by the order. We always go from most specific to least specific, so it should be root: .., flix: Flix.

This is just by convention, but useful when you have many implicit parameters.

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
magnus-madsen: I am a bit confused. Is this doing some job that perhaps should have been done in the Monomorpher?

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -48,13 +50,26 @@ object Simplifier {
       SimplifiedAst.Def(spec.ann, spec.mod, sym, fs, e, retType, eff, sym.loc)
   }
 
-  private def visitEffect(decl: MonoAst.Effect)(implicit flix: Flix, universe: Set[Symbol.EffectSym]): SimplifiedAst.Effect = decl match {
+  private def visitEffect(decl: MonoAst.Effect)(implicit flix: Flix, universe: Set[Symbol.EffectSym], root: MonoAst.Root): SimplifiedAst.Effect = decl match {
```
magnus-madsen, line 53: I dont know why the arguments here are in the wrong order: I would put them:

universe, root, flix

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
dghosef: I don't think so because the `Struct` type doesn't carry around field types. It only carries around tparams. It is only when we convert to monotype that we do the substitution - hence, in the simplifier.

magnus-madsen: OK; I am still not 100%, but you can explain to me on monday?

magnus-madsen (approved):

PR #8279, opened 2024-08-09
feat: add resolution error checking for struct fields declared out of order

dghosef:


magnus-madsen:
Ah- We should discuss this one. Should be fun.

This is the eternal question of how much style the compiler should enforce :)

(Ill mark as draft until we have discussed.)

dghosef:
Sounds good. 

The big benefit of this is it allows me to not have to do the sorting thing and it simplifies codegen a lot

magnus-madsen:
> Sounds good.
> 
> The big benefit of this is it allows me to not have to do the sorting thing and it simplifies codegen a lot

Ah yes, the implementers dream :)

magnus-madsen:
Where you sorting the fields with expressions?

What would `new S { b = println("first"); a = println("second") }` print with your sorting?

JonathanStarup:
with this pr we dont have to rearrange them, because they are checked to be in the declared order

magnus-madsen:
My question was whether we are changing the evaluation order.

dghosef:
No we aren't. Which is why I am enforcing the original ordering

I don't think let binding would be too hard to implement if we didn't enforce this however

magnus-madsen:
So your point is: by enforcing the order there is also not confusion or danger of _evaluation_ order.

dghosef:
Yes

main/src/ca/uwaterloo/flix/language/errors/ResolutionError.scala
```diff
@@ -1183,6 +1183,27 @@ object ResolutionError {
          |""".stripMargin
     }
   }
+  /**
+    * An error raised to indicate a `new` struct expression declares fields in the wrong order
```
magnus-madsen, line 1187: -> initializes its fields ...

main/src/ca/uwaterloo/flix/language/errors/ResolutionError.scala
```diff
@@ -1183,6 +1183,27 @@ object ResolutionError {
...
+    *
+    * @param providedFields the order in which fields were declared
+    * @param expectedFields the order in which fields were expected to be declared
+    * @param loc the location where the error occurred
+    */
+  case class WrongStructFieldOrdering(struct: Symbol.StructSym, providedFields: List[Name.Label], expectedFields: List[Name.Label], loc: SourceLocation) extends  ResolutionError with Recoverable {
```
magnus-madsen, line 1193: WrongStructFieldOrdering ->  IllegalNewStruct

main/src/ca/uwaterloo/flix/language/errors/ResolutionError.scala
```diff
@@ -1183,6 +1183,27 @@ object ResolutionError {
...
+    override def summary: String = s"Incorrect ordering of struct field declarations in initializer of struct `$struct`"
```
magnus-madsen, line 1194: -> Structs fields must be initialized in the same order as they are declared.

main/src/ca/uwaterloo/flix/language/errors/ResolutionError.scala
```diff
@@ -1183,6 +1183,27 @@ object ResolutionError {
...
+
+    def message(formatter: Formatter) : String = {
+      import formatter._
+      s""">> Incorrect ordering of struct field declarations in initializer of struct `$struct`
```
magnus-madsen, line 1198: Structs fields must be initialized in the same order as they are declared.

main/src/ca/uwaterloo/flix/language/errors/ResolutionError.scala
```diff
@@ -1183,6 +1183,27 @@ object ResolutionError {
...
+         |
+         |The struct $struct expected order '${expectedFields.mkString(",")}'
```
magnus-madsen, line 1200: Expected initialization order:
Actual initialized order:

main/src/ca/uwaterloo/flix/language/errors/ResolutionError.scala
```diff
@@ -1183,6 +1183,27 @@ object ResolutionError {
...
+         |but '${providedFields.mkString(",")}' was provided
+         |
+         |${code(loc, "incorrect field ordering")}
```
magnus-madsen, line 1203: -> incorrect order

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -1356,12 +1356,19 @@ object Resolver {
...
                 case st =>
-                  val providedFieldNames = fields.map { case (k, _) => k }.toSet
-                  val expectedFieldNames = st.fields.map(_.name).toSet
+                  val providedFieldNames = fields.map { case (k, _) => k }
+                  val expectedFieldNames = st.fields.map(_.name)
                   val extraFields = providedFieldNames.diff(expectedFieldNames)
                   val missingFields = expectedFieldNames.diff(providedFieldNames)
 
-                  val errors = extraFields.map(ResolutionError.ExtraStructField(st0.sym, _, loc)) ++ missingFields.map(ResolutionError.MissingStructField(st0.sym, _, loc))
+                  val errors = if (!extraFields.isEmpty || !missingFields.isEmpty) {
```
magnus-madsen, line 1364: I understand the logic here and I agree its correct. But it is still a bit hard to follow.

I wonder if we could spell it out:

if (expected == actual) {
  // all god
 } else {
  // Three things can be wrong:
  // Case 1: Missing Fields
  if (missingFields.nonEmpty) ...
  // Case 2: Extra fields
  if (extrafields.nonEmpty) { ... }

}

magnus-madsen: OK; I am not sure. Maybe you can have a go at it ?

magnus-madsen: I suppose a tricky thing is that you want to report both extra and missing fields at the same time.

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -1356,12 +1356,19 @@ object Resolver {
             case Result.Ok(st0) =>
               flatMapN(getStructIfAccessible(st0, ns0, loc)) {
                 case st =>
-                  val providedFieldNames = fields.map { case (k, _) => k }.toSet
-                  val expectedFieldNames = st.fields.map(_.name).toSet
+                  val providedFieldNames = fields.map { case (k, _) => k }
+                  val expectedFieldNames = st.fields.map(_.name)
                   val extraFields = providedFieldNames.diff(expectedFieldNames)
                   val missingFields = expectedFieldNames.diff(providedFieldNames)
```
magnus-madsen, line 1362: We typically use -- instead of diff. (if it works)

magnus-madsen (approved):

PR #8280, opened 2024-08-09
feat: support for empty interfaces in subtyping

chanattan:
Related to #8212.
@magnus-madsen 
I was forced implictly to check for explicit interfaces with no super-interfaces. As otherwise it would create duplicate methods, e.g., toString, in candidate methods with interfaces that have redefined it.
It could be done in other ways but this is a fast way.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
@@ -219,8 +219,26 @@ object TypeReduction {
    * Returns a JavaMethodResolutionResult either containing the Java method or a MethodNotFound object.
    */
   private def retrieveMethod(clazz: Class[_], methodName: String, ts: List[Type], isStatic: Boolean = false, loc: SourceLocation)(implicit flix: Flix): JavaMethodResolutionResult = {
-    val objectMethods = if (clazz.isInterface && clazz.getInterfaces.isEmpty) classOf[java.lang.Object].getMethods.toList else Nil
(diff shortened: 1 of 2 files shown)


chanattan:
Ok I need to fix something.

chanattan:
Since there may be interfaces, e.g., `java.util.Map` that have no superinterfaces and re-implement some methods from Object I have added this short fix that is applied in certain cases, considering performance improvement hopefully.

magnus-madsen:
Can we add some test cases that trigger this code? Thanks!

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
```diff
@@ -219,8 +219,26 @@ object TypeReduction {
    * Returns a JavaMethodResolutionResult either containing the Java method or a MethodNotFound object.
    */
   private def retrieveMethod(clazz: Class[_], methodName: String, ts: List[Type], isStatic: Boolean = false, loc: SourceLocation)(implicit flix: Flix): JavaMethodResolutionResult = {
-    // NB: this considers also static methods
-    val candidateMethods = clazz.getMethods.filter(m => isCandidateMethod(m, methodName, isStatic, ts))
+    def hasSimilarSignature(m1: Method, m2: Method): Boolean =
+      m1.getName.equals(m2.getName) && (m1.getParameterTypes zip m2.getParameterTypes).forall{ case (c1, c2) => c1 == c2 }
+
+    val candidateMethods =
```
magnus-madsen, line 225: I would extract this val into a separate helper method and document it.

PR #8283, opened 2024-08-09
feat: added struct tests to `TestSafety`

dghosef:


magnus-madsen:
Merge now or later?

dghosef:
Now I think

magnus-madsen (approved):

PR #8284, opened 2024-08-09
feat: add struct support to `EffectBinder`, `Inliner`, `LambdaLift`, `OccurrenceAnalyzer` and `Monomorpher`

dghosef:
All the additions to the phases are similar in nature


JonathanStarup:
Sorry maybe im just out of the loop, but why do the declaration still exist after changing it to simple structs with a list of fields?

dghosef:
We need it in `Simplifier` when  determining monotypes. Though I guess we can remove after that - Ill do that

magnus-madsen (approved):

Issue #6270, opened 2023-07-19
Upgrade to Scala 3


magnus-madsen:
- Get it to compile with minimal changes.
- Ensure all Flix tests pass.
- Find out if Scala3 compiler is faster/slower.
- Find out the produced flix.jar is smaller/larger.
- Find out if the runtime performance of the Flix compiler is faster/slower.

jaschdoc:
Close this and reopen whenever it is relevant again to reduce the number non-actionable issues?

PR #8286, opened 2024-08-10
chore: migrate to Scala 3

jaschdoc:
Closes https://github.com/flix/flix/issues/6270


jaschdoc:
Use regex `enum[^sS0 \.]` to search for enum occurrences

jaschdoc:
@magnus-madsen seems feasible. The only broken thing is the `jline` dependency where (as far as I can tell from a quick Google search) they have removed some type information in the class files which Scala 3 now complains about. The solution (again from the Google search) seems to be finding a different jar / version on Maven with the type information included.

magnus-madsen:
Can you try that? Including a different JLine from Maven should be easy.

magnus-madsen:
I don't plan to upgrade just yet. Many other things have to be decided.

But we can merge a PR where the codebase is updated.

magnus-madsen:
At least:

1. Get it to compile with minimal changes.
2. Ensure all Flix tests pass.
3. Find out if Scala3 compiler is faster/slower.
4. Find out the produced flix.jar is smaller/larger.
5. Find out if the runtime performance of the Flix compiler is faster/slower.

jaschdoc:
Roger

main/test/ca/uwaterloo/flix/util/FlixSuite.scala
```diff
@@ -90,10 +91,12 @@ class FlixSuite(incremental: Boolean) extends AnyFunSuite {
             // Evaluate the function.
             val result = run()
             // Expect the true value, if boolean.
-            if (result.isInstanceOf[java.lang.Boolean]) {
-              if (result != true) {
-                fail("Expected true, but got false.")
-              }
+            result match {
+              case res: java.lang.Boolean =>
+                if (res != true) {
```
magnus-madsen, line 95: Do you not need to unbox here?

jaschdoc: I think it's automatic... But I may be wrong? I can put in an else-branch to check :)

magnus-madsen: It just to be safe. I think res.value() or something should make it explicit.

jaschdoc:
Fyi, I have checked that bad changes to the Flix files actually result in test failures.

jaschdoc:
**Scala 3 (8f0f56cc49197f5e157230ae895604941b1b02bb)**
```
time ./gradlew jar

BUILD SUCCESSFUL in 1m 41s
3 actionable tasks: 3 executed

real    1m42,027s
user	0m1,657s
sys	0m0,197s

63M build/libs/flix.jar

Throughput (best): 14,700 lines/sec (with 8 threads.)

  min:  3,365, max: 14,700, avg: 10,002, median: 10,419

  The highest throughput was in iteration: 9 (out of 10).
  The ratio between the best and worst iteration was: 4.4x.

Finished 10 iterations on 66,420 lines of code in 79 seconds.
```

**Scala 2 (main 95bc4a5f05077df027ccc729be23ce859567f91a)**
```
time ./gradlew jar

BUILD SUCCESSFUL in 2m 27s
3 actionable tasks: 3 executed

real    2m28,387s
user	0m1,727s
sys	0m0,283s


36M build/libs/flix.jar

Throughput (best): 14,013 lines/sec (with 8 threads.)

  min:  2,906, max: 14,013, avg: 10,967, median: 12,457

  The highest throughput was in iteration: 9 (out of 10).
  The ratio between the best and worst iteration was: 4.8x.

Finished 10 iterations on 66,420 lines of code in 74 seconds.
```

jaschdoc:
Maybe inlining results in smaller jar and faster performance

magnus-madsen:
> Maybe inlining results in smaller jar and faster performance

Yes to performance, but not for JAR size.

jaschdoc:
Closing this PR for now. We can always refer to this if we decide to upgrade.

PR #8287, opened 2024-08-10
refactor: update files to be compatible with Scala 3

jaschdoc:
Related to https://github.com/flix/flix/pull/8286


magnus-madsen:
@mlutze @JonathanStarup Do you see any changes you do not "agree with"? I.e. feature loss that is unacceptable?

magnus-madsen:
@jaschdoc The use of implicit apply is gone it seems? @JonathanStarup was arguing to have that in Flix...

JonathanStarup:
hmm i think thats a simplified summary 🤔

main/src/ca/uwaterloo/flix/language/dbg/DocAstFormatter.scala
```diff
@@ -22,6 +22,7 @@ import ca.uwaterloo.flix.language.dbg.DocAst.Expr._
 import ca.uwaterloo.flix.language.dbg.DocAst._
 
 import scala.annotation.tailrec
+import scala.collection.immutable.::
```
mlutze, line 25: Seems redundant

main/src/ca/uwaterloo/flix/language/dbg/DocAstFormatter.scala
```diff
@@ -324,9 +325,9 @@ object DocAstFormatter {
       case arrow@Type.Arrow(_, _) =>
         val (curriedArgs, res) = collectArrowTypes(arrow)
         val formattedArgs = curriedArgs.map {
-          case ts@(Type.Tuple(_) :: Nil) =>
+          case ts@Type.Tuple(_) :: Nil =>
```
mlutze, line 328: So `::` has higher precedence than `@`? This confuses me.

JonathanStarup:
but it looks fine, its mostly the editor behavior thats worrying i think. but maybe its fine since we dont use fancy features

main/src/ca/uwaterloo/flix/language/dbg/DocAstFormatter.scala
magnus-madsen, line 25: Wait, this should not be necessary?

main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
```diff
@@ -200,7 +200,7 @@ object Reducer {
     * Returns all types contained in the given `Effect`.
     */
   private def typesOfEffect(e: Effect): Set[MonoType] = e.ops.toSet.map {
-    op: Op =>
+    (op: Op) =>
```
magnus-madsen, line 203: Does it work without parens, but with braces?

mlutze (commented):
I'm not concerned about these changes specifically, but I am concerned about Scala 3. In my experience, IntelliJ support is weaker and I am strongly opposed to indentation-sensitive syntax.

magnus-madsen:
> I'm not concerned about these changes specifically, but I am concerned about Scala 3. In my experience, IntelliJ support is weaker and I am strongly opposed to indentation-sensitive syntax.

Agreed. We will not switch without a major discussion. But if we can be compatible then we can at least check it out.

I would also wait some years before switching when tooling has caught up / the language has stabilized (?)

jaschdoc:
Btw all of these changes were required by Scala and not IntelliJ. I ran the gradle from the terminal

jaschdoc:
I think the explicit apply method goes away when we move to enums instead of case classes

main/src/ca/uwaterloo/flix/language/dbg/DocAstFormatter.scala
jaschdoc: Scala couldn't figure out that this belonged to List and not DocAst

main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
jaschdoc: Can try

main/src/ca/uwaterloo/flix/language/dbg/DocAstFormatter.scala
jaschdoc: See previous comment

main/src/ca/uwaterloo/flix/language/dbg/DocAstFormatter.scala
jaschdoc: Scala complained that the parentheses were redundant

magnus-madsen: Another reason not to use @

main/src/ca/uwaterloo/flix/language/dbg/DocAstFormatter.scala
magnus-madsen: Can we rename our Cons to FCons?

main/src/ca/uwaterloo/flix/language/dbg/DocAstFormatter.scala
mlutze: What was the first reason?

magnus-madsen: I dont like it :)

magnus-madsen:
@jaschdoc There are two things to fix/try to fix and then we can merge.

mlutze:
> @jaschdoc There are two things to fix/try to fix and then we can merge.

Sorry :P

magnus-madsen:
Is there a flag for scala 2.13 that enables forward compatability?

jaschdoc:
> Is there a flag for scala 2.13 that enables forward compatability?

I can look into it but they have a version of Scala 3 (which I think I'm using) intended for Scala 2 migration

main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
jaschdoc: It's already inside a call to `map` with braces.

jaschdoc: I can drop the type annotation, though.

magnus-madsen:
The Scala 2.13 compiler issues helpful migration warnings with the -Xsource:3 flag.

Before moving to the Scala 3 compiler, it’s recommended to enable this flag in Scala 2 and address the new warnings.

https://docs.scala-lang.org/scala3/guides/migration/tooling-scala2-xsource3.html

main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
jaschdoc: Nevermind, the type annotation was required, but I moved it to separate function.

jaschdoc:
There are more changes incoming (the Scala 3 compiler was not quite satisfied). I will add those to this PR and then resolve the errors from the Scala 2 compiler in migration mode.

magnus-madsen:
We definitely don't want significant whitespace. The goal here is to find the minimum number of changes.

magnus-madsen:
Hopefully this PR can be kept less than 500 lines. Otherwise we need to rethink.

magnus-madsen:
@JonathanStarup @mlutze let us all review it to: (1) learn about the changes and (2) see if we are unhappy with anything.

If you have no comment, you can just post a thumbs-up.

main/src/ca/uwaterloo/flix/api/lsp/CodeActionContext.scala
```diff
@@ -17,7 +17,8 @@ package ca.uwaterloo.flix.api.lsp
 
 import ca.uwaterloo.flix.util.Result
 import ca.uwaterloo.flix.util.Result.{Err, Ok}
-import org.json4s.JsonAST._
+import org.json4s.jvalue2monadic
+import org.json4s.JsonAST.*
```
magnus-madsen, line 21: Do we have to use *? I think we should stick with _ to remain Scala 2.0

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -2350,7 +2351,7 @@ object Resolver {
       // Case 1: Exactly one match. Success.
       case caze :: _ => Result.Ok(caze)
       // Case 2: Multiple matches. Error
-      case cazes => throw InternalCompilerException(s"unexpected duplicate tag: '$qname'.", qname.loc)
+      case null => throw InternalCompilerException(s"unexpected duplicate tag: '$qname'.", qname.loc)
```
magnus-madsen, line 2354: Something is wrong here!

magnus-madsen: The original case should have been: caze :: Nil not caze :: _

magnus-madsen: (Sidenote: We never use null. )

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -2368,7 +2369,7 @@ object Resolver {
       // Case 1: Exactly one match. Success.
       case st :: _ => Result.Ok(st)
       // Case 2: Multiple matches. Error
-      case sts => throw InternalCompilerException(s"unexpected duplicate struct: '$qname'.", qname.loc)
+      case null => throw InternalCompilerException(s"unexpected duplicate struct: '$qname'.", qname.loc)
```
magnus-madsen, line 2372: same comment

main/src/ca/uwaterloo/flix/language/phase/unification/CaseSetUnification.scala
```diff
@@ -235,12 +235,12 @@ object CaseSetUnification {
...
     case Cst(syms) =>
       val lits: Set[Nnf] = syms.map(sym => Nnf.Singleton(Literal.Positive(Atom.Case(sym))))
-      lits.reduceLeftOption(Nnf.Union).getOrElse(Nnf.Empty)
+      lits.reduceLeftOption(Nnf.Union.apply).getOrElse(Nnf.Empty)
     case Var(sym) => Nnf.Singleton(Literal.Positive(Atom.Var(sym)))
     case Not(tpe) => nnfNot(tpe)
     case Or(tpe1, tpe2) => Nnf.Union(nnf(tpe1), nnf(tpe2))
     case And(tpe1, tpe2) => Nnf.Intersection(nnf(tpe1), nnf(tpe2))
-    case _ => throw InternalCompilerException(s"unexpected type: $t", SourceLocation.Unknown)
+    case null => throw InternalCompilerException(s"unexpected type: $t", SourceLocation.Unknown)
```
magnus-madsen, line 243: This does not make sense to me. Is the pattern already exhaustive?

main/src/ca/uwaterloo/flix/language/phase/unification/CaseSetUnification.scala
```diff
@@ -260,7 +260,7 @@ object CaseSetUnification {
       nnf(mkNot(tpe1)),
       nnf(mkNot(tpe2))
     )
-    case _ => throw InternalCompilerException(s"unexpected type: $t", SourceLocation.Unknown)
+    case null => throw InternalCompilerException(s"unexpected type: $t", SourceLocation.Unknown)
```
magnus-madsen, line 263: Same comment.

main/src/ca/uwaterloo/flix/tools/SocketServer.scala
```diff
@@ -23,10 +23,11 @@ import ca.uwaterloo.flix.util._
 import org.java_websocket.WebSocket
 import org.java_websocket.handshake.ClientHandshake
 import org.java_websocket.server.WebSocketServer
-import org.json4s.JsonAST._
+import org.json4s.JsonAST.*
+import org.json4s.jvalue2monadic
 import org.json4s.ParserUtil.ParseException
 import org.json4s.native.JsonMethods
-import org.json4s.native.JsonMethods._
+import org.json4s.native.JsonMethods.*
```
magnus-madsen, line 30: Revert the two uses of *.

main/src/ca/uwaterloo/flix/tools/Tester.scala
```diff
@@ -148,7 +148,7 @@ object Tester {
             terminal.flush()
             finished = true
 
-          case _ => // nop
+          case null => // nop
```
magnus-madsen, line 151: I am confused again.

main/test/ca/uwaterloo/flix/util/FlixSuite.scala
```diff
@@ -21,7 +21,7 @@ import ca.uwaterloo.flix.runtime.{CompilationResult, TestFn}
 import org.scalatest.funsuite.AnyFunSuite
 
 import java.nio.file.{Files, Path, Paths}
-import scala.jdk.CollectionConverters._
+import scala.jdk.CollectionConverters.*
```
magnus-madsen, line 24: Undo

magnus-madsen (commented):
I added some comments. Two things standout:

- Let us not use * for imports
- Strange things are going on with null

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
jaschdoc: Hmm... true, but the Scala 3 compiler said that `cazes` was unreachable unless the match exp were `null`

main/src/ca/uwaterloo/flix/language/phase/unification/CaseSetUnification.scala
jaschdoc: Unreachable again as above

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
magnus-madsen: Here the original code is broken and should be fixed, i.e. the match before the wildcard should be caze :: Nil

main/src/ca/uwaterloo/flix/language/phase/unification/CaseSetUnification.scala
magnus-madsen: In this case then remove the throw

jaschdoc:
> We definitely don't want significant whitespace. The goal here is to find the minimum number of changes.

Funny thing is, if you disable significant whitespace it complains that things aren't formatted properly :thinking:

magnus-madsen:
> > We definitely don't want significant whitespace. The goal here is to find the minimum number of changes.
> 
> Funny thing is, if you disable significant whitespace it complains that things aren't formatted properly 🤔

WHAT?

jaschdoc:
> > > We definitely don't want significant whitespace. The goal here is to find the minimum number of changes.
> 
> > 
> 
> > Funny thing is, if you disable significant whitespace it complains that things aren't formatted properly 🤔
> 
> 
> 
> WHAT?

Haha yes then white space doesn't matter but it's still opinionated as to how it should be formatted

magnus-madsen:
As in it fails compilation?

magnus-madsen:
Should we try to get this merged in? We can do Resolver in a subsequent PR if that helps.

jaschdoc:
> As in it fails compilation?

I believe it's a warning, but we turn warnings into errors

magnus-madsen:
RFM?

magnus-madsen:
Wait. We need to deal with the case null =>

I don't want them. Can the cases be deleted or reverted back to _?

jaschdoc:
Still not ready. Just fixing up the last tidbits. I'll let you know :)

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -2362,9 +2363,9 @@ object Resolver {
       // Case 0: No matches. Error.
       case Nil => Result.Err(ResolutionError.UndefinedStruct(qname, qname.loc))
       // Case 1: Exactly one match. Success.
-      case st :: _ => Result.Ok(st)
+      case st :: Nil => Result.Ok(st)
```
jaschdoc, line 2366: Either we have discovered a bug in the compiler or the comments do not match what is actually intended here.

jaschdoc:
@magnus-madsen I think I need your help on the comment above

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -2977,7 +2978,7 @@ object Resolver {
       case Some(prefix) =>
         val ns = prefix ::: qname0.namespace.parts.tail
         val qname = Name.mkQName(ns, qname0.ident.name, SourceLocation.Unknown)
-        root.symbols.getOrElse(qname.namespace, Map.empty).get(qname.ident.name)
+        root.symbols.getOrElse(qname.namespace, Map.empty).get(qname.ident.name).map(_.distinct)
```
jaschdoc, line 2981: This change may be significant?

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -2949,20 +2950,20 @@ object Resolver {
...
           Nil
         }
       }
 
       // 4th priority: names in the root namespace
-      val rootNames = root.symbols.getOrElse(Name.RootNS, Map.empty).getOrElse(qname.ident.name, Nil).map(Resolution.Declaration)
+      val rootNames = root.symbols.getOrElse(Name.RootNS, Map.empty).getOrElse(qname.ident.name, Nil).map(Resolution.Declaration.apply)
 
-      envNames ::: localNames ::: currentNamespace ::: rootNames
+      (envNames ::: localNames ::: currentNamespace ::: rootNames).distinct
```
jaschdoc, line 2962: This change may be significant?

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
magnus-madsen: Why the extra map?

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
magnus-madsen: Same question: Why distinct?

jaschdoc: I inspected the results and the code above finds duplicate resolutions, e.g., it will find `Option.None` multiple times. However, the caller(s) of this function assumes that only one lookup should be returned if successful. This is (obviously) either intentional or a bug in this function but it seems like once we have found a resolution we shouldn't return it multiple times. Maybe it should be a set instead?

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
jaschdoc: Same as the other comment but for consistency. I did not observe any change in behavior on this line though so this could be removed maybe?

magnus-madsen:
We should ask @mlutze , but my first reaction is to leave the code unchanged: If there is a bug, we can find and fix it later.

Ideally, this PR should be a pure refactor.

jaschdoc:
> We should ask @mlutze , but my first reaction is to leave the code unchanged: If there is a bug, we can find and fix it later.
> 
> Ideally, this PR should be a pure refactor.

Ok sounds sensible :+1:

mlutze:
I agree that this should just be a pure refactor.

The distinct thing should not make a difference. The functions calling it should only be getting the first one they find, so it doesn't matter if there are duplicates. Any meaningful duplicates should have already been checked by the Namer.

jaschdoc:
Ah okay then the comments at the call sites are wrong

jaschdoc:
Alright let's try this again :smile: 
Matt convinced me, so I have removed the changes to Resolver and just kept it to a simple refactor :)

jaschdoc:
> Can we do those changes to lookup in another PR? Even if it _should_ be the same, it's better to be able to scrutinize those separately.

What do you mean? I only added the comment for clarity. The rest is specific to Scala 3 compatibility.

mlutze:
In commit aab42f6a0c89d0cf7eafdf54a41bf8f81b8d938d there appear to be semantic changes, no?

jaschdoc:
> In commit [aab42f6](https://github.com/flix/flix/commit/aab42f6a0c89d0cf7eafdf54a41bf8f81b8d938d) there appear to be semantic changes, no?

Added comments to the changes so you can see (via the link) what I've changed

magnus-madsen (approved):

PR #8291, opened 2024-08-12
feat: add `struct` support to `ConstraintGen` and fix `TypedAst` and `AtomicOp`

dghosef:
This PR:
- Changes `AtomicOp` to also have the index of the field. This simplifies codegen later down the line
  - changes `fields` of `TypedAst` to be a `Map` and adds an `idx` value to make this more efficient
- Adds `struct` support to `ConstraintGen` and adds tests
- Adjust `Typer` accordingly

This PR is relatively big, but the bulk of the interesting changes are in `ConstraintGen`


magnus-madsen:
@mlutze If you do first pass then I can do second pass :)

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -1228,4 +1278,17 @@ object ConstraintGen {
     }
   }
 
+  private def instantiateStruct(sym: Symbol.StructSym, structs: Map[Symbol.StructSym, KindedAst.Struct])(implicit flix: Flix) : (Map[Name.Label, Type], Type, Type.Var) = {
```
magnus-madsen, line 1281: I think some doc comments plus a concrete example would be helpful.

We often add some examples when something unusual is going on.

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -1228,4 +1278,17 @@ object ConstraintGen {
...
+    val struct = structs(sym)
+    assert(struct.tparams.last.sym.kind == Kind.Eff)
+    val fields = struct.fields
+    val (_, _, tpe, substMap) = Scheme.instantiate(struct.sc, struct.loc)
+    val subst = Substitution(substMap)
+    val instantiatedFields = fields.map( f => f match {
```
magnus-madsen, line 1287: extra space

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -1228,4 +1278,17 @@ object ConstraintGen {
...
+      case KindedAst.StructField(fieldSym, tpe, _) =>
+        fieldSym -> subst(tpe)
+    }
```
magnus-madsen, line 1290: move next paren on this line.

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -574,13 +575,62 @@ object ConstraintGen {
...
-        throw new RuntimeException("joe tbd")
+        val resTpe = tvar
+        val resEff = evar
+        (resTpe, resEff)
+
+      case Expr.StructPut(sym, exp1, field, exp2, tvar, evar, loc) =>
+        val (instantiatedFieldTpes, structTpe, regionVar) = instantiateStruct(sym, root.structs)
+        // lhs struct type
+        val (tpe1, eff1) = visitExp(exp1)
+        c.expectType(structTpe, tpe1, exp1.loc)
```
magnus-madsen, line 621: As a style choice, I believe we tend to do all the recursive calls before we start expectingTypes

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -574,13 +575,62 @@ object ConstraintGen {
...
 
       case Expr.StructNew(sym, fields, region, tvar, evar, loc) =>
-        throw new RuntimeException("joe tbd")
+        val (instantiatedFieldTpes, structTpe, regionVar) = instantiateStruct(sym, root.structs)
+        val visitedFields = fields.map { case (k, v) => visitExp(v) }
+        val (fieldTpes, fieldEffs) = visitedFields.unzip
+        // struct type
+        c.unifyType(tvar, structTpe, loc)
+        // field types
+        fields.zip(fieldTpes).foreach {
```
magnus-madsen, line 584: I think a normal for-loop is fine.

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -574,13 +575,62 @@ object ConstraintGen {
...
+          case ((fieldSym, expr), fieldTpe1) =>
+            instantiatedFieldTpes.get(fieldSym) match {
+              case None => () // Field does not exist. A resolution error will already have been generated
```
magnus-madsen, line 587: either just do the get directly or throw InternalCompilerException.

Here I think you can just directly apply instantiatedFieldTpes

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -574,13 +575,62 @@ object ConstraintGen {
...
+              case Some(fieldTpe2) => c.unifyType(fieldTpe1, fieldTpe2, expr.loc)
+            }
+        }
+        // region type
+        val regionType = Type.mkRegion(regionVar, loc)
+        val (regionTpe, regionEff) = visitExp(region)
+        c.unifyType(regionType, regionTpe, region.loc)
+        // effect
```
magnus-madsen, line 595: Drop this comment (and similar comments like "overall effect".

magnus-madsen: I think, overall, comments should either be more descriptive (longer) or not there at all.

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
dghosef: The reason I do this is that it is "valid" to get to this point with the field not existing, since the resolver generates a recoverable error when the field does not exist. Thus, if we have `new S {a = 3} @ r` if `a` does not exist, we would get to the constraint generation and don't want to crash the compiler

magnus-madsen: Oh, I see.

magnus-madsen: I think, where we detect the error, you should drop the extra field.

magnus-madsen: Is that easily doable or more cumbersome?

dghosef: I think that's probably not too hard, but what if we have
`new S {a = "hello" + 3} @ r`

Wouldn't we want `a` to persist to the typer?

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
dghosef: Sorry, what do you mean by next paren?

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
magnus-madsen: OK, I agree with your points.

Now turn those ideas into a comment or two :)

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
magnus-madsen: })

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
dghosef: What are your thoughts on this comment?

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
dghosef: I added a comment - I can separate them up into 2 comments(one at unify and one at the `visitExp` if that would be better

magnus-madsen:
@dghosef I will let @mlutze have a look and report back 👍 

What remains after this? Some CodeGen and more?

dghosef:
GenExpression and enabling all the tests are what's left to get the current iteration functional.

In addition,
- We were going to talk about the use of `Substiution` in the simplifier - I don't think it is a correctness issue but you know better than I, and it definitely could be a style issue
- I should also maybe add deriver support at some point
- More a tooling thing, but the indexer doesn't get struct names or fields yet

And of course, syntax, `mut`/`pub`/`const`/better error recovery,

main/src/ca/uwaterloo/flix/language/ast/AtomicOp.scala
```diff
@@ -62,9 +62,9 @@ object AtomicOp {
 
   case class StructNew(sym: Symbol.StructSym, fields: List[Name.Label]) extends AtomicOp
 
-  case class StructPut(sym: Symbol.StructSym, field: Name.Label) extends AtomicOp
+  case class StructPut(sym: Symbol.StructSym, idx: Int, field: Name.Label) extends AtomicOp
```
mlutze, line 65: Remind me, why do we not have a FieldSym? I would think the index could be part of the sym.

mlutze: Jonathan agrees.

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -574,13 +575,52 @@ object ConstraintGen {
...
+        val visitedFields = fields.map { case (k, v) => visitExp(v) }
+        val (regionTpe, regionEff) = visitExp(region)
+        val (fieldTpes, fieldEffs) = visitedFields.unzip
+        c.unifyType(tvar, structTpe, loc)
+        for {
+          (fieldSym, expr) <- fields
+          fieldTpe1 <- fieldTpes
+        } {
+          instantiatedFieldTpes.get(fieldSym) match {
+            case None => ()
```
mlutze, line 590: What does it mean if this is None?

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -574,13 +575,52 @@ object ConstraintGen {
...
+            case Some(fieldTpe2) => c.unifyType(fieldTpe1, fieldTpe2, expr.loc)
+          }
+        }
+        val regionType = Type.mkRegion(regionVar, loc)
```
mlutze, line 594: regionType vs regionTpe is a bit confusing

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -574,13 +575,52 @@ object ConstraintGen {
...
+        c.unifyType(regionType, regionTpe, region.loc)
+        c.unifyType(evar, Type.mkUnion(Type.mkUnion(fieldEffs, loc), regionEff, regionVar, loc), loc)
```
mlutze, line 596: You can make one big union instead of `Type.mkUnion` twice

main/src/ca/uwaterloo/flix/language/ast/AtomicOp.scala
dghosef: Yea, I am currently working on a PR that adds back the fieldsym. I think, though, that would be out of the scope of this one

magnus-madsen:
@mlutze RFM?

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
magnus-madsen: Looks good.

magnus-madsen (approved):

magnus-madsen:
I merged it. Yell if unhappy :)

PR #8298, opened 2024-08-12
refactor: enforce exception typing rules

saisoma123:
I added a check to see if exceptions in a try catch block are in correct order, i.e., an exception like FileNotFoundException cannot be after IOException, as FileNotFoundException is a subtype of IOException.


magnus-madsen:
Instead of throwing an exception, you need to create your own compiler error, see e.g. `NonExhaustiveMatchError`

saisoma123:
> Instead of throwing an exception, you need to create your own compiler error, see e.g. `NonExhaustiveMatchError`

I get what you're saying, but the `visitExp `method returns `List[NonExhaustiveMatchError]`, so wouldn't I have to return that in `case Expr.TryCatch`, as opposed to my own created error like `ExceptionTypeError`. I'm a bit confused.

magnus-madsen:
> > Instead of throwing an exception, you need to create your own compiler error, see e.g. `NonExhaustiveMatchError`
> 
> I get what you're saying, but the `visitExp `method returns `List[NonExhaustiveMatchError]`, so wouldn't I have to return that in `case Expr.TryCatch`, as opposed to my own created error like `ExceptionTypeMatchError`. I'm a bit confused.

I propose to introduce a PatMatchError trait and then have both errors extend that.

magnus-madsen:
Take a look at how e.g. TypeError is organized.

saisoma123:
Hi Magnus, I created a `PatMatchError` trait like you said, should I keep the trait sealed and have both `ExceptionTypeMatchError` and `NonExhaustiveMatchError` within the trait like `TypeError`? Or just keep the errors in separate files and have them extend with the trait. With the first option I think it would break the `TestPatExhaustiveness` test is the issue, that would have to be refactored to use `PatMatchError.NonExhaustiveMatchError`

saisoma123:
I have a working implementation with the first option

magnus-madsen:
Keep the trait sealed and keep everything in the same file.

saisoma123:
Added the sealed trait and reconfigured the `PatMatch.scala` and `TestPatExhaustiveness.scala`. I didn't see a `TestPatMatch` file, should I add the new tests to `Test.Exp.Jvm.TryCatch.flix` or `TestPatExhaustiveness.scala`? Or make a new test file.

magnus-madsen:
Is there not a `TestPatMatch` or similar?

magnus-madsen:
I correct myself: I would rename `TestPatExhaustiveness` to `TestPatchMatch` and put your tests there.

saisoma123:
Added the tests, but I am getting a unknown key error with `printUnlessUnit`

magnus-madsen:
What happens if you call the methods something other than main, e.g. just f?

saisoma123:
Renamed like you said, got this error now: ```
Expected an error of type ExceptionTypeMatchError, but found:
  
  List(ca.uwaterloo.flix.language.errors.ParseError$UnexpectedToken, ca.uwaterloo.flix.language.errors.ParseError$UnexpectedToken, ca.uwaterloo.flix.language.errors.ParseError$UnexpectedToken)
```

mlutze:
> Renamed like you said, got this error now: `Expected an error of type ExceptionTypeMatchError, but found: List(ca.uwaterloo.flix.language.errors.ParseError$UnexpectedToken, ca.uwaterloo.flix.language.errors.ParseError$UnexpectedToken, ca.uwaterloo.flix.language.errors.ParseError$UnexpectedToken)`

This indicates there is a parsing error in your tests. Do they parse correctly, for example at play.flix.dev ?

saisoma123:
Yeah looks like they are parsing fine on the play.flix.dev site

saisoma123:
Ah my bad, fixed the tests they are working now. Although in Test.Jvm.TryCatch.flix there seems to be a failing test, in which there is ArithmeticException after Exception.

saisoma123:
```def testPriority01(): Bool \ IO =
        try {
            exception();
            false
        } catch {
            case _: ##java.lang.Exception => true
            case _: ##java.lang.ArithmeticException => false
        }```, this test in particular I'm assuming

mlutze:
> ```
> def testPriority01(): Bool \ IO =
>         try {
>             exception();
>             false
>         } catch {
>             case _: ##java.lang.Exception => true
>             case _: ##java.lang.ArithmeticException => false
>         }
> ```
> 
> This test in particular I'm assuming

Yes, this test is no longer valid and can be deleted.

saisoma123:
Yep just removed that test, all checks passed

magnus-madsen:
@saisoma123 Are you still interested in working on this?

magnus-madsen:
If @saisoma123 is busy, @sockmaster27 you are welcome to take over this PR and polish it.

PR #8305, opened 2024-08-13
chore: enable struct related tests

dghosef:
Now that initial struct support is complete, enable tests that were previously ignored

Tests in the form of flix files will come in a different PR


magnus-madsen (approved):

magnus-madsen:
🚀

PR #8313, opened 2024-08-13
feat: propagate `SecurityContext`

magnus-madsen:
This PR propagates the SecurityContext from Flix add input methods into Input


main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -444,6 +444,8 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
   def reconfigureFlix(flix: Flix): Unit = {
     val previousSources = timestamps.keySet
 
+    implicit val defaultSctx: SecurityContext = SecurityContext.AllPermissions
```
magnus-madsen, line 447: @AStenbaek Here we will need to discover the right context based on the manifests.

mlutze (approved):

PR #8315, opened 2024-08-13
fix: Handle infinitely recursive struct types with some premature erasure

dghosef:
This is not the prettiest change. The problem is that the work of the `Monomorpher`, `Eraser`, and `Simplifier` must all be done at once.
- `Monomorpher` because we must determine the individual type of each field
- `Eraser` because we cannot infinitely expand types and therefore must erase at the same time
- `Simplifier` because this is all done during generation of `MonoType`

Though this duplicates work and isn't pretty, I don't think it leads to correctness issues, as it is just doing stuff that is done in earlier / later phases

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
@@ -280,7 +280,7 @@ object Simplifier {
             // We must do this here because the `MonoTypes` requires the individual types of each element
             // but the `Type` type only carries around the type arguments. i.e. for `struct S[v, r] {a: List[v]}`
             // at this point we would know `v` but we would need the type of `a`. We also erase to avoid infinitely
-            // expanding infinite types
+            // expanding recursive types
             val struct = root.structs(sym)
             val subst = Substitution(struct.tparams.zip(tpe.typeArguments).toMap)
             val substitutedStructFieldTypes = struct.fields.map { f =>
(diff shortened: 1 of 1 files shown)


magnus-madsen:
@dghosef @JonathanStarup Let me know the outcome of your deliberations.

dghosef:
@magnus-madsen, @JonathanStarup  and I talked and decided that we have a few options for this
- Keep this PR. The problem with this is now we do some erasing in the `Simplifier` which is not good. The plus side of this is that it is pretty easy to implement
- Move the substitution of type variables further back into the eraser phase or some phase beyond. The problem with this is that now we are going to need a new scala type for Flix types that is the same as `MonoType` but also has type variables(since we're doing a substitution). We'd also need to pass down a type parameter -> type map. The plus side of this is now all the erasing actually happens in the eraser
- Just lose the types and figure out how to get `GenExpression` to work without the type of the struct when getting/setting fields.

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -279,10 +279,37 @@ object Simplifier {
...
-            // at this point we would know `v` but we would need the type of `a`
+            // at this point we would know `v` but we would need the type of `a`. We also erase to avoid infinitely
+            // expanding recursive types
             val struct = root.structs(sym)
             val subst = Substitution(struct.tparams.zip(tpe.typeArguments).toMap)
-            val substitutedStructFieldTypes = struct.fields.map(f => visitType(subst(f.tpe)))
+            val substitutedStructFieldTypes = struct.fields.map { f =>
+              subst(f.tpe).typeConstructor match {
+                case Some(value) => value match {
+                  case TypeConstructor.AnyType | TypeConstructor.Unit | TypeConstructor.Null | TypeConstructor.BigDecimal |
```
magnus-madsen, line 289: Lets make this into a wildcard match _=>

magnus-madsen:
How about we go with this PR for now and then I would like to explore:

> Just lose the types and figure out how to get GenExpression to work without the type of the struct when getting/setting fields.

@JonathanStarup Are you OK with this PR as-is? (Modulo my one comment.)

JonathanStarup:
yes looks good. I think this belongs in MonoTypes conceptually but the existing coherence in general is not great, so i think we should merge this

magnus-madsen (approved):

Issue #8308, opened 2024-08-13
Structs post-MVP features


magnus-madsen:
I tried to sort the list from easy towards more difficult:

- [x] Fix missing semantic tokens for struct declarations
- [x] Add `struct` to `KeywordOtherCompleter`
- [x] Add `StructFieldCompleter` based on `UnexpectedField`
- [x] Refactor `Ref.flix` to use a struct
- [x] Refactor `MutDeque.flix` to use a struct
- [x] Add struct declarations to `Index` / `Indexer`
- [x] Add `StructCompleter` (like `EnumCompleter`) to enable auto-completion of struct types
- [x] Add support for struct fields and types to `GotoProvider`
- [x] Add support for structs fields and types to `FindReferencesProvider`
- [x] Add support for struct fields (but not types) to `RenameProvider`
- [x] Add one or two cool examples to `examples` folder
- [x] Add `mut` modifier to fields*.
- [ ] Add `pub` modifier to fields*.

If we can refactor `MutDeque` to use a struct we should feel confident in our system :)

* These should be discussed, but the code can already be written.

Lastly, but importantly:

- [x] We need to make a choice about syntax.

magnus-madsen:
Followup:

- Show field type for StructFieldCompleter

PR #8322, opened 2024-08-14
feat: Add a `mut` annotation to struct fields

dghosef:
#8308

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/errors/ResolutionError.scala
+++ b/main/src/ca/uwaterloo/flix/language/errors/ResolutionError.scala
@@ -1210,7 +1210,8 @@ object ResolutionError {
   /**
    * An error raised to indicate a `put` struct expression attempts to modify an immutable field
    *
-   * @param loc    the location where the error occurred
+   * @param field the immutable field
+   * @param loc   the location where the error occurred
    */
   case class NonModifiableField(field: Symbol.StructFieldSym, loc: SourceLocation) extends ResolutionError with Recoverable {
     override def summary: String = s"Modification of immutable field $field. Mark the field as `mut` to allow mutation."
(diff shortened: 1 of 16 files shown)


main/src/ca/uwaterloo/flix/language/errors/ResolutionError.scala
```diff
@@ -1207,6 +1207,23 @@ object ResolutionError {
     }
   }
 
+  /**
+   * An error raised to indicate a `put` struct expression attempts to modify an immutable field
+   *
+   * @param loc    the location where the error occurred
+   */
+  case class IllegalPut(loc: SourceLocation) extends ResolutionError with Recoverable {
+    override def summary: String = s"Modification of struct field not marked as `mut`"
```
magnus-madsen, line 1216: How about:

Modification of immutable field. Mark the field as `mut` to allow mutation.

main/src/ca/uwaterloo/flix/language/errors/ResolutionError.scala
```diff
@@ -1207,6 +1207,23 @@ object ResolutionError {
     }
   }
 
+  /**
+   * An error raised to indicate a `put` struct expression attempts to modify an immutable field
+   *
+   * @param loc    the location where the error occurred
+   */
+  case class IllegalPut(loc: SourceLocation) extends ResolutionError with Recoverable {
```
magnus-madsen, line 1215: Also, lets get the field as part of the error.

magnus-madsen, line 1215: IllegalPut -> NonModifiableField  or something like it

dghosef:
I'll look at those suggestions.

However, I've started working on lookup of struct(which is related to `pub`) and that one would conflict with this, so I will get that one in and then finish this one

magnus-madsen:
Let me know when I should look at this again.

dghosef:
You can look at it again. I feel this and #8334 should remain drafts until we discuss on Monday

PR #8329, opened 2024-08-15
feat: add `Indexer` support to `StructSym` and `StructFieldSym`

dghosef:


main/src/ca/uwaterloo/flix/api/lsp/provider/HighlightProvider.scala
```diff
@@ -30,12 +30,16 @@ object HighlightProvider {
       case Some(entity) => entity match {
         case Entity.Case(caze) => highlightCase(uri, caze.sym)
 
+        case Entity.StructField(field) => throw new RuntimeException("JOE TODO")
```
magnus-madsen, line 33: This is a bit dangerous because now VSCode crashes if you just hover. Ill merge for now, but it should be fixed fast.

Otherwise you could just return mkNotFound

magnus-madsen (approved):

main/src/ca/uwaterloo/flix/api/lsp/provider/HighlightProvider.scala
dghosef: I can push all the changes at once

dghosef: I'll be quick

magnus-madsen: No worries. Lets merge this and then you can make a new PR.

PR #8334, opened 2024-08-15
feat: Allow multiple structs in a mod as long as there are no duplicate field names

dghosef:
A future thing to think about would be what to do about structs w/ the same field names. But this is better than what we currently have and is necessary I think to start actually using structs in real life(e.g. to replace `Ref` and `MutDeque`

One thing to note is that this screws up field autocomplete. Basically, because the struct field is used to determine the type of the struct, rather than vice versa, we cannot autocomplete `struct.f` where `f` is an incomplete field because now we don't have a list of option of fields to pick.

#8308


magnus-madsen:
I suppose we can discuss these things on our Monday meeting.

JonathanStarup:
My understanding was that companion namespaces are used, i.e.
```
struct Cool[r] { pub mut x: Int32 }
mod Cool {
  pub def hereICanUseFields(): Int32 = region rc {
    let cool = new Cool{x = 42} @ rc;
    cool.x
  }
}
```
So that would mean that no structs could ever share namespace?

dghosef:
No structs with overlapping girls names can share namespace in this implementation

JonathanStarup:
But is my example program valid? because then I don't understand what the problem is

dghosef:
Yea your example should be fine. Problems only arise if for example someone makes a new struct with an x field inside the mod. 

Actually though on second thought, with the current implementation, even if someone makes a variable called x, we run into problems. But that's probably not too hard to fix

dghosef:
With this PR, the following code doesn't compile:
```
struct S[r] { foo: Int32 }
def foo(): Unit = ()
```

because the names `foo` collide and thus the `Namer` phase returns an error. But I think there is no reason why a struct field can't have the same name as a function.

Is there an easier way to fix this than passing around a second `SymbolTable` in `Namer` that is specifically for struct field symbols?

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -2353,6 +2339,24 @@ object Resolver {
...
+   */
+  private def lookupStructField(name: Name.Label, env: ListMap[String, Resolution], ns0: Name.NName, root: NamedAst.Root): Result[NamedAst.Declaration.StructField, ResolutionError.UndefinedStructField] = {
+    val matches = tryLookupName(Name.mkQName("€" + name.name, name.loc), env, ns0, root) collect {
+      case Resolution.Declaration(s: NamedAst.Declaration.StructField) => s
+    }
+    matches match {
+      // Case 0: No matches. Error.
+      case Nil => Result.Err(ResolutionError.UndefinedStructField(name, name.loc))
+      // Case 1: Exactly one match. Success.
+      case field :: _ => Result.Ok(field)
```
magnus-madsen, line 2353: This is that bug again right?

magnus-madsen (approved):

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
dghosef: Oh wait I have a fix. I'll put it in a different PR

PR #8337, opened 2024-08-15
feat: add security permission check to Safety phase

### LINKED ISSUE #8316: Extend the `Safety` phase to check the security context for all dangerous operations
In the `Safety` phase, guard sensitive operations. They are:

- [ ] UncheckedCast
- [ ] UncheckedMaskingCast
- [ ] Throw
- [ ] InvokeMethod
- [ ] InvokeStaticMethod
- [ ] GetField
- [ ] PutField
- [ ] GetStaticField
- [ ] PutStaticField
- [ ] NewObject (?)

InvokeConstructor is already implemented and can be used as a template.

AStenbaek:
resolve #8316

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Safety.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Safety.scala
@@ -385,9 +385,16 @@ object Safety {
             visit(exp)
         }
 
-      case e@Expr.UncheckedCast(exp, _, _, _, _, _) =>
-        val errors = verifyUncheckedCast(e)
-        visit(exp) ++ errors
+      case e@Expr.UncheckedCast(exp, _, _, _, _, loc) =>
+        val ctx = loc.security
+        if (ctx == SecurityContext.AllPermissions) {
+          // Permitted
+          val errors = verifyUncheckedCast(e)
+          visit(exp) ++ errors
+        } else {
+          // Forbidden
+          SafetyError.Forbidden(ctx, loc) :: Nil
+        }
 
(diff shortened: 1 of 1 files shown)


main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -385,12 +385,26 @@ object Safety {
...
+          // Forbidden
+          SafetyError.Forbidden(ctx, loc) :: Nil
+        }
 
-      case Expr.UncheckedMaskingCast(exp, _, _, _) =>
-        visit(exp)
+      case Expr.UncheckedMaskingCast(exp, _, _, loc) =>
+        val ctx = loc.security
+        if (ctx == SecurityContext.AllPermissions) {
+          // Allowed...
```
magnus-madsen, line 402: Why not use the same wording as elsewhere? i.e. Permitted / Forbidden?

magnus-madsen:
What about the other constructs?

magnus-madsen (approved):

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
AStenbaek: No good reason in all honesty. Mostly me being silly after finding out what `UncheckedmaskingCast`was.
I almost feel like this should never really be allowed, as in it should not be a thing, given it seems like an "unchecked unchecked cast", but there may be some uses I haven't been able to spot?

Issue #7983, opened 2024-06-26
Add `Index` and `IndexMut` syntax, traits, and instances


magnus-madsen:
- [ ] Extend the parser to support subscripting
- [ ] Add `Index` and `IndexMut` traits
- [ ] Add instances for `Array` and `Vector`
- [ ] Add tests

magnus-madsen:
@JonathanStarup is driving this, but @jaschdoc can help with the refactoring.

magnus-madsen:
@JonathanStarup What's the status on this? @jaschdoc is standing by to help

JonathanStarup:
We should start by adding `Index.flix` and `IndexMut.flix` based on https://dl.acm.org/doi/pdf/10.1145/3656393

The one change is that `IndexMut[t]` should have `with Index[t]`

@magnus-madsen should I do it or @jaschdoc ?

magnus-madsen:
@JonathanStarup I think you should add the traits -- with proper documentation etc.

magnus-madsen:
Then @jaschdoc can add some instances.

JonathanStarup:
#8339 now has the traits with some sample instances to make sure the design is sensible. What should happen now?

jaschdoc:
I can add some more instances once it's merged

JonathanStarup:
@magnus-madsen i don't know if you meant to assign @sockmaster27, but @jaschdoc its ready now :) I'm excited to have this, its gonna be cool 😎

magnus-madsen:
Holger is working full time this week, so I will let him work on this one.

magnus-madsen:
@JonathanStarup I think tests should not be added before the new syntax is in. Then we get more bang for the buck.

JonathanStarup:
@magnus-madsen so I should just start adding the parser support?

magnus-madsen:
> @magnus-madsen so I should just start adding the parser support?

Yes please 👍

jaschdoc:
> Add tests for `Array`, `Vector`, and `MutMap`

Do you want these as library tests or `Test.Exp.Indexable`?

magnus-madsen:
Most of these have been resolved.

PR #8339, opened 2024-08-15
feat: `Index` and `IndexMut` traits

JonathanStarup:
Part of #7983

Manual Test
```scala
def main(): Unit \ IO = region rc {
    use Index.get;
    use IndexMut.put;
    let l = "a" :: "b" :: "c" :: Nil;
    let v = l |> List.toVector;
    let a = l |> List.toArray(rc);
    let m = l |> List.zipWithIndex |> List.toMap |> Map.toMutMap(rc);
    testIt(() -> get(1, v));
    testIt(() -> get(3, v));

    println("");
    testIt(() -> get(1, a));
    testIt(() -> put(1, "z", a));
    testIt(() -> get(1, a));
    testIt(() -> get(4, a));
    testIt(() -> OutOfBounds.asRes(get(4))(a));

    println("");
    testIt(() -> get(0, m));
    testIt(() -> put(0, "q", m));
    testIt(() -> get(0, m));
    testIt(() -> get(4, m));
    ()
}

def testIt(f: Unit -> a \ ef): Unit \ IO + ((ef - KeyNotFound) - OutOfBounds) with ToString[a] = {
    match KeyNotFound.asRes(OutOfBounds.asRes(f))() |> Result.flatten {
        case Ok(v) => printUnlessUnit(v)
        case Err(msg) => printUnlessUnit(msg)
    }
}
```
Output
```
b
index 3 is out of bounds for Vector of length 3

b
z
index 4 is out of bounds for Array of length 3
Err(index 4 is out of bounds for Array of length 3)

a
q
key not found
```


JonathanStarup:
@mlutze can you guess why this occurs?
```
>> Mismatched signature 'put' required by 'IndexMut'.

40 |     pub def put(i: k, v: v, t: MutMap[k, v, r]): Unit \ r = MutMap.put!(i, v, t)
                 ^^^
                 mismatched signature.

Expected scheme: ∀(k, v, r). Idx[MutMap[k, v, r]] -> (Elm[MutMap[k, v, r]] -> (MutMap[k, v, r] -> Unit \ Aef[MutMap[k, v, r]])) with IndexMut[MutMap[k, v, r]]
Actual scheme:   k -> (v -> (MutMap[k, v, r] -> Unit \ r)) with IndexMut[MutMap[k, v, r]]
```

mlutze:
> @mlutze can you guess why this occurs? It looks like the same types to me based on the associated types
> 
> ```
> >> Mismatched signature 'put' required by 'IndexMut'.
> 
> 40 |     pub def put(i: k, v: v, t: MutMap[k, v, r]): Unit \ r = MutMap.put!(i, v, t)
>                  ^^^
>                  mismatched signature.
> 
> Expected scheme: ∀(k, v, r). Idx[MutMap[k, v, r]] -> (Elm[MutMap[k, v, r]] -> (MutMap[k, v, r] -> Unit \ Aef[MutMap[k, v, r]])) with IndexMut[MutMap[k, v, r]]
> Actual scheme:   k -> (v -> (MutMap[k, v, r] -> Unit \ r)) with IndexMut[MutMap[k, v, r]]
> ```

I'd guess the solver is broken...

magnus-madsen:
frowning-sad-face-large-size-of-yellow-emoji-smile-free-vector

magnus-madsen:
Is this a rigidity issue?

JonathanStarup:
maybe? since the second the type is not quantified and the first is

main/src/library/IndexMut.flix
```diff
@@ -0,0 +1,36 @@
...
+    type Elm: Type
+
+    ///
+    /// The associated effect of the data structure which represents the effect
+    /// of inserting elements.
+    ///
+    type Aef: Eff
+
+    /// Insert the element `v` at index `i` in `t`.
+    pub def put(i: Index.Idx[t], v: Index.Elm[t], t: t): Unit \ Index.Aef[t]
```
mlutze, line 34: Should this be `IndexMut.Aef`?

mlutze: Same for the others

main/src/library/InvalidKey.flix
```diff
@@ -0,0 +1,38 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+///
+/// Indicates that a key was invalid.
+/// Useful for instances of `Index` or `IndexMut`.
+///
+eff InvalidKey {
```
magnus-madsen, line 21: The key is not invalid. Rather it does not exist: I would propose we call it either "NoSuchElement" or "InvalidIndex" or "IndexOutOfBounds". I think I prefer the last as the most precise.

main/src/library/InvalidKey.flix
```diff
@@ -0,0 +1,38 @@
...
+
+    /// Stops execution with an error message.
+    pub def error(msg: String): Void
```
magnus-madsen, line 24: I think, when there is one operation, the effect and operations should have the same name. Then, later, we can add sugar.

main/src/library/InvalidKey.flix
```diff
@@ -0,0 +1,38 @@
...
+
+}
+
+mod InvalidKey {
+
+    /// Removes the `InvalidKey` effect from a function `f`, if `f` `do` `InvalidKey`.
+    pub def asRes(f: a -> b \ ef): (a -> Result[String, b] \ ef - InvalidKey) = {
```
magnus-madsen, line 31: 😮

main/src/library/MutMap.flix
```diff
@@ -21,6 +21,23 @@ pub enum MutMap[k: Type, v: Type, r: Region] {
     case MutMap(Region[r], Ref[Map[k, v], r])
 }
 
+instance Index[MutMap[k, v, r]] with Order[k] {
+    type Idx = k
+    type Elm = v
+    type Aef = r + InvalidKey
+    pub def get(i: k, t: MutMap[k, v, r]): v \ r + InvalidKey = {
```
magnus-madsen, line 28: I think a linebreak after associated types/effects is reasonable?

main/src/library/InvalidKey.flix
JonathanStarup: You're reading it as "key is not valid (forever)" but I think of it as "key is not valid (right now)". I think youre right that `InvalidKey` would be a better fit, compared to `InvalidIndex` if it was also called `Mapping` and `MappingMut`.

I don't think out of bounds is good, it sounds nonsense when you're working on `Map` fx

Another option is to have both `OutOfBounds` and `InvalidIndex`.

magnus-madsen: For Java familiarity, I think OutOfBounds would be most understandable?

magnus-madsen:
@JonathanStarup If you fix this first thing, I can have @sockmaster27 add instances.

Do we have a list of instances?

main/src/library/InvalidKey.flix
mlutze: `OutOfBounds` feels very weird to me for maps. What are the bounds?

magnus-madsen: The set of keys. Its just a collection of discrete bounds.

main/src/library/MutMap.flix
JonathanStarup: I did it now but I'm not a fan

JonathanStarup:
@magnus-madsen im happy now, ill make a list of instances to do

magnus-madsen (approved):

Issue #8268, opened 2024-08-08
Introduce `redef` keyword


magnus-madsen:
- [x] Introduce `redef` keyword as a short hand for `override def`.
- [x] Refactor the standard library and tests to use `redef`.
- [x] Remove support for `override`. (Or better put it behind `--Xdeprecated` flag)
- [x] Update the KeywordOtherCompleter (new in, old out)
- [x] Update VSCode keywords (new in, old out)
- [x] Update Highlight.js keywords (new in, old out)
- [x] Update doc.flix.dev
- [x] Update flix.dev

Note: This should not concern visibility, so this changes nothing w.r.t. `pub`.

magnus-madsen:
@jaschdoc Next I would drop the requirement that declarations in instances must be public. 

Then you can refactor instances to use `redef`.

magnus-madsen:
What do you mean? I suppose what I was trying to communicate is that first changing everything to use redef, and then changing everything to drop pub seems like double work. Might as well do it in one pass.

jaschdoc:
Should I also remove the `pub` constraint in the trait declaration?

magnus-madsen:
What do you mean?  I am only asking for changes to instances at this time.

jaschdoc:
Alright. In Rust the `pub` keyword is unused on trait declarations because it defines an interface where all members are public. I though that it would be relevant to this. I will, however, only rework the instances for now.

jaschdoc:
But do we allow private functions on a trait declaration?

magnus-madsen:
> But do we allow private functions on a trait declaration?

We have not decided. But it could make sense.

jaschdoc:
- [ ] Introduce redef keyword as a short hand for override def.
- [ ] Refactor the standard library and tests to use redef.
- [ ] Remove support for override. (Or better put it behind --Xdeprecated flag)
- [ ] Deprecate `pub` keyword
- [ ] Update the KeywordOtherCompleter (new in, old out)
- [ ] Update VSCode keywords (new in, old out)
- [ ] Update Highlight.js keywords (new in, old out)
- [ ] Update doc.flix.dev
- [ ] Update flix.dev

magnus-madsen:
I think we just need to get rid of the pub in `pub redef` for instances.

jaschdoc:
I think we can close this now

PR #8345, opened 2024-08-15
feat: add `redef` keyword

jaschdoc:

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
@@ -17,6 +17,7 @@
 package ca.uwaterloo.flix.language.phase
 
 import ca.uwaterloo.flix.api.Flix
+import ca.uwaterloo.flix.language.ast.Ast.Modifier
 import ca.uwaterloo.flix.language.ast.DesugaredAst.Expr
 import ca.uwaterloo.flix.language.ast.WeededAst.Predicate
 import ca.uwaterloo.flix.language.ast.shared.Fixity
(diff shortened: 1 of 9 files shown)


jaschdoc:
Should I just add another function to the parser called `redefinitionDecl` instead of extending the `definitionDecl` function and doing if-checks? Seems like a code smell

magnus-madsen:
> Should I just add another function to the parser called `redefinitionDecl` instead of extending the `definitionDecl` function and doing if-checks? Seems like a code smell

No, I think we should try do so a simple desugaring so that most the compiler stays the same.

magnus-madsen:
I suppose your question is: where should you do the desugaring? which is a good question.

jaschdoc:
Hmm we do have a separate desugaring phase so maybe there?

magnus-madsen:
> Hmm we do have a separate desugaring phase so maybe there?

I think we could, yes.

main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
```diff
@@ -120,6 +123,22 @@ object Desugar {
...
+    case WeededAst.Declaration.Redef(doc, ann, mod, ident, tparams0, fparams0, exp0, tpe0, eff0, tconstrs0, econstrs0, loc) =>
+      flix.subtask(ident.name, sample = true)
+      val tparams = tparams0.map(visitTypeParam)
+      val fparams = visitFormalParams(fparams0)
+      val exp = visitExp(exp0)
+      val tpe = visitType(tpe0)
+      val eff = eff0.map(visitType)
+      val tconstrs = tconstrs0.map(visitTraitConstraint)
+      val econstrs = econstrs0.map(visitEqualityConstraint)
+      DesugaredAst.Declaration.Def(doc, ann, mod, ident, tparams, fparams, exp, tpe, eff, tconstrs, econstrs, loc)
```
magnus-madsen, line 139: I think here we should add the override modifier, yes?

jaschdoc: Yeah oops forgot that

PR #8348, opened 2024-08-16
refactor: return type var symbols from Type.typeVars

mlutze:
In the vast majority of cases, we immediately map the vars to symbols. In other cases, IMO the location does not matter.


magnus-madsen:
Seems reasonable.

magnus-madsen:
Not sure why it breaks restrictable variants though

mlutze:
I'll try this again later.

PR #8350, opened 2024-08-16
feat: add a struct example with stack

dghosef:
This relies on https://github.com/flix/flix/pull/8334 to be merged first, so it is a draft for now

#8308


magnus-madsen:
I think this is a nice example. Do you have an idea for one more example?

magnus-madsen (commented):
I have some small comments like:

- Probably we should be returning Options
- I would order the operations by complexity (since it is an example), so: newStack, peek, pop, and push
- What's the relation between arr.length and capacity? Are they not the same?

dghosef:
>     * What's the relation between arr.length and capacity? Are they not the same?
Capacity is the total size of the underlying array. Length is the number of elements in the stack

dghosef:
> I think this is a nice example. Do you have an idea for one more example?

Could do any number of mutable data structures(queue, hashmap, etc), though that's probably a bit repetitive.

Maybe have an array of `r, g, b` structs represent an image and mutate the image somehow?

magnus-madsen:
> > I think this is a nice example. Do you have an idea for one more example?
> 
> Could do any number of mutable data structures(queue, hashmap, etc), though that's probably a bit repetitive.
> 
> Maybe have an array of `r, g, b` structs represent an image and mutate the image somehow?

Yeah. We want simple, not complex. It could be even simpler: just a struct that models some concept: Student? Animal? etc.

Take a look at https://doc.rust-lang.org/rust-by-example/custom_types/structs.html maybe?

magnus-madsen (approved):

PR #8361, opened 2024-08-19
fix: tryLookupName avoid duplicates

dghosef:
This fixes the problem for structs but not for enums.

For some reason, if you try to change the equivalent line in the enum class, you get a `unexpected duplicate enum: 'Nil'` error/


magnus-madsen:
Is this still relevant?

dghosef:
It's for the scala 3 migration

magnus-madsen (approved):

PR #8362, opened 2024-08-19
feat: allow effectful holes

### LINKED ISSUE #6851: Add a free effect to `?hole` or `???`
Currently, if I have a function where I know I'm going to need an effect, such as `IO`, in the future, I cannot mark it with that and use a hole to leave it unimplemented, as follows:
```
def my_future_io_function(): Unit \ IO = ?hole
```

The compiler will complain `Expected effect 'IO' but found effect: 'Pure'.`

I can work around this with `checked_ecast(?hole)` instead, but I feel like I shouldn't need to write extra code depending on what kind of unimplemented function I want.

Flix version `0.42.0`.

mlutze:
an experiment

Fixes #6851

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/Indexer.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/Indexer.scala
@@ -222,7 +222,7 @@ object Indexer {
       val parent = Entity.Exp(exp0)
       Index.occurrenceOf(exp0) ++ Index.useOf(sym, loc, parent) ++ Index.useOf(sym.trt, loc)
 
-    case Expr.Hole(_, _, _) =>
+    case Expr.Hole(_, _, _, _) =>
       Index.occurrenceOf(exp0)
 
     case Expr.HoleWithExp(exp, _, _, _) =>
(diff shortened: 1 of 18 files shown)


magnus-madsen:
Why do we need this? And won't sub-effecting solve it anyway?

magnus-madsen:
It was disabled for performance reasons. Not sure if those issues are fixed yet.

JonathanStarup (approved):

JonathanStarup:
I support this - I use `???()` anyway, which have the same behaviour as this `???`

JonathanStarup:
> It was disabled for performance reasons. Not sure if those issues are fixed yet.

is `???` used that often? one var per `???` doesn't seem that bad

mlutze:
> > It was disabled for performance reasons. Not sure if those issues are fixed yet.
> 
> is `???` used that often? one var per `???` doesn't seem that bad

I think it could be bad if e.g. we generate a match for a big enum and put `???` in all the bodies.

magnus-madsen:
Do we want to polish and merge?

JonathanStarup:
I vote yes

mlutze:
> What happened here?

Thought it was merged already so I deleted the branch

mlutze:
!!!rerun

magnus-madsen (approved):

PR #8367, opened 2024-08-20
fix: sizes to everything

### LINKED ISSUE #8049: uniform `size`/`length` on enums and traits
`MutDeque.size`, `Array.length`, `Iterable[t]` doesnt have length, etc

JonathanStarup:
I've chosen `size` as the universal word (instead of `length`) because size is more general.

The rule is: **if you have length you must have size**

Fixes #8049


main/src/library/Foldable.flix
```diff
@@ -49,6 +49,11 @@ pub trait Foldable[t : Type -> Type] {
     pub def length(t: t[a]): Int32 \ Foldable.Aef[t] =
         Foldable.foldLeft((acc, _) -> acc + 1, 0, t)
 
+    ///
+    /// Returns the length of `t`, i.e. the number of times it can be folded over.
+    ///
+    pub def size(t: t[a]): Int32 \ Foldable.Aef[t] = Foldable.length(t)
```
JonathanStarup, line 55: It might potentially be an issue to have both on the trait since if you only `redef` size, you get asymmetrical behaviour on what is supposed to be an alias

magnus-madsen:
You should also add tests.

magnus-madsen:
But first: Why is it good to have both? Java, for example, does not offer length on list.

main/src/library/Foldable.flix
```diff
@@ -388,4 +388,9 @@ mod Foldable {
     pub def fold(t: t[a]): a \ Foldable.Aef[t] with Foldable[t], Monoid[a] =
         Foldable.foldLeft((acc, x) -> Monoid.combine(acc, x), Monoid.empty(), t)
 
+    ///
+    /// Returns the length of `t`, i.e. the number of times it can be folded over.
```
magnus-madsen, line 392: Its not the number of times it can be folded over? Its the number of elements.

main/src/library/Iterable.flix
```diff
@@ -42,4 +42,11 @@ pub trait Iterable[t] {
     pub def enumerator(rc: Region[r], t: t): Iterator[(Int32, Iterable.Elm[t]), r + aef, r] \ (r + aef) where Iterable.Aef[t] ~ aef =
         Iterable.iterator(rc, t) |> Iterator.zipWithIndex
 
+    ///
+    /// Returns the size of `t`.
+    ///
+    pub def size(t: t): Int32 \ Iterable.Aef[t] = region rc {
```
magnus-madsen, line 48: I would NOT offer this function. Its seems error-prone, if the underlying collection changes.

magnus-madsen: Also, Iterable should NOT offer functions. It is just a factory.

main/src/library/Iterator.flix
```diff
@@ -127,6 +127,14 @@ mod Iterator {
     pub def sumWith(f: a -> Int32 \ ef2, iter: Iterator[a, ef1, r]): Int32 \ { ef1, ef2, r } =
         foldLeft((acc, x) -> acc + f(x), 0, iter)
 
+    ///
+    /// Returns the size of the iterator.
+    ///
+    /// Consumes the entire iterator.
+    ///
+    pub def size(iter: Iterator[a, ef1, r]): Int32 \ { ef1, r } =
```
magnus-madsen, line 135: I think this is dangerous and should not be offered.

main/src/library/MultiMap.flix
```diff
@@ -114,6 +114,13 @@ mod MultiMap {
     ///
     pub def nonEmpty(m: MultiMap[k, v]): Bool = not isEmpty(m)
 
+    ///
+    /// Returns the size of `m`.
```
magnus-madsen, line 118: number of keys?

main/src/library/MutList.flix
```diff
@@ -90,6 +90,11 @@ mod MutList {
         let MutList(_, _, l) = v;
         deref l
 
+    ///
+    /// Returns the number of elements in the given mutable list `v`.
```
magnus-madsen, line 94: currently in? (same for other mutable things?)

main/src/library/Option.flix
```diff
@@ -174,6 +174,14 @@ mod Option {
     ///
     pub def nonEmpty(o: Option[a]): Bool = not isEmpty(o)
 
+    ///
+    /// Returns `1` for `Some(_)` and `0` for `None`.
+    ///
+    pub def size(o: Option[a]): Int32 = match o {
+        case Some(_) => 1
+        case None => 0
```
magnus-madsen, line 182: Swap order

magnus-madsen: Also: Why? Remember we dont have isNone or isSome. So this should also not be here.

main/src/library/Result.flix
```diff
@@ -67,6 +67,14 @@ mod Result {
         case Err(_) => true
     }
 
+    ///
+    /// Returns `1` for `Ok(_)` and `0` for `Err(_)`.
+    ///
+    pub def size(r: Result[e, t]): Int32 = match r {
```
magnus-madsen, line 73: Same comment as for Option.

main/src/library/Validation.flix
```diff
@@ -186,6 +186,14 @@ mod Validation {
         }
     }
 
+    ///
+    /// Returns `1` for `Success(_)` and `0` for `Failure(_)`.
+    ///
+    pub def size(v: Validation[e, t]): Int32 = match v {
```
magnus-madsen, line 192: Same comment about Option

magnus-madsen (commented):
OK; I have several concerns that I have outlined.

main/src/library/Foldable.flix
JonathanStarup: It is copied from `Foldable.length`

main/src/library/MultiMap.flix
JonathanStarup: This is copied from `Map.flix`

main/src/library/MutList.flix
JonathanStarup: For the userfacing api there isn't anything else it could be, so it feels redundant

main/src/library/Option.flix
JonathanStarup: I think its nice, but not enough to champion it 🤷‍♂️

magnus-madsen:
Let's discuss at the meeting.

JonathanStarup:
@magnus-madsen now its just the non-controversial length functions with updated docs and tests

magnus-madsen:
> @magnus-madsen now its just the non-controversial length functions with updated docs and tests

Ill take a look soon.

magnus-madsen (approved):

PR #8373, opened 2024-08-20
refactor: introduce levels

mlutze:
This introduces levels to the appropriate phases, but does not use them for unification/constraint solving, so I'm calling it a refactor.


magnus-madsen (commented):
Lowering? Do we need to scope it?

magnus-madsen (approved):

magnus-madsen:
You can merge or fix lowering. I leave it to you.

PR #8376, opened 2024-08-20
refactor: make `Ref` module use `struct`

dghosef:
I could potentially do more work to also get rid of `ref` in the compiler if that would be helpful.


magnus-madsen:
> I could potentially do more work to also get rid of `ref` in the compiler if that would be helpful.

We leave that extra work for later.

magnus-madsen:
@mlutze Flixball needs to use Ref and not ref.

mlutze:
!!!rerun

magnus-madsen:
@stephentetley We are about to replace the internal implementation of `Ref` with a struct.

However, this breaks the `ref` and `deref` expressions. Instead one should use, `Ref.fresh`, `Ref.get`, and `Ref.put`.

Do you have a lot of code that uses `ref` and `deref`? If so, we can implement a grace period.

stephentetley:
Thanks Magus, I haven't used ref cells much so I'm fine not to have a transition period.

magnus-madsen (approved):

PR #8380, opened 2024-08-21
feat: add struct example with Person

dghosef:


examples/struct-person.flix
```diff
@@ -0,0 +1,29 @@
...
+struct Person[r] {
+    name: String,
+    age: Int32,
+    height: Int32
+}
+
+def giveBirth(name: String, rc: Region[r]): Person[r] \ r =
+    new Person { name = name, age = 0, height = 30 } @ rc
+
+def displayPerson(person: Person[r]): Unit \ {r, IO} =
```
magnus-madsen, line 11: Lets drop this and simply inline the print into main.

magnus-madsen: You can also use string interpolation "foo${person.blarg}"to make it smaller.

magnus-madsen (commented):
I like the example. Let us polish it up: Add some comments that gives a flow.

See the other examples on https://flix.dev/

examples/struct-person.flix
dghosef: Sorry, did you mean to drop `giveBirth` or `displayPerson`?

dghosef:
!!rerun

magnus-madsen (approved):

PR #8398, opened 2024-08-22
ci: give IDE tests their own concurrency group

mlutze:
This fixes interference between the normal testing suite and the IDE testing suite.


magnus-madsen (approved):

magnus-madsen:
What was the interference?

mlutze:
> What was the interference?

We use concurrency groups to cancel jobs in progress if a PR is updated. In this case, IDE tests had the same concurrency group as normal tests (`build-${issueorsomething}`). This resulted in GitHub thinking they were the same job, so IDE tests starting up would cancel normal tests, and vice versa.

PR #8410, opened 2024-08-23
chore: change struct access from euro to arrow

dghosef:
The frequency of me googling "euro symbol" will now drastically decrease


main/src/ca/uwaterloo/flix/language/ast/TokenKind.scala
```diff
@@ -660,7 +660,7 @@ object TokenKind {
 
   case object EqualEqual extends TokenKind
 
-  case object Euro extends TokenKind
+  case object StructArrow extends TokenKind
```
magnus-madsen, line 663: This should be moved, so its alphabetical again.

main/src/ca/uwaterloo/flix/language/ast/TokenKind.scala
```diff
@@ -56,7 +56,7 @@ sealed trait TokenKind {
       case TokenKind.DotCurlyL => "'.{'"
       case TokenKind.Equal => "'='"
       case TokenKind.EqualEqual => "'=='"
-      case TokenKind.Euro => "'€'"
+      case TokenKind.StructArrow => "'->'"
```
magnus-madsen, line 59: This should be moved, so its alphabetical again.

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -200,6 +200,17 @@ object Lexer {
     }
   }
 
+  /**
+   * Peeks the character two before the previous that state was on if available.
```
magnus-madsen, line 204: three? Sorry I don't even understand the comment, actually.

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -352,7 +362,13 @@ object Lexer {
       case _ if isOperator(":") => TokenKind.Colon
       case _ if isOperator("**") => TokenKind.StarStar
       case _ if isOperator("<-") => TokenKind.ArrowThinL
-      case _ if isOperator("->") => TokenKind.ArrowThinR
+      case _ if isOperator("->") =>
+        // Check for whitespace around arrow.
```
magnus-madsen, line 366: I think we want to explain here. A `->` string is a function arrow if there is whitespace before and after? Otherwise it is a structarrow.

So `foo ->a` is struct arrow?

Actually, in the comment, could you just write out all the cases with examples. Thanks!

magnus-madsen: Wait, it says or, so maybe I am confused. Definitely want a small table and better description 👍

main/src/library/MutDeque.flix
```diff
@@ -213,42 +213,42 @@ mod MutDeque {
     /// Pushes `x` to the front of `d`.
     ///
     pub def pushFront(x: a, d: MutDeque[a, r]): Unit \ r =
-        let f1 = (d€front - 1) `Int32.modulo` capacity(d);       // Update index such that it points to an empty index. This will never overlap with the back index.
-        Array.put(x, f1, d€values);                              // Store `x` in the array.
-        d€front = f1;                                            // Update the front index reference.
+        let f1 = (d->front - 1) `Int32.modulo` capacity(d);       // Update index such that it points to an empty index. This will never overlap with the back index.
```
magnus-madsen, line 216: Your search and replace broken indentation :/ I would like that fixed though :)

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
dghosef: It was basically what `previousPrevious` did but one more. I just generalized it to work for `n` characters - maybe that will be easier to understand. Alternatively, I could just inline the function and wouldn't need a separate definition

magnus-madsen:
Does -> work with the completion stuff?

dghosef:
Actually, for some reason it's not working right now. Let me look into it

dghosef:
For some reason, with `->`, `getCompletions` is still being invoked and returning the correct completions, but none are showing up on vscode. Do we have any mechanisms for debugging the completer?

magnus-madsen:
> For some reason, with `->`, `getCompletions` is still being invoked and returning the correct completions, but none are showing up on vscode. Do we have any mechanisms for debugging the completer?

Maybe it has something to do with "triggerChars" in the VSCode extension. But not sure.

magnus-madsen:
Or maybe you are returning labels that are being filtered because they dont match the text.

dghosef:
The error is still the same error

magnus-madsen:
Maybe the replacement text has to start with -> or something. (Now I am just guessing).

dghosef:
Yea that's a good idea - I'll try that

dghosef:
Alright, I got it working

magnus-madsen:
> Alright, I got it working

What did you learn?

dghosef:
I think it was a location problem. Previously, we were just giving it the location of the `SyntacticContext` which I think isn't correct with a 2 character struct accessor token but was correct with the Euro(and I would guess it would work with the dot, etc). Now I'm just manually passing the location to the completer

magnus-madsen:
OK, Ill take a look later.

magnus-madsen:
I played with the JAR:

- Autocompletion works 👍 
- Hovering on a field access seems to work 👍 
- Jump to definition on failed access seems not work 🔴  (Did it ever work?)
- Highligh provider works on the type, but no on the field access expr, maybe related to the above 🔴

I need to see if I can give `->` a different color in VSCode.

magnus-madsen:
These are not blockers for just PR. I am just checking where we are.

magnus-madsen (approved):

magnus-madsen:
I merged this so we can progress.

dghosef:
What do you mean by jump to definition on failed access?

dghosef:
Is it like:
```
struct S[r] {field1: Int32}
def foo(s: S[r]): Unit \ r = s->fiel
```
and jump to definition on the partially completed field?

magnus-madsen:
No, on a fully correct program.

dghosef:
Currently, go to definition isn't working for anything for me and the Indexer isn't working. Is it just me?

magnus-madsen:
> Currently, go to definition isn't working for anything for me and the Indexer isn't working. Is it just me?

Hmm. Maybe I broke it then. I will have a look.

magnus-madsen:
Indeed it seems broken.

PR #8411, opened 2024-08-23
fix: make struct field name resolution use the mod name

dghosef:
This relies on #8410


magnus-madsen:
I suppose this has to be rebased.

dghosef:
Got it

magnus-madsen (approved):

PR #8414, opened 2024-08-23
feat: run `Indexer` asynchronously in `LanguageServer`

### LINKED ISSUE #8413: Compute index asynchronously in LanguageServer

magnus-madsen:
Fixes #8413

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/LanguageServer.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/LanguageServer.scala
@@ -91,14 +91,14 @@ class LanguageServer(port: Int, o: Options) extends WebSocketServer(new InetSock
     *
     * Note: The index is updated *asynchronously* by a different thread, hence:
     *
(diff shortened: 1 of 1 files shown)


mlutze (approved):
Do I understand correctly: this is valid because the errors don't actually use the index, so it doesn't matter if they're out of sync. But other things like code actions can happen slightly later and that's ok?

magnus-madsen:
> Do I understand correctly: this is valid because the errors don't actually use the index, so it doesn't matter if they're out of sync. But other things like code actions can happen slightly later and that's ok?

Yes.

Except its not OK for the IDE tests because they are done by a robot: e.g. a function is added and then immediately renamed.

Hm.

magnus-madsen:
Now it is a hack on top of a hack. But this should help the IDE tests pass.

In the future, I think the plan should still be to get rid of the index. It is wasteful to recompute over and over. 
Better to search when needed.

PR #8421, opened 2024-08-25
feat: drop `pub` constraint for trait instances

jaschdoc:


jaschdoc:
Not sure why the vscode tests fail. Apparently it's renaming that fails but just tried manually in vscode and it seems to work just fine.

jaschdoc:
!!!rerun

jaschdoc:
Ok seems like it was just a bad run

magnus-madsen (approved):

PR #8424, opened 2024-08-25
refactor: use `redef` instead of `{pub, override} def` in library

jaschdoc:


magnus-madsen:
OK; so the pub keyword will be removed later?

magnus-madsen (approved):

PR #8433, opened 2024-08-26
ci: add Scala 3 build workflow

jaschdoc:
Just wanted to try this out. We can add it if we want to or leave it here. The Scala 3 compiler is more strict than the Scala 2 compiler in migration mode so this way we can always just see if new changes are compatible with Scala 3, i.e., nudges us to remain compatible.

Related to https://github.com/flix/flix/issues/6270


magnus-madsen:
We will stay with Scala 2 with Scala 3 compat, but we don't want to maintain two build configurations.

Issue #8087, opened 2024-07-27
Remove support for `##` fully-qualified Java names


magnus-madsen:

magnus-madsen:
I started on a PR here: https://github.com/flix/flix/pull/8088

magnus-madsen:
@jaschdoc You are also welcome to take a look at this.

The goal is to ensure that we never use ##, but instead import everything that is needed.

jaschdoc:
Do you want me to remove `##` imports from the std lib or remove support for it in the compiler?

magnus-madsen:
> Do you want me to remove `##` imports from the std lib or remove support for it in the compiler?

1. Refactor the standard library and tests
2. Remove support for ##

The latter may require changes to the community.

We may also discover that we are not fully ready to get rid of ##.

PR #8453, opened 2024-08-27
feat: remove support for `##`

jaschdoc:
Closes https://github.com/flix/flix/issues/8087


jaschdoc:
@mlutze @stephentetley the `##` syntax is on its way out so you should import all of your java types with the `import` mechanism once this is merged :)

magnus-madsen:
@jaschdoc Could we split this PR into two: one with the actual change and one with the changes to the tests? 
(Then the latter can be merged immediately). Thanks!

mlutze:
!!!rerun

jaschdoc:
Now depends on #8502

magnus-madsen:
Perhaps this breaks too much of @stephentetley code and we should put it behind the deprecated flag instead?

jaschdoc:
Yeah I thought about that too.

stephentetley:
I'm happy without a deprecated flag.

All my code that use Java interop was broken anyway and I'm starting to get it back compiling again. `charset-locale` which is the one in community build should be up to date with the new interop,

magnus-madsen:
Thanks @stephentetley. We appreciate your patience. It must be a lot of work keeping up with all of our changes. At least, I hope it feels like things are moving in the right direction.

magnus-madsen (approved):

PR #8464, opened 2024-08-28
feat: remove `Indexable.getSwapped` and `IndexableMut.putSwapped`

JonathanStarup:
okay it wasnt that cumbersome

### DIFF
--- a/main/src/library/Vector.flix
+++ b/main/src/library/Vector.flix
@@ -26,7 +26,7 @@ instance Indexable[Vector[a]] {
     type Elm = a
     type Aef = OutOfBounds
 
-    pub def get(i: Int32, t: Vector[a]): a \ OutOfBounds = {
+    pub def get(t: Vector[a], i: Int32): a \ OutOfBounds = {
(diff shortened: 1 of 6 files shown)


main/src/library/Vector.flix
```diff
@@ -26,7 +26,7 @@ instance Indexable[Vector[a]] {
     type Elm = a
     type Aef = OutOfBounds
 
-    pub def get(i: Int32, t: Vector[a]): a \ OutOfBounds = {
+    pub def get(t: Vector[a], i: Int32): a \ OutOfBounds = {
```
magnus-madsen, line 29: I would not swap the arguments for Array and Vector etc. That causes downstream breakages. I was only thinking of swapping them for the traits.

JonathanStarup: Yeah this is only on the instances

magnus-madsen: Ah. I saw wrong.

magnus-madsen (approved):

PR #8444, opened 2024-08-27
feat: java field parsing

JonathanStarup:
now `x.f` and `(exp..).f` parses as a `GetField2`.

This always gives `TypeError.FieldNotFound` since resolution is left as a TODO

PR #8465, opened 2024-08-28
refactor: change java solver interface

JonathanStarup:
Dont merge this before #8444 !

move known type checking into the solver instead of being external


magnus-madsen:
Maybe talk to Matt first? And maybe do some simpler cleanup first too?

JonathanStarup:
The diff looks big but ive just moved code around.

Instead of `isKnown -> lookup` its now `lookup (using isKnown, it might now answer UnresolvedTypes)`

JonathanStarup:
you think its okay @mlutze? its only concerning behaviour inside the java constraints

mlutze:
I think it's OK. I'm planning on ripping a lot of this out anyway.

magnus-madsen (approved):

magnus-madsen:
I merged in the wrong order 💀  sorry

PR #8515, opened 2024-08-30
feat: rerun typer on errors

JonathanStarup:
Simple idea: if there is a type error in a function, rerun without constraint sorting. This transforms this program (#7598)
```scala
def example(rc: Region[r]): Int32 = {
    let _r = Ref.fresh(rc, 42);
    42
}
```
currently has error
```
>> Unable to unify the effect formulas: 'Pure' and 'r'.

2 |     let _r = Ref.fresh(rc, 42);
                           ^^
                           mismatched effect formulas.

Type One: Region[Pure]
Type Two: Region[r]
```
in this PR it has error
```
>> Expected type: 'Pure' but found type: 'r'.

1 |> def example(rc: Region[r]): Int32 = {
2 |>     let _r = Ref.fresh(rc, 42);
3 |>     42
4 |> }

```
Fixes #8509


JonathanStarup:
@magnus-madsen this works very well and has very small code complexity cost and only has performance cost for wrong programs

magnus-madsen:
Did you discuss with Matthew? I assume we get this for free with his solver.

JonathanStarup:
why would it be free?

magnus-madsen:
I suppose I don't understand then.

magnus-madsen:
We can discuss.

PR #8518, opened 2024-08-30
refactor: implement `MutList` via struct

### LINKED ISSUE #8387: Refactor `MutList` to use a struct

sockmaster27:
Fixes #8387


sockmaster27:
@magnus-madsen 
How should we deal with tests for `reserve` and `compress` that currently accesses the capacity of the inner array?
These functions are hard to test when the capacity becomes private, but on the other hand, it probably should be.

magnus-madsen:
> @magnus-madsen How should we deal with tests for `reserve` and `compress` that currently accesses the capacity of the inner array? These functions are hard to test when the capacity becomes private, but on the other hand, it probably should be.

I suppose its OK not to test internals... Blackbox testing.

sockmaster27:
> > @magnus-madsen How should we deal with tests for `reserve` and `compress` that currently accesses the capacity of the inner array? These functions are hard to test when the capacity becomes private, but on the other hand, it probably should be.
> 
> I suppose its OK not to test internals... Blackbox testing.

But should we remove these tests entirely, or just have them assert that calling the function can be called but ignoring its output?

magnus-madsen:
How many tests

sockmaster27:
6\*reserve + 4\*compress = 10

magnus-madsen:
> 6*reserve + 4*compress = 10

Drop em

sockmaster27:
@magnus-madsen 
ping

main/src/library/MutList.flix
```diff
@@ -634,15 +580,13 @@ mod MutList {
...
+            reserve!(length(v), v)
         else ();
-        let sub = Array.copyOfRange(r, i, len, Ref.get(a));
-        Array.updateSequence!(i + 1, sub, Ref.get(a));
-        Array.put(x, i, Ref.get(a));
-        Ref.put(len + 1, l)
+        let sub = Array.copyOfRange(v->r, i, length(v), v->values);
+        Array.updateSequence!(i + 1, sub, v->values);
+        Array.put(x, i, v->values);
+        v->length = length(v) + 1
```
magnus-madsen, line 589: I would replace length(v) with v->length everywhere. I think of the length function as part of the API. There is no reason not to access the struct field directly. In particular this line shows that it would be more clear to access the field directly.

magnus-madsen:
Thanks- this looks good. Just one comment about using v->length instead of length(v).

magnus-madsen (approved):

PR #8523, opened 2024-09-01
feat: add additional base effects

magnus-madsen:
Additional scaffolding for @AStenbaek

### DIFF
--- a/main/src/library/Prelude.flix
+++ b/main/src/library/Prelude.flix
@@ -13,22 +13,58 @@ pub enum Proxy[_] {
 }
 
 ///
-/// The Generic IO Effect.
+/// The uninterpretable `Clock` effect.
 ///
-/// The `IO` effect is the most general effect of any
-/// operation that interacts with the environment.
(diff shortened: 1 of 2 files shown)


JonathanStarup:
its also specialcased in a lot of phases because its non-algebraic.

it would be useful to have any effect be non-algebraic instead of specialcasing a set of them

main/src/library/Prelude.flix
```diff
@@ -13,23 +13,96 @@ pub enum Proxy[_] {
...
 ///
-/// The Generic IO Effect.
+/// The uninterpretable `Clock` effect.
 ///
-/// The `IO` effect is the most general effect of any
-/// operation that interacts with the environment.
+/// The `Clock` effect represents the actions required to access the system clock.
 ///
-pub eff IO
+/// This includes use of `System.currentTimeMilis` and `System.nanoTime`, but also e.g. `Date.now()`.
```
mlutze, line 20: millis

mlutze (approved):

PR #8537, opened 2024-09-02
feat: set formula

JonathanStarup:


magnus-madsen:
Can you give me 3 sentences on to best review the code?

JonathanStarup:
I'm not sure I can offer a nice path. I could break it down into two PR's. One with constructors and one with definitions and operations (size, toString, etc.)

JonathanStarup:
subsumed by #8709

PR #8542, opened 2024-09-03
fix: fix indexer to accurately find struct field uses

dghosef:
Previously, we only used `StructFieldSym` that only had the location of the definition. Now we have the location of the definition and the use when it is used in expressions


magnus-madsen:
!!!jar

magnus-madsen:
@dghosef Unfortunately a conflict was introduced.

main/src/ca/uwaterloo/flix/language/ast/NamedAst.scala
```diff
@@ -49,7 +49,7 @@ object NamedAst {
 
     case class Enum(doc: Ast.Doc, ann: Ast.Annotations, mod: Ast.Modifiers, sym: Symbol.EnumSym, tparams: List[TypeParam], derives: Derivations, cases: List[Declaration.Case], loc: SourceLocation) extends Declaration
 
-    case class Struct(doc: Ast.Doc, ann: Ast.Annotations, mod: Ast.Modifiers, sym: Symbol.StructSym, tparams: List[TypeParam], fields: List[StructField], indices: Map[Name.Label, Int], loc: SourceLocation) extends Declaration
+    case class Struct(doc: Ast.Doc, ann: Ast.Annotations, mod: Ast.Modifiers, sym: Symbol.StructSym, tparams: List[TypeParam], fields: List[StructField], indicesAndLocs: Map[Name.Label, (SourceLocation, Int)], loc: SourceLocation) extends Declaration
```
magnus-madsen, line 52: I would have made the pair (Int, SourceLocation) since we always have source location as the last argument everywhere.

main/src/ca/uwaterloo/flix/language/phase/Namer.scala
```diff
@@ -368,10 +368,11 @@ object Namer {
   /**
     * Performs naming on the given field.
     */
-  private def visitField(struct: Symbol.StructSym, field0: DesugaredAst.StructField, indices: Map[Name.Label, Int])(implicit flix: Flix, sctx: SharedContext): NamedAst.Declaration.StructField = field0 match {
+  private def visitField(struct: Symbol.StructSym, field0: DesugaredAst.StructField, indices: Map[Name.Label, (SourceLocation, Int)])(implicit flix: Flix, sctx: SharedContext): NamedAst.Declaration.StructField = field0 match {
```
magnus-madsen, line 371: Same comment as earlier.

magnus-madsen (approved):

magnus-madsen:
I tested it and it works well, so we should be good to go 👍

PR #8556, opened 2024-09-04
style: match on lists rather than length in TypeReduction.scala

mlutze:
also adds some documentation


main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
```diff
@@ -210,24 +210,37 @@ object TypeReduction {
...
-    val candidateConstructors = clazz.getConstructors.filter(c => isCandidateConstructor(c, ts))
+    val candidateConstructors = clazz.getConstructors.filter(c => isCandidateConstructor(c, ts)).toList
 
-    candidateConstructors.length match {
-      case 0 => JavaConstructorResolutionResult.ConstructorNotFound
-      case 1 =>
-        val tpe = Type.Cst(TypeConstructor.JvmConstructor(candidateConstructors.head), loc)
+    candidateConstructors match {
+      // Case 1: No such constructor. Error.
+      case Nil => JavaConstructorResolutionResult.ConstructorNotFound
```
JonathanStarup, line 217: It adds linear a linear pass that wasn't there before in the cases for 0 and 1. Is this okay? i guess retrieveConstructor is called rarely?

mlutze: It's OK IMO since we're passing linearly over the constructors already in the `filter` function

magnus-madsen (approved):

PR #8557, opened 2024-09-04
style: avoid imperative loop in TypeReduction

mlutze:


main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
```diff
@@ -265,28 +265,34 @@ object TypeReduction {
...
-        for (existing <- methods) {
-          if (
-            existing.getName == method.getName &&
-            isStatic(existing) == isStatic(method) &&
-              existing.getParameterTypes.sameElements(method.getParameterTypes)
-          ) {
-            // could stop early
-            exists = true
-          }
+      // Case 1: Interface. We have to add the methods from Object.
```
JonathanStarup, line 268: I've dislike this `Case 1, 2, 3` or `Case A, B, C`. I don't think it adds anything but you have to maintain the numbering if you change the code. otherwise nice

JonathanStarup (approved):

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
mlutze: Please fix the 113 places in the compiler where we use this pattern.

JonathanStarup: I think its also only called once per constructor, so good point

magnus-madsen (approved):

PR #8560, opened 2024-09-04
proposal: better and simpler block comments

JonathanStarup:
Improves block comments as discussed in  #6531.
- This simplifies the implementation
- This has the property that you can _always_ add a block comment around anything, no matter its content

The rule is simple:
- There is no block comment nesting
- If you open with `/*`, you close with `*/`. If you open with `/**`, you close with `**/`. and so on

This means that there is always some number of `*` such that your block comment lexes correctly. An editor can quite easily find this number by just going through the selection naively, so with that, an editor block comment shorthand just always works.

### DIFF
--- a/examples/larger-examples/program-analysis/domains/Belnap.flix
+++ b/examples/larger-examples/program-analysis/domains/Belnap.flix
@@ -1,6 +1,6 @@
-/**
- * An implementation of the four-valued boolean lattice named for Nuel Belnap.
- */
+///
+/// An implementation of the four-valued boolean lattice named for Nuel Belnap.
+///
 mod Belnap {
 
     ///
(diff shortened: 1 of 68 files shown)


JonathanStarup:
This breaks some existing code because they use `/** some doc comments */`. This could easily be fixed with `/** **/` but all of these cases should actually just be proper flix doc comments with `///` so I did that instead.

magnus-madsen:
Does this support Java-style block comments? E.g.

```java
/**
 * bla bla
 */
```

https://google.github.io/styleguide/javaguide.html#s7.1.1-javadoc-multi-line

or that what we are giving up?

magnus-madsen:
I agree this change would simplify the language, but looking at the implementation: both are pretty small and self-contained?

JonathanStarup:
I think so yeah, its mostly that it is better and simpler to understand as a user. There is always an easy answer if someone has a block comment around some text that doesn't lex. "just use /***" or whatever fits in their case. In the current system there is sometimes no answer, like the rust ticket.

magnus-madsen:
I am open to further discussions. A convincing argument would be if this can significantly speed up incremental compilation. 

But until we take it open again, I am closing this PR.

Issue #8419, opened 2024-08-25
Refactor `KeywordOtherCompleter`


magnus-madsen:
- [ ] Break into `KeywordDeclCompleter` and `KeywordOtherCompleter`.
- [ ] Remove outdated keywords. Check if any keywords are missing.
- [ ] Refactor both completers to use explicit priorities for each keyword based on frequency.
- [ ] Move `SnippetCompleter` into `KeywordDeclCompleter`

magnus-madsen:
@lozander CC

LoZander:
I know the point is to have finer filtering of suggestions given a specific context. Here we want to have fewer and more relevant suggestions when in the decl context. But i'm unsure exactly what this context entails? Is it any top level declaration i.e. type, mod, def, pub etc. or is it just top level function declarations? I'm thinking the greatest level of contextual filtering occurs if we do the latter, as there is no need to be recommended keyword like "mod" when trying to define a top level function, for instance.

LoZander:
oh, but in most declarations, when it is certain what type, there usually aren't anymore keyword to be typed, I suppose. Like, when you've typed `pub def` there aren't really more keywords to be typed, I suppose. But when typing `pub` there are still keywords to be typed, maybe this is the kind of situation where we want to use decl completer?

LoZander:
Or maybe just having the cursor outside as function (either top level of in module) is enough to be in the context where decl completer should be used?

LoZander:
Maybe I started asking question rematurely ^^', because looking at `CompletionProvider.scala` I see that `KeywordOtherCompleter.getCompletions` is invoked specifically when we're in the context of
1. decl of trait
2. decl of enum
3. decl of instance
4. any ofther decl context.

So I assume the keyword to be included in `KeywordDeclCompleter` should be the ones relavent when we're in one of those context. I guess I answered my own question :p

LoZander:
I do have one question left though: in which situations does the completion context have the suntactic context of decl? Obviously whenever we've typed `def`, `instance`, etc., but are we in that context when we've typed `pub` for instance?

magnus-madsen:
Take a look at the parser and at the ASTs.

Declarations are things that go inside modules: defs, traits, etc. etc. etc.

To give an counter-example INSIDE an enum { ... } you cannot have any keywords other than case. So that is not a declaration context.

LoZander:
I see.

[...]

PR #8564, opened 2024-09-06
refactor: completion priority enum

LoZander:
Refators completion priority into an enum. Also simplifies parts of `TypeCompleter`, partly due to incompatibility with new priority enum. Also adds another completion type `TypeBuiltinPolyCompletion` for builtin polymorphic types.

Related to #8419


main/src/ca/uwaterloo/flix/api/lsp/provider/completion/KeywordCompleter.scala
```diff
@@ -29,25 +29,25 @@ object KeywordCompleter {
...
+      Completion.KeywordCompletion("enum"     , Priority.High),
+      Completion.KeywordCompletion("type"     , Priority.High),
+      Completion.KeywordCompletion("instance" , Priority.High),
+      Completion.KeywordCompletion("mod"      , Priority.Low),
+      Completion.KeywordCompletion("eff"      , Priority.Lower),
+      Completion.KeywordCompletion("struct"   , Priority.Lower),
+      Completion.KeywordCompletion("sealed"   , Priority.Lowest),
+      Completion.KeywordCompletion("trait"    , Priority.Lowest),
+      Completion.KeywordCompletion("import"   , Priority.Lowest),
     )
```
magnus-madsen, line 43: Much better. Now we are getting somewhere. 

Are the priorities here based on how frequently the keywords are?

magnus-madsen (approved):

magnus-madsen:
I think this is a good step forward, so I will merge it.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/KeywordCompleter.scala
LoZander: OH, sorry. Yes, the priorities here are. I thought you had referenced the priorities of builtin types

LoZander: but it's based on the priorities compared to all the other keywords in the list, not just the ones that they share a prefix with. That's the issue you pointed out in #8419. Currently, I'm working on that

PR #8573, opened 2024-09-06
refactor: better completion priorities based on occurances in std lib

LoZander:
Introduces better priorities based on occurances in std lib. Before, the priorities were based on the occurances relative to all other items in the set. Now it's only relative to the items that share a first letter with them. For instance, while `def` is more common than `pub`, there's no reason to give one a higher priority than the other, because they would never both occur in list of suggestions anyways.

This PR also introduces `Priority.Default`.

Related to #8419.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/KeywordCompleter.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/KeywordCompleter.scala
@@ -97,19 +97,25 @@ object KeywordCompleter {
     */
   def getOtherKeywords: Iterable[Completion] =
     List(
-      Completion.KeywordCompletion("with"             , Priority.Highest),
-      Completion.KeywordCompletion("law"              , Priority.Higher),
-      Completion.KeywordCompletion("@Test"            , Priority.High),
-      Completion.KeywordCompletion("where"            , Priority.Low),
-      Completion.KeywordCompletion("fix"              , Priority.Low),
+      // @
+      Completion.KeywordCompletion("@Test"            , Priority.Highest),
+      Completion.KeywordCompletion("@ParallelWhenPure", Priority.Low),
+      Completion.KeywordCompletion("@Parallel"        , Priority.Higher),
+      Completion.KeywordCompletion("@LazyWhenPure"    , Priority.Lower),
+      Completion.KeywordCompletion("@Lazy"            , Priority.High),
       Completion.KeywordCompletion("@Deprecated"      , Priority.Lowest),
-      Completion.KeywordCompletion("@Parallel"        , Priority.Lowest),
-      Completion.KeywordCompletion("@ParallelWhenPure", Priority.Lowest),
-      Completion.KeywordCompletion("@Lazy"            , Priority.Lowest),
-      Completion.KeywordCompletion("@LazyWhenPure"    , Priority.Lowest),
-      Completion.KeywordCompletion("Record"           , Priority.Lowest),
-      Completion.KeywordCompletion("redef"            , Priority.Lowest),
-      Completion.KeywordCompletion("Schema"           , Priority.Lowest),
+      // S
+      Completion.KeywordCompletion("Schema"           , Priority.Default),
+      // F
+      Completion.KeywordCompletion("fix"              , Priority.Default),
+      // L
+      Completion.KeywordCompletion("law"              , Priority.Default),
+      // R
+      Completion.KeywordCompletion("redef"            , Priority.High),
+      Completion.KeywordCompletion("Record"           , Priority.Low),
+      // W
+      Completion.KeywordCompletion("with"             , Priority.High),
+      Completion.KeywordCompletion("where"            , Priority.Low),
     )
 
   /**
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/KeywordCompleter.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/KeywordCompleter.scala
@@ -62,42 +62,62 @@ object KeywordCompleter {
     */
   def getExprKeywords: Iterable[Completion] =
     List(
-      "and",
-      "as",
-      "def",
-      "discard",
-      "do",
-      "else",
-      "false",
-      "forA",
-      "forM",
-      "force",
-      "foreach",
-      "from",
-      "if",
-      "inject",
-      "into",
-      "lazy",
-      "let",
-      "match",
-      "new",
-      "not",
-      "or",
-      "par",
-      "query",
-      "region",
-      "select",
-      "solve",
-      "spawn",
-      "struct",
-      "true",
-      "try",
-      "typematch",
-      "unsafe",
-      "use",
-      "without",
-      "yield"
-    ) map (name => Completion.KeywordCompletion(name, Priority.Lower))
+      // A
+      Completion.KeywordCompletion("and"      , Priority.High),
+      Completion.KeywordCompletion("as"       , Priority.Low),
+      // D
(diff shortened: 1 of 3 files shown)


main/src/ca/uwaterloo/flix/api/lsp/provider/completion/KeywordCompleter.scala
```diff
@@ -29,95 +29,128 @@ object KeywordCompleter {
...
+      Completion.KeywordCompletion("select"   , Priority.Higher),
+      Completion.KeywordCompletion("solve"    , Priority.High),
+      Completion.KeywordCompletion("spawn"    , Priority.Low),
+      Completion.KeywordCompletion("struct"   , Priority.Lower),
+      // T
+      Completion.KeywordCompletion("true"     , Priority.Higher),
+      Completion.KeywordCompletion("try"      , Priority.High),
+      Completion.KeywordCompletion("typematch", Priority.Low),
+      // U    
+      // experitments on occurances in stdlib shows *exact* same amount of
```
magnus-madsen, line 111: spelling: experiments and occurences.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/KeywordCompleter.scala
```diff
@@ -29,95 +29,128 @@ object KeywordCompleter {
...
     */
   def getOtherKeywords: Iterable[Completion] =
     List(
-      Completion.KeywordCompletion("with"             , Priority.Highest),
-      Completion.KeywordCompletion("law"              , Priority.Higher),
-      Completion.KeywordCompletion("@Test"            , Priority.High),
-      Completion.KeywordCompletion("where"            , Priority.Low),
-      Completion.KeywordCompletion("fix"              , Priority.Low),
+      // @
+      Completion.KeywordCompletion("@Test"            , Priority.Highest),
```
magnus-madsen, line 128: Sorted?

magnus-madsen:
Now we are cooking 👍

magnus-madsen:
I added some comments on some typos.

magnus-madsen:
I think this design and the way it is documented it much more solid than what was before. Do you agree?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/KeywordCompleter.scala
```diff
@@ -29,95 +29,128 @@ object KeywordCompleter {
...
+      Completion.KeywordCompletion("inject"   , Priority.Low),
+      Completion.KeywordCompletion("into"     , Priority.High),
+      // L
+      Completion.KeywordCompletion("lazy"     , Priority.Low),
+      Completion.KeywordCompletion("let"      , Priority.High),
+      // M
+      Completion.KeywordCompletion("match"    , Priority.Default),
+      // N
+      Completion.KeywordCompletion("new"      , Priority.Low),
+      Completion.KeywordCompletion("not"      , Priority.High),
```
magnus-madsen, line 92: Interesting: I would expect the race between new and not to be tight.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/KeywordCompleter.scala
```diff
@@ -29,95 +29,128 @@ object KeywordCompleter {
...
+      Completion.KeywordCompletion("par"      , Priority.Default),
+      // Q
+      Completion.KeywordCompletion("query"    , Priority.Default),
+      // R
+      Completion.KeywordCompletion("region"   , Priority.Default),
+      // S
+      Completion.KeywordCompletion("select"   , Priority.Higher),
+      Completion.KeywordCompletion("solve"    , Priority.High),
+      Completion.KeywordCompletion("spawn"    , Priority.Low),
+      Completion.KeywordCompletion("struct"   , Priority.Lower),
```
magnus-madsen, line 105: struct is not an expression keyword is it?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/KeywordCompleter.scala
LoZander: oh, right. I only really grouped it by first letter here. I'll sort it proper :)

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/KeywordCompleter.scala
```diff
@@ -29,95 +29,128 @@ object KeywordCompleter {
...
+      // Y
+      Completion.KeywordCompletion("yield"    , Priority.Default)
+    )
            
   /**
     * Miscellaneous keywords.
     */
   def getOtherKeywords: Iterable[Completion] =
     List(
-      Completion.KeywordCompletion("with"             , Priority.Highest),
```
magnus-madsen, line 100: When this PR is merged, next step could be to further breakdown other keywords.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/KeywordCompleter.scala
LoZander: It might have been close. I don't remember the exact number, but not did occur more often

magnus-madsen: Some of these consideration would be good for the eventual report.

LoZander: Yeah. I'll def do a section on my experiments and results. There was also one case where two keywords occured exactly the same amount of times, which was kinda wild and unexpected. It was `unsafe` and `use`. I still chose to give `use` higher priority because I reckon in most code bases `use` will be much more prevalent

LoZander:
> I think this design and the way it is documented it much more solid than what was before. Do you agree?

I think, this PR didn't make a huge difference in readable in and of itself, but the changes in readability from before the work on this keyword completer started entirely as opposed to now, I think that difference can be felt for sure. There was certain design choices (from before I touched and my early additions) which added unnecessary complexity for very little benefit :)

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/KeywordCompleter.scala
magnus-madsen: You are only using the library right? Ignoring test cases..

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/KeywordCompleter.scala
LoZander: Yep. I've started adding missing keywords as well, which has given a bit more insight into the lexer and parser. So I'm looking forward to working with syntactic contexts for finer keyword categories :D

magnus-madsen:
Hit me up when this is ready for another round.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/KeywordCompleter.scala
LoZander: right you are. I'll get on that :)

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/KeywordCompleter.scala
LoZander: yep. I only considered test cases to see how often `@Test` occurred. The other keyword occurances were only checked over the library

LoZander:
@magnus-madsen its ready for another review :)

magnus-madsen (approved):

magnus-madsen:
Thanks!

Issue #7871, opened 2024-06-11
Move `Ast.*` to `shared`


magnus-madsen:
The idea is to get rid of Ast.scala and instead have new files in the shared folder.

Lets do one thing at a time: We can start with, in separate PRs:

- Add shared/Denotation.scala
- Add shared/Polarity.scala

magnus-madsen:
@sockmaster27 You can slowly start on this. One PR per move.

The idea is simply to move things out of the Ast.scala file and into their own file in the shared package.

magnus-madsen:
@sockmaster27 Can we complete this?

sockmaster27:
3 to go

PR #8575, opened 2024-09-06
refactor: move `Denotation` from `Ast` to `shared`

sockmaster27:
Related to #7871


main/src/ca/uwaterloo/flix/language/ast/shared/Denotation.scala
```diff
@@ -0,0 +1,20 @@
+package ca.uwaterloo.flix.language.ast.shared
```
magnus-madsen, line 1: Needs copyright header

PR #8584, opened 2024-09-09
feat: add `Abort` effect

magnus-madsen:


magnus-madsen:
The signature of `run` will change once we have proper set effects.

mlutze (approved):

Issue #8540, opened 2024-09-03
Introduce SoftErrors for Weeder2


magnus-madsen:
I will go through each HardFailure and discuss what can be done:

I suggest to fix each of these in a separate PR

Line 370:
```
case (Some(_), _ :: _) => Validation.toHardFailure(IllegalEnum(ident.loc))
```

This happens when an enum is defined like `enum Foo(Int) { case Bar, case Baz}`. My suggestion: Use a soft failure where we combine all the cases in the body with the "implicit" case Foo. In other words, think of the enum as having cases `Foo, Bar, Baz`. 

Line 439: Same as above.

Line 679 and 685:

```
case _ => Validation.toHardFailure(IllegalEqualityConstraint(tree.loc))
```

Will need to ask @mlutze How to make this resilient? 

Line 781:

```
            case (None, Presence.Required) => Validation.toHardFailure(MissingFormalParamAscription(ident.name, tree.loc))
            case (Some(_), Presence.Forbidden) => Validation.toHardFailure(IllegalFormalParamAscription(tree.loc))
```

In case 1, the missing case, I suggest to simply add Type.Error. In case 2, I suggest to just ignore the ascribed type.

Line 1385 and friends:

```
case other => Validation.toHardFailure(UnsupportedRestrictedChoicePattern(isStar, other.loc))
case other => Validation.toHardFailure(UnsupportedRestrictedChoicePattern(isStar, other.loc))
case other => Validation.toHardFailure(UnsupportedRestrictedChoicePattern(isStar, other.loc))
```

I think -- in all three cases -- the solution is to add `WeededAst.RestrictableChoosePattern.Var` and just propagate that through the compiler.

magnus-madsen:
@mlutze Suggests that this:

```
    private def visitEqualityConstraint(tree: Tree): Validation[EqualityConstraint, CompilationMessage] = {
      flatMapN(traverse(pickAll(TreeKind.Type.Type, tree))(Types.visitType)) {
        case t1 :: t2 :: Nil => t1 match {
          case Type.Apply(Type.Ambiguous(qname, _), t11, _) => Validation.success(EqualityConstraint(qname, t11, t2, tree.loc))
          case _ => Validation.toHardFailure(IllegalEqualityConstraint(tree.loc))
        }
        case _ => Validation.toHardFailure(IllegalEqualityConstraint(tree.loc))
      }
    }
```

[...]

PR #8586, opened 2024-09-09
feat: use `SoftFailure` in `Weeder2` at `pickRestrictableChoosePattern`

jaschdoc:


main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1389,16 +1389,27 @@ object Weeder2 {
...
+              val e = UnsupportedRestrictedChoicePattern(isStar, other.loc)
+              val ident = Name.Ident(s"x${flix.genSym.freshId()}", other.loc.asSynthetic)
+              Validation.toSoftFailure(List(WeededAst.RestrictableChoosePattern.Var(ident, other.loc.asSynthetic)), e)
           }
           mapN(inner)(RestrictableChoosePattern.Tag(qname, _, loc))
-        case other => Validation.toHardFailure(UnsupportedRestrictedChoicePattern(isStar, other.loc))
+        case other =>
+          val e = UnsupportedRestrictedChoicePattern(isStar, other.loc)
+          val ident = Name.Ident(s"x${flix.genSym.freshId()}", other.loc.asSynthetic)
+          val patVar = WeededAst.RestrictableChoosePattern.Var(ident, other.loc)
```
mlutze, line 1406: Why are we using `Var` here instead of a proper error node?

magnus-madsen: Yeah, we should definitely have a proper error node.

jaschdoc: Tried adding an error node as a pattern but I can also try to do it as a var

jaschdoc:
Tried with the var in https://github.com/flix/flix/pull/8608

jaschdoc:
!!!rerun

magnus-madsen:
@jaschdoc Flixball is failing for unrelated reasons.

mlutze:
!!!rerun

magnus-madsen:
RFR?

jaschdoc:
Yes

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1389,16 +1389,24 @@ object Weeder2 {
                 case Pattern.Wild(loc) => Validation.success(WeededAst.RestrictableChoosePattern.Wild(loc))
                 case Pattern.Var(ident, loc) => Validation.success(WeededAst.RestrictableChoosePattern.Var(ident, loc))
                 case Pattern.Cst(Ast.Constant.Unit, loc) => Validation.success(WeededAst.RestrictableChoosePattern.Wild(loc))
-                case other => Validation.toHardFailure(UnsupportedRestrictedChoicePattern(isStar, other.loc))
+                case other =>
+                  val e = UnsupportedRestrictedChoicePattern(isStar, other.loc)
+                  val ident = Name.Ident(s"x${flix.genSym.freshId()}", other.loc.asSynthetic)
+                  Validation.toSoftFailure(WeededAst.RestrictableChoosePattern.Var(ident, other.loc.asSynthetic), e)
```
magnus-madsen, line 1395: This should use the error construct right?

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1389,16 +1389,24 @@ object Weeder2 {
...
               }
-            case other => Validation.toHardFailure(UnsupportedRestrictedChoicePattern(isStar, other.loc))
+            case other =>
+              val e = UnsupportedRestrictedChoicePattern(isStar, other.loc)
+              val ident = Name.Ident(s"x${flix.genSym.freshId()}", other.loc.asSynthetic)
+              Validation.toSoftFailure(List(WeededAst.RestrictableChoosePattern.Var(ident, other.loc.asSynthetic)), e)
```
magnus-madsen, line 1400: ditto

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
jaschdoc: See the other PR i opened

magnus-madsen:
Wait. This is outdated right?

magnus-madsen:
Or not. I am so confused.

jaschdoc:
2 types of error nodes. Here I added one where Tag would be used and in the other PR I added error nodes instead of vars. I'm not sure which one is preferred or if we need both (hence two PRs) 😀

magnus-madsen:
> 2 types of error nodes. Here I added one where Tag would be used and in the other PR I added error nodes instead of vars. I'm not sure which one is preferred or if we need both (hence two PRs) 😀

What do you mean a tag? It should always be an error node.

magnus-madsen:
I am still confused about these two PRs and which should be merged when and whatnot.

What I expected was:

- Introduce RestrictablePattern.Error (or whatever it is called) and propagate it in every phase.
- Refactor the error to use Pattern.Error.

jaschdoc:
I understand.

There are two patterns: a `Var` pattern and a `Tag` pattern which are disjoint patterns. In both paths of the code the same error is used which is a problem since one branch returns a `HardFailure` and the other one returns a `SoftFailure` which places different type requirements on the error.

I decided to split the work into two distinct PRs that deal with each path. This one has the minimal amount of changes (the `Tag` path was a bit more complex). So once this is merged into main, we can merge the other PR with these changes which should arrive at what you wanted.

This is all just to introduce the changes incrementally and to avoid breaking things.

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1400,13 +1400,11 @@ object Weeder2 {
           mapN(inner)(RestrictableChoosePattern.Tag(qname, _, loc))
         case other =>
           val e = UnsupportedRestrictedChoicePattern(isStar, other.loc)
-          val patVar = WeededAst.RestrictableChoosePattern.Error(other.loc)
-          val qname = Name.mkQName(s"choosePattern${flix.genSym.freshId()}", other.loc.asSynthetic)
-          Validation.toSoftFailure(RestrictableChoosePattern.Tag(qname, List(patVar), other.loc.asSynthetic), e)
+          Validation.toSoftFailure(RestrictableChoosePattern.Error(other.loc), e)
```
jaschdoc, line 1403: this is the important change

jaschdoc:
RFR :)

magnus-madsen:
@mlutze Can you do a first review?

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -968,6 +968,7 @@ object Lowering {
           val tagSym = visitRestrictableCaseSymUse(sym)
           val p = LoweredAst.Pattern.Tag(tagSym, pat1, tpe, loc)
           LoweredAst.MatchRule(p, None, e)
+        case TypedAst.RestrictableChoosePattern.Error(_, loc) => throw InternalCompilerException("unexpected error restrictable choose pattern near", loc)
```
mlutze, line 971: Is it standard to use "near" in ICEs?

main/src/ca/uwaterloo/flix/language/phase/typer/RestrictableChooseConstraintGen.scala
```diff
@@ -75,99 +76,115 @@ object RestrictableChooseConstraintGen {
       case KindedAst.Expr.RestrictableChoose(false, exp0, rules0, tpe0, loc) =>
 
         // Get the enum symbols for the matched type
-        val enumSym = rules0.headOption.getOrElse(throw InternalCompilerException("unexpected empty choose", loc)).pat match {
-          case KindedAst.RestrictableChoosePattern.Tag(sym, _, _, _) => sym.sym.enumSym
+        val enumSym0 = rules0.headOption.flatMap {
+          case rule => rule.pat match {
```
mlutze, line 80: Can we document a bit what's going on here? This place is already confusing :)

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
JonathanStarup: I don't think so, especially since ICE doesn't print it location, so it will always just say `unexpected error restrictable choose pattern near`

magnus-madsen: Not really. In some places we do it. We should probably think of something. 

The location is already there, so not super relevant to repeat.

magnus-madsen:
Lets fix Matts second comment and then we can go.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
JonathanStarup: the location is not not printed though, so you cant see it

JonathanStarup: throw ICE(msg, loc) and throw ICE(msg, SourceLocation.Unknown) gives the exact same information

jaschdoc: So what do we agree on?

jaschdoc: I'll remove `near` for now but we should figure something out. I'll bring this up next meeting, maybe just a quick vote?

jaschdoc:
RFR

main/src/ca/uwaterloo/flix/language/phase/typer/RestrictableChooseConstraintGen.scala
```diff
@@ -75,99 +76,114 @@ object RestrictableChooseConstraintGen {
       case KindedAst.Expr.RestrictableChoose(false, exp0, rules0, tpe0, loc) =>
 
         // Get the enum symbols for the matched type
-        val enumSym = rules0.headOption.getOrElse(throw InternalCompilerException("unexpected empty choose", loc)).pat match {
-          case KindedAst.RestrictableChoosePattern.Tag(sym, _, _, _) => sym.sym.enumSym
+        rules0.headOption match {
+          case None => throw InternalCompilerException("unexpected empty choose", loc)
+          case Some(rule) => rule.pat match {
+            case RestrictableChoosePattern.Error(_, loc) => // If the pattern is an error we just return fresh error variables
+              (Type.freshError(Kind.Error, loc), Type.freshError(Kind.Error, loc))
```
mlutze, line 83: Do we use `Kind.Error` elsewhere like this? I thought `Kind.Error` is only when we don't know the kind(?)

main/src/ca/uwaterloo/flix/language/phase/typer/RestrictableChooseConstraintGen.scala
```diff
@@ -75,99 +76,114 @@ object RestrictableChooseConstraintGen {
...
-        val resTpe = enumTypeOut
-        val resEff = Type.mkUnion(eff :: effs, loc)
-        (resTpe, resEff)
+            // τ_out
+            c.unifyType(enumTypeOut, tpe0, loc)
+            val resTpe = enumTypeOut
+            val resEff = Type.mkUnion(eff :: effs, loc)
+            (resTpe, resEff)
 
+          case None => (Type.freshError(Kind.Error, loc), Type.freshError(Kind.Error, loc))
```
mlutze, line 185: Ditto

main/src/ca/uwaterloo/flix/language/phase/typer/RestrictableChooseConstraintGen.scala
jaschdoc: I think I just assumed that this was the right thing to do. I can turn it into a `Star` instead :)

main/src/ca/uwaterloo/flix/language/phase/typer/RestrictableChooseConstraintGen.scala
```diff
@@ -75,99 +76,114 @@ object RestrictableChooseConstraintGen {
...
-        val resTpe = enumTypeOut
-        val resEff = Type.mkUnion(eff :: effs, loc)
-        (resTpe, resEff)
+            // τ_out
+            c.unifyType(enumTypeOut, tpe0, loc)
+            val resTpe = enumTypeOut
+            val resEff = Type.mkUnion(eff :: effs, loc)
+            (resTpe, resEff)
 
+          case None => (Type.freshError(Kind.Star, loc), Type.freshError(Kind.Star, loc))
```
mlutze, line 185: The second one should actually be Kind.Eff, since that's the inferred effect.

mlutze (approved):

magnus-madsen (approved):

PR #8591, opened 2024-09-09
feat: add `Console` effect

magnus-madsen:


magnus-madsen:
The diff is hard to read, but this is the essence:

```scala
///
/// An effect used to interact with the console.
///
eff Console {

    ///
    /// Reads a single line from the console.
    ///
    pub def readln(): String

    ///
    /// Prints the given string `s` to the standard out.
    ///
    pub def print(s: String): Unit

    ///
    /// Prints the given string `s` to the standard err.
    ///
    pub def printErr(s: String): Unit

    ///
    /// Prints the given string `s` to the standard out followed by a new line.
    ///
    pub def println(s: String): Unit

    ///
    /// Prints the given string `s` to the standard err followed by a new line.
    ///
    pub def printlnErr(s: String): Unit

}

mod Console {

    import java.lang.System

    ///
    /// Runs the `Console` effect of the given function `f`.
    ///
    /// In other words, re-interprets the `Console` effect using the `IO` effect.
    ///
    pub def run(f: Unit -> a \ Console): a \ IO =
        try {
            f()
        } with Console {
            def readln(k)        = k(System.console().readLine())
            def print(s, k)      = { System.out.print(s); k() }
            def printErr(s, k)   = { System.err.print(s); k() }
            def println(s, k)    = { System.out.println(s); k() }
            def printlnErr(s, k) = { System.err.println(s); k() }
        }

}
```

examples/effects/InputOutput.flix
```diff
@@ -21,11 +23,8 @@ mod InputOutputExample {
 
     def runWithConsole(f: Unit -> t \ ef + InputOutput): Result[String, t] \ ef + IO = {
         try Ok(f()) with InputOutput {
-            def read(k) = match Console.readLine() {
-                case Err(msg) => Err(msg)
-                case Ok(line) => k(line)
-            }
-            def write(s, k) = {Console.println(s); Console.flush(); k()}
+            def read(k) = k(System.console().readLine())
```
JonathanStarup, line 26: This can give nullpointer exception, `.console()` can be null

JonathanStarup:
The effect needs flush? otherwise the handler should flush after each print i think

JonathanStarup (approved):

PR #8603, opened 2024-09-10
fix: `Console.run`

### LINKED ISSUE #8602: Can't call `Console.run`
The definition of run
```
pub def run(f: Unit -> a \ Console): a \ IO
```
is very limited. I cant use it in Flix-ANSI-Terminal because i have effect `Console + IO`

JonathanStarup:
Fixes #8602


JonathanStarup:
@magnus-madsen this is needed to make my library run (I've verified that this works)

magnus-madsen (approved):

PR #8604, opened 2024-09-10
feat: thunk shorthand `-> exp`

JonathanStarup:
parse `-> exp` as `() -> exp`

This is very nice with handler-type functions like `def run(f: Unit -> a \ ef): a \ (ef - Console) + IO`

`run(() -> exp)` becomes `run(-> exp)`

a very small change, most of the diff is changing `() -> exp` to `-> exp` in lib


magnus-madsen:
Lets talk more. In particular, what are your thoughts on going full "block mode"? 

In particular, I think `while(-> ...)(-> ...)` is just not good enough. Thoughts?

JonathanStarup:
yeah I've also though about that. but `while{..}{..}` also seems to implicit and magic

magnus-madsen:
We can discuss, but I don't think any programmers will accept having to write `while(-> ...)(-> ...)`.

JonathanStarup:
I think its better than `while{..}{..}` unless its built-in syntax

PR #8612, opened 2024-09-11
feat: Set formula part 1

JonathanStarup:
Add SetFormula except the smart constructors


JonathanStarup:
@magnus-madsen this is more digestable

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
```diff
@@ -0,0 +1,338 @@
...
+
+package ca.uwaterloo.flix.language.phase.unification
+
+import scala.annotation.nowarn
+import scala.collection.immutable.SortedSet
+
+object FastSetUnification {
+
+  /**
+    * A common super-type for set formulas `f`, like `x1 ∩ x2 ∪ (e4 ∪ !c17)`.
```
magnus-madsen, line 25: Here !c17 means the complement of c17, correct?

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
```diff
@@ -0,0 +1,338 @@
...
+    *
+    * A set formula can contain elements, constants (fixed unknown sets), variables (unknown sets),
+    * complements, unions, and intersections.
+    *
+    * Variables and constants are collectively referred to as unknowns.
+    * A formula without unknowns is called ground.
```
magnus-madsen, line 31: OK

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
```diff
@@ -0,0 +1,338 @@
...
+    /** The set of [[Var]] in `this`. */
+    final def variables: SortedSet[Int] = this match {
+      case Univ => SortedSet.empty
+      case Empty => SortedSet.empty
+      case Cst(_) => SortedSet.empty
+      case Var(x) => SortedSet(x)
+      case ElemSet(_) => SortedSet.empty
+      case Compl(t) => t.variables
+      case Inter(_, _, varsPos, _, _, varsNeg, other) =>
+        SortedSet.from(
```
magnus-madsen, line 52: Whats your reason for using iterators? As opposed to either folding or just ++?

magnus-madsen: My worry is that -- some refactorings later -- iterators can be error-prone. And that the performance gain is not worth it?

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
```diff
@@ -0,0 +1,338 @@
...
+    *
+    * Invariant: [[SetFormula.ElemSet]], [[SetFormula.Cst]], and [[SetFormula.Var]] must use
+    * disjoint integers (see [[unknowns]]).
+    */
+  sealed trait SetFormula {
+
+    import SetFormula.*
+
+    /** The set of [[Var]] in `this`. */
+    final def variables: SortedSet[Int] = this match {
```
magnus-madsen, line 44: May I propose freeVars? (We use that terminology elsewhere).

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
```diff
@@ -0,0 +1,338 @@
...
+      case Union(_, _, varsPos, _, _, varsNeg, other) =>
+        SortedSet.from(
+          varsPos.iterator.map(_.x) ++
+            varsNeg.iterator.map(_.x) ++
+            other.iterator.flatMap(_.variables)
+        )
+    }
+
+    /** `true` if `this` contains neither [[Var]] nor [[Cst]]. */
+    final def ground: Boolean = this match {
```
magnus-madsen, line 66: -> isGround

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
```diff
@@ -0,0 +1,338 @@
...
+          other.forall(_.ground)
+      case Union(_, cstsPos, varsPos, _, cstsNeg, varsNeg, other) =>
+        cstsPos.isEmpty &&
+          varsPos.isEmpty &&
+          cstsNeg.isEmpty &&
+          varsNeg.isEmpty &&
+          other.forall(_.ground)
+    }
+
+    /** Returns all [[Var]] and [[Cst]] that occur in `this`. */
```
magnus-madsen, line 87: Add a comment that the domains ofVar and Cst are assumed to be disjoint, so it is okay to return them in one set instead of in a pair of sets.

magnus-madsen: Nevertheless, when do you need this function? Why is it not "better" to have varsOf, cstsOf etc?

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
```diff
@@ -0,0 +1,338 @@
...
+            other.iterator.flatMap(_.unknowns)
+        )
+    }
+
+    /**
+      * Returns the number of connectives in the unary/binary representation of `this`.
+      *
+      * [[Compl]], [[Union]], and [[Inter]] are connectives.
+      */
+    final def size: Int = {
```
magnus-madsen, line 118: What's this use for? For the complexity measure in TooComplex etc?

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
```diff
@@ -0,0 +1,338 @@
...
+        val negCstsSize = cstsNeg.size
+        val negVarsSize = varsNeg.size
+        val subformulas = elemPos.size + cstsPos.size + varsPos.size +
+          negElemSize + negCstsSize + negVarsSize + other.size
+        val connectives = negElemSize + negCstsSize + negVarsSize
+        counter += (subformulas - 1) + connectives
+        workList = other ++ workList
+      }
+
+      while (workList.nonEmpty) {
```
magnus-madsen, line 139: Why is this a fixpoint? I mean its okay, but could have been a recursive definition?

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
```diff
@@ -0,0 +1,338 @@
...
+
+  }
+
+  object SetFormula {
+
+    /** Skip invariant checks if `false`. */
+    private val CHECK_INVARIANTS: Boolean = true
+
+    /** A trait used to refer to [[Univ]] and [[Empty]] collectively. */
+    sealed trait UnivOrEmpty extends SetFormula
```
magnus-madsen, line 188: How is this later used?

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
```diff
@@ -0,0 +1,338 @@
...
+    final case object Univ extends SetFormula with UnivOrEmpty
+
+    /** The empty set (`empty`). */
+    final case object Empty extends SetFormula with UnivOrEmpty
+
+    /**
+      * An uninterpreted constant set (`c42`), i.e. a set that we do not know and cannot make
+      * assumptions about.
+      *
+      * Invariant: [[ElemSet]], [[Cst]], and [[Var]] must use disjoint integers.
```
magnus-madsen, line 200: Thanks for repeating this 👍

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
```diff
@@ -0,0 +1,338 @@
...
+      */
+    final case class Cst(c: Int) extends SetFormula
+
+    /**
+      * A set variable (`x42`), i.e. a set that we do not know but should learn assumptions about.
```
magnus-madsen, line 205: > but should learn assumptions about.

Perhaps something more like, a variable that can be instantiated by unification?

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
```diff
@@ -0,0 +1,338 @@
...
+    @nowarn
+    final case class Inter private(
+                                    elemPos: Option[ElemSet], cstsPos: Set[Cst], varsPos: Set[Var],
+                                    elemNeg: Option[ElemSet], cstsNeg: Set[Cst], varsNeg: Set[Var],
+                                    other: List[SetFormula]
+                                  ) extends SetFormula {
+      if (CHECK_INVARIANTS) {
+        assert(!varsPos.exists(varsNeg.contains), message = this.toString)
+        assert(!cstsPos.exists(cstsNeg.contains), message = this.toString)
+        assert(subformulasOf(elemPos, cstsPos, varsPos, elemNeg, cstsNeg, varsNeg, other).take(2).toList.size == 2, message = this.toString)
```
magnus-madsen, line 277: Whats this?

magnus-madsen:
Mostly questions with a few minor change requests.

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
JonathanStarup: In propagation, there is a partial substitution only to Univ or Empty so its just for added type safety / pattern match exhaustiveness

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
JonathanStarup: yes

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
JonathanStarup: I think its a bit misleading when there is no of bound/free distinction here, variables are just variables

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
JonathanStarup: I don't know if the performance is noticeable, but I don't think it hurts here because the iterator is immediately created and given to `SortedSet`. it doesn't seem error prone

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
JonathanStarup: constants are never needed by themselves, so having unknowns seems better (one traversal for the union)

Its a general pattern i use to ask questions directly (t.freeVarsContains(x)) instead of general methods (t.variables.contains(x)) to have better performance

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
JonathanStarup: yea just debugging and tooComplex

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
JonathanStarup: It can't be tail-recursive if it is a recursive function, so I did this instead

JonathanStarup: its a bit more complex but its also self-contained so this complexity doesn't influence other functions etc.

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
JonathanStarup: Checking the invariants mentioned in the doc comment

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
magnus-madsen: My question is this: Why not just write a standard recursive function? That's what we do elsewhere. In only one place to we use iterators (SemanticTokens) and there we do it because benchmarking shows its critical for performance (the lists are very long).

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
magnus-madsen: OK. We keep it then. But in most places we use freeVars irrespective of whether there are binders.

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
magnus-madsen: But again, does it have to be tail recursive?

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
magnus-madsen: My question was as to: specifically what do we check for formulas? And why is not a recursive call perhaps?

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
JonathanStarup: i like it because
- faster: this is a bottle neck of performance so everything counts
- low maintenance cost: its selfcontained
- better errors: its tailrecursive, so it allows us to give TooComplex more often instead of StackOverflow
- low effort: its already written

If anything is wrong/weird its easy to go back to recursive def. I think this is better

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
JonathanStarup: i like it because
- faster: this is a bottle neck of performance so everything counts
- low maintenance cost: its selfcontained
- better errors: its tailrecursive, so it allows us to give TooComplex more often instead of StackOverflow
- low effort: its already written

If anything is wrong/weird its easy to go back to recursive def. I think this is better

magnus-madsen: Then add those comments.

main/src/ca/uwaterloo/flix/language/phase/unification/FastSetUnification.scala
JonathanStarup: I've added comments. I don't follow with the recursive call

JonathanStarup:
@magnus-madsen done a new pass over

magnus-madsen (approved):

PR #8621, opened 2024-09-11
feat: add `Exit` effect

magnus-madsen:


magnus-madsen:
We should rework the check for permitted top-level effects, but I think we need to wait for proper sets.

magnus-madsen:
RFR

JonathanStarup:
Note that all functions that have effects other than IO will have full CPS instrumentation since these changes are purely frontend

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -216,13 +216,14 @@ object Safety {
   }
 
   /**
-    * Returns `true` if the given `defn` is pure or has the IO effect or has the NonDet effect.
+    * Returns `true` if the given `defn` is pure or has an effect that is allowed for a top-level function.
     */
-  private def isPureOrIOOrNonDet(defn: Def): Boolean = {
+  private def isAllowedEffect(defn: Def): Boolean = {
```
mlutze, line 221: `isAllowedEntrypointEffect`?

main/src/library/Exit.flix
```diff
@@ -0,0 +1,51 @@
...
+
+///
+/// An effect used to exit the JVM.
+///
+eff Exit {
+
+    ///
+    /// Immediately exits the JVM with the specified `exitCode`.
+    ///
+    pub def exit(exitCode: Int32): Void
```
mlutze, line 25: Cool, now we can add exit hooks!

main/src/ca/uwaterloo/flix/language/phase/EntryPoint.scala
```diff
@@ -185,7 +185,7 @@ object EntryPoint {
       val resultSc = Scheme.generalize(Nil, Nil, resultTpe, RigidityEnv.empty)
 
       // Check for [[IllegalEntryPointEffect]]
-      if (declaredEff != Type.Pure && (declaredEff != Type.IO && declaredEff != Type.NonDet)) {
+      if (declaredEff != Type.Pure && (declaredEff != Type.IO && declaredEff != Type.NonDet && declaredEff != Type.Sys)) {
```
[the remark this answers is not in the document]
magnus-madsen: We really need sets here.

magnus-madsen:
> Note that all functions that have effects other than IO will have full CPS instrumentation since these changes are purely frontend
> 
> Do you want to add all these first and the refactor to make them behave consistently with IO?

Yes, we need to fix that too. But we need sets to accurately compute the actual effects.

JonathanStarup:
Once the effects are ground, they can be computed exactly, which is done in purity.fromType. it only has IO. i can make a pr tomorrow

Issue #8626, opened 2024-09-12
Rename `debug` commands to `dbg` to free up `debug` for `Logger`


magnus-madsen:
- [x] Rename `debug` and friends to `dbg` etc.
- [x] Rename `Logger.debug1` to `Logger.debug`
- [x] Update doc.flix.dev

magnus-madsen:
Requires some changes to the Lexer/Parser plus Prelude.flix.

JonathanStarup:
couldnt we add a ! to each, so `debug!`, `debug!!`, and `debug!!!`

magnus-madsen:
That's also possible.

sockmaster27:
> Is debug treated as a keyword?

Yes. Both in the compiler and in highlighting.

sockmaster27:
What's the strategy with keeping highlight.js etc. up to date without spamming the maintainers?

magnus-madsen:
We maintain our own fork of highlight.js-- they are not accepting upstream PRs for a while now.

For other packages, I think the point is just to update more rarely, so we can bundle more changes into one PR.

sockmaster27:
Ok, so it seems like the only upstream highlighting we have is Ace, right?
Can someone with permissions make a flix/ace fork where we can buffer the changes?

magnus-madsen:
> Ok, so it seems like the only upstream highlighting we have is Ace, right? Can someone with permissions make a flix/ace fork where we can buffer the changes?

I think it can wait. I propose when Linquist is eventually merge we can do a big round of updating all the other stuff.

PR #8627, opened 2024-09-12
feat: add `Logger` effect

magnus-madsen:


main/src/library/Logger.flix
```diff
@@ -0,0 +1,73 @@
...
+    pub def log(s: Severity, m: String): Unit
+
+}
+
+mod Logger {
+
+    ///
+    /// Logs the message `m` at the `Trace` level.
+    ///
+    pub def trace(m: a): Unit \ Logger with ToString[a] = do Logger.log(Severity.Trace, "${m}")
```
mlutze, line 34: Is this interface typical? I would expect only to debug strings.

main/src/library/Logger.flix
```diff
@@ -0,0 +1,73 @@
...
+
+    ///
+    /// Logs the message `m` at the `Debug` level.
+    ///
+    pub def debug1(m: a): Unit \ Logger with ToString[a] = do Logger.log(Severity.Debug, "${m}")
```
mlutze, line 39: I guess there is a TODO here

mlutze (approved):

main/src/library/Logger.flix
magnus-madsen: No strong opinion, but this way you can debug a number too

main/src/library/Logger.flix
magnus-madsen: Yes, with ticket: https://github.com/flix/flix/issues/8626

Issue #8399, opened 2024-08-22
Resolver: introduce ApplyDef, ApplySig, ApplyClo


mlutze:
Maybe generalize over them

jaschdoc:
Maybe this could be a field in `Apply`?

JonathanStarup:
They have different lengths of arguments so it seems nice to have three cases

magnus-madsen:
What @JonathanStarup said.

magnus-madsen:
Also backend uses this naming convention.

jaschdoc:
Should we also add an `ApplyVar` for the case below? I assume `ApplyClo` is for direct application of a lambda, correct?
```scala
let f = x -> x + 1;
f(1)
```

mlutze:
> Should we also add an `ApplyVar` for the case below? I assume `ApplyClo` is for direct application of a lambda, correct?
> ```scala
> let f = x -> x + 1;
> f(1)
> ```

I don't think so. `Clo` is any applied expression, including e.g., another application:

`(id(foo))(xyz)`

jaschdoc:
Should we also rename `Lambda` to `LambdaClosure`? It seems like it is the same thing situation (in terms of naming) when it comes to `ClosureConversion`, i.e., CC eliminates `Lambda`. We could also just rename `LambdaClosure` to `Lambda` in the backend.

magnus-madsen:
> Should we also rename `Lambda` to `LambdaClosure`? It seems like it is the same thing situation (in terms of naming) when it comes to `ClosureConversion`, i.e., CC eliminates `Lambda`. We could also just rename `LambdaClosure` to `Lambda` in the backend.

No, leave as is. Its not the same.

jaschdoc:
Yeah but that would mean fewer `ICE`s

magnus-madsen:
> Yeah but changing it would mean fewer `ICE`s

OK, maybe I am stupid then. Lets discuss at the meeting.

PR #8635, opened 2024-09-13
fix: allow importing ops

mlutze:


magnus-madsen:
Everything has failed

mlutze:
I will wait for https://github.com/flix/flix/issues/8399 so I can use the same system for ops.

magnus-madsen:
I have created a ticket here: #9036 so we can resume later.

PR #8641, opened 2024-09-14
feat: add `ApplyDef` in `ResolvedAst`

jaschdoc:
Related to https://github.com/flix/flix/issues/8399


main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
```diff
@@ -1022,6 +1031,10 @@ object Kinder {
       Validation.success(KindedAst.Expr.Error(m, tvar, evar))
   }
 
+  private def visitDefn(defn0: ResolvedAst.Expr.Def)(implicit scope: Scope, flix: Flix): Validation[KindedAst.Expr.Def, KindError] = defn0 match {
```
magnus-madsen, line 1034: Why a separate function here?

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -157,6 +158,24 @@ object ConstraintGen {
             (resTpe, resEff)
         }
 
+      case Expr.ApplyDef(KindedAst.Expr.Def(sym, tvar1, loc1), exps, tvar, evar, loc) =>
```
magnus-madsen, line 161: The logic from above, under known target, should now be redundant

main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
jaschdoc: If we want to avoid a type cast in `visitExp` in the `ApplyDef` case then we need a function that returns a `KindedAst.Expr.Def` so we can construct a `KindedAst.Expr.ApplyDef` which takes not just an `Expr` but a `Expr.Def`

jaschdoc:
I'm not quite sure why it crashes now. Seems there is an invariant that is no longer maintained in `SimplifiedAst` since `ClosureConv` previously added `ApplyDef` by pattern matching on `exp` in `Apply` so there was a path in the compiler that was impossible which has now been made possible by constructing the `ApplyDef` in `Simplifier`. Maybe @JonathanStarup knows what's wrong?

JonathanStarup:
- put defSymUse into APplyDef instead of exp
- maybe it works out now?

jaschdoc:
> * put defSymUse into APplyDef instead of exp
> 
>     * maybe it works out now?

Didn't really seem to do anything:
```
A needed class was not found. This could be due to an error in your runpath. Missing class: Test/Currying/Clo$g
java.lang.NoClassDefFoundError: Test/Currying/Clo$g
	at Test.Currying.Def$testCurry03.applyFrame(Unknown Source)
	at Test.Currying.Def$testCurry03.invoke(Unknown Source)
	at Test.Currying.m_testCurry03(Unknown Source)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at ca.uwaterloo.flix.language.phase.jvm.JvmBackend$.$anonfun$link$2(JvmBackend.scala:240)
...
```

jaschdoc:
Maybe the problem is now in `ApplyClo` because it seems like it's missing `g`

main/src/ca/uwaterloo/flix/language/phase/Monomorpher.scala
```diff
@@ -520,6 +520,15 @@ object Monomorpher {
       val es = exps.map(visitExp(_, env0, subst))
       MonoAst.Expr.Apply(e, es, subst(tpe), subst(eff), loc)
 
+    case LoweredAst.Expr.ApplyDef(Ast.DefSymUse(sym, loc1), exps, tpe, eff, loc2) =>
+      val spec = root.defs(sym).spec
+      val declaredTypes = spec.fparams.map(_.tpe)
+      // Is this correct or should this be curried? Maybe done in ConstraintGen?
+      val tpe1 = Type.mkApply(Type.Cst(TypeConstructor.Arrow(declaredTypes.length), loc1), spec.eff :: declaredTypes, loc1)
```
mlutze, line 527: Looks like you're missing a return type here. There's something like Type.mkArrowWithEffect, which you may find useful

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -102,13 +102,6 @@ object ConstraintGen {
         // - have better performance (we don't generate unnecessary type variables)
         //
         val knownTarget = exp match {
-          case KindedAst.Expr.Def(sym, tvar1, loc1) =>
```
mlutze, line 105: I wonder... do we even need Expr.Def anymore?

JonathanStarup: we do not no, pretty cool

magnus-madsen: Very cool. Remove it.

magnus-madsen:
@mlutze and @JonathanStarup you can help @jaschdoc with this. I would still like to review the code before its merged.

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
jaschdoc: Yeah really cool. I definitely will :+1:

mlutze:
@jaschdoc Can we do the removal in a separate PR?

jaschdoc:
Removal of what?

mlutze:
> Removal of what?

Removal of `*Ast.Def`

jaschdoc:
Expr.Def?

I think it's going to be difficult though. What would you replace it with if not ApplyDef?

mlutze:
What I mean is, we do this thing where we construct the ApplyDef from a Def node in Resolver (or whatever it is) in this PR and we handle it in all the phases. But otherwise we do not touch Ast.Expr.Def. 

This keeps this PR clean and focused, and then we can deal with complications in a separate PR.

Does that seem reasonable?

JonathanStarup:
(i still think it can be a good intermediate step to weed out all the phases that create wrong applies. The Def classes can be readded at the end if wanted)

jaschdoc:
Moved all changes to https://github.com/flix/flix/pull/8667 and trimmed this one down

main/src/ca/uwaterloo/flix/language/ast/ResolvedAst.scala
```diff
@@ -105,6 +105,8 @@ object ResolvedAst {
 
     case class Apply(exp: Expr, exps: List[Expr], loc: SourceLocation) extends Expr
 
+    case class ApplyDef(exp: Expr.Def, exps: List[Expr], loc: SourceLocation) extends Expr
```
magnus-madsen, line 108: This is not right. You should not be referring to Expr.Def. Instead, you should always refer to the relevant symbol.

magnus-madsen: We generally never have ADTs that refer to specific subcases of (other) ADTs -- because it creates a tight coupling.

The point of ApplyDef is that it targets a Def, and a Def has a DefnSym

jaschdoc: Roger that

jaschdoc: I think the important thing here was to make this compile and run with minimal changes (everything is in https://github.com/flix/flix/pull/8667 )

magnus-madsen: Agree on minimalism. But its about having the right abstraction. In this PR you have this weird ApplyDef is not its own thing but a reference to another thing, yet it still has some extra type vars etc.

main/src/ca/uwaterloo/flix/language/phase/Stratifier.scala
```diff
@@ -131,6 +131,10 @@ object Stratifier {
       val es = exps.map(visitExp)
       Expr.Apply(e, es, tpe, eff, loc)
 
+    case Expr.ApplyDef(exp, exps, ftpe, tpe, eff, loc) =>
```
magnus-madsen, line 134: exp is no longer an exp. This applies in many places.

magnus-madsen: it should be symUse

main/src/ca/uwaterloo/flix/language/phase/EffectVerifier.scala
```diff
@@ -79,6 +79,11 @@ object EffectVerifier {
       val expected = Type.mkUnion(Type.eraseTopAliases(exp.tpe).arrowEffectType :: exp.eff :: exps.map(_.eff), loc)
       val actual = eff
       expectType(expected, actual, loc)
+    case Expr.ApplyDef(_, exps, _, _, eff, loc) =>
+      exps.foreach(visitExp)
+      val expected = Type.mkUnion(exps.map(_.eff), loc)
```
magnus-madsen, line 84: This seems incorrect. Presumably the def itself can have effects?

main/src/ca/uwaterloo/flix/language/ast/TypedAst.scala
```diff
@@ -122,6 +122,8 @@ object TypedAst {
 
     case class Apply(exp: Expr, exps: List[Expr], tpe: Type, eff: Type, loc: SourceLocation) extends Expr
 
+    case class ApplyDef(symUse: Ast.DefSymUse, exps: List[Expr], ftpe: Type, tpe: Type, eff: Type, loc: SourceLocation) extends Expr
```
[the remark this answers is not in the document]
jaschdoc: Because we have lost the type of the def itself which was located in `Expr.Def`. I think @JonathanStarup can explain a little better

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -1746,7 +1746,8 @@ private def visitDef(defn: NamedAst.Declaration.Def, loc: SourceLocation)(implic
     */
 private def visitApplyDef(defn: NamedAst.Declaration.Def, exps: List[NamedAst.Expr], env: ListMap[String, Resolver.Resolution], innerLoc: SourceLocation, outerLoc: SourceLocation)(implicit scope: Scope, ns0: Name.NName, taenv: Map[Symbol.TypeAliasSym, ResolvedAst.Declaration.TypeAlias], root: NamedAst.Root, flix: Flix): Validation[ResolvedAst.Expr, ResolutionError] = {
     mapN(traverse(exps)(resolveExp(_, env))) {
-      es => visitApplyToplevelFull(ResolvedAst.Expr.Def(defn.sym, innerLoc), defn.spec.fparams.length, es, outerLoc)
+      case es if defn.spec.fparams.length == es.length => ResolvedAst.Expr.ApplyDef(Ast.DefSymUse(defn.sym, innerLoc), es, outerLoc)
```
magnus-madsen, line 1749: What happens here (I know myself), but add some comments.

main/src/ca/uwaterloo/flix/language/ast/TypedAst.scala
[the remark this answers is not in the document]
magnus-madsen: I am 95% sure this type can be reconstructed from the available information, e.g. from the types of exps, from the eff and tpe.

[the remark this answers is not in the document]
jaschdoc: I think there was an issue with this example:

```
def f(_: Int32, _: Int32): Bool \ ef = ???

f(1, g()) // g returns Int but does IO
```

The `f` has the variable `ef` and the call has `IO` effect so the call to `f` has type `ef + IO` then we cannot reconstruct the type of `f` because `ef` unifies with `IO` but it could also just be `ef`.

We can, however, easily get rid of the result type that we carry everywhere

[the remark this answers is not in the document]
magnus-madsen: We had something like this before and it was an absolute mess because the types would come out of sync.

My question is: What information is it that you need? The type scheme of f? The instantiated type of f? Something else?

Strictly speaking, in your example, the effect of `f`can be found by simply looking up its type scheme.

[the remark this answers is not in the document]
magnus-madsen: I think I understand the problem. I will give you a chance to think carefully about it, and describe it, so you understand it too :)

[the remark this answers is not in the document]
JonathanStarup: So i shouldn't explain? 

> I am 95% sure this type can be reconstructed from the available information, e.g. from the types of exps, from the eff and tpe.

But yea this is not true in monomorph. we don't know the effect of the def to specialize to without more information (ftpe or intrinsicEff or something third)

main/src/ca/uwaterloo/flix/language/ast/Ast.scala
```diff
@@ -540,6 +540,12 @@ object Ast {
     */
   case class RestrictableEnumSymUse(sym: Symbol.RestrictableEnumSym, loc: SourceLocation)
 
+  /**
+    * Represents a use of an defn sym.
```
JonathanStarup, line 544: *a* defn sym

main/src/ca/uwaterloo/flix/language/ast/LiftedAst.scala
```diff
@@ -61,7 +61,7 @@ object LiftedAst {
 
     case class ApplyClo(exp: Expr, exps: List[Expr], tpe: MonoType, purity: Purity, loc: SourceLocation) extends Expr
 
-    case class ApplyDef(sym: Symbol.DefnSym, exps: List[Expr], tpe: MonoType, purity: Purity, loc: SourceLocation) extends Expr
+    case class ApplyDef(symUse: Symbol.DefnSym, exps: List[Expr], tpe: MonoType, purity: Purity, loc: SourceLocation) extends Expr
```
JonathanStarup, line 64: should all ApplyDef use DefnSymUse or im I missing something?

main/src/ca/uwaterloo/flix/language/dbg/printer/ResolvedAstPrinter.scala
```diff
@@ -43,6 +43,7 @@ object ResolvedAstPrinter {
     case Expr.Use(_, _, _, _) => DocAst.Expr.Unknown
     case Expr.Cst(cst, _) => ConstantPrinter.print(cst)
     case Expr.Apply(exp, exps, _) => DocAst.Expr.App(print(exp), exps.map(print))
+    case Expr.ApplyDef(Ast.DefSymUse(sym, _), exps, _) => DocAst.Expr.App(DocAst.Expr.AsIs(sym.name), exps.map(print))
```
JonathanStarup, line 46: DocAst.Expr.ApplyDef?

main/src/ca/uwaterloo/flix/language/phase/EffectVerifier.scala
JonathanStarup: Oh yeah we call pull it off ftpe

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -1746,7 +1746,8 @@ private def visitDef(defn: NamedAst.Declaration.Def, loc: SourceLocation)(implic
...
+      case es => visitApplyToplevelFull(ResolvedAst.Expr.Def(defn.sym, innerLoc), defn.spec.fparams.length, es, outerLoc)
```
JonathanStarup, line 1750: This call will make an Apply(Def(..), ..) which shouldve been and ApplyDef. Essentially we should copy visitApplyToplevelFull twice, one where `base: DefSymUse` and one where `base: ResolvedAst.Expr.Sig`. these could stay as one function if we give a full application constructor function as `base: List[Exp] => Exp`.

main/src/ca/uwaterloo/flix/language/ast/LiftedAst.scala
magnus-madsen: DefnSymUse is to store the location of the symbol reference for IDE support.

So its not needed in the backend,

But this PR does not require DefnSymUse... It is just something Jakob decided to do.

magnus-madsen:
If you rename ftpe to itpe (for instantiated type) everywhere and fix the other comments then we can merge.

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
jaschdoc: I see but right now we support both (in this pr)

jaschdoc:
RFR

main/src/ca/uwaterloo/flix/language/ast/KindedAst.scala
```diff
@@ -89,6 +89,8 @@ object KindedAst {
 
     case class Apply(exp: Expr, exps: List[Expr], tvar: Type.Var, evar: Type.Var, loc: SourceLocation) extends Expr
 
+    case class ApplyDef(symUse: Ast.DefSymUse, exps: List[Expr], ftvar: Type, tvar: Type.Var, evar: Type.Var, loc: SourceLocation) extends Expr
```
magnus-madsen, line 92: search for ftvar

main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
```diff
@@ -410,6 +409,16 @@ object Kinder {
           KindedAst.Expr.Apply(exp, exps, tvar, evar, loc)
       }
 
+    case ResolvedAst.Expr.ApplyDef(Ast.DefSymUse(sym, loc1), exps0, loc2) =>
+      val expsVal = traverse(exps0)(visitExp(_, kenv0, taenv, henv0, root))
+      mapN(expsVal) {
+        case exps =>
+          val ftvar = Type.freshVar(Kind.Star, loc1.asSynthetic)
```
magnus-madsen, line 416: more ftvar

main/src/ca/uwaterloo/flix/language/phase/Monomorpher.scala
```diff
@@ -520,6 +520,12 @@ object Monomorpher {
       val es = exps.map(visitExp(_, env0, subst))
       MonoAst.Expr.Apply(e, es, subst(tpe), subst(eff), loc)
 
+    case LoweredAst.Expr.ApplyDef(sym, exps, itpe, tpe, eff, loc2) =>
+      val ft = subst(itpe)
```
magnus-madsen, line 524: ft

magnus-madsen:
@jaschdoc It may be useful to scan the diff.

main/src/ca/uwaterloo/flix/language/ast/KindedAst.scala
jaschdoc: oh thought you only wanted ftpe because it's not instantiated at this point

magnus-madsen (approved):

PR #8646, opened 2024-09-15
feat: resolved ast printer

JonathanStarup:
some parts missing, but better than nothing


mlutze (approved):
Should we open a ticket to fill in the gaps?

PR #8647, opened 2024-09-15
feat: better currying of defs

JonathanStarup:
Reduces lambdas in std lib by ~6.5 %

This improvement means that lambda creation becomes nested, which means that we need unique names to avoid shadowing (in Redundancy checking)

- have unique names with LocalContext
- smarter partial application of defs/sigs


main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -4034,4 +4027,22 @@ private def mkUnappliedAssocType(sym: Symbol.AssocTypeSym, loc: SourceLocation):
...
   }
+
+  private object LocalContext {
+    def mk(): LocalContext = new LocalContext(0)
+  }
+
+  /**
+    * Used to generated [[Symbol.VarSym]] with unique names.
+    */
+  private class LocalContext(private var id: Int) {
```
mlutze, line 4038: Why can't we just use GenSym?

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -1747,45 +1695,90 @@ private def visitApply(exp: NamedAst.Expr.Apply, env0: ListMap[String, Resolutio
...
+    val closureApplication = cloArgs.foldLeft(fullDefLambda){
+      case (acc, cloArg) => ResolvedAst.Expr.Apply(acc, List(cloArg), loc)
     }
+
+    closureApplication
   }
 
   /**
-    * Resolve the application expression, applying `sig` to `exps`.
+    * Returns a curried lambda and applicatoin of `defn`.
```
mlutze, line 1729: s/toin/tion/

JonathanStarup:
> Can you give an example of what happened before and what happens now?

The doc comments show the current behaviour, ill add the old in parens
- `   f     ===> x -> y -> f(x, y) ===> x -> y -> f(x, y)`
- `  f(a)   ===> x -> f(a, x)       ===> (x -> y -> f(x, y))(a)`
- ` f(a,b)  ===> f(a, b)           ===> f(a, b)`
- `f(a,b,c) ===> f(a, b)(c)        ===> (x -> y -> f(x, y))(a)(b)(c)`

So you can see its only optimal with zero or exactly all arguments applied

JonathanStarup:
> How does this interact with Jakob's new thing.

Well. Now it's more apparent when closure application and full def/sig application happen. It should make programs faster but I could be anywhere from insignificant to a lot better

JonathanStarup:
> Bonus question: How does this interact with a future Symbol.LocalDefSym

Doesn't matter much, expect code is easier to reuse now

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
JonathanStarup: fixed per discussion

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -1746,39 +1693,82 @@ private def visitApply(exp: NamedAst.Expr.Apply, env0: ListMap[String, Resolutio
...
+
+    val fparamsPadding = mkFreshFparams(arity - directArgs.length, loc.asSynthetic)
+    val argsPadding = fparamsPadding.map(fp => ResolvedAst.Expr.Var(fp.sym, loc.asSynthetic))
+
+    val fullArgs = directArgs ++ argsPadding
+    val fullDefApplication = ResolvedAst.Expr.Apply(base, fullArgs, loc)
+
+    // The ordering of lambdas and closure application doesn't matter,
+    // `fparamsPadding.isEmpty` iff `cloArgs.nonEmpty`.
+    val fullDefLambda = fparamsPadding.foldRight(fullDefApplication: ResolvedAst.Expr){
```
mlutze, line 1716: You can earn bonus points with inline comments showing what's built up along the way.

mlutze (approved):

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
JonathanStarup: Yeah i thought about it, but its hard to show each version at each point

Issue #8638, opened 2024-09-14
Annotation completions seems broken


magnus-madsen:
E.g. `@Test`

magnus-madsen:
@LoZander CC

JonathanStarup:
i thought it was `@test`, thats what ive been using

magnus-madsen:
Neither works for me.

LoZander:
The issue is specifically this: if you type @ nothing is suggest ed but if you type the first letter of the an notation text, the annotation is suggested.

E. g. typing `@T` gives nothing while `T` suggests (among others) `@Test`.

For some reason, if the annotation is made into an expression completion suggestion, it will actually be suggested when typing `@T` in an expression. I additionally tried to make it a trait decl completion, but here it behaved attrangely, just like for module decl completions.

I don't know why these suggestions work differently in this way when classifying them as a expr completions as supposed to decl completions. Perhaps some result of how decls vs exprs are parsed? I don't know. It might also be an issue in the completers but I don't know where if so

magnus-madsen:
Maybe @ is not a trigger char?

JonathanStarup:
i think its the parser thing where it only goes into annotations if it sees @ *and* a name. its a problem for many autocompletes

magnus-madsen:
I can type @T and not see anything.

LoZander:
yeah, thats the weird part. It's also strange that it *does* work as expected in expression context. It light be the trigger word thing, but even if @ isn't a trigger word, wouldn't er get suggestions you the type t and are standing with the unfinished line @T? Or perhaps if a word starts with something that isn't a trigger char, nothing will be suggested even after you type the next letter?

JonathanStarup:
as we discussed: add a hook in completionProvider in getContext (?) to have UndefinedAnnotation give the module context. This (along with ctrl+space) should allow nice completions (except maybe for @ by itself)

LoZander:
Is this issue done, or is there more to do?

PR #8654, opened 2024-09-16
fix: broken annotation completions

LoZander:
This fixes an issue where annotation completions aren't suggested as expect. For instance, typing `@T` would not give us the suggestion `@Test`. Now it does.

If only typing `@` we only get suggestions by invoking the suggestions using ctrl+space. However, when typing for instance `@D` we immediately get the suggestion `@Deprecated`.

For now I've given the priority 1 to the new `getSyntacticContext` case. I believe that what really matters is that it's prioritised above parse error, but I'm unsure. Currently, it seems to have the intended behaviour.

Related to #8638

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
@@ -246,6 +246,7 @@ object CompletionProvider {
       case ResolutionError.UndefinedOp(_, _) => (1, SyntacticContext.Expr.Do)
       case WeederError.MalformedIdentifier(_, _) => (2, SyntacticContext.Import)
       case WeederError.UnappliedIntrinsic(_, _) => (5, SyntacticContext.Expr.OtherExpr)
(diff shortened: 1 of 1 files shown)


mlutze:
!!!rerun

JonathanStarup:
I think the line
```scala
case err => pos.line <= err.loc.beginLine
```
is pretty weird, maybe you should keep it mind that restricting this range might lead to better contexts (it looks like all errors after your cursor can decide the context, instead of just the nearby errors)

LoZander:
Hm. I havent really looked much at this `getSyntacticContexts` function. It was from before my time. I'm not sure why the tests fail, I simply added a case, didn't touch anything else...
and I agree about that one line with that bound. That's weird. I don't know what to make of that, but I can look into it

LoZander:
yeah, I think the tests failing was something else somewhere else

JonathanStarup:
I don't think you have to change anything now, but just keep it in mind if you have multiple errors and have bad contexts

JonathanStarup (approved):

magnus-madsen:
@LoZander Would it make sense to add a specific AnnotationCompleter?

PR #8655, opened 2024-09-17
chore: add inlining benchmark

jaschdoc:


jaschdoc:
@mlutze seems like the test tool wont compile the benchmark suite when it refers to another module outside the file. Should I just put the modules into the benchmark file, i.e., `BenchmarkSuite.flix`?

JonathanStarup:
The testing framework assumes that files are selfcontained so it seems sensible to just put it into the one file

jaschdoc:
Thanks

main/src/resources/benchmark/BenchmarkSuite.flix
```diff
@@ -21,12 +21,25 @@ import java.lang.System
...
-    "List.map"          => () -> List.range(0, 10_000) |> List.map(x -> x + 1) |> blackhole,
-    "List.length"       => () -> List.range(0, 10_000) |> List.length |> blackhole,
-    "List.reverse"      => () -> List.range(0, 10_000) |> List.reverse |> blackhole
+    "List.filter"    => () -> List.range(0, 10_000) |> List.filter(x -> Int32.modulo(x, 2) == 0) |> blackhole,
+    "List.foldLeft"  => () -> List.range(0, 10_000) |> List.foldLeft(Add.add, 0) |> blackhole,
+    "List.foldRight" => () -> List.range(0, 10_000) |> List.foldRight(Add.add, 0) |> blackhole,
+    "List.map"       => () -> List.range(0, 10_000) |> List.map(x -> x + 1) |> blackhole,
+    "List.length"    => () -> List.range(0, 10_000) |> List.length |> blackhole,
+    "List.reverse"   => () -> List.range(0, 10_000) |> List.reverse |> blackhole,
+    "(inliner) Map 100_000 normal" => () -> {
```
magnus-madsen, line 30: I suggest to use a name without special chars or spaces. Otherwise it might break our scripts on arewefast.

Suggestion: InlinerMap100k or whatever.

magnus-madsen:
I think it should be possible to split BenchmarkSuite into multiple files, e.g. `Benchmarks` and `BenchmarkSuite`.

Would that work @mlutze ? We copy the whole directory right? So as long as everything is self-contained?

mlutze:
> I think it should be possible to split BenchmarkSuite into multiple files, e.g. `Benchmarks.flix` and `BenchmarkSuite.flix`.
> 
> Would that work @mlutze ? We copy the whole directory right? So as long as everything is self-contained?

Yes, it should work.

jaschdoc:
_should_ :)

jaschdoc:
I tried it but it was unable to resolve the names in BenchmarkSuite that referred to the other file (using fully qualified names)

magnus-madsen:
> I tried it but it was unable to resolve the names in BenchmarkSuite that referred to the other file (using fully qualified names)

Maybe @mlutze can help here?

magnus-madsen:
@mlutze You can merge this PR once satisfied.

mlutze:
> I tried it but it was unable to resolve the names in BenchmarkSuite that referred to the other file (using fully qualified names)

How did you test?

mlutze:
> > I tried it but it was unable to resolve the names in BenchmarkSuite that referred to the other file (using fully qualified names)
> 
> How did you test?

As Jonathan pointed out to me, the test suite tests files individually. If you look for `mkTestDir` you can see the implementation.

In a separate PR you can create a `mkTestDirTogether` (or whatever) function that will test the files in the directory together. Then we can use it here.

magnus-madsen (approved):

Issue #8639, opened 2024-09-14
Rework HoverProvider to use search


magnus-madsen:
We should refactor HoverProvider to search through the AST instead of using the index.

The core idea is to introduce a new family of functions like:

```scala
def visitExpr(expr: Expr, visit: Expr => Unit, accept: Expr => Boolean): Unit = expr match {
    case Expr.IfThenElse(exp1, exp2, exp3, tpe, eff, loc) =>
      visitExpr(expr)
      if (accept(exp1)) {
        visitExpr(exp1, visit, accept)
      }
      if (accept(exp2)) {
        visit(exp2, visit, accept)
      } if (accept(exp3)) {
        visitExpr(exp3, visit, accept)
      }
    // and so forth
}
```

We will also need visitors for almost everything in the root:

```scala
case class Root(modules: Map[Symbol.ModuleSym, List[Symbol]],
                  traits: Map[Symbol.TraitSym, Trait],
                  instances: Map[Symbol.TraitSym, List[Instance]],
                  sigs: Map[Symbol.SigSym, Sig],
                  defs: Map[Symbol.DefnSym, Def],
                  enums: Map[Symbol.EnumSym, Enum],
                  structs: Map[Symbol.StructSym, Struct],
                  restrictableEnums: Map[Symbol.RestrictableEnumSym, RestrictableEnum],
                  effects: Map[Symbol.EffectSym, Effect],
                  typeAliases: Map[Symbol.TypeAliasSym, TypeAlias],
                  uses: Map[Symbol.ModuleSym, List[Ast.UseOrImport]],
                  entryPoint: Option[Symbol.DefnSym],
                  reachable: Set[Symbol.DefnSym],
                  sources: Map[Source, SourceLocation],
                  traitEnv: Map[Symbol.TraitSym, Ast.TraitContext],
                  eqEnv: ListMap[Symbol.AssocTypeSym, Ast.AssocTypeDef],
                  names: MultiMap[List[String], String],
                  precedenceGraph: LabelledPrecedenceGraph)
```

We can start with visitDef and visitExpr.

We will also need functions like:

```scala
def inside(loc1: SourceLocation, loc2: SourceLocation): Boolean = ...
def inside(uri: String, pos: Position, loc: SourceLocation): Boolean = ...
```

With all of this, we should be able to replace:

```scala
    index.query(uri, pos) match {
      case None => mkNotFound(uri, pos)
      case Some(entity) => hoverEntity(entity, uri, pos)
    }
```

[...]

PR #8657, opened 2024-09-17
feat: hover search

LoZander:
This changes the approach to the hover provider from using the index to search based approach. 

This is the first step in process with the ultimate goal of  completely elliminating the use of index. We want to elliminate index because it's costly to maintain it, since doing so means rebuilding it upon every keystroke.

Related to #8639


LoZander:
@magnus-madsen i've made this draft PR so you can follow my progress, if you want :)

LoZander:
im still in the very early stages right now. I kinda started from above, which might have been a mistake. I think I'll focus on `visitExpr` for a bit. I also wanna see if I can test that function systematically

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitors.scala
```diff
@@ -0,0 +1,316 @@
...
+
+    // sp1 and sp2, by invariant, has the same source, so we can use either
+    (uri == loc.sp1.source.name) &&
+    (posLine >= loc.beginLine) &&
+    (posLine <= loc.endLine) &&
+    !(posLine == loc.beginLine && posCol < loc.beginCol) &&
+    !(posLine == loc.endLine && posCol > loc.endCol)
+  }
+
+  def inside(other: SourceLocation => Boolean, defn: Def): Boolean = {
```
magnus-madsen, line 58: Why is this function curried? And why does it take a Def?

magnus-madsen: Just define _one function_ on locs and use that everywhere.

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitors.scala
```diff
@@ -0,0 +1,316 @@
...
+    if (accept(defn.spec.loc)) {
+      ???
+    }
+
+    if (inside(accept, defn.exp)) {
+      visitExpr(visit, accept)(defn.exp)
+    }
+  }
+
+  def visitExpr(visit: Entity => Unit, accept: SourceLocation => Boolean)(expr: Expr): Unit = {
```
magnus-madsen, line 94: I know its tempting to use Entity, but I think that abstraction is broken too.

Just go straight for the ball and use Expr.

magnus-madsen: This means that you will need multiple visit functions at the top, e.g. both acceptExpr, acceptDefn, acceptBlah. But dont worry, later we can put them behind an interface. For now just make the signatures correct (even if long).

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitors.scala
```diff
@@ -0,0 +1,316 @@
...
+    // TODO: handle mutually recursive calls to non expressions in expressions (fx annotations)
+    val recur = visitExpr(visit, accept)
```
magnus-madsen, line 96: OK. I need to teach you something:

_first_ you write it in the obvious but long way.

_only then_ can you introduce the right abstractions and refactor.

You don't try to be smart from the beginning because you paint yourself into a corner because you do not yet fully understand the problem.

magnus-madsen: So forget about this recurIf blah and just write the code in the obvious way with minimal fuss.

Only then can we discuss a better way. 

Do not immediately reach for: higher order functions, polymorphism, traits. Just plain functions and plain recursion.

magnus-madsen: Wax on, wax off:

CleanShot 2024-09-18 at 10 19 58@2x

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
```diff
@@ -26,19 +26,31 @@ import org.json4s.JsonAST.JObject
...
-
   def processHover(uri: String, pos: Position)(implicit index: Index, root: Root, flix: Flix): JObject = {
-    index.query(uri, pos) match {
-      case None => mkNotFound(uri, pos)
+    hover(uri, pos)
+  }
+
 
-      case Some(entity) => hoverEntity(entity, uri, pos)
+  def hover(uri: String, pos: Position)(implicit root: Root, flix: Flix): JObject = {
```
magnus-madsen, line 41: This probably belongs in the visitor.

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitors.scala
```diff
@@ -0,0 +1,316 @@
+package ca.uwaterloo.flix.api.lsp.provider
```
magnus-madsen, line 1: Add copyright header.

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitors.scala
LoZander: i see what you mean. I didn't think it was an issue here, since I considered the abstraction fairly simple. but maybe it's better to not use it anyways. I'll revert to the simpler way :)

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitors.scala
LoZander: got it

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitors.scala
LoZander: ive removed all the higher order inside functions. Now it's just the uri-pos and loc ones that are left. Since we don't have loc at the top level and we kinda just let the partially applied `inside` using uri and pos propegate down, we actually never use the one that takes in two locs

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
LoZander: im not sure I understand why it would. It feels specific to the hover provider, whereas the content of visitors, at least in theory, should be more general. Although, I can see that it might be a little unnatural for the user of visitors to have to manually take the head of the list and such. But in theory, a user might want the entire back search trace

LoZander: if you think its cleaner, I can move it into visitors

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitors.scala
LoZander: your point about painting myself into a corner, I think you're absolutely right. Thats the danger of premature abstraction. I think I do fall for that sometimes

LoZander: ive actually never watched karate kid. I think I will :p

LoZander: I've watched it now. I danced too much, I see that now *bows while maintaining eyecontact

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
magnus-madsen: I think maybe my comment was misplaced. I was thinking the stack stuff might be reusable for other providers?

LoZander: true. While it should be seperate from the visit functions, it might make sense to put it might be nice to have it in the module. I think you're right, that other providers might be able to reuse it :)

LoZander:
Now, hovering on an exp works!

LoZander:
Thought there is some strange behaviour when hovering on semantic operations. Because these aren't visited by the visitor, the hover is instead on the whole expression. So hovering on `and` in `x and y` just says the type of the entire expression and nothing about `and`. This might be misleading behaviour, because it might seem like it's saying that `and` has type `Bool`, which is ofc. not the case. `and` doesn't strictly have any type (though in the abstract its `Bool -> Bool -> Bool` ofc)

LoZander:
Alright, so *now* it works. I had yet to pull the merge before :p

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
```diff
@@ -26,19 +26,58 @@ import org.json4s.JsonAST.JObject
...
 
-      case Some(entity) => hoverEntity(entity, uri, pos)
+    def seenDef(defn: Def): Unit = {
+      stack = defn :: stack
     }
+
+    Visitor.visitRoot(root,
+                      seenDef,
+                      eff => (),
+                      enum => (),
```
JonathanStarup, line 56: enum -> enm

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
```diff
@@ -26,19 +26,58 @@ import org.json4s.JsonAST.JObject
...
+  def processHover(uri: String, pos: Position)(implicit root: Root, flix: Flix): JObject = {
+    hover(uri, pos)
+  }
+
 
-  def processHover(uri: String, pos: Position)(implicit index: Index, root: Root, flix: Flix): JObject = {
-    index.query(uri, pos) match {
-      case None => mkNotFound(uri, pos)
+  def hover(uri: String, pos: Position)(implicit root: Root, flix: Flix): JObject = {
+    var stack: List[Any] = Nil
```
JonathanStarup, line 43: Any -> AnyRef

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
```diff
@@ -26,19 +26,58 @@ import org.json4s.JsonAST.JObject
...
+                      sig => (),
+                      struct => (),
+                      traitt => (),
+                      typeAlias => (),
+                      Visitor.inside(uri, pos))
+
+    hoverAny(stack, uri, pos)
+  }
+
+  private def hoverAny(trace: List[Any], uri: String, pos: Position)(implicit root: Root, flix: Flix): JObject = trace match {
```
JonathanStarup, line 70: trace -> trace.headOpt

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
```diff
@@ -0,0 +1,529 @@
...
+    }}
+
+    root.typeAliases.foreach{ case (_, alias) => {
+      if (accept(alias.loc)) { visitTypeAlias(alias, seenTypeAlias, accept) }
+    }}
+
+    // root.uses
+  }
+
+  def inside(uri: String, pos: Position)(loc: SourceLocation): Boolean = {
```
JonathanStarup, line 127: all defs private

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
```diff
@@ -0,0 +1,529 @@
...
+  /**
+    * Applies a `seenRoot` to the root AST node and recursively visits
+    * all children, applying the corresponding `seenX` functions.
+    *
+    * @param root    The AST root node.
+    * @param seenX   The `seen` function that is to be applied to X type AST nodes.
+    * @param accept  A predicate determining whether to visit a child node.
+    */
+  def visitRoot(root: Root, 
+                seenDef: Def => Unit,
```
JonathanStarup, line 58: `SeenX` trait called `Consumer` with `consumeX` defs.

```
trait Consumer {
    def consumeDef(defn: Defn): Unit = ()
    /* def consumeX(x: X): Unit = () */
}
trait Acceptor {
    def accept(loc: SourceLocation): Boolean
}
val allAcceptor = ???
val fileAcceptor = ???
```

```
def visitExpr(exp: Expr)(implicit a: Acceptor, c: Consumer): Unit =  ???
```

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
```diff
@@ -0,0 +1,529 @@
...
+  }
+
+  def visitUse(use: UseOrImport, seen: UseOrImport => Unit, accept: SourceLocation => Boolean): Unit = {
+    // TODO
+    // visit(use)
+  }
+
+  def visitExpr(expr: Expr, seen: Expr => Unit, accept: SourceLocation => Boolean): Unit = {
+    // TODO: handle mutually recursive calls to other visit functions
+
```
JonathanStarup, line 252:
```
if (!accept(expr)) return;
seen(expr)
```

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
```diff
@@ -0,0 +1,529 @@
...
+    seen(expr)
+    expr match {
+      case Expr.Cst(cst, tpe, loc) => ()
+      case Expr.Var(sym, tpe, loc) => ()
+      case Expr.Def(sym, tpe, loc) => ()
+      case Expr.Sig(sym, tpe, loc) => ()
+      case Expr.Hole(sym, tpe, eff, loc) => ()
+
+      case Expr.HoleWithExp(exp, tpe, eff, loc) => {
+        if (accept(exp.loc)) { visitExpr(exp, seen, accept) }
```
JonathanStarup, line 262: this becomes `visitExpr(exp)` 🥂

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
```diff
@@ -0,0 +1,529 @@
...
+                seenEnum: Enum => Unit,
+                seenExpr: Expr => Unit,
+                seenInstance: Instance => Unit,
+                seenResEnum: RestrictableEnum => Unit,
+                seenRoot: Root => Unit,
+                seenSig: Sig => Unit,
+                seenStruct: Struct => Unit,
+                seenTrait: Trait => Unit,
+                seenTypeAlias: TypeAlias => Unit,
+                accept: SourceLocation => Boolean): Unit = {
```
JonathanStarup, line 69: add "seenSpec" but in the trait, `consumeSpec`

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
```diff
@@ -0,0 +1,529 @@
...
+
+    root.defs.foreach{ case (_, defn) => {
+      val insideDefn = accept(defn.spec.loc) || accept(defn.exp.loc) || accept(defn.sym.loc)
```
JonathanStarup, line 72: use the "clear" location `defn.loc` or `eff.loc`.

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
LoZander: except for the `inside` functions and ofc. the `visitRoot` function

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
LoZander: There is the issue that `Def` do not have locations at all. Their `DefnSym`s have, the `Spec`s have and their `Expr`s have, but `Def`s in and of themselves, do not have `SourceLocation`s. Should I use the symbols locations or something?

magnus-madsen:
Everything related to:

- Restrictable variants (e.g. all their declarations, choose, choose star, and open_as) should be completely ignored.

LoZander:
> Everything related to:
> 
> * Restrictable variants (e.g. all their declarations, choose, choose star, and open_as) should be completely ignored.

Got it. I knew it was kind of outdated, but I wasn't sure how to handle it in the visitor since it is still in the AST. Makes my life easier to not have to deal with it, so I'm glad to hear it. I'll remove the visitor stuff related to it :)

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
JonathanStarup: Yeah good point. I also understand your point about the locations not covering the entire "thing" defs/effect/etc. Maybe we should have some extra PRs to add locations to defs fx.

LoZander: Yeah. It's an issue I've run into a couple of times. How should I handle these situations for now?

magnus-madsen (commented):
Lets break this PR into several smaller PRs:

- One that simply adds Vistor.scala,
- One that adds inside functions to `SourceLocation`, and 
- One that reworks HoverProvider

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
```diff
@@ -0,0 +1,850 @@
...
+    !(loc1.endLine == loc2.endLine && loc1.endCol > loc2.endCol)
+  }
+
+  private def visitEnum(enm: Enum)(implicit a: Acceptor, c: Consumer): Unit = {
+    if (!a.accept(enm.loc)) { return }
+
+    c.consumeEnum(enm)
+
+    visitAnnotations(enm.ann)
+    // TODO visit modifiers?
```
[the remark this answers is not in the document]
magnus-madsen: If you want them to have source locations you can add that in a separate PR.

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
```diff
@@ -0,0 +1,850 @@
...
+    enm.tparams.foreach(visitTypeParam)
+    visitDeriveList(enm.derives)
+    enm.cases.map(_._2).foreach(visitCase)
+  }
+
+  private def visitDeriveList(deriveList: Derivations)(implicit a: Acceptor, c: Consumer): Unit = {
+    if (!a.accept(deriveList.loc)) { return }
```
magnus-madsen, line 236: You had something nice going where there was always a line break after the check. Stick to it!

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
```diff
@@ -0,0 +1,850 @@
...
+  private def visitCase(cse: Case)(implicit a: Acceptor, c: Consumer): Unit = {
+    if (!a.accept(cse.loc)) { return }
+    c.consumeCase(cse)
+    // TODO visit scheme?
+  }
+
+  private def visitInstance(ins: Instance)(implicit a: Acceptor, c: Consumer): Unit = {
+    if (!a.accept(ins.loc)) { return }
+
+    visitAnnotations(ins.ann)
```
magnus-madsen, line 255: You know, if you pattern match on Instance then you can see what components are there.

This would make it much easier to immediately see if you have visited everything.

magnus-madsen: Moreover, if the definition of Instance changes then this code would break, but your code would not.

magnus-madsen: So I suggest to use pattern matching everywhere, i.e. def visitX() should start with a pattern match on `X`

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
```diff
@@ -0,0 +1,850 @@
...
+    visitType(tdefn.arg)
+  }
+
+  private def visitAssocTypeSymUse(symUse: AssocTypeSymUse)(implicit a: Acceptor, c: Consumer): Unit = {
+    if (!a.accept(symUse.loc)) { return }
+    c.consumeAssocTypeSymUse(symUse)
+  }
+
+  private def visitSig(sig: Sig)(implicit a: Acceptor, c: Consumer): Unit = {
+    // TODO fix hack. This hack is necessary because a signature currently does not have a location.
```
magnus-madsen, line 293: Refactor into an actual comment.

magnus-madsen: and consider a PR to fix the issue.

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
```diff
@@ -0,0 +1,850 @@
...
+    sig.exp.foreach(visitExpr)
+  }
+
+  private def visitStruct(struct: Struct)(implicit a: Acceptor, c: Consumer): Unit = {
+    if (!a.accept(struct.loc)) { return }
+    c.consumeStruct(struct)
+
+    visitAnnotations(struct.ann)
+    struct.tparams.foreach(visitTypeParam)
+    // TODO visit scheme?
```
magnus-madsen, line 307: Matt says not to visit schemes.

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
```diff
@@ -0,0 +1,850 @@
...
+    c.consumeAssocTypeSig(assoc)
+
+    visitTypeParam(assoc.tparam)
+    visitKind(assoc.kind)
+    assoc.tpe.foreach(visitType)
+    
+  }
+
+  private def visitKind(kind: Kind)(implicit a: Acceptor, c: Consumer): Unit = {
+    // TODO it doesn't have a location, so does this even make sense?
```
magnus-madsen, line 342: If there is no location then there is nothing to visit. Correct.

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
```diff
@@ -0,0 +1,850 @@
...
+  private def visitEffect(eff: Effect)(implicit a: Acceptor, c: Consumer): Unit = {
+    if (!(a.accept(eff.loc))) { return }
+    c.consumeEff(eff)
+
+    visitAnnotations(eff.ann)
+    eff.ops.foreach(visitOp)
+  }
+
+  private def visitOp(op: Op)(implicit a: Acceptor, c: Consumer): Unit = {
+    // TODO: hack that should eventually be fixed. Really it should be `op.loc`, but since op decls don't have locations, this hack is necessary for now
```
magnus-madsen, line 354: Same comment as above.

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
```diff
@@ -0,0 +1,850 @@
...
+    if (!a.accept(op.spec.loc)) { return }
+    c.consumeOp(op)
+    visitSpec(op.spec)
+  }
+
+  private def visitDef(defn: Def)(implicit a: Acceptor, c: Consumer): Unit = {
+    val insideDefn = a.accept(defn.spec.loc) || a.accept(defn.exp.loc) || a.accept(defn.sym.loc)
```
magnus-madsen, line 361: No comment?

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
```diff
@@ -0,0 +1,850 @@
...
+
+  private def visitSpec(spec: Spec)(implicit a: Acceptor, c: Consumer): Unit = {
+    if (!a.accept(spec.loc)) { return }
+
+    c.consumeSpec(spec)
+
+    visitAnnotations(spec.ann)
+    spec.tparams.foreach(visitTypeParam)
+    spec.fparams.foreach(visitFormalParam)
+    // TODO visit scheme
```
magnus-madsen, line 378: No need

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
```diff
@@ -0,0 +1,850 @@
...
+      case UseOrImport.Import(_, _, loc) => loc
+    }
+
+    if (!a.accept(loc)) { return }
+
+    c.consumeUseOrImport(use)
+  }
+
+  private def visitExpr(expr: Expr)(implicit a: Acceptor, c: Consumer): Unit = {
+    // TODO: handle mutually recursive calls to other visit functions
```
magnus-madsen, line 427: Not needed.

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
```diff
@@ -0,0 +1,850 @@
...
+        visitAnnotations(ann)
+        visitExpr(exp1) 
+        visitExpr(exp2) 
+
+      case Expr.Region(tpe, loc) => ()
+
+      case Expr.Scope(sym, regionVar, exp, tpe, eff, loc) =>
+        visitExpr(exp) 
+
+      case Expr.IfThenElse(exp1, exp2, exp3, tpe, eff, loc) =>
```
magnus-madsen, line 486: You should not have unused local variables. Intellij IDEA can warn about this.

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
```diff
@@ -0,0 +1,850 @@
...
+
+    root.instances.foreach{ case (_, l) => l.foreach(visitInstance) }
+
+    // root.modules.map { case (_, v) => visitModule(v, ???, ???) }
+
+    root.sigs.foreach{ case (_, sig) => visitSig(sig) }
+
+    root.structs.foreach{ case (_, struct) => visitStruct(struct) }
+
+    // root.traitEnv.map{ case (_, v) => visitTraitEnv(???, ???)(v) };
```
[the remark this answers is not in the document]
mlutze: Yes

main/src/ca/uwaterloo/flix/api/lsp/Position.scala
```diff
@@ -26,6 +26,14 @@ import org.json4s.*
   */
 object Position {
 
+  /**
+    * Returns a position from zero-indexed input
+    *
+    * Temporary fix due to positions currently being one-indexed.
+    */
+  def fromZeroIndexed(line: Int, character: Int): Position =
```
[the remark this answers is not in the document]
LoZander: So, currently the `Position`s are one-indexed, despite this disagreeing with both documentation in the code and the LSP spec. I want them to be zero-indexed, but refactoring Positions entirely feels like a PR in its own right. So right now, I just want to be able to *use* them as if they are zero-indexed, so I don't compound the issue by implementing things that depend on them being one-indexed. So I added two functions that allow me to create and use positions as if they were zero-indexed: `fromZeroIndexed` and `toZeroIndex` (I believe that's what I called the other one). That way, at least the new code I've added will be easy to refactor when I do refactor Position to be zero-indexed, as I can simply delete these functions, and errors will show where I need to adjust the code. I know it's a kind of hack solution.

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
```diff
@@ -0,0 +1,850 @@
...
+
+    // sp1 and sp2, by invariant, has the same source, so we can use either
+    (uri == loc.sp1.source.name) &&
+    (posLine >= loc.beginLine) &&
+    (posLine <= loc.endLine) &&
+    (!(posLine == loc.beginLine && posCol < loc.beginCol)) &&
+    (!(posLine == loc.endLine && posCol >= loc.endCol)) // geq because end column is exclusive
+  }
+
+  def inside(loc1: SourceLocation, loc2: SourceLocation): Boolean = {
```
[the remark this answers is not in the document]
LoZander: Yeah, you're right. I'll remove it

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
LoZander: Good point. I think I'll do that

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
```diff
@@ -0,0 +1,850 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ca.uwaterloo.flix.api.lsp.provider
+
+import ca.uwaterloo.flix.api.lsp.Entity
+import ca.uwaterloo.flix.language.ast.TypedAst.{
```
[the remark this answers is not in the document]
LoZander: I have not been using intellij, but I'm giving it a try now. I have asked intellij to optimise the imports and used more wildcards

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
```diff
@@ -0,0 +1,918 @@
...
+    def consumeConstraint(c: Constraint): Unit = ()
+    def consumeConstraintParam(cparam: ConstraintParam): Unit = ()
+    def consumeDef(defn: Def): Unit = ()
+    def consumeDefSymUse(sym: DefSymUse): Unit = ()
+    def consumeDerivation(derive: Derivation): Unit = ()
+    def consumeDerivations(derives: Derivations): Unit = ()
+    def consumeEff(eff: Effect): Unit = ()
+    def consumeEffectSymUse(effUse: EffectSymUse): Unit = ()
+    def consumeEnum(enm: Enum): Unit = ()
+    def consumeEqConstraint(ec: EqualityConstraint): Unit = ()
```
magnus-madsen, line 69: EqualityConstraint

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
```diff
@@ -0,0 +1,918 @@
...
+
+    val sameSource = uri == loc.source.name
+    val posWithinLocLines = (loc.beginLine <= posLine) && (posLine <= loc.endLine)
+
+    val beginSameLine = posLine == loc.beginLine
+    val endSameLine = posLine == loc.endLine
+    val beginSameLineImpliesPosColGeq = !beginSameLine || loc.beginCol <= posCol
+    val endSameLineImpliesPosColLess = !endSameLine || posCol < loc.endCol // pos column must be *strictly* less than since location range is exclusive with regard to columns
+
+    sameSource &&
```
magnus-madsen, line 208: Add if statements.

magnus-madsen: @JonathanStarup Will give a description of how to do this clearly.

LoZander:
> Lets break this PR into several smaller PRs:
> 
> * One that simply adds Vistor.scala,
> * One that adds inside functions to `SourceLocation`, and
> * One that reworks HoverProvider

I've split this into the PR drafts:

* #8837

I will add to this list as I create the PRs

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
LoZander: what do you by refactoring into actual comment? Isn't it an actual comment? Just that the comment should be more clear?

main/src/ca/uwaterloo/flix/api/lsp/provider/Visitor.scala
LoZander: We ended up agreeing that this should infact be visited, right? Just like sigs and defs

PR #8658, opened 2024-09-18
feat: summary includes NonDet, etc.

JonathanStarup:


magnus-madsen:
@mlutze You can discuss and merge when satisfied.

magnus-madsen:
Note that the new base effects are currently used very very sparingly. See BaseEffects.

main/src/ca/uwaterloo/flix/language/ast/Purity.scala
```diff
@@ -114,19 +114,32 @@ object Purity {
...
   def fromType(eff: Type)(implicit universe: Set[Symbol.EffectSym]): Purity = {
     evaluateFormula(eff) match {
       case set if set.isEmpty => Purity.Pure
-      case set if set.sizeIs == 1 && set.contains(Symbol.IO) => Purity.Impure
+      case set if set.forall(isJavaEffect) => Purity.Impure
       case _ => Purity.ControlImpure
     }
   }
 
+  def isJavaEffect(sym: Symbol.EffectSym): Boolean = sym match {
```
mlutze, line 130: Surely this should be somewhere else, fx right next to where these all are declared.

magnus-madsen: @sockmaster27 Is adding this function also. You should coordinate and add it in one place.

sockmaster27: I propose adding it to `Symbol`:  #8699

PR #8660, opened 2024-09-18
feat: add two struct examples

magnus-madsen:


JonathanStarup (approved):
Maybe they should be named `OrderedTree`

magnus-madsen:
> Maybe they should be named `OrderedTree`

I think search tree is standard terminology. Anyway, it is just an example.

JonathanStarup:
But they are also not called SearchTree, `Tree` just seemed a bit general

PR #8665, opened 2024-09-20
test: add `mkTestDirCollected`

jaschdoc:
Add `mkTestDirCollected` which compiles all files in a directory together and runs all tests in the program

Related to https://github.com/flix/flix/pull/8655

### DIFF
--- a/main/test/ca/uwaterloo/flix/util/FlixSuite.scala
+++ b/main/test/ca/uwaterloo/flix/util/FlixSuite.scala
@@ -45,10 +45,10 @@ class FlixSuite(incremental: Boolean) extends AnyFunSuite {
   def mkTest(path: String)(implicit options: Options): Unit = {
     val p = Paths.get(path)
     val n = p.getFileName.toString
-    test(n)(compileAndRun(p))
+    test(n)(compileAndRun(List(p)))
   }
 
-  private def compileAndRun(path: Path)(implicit options: Options): Unit = {
+  private def compileAndRun(paths: List[Path])(implicit options: Options): Unit = {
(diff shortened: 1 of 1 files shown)


main/test/ca/uwaterloo/flix/util/FlixSuite.scala
```diff
@@ -31,6 +31,15 @@ class FlixSuite(incremental: Boolean) extends AnyFunSuite {
     */
   var flix = new Flix()
 
+  def mkTestDirCollected(path: String)(implicit options: Options): Unit = {
+    val iter = Files.walk(Paths.get(path), 1)
```
magnus-madsen, line 35: We have a class called something like FileOps or FileUtils. Could you move the Files.walk thingy to there?

I suppose it takes a path and returns a Scala list of paths.

main/test/ca/uwaterloo/flix/util/FlixSuite.scala
```diff
@@ -31,6 +31,15 @@ class FlixSuite(incremental: Boolean) extends AnyFunSuite {
     */
   var flix = new Flix()
 
+  def mkTestDirCollected(path: String)(implicit options: Options): Unit = {
```
magnus-madsen, line 34: I think a doc comment would be good too.

main/test/ca/uwaterloo/flix/util/FlixSuite.scala
jaschdoc: Sure thing :)

magnus-madsen (approved):

Issue #8606, opened 2024-09-10
Improve letrec/local def representation


JonathanStarup:
local defs are slow in many aspects (typing / runtime) and right now we treat them very "by the book"

- `def inner(a, b) = inner(a, b + c)` becomes `inner = a -> b -> loop -> c -> loop(a)(b + c)` where the backend wires loop to be the function itself.

We need to treat local defs more like defs and less like let binders.

- [ ] Modify LetRec in weeded AST and onwards to have fparams (`def f(x, y) = f(y)(x)` instead of `letrec f = x -> y -> f(y)(x)`)
- [ ] Rename LetRec to LocalDef
- [ ] Add LocalDefSym for these symbols
- [ ] handle applications of LocalDef in resolver by either fully applying or currying, like for defs with ApplyLocalDef or applies of the curried thing
- [ ] convert other phases to do as they did before but on the new structure
- [ ] in ClosureConv (i think) handle the captured variables and hoist into a top-level def

hopefully this means that the backend then generates tail calls for local defs

JonathanStarup:
hmm its not too easy

magnus-madsen:
Sounds good. I agree with the approach, just worried about the details.

We should discuss @jaschdoc, you, and me.

jaschdoc:
Can you assign this to me and maybe add a `project` label? Or as long as it does not have the `student-programmer` label then I can distinguish between the two hats :)

jaschdoc:
Done? :)

PR #8669, opened 2024-09-20
feat: add `LocalDef`

jaschdoc:
`LocalDef` is meant to supersede `LetRec` and treats local defs more like functions. This is the first step in the inliner project.

This PR has the following changes:
- Adds`LocalDef` which is meant to supersede `LetRec`. It moves the representation of local defs closer to same representation as top level defs
- `LocalDef`s are allowed to capture any number of variables from their enclosing scope. This is handled in `ClosureConv` where they are rewritten to have their captured variables introduced as extra formal parameters before their original formal params, i.e., the new list of formal parameters are `captured variables + original formal parameters`. They are lifted to top level defs in `LambdaLifting`
- Adds `ApplyLocalDef` is similar to `ApplyDef` in that it is a fully applied local function. During `ClosureConv` any occurrence of an `ApplyLocalDef` is rewritten to also apply with the captured parameters. More precisely, when visiting a `LocalDef` in `ClosureConv`, the body of the `LocalDef` that defines a function with symbol `f` is rewritten such that any occurrence of an `ApplyLocalDef` with symbol `f` is rewritten to also include the new parameter list that represent the captured variables. In the continuation expression of the `LocalDef`, i.e., the expression that comes after the definition of a local function (as in `LetRec`), any occurrence of an `ApplyLocalDef` node is rewritten such that it also contains the captured variables as arguments. They are rewritten to `ApplyDef` in `LambdaLift`.


JonathanStarup:
I wouldn't mind it being called `LocalDef` to match surface flix better

jaschdoc:
@JonathanStarup Can I ask you to take a look through things so far? Just a check to see if I have missed something critical

JonathanStarup:
maybe we can take it irl Monday? or do you want comments before then?

jaschdoc:
> maybe we can take it irl Monday? or do you want comments before then?

Monday is fine (without comments before then)

magnus-madsen:
@jaschdoc It might be useful if you outline the most significant changes / files where we should look.

magnus-madsen:
Also, what's the status on the test?

magnus-madsen:
The compiler tests

jaschdoc:
`Verifier` blocks codegen for some tests because there is a bug currently

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -91,6 +91,13 @@ object Simplifier {
       val t = visitType(tpe)
       SimplifiedAst.Expr.Apply(e, es, t, simplifyEffect(eff), loc)
 
+    case MonoAst.Expr.ApplyLocalDef(sym, exps, itpe, tpe, eff, loc) =>
+      val ft = visitType(itpe)
+      val e = SimplifiedAst.Expr.Var(sym, ft, loc)
+      val es = exps.map(visitExp)
+      val t = visitType(tpe)
```
JonathanStarup, line 98: Just to repeat what we discussed - this `t` should be curried if we go back into a general apply

main/src/ca/uwaterloo/flix/language/ast/WeededAst.scala
```diff
@@ -119,6 +119,8 @@ object WeededAst {
 
     case class LetRec(ident: Name.Ident, ann: Annotations, mod: Ast.Modifiers, exp1: Expr, exp2: Expr, loc: SourceLocation) extends Expr
 
+    case class LocalDef(ann: Annotations, ident: Name.Ident, fparams: List[FormalParam], declaredTpe: Option[Type], declaredEff: Option[Type], exp1: Expr, exp2: Expr, loc: SourceLocation) extends Expr
```
jaschdoc, line 122: maybe rename to `declTpe` and `declEff` instead of `declaredTpe` and `declaredEff`

jaschdoc:
@JonathanStarup can I ask you to take a look at `ClosureConv` and `LambdaLift`? I think I'm on the right track, but I'm not quite sure what I'm missing

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -85,6 +85,12 @@ object Simplifier {
       val t = visitType(tpe)
       SimplifiedAst.Expr.ApplyDef(sym, es, t, simplifyEffect(eff), loc)
 
+    case MonoAst.Expr.ApplyLocalDef(sym, exps, itpe, tpe, eff, loc) =>
+      val ft = visitType(itpe) // consider dropping this
+      val es = exps.map(visitExp)
+      val t = visitType(tpe)
+      SimplifiedAst.Expr.ApplyLocalDef(sym, es, ft, t, simplifyEffect(eff), loc)
```
jaschdoc, line 92: drop `ft` here

main/src/ca/uwaterloo/flix/language/dbg/printer/LoweredAstPrinter.scala
```diff
@@ -58,9 +58,11 @@ object LoweredAstPrinter {
     case Expr.Lambda(fparam, exp, tpe, loc) => DocAst.Expr.Lambda(List(printFormalParam(fparam)), print(exp))
     case Expr.Apply(exp, exps, tpe, eff, loc) => DocAst.Expr.ApplyClo(print(exp), exps.map(print), None)
     case Expr.ApplyDef(sym, exps, _, _, _, _) => DocAst.Expr.ApplyDef(sym, exps.map(print), None)
+    case Expr.ApplyLocalDef(sym, exps, _, _, _, _) => DocAst.Expr.ApplyClo(DocAst.Expr.Var(sym), exps.map(print), None)
     case Expr.ApplyAtomic(op, exps, tpe, _, loc) => OpPrinter.print(op, exps.map(print), TypePrinter.print(tpe))
     case Expr.Let(sym, mod, exp1, exp2, tpe, eff, loc) => DocAst.Expr.Let(DocAst.Expr.Var(sym), None, print(exp1), print(exp2))
     case Expr.LetRec(sym, mod, exp1, exp2, tpe, eff, loc) => DocAst.Expr.LetRec(DocAst.Expr.Var(sym), None, print(exp1), print(exp2))
+    case Expr.LocalDef(sym, fparams, exp1, exp2, tpe, eff, loc) => DocAst.Expr.LetRec(DocAst.Expr.Var(sym), None, print(exp1), print(exp2))
```
jaschdoc, line 65: update `DocAst` in subsequent PR?

jaschdoc:
It works now :)

magnus-madsen:
@jaschdoc Can you write a summery of the changes, and where the most important parts are that should be reviewed?

magnus-madsen:
Also; is this code live yet, or not?

jaschdoc:
> Also; is this code live yet, or not?

Yes, it is all the way from the `Parser2` to `LambdaLift`

magnus-madsen:
> > Also; is this code live yet, or not?
> 
> Yes, it is all the way from the `Parser2` to `LambdaLift`, i.e., all the way to codegen

So it replaces the current LetRec?

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1307,12 +1307,8 @@ object Weeder2 {
...
         Types.tryPickEffect(tree),
       ) {
-        case (ann, (exp1, exp2), fparams, ident, tpe, eff) =>
-          val e = if (tpe.isDefined || eff.isDefined) Expr.Ascribe(exp1, tpe, eff, exp1.loc) else exp1
-          val lambda = fparams.foldRight(e) {
-            case (fparam, acc) => WeededAst.Expr.Lambda(fparam, acc, exp1.loc.asSynthetic)
-          }
-          Expr.LetRec(ident, ann, Ast.Modifiers.Empty, lambda, exp2, tree.loc)
+        case (ann, (exp1, exp2), fparams, ident, declaredTpe, declaredEff) =>
+          Expr.LocalDef(ann, ident, fparams, declaredTpe, declaredEff, exp1, exp2, tree.loc)
```
magnus-madsen, line 1311: Right -- this is where it goes live

magnus-madsen:
Does it mean that LetRec is now deprecated or?

magnus-madsen:
@JonathanStarup Can do a rigorous review and then I can do one after.

main/src/ca/uwaterloo/flix/api/lsp/Indexer.scala
```diff
@@ -252,6 +256,17 @@ object Indexer {
     case Expr.LetRec(sym, _, _, exp1, exp2, _, _, _) =>
       Index.occurrenceOf(sym, exp1.tpe) ++ visitExp(exp1) ++ visitExp(exp2) ++ Index.occurrenceOf(exp0)
 
+    case Expr.LocalDef(_, sym, fparams, declaredType, declaredEff, exp1, exp2, _, _, _) =>
+      Index.all(
+        traverse(fparams)(visitFormalParam),
+        visitType(declaredType),
+        visitType(declaredEff),
+        Index.occurrenceOf(sym, exp1.tpe),
```
JonathanStarup, line 264: Doesn't this imply that for `def example(x: Int32): Int32 = ???`, `example` would be given type `Int32`? I think you should use the `itpe` or whatever its called

main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
```diff
@@ -357,6 +357,13 @@ object SemanticTokensProvider {
       exps.foldLeft(Iterator(t)) {
         case (acc, exp) => acc ++ visitExp(exp)
       }
+    case Expr.ApplyLocalDef(sym, exps, _, _, _, loc) =>
+      val o = SemanticTokenType.Function
+      val t = SemanticToken(o, Nil, loc)
+      Iterator(t)
```
JonathanStarup, line 363: Delete line 363 and inline `o`

main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
```diff
@@ -374,6 +381,18 @@ object SemanticTokensProvider {
       val t = SemanticToken(o, Nil, sym.loc)
       Iterator(t) ++ visitExp(exp1) ++ visitExp(exp2)
 
+    case Expr.LocalDef(_, sym, fparams, declaredType, declaredEff, exp1, exp2, _, _, loc) =>
+      val o = SemanticTokenType.Function
+      val t = SemanticToken(o, Nil, sym.loc)
```
JonathanStarup, line 386: inline `o`

main/src/ca/uwaterloo/flix/language/ast/DesugaredAst.scala
```diff
@@ -111,6 +111,8 @@ object DesugaredAst {
 
     case class LetRec(ident: Name.Ident, ann: Annotations, mod: Ast.Modifiers, exp1: Expr, exp2: Expr, loc: SourceLocation) extends Expr
 
+    case class LocalDef(ann: Annotations, ident: Name.Ident, fparams: List[FormalParam], declaredTpe: Option[Type], declaredEff: Option[Type], exp1: Expr, exp2: Expr, loc: SourceLocation) extends Expr
```
JonathanStarup, line 114: I think, like with `Let`, that in these phases it should just be called `tpe` and `eff`

main/src/ca/uwaterloo/flix/language/ast/KindedAst.scala
```diff
@@ -105,6 +107,8 @@ object KindedAst {
 
     case class LetRec(sym: Symbol.VarSym, ann: Annotations, mod: Ast.Modifiers, exp1: Expr, exp2: Expr, loc: SourceLocation) extends Expr
 
+    case class LocalDef(ann: Annotations, sym: Symbol.VarSym, fparams: List[FormalParam], declaredTvar: Type, declaredEvar: Type, exp1: Expr, exp2: Expr, loc: SourceLocation) extends Expr
```
JonathanStarup, line 110: Ascribe calls them `ExpectedType` and `ExpectedPur`. Not great names but I think the two should be consistent

main/src/ca/uwaterloo/flix/language/ast/KindedAst.scala
```diff
@@ -89,6 +89,8 @@ object KindedAst {
 
     case class ApplyDef(symUse: Ast.DefSymUse, exps: List[Expr], itvar: Type, tvar: Type.Var, evar: Type.Var, loc: SourceLocation) extends Expr
 
+    case class ApplyLocalDef(sym: Symbol.VarSym, exps: List[Expr], itvar: Type.Var, tvar: Type.Var, evar: Type.Var, loc: SourceLocation) extends Expr
```
JonathanStarup, line 92: I think magnus wanted `itvar` but every `tvar` and `evar` are also instantiated so it feels very arbitrary. can we do `arrowTvar`

main/src/ca/uwaterloo/flix/language/ast/LoweredAst.scala
```diff
@@ -96,12 +96,16 @@ object LoweredAst {
 
     case class ApplyDef(sym: Symbol.DefnSym, exps: List[Expr], itpe: Type, tpe: Type, eff: Type, loc: SourceLocation) extends Expr
 
+    case class ApplyLocalDef(sym: Symbol.VarSym, exps: List[Expr], ftpe: Type, tpe: Type, eff: Type, loc: SourceLocation) extends Expr
```
JonathanStarup, line 99: here its called ftpe

main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
```diff
@@ -555,6 +555,14 @@ object Desugar {
       val e2 = visitExp(exp2)
       Expr.LetRec(ident, ann, mod, e1, e2, loc)
 
+    case WeededAst.Expr.LocalDef(ann, ident, fparams, dtpe, deff, exp1, exp2, loc) =>
```
JonathanStarup, line 558: I don't think we should have dtpe and deff after this actually. shouldn't it just become `def (...) = (body: dtpe \ deff)` like lets do in this phase?

jaschdoc:
Added synposis at the top

jaschdoc:
I could also use a double check on `ConstraintGen`. I think it's ok, but I can't shake the feeling that I overlooked something.

main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
jaschdoc: I'm not quite sure I understand what you mean :)

main/src/ca/uwaterloo/flix/api/lsp/Indexer.scala
magnus-madsen: By given he means shown on hover in VSCode

magnus-madsen: I think it might be good to build a JAR and do a bit of hover + renaming to see that everything works as expected.

main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
magnus-madsen: Agree

main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
magnus-madsen: Agree

main/src/ca/uwaterloo/flix/language/ast/TypedAst.scala
```diff
@@ -126,6 +128,8 @@ object TypedAst {
 
     case class LetRec(sym: Symbol.VarSym, ann: Annotations, mod: Ast.Modifiers, exp1: Expr, exp2: Expr, tpe: Type, eff: Type, loc: SourceLocation) extends Expr
 
+    case class LocalDef(ann: Annotations, sym: Symbol.VarSym, fparams: List[FormalParam], declaredType: Type, declaredEff: Type, exp1: Expr, exp2: Expr, tpe: Type, eff: Type, loc: SourceLocation) extends Expr
```
magnus-madsen, line 131: Do we need to keep the declared type/eff around? We have the real type and effect now.

main/src/ca/uwaterloo/flix/language/ast/WeededAst.scala
magnus-madsen: I am ok with declaredTpe and declaredEff. What is used elsewhere in this file?

main/src/ca/uwaterloo/flix/language/ast/ops/TypedAstOps.scala
```diff
@@ -169,6 +176,9 @@ object TypedAstOps {
     case Expr.LetRec(sym, _, _, exp1, exp2, _, _, _) =>
       (freeVars(exp1) ++ freeVars(exp2)) - sym
 
+    case Expr.LocalDef(_, sym, fparams, _, _, exp1, exp2, _, _, _) =>
+      (freeVars(exp1) -- fparams.map(_.sym) - sym) ++ freeVars(exp2)
```
magnus-madsen, line 180: The minuses get me. Add some extra parens.

main/src/ca/uwaterloo/flix/language/dbg/printer/LoweredAstPrinter.scala
magnus-madsen: Yes

main/src/ca/uwaterloo/flix/language/phase/ClosureConv.scala
```diff
@@ -146,6 +170,41 @@ object ClosureConv {
 
   }
 
+  private def rewriteApplyLocalDef(expr00: Expr, sym0: Symbol.VarSym, freeVars: List[FreeVar]): Expr = {
```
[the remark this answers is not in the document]
magnus-madsen: Yes, and maybe a small example in the doc.

main/src/ca/uwaterloo/flix/language/phase/ClosureConv.scala
```diff
@@ -146,6 +170,41 @@ object ClosureConv {
...
+      case Expr.ApplyDef(sym, exps, tpe, purity, loc) => Expr.ApplyDef(sym, exps.map(visit), tpe, purity, loc)
+      case Expr.ApplyLocalDef(sym, exps, itpe, tpe, purity, loc) =>
+        val es = exps.map(visit)
+        if (sym == sym0) {
+          val all = freeVars.map(fv => Expr.Var(fv.sym, fv.tpe, loc.asSynthetic)) ++ es
+          Expr.ApplyLocalDef(sym, all, itpe, tpe, purity, loc)
+        } else {
+          Expr.ApplyLocalDef(sym, es, itpe, tpe, purity, loc)
+        }
+      case Expr.IfThenElse(exp1, exp2, exp3, tpe, purity, loc) => Expr.IfThenElse(visit(exp1), visit(exp2), visit(exp3), tpe, purity, loc)
```
magnus-madsen, line 191: I think we should add "some space" so its easier to read. Maybe also introduce local variables. 

Otherwise this function is just very crammed and it would be easy to miss a small mistake.

main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
magnus-madsen: I think he is questioning whether we should allow/support return type/return effect declarations.

magnus-madsen: But I don't see why we should not support them.

main/test/flix/Test.Exp.Let.Rec.flix
```diff
@@ -58,6 +58,45 @@ mod Test.Exp.LetRec {
...
+    def testLetRecCapture09(): Bool =
+        let a = 1;
+        let b = 2;
+        let g = {
+            let d = 3;
+            def f(l) = if (l == 4) () -> d + l else f(a + b + 1);
+            () -> f
+        };
+        g()(5)() == 7
+
```
magnus-madsen, line 99: What about a def capturing another def?

What about a def with a nested def?

main/test/ca/uwaterloo/flix/language/phase/TestRedundancy.scala
```diff
@@ -2202,4 +2202,14 @@ class TestRedundancy extends AnyFunSuite with TestUtils {
...
 
+  test("ShadowedVariable.LocalDef.01") {
+    val input =
+      """
+        |def f(): Int32 =
+        |   def g() = { def g() = 1; g() };
+        |   g()
+        |""".stripMargin
+    val result = compile(input, Options.TestWithLibNix)
+    expectError[RedundancyError.ShadowingName](result)
```
magnus-madsen, line 2213: Great- thanks!

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
magnus-madsen: So Expr.LetRec can be removed in a follow-up PR?

magnus-madsen (commented):
I am not done, but here are some preliminary comments. I will add more later.

main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
JonathanStarup: no i mean that `let x: t = e` becomes `let x = (e: t)` in this phase

I don't see why we shouldn't do the same translation here. I was looking at why `let` didn't have `dtpe` and `deff` and that's the reason. So to be consistent we should do the same or change let

main/src/ca/uwaterloo/flix/language/ast/TypedAst.scala
JonathanStarup: thats also related to my comment about the desugaring phase

main/src/ca/uwaterloo/flix/language/ast/ops/TypedAstOps.scala
JonathanStarup: maybe `let bound = fparams.map(_.sym) + sym`

main/src/ca/uwaterloo/flix/language/phase/ClosureConv.scala
```diff
@@ -101,6 +105,26 @@ object ClosureConv {
...
+      //         (Remove the variables bound by the function itself).
+      val bound = sym :: fparams.map(_.sym)
+      val fvs = filterBoundVars(freeVars(exp1), bound).toList
+
+      // Step 2: Convert the free variables into a new parameter list and substitution.
+      val (cloParams, subst) = getFormalParamsAndSubst(fvs, loc)
+
+      // Step 3: Rewrite every recursive call to include free vars and then replace
+      //         every occurrence of the free vars with their substitution.
+      val e1 = visitExp(applySubst(rewriteApplyLocalDef(exp1, sym, fvs), subst))
```
JonathanStarup, line 119: Does it matter if you `rewriteApplyLocalDef` first or `applySubst` first? it feels like both would work. Was just curious

jaschdoc: The order matters since we rewrite with `fvs`, i.e., introducing them as args, and then renaming them via the substitution afterwards. My idea was that I wanted to first do one rewriting under the same variable naming scheme and then afterwards apply the substitution. I guess you could rewrite with the variables in `cloParams` directly but I like the consistency with this and we have to apply the substitution anyway so we do not really gain anything by being "smart" that way.

jaschdoc:
Btw, I just ran our inliner benchmark and on my machine and compared it to main. With this PR the running time is halved, i.e., it 2x faster :)))

main/src/ca/uwaterloo/flix/language/phase/ClosureConv.scala
JonathanStarup: `def f(x) = f(x +  y)` becomes `def f(q, x) = f(q, x + q)` right? 

So subst first would be `f(x + y) --> f(x + q) --> f(q, x + q)`

and rewrite first would be `f(x + y) --> f(y, x + y) --> f(q, x + q)`

I think subst first feels better to me, but I don't have any concrete reason

JonathanStarup (approved):

main/src/ca/uwaterloo/flix/language/phase/ClosureConv.scala
JonathanStarup: I see. Yeah and getFormalParamsAndSubst doesnt return the new names. good point

JonathanStarup: but yeah an example in docs somewhere in this file feels required

main/src/ca/uwaterloo/flix/language/phase/ClosureConv.scala
```diff
@@ -308,6 +381,14 @@ object ClosureConv {
         val es = exps.map(visitExp)
         Expr.ApplyDef(sym, es, tpe, purity, loc)
 
+      case Expr.ApplyLocalDef(sym, exps, itpe, tpe, purity, loc) =>
+        val es = exps.map(visitExp)
+        val sym1 = subst.get(sym) match {
```
JonathanStarup, line 386: I think this implies that 
```
def f(x) = x;
def g(x) = f(x);
g(12)
```
will become
```
def f(x) = x;
def g(q, x) = q(x);
g(f, x)
```
Or something like that? Wouldn't that mean that the `ApplyLocalDef` now contains a normal varsym and that (less importantly) we don't have a direct call?

JonathanStarup: Why are binders are replaced by subst at all? seems weird

main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
```diff
@@ -417,6 +424,19 @@ object Kinder {
       val exp2 = visitExp(exp20, kenv0, taenv, henv0, root)
       KindedAst.Expr.LetRec(sym, ann, mod, exp1, exp2, loc)
 
+    case ResolvedAst.Expr.LocalDef(ann, sym, fparams0, dtpe0, deff0, exp10, exp20, loc) =>
```
JonathanStarup, line 427: I've never looked at the kinder before, maybe @mlutze can have a guest appearance 😄

main/src/ca/uwaterloo/flix/language/phase/LambdaLift.scala
```diff
@@ -222,6 +246,10 @@ object LambdaLift {
     *
     * We use a concurrent (non-blocking) linked queue to ensure thread-safety.
     */
-  private case class SharedContext(liftedDefs: ConcurrentLinkedQueue[(Symbol.DefnSym, LiftedAst.Def)])
+  private case class SharedContext(liftedDefs: ConcurrentLinkedQueue[(Symbol.DefnSym, LiftedAst.Def)], liftedLocalDefs: ConcurrentLinkedQueue[(Symbol.VarSym, Symbol.DefnSym)])
+
+  private object SharedContext {
+    def mk(): SharedContext = SharedContext(new ConcurrentLinkedQueue(), new ConcurrentLinkedQueue())
```
JonathanStarup, line 252: All the other ones have 
```
/**
  * Returns a fresh shared context.
  */
```

main/src/ca/uwaterloo/flix/language/phase/Monomorpher.scala
```diff
@@ -471,6 +471,11 @@ object Monomorpher {
       val es = exps.map(visitExp(_, env0, subst))
       MonoAst.Expr.ApplyDef(newSym, es, it, subst(tpe), subst(eff), loc2)
 
+    case LoweredAst.Expr.ApplyLocalDef(sym, exps, itpe, tpe, eff, loc) =>
+      val it = subst(itpe)
+      val es = exps.map(visitExp(_, env0, subst))
+      MonoAst.Expr.ApplyLocalDef(env0(sym), es, it, subst(tpe), subst(eff), loc)
```
JonathanStarup, line 477: you should inline the one or un-inline the other two

magnus-madsen:
Is this PR "heavy" enough that we should look at it together on Monday before merge?

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -979,6 +982,24 @@ object Resolver {
         case (e1, e2) => ResolvedAst.Expr.LetRec(sym, ann, mod, e1, e2, loc)
       }
 
+    case NamedAst.Expr.LocalDef(ann, sym, fparams0, dtpe0, deff0, exp1, exp2, loc) =>
+      val fparamsVal = traverse(fparams0)(resolveFormalParam(_, env0, taenv, ns0, root))
+      flatMapN(fparamsVal) {
+        fparams =>
+          val env1 = env0 ++ mkLocalDefEnv(sym, fparams) ++ mkFormalParamEnv(fparams)
+          val exp1Val = resolveExp(exp1, env1)
+          flatMapN(exp1Val) {
```
JonathanStarup, line 991: cant this mapping be moved to the later mapN?

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -1777,7 +1798,34 @@ object Resolver {
...
+    * Example with `def f(a: Char, b: Char): Char -> Char`
+    *   - `   f     ===> x -> y -> f(x, y)`
+    *   - `  f(a)   ===> x -> f(a, x)`
+    *   - ` f(a,b)  ===> f(a, b)`
+    *   - `f(a,b,c) ===> f(a, b)(c)`
+    */
+  private def visitApplyLocalDef(sym: Symbol.VarSym, arity: Int, exps: List[NamedAst.Expr], env: ListMap[String, Resolution], innerLoc: SourceLocation, outerLoc: SourceLocation)(implicit scope: Scope, ns0: Name.NName, taenv: Map[Symbol.TypeAliasSym, ResolvedAst.Declaration.TypeAlias], root: NamedAst.Root, flix: Flix): Validation[ResolvedAst.Expr, ResolutionError] = {
+    mapN(traverse(exps)(resolveExp(_, env))) {
+      es =>
+        val base = args => ResolvedAst.Expr.ApplyLocalDef(sym, args, outerLoc) // TODO: Add Ast.LocalDefSymUse?
```
JonathanStarup, line 1827: old todo? or still relevant?

main/src/ca/uwaterloo/flix/language/phase/Stratifier.scala
```diff
@@ -152,6 +156,11 @@ object Stratifier {
       val e2 = visitExp(exp2)
       Expr.LetRec(sym, ann, mod, e1, e2, tpe, eff, loc)
 
+    case Expr.LocalDef(ann, sym, fparams, declaredType, declaredEff, exp1, exp2, tpe, eff, loc) =>
```
JonathanStarup, line 159: yeah I think the double types and effects should go away via Desugar

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -158,6 +158,19 @@ object ConstraintGen {
         val resEff = evar
         (resTpe, resEff)
 
+      case Expr.ApplyLocalDef(sym, exps, itvar, tvar, evar, loc) =>
+        val (tpes, effs) = exps.map(visitExp).unzip
+        val defResTpe = Type.freshVar(Kind.Star, loc.asSynthetic)
```
JonathanStarup, line 163: since defResTpe is exactly the same as tvar, you don't have to make a new typevariable

delete `defResTpe` and use `tvar` instead where it was used

JonathanStarup: but it looks correct

JonathanStarup:
> Is this PR "heavy" enough that we should look at it together on Monday before merge?

Could be - but I think jakob should first do the overall change with Desugaring ascribes and removing the duplicate types and effects for the AST afterwards

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
jaschdoc: old :)

main/src/ca/uwaterloo/flix/language/phase/Stratifier.scala
jaschdoc: Yeah true this is mistake. Just forgot to delete the old `tpe` and `eff` from the AST node

main/src/ca/uwaterloo/flix/language/ast/TypedAst.scala
jaschdoc: Yes, because `tpe` and `eff` is related to `exp2`, i.e., the type and effect of the whole expression, whereas the declared type is the result type of the local def (same with declared effect). Maybe they should be renamed to `arrowResTpe` and `arrowResEff`?

main/src/ca/uwaterloo/flix/language/ast/WeededAst.scala
jaschdoc: What about Jonathan's comment that it should be `expectedTpe` and `expectedEff` like `Ascribe`?

main/src/ca/uwaterloo/flix/language/phase/ClosureConv.scala
jaschdoc: Good point

jaschdoc: But then local def symbols can never be free, i.e., in `freevars` we should always exclude the symbol of a local def... but then a lambda may capture it? I guess that's okay since it's replaced by a `DefnSym` in `LambdaLift`

main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
jaschdoc: Ah I see, that way we also avoid generating the type constraint `t = e` manually

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
jaschdoc: Yes :)

main/test/flix/Test.Exp.Let.Rec.flix
jaschdoc: They are both tested above and below these tests, respectively :)

main/src/ca/uwaterloo/flix/api/lsp/Indexer.scala
```diff
@@ -252,6 +256,17 @@ object Indexer {
     case Expr.LetRec(sym, _, _, exp1, exp2, _, _, _) =>
       Index.occurrenceOf(sym, exp1.tpe) ++ visitExp(exp1) ++ visitExp(exp2) ++ Index.occurrenceOf(exp0)
 
+    case Expr.LocalDef(_, sym, fparams, declaredType, declaredEff, exp1, exp2, _, _, _) =>
+      Index.all(
+        traverse(fparams)(visitFormalParam),
+        visitType(declaredType),
+        visitType(declaredEff),
+        Index.occurrenceOf(sym, declaredType),
```
jaschdoc, line 264: oops, should be an arrow type :)))

main/src/ca/uwaterloo/flix/language/ast/TypedAst.scala
jaschdoc: Ah okay my bad, we have the in `exp1` of course

main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
mlutze: I think this looks reasonable (I haven't looked at Kinder in a long time)

But we should add a test or two since this is a new special case.

main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
JonathanStarup: yeah 👍

main/src/ca/uwaterloo/flix/language/ast/WeededAst.scala
jaschdoc, line 122: The `ann` field is completely unused through all phases, so maybe we just drop it after `Weeder`? If we ever decide to get a termination check or tail recursion check, it should be somewhat easy to re-add it, since it's handled / checked in the Weeder, just not carried any further.

magnus-madsen: OK. Dump it.

jaschdoc:
RFR again

jaschdoc:
Ok RFR now :)

jaschdoc:
`f` gets an effect as seen below but this is also what happens on `main`.

Screenshot from 2024-10-04 11-53-35

jaschdoc:
Manual tests in vscode of hover, find usages, find definition, and rename all work fine

jaschdoc:
Ok it still needs a `Ast.LocalDefSymUse` from `Resolver` until `Lowering` for LSP to work correctly

jaschdoc:
Ok manal vscode tests new work: hover, rename, go to definition, go to references (also with top level functions that have the same (surface) name, i.e., top level functions are not touched since they have different underlying symbols)

main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
```diff
@@ -555,6 +555,15 @@ object Desugar {
       val e2 = visitExp(exp2)
       Expr.LetRec(ident, ann, mod, e1, e2, loc)
 
+    case WeededAst.Expr.LocalDef(ident, fparams, dtpe, deff, exp1, exp2, loc) =>
+      val fps = visitFormalParams(fparams)
+      val t = dtpe.map(visitType)
+      val ef = deff.map(visitType)
+      val e1 = visitExp(exp1)
+      val e11 = Expr.Ascribe(e1, t, ef, e1.loc)
```
JonathanStarup, line 563: I think there's an invariant that t and ef cannot both be None (at least Let checks so i think you should too)

magnus-madsen: Yes 👍

main/src/ca/uwaterloo/flix/language/ast/TypedAst.scala
```diff
@@ -118,6 +118,8 @@ object TypedAst {
 
     case class ApplyDef(symUse: Ast.DefSymUse, exps: List[Expr], itpe: Type, tpe: Type, eff: Type, loc: SourceLocation) extends Expr
 
+    case class ApplyLocalDef(symUse: Ast.LocalDefSymUse, exps: List[Expr], arrowTpe: Type, tpe: Type, eff: Type, loc: SourceLocation) extends Expr
```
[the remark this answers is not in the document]
JonathanStarup:
```
def f(x: Int32): Int32 \ IO = {println(2); x};
f(Ref.get(p))
```
here arrowtype is `Int32 -> Int32 \ IO` and tpe is `Int32` and eff is `IO + r`.

main/src/ca/uwaterloo/flix/api/lsp/Indexer.scala
```diff
@@ -252,6 +256,16 @@ object Indexer {
     case Expr.LetRec(sym, _, _, exp1, exp2, _, _, _) =>
       Index.occurrenceOf(sym, exp1.tpe) ++ visitExp(exp1) ++ visitExp(exp2) ++ Index.occurrenceOf(exp0)
 
+    case Expr.LocalDef(sym, fparams, exp1, exp2, _, _, _) =>
+      val arrowType = Type.mkCurriedArrowWithEffect(fparams.map(_.tpe), exp1.eff, exp1.tpe, sym.loc)
```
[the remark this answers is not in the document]
JonathanStarup: arrowTpe is on the apply. ArrowTpe is not on the def because its redundant there. but it is a bit irregular yea

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -158,6 +158,17 @@ object ConstraintGen {
         val resEff = evar
         (resTpe, resEff)
 
+      case Expr.ApplyLocalDef(Ast.LocalDefSymUse(sym, loc1), exps, arrowTvar, tvar, evar, loc2) =>
+        val (tpes, effs) = exps.map(visitExp).unzip
+        val defEff = Type.freshVar(Kind.Eff, loc1.asSynthetic)
+        val actualDefTpe = Type.mkCurriedArrowWithEffect(tpes, defEff, tvar, loc1)
```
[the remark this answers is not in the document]
JonathanStarup: I agree, it shouldnt be curried. i talked to jakob about doing it but that was only relevant if we wanted to covert into the old AST nodes

main/src/ca/uwaterloo/flix/language/phase/ClosureConv.scala
```diff
@@ -398,4 +433,224 @@ object ClosureConv {
...
+    *
+    * becomes
+    *
+    * {{{
+    *   def f(): Int32 = {
+    *       let a = 1;
+    *       let b = 2;
+    *       let c = 3;
+    *       def g(x, y, z, l) = if (l == 4) x + y + z + l else g(x, y, z, 4);
+    *       g(a, b, c, 10)
```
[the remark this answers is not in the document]
magnus-madsen: I think tail recursive functions breaks this: Something like:

```
def foo(x: Int32, r: Unit -> Int32): Int32 -> Int32 = 
  def local() = x;
  if (x == 0) r else foo(x - 1, _ -> local())
 ```
which gets rewritten to a loop where the value of `x` changes (by mutation).

The point I am trying get across is that I will modify `x` between the time when the closure should have been created and when it is currently created.

File: main/src/ca/uwaterloo/flix/language/phase/LambdaLift.scala [hunk 35]
```diff
@@ -126,6 +126,14 @@ object LambdaLift {
       val es = exps.map(visitExp)
       LiftedAst.Expr.ApplyDef(sym, es, tpe, purity, loc)
 
+    case SimplifiedAst.Expr.ApplyLocalDef(sym, exps, tpe, purity, loc) =>
+      val es = exps.map(visitExp)
+      val newDefnSym = ctx.liftedLocalDefs.asScala.toMap.get(sym)
```
magnus-madsen, line 131: Doing asScala.toMap.get sounds computationally expensive.

magnus-madsen:
Finish my review. My one major concern is that your rewrite of ApplyLocalDef outside of the letrec body goes too far.

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1300,19 +1300,15 @@ object Weeder2 {
       }
 
       mapN(
-        annVal,
+        annVal, // Even though the annotation is unused, we still need to map over it to collect possible errors.
```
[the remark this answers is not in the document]
jaschdoc: My plan was to allow people to write them and handle them in Weeder but have no effect but maybe that's a bad idea?

main/src/ca/uwaterloo/flix/language/ast/TypedAst.scala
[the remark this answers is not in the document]
jaschdoc: Also, ApplyLocalDef does not have access to the definition the same way an ApplyDef has (via the root) so we cannot derive the arrow type from the formal parameters here.

main/src/ca/uwaterloo/flix/language/phase/LambdaLift.scala
jaschdoc: Do you want a call to `find` instead?

main/src/ca/uwaterloo/flix/language/phase/ClosureConv.scala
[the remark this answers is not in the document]
JonathanStarup: Im pretty sure your example doesn't break because the tail call is the last instruction, so the `_ -> local()` is still a closure creation that happens before the tail call.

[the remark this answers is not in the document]
jaschdoc: We can add it as a test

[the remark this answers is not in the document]
jaschdoc: Also, it just adds `x` to the list of arguments in the call but, crucially, before the lambda is closure converted so when visiting the lambda `_ -> local()` it will have `x` as a free variable of the body of the lambda and introduce it as a formal parameter of the lambda. That's why we do rewriting of `ApplyLocalDef` before anything else

main/src/ca/uwaterloo/flix/language/phase/ClosureConv.scala
```diff
@@ -398,4 +433,224 @@ object ClosureConv {
...
```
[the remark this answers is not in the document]
jaschdoc: Just a random number

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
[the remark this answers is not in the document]
magnus-madsen: You want to give a specific error message or?

main/src/ca/uwaterloo/flix/language/phase/ClosureConv.scala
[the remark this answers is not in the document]
JonathanStarup: but i think we have good material for Monday 👨‍🔬

main/src/ca/uwaterloo/flix/language/phase/ClosureConv.scala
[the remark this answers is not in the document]
magnus-madsen: But i was not in the program above?

main/src/ca/uwaterloo/flix/language/phase/LambdaLift.scala
magnus-madsen: Just not convert the entire thing to a Scala map

magnus-madsen: Find sounds fine

magnus-madsen: If there is some null check involved you can add a helper method to SharedContext.

jaschdoc: I mean `cttx.liftedLocalDefs.asScala.find(...)` is ok?

main/src/ca/uwaterloo/flix/language/phase/ClosureConv.scala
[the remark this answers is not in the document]
jaschdoc: True, that's a blunder.

main/src/ca/uwaterloo/flix/language/phase/LambdaLift.scala
magnus-madsen: Without the as scala

main/src/ca/uwaterloo/flix/language/phase/ClosureConv.scala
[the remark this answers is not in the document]
magnus-madsen: My counter-example may not be a counter-example, but you still need to convince yourself and me that the transformation is correct (and that it does not rely on informal invariants).

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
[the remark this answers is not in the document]
magnus-madsen: How about:

- Drop the comment
- Add an error saying that annotations are not allowed here

main/src/ca/uwaterloo/flix/language/phase/ClosureConv.scala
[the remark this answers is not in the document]
JonathanStarup: not relying on informal invariants seems like a higher constraint then many parts of the compiler 😄

main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
magnus-madsen: We should just do the same as we do for let, whatever that is.

main/src/ca/uwaterloo/flix/language/ast/TypedAst.scala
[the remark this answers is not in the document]
magnus-madsen: I think I understand now.

main/src/ca/uwaterloo/flix/api/lsp/Indexer.scala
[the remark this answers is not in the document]
magnus-madsen: I suppose a comment would be useful to explain this.

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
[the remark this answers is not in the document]
magnus-madsen: You answer confuses me. Do you mean that the code should be changed or not?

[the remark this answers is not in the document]
magnus-madsen: We agree it should just be: Type.mkUncurriedArrow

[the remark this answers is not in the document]
jaschdoc: This is in both `LocalDef` and `ApplyLocalDef`, yeah?

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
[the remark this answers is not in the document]
jaschdoc: I'm confused... the code stays more or less the same if we want to give an error saying that is not allowed. This comment is there because we do not use the `annVal` for anything other than including it in the map call so we collect its errors if there are any.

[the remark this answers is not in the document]
jaschdoc: It will go away when we introduce a `SharedContext`

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1304,19 +1304,15 @@ object Weeder2 {
...
         Types.tryPickType(tree),
         Types.tryPickEffect(tree),
       ) {
-        case (ann, (exp1, exp2), fparams, ident, tpe, eff) =>
-          val e = if (tpe.isDefined || eff.isDefined) Expr.Ascribe(exp1, tpe, eff, exp1.loc) else exp1
-          val lambda = fparams.foldRight(e) {
-            case (fparam, acc) => WeededAst.Expr.Lambda(fparam, acc, exp1.loc.asSynthetic)
-          }
-          Expr.LetRec(ident, ann, Modifiers.Empty, lambda, exp2, tree.loc)
+        case (_, (exp1, exp2), fparams, ident, declaredTpe, declaredEff) =>
```
jaschdoc, line 1314: @magnus-madsen notice the first entry here is `_`

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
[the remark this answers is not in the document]
magnus-madsen: Ok. Then leave it

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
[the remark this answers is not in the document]
magnus-madsen: I think so?

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
magnus-madsen: Kay

main/src/ca/uwaterloo/flix/api/lsp/Indexer.scala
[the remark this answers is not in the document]
jaschdoc: This one should be uncurried too, correct?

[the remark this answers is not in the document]
magnus-madsen: I think so.

jaschdoc:
We are ready for liftoff

magnus-madsen (approved):

magnus-madsen:
To the moon! Great job

mlutze:
Time to [watch the skies](https://flix.cs.au.dk/d/agddciz4k/main?orgId=1)

jaschdoc:
Related to https://github.com/flix/flix/issues/8606

PR #8679, opened 2024-09-21
proposal: magic match completions

magnus-madsen:
This PR demonstrates that we can support `expr.match` magic completions ala 

https://rust-analyzer.github.io/manual.html#magic-completions

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
@@ -111,7 +111,13 @@ object CompletionProvider {
       //
       case SyntacticContext.Expr.Constraint => PredicateCompleter.getCompletions(context) ++ KeywordCompleter.getConstraintKeywords
       case SyntacticContext.Expr.Do => OpCompleter.getCompletions(context)
-      case SyntacticContext.Expr.InvokeMethod(tpe, name) => InvokeMethodCompleter.getCompletions(tpe, name, context)
+      case SyntacticContext.Expr.InvokeMethod(tpe, name) =>
+        if ("match".startsWith(name.name)) {
+          MagicMatchCompleter.getCompletions(name, tpe)
+        } else {
+          InvokeMethodCompleter.getCompletions(tpe, name, context)
+        }
+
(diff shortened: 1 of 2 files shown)


magnus-madsen:
@LoZander Take a look.

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -111,7 +111,13 @@ object CompletionProvider {
       //
       case SyntacticContext.Expr.Constraint => PredicateCompleter.getCompletions(context) ++ KeywordCompleter.getConstraintKeywords
       case SyntacticContext.Expr.Do => OpCompleter.getCompletions(context)
-      case SyntacticContext.Expr.InvokeMethod(tpe, name) => InvokeMethodCompleter.getCompletions(tpe, name, context)
+      case SyntacticContext.Expr.InvokeMethod(tpe, name) =>
+        if ("match".startsWith(name.name)) {
```
magnus-madsen, line 115: This should not really be a branch, instead it should just add the two lists.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/MagicMatchCompleter.scala
```diff
@@ -0,0 +1,25 @@
...
+object MagicMatchCompleter {
+
+  def getCompletions(ident: Name.Ident, tpe: Type)(implicit root: TypedAst.Root): Iterable[Completion] = {
+    println("magic match")
+    getEnumSym(tpe) match {
+      case Some(sym) =>
+        val cases = root.enums(sym).cases
+        val s = "match " + ident.name + "{" + cases.map(_._1.name).mkString(" => \n ") + "}"
+
+        Completion.SnippetCompletion("color.match", s, "here is my documentation") :: Nil
```
magnus-madsen, line 14: I assumed an enum:

enum Color { case Red, case Green }

magnus-madsen:
image


image

magnus-madsen:
@sockmaster27 You can take over here, so @LoZander can focus on visitor.

A tricky thing here is that when we write `expr.match` it should become `match expr { ... `

So the question is how to create a textedit that somehow adds the `match` to _the left_ and the body _to the right_.

magnus-madsen:
This ticket was just for demonstration.

Issue #8676, opened 2024-09-21
Simple autocompletion improvements


magnus-madsen:
In reference to: https://github.com/flix/flix/blob/master/examples/struct-tree.flix

- [ ] `pub def search(k: k, t: Tree[k, v, r]): Option[v] \ r with O|` suggests impossible completions.
- [ ] Empty match does not suggest case.
- [ ] `None`/`Some` is not suggested as expr.
- [ ] `None`/`Some` is not suggested in pattern.
- [ ] `main` completion is suggested in the middle of an expression.

In reference to: https://github.com/flix/flix/blob/master/examples/struct-person.flix

- [ ] `mod P|` suggests annotations.
- [x] `Region[r]` missing as a type completion.
- [x] `new Person @ rc { name = name, ag| }` suggests match? When it is a missing field.
- [x] `pub def birthday(person: Person[r]): Unit \ r = p|` suggest project over person.

magnus-madsen:
@LoZander While the visitors should be your main focus, I suggest to also start looking into some of the above.

LoZander:
This first PR relates to the first point with `mod P|`. There was a deeper issue at play that I've solved

magnus-madsen:
@stormckey This ticket is a bit dated, but you could look into the above list.

For example, Some now seems to complete, but it does not insert `Some(_elm0)`.

PR #8685, opened 2024-09-23
fix: unfitting keyword completion suggestions in element headers (function signatures, module names, etc.)

LoZander:
A problem arised for keyword completion suggestions in which keywords that may occur within the body of some declaration would actually also occur within it's "header". Here header is everything that comes before the body.

For instance, we'd get the suggestion `def` for the following (where | is the cursor)

```
pub def d|
```

This PR fixes this issue by introducing distinct syntactic contexts for the headers such that keyword suggestions for these can be different from those of the corresponding bodies.

Related to #8676, specifically the first point

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
@@ -119,18 +119,18 @@ object CompletionProvider {
       //
       // Declarations.
       //
-      case SyntacticContext.Decl.Def      => Nil
-      case SyntacticContext.Decl.Enum     => KeywordCompleter.getEnumKeywords
-      case SyntacticContext.Decl.EnumHeader => Nil
-      case SyntacticContext.Decl.Instance => InstanceCompleter.getCompletions(context) ++ KeywordCompleter.getInstanceKeywords
+      case SyntacticContext.Decl.Def            => Nil
+      case SyntacticContext.Decl.Enum           => KeywordCompleter.getEnumKeywords
+      case SyntacticContext.Decl.EnumHeader     => Nil
+      case SyntacticContext.Decl.Instance       => InstanceCompleter.getCompletions(context) ++ KeywordCompleter.getInstanceKeywords
       case SyntacticContext.Decl.InstanceHeader => Nil
-      case SyntacticContext.Decl.Module   => KeywordCompleter.getModKeywords ++ SnippetCompleter.getCompletions(context)
-      case SyntacticContext.Decl.ModuleHeader => Nil
-      case SyntacticContext.Decl.Struct   => KeywordCompleter.getStructKeywords
-      case SyntacticContext.Decl.StructHeader => Nil
-      case SyntacticContext.Decl.Trait    => KeywordCompleter.getTraitKeywords
-      case SyntacticContext.Decl.TraitHeader => Nil
-      case SyntacticContext.Decl.Type     => KeywordCompleter.getTypeKeywords
+      case SyntacticContext.Decl.Module         => KeywordCompleter.getModKeywords ++ SnippetCompleter.getCompletions(context)
+      case SyntacticContext.Decl.ModuleHeader   => Nil
+      case SyntacticContext.Decl.Struct         => KeywordCompleter.getStructKeywords
+      case SyntacticContext.Decl.StructHeader   => Nil
+      case SyntacticContext.Decl.Trait          => KeywordCompleter.getTraitKeywords
+      case SyntacticContext.Decl.TraitHeader    => Nil
+      case SyntacticContext.Decl.Type           => KeywordCompleter.getTypeKeywords
 
       //
       // Imports.
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
@@ -119,11 +119,17 @@ object CompletionProvider {
       //
       // Declarations.
       //
+      case SyntacticContext.Decl.Def      => Nil
       case SyntacticContext.Decl.Enum     => KeywordCompleter.getEnumKeywords
+      case SyntacticContext.Decl.EnumHeader => Nil
       case SyntacticContext.Decl.Instance => InstanceCompleter.getCompletions(context) ++ KeywordCompleter.getInstanceKeywords
+      case SyntacticContext.Decl.InstanceHeader => Nil
       case SyntacticContext.Decl.Module   => KeywordCompleter.getModKeywords ++ SnippetCompleter.getCompletions(context)
+      case SyntacticContext.Decl.ModuleHeader => Nil
       case SyntacticContext.Decl.Struct   => KeywordCompleter.getStructKeywords
+      case SyntacticContext.Decl.StructHeader => Nil
       case SyntacticContext.Decl.Trait    => KeywordCompleter.getTraitKeywords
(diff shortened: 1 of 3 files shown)


main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -879,8 +879,8 @@ object Parser2 {
 
     private def moduleDecl(mark: Mark.Opened, nestingLevel: Int = 0)(implicit s: State): Mark.Closed = {
       assert(at(TokenKind.KeywordMod))
-      expect(TokenKind.KeywordMod, SyntacticContext.Decl.Module)
-      name(NAME_MODULE, allowQualified = true, context = SyntacticContext.Decl.Module)
+      expect(TokenKind.KeywordMod, SyntacticContext.Decl.ModuleHeader)
+      name(NAME_MODULE, allowQualified = true, context = SyntacticContext.Decl.ModuleHeader)
       expect(TokenKind.CurlyL, SyntacticContext.Decl.Module)
```
magnus-madsen, line 884: It seems to me it would be more appropriate to have:

SyntacticContext.Name
SyntacticContext.Keyword

LoZander: That makes sense. In all but a few cases, we don't care *what* kind of "header" we're in anyways. There might still be a reason to have a special context for the headers where "where" and "with" suggestions make sense. I know we decided to drop them suggestion support for them, but I think we might be able support them that way

LoZander:
I didn't really know what would be a fitting context for the parameters of a function, so I kept the Def context for that. But I could make a parameter context or just use some other context. I didn't think any of the contexts that existed before this really made sense for those parameters. Do you think a "parameter" context or something like that would be better?

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
magnus-madsen: There is no such thing as a header. It is an abstraction you have invented.

magnus-madsen:
> I didn't really know what would be a fitting context for the parameters of a function, so I kept the Def context for that. But I could make a parameter context or just use some other context. I didn't think any of the contexts that existed before this really made sense for those parameters. Do you think a "parameter" context or something like that would be better?

What would it add?

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -131,6 +132,16 @@ object CompletionProvider {
       //
       case SyntacticContext.Import => ImportCompleter.getCompletions(context)
 
+      //
+      // Keywords. Note that this is not the context in which keywords can occur, but the context of a keyword, i.e. inside a keyword.
```
magnus-madsen, line 136: Perhaps all context would benefit from such a description? And perhaps not here, but where the contexts are declared.

main/src/ca/uwaterloo/flix/language/ast/Ast.scala
```diff
@@ -564,7 +568,7 @@ object Ast {
 
       case object Do extends Expr
 
-      case class InvokeMethod(tpe: ca.uwaterloo.flix.language.ast.Type, name: Name.Ident) extends Expr
+      case class InvokeMethod(tpe: ca.uwaterloo.flix.language.ast.Type, name: ast.Name.Ident) extends Expr
```
magnus-madsen, line 571: What happened here?

magnus-madsen (commented):
This looks better. I will take a more detailed look later. Meanwhile you can fix my comments.

main/src/ca/uwaterloo/flix/language/ast/Ast.scala
LoZander: Strange name conflict. In this file, there's a `Name` object and now `SyntacticContext.Name`. That line is referring to the former, but since it's in the `SyntacticContext` object, the latter shadows the former. So to refer to the outer `Name` object, I'm importing ast and then doing `ast.Name`. It's a little clunky

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
LoZander: I'm aware that there's nothing named header in the code (except for what I added before). It's just a term I've been using to refer to a specific part of declarations. Most declarations have some kind of body or block and something that comes before that. For a module there's the name. For a function, there's the signature. For trait, there's name and optionally the with clause. For instances there's the name of the instance and what it's being implemented on. This part before the body or block, that's what I call the header. I didn't know what else to call it. A signature? But that's only what you'd call it for a function, right? I don't know if there is some other common term for it

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
LoZander: Hmm. Yes. You might be right. I added this here because I sensed a danger of confusion due to there being a KeywordCompleter, which has no relation to SyntacticContext.Keyword. But you're right, I think it would be good to added such descriptions as docs for at the decl site

main/src/ca/uwaterloo/flix/language/ast/Ast.scala
magnus-madsen: I see.

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
magnus-madsen: OK, so lets move the comments to there and expand.

magnus-madsen:
Did you test that keyword completions still work as expected?

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -879,8 +879,8 @@ object Parser2 {
 
     private def moduleDecl(mark: Mark.Opened, nestingLevel: Int = 0)(implicit s: State): Mark.Closed = {
       assert(at(TokenKind.KeywordMod))
-      expect(TokenKind.KeywordMod, SyntacticContext.Decl.Module)
-      name(NAME_MODULE, allowQualified = true, context = SyntacticContext.Decl.Module)
+      expect(TokenKind.KeywordMod, SyntacticContext.Keyword)
```
magnus-madsen, line 882: Did you change it to use SyntacticContext everywhere a TokenKind.Keyword is used?

LoZander: i changed it to use `SyntacticContext.Keyword` for expect denoting the start of a declaration, and `SyntacticContext.Name` for every name that follows. Potential other cases of `expect` on `TokenKind.Keyword`, I didn't touch. But perhaps I should?

magnus-madsen: You should be consistent

LoZander: I'll go over it again to check for other `expect` with TokenKind.Keyword and name and make sure it's consistent :)

LoZander:
> Did you test that keyword completions still work as expected?

damn. I just did, and it seems I broke it somewhere along the line. Well, I made it like this PR had never started, is what I mean. It seems we're not in the name context when expected. Hm.

magnus-madsen:
Seems difficult.

LoZander:
So, as we discussed, we'll just let this one go, at least for now. Too difficult. So I'll move onto other things :)

PR #8687, opened 2024-09-23
fix: match completion suggestions at odd times

LoZander:
This PR fixed an issue where match completions are suggested at odd times. This was due to a too "forgiving" regex which has now been fixed.

Now match completions are only suggested if you've just typed "match" or if the previous word was "match" and the current word matches an enum.

Related to #8676, specifically item 3


magnus-madsen (approved):

magnus-madsen:
I think ultimately the match completer should be removed in favor of the the magic match completer.

LoZander:
> I think ultimately the match completer should be removed in favor of the the magic match completer.

I was thinking that too :)

PR #8688, opened 2024-09-23
fix: priority for vars generally too low

LoZander:
Priorities for variables were too low relative to keywords. Priorities for variables have now been raised to Higher and the priority of keywords that can occur in expressions have been lowered to high or below.

Related to #8676, item 4


magnus-madsen:
You want to change all priorities every time there is a conflict? That seems unsustainable. 

How about:

- Either just boosting local variables to maximum, or 
- Adding increase/decrease priority functions and using them when multiple providers are in play?

LoZander:
I've run into an issue. A varSym tells you very little. When hovering on a varSym, I'd except the type of the variable. But the var sym doesn't contain that information, so you'll have to backtrack on the stack to find it's parent element, and then depending on what kind of element that is, determine the type. That's kind of a mess since we're operating on `AnyRef`

PR #8703, opened 2024-09-24
feat: add support for varargs

magnus-madsen:
Example:

```scala
import java.nio.file.Path

@Test
def testInvokeVarArgs01(): Path \ IO =
    Path.of("hello", (...{}: Vector[String]))

@Test
def testInvokeVarArgs02(): Path \ IO =
    Path.of("hello", ...{"foo.txt"})

@Test
def testInvokeVarArgs03(): Path \ IO =
    Path.of("hello", ...{"foo.txt", "bar.txt"})
```

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
@@ -322,6 +322,17 @@ object TypeReduction {
       val arrType = phantomArr.getClass
       arrType
 
+    // Vectors
+    case Type.Apply(Type.Cst(TypeConstructor.Vector, _), elmType, _) =>
+      // Find the Java class of the element type.
+      val t = getJavaType(elmType)
+
+      // Find the Java class of the array type.
+      // See: https://stackoverflow.com/questions/1679421/how-to-get-the-array-class-for-a-given-class-in-java
+      val phantomArr = java.lang.reflect.Array.newInstance(t, 0)
+      val arrType = phantomArr.getClass
+      arrType
+
     // Functions
     case Type.Apply(Type.Apply(Type.Apply(Type.Cst(TypeConstructor.Arrow(2), _), _, _), varArg, _), varRet, _) =>
       (varArg, varRet) match {
--- a/main/test/flix/Test.Exp.Jvm.InvokeMethod.VarArgs.flix
+++ b/main/test/flix/Test.Exp.Jvm.InvokeMethod.VarArgs.flix
@@ -1,0 +1,9 @@
+mod Test.Exp.Jvm.InvokeMethod.VarArgs {
+
+    import java.nio.file.Path
+
+    def main(): Unit \ IO =
+        let foo = Path.of("hello", ...{"abc"});
+        println(foo.toString())
+
+}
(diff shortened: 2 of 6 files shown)


magnus-madsen:
@mlutze You may not like it... I don't even like it... But I want a solution for varargs which does not require a region.

I think this is a useful start. And we can iterate.

mlutze:
I'm confused. Why are you calling a varargs function with an array? Why not use it as a varargs function?

magnus-madsen:
> I'm confused. Why are you calling a varargs function with an array? Why not use it as a varargs function?

The JVM has no concept of varargs. Moreover, commons-lang3 has no concept of varargs.

mlutze:
> > I'm confused. Why are you calling a varargs function with an array? Why not use it as a varargs function?
> 
> The JVM has no concept of varargs. Moreover, commons-lang3 has no concept of varargs.

The JVM has no concept of generics, either. Ihe JVM isn't our interface; whatever they put in JAR files is our interface.


That said, as long as we only allow putting literals in, I guess it's safe.

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
```diff
@@ -322,6 +322,17 @@ object TypeReduction {
       val arrType = phantomArr.getClass
       arrType
 
+    // Vectors
+    case Type.Apply(Type.Cst(TypeConstructor.Vector, _), elmType, _) =>
+      // Find the Java class of the element type.
+      val t = getJavaType(elmType)
+
+      // Find the Java class of the array type.
+      // See: https://stackoverflow.com/questions/1679421/how-to-get-the-array-class-for-a-given-class-in-java
```
mlutze, line 331: This is a gammeldags solution. Class arrayType is more direct. Maybe we can go through all instances of this in another PR

main/test/flix/Test.Exp.Jvm.InvokeMethod.VarArgs.flix
```diff
@@ -0,0 +1,17 @@
+mod Test.Exp.Jvm.InvokeMethod.VarArgs {
+
+    import java.nio.file.Path
+
+    @Test
+    def testInvokeVarArgs01(): Path \ IO =
+        Path.of("hello", (...{}: Vector[String]))
```
mlutze, line 7: IMO it's best if we require a literal vector here, to ensure no one else owns the vector and depends on its immutability.

But I guess that does not work for the empty case.

mlutze (approved):

magnus-madsen:
This PR is allows programmers to use ordinary Java constructors/methods with varargs with requiring them to learn/understand regions. Moreover, it provides some new syntax so there is a path towards a better implementation.

main/test/flix/Test.Exp.Jvm.InvokeMethod.VarArgs.flix
magnus-madsen: Sadly the common case of empty varargs is a case we cannot easily support :(

main/src/ca/uwaterloo/flix/language/phase/typer/TypeReduction.scala
magnus-madsen: See if you can get it to work. I could not!

PR #8705, opened 2024-09-24
refactor: rework constructor resolution

magnus-madsen:


magnus-madsen:
@mlutze I think it would be nice to have this in before next release.

magnus-madsen:
I did some cleanup too.

magnus-madsen:
@mlutze Ping

mlutze (approved):

PR #8707, opened 2024-09-24
refactor: extend FilePath

magnus-madsen:


magnus-madsen:
The intention is for a student programmer to fill in more details later. I am just adding some skeleton code.

main/src/library/FilePath.flix
```diff
@@ -25,38 +28,130 @@ pub eff FilePath {
...
+
+    ///
+    /// Returns the last access time of the given file `f` in milliseconds since the epoch.
+    ///
+    pub def accessTime(f: String): Int64
+
+    ///
+    /// Returns the creation time of the given file `f` in milliseconds since the epoch.
+    ///
+    pub def creationTime(f: String): Int64
```
mlutze, line 68: I guess we can cheat and get an almost-clock effect here.

mlutze (approved):

PR #8708, opened 2024-09-24
docs: single line doc comments

JonathanStarup:
As jakob pointed out, it was easy with a regex. @magnus-madsen you can merge if you want or just close


magnus-madsen:
Thanks for the experiment-- but I think this goes a bit too far for my taste.

Issue #8715, opened 2024-09-25
Benchmarks: don't print anything


mlutze:
Or print to stderr if needed. Otherwise the output is invalid.

PR #8716, opened 2024-09-25
test: do not print in benchmarks

jaschdoc:
Closes https://github.com/flix/flix/issues/8715


magnus-madsen:
Can I ask why you print? Just use blackhole

magnus-madsen (approved):

PR #8719, opened 2024-09-25
refactor: remove unused `Enum.tpe` from `TypedAst` and `LoweredAst`

JonathanStarup:
This is redundant (and unused information) because tparams is also kept.


JonathanStarup:
@mlutze if you have the time, could you approve?

mlutze (approved):

PR #8722, opened 2024-09-25
fix: simplificy struct declaration types in monomoprh

JonathanStarup:
the substitution must be applied to handle assoc types, aliases, etc


JonathanStarup:
Broken for complicated reasons - will fix in monomorph revamp

PR #8727, opened 2024-09-25
refactor: principled Monomorpher

JonathanStarup:
blocked by  #8728


JonathanStarup:
@magnus-madsen here is a version 0 of the new monomorpher - ready for optimized enums

magnus-madsen:
Quick service! 👍 

What's the cliffnotes? What is the most important changes I need to understand? What makes it version 0?

What's the expected impact on compiler performance? On code size?

I see some test failures. Are those easy or is there some more fundamental problem?

main/src/ca/uwaterloo/flix/language/phase/Eraser.scala
```diff
@@ -201,16 +226,16 @@ object Eraser {
...
+      case MonoType.Struct(sym, tparams) => MonoType.Struct(sym, tparams.map(erase))
       case Arrow(args, result) => Arrow(args.map(visitType), box(result))
       case RecordEmpty => RecordEmpty
       case RecordExtend(label, value, rest) => RecordExtend(label, erase(value), visitType(rest))
       case Native(clazz) => Native(clazz)
     }
   }
 
-  private def erase(tpe: MonoType): MonoType = {
+  def erase(tpe: MonoType): MonoType = {
```
magnus-madsen, line 238: Why is this public?

magnus-madsen: If it used elsewhere maybe it should be moved?

main/src/ca/uwaterloo/flix/language/phase/Eraser.scala
```diff
@@ -221,13 +246,34 @@ object Eraser {
...
-           Region | Array(_) | Lazy(_) | Tuple(_) | MonoType.Enum(_) |
-           MonoType.Struct(_, _, _) | Arrow(_, _) | RecordEmpty | RecordExtend(_, _, _) | Native(_) | Null =>
+      case Void | AnyType | Unit | BigDecimal | BigInt | String | Regex | Region | Array(_) |
+           Lazy(_) | Tuple(_) | MonoType.Enum(_, _) | MonoType.Struct(_, _) | Arrow(_, _) |
+           RecordEmpty | RecordExtend(_, _, _) | Native(_) | Null =>
         MonoType.Object
     }
   }
 
+  private def erase(tpe: Type): Type = tpe match {
```
magnus-madsen, line 256: Add comment

magnus-madsen: I think both erase functions deserve some detailed comments to explain why there are two. And some examples to demonstrates what they translate.

main/src/ca/uwaterloo/flix/language/phase/Eraser.scala
```diff
@@ -221,13 +246,34 @@ object Eraser {
...
+      case TypeConstructor.Float32 => c
+      case TypeConstructor.Float64 => c
+      case TypeConstructor.Int8 => c
+      case TypeConstructor.Int16 => c
+      case TypeConstructor.Int32 => c
+      case TypeConstructor.Int64 => c
+      case _ => Type.Cst(TypeConstructor.Native(classOf[Object]), loc)
+    }
+    case Type.Apply(_, _, loc) => Type.Cst(TypeConstructor.Native(classOf[Object]), loc)
+    case Type.Alias(_, _, _, _) => throw InternalCompilerException(s"Unexpected type $tpe", tpe.loc)
```
magnus-madsen, line 270: add linebreak to separate error cases.

main/src/ca/uwaterloo/flix/language/phase/Eraser.scala
```diff
@@ -221,13 +246,34 @@ object Eraser {
...
+    case c@Type.Cst(tc, loc) => tc match {
+      case TypeConstructor.Bool => c
+      case TypeConstructor.Char => c
+      case TypeConstructor.Float32 => c
+      case TypeConstructor.Float64 => c
+      case TypeConstructor.Int8 => c
+      case TypeConstructor.Int16 => c
+      case TypeConstructor.Int32 => c
+      case TypeConstructor.Int64 => c
+      case _ => Type.Cst(TypeConstructor.Native(classOf[Object]), loc)
```
magnus-madsen, line 267: Add a comment why this is correct

main/src/ca/uwaterloo/flix/language/phase/Eraser.scala
```diff
@@ -221,13 +246,34 @@ object Eraser {
...
+    }
+    case Type.Apply(_, _, loc) => Type.Cst(TypeConstructor.Native(classOf[Object]), loc)
```
magnus-madsen, line 269: Add a comment why this is correct.

main/src/ca/uwaterloo/flix/language/phase/Eraser.scala
```diff
@@ -221,13 +246,34 @@ object Eraser {
...
+    case v@Type.Var(_, _) => v
+    case c@Type.Cst(tc, loc) => tc match {
+      case TypeConstructor.Bool => c
+      case TypeConstructor.Char => c
+      case TypeConstructor.Float32 => c
+      case TypeConstructor.Float64 => c
+      case TypeConstructor.Int8 => c
+      case TypeConstructor.Int16 => c
+      case TypeConstructor.Int32 => c
+      case TypeConstructor.Int64 => c
```
magnus-madsen, line 266: What about TypeConstructors like: Regex, functions, records, etc? Native?

magnus-madsen: Perhaps some of these should be explicitly mapped to Object, and the rest (which are assumed to have been erased) should throw ICE?

main/src/ca/uwaterloo/flix/language/phase/MonoTypes.scala
```diff
@@ -1,355 +0,0 @@
-/*
```
magnus-madsen, line 1: This is gone for the simply reason that we no longer monomorph types?

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -211,16 +211,14 @@ object Verifier {
 
         case AtomicOp.Is(sym) =>
           val List(t1) = ts
-          check(expected = MonoType.Enum(sym.enumSym))(actual = t1, loc)
           check(expected = MonoType.Bool)(actual = tpe, loc)
 
         case AtomicOp.Tag(sym) =>
           val List(t1) = ts
-          check(expected = MonoType.Enum(sym.enumSym))(actual = tpe, loc)
+          tpe
```
magnus-madsen, line 218: Is there no check we can add?

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
```diff
@@ -211,16 +211,14 @@ object Verifier {
...
 
         case AtomicOp.Untag(sym) =>
           val List(t1) = ts
-          check(expected = MonoType.Enum(sym.enumSym))(actual = t1, loc)
```
magnus-madsen, line 223: Ditto?

main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
```diff
@@ -379,38 +379,60 @@ object BackendObjType {
     }
   }
 
-  case class Tag(tpe: BackendType) extends BackendObjType with Generatable {
+  case class Tag(elms: List[BackendType]) extends BackendObjType with Generatable {
```
magnus-madsen, line 382: So this already preparing for n-ary tags?

main/src/ca/uwaterloo/flix/language/phase/Monomorpher.scala
```diff
@@ -128,21 +128,25 @@ object Monomorpher {
     /**
```
magnus-madsen, line 128: I think this class has some documentation at the top that should now be updated?

main/src/ca/uwaterloo/flix/language/phase/Monomorpher.scala
```diff
@@ -128,21 +128,25 @@ object Monomorpher {
...
       * Applies `this` substitution to the given type `tpe`, returning a normalized type.
       *
+      * `keepVariables` can only be `true` if `this.s` is empty. It will avoid removing variables.
```
magnus-madsen, line 131: __First__ describe what the flag does. __Then__ describe its invariants.

magnus-madsen: This change seem to erode the subst. abstraction. 

If you want the effects of the Type.Apply cases-- why not extract them into a separate function?

magnus-madsen: __Alternatively__ why not parameterize over the default case?

magnus-madsen: But I much prefer to not break Subst. like this. Better to extract the functionality.

main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
```diff
@@ -42,10 +42,12 @@ object Reducer {
 
     val newDefs = ParOps.parMapValues(root.defs)(visitDef)
     val defTypes = ctx.defTypes.keys.asScala.toSet
-
+    // This is an over approximation of the types in enums and structs since they are erased.
+    val enumTypes = Set(MonoType.Bool, MonoType.Char, MonoType.Int8, MonoType.Int16, MonoType.Int32, MonoType.Int64, MonoType.Float32, MonoType.Float64, MonoType.Object)
```
magnus-madsen, line 46: This looks like all the JVM types? Should this be in the Jvm util thing?

main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
```diff
@@ -42,10 +42,12 @@ object Reducer {
...
+    val structTypes = enumTypes
     val effectTypes = root.effects.values.toSet.flatMap(typesOfEffect)
 
-    val types = nestedTypesOf(Set.empty, Queue.from(defTypes ++ effectTypes))
+    val types = nestedTypesOf(Set.empty, Queue.from(effectTypes ++ structTypes ++ enumTypes ++ defTypes))
```
magnus-madsen, line 50: I dream of the day we can get rid of nestedTypesOf :O

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -373,6 +374,137 @@ object Simplifier {
...
+          case TypeConstructor.Receiver => throw InternalCompilerException("Unexpected Receiver", tpe.loc)
+
+          case TypeConstructor.Lazy => Type.mkApply(Type.Cst(tc, tpe.loc), args, tpe.loc)
+
+          case TypeConstructor.Enum(sym, _) => Type.mkApply(Type.Cst(tc, tpe.loc), args, tpe.loc)
+
+          case TypeConstructor.Struct(sym, _) => Type.mkApply(Type.Cst(tc, tpe.loc), args, tpe.loc)
+
+          case TypeConstructor.RestrictableEnum(sym, _) =>
+            val enumSym = new Symbol.EnumSym(sym.namespace, sym.name, sym.loc)
```
magnus-madsen, line 432: Add a comment to explain

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -373,6 +374,137 @@ object Simplifier {
...
+            Type.mkEnum(enumSym, args, tpe.loc)
+
+          case TypeConstructor.Native(clazz) => Type.Cst(tc, tpe.loc)
+
+          case TypeConstructor.Array =>
+            // remove region
+            // OBS: non-well kinded type
```
magnus-madsen, line 439: Yikes, okay

magnus-madsen: perhaps WARNING: This construct a ...

magnus-madsen: I assume the reason is that we don't want to monomorph w.r.t. regions?

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -373,6 +374,137 @@ object Simplifier {
...
+            Type.Apply(Type.Cst(TypeConstructor.Array, tpe.loc), args.head, tpe.loc)
+
+          case TypeConstructor.RegionToStar => Type.mkApply(Type.Cst(tc, tpe.loc), args, tpe.loc)
+
+          case TypeConstructor.Tuple(_) => Type.mkApply(Type.Cst(tc, tpe.loc), args, tpe.loc)
+
+          case TypeConstructor.Arrow(_) =>
+            // remove effect
+            // OBS: non-well kinded type
+            Type.mkApply(Type.Cst(TypeConstructor.Arrow(args.length - 1), tpe.loc), args.drop(1), tpe.loc)
```
magnus-madsen, line 453: I think breaking these invariants is terrible. What can be done?

magnus-madsen: Perhaps we need a non-Type, non-MonoType type?

magnus-madsen: because these violations seems to extreme

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmOps.scala
```diff
@@ -344,13 +346,62 @@ object JvmOps {
...
+    types.foldLeft(Set.empty[BackendObjType.Tag]) {
+      case (acc, MonoType.Enum(sym, targs)) =>
+        val tags = instantiateEnum(root.enums(sym), targs)
+        tags.foldLeft(acc){
+          case (acc, tagElms) => acc + BackendObjType.Tag(tagElms)
+        }
+      case (acc, _) => acc
+    }
+
+  private def instantiateEnum(enm: ReducedAst.Enum, targs: List[MonoType]): List[List[BackendType]] = {
```
magnus-madsen, line 375: Can we get some documentation?

Also, when/where is this called.

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmOps.scala
```diff
@@ -344,13 +346,62 @@ object JvmOps {
...
+    assert(enm.tparams.length == targs.length)
+    val map = enm.tparams.map(_.sym).zip(targs).toMap
+    enm.cases.map{
+      case (_, caze) => List(caze.tpe).map(instantiateType(map, _))
+    }.toList
+  }
+
+  private def instantiateType(map: Map[Symbol.KindedTypeVarSym, MonoType], tpe: Type): BackendType = tpe match {
```
magnus-madsen, line 383: Can we get some documentation?

Also, when/where is this called.

magnus-madsen:
Would it work to add:

```
sealed trait PolyType
object PolyType { 
  case class Var(...) extends PolyType
  case class Mono(tpe: MonoType, ...) extends PolyType
}
```

That seems elegant, if it works?

magnus-madsen:
MonoType is less than 100 lines. Perhaps we just have to copy it. Alternatively, we have to add a variable to it.

magnus-madsen:
Is this the use case for restrictable variants?

JonathanStarup:
> What's the cliffnotes? What is the most important changes I need to understand? What makes it version 0?

Instead of the current handful of hacks to monomorphize datatypes pre-eraser, we keep datatypes polymorphic. This means that phases such as Lowering and Simplicifier must also perform their logic on types with variables. Once we hit Eraser and afterwards, it is easy to collapse the recursive datatypes via Object erasure, no tricks needed.

It is version zero because of lacking documentation and lack of polish.

> What's the expected impact on compiler performance? On code size?

The additional work is just traversing enums and structs in a couple more phases, but no complicated logic.

The code generated is identical, so no changes there. This just allows optimized enums in the future.

> I see some test failures. Are those easy or is there some more fundamental problem?

small issues:
- One stems from my wrong assumption about Simplifier
- One is some unknown assertion in Verfier in a datalog - lambda, might be a mistyping of generated code, I need more debugging.

> What about making MonoType + Variable?

It is tempting, but since we have variables, we also can't "flatten" types, because of enums like this
```
enum E[t: Type -> Type] { case T(t[Int32]) }
```
So since we would need both variables, general type application, and schema rows fx, I don't think it's worth it

main/src/ca/uwaterloo/flix/language/phase/MonoTypes.scala
JonathanStarup: currently we convert `Option[Int32]` into `Option`. This new approach just leaves `Option[Int32]` as it is. The same goes for structs, they are maintained as they are instead of unfolded into their element types.

main/src/ca/uwaterloo/flix/language/phase/Verifier.scala
JonathanStarup: There is, but I wanted to work before spending the effort

main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
JonathanStarup: Yeah exactly - will be its own PR, but here the list is always just one element.

main/src/ca/uwaterloo/flix/language/phase/Monomorpher.scala
JonathanStarup: yes, have yet to do a documentation pass

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
JonathanStarup: Monomorphization has already happened. It would still work if left out, but I think the only sane option is to apply the phase logic to both the MonoTypes and the enum/struct types. Otherwise, you have cross-phase invariants, and types become out of sync

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
JonathanStarup: Yeah its not great. Here is a sorts of types I have:
- Monomoprh: Types without Alias, Associated types, Jvm Types (here with variables, but monomorpher would like one version of this also without variables)
- Simplifier: The above, but now also with different Kinding of array and arrows, 
- Eraser: A type that can only be a variable, a java primite, or object

main/src/ca/uwaterloo/flix/language/phase/Eraser.scala
JonathanStarup: The constructors you mention are still erased to object. 

I think its too much work checking the shape of the type here. It also couples the phases since the eraser really doesn't care for correctness

main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
JonathanStarup: It's not pretty but it seems required in some form to figure out what tuples and functions to generate

magnus-madsen:
@JonathanStarup We still want to do this right?

JonathanStarup:
yes, it is on my list once the typer stuff is under control

JonathanStarup:
new version #9077

PR #8741, opened 2024-09-28
feat: add with-handler completion

### LINKED ISSUE #5995: Add `WithHandler` completion
Add completion for:

```
def f(): Unit = try { ... } with [EffectList + Operations]
```

sockmaster27:
Fixes #5995

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/WithHandlerCompleter.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/WithHandlerCompleter.scala
@@ -1,0 +1,25 @@
+package ca.uwaterloo.flix.api.lsp.provider.completion
+
+import ca.uwaterloo.flix.api.Flix
+import ca.uwaterloo.flix.api.lsp.{Index, InsertTextFormat, TextEdit}
+import ca.uwaterloo.flix.language.ast.TypedAst
+
+object WithHandlerCompleter {
+  def getCompletions(context: CompletionContext)(implicit flix: Flix, index: Index, root: TypedAst.Root): Iterable[Completion] = {
+    if (context.previousWord != "with") {
+      return Nil
+    }
+
+    // TODO: infer the effect of the expression
+    root.effects.map { case (sym, eff) =>
+      val effString = sym.name
+      val opStrings = eff.ops.map {
+        case TypedAst.Op(sym, spec) =>
+        val fparamsString = (spec.fparams.map(p => p.sym.text) :+ "k").mkString(", ")
+        s"    def ${sym.name}($fparamsString) = ???"
+      }
+      val bodyString = s"$effString {\n${opStrings.mkString("\n")}\n}"
+      Completion.WithHandlerCompletion(effString, TextEdit(context.range, bodyString))
+    }
+  }
+}
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
@@ -16,8 +16,8 @@
 package ca.uwaterloo.flix.api.lsp.provider
 
 import ca.uwaterloo.flix.api.Flix
-import ca.uwaterloo.flix.api.lsp._
-import ca.uwaterloo.flix.api.lsp.provider.completion._
+import ca.uwaterloo.flix.api.lsp.*
+import ca.uwaterloo.flix.api.lsp.provider.completion.*
 import ca.uwaterloo.flix.language.CompilationMessage
 import ca.uwaterloo.flix.language.ast.Ast.SyntacticContext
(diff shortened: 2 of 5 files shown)


sockmaster27:
Generates this part:
image

@magnus-madsen What should I do about the `_unit`?

magnus-madsen:
I think you leave unit as is.

magnus-madsen:
the unit thing is a quirk, but not something to fix in this PR.

magnus-madsen:
How does the suggestion list look when you type with M

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/WithHandlerCompleter.scala
```diff
@@ -0,0 +1,25 @@
+package ca.uwaterloo.flix.api.lsp.provider.completion
```
magnus-madsen, line 1: Add copyright header

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/WithHandlerCompleter.scala
```diff
@@ -0,0 +1,25 @@
...
+
+import ca.uwaterloo.flix.api.Flix
+import ca.uwaterloo.flix.api.lsp.{Index, TextEdit}
+import ca.uwaterloo.flix.language.ast.TypedAst
+
+object WithHandlerCompleter {
+  def getCompletions(context: CompletionContext)(implicit flix: Flix, index: Index, root: TypedAst.Root): Iterable[Completion] = {
+    if (context.previousWord != "with") {
```
magnus-madsen, line 9: Ideally this would not be necessary. Do we sometime get the wrong syntactic context or what?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/WithHandlerCompleter.scala
```diff
@@ -0,0 +1,25 @@
...
+      return Nil
+    }
+
+    // TODO: infer the effect of the expression
```
magnus-madsen, line 13: What this about?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/WithHandlerCompleter.scala
```diff
@@ -0,0 +1,25 @@
...
+    root.effects.map { case (sym, eff) =>
+      val effString = sym.name
+      val opStrings = eff.ops.map {
+        case TypedAst.Op(sym, spec) =>
+        val fparamsString = (spec.fparams.map(p => p.sym.text) :+ "k").mkString(", ")
+        s"    def ${sym.name}($fparamsString) = ???"
```
magnus-madsen, line 19: Can we add a private helper to construct each def-case string?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/WithHandlerCompleter.scala
sockmaster27: I just saw it in some other instances and thought I'd be consistent

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/WithHandlerCompleter.scala
sockmaster27: Ideally when writing 
```flix
try f() with 
```
you would only get the snippet for handling the effect that `f` has, if it was known. I couldn't get that to work unfortunately, and I don't know if we even have that information when the full statement isn't parsable, so now it's just suggesting all effects.

sockmaster27:
> How does the suggestion list look when you type with M

The biggest problem right now is that for some reason it only activates on lowercase:
image

image
image

I'm starting to suspect it might have something to do with the parser treating uppercase and lowercase differently?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/WithHandlerCompleter.scala
magnus-madsen: OK; we are moving away from that "old style". Ideally the syntactic context should be enough.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/WithHandlerCompleter.scala
magnus-madsen: I see. Thats going to be difficult. I would just focus on offering a list of suggestions once a letter has been typed.

sockmaster27:
> The biggest problem right now is that for some reason it only activates on lowercase

Found out this is due to an uppercase letter triggering an `OtherExpr` syntactic context. Trying to look into why.

JonathanStarup:
Yeah I think we should move away for checking casing in the parser when somebody has time to work on it. Makes worse errors

magnus-madsen (approved):

PR #8746, opened 2024-09-30
refactor: remove `Validation` from `Kinder`

jaschdoc:
This is the final step in removing `Validation` from `Kinder`.

Closes https://github.com/flix/flix/issues/8546

PR #8748, opened 2024-09-30
refactor: use parallelism on both levels of `root.instances` in `Kinder`

jaschdoc:
Related to https://github.com/flix/flix/pull/8746

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
@@ -74,7 +74,11 @@ object Kinder {
 
     val defs = visitDefs(root, taenv, oldRoot, changeSet)
 
-    val instances = ParOps.parMapValues(root.instances)(_.map(visitInstance(_, taenv, root)))
+    val instances = root.instances.map {
+      case (sym, instances0) =>
+        val is = ParOps.parMap(instances0)(visitInstance(_, taenv, root)).toList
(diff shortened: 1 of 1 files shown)


main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
```diff
@@ -74,10 +74,7 @@ object Kinder {
 
     val defs = visitDefs(root, taenv, oldRoot, changeSet)
 
-    // Should `instances` not be root.instances.map(ParOps.parMap)?
-    // I.e., for every trait, we check all instances in parallel so we get a parallel task for every instance
-    // instead of check all traits in parallel which spawns k threads each does synchronous code
-    val instances = ParOps.parMapValues(root.instances)(_.map(visitInstance(_, taenv, root)))
+    val instances = ParOps.parMapValues(root.instances)(ParOps.parMap(_)(visitInstance(_, taenv, root)))
```
magnus-madsen, line 77: Oooh. I misunderstood. Don't do _nested parallelism_ that is not good. 

I thought you wanted to move the level, not do nested parallelism.

jaschdoc: Alright :)

jaschdoc: That was actually my initial idea but then I realized I could do this :D

PR #8753, opened 2024-09-30
fix: splitting the side-condition check in a T-Choose-* rule to increase performance.

Alex1005a:
The side-condition in the T-Choose-* rule results in very large formulas that slow down compilation very much. For example, this code cannot compile on my computer in any acceptable time.
```scala
restrictable enum Exp[s] {
        case Var(Int32)
        case Cst(Bool)
        case Not(Exp[s])
        case Or(Exp[s], Exp[s])
        case And(Exp[s], Exp[s])
        case Xor(Exp[s], Exp[s])
    }

def simplify(e: Exp[s rvsub <Exp.Var>]): Exp[(s rvsub <Exp.Xor>) rvadd <Exp.And>] = 
    choose* e {
        case Exp.Cst(b)     => open_variant Exp.Cst(b)
        case Exp.Not(x)     => open_variant Exp.Not(simplify(x))
        case Exp.Or(x, y)   => open_variant Exp.Or(simplify(x), simplify(y))
        case Exp.And(x, y)  => open_variant Exp.And(simplify(x), simplify(y))
        case Exp.Xor(x, y)  => open_variant Exp.And(simplify(x), simplify(y))
    }
```
I split the side-condition check into checking multiple conditions. I did not formally prove that this is correct, but I came to this conclusion based on that intersection distributes over union and union of any set with the empty set is the set itself.
This change has given a significant performance boost (at least now compilation completes in a few seconds).


magnus-madsen:
Very cool.

magnus-madsen:
@Alex1005a Would you mind swinging by https://app.gitter.im/#/room/#flix_Lobby:gitter.im ?

This is obviously non-trivial stuff, so we'd like to chat.

mlutze:
Interesting, this change causes a failure in other tests.

JonathanStarup:
It is definitely a good idea, but it also has downsides. Each call to unifySubset introduces a new type variable. The solver takes exponential time based on the number of variables in the worst case. That is why this might be slower in some cases and faster in others. We could avoid those additional type variables in this case (I think), but that requires the new solver, which is currently being merged bit by bit

Alex1005a:
@JonathanStarup I think for now we can add heuristics to solve this problem, for example, split the formula if it is larger than a certain size. At the very least, this solves the Out of memory problem in tests.

JonathanStarup:
Even though the individual formulas are small, their solution might be very big. The solution for `φ_out` in the first equation is substituted into the next formula, and so on and so on.

Alex1005a:
Do you mean that splitting the formula will increase the substitution costs? I agree with this, but for very large formulas it turns out that it is better to accept the cost of substitution, since the final performance improves.

JonathanStarup:
I mean that, since your reformulation is equivalent, the solution for the variables in the system will remain equivalent. I'd think that the separation of the constraints could only lead to coincidental performance gain (solver ordering changes - which might yield smaller formulas) but in the general case it could also be worse

again - with the new solver I think your approach is just better

Alex1005a:
Probably, the increase in perforamance occurs in case of splitting large formulas with several flexible variables. Because in the function `successiveVariableElimination` the formula increases approximately twice for each variable each time. It turns out to be an exponential growth of the already large formula. Therefore, the final formula obtained from small formulas with one or two flexible variables should essentially be smaller.

However, for now we can wait for a new solver.

JonathanStarup:
I hope I'm not misunderstanding, but if we have three small formulas instead of one of thrice the size, let's say `x ~ f1`, `x ~ f2`, `x ~ f3`, then the solving of the first constraint is easy as you say. So we solve `x ~ f1` and figure out that the most general solution for x is `x |-> y + {Cst} + not z` fx. When we go to the next constraint, we must solve ` y + {Cst} + not z ~ f2` because `x` is no longer just `x`. This then repeats and might make the third constraint even larger. 

So even though you start small, once you get to the nth small formula, it might not be small anymore. It depends on the running substitution for the shared variable (among other things, when considering the whole constraint system).

Alex1005a:
The equations are definitely increasing. But as far as I understand there are several factors why this PR improves performance by split formulas.
1. A lot of formulas of the form `x ~ {}` are created. As far as I understand, this means that we cannot get a solution in which the number of free variables is greater than in `x`.
2. In small formulas the number of free variables is less than in the overall large one (most likely not always, so we need to add a check before splitting the formula). If point 1 is true, then the number of free variables after substitution will not increase.

This improves the performance of small formulas in the `successiveVariableElimination` function, since fewer free variables help avoid exponential growth.

JonathanStarup:
The checking should very soon be redundant since we can (and should) change unifySubset from `t1 + a ~ t2` to `t1 - t2 ~ empty` with the new set solver (cant link the PR on phone, but its called EffUnification3 i think)

Alex1005a:
I added this check because the type check for the [sequences.flix](https://github.com/flix/flix/blob/master/examples/larger-examples/restrictable-variants/sequences.flix) file resulted in an `OutOfMemoryError` exception. If the new solver fixes this problem I will remove it.

Alex1005a:
@JonathanStarup It seems that this check is not redundant. Without it, the sequences.flix type check still ends in an error.

Alex1005a:
@magnus-madsen PR is ready to merge. Did you mark it as draft because branch is out-of-date with the base branch?

magnus-madsen:
> @magnus-madsen PR is ready to merge. Did you mark it as draft because branch is out-of-date with the base branch?

I just marked it as draft because I wanted to wait for the new solver to land. Once that happens I want to benchmark this.

magnus-madsen:
We have introduced a new Boolean unification solver. We should port the restrictable variant codebase to use that. Once that is done, then we can resurrect this experiment to see if it is worth it.

PR #8754, opened 2024-09-30
fix: introduce syntactic context for `OpCompleter`

magnus-madsen:


magnus-madsen:
@mlutze Also this one

mlutze (approved):

PR #8760, opened 2024-09-30
feat: add `ApplySig`

jaschdoc:
Related to https://github.com/flix/flix/issues/8399


jaschdoc:
@JonathanStarup do we want to return an `ApplyDef` node in `Monomorpher` or do we want to carry `ApplySig` further into the backend? Would it give us anything?

```scala
    case LoweredAst.Expr.ApplySig(sym, exps, itpe, tpe, eff, loc) =>
      val it = subst(itpe)
      val newSym = specializeSigSym(sym, it)
      MonoAst.Expr.ApplyDef(newSym, es, it, subst(tpe), subst(eff), loc)
```

 The current `Apply` node already has this behavior (which makes sense since the a `Sig` is specialized to a `Def` in `Monomorpher`):

```scala
    case LoweredAst.Expr.Apply(exp, exps, tpe, eff, loc) =>
      val e = visitExp(exp, env0, subst)
      e match {
        case MonoAst.Expr.Sig(sym, itpe, _) => MonoAst.Expr.ApplyDef(sym, es, itpe, subst(tpe), subst(eff), loc)
        case _ => MonoAst.Expr.Apply(e, es, subst(tpe), subst(eff), loc)
      }
```

jaschdoc:
!!!rerun

magnus-madsen:
We want to return `ApplyDef` is Monomorpher. The reason is that monomorphization eliminates type classes (traits) via translation.

magnus-madsen:
@jaschdoc It seems some of the IDE tests are actually broken. Possibly some source locations shifted. Either this PR needs some tweaks or the tests will need tweaks.

main/src/ca/uwaterloo/flix/api/lsp/Indexer.scala
```diff
@@ -240,6 +240,10 @@ object Indexer {
       val parent = Entity.Exp(exp0)
       visitExps(exps) ++ Index.occurrenceOf(exp0) ++ Index.useOf(sym, loc, parent)
 
+    case Expr.ApplySig(Ast.SigSymUse(sym, loc1), exps, _, _, _, loc2) =>
+      val parent = Entity.Exp(exp0)
+      visitExps(exps) ++ Index.occurrenceOf(exp0) ++ Index.useOf(sym, loc1, parent) ++ Index.useOf(sym.trt, loc1)
```
magnus-madsen, line 245: I am guessing this line is the one which needs to be scrutinized.

magnus-madsen:
Anything non-standard here?

jaschdoc:
> Anything non-standard here?

Nope, just fixed the location issue. I had used the wrong location in `SigSymUse` to begin with, so just a small fix

magnus-madsen (approved):

PR #8776, opened 2024-10-02
chore: remove `Time` base effect

magnus-madsen:


magnus-madsen:
Its definition is too nebulous. Moreover, the Java SecurityManager has no notion of time.

mlutze (approved):

PR #8779, opened 2024-10-03
feat: RecordConstraintSolver

mlutze:
This PR introduces a new unification function for records. This is important for the new solver:
- it uses the resilient strategy of giving back constraints instead of creating an error
- it does not call `Unification.xyz`

I've also added some type rule comments to annotate each match rule. Since the Scoped Labels paper does not use separated constraint generation and solving, I had to make up my own syntax.


magnus-madsen (approved):

magnus-madsen:
Very nice

PR #8784, opened 2024-10-04
refactor: `Simplifier.visitType`

JonathanStarup:
Instead of converting args up front and the destructing the list. I now destruct the list first and then convert. This achieves:
- More checking of expected list lengths
- allows adding array regions and arrow effects to monotypes in follow up PR


JonathanStarup:
@magnus-madsen this allows me to add the array regions and arrow effects to monotype which in turn means that the monomorpher PR can be well-kinded and also means that the verifier can check more

magnus-madsen (approved):

PR #8785, opened 2024-10-04
feat: `SetFormula.successiveVariableElimination`

JonathanStarup:


main/src/ca/uwaterloo/flix/language/phase/unification/set/SetSubstitution.scala
```diff
@@ -143,6 +143,14 @@ case class SetSubstitution(m: Map[Int, SetFormula]) {
     }
   }
 
+  /**
+    * Adds the disjoint binding `[x -> f]` to `this`.
+    *
+    * OBS: The [[SetFormula.Var]] of `this`, `x`, and `f` must not overlap.
```
magnus-madsen, line 149: This would be costly to check, right?

magnus-madsen: Like with an assertion.

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetFormula.scala
```diff
@@ -872,4 +872,146 @@ object SetFormula {
...
 
+  /**
+    * The Successive Variable Elimination algorithm.
+    *
+    * Returns the most-general unifier of the equation `f ~ empty` where `fvs` is the free
+    * variables in `f`. If there is no unifier then [[NoSolutionException]] is thrown.
+    *
+    * Eliminates variables one-by-one from the given list `fvs` recursively.
+    *
+    * If the recursively building formula is ever larger than `recSizeThreshold` then
```
magnus-madsen, line 883: Grammar. At least it reads strange to me.

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetFormula.scala
```diff
@@ -872,4 +872,146 @@ object SetFormula {
...
+    * [[ComplexException]] is thrown. If `recSizeThreshold` is non-positive then there is no
```
magnus-madsen, line 884: Why do it like this? I would just prefer people can use Int.MaxValue or whatever

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetFormula.scala
```diff
@@ -872,4 +872,146 @@ object SetFormula {
...
+    * checking.
+    */
+  def successiveVariableElimination(f: SetFormula, fvs: List[Int], recSizeThreshold: Int): SetSubstitution = fvs match {
+    case Nil =>
+      // `emptyEquivalent` now treats the remaining constants as variables and goes through all
```
magnus-madsen, line 889: Can you rephrase. Its hard to understand what should happen here. (I know what should happen, but re-read your sentence).

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetFormula.scala
```diff
@@ -872,4 +872,146 @@ object SetFormula {
...
+      // assignments of those.
+      if (emptyEquivalent(f)) SetSubstitution.empty
```
magnus-madsen, line 891: why not call it `isEmpty?` or `isEmptyEquiv`?  or similar.

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetFormula.scala
```diff
@@ -872,4 +872,146 @@ object SetFormula {
...
+      else throw NoSolutionException()
+
+    case x :: xs =>
+      val fEmpty = SetSubstitution.singleton(x, Empty)(f)
+      val fUniv = SetSubstitution.singleton(x, Univ)(f)
```
magnus-madsen, line 896: I think it would really be better to keep the short names used elsewhere. They match those used in the papers; so its possible to lookup and compare.

magnus-madsen (commented):
Mostly complaints about documentation :)

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetSubstitution.scala
JonathanStarup: yeah I didn't think it was worth it. And technically the mapping `x -> x` is the same as not having `x` so we should also account for that in the check

JonathanStarup: But we could constrain it to its less general form, and check that all mappings x -> f have f.isground and that the keys are disjoint. Also expensive but less cumbersome and its enough for the sve use

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetFormula.scala
JonathanStarup: that would mean that we always call `f.size` which is expensive. But you can still call it with `MaxValue` if you want to.

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetFormula.scala
JonathanStarup: hmm, i dont agree. but not enough to insist. ill change it back

magnus-madsen: This I insist on. If I ever take it to Jaco he will not be able to read it :)

magnus-madsen (approved):

PR #8786, opened 2024-10-04
feat: `MonoType.Arrow.purity` and `MonoType.Array.purity`

JonathanStarup:


main/src/ca/uwaterloo/flix/language/ast/MonoType.scala
```diff
@@ -69,7 +69,7 @@ object MonoType {
   // Compound Types.
   //
 
-  case class Array(tpe: MonoType) extends MonoType
+  case class Array(tpe: MonoType, purity: Purity) extends MonoType
```
magnus-madsen, line 72: What????

JonathanStarup: Looks crazy but its really just exposing how it is (we monomorphize regions into IO)

JonathanStarup:
I think this is a good enough plan long term, but I can progress with monomorpher by creating TypeConstructor.SimpleArray and TypeConstructor.SimpleArrow instead?

magnus-madsen:
@JonathanStarup I see some community build failures.

magnus-madsen:
What's the plan? This first and then cleanup or in the opposite order?

JonathanStarup:
there is actual errors right now because the verifier has more information

we could do that yes, but we can also go another way with monomorph. you want this approach (arrow and array purity) or the other (TC.SimpleArray and TC.SimpleArrow)?

magnus-madsen:
What are the two approaches? I would like the one with well-kinded types and new type constructors. What's the other approach?

magnus-madsen:
@JonathanStarup What about this? What is its relevance to the current set effect work?

JonathanStarup:
Too annoying to add for now with the backend dependence on monotype instead of backendType

PR #8787, opened 2024-10-04
feat: introduce SchemaConstraintSolver

mlutze:
The same as the last PR. I will follow up to change everywhere it says "record" to "row", but I will leave the type rules as they are.

I've bumped up the solver iteration limit since some schema cases were struggling.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/typer/SchemaConstraintSolver.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/typer/SchemaConstraintSolver.scala
@@ -1,0 +1,122 @@
+/*
+ * Copyright 2024 Matthew Lutze
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ca.uwaterloo.flix.language.phase.typer
+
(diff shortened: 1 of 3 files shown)


main/src/ca/uwaterloo/flix/language/phase/typer/SchemaConstraintSolver.scala
```diff
@@ -0,0 +1,122 @@
...
+import ca.uwaterloo.flix.api.Flix
+import ca.uwaterloo.flix.language.ast.shared.Scope
+import ca.uwaterloo.flix.language.ast.{Kind, Name, RigidityEnv, Symbol, Type, TypeConstructor}
+import ca.uwaterloo.flix.language.phase.typer.ConstraintSolver.ResolutionResult
+import ca.uwaterloo.flix.language.phase.typer.TypeConstraint.{Equality, Provenance}
+import ca.uwaterloo.flix.language.phase.unification.Substitution
+import ca.uwaterloo.flix.util.InternalCompilerException
+
+//
+// This is a copy of the SchemaConstraintSolver. Because the shapes are slightly different, we cannot reuse the same solver.
```
magnus-madsen, line 27: Copy of what now? :)

magnus-madsen:
Increasing the iteration limit seems dubious-- but lets see what iteration limit is actually needed for the real solver.

magnus-madsen:
Hmm. You iteration threshold is not high enough... Hmm

JonathanStarup:
I think we should have a flag that runs prints the thresholds needed for each function. that would be a good debugging tool

mlutze:
Yeah, I'll adjust the recursive case to do a little more work, which should reduce costs.

magnus-madsen:
> Yeah, I'll adjust the recursive case to do a little more work, which should reduce costs.

Will you do the same for records or no?

mlutze:
> > Yeah, I'll adjust the recursive case to do a little more work, which should reduce costs.
> 
> Will you do the same for records or no?

Yep, I want to keep them exactly 1:1

magnus-madsen:
Apparently it is still unhappy.

PR #8788, opened 2024-10-04
feat: Set Unification Rules

JonathanStarup:
@magnus-madsen before you read closely, what do you think about:

- Equation now has a status, so you know whether its been judged unsolvable or have caused a timeout
- The signature of the rules

I think this is quite nice and handles both progress and conflicts with minimal fuzz


main/src/ca/uwaterloo/flix/language/phase/unification/set/Equation.scala
```diff
@@ -17,31 +17,47 @@
...
   */
 @nowarn
-case class Equation private(f1: SetFormula, f2: SetFormula, loc: SourceLocation) {
+case class Equation private(f1: SetFormula, f2: SetFormula, status: Equation.Status, loc: SourceLocation) {
 
   /** Returns the sum of the sizes of the formulas in `this`. */
   final def size: Int = f1.size + f2.size
 
   /** Returns a human-readable string of `this`. */
   override final def toString: String = s"$f1 ~ $f2"
```
magnus-madsen, line 41: I can see arguments both for and against including the status here.

main/src/ca/uwaterloo/flix/language/phase/unification/set/Equation.scala
```diff
@@ -17,31 +17,47 @@
...
 
+  /** Returns a copy of `this` with the new `status` */
+  final def copyWithStatus(status: Equation.Status): Equation = {
+    if (status == this.status) this
+    else Equation(f1, f2, status, loc)
+  }
+
+  /** Returns a copy of `this` with status [[Equation.Status.Unsolvable]]. */
+  @inline
```
magnus-madsen, line 50: I don't think these inline annotations should matter. The JVM is pretty much guaranteed to inline here.

main/src/ca/uwaterloo/flix/language/phase/unification/set/Equation.scala
```diff
@@ -55,13 +71,28 @@ object Equation {
...
+  object Status {
+
+    /** Equation might be solvable or unsolvable. */
+    case object Unknown extends Status
+
+    /** Equation is unsolvable. */
+    case object Unsolvable extends Status
+
+    /** The equation is unsolvable due to a timeout or a size threshold. */
+    case class Timeout(msg: String) extends Status
```
magnus-madsen, line 94: I like the overall idea 👍

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetFormula.scala
```diff
@@ -56,6 +56,20 @@ sealed trait SetFormula {
       SortedSet.from(varsPos.map(_.x)) ++ varsNeg.map(_.x) ++ other.flatMap(_.variables)
   }
 
+  /** Faster alternative to `this.variables.contains(v)`. */
```
magnus-madsen, line 59: Faster in practice?

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
```diff
@@ -0,0 +1,235 @@
...
+      *
+      *   - `univ ~ univ` becomes `{}`
+      *   - `univ ~ empty` becomes `{univ !~ empty}`
+      *   - `x1 ~ x1` becomes `{}`
+      *   - `x2 ~ univ` becomes `{x2 ~ univ}`
+      */
+    def trivial(eq: Equation)(implicit p: Progress): (List[Equation], SetSubstitution) = {
+      val Equation(t1, t2, _, _) = eq
+
+      @inline
```
magnus-madsen, line 54: Note that: Inlining these probably does little for performance. However, it does increase the size of `trivial` so that it cannot be inlined.

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
```diff
@@ -0,0 +1,235 @@
...
+      }
+    }
+
+    /**
+      * Solves equations of ground assignments to variables (e.g. `x ~ c1 ∪ e2`).
+      *
+      *   - `x ~ t` where [[SetFormula.isGround]] on `t` becomes `({}, [x -> t])`
+      *   - `!x ~ t` where [[SetFormula.isGround]] on `t` becomes `({}, [x -> !t])`
+      *   - `t1 ∩ t2 ∩ .. ~ univ` becomes `({t1 ~ univ, t2 ~ univ, ..}, [])`
+      *   - `t1 ∪ t2 ∪ .. ~ univ` becomes `({t1 ~ empty, t2 ~ empty, ..}, [])`
```
magnus-madsen, line 99: This seems wrong. Typo?

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
```diff
@@ -0,0 +1,235 @@
...
+    def constantAssignment(eq: Equation)(implicit p: Progress): (List[Equation], SetSubstitution) = {
+      val Equation(f1, f2, _, loc) = eq
+      (f1, f2) match {
+        // x ~ t, where t is ground
+        // ---
+        // {},
+        // [x -> t]
+        case (Var(x), f) if f.isGround =>
+          p.progress()
+          (Nil, SetSubstitution.singleton(x, f))
```
magnus-madsen, line 110: I _much_ prefer this style compared to the code I saw last 👍

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    /**
+      * Solves equations of ground assignments to variables (e.g. `x ~ c1 ∪ e2`).
+      *
+      *   - `x ~ t` where [[SetFormula.isGround]] on `t` becomes `({}, [x -> t])`
+      *   - `!x ~ t` where [[SetFormula.isGround]] on `t` becomes `({}, [x -> !t])`
+      *   - `t1 ∩ t2 ∩ .. ~ univ` becomes `({t1 ~ univ, t2 ~ univ, ..}, [])`
+      *   - `t1 ∪ t2 ∪ .. ~ univ` becomes `({t1 ~ empty, t2 ~ empty, ..}, [])`
+      */
```
magnus-madsen, line 100: Do we need some comment about symmetric cases being absent because we assume something of Eq.mk?

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
```diff
@@ -0,0 +1,235 @@
...
+package ca.uwaterloo.flix.language.phase.unification.set
+
+import SetFormula.*
+
+object SetUnification {
+
+  /** A class to track progress during set unification. */
+  class Progress private(var somethingHappened: Boolean = false) {
+    /** Mark that progress has been made. */
+    def progress(): Unit = somethingHappened = true
```
magnus-madsen, line 26: -> markProgress?

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
```diff
@@ -0,0 +1,235 @@
...
+          else (List(eq.toUnsolvable), SetSubstitution.empty)
+
+        case _ =>
+          // Can't do anything.
+          (List(eq), SetSubstitution.empty)
+      }
+    }
+
+    /**
+      * Solves equations variable alias equations (e.g. `x1 ~ x2`).
```
magnus-madsen, line 154: Grammar

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
```diff
@@ -0,0 +1,235 @@
...
+      *   - `!x1 ~ !x2` becomes `({}, [x1 -> x2])`
+      */
+    def variableAlias(eq: Equation)(implicit p: Progress): (List[Equation], SetSubstitution) = {
+      val Equation(t1, t2, _, _) = eq
+      (t1, t2) match {
+        // x1 ~ x2
+        // ---
+        // {},
+        // [x1 -> x2]
+        case (Var(x), y@Var(_)) =>
```
magnus-madsen, line 166: What about literal x -> x? Is that gone? Can it happen here?

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
```diff
@@ -0,0 +1,235 @@
...
+      }
+    }
+
+    /**
+      * Solves equations using successive-variable-elimination, i.e. exhaustive instantiation.
+      *
+      * Always returns no equations or `eq` marked as [[Equation.Status.Unsolvable]] or
+      * [[Equation.Status.Timeout]].
+      */
+    def sve(recSizeThreshold: Int)(eq: Equation)(implicit p: Progress): (List[Equation], SetSubstitution) = {
```
magnus-madsen, line 221: Any reason recSizeThreshold is an argument? Could it not be a static final field, so to speak?

Alternatively, it could be a property on `Progress` which could potentially be renamed to `Solver`?

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
```diff
@@ -0,0 +1,235 @@
...
+      p.progress()
+      val query = mkEquivalenceTestToEmpty(eq.f1, eq.f2)
```
magnus-madsen, line 223: mkEquivalenceTestToEmpty - I find this name hard to understand.

magnus-madsen (commented):
I much prefer this refactor 👍

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
```diff
@@ -0,0 +1,235 @@
...
+    def hasProgressed: Boolean = somethingHappened
+  }
+
+  object Progress {
+    /** Make a new [[Progress]] object with progress set to `false`. */
+    def apply(): Progress = new Progress()
+  }
+
+  /** A collection of rules that can be used to solve individual [[Equation]]. */
+  private object Rules {
```
magnus-madsen, line 38: Do we have to nest these functions? Let us unnest them please, thanks.

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
```diff
@@ -0,0 +1,235 @@
...
+  /** A class to track progress during set unification. */
+  class Progress private(var somethingHappened: Boolean = false) {
+    /** Mark that progress has been made. */
+    def progress(): Unit = somethingHappened = true
+
+    /** Returns `true` if there has been progress. */
+    def hasProgressed: Boolean = somethingHappened
+  }
+
+  object Progress {
```
magnus-madsen, line 32: companion object before class, by convention.

magnus-madsen: Also: Maybe its better to be explicit where the progress object is created? I would do that and drop the apply method entirely.

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
JonathanStarup: yea i found it hard to name - I could just call it `mkXor` but that also doesn't why that is what we want.

`mkEmptyTester` ?

magnus-madsen: Oh no, not mkEmptyTester

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
JonathanStarup: To have it be changable - i was annoyed by the static thresholds of the old solver

Hmm we could make `Progress` the "settings" object but I want to make the rules as selfcontained and simple as possible - removed from the other infrastructure. So I think direct arguments and simple objects achieve that better than syntactic brevity

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
JonathanStarup: if you mean `x42 ~ x42` then that is taken care of by `trivial` but we could also add it here yes

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
magnus-madsen: Changeable by who?

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
JonathanStarup: I mean `def progress` as the verb "progress" but maybe markProgress is clearer yeah

magnus-madsen:
How about you add a private lass:

```
private lass SolverContext(var progress: Boolean, threshold: Int)
```

and then pass it implicitly everywhere needed.

I think that would be nice. Then we both get what we want :)

main/src/ca/uwaterloo/flix/language/phase/unification/set/Equation.scala
JonathanStarup: I think that `f1 !~ f2` would be read as an inequality constraint to new people. We could maybe have `f1 ~(error) f2` and `f1 ~(timeout) f2` to not hide the information? or `f1 ~err f2` and `f1 ~to f2`

main/src/ca/uwaterloo/flix/language/phase/unification/set/Equation.scala
JonathanStarup: It is usefull though - I did not put `final` and the optimizer warned me that this could be overridden so It couldn't inline.

I'd guess it'd be jitted fine afterwards, but it shows that writing your intent is useful.

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetFormula.scala
JonathanStarup: Ignoring compiler magic, yes. You 1) don't have to build the variable set and 2) you can early exit

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
JonathanStarup: > [..] so that it cannot be inlined.

I mean who knows the size threshold? might also not be inlined anyway and the straightline, inlined program is faster?

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
JonathanStarup: hmm yeah. And the `!x ~ f` case need to be mirrored. I've though about not having `Equation` ordering invariants. It makes the code hard to read because you have to think about them. Wouldn't it be nicer just to have the mirrored cases? fx `trivial` right now has all cases, disregarding the `Equation` invariants and that seems easier to read to deleting some of lines because of invariants in another file

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
magnus-madsen: There is a book on Jvm Internals that you can read :) I read it, but cannot remember the name. Just saying, dont go by folklore. At least you should bench it.

magnus-madsen: Maybe its this one: https://www.amazon.com/JVM-Performance-Engineering-OpenJDK-Developers-ebook/dp/B08F5J6V4T but I am not sure

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
magnus-madsen: OK, I leave it to you.

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
JonathanStarup: I've tried to rewrite `constantAssignment` to be selfcontained without duplicate cases. Then we could remove `Eq.mk`. What do you think?

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
JonathanStarup: without static mutation, by anyone. If we want it as a compiler flag, need to disable it for testing, etc

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
magnus-madsen: We could do both:

We still order (for our own sake in debugging)
But the code handles mirrored cases.

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
JonathanStarup: I mean right now it looks goofy, but it think its nice to put them together. They will also probably have a corresponding phase called the same thing, so then it'd be `variableAssignmentPhase` and `variableAssignmentRule` or something like that

magnus-madsen:
> Isn't that just what you want? not a lot of compromise in that compromise 😄

I wanted you to remove the threshold or hardcode it. Now you get to keep it as a parameter you can scoop from some Xproperty.

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
JonathanStarup: mkEmptyQuery?

magnus-madsen: Yes

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
JonathanStarup: right, my intention was not to performance game, my intention was just to reduce clutter for the reader without changing its actual behaviour/performance. I just wanted "syntactic functions" if that makes sense. That's why I felt that `@inline` was the neutral choice. But I can remove it

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
JonathanStarup: is ordering good for debugging? to have small left sides?

sounds good

magnus-madsen: Its good to have variables on the left, yes

main/src/ca/uwaterloo/flix/language/phase/unification/set/Equation.scala
```diff
@@ -17,51 +17,74 @@
...
+case class Equation private(f1: SetFormula, f2: SetFormula, status: Equation.Status, loc: SourceLocation) {
 
   /** Returns the sum of the sizes of the formulas in `this`. */
   final def size: Int = f1.size + f2.size
 
   /** Returns a human-readable string of `this`. */
-  override final def toString: String = s"$f1 ~ $f2"
+  override final def toString: String = status match {
+    case Status.Unknown => s"$f1 ~ $f2"
+    case Status.Unsolvable => s"$f1 ~error $f2"
```
magnus-madsen, line 44: error -> unsolvable

main/src/ca/uwaterloo/flix/language/phase/unification/set/Equation.scala
```diff
@@ -17,51 +17,74 @@
...
-case class Equation private(f1: SetFormula, f2: SetFormula, loc: SourceLocation) {
+case class Equation private(f1: SetFormula, f2: SetFormula, status: Equation.Status, loc: SourceLocation) {
 
   /** Returns the sum of the sizes of the formulas in `this`. */
   final def size: Int = f1.size + f2.size
 
   /** Returns a human-readable string of `this`. */
-  override final def toString: String = s"$f1 ~ $f2"
+  override final def toString: String = status match {
+    case Status.Unknown => s"$f1 ~ $f2"
```
magnus-madsen, line 43: Unknown -> Pending

main/src/ca/uwaterloo/flix/language/phase/unification/set/Equation.scala
```diff
@@ -17,51 +17,74 @@
...
+    if (status == this.status) this
+    else Equation(f1, f2, status, loc)
+  }
+
+  /** Returns a copy of `this` with status [[Equation.Status.Unsolvable]]. */
+  @inline
+  final def toUnsolvable: Equation = copyWithStatus(Equation.Status.Unsolvable)
+
+  /** Returns a copy of `this` with status [[Equation.Status.Timeout]]. */
+  @inline
```
magnus-madsen, line 59: We agree to remove inline.

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
```diff
@@ -0,0 +1,279 @@
...
+    }
+  }
+
+  /**
+    * Solves equations using successive-variable-elimination, i.e. exhaustive instantiation.
+    *
+    * Always returns no equations or `eq` marked as [[Equation.Status.Unsolvable]] or
+    * [[Equation.Status.Timeout]].
+    */
+  def sve(recSizeThreshold: Int)(eq: Equation)(implicit p: Progress): (List[Equation], SetSubstitution) = {
```
magnus-madsen, line 266: We add two implicit parameters: Progress, SolverOptions

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
```diff
@@ -0,0 +1,279 @@
...
+    p.markProgress()
+    val query = mkEmptyQuery(eq.f1, eq.f2)
+    val fvs = query.variables.toList
+    try {
+      val subst = successiveVariableElimination(query, fvs, recSizeThreshold)
```
magnus-madsen, line 271: Move SVE after this function

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
```diff
@@ -0,0 +1,279 @@
...
+    *
+    * This also applies to the symmetric equations.
+    */
+  def constantAssignment(eq: Equation)(implicit p: Progress): (List[Equation], SetSubstitution) = {
+    val Equation(f01, f02, _, loc) = eq
+
+    // Reorder formulas to avoid symmetric matches below.
+    //   - `x` and `!x` to the left
+    //   - `empty` and `univ` to the right
+    val (f1, f2) = (f01, f02) match {
```
magnus-madsen, line 104: Remove and duplicate

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
```diff
@@ -0,0 +1,279 @@
...
+    *   - `!x ~ f` where `f` does not contain `x` becomes `({}, [x -> !f])`
+    *
+    * This also applies to the symmetric equations.
+    */
+  def variableAssignment(eq: Equation)(implicit p: Progress): (List[Equation], SetSubstitution) = {
+    val Equation(t01, t02, _, _) = eq
+
+    // Reorder formulas to avoid symmetric matches below.
+    //   - `x` and `!x` to the left
+    val (t1, t2) = (t01, t02) match {
```
magnus-madsen, line 231: Kill it

JonathanStarup:
@magnus-madsen ready for a new round

magnus-madsen (approved):

magnus-madsen:
OK, on to the next 🚀

PR #8809, opened 2024-10-07
feat: return `SoftFailure` in `visitSelectRule` in `Weeder`

jaschdoc:


main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1908,9 +1908,10 @@ object Weeder2 {
       flatMapN(pickNameIdent(tree), exprs) {
         case (ident, channel :: body :: Nil) =>
           Validation.success(SelectChannelRule(ident, channel, body))
-        case _ =>
+        case _ => // Unreachable case, since this is a parser error
           val err = Malformed(NamedTokenSet.MatchRule, SyntacticContext.Expr.OtherExpr, loc = tree.loc)
-          Validation.HardFailure(Chain(err))
+          val ident = Name.Ident("error", tree.loc.asSynthetic)
+          Validation.toSoftFailure(SelectChannelRule(ident, Expr.Error(err), Expr.Error(err)), err)
```
jaschdoc, line 1914: we could also throw an ICE but my hope is that if we decide to change parsing at some point then this will catch it

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1908,9 +1908,10 @@ object Weeder2 {
       flatMapN(pickNameIdent(tree), exprs) {
         case (ident, channel :: body :: Nil) =>
           Validation.success(SelectChannelRule(ident, channel, body))
-        case _ =>
+        case _ => // Unreachable case, since this is a parser error
```
magnus-madsen, line 1911: Are you sure the case is unreachable? I don't think it is... 

I would guess if you write a partial select rule it is reachable.

If it was truely unreachable then we should just throw ICE

jaschdoc: I added tests for that but they all failed with parser failures

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
magnus-madsen: Ditto for comment about ICE

jaschdoc:
I have manually tried messing up the select-rule in vscode with this version and was unable to do that with
```
def main(): Unit \ IO =
    select {
        case wad <- Cwf.awkf(???) => ???
    }
```

However, it makes it all the way to Monomorpher which throws an ICE

jaschdoc:
I think Resolver might need to handle this or at least we have to check somewhere that this is a call to `Channel.recv`

magnus-madsen:
> I have manually tried messing up the select-rule in vscode with this version and was unable to do that with
> 
> ```
> def main(): Unit \ IO =
>     select {
>         case wad <- Cwf.awkf(???) => ???
>     }
> ```
> 
> However, it makes it all the way to Monomorpher which throws an ICE

I think it should be with an incomplete rule, right? So something where some of the `case ...` is missing?

jaschdoc:
Ok I think this requires at bit more investigation... I will come back to this

magnus-madsen (approved):

PR #8820, opened 2024-10-08
fix: perform kind inference on types in typematch

mlutze:


mlutze:
@magnus-madsen Are we ok with disallowing typematch with wildcard heads?

```
    def hasListSomewhere(x: a): Bool = typematch x {
        case _: List[_] => true
        case _: _[List] => true
        case _: _[List[_]] => true
        case _: _[List, _] => true
        case _: _[List[_], _] => true
        case _: _ => false
    }
```

magnus-madsen:
> @magnus-madsen Are we ok with disallowing typematch with wildcard heads?
> 
> ```
>     def hasListSomewhere(x: a): Bool = typematch x {
>         case _: List[_] => true
>         case _: _[List] => true
>         case _: _[List[_]] => true
>         case _: _[List, _] => true
>         case _: _[List[_], _] => true
>         case _: _ => false
>     }
> ```

I suppose its OK

magnus-madsen (approved):

PR #8824, opened 2024-10-08
feat: use `SoftFailure` in `pick` in `Weeder2`

jaschdoc:


jaschdoc:
@magnus-madsen what do you think of this as a solution (modulo refactoring and rough edges)?

jaschdoc:
It *seems* to work

magnus-madsen:
> @magnus-madsen what do you think of this as a solution (modulo refactoring and rough edges)?

Can you explain it in words? Or even better in comments :)

magnus-madsen:
Is the solution to "return random garbage" ??? :)

jaschdoc:
Yes basically... or well something obviously bad that should fail / crash later on

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -3064,17 +3064,17 @@ object Weeder2 {
...
-        assert(idents.nonEmpty) // We require atleast one element to construct a qname
+      case idents if idents.nonEmpty => // We require at least one element to construct a qname
         val first = idents.head
         val ident = idents.last
         val nnameIdents = idents.dropRight(1)
         val loc = SourceLocation(isReal = true, first.loc.sp1, ident.loc.sp2)
         val nname = Name.NName(nnameIdents, loc)
         Name.QName(nname, ident, loc)
+      case nothing => // Return None instead?
+        Name.mkQName("error", SourceLocation.Unknown)
```
magnus-madsen, line 3077: This is a big problem. Because you are pretending that everything is well and there is a name. Just that it has the text "error". That, unfortunately, won't work.

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -3123,7 +3123,10 @@ object Weeder2 {
         val name = text(tree).mkString("")
         val error = MisplacedComments(SyntacticContext.Unknown, t.loc)
         Validation.toSoftFailure(Name.Ident(name, tree.loc), error)
-      case _ => throw InternalCompilerException(s"Parse failure: expected first child of '${tree.kind}' to be Child.Token", tree.loc)
```
magnus-madsen, line 3126: Does it mean that this exception can happen?

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -3207,7 +3210,10 @@ object Weeder2 {
       case Some(t) => Validation.success(t)
       case None =>
         val error = NeedAtleastOne(NamedTokenSet.FromTreeKinds(Set(kind)), sctx, loc = tree.loc)
```
magnus-madsen, line 3212: Is need  NeedAtleastOne not an error? Should we not preserve it?

magnus-madsen: My bad, you do preserve it.

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
magnus-madsen: It is not so clear how to fix it though. It may require a Name.Error which would require _a ton_ of refactoring... So let us not do that just yet.

jaschdoc: What if we return `None` instead in this case and below? Then we do not introduce bad values but we also signal that an error occurred, so we can then collect the error and figure out what to do at the call site

magnus-madsen: Can this actually happen? The old code has some asserts?

jaschdoc: yes I removed the `expect` call to prevent it from crashing when this case was hit (to test out my implementation)

magnus-madsen: If you return an Option, I suppose you still have to do a lot of refactoring? How many call sites would break?

jaschdoc:
I tried returning `Option` and it seems to work fine. There are only one or two places where special handling seems to be required. Other than that the call to `expect` at the top of most functions ensures that the syntax tree is ok (I think?)

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1699,7 +1703,7 @@ object Weeder2 {
     private def visitWithoutExpr(tree: Tree): Validation[Expr, CompilationMessage] = {
       expect(tree, TreeKind.Expr.Without)
       val effectSet = pick(TreeKind.Type.EffectSet, tree)
-      val effects = flatMapN(effectSet)(effectSetTree => traverse(pickAll(TreeKind.QName, effectSetTree))(visitQName))
+      val effects = flatMapN(effectSet)(effectSetTree => mapN(traverse(pickAll(TreeKind.QName, effectSetTree))(visitQName))(_.flatten))
```
jaschdoc, line 1706: should probably be handled better here

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1244,7 +1248,7 @@ object Weeder2 {
             case "<+>" => Validation.success(Expr.FixpointMerge(e1, e2, tree.loc))
             case "instanceof" =>
               flatMapN(pickQName(exprs(1))) {
-                case qname if qname.isUnqualified => Validation.success(Expr.InstanceOf(e1, qname.ident, tree.loc))
+                case Some(qname) if qname.isUnqualified => Validation.success(Expr.InstanceOf(e1, qname.ident, tree.loc))
```
jaschdoc, line 1251: this falls through to the error case below

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -2696,15 +2704,15 @@ object Weeder2 {
...
               }
 
             case (tree, acc) if tree.kind == TreeKind.Type.PredicateWithTypes =>
               val types = pickAll(TreeKind.Type.Type, tree)
               val maybeLatTerm = tryPick(TreeKind.Predicate.LatticeTerm, tree)
               mapN(pickQName(tree), traverse(types)(Types.visitType), traverseOpt(maybeLatTerm)(Types.pickType)) {
-                case (qname, tps, None) => Type.SchemaRowExtendByTypes(qname.ident, Denotation.Relational, tps, acc, tree.loc)
-                case (qname, tps, Some(t)) => Type.SchemaRowExtendByTypes(qname.ident, Denotation.Latticenal, tps :+ t, acc, tree.loc)
+                case (qname, tps, None) => Type.SchemaRowExtendByTypes(qname.get.ident, Denotation.Relational, tps, acc, tree.loc)
+                case (qname, tps, Some(t)) => Type.SchemaRowExtendByTypes(qname.get.ident, Denotation.Latticenal, tps :+ t, acc, tree.loc)
```
jaschdoc, line 2715: not sure this is entirely safe either

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -2854,8 +2870,7 @@ object Weeder2 {
       val derivations = maybeDerivations
         .map(tree => traverse(pickAll(TreeKind.QName, tree))(visitQName))
         .getOrElse(Validation.success(List.empty))
-
-      mapN(derivations)(Derivations(_, loc))
+      mapN(derivations)(l => Derivations(l.filter(_.isDefined).map(_.get), loc)) // should maybe check that does not contain none
```
jaschdoc, line 2873: also check that length of `l` did not shrink in size

magnus-madsen:
I think we have to brainstorm this at one of the meetings.

PR #8833, opened 2024-10-09
chore: add benchmarks

jaschdoc:
Adds `filterMap` benchmark

Also refactors the `InlinerMap10k` benchmark: the call to `rev` should be outside the local def in `map`


main/src/resources/benchmark/Benchmark.Inliner.Optimized.flix
```diff
@@ -21,6 +21,15 @@ mod Benchmark.Inliner.Optimized {
         case z :: zs => mp(zs, z + 1 :: acc)
     }
 
+    pub def fmp(l: List[Int32], acc: List[Int32]): List[Int32] = match l {
```
magnus-madsen, line 24: Not really sure why these functions could not have been given proper names? :)

jaschdoc: I thought it was easier to see that it was the (manually) lifted local defs

magnus-madsen: Alright

PR #8834, opened 2024-10-09
refactor: rename `Optimizer` and friends

jaschdoc:
Rename to `Optimizer2`, `OccurrenceAnalyzer2`, `Inliner2`

This is to allow for the new `Optimizer` which runs before `ClosureConv`, following the same naming scheme as `TreeShaker`


magnus-madsen:
But, I thought the old Optimizer would be removed?

jaschdoc:
It will, I just wanted to rename it before adding the new one, which could then safely allow me to remove the old one

magnus-madsen:
Hmm. Just call the new one for 1 or 2? :)

magnus-madsen:
Then we don't break arewefast etc.

PR #8837, opened 2024-10-09
feat: visitor

LoZander:
this PR adds the visitor, which allows traversing the AST selectively, performing some action for each visited node depending on node type.

related to #8639

PR #8839, opened 2024-10-09
refactor: added source locations for `Def`, `Sig` and `Op`

LoZander:
added source locations for Def, Sig and Op AST nodes where these are missing. These already do exist for some ASTs, but are removed during the Naming phase `Namer.scala`.

Related to the work being done in #8837

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
@@ -217,7 +217,7 @@ object SemanticTokensProvider {
     * Returns all semantic tokens in the given definition `defn0`.
     */
   private def visitDef(defn0: TypedAst.Def): Iterator[SemanticToken] = defn0 match {
-    case Def(sym, spec, exp, loc) =>
+    case Def(sym, spec, exp, _) =>
       val t = SemanticToken(SemanticTokenType.Function, Nil, sym.loc)
       IteratorOps.all(
         Iterator(t),
@@ -230,7 +230,7 @@ object SemanticTokensProvider {
     * Returns all semantic tokens in the given signature `sig0`.
     */
   private def visitSig(sig0: TypedAst.Sig): Iterator[SemanticToken] = sig0 match {
-    case TypedAst.Sig(sym, spec, exp, loc) =>
+    case TypedAst.Sig(sym, spec, exp, _) =>
       val t = SemanticToken(SemanticTokenType.Function, Nil, sym.loc)
       IteratorOps.all(
         Iterator(t),
@@ -309,7 +309,7 @@ object SemanticTokensProvider {
     * Returns all semantic tokens in the given effect operation.
     */
   private def visitOp(op: TypedAst.Op): Iterator[SemanticToken] = op match {
-    case TypedAst.Op(sym, spec, loc) =>
+    case TypedAst.Op(sym, spec, _) =>
       val t = SemanticToken(SemanticTokenType.Function, Nil, sym.loc)
       IteratorOps.all(
         Iterator(t),
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
@@ -217,7 +217,7 @@ object SemanticTokensProvider {
     * Returns all semantic tokens in the given definition `defn0`.
     */
   private def visitDef(defn0: TypedAst.Def): Iterator[SemanticToken] = defn0 match {
-    case Def(sym, spec, exp) =>
+    case Def(sym, spec, exp, loc) =>
       val t = SemanticToken(SemanticTokenType.Function, Nil, sym.loc)
       IteratorOps.all(
         Iterator(t),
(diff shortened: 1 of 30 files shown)


LoZander:
While working on this, I have noticed multiple locations where the locations of the `sym` or `spec` of1 `Def`, `Sig` and `Op` are used where it seems like it ought to have been the location of the `Def`, `Sig` and `Op` themselves. Perhaps, these places should be investigated.

magnus-madsen:
@mlutze Can you review?

main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
```diff
@@ -217,7 +217,7 @@ object SemanticTokensProvider {
     * Returns all semantic tokens in the given definition `defn0`.
     */
   private def visitDef(defn0: TypedAst.Def): Iterator[SemanticToken] = defn0 match {
-    case Def(sym, spec, exp) =>
+    case Def(sym, spec, exp, loc) =>
```
mlutze, line 220: unused?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/HoleCompletion.scala
```diff
@@ -55,7 +55,7 @@ object HoleCompletion {
     )
 
     val matches = root.defs.values.flatMap {
-      case TypedAst.Def(sym, spec, _) =>
+      case TypedAst.Def(sym, spec, _, loc) =>
```
mlutze, line 58: unused?

mlutze (approved):

mlutze:
We can remove the loc from `spec` as a follow-up.

LoZander:
> We can remove the loc from `spec` as a follow-up.

I'm not sure I understand why we'd removed it from `spec`. For instance, shouldn't location for `Def` be different from that of it's `Spec`?

LoZander:
> > We can remove the loc from `spec` as a follow-up.
> 
> I'm not sure I understand why we'd removed it from `spec`. For instance, shouldn't location for `Def` be different from that of it's `Spec`?

I was under the impression that the `Spec` was only the signature. So that a the location of a `spec` for a `def` would not contain the location for the `exp` for instance. But I see now that during naming, the `loc` of the `def` is actually just "moved" to the spec. So the `spec` of a `def` will always have the same location as that of the `def`. I assume the same is true for `Op` and `Sig`. And in such a case, I suppose there is no reason to have loc in both the `Def`/`Sig`/`Op` and its spec

LoZander:
I guess my final question is whether or not the location for a `Def`/its `Spec` contains the location for the body exp?

magnus-madsen:
@LoZander Will you fix Matt's comments and remove loc from spec?

magnus-madsen:
> > We can remove the loc from `spec` as a follow-up.
> 
> I'm not sure I understand why we'd removed it from `spec`. For instance, shouldn't location for `Def` be different from that of it's `Spec`?

No. Its not. That's why we want to move it back to the Def/Sig/Op.

main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
```diff
@@ -230,7 +230,7 @@ object SemanticTokensProvider {
     * Returns all semantic tokens in the given signature `sig0`.
     */
   private def visitSig(sig0: TypedAst.Sig): Iterator[SemanticToken] = sig0 match {
-    case TypedAst.Sig(sym, spec, exp) =>
+    case TypedAst.Sig(sym, spec, exp, loc) =>
```
magnus-madsen, line 233: This probably also unused.

magnus-madsen: @LoZander If you use intellij idea it should warn you about unused variables.

mlutze: > @LoZander If you use intellij idea it should warn you about unused variables.

But don't bother fixing all unused variables that you did not introduce yourself. That's out of scope for this PR.

magnus-madsen: Yes, just focus on your code.

LoZander:
> @LoZander Will you fix Matt's comments and remove loc from spec?

Yep, sure. I was just waiting because I got the impression that the two of you were still mid-discussion on whether or not it should be done now or later :)

LoZander:
> > > We can remove the loc from `spec` as a follow-up.
> > 
> > 
> > I'm not sure I understand why we'd removed it from `spec`. For instance, shouldn't location for `Def` be different from that of it's `Spec`?
> 
> No. Its not. That's why we want to move it back to the Def/Sig/Op.

yep. Realised a bit after posting that. My understanding of spec was a bit wrong before :)

LoZander:
Is it a mistake that the SimplfiedAst created in `visitDef` in `Simplifier` is given the sourcelocation of the originals symbol (not spec or the def itself)?

main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
LoZander: Yes, these ones slipped by me. I'll fix it :)

LoZander:
> * [ ]  Remove loc from spec
> * [ ]  Remove unused local variables you have introduced

Both have been completed :)

magnus-madsen:
Merged!

PR #8841, opened 2024-10-09
style: reorder `MonoAst`

jaschdoc:
Moved `ApplyAtomic` before `ApplyClo` because that's the alphabetical ordering.


magnus-madsen:
Wait... I think pattern matches should also be reordered then or?

magnus-madsen (changes requested):
.

jaschdoc:
Ah okay I was only thinking about the apply ordering

magnus-madsen:
We try to keep declaration order and pattern match order in synch.

magnus-madsen (approved):

PR #8845, opened 2024-10-10
feat: jdk base effects

AStenbaek:


AStenbaek:
How would we categorize `finalize` in the `Object` class? 
`finalize` is called by the garbage collector during clean-up and should therefore be associated with the `Sys` effect.
This method is only really a security issue when overwritten by a sub-class to manipulate the JVM.

- Since every object is sub-class of `Object`, we cannot prevent this by adding `Object` to the BaseEffect.Class table.
- Since the method is invoked by the garbage collector rather than code, I do not believe adding it to the BaseEffect.Method table would solve the issue either.

magnus-madsen:
> How would we categorize `finalize` in the `Object` class? `finalize` is called by the garbage collector during clean-up and should therefore be associated with the `Sys` effect. This method is only really a security issue when overwritten by a sub-class to manipulate the JVM.

1. It does not matter.
2. Do you have examples of a finalizer calling a SecurityManager on a permissions it does not already have?
3. The effect of the finalizer can just be added to the effect of the constructor.

main/src/ca/uwaterloo/flix/language/phase/typer/BaseEffects.Classes.json
```diff
@@ -3,6 +3,7 @@
     "java.lang.Process": "Exec",
     "java.lang.ProcessBuilder": "Exec",
     "java.lang.ProcessHandle": "Exec",
-    "java.lang.reflect.Method": "Sys"
+    "java.lang.reflect.Method": "Sys",
+    "java.lang.Thread": "Exec"
```
magnus-madsen, line 7: Does the Thread class allow one to start a new process outside the VM?

magnus-madsen:
We discussed:
- Adding Env effect
- Boolean.getBoolean needs some effect, Sys (or Env if we have it)
- Classes as capabilities is not great if System is a capability.

magnus-madsen:
I merged something similar earlier.

PR #8848, opened 2024-10-10
feat: EffUnification3

JonathanStarup:
@mlutze I've replaced calls to `EffUnification.solve` with my `EffUnification3.solve`, is that sufficient to be "fully online"

PR #8850, opened 2024-10-11
fix: do not minimize types with errors

magnus-madsen:


magnus-madsen:
@JonathanStarup @mlutze This is a proposed work-around for #8848. 

In any case, I think TypeMinimization should also die.

mlutze (approved):

PR #8854, opened 2024-10-11
fix: disallow control-effects in spawn and new expressions

### LINKED ISSUE #8853: Enforce that spawn and new object expressions cannot have control-effects

magnus-madsen:
Related to #8853 but does not fully solve it

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Safety.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Safety.scala
@@ -370,7 +370,7 @@ object Safety {
       case Expr.InstanceOf(exp, _, _) =>
         visit(exp)
 
-      case Expr.CheckedCast(cast, exp, tpe, eff, loc) =>
+      case Expr.CheckedCast(cast, exp, tpe, _, loc) =>
         cast match {
           case CheckedCastType.TypeCast =>
             val from = Type.eraseAliases(exp.tpe)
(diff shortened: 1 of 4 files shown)


magnus-madsen:
@mlutze This is neither sufficient nor complete. But its better than nothing.

main/src/library/Adaptor.flix
```diff
@@ -40,11 +40,11 @@ mod Adaptor {
     /// Note: Currently requires a Proxy (until Java types support generics).
     ///
     pub def comparator(_: Proxy[a]): Comparator \ IO with Order[a] =
-        let f1 = (o1, o2) -> {
+        let f1 = (o1, o2) -> unchecked_cast ({
```
mlutze, line 43: Why is this one necessary?

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -1073,4 +1083,13 @@ object Safety {
   private def isAbstractMethod(m: java.lang.reflect.Method): Boolean =
     java.lang.reflect.Modifier.isAbstract(m.getModifiers)
 
+  /**
+    * Returns the control-effects inside `tpe`.
+    */
+  private def getControlEffs(tpe: Type): Set[Symbol.EffectSym] =
```
mlutze, line 1089: Should this be in Type? We already have several similar functions there.

mlutze (approved):

main/src/library/Adaptor.flix
magnus-madsen: Fixed.

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
magnus-madsen: I think this function is stupid and should be reworked. So I would keep it here for now.

PR #8856, opened 2024-10-11
feat: visitor

LoZander:
this PR adds the visitor, which allows traversing the AST selectively, performing some action for each visited node depending on node type.

related to #8639

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package ca.uwaterloo.flix.api.lsp.provider
+package ca.uwaterloo.flix.api.lsp
 
 import ca.uwaterloo.flix.api.lsp.Position
 import ca.uwaterloo.flix.language.ast.Ast.*
@@ -73,6 +73,7 @@ object Visitor {
     def consumeHandlerRule(rule: HandlerRule): Unit = ()
     def consumeInstance(ins: Instance): Unit = ()
     def consumeJvmMethod(method: JvmMethod): Unit = ()
+    def consumeLocalDefSym(symUse: LocalDefSymUse): Unit = ()
     def consumeMatchRule(rule: MatchRule): Unit = ()
     def consumeOp(op: Op): Unit = ()
     def consumeOpSymUse(sym: OpSymUse): Unit = ()
@@ -82,6 +83,7 @@ object Visitor {
     def consumeRecordLabelPattern(pat: RecordLabelPattern): Unit = ()
     def consumeSelectChannelRule(rule: SelectChannelRule): Unit = ()
     def consumeSig(sig: Sig): Unit = ()
+    def consumeSigSymUse(symUse: SigSymUse): Unit = ()
     def consumeSpec(spec: Spec): Unit = ()
     def consumeStruct(struct: Struct): Unit = ()
     def consumeStructField(field: StructField): Unit = ()
@@ -443,7 +445,6 @@ object Visitor {
     expr match {
       case Expr.Cst(_, _, _) => ()
       case Expr.Var(_, _, _) => ()
-      case Expr.Sig(_, _, _) => ()
       case Expr.Hole(_, _, _, _) => ()
 
       case Expr.HoleWithExp(exp, _, _, _) =>
@@ -459,12 +460,20 @@ object Visitor {
         visitFormalParam(fparam)
         visitExpr(exp)
 
-      case Expr.Apply(exp, exps, _, _, _) =>
+      case Expr.ApplyClo(exp, exps, _, _, _) =>
         visitExpr(exp)
         exps.foreach(visitExpr)
 
-      case Expr.ApplyDef(sym, exps, _, _, _, _) =>
-        visitDefSymUse(sym)
+      case Expr.ApplyDef(symUse, exps, _, _, _, _) =>
+        visitDefSymUse(symUse)
+        exps.foreach(visitExpr)
+
+      case Expr.ApplyLocalDef(symUse, exps, _, _, _, _) =>
+        visitLocalDefSymUse(symUse)
+        exps.foreach(visitExpr)
+
+      case Expr.ApplySig(symUse, exps, _, _, _, _) =>
+        visitSigSymUse(symUse)
         exps.foreach(visitExpr)
 
       case Expr.Unary(_, exp, _, _, _) =>
@@ -474,12 +483,7 @@ object Visitor {
         visitExpr(exp1)
         visitExpr(exp2)
 
-      case Expr.Let(_, _, exp1, exp2, _, _, _) =>
-        visitExpr(exp1)
-        visitExpr(exp2)
-
-      case Expr.LetRec(_, ann, _, exp1, exp2, _, _, _) =>
-        visitAnnotations(ann)
+      case Expr.Let(_, exp1, exp2, _, _, _) =>
         visitExpr(exp1)
         visitExpr(exp2)
 
@@ -697,6 +701,20 @@ object Visitor {
 
       case Expr.Error(_, _, _) => ()
     }
+  }
+
+  private def visitSigSymUse(symUse: SigSymUse)(implicit a: Acceptor, c: Consumer): Unit = {
+    val SigSymUse(_, loc) = symUse
+    if (!a.accept(loc)) { return }
+
+    c.consumeSigSymUse(symUse)
+  }
+
+  private def visitLocalDefSymUse(symUse: LocalDefSymUse)(implicit a: Acceptor, c: Consumer): Unit = {
+    val LocalDefSymUse(_, loc) = symUse
+    if (!a.accept(loc)) { return }
+
+    c.consumeLocalDefSym(symUse)
   }
 
   private def visitPredicateParam(pparam: PredicateParam)(implicit a: Acceptor, c: Consumer): Unit = {
--- a/main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
@@ -1,0 +1,914 @@
+/*
+ * Copyright 2024 Alexander Dybdahl Troelsen
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ca.uwaterloo.flix.api.lsp.provider
+
+import ca.uwaterloo.flix.api.lsp.Position
+import ca.uwaterloo.flix.language.ast.Ast.*
+import ca.uwaterloo.flix.language.ast.TypedAst.Pattern.*
+import ca.uwaterloo.flix.language.ast.TypedAst.Pattern.Record.RecordLabelPattern
+import ca.uwaterloo.flix.language.ast.TypedAst.{AssocTypeDef, Instance, *}
+import ca.uwaterloo.flix.language.ast.shared.{Annotation, Annotations}
+import ca.uwaterloo.flix.language.ast.{SourceLocation, Type}
+
+object Visitor {
+
+  /**
+    * Defines how each AST node type is handled when it's visited.
+    *
+    * Each consume function `consumeX` takes a value of type `X` and returns nothing.
+    * Thus, anything it does is through effects.
+    *
+    * Implementations need only implement the consume functions that they care about,
+    * Any consume function not implemented falls back on the default of doing nothing
+    * upon consuming the the corresponding AST node.
+    *
+    * Example
+    * {{{
+    * object ExprListConsumer extends Consumer {
+    *   var stack: List[Expr] = Nil
+    *   override consumeExpr(exp: Expr): Unit = {
+    *     stack = exp :: stack
+    *   }
+    * }
+    * }}}
+    *
+    * This consumer only cares about `Expr`s and simply collects all expressions visited.
+    */
+  trait Consumer {
+    def consumeAnnotation(ann: Annotation): Unit = ()
+    def consumeAnnotations(anns: Annotations): Unit = ()
+    def consumeAssocTypeConstructor(tcst: AssocTypeConstructor): Unit = ()
+    def consumeAssocTypeDef(tdefn: AssocTypeDef): Unit = ()
+    def consumeAssocTypeSig(tsig: AssocTypeSig): Unit = ()
+    def consumeAssocTypeSymUse(symUse: AssocTypeSymUse): Unit = ()
+    def consumeCase(cse: Case): Unit = ()
+    def consumeCaseSymUse(sym: CaseSymUse): Unit = ()
+    def consumeCatchRule(rule: CatchRule): Unit = ()
+    def consumeConstraint(c: Constraint): Unit = ()
+    def consumeConstraintParam(cparam: ConstraintParam): Unit = ()
+    def consumeDef(defn: Def): Unit = ()
+    def consumeDefSymUse(sym: DefSymUse): Unit = ()
+    def consumeDerivation(derive: Derivation): Unit = ()
+    def consumeDerivations(derives: Derivations): Unit = ()
+    def consumeEff(eff: Effect): Unit = ()
+    def consumeEffectSymUse(effUse: EffectSymUse): Unit = ()
+    def consumeEnum(enm: Enum): Unit = ()
+    def consumeEqualityConstraint(ec: EqualityConstraint): Unit = ()
+    def consumeExpr(exp: Expr): Unit = ()
+    def consumeFormalParam(fparam: FormalParam): Unit = ()
+    def consumeParYieldFragment(frag: ParYieldFragment): Unit = ()
+    def consumeHandlerRule(rule: HandlerRule): Unit = ()
+    def consumeInstance(ins: Instance): Unit = ()
+    def consumeJvmMethod(method: JvmMethod): Unit = ()
+    def consumeMatchRule(rule: MatchRule): Unit = ()
+    def consumeOp(op: Op): Unit = ()
+    def consumeOpSymUse(sym: OpSymUse): Unit = ()
+    def consumePattern(pat: Pattern): Unit = ()
+    def consumePredicate(p: Predicate): Unit = ()
+    def consumePredicateParam(pparam: PredicateParam): Unit = ()
+    def consumeRecordLabelPattern(pat: RecordLabelPattern): Unit = ()
+    def consumeSelectChannelRule(rule: SelectChannelRule): Unit = ()
+    def consumeSig(sig: Sig): Unit = ()
+    def consumeSpec(spec: Spec): Unit = ()
+    def consumeStruct(struct: Struct): Unit = ()
+    def consumeStructField(field: StructField): Unit = ()
+    def consumeStructFieldSymUse(symUse: StructFieldSymUse): Unit = ()
+    def consumeTypeMatchRule(rule: TypeMatchRule): Unit = ()
+    def consumeTrait(traitt: Trait): Unit = ()
+    def consumeTraitConstraint(tc: TraitConstraint): Unit = ()
+    def consumeTraitConstraintHead(tcHead: TraitConstraint.Head): Unit = ()
+    def consumeTraitSymUse(symUse: TraitSymUse): Unit= ()
+    def consumeType(tpe: Type): Unit = ()
+    def consumeTypeAlias(alias: TypeAlias): Unit = ()
+    def consumeTypeParam(tparam: TypeParam): Unit = ()
+  }
+
+  /**
+    * Defines whether AST nodes are visited.
+    */
+  trait Acceptor {
+    /**
+      * Defines whether an AST node is visited based on its `SourceLocation`
+      *
+      * @param loc  `SourceLocation` of the AST node
+      * @return     `true` if the AST node should be visited, `false` otherwise
+      */
+    def accept(loc: SourceLocation): Boolean;
+  }
+
+  /**
+    * Acceptor that accepts all AST nodes.
+    */
+  case object AllAcceptor extends Acceptor {
+    def accept(loc: SourceLocation): Boolean = true
+  }
+
+  /**
+    * Acceptor that accepts all AST nodes whose `SourceLocation` is within
+    * the file given by the path `uri`.
+    *
+    * @param uri  the path of the file that an AST node `SourceLocation` must be within to be accepted.
+    */
+  case class FileAcceptor(uri: String) extends Acceptor {
+    def accept(loc: SourceLocation): Boolean = uri == loc.source.name
+  }
+
+  /**
+    * Acceptor that accepts an AST node if it contains a given position.
+    *
+    * `InsideAcceptor` accepts an AST if it's `SourceLocation` is within the file given by `uri`
+    * and the position `pos` is within the `SourceLocation` of the AST.
+    *
+    * @param uri  the path to the file that the AST node `SourceLocation` must be in to be accepted.
+    * @param pos  the position that must be within the AST node `SourceLocation` for the node to be accepted.
+    */
+  case class InsideAcceptor(uri: String, pos: Position) extends Acceptor {
+    def accept(loc: SourceLocation): Boolean = inside(uri, pos)(loc)
+  }
+
+  /**
+    * Visits the root AST node and recursively visits
+    * all children that are accepted by the acceptor,
+    * "consuming" each node visited (including the root).
+    *
+    * By "consuming" what is meant is that the node is used
+    * as the input to a function of the consumer associated with
+    * the AST node type. Such consumer funcitons have no output,
+    * but can have effects. For instance, the consumer can be defined such that each
+    * expression containing a specific variable is collected in a list, via
+    * mutation a variable.
+    *
+    * Note that if a node is not accepted, none of the nodes in it's
+    * subtree will be visited either. For instance, if a `def` is not
+    * accepted, none of the expressions in the body will be visited eiter,
+    * even if they would otherwise be accepted by the acceptor.
+    *
+    * @param root      The AST root node.
+    * @param consumer  A consumer which defines what to do when visiting different types of AST nodes.
+    * @param acceptor  An acceptor which defines the criteria for whether an AST node should be visited.
+    */
+  def visitRoot(root: Root, consumer: Consumer, acceptor: Acceptor): Unit = {
+
+    implicit val c: Consumer = consumer
+    implicit val a: Acceptor = acceptor
+
+    root.defs.values.foreach(visitDef)
+
+    root.effects.values.foreach(visitEffect)
+
+    root.enums.values.foreach(visitEnum)
+
+    root.instances.values.flatten.foreach(visitInstance)
+
+    root.sigs.values.foreach(visitSig)
+
+    root.structs.values.foreach(visitStruct)
+
+    root.traits.values.foreach(visitTrait)
+
+    root.typeAliases.values.foreach(visitTypeAlias)
+  }
+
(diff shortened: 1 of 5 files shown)


LoZander:
There are changes I want to make, but most of these are dependent on #8839. Otherwise, maybe renaming some of the tests is in order. But otherwise, I'm happy with this. Let me know if there's anything to be fixed, otherwise it might be merge time :)

main/test/ca/uwaterloo/flix/api/lsp/VisitorSuite.scala
```diff
@@ -0,0 +1,137 @@
...
+import ca.uwaterloo.flix.language.ast.shared.{Input, SecurityContext, Source}
+import ca.uwaterloo.flix.language.ast.{SourceLocation, SourcePosition}
+import org.scalatest.funsuite.AnyFunSuite
+
+
+class VisitorSuite extends AnyFunSuite {
+  val source: Source = Source(Input.Text("test", "test", stable = true, SecurityContext.AllPermissions), Array.emptyCharArray)
+  val uri = "test"
+
+  test("Inside.01") {
```
magnus-madsen, line 27: Inside -> inside

main/test/ca/uwaterloo/flix/api/lsp/VisitorSuite.scala
```diff
@@ -0,0 +1,137 @@
...
+    val loc = SourceLocation(
+      isReal = false,
```
magnus-madsen, line 29: I think you want these source locations to be real. And then I think the visitor should only accept real source locations.

main/test/ca/uwaterloo/flix/api/lsp/VisitorSuite.scala
```diff
@@ -0,0 +1,137 @@
...
+
+  test("Inside.02") {
+    val loc = SourceLocation(
+      isReal = false,
+      SourcePosition(source, 5, 2),
+      SourcePosition(source, 5, 5),
+    )
+    val pos = Position(5, 4)
+
+    assert(Visitor.inside(uri, pos)(loc) === true)
```
magnus-madsen, line 45: is there not an assertFalse? Otherwise assert(! should do)

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
```diff
@@ -0,0 +1,937 @@
...
+import ca.uwaterloo.flix.language.ast.TypedAst.{AssocTypeDef, Instance, *}
+import ca.uwaterloo.flix.language.ast.shared.{Annotation, Annotations}
+import ca.uwaterloo.flix.language.ast.{SourceLocation, Type}
+
+object Visitor {
+
+  /**
+    * Defines how each AST node type is handled when it's visited.
+    *
+    * Each consume function `consumeX` takes a value of type `X` and returns nothing.
```
magnus-madsen, line 31: returns unit

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
```diff
@@ -0,0 +1,937 @@
...
+    val afterStart = loc.beginLine < pos.line ||
+                     loc.beginLine == pos.line && loc.beginCol <= pos.character
+    if (!afterStart) { return false }
+
+
+    val beforeEnd = pos.line < loc.endLine ||
+                    pos.line == loc.beginLine && pos.character < loc.endCol
+    if (!beforeEnd) { return false }
+
+    true
```
magnus-madsen, line 209: Much better- thanks!

magnus-madsen: Do you have 100% test coverage?

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
```diff
@@ -0,0 +1,937 @@
...
+    * @param pos  the position that we want to know whether is within `loc`.
+    * @param loc  the `SourceLocation` that want to know if `pos` is within.
+    * @return `true` if `pos` in file at path `uri` is within `loc`. `false` otherwise.
+    */
+  def inside(uri: String, pos: Position)(loc: SourceLocation): Boolean = {
+    val sameSource = uri == loc.source.name
+    if (!sameSource) { return false }
+
+    val afterStart = loc.beginLine < pos.line ||
+                     loc.beginLine == pos.line && loc.beginCol <= pos.character
```
magnus-madsen, line 201: I think extra parens would be useful

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
```diff
@@ -0,0 +1,937 @@
...
+    if (!afterStart) { return false }
+
+
+    val beforeEnd = pos.line < loc.endLine ||
+                    pos.line == loc.beginLine && pos.character < loc.endCol
```
magnus-madsen, line 206: I think extra parens would be useful

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
magnus-madsen: I dont recall whether || or && binds more tightly, so its hard to see if the code is correct.

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
```diff
@@ -0,0 +1,937 @@
...
+  /**
+    * Returns `true` if the position `pos` within the file given by path `uri` is contained within the `SourceLocation` `loc`.
+    * Returns `false` otherwise.
+    *
+    * @param uri  the path of the file that `pos` is within.
+    * @param pos  the position that we want to know whether is within `loc`.
+    * @param loc  the `SourceLocation` that want to know if `pos` is within.
+    * @return `true` if `pos` in file at path `uri` is within `loc`. `false` otherwise.
+    */
+  def inside(uri: String, pos: Position)(loc: SourceLocation): Boolean = {
```
magnus-madsen, line 196: This is a helper function, so I would move it to the very bottom of this file, since it is not related to all the visitX functions.

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
```diff
@@ -0,0 +1,937 @@
...
+  private def visitAssocTypeSymUse(symUse: AssocTypeSymUse)(implicit a: Acceptor, c: Consumer): Unit = {
+    val AssocTypeSymUse(_, loc) = symUse
+    if (!a.accept(loc)) { return }
+
+    c.consumeAssocTypeSymUse(symUse)
+  }
+
+  private def visitSig(sig: Sig)(implicit a: Acceptor, c: Consumer): Unit = {
+    val Sig(_, spec, exp) = sig
+    // TODO `insideSig` is a hack and should eventually be removed. This hack is necessary because a signature currently does not have a location.
```
magnus-madsen, line 303: Once #8839 is in this can be fixed 👍

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
```diff
@@ -0,0 +1,937 @@
...
+
+    c.consumeEff(eff)
+
+    visitAnnotations(ann)
+    ops.foreach(visitOp)
+  }
+
+  private def visitOp(op: Op)(implicit a: Acceptor, c: Consumer): Unit = {
+    val Op(_, spec) = op
+    // TODO: hack that should eventually be fixed. Really it should be `op.loc`, but since op decls don't have locations, this hack is necessary for now
```
magnus-madsen, line 369: Once #8839 is in this can be fixed 👍

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
```diff
@@ -0,0 +1,937 @@
...
+    if (!a.accept(spec.loc)) { return }
+
+    c.consumeOp(op)
+
+    visitSpec(spec)
+  }
+
+  private def visitDef(defn: Def)(implicit a: Acceptor, c: Consumer): Unit = {
+    val Def(sym, spec, exp) = defn
+    // TODO `insideDef` is a hack and should be removed eventually. Necessary for now since Defs  don't have locations
```
magnus-madsen, line 379: Once #8839 is in this can be fixed 👍

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
```diff
@@ -0,0 +1,937 @@
...
+    expr match {
+      case Expr.Cst(_, _, _) => ()
+      case Expr.Var(_, _, _) => ()
+      case Expr.Hole(_, _, _, _) => ()
+
+      case Expr.HoleWithExp(exp, _, _, _) =>
+        visitExpr(exp)
+
+      // we do nothing here, because open as is a restrictable enum feature and thus experimental
+      case Expr.OpenAs(_, _, _, _) => ()
```
magnus-madsen, line 454: Lets just write // Unsupported feature on the same line, i.e.

```
case Expr.OpenAs(_, _, _, _) => () // Not visited, unsupported feature.
```

magnus-madsen: And use the same style everywhere else.

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
```diff
@@ -0,0 +1,937 @@
...
+      case Expr.Match(exp, rules, _, _, _) =>
+        visitExpr(exp)
+        rules.foreach(visitMatchRule)
+
+      case Expr.TypeMatch(exp, rules, _, _, _) =>
+        visitExpr(exp)
+        rules.foreach(visitTypeMatchRule)
+
+      case Expr.RestrictableChoose(_, _, _, _, _, _) =>
+        // Does nothing because feature is experimental
```
magnus-madsen, line 522: See above

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
```diff
@@ -0,0 +1,937 @@
...
+
+      case Expr.Tag(symUse, exp, _, _, _) =>
+        visitCaseSymUse(symUse)
+        visitExpr(exp)
+
+      case Expr.RestrictableTag(_, exp, _, _, _) =>
```
magnus-madsen, line 528: make this unsupported.

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
```diff
@@ -0,0 +1,937 @@
...
+      case Expr.ArrayLength(exp, _, _) =>
+        visitExpr(exp)
+
+      case Expr.ArrayStore(exp1, exp2, exp3, _, _) =>
+        visitExpr(exp1)
+        visitExpr(exp2)
+        visitExpr(exp3)
+
+      case Expr.StructNew(_, fields, region, _, _, _) =>
+        fields.foreach{case (symUse, exp) =>
```
magnus-madsen, line 568: linebreak and indent before case.

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
```diff
@@ -0,0 +1,937 @@
...
+  private def visitDefSymUse(symUse: DefSymUse)(implicit a: Acceptor, c: Consumer): Unit = {
+    val DefSymUse(_, loc) = symUse
+    if(!a.accept(loc)) { return }
+
+    c.consumeDefSymUse(symUse)
+  }
+
+  private def visitSelectChannelRule(rule: SelectChannelRule)(implicit a: Acceptor, c: Consumer): Unit = {
+    val SelectChannelRule(_, chan, exp) = rule
+    // TODO `insideRule` is a hack, should be removed eventually. Necessary for now since SelectChannelRule don't have locations
```
magnus-madsen, line 766: Does #8839 fix this? Otherwise I suppose this can be a follow-up PR.

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
```diff
@@ -0,0 +1,937 @@
...
+    if (!a.accept(loc)) { return }
+
+    c.consumeFormalParam(fparam)
+
+    visitType(tpe)
+  }
+
+  private def visitHandlerRule(rule: HandlerRule)(implicit a: Acceptor, c: Consumer): Unit = {
+    val HandlerRule(op, fparams, exp) = rule
+    // TODO `insideRule` is a hack, should be removed eventually. Necessary for now since HandlerRules don't have locations
```
magnus-madsen, line 787: Does #8839 fix this? Otherwise I suppose this can be a follow-up PR.

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
```diff
@@ -0,0 +1,937 @@
...
+
+    c.consumeParYieldFragment(frag)
+
+    visitPattern(pat)
+    visitExpr(exp)
+  }
+
+  private def visitMatchRule(rule: MatchRule)(implicit a: Acceptor, c: Consumer): Unit = {
+    val MatchRule(pat, guard, exp) = rule
+    // TODO `insideRule` is a hack, should be removed eventually. Necessary for now since MatchRules don't have locations
```
magnus-madsen, line 817: Does #8839 fix this? Otherwise I suppose this can be a follow-up PR.

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
```diff
@@ -0,0 +1,937 @@
...
+    c.consumeMatchRule(rule)
+
+    visitPattern(pat)
+    guard.foreach(visitExpr)
+    visitExpr(exp)
+  }
+
+  private def visitTypeMatchRule(rule: TypeMatchRule)(implicit a: Acceptor, c: Consumer): Unit = {
+    val TypeMatchRule(sym, tpe, exp) = rule
+    // TODO `insideRule` is a hack, should be removed eventually. Necessary for now since TypeMatchRules don't have locations
```
magnus-madsen, line 830: Does #8839 fix this? Otherwise I suppose this can be a follow-up PR.

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
```diff
@@ -0,0 +1,937 @@
...
+  }
+
+  private def visitType(tpe: Type)(implicit a: Acceptor, c: Consumer): Unit = {
+    if (!a.accept(tpe.loc)) { return }
+    c.consumeType(tpe)
+  }
+
+  private def visitAnnotations(anns: Annotations)(implicit a: Acceptor, c: Consumer): Unit = {
+    val Annotations(ls) = anns
+    // TODO `insideAnns` is a hack, should be removed eventually. Necessary for now since Annotations (not to be confused with Annotation) don't have locations
```
magnus-madsen, line 847: I think this is more tricky. So I would leave the comment, but without a TODO. This is just the way it is.

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
```diff
@@ -0,0 +1,937 @@
...
+  }
+
+  private def visitAnnotation(ann: Annotation)(implicit a: Acceptor, c: Consumer): Unit = {
+    if (!a.accept(ann.loc)) { return }
+    c.consumeAnnotation(ann)
+  }
+
+  private def visitCatchRule(rule: CatchRule)(implicit a: Acceptor, c: Consumer): Unit = {
+    val CatchRule(sym, _, exp) = rule
+    // TODO `insideRule` is a hack, should be removed eventually. Necessary for now since CatchRules don't have locations
```
magnus-madsen, line 863: Does #8839 fix this? Otherwise I suppose this can be a follow-up PR.

magnus-madsen:
- [ ] Fix easy comments
- [ ] Wait for #8839 to merge 
- [ ] Rework those acceptance checks that #8839 helps with 
- [ ] (Later) Add source locations to MatchRules etc. (I suggest one PR per AST node).

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
LoZander: Unfortunately, #8839 does not fix this :/

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
LoZander: Same as below

main/test/ca/uwaterloo/flix/api/lsp/VisitorSuite.scala
```diff
@@ -0,0 +1,137 @@
...
+  val uri = "test"
+
+  test("Inside.01") {
+    val loc = SourceLocation(
+      isReal = false,
+      SourcePosition(source, 3, 10),
+      SourcePosition(source, 6, 2))
+    val pos = Position(5, 5)
+
+    assert(Visitor.inside(uri, pos)(loc) === true)
```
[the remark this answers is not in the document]
LoZander: I thought === was necessary for the scala test assertions to give more information about failure. But perhaps it's not? It not, then ofc. I agree with you that it's silly compare booleans like that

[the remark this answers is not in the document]
LoZander: Reading a bit up on scala tests tells me that I did just misunderstand. I'll fix it :)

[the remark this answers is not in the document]
LoZander: The benefits of === seem irrelevant for booleans

main/test/ca/uwaterloo/flix/api/lsp/VisitorSuite.scala
LoZander: Good point. It's perhaps worth adding a test to ensure that `inside` rejects non-real source locations?

main/test/ca/uwaterloo/flix/api/lsp/VisitorSuite.scala
LoZander: There is not an assertFalse, but yeah, assert and ! works

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
LoZander: Good point. Better to make it as unambiguous as possible

main/test/ca/uwaterloo/flix/api/lsp/VisitorSuite.scala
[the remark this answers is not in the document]
LoZander: Anyways, I'ved simplified it as you suggested

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
LoZander: No, I did not. I have added more extensive testing (maybe too extensive tbh), and now there should be coverage on the entire inside function. These new tests also did shed light on a dumb mistake I made in `inside` that went over my head

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
LoZander: You could maybe argue that `Annotations`s do not require visting, and that only each individual annotation ought to be visited. But then that would make it impossible to do something regarding a set of annotations (as opposed to the individual) and it would create a somewhat mismatch between the visitor and the actual structure of the AST, since `ANnotations` is an abstraction in the AST at the end of the day. That's why I added both visitAnnotations and visitAnnotation. What do you think?

magnus-madsen:
@LoZander PR #8839 has now been merged, so you can quickly iterate on this

main/test/ca/uwaterloo/flix/api/lsp/VisitorSuite.scala
magnus-madsen: Sure, why not!

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
magnus-madsen: Roger. Follow-up  PR.

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
magnus-madsen: I agree with your observations. For now, I would remove visitAnnotations (plural). Later, if Annotations were to have a source location itself, we could support your use case.

magnus-madsen:
@LoZander I would like to merge this today. The only blocker is see is fixing these:

```
// TODO `insideSig` is a hack and should eventually be removed. This hack is necessary because a signature currently does not have a location.
```

which should be quick.

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
LoZander: same as above

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
LoZander: same as the other rule visitations, sadly

LoZander:
> @LoZander I would like to merge this today. The only blocker is see is fixing these:
> 
> ```
> // TODO `insideSig` is a hack and should eventually be removed. This hack is necessary because a signature currently does not have a location.
> ```
> 
> which should be quick.

Yep. I want to also cover some of the other remaining conversations, but I don't imagine any of it will take long. I'll get it done within the day, most likely before noon

magnus-madsen:
> I'll get it done within the day, most likely before noon

Sounds good. I am here.

LoZander:
> > I'll get it done within the day, most likely before noon
> 
> Sounds good. I am here.

Right, cool :)
So then I'll come directly to you in case some issue does somehow present itself (although I don't imagine any will)

LoZander:
There is actually one issue. The issue of what happens to visitSpec. Specs don't have a location anymore. But maybe it's fine that they are neither checked with the acceptor nor consumed, because Specs aren't really interesting in their own right. If my understanding is correct, it's simply an abstraction for bundling data common across Sigs, Defs and Ops. In a way, Specs could be seen as not "real" AST elements. In the same way that Annotations (not to be confused with Annotation) is not a "real" AST element. What do you think?

LoZander:
My current idea is to keep visitSpec, but simply not do the acceptor check nor consume. Make visitSpec a simpler helper for visiting Defs, Sigs and Ops

magnus-madsen:
Drop visitSpec

magnus-madsen:
EDIT: We should still visitSpec (to visit its formal params etc), but just always visit it.

magnus-madsen:
So what I mean is simply: If are visiting the e.g. def then you always visit the spec.

LoZander:
Do you think we should keep visitSpec around without consuming or checking with acceptor, just to avoid code duplication across visitDef, visitSig and visitOp, or do you think it would be too confusing for there to be a function called visitSpec that doesn't follow the same pattern (or serve the same function) as the other visit functions? Maybe it's better to just have the code duplication or maybe name the function something else to make it clear that it serves a different purpose?

LoZander:
> Do you think we should keep visitSpec around without consuming or checking with acceptor, just to avoid code duplication across visitDef, visitSig and visitOp, or do you think it would be too confusing for there to be a function called visitSpec that doesn't follow the same pattern (or serve the same function) as the other visit functions? Maybe it's better to just have the code duplication or maybe name the function something else to make it clear that it serves a different purpose?

To be clear, what I mean is this: we could keep a function around called `visitSpec`, which just always visits the Spec. It doesn't consume the Spec and doesn't do the acceptor check, but simply serves the purpose of avoiding the code duplication that would occur if we added lines for visiting everything in spec in visitDef, visitSig and visitOp. The concern here, however, would be that maybe it would be confusing if it were called `visitSpec`, since this function clearly would break the pattern established by the other `visitX` functions, because really it serves a subtly different purpose.

Do you think we should inline visiting the parts of spec in the three visit functions and just accept the code duplication, keep the `visitSpec` like I described or keep something like `visitSpec` around, but maybe give it a different name?

magnus-madsen:
> To be clear, what I mean is this: we could keep a function around called visitSpec, which just always visits the Spec. It doesn't consume the Spec and doesn't do the acceptor check, but simply serves the purpose of avoiding the code duplication that would occur

I agree with this.

magnus-madsen:
Its a bit like the situation with Annotations. We have visitAnnotations, but no accept check nor any consumeAnnotations

LoZander:
I see, then I'll handle it that way.

But what you write actually raises another question. Currently, we do consume `Annotations`s as well as `Annotation`s, but maybe that is actually a mistake? My thought was that Annotations do represent something subtly different. They represent a specific set of annotations for a specific thing. And so, they might be worth consuming. But maybe I should remove the consumer for it?

magnus-madsen:
> I see, then I'll handle it that way.
> 
> But what you write actually raises another question. Currently, we do consume `Annotations`s as well as `Annotation`s, but maybe that is actually a mistake? My thought was that Annotations do represent something subtly different. They represent a specific set of annotations for a specific thing. And so, they might be worth consuming. But maybe I should remove the consumer for it?

Lets remove the Annotations consumer

LoZander:
> > I see, then I'll handle it that way.
> > But what you write actually raises another question. Currently, we do consume `Annotations`s as well as `Annotation`s, but maybe that is actually a mistake? My thought was that Annotations do represent something subtly different. They represent a specific set of annotations for a specific thing. And so, they might be worth consuming. But maybe I should remove the consumer for it?
> 
> Lets remove the Annotations consumer

Got it. Will do :)

LoZander:
Then I guess we also don't really need to do the acceptor check in visitAnnotations, since all it does not is call visitAnnotation for each annotation, which will ofc. do a check for each, making the check in visitAnnotations completely redundant.

magnus-madsen:
> Then I guess we also don't really need to do the acceptor check in visitAnnotations, since all the function does now is call visitAnnotation for each annotation, which will ofc. do a check for each, making the check in visitAnnotations completely redundant.

Agreed

magnus-madsen:
Lets go!

LoZander:
Yeah!

At this point, you could claim that visitAnnotations is kinda redundant in its own right, since we could just inline that `.foreach(visitAnnotation)`. But maybe it's nice enough to not have to do that `.foreach` everywhere annotations occur

LoZander:
hm. Maybe that indirection of an additional call is just confusing at that point though

magnus-madsen:
> Yeah!
> 
> At this point, you could maybe argue that visitAnnotations is kinda redundant in its own right, since we could just inline that `.foreach(visitAnnotation)`. But maybe it's nice enough to not have to do that `.foreach` everywhere annotations occur

No keep it because then you are structurally recursive, and later we might add source locations to Annotations (and spec and MatchRule and so on)

LoZander:
> > Yeah!
> > At this point, you could maybe argue that visitAnnotations is kinda redundant in its own right, since we could just inline that `.foreach(visitAnnotation)`. But maybe it's nice enough to not have to do that `.foreach` everywhere annotations occur
> 
> No keep it because then you are structurally recursive, and later we might add source locations to Annotations (and spec and MatchRule and so on)

oh, good point! I'll keep then :)

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
LoZander: As we've been discussing, I'll keep it around with no consumer and no check :)

magnus-madsen (approved):

LoZander:
Weirdly the line `val sameSource = uri == loc.source.name` in `inside` is only "partially" executed according to intellij coverage report :shrug:

LoZander:
Feels awesome to have the Visitor done (with a few caveats)! Then the work on the hover provider can truly get going :D

magnus-madsen:
> Feels awesome to have the Visitor done (with a few caveats)! Then the work on the hover provider can truly get going :D

Yes. Now its time to quickly create a HoverProvider PR that uses the visitors so we can start experimenting. Then from here it should be more straightforward and changes can come in smaller PRs.

LoZander:
> > Feels awesome to have the Visitor done (with a few caveats)! Then the work on the hover provider can truly get going :D
> 
> Yes. Now its time to quickly create a HoverProvider PR that uses the visitors so we can start experimenting. Then from here it should be more straightforward and changes can come in smaller PRs.

Yeah. Sounds great. And then we can start churning out other provider refactorings to eliminate uses of Index :D

LoZander:
But yeah, one thing at a time. I'll get a HoverProvider PR going soon

magnus-madsen:
Merged! 🚀

Issue #8817, opened 2024-10-07
`Position`, `Range`, `SourceLocation`, and `SourcePosition`


JonathanStarup:
```scala
case class Position(line: Int, character: Int)
case class Range(start: Position, end: Position)
case class SourceLocation(isReal: Boolean, sp1: SourcePosition, sp2: SourcePosition) {
    assert(sp1.source eq sp2.source)
    ..
}
case class SourcePosition(source: Source, line: Int, col: Short)
```
Shouldn't it really just be 
```scala
case class Position(line: Int, character Int) // or Short, doesn't matter
case class Range(start: Position, end: Position) {
    def  contains(p: Position): Boolean
case class SourceLocation(isReal: Boolean, source: Source, range: Range) {
    def start: Position = range.start
    def end: Position = range.end
    def contains(p: Position): Boolean = range.contains(p)
}
```
and then we implement `<, <=, >, >=` on position together with some shorthand functions so you can write
```scala
pos1 <= pos2
loc.contains(pos)
pos1.sameLineAs(pos2)
..
```
instead of direct access stuff like 
```scala
traitSym.loc.source.name == uri
      && (traitSym.loc.beginLine < p.line
      || (traitSym.loc.beginLine == p.line && traitSym.loc.beginCol <= p.character))
      && (traitSym.loc.endLine > p.line
      || (traitSym.loc.endLine == p.line && traitSym.loc.endCol >= p.character))
```
which would be 
```scala
traitSym.loc.source.name == uri
    && traitSym.loc.begin <= p && traitSym.loc.end > p // is it inclusive, exclusive? who knows, doc comment doesn't.
```
or just
```scala
traitSym.loc.source.name == uri && traitSym.loc.contains(p)
```

magnus-madsen:
@LoZander CC

LoZander:
Oh, so you mean to add Positions as a thing not just in the LSP side of things but in the very definition of SourceLocation, so that we may use Positions in the LSP side of things (where there is not "source") and still be able to easily compare SouceLocations with Positions?

LoZander:
Your proposed solution would certainly make it easier to use Positions. I'll look into it :)

magnus-madsen:
We should discuss this: one issue is the increased memory cost.

JonathanStarup:
the actual representation is whatever, i just think we should code with that kind of interface instead of comparing integers

LoZander:
[...]

PR #8857, opened 2024-10-12
proposal: from SourcePosition to Position

LoZander:
This PR replaces `SourcePosition` with `Position`, moving the `Source` to the `SourceLocation` that the given positions are associated with. This means that

`SourcePosition(isReal: Boolean, sp1: SourcePosition, sp2: SourcePosition)` becomes `SourcePosition(isReal: Boolean, source: Source, range: Range)

Where `Range` is `Range(start: Position, end: Position)`

This change serves two purposes:
- simplifying the notion of positions, since these do not *need* to include the source
- bridge the gap between the LSP notion of positions and the compiler's, enabling direct use of compiler locations on the LSP side of things.

Additionally, an order on positions is implemented, allowing easy comparison.

Related to #8817


magnus-madsen:
This is a much more invasive refactoring which should be discussed. 

So I would hold off until we have analyzed the situation together.

PR #8859, opened 2024-10-12
chore: remove `TypeMinimization`

magnus-madsen:
Experimental results shows that it has no impact on performance.

While it might simplify some types shown to the user -- all of that will break once we switch to set effects anyway.


magnus-madsen:
The advantage of removing this and doing it now is that it simplifies BU in the compiler and understanding performance implications.

magnus-madsen:
In particular this would probably skew some performance experiments slightly in unpredictable directions.

magnus-madsen:
!!!jar

magnus-madsen:
A concern is memory consumption. I need to check that.

magnus-madsen:
Memory usage is fine too.

mlutze (approved):

PR #8869, opened 2024-10-13
refactor: HoverProvider using LSP visitor

LoZander:
This PR refactors the HoverProvider to use the LSP visitor instead of the index.

Related to #8639


LoZander:
alternatively, you could also create the JObjects in the consumer, and then just return the head at the end of processHover, but then we have to create those JObjects for each consumed AST node

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
```diff
@@ -28,16 +30,46 @@ import scala.annotation.tailrec
...
     }
   }
 
+  private def hoverAny(x: AnyRef, uri: String, pos: Position)(implicit root: Root, flix: Flix): JObject = x match {
+    case tpe: Type => hoverKind(tpe)
+    case exp: Expr => hoverTypeAndEff(exp.tpe, exp.eff, exp.loc)
+    case DefSymUse(sym, loc) => hoverDef(sym, loc)
+    case SigSymUse(sym, loc) => hoverSig(sym, loc)
+    case OpSymUse(symUse, loc) => hoverOp(symUse, loc)
+    case Def(sym, _, _, loc) => hoverDef(sym, loc)
```
magnus-madsen, line 64: I think these should be prefixed by their type. Is it Declaration.Def or something else?

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
```diff
@@ -28,16 +30,46 @@ import scala.annotation.tailrec
...
+    case Sig(sym, _, _, loc) => hoverSig(sym, loc)
+    case Op(sym, _, loc) => hoverOp(sym, loc)
+    case FormalParam(_, _, tpe, _, loc) => hoverType(tpe, loc)
+    case _ => mkNotFound(uri, pos)
+  }
+
   @tailrec
-  private def hoverEntity(entity: Entity, uri: String, pos: Position)(implicit Index: Index, root: Root, flix: Flix): JObject = entity match {
+  private def hoverEntity(entity: Entity, uri: String, pos: Position)(implicit root: Root, flix: Flix): JObject = entity match {
```
magnus-madsen, line 72: Presumably this can be deleted.

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
```diff
@@ -28,16 +30,46 @@ import scala.annotation.tailrec
...
 
-  def processHover(uri: String, pos: Position)(implicit index: Index, root: Root, flix: Flix): JObject = {
-    index.query(uri, pos) match {
-      case None => mkNotFound(uri, pos)
+  def processHover(uri: String, pos: Position)(implicit root: Root, flix: Flix): JObject = {
+    var stack: List[AnyRef] = Nil
+    def push(x: AnyRef): Unit = {
+      stack = x :: stack
+    }
+    object HoverConsumer extends Visitor.Consumer {
```
magnus-madsen, line 38: I think this can be encapsulated in one reusable abstraction. 

A class that both extends Consumer and has the stack inside it.

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
LoZander: yes, but in fact `hoverEntity` will be removed entirely. Currently I'm just using it as reference for the current behaviour of the hover provider

LoZander: I don't believe the `Entity` abstraction will be of much use moving forward

magnus-madsen: Indeed. Entity will be removed together with Index.

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
LoZander: Yeah, we could do that, but wouldn't we then need to implement each consumer such that everything is added to the stack? We might not want that. In this casee in particular, I don't think that's useful. I guess we could add everything to the stack and just ignore the things we don't care about by looking at the next element if the head is "uninteresting" to us.

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
LoZander: Good call :)

LoZander: Actually, all of these are top most things within the TypedAst and Ast. So we'd have to prefix them with these

LoZander: We could do that, but I don't know that that would add much towards analysability

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
magnus-madsen: I don't understand. I am just saying that the current code can be encapsulated in a reusable class.

magnus-madsen: Like you will need the same functionality in e.g. HighlightProvider.

magnus-madsen: "StackConsumer"

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
magnus-madsen: Prefix with TypedAst. Dont prefix anything from Ast.

magnus-madsen: Like TypedAst.Def, but Expr alone is fine.

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
LoZander: My problem with the abstraction is that what needs to be added to the stack is dependent on the use case. What we want to add to the stack and what we want to ignore, might be different between HoverProvider and HightlightProvider. Then this abstraction would need to still provide the same default implementations as the Consumer, so that each use site may implement just the consumers they need. But at that point, the abstraction gives us nothing, except maybe packaging the stack itself within the consumer behind some interface

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
magnus-madsen: Wait. Maybe I am just being silly. I suppose there is only one thing called e.g. `Op`?

magnus-madsen:
!!!jar

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
LoZander: There might be some overlap between Expr.X and TypedAst.X, but I'm not sure. So you might have been onto something

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
LoZander: and hoverEntity will then be removed as well. Currently, it's not called at all

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
LoZander: Unless ofc. we just want everything on the stack. Then yeah, that abstraction would make sense

magnus-madsen: The HighlightProvider will need to visitors, but it seems to me that the first one would very much be exactly equivalent to this one-- i.e. to find the item that the cursor is on.

LoZander:
there is some weird behaviour when hovering on the return type of a def when the return type is an arrow type. Normally, we'd get the kind of the type when hovering on a type like that, but for some reason, while hovering on the return type in this specific case, it's as if we're hovering on the entire def. Since the visitor doesn't differentiate between these cases, I worry there might be some problem with the generation of the locations or something like that.

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
LoZander: What I'm saying is, we might not want to add every element visited to the stack. I don't think it's useful to add all the elements visited to the stack in the hover provider for instance. But what that means is that such a StackConsumer can't actually implement any consumers, since that would lock-in what's added to the stack. Only if we always wish to add everything visited to the stack, does this abstraction then make sense

LoZander: any if StackConsumer can't implement ant consumeX functions, then StackConsumer adds very little if anything

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
magnus-madsen: Just remove it then

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
magnus-madsen: Do you agree that HoverProvider and HighlightProvider need to perform the same search?

LoZander: their searches will visit the same elements, but does that necessarily mean they want the same consumers?

LoZander: I suppose, at the end of the day, it's only the very last (most specific) element that matters in both cases. But even here, there might be reasons that we want the two cases to be different. Imagine that that you're hovering on some element `x` but actually within that element is a subelement `y` that you're more specifically hovering on. Now lets say that the hover can't say anything about type of AST node. Then we might want to "backtrack" and actually give information about the parent `x`. We could add it all to the stack and backtrack, or, if we just never add the elements we can't say anything about, we get the same effect but without having to actively backtrack. Now lets say we're highlighting `x`, maybe in this case highlighting the most specific `y` object does make sense, and then we don't want to backtrack. We can ofc. use the same Consumer that adds everything visited (including `y`) to the stack in both cases, and backtrack in the hover provider case. But instead we could simply use two different Consumers. For the hover provider we could use one that implements consumeX and but doesn't implement consumeY, so that y (type Y) is just never added to the stack. For the hightlight provider, we could then implement both consumeX and consumeY, such that y is added to the stack. This is what I mean by different situations maybe calling for different things to be added to the stack. With a `StackConsumer`, we'd have to add everything to the stack and backtrack if necessary.

LoZander: But maybe its better to just always find the most specific element and just backtrack later?

LoZander: In that case, then yeah, the StackConsumer makes perfect sense

LoZander: sorry for the text wall ^^'

LoZander: Am I just over complicating it again?

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
LoZander: yep

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
magnus-madsen: YAGNI. You are designing for what-ifs

magnus-madsen: In both cases you may consider something simpler than a stack: Just keep one element.

magnus-madsen:
> there is some weird behaviour when hovering on the return type of a def when the return type is an arrow type. Normally, we'd get the kind of the type when hovering on a type like that, but for some reason, while hovering on the return type in this specific case, it's as if we're hovering on the entire def. Since the visitor doesn't differentiate between these cases, I worry there might be some problem with the generation of the locations or something like that.

It sounds like you can start debugging.

magnus-madsen:
But perhaps you should do some extensive manual testing and come up with a list of bugs and then afterwards prioritize and address the most important.

magnus-madsen:
How do you find the hover speed?

magnus-madsen (commented):
Actionable items for PR:

- [ ] Extensive manual testing to ensure that we did not break anything. Make a bug list with priorities. 
- [ ] Extract StackConsumer into its own abstraction, so it is reusable. Possibly simplify it.
- [ ] Can we sort the cases in hoverAny? I am not exactly sure how, but right now seems too unstructured.
- [ ] Are there things you cannot hover, but you should be able to hover? (Goes back to testing).

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
```diff
@@ -28,56 +30,42 @@ import scala.annotation.tailrec
...
+  private def hoverAny(x: AnyRef, uri: String, pos: Position)(implicit root: Root, flix: Flix): JObject = x match {
+    case tpe: Type => hoverKind(tpe)
+    case exp: Expr => hoverTypeAndEff(exp.tpe, exp.eff, exp.loc)
+    case DefSymUse(sym, loc) => hoverDef(sym, loc)
+    case SigSymUse(sym, loc) => hoverSig(sym, loc)
+    case OpSymUse(symUse, loc) => hoverOp(symUse, loc)
+    case Def(sym, _, _, loc) => hoverDef(sym, loc)
+    case Sig(sym, _, _, loc) => hoverSig(sym, loc)
+    case Op(sym, _, loc) => hoverOp(sym, loc)
+    case FormalParam(_, _, tpe, _, loc) => hoverType(tpe, loc)
```
magnus-madsen, line 67: Would this case not already be covered by the tpe: Type case? I thought that was part of the advantage of the visitor.

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
```diff
@@ -28,56 +30,42 @@ import scala.annotation.tailrec
...
-    case Entity.TypeAlias(_) => mkNotFound(uri, pos)
-    case Entity.AssocType(_) => mkNotFound(uri, pos)
-    case Entity.Label(_) => mkNotFound(uri, pos)
-    case Entity.Op(_) => mkNotFound(uri, pos)
-    case Entity.Sig(_) => mkNotFound(uri, pos)
-    case Entity.TypeVar(_) => mkNotFound(uri, pos)
+  private def hoverAny(x: AnyRef, uri: String, pos: Position)(implicit root: Root, flix: Flix): JObject = x match {
+    case tpe: Type => hoverKind(tpe)
+    case exp: Expr => hoverTypeAndEff(exp.tpe, exp.eff, exp.loc)
+    case DefSymUse(sym, loc) => hoverDef(sym, loc)
```
magnus-madsen, line 61: What's the argument to support both Def and DefSymUse? I think that hover on a def, when it is right there, should not show the same information...

magnus-madsen:
Lets iterate quickly on this so we can get some experience with the new visitor.

magnus-madsen:
!!!jar

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
magnus-madsen: But I think it would be best with a stack that can consume everything. That seems very reusable.

magnus-madsen:
This program is good to test on:

```
///
/// We can define pure functions that use
/// internal mutability (impurity) with regions.
/// Regions encapsulate mutability to its declared scope.
///
def deduplicate(l: List[a]): List[a] with Order[a] =
    // Declare a new region `rc`.
    region rc {

        // Create a new `MutSet` at region `rc`.
        // This will be used to keep track of
        // unique elements in `l`.
        let s = MutSet.empty(rc);

        // The lambda used in the call to `filter`
        // would be impure without a region.
        List.filter(x -> {
            if (MutSet.memberOf(x, s))
                false // `x` has already been seen.
            else {
                MutSet.add!(x, s);
                true
            }
        }, l)
    }

///
/// Create a list `l` with duplicates and
/// call `deduplicate` that returns a new list
/// with only unique elements.
///
def main(): Unit \ IO =
    let l = 1 :: 1 :: 2 :: 2 :: 3 :: 3 :: Nil;
    println(deduplicate(l))
```

magnus-madsen:
It reveals several issues, including hovering on literals and on variables.

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
LoZander: I thought that would be expected. But I can see your point that it's redundant. Alternatively, howering on the definition itself, could do nothing or show like the arrow type of the function (although this is also somewhat redundant as it can easily be gleaned from the signature)

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
LoZander: Howering on a type and hovering on a formal param is not quite the same, though you might be right that there is little reason to support both. For some formal param `x: T`, hovering on `x` is hovering on the formal param, but not the type `T`. If we only support `tpe: Type`, then hovering on the `x` itself, would do nothing. So the FormalParam case does add *something*. Right now, hovering on `x: T` but *not* `T` would give the type of `x` (T itself), which is arguably redundant, as it's a repition of the information you're hovering on. Hovering on the `T` itself, gives the kind of `T`.

We could remove FormalParam, I think that might be reasonable to do, since it doesn't really add anything. I added it based on the current behaviour of the hover provider. But maybe I shouldn't base this new provider off of the behaviour of the old. Better to rethink it entirely, perhaps?

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
LoZander: Yeah, I think you're right. That is probably better

LoZander: Should I add this in a new PR since its a change in HoverProvider.scala, or is it fine to do it as a part of this PR?

LoZander:
> It reveals several issues, including hovering on literals and on variables.

I've been looking at the example, and I see what you mean. I think the issue lies with whether locations have `isReal = true` or `isReal = false`. Inspecting how hovering behaved, I noticed that lambdas seemed to just not be visited at all. I thought that maybe it was the `isReal` check in `inside` that filtered it out. I commented the line out, and now not only are lambdas properly visited, but so are the literals and variables.

This leads me to believe that either things that should have `isReal = true` have `isReal = false` or the `isReal` check in `inside` is simply wrong. Maybe `isReal` doesn't simply denote whether a location corresponds to an actual place in the code? For now, the simplest solution is to just remove that `isReal` check in `inside`

LoZander:
For now, I've commented out the the `isReal` check in inside and set the corresponding test to be ignored. If the check all together is misguided, then maybe a fix PR is warranted

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
magnus-madsen: It seems redundant to me.

magnus-madsen:
Let's do more focused PRs. It seems this PR is now covering many different things.

magnus-madsen:
!!!jar

magnus-madsen:
I talked to @mlutze and we (tentatively) think it makes sense to include source location no matter whether they are real or synthetic.

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
```diff
@@ -97,6 +97,65 @@ object Visitor {
     def consumeTypeParam(tparam: TypeParam): Unit = ()
   }
 
+  case class StackConsumer() extends Consumer {
```
magnus-madsen, line 100: I think this should go in a separate file.

Probably we should also more the Consumer and Acceptor to separate files.

magnus-madsen: Also need some documentation of how this works.

LoZander:
> Let's do more focused PRs. It seems this PR is now covering many different things.

Sure. I can wrap this up. The most basic hover provider functionality works now. Then I can do a new PR for structs and enums etc. Something like that? I could also try to split the work already done here into multiple PRs, but I think a lot of it is too interconnected for that to work well

main/src/ca/uwaterloo/flix/api/lsp/provider/HoverProvider.scala
LoZander: yep, Def, Sig and Op hover has been removed

LoZander:
Ill try to make future PRs smaller :)

magnus-madsen:
Please break this PR into smaller pieces too. Thanks!

LoZander:
> Please break this PR into smaller pieces too. Thanks!

Alright, I will. I think, there will naturally be some dependency between some of these PRs though. I can do a PR for the a) stack consumer, b) the hoverprovider and c) the refactoring of ASTs, but a) will depend on b) which will depend on c). Is that fine?

LoZander:
or maybe it's better to just break it into stack consumer PR and a couple of PRs for the hover provider?

LoZander:
Breaking PR into:
- #8934

LoZander:
> I talked to @mlutze and we (tentatively) think it makes sense to include source location no matter whether they are real or synthetic.

should I also completely remove the test relating to realness of source locations, or should I just set it to be ignored?

magnus-madsen:
I would remote the test

magnus-madsen:
Remove

Issue #8815, opened 2024-10-07
Refactor return type of phases


magnus-madsen:
Here is the signature of our good friend `Kinder`:

```scala
def run(root: ResolvedAst.Root, oldRoot: KindedAst.Root, changeSet: ChangeSet)(implicit flix: Flix): 
    Validation[KindedAst.Root, KindError] = flix.phase("Kinder") { ... } (DebugValidation())
```

I would like this to become:

```scala
def run(root: ResolvedAst.Root, oldRoot: KindedAst.Root, changeSet: ChangeSet)(implicit flix: Flix): 
     (KindedAst.Root, List[KindError]]) = flix.phase("Kinder") { ... }
```

This needs a few changes. Most importantly, we can stop using Monadic style in Flix. Instead, something like:

```scala
val errors: mutable.ListBuffer[CompilationError] = mutable.ListBuffer.empty

// previous phase
val (afterKinder, kinderErrors) = Kinder.run(afterResolver, cachedKinderAst, changeSet)
errors += kinderErrors
// next phase
```

I think this refactoring can be done gradually (i.e. phase by phase). In particular, we can keep the giant `for` during refactoring because a `for` can contain `val` decls in Scala (I think).

We might need a temporary new `flix.phase` command:

```scala
def phaseNew[A, B](phase: String)(f: => (A, B))(implicit d: Debug[A]): (A, B) = ...
```

EDIT: If a phase does not change an AST I think it should just return the list of errors.

magnus-madsen:
@jaschdoc Does it make sense?

magnus-madsen:
@jaschdoc I did an example PR here: https://github.com/flix/flix/pull/8816 to show the pattern.

jaschdoc:
Yes looks good :)

jaschdoc:
What about changing the return type to `(XAst.Root, Iterable[XError])` so we avoid the `toList` on `sctx.error.asScala.toList`? Maybe it's not that expensive really since most of the time it's just `Nil` but it's free performance

magnus-madsen:
> What about changing the return type to `(XAst.Root, Iterable[XError])` so we avoid the `toList` on `sctx.error.asScala.toList`? Maybe it's not that expensive really since most of the time it's just `Nil` but it's free performance

No fucking way.

magnus-madsen:
[...]

PR #8873, opened 2024-10-13
refactor: remove `Validation` from `PredDeps`

jaschdoc:


main/src/ca/uwaterloo/flix/api/Flix.scala
```diff
@@ -522,8 +522,8 @@ class Flix {
       (_, regionErrors) = Regions.run(afterTyper)
       afterEntryPoint <- EntryPoint.run(afterTyper).withSoftFailures(regionErrors)
       _ <- Instances.run(afterEntryPoint, cachedTyperAst, changeSet)
-      afterPredDeps <- PredDeps.run(afterEntryPoint)
-      afterStratifier <- Stratifier.run(afterPredDeps)
+      (afterPredDeps, predDepErrors) = PredDeps.run(afterEntryPoint)
+      afterStratifier <- Stratifier.run(afterPredDeps).withSoftFailures(predDepErrors)
```
jaschdoc, line 526: `predDepErrors` is always `Nil` since `PredDeps` never fails. Any idea in carrying `predDepErrors` around then? Should we ever introduce errors into the phase, it would automatically be picked up by this, but enforcing that it only return `Root` makes it impossible to miss anything. Just thoughts

magnus-madsen (approved):

PR #8878, opened 2024-10-13
refactor: remove `Validation` internally from `Lexer`

jaschdoc:


magnus-madsen:
Can we avoid all the doc comment reformatting? 

We should fix such things separately, but probably not before we have a style checker in place.

magnus-madsen (approved):

PR #8880, opened 2024-10-13
refactor: use `phase` in phases that return errors

jaschdoc:
Uses the old `phase` instead of `phaseNew` to avoid redundant let-matches. Makes the code more readable / less cluttered.


magnus-madsen:
My plan is to get rid of old phase and only have new phase -- which should be renamed.

jaschdoc:
> My plan is to get rid of old phase and only have new phase -- which should be renamed.

But isn't it sort of awkward to always return a tuple? I think it's prettier to only return one thing when there is only one result of the computation

magnus-madsen:
> > My plan is to get rid of old phase and only have new phase -- which should be renamed.
> 
> But isn't it sort of awkward to always return a tuple? I think it's prettier to only return one thing when there is only one result of the computation

Possibly. But I did code duplication in newPhase/phase so that would have to be fixed.

magnus-madsen:
I suggest we do the Unit, Errors thing first and then refactor later.

jaschdoc:
> I suggest we do the Unit, Errors thing first and then refactor later.

Roger that

PR #8885, opened 2024-10-14
refactor: use `phase` in `phaseNew`

jaschdoc:
Here is how I would refactor `phaseNew` to use `phase` so we can keep both. Note that I also updated the parameter names to avoid shadowing the function name, e.g., `phase(phase)(f)` is not resolved correctly and that I renamed `phaseNew` to `phase2` implying that both functions should be there with different functionality

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/Flix.scala
+++ b/main/src/ca/uwaterloo/flix/api/Flix.scala
@@ -618,7 +618,7 @@ class Flix {
   /**
     * Enters the phase with the given name.
     */
-  def phaseNew[A, B](phaseName: String)(f: => (A, B))(implicit d: Debug[A]): (A, B) = {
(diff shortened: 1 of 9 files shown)


magnus-madsen:
Can you come up with a better name and document better? That would help.

jaschdoc:
Better name for `phase2`? :)

jaschdoc:
> Remind me again, what is the benefit to having two phase functions?

1. We drop errors that do not exist for phases that cannot fail (like `Reader` and `Desugar`). This is nice since it shows that the phase does not produce any errors
2. Clean up the syntax a bit more:
```scala
val (afterReader, _) = Reader.run(getInputs, knownClassesAndInterfaces)
```
becomes
```scala
val afterReader = Reader.run(getInputs, knownClassesAndInterfaces) // it now uses phase internally so it only returns 1 value
```

3. Nice code reuse

PR #8904, opened 2024-10-15
fix: printing of dbg keyword

mlutze:


mlutze:
Fixes this:

```
>> Expected <Name>, <greek name>, <name>, <user-defined operator> or <math name> before 'debug'.

8 |     pub def dbg(x: a): a with ToString[a] = unchecked_cast({
                ^^^
                Here
Hint: 'debug' is a keyword.
```

magnus-madsen (approved):

PR #8907, opened 2024-10-15
refactor: turn import completions into semantic completions

### LINKED ISSUE #8758: Refactor `ImportCompleter` to use semantic information
The `ImportCompleter` currently uses a regex like: `raw"\s*import\s+(?:.*\s+)*(.*)".r`.

The goal instead should be that it takes the error `UndefinedJvmClass` and uses that information directly.

sockmaster27:
Fixes #8758


sockmaster27:
This splits completions into syntactic completions and semantic completions, where the syntactic are based on the current syntactic context and the semantic are based on the error at the current position.

@magnus-madsen Was this what you were thinking?

magnus-madsen:
Yes, but I think I want smaller steps. Can we do one that only touches ImportCompleter?

sockmaster27:
Not strictly, since we need to pass the error-info through somehow.

magnus-madsen:
I will have a more detailed look later then. I just felt like too many things were going on, but perhaps they are all necessary.

magnus-madsen:
@sockmaster27 Conflicted. Also can we break it into smaller steps? Or alternatively explain each refactoring.

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -194,19 +206,14 @@ object CompletionProvider {
...
+    errorsAt(uri, pos, errors).map({
       // We can have multiple errors, so we rank them, and pick the highest priority.
       case WeederError.UnqualifiedUse(_) => (1, SyntacticContext.Use)
-      case ResolutionError.UndefinedJvmClass(_, _, _) => (1, SyntacticContext.Import)
       case ResolutionError.UndefinedName(_, _, _, isUse, _) => if (isUse) (1, SyntacticContext.Use) else (2, SyntacticContext.Expr.OtherExpr)
       case ResolutionError.UndefinedNameUnrecoverable(_, _, _, isUse, _) => if (isUse) (1, SyntacticContext.Use) else (2, SyntacticContext.Expr.OtherExpr)
       case ResolutionError.UndefinedType(_, _, _) => (1, SyntacticContext.Type.OtherType)
       case ResolutionError.UndefinedTag(_, _, _) => (1, SyntacticContext.Pat.OtherPat)
       case ResolutionError.UndefinedOp(_, _) => (1, SyntacticContext.Expr.Do)
-      case WeederError.MalformedIdentifier(_, _) => (2, SyntacticContext.Import)
```
magnus-madsen, line 209: Why is it safe to remove this line?

magnus-madsen (approved):

magnus-madsen:
@sockmaster27 I approve, but unfortunately some conflict snuck in.

magnus-madsen (approved):

PR #8924, opened 2024-10-18
fix: allow importing ops to shadow definitions

### LINKED ISSUE #8600: Cannot `use` effect operations
```
use Console.println
def example(): Unit \ Console =
    do println("hey")
```

mlutze:
fixes #8600


magnus-madsen (approved):

magnus-madsen:
This is unrelated to getting rid of `do`?

PR #8931, opened 2024-10-18
feat: remove `do` keyword from frontend

magnus-madsen:


mlutze (approved):

magnus-madsen:
@mlutze @JonathanStarup @stephentetley We no longer need `do` in front of a call to an effect operation.

In fact this PR removes `do`. Could you update your community packages? All you need is to go from `do F.f` to `F.f`

stephentetley:
Thanks for the heads up Magnus.

Is Community Build looking at the head of the repository or the latest release tarball? I've updated both, but I doth think I was using do in the previously release version.

magnus-madsen:
> Thanks for the heads up Magnus.
> 
> Is Community Build looking at the head of the repository or the latest release tarball? I've updated both, but I doth think I was using do in the previously release version.

I think it uses the latest master. (so one does not have to create new releases).

magnus-madsen:
@stephentetley 
On a related note, I am hoping #8848 lands this week. This master will support proper set effects - which should enable effects and handlers. The documentation has already been updated here: https://doc.flix.dev/effect-system.html

magnus-madsen:
@mlutze @JonathanStarup I am merging this. Please update your packages :)

PR #8936, opened 2024-10-19
feat: hover provider using LSP visitor

LoZander:
This PR refactors the HoverProvider to use the LSP visitor instead of the index. In this initial PR, only basic hover functionality is supported.

Related to #8639


LoZander:
This suffers from issues relating to #8935, so that has to go through before a proper review of this PR

magnus-madsen:
@LoZander 
I think we can merge this. Anything you want to change before merge?

magnus-madsen:
!!!jar

magnus-madsen:
It seems there is at least one issue to be fixed: I cannot hover on a variable bound by a let.

magnus-madsen:
I also cannot hover on a formal parameter.

LoZander:
> I also cannot hover on a formal parameter.

Isnt it enough to be able to just hover on its type? What should hovering on the parameter as a whole tell you when its type is already right there? Maybe theres some other info about the param that it should tell you?

LoZander:
> It seems there is at least one issue to be fixed: I cannot hover on a variable bound by a let.

Do you mean hovering on the variable in the bind itself, like for `let b = ...` hovering on `b` or do you mean hovering on `b` in some following expression or statement?

magnus-madsen:
I want to hover on variables and formal parameters to see their types.

magnus-madsen:
This includes binders in lambdas

LoZander:
The complication for the variable in `let x = ...` is that there is no AST node for it. In `Expr.Let`, it's just a variable symbol. So that's why there's a case in `hoverAny` for `Expr.Let` with a guard checking manually if you're on the symbol. Not the cleanest, but it works. The new FormalParam case also covers hovering on lambda params

LoZander:
Is it confusing that hovering on the type annotation for formal params gives the kind of the type? It's simply always the case that hovering on an explicit type gives you the kind, but I could see it being possibly misleading here. What do you think?

LoZander:
Hovering is not yet supported for everything. Stuff like struct and record fields is not supported yet, but I'm thinking things like that will be done in follow up PRs

LoZander:
there is one more thing. Currently the visitor visits the regionVar of a region scope. I think this is a mistake in the visitor and it affects the result of hovering on a region scope. I want to fix that now, but does it fall outside of the scope of this PR?

magnus-madsen:
!!!jar

magnus-madsen:
> The complication for the variable in `let x = ...` is that there is no AST node for it. 

For what exactly? For VarSym?

> In `Expr.Let`, it's just a variable symbol. So that's why there's a case in `hoverAny` for `Expr.Let` with a guard checking manually if you're on the symbol. Not the cleanest, but it works. 

It seems to me this break the abstractions we just created.

magnus-madsen:
I think some more testing is warranted. Try to take this program:

```
/// An algebraic data type for shapes.
enum Shape {
    case Circle(Int32),          // circle radius
    case Square(Int32),          // side length
    case Rectangle(Int32, Int32) // height and width
}

/// Computes the area of the given shape using
/// pattern matching and basic arithmetic.
def area(s: Shape): Int32 = match s {
    case Shape.Circle(r)       => 3 * (r * r)
    case Shape.Square(w)       => w * w
    case Shape.Rectangle(h, w) => h * w
}

// Computes the area of a 2 by 4.
def main(): Unit \ IO =
    println(area(Shape.Rectangle(2, 4)))
```

Does the hovering always work as you expect?

LoZander:
I've run into an issue. A varSym tells you very little. When hovering on a varSym, I'd except the type of the variable. But the var sym doesn't contain that information, so you'll have to backtrack on the stack to find it's parent element, and then depending on what kind of element that is, determine the type. That's kind of a mess since we're operating on AnyRef

LoZander:
oh, but we do have the `BoundBy`, which can at least tell us the type of AST node the parent is

magnus-madsen:
What if we had: consumeVarBinder(sym: Symbol.VarSym, tpe: Type)?

LoZander:
I think that would work. But then we're really starting to diverge from the interface of all the other consumers. But maybe that's okay, at least for now?

LoZander:
I mean, we could do it. It would work, at least if the type of the bound value is the only thing ever of interest that's not contained in the var sym. Do you think that's what we should do?

It just feels like we're creating a pseudo AST node on the spot . If we really want the type of the bound value alongside the var sym, shouldn't the two just already be packeded in some AST node in place of these occurances of varSyms in the AST?

LoZander:
It feels like we're working around a problem in the AST

magnus-madsen:
Add the type argument now. Then we can add a new AST node later.

LoZander:
Alright, sounds good :)

LoZander:
Hovering on varSyms should work now :)

LoZander:
Enum, Record and Struct related hovering still doesn't work. I'll do those in follow up PRs :)

magnus-madsen:
!!!jar

magnus-madsen (approved):

magnus-madsen:
🚀

PR #8940, opened 2024-10-20
refactor: drop usage of BDDs

magnus-madsen:
Related #8939


magnus-madsen:
@mlutze This is the first in a series of PRs I plan to make.

mlutze (approved):

Issue #8813, opened 2024-10-07
Refactor `Validation` to remove soft failures


magnus-madsen:
When #8540, #8811, and #8810 are done, we can refactor `Validation`:

We basically turn back the clock and make it as it was. Specifically, it becomes like Validation in Flix. That is:

- [x] Remove `case class SoftFailure[T, E](t: T, errors: Chain[E])`
- [x] Remove `withSoftFailure`
- [x] Remove `withSoftFailures`
- [ ] Merge `toSoftResult` and `toHardResult`
- [ ] Remove `toSoftFailure` (all uses should be gone by this point)
- [x] Remove `toSuccessOrSoftFailure` (all uses should be gone by this point)
- [ ] Rename `HardFailure` to simply `Failure`
- [ ] Remove `Recoverable` and `Unrecoverable`
- [ ] Remove implicit class `AsMonad`

Note that the above tickets do not have to be fully done. We just need to have all uses of `toSoftFailure` eliminated.

PR #8944, opened 2024-10-21
refactor: remove `SoftFailure`

jaschdoc:
Just making this PR to keep track of progress towards removing soft failure

Related to https://github.com/flix/flix/issues/8813


magnus-madsen:
I think it is just one place in Typer and then Weeder. Typer can be refactored by returning a pair (Thing, List[Error]) whereas Weeder can use the trick of adding a shared context and using that __before__ removing Validation from it (since that seems tricky).

magnus-madsen:
This PR is gonna feel good :rocket:

magnus-madsen:
What is left:

- Typer.checkAssocTypes
- ConstraintSolver.visitSpec

main/src/ca/uwaterloo/flix/api/Flix.scala
```diff
@@ -540,7 +541,24 @@ class Flix {
         this.cachedResolverAst = afterResolver
         this.cachedTyperAst = afterTyper
       }
-      output
+      errors ++= readerErrors
```
magnus-madsen, line 544: This only gets you the errors on the happy path, but if resolver fails you lose any errors that occurred earlier. 

I think you need to get rid of the for-comprehension and then add the errors to the error set as you go.

jaschdoc: This is not actually the issue, but I guess it is better to remove the for-comprehension now rather than later

jaschdoc: The issue was that it now returned `HardFailure` only containing errors when the parser recovery tests assumed that it would get a soft failure which contains a root

main/src/ca/uwaterloo/flix/util/Validation.scala
```diff
@@ -113,47 +79,30 @@ object Validation {
...
-          case Validation.SoftFailure(value, thatErrors) => Validation.SoftFailure(value, errors ++ thatErrors)
-          case Validation.HardFailure(thatErrors) => Validation.HardFailure(errors ++ thatErrors)
-        }
-
         case Validation.HardFailure(errors) => Validation.HardFailure(errors)
       }
     }
+
+    implicit class AsHardFailure[T, E](v: (Validation[T, E], Iterable[E])) {
+      final def toHardFailure: Validation[T, E] = v match {
```
jaschdoc, line 89: Added this to avoid lots of changes to call sites.

main/src/ca/uwaterloo/flix/api/Flix.scala
magnus-madsen: Kind of confused. Does this not make the parser tests less valuable or?

magnus-madsen: Perhaps there is no alternative?

main/src/ca/uwaterloo/flix/util/Validation.scala
```diff
@@ -168,15 +117,10 @@ object Validation {
   /**
     * Represents a success `value`.
     */
-  private case class Success[T, E](t: T) extends Validation[T, E] {
+  case class Success[T, E](t: T) extends Validation[T, E] {
```
jaschdoc, line 120: Made this public again. It is easier to pattern match directly on this instead of `Result` when the constructors are equivalent. Also, this is a step in the direction of refactoring `Validation` later to be what it was before `SoftFailure`.

main/src/ca/uwaterloo/flix/api/Flix.scala
jaschdoc: From the documentation of `TestParser`

> Note that these tests use check rather than compile.
> That's because a compile converts any check failure into a HardFailure before running, codegen so the result we would like to expect is lost.

main/src/ca/uwaterloo/flix/api/Flix.scala
```diff
@@ -485,84 +487,126 @@ class Flix {
   /**
     * Compiles the Flix program and returns a typed ast.
     */
-  def check(): Validation[TypedAst.Root, CompilationMessage] = try {
-    import Validation.Implicit.AsMonad
+  def check(): (Validation[TypedAst.Root, CompilationMessage], List[CompilationMessage & Recoverable]) = {
```
magnus-madsen, line 490: I think when making big changes like this it is better to discuss first.

magnus-madsen: I can imagine two better signatures:

- Validation[(TypedAst.Root, Message), Message]
- (Option[TypedAst.Root], List[Message])

main/src/ca/uwaterloo/flix/api/Flix.scala
jaschdoc: So they want to assert that the front-end succeeded but with errors whereas there are other tests that assert that `compile` fails. However, `compile` removes the `root` if the front-end fails

main/src/ca/uwaterloo/flix/api/Flix.scala
magnus-madsen: What are your thoughts on these?

jaschdoc: This is also just a proposal.

I see a couple of options:
a) This solution: return a Validation along with recoverable failures which the caller can decide what to do with. Only return Hard failure if the front-end produced one.
b) Only return success if no error was present at all (then we should probably add a separate test function for `TestParser`)

magnus-madsen: I don't understand why it is just TestParser- should all the frontend test not just call check?

magnus-madsen: And what about the convenience functions we have, e.g. expectError etc?

magnus-madsen: I have to, but could you write up a small blob of text to outline the problem and discuss the trade-offs of the different signatures?

jaschdoc: Most tests just call `compile` because will catch the expected error anyway. The way I understand it is that the intention was that now that the parser is resilient, we should check that the front-end can succeed. Of course we should probably always check that but since most tests only really care about producing the expected error it does not really make a difference.

magnus-madsen: OK, but what are the implications of that?

magnus-madsen:
I propose to split this PR into two: one that refactors flix.check and one that later removes soft failure.

PR #8952, opened 2024-10-21
feat: extension to LSP visitor to visit and consume var bindings

LoZander:
This PR extends the LSP visitor to also visit `VarSym`, specifically where these denote a binding of a variable.

Related to #8639


magnus-madsen (approved):

magnus-madsen:
@LoZander Merged! Lets continue.

PR #8956, opened 2024-10-21
fix: changed consumeVarBinder to also take the type of the bound value

LoZander:
This PR changes `consumeVarBinder` to also take the type of the bound value.

Related to #8639


LoZander:
Some versioning weirdness occured where the impl of consumerVarBinder was not implemented in the code that this built on, so there the change created no local issue but in the remote, it does ofc. break that impl. I'll fix that in a new PR

LoZander:
New PR #8957

PR #8957, opened 2024-10-21
fix: changed consumeVarBinder to also take the type of the bound value

LoZander:
This PR changes consumeVarBinder to also take the type of the bound value.

Related to https://github.com/flix/flix/issues/8639


magnus-madsen (approved):

LoZander:
Oops, missed something in stackconsumer

Issue #8955, opened 2024-10-21
Introduce `TypedAst.Binder`


magnus-madsen:

magnus-madsen:
Fixed in master.

PR #8958, opened 2024-10-21
refactor: introduced `TypedAst.Binder`

magnus-madsen:
Related #8955

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/Indexer.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/Indexer.scala
@@ -251,8 +251,8 @@ object Indexer {
     case Expr.Binary(_, exp1, exp2, _, _, _) =>
       visitExp(exp1) ++ visitExp(exp2) ++ Index.occurrenceOf(exp0)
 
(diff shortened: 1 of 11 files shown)


magnus-madsen:
@mlutze This seems reasonable, what do you think?

magnus-madsen:
(Obv. more needs to be refactored- but I was going to let Alexander do that)

mlutze:
Is the type necessary? It may be confusing, as one's assumption may be that it corresponds to the `Foo` of `let x: Foo = ...`

mlutze (approved):

magnus-madsen:
> Is the type necessary? It may be confusing, as one's assumption may be that it corresponds to the `Foo` of `let x: Foo = ...`

It is because we need the type for e.g. hovering, but the VarSym only has the type variable.

magnus-madsen:
@LoZander Please take a look at the diff here and then refactor to use `Binder` for all occurrences of `VarSym` in `TypedAst`. 

Makes sense?

LoZander:
> @LoZander Please take a look at the diff here and then refactor to use `Binder` for all occurrences of `VarSym` in `TypedAst`.
> 
> Makes sense?

Yep, that makes sense :)

PR #8966, opened 2024-10-23
refactor: generic `CofiniteSet`

JonathanStarup:


magnus-madsen:
Why does it need to be generic?

magnus-madsen:
YAGNI?

JonathanStarup:
Yagni? We have both `CofiniteIntSet` and `CofiniteEffSet` which is bad for code maintenance

magnus-madsen:
OK; then lets do it! - Add the generic one

PR #8967, opened 2024-10-23
feat: add `App.run`

magnus-madsen:


magnus-madsen:
We can extend the set of handled effects later. For now, I just wanted to handle those that are only in IO.

mlutze (approved):

PR #8983, opened 2024-10-24
refactor: `BoolAlg` independent of `BoolFormula` and `Type`

magnus-madsen:


magnus-madsen:
This is also just a refactoring.

mlutze (approved):

PR #8992, opened 2024-10-24
refactor: introduce TraitEnv

mlutze:
This will allow us to treat the trait environment more abstractly. Specifically, we need to manipulate it by adding constraints, etc., but the implementation should remain hidden.


magnus-madsen (approved):

magnus-madsen:
@mlutze Sorry, conflicted.

PR #8997, opened 2024-10-24
NeverMerge: track effect variables

JonathanStarup:


magnus-madsen:
This is the paper PR right? So its OK if I close?

PR #8998, opened 2024-10-24
feat: Annotate `Resolver` lambdas pure

### LINKED ISSUE #8653: Annotate generated exps
In resolver fx we create lambda aliases for defs. these have known types and effects so it would help errors and performance to put these known types into the AST (is my conjecture)

JonathanStarup:
Most lambdas created in `Resolver` must be pure, so know the type will also know that. This should gain some (perhaps not much) typer performance and will allow better subeffecting for lambdas.

Fixes #8653


magnus-madsen (approved):

magnus-madsen:
Let's see what this does. It is at least a good idea 👍

PR #8999, opened 2024-10-24
feat: Optimize generated lambda subeffecting

JonathanStarup:
Based on #8998, now we can omit some redundant subeffecting on lambdas (can be merged independently)


magnus-madsen:
When does this happen? Like when would this trigger?

JonathanStarup:
From the lambdas that Resolver creates

magnus-madsen (approved):

PR #9013, opened 2024-10-25
feat: ignore redundant `checked_ecast`

JonathanStarup:
if lambda subeffecting is enabled, `x -> checked_ecast(x)` is typed as `x -> x`. The same for defs and instances

Even though this seems minor, I think its pretty important


JonathanStarup:
@magnus-madsen After this I can pull the numbers and they will not have any noise (afaik)

magnus-madsen (approved):

PR #9021, opened 2024-10-25
proposal: use stable order of effect variables

magnus-madsen:
This PR demonstrates that the problem is in unstable ordering of effect variables during translation from types to atoms to numbers.

With this PR I can run: 

```
Xperf --par --n 51 --frontend --Xsubeffecting lambdas
```

which does consume a ton of memory (at least 9GB), but it returns with:

```

Throughput (best): 19.350 lines/sec (with 24 threads.)

  min:  8.903, max: 19.350, avg: 18.124, median: 18.354

Finished 51 iterations on 67.196 lines of code in 191 seconds.
```

The PR is an ugly hack. It does two things:

- It hacks the Bimap creation to ensure that stable IDs are used for the effect variables.
- It ensure that subeffect variables are given very low (negative numbers), hence they are solved "first".

I think before we attempt any other performance trick, we need to resolve the issue about stable numbering. Otherwise we are essentially always operating under adverse scheduling. 

@JonathanStarup @mlutze Does it make sense?

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/Type.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/Type.scala
@@ -666,6 +666,19 @@ object Type {
   }
 
   /**
+    * Returns a fresh type variable of the given kind `k` and rigidity `r`.
+    */
+  def freshSubEffectVar(loc: SourceLocation)(implicit scope: Scope, flix: Flix): Type.Var = {
+    import ca.uwaterloo.flix.language.ast.Symbol.KindedTypeVarSym
+    val offset = -10_000_000
+    val text = Ast.VarText.Absent
+    val kind = Kind.Eff
+    val isRegion = false
+    val sym = new KindedTypeVarSym(flix.genSym.freshId() + offset, text, kind, isRegion, scope, loc)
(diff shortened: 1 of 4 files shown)


magnus-madsen:
Master perf:

```

Throughput (best): 70.650 lines/sec (with 24 threads.)

  min: 13.654, max: 70.650, avg: 64.732, median: 66.944

Finished 51 iterations on 67.196 lines of code in 56 seconds.
```

Slowdown (median): **3.6x**

magnus-madsen:
Reversing the variable resolution order boosts performance to:

```

Throughput (best): 78.839 lines/sec (with 24 threads.)

  min: 13.983, max: 78.839, avg: 70.285, median: 72.693

Finished 51 iterations on 67.196 lines of code in 52 seconds.
```

The new baseline is

```

Throughput (best): 86.066 lines/sec (with 24 threads.)

  min: 14.207, max: 86.066, avg: 73.625, median: 76.299

Finished 51 iterations on 67.196 lines of code in 50 seconds.
```

For a median slowdown of: **1.05x**

mlutze:
Cool

magnus-madsen:
!!!jar

PR #9028, opened 2024-10-27
feat: improve performance of unification in `Monomorph`

magnus-madsen:


magnus-madsen:
@mlutze This has a decent impact, so it is worth merging, but coming up with a more clean solution would be nice.

mlutze (approved):

PR #9030, opened 2024-10-27
refactor: integrating Binder in TypedAst

LoZander:
This PR integrates `TypedAst.Binder`, replacing occurances of `Symbol.VarSym` that "binds" in `TypedAst`. 

Related to #8955.


LoZander:
There is one remaining issue, the way I see it. In `Predicate.Functional`, I don't know if the list of VarSym should be a list of Binders

LoZander:
I think it can be done, we'd just have to inspect the type of the exp of the functional predicate to see if it's a tuple type and then couple those to the var syms in binders. If there is a mismatch between the arity, ofc. that's already a broken program and this should have been caught before this.

main/src/ca/uwaterloo/flix/language/phase/TypeReconstruction.scala
```diff
@@ -680,7 +687,7 @@ object TypeReconstruction {
     */
   private def visitPattern(pat0: KindedAst.Pattern)(implicit subst: Substitution): TypedAst.Pattern = pat0 match {
     case KindedAst.Pattern.Wild(tvar, loc) => TypedAst.Pattern.Wild(subst(tvar), loc)
-    case KindedAst.Pattern.Var(sym, tvar, loc) => TypedAst.Pattern.Var(sym, subst(tvar), loc)
+    case KindedAst.Pattern.Var(sym, tvar, loc) => TypedAst.Pattern.Var(TypedAst.Binder(sym, tvar), subst(tvar), loc)
```
magnus-madsen, line 690: Everywhere need to do: subst(tvar)

magnus-madsen:
- Also LocalDef

main/src/ca/uwaterloo/flix/language/phase/TypeReconstruction.scala
LoZander: done

LoZander:
There is the quesiton of what should be the tpe of the binder for Scope. The variable binds the effect, right? Currently the binder takes the form Binder(sym, eff). But this stands out in a strange way, since every other Binder binds a type. Is this fine?

magnus-madsen (approved):

PR #9049, opened 2024-10-28
refactor: clearer US formatting in `Summary`

JonathanStarup:


JonathanStarup:
@magnus-madsen maybe we could consider this for stats in general. It seems pretty annoying that `f".."` is locale dependent

magnus-madsen (approved):

PR #9055, opened 2024-10-29
test: avoid `Options.Default` in tests (use `Options.DefaultTest` instead)

JonathanStarup:


magnus-madsen:
@JonathanStarup Test failures

magnus-madsen (approved):

PR #9057, opened 2024-10-29
proposal: SVE on table form

JonathanStarup:
frontend throughput with all subeffecting is 10% faster on my PC


magnus-madsen:
@JonathanStarup There is some conflict. If you resolve it and rebase/merge then I can run it on the server to see if there is any real impact.

magnus-madsen:
The conclusion was that this is faster, but some tests seem to fail.

JonathanStarup:
Since `tableForm` is actually tested, my current thinking is that it triggers the purification error. The failures look similar to the last time I did it with the variable bias.

magnus-madsen:
Do we just mark it as blocked then and leave it be?

PR #9058, opened 2024-10-29
feat: Zhegalkin to SetFormula

JonathanStarup:
now without superexponential blowup

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
@@ -189,7 +189,8 @@ object SetUnification {
   }
 
   /** Solves `eqs` with [[sve]], trying multiple different orderings to minimize substitution size. */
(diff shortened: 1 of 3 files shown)


magnus-madsen:
I suppose this helps, but how can I know if the translation is still too costly? (since it is not linear)

JonathanStarup:
to costly compared to what? They are both worst case `O(|vars| * 2 ^ |vars|)` big so it should be around the same size

magnus-madsen:
The translation is not linear, is what I meant.

JonathanStarup:
The runtime of the translation is `O((|vars| * 2 ^ |vars|)^2)` i think. but i think its the best we can do, going directly to set formula

magnus-madsen (approved):

magnus-madsen:
> The runtime of the translation is `O( (|vars| * 2 ^ |vars|) * (2^|vars|) )` i think. but i think its the best we can do, going directly to set formula

What I mean is that we should add `Xor` to SetFormula so the translation is one-to-one.

I can merge this PR and do experiments, but they will not really tell me whether it performs well, right?

PR #9059, opened 2024-10-29
feat: optimize easy `svePermutations` calls

JonathanStarup:
5% speedup for me. Not shortcircuiting requires checking the size of the substitution, that's why this is not just a fluff optimization.


magnus-madsen:
!!!jar

magnus-madsen:
Do we know if it is worth computing the permutations?

JonathanStarup:
> Do we know if it is worth computing the permutations?

With current settings we don't try any permutations. Since we only call `EffUnification3.unify` (one-by-one) it also doesn't do anything. (constant propagation can produce multiple output constraints, but they don't seem to hit SVE)

magnus-madsen:
Perhaps we should just remove all the code then?

JonathanStarup:
We could yeah, depends on how far away the unify-at-once solver is

magnus-madsen:
I am just going to drop the permutation idea. We never had any evidence that it could work.

Once we have the new solver, we can **inspect the equations and then determine if it would be useful -- based on actual data**.

PR #9060, opened 2024-10-29
proposal: add `SolverListener` throughout typing

JonathanStarup:
This is required to get total metrics for the whole of typing (without mutable state)

@mlutze approves this PR


magnus-madsen:
Let's keep this on a branch

JonathanStarup:
But it'd be very nice to have on master, then you can actually use the listener to get stats. Matt also says that it will be useful for the general solver (with more events)

magnus-madsen:
What do you need it for? Is it purely for debugging? What would a reusable design look like?

PR #9062, opened 2024-10-29
refactor: drop SVE permutations

magnus-madsen:


magnus-madsen:
The point here is to avoid code that is not used and has not even proved itself useful.

mlutze (approved):

PR #9065, opened 2024-10-30
feat: add `FilePath.handle`

### LINKED ISSUE #9032: Implement `handler` for `FilePath`
Implement a `handler` for the `FilePath` effect. (See e.g. `Clock` for an example).

The handler should return a `Result` so as to account for any thrown exception by the Java code.

sockmaster27:
Fixes #9032

### DIFF
--- a/main/src/library/App.flix
+++ b/main/src/library/App.flix
@@ -18,11 +18,10 @@ mod App {
     ///
     /// Runs the given function `f` handling the `Clock`, `Console`, `FilePath` and `Logger` effects.
     ///
-    pub def run(f: Unit -> a \ ef): a \ (ef - {Clock, Console, FilePath, Logger}) + IO =
+    pub def run(f: Unit -> a \ ef): a \ (ef - {Clock, Console, Logger}) + IO =
         let r = f |>
             Clock.handle |>
             Console.handle |>
-            FilePath.handle |>
             Logger.handle;
         r()
 
--- a/main/src/library/App.flix
+++ b/main/src/library/App.flix
@@ -16,12 +16,13 @@
 mod App {
 
     ///
-    /// Runs the given function `f` handling the `Clock`, `Console`, and `Logger` effects.
+    /// Runs the given function `f` handling the `Clock`, `Console`, `FilePath` and `Logger` effects.
     ///
-    pub def run(f: Unit -> a \ ef): a \ (ef - {Clock, Console, Logger}) + IO =
+    pub def run(f: Unit -> a \ ef): a \ (ef - {Clock, Console, FilePath, Logger}) + IO =
         let r = f |>
             Clock.handle |>
             Console.handle |>
+            FilePath.handle |>
             Logger.handle;
         r()
 
--- a/main/src/library/App.flix
+++ b/main/src/library/App.flix
@@ -16,7 +16,7 @@
 mod App {
 
     ///
-    /// Runs the given function `f` handling the `Clock`, `Console`, `FilePath` and `Logger` effects.
+    /// Runs the given function `f` handling the `Clock`, `Console`, and `Logger` effects.
     ///
     pub def run(f: Unit -> a \ ef): a \ (ef - {Clock, Console, Logger}) + IO =
         let r = f |>
--- a/main/src/library/FilePath.flix
+++ b/main/src/library/FilePath.flix
@@ -434,7 +434,7 @@ mod FilePath {
 
             def list(f, k) = {
                 try {
-                    // TODO: Use java.nio.file.Files.list, but it returns a generic Stream<Path> which Flix's type system currently doesn't support.
+                    // TODO: Use java.nio.file.Files.list when we have support for Java generics.
                     let file = new File(f);
                     k(Array.toList(file.list()))
                 } catch {
--- a/main/src/library/FilePath.flix
+++ b/main/src/library/FilePath.flix
@@ -158,6 +158,7 @@ pub eff FilePath {
 
 mod FilePath {
 
+    import java.io.File
     import java.io.IOException
     import java.lang.Class
     import java.lang.{String => JString}
@@ -431,15 +432,17 @@ mod FilePath {
                 }
             }
 
-//            def list(f, k) = {
-//                try {
-//                    k(FromJava.fromJava(Files.list(Paths.get(f))))
-//                } catch {
-//                    case ex: InvalidPathException => Err(IOError.Generic(ex.getMessage()))
-//                    case ex: IOException => Err(IOError.Generic(ex.getMessage()))
-//                    case ex: NotDirectoryException => Err(IOError.Generic(ex.getMessage()))
-//                }
-//            }
+            def list(f, k) = {
+                try {
+                    // TODO: Use java.nio.file.Files.list, but it returns a generic Stream<Path> which Flix's type system currently doesn't support.
+                    let file = new File(f);
+                    k(Array.toList(file.list()))
+                } catch {
+                    case ex: InvalidPathException => Err(IOError.Generic(ex.getMessage()))
+                    case ex: IOException => Err(IOError.Generic(ex.getMessage()))
+                    case ex: NotDirectoryException => Err(IOError.Generic(ex.getMessage()))
+                }
+            }
         }
 
         ///
--- a/main/src/library/FilePath.flix
+++ b/main/src/library/FilePath.flix
@@ -313,8 +313,8 @@ mod FilePath {
                     k(())
                 } catch {
                     case ex: InvalidPathException => Err(IOError.Generic(ex.getMessage()))
-                    case ex: IOException => Err(IOError.Generic(ex.getMessage()))
                     case ex: UnsupportedOperationException => Err(IOError.Generic(ex.getMessage()))
+                    case ex: IOException => Err(IOError.Generic(ex.getMessage()))
                 }
             }
 
@@ -329,8 +329,8 @@ mod FilePath {
                     k(())
                 } catch {
                     case ex: InvalidPathException => Err(IOError.Generic(ex.getMessage()))
-                    case ex: IOException => Err(IOError.Generic(ex.getMessage()))
                     case ex: UnsupportedOperationException => Err(IOError.Generic(ex.getMessage()))
+                    case ex: IOException => Err(IOError.Generic(ex.getMessage()))
                 }
             }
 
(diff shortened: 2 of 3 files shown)


sockmaster27:
@magnus-madsen 
Is this the correct approach

magnus-madsen:
Yes, but please use new Java IO (nio).

https://docs.oracle.com/javase/8/docs/api/index.html?java/nio/file/Path.html
https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#exists-java.nio.file.Path-java.nio.file.LinkOption...-

main/src/library/FilePath.flix
```diff
@@ -155,3 +155,25 @@ pub eff FilePath {
...
+    ///
+    /// Handles the `FilePath` effect of the given function `f`.
+    ///
+    /// In other words, re-interprets the `FilePath` effect using the `IO` effect.
+    ///
+    pub def handle(f: a -> b \ ef): a -> b \ (ef - FilePath) + IO = x ->
+        try {
+            f(x)
+        } with FilePath {
+            def exists(f, k) = {
```
magnus-madsen, line 172: Actually this is not totally correct. You need a try-catch around the block starting on this line.

Then you need to return a Result[FileError, b].

main/src/library/App.flix
```diff
@@ -16,12 +16,13 @@
...
     ///
-    /// Runs the given function `f` handling the `Clock`, `Console`, and `Logger` effects.
+    /// Runs the given function `f` handling the `Clock`, `Console`, `FilePath` and `Logger` effects.
     ///
-    pub def run(f: Unit -> a \ ef): a \ (ef - {Clock, Console, Logger}) + IO =
+    pub def run(f: Unit -> a \ ef): a \ (ef - {Clock, Console, FilePath, Logger}) + IO =
         let r = f |>
             Clock.handle |>
             Console.handle |>
+            FilePath.handle |>
```
magnus-madsen, line 25: We can worry about this later, I would lave it out for now. It will cause issues with the Result return type.

main/src/library/FilePath.flix
```diff
@@ -155,3 +155,188 @@ pub eff FilePath {
...
+    import java.nio.file.Path
+
+    ///
+    /// Handles the `FilePath` effect of the given function `f`.
+    ///
+    /// In other words, re-interprets the `FilePath` effect using the `IO` effect.
+    ///
+    pub def handle(f: a -> b \ ef): a -> Result[IOError, b] \ (ef - FilePath) + IO = x ->
+        try {
+            try {
```
magnus-madsen, line 178: The catch is in the wrong place. Not here.

main/src/library/FilePath.flix
```diff
@@ -155,3 +155,188 @@ pub eff FilePath {
...
+                Result.Ok(f(x))
+            } catch {
+                case ex: IOException =>
+                    Result.Err(IOError.Generic(ex.getMessage()))
+            }
+        } with FilePath {
+            def exists(f, k) = {
+                let file = new File(f);
```
magnus-madsen, line 186: but the catch should be here, and in all the other handlers.

sockmaster27:
@magnus-madsen 
So like this?

main/src/library/FilePath.flix
```diff
@@ -155,3 +155,49 @@ pub eff FilePath {
...
+                let existsRes = try {
+                    let path = Paths.get(f);
+                    let linkOpts: Vector[LinkOption] = Vector.empty();
+                    let exists = Files.exists(path, linkOpts);
+                    Ok(exists)
+                } catch {
+                    case ex: Exception =>
+                        Err(IOError.Generic(ex.getMessage()))
+                };
+                match existsRes {
```
magnus-madsen, line 191: This match can be inlined, right?

main/src/library/FilePath.flix
```diff
@@ -155,3 +155,49 @@ pub eff FilePath {
...
+    ///
+    pub def handle(f: a -> b \ ef): a -> Result[IOError, b] \ (ef - FilePath) + IO = x ->
+        try {
+            Ok(f(x))
+        } with FilePath {
+            def exists(f, k) = {
+                let existsRes = try {
+                    let path = Paths.get(f);
+                    let linkOpts: Vector[LinkOption] = Vector.empty();
+                    let exists = Files.exists(path, linkOpts);
```
magnus-madsen, line 185: Can Flix not resolve  the call to exists without the Vector?

sockmaster27: It depends. I've found that excluding the varargs is flaky in general, partly because of https://github.com/flix/flix/issues/9067, but also because it will sometimes randomly decide that the function override doesn't exist when changing unrelated parts of the program.

So actually, we should probably add one more for the `Paths.get()` call.

main/src/library/FilePath.flix
sockmaster27: How do you mean? Omitting the `existRes` binding?

sockmaster27: Would it be better to use `Result.flatMap`?

magnus-madsen:
Approximately like so:

```
    pub def handle(f: a -> b \ ef): a -> Result[IOError, b] \ (ef - FilePath) + IO = x ->
        try {
            Ok(f(x))
        } with FilePath {
            def exists(f, k) = {
                try {
                    k(Files.exists(Paths.get(f), (...{}: Vector[LinkOption])))
                } catch {
                    case ex: IOException => Err(IOError.Generic(ex.getMessage()))
                }
            }
        }
```

Note also the use of IOException.

sockmaster27:
> Approximately like so:
> 
> ```
>     pub def handle(f: a -> b \ ef): a -> Result[IOError, b] \ (ef - FilePath) + IO = x ->
>         try {
>             Ok(f(x))
>         } with FilePath {
>             def exists(f, k) = {
>                 try {
>                     k(Files.exists(Paths.get(f), (...{}: Vector[LinkOption])))
>                 } catch {
>                     case ex: IOException => Err(IOError.Generic(ex.getMessage()))
>                 }
>             }
>         }
> ```
> 
> Note also the use of IOException.

Wouldn't this catch other unrelated exceptions thrown by the continuation?

sockmaster27:
> Note also the use of IOException.

What about the [`InvalidPathException`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/InvalidPathException.html)?

magnus-madsen:
> > Note also the use of IOException.
> 
> What about the [`InvalidPathException`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/InvalidPathException.html)?

Good catch, add it too.

sockmaster27:
Actually, it looks like this doesn't throw an `IOException` at all 😵‍💫.
However, `Files.exists` throws a [`SecurityException`](https://docs.oracle.com/javase/8/docs/api/java/lang/SecurityException.html).

magnus-madsen:
We ignore security exceptions. The Files class throws IOExs on general.

magnus-madsen:
Anyway, I think you look at each Files.X op, add its exceptions (except for SecurityEx) plus the InvalidPath thing and that is what you should catch.

magnus-madsen:
So with that I think it should be possible to fill out this handler. And then we can iterate.

sockmaster27:
> > Approximately like so:
> > ```
> >     pub def handle(f: a -> b \ ef): a -> Result[IOError, b] \ (ef - FilePath) + IO = x ->
> >         try {
> >             Ok(f(x))
> >         } with FilePath {
> >             def exists(f, k) = {
> >                 try {
> >                     k(Files.exists(Paths.get(f), (...{}: Vector[LinkOption])))
> >                 } catch {
> >                     case ex: IOException => Err(IOError.Generic(ex.getMessage()))
> >                 }
> >             }
> >         }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Note also the use of IOException.
> 
> Wouldn't this catch other unrelated exceptions thrown by the continuation?

What about this?

magnus-madsen:
That's ok for now. We will deal with that later when we add support for java exceptions to the effect system.

sockmaster27:
@magnus-madsen 
This good?

magnus-madsen:
Don't extend IOError. We can do it later, but doing it now causes breakages.
Second, no reason to use varargs where there is no problem. (The problem is only if the varargs array type is an interface).

sockmaster27:
I assume it's okay now?
Then I will start to implement it for the other operations

sockmaster27:
I assume handling `OutOfMemoryException` is also out of scope?

magnus-madsen:
> I assume it's okay now? Then I will start to implement it for the other operations

Please go ahead.

main/src/library/FilePath.flix
```diff
@@ -147,11 +147,311 @@ pub eff FilePath {
...
+
+            def mkDirs(d, k) = {
+                try {
+                    Files.createDirectories(Paths.get(d), (...{}: Vector[FileAttribute]));
+                    k(())
+                } catch {
+                    case ex: InvalidPathException => Err(IOError.Generic(ex.getMessage()))
+                    case ex: IOException => Err(IOError.Generic(ex.getMessage()))
+                    case ex: UnsupportedOperationException => Err(IOError.Generic(ex.getMessage()))
+                    case ex: FileAlreadyExistsException => Err(IOError.Generic(ex.getMessage()))
```
magnus-madsen, line 430: This inherits from IOException, so it will never be reached. It has to be moved upwards. I suggest to always have IOException last.

magnus-madsen:
@sockmaster27 This looks almost ready? If the list() operation is giving trouble, we could defer it.

sockmaster27:
I got list to work by using the old `java.io.File`. I guess we might be able to fix it later, but this should work for now.

magnus-madsen:
> I got list to work by using the old `java.io.File`. I guess we might be able to fix it later, but this should work for now.

That's acceptable. We prefer nio, but its totally OK to use file if necessary. Similarly to using FileReader etc.

main/src/library/FilePath.flix
```diff
@@ -431,15 +432,17 @@ mod FilePath {
...
-//                    k(FromJava.fromJava(Files.list(Paths.get(f))))
-//                } catch {
-//                    case ex: InvalidPathException => Err(IOError.Generic(ex.getMessage()))
-//                    case ex: IOException => Err(IOError.Generic(ex.getMessage()))
-//                    case ex: NotDirectoryException => Err(IOError.Generic(ex.getMessage()))
-//                }
-//            }
+            def list(f, k) = {
+                try {
+                    // TODO: Use java.nio.file.Files.list, but it returns a generic Stream<Path> which Flix's type system currently doesn't support.
```
magnus-madsen, line 437: -> TODO: Use java.nio.file.Files.list when we have support for Java generics.

magnus-madsen (approved):

magnus-madsen:
Looks good: I suggest to next duplicate and split or split and duplicate, if you prefer.

magnus-madsen:
In parallel you can also start looking into the Http effect. The design here is less clear, though.

PR #9071, opened 2024-10-31
feat: highlight provider using lsp visitor

LoZander:
This PR introduces a new highlight provider using the LSP visitor.

The approach includes two searches
- Search 1: find symbol hovered on
- Search 2: find all (other) occurrences of this symbol in the file

The first search uses the `InsideAcceptor` and `StackConsumer`. The second search uses `FileAcceptor` and the new `SymbolOccurrenceConsumer`, which collects all other occurrences of the symbol (does not include the initial occurrence).

Related to #9114


magnus-madsen:
!!!jar

LoZander:
> @LoZander I notice the IDE tests fail. Any idea why?

Hm. Maybe I somehow broke either the LSP spec by illformed JSON responses? But I don't see where the issue would be

LoZander:
On a second thought, it's probably just that there are some cases that the original supports that the new doesn't. So it might not highlight as many kinds of things

LoZander:
Currently looking into it

LoZander:
It turns out it was a silly match case error I made that slipped by (I must've forgotten to test after refactoring)

magnus-madsen:
You seem to assume that you can treat all symbols as the same, but that's probably going too far.

Here is how I imagined it:

I call StackConsumer to find where the cursor is. I pattern match.

If I see `Expr.Var` then I know its a local variable. I can highlight those, so I create a new Visitor where I consume on `Expr.Var` and `Binder`.

If I see `Expr.Tag` then I know its a tag. I can highlight those, so I create a new Visitor where I consume `Expr.Tag`, `Pattern.Tag`, and `Case`.

Does it make sense? You seem to want to force everything into one umbrella which is problematic because (1) You lose some type safety, but (2) more annoyingly not all symbols really work the same way, so you paint yourself into a corner.

LoZander:
> You seem to assume that you can treat all symbols as the same, but that's probably going too far.
> 
> Here is how I imagined it:
> 
> I call StackConsumer to find where the cursor is. I pattern match.
> 
> If I see `Expr.Var` then I know its a local variable. I can highlight those, so I create a new Visitor where I consume on `Expr.Var` and `Binder`.
> 
> If I see `Expr.Tag` then I know its a tag. I can highlight those, so I create a new Visitor where I consume `Expr.Tag`, `Pattern.Tag`, and `Case`.
> 
> Does it make sense? You seem to want to force everything into one umbrella which is problematic because (1) You lose some type safety, but (2) more annoyingly not all symbols really work the same way, so you paint yourself into a corner.

Yeah, that makese sense. After fixing that previous issue, I could also see how for instance way too many things were highlighted as the approach of treating all symbols the same means that symbols that contain the same source text but in different contexts, are also highlighted

magnus-madsen:
Also, if you do it this way, you can refactor one small piece at the time. This will probably help with the IDE tests.

LoZander:
Yeah. Easier to locate problems if nothing else probably

LoZander:
Most of it's there now. What's left (unless I missed something) is
- [ ] Highlight on trait symbols
- [ ] Highlight on Typevars
- [ ] Highlight on general types besides structs and enums
- [ ] Highlight on signatures
- [ ] More manual testing

LoZander:
Ill see if I can get the rest of this done before the next meeting so I can move on to the rename provider thereafter :)

magnus-madsen:
What about the tests?

I think everything on your list should work, except for ModuleSyms. Skip those (we can discuss).

magnus-madsen (commented):
Overall the new code structure looks much better 👍 

You see how everything is structured as one simple pattern match and then a function for each thing?

main/src/ca/uwaterloo/flix/api/lsp/provider/HighlightProvider.scala
```diff
@@ -15,195 +15,251 @@
...
-          case Type.Var(sym, _) => highlightTypeVar(uri, sym)
-          case _ => mkNotFound(uri, pos)
-        }
+    mkHighlights(loc, occurs)
+  }
 
-        case Entity.OpUse(sym, _, _) => highlightOp(uri, sym)
+  private def highlightStructSym(uri: String, sym: Symbol.StructSym, loc: SourceLocation)(implicit root: Root): JObject = {
+    var occurs: List[SourceLocation] = Nil
+    def add(x: SourceLocation): Unit = {
```
magnus-madsen, line 185: I don't think these small add functions -- excuse the pun -- add anything

LoZander: Yeah, no. They don't. Might as well just inline them in "check" which should also be given a better name tbh

main/src/ca/uwaterloo/flix/api/lsp/provider/HighlightProvider.scala
```diff
@@ -15,195 +15,273 @@
...
 
-  private def highlightEffect(uri: String, sym: Symbol.EffectSym)(implicit index: Index): JObject = {
-    val write = (sym.loc, DocumentHighlightKind.Write)
-    val reads = index.usesOf(sym).toList.map(loc => (loc, DocumentHighlightKind.Read))
-    highlight(uri, write :: reads)
-  }
+  private def highlightVarSym(uri: String, sym: Symbol.VarSym, loc: SourceLocation)(implicit root: Root): JObject = {
+    var occurs: List[SourceLocation] = Nil
+
+    def consider(x: Symbol.VarSym, loc: SourceLocation): Unit = {
```
magnus-madsen, line 267: consider becomes:

maybeAddRead
maybeAddWrite

main/src/ca/uwaterloo/flix/api/lsp/provider/HighlightProvider.scala
```diff
@@ -15,195 +15,273 @@
...
-    highlight(uri, write :: reads)
+    mkHighlights(loc, occurs)
   }
 
-  private def highlightEffect(uri: String, sym: Symbol.EffectSym)(implicit index: Index): JObject = {
-    val write = (sym.loc, DocumentHighlightKind.Write)
-    val reads = index.usesOf(sym).toList.map(loc => (loc, DocumentHighlightKind.Read))
-    highlight(uri, write :: reads)
-  }
+  private def highlightVarSym(uri: String, sym: Symbol.VarSym, loc: SourceLocation)(implicit root: Root): JObject = {
```
magnus-madsen, line 264: loc goes away as an argument because we recover all information we want.

main/src/ca/uwaterloo/flix/api/lsp/provider/HighlightProvider.scala
```diff
@@ -15,195 +15,273 @@
...
+    // Traits
+    case TypedAst.Trait(_, _, _, sym, _, _, _, _, _, _) => highlightTraitSym(uri, sym, sym.loc)
+    case SymUse.TraitSymUse(sym, loc) => highlightTraitSym(uri, sym, loc)
+    case TraitConstraint.Head(sym, loc) => highlightTraitSym(uri, sym, loc)
+    case Ast.Derivation(sym, loc) => highlightTraitSym(uri, sym, loc)
+    // Signatures
+    case TypedAst.Sig(sym, _, _, _) => highlightSigSym(uri, sym, sym.loc)
+    case SymUse.SigSymUse(sym, loc) => highlightSigSym(uri, sym, loc)
+    // Symbols in types
+    case tpe: Type => getTypeSymOccur(tpe) match {
```
magnus-madsen, line 78: Extend visitor with:

visitTypeCst
visitTypeVar
visitTypeApply
... type alias and assoc type ...

LoZander: Yes. This has been done in #9115

main/src/ca/uwaterloo/flix/api/lsp/provider/HighlightProvider.scala
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Magnus Madsen
+ * Copyright 2024 Magnus Madsen, Alexander Dybdahl Troelsen
```
magnus-madsen, line 2: I think putting on two lines would be more readable and also preserve the dates.

LoZander: Oh, right. Yeah, I'll do that. Like A or B?
A:
```
Copyright 2020 Magnus Madsen
Copyright 2024 Alexander Dybdahl Troelsen?
```

B:
```
Copyright 2020 Magnus Madsen
          2024 Alexander Dybdahl Troelsen
```

magnus-madsen: I think B is the easiest

magnus-madsen:
What's the status on this? I can review, but maybe the tests should be fixed first?

LoZander:
> What's the status on this? I can review, but maybe the tests should be fixed first?

I'll look into the tests and why they're failing today, and then hopefully I can set it as ready for review by tonight :)

magnus-madsen:
> > What's the status on this? I can review, but maybe the tests should be fixed first?
> 
> I'll look into the tests and why they're failing today, and then hopefully I can set it as ready for review by tonight :)

Sounds good. If the tests need to be updated then please create a PR targetting the VSCode repo.

LoZander:
I'm seeing some very strange behaviour. I know that vscode seems to have some kind of "fallback" highlighter that is invoked if the highlight provider doesn't return anything. This one simply highlights text that is equal. It seems that the testing environment is not actually invoking the highlightprovider but this fallback instead.

For instance, in the test "Should find highlights of Equatable.equals signature-use" the set of highlights returned includes the instance impl on line 16, which it should not. But testing the new highlight provider manually, it doesn't highlight this line. I suspect it for some reason uses the default fallback highlighting. So for some reason, the testing environment fails to capture the actual behaviour. I don't know why. If I go back to the master branch, the tests don't fail. For some reason, my changes mean that the highlighter works differently for manual tests as opposed to the vscode tests. Maybe I broke the JSON objects in such a way that it's fragile now or something?

magnus-madsen:
You can always print to see if your code is being run and returns the right result? Maybe you are code is incorrect (offers no suggestions) and then the fallback kick in? I don't know.

LoZander:
I'll try that

LoZander:
Oh, I found the specific error. So, the highlighter is being invoked, BUT the test catches an edge case where the cursor is on the first parentheses __next__ to the symbol in question. Thus, in the eyes of the new provider, you're not __on__ the symbol, so the highlight provider returns nothing. However, the vscode callback is a little more "forgiving" so is actually invoked.

The failure happens in `highlight/Equatable.flix` on line 30
```flix
Equatable.equals(first, second)
                ^
```

The problem is that the highlight provider only considers the cursor on `Equatable.equals` when it's on the actual symbol, not when the cursor is on the apply of the signature in general.

I'm unsure if this is a problem with the test or my provider, I suppose that's up to interpretation. But the fact that the fallback is invoked, indicated that maybe my provider should be accept being on apply as being on the symbol (as long as you're not on an arg ofc.)

magnus-madsen:
I would argue the paren is not part of the symbol. So it sounds like the test should be changed?

magnus-madsen:
> It depends on whether the highlight is a highlight of the symbol or the AST the apply it's assocaited with. I could see an argument for either. It's the same question in the enum case situation

Think critically. What would you do here:

```
foo 


/* blabla */ 


      (123)
```

LoZander:
> > It depends on whether the highlight is a highlight of the symbol or the AST the apply it's assocaited with. I could see an argument for either. It's the same question in the enum case situation
> 
> Think critically. What would you do here:
> 
> ```
> foo 
> 
> 
> /* blabla */ 
> 
> 
>       (123)
> ```

Yeah, I see. It's also strange if an application would be treated differently. Really application is just an operation involving the symbol. When having the cursor on `x`, it'd be weird if  `+` in  `x + 2` was included in the highlight.

LoZander:
Wish we could turn off the fallback highlighter though. I could see it being invoked being possibly confusing

LoZander:
> In general, if you were to establish a principle, would you say that highlights should be narrow or wide?

what do you mean by narrow and wide?

magnus-madsen:
> > In general, if you were to establish a principle, would you say that highlights should be narrow or wide?
> 
> what do you mean by narrow and wide?

How much it highlights. Is it like a precision strike or an area bombardment? When you click on something, what do your eyes want to see?

LoZander:
> > > In general, if you were to establish a principle, would you say that highlights should be narrow or wide?
> > 
> > 
> > what do you mean by narrow and wide?
> 
> How much it highlights. Is it like a precision strike or an area bombardment? When you click on something, what do your eyes want to see?

Well, we're highlight symbols, and those are usually fairly clear-cut as to where they start and end. For something like a hover, there might be cases where a result could be "too precise" in the sense that you might've meant to hover on a more general expression (like hovering on the `+` in `2 + 3` might be meant as a hover on the entire expression). But with highlights, a symbol is a symbol. I would never expect it to include the surrounding expressions.

LoZander:
So at least as a general principle, I'd say they should be narrow

LoZander:
... I completely forgot about records. So I'm doing them now. Problem is the Visitor doesn't actually visit the labels of record expressions at all. So I'll do a PR to add that to the visitor.

Then there are just three things left:
- [ ] make visitor visit record labels
- [ ] make Record highlight work
- [ ] remove highlight repetition in certain case

LoZander:
So there is a complication. The SourceLocations for records are weird. I modified the printing of locations to include the end line and column, and for
```
    let r1 = {x = 1, y = 2};
```
when inspecting the record, I got
```
RecordExtend(
    x,
    Cst(Int32(1),Int32,file://[censored]/src/testtesttest.flix:14-14:19-21),
    RecordExtend(
        y,
        Cst(Int32(2),Int32,file://[censored]/src/testtesttest.flix:14-14:26-28),
        RecordEmpty({  },file://[censored]/src/testtesttest.flix:14-14:14-28),
        { y = Int32 },
        Pure,
        file://[censored]/src/testtesttest.flix:14-14:22-28
    ),
    { x = Int32, y = Int32 },
    Pure,
    file://[censored]/testtesttest.flix:14-14:15-21
)
```

It does not make sense that the outer-mst RecordExtend doesn't actually cover the sourcelocation for everything inside. In fact, the outer Record extend only covers the `{x = 1` part. Since the visitor only visits the inner RecordExtend if it visits the outer, this means that the `the `y = 2}` part is never visited, even if the cursor is on it.

LoZander:
Also, I think the visit to labels is actually unnecessary. My bad. I don't know if I can justify it, since a label is worth very little in and of itself. Like we also don't visit symbols. I thought I needed it, but I don't. I might do a PR to remove it

LoZander:
So the record higlighting is done. It doesn't quite work, but that should be fixed when the SourceLocation issues of records is fixed. As for this PR, I think it's done, except that #9153 needs to go through and then a slight modification is needed to adjust for that.

The remaining test failures are due to the tests expecting the wrong behaviour.

LoZander:
I'll just mark this ready. If you check this before #9153 goes through, I'll just do a very short follow up PR. If that one goes through first, then I'll just do an additional commit to account for the removal of label visits :)

magnus-madsen:
> The remaining test failures are due to the tests expecting the wrong behaviour, except for the record label test, which fails presumably fails due to aforemetioned issues.

Do you have an updated PR for the tests?

> It does not make sense that the outer-mst RecordExtend doesn't actually cover the sourcelocation for everything [...]

Will you submit a PR to fix this?

magnus-madsen:
!!!jar

main/src/ca/uwaterloo/flix/api/lsp/provider/HighlightProvider.scala
```diff
@@ -15,195 +16,369 @@
...
 
-        case Entity.CaseUse(sym, _, _) => highlightCase(uri, sym)
+    builder.build
+  }
+
+  /**
+    * A builder for creating an LSP highlight response containing a [[DocumentHighlight]] for each recorded
+    * occurrence of a given token [[tok]].
+    *
+    * An occurrence of a token of type [[T]] is "considered" by invoking [[considerRead]] or [[considerWrite]] for resp.
```
magnus-madsen, line 177: The word token has a very specific meaning in Flix. What is this token?

main/src/ca/uwaterloo/flix/api/lsp/provider/HighlightProvider.scala
```diff
@@ -15,195 +16,369 @@
...
+    * "write" and "read" occurrences. By a "write" occurrence, we mean an occurrence where the token is being
```
magnus-madsen, line 178: you flipped write and read.

main/src/ca/uwaterloo/flix/api/lsp/provider/HighlightProvider.scala
```diff
@@ -15,195 +16,369 @@
...
+    * defined or otherwise bound to something. By "read" we mean an occurrence where the token is merely read.
+    * When we say "consider", we mean first checking if the occurrence is an occurrence of [[tok]] specifically.
+    * If so, it's added to our list of either "read" or "write" occurrences, depending on the type.
+    *
+    * When we're done considering tokens, we can construct the LSP response by calling [[build]]
+    *
+    * @param tok  the token we're finding occurrences of.
+    * @tparam T   the type of token that [[tok]] is.
+    */
+  private case class HighlightBuilder[T](tok: T) {
```
magnus-madsen, line 188: Move this to the very end of the file.

magnus-madsen: I dont think this should be a case class since its more of an object with state.

main/src/ca/uwaterloo/flix/api/lsp/provider/HighlightProvider.scala
```diff
@@ -15,195 +16,369 @@
...
+    private var writes: List[SourceLocation] = Nil
```
magnus-madsen, line 189: I think it is more natural to have reads before writes. So reorder everything below in this class according to that. Thanks!

magnus-madsen:
A few comments from running it:

- It seems fast enough. Agree?
- I notice that clicking "Plus" in a pattern "AExp.Plus" will highlight the whole thing.
- I notice that clicking at the end of a variable does not work. For example, try clicking here "foo|". This seems to trigger the default highlighter. What are your thoughts on this?
- Highlighting effect operations does not seem to work.

main/src/ca/uwaterloo/flix/api/lsp/provider/HighlightProvider.scala
LoZander: It used to be Symbol, when we were just finding occurrences of symbols. But now were findind occurrences of labels as well. So it's more like an sequence of letters that mark or name some unit that might occur repeatedly. Token was the best name I could come up with, but I'm open to alternatives. Maybe identifier or something?

main/src/ca/uwaterloo/flix/api/lsp/provider/HighlightProvider.scala
LoZander: That I did. Good catch. I'll fix that :)

main/src/ca/uwaterloo/flix/api/lsp/provider/HighlightProvider.scala
```diff
@@ -15,195 +16,369 @@
...
+      // Variables
+      case Binder(sym, _) => highlightVarSym(sym)
+      case TypedAst.Expr.Var(varSym, _, _) => highlightVarSym(varSym)
+
+      case _ => mkNotFound(uri, pos)
+    }
+  }
 
-        case Entity.Def(defn) => highlightDef(uri, defn.sym)
+  private def highlightLabel(label: Name.Label)(implicit root: Root, acceptor: Visitor.Acceptor): JObject = {
```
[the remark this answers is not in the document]
LoZander: yep, that makes sense

main/src/ca/uwaterloo/flix/api/lsp/provider/HighlightProvider.scala
```diff
@@ -1,5 +1,6 @@
 /*
- * Copyright 2020 Magnus Madsen
+ * Copyright 2020 Magnus Madsen,
+ *           2024 Alexander Dybdahl Troelsen
```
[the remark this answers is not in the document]
LoZander: But you wrote that B (copyright only being written o nthe first line) was better? I mean, I don't have a strong preference. If you think it's better to right Copyright on both lines, we can do that

main/src/ca/uwaterloo/flix/api/lsp/provider/HighlightProvider.scala
```diff
@@ -15,195 +16,369 @@
...
+    *             for each occurrence of the symbol under the cursor.
+    */
+  def processHighlight(uri: String, pos: Position)(implicit root: Root): JObject = {
+    val stackConsumer = StackConsumer()
+    Visitor.visitRoot(root, stackConsumer, Visitor.InsideAcceptor(uri, pos))
+
+
+    stackConsumer.getStack.headOption match {
       case None => mkNotFound(uri, pos)
+      case Some(x) => println(x); highlightAny(x, uri, pos)
```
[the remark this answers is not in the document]
LoZander: just temporary for testing purposes. I just forgot to remove it ^^'

LoZander:
> A few comments from running it:
> 
> * It seems fast enough. Agree?
> * I notice that clicking "Plus" in a pattern "AExp.Plus" will highlight the whole thing.
> * I notice that clicking at the end of a variable does not work. For example, try clicking here "foo|". This seems to trigger the default highlighter. What are your thoughts on this?
> * Highlighting effect operations does not seem to work.

1)  I agree. It has the same delay that anything else has, which I think we agree was probably an intended delay in the end of vscode. Eiter way, it's def fast enough.

2) strange, I'll look into that.

3) Yes, so this is something I've noticed but due to the kind of cursor I use (Vim style) I thought it was the default highlighter that was weird. For me the "foo|" situation would look more like "foo_", which makes it seem like highlighting shouldn't occur. But I agree, for "foo|", "foo" should be highlighted. Most people use the thing cursor afterall, and that agrees with the default highlighter as well (meaning we won't get triggers of it as much).

Currently, the issue actually lies with how `InsideAcceptor` interprets the position of the cursor. Currently it interprets it as the part that the fat cursor highlights, but for the thing cursor, that means only what's directly to the right is considered "at the cursor" and not what's directly on it's left. The fat cursor is clearly the outlier here, so the easiest solution is just to make `InsideAcceptor` interpret the position as a range from the position directly to the left of | and through the one on the right of it.

LoZander:
> > The remaining test failures are due to the tests expecting the wrong behaviour, except for the record label test, which fails presumably fails due to aforemetioned issues.
> 
> Do you have an updated PR for the tests?
> 
> > It does not make sense that the outer-mst RecordExtend doesn't actually cover the sourcelocation for everything [...]
> 
> Will you submit a PR to fix this?

I can def make a PR for the tests. I can look into the record sourcelocation issue as well

LoZander:
So, there are a couple of issues left. One big one, that will affect not just this but many other providers I'm sure, is the issue of the thin cursor position. The issue is that a thin cursor may be interpreted as being at the position before or after it. That then raises an issue: if there are two different AST nodes, both of which border the thin cursor one at each side, then which one is the cursor at? Most expressions and such have a space between, so here the issue cannot occur. But will this always be the case? What about a call? where the cursor is between the final letter of the function name and the opening parenthesis? There might be a correct answer to that one, but it my point is the issue seems like a more general issue. What about `2|+2`? Are we at `2` or `+`? and so on

LoZander:
> So, there are a couple of issues left. One big one, that will affect not just this but many other providers I'm sure, is the issue of the thin cursor position. The issue is that a thin cursor may be interpreted as being at the position before or after it. That then raises an issue: if there are two different AST nodes, both of which border the thin cursor one at each side, then which one is the cursor at? Most expressions and such have a space between, so here the issue cannot occur. But will this always be the case? What about a call? where the cursor is between the final letter of the function name and the opening parenthesis? There might be a correct answer to that one, but it my point is the issue seems like a more general issue. What about `2|+2`? Are we at `2` or `+`? and so on

for now, I could just assert some rule like "if there is something immediately to the left then we're in that, if not then we check immediately right". That would always fevour the left, which I suspect will at least often give expected behaviour. That might be a sufficient way to handle it until we find a better solution

LoZander:
A similar but somewhat better solution would be a new variant on the inside acceptor that takes two positions (left and right) and accepts on _either_. Then we're branching a bit, but at least we aren't doing a complete additional search

LoZander:
but actually, not sure that would suffice. We'd need some way of finding the most precise thing under _either_ location. What I described above would still provide a "last visited" node and not give any indication whether it's the one under left or right. Hm. Some combination of a more sophisticated acceptor _and_ consumer might work

LoZander:
All in all, I guess three searches isn't _that_ bad, when two of them are bounded by the height of the AST. Could be worse

magnus-madsen:
These are some good reflections. But what do you think is the correct solution?

Also, why not modify the acceptor?

Also, why not consider the difference between an atomic expression like Expr.Var and a compound one like Binary Plus?

LoZander:
> These are some good reflections. But what do you think is the correct solution?
> 
> Also, why not modify the acceptor?
> 
> Also, why not consider the difference between an atomic expression like Expr.Var and a compound one like Binary Plus?

I'll spend this morning thinking about this. Because I feel there has to be a better solution than what I did. Probably something with a new acceptor

main/src/ca/uwaterloo/flix/api/lsp/provider/HighlightProvider.scala
```diff
@@ -15,201 +16,412 @@
...
+    * @param pos  the [[Position]] of the cursor.
+    * @param root the [[Root]] AST node of the Flix project.
+    * @return     A [[JObject]] representing an LSP highlight response. On success, contains [[DocumentHighlight]]
+    *             for each occurrence of the symbol under the cursor.
+    */
+  def processHighlight(uri: String, pos: Position)(implicit root: Root): JObject = {
 
-      case Some(entity) => entity match {
-        case Entity.Case(caze) => highlightCase(uri, caze.sym)
+    // TODO both left and right searches are only necessary in some cases. Test if optimising to only do both when necessary actually improves performance in a measurable way
```
magnus-madsen, line 55: Remove TODO. We track TODOs in tickets. Otherwise the code is just what it is.

magnus-madsen:
!!!jar

main/src/ca/uwaterloo/flix/api/lsp/provider/HighlightProvider.scala
```diff
@@ -15,201 +16,412 @@
...
+      case (Some(x), _) => highlightAny(x, uri)
+      case (None, None) => None
+    }
 
-        case Entity.Def(defn) => highlightDef(uri, defn.sym)
+    highlights.getOrElse(mkNotFound(uri, pos))
+  }
 
-        case Entity.Sig(sig0) => highlightSig(uri, sig0.sym)
+  private def searchLeftOfCursor(uri: String, pos: Position)(implicit root: Root): Option[AnyRef] = pos match {
```
magnus-madsen, line 68: Add doc comments and example

main/src/ca/uwaterloo/flix/api/lsp/provider/HighlightProvider.scala
```diff
@@ -15,201 +16,412 @@
...
+      case Position(line, character) if character >= 2 =>
+        val leftOfCursor = Position(line, character - 1)
+        search(uri, leftOfCursor)
+      case _ => None
+    }
 
-        case Entity.Enum(enum0) => highlightEnum(uri, enum0.sym)
+  private def searchRightOfCursor(uri: String, pos: Position)(implicit root: Root): Option[AnyRef] = {
```
magnus-madsen, line 75: Add doc comments and example

main/src/ca/uwaterloo/flix/api/lsp/provider/HighlightProvider.scala
```diff
@@ -15,201 +16,412 @@
...
+    search(uri, pos)
+  }
 
-        case Entity.Struct(struct) => highlightStruct(uri, struct.sym)
+  private def search(uri: String, pos: Position)(implicit root: Root): Option[AnyRef] = {
```
magnus-madsen, line 79: Add doc comments

magnus-madsen (commented):
I added some minor comments.

magnus-madsen:
Once these are fixed, this is ready for merge right?

magnus-madsen:
The search left does not really seem to work? Did you try it on an example?

magnus-madsen:
For example:

```
/// An algebraic data type for shapes.
enum Shape {
    case Circle(Int32),          // circle radius
    case Square(Int32),          // side length
    case Rectangle(Int32, Int32) // height and width
}

/// Computes the area of the given shape using
/// pattern matching and basic arithmetic.
def area(s: Shape): Int32 = match s {
    case Shape.Circle(r)       => 3 * (r * r)
    case Shape.Square(w)       => w * w
    case Shape.Rectangle(h, w) => h * w
}
```

LoZander:
> The search left does not really seem to work? Did you try it on an example?

Yes, I did but I was not thorough enough. I see the issue. A very simple and dumb mistake. I accidentally got it backwards so it prioritises searching left before right.

LoZander:
Beyond that, there are two other issues. On is that I forgot to filter out highlights on enums when you're in the enum but not in it's symbol. So that gives some strange behaviour when on white space in enum.

There's also an issue with Tag expressions. I suspect some source location stuff again, but I'll look into it.

I swear, some day the highlighter will actually be done ^^'

LoZander:
Update on the Tag expression stuff. Yes, it's a source location issue again. At least hopefully these issues will all or mostly be solved by the time this highlight provider is done. The issue here is that the inner expression in a Tag expression has a source location that starts where the Tag expression starts and so it overlaps the CaseSymUse completely. Since the visitor visits the subexpression last, it means the CaseSymUse is never visited. I'll do a new PR to fix this

magnus-madsen:
Do you think it would make sense to merge now and then iterate?

LoZander:
> Do you think it would make sense to merge now and then iterate?

I think that would make sense. As far as I can see, there are only two remaining issues 

The first is that records stil aren't perfect. recordRestrict and recordTypes still don't work.

The second is this strange issue I stumble upon. For
```flix
def main(): Unit \ IO = 
    try {
        println(divide(3, 2));
        println(divide(3, 0))
    } with DivByZero {
        def divByZero(_resume) = println("Oops: Division by Zero!")
    }
```
having the cursor on IO also highlights the entire function body. Not sure why. But it seems to be specific to the try expression. If I change the body to for instance `println("")`, the issue doesn't occur.

But otherwise, I think it works. These two issues could easily just be fixed in a follow up PR :)

LoZander:
also, the record issue accounts for 1 of the 4 remaining vscode test failures. The other three are due to problems in the tests, problems that vscode-flix#457 should fix

magnus-madsen (approved):

magnus-madsen:
I am merging this so we can iterate and start on the next provider.

LoZander:
the compiler did crash at one point. But I could not recreate the issue, so I don't know if it had anything to do with the provider. Hard to tell. Might have just been a fluke

PR #9073, opened 2024-10-31
feat: struct get for non-mut fields is pure

JonathanStarup:


magnus-madsen:
What about some tests?

magnus-madsen:
@mlutze Do you agree we should have this feature?

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -605,9 +605,10 @@ object ConstraintGen {
         val (instantiatedFieldTpes, structTpe, regionVar) = instantiateStruct(field.sym.structSym, root.structs)
         val (tpe, eff) = visitExp(exp)
         c.expectType(structTpe, tpe, exp.loc)
-        val fieldTpe = instantiatedFieldTpes(field.sym)
+        val (mutable, fieldTpe) = instantiatedFieldTpes(field.sym)
         c.unifyType(fieldTpe, tvar, loc)
-        c.unifyType(Type.mkUnion(eff, regionVar, loc), evar, exp.loc)
+        if (mutable) c.unifyType(Type.mkUnion(eff, regionVar, loc), evar, exp.loc)
```
magnus-madsen, line 610: I think a comment would be appropriate.

mlutze:
> @mlutze Do you agree we should have this feature?

I think our effect system should be as honest as possible

JonathanStarup:
I've tried to make it more clear

I also think of this just as a more precise type rather than a feature

magnus-madsen:
> > @mlutze Do you agree we should have this feature?
> 
> I think our effect system should be as honest as possible

Is  that a yes? Doesnt it depend on what a heap effect is?

mlutze:
> > > @mlutze Do you agree we should have this feature?
> > 
> > 
> > I think our effect system should be as honest as possible
> 
> Is that a yes? Doesnt it depend on what a heap effect is?

It is a yes.

JonathanStarup:
> > > @mlutze Do you agree we should have this feature?
> > 
> > 
> > I think our effect system should be as honest as possible
> 
> Is that a yes? Doesnt it depend on what a heap effect is?

It is an immutable value so It can definitely be pure. The only reason it wouldn't be is if we deallocate the object on scope exit, but I deem that impossible because of algebraic effects

magnus-madsen:
If you add a few tests I would be happy to merge this.

main/test/flix/Test.Exp.Struct.Get.flix
```diff
@@ -78,12 +77,21 @@ mod Get {
...
+        };
+        f("Hello") == "Hello Lucky"
+    }
 }
 
 struct BinaryTree[t, r] {
-    left: Option[BinaryTree[t, r]],
-    right: Option[BinaryTree[t, r]],
-    value: t
+    mut left: Option[BinaryTree[t, r]],
```
JonathanStarup, line 92: I've added `mut` to the structs since its weird only to have immutable fields

magnus-madsen (approved):

magnus-madsen:
@JonathanStarup Tests failures

PR #9081, opened 2024-11-01
refactor: remove `Validation` from `pick` in `Weeder2`

jaschdoc:
Supersedes https://github.com/flix/flix/pull/8824

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
@@ -90,26 +90,32 @@ object Weeder2 {
   private def visitUse(tree: Tree)(implicit sctx: SharedContext): Validation[List[UseOrImport], CompilationMessage] = {
     expect(tree, TreeKind.UsesOrImports.Use)
     val maybeUseMany = tryPick(TreeKind.UsesOrImports.UseMany, tree)
-    mapN(pickQName(tree)) { qname =>
-      val isTopLevelName = qname.namespace.idents.isEmpty
-      val isNotImportedByUse = maybeUseMany.isEmpty
-      val isUnqualifiedUse = isTopLevelName && isNotImportedByUse
-      if (isUnqualifiedUse) {
-        val error = UnqualifiedUse(qname.loc)
-        sctx.errors.add(error)
-        List.empty
-      } else {
-        val nname = Name.NName(qname.namespace.idents :+ qname.ident, qname.loc)
-        val optUses = maybeUseMany.map(visitUseMany(_, nname))
-        optUses match {
-          // case: use many, if uses is empty the parser will have reported an error
-          case Some(uses) => uses
+    val optQname = pickQName(tree)
+    val result = optQname match {
+      case Some(qname) =>
+        val isTopLevelName = qname.namespace.idents.isEmpty
+        val isNotImportedByUse = maybeUseMany.isEmpty
+        val isUnqualifiedUse = isTopLevelName && isNotImportedByUse
+        if (isUnqualifiedUse) {
+          val error = UnqualifiedUse(qname.loc)
+          sctx.errors.add(error)
+          List.empty
+        } else {
+          val nname = Name.NName(qname.namespace.idents :+ qname.ident, qname.loc)
+          val optUses = maybeUseMany.map(visitUseMany(_, nname))
+          optUses match {
+            // case: use many, if uses is empty the parser will have reported an error
+            case Some(uses) => uses
 
-          // case: use one, use the qname
-          case None => List(UseOrImport.Use(qname, qname.ident, qname.loc))
+            // case: use one, use the qname
+            case None => List(UseOrImport.Use(qname, qname.ident, qname.loc))
+          }
         }
-      }
+
+      // Parser has already reported an error
+      case None => List.empty
     }
+    Validation.success(result)
   }
 
   private def visitUseMany(tree: Tree, namespace: Name.NName)(implicit sctx: SharedContext): List[UseOrImport] = {
@@ -222,12 +228,19 @@ object Weeder2 {
 
     private def visitModuleDecl(tree: Tree)(implicit sctx: SharedContext, flix: Flix): Validation[Declaration.Namespace, CompilationMessage] = {
       expect(tree, TreeKind.Decl.Module)
+      val qname = pickQName(tree) match {
+        case Some(name) => name
+        case None =>
+          // Parser has reported an error so return some possibly invalid name
+          val name = text(tree)
+          Name.mkQName(name.mkString)
+      }
+
       mapN(
-        pickQName(tree),
         pickAllUsesAndImports(tree),
         pickAllDeclarations(tree)
       ) {
-        (qname, usesAndImports, declarations) =>
+        case (usesAndImports, declarations) =>
           val base = Declaration.Namespace(qname.ident, usesAndImports, declarations, tree.loc)
           qname.namespace.idents.foldRight(base: Declaration.Namespace) {
             case (ident, acc) => Declaration.Namespace(ident, Nil, List(acc), tree.loc)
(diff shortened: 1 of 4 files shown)


jaschdoc:
@magnus-madsen I think this may be a better solution for the hard failure in the `pick` function. I think it is "just" a matter of a) removing calls to `expect` since an arbitrary `tree` may now be an error node and then b) turn ICEs into actual errors.

One thing we could consider changing is allowing a tree to have 0 children (the code assumes that a node always has children right now). Then we do not carry the entire tree around. The nice thing about the current solution is that any other function can in principle have access to the original tree and continue from that (but right now it does not do that)

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1180,35 +1216,34 @@ object Weeder2 {
...
+                case "+" => expr
+                case op => Expr.Apply(Expr.Ambiguous(Name.mkQName(op, tree.loc), opTree.loc), List(expr), tree.loc)
               }
-            case Some(_) => throw InternalCompilerException(s"Expected unary operator but found tree", tree.loc)
-            case None => throw InternalCompilerException(s"Parser produced tree of kind 'Op' without child", tree.loc)
+            })
           }
-        }
-      )
+        // TODO Make these into errors?
```
jaschdoc, line 1243: undecided todo

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -843,6 +863,7 @@ object Weeder2 {
         case None =>
           // Fall back on Expr.Error. Parser has reported an error here.
           val err = UnexpectedToken(expected = NamedTokenSet.Expression, actual = None, SyntacticContext.Expr.OtherExpr, loc = tree.loc)
+          // TODO: add to sctx?
```
jaschdoc, line 866: undecided todo

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -3123,32 +3192,36 @@ object Weeder2 {
...
+    if (idents.nonEmpty) {
+      val first = idents.head
+      val ident = idents.last
+      val nnameIdents = idents.dropRight(1)
+      val loc = SourceLocation(isReal = true, first.loc.sp1, ident.loc.sp2)
+      val nname = Name.NName(nnameIdents, loc)
+      Some(Name.QName(nname, ident, loc))
+    } else {
+      // Parser has already reported an error so we just return None
+      // TODO: NeedAtLeastOne? Error?
```
jaschdoc, line 3218: undecided todo

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -3123,32 +3192,36 @@ object Weeder2 {
...
       val idents = pickQNameIdents(tree)
       mapN(idents) {
         idents => Name.JavaName(idents, tree.loc)
       }
     }
   }
 
-  private def pickQName(tree: Tree)(implicit sctx: SharedContext): Validation[Name.QName, CompilationMessage] = {
-    mapN(pick(TreeKind.QName, tree))(visitQName)
+  private def pickQName(tree: Tree)(implicit sctx: SharedContext): Option[Name.QName] = {
```
jaschdoc, line 3203: this is a key change since it allows `pickQName` to "fail" after `pick` has failed. However, changing this function requires changing all call sites. This is the reason why the PR is so big. I have tried to the best of my ability to carefully consider each call site

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -3123,32 +3192,36 @@ object Weeder2 {
...
+    } else {
+      // Parser has already reported an error so we just return None
+      // TODO: NeedAtLeastOne? Error?
+      None
+    }
   }
 
   private def pickNameIdent(tree: Tree)(implicit sctx: SharedContext): Validation[Name.Ident, CompilationMessage] = {
-    mapN(pick(TreeKind.Ident, tree))(tokenToIdent)
+    Validation.success(tokenToIdent(pick(TreeKind.Ident, tree)))
```
jaschdoc, line 3224: We can now start to remove more Validation in subsequent PRs. Of course, we still have 1 `HardFailure` left.

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -3178,22 +3251,32 @@ object Weeder2 {
...
+          Name.Ident(name, tree.loc)
+
+        case Some(t: Tree) if t.kind == TreeKind.CommentList =>
+          // We hit a misplaced comment.
+          val name = text(tree).mkString("")
+          val error = MisplacedComments(SyntacticContext.Unknown, t.loc)
+          sctx.errors.add(error)
+          Name.Ident(name, tree.loc)
+
+        // TODO: Replace exception below with error?
```
jaschdoc, line 3277: undecided todo

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -3178,22 +3251,32 @@ object Weeder2 {
...
-        Name.Ident(token.text, SourceLocation(isReal = true, sp1, sp2))
-      // If child is an ErrorTree, that means the parse already reported and error.
-      // We can avoid double reporting by returning a success here.
-      // Doing it this way is most resilient, but phases down the line might have trouble with this sort of thing.
-      case Some(t: Tree) if t.kind.isInstanceOf[TreeKind.ErrorTree] =>
-        val name = text(tree).mkString("")
-        Name.Ident(name, tree.loc)
-      case Some(t: Tree) if t.kind == TreeKind.CommentList =>
-        // We hit a misplaced comment.
+    tree.kind match {
```
jaschdoc, line 3254: This is also an important change. Before this, it just checked directly the children. However, now it checks if the tree is an error node itself and handles it same way it would handle a child error node.

jaschdoc: This means the same disclaimer applies:
```
        // If child is an ErrorTree, that means the parse already reported and error.
        // We can avoid double reporting by returning a success here.
        // Doing it this way is most resilient, but phases down the line might have trouble with this sort of thing.
```

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -3323,6 +3410,24 @@ object Weeder2 {
     List.from(pairs.flatten)
   }
 
+  private def sequenceOpt[A](xs: List[Option[A]]): Option[List[A]] = {
```
jaschdoc, line 3413: todo: add doc

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -3271,13 +3354,17 @@ object Weeder2 {
...
     tryPick(kind, tree) match {
-      case Some(t) => Validation.success(t)
+      case Some(t) => t
       case None =>
-        val error = NeedAtleastOne(NamedTokenSet.FromTreeKinds(Set(kind)), synctx, loc = tree.loc)
-        Validation.HardFailure(Chain(error))
+        val error = NeedAtLeastOne(NamedTokenSet.FromTreeKinds(Set(kind)), synctx, loc = tree.loc)
+        sctx.errors.add(error)
+        val treeKind = TreeKind.ErrorTree(error)
+        Tree(treeKind, Array(tree), tree.loc.asSynthetic)
```
jaschdoc, line 3367: This is the primary change and the reason for all the other changes. The idea here is just to return an error node that wraps the original tree.

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -222,12 +228,19 @@ object Weeder2 {
 
     private def visitModuleDecl(tree: Tree)(implicit sctx: SharedContext, flix: Flix): Validation[Declaration.Namespace, CompilationMessage] = {
       expect(tree, TreeKind.Decl.Module)
+      val qname = pickQName(tree) match {
+        case Some(name) => name
+        case None =>
+          // Parser has reported an error so return some possibly invalid name
+          val name = text(tree)
```
jaschdoc, line 235: everywhere we use `text` for error handling we could consider using `flix` or `Symbol` to generate a unique name

jaschdoc:
Still a lot of comments, but I highlighted the important changes. So if this seems like right thing, we can tidy it up a bit and merge it.

magnus-madsen:
We agreed that a simple `ICE` should be sufficient + fix a few things here and there.

jaschdoc:
Will open a new PR instead

PR #9086, opened 2024-11-02
refactor: rename `FileRead` to `FsRead` and `FileWrite` to `FsWrite`

magnus-madsen:


magnus-madsen:
The idea is to free these names for the algebraic effect.

mlutze (approved):

PR #9097, opened 2024-11-03
feat: introduce `FilePathWithError`

### LINKED ISSUE #9034: Duplicate `FilePath` into `FilePathWithError`
The idea is to expose the same interface, except the return types should now all be `Result`s. 

For example, instead of:

```
    ///
    /// Returns `true` if the given file `f` exists.
    ///
    pub def exists(f: String): Bool

    ///
    /// Returns `true` is the given file `f` is a directory.
    ///
    pub def isDirectory(f: String): Bool
```

We would have

```
    ///
    /// Returns `true` if the given file `f` exists.
    ///
    pub def exists(f: String): Result[IOError, Bool]

    ///
    /// Returns `true` is the given file `f` is a directory.
    ///
    pub def isDirectory(f: String): Result[IOError, Bool]
```

#9032 should be completed first.

sockmaster27:
Fixes #9034

### DIFF
--- a/main/src/library/FilePathWithError.flix
+++ b/main/src/library/FilePathWithError.flix
@@ -1,0 +1,481 @@
+/*
+ *  Copyright 2024 Holger Dal Mogensen
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+///
+/// An effect used to interact with the file system.
+///
+pub eff FilePathWithError {
+
+    ///
+    /// Returns `true` if the given file `f` exists.
+    ///
+    pub def exists(f: String): Result[IOError, Bool]
+
+    ///
+    /// Returns `true` is the given file `f` is a directory.
+    ///
+    pub def isDirectory(f: String): Result[IOError, Bool]
+
+    ///
+    /// Returns `true` if the given file `f` is a regular file.
+    ///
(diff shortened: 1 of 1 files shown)


main/src/library/FilePathWithError.flix
```diff
@@ -0,0 +1,481 @@
...
+                    case ex: IOException => Err(IOError.Generic(ex.getMessage()))
+                };
+                k(res)
+            }
+
+            def write(data, f, k) = {
+                let res = try {
+                    let bytes = data#str.getBytes(StandardCharsets.UTF_8);
+                    Files.write(Paths.get(f), bytes, (...{}: Vector[OpenOption]));
+                    k(())
```
magnus-madsen, line 324: Why call k here? (I am also surprised it type checks).

magnus-madsen: (This pattern repeats-- but k should only be called once, and it should always be called.).

sockmaster27: > Why call k here? (I am also surprised it type checks). (EDIT: Its because both args are Unit)

But the continuation takes `Result[IOError, Unit]`, right? So why can we just pass it unit?

magnus-madsen: You figured it out right? :)

main/src/library/FilePathWithError.flix
```diff
@@ -0,0 +1,481 @@
...
+        try {
+            Ok(f(x))
+        } with FilePath {
+            def exists(f, k) = {
+                let res = try {
+                    Ok(Files.exists(Paths.get(f), (...{}: Vector[LinkOption])))
+                } catch {
+                    case ex: InvalidPathException => Err(IOError.Generic(ex.getMessage()))
+                };
+                k(res)
```
magnus-madsen, line 193: I am thinking it would be cleaner with out the res local variable. Then you can immediately see `k(Ok(...))` or `k(Err(...))``

magnus-madsen: But we can do this refactor as part of the split.

magnus-madsen (approved):

main/src/library/FilePathWithError.flix
sockmaster27: I made it work, but why was this not caught?

magnus-madsen: I think there are still some bugs (?)

PR #9103, opened 2024-11-03
chore: code cleanup

magnus-madsen:


magnus-madsen:
Found by Intellij IDEA, but each manually inspected.

PR #9121, opened 2024-11-05
test: add tests for large literals

magnus-madsen:


JonathanStarup:
I've fixed the fuel issue in #9151

magnus-madsen:
> I've fixed the fuel issue in #9151. It can parse infinite lists now

Thanks- so with this @jaschdoc can now do the rewrites in Desugar.

PR #9128, opened 2024-11-06
fix: automatically introduce empty array for VarArgs

magnus-madsen:


mlutze (commented):
Can we get a test?

Should there not be a corresponding change to type inference?

PR #9134, opened 2024-11-06
feat: improve `instanceof` parsing

JonathanStarup:
Improve errors and avoid pickQName
```
>> Expected <QName>.

1 | def main(): Int32 = 32 instanceof java.lang.Object
                                      ^^^^^^^^^
                                      Here
Hint: Use a single unqualified Java type like 'Object' instead of 'java.lang.object'.
```

```
>> Unexpected qualified name. Java names must be imported, e.g., `import java.lang.Object`.

1 | def main(): Int32 = 32 instanceof Java.Lang.Object
                                      ^^^^^^^^^^^^^^^^
                                      illegal qualified name
```


magnus-madsen (approved):

magnus-madsen:
This helps @jaschdoc right?

PR #9135, opened 2024-11-06
feat: remove `Validation` from `pick` in `Weeder2`

jaschdoc:
Supersedes https://github.com/flix/flix/pull/9081


magnus-madsen:
!!!jar

magnus-madsen:
@jaschdoc It seems the fuzzer found an error.

magnus-madsen:
Ignore my previous comment.

magnus-madsen:
Consider the program:

```
enum Foo {
  case case
}
```

I don't see how we can expect the tree to always contain an ident :-/

magnus-madsen:
If you look at Weeder2, see how pickExpr and pickType are implemented. They always account for an error tree.

This strongly suggests that must deal with the fact that Name.Ident etc can be absent.

But how to do that is the question. It seems painful.

magnus-madsen:
I have made https://github.com/flix/flix/pull/9139

It is similar to a solution you had earlier. But now I understand it, and it is just a stop gap until we figure out a better solution.

Issue #9123, opened 2024-11-05
Run `FuzzerSuite` in parallel to all other tests on GitHub CI


magnus-madsen:
I think it makes sense to separate out `FuzzerSuite` (and only fuzzersuite!)

Then we can crank up the time it spends so it equals that of TestAll

PR #9136, opened 2024-11-06
ci: run `FuzzerSuite` in parallel with more gas

jaschdoc:
Based on https://github.com/flix/flix/pull/9083
Closes https://github.com/flix/flix/issues/9123


jaschdoc:
I do not understand why `build` is still pending... it was the same issue on the other PR

mlutze:
I think it is because an internal configuration depends on a job called "build", but here you have renamed it to "build and test".
If you name it back, we can do the name change separately

jaschdoc:
> I think it is because an internal configuration depends on a job called "build", but here you have renamed it to "build and test".
> If you name it back, we can do the name change separately 

The same thing happened before I renamed it

jaschdoc:
> I think it is because an internal configuration depends on a job called "build", but here you have renamed it to "build and test".
> If you name it back, we can do the name change separately 

The same thing happened before I renamed it

magnus-madsen:
I cannot even restart it. Its like it does not exist. But perhaps try as Matt suggests and rename it back?

jaschdoc:
Sure :)

jaschdoc:
Oh we found an error in the compiler I think

jaschdoc:
Okay now we are just down to the weird `build` issue and the expected failures from community build and IDE tests

jaschdoc:
Ok I will just split them into separate yaml files

jaschdoc:
If you want to, you can make the fuzzing job required

magnus-madsen:
> If you want to, you can make the fuzzing job required

Lets leave it optional for now.

build.gradle
```diff
@@ -104,6 +104,19 @@ tasks.register('testAll', JavaExec) {
     standardInput = System.in
 }
 
+task testFuzzerSuite(dependsOn: ['testClasses'], type: JavaExec) {
+    mainClass = 'org.scalatest.tools.Runner'
+    // Note: We do not use the -P (parallelism) flag for two reasons:
```
magnus-madsen, line 109: I would drop this comment since this is just a single test suite.

build.gradle
```diff
@@ -104,6 +104,19 @@ tasks.register('testAll', JavaExec) {
...
+    // (A) It may use excessive amounts of memory (running many instances of the compiler).
+    // (B) It may garble the order of the output (when an error occurs).
+    // Of course we still benefit from parallelism inside Flix itself.
+    args = ['-s', 'flix.fuzzers.FuzzerSuite', '-o']
+    // We also enforce that all tests can complete with 4GB of memory:
```
magnus-madsen, line 114: Drop this comment and the memory requirement too.

magnus-madsen (approved):

Issue #9118, opened 2024-11-05
Improve `MagicCompleter`


magnus-madsen:
We want to be able to support e.g.:

```scala
enum Foo {
    case Bar,
    case Baz,
    case Quix(String)
}

def f(): Foo = Foo.Bar

def b(www: Foo): Unit = f().ma
```

We probably need to at least:

- Rework to be guided by `TypeError.FieldNotFound` 
- Possibly need to return a richer form of text edit.

magnus-madsen:
@stormckey I wrote some thoughts.

PR #9137, opened 2024-11-07
feat: improve magic match

stormckey:
Fulfilled the two improvement suggestions in #9118.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
@@ -67,7 +67,7 @@ object CompletionProvider {
       //
       case SyntacticContext.Expr.Constraint => PredicateCompleter.getCompletions(ctx) ++ KeywordCompleter.getConstraintKeywords
       case SyntacticContext.Expr.InvokeMethod(tpe, name) => InvokeMethodCompleter.getCompletions(tpe, name)
-      case SyntacticContext.Expr.AccessField(tpe, name) => MagicMatchCompleter.getCompletions(name.name, tpe, ctx)
+      case SyntacticContext.Expr.AccessField(tpe, name) => MagicMatchCompleter.getCompletions(tpe, ctx)
       case SyntacticContext.Expr.StaticFieldOrMethod(e) => GetStaticFieldCompleter.getCompletions(e) ++ InvokeStaticMethodCompleter.getCompletions(e)
       case SyntacticContext.Expr.StructAccess(e) => StructFieldCompleter.getCompletions(e, root)
       case _: SyntacticContext.Expr => ExprCompleter.getCompletions(ctx)
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
@@ -67,7 +67,6 @@ object CompletionProvider {
       //
       case SyntacticContext.Expr.Constraint => PredicateCompleter.getCompletions(ctx) ++ KeywordCompleter.getConstraintKeywords
       case SyntacticContext.Expr.InvokeMethod(tpe, name) => InvokeMethodCompleter.getCompletions(tpe, name)
-      case SyntacticContext.Expr.AccessField(tpe, name) => MagicMatchCompleter.getCompletions(tpe, ctx)
       case SyntacticContext.Expr.StaticFieldOrMethod(e) => GetStaticFieldCompleter.getCompletions(e) ++ InvokeStaticMethodCompleter.getCompletions(e)
       case SyntacticContext.Expr.StructAccess(e) => StructFieldCompleter.getCompletions(e, root)
       case _: SyntacticContext.Expr => ExprCompleter.getCompletions(ctx)
@@ -129,6 +128,7 @@ object CompletionProvider {
       // Imports.
       //
       case err: ResolutionError.UndefinedJvmClass => ImportCompleter.getCompletions(err)
+      case err: TypeError.FieldNotFound => MagicMatchCompleter.getCompletions(err.tpe, ctx)
 
       case _ => Nil
     })
@@ -217,7 +217,7 @@ object CompletionProvider {
       case WeederError.UndefinedAnnotation(_, _) => (1, SyntacticContext.Decl.Module)
       case err: ResolutionError.UndefinedJvmStaticField => (1, SyntacticContext.Expr.StaticFieldOrMethod(err))
       case err: TypeError.MethodNotFound => (1, SyntacticContext.Expr.InvokeMethod(err.tpe, err.methodName))
-      case err: TypeError.FieldNotFound => (1, SyntacticContext.Expr.AccessField(err.tpe, err.fieldName))
+      case err: TypeError.FieldNotFound => (1, SyntacticContext.Expr.InvokeMethod(err.tpe, err.fieldName))
       case err: ResolutionError.UndefinedStructField => (1, SyntacticContext.Expr.StructAccess(err))
       case err: ParseError => (5, err.sctx)
       case _ => (999, SyntacticContext.Unknown)
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
@@ -128,7 +128,7 @@ object CompletionProvider {
       // Imports.
       //
       case err: ResolutionError.UndefinedJvmClass => ImportCompleter.getCompletions(err)
-      case err: TypeError.FieldNotFound => MagicMatchCompleter.getCompletions(err.tpe, ctx)
+      case err: TypeError.FieldNotFound => MagicMatchCompleter.getCompletions(err)
 
       case _ => Nil
     })
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
@@ -66,8 +66,8 @@ object CompletionProvider {
       // Expressions.
       //
       case SyntacticContext.Expr.Constraint => PredicateCompleter.getCompletions(ctx) ++ KeywordCompleter.getConstraintKeywords
-      case SyntacticContext.Expr.InvokeMethod(tpe, name) =>
-        MagicMatchCompleter.getCompletions(name.name, tpe, ctx) ++ InvokeMethodCompleter.getCompletions(tpe, name)
+      case SyntacticContext.Expr.InvokeMethod(tpe, name) => InvokeMethodCompleter.getCompletions(tpe, name)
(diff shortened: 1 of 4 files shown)


main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
```diff
@@ -524,6 +534,15 @@ object Completion {
...
   case class SnippetCompletion(name: String, snippet: String, documentation: String) extends Completion
 
+  /**
+   * Represents a Snippet completion
+   *
+   * @param name          the name of the snippet.
+   * @param textEdit      the TextEdit.
+   * @param documentation a human-readable string that represents a doc-comment.
+   */
+  case class MagicMatchCompletion(name: String, textEdit: TextEdit, documentation: String) extends Completion
```
magnus-madsen, line 544: The idea with this case class would be not to store the textEdit but rather to store information enough to compute the text from that information. Otherwise the class does not do much. Makes sense?

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -217,7 +217,7 @@ object CompletionProvider {
       case WeederError.UndefinedAnnotation(_, _) => (1, SyntacticContext.Decl.Module)
       case err: ResolutionError.UndefinedJvmStaticField => (1, SyntacticContext.Expr.StaticFieldOrMethod(err))
       case err: TypeError.MethodNotFound => (1, SyntacticContext.Expr.InvokeMethod(err.tpe, err.methodName))
-      case err: TypeError.FieldNotFound => (1, SyntacticContext.Expr.InvokeMethod(err.tpe, err.fieldName))
+      case err: TypeError.FieldNotFound => (1, SyntacticContext.Expr.AccessField(err.tpe, err.fieldName))
```
magnus-madsen, line 220: You don't need to extend SyntacticContext.

Instead, try to extend the getSemanticCompletions function.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/MagicMatchCompleter.scala
```diff
@@ -33,16 +34,16 @@ object MagicMatchCompleter {
     *   case Blue => ???
     * }
     */
-  def getCompletions(name: String, tpe: Type ,ctx: CompletionContext)(implicit root: TypedAst.Root): Iterable[Completion] = {
-    if (!"match".startsWith(name)) return Nil
-
-    val ident = extractIdentifier(ctx.word)
+  def getCompletions(tpe: Type ,ctx: CompletionContext)(implicit root: TypedAst.Root): Iterable[Completion] = {
+    val ident = extractIdentifier(ctx)
+    val range = extractRange(ctx)
```
magnus-madsen, line 39: Instead of thinking in strings, we should try to rely on the source location in the FieldNotFound error. 

Then we can simply insert "match" right before that location and insert "{ case .." right after.

magnus-madsen: If it helps, the goal is very much to __avoid__ lots of string computations.

magnus-madsen:
Remember, ultimately the goal is to support `<any expression>.match` which we cannot solve purely hacking with strings. We have to rely the parser producing source locations, the type inference giving us an error message to build on, and then constructing (two?) text edits.

stormckey:
> Remember, ultimately the goal is to support `<any expression>.match` which we cannot solve purely hacking with strings. We have to rely the parser producing source locations, the type inference giving us an error message to build on, and then constructing (two?) text edits.
> 
> EDIT: It seems LSP does allow us to send multiple text edits. But my point still remains. We should then try to use the source location to construct basically `"match" + loc.text + {...}`.

I’ve removed most string operations, except for extracting the identifier, where I delete the field to obtain it.

Currently, only one TextEdit is used, as each CompletionItem accepts only one. To allow adding two TextEdit instances simultaneously, we’ll need to modify CompletionItem accordingly.

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -66,8 +66,7 @@ object CompletionProvider {
       // Expressions.
       //
       case SyntacticContext.Expr.Constraint => PredicateCompleter.getCompletions(ctx) ++ KeywordCompleter.getConstraintKeywords
-      case SyntacticContext.Expr.InvokeMethod(tpe, name) =>
-        MagicMatchCompleter.getCompletions(name.name, tpe, ctx) ++ InvokeMethodCompleter.getCompletions(tpe, name)
+      case SyntacticContext.Expr.InvokeMethod(tpe, name) => InvokeMethodCompleter.getCompletions(tpe, name)
```
magnus-madsen, line 69: This should no longer be needed right?

magnus-madsen: My bad. I misread. The code is correct.

magnus-madsen:
It must be late for you. I think I need to help you by extending FieldNotFound.

In particular, I think you need something like:

```
(complexExp).mat
^^^^^^^^^^^^
  loc1
             ^^^
             loc2
```

That would help a lot right? Then you can just call loc1.text

stormckey:
> Can you handle the example in the ticket now?

Yes
image
image

stormckey:
> It must be late for you. I think I need to help you by extending FieldNotFound.
> 
> In particular, I think you need something like:
> 
> ```
> (complexExp).mat
> ^^^^^^^^^^^^
>   loc1
>              ^^^
>              loc2
> ```
> 
> That would help a lot right? Then you can just call loc1.text

Of course! It helps a lot.

magnus-madsen:
OK; I will make that change tonight. With that in hand, you should be able to handle all kinds of expressions 👍

magnus-madsen:
Here is the ticket: https://github.com/flix/flix/issues/9144 I will work on it later today, so by tomorrow you should be able to build on it.

magnus-madsen:
When this https://github.com/flix/flix/pull/9152 is merged, you should be able to continue.

stormckey:
> When this #9152 is merged, you should be able to continue.

Done. I have one remaining question: should we place the conversion function from sourceLocation to Range within MagicMatchCompleter, or would it be more versatile if positioned elsewhere?

magnus-madsen:
> > When this #9152 is merged, you should be able to continue.
> 
> Done. I have one remaining question: should we place the conversion function from sourceLocation to Range within MagicMatchCompleter, or would it be more versatile if positioned elsewhere?

Lets keep it there for now.

magnus-madsen:
!!!jar

magnus-madsen:
Ill take a look. Does it seem to work well?

magnus-madsen:
Looks like it works well. I will do more testing later. I do have one request. Would it be possible to align the => by adding extra spaces?

magnus-madsen:
Like so:

```
def b(www: Foo): Unit = match www {
    case Foo.Bar                            => ???
    case Foo.Baz                            => ???
    case Foo.Quix(_elem)                    => ???
    case Foo.Quuuux(_elem1, _elem2, _elem3) => ???
}
```

magnus-madsen:
I think this is tricky to do cleanly because of the placeholder braces, but can you come up with an elegant solution? I think you need to do some counting, but __NOT__ on the string, but rather on the case before  you construct the string.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/MagicMatchCompleter.scala
```diff
@@ -33,17 +36,16 @@ object MagicMatchCompleter {
...
-      case Some(sym) =>
-        val casesString = generateCasesString(root.enums(sym).cases)
-        val matchExpr = s"match $ident {\n$casesString}"
-        val prompt = s"$ident.match"
-        Completion.SnippetCompletion(prompt, matchExpr, "Expand to a full match expression.") :: Nil
-      case None => Nil
+  def getCompletions(err: TypeError.FieldNotFound)(implicit root: TypedAst.Root): Iterable[Completion] = {
+    for {
+      sym <- getEnumSym(err.tpe)
+      ident <- err.base.text
```
magnus-madsen, line 42: I would rename ident to exp or baseExp or base

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/MagicMatchCompleter.scala
```diff
@@ -64,13 +66,22 @@ object MagicMatchCompleter {
...
 
+
   /**
-    * Extract the substring before the last . as the identifier
-    */
-  private def extractIdentifier(word: String): String = {
-    word.substring(0, word.lastIndexOf("."))
+   * Converts a [[SourceLocation]] to an [[Range]].
+   */
+
```
magnus-madsen, line 73: extra linebreak

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/MagicMatchCompleter.scala
```diff
@@ -64,13 +66,22 @@ object MagicMatchCompleter {
...
+  def sourceLocation2Range(loc: SourceLocation): Range = {
```
magnus-madsen, line 74: private

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/MagicMatchCompleter.scala
```diff
@@ -64,13 +66,22 @@ object MagicMatchCompleter {
...
+    Range(sourcePosition2Position(loc.sp1), sourcePosition2Position(loc.sp2))
   }
 
+  /**
+   * Converts a [[SourcePosition]] to a [[Position]].
+   */
+
```
magnus-madsen, line 81: extra linebreak

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/MagicMatchCompleter.scala
```diff
@@ -64,13 +66,22 @@ object MagicMatchCompleter {
...
+  def sourcePosition2Position(pos: SourcePosition): Position =
```
magnus-madsen, line 82: private

magnus-madsen (commented):
Add minor comments

stormckey:
I've aligned the "=>" for all cases.

One point of uncertainty is whether we should replace the magic numbers in `getCaseLength` with meaningful constants. The challenge lies in the formula for tuples: it’s actually 2 + arity * 6 + (arity - 1), accounting for braces, elements, and commas. Using named constants for each part could make the expression verbose and potentially harder to read.

stormckey:
Here is the reuslt:
image

magnus-madsen (approved):

magnus-madsen:
Thanks- I have merged it. Will take a detailed look later.

(PS: Remember to mark conversations as resolved as you fix them.)

stormckey:
> (PS: Remember to mark conversations as resolved as you fix them.)

Got it. 

I've found a little bug and fixed it. Maybe you can continue to review the code and we can discuss in the new pr #9167.

magnus-madsen:
I'll take a look later. Please also look at the new ticket I created.

PR #9139, opened 2024-11-07
refactor: throw ICE in `pickIdent`

magnus-madsen:


main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -3067,14 +3067,23 @@ object Weeder2 {
...
 
   private def pickNameIdent(tree: Tree)(implicit sctx: SharedContext): Validation[Name.Ident, CompilationMessage] = {
-    mapN(pick(TreeKind.Ident, tree))(tokenToIdent)
+    tryPick(TreeKind.Ident, tree) match {
+      case None =>
+        // The tree is missing a child of kind `TreeKind.Ident`.
+        // We should return something like Name.IdentWithError, but we lack such a thing.
+        // Instead, we construct a bogus Name.Ident-- which is not great, but should not crash the compiler.
+        // We do not have to add a compiler error because the Parser should already have emitted one.
+        val ident = Name.Ident("", tree.loc)
```
mlutze, line 3076: Should this not have some other name? Something searchable?

magnus-madsen: Its because -- sometimes not always -- a missing TreeKind.Ident means that the user did not type anything. So the empty string is "correct" and works "correctly" with auto-complete. But more though is needed. This is just a stopgap.

jaschdoc: What about a fresh name?

mlutze (approved):

PR #9143, opened 2024-11-07
chore: remove TypeReconstruction2

mlutze:


magnus-madsen:
Why? Also, you will need to redo because I just merged some of Jonathans code :)

mlutze:
Just because it's an unnecessary pain for others at the moment.

magnus-madsen:
I think it's fine

PR #9153, opened 2024-11-07
fix: removed label visit and consumer from visitor and consumer interface

LoZander:
This PR removes the visit to labels from the LSP visitor. The inclusion of this visitation was a mistake since labels don't carry any interesting information in and of themselves, but a name and their location. In this sense, they are much like symbols, which we also do not visit. Where labels are of interest, they can be extracted from their parent node. In most cases where a label is of interest, the parent node context is likely needed anyways.

Related to #9114 and undoes #9145 and #9149


LoZander:
I'm sorry it got a little messy. I added the label visit prematurely ^^'
I thought it was necessary at the time.

magnus-madsen (approved):

PR #9157, opened 2024-11-08
feat: add `Http` effect

### LINKED ISSUE #9035: Introduce `Http` effect
Introduce an effect like:

```
type alias HttpResponse = {status = Int32, statusText = String, url = String, body = String}

eff Http {
    pub def get(url: String, headers: Map[String, String]): HttpResponse
}
```

We will probably need to iterate on the API.

sockmaster27:
Fixes #9035

What do we think of this design of having `Http.Response`?

Do we want to add handler in this PR or another?

### DIFF
--- a/main/src/library/Http.flix
+++ b/main/src/library/Http.flix
@@ -26,22 +26,20 @@ mod Http {
     ///
     /// A response from an HTTP request.
     ///
-    /// The response consists of: `(status, statusText, headers, body)`
-    ///
-    pub enum Response(Int32, String, String, String)
+    pub enum Response({status = Int32, statusText = String, headers = String, body = String})
 
     mod Response {
         pub def status(r: Response): Int32 = match r {
-            case Response(status, _, _, _) => status
+            case Response(x) => x#status
         }
         pub def statusText(r: Response): String = match r {
-            case Response(_, statusText, _, _) => statusText
+            case Response(x) => x#statusText
         }
-        pub def url(r: Response): String = match r {
-            case Response(_, _, url, _) => url
+        pub def headers(r: Response): String = match r {
+            case Response(x) => x#headers
         }
         pub def body(r: Response): String = match r {
-            case Response(_, _, _, body) => body
+            case Response(x) => x#body
         }
     }
 
--- a/main/src/library/Http.flix
+++ b/main/src/library/Http.flix
@@ -1,0 +1,48 @@
+/*
+ *  Copyright 2024 Holger Dal Mogensen
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


magnus-madsen:
@JonathanStarup @mlutze Design feedback?

I could see the argument for a struct-- but that won't work right now.

In the interim is this fine? Do we want a record inside the enum?

magnus-madsen:
(Lets do the handler later- first we need to design)

main/src/library/Http.flix
```diff
@@ -0,0 +1,48 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+///
+/// The effect used to interact with the HTTP protocol.
+///
+eff Http {
+    pub def get(url: String, headers: Map[String, String]): Http.Response
```
magnus-madsen, line 21: I think we also want: head, post, put, and delete

main/src/library/Http.flix
```diff
@@ -0,0 +1,48 @@
...
+}
+
+mod Http {
+
+    ///
+    /// A response from an HTTP request.
+    ///
+    /// The response consists of: `(status, statusText, headers, body)`
```
magnus-madsen, line 29: The response headers should be a map right?

main/src/library/Http.flix
```diff
@@ -0,0 +1,48 @@
...
+    ///
+    pub enum Response(Int32, String, String, String)
```
magnus-madsen, line 31: Can the response be binary? Or would it then be encoded somehow?

main/src/library/Http.flix
magnus-madsen: Possibly that could wait till iteration 2

main/src/library/Http.flix
sockmaster27: The whole message is always encoded as a string so that would be the most general case.

However, it looks like there are different cases: https://developer.mozilla.org/en-US/docs/Web/HTTP/Messages#body_2

There's 
- the regular single string
- the stream of unknown length, (but this is disallowed with HTTP/2)
- the multiple-resource body, with multiple sections delimited by `--myBoundary`, but even MDN admits that these are rare

I definitely think we can just handle the third case by letting users manually decode the string, but I'm not sure about the second case.

magnus-madsen: OK. Lets just support a string for now.

main/src/library/Http.flix
sockmaster27: Alternatively, we could add an `HttpMethod` enum and take it as a parameter, while renaming the funciton to `request`.

magnus-madsen: Lets keep it like this for now.

magnus-madsen (approved):

PR #9162, opened 2024-11-08
fix: check entrypoint for primitive effects

### LINKED ISSUE #8921: Unsound Entrypoint effect check
Code here https://github.com/flix/flix/blob/7b590a8a2c6bff1c172fdfd8619444401f9cfa9e/main/src/ca/uwaterloo/flix/language/phase/EntryPoint.scala#L219

```scala
private def isBadEntryPointEffect(declaredEff: Type) = {
    declaredEff.effects.exists(sym => !Symbol.isBaseEff(sym))
  }
```

This code allows `not Crash` which would allow control effects in the entrypoint (it also allows `ef` but that doesn't matter as much)

JonathanStarup:
This is is a working but unpolished PR since:

- In EntryPoint, when we add our own main, we check main and only main
- In Safety, we check all entrypoints (which does not include the users main but our generated main)

So there is some awkward special treatment for main compared to tests/exports/..

Fixes #8921


main/src/ca/uwaterloo/flix/language/phase/EntryPoint.scala
```diff
@@ -216,8 +217,46 @@ object EntryPoint {
   /**
     * Returns `true` iff `declaredEff` is not `Pure` and contains some non-primitive effect.
     */
-  private def isBadEntryPointEffect(declaredEff: Type) = {
-    declaredEff.effects.exists(sym => !Symbol.isPrimitiveEff(sym))
+  private def isBadEntryPointEffect(declaredEff: Type): Boolean = {
```
JonathanStarup, line 220: @magnus-madsen This check is currently duplicated. I almost suggest that we just make it public in `Safety` and then `Entrypoint` calls that function. Then its also clear that `Entrypoint` does safety checking "in advance". Making the special case explicit

magnus-madsen: Can we just move the function to Type?

JonathanStarup: Hmm theyre a bit specfic. They require types that are ground, have no assoc, and no JVM types.

magnus-madsen: But it is a pure function just Type -> Bool right?

JonathanStarup: technically `Type -> Result[Option[CES], err]` or something, but yes

JonathanStarup: oh he whole function, not just the evaluater. yeah that one has a safe signature. i guess we can do that

magnus-madsen: I think phases should be independent.

JonathanStarup: me too, but pretending that they aren't seems less productive then making it explicit

JonathanStarup: A proper fix would be to delay the entrypoint function decoration I think. I has also obscured a couple bugs so it might be the correct choice

magnus-madsen: Sure, but in the interim we can move it to Type?

JonathanStarup: > pretending that they aren't [dependent] seems less productive then making it explicit

without thinking about code - the entrypoint conversion should first check safety before making a new entrypoint. Why not make the code reflect that directly? they somebody can't mess up the local code in entrypoint

magnus-madsen: Does entry point deal with all entry points and not just main?

In any case, if you can collect everything in EntryPoint then that is fine 👍

JonathanStarup:
oh, no, superseded by #9203

Issue #9117, opened 2024-11-05
Rewrite desugaring of literals to avoid StackOverflow


magnus-madsen:
The idea is to update:

- [ ] `desugarFCons`
- [ ] `desugarListLit`
- [ ] `desugarSetLit`
- [ ] `desugarMapLit`

(desugarFAppend is left unchanged)

to use the following translation scheme:

```scala
// 1 :: 2 :: 3 :: Nil -> (ONLY IF: the list goes deeper than 20)
let _ = Vector.toList(Vector#{1, 2, 3});

// List#{1, 2, 3} -> 
let _ = Vector.toList(Vector#{1, 2, 3});

// Set#{1, 2, 3} -> 
let _ = Vector.toSet(Vector#{1, 2, 3});

// Map#{"a" => 123, "b" => 456}
let _ = Vector.toMap(Vector#{("a", 123), ("b", 456)});
```

The use of vectors means that we avoid a very deep expression that triggers `StackOverflow`.

JonathanStarup:
Why conditionally? surely it is inlined/doesn't matter when the list is small

magnus-madsen:
> Why conditionally? surely it is inlined/doesn't matter when the list is small

Not easy to avoid the vector allocation. Also we must not rewrite the base case!!

JonathanStarup:
What do you mean by base case? 

What happens for `1 :: 2 :: .. :: 999 :: list`? That isn't covered by the above right?

JonathanStarup:
Also: the stack overflow is involving validation. It might be a significantly higher nesting limit when we get rid of validation

magnus-madsen:
> Also: the stack overflow is involving validation. It might be a significantly higher nesting limit when we get rid of validation

Not enough I am afraid. Its very short today, without validation its longer, but still very short.

> What happens for 1 :: 2 :: .. :: 999 :: list? That isn't covered by the above right?

That should work-- i.e. its longer than 20 so should be rewritten.

magnus-madsen:
> > > What happens for 1 :: 2 :: .. :: 999 :: list? That isn't covered by the above right?
> > 
> > 
> > That should work-- i.e. its longer than 20 so should be rewritten.
> 
> To what exactly? you can't write `Vector.toList(Vector#{1, 2, .., list})`

If is only cons and nil (i.e. a literal list) then I can. (This is where people get tripped up.)

JonathanStarup:
> without validation its longer, but still very short.

Have you tested it?

[...]

PR #9164, opened 2024-11-08
feat: desugar collection literals to vector literals

jaschdoc:
Closes https://github.com/flix/flix/issues/9117


main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
```diff
@@ -1315,12 +1315,18 @@ object Desugar {
 
   /**
     * Rewrites [[WeededAst.Expr.ListLit]] (`1 :: 2 :: Nil`) expression into `List.Nil` with `List.Cons`.
+    * If there are over 20 literals we translate it to `Vector.toList(Vector#{1, 2, ...})`.
```
magnus-madsen, line 1318: You are confusing two cases. THIS should ALWAYS be desugared.

The other case I am talking about is called FCons where you need a more tricky check.

magnus-madsen: This case is `List#{1, 2, 3}` which is always desugared.

main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
```diff
@@ -1295,13 +1295,26 @@ object Desugar {
 
   /**
     * Rewrites [[WeededAst.Expr.FCons]] (`x :: xs`) into a call to `List.Cons(x, xs)`.
+    * If there are over 20 literals we translate it to `Vector.toList(Vector#{1, 2, ...})`.
```
JonathanStarup, line 1298: I still think we should leave cons untouched - it's also redundant when we add the weeder recursion check

magnus-madsen: Its a footgun for our users. What's the downside?

magnus-madsen:
Feel free to copy from https://github.com/flix/flix/pull/9121 and then we can have this in one PR

main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
```diff
@@ -1295,11 +1297,58 @@ object Desugar {
...
+    *
+    * Additionally, if there are over 20 literals and the FCons sequence does not end with the literal `Nil`,
+    * we translate it to `List.append(Vector.toList(Vector#{literals}), nonLiteral)`.
```
magnus-madsen, line 1303: 👍

magnus-madsen (approved):

Issue #9166, opened 2024-11-09
Additional improvements to `MagicCompleter`


magnus-madsen:
We need one more pass over the magic completer. Here is my TODO list:

- [x] Fix a crash where if an enum is empty `enum Empty {}` we get a `java.lang.UnsupportedOperationException: empty.max`
- [x] Remove unused imports
- [x] Remove extra linebreaks between method comments and their bodies
- [x] Update the doc/comment for getCompletions to accurately reflect the indentation
- [x] Update getCaseLength to give an actual example with the computed lengths
- [x] Add explicit return types to all methods (e.g. generatedCasesString is missing one)
- [x] I think using a string builder in generateCasesString is the way to go
- [x] I would rename generateCasesString to patternMatchBody
- [x] I don't understand why padLhs needs to create a string and take its length. Why can it not be a pure arithmetic computation?

magnus-madsen:
@stormckey I have written down several suggestions for improvements. One concerns a serious bug, the others are about code quality and maintainability. If you could take a look, thanks!

stormckey:
Thanks for your review and all the suggestions! I will address these problems soon.

PR #9169, opened 2024-11-09
feat: improve magic match based on #9166

stormckey:
Resolved all the issues mentioned in #9166.
Note:
- Now we won't return a magic match completion if the enum is empty.
- Now we use `Math.log10` to compute the length of an Int instead of by building a string.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/MagicMatchCompleter.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/MagicMatchCompleter.scala
@@ -15,11 +15,9 @@
  */
 package ca.uwaterloo.flix.api.lsp.provider.completion
 
-import ca.uwaterloo.flix.language.ast.{Name, SourceLocation, SourcePosition, Symbol, Type, TypeConstructor, TypedAst}
+import ca.uwaterloo.flix.language.ast.{SourceLocation, SourcePosition, Symbol, Type, TypeConstructor, TypedAst}
 import ca.uwaterloo.flix.api.lsp.{Position, Range}
 import ca.uwaterloo.flix.language.errors.TypeError
-
-import scala.annotation.TypeConstraint
 
 object MagicMatchCompleter {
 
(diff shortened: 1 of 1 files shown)


main/src/ca/uwaterloo/flix/api/lsp/provider/completion/MagicMatchCompleter.scala
```diff
@@ -86,51 +84,55 @@ object MagicMatchCompleter {
...
-       val (lhs, rhs, nextZ) = createCase(sym, cas, z)
-       val paddedLhs = padLhs(lhs, maxCaseLength)
-       (acc + s"    case $paddedLhs => $rhs\n", nextZ)
-     }._1
+  private def patternMatchBody(cases: Map[Symbol.CaseSym, TypedAst.Case]): String = {
+    val maxCaseLength = cases.values.map(getCaseLength).max
+    val sb = new StringBuilder
+    cases.toList
+      .sortBy(_._1.loc)
+      .foldLeft(1) { case (z, (sym, cas)) =>
```
magnus-madsen, line 92: This longer has to be a fold right? It can even just be an ordinary for loop? I think

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/MagicMatchCompleter.scala
```diff
@@ -31,46 +29,46 @@ object MagicMatchCompleter {
...
+    *   case Blue  => ???
     * }
     */
-
   def getCompletions(err: TypeError.FieldNotFound)(implicit root: TypedAst.Root): Iterable[Completion] = {
     for {
       sym <- getEnumSym(err.tpe)
       baseExp <- err.base.text
+      cases = root.enums(sym).cases
+      if cases.nonEmpty
```
magnus-madsen, line 42: Add a comment at the of the line: // We skip empty enums (offer no suggestion)

magnus-madsen (commented):
OK. I think this is almost there.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/MagicMatchCompleter.scala
stormckey: The increment of z is not linear.
To use a for loop here, we'll need a mutable `var z` to track it. I've implemented it in the new commit. Shall we adopt this?

magnus-madsen: Yeah, I think its good. Shall I merge?

magnus-madsen (approved):

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/MagicMatchCompleter.scala
stormckey: OK, let's merge.

PR #9173, opened 2024-11-09
refactor: remove `SoftFailure`

jaschdoc:
Related to https://github.com/flix/flix/issues/8813


main/src/ca/uwaterloo/flix/api/Flix.scala
```diff
@@ -509,31 +509,52 @@ class Flix {
...
+    errors ++= weederErrors
 
     /** Remember to update [[AstPrinter]] about the list of phases. */
     val result = for {
-      afterWeeder <- Weeder2.run(afterReader, entryPoint, afterParser, cachedWeederAst, changeSet).withSoftFailures(readerErrors).withSoftFailures(lexerErrors).withSoftFailures(parserErrors)
+      afterWeeder <- weederValidation
       afterDesugar = Desugar.run(afterWeeder, cachedDesugarAst, changeSet)
-      (afterNamer, namerErrors) = Namer.run(afterDesugar)
+      (afterNamer, nameErrors) = Namer.run(afterDesugar)
+      _ = errors ++= nameErrors
```
magnus-madsen, line 531: What is this magic?... Ah its because its in a for-comprehension. But perhaps we should just call weeder and pattern match? And then again for Resolver. Yes its two layers of indentation, but it feels more clear to me, and then we can also drop the monadic stuff from Validation (AsMonad)

magnus-madsen: Ah- lets do it in the next then 👍

magnus-madsen (approved):

magnus-madsen:
Conflicted, but otherwise good to go

PR #9177, opened 2024-11-10
feat: extend `App` with additional effect handlers

magnus-madsen:


magnus-madsen:
Credit to @JonathanStarup for the flatMap

JonathanStarup (approved):

PR #9178, opened 2024-11-10
test: add testing of new file operations

### LINKED ISSUE #9131: Add basic tests for `FilePath` and friends
We don't need tests for everything, but I would like 5-8 tests like:

```
@Test
def exists01(): Unit \ IO = 
  FilePath.handle(() -> FilePath.exists("foo.txt"))()
```

or something like that.

sockmaster27:
Fixes  #9131


magnus-madsen:
I think this is good enough now. Later we can return to test every operator, but for now we just want to ensure that the whole setup with effect handlers works.

main/test/ca/uwaterloo/flix/library/TestFileRead.flix
```diff
@@ -0,0 +1,45 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+mod TestFileRead {
+
+    import java.io.IOException
+    import java.nio.file.Path
```
magnus-madsen, line 20: These are not needed, right?

sockmaster27:
> I think this is good enough now. Later we can return to test every operator, but for now we just want to ensure that the whole setup with effect handlers works.

Ah, ok. But we can still do one for each `FileRead`, `FileReadWithResult`, `FileWrite` and `FileWriteWithResult`, right?

magnus-madsen:
> > I think this is good enough now. Later we can return to test every operator, but for now we just want to ensure that the whole setup with effect handlers works.
> 
> Ah, ok. But we can still do one for each `FileRead`, `FileReadWithResult`, `FileWrite` and `FileWriteWithResult`, right?

Lets just do FileReadWithResult

magnus-madsen (approved):

magnus-madsen:
@sockmaster27 test failures

sockmaster27:
@magnus-madsen 
Looks like the working directory is different for IntelliJ tests and CI tests.
Is it ok if we leave out the test case of finding an existing file? Then it is ready.

magnus-madsen:
I don't care too much about the invalid path. Lets just check that an existing file exists and can be read?

sockmaster27:
@magnus-madsen RFR

main/test/ca/uwaterloo/flix/library/TestFileRead.flix
```diff
@@ -0,0 +1,74 @@
...
+
+
+    /////////////////////////////////////////////////////////////////////////////
+    // exists                                                                  //
+    /////////////////////////////////////////////////////////////////////////////
+
+    @Test
+    def exists01(): Bool \ IO = {
+        (
+            makeTempPath("some-file") >>= path ->
```
magnus-madsen, line 55: Lets make this some name that is definitely unique. I propose:

`flix-tmpfile-68975`

(I am just scared overwriting data)

main/test/ca/uwaterloo/flix/library/TestFileRead.flix
```diff
@@ -0,0 +1,74 @@
...
+    use Monad.{>>=}
+
+    import java.io.IOException
+    import java.nio.file.Path
+
+    ///
+    /// Helper - make a filepath within the temp directory.
+    /// Use `java.nio.file.Path` functions for platform independence.
+    ///
+    def makeTempPath(fileName: String): Result[IoError, String] = region rc {
```
magnus-madsen, line 30: Move this function after the tests

magnus-madsen: Do you think this should be a AE? Part of FileWrite?

magnus-madsen: AE = algebraic effect, i.e. part of FileWrite.

magnus-madsen (approved):

PR #9184, opened 2024-11-11
Experiment: Binding Blocks in Desugar

JonathanStarup:
Have flat lists of binders instead of nested singular binders

This seems pretty doable - one issue is that source locations get muddled between the different representations


magnus-madsen:
> This seems pretty doable - one issue is that source locations get muddled between the different representations

What do you mean?

magnus-madsen:
Ticket would be nice. But not sure how relevant this is any more.

PR #9192, opened 2024-11-11
refactor: use `pickExpr` in `visitScopeExpr`

jaschdoc:
Introduces a default parameter for `pickExpr` that can be overrriden so callers can pick expressions of a specific tree kind. Also changes `unfold` to immediately return the tree if it is not unfoldable. This causes `visitExpr` to just match directly on the kind of its parameter and call the corresponding function which in turn may call `visitExpr` but this time with another tree kind.

In short, it makes `pickExpr` and `visitExpr` a bit more flexible so we can start to move more uses of `pick` to `pickExpr`.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
@@ -834,8 +834,8 @@ object Weeder2 {
   }
 
   private object Exprs {
-    def pickExpr(tree: Tree)(implicit sctx: SharedContext): Validation[Expr, CompilationMessage] = {
-      val maybeExpression = tryPick(TreeKind.Expr.Expr, tree)
+    def pickExpr(tree: Tree, kind: TreeKind = TreeKind.Expr.Expr)(implicit sctx: SharedContext): Validation[Expr, CompilationMessage] = {
+      val maybeExpression = tryPick(kind, tree)
       flatMapN(
         traverseOpt(maybeExpression)(visitExpr)
       ) {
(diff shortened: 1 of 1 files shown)


main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -848,7 +848,6 @@ object Weeder2 {
     }
 
     def visitExpr(exprTree: Tree)(implicit sctx: SharedContext): Validation[Expr, CompilationMessage] = {
-      assert(exprTree.kind == TreeKind.Expr.Expr)
```
JonathanStarup, line 851: why do you remove this line?

magnus-madsen: I think we should keep this line.

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1095,7 +1094,7 @@ object Weeder2 {
 
     private def visitArguments(tree: Tree)(implicit sctx: SharedContext): Validation[List[Expr], CompilationMessage] = {
       mapN(
-        traverse(pickAll(TreeKind.Argument, tree))(pickExpr),
+        traverse(pickAll(TreeKind.Argument, tree))(pickExpr(_)),
```
magnus-madsen, line 1097: Scala requires this?

magnus-madsen:
I am sorry, I don't understand the idea

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
jaschdoc: No because now `visitExpr` can accept trees that do not have this kind.

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
jaschdoc: Yes, it's a type error otherwise (with the default parameter it apparently needs explicit application)

Issue #9125, opened 2024-11-06
Extend `CodeActionProvider`


magnus-madsen:
- [x] Add `mkNewStruct` which is conceptually similar to `mkNewEnum`
- [x] Add a `mkNewDef` which introduces a new def for an `UndefinedName` error
- [x] Add `mkImportJava` which introduces a new `import` statement (using `ClassList`)
- [x] Fix the TODO related to `onSameLine`.

magnus-madsen:
@stormckey When you are done with the magic completer, I think you should start looking at the first bullet point here.

The starting point is simply a program like:

```
def foo(): TypeNotYetDefined
```

Then when you click on TypeNotYetDefined a lightbulb pops up offering to define TypeNotYetDefined as an enum.

magnus-madsen:
We have the base versions and can now iterate.

PR #9197, opened 2024-11-12
feat: add mkNewStruct to CodeActionProvider

stormckey:
According to #9125, mkNewStruct is added to automatically introduce a new struct if a undefined type is encountered.

A minor problem: we will not check if the undefined type is followed by a `[r]` even thouth it's necessary for a struct, since this info is not included in `UndefinedType`. But we will add `[r]` manually to the generated struct definition, otherwise the flix compiler will crash.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
@@ -56,7 +56,7 @@ object CodeActionProvider {
         Nil
 
     case ResolutionError.UndefinedType(qn, _, loc) if onSameLine(range, loc) =>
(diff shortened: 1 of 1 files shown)


stormckey:
Here is the preview:
image

image

magnus-madsen:
> According to #9125, mkNewStruct is added to automatically introduce a new struct if a undefined type is encountered.
> 
> A minor problem: we will not check if the undefined type is followed by a `[r]` even thouth it's necessary for a struct, since this info is not included in `UndefinedType`. But we will add `[r]` manually to the generated struct definition, otherwise the flix compiler will crash.

Yeah, thats a limitation. I think we can live with it for now.

magnus-madsen (approved):

PR #9198, opened 2024-11-12
feat: add mkNewDef to CodeActionProvider

stormckey:
According to #9125, we introduce mkNewDef, a quickfix that will introduce a function definition for a undefined name.


stormckey:
> Can we add some snippets so that one can fill out the return type and body by pressing tab? Is that possible?

Maybe not? I'm not 100% sure. But when returning a CodeAction of kind quickfix, there is no place for us to make it a snippet. Also, there is no such CodeActionKind that is relative to snippet.

magnus-madsen:
> > Can we add some snippets so that one can fill out the return type and body by pressing tab? Is that possible?
> 
> Maybe not? I'm not 100% sure. But when returning a CodeAction of kind quickfix, there is no place for us to make it a snippet. Also, there is no such CodeActionKind that is relative to snippet.

What does the language server protocol spec say? Is it support or no? (It could be our data types are too simple...)

stormckey:
> What does the language server protocol spec say? Is it support or no? (It could be our data types are too simple...)

Still I can't find fields that can make quickfix a snippet. You could double check through this [link](https://code.visualstudio.com/api/references/vscode-api#CodeAction).

magnus-madsen:
Hmm okay. Could you try some other language server, e.g. the one for Rust and see what they support?

If we cannot add holes/snippets then I would say this feature is less useful.

stormckey:
Well it do works for rust, although it's not exactly snippet:
image

image

image

We can not jump between holes by striking tab, but it do help us select the region to be edited(only one place)

stormckey:
You can search `generate_function` on  the rust-analyzer [manual](https://rust-analyzer.github.io/manual.html). I can't tell which CodeActionKind they are using. But it's possible for us to first fill in the parameters automatically as shown in the manual.

magnus-madsen (approved):

PR #9200, opened 2024-11-12
feat: add mkImportJava for CodeActionProvider

stormckey:
According to #9125, we add `mkImportJava` to automatically import a undefined name/class if it's an existing class on the Java platform.

Note that a new object ClassMap is introduced to quickly lookup the path for a  given name. Currently `ClassMap` is built by processing the string list of java files from `ClassList`.


stormckey:
For undefined class:
image
image

For undefined name:
image
image

Currently we will not offer completion for cases like `Math.sin` since it is treated as a whole name by the error, which means we will need to process the string to find possible imports.

magnus-madsen:
Looks great. I added some comments on how to improve the code.

magnus-madsen:
Conflicted.

magnus-madsen:
I ran out of time today, so I will have to give more feedback tomorrow.

However, one thing you can already look into is inserting the import in the right place.

Here is what I mean:

```
mod Abc {
  mod Def {
    def f(): FileReader
  }
}
```

then the import should go here:

```
mod Abc {
  mod Def {
    import java......FileReader
    def f(): FileReader
  }
}
```

To help with this, I have added `EnclosingMod` to UndefinedName. See here:

https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/errors/ResolutionError.scala#L672

This tells you the position of "mod Def" in the above example. You can then insert the import on the next line (properly indented).

stormckey:
I've adjusted the place to insert the import, preview:
image
image

Note:
- I've added yet another `sourcePosition2Position` here, I think it's the third time.
- I've commented out the `mkImportJava` under `UndefinedJvmClass`, where the enclosing module is not available. Thus cases like `new File("data.txt")` is no longer supported. We need to first update `UndefinedJvmClass` to support this.

stormckey:
Known bug:

where there is an enclosing module from the start of the file, the position provided by `em` is  (1, 1). So we should add the import in the second line:
image

But when there are no enclosing module at all , the position provided by `em` is still (1, 1). However, this time we can't always insert in the second line:
image
image

I think we need a mechanism to distinguish them.

magnus-madsen:
> I've commented out the mkImportJava under UndefinedJvmClass, where the enclosing module is not available. Thus cases like new File("data.txt") is no longer supported. We need to first update UndefinedJvmClass to support this.

What is this about?

magnus-madsen:
For the known bug, perhaps I should give the source location of the entire module? Plus some information of whether it is root.

magnus-madsen:
We have added: https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/ast/shared/AnchorPosition.scala for your use.

magnus-madsen:
An UndefinedName/UndefinedType now has an anchor position and you can use that directly. It should have the "right" line number.

magnus-madsen:
> I've commented out the mkImportJava under UndefinedJvmClass, where the enclosing module is not available. Thus cases like new File("data.txt") is no longer supported. We need to first update UndefinedJvmClass to support this.

I understand your comment now. 

I think a very small extension here:

https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/Resolver.scala#L1333

similar to 

https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/Resolver.scala#L2262

Will get you exactly what you need.

main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
```diff
@@ -218,6 +222,44 @@ object CodeActionProvider {
...
+    * {{{
+    *   def foo(): = new File("data.txt")
+    * }}}
+    *
+    * where the undefined class `File` is a valid Java class, this code action proposes to add:
+    * {{{
+    *   import java.io.File
+    * }}}
+    */
+  private def mkImportJava(name: String, uri: String, em: EnclosingMod): List[CodeAction] = {
```
magnus-madsen, line 239: I renamed EnclosingMod to AnchorPosition and changed its definition, so you will need to merge/rebase and then refactor.

stormckey:
Now `mkImportJava` uses `AnchorPosition` :) . But there is still something to note:

When inserting in `(line+1, col)`, there are several problems:

No one could assure that there is nothing in this place(but it's abnormal), for example:
image
Then the import will break the comment:
image

Also, we need to manually add spaces after the import to make the indentation of next line works fine(because our import takes part of its indentation).

I used to insert in the beginning of the next line and add the spaces for import manually, which do you think is better?

magnus-madsen:
Is it possible to ask to insert and entire new line? Or maybe we need to insert "\n<spaces>import foo" ?

magnus-madsen:
(There is also a conflict for you to fix)

stormckey:
> Is it possible to ask to insert and entire new line? Or maybe we need to insert "\nimport foo" ?

If we want to avoid the first problem totally, we'll need to insert in the beginning the next line, so the text after insertion will be"import xxx\n//This is B!". Otherwise it will still be "//This\nimport xxx\nis B!"

stormckey:
> (There is also a conflict for you to fix)

It keeps to think of `mkImportJava` and `mkNewDef` as a conflict, which is weird. Why not just keep both🤔.

magnus-madsen:
(We can meet on Zoom now, if you have time)

stormckey:
> (We can meet on Zoom now, if you have time)

let'e meet.

magnus-madsen:
See https://github.com/flix/flix/pull/9222

stormckey:
My bad, this should be sp.col + 3 (short for - 1 + 4) to make the number of spaces correct.
https://github.com/flix/flix/blob/97b568c87d30da6bf1bb1fe3947c570d5e5fab18/main/src/ca/uwaterloo/flix/language/ast/shared/AnchorPosition.scala#L37-L43

Shall I change the code here in this pr?

magnus-madsen:
Yes, please make the change. Please write is as +4 -1

magnus-madsen:
Also post some screenshots when you have it working :)

stormckey:
Here is the preview:
image
image

on Root:
image
image

main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
```diff
@@ -43,8 +44,11 @@ object CodeActionProvider {
       else
         Nil
 
-    case ResolutionError.UndefinedName(qn, _, env, _, loc) if overlaps(range, loc) =>
-      mkNewDef(qn.ident.name, uri) :: {
+//    case ResolutionError.UndefinedJvmClass(name, _, loc) if onSameLine(range, loc) =>
```
magnus-madsen, line 47: Is this for a followup PR?

main/src/ca/uwaterloo/flix/util/ClassList.scala
```diff
@@ -15,22 +15,37 @@
...
+   * $ find . -name "*.java"|sort|grep -v "jdk/internal"|grep -v "module-info.java" | grep -v "package-info.java"
+   * }}}
+   */
   val TheList: List[String] = LocalResource.get("/src/ca/uwaterloo/flix/util/ClassList.txt").split('\n').map(_.trim).toList
 
+  /**
+   * The map from class name to class path, built from TheList
+   *
+   * Example:
+   *  - List -> java.util.List
```
magnus-madsen, line 42: The right should be a set, e.g. List -> {java.util.List} to highlight the multimappedness

magnus-madsen:
Thanks, I will take a more detailed look later.

main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
stormckey: Yeah, we'll need to first extend UndefinedJvmClass and then complete the code, as discussed above .

magnus-madsen: We do it after this has been reviewed and merged 👍

main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
```diff
@@ -249,6 +253,39 @@ object CodeActionProvider {
...
+    * where the undefined class `File` is a valid Java class, this code action proposes to add:
+    * {{{
+    *   import java.io.File
+    * }}}
+    */
+  private def mkImportJava(name: String, uri: String, ap: AnchorPosition): List[CodeAction] = {
+    val startPosition = Position(line = ap.line, character = ap.col)
+    val insertRange = Range(startPosition, startPosition)
+    val leadingSpaces = " " * ap.spaces
+    ClassList.TheMap.get(name).toList.flatten.map { path =>
```
magnus-madsen, line 274: toList.flatten.map -- I think this can be simplified? If nothing else, with just an ordinary pattern match

magnus-madsen (approved):

magnus-madsen:
I added one comment. But I will merge now, so you can fix it later.

main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
stormckey: The type after get is Option[Set[String]]. If using pattern match , we'll need first to process the option then map the Set to List, which I think is more complicated than this.🤔 Or you have other suggestions?

PR #9209, opened 2024-11-12
fix: record select source locations are wrong

LoZander:
This PR fixes an issue where the `SourceLocation` of `RecordSelect`s are wrong. The issue is that the beginning of the `SourceLocation`s currently start at the selected field instead of the record expression.

For instance, for
```flix
r#x
```
the `SourceLocation` only includes `x`.

This is particularly an issue for #9114.


main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1608,7 +1608,7 @@ object Weeder2 {
       mapN(pickExpr(tree)) {
         expr =>
           idents.foldLeft(expr) {
-            case (acc, ident) => Expr.RecordSelect(acc, Name.mkLabel(ident), ident.loc)
+            case (acc, ident) => Expr.RecordSelect(acc, Name.mkLabel(ident), tree.loc)
```
JonathanStarup, line 1611: does this fix work for `r#x#y#z`?

LoZander: Yeah, you're right. I think I've fixed it. We want the location that from tree.loc through ident.loc for each

LoZander: I believe I've fixed it now. I've tested it on a nested example (similar to what you proposed. Good catch. Before they all had the same sourcelocation, which ofc. makes no sense.

LoZander:
Not sure why the IDE tests are failing. But it doesn't seem related to this

magnus-madsen (approved):

PR #9210, opened 2024-11-13
feat: annotate more class effects

magnus-madsen:


main/src/library/Environment.flix
```diff
@@ -37,7 +37,7 @@ mod Environment {
     pub def getEnv(): Map[String, String] = region rc {
         try {
             let _ = rc;
-            let iter = unchecked_cast ((System.getenv().entrySet().iterator(): _ \ IO) as _ \ rc);
+            let iter = unchecked_cast ((System.getenv().entrySet().iterator()) as _ \ rc);
```
mlutze, line 40: I assume the signature of this function will change at some point?

mlutze (approved):

main/src/library/Environment.flix
magnus-madsen: The whole file will be deleted.

PR #9217, opened 2024-11-14
feat: introduce HTTP handler

### LINKED ISSUE #9035: Introduce `Http` effect

sockmaster27:
Fixes #9035

### DIFF
--- a/main/src/library/Http.flix
+++ b/main/src/library/Http.flix
@@ -66,10 +66,13 @@ mod Http {
                 let client = HttpClient.newHttpClient();
                 let response = client.send(request, BodyHandlers.ofString());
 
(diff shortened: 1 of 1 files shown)


sockmaster27:
@magnus-madsen 
It looks like headers should actually be a multimap

magnus-madsen:
> @magnus-madsen Looks like the status text is hard to get without libraries. Can we drop it?

OK for now

sockmaster27:
@magnus-madsen RFR

magnus-madsen (approved):

Issue #9202, opened 2024-11-12
Improve magic match completion when working with brackets


stormckey:
A problem with the current magic match completion is that it won't work with expressions in brackets.

Here is a example:
```
enum Color {
    case Red
}

def b(c: Color): Unit = 
    (c).ma 

```

Although a CompletionItem is sent back by the lsp, no suggestions pop up when typing `match`.

Another problem is that the generated match expression will not include the brackets, since the base expression stored in `FieldNotFound` ignores the brackets outside.

stormckey:
@magnus-madsen Here is the ticket, any ideas?

magnus-madsen:
I wonder if it is because the parser ignores the parens in this case?

What happens if you write:

```
f(1, 2, 3).mat
```

Does that work?

What about

```
(123, 456).mat
```
Does that work?

(By the way, we should probably extend the magic completer to support tuples as above)

stormckey:
```
f(1, 2, 3).mat
```
works just fine. But if the brackets affect only precedence, they seem to be ignored. By the way, generating match expression without brackets works fine too. It's just a formatting problem. The main problem, I think, may be that the completion suggestion is somewhat rejected by the client.

The tuple example doesn't work since we only offer completion for enums. So adding support for tuples means a pattern matching with only one case, right?

magnus-madsen:
>  But if the brackets affect only precedence, they seem to be ignored. 

Yes, thats probably the case.

> By the way, generating match expression without brackets works fine too. It's just a formatting problem. The main problem, I think, may be that the completion suggestion is somewhat rejected by the client.

Yeah, I wonder why VSCode skips it / does not show it?

stormckey:
> Do you think that would be possible?

I think it's possible. I'll start a new pr to support tuples later :)

JonathanStarup:
no judgement on their choice, but IntelliJ/scala does have the tuple version. looks cool though.

magnus-madsen:
> no judgement on their choice, but IntelliJ/scala does have the tuple version. looks cool though.

I don't understand.

JonathanStarup:
if you have `(e1, e2) mat|` then intellij doesnt suggest `match (exhaustive)`

magnus-madsen:
> if you have `(e1, e2) mat|` then intellij doesnt suggest `match (exhaustive)`

OK. But we can, and you think it would be cool?

PR #9218, opened 2024-11-14
feat: support magic match completion for tuple

stormckey:
According to #9202, we now support magic match for tuple!

Here is how it looks like:

image

image

Note:
- I introduce a new enum `Member` to represent either a enum or other type in the tuple.
- Maybe we could consider using the same marker for all the `_member2` placeholder so that user can edit them simultaneously.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/MagicMatchCompleter.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/MagicMatchCompleter.scala
@@ -36,24 +36,137 @@ object MagicMatchCompleter {
     */
   def getCompletions(err: TypeError.FieldNotFound)(implicit root: TypedAst.Root): Iterable[Completion] = {
     for {
-      sym <- getEnumSym(err.tpe)
       baseExp <- err.base.text
-      cases = root.enums(sym).cases
-      if cases.nonEmpty  // We skip empty enums (offer no suggestion)
+      patternMatchBody <- mkPatternMatchBody(err.tpe)
     } yield {
       val name = s"$baseExp.match"
       val range = sourceLocation2Range(err.loc)
-      val casesString = patternMatchBody(cases)
-      val snippet = s"match $baseExp {\n$casesString}"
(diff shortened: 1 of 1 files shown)


magnus-madsen:
!!!jar

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/MagicMatchCompleter.scala
```diff
@@ -36,78 +36,164 @@ object MagicMatchCompleter {
...
+    */
+  private def mkTupleMatchBody(members: List[List[Member]]): String = {
     val sb = new StringBuilder
-    var z = 1
-    for ((sym, cas) <- cases.toList.sortBy(_._1.loc)) {
-      val (lhs, rhs, newZ) = createCase(sym, cas, z)
-      val paddedLhs = padLhs(lhs, maxCaseLength)
-      sb.append(s"    case $paddedLhs => $rhs\n")
-      z = newZ
+    val maxCaseLength = members.map(getMemberLength).max
```
magnus-madsen, line 117: Do we know this .max cannot cause an exception?

magnus-madsen (commented):
The code seems to work well, but I am struggling a bit to understand it. 

How do you think we can document it better?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/MagicMatchCompleter.scala
stormckey: The .max will not cause an exception as long as the length of the tuple is not zero.

Is it possible for us to create an empty tuple in flix?

stormckey:
I've changed part of the name to make the code easier to understand.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/MagicMatchCompleter.scala
magnus-madsen: No. So that should be good.

magnus-madsen (approved):

PR #9219, opened 2024-11-14
fix: Enclosing mod position

JonathanStarup:


JonathanStarup:
Hi @stormckey, this PR is relevant to you, the use/import should now be inserted on the line before the source position

PR #9234, opened 2024-11-15
feat: Type difference

JonathanStarup:


magnus-madsen:
RFR?

main/src/ca/uwaterloo/flix/language/fmt/SimpleType.scala
```diff
@@ -570,6 +570,19 @@ object SimpleType {
...
 
+        case TypeConstructor.Difference =>
+          // collapse into a chain of intersections
+          t.typeArguments.map(visit).map(splitDifference) match {
+            // Case 1: No args. ? - ?
+            case Nil => Difference(Hole :: Hole :: Nil)
+            // Case 2: One arg. Take the left and put a hole at the end: tpe1 - tpe2 - ?
+            case args :: Nil => Difference(args :+ Hole)
+            // Case 3: Multiple args. Concatenate them: tpe1 - tpe2 - tpe3 - tpe4
+            case args1 :: args2 :: Nil => Difference(args1 ++ args2)
```
mlutze, line 581: I don't think it makes sense to do this transformation because difference is not associative. X - (Y - Z) ≠ X - Y - Z

JonathanStarup: good catch, this was the stuff i was concerned I missed

JonathanStarup: now its just for the left part

Issue #8791, opened 2024-10-05
Missing `import javax.swing.JFrame` autocomplete


JonathanStarup:
For some reason, it is not in the autocomplete, but the code works fine (Flix 0.51.0)

magnus-madsen:
I think its because that code is not part of our ClassList

JonathanStarup:
I didn't download anything. `java.awt` also isn't shown

magnus-madsen:
Did you check the class list file?

magnus-madsen:
https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/util/ClassList.scala

JonathanStarup:
Yeah they are missing from there. Is it autogenerated?

magnus-madsen:
Yes. It says how in the comments.

JonathanStarup:
We pull from `src/java.base/share/classes` but the classes I need are in `src/java.desktop/share/classes/`. There is also a bunch of other modules (https://docs.oracle.com/en/java/javase/22/docs/api/)

should we include all of them?

magnus-madsen:
I think only the most relevant.

magnus-madsen:
Could you extend with java.desktop and java.net for now. Thanks

magnus-madsen:
Fixed.

PR #9243, opened 2024-11-16
chore: update `ClassList.txt` with `java.desktop` and `java.net`

jaschdoc:
Maybe you should take a quick look and check that it is ok?

Closes https://github.com/flix/flix/issues/8791


magnus-madsen:
I am thinking we should drop the things in `com` and `sun`

magnus-madsen:
I am also confused why the file is sorted slightly different? I think maybe its a lower and uppercase issue and/or platform issue? (See the diff).

jaschdoc:
> I am thinking we should drop the things in `com` and `sun`

All of it or just the stuff related to `java.desktop`? We already had some `com/sun` classes in there.

magnus-madsen:
> > I am thinking we should drop the things in `com` and `sun`
> 
> All of it or just the stuff related to `java.desktop`? We already had some `com/sun` classes in there.

I think we can keep com and sun. Indeed we already have them.

Lets just see if we can figure out the sort issue.

jaschdoc:
> I think we can keep com and sun. Indeed we already have them.

Oh ok too late hehe. I can still revert it?

jaschdoc:
I think the sorting issue was both because it may have been case sensitive but also because the version of OpenJDK I pulled from may have new classes added to it

magnus-madsen:
> I think the sorting issue was both because it may have been case sensitive but also because the version of OpenJDK I pulled from may have had new classes added to it

You need to use 21.

jaschdoc:
Okay updated with files from the `jdk-21+0` tag

jaschdoc:
Oops updated with `jdk-21-ga` because that seems like the correct commit?

jaschdoc:
Can still revert

PR #9245, opened 2024-11-16
feat: n-ary enums (frontend)

JonathanStarup:
Updated version of #8265

- `Tag` does not mean `Tag(())` any more, so remove all special handling of `Tag(Unit)`
- `Tag(a, b)` is now a tag of two things instead of a tag of a tuple, so remove all special handling of `Tag(Tuple(..))`
- Convert the previous tuple behavior into the default term list behavior

Enums are now n-ary, and n can be zero
```scala
enum E {
  case A1 // no terms
  // case A2() // parser error
  case A3(Unit) // an actual unit term
  case A4(String, String) // two string terms
  case A5((String, String)) // one tuple term
  case A6(((String, String))) // one tuple term
}

mod E {
def _example1(): Unit = {
  let _ = A1; // Creating a case of zero terms
  let _ = A3(); // creating a case of one unit term
  let _ = A3(()); // creating a case of one unit term
  let _ = A4("a", "b"); // creating a case of two string terms
  let _ = A5(("a", "b")); // creating a case of one tuple term
  let _ = A5((("a", "b"))); // creating a case of one tuple term
  let _ = A6(("a", "b")); // creating a case of one tuple term
  let _ = A6((("a", "b"))); // creating a case of one tuple term
  ()
}
def _example2(): Unit = match ??? {
  case A1 => () // A1() is a type error
  case A3() => ()
  case A3(()) => ()
  case A4(_, _) => () // A4(_) is a type error
  case A5((_, _)) => ()
  case A5(_) => ()
  case A5(((_, _))) => ()
  case A6((_, _)) => ()
  case A6(_) => ()
  case A6(((_, _))) => ()
}
}
```


JonathanStarup:
@magnus-madsen rfr. There is as little documentation as before, but i should probably add some. you can look through the .flix files to see the changes in surface

magnus-madsen:
@JonathanStarup A quick glance suggests that the library changes look fine.

magnus-madsen:
In fact, I would say it sometimes clarifies things.

main/src/ca/uwaterloo/flix/language/phase/Deriver.scala
```diff
@@ -802,7 +803,7 @@ object Deriver {
         val (_, caze) = cases.head
 
         val outSym = new Symbol.AssocTypeSym(coerceTraitSym, "Out", loc)
-        val outTpe = caze.tpe
+        val outTpe = Type.mkTuplish(caze.tpes, loc)
```
JonathanStarup, line 806: Coerce will extract the old type, i.e. unit for empty, tpe for single term cases, and a tuple for n-ary enums

main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
```diff
@@ -167,11 +167,12 @@ object Kinder {
...
     */
   private def visitCase(caze0: ResolvedAst.Declaration.Case, tparams: List[KindedAst.TypeParam], resTpe: Type, kenv: KindEnv, taenv: Map[Symbol.TypeAliasSym, KindedAst.TypeAlias], root: ResolvedAst.Root)(implicit sctx: SharedContext, flix: Flix): KindedAst.Case = caze0 match {
-    case ResolvedAst.Declaration.Case(sym, tpe0, loc) =>
-      val t = visitType(tpe0, Kind.Star, kenv, taenv, root)
+    case ResolvedAst.Declaration.Case(sym, tpes0, loc) =>
+      val ts = tpes0.map(visitType(_, Kind.Star, kenv, taenv, root))
       val quants = tparams.map(_.sym)
-      val sc = Scheme(quants, Nil, Nil, Type.mkPureArrow(t, resTpe, sym.loc.asSynthetic))
-      KindedAst.Case(sym, t, sc, loc)
+      val schemeBase = if (ts.nonEmpty) Type.mkPureUncurriedArrow(ts, resTpe, sym.loc.asSynthetic) else resTpe
```
JonathanStarup, line 173: Type.mkPureUncurriedArrow makes a nonsense type for empty arguments. I could also make it return a non-arrow for empty arguments so I don't have to check multiple places?

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -466,17 +466,17 @@ object Lowering {
       val t = visitType(tpe)
       LoweredAst.Expr.Match(e, rs, t, eff, loc)
 
-    case TypedAst.Expr.Tag(sym, exp, tpe, eff, loc) =>
-      val e = visitExp(exp)
+    case TypedAst.Expr.Tag(sym, exps, tpe, eff, loc) =>
+      val es = exps.map(visitExp)
       val t = visitType(tpe)
-      LoweredAst.Expr.ApplyAtomic(AtomicOp.Tag(sym.sym), List(e), t, eff, loc)
+      LoweredAst.Expr.ApplyAtomic(AtomicOp.Tag(sym.sym), es, t, eff, loc)
```
JonathanStarup, line 472: Here the backend support for n-ary comes online

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1630,23 +1619,22 @@ object Lowering {
...
-    val tuple = mkTuple(hd :: tail :: Nil, loc)
-    mkTag(Enums.FList, "Cons", tuple, tail.tpe, loc)
+    mkTag(Enums.FList, "Cons", List(hd, tail), tail.tpe, loc)
   }
 
   /**
     * Returns a pure tag expression for the given `sym` and given `tag` with the given inner expression `exp`.
     */
-  private def mkTag(sym: Symbol.EnumSym, tag: String, exp: LoweredAst.Expr, tpe: Type, loc: SourceLocation): LoweredAst.Expr = {
+  private def mkTag(sym: Symbol.EnumSym, tag: String, exps: List[LoweredAst.Expr], tpe: Type, loc: SourceLocation): LoweredAst.Expr = {
```
JonathanStarup, line 1635: All the tags created in lowering now needs to follow the new rules

main/src/ca/uwaterloo/flix/language/phase/Monomorpher.scala
```diff
@@ -812,6 +812,12 @@ object Monomorpher {
 
   }
 
+  private def combineEnvs(envs: List[Map[Symbol.VarSym, Symbol.VarSym]]): Map[Symbol.VarSym, Symbol.VarSym] = {
```
JonathanStarup, line 815: Cases can have zero terms, so `envs.reduce(_ ++ _)` crashes sometimes because of an empty list

JonathanStarup:
@magnus-madsen i cleaned up a bit and now its actually mostly straight forward things. Ill update the main text with more information. Weeder is the worst part I think. It could be clearer but that would involve rewriting the parsing which i would prefer in another PR

main/src/ca/uwaterloo/flix/language/phase/Inliner.scala
```diff
@@ -148,10 +148,9 @@ object Inliner {
       op match {
         case AtomicOp.Untag(_, idx) =>
           val List(e) = es
-          // Inline expressions of the form Untag(Tag(e0, ... e_i, ..., e_n)) if all e_i are pure
           e match {
-            case LiftedAst.Expr.ApplyAtomic(AtomicOp.Tag(_), innerExps, _, _, _) if innerExps.map(_.purity).forall(Purity.isPure) =>
-              innerExps(idx)
+            case LiftedAst.Expr.ApplyAtomic(AtomicOp.Tag(_), innerExps, _, _, _) =>
+              ???
```
magnus-madsen: TODO

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -444,20 +444,22 @@ object ConstraintGen {
 
       case e: Expr.RestrictableChoose => RestrictableChooseConstraintGen.visitRestrictableChoose(e)
 
-      case KindedAst.Expr.Tag(symUse, exp, tvar, loc) =>
+      case KindedAst.Expr.Tag(symUse, exps, tvar, evar, loc) =>
```
mlutze, line 447: We might consider dividing up `ApplyTag` and `Tag`, but that can be future work.

mlutze (approved):

magnus-madsen:
I have not reviewed yet, but I will be asking for 5-10 tests

magnus-madsen (commented):
Phew. This was a lot of effort 👍

magnus-madsen:
I do wonder whether we should not fix the Parser as part of this PR? How difficult can it be, I think it should make parsing easier. Just ask for ArgumentList (or what it is called).

magnus-madsen:
@JonathanStarup Will you write a small paragraph about the implications of this PR and ping Stephen?

JonathanStarup:
> I do wonder whether we should not fix the Parser as part of this PR? How difficult can it be, I think it should make parsing easier. Just ask for ArgumentList (or what it is called).

I'd prefer not to - lets make a follow up PR and give it more more careful attention

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
JonathanStarup: This could conceptually be called ApplyTag. There is only fully applied tags just like with defs

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -1734,36 +1693,36 @@ object Resolver {
...
-      // Case 1: one expression. No tuple.
-      case e :: Nil =>
-        ResolvedAst.Expr.Tag(CaseSymUse(caze.sym, innerLoc), e, outerLoc)
-      // Case 2: multiple expressions. Make them a tuple
-      case es =>
-        val lastLoc = es.last.loc
-        val tupleLoc = SourceLocation(lastLoc.isReal, innerLoc.sp2, lastLoc.sp2)
-        val exp = ResolvedAst.Expr.Tuple(es, tupleLoc)
-        ResolvedAst.Expr.Tag(CaseSymUse(caze.sym, innerLoc), exp, outerLoc)
+  private def visitApplyTag(caze: NamedAst.Declaration.Case, exps: List[NamedAst.Expr], env: ListMap[String, Resolver.Resolution], innerLoc: SourceLocation, outerLoc: SourceLocation)(implicit scope: Scope, ns0: Name.NName, taenv: Map[Symbol.TypeAliasSym, ResolvedAst.Declaration.TypeAlias], sctx: SharedContext, root: NamedAst.Root, flix: Flix): Validation[ResolvedAst.Expr, ResolutionError] = {
```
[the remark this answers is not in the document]
JonathanStarup: This tag note is essentailly ApplyTag (i didn't rename since it would make the diff even bigger. We only have fully applied tags

JonathanStarup:
!!!jar

JonathanStarup:
Seems to work as expected. although is the rhs meant to be _elem0? 

image

magnus-madsen (approved):

magnus-madsen:
@JonathanStarup Oops. I broke something.

Issue #8798, opened 2024-10-05
Number literal source locations off-by-one problem


LoZander:
The `SourceLocation` for number literals erroneously includes the symbol right after. fx the location for `2` in `let x = 2;` would actually include `;`. This is not the case for all number literals. It's not the case for number literals with an explicit type, like `2i32`. It is however the case for both number literals with and without decimals.

It could be that this issue resides somewhere in in the `acceptNumber` function in the lexer. Maybe a problem with how advance and retreat is used here or affects locations. But this is just speculation.

jaschdoc:
Two questions:
1) How do you observe the behavior?
2) Does this also happen for floats / decimal numbers?

PR #9246, opened 2024-11-16
fix: produce correct source position for number literals

jaschdoc:
Tested this manually by hovering in VSCode. It works now (also for decimals which suffered from the same issue).

Closes https://github.com/flix/flix/issues/8798


magnus-madsen:
I don't understand the fix. Does this work for:

```
def main(): Unit \ IO = 
    let x = 123 + 234;
    println("abc ${x}")
```

jaschdoc:
Updated

magnus-madsen (approved):

Issue #9260, opened 2024-11-19
Rework GotoProvider to use LSP visitor instead of index


LoZander:

magnus-madsen:
Fixed.

PR #9261, opened 2024-11-19
feat: goto provider using lsp visitor

LoZander:
This PR changes the GotoProvider to use the LSP visitor instead of the index.

In theory, the approach is to do a singular search for where the cursor is, and then if there is an occurance of the symbol there that is defined elsewhere, we create a LocationLink (according to the LSP spec) from this occurrence of the symbol to it's definition site. Finding where a symbol is defined does not require an additional search, as this information is stored within the symbol.

In practice, however, since there two positions associated with the thin cursor, we might have to do up to two searches. If something is found on the right side of the cursor, we're done, if not, we have to search again on the left. Currently, we're naively doing both searches always, under the assumption that this additional cost is negligable. Further testing on this might be called for though.

Related to #9260.


LoZander:
I think this is actually more or less done, but I wanna give it another look over to make sure I didn't forget something.

LoZander:
Let me know if yall spot something you should be able invoke "go to definition" on that I've missed

magnus-madsen:
!!!jar

main/src/ca/uwaterloo/flix/api/lsp/provider/GotoProvider.scala
```diff
@@ -27,74 +29,129 @@ object GotoProvider {
...
+    case TraitConstraint.Head(sym, loc) => Some(mkGoto(LocationLink.fromTraitSym(sym, loc)))
+    case SymUse.SigSymUse(sym, loc) => Some(mkGoto(LocationLink.fromSigSym(sym, loc)))
+    // Type Vars
+    case Type.Var(sym, loc) => Some(mkGoto(LocationLink.fromTypeVarSym(sym, loc)))
+    // Vars
+    case TypedAst.Expr.Var(sym, _, loc) => Some(mkGoto(LocationLink.fromVarSym(sym, loc)))
+    case _ => None
+  }
+
+  private def isReal(x: AnyRef): Boolean = x match {
```
magnus-madsen, line 86: Is this function similar to the one from the other provider? Or does it have some subtle differences?

main/src/ca/uwaterloo/flix/api/lsp/provider/GotoProvider.scala
```diff
@@ -27,74 +29,129 @@ object GotoProvider {
...
+    case SymUse.StructFieldSymUse(_, loc) => loc.isReal
+    case SymUse.TraitSymUse(_, loc) => loc.isReal
+
+    case TraitConstraint(_, _, loc) => loc.isReal
+    case TraitConstraint.Head(_, loc) => loc.isReal
+
+    case AssocTypeConstructor(_, loc) => loc.isReal
+    case EqualityConstraint(_, _, _, loc) => loc.isReal
+
+    case _: Symbol => true
```
magnus-madsen, line 140: What symbol is not covered by the above?

magnus-madsen:
@LoZander Looks good! Polish (e.g. add documentation etc) and then we can merge 🚀

main/src/ca/uwaterloo/flix/api/lsp/provider/GotoProvider.scala
LoZander: I might not literally be the same, but it should be. It's really just answering the question: does this "thing" `x` have a real sourcelocation.

main/src/ca/uwaterloo/flix/api/lsp/provider/GotoProvider.scala
LoZander: Now that I think about it again, none. `x` should never be a Symbol at all, really, since the we're filtering things on the stack of the StackConsumer and the visitor never consumes a symbol (only ever a symbol occurrrence). So it should be safe to remove that line, which I'll do

LoZander:
There are still a couple of issues. Some stuff with record extensions/restrictions for instance, and an issue I just found regarding instance function definitions. These issues affect this provider but they are kinda beyond the scope of this PR. So I'd like to handle those in their own PRs and then maybe do a follow up PR to this if necessary.

As for this PR, I'd say it's done.

LoZander:
Also, currently goto definition for a call to a signature takes you to the trait sig definition site, not the instance. There are cases where we know exactly which type such a function is called on, and you might argue that in those cases, goto should take you to the corresponding instance def, not the trait sig. But that's sort of up for debate I suppose. Currently, it always just takes you to the trait sig, which I think is fine, at least for now

main/src/ca/uwaterloo/flix/api/lsp/provider/GotoProvider.scala
LoZander: Yes, we could extract it from both and have a single definition somewhere. Not sure where it should be though. Maybe it's own file somewhere in LSP directory?

LoZander: If `x` was just AST nodes,  I would say put it in TypedAst.scala, but it's not. It's also SymUse and such. Really, it's anything that the Consumer might consume.

LoZander:
> @LoZander Looks good! Polish (e.g. add documentation etc) and then we can merge 🚀

Yep, it's done. Barring my two previous comments, I'm happy with this as is. And I think those issues are better handled in other PRs. This, as it is now, is good for merging imo

main/src/ca/uwaterloo/flix/api/lsp/provider/GotoProvider.scala
LoZander: Since that change will refactor both providers, I think I'll do that in another PR :)

LoZander:
So I think this PR is done :)

magnus-madsen:
> Also, currently goto definition for a call to a signature takes you to the trait sig definition site, not the instance. There are cases where we know exactly which type such a function is called on, and you might argue that in those cases, goto should take you to the corresponding instance def, not the trait sig. But that's sort of up for debate I suppose. Currently, it always just takes you to the trait sig, which I think is fine, at least for now

This sounds fine to me. Thats what I would expect.

magnus-madsen:
> P.S. expect for that one onging conversation. That should be resolved first. I forgot about that for a sec :p

which?

magnus-madsen (approved):

LoZander:
> > P.S. expect for that one onging conversation. That should be resolved first. I forgot about that for a sec :p
> 
> which?

The one about extracting the `isReal` function. It was promptly resolved after that edit

PR #9263, opened 2024-11-19
feat: add available classes

### LINKED ISSUE #9228: Rework Root.names to hold two maps (and use in CodeActionProvider)
We are currently using `ClassList.TheList` for import quickfixes. But actually this is not entirely correct because JARs that are included in a project do not appear in `ClassList.TheList`. Instead,  `TypedAst.Root` has a field: `names: MultiMap[List[String], String]`. This maps each package name to its classes. The problem is that we need the reverse map (from a class name to a set of packages). We could compute this map from `names` all the time, but that is too slow.

Instead, I propose we introduce (in package `ast.shared`):

```scala
/**
  * @param byPackage a map from a package name to a set of classes (and interfaces) in that package.
  * @param byClass a map from a class (or interface) to the packages that it occurs in.
  */
case class AvailableClasses(byPackage: MultiMap[List[String], String], byClass: MultiMap[String, List[String]])
```

We then need to replace `private var knownClassesAndInterfaces` in `Flix.scala` with this new type.

Then we can use `root.availableClasses` in CodeActionProvider.

stormckey:
fixes #9228

I've already replaces `root.names` with `root.avialableClasses`, and it works fine.

I'm still not sure what we should do with `knownClassesAndInterfaces` in Flix.scala


main/src/ca/uwaterloo/flix/api/Flix.scala
```diff
@@ -80,6 +80,7 @@ class Flix {
     * The set of known Java classes and interfaces.
     */
   private var knownClassesAndInterfaces: MultiMap[List[String], String] = getJavaPlatformClassesAndInterfaces()
+//  private var knownClassesAndInterfaces: AvailableClasses = AvailableClasses() TODO: which root should we use?
```
magnus-madsen, line 83: We should construct and maintain it here. See how knownClassesAndInterfaces is used.

main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
```diff
@@ -267,18 +267,19 @@ object CodeActionProvider {
     *   import java.io.File
     * }}}
     */
-  private def mkImportJava(name: String, uri: String, ap: AnchorPosition): List[CodeAction] = {
+  private def mkImportJava(name: String, uri: String, ap: AnchorPosition)(implicit root: Root): List[CodeAction] = {
     val startPosition = Position(line = ap.line, character = ap.col)
     val insertRange = Range(startPosition, startPosition)
     val leadingSpaces = " " * ap.spaces
-    ClassList.TheMap.get(name).toList.flatten.map { path =>
```
magnus-madsen, line 274: TheMap can then be deleted from class list.

main/src/ca/uwaterloo/flix/language/ast/TypedAst.scala
```diff
@@ -43,7 +43,8 @@ object TypedAst {
                   sources: Map[Source, SourceLocation],
                   traitEnv: Map[Symbol.TraitSym, TraitContext],
                   eqEnv: ListMap[Symbol.AssocTypeSym, shared.AssocTypeDef],
-                  names: MultiMap[List[String], String],
+//                  names: MultiMap[List[String], String],
```
magnus-madsen, line 46: Delete lines

main/src/ca/uwaterloo/flix/language/ast/TypedAst.scala
```diff
@@ -43,7 +43,8 @@ object TypedAst {
...
+                  availableClasses: AvailableClasses,
```
magnus-madsen, line 47: This line should also be added to the previous ASTs:

- ParsedAst
- ...
- NamedAst
- ResolvedAst
- KindedAst

magnus-madsen: You can see how "names" is used-- which should be deleted.

magnus-madsen (commented):
I gave some feedback for iteration 2

magnus-madsen:
Thanks-- ill try to review soon. Note that there is a conflict :/

stormckey:
Should we keep the path as `List[String]`? Or we can concatenate it the form the full path, which I think would be more convenient in our use case.

magnus-madsen:
@stormckey CI still fails.

magnus-madsen (commented):
Thanks-- this is exactly what I wanted 👍 

I left some minor stylistic comments that should be easy to fix.

PR #9265, opened 2024-11-19
fix: Algebraic effects in REPL expressions

### LINKED ISSUE #8601: No check for unhandled effects on entrypoint
```
pub def main(): Unit \ Console = Console.println("hey")
```
Running in vscode (`:eval main()`) crashes at runtime

JonathanStarup:
Fixes #8601

This PR correctly checks for algebraic effects, but also disallows any non-primitive effect. I think this is okay, because how would you even use regions in the repl?

@magnus-madsen Env is missing in Prelude - do you have a documentation comment in mind that I can insert?

### DIFF
--- a/main/src/library/Prelude.flix
+++ b/main/src/library/Prelude.flix
@@ -18,6 +18,15 @@ pub enum Proxy[_] {
 }
 
 ///
+/// The uninterpretable `Env` effect.
+///
+/// The `Env` effect represents ..
+///
+/// The `Env` effect is uninterpretable which means that it cannot be handled.
+///
+pub eff Env
+
+///
 /// The uninterpretable `Exec` effect.
 ///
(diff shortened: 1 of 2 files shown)


magnus-madsen:
>@magnus-madsen Env is missing in Prelude - do you have a documentation comment in mind that I can insert?

Something about accessing properties of the system?

main/src/library/Prelude.flix
```diff
@@ -17,6 +17,15 @@ pub enum Proxy[_] {
     case Proxy
 }
 
+///
+/// The uninterpretable `Env` effect.
+///
+/// The `Env` effect represents ..
```
magnus-madsen, line 23: ... actions that read from the environment (e.g. information about the shell, operating system, or hardware)

magnus-madsen:
@JonathanStarup Ping

JonathanStarup:
@magnus-madsen sorry, got it done now

PR #9267, opened 2024-11-19
feat: Add CodeAction for UndefinedTag

### LINKED ISSUE #9230: Extend `CodeActionProvider` to support `UndefinedTag`
Example:

```
enum Color {
  case Red,
  case Green
}

def foo(): Color = Red // propose to "use Color.Red"
```

stormckey:
Relaed #9230

Preview:
image
image

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
@@ -53,8 +53,8 @@ object CodeActionProvider {
     case ResolutionError.UndefinedTrait(qn, ap,  _, loc) if overlaps(range, loc) =>
       mkUseTrait(qn.ident, uri, ap)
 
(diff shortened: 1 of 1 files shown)


stormckey:
Some questions:
- Should we match the case if they only have the same prefix?
- Should/Can we fill in the rest of the cases after choosing the CodeAction?

For the latter, I think the current `UndefinedTag` is not enough.

magnus-madsen:
>     * Should we match the case if they only have the same prefix?

What do you mean? We should just match on tag name.

>     * Should/Can we fill in the rest of the cases after choosing the CodeAction?

Do you mean to apply to more than one case? I think going one case at time is fine.

magnus-madsen (approved):

magnus-madsen:
I will merge this because it seems useful, but actually had something totally different in mind.

Instead of prefixing each case by the namespace, I want to use the tag, i.e. insert:

```
use Color.Red
```

Just as we have been doing in other tickets.

Makes sense?

Can we support that too?

PR #9268, opened 2024-11-20
feat: add CodeAction to use UndefinedTag

stormckey:
further update to #9267

Now if there is an undefined tag, we also propose to use it:

image

image

Some left problems:
- Shall we check by prefix, like also providing those for 'Re', 'Gree'
- Shall we propose to use all the cases in the same enum?

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
@@ -54,7 +54,7 @@ object CodeActionProvider {
       mkUseTrait(qn.ident, uri, ap)
 
     case ResolutionError.UndefinedTag(name, ap, _, loc) if overlaps(range, loc) =>
-      mkUseTag(name, uri, ap) ++ replaceFullTag(name, uri, loc)
+      mkUseTag(name, uri, ap) ++ mkQualifyTag(name, uri, loc)
 
     case ResolutionError.UndefinedType(qn, ap, loc) if overlaps(range, loc) =>
       mkUseType(qn.ident, uri, ap) ++ mkImportJava(qn.ident.name, uri, ap) ++ mkNewEnum(qn.ident.name, uri, ap) ++ mkNewStruct(qn.ident.name, uri, ap)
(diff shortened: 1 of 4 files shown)


main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
```diff
@@ -53,8 +53,8 @@ object CodeActionProvider {
     case ResolutionError.UndefinedTrait(qn, ap,  _, loc) if overlaps(range, loc) =>
       mkUseTrait(qn.ident, uri, ap)
 
-    case ResolutionError.UndefinedTag(name, _, loc) if overlaps(range, loc) =>
-      replaceFullTag(name, uri, loc)
+    case ResolutionError.UndefinedTag(name, ap, _, loc) if overlaps(range, loc) =>
+      mkUseTag(name, uri, ap) ++ replaceFullTag(name, uri, loc)
```
magnus-madsen, line 57: replaceFullTag -> mkQualifyTag ?

magnus-madsen: Also lets change "Replace" in the message to "Prefix with"

magnus-madsen: (I know these two comments applied more to the old PR, but I just saw these things now.)

stormckey:
> > Shall we propose to use all the cases in the same enum?
> 
> We could yes!

We could add that as an extra codeaction, then there will be 9 code actions in the example above.

Another choice is that we by default use all the cases when users choose the use the enum.

magnus-madsen:
> > > Shall we propose to use all the cases in the same enum?
> > 
> > 
> > We could yes!
> 
> We could add that as an extra codeaction, then there will be 9 code actions in the example above.
> 
> Another choice is that we by default use all the cases when users choose to use the enum.

Not sure. What do you think?

magnus-madsen:
Using the cases in magic match completer can be problematic. They could already be used and then it would be annoying.

magnus-madsen (approved):

Issue #9248, opened 2024-11-17
Rework `CompletionProvider` to be scope aware


magnus-madsen:
We would like that if we write:

```
use Validation.map5

def main(): Unit \ IO =
    let foobar = 123;
    map
```

then we should be suggested "map5". This will require us to extend `ResolutionError.UndefinedName` (among others) with additional information.

Action plan:

- [x] Move `Resolver.Resolution` to `ast.shared`.
- [x] Introduce `case class Scope(env: ListMap[String, Resolution])` in `ast.shared` and refactor `Resolver` to use it.
- [x] Add `Scope` to `UndefinedName` (and similar errors).
- [x] Extend `CompletionProvider` with some `ScopeCompletionProvider`. 
- [x] Drop `VarCompleter`

magnus-madsen:
@stormckey We should discuss this in more detail, but I wanted to write down the steps while I had them in my mind. I suggest to wait with this ticket until we have made more progress on the other tickets.

magnus-madsen:
I think each bullet point should be a separate PR to keep things simple-- but lets talk over Zoom.

stormckey:
oops, there is already a scope in ast.shared:
<img width="793" alt="image" src="https://github.com/user-attachments/assets/4652881f-40cf-4c46-90b9-17cf561548fb">

magnus-madsen:
[...]

PR #9276, opened 2024-11-20
feat: support completion with auto import

### LINKED ISSUE #9247: Add `additionalTextEdits` field to `CompletionItem`
See https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem

This will allow us to insert import/use declarations together with auto-completions in the future.

stormckey:
fixes #9247
Preview:
for UndefinedType
image

image

for UndefinedName
image

image

known bug:
Currently, the code imports the class every time we select the completion, regardless of whether the class is already imported or not.


magnus-madsen:
> Currently, the code imports the class every time we select the completion, regardless of whether the class is already imported or not.

Yes. This is a bit of show stopper. (It is super-annoying right?)

I think to fix it we need to fix https://github.com/flix/flix/issues/9248 first. Then we can filter on the already imported names.

magnus-madsen:
Can we extend the label, so it says Math (add import)?

stormckey:
> Can we extend the label, so it says Math (add import)?

Sure

magnus-madsen:
> > Can we extend the label, so it says Math (add import)?
> 
> Sure, like this? image

Yes, very nice.

What we need know is the scope-thing from the other ticket. This will allow us to filter suggestions where no import/use is needed.

magnus-madsen:
!!!jar

magnus-madsen:
We have to ensure a good ordering here:

image

magnus-madsen:
It feels pretty good to use, doesn't it?

magnus-madsen:
One more thing: For technical reasons we have to filter out classes that have a lowercase name. 

We could also fix that in ClassList.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/AutoImportCompleter.scala
```diff
@@ -0,0 +1,48 @@
+package ca.uwaterloo.flix.api.lsp.provider.completion
```
magnus-madsen, line 1: Add copyright header

stormckey:
The current order is:
image

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -122,14 +122,16 @@ object CompletionProvider {
   }
 
   private def getSemanticCompletions(ctx: CompletionContext, errors: List[CompilationMessage])(implicit root: TypedAst.Root): Iterable[Completion] = {
+    println(errors)
```
magnus-madsen, line 125: remove print debugging.

main/src/ca/uwaterloo/flix/util/ClassList.scala
```diff
@@ -38,6 +38,10 @@ object ClassList {
     *
     * Finally, remove the `./` prefix, e.g., `./com/sun/...` should be `com/sun/...`.
     *
+    * Note that we will remove all classes with a lowercase name.
     */
-  val TheList: List[String] = LocalResource.get("/src/ca/uwaterloo/flix/util/ClassList.txt").split('\n').map(_.trim).toList
+  val TheList: List[String] = {
```
magnus-madsen, line 43: I would revert the changes to ClassList.scala.

it should be the completer that filters out the non-capital classes. 

The reason is that for such classes we will eventually need to importer to be even smarter (I will explain later).

So If you could move the check

magnus-madsen: Move into AutoImportCompleter.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
```diff
@@ -157,6 +158,18 @@ sealed trait Completion {
         kind             = CompletionItemKind.Class
       )
 
+    case Completion.AutoImportCompletion(label, name, path, ap, documentation, shouldImport) =>
+      val priority = if (shouldImport) Priority.Low else Priority.High
```
magnus-madsen, line 162: I think I would make it 

if (shouldImport) Priority.Lower else Priority.Low

because the importing suggestions should not be high

magnus-madsen:
Ready to merge?

PR #9277, opened 2024-11-20
fix: ImportJava CodeAction now works with '.'

stormckey:
Now the CodeAction to import java class works with '.':
image

image

image

If there is no '.' in the undefined name, we will find the java class using the whole string. If there is '.' in the undefined name, we will find the java class using the substring before the first '.'

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
@@ -44,11 +44,11 @@ object CodeActionProvider {
     case ResolutionError.UndefinedStruct(qn, ap, loc) if overlaps(range, loc) =>
       mkNewStruct(qn.ident.name, uri, ap)
 
-    case ResolutionError.UndefinedJvmClass(name, ap, _, loc) if overlaps(range, loc) =>
-      mkImportJava(name, uri, ap)
+    case ResolutionError.UndefinedJvmClass(_, ap, _, loc) if overlaps(range, loc) =>
+      mkImportJava(loc.text, uri, ap)
 
     case ResolutionError.UndefinedName(qn, ap, env, _, loc) if overlaps(range, loc) =>
-      mkFixMisspelling(qn, loc, env, uri) ++ mkUseDef(qn.ident, uri, ap) ++ mkImportJava(qn.ident.name, uri, ap) ++ mkNewDef(qn.ident.name, uri, ap)
+      mkFixMisspelling(qn, loc, env, uri) ++ mkUseDef(qn.ident, uri, ap) ++ mkImportJava(loc.text, uri, ap) ++ mkNewDef(qn.ident.name, uri, ap)
 
     case ResolutionError.UndefinedTrait(qn, ap,  _, loc) if overlaps(range, loc) =>
       mkUseTrait(qn.ident, uri, ap)
@@ -57,7 +57,7 @@ object CodeActionProvider {
       mkUseTag(name, uri, ap) ++ mkQualifyTag(name, uri, loc)
 
     case ResolutionError.UndefinedType(qn, ap, loc) if overlaps(range, loc) =>
-      mkUseType(qn.ident, uri, ap) ++ mkImportJava(qn.ident.name, uri, ap) ++ mkNewEnum(qn.ident.name, uri, ap) ++ mkNewStruct(qn.ident.name, uri, ap)
+      mkUseType(qn.ident, uri, ap) ++ mkImportJava(loc.text, uri, ap) ++ mkNewEnum(qn.ident.name, uri, ap) ++ mkNewStruct(qn.ident.name, uri, ap)
 
     case TypeError.MissingInstanceEq(tpe, _, loc) if overlaps(range, loc) =>
       mkDeriveMissingEq(tpe, uri)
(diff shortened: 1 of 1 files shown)


magnus-madsen:
This is a nice idea! 

But can we do it without relying on loc.text? We really should not rely on source locations for the text (because it could contain all kinds of things we are not prepared to deal with, e.g. escape chars). We must use information from the error itself. So what is the extra information we need?

magnus-madsen:
@stormckey Can we use the qname.namespace? which is an NName

stormckey:
> @stormckey Can we use the qname.namespace? which is an NName

Oh yeah, it works.

magnus-madsen:
> Oh yeah, it works. But UndefinedJvmClass only has `name: String`, I think we don't support something like `new Map.Entry()`?

The `Map.Entry` is a corner case. Not sure what to do there (its an inner class).

But w.r.t. to `Math.sin` I think you can do it 👍

stormckey:
> The `Map.Entry` is a corner case. Not sure what to do there (its an inner class).

It will trigger an error, so I think this is not allowed:
image

stormckey:
Now it works with `qn.namespace`:
image
image
image

magnus-madsen:
Nice-- what's with the double import suggestion?

main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
```diff
@@ -325,16 +325,18 @@ object CodeActionProvider {
     *   import java.io.File
     * }}}
     */
-  private def mkImportJava(name: String, uri: String, ap: AnchorPosition)(implicit root: Root): List[CodeAction] =
-    root.availableClasses.byClass.get(name).toList.flatten.map { path =>
-      val completePath = path.mkString(".") + "." + name
+  private def mkImportJava(name: String, namespace: List[Name.Ident], uri: String, ap: AnchorPosition)(implicit root: Root): List[CodeAction] = {
```
magnus-madsen, line 328: I think we should update both the doc comment above and add a code comment below to explain what is happening.

main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
```diff
@@ -44,11 +44,11 @@ object CodeActionProvider {
     case ResolutionError.UndefinedStruct(qn, ap, loc) if overlaps(range, loc) =>
       mkNewStruct(qn.ident.name, uri, ap)
 
-    case ResolutionError.UndefinedJvmClass(name, ap, _, loc) if overlaps(range, loc) =>
-      mkImportJava(name, uri, ap)
+    case ResolutionError.UndefinedJvmClass(qn, ap, _, loc) if overlaps(range, loc) =>
+      mkImportJava(qn, Nil, uri, ap)
```
magnus-madsen, line 48: I think we can change mkImportJava to take a Name.QName as argument, yes?

stormckey: Then we'll need to construct `Name.QName` for `UndefinedJvmClass`, where we only have `name: String`.

Which do you think is better?

stormckey:
> Nice-- what's with the double import suggestion?

UndefinedName Error is triggered twice:
image

main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
stormckey: OK, let's use qn

stormckey:
Do we still need to add some comment inside `mkImporetJava`?

main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
magnus-madsen: Do we have to construct a fake QName? If so, it may be better to leave as is.

main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
```diff
@@ -312,29 +312,35 @@ object CodeActionProvider {
...
     * {{{
-    *   import java.io.File
+    *  import java.lang.Math
     * }}}
     */
-  private def mkImportJava(name: String, uri: String, ap: AnchorPosition)(implicit root: Root): List[CodeAction] =
-    root.availableClasses.byClass.get(name).toList.flatten.map { path =>
-      val completePath = path.mkString(".") + "." + name
+  private def mkImportJava(qn: Name.QName, uri: String, ap: AnchorPosition)(implicit root: Root): List[CodeAction] = {
+    val className = qn.namespace.idents.headOption.map(_.name).getOrElse(qn.ident.name)
```
magnus-madsen, line 333: I was just thinking a line like this could use a comment or example. All the headOptionMap getOrElse etc.

magnus-madsen:
This https://github.com/flix/flix/pull/9280 will fix the duplicate error problem.

magnus-madsen:
If you polish this a bit more then I will merge it.

main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
stormckey: I add some comment, or we can match over headOption to make this more clear.

magnus-madsen:
@stormckey I see some CI errors

magnus-madsen (approved):

PR #9285, opened 2024-11-20
fix: OccuranceAnalyzer

JonathanStarup:
This looks like a typo


magnus-madsen (approved):

magnus-madsen:
@jaschdoc Maybe this is the source of the bugs?

jaschdoc:
I don't think so but I will be very happy if it is. The analyzer does not perform any rewriting so matching on e or exp should be the same, no? :)

magnus-madsen:
> I don't think so but I will be very happy if it is. The analyzer does not perform any rewriting so matching on e or exp should be the same, no? :)

You tell me. You are the expert :)

PR #9289, opened 2024-11-20
refactor: remove App evar

JonathanStarup:
Removing the evar requires destructing an arrow type which reveals three enemies
- type aliases (not too bad)
- associated types (hmmm)
- wrong typing?? (how can it be a variable after substitution when its unified with an arrow type??)


magnus-madsen:
I think App is one of those cases where you probably want the evar...

PR #9296, opened 2024-11-21
fix: make `Throw` impure in `Simplifier`

jaschdoc:
I'm not sure if this is a bug or not, but it seems that if a `HoleError` is impure, then it should also hold for `Throw`?


magnus-madsen (approved):

magnus-madsen:
Yeah, I think so

JonathanStarup:
I think we should undo this - throw is already impure because it always has IO. Hole is just here to undo the lying a bit (it can be typed as pure) (although I'm unsure what that would do)

PR #9298, opened 2024-11-21
feat: simplify application evars

JonathanStarup:
for `Xperf --frontend --n 11` this PR moves the median throughput from 19.985 to  22.644 (+13%)

when you have an application we currently have
```
outsideconstraints ~ evar
evar ~ Union(.. funEff ..)
```
but with this PR it is instead (requiring TypeReduction to rebuild the union)
```
outsideconstraints ~ Union(.. funEff ..)
evar ~ funEff
```
which has the very trivial `var ~ var` constraint


JonathanStarup:
Here's some performance @magnus-madsen, no type destruction needed

magnus-madsen:
Interesting. It would be nice to know what would happen with the new solver.

JonathanStarup:
Yeah its equivalent so its in some sense up to the solver if it matters or not. But simple alias constraints seem easier no matter what

magnus-madsen:
Wait for new solver. Discuss in ticket. See if it is worth it.

PR #9300, opened 2024-11-21
feat: simplify region evars

JonathanStarup:
for `Xperf --frontend --n 11` this PR moves the median throughput from 21.228 to  20.410 (-4%)


magnus-madsen:
> for `Xperf --frontend --n 11` this PR moves the median throughput from 21.228 to 20.410 (-4%)

Hmm. You could try n 51 to see if it is really true that it is slower.

JonathanStarup:
I should probably also use the server instead of my laptop

magnus-madsen:
> I should probably also use the server instead of my laptop

Use server, fix n at 250, grab beverage of choice.

mlutze:
!!!rerun

magnus-madsen:
I am not against this, but it should be a principled an step-wise approach. Also we should have some evidence that it actually improves performance since it just complicates the code.

magnus-madsen:
Also probably best to wait for the new solver.

PR #9303, opened 2024-11-21
feat: singleton nullary enum cases

### LINKED ISSUE #9259: singleton enums
Follwing #9245 it should be easy to reuse a single object for tags without terms

This would make it viable to have `Unit` be a surface defined `enum Unit { case Unit }`

JonathanStarup:
Idea: for nullary enum tags like `Option.None`, make a `Option$dotNone$` class with a single tag instance. This means that we only have one `Option.None` object and that pattern match checking for `Option.None` can be done with `instanceOf Option$dotNone$`

Fixes #9259


magnus-madsen:
Could we call it Nullary everywhere?

JonathanStarup:
done

magnus-madsen (approved):

PR #9308, opened 2024-11-22
refactor: make `Environment` an algebraic effect

### LINKED ISSUE #9224: Refactor `Environment` to be an algebraic effect

sockmaster27:
Fixes  #9224


magnus-madsen:
@JonathanStarup @mlutze We have some design choices to make here. See the diff. Lets talk at some point.

magnus-madsen:
Marked as draft until we decide what to do about `threads()`

JonathanStarup:
I like it - clear environment dependence and it makes it possible to inject your own implementation

magnus-madsen:
> I like it - clear environment dependence and it makes it possible to inject your own implementation

Did you see the casts?

magnus-madsen:
(Wrong button, fat fingers)

JonathanStarup:
oh, right. The easy answer is just to not have the cast - but I suppose that touches a lot of code

magnus-madsen:
> oh, right. The easy answer is just to not have the cast - but I suppose that touches a lot of code

Do you want your Set.insert to have the Environment effect? :)

JonathanStarup:
I mean it is dependent on the environment 🤷 the consequences of the current design

mlutze:
> I mean it is dependent on the environment 🤷 the consequences of the current design

I'd say that its dependency on the environment is local, like a region. It's not observable from the outside.

JonathanStarup:
hmm right because time and cpu usage is not effects

sockmaster27:
Any conclusions?

main/src/library/DelayMap.flix
```diff
@@ -166,7 +166,8 @@ mod DelayMap {
                 if (useParallelEvaluation(m))
                     let h = (k, v) -> g(k, force v);
                     let DMap(t) = m;
-                    RedBlackTree.parCount(threads() - 1, h, t)
+                    let threads = unchecked_cast(Environment.handle(threads)() as _ \ {});
```
JonathanStarup, line 169: magnus has the final decision, but as matt suggested, if we put these two lines into a helper function, and use unchecked_cast around its body I think it could be okay. That function would then doc comment about why its okay to cast it away (System + IO is unobservable because performance is not observable in our model)

JonathanStarup: hmm its used with multiple methods, not as easy

magnus-madsen: I agree with @JonathanStarup 

Add a PRIVATE function to each of DelayList, DelayMap, and RedBlackTree:

```
def threads(): Int32 = unsafe ...
```

and use it there, so we don't have to use the Env effect.

main/src/library/Set.flix
```diff
@@ -229,7 +229,8 @@ mod Set {
                 if (useParallelEvaluation(s))
                     let h = (kl, _, kr, _) -> g(kl, kr);
                     let Set(t) = s;
-                    let o = RedBlackTree.parMinimumBy(threads() - 1, h, t);
+                    let threads = unchecked_cast(Environment.handle(threads)() as _ \ {});
+                    let o = RedBlackTree.parMinimumBy(threads - 1, h, t);
```
magnus-madsen, line 233: I think we should move the threads argument into RedBlackTree, so here we should just call parMinimumBy

main/src/library/Map.flix
```diff
@@ -236,7 +236,8 @@ mod Map {
                 if (useParallelEvaluation(m))
                     let h = (kl, _, kr, _) -> g(kl, kr);
                     let Map(t) = m;
-                    RedBlackTree.parMinimumBy(threads() - 1, h, t)
+                    let threads = unchecked_cast(Environment.handle(threads)() as _ \ {});
```
magnus-madsen, line 239: Same comment about not passing the number of threads here. RedBlackTree can get them by themselves.

main/src/library/DelayMap.flix
JonathanStarup: (i mean technically its not sound at that scope because the integer returned very much is dependent on Environment, its only after calling the relevant function and threads goes out of scope that the only observable effect is performance)

magnus-madsen:
Lets split this PR into smaller steps:

- [ ] Introduce a private `threads()` function in RedBlackTree where it is used in the parallel functions.
- [ ] Refactor so Set/Map do not pass any thread argument.
- [ ] Refactor Environment to be an AE

What's the story with Chalk, why does it need AE?

sockmaster27:
> What's the story with Chalk, why does it need AE?

It uses the environment to determine how to format the output for the given system.
I think it makes sense in principle to have this dependency declared, to make it clear that the string can't immediately be sent over the internet or anything like that.

sockmaster27:
>     * [ ]  Introduce a private `threads()` function in RedBlackTree where it is used in the parallel 

It's already using one, right? Or what do you mean?

sockmaster27:
What should we do about `DelayMap.forceAll`? It's the only method that doesn't use `RedBlackTree`s parallelism, and need to read threads by itself.

magnus-madsen:
> > ```
> > * [ ]  Introduce a private `threads()` function in RedBlackTree where it is used in the parallel 
> > ```
> 
> It's already using one, right? Or what do you mean?

It should not use Environment. Instead it should use Java interop + unsafe.

Then we can avoid Environment for these things.

sockmaster27:
>     * [ ]  Refactor so Set/Map do not pass any thread argument.

Did this part here: #9498

magnus-madsen:
@sockmaster27 Conflicted. Can we get this PR up to date and merged?

magnus-madsen:
This PR will soon block others, so it would be good to get it in soon. 

The alternative is that it will have even more conflicts.

sockmaster27:
@magnus-madsen 
RFM

magnus-madsen (approved):

PR #9313, opened 2024-11-24
docs: update Swing example

jaschdoc:
Now updated to the same style as the original example: https://docs.oracle.com/javase/tutorial/uiswing/examples/start/HelloWorldSwingProject/src/start/HelloWorldSwing.java

### DIFF
--- a/examples/swing/simple-swing-app.flix
+++ b/examples/swing/simple-swing-app.flix
@@ -4,7 +4,7 @@ import javax.swing.JLabel
 import javax.swing.SwingUtilities
 
 def createAndShowGUI(): Unit \ IO = {
-    let frame = new JFrame("Dialog App");
+    let frame = new JFrame("Hello App");
     frame.setDefaultCloseOperation(JFrame.EXIT_ON_CLOSE);
 
     let label = new JLabel("Hello World!");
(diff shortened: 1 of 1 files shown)


magnus-madsen:
Not sure I like that. As @JonathanStarup said, why not just do it in the straightforward way. It feels like THEIR docs are outdated.

jaschdoc:
Maybe `SwingUtilities.invokeLater` manages the lifecycle of Swing stuff? Also, if the app does several other things, you would not want the main thread to be handling UI stuff, right?

jaschdoc:
Of course this is just a hello world app but still

magnus-madsen:
@JonathanStarup Thoughts?

magnus-madsen:
I would prefer to keep it simple unless there is an argument against.

JonathanStarup:
maybe this is the more proper way after some quick reading. I think there is both pros for not using invokelater (simplicity) and for using it (showing people how to use new object, which is pretty relevant when using swing with listeners etc)

magnus-madsen:
> maybe this is the more proper way after some quick reading. I think there is both pros for not using invokelater (simplicity) and for using it (showing people how to use new object, which is pretty relevant when using swing with listeners etc)

We will have listener examples though

magnus-madsen:
I have not heard a good argument yet for the extra complexity, so let us keep it simple.

PR #9316, opened 2024-11-24
feat: implement different HTTP methods

sockmaster27:
The only left out method for now is [`CONNECT`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/CONNECT).

Related to  #9035

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
+++ b/main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
@@ -8,6 +8,8 @@
     "java.lang.System::console": "IO",
     "java.lang.System::currentTimeMillis": "IO",
     "java.lang.System::getenv": "Env, IO",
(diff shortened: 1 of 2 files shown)


magnus-madsen:
Check the design of Logger:

https://github.com/flix/flix/blob/master/main/src/library/Logger.flix#L25

I think it makes sense for the eff to have one general operation and the offer helpers in the companion module.

magnus-madsen:
RFR/RFM?

magnus-madsen (approved):

magnus-madsen:
Ill merge and then you can iterate. See also #9391 which is probably blocking for @jaschdoc

jaschdoc:
> Ill merge and then you can iterate. See also #9391 which is probably blocking for @jaschdoc

No worries it was just an edge case I wanted to try by manually disconnecting. Works fine when I'm online

magnus-madsen:
OK, but its still a pretty big issue, so lets get it resolved ASAP.

PR #9318, opened 2024-11-24
refactor: reorganize examples around features

magnus-madsen:


magnus-madsen:
@mlutze This is just a first step.

PR #9319, opened 2024-11-24
feat: add Java `interoperability` examples

magnus-madsen:


magnus-madsen:
The new style is just to show the code. For detailed explanations one should read the docs. 

We will only add comments when there is something special going on.

Moreover, the new examples will be more focused on showing a specific features. 

This way the examples serve as a nice supplement to the docs.

(We will still have a handful of larger examples to show case the language).

magnus-madsen:
Oh- and the examples should be runnable.

mlutze (approved):

Issue #8937, opened 2024-10-20
Move Consumer and Acceptor to their own source files


magnus-madsen:

magnus-madsen:
@LoZander CC

LoZander:
I'm looking at doing this and I'm about to move the `InsideAcceptor` into its own file, and it's making me think about the `inside` function. Really, it should be in some other module somewhere. It's not just used by the acceptor at this point either. It's used and useful beyond this. I just don't know where the function should be. I almost wanna put it in a 

It doesn't make sense to put it in `Position.scala`, since it's not really asking about positions, but pairs of positions and URIs. If we change the function a bit to operate on LSP Locations instead of SourceLocations, then we could put it in Location.scala as `Location.contains(uri: String, pos: Position)`. But then that would mean whenever `InsideAcceptor.accept` is invoked, it has to convert the SourceLocation to an LSP Location to then use `Location.contains`, which is kinda awkward.

I could also just make a Inside.scala module, I suppose. It just feels a bit awkward as well.

For now I'll just leave it inside the `Visitor` module. But let me know what yall think is best.

magnus-madsen:
I think its good enough for now.

PR #9320, opened 2024-11-24
refactor: extracting consumer and acceptor

LoZander:
This PR moves `Consumer` and `Acceptor`, as well as various consumers and acceptors, to their own files.

This is related to #8937.


magnus-madsen:
- Add copyright headers to all files
- Use wildcard imports where appropriate

LoZander:
> * Add copyright headers to all files
> * Use wildcard imports where appropriate

oh, right. Ofc. Should be fixed now :)

PR #9321, opened 2024-11-25
feat: support mkUseStruct as a code action

### LINKED ISSUE #9270: Extend `CodeActionProvider` to support `UndefinedStruct`
```
mod Foo {
   pub struct Bar[rc] {}
}

def f(): Unit = {
    region rc {
        new Bar @ rc { };
        ()
    }
}
```

We want a quickfix that proposes a use.

stormckey:
fixes #9270
Preview:
image
image

Shall we keep the trailing ';'?


stormckey:
> > Shall we keep the trailing ';'?
> 
> What do you mean?

The trailing ; in `use Foo.Bar;`

magnus-madsen (approved):

PR #9325, opened 2024-11-25
feat: mkUseX only propose when the name is visible

### LINKED ISSUE #9249: All `mkUseX` in `CodeActionProvider` should only propose visible names
We should not propose private names so: A name should

- be public or
- be in the root module or
- be in the same namespace

stormckey:
fixes #9249

now we only filter through the public modifier.

image

image

Now the code is verbose and duplicated, since there is no base class for all the symbols where we can access the name and no base class for all the cases where we can access mod. So we have to write code for each type.

Should we use some mechanism to improve that?

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
@@ -103,7 +103,7 @@ object CodeActionProvider {
     */
   private def mkUseDef(ident: Name.Ident, uri: String, ap: AnchorPosition)(implicit root: Root): List[CodeAction] = {
     val syms = root.defs.collect {
-      case (sym, defi) if specIsPublic(defi.spec) => sym
+      case (sym, defi) if isPublic(defi.spec) => sym
     }
     mkUseSym(ident, syms.map(_.name), syms, uri, ap)
   }
@@ -113,7 +113,7 @@ object CodeActionProvider {
     */
   private def mkUseTrait(ident: Name.Ident, uri: String, ap: AnchorPosition)(implicit root: Root): List[CodeAction] = {
     val syms = root.traits.collect {
-      case (sym, trt) if traitIsPublic(trt) => sym
+      case (sym, trt) if isPublic(trt) => sym
     }
     mkUseSym(ident, syms.map(_.name), syms, uri, ap)
   }
@@ -123,7 +123,7 @@ object CodeActionProvider {
     */
   private def mkUseEffect(ident: Name.Ident, uri: String, ap: AnchorPosition)(implicit root: Root): List[CodeAction] = {
     val syms = root.effects.collect {
-      case (sym, eff) if effectIsPublic(eff) => sym
+      case (sym, eff) if isPublic(eff) => sym
     }
     mkUseSym(ident, syms.map(_.name), syms, uri, ap)
   }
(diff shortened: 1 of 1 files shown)


main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
```diff
@@ -101,8 +102,8 @@ object CodeActionProvider {
     * }}}
     */
   private def mkUseDef(ident: Name.Ident, uri: String, ap: AnchorPosition)(implicit root: Root): List[CodeAction] = {
-    val syms = root.defs.map {
-      case (sym, _) => sym
+    val syms = root.defs.collect {
+      case (sym, defi) if defi.spec.mod.mod.contains(Modifier.Public) => sym
```
magnus-madsen, line 106: Can we please add one or more helper functions :) Thanks!

main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
```diff
@@ -187,14 +188,15 @@ object CodeActionProvider {
...
-      root.traits.map { case (sym, _) => sym.name } ++
-      root.typeAliases.map { case (sym, _) => sym.name }
-
-    val syms = (root.enums ++ root.restrictableEnums ++ root.traits ++ root.typeAliases).map {
-      case (sym, _) => sym
-    }
+    val names = root.enums.collect { case (sym, enm) if enm.mod.mod.contains(Modifier.Public) => sym.name } ++
+      root.restrictableEnums.collect { case (sym, enm) if enm.mod.mod.contains(Modifier.Public) => sym.name } ++
+      root.traits.collect { case (sym, trt) if trt.mod.mod.contains(Modifier.Public) => sym.name } ++
+      root.typeAliases.collect { case (sym, alias) if alias.mod.mod.contains(Modifier.Public) => sym.name }
```
magnus-madsen, line 194: I propose to introduce a local variable for each set.

main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
```diff
@@ -187,14 +188,15 @@ object CodeActionProvider {
...
+
+    val syms = (root.enums.collect { case (sym, enm) if enm.mod.mod.contains(Modifier.Public) => sym } ++
```
magnus-madsen, line 196: And a local variable for each here too.

magnus-madsen (approved):

magnus-madsen:
Conflicted.

main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
```diff
@@ -187,18 +188,54 @@ object CodeActionProvider {
...
+
+  /**
+    * Checks if the given restrictable enum is public.
+    */
+  private def restrictableEnumIsPublic(enm: TypedAst.RestrictableEnum): Boolean = enm.mod.mod.contains(Modifier.Public)
+
+  /**
+    * Checks if the given type alias is public.
+    */
+  private def typeAliasIsPublic(alias: TypedAst.TypeAlias): Boolean = alias.mod.mod.contains(Modifier.Public)
```
magnus-madsen, line 237: Lets rename all of the above to just `isPublic` and rely on overloading to pick the right one.

magnus-madsen (approved):

PR #9326, opened 2024-11-25
feat: introduce LocalScopeCompleter

### LINKED ISSUE #9299: Introduce `LocalScopeCompleter`
- [x] Rename `VarCompleter` to `LocalScopeCompleter`
- [x] Add completions for:
  - [x] case class Declaration(decl: NamedAst.Declaration) extends Resolution
  - [x] case class JavaClass(clazz: Class[?]) extends Resolution
  - [x] case class Var(sym: Symbol.VarSym) extends Resolution - **ALREADY DONE**
  - [x] case class LocalDef(sym: Symbol.VarSym, fparams: List[ResolvedAst.FormalParam]) extends Resolution
  - [x] case class TypeVar(sym: Symbol.UnkindedTypeVarSym) extends Resolution - **SKIP**

stormckey:
fixes #9299

Now we provide completion for items in local scope when meeting `UndefinedName` or `UndefinedType`

image

image

image

Potential update and problem:
- The current filter is coarse. Some `Resolution.Declaration`(like def) should pop up only with `UndefinedName`, while some others(structs, enums) should only pop up with `UndefinedType` or `UndefinedStruct` etc.
- We can still see some duplication like `bar` and `Foo.bar`, where `Foo.bar` is proposed by some existing completer. I think we can disable `Foo.bar` when it's in scope, but preserve it when it's not in scope.(like what we did to `AutoImportCompleter`)


magnus-madsen:
@stormckey Conflicted, but looks good. I will give some more feedback later.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/LocalScopeCompleter.scala
```diff
@@ -19,17 +19,31 @@ import ca.uwaterloo.flix.language.errors.ResolutionError
...
-        case Resolution.Var(_) => true
+        case Resolution.LocalDef(_, _)
+        | Resolution.Var(_)
+        | Resolution.JavaClass(_)
+        | Resolution.Declaration(_) => true
         case _ => false
       } => k
-    }
-    matchingVars.map(Completion.VarCompletion(_))
+    }.map(Completion.LocalScopeCompletion(_))
```
magnus-madsen, line 37: We should not  have a thing called "LocalScopeCompletion" -- because there is no such thing externally.

We should have different completions. For example, we should still have Completion.VarCompletion.

You have to remember each completion comes with its own icon, and we want different icons for different things!

stormckey:
Now we provide different completions for different resolutions.

So how should we process all the different types in `Resolution.Declaration`. 
As you can see, there are cases that we should only offer in `UndefinedName` or `UndefinedType`, and some other things that I am not sure about.
image

stormckey:
> * We can still see some duplication like `bar` and `Foo.bar`, where `Foo.bar` is proposed by some existing completer. I think we can disable `Foo.bar` when it's in scope, but preserve it when it's not in scope.(like what we did to `AutoImportCompleter`)

That extra completion is provided by `DefCompleter.getCompletions(context)` in `ExprCompleter`,  if we still adopt the same structure, we'll need to change some types to pass `env` deep into `DefCompleter`

magnus-madsen:
> I make `LocalDefCompletion` a snippet, it looks like: image image
> 
> Shall we use this? Or we can provide both options?
> 
> Also we could do this for `Resolution.Declaration(Def(...))`, But then we will need to divide `DeclarationCompletion` into pieces.

Yes, I think this is exactly why we want to reuse e.g. LocalDefCompletion

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
```diff
@@ -563,6 +588,28 @@ object Completion {
...
+   */
+  case class JavaClassCompletion(name: String) extends Completion
+
+  /**
+   * Represents a local def completion
+   *
+   * @param sym the symbol of the local function
+   * @param fparams the formal parameters of the local function
+   */
+  case class LocalDefCompletion(sym: Symbol.VarSym, fparams: List[ResolvedAst.FormalParam]) extends Completion
```
magnus-madsen, line 611: Do we need to duplicate these? Why can we not reuse what we have?

magnus-madsen:
I will have to look at this in detail. Perhaps we can zoom (but I have to teach now).

My main message is: it is important not just to add code destroying any structure in the process.

We would like to have reusable components-- it seems, to me, some of the completions offered by LocalScopeCompleter overlap with existing ones. What can we do to reuse those?

stormckey:
> We would like to have reusable components-- it seems, to me, some of the completions offered by LocalScopeCompleter overlap with existing ones. What can we do to reuse those?

Yeah I understand, we would need reusable components to make the code more elegant and easier to maintain. I love the feeling when writing beautiful and reusable codes.

For the overlapping problem, I don't think it's totally a problem about reuse. Let'e take this example:
image
I think the our ultimate goal would be:
- when `bar` is in scope, we only propose to use bar(or bar with snippet)
- when `bar` is not in scope, we propose to use bar and `use Foo.bar;` automatically, or we propose to use `Foo.bar` since that may suits better in some cases.

For the first situation, `LocalScopeCompletion` is the first choice. But for the second situation, We can't find `bar` in `LocalScope`, so there must be some other code(DefCompleter) to offer that completion. Currently, we just need to pass `LocalScope` to `DefCompleter` so that it will keep silent when bar is in scope. So there is not so much thing to reuse.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/LocalScopeCompleter.scala
```diff
@@ -0,0 +1,71 @@
...
+  def getCompletions(err: ResolutionError.UndefinedType): Iterable[Completion] =
+    err.env.env.m.foldLeft((List.empty[Completion])){case (acc, (name, resolutions)) =>
+       acc ++ mkDeclarationCompletion(name, resolutions) ++ mkJavaClassCompletion(name, resolutions)
+    }
+
+  /**
+    * Tries to create a DeclarationCompletion for the given name and resolutions.
+    */
+  private def mkDeclarationCompletion(k: String, v: List[Resolution]): Iterable[Completion] =
+    if (v.exists(_.isInstanceOf[Resolution.Declaration])) Completion.DeclarationCompletion(k) :: Nil else Nil
```
magnus-madsen, line 51: We try not to use instanceOf. Instead use a pattern match or use collect.

magnus-madsen:
You have a dangling .DS_STORE file

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/AutoImportCompleter.scala
```diff
@@ -54,12 +54,11 @@ object AutoImportCompleter {
   private def javaClassCompletionsByClass(prefix: String, ap: AnchorPosition, env: LocalScope)(implicit root: Root): Iterable[AutoImportCompletion] = {
     val availableClasses = root.availableClasses.byClass.m.filter(_._1.exists(_.isUpper))
     availableClasses.keys.filter(_.startsWith(prefix)).flatMap { className =>
-      availableClasses(className).map{path =>
+      availableClasses(className).collect { case path if (!env.m.contains(className)) =>
```
magnus-madsen, line 57: Right

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
magnus-madsen: If we really need to above. Perhaps we should call them `LocalDeclCompletion`, `LocalJavaClassCompletion`, and so on-- as you started to do.

magnus-madsen:
@stormckey Remember to remove the .DS_STORE file

stormckey:
> @stormckey Remember to remove the .DS_STORE file

Done, can I add that to the .gitignore? Sometimes it's hard to notice that

magnus-madsen:
> > @stormckey Remember to remove the .DS_STORE file
> 
> Done, can I add that to the .gitignore? Sometimes it's hard to notice that

Yeah sure, but please make a separate PR for that.

stormckey:
> Yeah sure, but please make a separate PR for that.

OK

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
```diff
@@ -198,6 +197,32 @@ sealed trait Completion {
...
 
+    case Completion.DeclarationCompletion(name) =>
+        CompletionItem(
+          label    = name,
+          sortText = Priority.toSortText(Priority.High, name),
+          textEdit = TextEdit(context.range, name),
+          kind     = CompletionItemKind.Enum
+        )
+
+    case Completion.JavaClassCompletion(name) =>
```
magnus-madsen, line 208: -> LocalJavaClassCompl

stormckey:
@magnus-madsen If there is no problem I think we can first merge this pr, and I can continue to work on relative things.

magnus-madsen (approved):

magnus-madsen:
OK; I will merge and then we can iterate.

PR #9331, opened 2024-11-25
proposal: Delete Unit

JonathanStarup:
Now Unit can just be an enum! :D

### DIFF
--- a/main/src/library/Prelude.flix
+++ b/main/src/library/Prelude.flix
@@ -1,5 +1,9 @@
 import java.lang.System
 
+pub enum Unit {
+    case Unit
+}
+
 ///
 /// Static denotes the region of global lifetime.
 ///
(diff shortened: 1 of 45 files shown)


main/src/library/Prelude.flix
```diff
@@ -1,5 +1,9 @@
 import java.lang.System
 
+pub enum Unit {
```
JonathanStarup, line 3: This is where it happens

mlutze:
Now do it for Booleans

magnus-madsen:
You can just start here: https://en.wikipedia.org/wiki/Church_encoding

JonathanStarup:
Ah nevermind we don't want to objectify Booleans

magnus-madsen:
This was fun, but we are not serious right?

JonathanStarup:
half and half :P

PR #9359, opened 2024-11-28
feat: introduce CompletionItemLabelDetails

### LINKED ISSUE #9349: Add support for `CompletionItemLabelDetails`?
LSP now supports something called `CompletionItemLabelDetails`.

See

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItemLabelDetails

How does it work? What does it look like in the completion box if this extra information is provided?

stormckey:
fixes #9349

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/CompletionItemLabelDetails.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/CompletionItemLabelDetails.scala
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2024 Chenhao Gao
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


main/src/ca/uwaterloo/flix/api/lsp/CompletionItemLabelDetails.scala
```diff
@@ -0,0 +1,11 @@
+package ca.uwaterloo.flix.api.lsp
```
magnus-madsen, line 1: Copyright header

main/src/ca/uwaterloo/flix/api/lsp/CompletionItemLabelDetails.scala
```diff
@@ -0,0 +1,11 @@
...
+
+import org.json4s.JsonDSL.*
+import org.json4s.*
+
+/**
+  * Represents a `CompletionItemLabelDetails` in LSP.
+  */
+case class CompletionItemLabelDetails(detail: Option[String], description: Option[String]) {
```
magnus-madsen, line 9: So CompletionItemLabelDetails is optional and its two fields are also optional? 

Thats a lot of optionality! But if thats how they do it, I suppose we should follow that.

magnus-madsen (approved):

magnus-madsen:
Will you make a ticket with a check list of all the completers where it would make sense to use CompletionItemLabelDetails?

Then you could go through them one-by-one.

PR #9360, opened 2024-11-28
feat: delete `masked_cast`

JonathanStarup:
it is veeery brittle and both requires - and it broken by - inlining


main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
```diff
@@ -1541,8 +1537,7 @@ object Desugar {
     val e = visitExp(exp0)
     val prefix = mkDebugPrefix(e, kind0, loc0)
     val e1 = DesugaredAst.Expr.Cst(Constant.Str(prefix), loc0)
-    val call = mkApplyFqn("Debug.debugWithPrefix", List(e1, e), loc0)
-    DesugaredAst.Expr.UncheckedMaskingCast(call, loc0)
+    mkApplyFqn("Debug.debugWithPrefix", List(e1, e), loc0)
```
magnus-madsen, line 1540: Does this require a cast or no?

magnus-madsen:
@JonathanStarup Merging, but do answer my Q

main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
JonathanStarup: I've changed debugWithPrefix to be pure and have a cast inside .. but actually maybe that's a bad idea

magnus-madsen: Thats seems worse to me.

PR #9361, opened 2024-11-28
feat: inhibit AutoUseCompletion and AutoImportCompletion

### LINKED ISSUE #9357: Inhibit AutoUseCompleter and AutoImportCompleter until some heuristic is true
I propose to start with the following simple inhibition:

AutoUseCompleter and AutoImportCompleter should not offer any suggestions until:

- [ ] At least three characters have been entered (i.e. silent for `S`, silent for `St`, but offers suggestions for `Str`)

Can we put this in a function in `CompletionUtils.scala`?

Again, this is only for AutoUseCompleter and AutoImportCompleter. Rest is as before.

stormckey:
fixes #9357

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
@@ -144,8 +144,8 @@ object CompletionUtils {
   }
 
   /**
-    * Checks if we are ready to offer AutoUseCompletion or AutoImportCompletion.
+    * Checks if we should offer AutoUseCompletion or AutoImportCompletion.
     * Currently, we will only offer them if at least three characters have been typed.
     */
-  def readyToComplete(word: String): Boolean = word.length >= 3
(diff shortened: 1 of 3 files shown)


main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
```diff
@@ -143,4 +143,9 @@ object CompletionUtils {
     }
   }
 
+  /**
+    * Checks if we are ready to offer AutoUseCompletion or AutoImportCompletion.
+    * Currently, we will only offer them if at least three characters have been typed.
+    */
+  def readyToComplete(word: String): Boolean = word.length >= 3
```
magnus-madsen, line 150: I think the word "ready" is not exactly right. How about something like: "shouldComplete" or "includeCompletion"?

stormckey: vote for shouldComplete

magnus-madsen (approved):

Issue #9365, opened 2024-11-28
Improve Completion Item Readability with CompletionItemLabelDetails


stormckey:
Here are the Completions that I think can be improved with a CompletionItemLabelDetails:

- [x] AutoImportCompleter
- [x] AutoUseCompleter
- [x] DefCompleter
- [x] LocalScopeCompleter

Any other ideas?

magnus-madsen:
It looks like a good start at least 👍

magnus-madsen:
- [ ] InvokeMethodCompleter (similar to DefCompleter)
- [ ] SignatureCompleter (similar to DefCompleter)
- [ ] OpCompleter (similar to DefCompleter)

stormckey:
I think this ticket has been done.

The label for most Completion is concise now.

PR #9366, opened 2024-11-28
feat: make AutoImportCompletion more readable

stormckey:
first step for #9365
Preview:
image


magnus-madsen:
Very nice-- I still think we need to say that we are adding an import. Where do you propose to put that text? I guess it could just go in the text on the right, "import java.util..." 

But it looks much better!

stormckey:
Now it looks like:
image

magnus-madsen (approved):

PR #9367, opened 2024-11-28
feat: make AutoUseCompletion more readable

stormckey:
second step for #9365
Preview:
image

Shall we add an extra space between the function name and the parameters? Or anything others?

Another example:
image
sometimes a long parameter list will obscure the qualified name :( Maybe we should set an upper bound for the parameter list?

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
@@ -176,7 +176,7 @@ sealed trait Completion {
       val useTextEdit = Completion.mkTextEdit(ap, s"use $qualifiedName;")
       val labelDetails = CompletionItemLabelDetails(
         Some(CompletionUtils.getLabelForSpec(decl.spec)(flix)),
-        Some(qualifiedName))
+        Some(s" use $qualifiedName"))
       CompletionItem(
         label               = name,
         labelDetails        = Some(labelDetails),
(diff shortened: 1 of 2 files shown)


magnus-madsen:
I think

- We need to prefix the text on the right with `use ...` (like for import).
- We could consider dropping the types. For example: the entry for `map` becomes `map(f, l)`. I am less sure about this.

@JonathanStarup @mlutze The picture shows the new and improved auto-completer (notice the text on the left, the grayed args, and the text on the right). But, as you can see, there is not really enough space. What would you propose to do?

mlutze:
Is it possible to make the types come up only for the selected item?

magnus-madsen:
> Is it possible to make the types come up only for the selected item?

No

magnus-madsen:
> > Is it possible to make the types come up only for the selected item?
> 
> No

Although there is a "detailed textbox". You can barely see it on the right. This text box is not always shown (depends on user settings etc).

magnus-madsen:
We could also consider dropping just the return type.

or dropping the argument types and keeping the return type.

magnus-madsen:
Intellij IDEA seems to just accept the mess.

stormckey:
The choice of Rust:
image

Which is opposite to the recommendation of lsp spec. But I think it somewhat makes sense. The qualified name of an item is displayed in the most prominent place, which is also the thing that users cases most when they try to find their target completion.

I don't think the parameter list matters that much, as we will always display it in for the select item on the extra panel, together with the doc(if applicable):
image

stormckey:
If we adopt the Rust way:
image

The signature still takes a lot of place.

magnus-madsen:
I think this would be good to discuss at the meeting.

For now, can we merge update and where the "use" is on the right?

Then you can create a ticket to discuss the two screenshots above and ping Matt and Jonathan to hear what they think.

magnus-madsen (approved):

PR #9370, opened 2024-11-28
refactor: better enum parsing

JonathanStarup:


JonathanStarup:
@magnus-madsen I was looking at what would be possible with more fine-grained break conditions. You can really make it quite good (none of the nine added tests pass in the current parser), but I don't know if anyone will ever have the time to be this detailed.

JonathanStarup:
Yeah, I agree 👍

Issue #9328, opened 2024-11-25
rework spawn


JonathanStarup:

magnus-madsen:
@JonathanStarup Will you fill in some ideas? :)

magnus-madsen:
New proposal:

- [ ] We change the type rule of `spawn e` as follows: 
  - [ ] We require `eff(e) - PrimEff = \emptyset` (i.e. `e` can only have primitive effects).
  - [ ] The effect of `spawn e` is: `eff(s) \cap PrimEff` (i.e. we propagate effects).
- [ ] We can drop the `Sendable` trait.

The above two changes ensure: 
1. We cannot have races because we cannot have _any_ non-local heap effects in spawn. 
2. We cannot have any unhandled AE. 
3. We actually track primitive effects across spawn, e.g. `Net`.

The above breaks channels. We fix this as follows:

- [ ] We introduce a new effect `Global` (I dislike the name. Alternatives? `Static`? `World`? `Comm`unication? `SendRecv`?)
- [ ] We update all channel signatures as follows:
```scala
def recv(receiver: Receiver[t, r]): t \ NonDet + Global = cast (...)
def send(m: t, sender: Sender[t, r]): Unit \ NonDet + Global = cast (...)
```

magnus-madsen:
We now have something like this; spawn only allows Chan, NonDet, IO-- IIRC.

PR #9371, opened 2024-11-28
feat: Rework Channel, Sender, and Receiver

JonathanStarup:
Related to #9328

Some typing is weird because of spawn, but that is next PR


main/src/library/Channel.flix
```diff
@@ -20,28 +20,28 @@ mod Channel {
     ///
```
JonathanStarup, line 20: @magnus-madsen In this file is the typing change clearest - all the other code is just consequences

magnus-madsen: Thanks 👍

magnus-madsen:
Do you mind if this sits for a few days while I mull over the changes? 

I mean, I think this is the way to go, I just dislike "Global". 

Meanwhile I will also give Matt a chance to review since he stands to benefit :)

JonathanStarup:
Sure - if its just the name you're pondering, it's easy to change

magnus-madsen:
I think this is structurally sound. @mlutze Agree?

I am just left with two stylistic issues:

- What should the `Global` effect be called?
- What should be order we write the effects in?

That said, we should also be sure to drop `Sendable` since the new system is sound without it.

magnus-madsen:
Brainstorming names:

- Channel (obvious) -- note that the types are called Sender and Receiver (so there is no clash)
- Comm (Communication)
- Relay
- Talk
- Message

JonathanStarup:
> That said, we should also be sure to drop Sendable since the new system is sound without it.

Lets do in a separate PR

magnus-madsen:
I propose:

- We call the effect `Chan` -- this does not clash with any existing name.
- We always write the effect set as a concrete set in this order: `{Chan, NonDet, IO}` (most specific to least specific)

magnus-madsen:
I would merge this.

JonathanStarup:
@magnus-madsen ready

magnus-madsen:
@mlutze Can you take a quick look? Then I will follow.

magnus-madsen:
@JonathanStarup Before I can merge this, I also need a PR that updates doc.flix.dev

JonathanStarup:
done

magnus-madsen (approved):

magnus-madsen:
Next: 

- Fix sending-and-receiving-on-channels.flix
- Sendable is dead
- Type rule of spawn

PR #9376, opened 2024-11-28
feat: rework spawn

JonathanStarup:
This is a follow up to #9371

In this PR channels lose their region - which means that the exception propagation doesn't work anymore. I suppose its more sensible to keep the region on spawn. something to consider @magnus-madsen


magnus-madsen:
I would have thought that channels are separate from threads? i.e. we can still keep a region for spawn and its fine?

PR #9377, opened 2024-11-29
fix: stop AutoUseCompleter from using functions in root module

### LINKED ISSUE #9375: Autocomplete of `println` broke
We have a function called `println` that is defined in the root module, i.e. it does not have to be use'd to be, well, used.

But currently we get this behavior:

image

image

stormckey:
fixes #9375.
Preview:
image


magnus-madsen:
Conflicted, sorry >\

stormckey:
The main idea is calling `filterDefsByScope` to get all the completions that AutoUseCompleter or LocalScopeCompleter should offer. And `filterDefsByScope` is responsible for considering root modules and local scope at the same time.

magnus-madsen:
OK. Can you rebase and then I will take a look?

PR #9380, opened 2024-11-29
fix: Enum singleton parsing

JonathanStarup:


main/test/flix/Test.Dec.Enum.flix
```diff
@@ -5,6 +5,27 @@ mod Test.Dec.Enum {
 
     pub enum Singleton(Int32)
 
+    pub enum Singleton2(Int32, String)
```
mlutze, line 8: Thanks for tests :)

JonathanStarup: rare occasion

magnus-madsen (approved):

PR #9385, opened 2024-11-29
feat: improve SigCompletion with labelDetails

stormckey:
Not sure about how to trigger that.
But it should work fine.


magnus-madsen:
A signature comes from a trait.

So try `Eq.eq`  or `Foldable.f`

stormckey:
Here it is:
image

PR #9386, opened 2024-11-29
feat: improve OpCompl EnumTagCompl and HoleCompl with labelDetails

stormckey:
Similar update for three completions.

Is it okay to put them in the same pr, or separate them?


stormckey:
For EnumTag:
image
image

magnus-madsen (approved):

magnus-madsen:
> Is it okay to put them in the same pr, or separate them?

Its OK since the changes are so small and are related. But in general we do try to keep it: 1 feature = 1 PR

PR #9387, opened 2024-11-29
feat: `0phases.txt`

JonathanStarup:
if you use `--Xprint-phases` then it will also create `0phases.txt` in that folder, which lists the phases in order.

Additionally I removed some duplicated code that was already checked in FileOps


magnus-madsen:
> if you use --Xprint-phases then it will also create 0phases.txt in that folder, which lists the phases in order.

What does it look like-- and why?

magnus-madsen (approved):

Issue #9348, opened 2024-11-27
Rework `RenameProvider` to use LSP visitor instead of Index


LoZander:

LoZander:
While working on this, I've run into an issue relating to `use`s. When renaming, for instance, enum cases, if these are imported, we have to rename them in the `use` import.

For instance, for
```
enum Color {
  case Red,
  case Blue,
  case Green
}

def f(c: Color): Color = {
  use Color.{Red, Blue, Green}
  
  match c {
    case Red => ???
    ...
  }
}
```

If we rename the `Red` case, it should also change the name in the `use`, but currently the LSP visitor doesn't visit uses at all, so it doesn't allow us do this. I think this should be fixed, but I don't know how important this is right off the bat. Should I immediately expand the visitor or should this issue be left alone for now, to be handled in a follow up for instance? Would it be too big an issue for `use`s to not be handled by the rename provider in the initial version of the rename provider?

magnus-madsen:
I believe this is currently broken right? So you would just leave it as broken?

LoZander:
Yes. It is currently broken. So for now, I'll leave it broken :)

LoZander:
There're also some deeper question about enums and cases relating to how their occurrences are represented in the AST.

Currently, something like `Color.Red` counts as an occurrence of the case `Red` with no information regarding the occurrence of the enum symbol `Color`. There is not specific location for the occurrence of the EnumSymbol `Color` either. This has implications for highlighting and hover as well, but it's more noticable here for the rename provider.

1) if you're trying to rename the case `Red` then in all occurrences of `Color.Red` we need to manually extract the source location for just the `Red` part, since currently the sourcelocation of the occurrence of red includes `Color.`. This can be worked around, but it's a little awkward, especially since an occurrence of `Red` might *not* include `Color.` if the case is imported in the current scope.

2) when trying to rename with the cursor at `C|olor.Red`, wouldn't you assume that you're trying to rename the enum itself? Currently, there is no easy way to support that, since you'd have to recognise that you're actually in the occurrence of the enum symbol not the case itself and with the current model we don't distinguish those. If you're anywhere in `Color.Red`, you're in the occurrence of the case symbol `Red`. You could with some extra logic figure this out, but it's a bit awkward, much like the for previous point.

[...]

PR #9388, opened 2024-11-30
feat: rename provider with lsp visitor

LoZander:
This PR changes the `RenameProvider` to use the LSP visitor instead of the index.

The provider consists of two different searches using the LSP provider
1. Search for the most precise LSP node under the cursor.
2. Search for all occurrencess of a symbol.

The result of the first search, if it is a symbol of occurrence of one, is the one that the second search finds all occurrences of.

In practice, since we assume a thin cursor, we repeat the first search for the symbol under the cursor for both the location on the immediate left and right of the cursor. This means that a total of three searches take place. In theory, it would only be necessary to start searching on the opposite side of the cursor if the initial searches on the first side finds nothing. This means that we don't always have to do three searches but instead sometimes only two. In fact, only when the cursor is at the very end of a symbol is it necessary to do three searches. But in practice we simply always do search 1) twice, based on the assumption that the cost of this redudent search is negligble. Further testing to see if this assumption holds might be warranted however.

Related to #9348.


magnus-madsen:
Why is it not 2 searches to find what the cursor is and then 1 search to perform the renaming?

main/src/ca/uwaterloo/flix/api/lsp/provider/RenameProvider.scala
```diff
@@ -27,145 +30,517 @@ object RenameProvider {
...
-    val defs = index.defsOf(label)
-    val uses = index.usesOf(label)
-    rename(newName, defs ++ uses)
+  private def getEnumSymOccurs(sym: Symbol.EnumSym)(implicit root: Root): Set[SourceLocation] = {
+    var occurs: Set[SourceLocation] = Set.empty
+    def consider(s: Symbol.EnumSym, loc: SourceLocation): Unit = {
+      if (s == sym) { occurs += loc }
+    }
+    object EnumSymConsumer extends Consumer {
+      override def consumeEnum(enm: TypedAst.Enum): Unit = consider(enm.sym, enm.sym.loc)
```
magnus-madsen, line 431: We already have the enum symbol right? Here TypedAst.Enum is literally the same symbol -- its the declaration site.

magnus-madsen: We just need the uses which we get from consumeCaseSymUse

magnus-madsen: (This comment applies to many of the other functions too)

main/src/ca/uwaterloo/flix/api/lsp/provider/RenameProvider.scala
```diff
@@ -27,145 +30,517 @@ object RenameProvider {
...
+    Visitor.visitRoot(root, CaseSymConsumer, AllAcceptor)
+
+    occurs
   }
 
-  private def renameCase(sym: Symbol.CaseSym, newName: String)(implicit index: Index): JObject = {
-    val defn = sym.loc
-    val uses = index.usesOf(sym)
-    rename(newName, uses + defn)
+  private def seperateEnumAndCaseSymOccur(use: SymUse.CaseSymUse): (Option[SourceLocation], SourceLocation) = {
```
magnus-madsen, line 461: This is not a good solution.

magnus-madsen: We should leave it broken (as it already is), or disable the feature, or implement it correctly.

main/src/ca/uwaterloo/flix/api/lsp/provider/RenameProvider.scala
LoZander: Yes. But then we'd need a different `getOccurs` function for when the cursor is on the definition site vs. use site of a symbol for each symbol type.  I figured it it was simpler to have *one* function that gives the occurrences for each relevant symbol type and just repeat a little bit of search. But maybe it's better to to split into `getXOccurFromDeclSite` and `getXOccurFromUseSite` (or something like that) to avoid some of that search repetition?

magnus-madsen:
@LoZander Give me a ping when this is polished and the conflict resolved. Thanks!

main/src/ca/uwaterloo/flix/api/lsp/provider/RenameProvider.scala
LoZander: What do you think?

magnus-madsen: I don't understand. We use on search to find the symbol and the one search to find all the SymUses. Why does the second search need to include the declaration?

LoZander: Oh, right, because when we've found the symbol, where the cursor is on a use of it or the declaration, we know the `SourceLocation` of the declaration, because it's baked into the symbol. So yeah, you're right. So in the second search we only ever need to find sym uses. I though that if we found the symbol through the cursor being on a SymUse, then we'd have to search for the declaration site, but ofc. we don't. So yeah, it can be simplified. Good cool :)

magnus-madsen: Ping me when you have a refreshed PR

LoZander: Yep. Will do. I'll also be publishing a new PR soon with the FindReferencesProvider changes as well. Probably both will be done by tonight :)

magnus-madsen: I see we are both burning the midnight oil

LoZander: Always XD

magnus-madsen:
This looks good  to me. Resolve the conflict and we can merge.

magnus-madsen (approved):

magnus-madsen:
Can you create PR towards flix/vscode-flix that disables/ignores the now broken tests? Thanks!

LoZander:
> Can you create PR towards flix/vscode-flix that disables/ignores the now broken tests? Thanks!

yes, ofc. I'll look into that soon :)

Issue #9214, opened 2024-11-13
Remove support for `try-with`


magnus-madsen:
Two PRs:

- One that reworks all tests
- One that actually removes support in the Parser/Weeder

Probably easier to start with 2 and then do 1.

PR #9398, opened 2024-12-01
refactor: remove use of `try-with`

jaschdoc:
Depends on https://github.com/flix/flix/pull/9409
Related to https://github.com/flix/flix/issues/9214

### DIFF
--- a/main/test/flix/Test.Handler.NewObject.flix
+++ b/main/test/flix/Test.Handler.NewObject.flix
@@ -12,7 +12,7 @@ mod Test.Handler.NewObject {
 
     def newAskRunnable(r: Ref[Int32, rc]): Runnable \ IO = new Runnable {
         def $run(_this: Runnable): Unit \ { rc } =
-            try {
+            run {
               Ask.ask(42)
             } with Ask {
                 def ask(x, _) = Ref.put(x, r)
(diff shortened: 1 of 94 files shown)


main/test/flix/Test.Handler.NewObject.flix
```diff
@@ -30,9 +30,9 @@ mod Test.Handler.NewObject {
 
     def newGenAskRunnable(r: Ref[Int32, rc]): Runnable \ IO = new Runnable {
         def $run(_this: Runnable): Unit \ { rc } =
-            try {
+            run {
                 Ask.ask(
-                    try {
+                    run {
```
jaschdoc, line 35: Weird, parsing does not allow nested run-with

magnus-madsen: Uh oh. You may have to check my stupid code!

jaschdoc: I will revert these to errors to `try-with` for now and fix it when we also remove the old `try-with` syntax completely (it's on 2 places in the entire code base that needs to be updated)

magnus-madsen: Can we not fix it now? Separately?

magnus-madsen: Its seems like a pretty bad bug, so would be good to have a fix in.

jaschdoc: True I have been working on it for 45 minutes now at it is not straightforward why this happens

magnus-madsen: Gosh.

magnus-madsen (approved):

Issue #9401, opened 2024-12-01
Disallow capitalized function


magnus-madsen:
```
def MyFunction(x: Int32): Int32 = ???
```

Can we add a new weeder error for this?

PR #9404, opened 2024-12-01
feat: add `UnexpectedNonLowerCaseName`

jaschdoc:
Closes https://github.com/flix/flix/issues/9401


JonathanStarup:
I think its a parser problem in general - IMO the parser should be case insensitive and then the weeder sorts it out

magnus-madsen (approved):

PR #9412, opened 2024-12-01
feat: make `Http` fallible and introduce `HttpWithResult`

### LINKED ISSUE #9391: Unhandled `Http` exception
If you use the `Http` effect and handle it using `Http.handle` and you disconnect from the internet you get `java.net.ConnectException` from Java that is not handled by the `Http.handle` function. I think this should be handled gracefully by wrapping the response in a `Result`

sockmaster27:
Fixes #9391


sockmaster27:
@jaschdoc
Can you check if this fixes what you were doing before?
I can't find anywhere in the documentation that a `ConnectException` can be thrown, but I've tried including it anyway.

main/src/library/Http.flix
```diff
@@ -23,13 +23,25 @@ eff Http {
...
     import java.net.http.HttpRequest
     import java.net.http.{HttpRequest$BodyPublishers => BodyPublishers}
     import java.net.URI
     import java.util.{List => JList}
 
+    ///
+    /// An error that can occur when performing HTTP operations.
+    ///
+    pub enum Error {
+        case Generic(String)
```
magnus-madsen, line 42: Instead of this, we should reuse our IoError that we already have.

magnus-madsen:
It would be good if @jaschdoc could offer some more information.

jaschdoc:
It works but the error message is just `null` so maybe we should add some more info manually? Maybe leave in the stack trace?

magnus-madsen:
> It works but the error message is just `null` so maybe we should add some more info manually? Maybe leave in the stack trace?

No stack trace, but he should use ErrorKind and a proper error message.

magnus-madsen:
Test failures

main/src/library/IoError.flix
```diff
@@ -38,6 +38,7 @@ mod IoError {
         case NotDirectory,
         case UnexpectedEof,
         case Unsupported,
+        case ConnectionFailed,
```
magnus-madsen, line 41: Keep sorted?

magnus-madsen:
What is the status on this?

magnus-madsen:
The compiler tests are still failing.

main/src/library/App.flix
sockmaster27, line 1: What's the clean FP way of doing this in Flix?

magnus-madsen: Lets ask @JonathanStarup

main/src/ca/uwaterloo/flix/api/Flix.scala
```diff
@@ -260,6 +260,7 @@ class Flix {
     "Abort.flix" -> LocalResource.get("/src/library/Abort.flix"),
     "Clock.flix" -> LocalResource.get("/src/library/Clock.flix"),
     "Http.flix" -> LocalResource.get("/src/library/Http.flix"),
+    "HttpWithResult.flix" -> LocalResource.get("/src/library/HttpWithResult.flix"),
```
magnus-madsen, line 263: Can we do this in separate PR?

This one is supposed to fix a bug.

sockmaster27: Or, well, there's sort of a circular dependency, since this PR won't compile without this, and introducing `HttpWithResult` is weird when there's no error handling in it.

magnus-madsen: OK. Rename the PR then

magnus-madsen (approved):

PR #9422, opened 2024-12-02
feat: pre-ground strict substitutions

magnus-madsen:


magnus-madsen:
```
PR:
  Throughput (best): 68.632 lines/sec (with 24 threads.)
    min:  9.159, max: 68.632, avg: 58.330, median: 61.168
  Finished 100 iterations on 67.274 lines of code in 122 seconds.

Master:
  Throughput (best): 60.484 lines/sec (with 24 threads.)
    min:  8.693, max: 60.484, avg: 54.461, median: 56.388
  Finished 100 iterations on 67.356 lines of code in 130 seconds.
```

magnus-madsen:
@JonathanStarup I was tired when I wrote this. Please think about it :)

main/src/ca/uwaterloo/flix/language/phase/Monomorpher.scala
```diff
@@ -114,18 +131,9 @@ object Monomorpher {
       * Performance Note: We are on a hot path. We take extra care to avoid redundant type objects.
```
JonathanStarup, line 131: can't comment the correct location, but StrictSubstitution doc comment in "Notes on `s`:" should be updated

main/src/ca/uwaterloo/flix/language/phase/Monomorpher.scala
```diff
@@ -590,9 +600,9 @@ object Monomorpher {
               val env1 = env0 + (sym -> freshSym)
               val subst1 = caseSubst @@ subst.nonStrict
```
JonathanStarup, line 601: It's hard to know if this is now doing the same thing, but I think its fine

JonathanStarup (approved):

main/src/ca/uwaterloo/flix/language/phase/Monomorpher.scala
magnus-madsen: @mlutze Why do we need this? What's the consequence if it is made strict?

JonathanStarup: If you typematch with the pattern `List[_]` and make that substitution strict, then you effectively match for `List[AnyType]` which ruins the pattern

magnus-madsen: What's the example? (i.e. something that be a test case that breaks)

magnus-madsen: 1. I remember some discussion about List[_]. What was it?
2. Is my solution broken or not? I remain the same dom(s) for the two substitutions, so if x \notin dom(s) then x \notin dom(s') where s' is a subst. with my new implementation.

PR #9425, opened 2024-12-02
refactor: lazy val `isGround`

JonathanStarup:
~2.3 % frontend throughput increase

(doing just `val` was 0.9 % faster)


magnus-madsen:
Is this measured on the server?

JonathanStarup:
my (non laptop) pc. I am always hesitant to use server since we have no protocol for concurrent use

magnus-madsen:
You can use `w` and `top` see if if anyone is logged in.

JonathanStarup:
> You can use `w` and `top` see if if anyone is logged in.

Ah nice

---

Xperf --frontend --n 100
- master: 118.940 median
- PR: median X (+Y%)

magnus-madsen (approved):

PR #9426, opened 2024-12-02
refactor: optional fixpoint

JonathanStarup:
~1.5 % frontend throughput speedup

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
@@ -87,13 +87,13 @@ object SetUnification {
     val trivialPhaseName = "Trivial Equations"
 
     runWithState(state, runRule(constantAssignment), "Constant Assignment")
-    runWithState(state, runRule(trivial), trivialPhaseName)
-    runWithState(state, runRule(variableAlias), "Variable Aliases")
-    runWithState(state, runRule(trivial), trivialPhaseName)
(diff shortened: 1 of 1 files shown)


magnus-madsen:
Why would this give a speedup?

JonathanStarup:
I would've also guessed it would do nothing since the lists are size 1 basically always - but it does avoid re-applying the substitution

magnus-madsen:
I don't think we should have this optimization. And if we really should then it suggests the architecture is broken and should be reworked.

JonathanStarup:
It's fine to close, yeah - but certainly, knowing things upfront about fixpoint behavior will allow that rule to run faster. i don't think that means that anything is broken

magnus-madsen:
> It's fine to close, yeah - but certainly, knowing things upfront about fixpoint behavior will allow that rule to run faster. i don't think that means that anything is broken

Perhaps. But this suggests the old approach -- with more distinct phases -- was better then?

JonathanStarup:
I mean you can have both like shown here. you can have both control and avoid duplication

magnus-madsen:
> I mean you can have both like shown here. you can have both control and avoid duplication

Yes... But my point was that the other implementation was even more fine-grained about how it applied the substitution. You could support that by adding even more Boolean flags, but at some point it becomes unwieldy.

JonathanStarup:
at some point everything becomes unwieldy. I think its possible not to have duplicated fixpoints with individual bugs and unclear differences

PR #9431, opened 2024-12-03
refactor: `Unsafe` -> `UnsafeOld`

JonathanStarup:


magnus-madsen:
The name seems confusing since it is specifically _not_ a block. Would it not be better with `UnsafeExpr` or `UnsafeOld`?

magnus-madsen (approved):

PR #9432, opened 2024-12-03
feat: add type arguments to intrinsic

JonathanStarup:
Allows syntax `$Whatever$[types..](args..)` (and also `$Whatever$(args..)`

This allows for easier experimentation with new language constructs without requiring new parsing from the start.


magnus-madsen:
Sorry. I am really against this. 

This is taking a feature that has high cohesion and then polluting it with new stuff. Moreover, the new stuff is only to support some vague future use cases that will be refactored anyway.

With the old parser and frontend, whenever we wanted to introduce something new, we always just keep up with some new keyword `experimental_babblabla`. This gave maximum freedom and kept everything isolated. I don't see why that approach is still not superior.

magnus-madsen:
(This is also a classic example of putting convenience over just doing it right.)

JonathanStarup:
I am thinking of the java interop project and the struct project and to a smaller degree this unsafe_remove thing. They all have to add some random parsing as the first step, and if its a student they might not be very comfortable in that (its also not easy). I think its valuable to have a WIP construct even though its only for WIP things.

Issue #9347, opened 2024-11-27
Add `AutoUseCompleter`


magnus-madsen:
We have `AutoImportCompleter`. We would like the same for uses, i.e. an `AutoUseCompleter`

For example, if I write:

```scala
mod Foo {
    pub def foo(): Unit = ()
}
mod Qux {
    pub def main(): Unit \ IO = 
        fo|
}
```

Then I would like a suggestion `foo (use Foo.foo)`.

We should have such "auto uses" for each Flix construct that can be `use`d. They are:

- [ ] defs
- [x] effects
- [ ] enums
- [ ] structs
- [ ] traits
- [ ] type aliases
- [ ] tags (the components of an enum)
- [ ] operations (a bit like a def but lives *inside* an effect)
- [ ] signatures (a bit like a def but lives *inside* a trait)

To understand what I mean by lives *inside* X, here is an example:

```scala
mod Foo {
    pub enum Color { case Red }
}

def f(): Foo.Color = Re|
```

Here I want the suggestion: `Red (use Color.Red)`.

PR #9434, opened 2024-12-03
feat: support AutoUse for effects

stormckey:
Another step for #9347
Preview:
image
image
image

A problem is that the former completer will still suggest to complete `A.FooBar` even when `FooBar` is in scope. We'll need to change `EffSymCompleter`, but that would be another PR.

By the way, should we keep `AutoUseEffCompletion` and `EffectCompletion` two different things? We can otherwise use `Option[AnchorPosition]` to indicate if we should auto use and merge them.


magnus-madsen:
> A problem is that the former completer will still suggest to complete A.FooBar even when FooBar is in scope. We'll need to change EffSymCompleter, but that would be another PR.

OK

> By the way, should we keep AutoUseEffCompletion and EffectCompletion two different things? We can otherwise use Option[AnchorPosition] to indicate if we should auto use and merge them.

How about we do these things as best as we can separately, and then when you arrive, we can sketch out a new design.

magnus-madsen (approved):

Issue #9449, opened 2024-12-04
Rework `FindReferencesProvider` to use LSP visitor instead of Index


LoZander:

LoZander:
I've come to the conclusion that "find references" does not make much sense for record labels. To justify this, I'll repeat the argument I made in flix/vscode-flix#465.

> The test regarding record labels has been removed. This is because we will not, at least at this time, support "find references" for record labels. This is because, since all labels of the same name are indistinguishable, the best we could hope for is returning a list of all labels with the given name either in the entire project or some limited part of it. Since it's very likely that a label name might be used in multiple unrelated contexts, it's not useful to return a list of all of these and worse yet, it might mislead a user to believe that there's a meaningful connection between all of these labels. At best, we could limit this by restricting the scope that we search for references in, but this doesn't really address the underlying issue. I argue, that it's better to just not support "find references" for record labels. Any reference to label that is likely to be relevant would be in the surrounding code anyways, such as within the same function or similar.

LoZander:
Ofc I'm more than open for arguments against this, but this is the thought process behind my choice

PR #9450, opened 2024-12-04
feat: `FindReferencesProvider` with LSP visitor instead of Index

LoZander:
This PR changes the `FindReferenceProvider` to use the LSP visitor instead of the Index.

The provider uses two different searches using the LSP provider

1. Search for the most precise LSP node under the cursor. If this is a symbol that Flix supports "find references" for, then this will be the one we'll find the references to.
2. Search for all occurrencess of the aforementioned symbol.

In practice, since we assume a thin cursor, we repeat the first search for the symbol under the cursor for both the location on the immediate left and right of the cursor. This means that a total of three searches take place. In theory, it would only be necessary to start searching on the opposite side of the cursor if the initial searches on the first side finds nothing. This means that we don't always have to do three searches but instead sometimes only two. In fact, only when the cursor is at the very end of a symbol is it necessary to do three searches. But in practice we simply always do search 1) twice, based on the assumption that the cost of this redudent search is negligble. Further testing to see if this assumption holds might be warranted however.

Related to #9449.


LoZander:
I still need to filter out references to the std library. But then I think this one is basically done

magnus-madsen:
Here are also some tests to look into.

magnus-madsen:
I think this just needs some polish and to figure out what is going on with the tests and then we are good to go?

LoZander:
> I think this just needs some polish and to figure out what is going on with the tests and then we are good to go?

Yes. I believe that's what's left. I'll look into that now :)

magnus-madsen (approved):

magnus-madsen:
Merged 🚀

PR #9456, opened 2024-12-04
refactor: remove propagation

JonathanStarup:
-0.37 % throughput on server (Xperf --frontend --n 100)


magnus-madsen (approved):

magnus-madsen:
Interesting. What do you think of the fact that propagation has almost no effect?

JonathanStarup:
It think its makes sense now that we have canonical form, since propagation doesn't do anything on normal form formulas

PR #9466, opened 2024-12-04
refactor: simplify `SetFormula.Inter`

magnus-madsen:


main/src/ca/uwaterloo/flix/language/phase/unification/set/SetFormula.scala
```diff
@@ -406,116 +385,11 @@ object SetFormula {
...
-        }
-    }
-
-    // Avoid intersections with zero or one subformula.
-    subformulasOf(elemPos, cstsPos, varsPos, elemNeg, cstsNeg, varsNeg, other).take(2).toList match {
-      case Nil => return Univ
-      case one :: Nil => return one
-      case _ => ()
+    fs match {
+      case Nil => Empty
```
JonathanStarup, line 389: I guess this case never happens but I think it should be Univ since its the neutral set for intersection

magnus-madsen: I am still programming

magnus-madsen:
@mlutze @JonathanStarup This is not the final refactor -- more can be done. But this a good step.

magnus-madsen:
You can review and approve, but don't merge yet. I have to look into some performance stuff.

mlutze (approved):

PR #9468, opened 2024-12-05
fix: tuple parsing crash ( #9441)

### LINKED ISSUE #9441: Parser crash
```
mod Day1 {
    use Int32.fromString;
    use String.splitOn;
    use List.sum
    use List.map
    use List.zip
    use List.sort
    use Util.Option.get

    pub def solv(l1: List[Int32], l2: List[Int32]): Int32 = {
        (sort(l1), sort(l2)) ||> zip |> map(match (x, y) -> x + y) |> sum
    }

    pub def parse(l: List[String]): (List[Int32], List[Int32]) \ Abort = {
        l |> map(s -> {
            match splitOn(substr = " ", s) {
                case one :: two :: Nil => (fromString(one) |> get, fromString(two) |)
                case other => Abort.abort("Expected two ints, not %!!{other}.")
            }
        })
    }

}
```

JonathanStarup:
Fixes  #9441

This costs resilience in the current version. The situation is that for `(x, |)` (where `|` can be any specific non-expr token), the expressions parser cannot parse `|` but also does not advance past it when it produces an error - infinite loop.

Advancing past it (saying `|` is wrong and basically ignoring it) is bad for resilience however, since the wrong token could be `)`, e.g. `(x, )` but here we want to let the tuple rule eat the `)` to restore a sensible AST.


JonathanStarup:
The bahaviour in this PR also follows the blogpost, which advances the parser for errors in expr_delimited

magnus-madsen:
Test case?

JonathanStarup:
yeah i think it just breaks resilience for tuple like things with a trailing `,`, like `f(x, )` and `(x, )`

JonathanStarup:
To fix this proper we need more finegrained break/first sets

magnus-madsen:
Does autocomplete for `(x, |` and `(x, |)` still work?

JonathanStarup:
What should autocomplete do in that case?

magnus-madsen:
I suppose we can merge this.

magnus-madsen:
But can you just check that autocomplete works fine?

JonathanStarup:
There is no autocomplete inside or outside the tuple on both master and PR, but `(x, String.)` fx still gives you string autocomplete. It looks good to me

magnus-madsen (approved):

PR #9478, opened 2024-12-06
feat: support AutoUse for enums

stormckey:
Another step for #9347
Preview:
image

image

Note:
- The priority for auto use is `Lower`, the same as normal java package import.
- We use the same completion for Effects and Enums.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
@@ -44,7 +44,9 @@ sealed trait Completion {
         kind             = CompletionItemKind.Enum
       )
 
-    case Completion.AutoUseCompletion(name, qualifiedName, doc, ap) =>
+    case Completion.AutoUseEffCompletion(sym, doc, ap) =>
+      val name = sym.name
+      val qualifiedName = sym.toString
       val additionalTextEdits = List(Completion.mkTextEdit(ap, s"use $qualifiedName"))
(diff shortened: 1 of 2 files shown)


magnus-madsen:
@stormckey Conflicted.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
```diff
@@ -639,11 +637,12 @@ object Completion {
...
    * Represents an auto-import completion.
    *
-   * @param eff           the effect to complete and use.
+   * @param name          the name to complete.
+   * @param qualifiedName the qualified name to use.
    * @param doc           the documentation associated with the effect.
    * @param ap            the anchor position for the use statement.
    */
-  case class AutoUseEffCompletion(eff: Symbol.EffectSym, doc: String, ap: AnchorPosition) extends Completion
+  case class AutoUseCompletion(name: String, qualifiedName: String,  doc: String, ap: AnchorPosition) extends Completion
```
magnus-madsen, line 645: I don't think this is a good idea. Its better to keep them separate-- that way we retain more precise information and we can use different icons etc.

stormckey:
A small problem, we don't have a completionItemKind for Effect. Is there a corresponding icon is VSCode?

magnus-madsen:
> A small problem, we don't have a completionItemKind for Effect. Is there a corresponding icon is VSCode?

No, but we can use something else as stand in.

magnus-madsen (approved):

PR #9480, opened 2024-12-06
fix: disable EffSymCompleter when the user is not typing a qualified name

stormckey:
As suggested in #9434
Preview:
image

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
@@ -84,8 +84,7 @@ object CompletionProvider {
       //
       // Types.
       //
-      case SyntacticContext.Type.Eff => EffSymCompleter.getCompletions()
-      case SyntacticContext.Type.OtherType => TypeCompleter.getCompletions(ctx) ++ EffSymCompleter.getCompletions()
+      case SyntacticContext.Type.OtherType => TypeCompleter.getCompletions(ctx)
 
       //
       // Patterns.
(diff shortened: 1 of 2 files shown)


magnus-madsen:
I am not sure I understand. Should it not check whether that specific effect is in scope?

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -84,8 +84,7 @@ object CompletionProvider {
       //
       // Types.
       //
-      case SyntacticContext.Type.Eff => EffSymCompleter.getCompletions()
-      case SyntacticContext.Type.OtherType => TypeCompleter.getCompletions(ctx) ++ EffSymCompleter.getCompletions()
```
magnus-madsen, line 88: Also, are we sure we are covering all cases here?

stormckey: Yes, the only way a `SyntacticContext.Type.*` is built is from `UndefinedType`.
https://github.com/flix/flix/blob/983766f6689ad454e6d4f1d5757d154d15f62dc9/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala#L216C1-L216C93
So we just move the EffSymCompleter to SemanticCompleter. As a result, EffSymCompleter can access the `err.qn` and tell if the undefined type is qualified.

stormckey:
> I am not sure I understand. Should it not check whether that specific effect is in scope?

You're right. Actually we will disable this completer when the user is typing a unqualified name. At that time, it's AutoUseCompleter and LocalScopeCompleter to do their work.

stormckey:
image
image
image

magnus-madsen (approved):

PR #9481, opened 2024-12-06
fix: add missing priority for some completions

stormckey:
I've found that the sortText of some completion is bare String.
I've changed that of EffectCompletion to `Lower`, same as DefCompletion.

Some completion remains:
- most use completion image
- hole completion, where priority is a string image

Shall we change them to use Priority?


magnus-madsen (approved):

magnus-madsen:
I suppose. In general, I don't know of a good way to handle priority. Do you see some clever design we can use?

Issue #9482, opened 2024-12-06
Ignore synthethic locations in visitor


JonathanStarup:
Follow up to #8935 and solves #8935

For resolver generated partial applications to work with hover, we need to 1) recurse into synthetic locations but 2) not consume them. 2) is the missing part

magnus-madsen:
I think it is more complicated than that.

We need an invariant that says when to visit/not to visit a synthetic location and when to recurse. 

Do you have a suggestion?

JonathanStarup:
I think that it sounds like our use or synthetic/real is wrong then. Why should lsp things care about compiler generated stuff?

magnus-madsen:
> I think that it sounds like our use of synthetic/real is wrong then. Why should lsp things care about compiler generated stuff?

The problem is all our desugarings etc.

JonathanStarup:
yeah that makes sense, but  the AST should know whether its synthetic or not, otherwise the visitor has the impossible task of undoing the desugaring that happened

magnus-madsen:
> yeah that makes sense, but the AST should know whether its synthetic or not, otherwise the visitor has the impossible task of undoing the desugaring that happened

Agreed. But I don't know how to solve it. Consider par, forM, etc. What should the AST look like and what information should it carry?

JonathanStarup:
yeah for those its not easy. i guess the best solution in this perspective is to cut down on desugaring. i remember you also saying that a while ago, do you still feel like thats the direction it should go?

magnus-madsen:
I think we need come up up a plan like this:

Write a verifier that enforces:

- Source locations properly nested within each other
- ??? invariant about synthetic locations.

Even if we "drop desugar" there are still many pieces of the AST that are synthetic. Just small things here and there.

JonathanStarup:
> Source locations properly nested within each other

Sounds like some nice coding on a lazy Sunday - is it okay if i do it at some point? 

> ??? invariant about synthetic locations.

I can't imagine that we can do much better than `don't show these but still explore them` with the current amount of desugaring

PR #9483, opened 2024-12-06
fix: hover on partial application

### LINKED ISSUE #9461: resolver ruins hover information
anything in `List.map(x -> ???)` just says `List[t]`
```
def example(l: List[t]): t = l |> List.map(x -> ???) |> ???
```

JonathanStarup:
Fixes  #9461 and #9482

- partial application produced too many real locations and
- visitor will consume synthetic locations

so this was fixed


JonathanStarup:
This PR allows you too, on `List.map(f)` hover on `List.map` and see the docs, `f` and see its type, but for some reason hovering on the full expression (one of the parens) gives you its argument missing argument type i think. So better but not fully fixed.

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
```diff
@@ -295,7 +295,7 @@ object Visitor {
   private def visitExpr(expr: Expr)(implicit a: Acceptor, c: Consumer): Unit = {
     if (!a.accept(expr.loc)) { return }
 
-    c.consumeExpr(expr)
+    if (!expr.loc.isSynthetic) c.consumeExpr(expr)
```
magnus-madsen, line 298: This is not correct, generally speaking. It might work here, but it probably breaks other things that are desugared.

We need a more general and robust solution.

Also one that is applied to all nodes, not just exprs.

JonathanStarup: Why is it not correct? isn't lsp about the stuff that the user has in their program?

magnus-madsen: There are synthetic node with real nodes inside.

JonathanStarup: right, but the synthetic one is still explored, just not consumed

Issue #9508, opened 2024-12-13
Rework `CodeHinter` to use LSP visitor


LoZander:

magnus-madsen:
Fixed

PR #9509, opened 2024-12-13
feat: `CodeHinter` with LSP visitor

LoZander:
This PR changes `CodeHinter` to use the LSP visitor instead of `Index`.

Related to #9508.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/extra/CodeHinter.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/extra/CodeHinter.scala
@@ -65,12 +65,30 @@ object CodeHinter {
     hints.filter(include(_, sources))
   }
 
+  /**
+    * Returns a collection of code quality hints related to a given trait's annotations.
+    *
+    * @param root The root AST node of the project.
+    * @param sym  The [[Symbol.TraitSym]] for the trait in question.
+    * @param loc  The [[SourceLocation]] for the occurrence of `sym`.
+    * @return     A collection of code quality hints.
+    */
   private def considerTrait(root: Root, sym: Symbol.TraitSym, loc: SourceLocation): List[CodeHint] = {
     val trt = root.traits(sym)
     val ann = trt.ann
     checkDeprecated(ann, loc) ++ checkExperimental(ann, loc)
   }
 
+  /**
+    * Returns a collection of code quality hints related to the given def's annotations.
+    *
+    * Note that hints related to `@LazyWhenPure` and `@ParallelWhenPure` are not included in this colleciton.
+    *
+    * @param root The root AST node for the Flix project.
+    * @param sym  The [[Symbol.DefnSym]] for the Def in question.
+    * @param loc  The [[SourceLocation]] for the occurrence of the `sym`.
+    * @return     A collection of code quality hints
+    */
   private def considerDef(root: Root, sym: Symbol.DefnSym, loc: SourceLocation): List[CodeHint] = {
     val defn = root.defs(sym)
     val ann = defn.spec.ann
@@ -80,6 +98,14 @@ object CodeHinter {
       checkParallel(ann, loc)
   }
 
+  /**
+    * Returns a collection of code quality hints related to the given enum's annotations.
+    *
+    * @param root The root AST node for the Flix project.
+    * @param sym  The [[Symbol.EnumSym]] for the enum in quesiton.
+    * @param loc  The [[SourceLocation]] for the occurrence of `sym`.
+    * @return     A collection of code hints.
+    */
   private def considerEnum(root: Root, sym: Symbol.EnumSym, loc: SourceLocation): List[CodeHint] = {
     val enm = root.enums(sym)
     val ann = enm.ann
--- a/main/src/ca/uwaterloo/flix/language/phase/extra/CodeHinter.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/extra/CodeHinter.scala
@@ -122,18 +122,18 @@ object CodeHinter {
     var traitUses: List[TraitSymUse] = Nil
 
     object UseConsumer extends Consumer {
-      override def consumeTraitSymUse(symUse: TraitSymUse): Unit = traitUses = symUse :: traitUses
-      override def consumeTraitConstraintHead(tcHead: TraitConstraint.Head): Unit = traitUses = TraitSymUse(tcHead.sym, tcHead.loc) :: traitUses
+      override def consumeCaseSymUse(sym: SymUse.CaseSymUse): Unit = enumUses = (sym.sym.enumSym, sym.loc) :: enumUses
       override def consumeDefSymUse(sym: DefSymUse): Unit = defUses = sym :: defUses
       override def consumeExpr(exp: Expr): Unit = exp match {
         case TypedAst.Expr.ApplyDef(symUse, exps, _, _, _, _) => defCalls = (symUse.sym, exps) :: defCalls
         case _ => ()
       }
+      override def consumeTraitSymUse(symUse: TraitSymUse): Unit = traitUses = symUse :: traitUses
+      override def consumeTraitConstraintHead(tcHead: TraitConstraint.Head): Unit = traitUses = TraitSymUse(tcHead.sym, tcHead.loc) :: traitUses
       override def consumeType(tpe: Type): Unit = tpe match {
         case Type.Cst(TypeConstructor.Enum(sym, _), loc) => enumUses = (sym, loc) :: enumUses
         case _ => ()
       }
-      override def consumeCaseSymUse(sym: SymUse.CaseSymUse): Unit = enumUses = (sym.sym.enumSym, sym.loc) :: enumUses
     }
 
     Visitor.visitRoot(root, UseConsumer, AllAcceptor)
--- a/main/src/ca/uwaterloo/flix/language/phase/extra/CodeHinter.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/extra/CodeHinter.scala
@@ -1,5 +1,6 @@
 /*
  * Copyright 2021 Magnus Madsen
+ * Copyright 2024 Alexander Dybdahl Troelsen
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/main/src/ca/uwaterloo/flix/language/phase/extra/CodeHinter.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/extra/CodeHinter.scala
@@ -29,26 +29,26 @@ object CodeHinter {
   /**
     * Returns a collection of code quality hints for the given AST `root`.
     */
-  def run(root: TypedAst.Root, sources: Set[String]): List[CodeHint] = {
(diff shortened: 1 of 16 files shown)


main/src/ca/uwaterloo/flix/language/phase/extra/CodeHinter.scala
```diff
@@ -16,349 +17,113 @@
...
-
-    case Expr.Discard(exp, _, _) =>
-      visitExp(exp)
-
-    case Expr.Match(matchExp, rules, _, _, _) =>
-      visitExp(matchExp) ++ rules.flatMap {
-        case MatchRule(_, guard, exp) => guard.toList.flatMap(visitExp) ::: visitExp(exp)
+    object HintConsumer extends Consumer {
+      override def consumeTraitSymUse(symUse: SymUse.TraitSymUse): Unit = {
+        hints = hints ++ considerTrait(root, symUse.sym, symUse.loc)
```
magnus-madsen, line 40: Why interleave different concerns?

You could compute uses separate (in a helper function) and just use that

magnus-madsen: Instead now your Consumers are producing hints. Seems like the opposite of separation of concerns.

LoZander: Doing so would mean "throwing away" the type information we have at hand. When creating hints from these uses, we'd have to then pattern match to regain that information to find the def, enum or trait in question to gain it's annotation.

Ideas 
1. different collections for defs, traits, enums and def calls
2. one stack of `(Annotation, SourceLocation)` and treat all the same (check for each if it has an annotation). Separate stack for def call info `(Symbol, Type, SourceLocation)`

LoZander: i agree that more separation would be good

magnus-madsen: You need to collect uses right? Can you not write a visitors that collects those uses in separate maps without worrying that downstream they will be used for hints?

LoZander: What I could do is collect every occurrence of every trait, enum, def use and def call, each in their own type of collection and then afterwards "consider" each. You mean something like that?

magnus-madsen: Yes

LoZander: How about now?

main/src/ca/uwaterloo/flix/language/phase/extra/CodeHinter.scala
```diff
@@ -15,350 +16,170 @@
...
-    case Body.Functional(_, exp, _) => visitExp(exp)
-    case Body.Guard(exp, _) => visitExp(exp)
+  private def checkParallel(ann: Annotations, loc: SourceLocation): List[CodeHint] = {
+    if (ann.isParallel) { CodeHint.Parallel(loc) :: Nil } else { Nil }
   }
 
   /**
     * Checks whether `sym` would benefit from `tpe` being pure.
     */
   private def checkEffect(sym: Symbol.DefnSym, tpe: Type, loc: SourceLocation)(implicit root: Root): List[CodeHint] = {
```
magnus-madsen, line 164: You should undo this change.

magnus-madsen: This PR is not about changing the meaning of the phase.

main/src/ca/uwaterloo/flix/language/phase/extra/CodeHinter.scala
```diff
@@ -15,350 +16,170 @@
...
-      visitExp(exp) ++ rules.flatMap {
-        case RestrictableChooseRule(_, body) => visitExp(body)
-      }
-
-    case Expr.Tag(_, exps, _, _, _) =>
-      visitExps(exps)
-
-    case Expr.RestrictableTag(_, exps, _, _, _) =>
-      visitExps(exps)
+  private case class CodeHintCandidates(traitOccurs: List[TraitSymUse], enumOccurs: List[(Symbol.EnumSym, SourceLocation)], defOccurs: List[DefSymUse], defCalls: List[(Symbol.DefnSym, List[Expr])])
```
magnus-madsen, line 110: Lets sort the members alphabetically.

main/src/ca/uwaterloo/flix/language/phase/extra/CodeHinter.scala
```diff
@@ -15,350 +16,170 @@
...
+    * Returns a [[CodeHintCandidates]] for the Flix project.
+    *
+    * @param root The root AST node of the Flix project
+    * @return     A [[CodeHintCandidates]] for the Flix project.
+    */
+  private def getCandidates(implicit root: Root): CodeHintCandidates = {
+    var traitUses: List[TraitSymUse] = Nil
+    var defUses: List[DefSymUse] = Nil
+    var enumUses: List[(Symbol.EnumSym, SourceLocation)] = Nil
+    var defCalls: List[(Symbol.DefnSym, List[Expr])] = Nil
```
magnus-madsen, line 122: sort fields alphabetically

main/src/ca/uwaterloo/flix/language/phase/extra/CodeHinter.scala
```diff
@@ -15,350 +16,170 @@
...
-    case Expr.TryCatch(exp, rules, _, _, _) =>
-      visitExp(exp) ++ rules.flatMap {
-        case CatchRule(_, _, exp) => visitExp(exp)
+    object UseConsumer extends Consumer {
+      override def consumeTraitSymUse(symUse: TraitSymUse): Unit = traitUses = symUse :: traitUses
+      override def consumeTraitConstraintHead(tcHead: TraitConstraint.Head): Unit = traitUses = TraitSymUse(tcHead.sym, tcHead.loc) :: traitUses
+      override def consumeDefSymUse(sym: DefSymUse): Unit = defUses = sym :: defUses
+      override def consumeExpr(exp: Expr): Unit = exp match {
+        case TypedAst.Expr.ApplyDef(symUse, exps, _, _, _, _) => defCalls = (symUse.sym, exps) :: defCalls
+        case _ => ()
```
magnus-madsen, line 130: sort the above.

main/src/ca/uwaterloo/flix/language/phase/extra/CodeHinter.scala
```diff
@@ -15,350 +16,170 @@
...
 
-    case Expr.TryCatch(exp, rules, _, _, _) =>
-      visitExp(exp) ++ rules.flatMap {
-        case CatchRule(_, _, exp) => visitExp(exp)
+    object UseConsumer extends Consumer {
+      override def consumeTraitSymUse(symUse: TraitSymUse): Unit = traitUses = symUse :: traitUses
+      override def consumeTraitConstraintHead(tcHead: TraitConstraint.Head): Unit = traitUses = TraitSymUse(tcHead.sym, tcHead.loc) :: traitUses
+      override def consumeDefSymUse(sym: DefSymUse): Unit = defUses = sym :: defUses
+      override def consumeExpr(exp: Expr): Unit = exp match {
+        case TypedAst.Expr.ApplyDef(symUse, exps, _, _, _, _) => defCalls = (symUse.sym, exps) :: defCalls
```
magnus-madsen, line 129: We don't use the term "call". We use "Apply" so call it applyDefs (this comment applies everywhere).

main/src/ca/uwaterloo/flix/language/phase/extra/CodeHinter.scala
```diff
@@ -15,350 +16,170 @@
...
-  private def visitEnum(enum0: TypedAst.Enum)(implicit index: Index): List[CodeHint] = {
-    val tagUses = enum0.cases.keys.flatMap(sym => index.usesOf(sym))
-    val enumUses = index.usesOf(enum0.sym)
-    val uses = enumUses ++ tagUses
-    val isDeprecated = enum0.ann.isDeprecated
-    val isExperimental = enum0.ann.isExperimental
-    val deprecated = if (isDeprecated) uses.map(CodeHint.Deprecated.apply) else Nil
-    val experimental = if (isExperimental) uses.map(CodeHint.Experimental.apply) else Nil
-    (deprecated ++ experimental).toList
+  private def considerDefCall(sym: Symbol.DefnSym, exps: List[Expr])(implicit root: Root): List[CodeHint] = {
```
magnus-madsen, line 71: These considerX should just be visitX as before

main/src/ca/uwaterloo/flix/language/phase/extra/CodeHinter.scala
```diff
@@ -15,350 +16,170 @@
...
   /**
     * Returns a collection of code quality hints for the given AST `root`.
     */
-  def run(root: TypedAst.Root, sources: Set[String])(implicit flix: Flix, index: Index): List[CodeHint] = {
-    val traitHints = root.traits.values.flatMap(visitTrait(_)(index)).toList
-    val defsHints = root.defs.values.flatMap(visitDef(_)(root, flix)).toList
-    val enumsHints = root.enums.values.flatMap(visitEnum(_)(index)).toList
-    (traitHints ++ defsHints ++ enumsHints).filter(include(_, sources))
+  def run(sources: Set[String])(implicit root: Root): List[CodeHint] = {
+    val cands = getCandidates
```
magnus-madsen, line 33: why not something like: getOccurrences? or getUses?

Then the case class could be "Uses"

magnus-madsen (commented):
Much better architecture. Just needs some cleanup and polish.

main/src/ca/uwaterloo/flix/language/phase/extra/CodeHinter.scala
LoZander: Yeah, okay. That's fair. If this indeed ought to be changed, it should be in a different PR. I get it :)

main/src/ca/uwaterloo/flix/language/phase/extra/CodeHinter.scala
LoZander: I did originally call it Uses, but I figured it didn't really reflect what it was that cleanly, since it also contains def application stuff, which are not really "uses" as such.

magnus-madsen: Occurrences then?

LoZander:
Other than the one ongoing conversation, I've implemented the suggested changes :)

magnus-madsen:
> Other than the one ongoing conversation, I've implemented the suggested changes :)

I suggest to call it Occurrences and then move on

LoZander:
I think it's all done now :)

LoZander:
Onto removing the index once and forall!

LoZander:
(unless there's something else that should be addressed in this code)

magnus-madsen (approved):

magnus-madsen:
Merged 👍

PR #9516, opened 2024-12-14
refactor: specialize `Parser2.name`

JonathanStarup:
the `tail` argument did nothing if `allowQualified = false`. so it seemed due to splitting.

- `name(kinds, allowQualified = true, tail, context)` -> `qualifiedName(kinds, tail, context)`
- `name(kinds, allowQualified = false, tail, context)` -> `name(kinds, context)`

I also documented the functions behaviour, and the documentation mirrors the complexity of the function


main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -639,11 +639,48 @@ object Parser2 {
...
+  /**
+    *   - `KIND` is consumed and produces [[TreeKind.Ident]]
+    *   - `KIND.KIND.KIND` is consumed and produces [[TreeKind.QName]]([[TreeKind.Ident]]..)
+    *   - `KIND.KIND.` is consumed and produces
+    *     [[TreeKind.QName]]([[TreeKind.Ident]].., [[TreeKind.ErrorTree]])
+    *   - `KIND. KIND.KIND` is consumed and produces
+    *     [[TreeKind.QName]]([[TreeKind.Ident]].., [[TreeKind.ErrorTree]], [[TreeKind.Ident]]..)
+    *   - [[TokenKind.isFirstDecl]] and [[TokenKind.isKeyword]] (not consumed)
+    *     produces [[TreeKind.ErrorTree]]
+    *   - `ANY` is consumed and produces [[TreeKind.ErrorTree]]
```
JonathanStarup, line 677: I am not sure how to do these comments without filling the screen, feels like i need a parsing rule specification language

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -639,11 +639,48 @@ object Parser2 {
...
+    *
```
magnus-madsen, line 661: It is really hard to understand the above.

magnus-madsen:
What is the goal of this change?

JonathanStarup:
1. To document and understand what `name` (and its new variants) does
2. To remove direct use of `name` since it has weirdly dependent flags

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
JonathanStarup: I am trying to describe 1) what is parsed 2) what is consumed 3) what tree is produced 4) are errors actively reported

And for rules like this, it doesn't just do one thing, but chooses among several different options

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -638,11 +638,48 @@ object Parser2 {
...
+    * @param tail if any KIND found is in `tail` then no more dots are parsed
+    *             (does nothing if `allowQualified` is false)
+    * @param allowQualified if `false` only one KIND is parsed
+    */
+  private def nameGeneric(kinds: Set[TokenKind], tail: Set[TokenKind], allowQualified: Boolean = false, context: SyntacticContext)(implicit s: State): Mark.Closed = {
```
magnus-madsen, line 682: I think nameGeneric is a very bad name. Any better suggestions?

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
magnus-madsen: But what does KIND.KIND.KIND even mean? And by Kind its not the kind of a type or?

magnus-madsen:
I will ask @mlutze if he understands this.

mlutze:
> I will ask @mlutze if he understands this.

I will try to understand it at my understanding prime time.

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
mlutze: I agree, it's confusing. I think it would be a bit easier if we had example strings instead of `KIND`

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
mlutze, line 682: Maybe it would be helpful if we compared this to `qualifiedName` to highlight when one would use one or the other.

JonathanStarup: you should never use nameGeneric - its just the shared implementation of name and nameQualified

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
JonathanStarup: @magnus-madsen but do you agree with the change, ignoring the comment? I can always make the comment as vague as it was before

magnus-madsen:
Sorry. I don't understand this refactor. By introducing `nameGeneric` it seems you still have the same problem you outline, but now instead we have to choose between `name`, `qualifiedName`, and `nameGeneric`. It seems to me this is just more likely to go wrong / lead to confusion. I am all for cleaning up the code, but this PR just leaves me more puzzled than before. Adding indirection is not always helpful, it can also lead to confusion.

JonathanStarup:
Okay i've repainted the picture - the change is the same (nameGeneric was only a private private helper function for the two others)

you either call `nameUnqualified` or `nameAllowQualified`. that's it

magnus-madsen:
> Okay i've repainted the picture at the cost of some code duplication - the change is the same (nameGeneric was only a private private helper function for the two others)
> 
> you either call `nameUnqualified` or `nameAllowQualified`. that's it

That sounds better. I will take a more detailed look later. 

Maybe @mlutze can chip in

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -783,13 +806,13 @@ object Parser2 {
...
     expect(TokenKind.KeywordUse, SyntacticContext.Use)
-    name(NAME_USE, allowQualified = true, context = SyntacticContext.Use)
+    nameAllowQualified(NAME_USE, context = SyntacticContext.Use)
     // handle use many case
     if (at(TokenKind.DotCurlyL)) {
       val mark = open()
       oneOrMore(
         namedTokenSet = NamedTokenSet.Name,
-        getItem = () => aliasedName(NAME_USE, tail = Set(TokenKind.NameLowerCase), SyntacticContext.Use),
+        getItem = () => aliasedName(NAME_USE, SyntacticContext.Use),
```
JonathanStarup, line 815: Here was an example of a given tail set that is unused because allowQualified was false - just as a example of why this is needed

magnus-madsen:
The IDE tests failing is that unrelated?

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -639,10 +639,36 @@ object Parser2 {
...
-    * If the found kind is in the tail set, then no further dot separated tokens will be consume
     */
-  private def name(kinds: Set[TokenKind], tail: Set[TokenKind] = Set(TokenKind.NameLowerCase), allowQualified: Boolean = false, context: SyntacticContext)(implicit s: State): Mark.Closed = {
+  private def nameUnqualified(kinds: Set[TokenKind], context: SyntacticContext)(implicit s: State): Mark.Closed = {
+    val mark = open(consumeDocComments = false)
+
+    // Check if we are at a keyword and emit nice error if so.
+    val current = nth(0)
+    if (current.isKeyword) {
+      // If the keyword is leads a declaration it's best to leave it be.
```
mlutze, line 649: "is leads"

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -639,10 +639,36 @@ object Parser2 {
...
+      if (!current.isFirstDecl) {
+        advance()
+      }
+      return closeWithError(mark, UnexpectedToken(
+        NamedTokenSet.FromKinds(kinds),
+        actual = Some(current),
+        sctx = context,
+        hint = Some(s"${current.display} is a keyword."),
```
mlutze, line 657: This seems like it will be very nice.

mlutze (approved):

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -639,10 +639,36 @@ object Parser2 {
 
   /**
     * Consumes a token if kind is in `kinds`.
-    * If `allowQualified` is passed also consume subsequent dot-separated tokens with kind in `kinds`.
-    * If the found kind is in the tail set, then no further dot separated tokens will be consume
     */
-  private def name(kinds: Set[TokenKind], tail: Set[TokenKind] = Set(TokenKind.NameLowerCase), allowQualified: Boolean = false, context: SyntacticContext)(implicit s: State): Mark.Closed = {
+  private def nameUnqualified(kinds: Set[TokenKind], context: SyntacticContext)(implicit s: State): Mark.Closed = {
```
magnus-madsen, line 643: Update doc comment

magnus-madsen:
@JonathanStarup Two minor things and then we can merge.

JonathanStarup:
Fixed

magnus-madsen (approved):

Issue #9511, opened 2024-12-14
Annotate Java's wrapper types with effects


magnus-madsen:
We need to annotate Java's wrapper types with effects in the JSON files. See #9470 for an example.

I suggest one PR per

- [ ] `java.lang.Char`
- [ ] `java.lang.Float`
- [ ] `java.lang.Double`
- [ ] `java.lang.Byte`
- [ ]  `java.lang.Short`
- [ ]  `java.lang.Integer`
- [ ]  `java.lang.Long`

Note that most operations are simply pure. 

But we need to read through Java doc to find silly things like `Boolean.getBoolean` which has the `Env` and `IO` effects.

When we start marking things as pure, we will need to remove `unsafe` expressions as seen in #9470

magnus-madsen:
@jaschdoc Brevity is good. So from my pov, it makes sense to mark the entire class as pure and then pick out the methods that have more effects and assign those specifically. 

(As opposed to verbosely declaring every method as pure).

jaschdoc:
- [ ] java.lang.Char
- [ ] java.lang.Float
- [ ] java.lang.Double
- [ ] java.lang.Byte
- [ ] java.lang.Short
- [ ] java.lang.Integer
- [ ] java.lang.Long

PR #9517, opened 2024-12-14
refactor: `unsafe exp` -> `unsafely effs run exp`

JonathanStarup:


main/src/library/Environment.flix
```diff
@@ -123,13 +123,13 @@ mod Environment {
     /// The returned value is never less than one.
     ///
     pub def getVirtualProcessors(): Int32 =
-        unsafe Runtime.getRuntime().availableProcessors()
+        unsafely {Env, IO, Sys} run Runtime.getRuntime().availableProcessors()
```
JonathanStarup, line 126: Its always `unsafely IO run exp` expect some places in Environment and one place ins String

magnus-madsen:
I want this PR-- but first I want Jakob to work on adding more annotations, see e.g. #9511.

Adding these annotations will significantly cut down on the unsafe/unsafely things.

JonathanStarup:
I mean if we don't merge he can remove unsafe blocks and if we do he can remove unsafely blocks

If we add the redundancy PR he will even get help if they are using unsafely

PR #9520, opened 2024-12-14
feat: `unsafely run` redundancy

JonathanStarup:
just for the very clear cases for now where either the effect or the expression is pure


main/src/ca/uwaterloo/flix/language/phase/Redundancy.scala
```diff
@@ -664,6 +664,13 @@ object Redundancy {
         case _ => visitExp(exp, env0, rc)
       }
 
+    case Expr.Unsafe(exp, runEff, _, _, loc) =>
+      (runEff, exp.eff) match {
+        case (Type.Pure, _) => visitExp(exp, env0, rc) + UselessUnsafe(loc)
+        case (_, Type.Pure) => visitExp(exp, env0, rc) + RedundantUnsafe(loc)
```
magnus-madsen, line 670: I think these checks are not strong enough.

We should check that the effect removed is actually present.

JonathanStarup: That is not very easy. I think this PR is a good step 0, and then the more advanced checks can get its own PR without all the AST stuff

magnus-madsen: Okay, but I don't want to start migrating to usafely before this check is in place.

magnus-madsen: What's the problem? Can we not check that exp.eff contains runEff?

JonathanStarup: both sides can contain variables. and what if you unsafely run IO on an expression with effect `ef`. is that redundant?

magnus-madsen: But can we at least at one more case where both sets are known?

JonathanStarup: Known in what sense? A ground formula is still complicated. I could do ground unions but then i still need a recursive type function to collect it into an n-ary union

I think it better to have this pr and then workshop it in another PR. this is simple and well-defined and has all the boilerplate

magnus-madsen: I can merge this PR but I will not accept any refactorings that use `unsafely` before this is in.

magnus-madsen:
You need to delete the `Indexer.scala` file

JonathanStarup:
fixed

magnus-madsen (approved):

PR #9529, opened 2024-12-15
feat: add `FlixListener`

magnus-madsen:


magnus-madsen:
The single event is just intended as an example.

mlutze (approved):

main/src/ca/uwaterloo/flix/api/FlixListener.scala
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+package ca.uwaterloo.flix.api
+
+trait FlixListener {
+
+  /**
+    * Invoked to notify the listener of the given Flix event.
+    */
+  def notify(e: FlixEvent): Unit
```
JonathanStarup, line 23: we should mention that the implementation must be threadsafe

PR #9539, opened 2024-12-16
feat: `instance Indexable[String]`

JonathanStarup:


mlutze:
I'm not sure we should do this. Indexing into strings is dangerous because you could end up in the middle of a codepoint.

magnus-madsen:
> I'm not sure we should do this. Indexing into strings is dangerous because you could end up in the middle of a codepoint.

Agreed. This is often considered an anti pattern.

JonathanStarup:
But what's the difference between using `s[i]` and `String.charAt(s, i)`

We use it throughout our parser fx. it think its a very natural thing to do

mlutze:
> But what's the difference between using `s[i]` and `String.charAt(s, i)`
> 
> We use it throughout our parser fx. it think its a very natural thing to do. It exists and its messy but that's sort of inherited from Java unless we completely use codepoints

`charAt` should go too :)

JonathanStarup:
I mean that's a bit like saying our parser is a program that is invalid and shouldn't exist - isn't that a bit much?

mlutze:
> I mean that's a bit like saying our parser is a program that is invalid and shouldn't exist - isn't that a bit much?

Maybe. But maybe our parser would get into trouble if we feed it e.g. Japanese characters

JonathanStarup:
Yeah probably, or emojis who knows, but it works for its expected input. who does regex work with codepoints? I don't think we can change the world from our side in this regard. This pr is just about if people do it in an ugly way or a nice way

JonathanStarup:
> But bad things should be ugly!

I think ideally that bad things should be clear and explicit - but it also shouldn't make every line of code painful. In some sense a "I am using chars directly and I am aware of my actions" notice at the top of the file makes it clear and explicit and then the code could just use `s[i]`. This is just the illustrate the spirit of it

mlutze:
Maybe we should introduce an `ugly` block.

magnus-madsen:
I am strongly against this. It is widely considered a mistake to index into UTF8 strings. 
Adding an Indexable trait seems to just make everything worse.

JonathanStarup:
why does our lexer do it then? 

also - you dont know what the user knows. they might not have a utf string.

JonathanStarup:
im not saying this pr neccessirily have to be merged - im just saying that our absolutist ban on conditionally useful functions (list indexing, string char lookup, etc) are not sustainable imo.

sometimes a slow random index is exactly and precisely what you need, likewise with char lookups in strings

mlutze:
> im not saying this pr neccessirily have to be merged - im just saying that our absolutist ban on conditionally useful functions (list indexing, string char lookup, etc) are not sustainable imo.
> 
> sometimes a slow random index is exactly and precisely what you need, likewise with char lookups in strings

Sounds like a topic for a group meeting :)

magnus-madsen:
Pending further discussion

Issue #8762, opened 2024-10-01
Introduce `syntactic` and `semantic` completers


magnus-madsen:
Refactor our completers to fall into one of the two categories.

PR #9543, opened 2024-12-17
refactor: introduce SemanticCompletionProvider

stormckey:
As said in #8762. We introduce SemanticCompletionProvider to hold the method to provide semantic completions.


magnus-madsen:
Conflicted.

stormckey:
Solved

magnus-madsen:
Do we want this now given that we are going for a big switch?

stormckey:
Emmm you're right, I think we no longer need to refactor the code this way. Maybe close this PR?

PR #9544, opened 2024-12-17
feat: introduce KindCompleter

### LINKED ISSUE #8582: Add `KindCompleter`
Based on `UndefinedKind`

stormckey:
fixes #8582
Preview: 
image
image
image

Note:
- I use the Highest Priority since this is the most useful completion for UndefinedKind.
- There is no CompletionItemKind for Kind so I just use TypeParameter. Or should we use something others?


magnus-madsen (approved):

magnus-madsen:
I agree with both of your notes.

PR #9547, opened 2024-12-17
Proposal: Pretty print types in DNF

JonathanStarup:
This is definitely more useful than existing types, but also not fantastic.

Before merging we should also test it thoroughly - would suck to show wrong types

Example
```
def bain(f: Unit -> Unit \ ef, g: Unit -> Unit \ ef2, h: Unit -> Unit \ ef3): Unit \ IO = // IO should be {Net, IO}
    f(); g(); h();
    run {
        let url = "http://example.com/";
        Logger.info("Downloading URL: '${url}'");

        match HttpWithResult.get(url, Map.empty()) {
            case Result.Ok(response) =>
                let file = "data.txt";
                Logger.info("Saving response to file: '${file}'");

                let body = Http.Response.body(response);
                match FileWriteWithResult.write(str = body, file) {
                    case Result.Ok(_) =>
                        Logger.info("Response saved to file: '${file}'")
                    case Result.Err(err) =>
                        Logger.fatal("Unable to write file: '${err}'")
                }
            case Result.Err(err) =>
                Logger.fatal("Unable to download URL: '${err}'")
        }
    } with FileWriteWithResult.runWithIO
      with HttpWithResult.runWithIO
      with Logger.runWithIO
```
On master the shown effect is `ef3 + ((Net + IO + ((((Logger + Logger + Logger + HttpWithResult + Logger + Logger + FileWriteWithResult) & (~FileWriteWithResult)) + IO) & (~HttpWithResult))) & (~Logger)) + IO + ef + ef2`

This pr the shown effect is `ef3 + ef2 + IO + Net + ef`


JonathanStarup:
@magnus-madsen Its a bit messy atm, but i programmed it close to zhegalkin, skipping explicit constants just to make it shorter. I also handle non-wellformed things in a nice way with Chunk

I think this is a good step 1 (with polish and testing) and then we can do smarter things later

main/src/ca/uwaterloo/flix/language/phase/TypeSimplifier.scala
```diff
@@ -0,0 +1,284 @@
+package ca.uwaterloo.flix.language.phase
+
+import ca.uwaterloo.flix.language.ast.{Kind, SourceLocation, Symbol, Type, TypeConstructor}
+import ca.uwaterloo.flix.util.CofiniteEffSet
+
+import scala.collection.mutable
+
+
```
magnus-madsen, line 8: A sketch of the algorithm would be nice.

magnus-madsen:
> Before merging we should also test it thoroughly - would suck to show wrong types

Yes that would suck.

JonathanStarup:
!!!jar

magnus-madsen:
- Minus counter
- Plus some more sorting of "effect kinds"
- Polish

PR #9548, opened 2024-12-17
fix: lexer positions

JonathanStarup:
`Lexer.retreat` didn't handle lines/columns properly and I couldn't comprehend how `State.end` was tracked.

Now i save the width of the previous line, making the end position computable on the spot when building tokens instead of keeping track of it dynamically. This width of the previous line needs to be recomputed if you retreat across multiple lines. In a follow up PR i think we should remove `retreat` and instead store checkpoints (line/col/offset/prevWidth) and just directly reset instead of going back one by one. So i judge this potential performance cost small and gone in the next PR.

These new positions agree with all locations of the existing lexer, except for some lines ending with an integer literal, in which the existing lexer gives a wrong location.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
@@ -69,20 +69,33 @@ object Lexer {
     */
   private def isDigit(c: Char): Boolean = '0' <= c && c <= '9' || 'a' <= c && c <= 'f' || 'A' <= c && c <= 'F'
 
-  /**
-    * The internal state of the lexer as it tokenizes a single source.
-    * At any point execution `start` represents the start of the token currently being considered.
-    * Likewise `end` represents the end of the token currently being considered,
-    * while `current` is the current read head of the lexer.
-    * Note that both start and current are `Position`s since they are not necessarily on the same line.
-    * `current` will always be on the same character as or past `start`.
-    * As tokens are produced they are placed in `tokens`.
-    */
+  /** The internal state of the lexer as it tokenizes a single source. */
   private class State(val src: Source) {
+
+    /** `start` is the first position of the token that is currently being lexed. */
     var start: Position = new Position(0, 0, 0)
+
+    /**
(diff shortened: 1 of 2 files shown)


JonathanStarup:
!!!jar

JonathanStarup:
Also looks good with manual review

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -175,15 +187,38 @@ object Lexer {
...
       s.current.column -= 1
-      s.end.column -= 1
     }
   }
 
+  /**
+    * Assuming that `s.current` is on the end of a line (i.e. on `\n`), recomputes the length of
+    * the previous line.
+    */
+  private def recomputePrevWidth()(implicit s: State): Unit = {
```
magnus-madsen, line 201: Will this be slow?

magnus-madsen:
How can we test this? It could break source locations in non-obvious ways.

magnus-madsen:
I feel like we really need that source location checker...

JonathanStarup:
I ran it with the old `s.end` location and checked that they agreed. Then manually checked a random subset of them. They were all on literal int tokens and related to end-of-line things on which the existing lexer has a bug. 

But yeah working forward we should have a way to test it

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
JonathanStarup: It traverses a line, so somewhat. but 1) there is a (a little) less overhead during normal advance nad 2) I think we should delete retreat and do it with checkpoints, then this function will also be deleted 3) it is only triggered if you retreat across a line 4) it is stored in state, so its not necessarily spending effort. 5) i could do it on-demand if needed, but due to point 2 i don't think its worth it

magnus-madsen:
I am not against, but I think I need a more in depth in person description.

mlutze (approved):
I approve on condition that we move to checkpoints afterward

magnus-madsen (approved):

PR #9556, opened 2024-12-18
feat: remove `Sendable`

JonathanStarup:


JonathanStarup:
@magnus-madsen should i go through with this? I can PR peoples repos to make it less annoying

magnus-madsen:
Yes. I want this out. We should strive towards a solid foundation and Sendable is not part of that for now.

JonathanStarup:
Oh actually, we cant fix the PRs before this is merged, then they wouldn't type check. So we just need to merge and fix afterwards

magnus-madsen (approved):

magnus-madsen:
@JonathanStarup I have approved it, so once you have the PRs open you can just press merge.

PR #9563, opened 2024-12-18
feat: rework incremental plot

magnus-madsen:
image


mlutze (approved):

magnus-madsen:
> could we have this on arewefast?

If someone adds it.

Issue #4276, opened 2022-07-15
Feature idea: Effect highlighting


JonathanStarup:
I want to be able to click on an effect in the header of a function and have the expressions are cause that effect to be highlighted, like so:
![image](https://user-images.githubusercontent.com/32037926/179214501-ccd7a46c-dfde-42e0-8ca3-596bd76b46d0.png)

This could be implemented in gradually more complex steps:
1) Highlight base expressions: `ref 2 @ r`, `[] @ r`, `do F()`
2) Highlight direct function calls `f(1, 2)` where `def f(x: Int32, y: Int32): Int32 & r`
3) Highlight all non-composite expressions that cause the effect.

We've discussed implementing it by gathering all expressions that has the effect in question and the filtering out expressions that cannot have caused it. `e1 ; e2` would be filtered away fx. since it must be from `e1` or `e2`.

magnus-madsen:
I think an easy starting point is to look at the do-effect.

magnus-madsen:
@LoZander Would it be easy to extend the HighlightProvider with this as a proof of concept?

The simplest scenario I can think of is that I want to click on an effect and then it should highlight all uses of that effect.

LoZander:
Yeah, I think that should be doable. I'll take a look at it once it soon :)

LoZander:
I've started work on this :)

LoZander:
Oh, I missed the part where both of you suggested doing this incrementally. I kinda just went all out ^^'

LoZander:
My approach so far is to gather all expressions in the file and then depending on the type of expression, decide if we should consider highlighting it. For instance, for `2 + f(3)` for some impure `f`, we shouldn't highlight the entire expression, but just the most "precise" source which would be `f(3)`. When we "consider" the expression, we also filter it out if it's not in the function we care about.

LoZander:
> I want to be able to click on an effect in the header of a function and have the expressions are cause that effect to be highlighted, like so: ![image](https://user-images.githubusercontent.com/32037926/179214501-ccd7a46c-dfde-42e0-8ca3-596bd76b46d0.png)
> 
> This could be implemented in gradually more complex steps:
> 
> 1. Highlight base expressions: `ref 2 @ r`, `[] @ r`, `do F()`
> 2. Highlight direct function calls `f(1, 2)` where `def f(x: Int32, y: Int32): Int32 & r`
> 3. Highlight all non-composite expressions that cause the effect.
> 
[...]

PR #9566, opened 2024-12-19
feat: contains function for SourceLocation

LoZander:
This adds a `contains` function to `SourceLocation`s as well as a partial ordering on `SourcePosition`s used by the aforementioned.

This is related to #4276 and will be used by the `HighlightProvider`.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/SourceLocation.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/SourceLocation.scala
@@ -54,6 +54,12 @@ case class SourceLocation(isReal: Boolean, sp1: SourcePosition, sp2: SourcePosit
     */
   def beginCol: Int = sp1.col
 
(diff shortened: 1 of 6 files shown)


LoZander:
Right, yes. I'll get on that. Late night :p

LoZander:
Documentation and tests have been added :)

magnus-madsen (approved):

JonathanStarup:
i've though about this for a while, nice :)

Is this for the location verifier? or just for nicer code in general?

LoZander:
> i've though about this for a while, nice :)
> 
> Is this for the location verifier? or just for nicer code in general?

Do you mean the `InsideAcceptor`?

I just needed it for the effect highlighting to filter out expressions not in the function we're concerned with, and instead of adding it as a local private function, I thought I might as well add it to SourceLocations in general, since arguably that is where this kinda thing belongs and I could see it being useful in general :)

PR #9567, opened 2024-12-19
feat: effect highlighting

LoZander:
This PR adds effect highlighting. This means that when the cursor is on an effect in a function signature, any expressions that produce said effect will be highlighted.

Related to #4276.


LoZander:
I think it's mostly done now. All that's left is filtering out highlights from outside the function in question, which depends on #9566 and then being able to handle the case where there are multiple effects in the signature. Other then that, it's just polishing, a bit of manual sanity testing and another lookover to be sure I didn't miss something. When it comes to polishing, there's def some stuff I want to refactor.

main/src/ca/uwaterloo/flix/api/lsp/provider/HighlightProvider.scala
```diff
@@ -261,58 +264,180 @@ object HighlightProvider {
...
+    val retEffHighlight = DocumentHighlight(Range.from(loc), DocumentHighlightKind.Write)
+    val expHighlights = exps.map(loc => DocumentHighlight(Range.from(loc), DocumentHighlightKind.Read))
+    val highlights = retEffHighlight :: expHighlights
+
+    ("status" -> ResponseStatus.Success) ~ ("result" -> JArray(highlights.map(_.toJSON)))
+  }
+
+  private case class EffectConsumer(consider: Expr => Unit) extends Consumer {
+    override def consumeExpr(exp: Expr): Unit = exp match {
+      case Expr.Cst(_, _, _) => ()
```
magnus-madsen, line 363: This seems to defeat the entire point of using the Visitor?

LoZander: I agree that it's a lot, but it's just about identifying what kind of expression it is, so it's not really doing what the Visitors job. The reason I do it is so we can filter out the kinds of expression that we don't want to highlight. If we highlighted every Expr with the right effect, we'd highlight all the composite ones too, like `e1; e2` when we just care about `e2` etc.

LoZander: It could be make less verbose by introducing a catch-all case for all the kinds of expressions we don't care about. But if your concern is about separating concerns and responsibility then I suppose that doesn't solve anything

magnus-madsen: Why not highlight just ApplyOp?

LoZander: There is no ApplyOp. There is `do` which does that I suppose, but it would just not cover all the sources of effects in a function body. For instance it wouldn't cover calling a function that has the effect in question. Then there are Array and Struct operations, effect casts and `do`. All of these are or may be impure. Then there's java interop stuff.

There are a couple of things on there that I could maybe remove. I don't believe  `Unary and Binary ops` can be impure. And then there are a couple of intrinsics that there is just no reason to support since they aren't allowed.

LoZander: What we could do is make the visitor gather all the expressions with the effect in question and then filter them afterwards. We would still need a big match case like that, but it would be separate from the consumer. Maybe that's better?

LoZander: @JonathanStarup also mentioned that that had been brought up as an idea for how to approach it

LoZander: we could do that, but then we only highlight do expressions. We want all sources of the effect to be highlighted, at least eventually, right?

LoZander: Do you just want a smaller pr initially?

JonathanStarup: Doing the general "where does this effect come from" seems very hard (fx if `f(exp1, exp2)` has effect Crash, it could be because `f` introduces Crash, `f` could introduce crash because of a higher order effect of either exp1 or exp2, the application could have crash because exp1 or exp2 or both has Crash, in which case the application of f is not the cause of Crash). So even with this approach you can't get the full ideal solution. 

That why I think magnus just want the simple thing first - get it started with small steps

LoZander: oh, good point. I see. Alright. I'll just remove the other cases for now then

LoZander: Then we can maybe thing about a more sophisticated and full solution afterwards

LoZander: maybe some kind of consumer-acceptor combo that finds the precise origins of an effect

LoZander: I've limited the support to highlighting `do` expressions

JonathanStarup: Sounds good - yeah i hope we get to it, seems complicated but also useful

LoZander:
I thought it would be that simple, but a lot of the complexity comes from the fact that we need more information to determine if we want to highlight the effects than for the other highlights.

It is not enough to know that the cursor is on an effect, we need to know that it is on an effect that is within a function spec. So we need to consider more than just the top element of the stack. But it gets worse. Since effect union is represented as an apply type, `Type.Cst(TypeConstructor.Effect(...),...)` can be nested within an arbitrary number of other types. This means that the stack will look something like `Def :: Type :: ... :: Type`.

So when the cursor is on an effect, to figure out if we want this specific kind of effect highlighting, we ha we have to search back through the stack until we either find the def that contains the effect or something that isn't a type. If the former happens, then we know this effect is in a def decl, if the latter happens, then we know we're not.

I suppose that we could simply filter out effects from the stack and then in these cases the top element would be the def and then we could just manually check, if we're actually more specifically in the effect. But do `Type.Cst(TypeConstructor.Effect(...),...) not occur in other contexts where the cursor would also be? If so, we'd lose that or at least add multiple other cases where we need to manually check if we're actually in an effect and then it gets messy.

An additional complication was that some of the filtering I was doing on the stack caused problems, so I had to change it to just filter the top element.

LoZander:
It can certainly be refactored to look cleaner though

LoZander:
For instance, these new `highlightX` functions don't work with the `HighlightBuilder` so I had to repeat some of that code. Generally, I wanna refactor this provider to use the same pipeline structure as the others, where each of the `highlightXSym` function is instead a `getXOccurs` that returns a simple list or set of `SourceLocation`. It gives a better separation of concerns and then these functions don't need to repeat the code for the creation of highlights

LoZander:
and then there's also the complication of concrete effects vs. effect type vars

LoZander:
> I thought it would be that simple, but a lot of the complexity comes from the fact that we need more information to determine if we want to highlight the effects than for the other highlights.
> 
> It is not enough to know that the cursor is on an effect, we need to know that it is on an effect that is within a function spec. So we need to consider more than just the top element of the stack. But it gets worse. Since effect union is represented as an apply type, `Type.Cst(TypeConstructor.Effect(...),...)` can be nested within an arbitrary number of other types. This means that the stack will look something like `Type :: ... :: Type :: Def :: ...`.
> 
> So when the cursor is on an effect, to figure out if we want this specific kind of effect highlighting (and to find what function we then have to consider expressions from), we have to search back through the stack until we either find the def that contains the effect or something that isn't a type. If the former happens, then we know this effect is in a def spec, if the latter happens then we know it's not.
> 
> I suppose that we could simply filter out effects from the stack and then in these cases the top element would be the def and then we could just manually check, if we're actually more specifically in the effect. But do `Type.Cst(TypeConstructor.Effect(...),...) not occur in other contexts where the cursor would also be on it? If so, we'd lose that or at least add multiple other cases where we need to manually check if we're actually in an effect and then it gets messy.
> 
> An additional complication was that some of the filtering I was doing on the stack caused problems for this highlighting (it removed too much information), so I had to change it to just filter the top element.

I actually just realised I haven't pushed the code that handles the nesting of effects yet, so unfortunately it is gonna get a little worse ^^'

LoZander:
Maybe a smarter consumer and/or acceptor could simplify things. I don't like having to backtrack on the stack to find the information I need. I mean, I suppose that is why it's a stack rather than a single most "recent" element, but still

magnus-madsen:
This is a good example of incidental complexity. The change here is conceptually straightforward, so why does the architecture not support a 5 line change? And what architecture would support it?

In my mind this should be a straightforward lookup followed by a straightforward search.

LoZander:
> This is a good example of incidental complexity. The change here is conceptually straightforward, so why does the architecture not support a 5 line change? And what architecture would support it?
> 
> In my mind this should be a straightforward lookup followed by a straightforward search.

I see your point. If we want the architecture to better support this, then I think we need to either change the Visitor or the AST. If the effect knows that it's in a def spec, then this would be easy. But then again, there might be any number of cases where we need that kind of information and that information is in the ASTs structure already. I think it would make sense to consider if the Visitor can be generalised. The problem currently is that we need a way to tell child-parent relations between things visited. This information lies in the structure of the AST and the Visitor throws that information away. There is no direct way to tell if two things visited are siblings, cousins or descendents/ancestors, except manually inspecting their SourceLocations. The AST has this information, we just need a way for the Visitor to preserve it, for cases where it might be relevant, like here.

LoZander:
Alternatively, if ASTs had a reference to their parent, that would work, but that's not very in line with the generally functional approach of the compiler and not a particularly attractive idea.

There might also be some way to do a smarter consumer that could solve this issue. Maybe something like everytime you find a def, you associate the def with all of it's effects in some way, like a map or something. That way you can just look up an effects def (if there is one) in `Type.Cst(TypeConstructor.Effect(...),...)` case. I think that actually kinda makes sense. It won't be too difficult, complicated or costly I reckon

LoZander:
or would that also be incidental complexity? Do you think I should dig deeper than that?

JonathanStarup:
what are they things that would want to highlight now? it sounds like you are looking at arbitrary function calls? (note that you cannot syntactically look through effects, since `ef - IO` does not produce the IO effect)

I think the version 0 of this functionality is just for `eff Abort { def abort(msg: String): Void }` to highlight concrete calls to `Abort.abort`. Then you only have to care about def symbols

another PR could add java effects (again only needs to reason about methods, not arbitrary effects)

LoZander:
> what are they things that would want to highlight now? it sounds like you are looking at arbitrary function calls? (note that you cannot syntactically look through effects, since `ef - IO` does not produce the IO effect)
> 
> I think the version 0 of this functionality is just for `eff Abort { def abort(msg: String): Void }` to highlight concrete calls to `Abort.abort`. Then you only have to care about def symbols
> 
> another PR could add java effects (again only needs to reason about methods, not arbitrary effects)

Oh, no that's not what I'm trying to get. I might have not been clear about it. The problem I'm talking about is how to determine, when the cursor is on an effect, if this effect is in a def spec. We only want to do this effect highlighting when the cursor is specifically on an effect annotation in a def spec, and furthermore, we need information about the def in question to know where what scope to limit the highlighting to.

This sounds simple, and it should be, but it's not given the current stack consumer approach.

So the `HighlightProvider` generally works as follows: We search through the AST using the visitor with the `StackConsumer` and `InsideAcceptor`. This gives as a stack which acts as a kind of trace of the search, where the top element is the final AST node visited. We then inspect this node, and if it's a node for which highlighting makes sense, we find the occurrences of the symbol and highlight them.

This works fine for the Highlighting up to this point, but for this new effect highlighting, there is a complication. The top most element would here be `Type.Cst(TypeConstructor.Effect(...),...)`. But that doesn't tell us if it's an effect annotation in a def spec or in some other case where one might occur. Furthermore, it doesn't tell us _what_ def it describes, which we need to know as mentioned before.

Currently, the solution in the PR is that when we find a `Type.Cst(TypeConstructor.Effect(...),...)` (or a polymorphic effect) we search further down in the stack, basically tracing back up through the nodes visited. If this effect annotation is in a def spec, we should find only other types (other effects) until we find a Def. If this effect annotation is not in a def, then we should at some point find something that is neither a type nor def. Depending on this, we know whether to do the effect highlighting and the def whose body to highlight in.

This is annoying complicated for how simple the idea is. Hence the incidental complexity that @magnus-madsen mentions. That it is this complicated seems to suggest that there is some architectural issue that I'm working around. I agree. It should be simple.

I think the solution might be a different consumer. Perhaps one that maps effects annotations to their def (if they have one). Then in the effect annotation case, we can simply query this map on whether it is part of a def spec and for what def if it is. But I don't know if this is the best solution or if this is also working around a deeper issue.

LoZander:
Sorry for being long-winded

magnus-madsen:
> Oh, no that's not what I'm trying to get at. I might have not been clear about it. The problem I'm talking about is how to determine, when the cursor is on an effect annotation, if this annotation is in a def spec. We only want to do this effect highlighting when the cursor is specifically on an effect annotation in a def spec, and furthermore, we need information about the def in question to know where what scope to limit the highlighting to.

Why? Where is this design requirement coming from?

magnus-madsen:
Think about this: What is the simplest thing you can come up with that still adds value?

LoZander:
the simplest thing would be to just highlight all matching do expressions in the file when the cursor is on some effect annotation. maybe that would be fine for now? Maybe Ive been too hung up on matching the original idea exactly?

LoZander:
I'm gonna go with the approach I outlined above. At this point, it's easier to just do a new PR, so I will close this one

Issue #9355, opened 2024-11-28
Merge completions from LocalScopeCompleter, ExprCompleter and AutoUseCompleter


stormckey:
As discussed in last pr, we can divide the situations into three cases where a def completion is needed:
- List.ma: we want a completion to List.map
- ma (List.map not in scope): we want a completion to map and auto use List.map
- ma (List.map in scope): we want a completion to map

But there are three difference completers responsible for those:
- List.ma: DefCompleter
- ma (List.map not in scope): AutoUseCompleter
- ma (List.map in scope): LocalScopeCompleter

And current we just disable other Completers when it's not their time to work.
But that brings several problems:
- it makes code harder to understand and maintain: this division is not documented anywhere, and manually disable other completer is verbose and error-prone, it should be something like a branch.
- duplicated code: for case 2 and 3, we all need to check the existence of List.map in scope, and they both need to find information from root.defs to build a informative completion.

What do you think? I think keeping these functionalities as different branches in the same completer could help to keep code clear and concise. Or is there something that I miss?

magnus-madsen:
One problem is that they rely on different information-- which is an argument for keeping them separate.

I think this would be good to discuss at the next meeting to get some input.

stormckey:
Here is a checklist of completers to be merged.

- [ ] Def
- [ ] Enum
- [ ] Struct
- [ ] Effect

I will figure out how to merge the rest things and extend the list.

PR #9570, opened 2024-12-19
refactor: merge DefCompleter

stormckey:
According to #9355, we are now merge different parts of DefCompleter from LocalScopeCompleter, ExprCompleter and AutoUseCompleter into a single DefCompleter.
Preview:
image
image
image
image

But some problems remain:
- There used to be another guard named `canApplySnippet` that will check if we should use a snippet in situations with `|>` or similar things. But that needs the `previousWord` information from `CompletionContext`. To preserve that guard, we should add this to the error or find the `previousWord` anyway in the completer.
- Some ParseError will carry a `SyntacticContext.Expr.OtherExpr` that will now trigger `ExprCompleter`. If we remove `DefCompleter` and other completers gradually from `ExprCompleter`, how should we handle these `ParseError`? This applies to many other completers too. Something used in `DefCompleter`, like LocalScope or AnchorPosition, is not available just from `SyntacticContext.Expr.OtherExpr`.
- `DefCompleter` now provides completions for defs in scope, defs out of scope and qualified defs. But offering completions for qualified defs still suffers from the dot problem we met before. I suggest that we just provide completions for defs in scope and out of scope in `DefCompleter`, while delegating the completions for qualified defs and other qualified things to the something like a "NameSpaceCompleter", as we can handle with the dot problem in one place and provide completions for all sorts of things in the namespace.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
@@ -231,21 +231,21 @@ object CompletionUtils {
   }
 
   /**
-    * Updates the given `qn` based on the character immediately following the location.
+    * Parse given `qn` based on the character immediately following the location.
     * If the character is a dot, we move the ident to the namespace.
-    * Otherwise, we will keep the `qn` as it is.
+    * Otherwise, we will just take namespace and ident out of qn.
     *
     * Example:
-    *   - Source "AA.BB", QName(["AA"], "BB") -> QName(["AA"], "BB")
-    *   - Source "AA.BB.", QName(["AA"], "BB") -> QName(["AA", "BB"], "")
+    *   - Source "A.B.C", QName(["A", "B"], "C") -> ("A.B", "C")
+    *   - Source "A.B.C.", QName(["A", "B"], "C") -> ("A.B.C", "")
     */
-  def updateQNameBasedOnDot(qn: QName, loc: SourceLocation): (List[String], String) = {
+  def getNamespaceAndIdentFromQName(qn: QName, loc: SourceLocation): (String, String) = {
     val ident = if (followedByDot(loc)) "" else qn.ident.name
     val namespace = qn.namespace.idents.map(_.name) ++ {
       if (followedByDot(loc)) List(qn.ident.name)
       else Nil
     }
-    (namespace, ident)
+    (namespace.mkString("."), ident)
   }
 
   /**
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
@@ -239,13 +239,13 @@ object CompletionUtils {
     *   - Source "AA.BB", QName(["AA"], "BB") -> QName(["AA"], "BB")
     *   - Source "AA.BB.", QName(["AA"], "BB") -> QName(["AA", "BB"], "")
     */
-  def updateQNameBasedOnDot(qn: QName, loc: SourceLocation): QName = {
+  def updateQNameBasedOnDot(qn: QName, loc: SourceLocation): (List[String], String) = {
     val ident = if (followedByDot(loc)) "" else qn.ident.name
     val namespace = qn.namespace.idents.map(_.name) ++ {
       if (followedByDot(loc)) List(qn.ident.name)
       else Nil
     }
-    Name.mkQName(namespace, ident, SourceLocation.Unknown)
+    (namespace, ident)
   }
 
   /**
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
@@ -1,5 +1,6 @@
 /*
  * Copyright 2022 Paul Butcher, Lukas Rønn, Magnus Madsen
+ * Copyright 2025 Chenhao Gao
  *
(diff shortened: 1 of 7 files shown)


stormckey:
I agree that we need to provide detailed completions for each construct. So I think it's fine to deal with the dot thing in each completer. I've written a util function `updateQNameBasedOnDot` that can be shared between completers.

stormckey:
The fuzzyMatch also works.
image
image
I've thought of a possible update for fuzzyMatch and I will talk about that in another ticket.

JonathanStarup:
I think that there are two jobs:
- figuring out what context we are in / what is happening (`CompletionProvider.autoComplete`)
- Computing which autocompletions to give (`XYZCompleter`)

And things like "does this name end with a dot" and "does this this have |> before it" belongs to the first job, so it should go in the autoComplete function, not the completers

stormckey:
> Things like "does this name end with a dot" and "does this this have |> before it" belongs to the first job, so it should go in the autoComplete function, not the completers

It would be great if we could separate these two phases completely and make the architecture clean. But the problem is the information needed to compute the context is specific to the error. For dot thing, we need to QName and SourceLocation from some errors. For "|>" thing, we also need the specific Def used there.

I am not sure if there is an easy way to collect some information from some errors and give a context for all completers. But now, for example, we just need to call `getQNameBasedOnDot` every time we want to check for the dot, which is relatively lightweight.

JonathanStarup:
> But the problem is the information needed to compute the context is specific to the error. For dot thing, we need to QName and SourceLocation from some errors. For "|>" thing, we also need the specific Def used there.

Maybe I'm misunderstanding, but couldn't autocomplete figure this out? maybe with a helper function. And the call the appropriate completer that now doesn't have to search for `|>` or multiple errors etc?

stormckey:
I may get your point and have lift the updating for QName to autoComplete. The updated QName will be shared with following completers.

Let me know if I get it wrong.

magnus-madsen:
I think the agreement was to drop support for `|>` special completions for now.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
```diff
@@ -228,4 +229,33 @@ object CompletionUtils {
...
+  /**
+    * Updates the given `qn` based on the character immediately following the location.
+    * If the character is a dot, we move the ident to the namespace.
+    * Otherwise, we will keep the `qn` as it is.
+    *
+    * Example:
+    *   - Source "AA.BB", QName(["AA"], "BB") -> QName(["AA"], "BB")
+    *   - Source "AA.BB.", QName(["AA"], "BB") -> QName(["AA", "BB"], "")
+    */
+  def updateQNameBasedOnDot(qn: QName, loc: SourceLocation): QName = {
```
magnus-madsen, line 242: As a general rule, we should __never__ construct QNames as they come from the AST. 

Why do you need to do that?

magnus-madsen:
This behavior is strange:

image

magnus-madsen:
This is also strange:

image

magnus-madsen:
There are some things to work out

magnus-madsen:
And this gives no suggestions:

```
use Fixpoint.Ast.
```

stormckey:
> This behavior is strange:
> 
> image

I can not repeat this. How did you trigger this?

stormckey:
> This is also strange:
> 
> image

Sorry but I think this comment and the following ones should refer to #9486 since we are refactoring the UseCompleter there.

JonathanStarup:
> I think the agreement was to drop support for `|>` special completions for now.

i think its fine as long as the logic is in the dispatcher. but removing is also fine

magnus-madsen:
Yes. I messed up the tickets.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
stormckey: OK. I will use a List[String] for the namespace and a String for the ident.

stormckey:
Is there any further problem with this PR?

One problem is that ExprCompleter (still used by ParseError) now will not provide completions for defs, since the current DefCompleter can't be called with a context. 

Maybe we could gradually remove completers from ExprCompleter and see if we can still have a good life.

On the other hand, it would be great that we merge this PR so that I can continue to merge other completers.

magnus-madsen:
I will have to a look at it again in detail-- but my understanding was that it was broken (?) Does it work well?

Also, there is the dot-problem which we should try to resolve, but that may require parser changes.

magnus-madsen:
!!!jar

magnus-madsen:
@stormckey I may come find you and we can look at it together.

stormckey:
> @stormckey I may come find you and we can look at it together.

OK

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
```diff
@@ -228,4 +229,33 @@ object CompletionUtils {
...
+    (namespace, ident)
+  }
+
+  /**
+    * Returns true if the character immediately following the location is a dot.
+    * Note:
+    *   - loc.endCol will point to the next character after QName.
+    *   - loc is 1-indexed, so we are actually checking the character at loc.endCol-1.
+    */
+  private def followedByDot(loc: SourceLocation): Boolean = {
```
magnus-madsen, line 257: Change this as follows:
- Instead of loc, we take the qname of the error.
- We then check if qname.loc.column + 1 is a dot (we have to check its within bounds)

magnus-madsen (approved):

PR #9571, opened 2024-12-19
refactor: get rid of `Lexer.retreat`

JonathanStarup:
retreat has a bug with positions so now there is a simpler checkpoint mechanism. This might be a bit slower but its used rarely. 2/3 places could also avoid using checkpoints with a new version of the isMatch function

This PR changes some source locations a la #9548, but it that PR is merged first, then this will have no location changes


magnus-madsen:
@JonathanStarup Conflicted

magnus-madsen (approved):

Issue #9414, opened 2024-12-02
Extend `Process` effect


magnus-madsen:
We currently have: https://api.flix.dev/Process.html

I think it would be good to replace that single operation with  a more general form:
- [ ] Add operation `execWithCwdAndEnv` (takes something like `cmd: String, args: List[String], cwd: String, env: Map[String, String]`)

The companion module could then contain these helper functions that forward to the operation:
- [ ] Add module function `exec` (current signature)
- [ ] Add module function `execWithCwd` (adds a String for the path)
- [ ] Add module function `execWithEnv` (adds a map for the local env)

The `exec` operation should return a `ProcessHandle`. We probably want it to support operations like on: 

https://docs.oracle.com/javase/8/docs/api/java/lang/Process.html

The question is whether we should have a new algebraic effect for operations on ProcessHandle? Or should we have more operations on Process? Or something third.

magnus-madsen:
@Emaisty Did we complete this or is there more to be done?

magnus-madsen:
@CadeMichael Could you take a look and see if the Process effect can be improved? 

I suggest to write ideas here before programming.

PR #9573, opened 2024-12-20
feat: extend `Process` effect (#9414)

Emaisty:
Related to #9414 .

The process effect has been changed to a more general one. The module's help functions allow you to call exec with fewer arguments. 

Now, the process returns a `ProcessHelper` -- an interface to manipulate with created processes.

### DIFF
--- a/main/src/library/Process.flix
+++ b/main/src/library/Process.flix
@@ -27,6 +27,12 @@ eff Process {
 
 mod Process {
 
+    use IoError.ErrorKind
+    use IoError.IoError
+
+    import java.lang.IllegalArgumentException
+    import java.lang.SecurityException
+    import java.io.IOException
     import java.lang.ProcessBuilder
     import java.lang.Runtime
     import java.io.{File => JFile}
@@ -54,26 +60,31 @@ mod Process {
     ///
     /// In other words, re-interprets the `Process` effect using the `Exec` effect.
     ///
-    pub def handle(f: a -> b \ ef): a -> b \ (ef - Process) + {Exec, IO, Abort} = x ->
+    pub def handle(f: a -> b \ ef): a -> Result[IoError,b] \ (ef - Process) + {Exec, IO} = x ->
         run {
-            f(x)
+            Ok(f(x))
         } with Process {
             def execWithCwdAndEnv(cmd, args, cwd, env, handler) = region rc {
-                let arr = List.toArray(rc, cmd :: args);
-                let pb = new ProcessBuilder(arr);
-                match cwd {
-                    case Some("") => Abort.abort("given an empty directory as an argument")
-                    case Some(cwd) => pb.directory(new JFile(cwd)); ()
-                    case None => ()
-                };
-                foreach(e <- Map.toList(env)) {
-                    pb.environment().put(fst(e), snd(e));
-                    ()
-                };
-                let process = pb.start();
-                handler(ProcessHandle.ProcessHandle({stop = () -> process.destroy(), exitValue = () -> process.exitValue(),  
-                    isAlive = () -> process.isAlive()})
-                    )
+                try {
+                    let arr = List.toArray(rc, cmd :: args);
+                    let pb = new ProcessBuilder(arr);
+                    foreach(e <- Map.toList(env)) {
+                        pb.environment().put(fst(e), snd(e));
+                        ()
+                    };
+                    match cwd {
+                        case Some("") => throw new IllegalArgumentException("given an empty directory as an argument")
+                        case Some(cwd) => pb.directory(new JFile(cwd)); ()
+                        case None => ()
+                    };
+                    let process = pb.start();
+                    handler(ProcessHandle.ProcessHandle({stop = () -> process.destroy(),
+                        exitValue = () -> process.exitValue(), isAlive = () -> process.isAlive()}))
+                } catch {
+                    case ex: IllegalArgumentException  => Err(IoError(ErrorKind.InvalidInput, ex.getMessage()))
+                    case ex: SecurityException         => Err(IoError(ErrorKind.PermissionDenied, ex.getMessage()))
+                    case ex: IOException               => Err(IoError(ErrorKind.Other, ex.getMessage()))
+                }
             }
         }
 
@@ -82,7 +93,7 @@ mod Process {
     ///
     /// In other words, re-interprets the `Process` effect using the `Exec` effect.
     ///
-    pub def runWithIO(f: Unit -> a \ ef): a \ (ef - Process) + {Exec, IO} = handle(f)()
+    pub def runWithIO(f: Unit -> a \ ef): Result[IoError,a] \ (ef - Process) + {Exec, IO} = handle(f)()
 
 }
 
--- a/main/src/library/Process.flix
+++ b/main/src/library/Process.flix
@@ -1,6 +1,5 @@
-
 /*
- *  Copyright 2024 Magnus Madsen
+ *  Copyright 2024 Magnus Madsen, Maksim Gusev
  *
  *  Licensed under the Apache License, Version 2.0 (the "License");
  *  you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


main/src/library/Process.flix
```diff
@@ -1,3 +1,4 @@
+
 /*
  *  Copyright 2024 Magnus Madsen
```
magnus-madsen, line 3: Add your name :)

And also add your name to AUTHORS.md

main/src/library/Process.flix
```diff
@@ -14,22 +15,73 @@
  *  limitations under the License.
  */
 
+///
+/// Handles the created Process
+///
+enum ProcessHandle({stop = Unit -> Unit \ {Exec, IO}}, {exitValue = Unit -> Int32 \ {Exec, IO}},
```
magnus-madsen, line 21: Any reason to have three records instead of one?

main/src/library/Process.flix
```diff
@@ -14,22 +15,73 @@
  *  limitations under the License.
  */
 
+///
+/// Handles the created Process
+///
+enum ProcessHandle({stop = Unit -> Unit \ {Exec, IO}}, {exitValue = Unit -> Int32 \ {Exec, IO}}, 
+    {isAlive = Unit -> Bool \ {Exec, IO}})
+
+mod ProcessHandle {
```
magnus-madsen, line 24: This should probably go in ProcessHandle.flix

main/src/library/Process.flix
```diff
@@ -14,22 +15,73 @@
...
+    pub def exitValue(ph: ProcessHandle) : Int32 \ {IO, Exec} = match ph {
+        case ProcessHandle.ProcessHandle(_, {exitValue}, _) => exitValue()
+    }
+
+    ///
+    ///  Returns status of the process, if it alive or not
+    ///
+    pub def isAlive(ph: ProcessHandle) : Bool \ {IO, Exec} = match ph {
+        case ProcessHandle.ProcessHandle(_, _, {isAlive}) => isAlive()
+    }
```
magnus-madsen, line 45: I would sort the functions alphabetically.

main/src/library/Process.flix
```diff
@@ -14,22 +15,73 @@
...
 
+
+
 mod Process {
 
     import java.lang.ProcessBuilder
     import java.lang.Runtime
+    import java.io.{File => JFile}
+
+
```
magnus-madsen, line 67: extra line break?

main/src/library/Process.flix
```diff
@@ -14,22 +15,73 @@
...
+    ///
+    /// Executes the command `cmd` with the arguments `args`
+    ///
+    pub def exec(cmd: String, args: List[String]): ProcessHandle \ Process =
+        Process.execWithCwdAndEnv(cmd, args, "", Map.empty())
```
magnus-madsen, line 72: I think there should be a difference between setting cwd to the empty string and the actual cwd?

main/src/library/Process.flix
```diff
@@ -40,11 +92,17 @@ mod Process {
         run {
             f(x)
         } with Process {
-            def exec(cmd, args, k) = region rc {
+            def execWithCwdAndEnv(cmd, args, cwd, env, handler) = region rc {
                 let arr = List.toArray(rc, cmd :: args);
                 let pb = new ProcessBuilder(arr);
-                pb.start();
-                k()
+                pb.directory(new JFile(cwd));
```
magnus-madsen, line 98: What happens if cwd is the empty string?

main/src/library/Process.flix
```diff
@@ -40,11 +92,17 @@ mod Process {
...
+                foreach(e <- Map.toList(env)) {
+                    pb.environment().put(fst(e),snd(e));
```
magnus-madsen, line 100: missing space after ,

magnus-madsen (commented):
Looks good! I added some minor comments.

main/src/library/Process.flix
Emaisty: I think it is better to leave it like that because the user can give an empty string as an argument anyway.

If we decide to separate them, it will probably be better to move `execWithEnv` to the `Process` effect itself since `execWithEnv` and `execWithCwdAndEnv` have kinda different behavior. But for me, it seems unnecessary because the user can provide an empty string as a `cwd`, and we will need to address it in the second function, which will make the work of the first one.

magnus-madsen: Why is it correct to set CWD to the empty string?

Emaisty: I c. In that case, it makes sense to separate them and throw an error in the second function. And I think it still makes sense to move `execWithEnv ` to the effect since it will have a different behavior.

Or we can change the type of `cwd` to `Option[String]` and it can solve that issue

magnus-madsen: I think it would make sense to make cwd Option[String]

magnus-madsen:
Remember to mark comments as resolved when they are fixed.

main/src/library/ProcessHandle.flix
```diff
@@ -0,0 +1,45 @@
...
+///
+enum ProcessHandle({exitValue = Unit -> Int32 \ {Exec, IO}, isAlive = Unit -> Bool \ {Exec, IO}, 
+    stop = Unit -> Unit \ {Exec, IO}})
+
+mod ProcessHandle {
+
+    ///
+    /// Returns the exit value of the process
+    ///
+    pub def exitValue(ph: ProcessHandle) : Int32 \ {IO, Exec} = match ph {
```
magnus-madsen, line 28: Extra space before :

magnus-madsen:
If you could take a look at Http and HttpWithResult. 

Do we want to use the same pattern here? Why / why not?

Emaisty:
> If you could take a look at Http and HttpWithResult.
> 
> Do we want to use the same pattern here? Why / why not?

I don't think it is necessary to create `Process` and `ProcessWithResult` because the error occurs not during the execution of the process but when we are handling it. So in my opinion, it makes sense to change the `Process.handle` return value to the `a -> Result[IoError,b]`.

Emaisty:
Another way is to return the `throw` statement to the wrong cwd (`case Some("") => throw ...`) and add a try catch block to the whole function. Since `pb.start()` also can throw exceptions, it can be useful to catch its own errors.

main/src/library/Process.flix
```diff
@@ -17,34 +17,77 @@
...
+            def execWithCwdAndEnv(cmd, args, cwd, env, handler) = region rc {
+                try {
+                    let arr = List.toArray(rc, cmd :: args);
+                    let pb = new ProcessBuilder(arr);
+                    foreach(e <- Map.toList(env)) {
+                        pb.environment().put(fst(e), snd(e));
+                        ()
+                    };
+                    match cwd {
+                        case Some("") => throw new IllegalArgumentException("given an empty directory as an argument")
```
magnus-madsen, line 79: I think we should just drop this case. Its not the just the empty string that is illegal right?

So either you have to check that it is a real path or we let Java deal with it and we catch the appropriate exception and translate it to Result.Err.

main/src/library/Process.flix
```diff
@@ -17,34 +17,77 @@
...
+                        case Some(cwd) => pb.directory(new JFile(cwd)); ()
+                        case None => ()
+                    };
+                    let process = pb.start();
+                    handler(ProcessHandle.ProcessHandle({stop = () -> process.destroy(),
```
magnus-madsen, line 84: Can we format this better with some linebreaks and better alignment?

main/src/library/ProcessHandle.flix
```diff
@@ -0,0 +1,45 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+///
+/// Handles the created Process
+///
+enum ProcessHandle({exitValue = Unit -> Int32 \ {Exec, IO}, isAlive = Unit -> Bool \ {Exec, IO},
```
magnus-madsen, line 20: Can we format this better with some linebreaks and alignment?

main/src/library/ProcessHandle.flix
```diff
@@ -0,0 +1,45 @@
...
+///
+/// Handles the created Process
+///
+enum ProcessHandle({exitValue = Unit -> Int32 \ {Exec, IO}, isAlive = Unit -> Bool \ {Exec, IO}, 
+    stop = Unit -> Unit \ {Exec, IO}})
+
+mod ProcessHandle {
+
+    ///
+    /// Returns the exit value of the process
```
magnus-madsen, line 26: All comments should end with period (this applies everywhere btw).

main/src/library/ProcessHandle.flix
```diff
@@ -0,0 +1,45 @@
...
+    ///
+    pub def exitValue(ph: ProcessHandle): Int32 \ {IO, Exec} = match ph {
```
magnus-madsen, line 28: We use the following information rule: effects are listed from most specific to least, so IO should always be last (this comment applies everywhere)

magnus-madsen (commented):
I have added a few more comments.

They are polish, but that is important for a standard library.

Once these are fixed, we can merge in the PR 👍

magnus-madsen (approved):

magnus-madsen:
@Emaisty Looks good. Just need to add ProcessHandle to the right place in Flix.scala so tests pass.

magnus-madsen:
@Emaisty Congrats on the first PR

PR #9576, opened 2024-12-20
refactor: limit `State.restore` with `isMatchCurrent`

JonathanStarup:
- The old `isMatch` considered the previously read char, so it is now called `isMatchPrev`
- I've added `isMatchCurrent` that just reads from the current position
- `isMatchCurrent` allows us to only use checkpoints for `$`-related tokens

Possibilities
- merge the two matches into `isMatch(offset: Int, keyword: String) ..` and then most use-sites will call `isMatch(-1, ..)`
- let the old function remain `isMatch` (the obvious name was partially for the diff to be understandable)


main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -974,21 +1005,22 @@ object Lexer {
             advance()
           }
           error = Some(TokenKind.Err(LexerError.DoubleUnderscoreInNumber(sourceLocationAtCurrent())))
+        // underscore in tail position
+        case '_' =>
+          advance()
+          return TokenKind.Err(LexerError.TrailingUnderscoreInNumber(sourceLocationAtCurrent()))
```
JonathanStarup, line 1011: I moved this so the code of the two places we parse numbers are written in the same way (it also makes the following comment tru-er)

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -974,21 +1005,22 @@ object Lexer {
...
-            case '_' => TokenKind.Err(LexerError.TrailingUnderscoreInNumber(sourceLocationAtCurrent()))
-            case _ if isMatch("f32") => error.getOrElse(TokenKind.LiteralFloat32)
-            case _ if isMatch("f64") => error.getOrElse(TokenKind.LiteralFloat64)
-            case _ if isMatch("i8") => error.getOrElse(TokenKind.LiteralInt8)
-            case _ if isMatch("i16") => error.getOrElse(TokenKind.LiteralInt16)
-            case _ if isMatch("i32") => error.getOrElse(TokenKind.LiteralInt32)
-            case _ if isMatch("i64") => error.getOrElse(TokenKind.LiteralInt64)
-            case _ if isMatch("ii") => error.getOrElse(TokenKind.LiteralBigInt)
-            case _ if isMatch("ff") => error.getOrElse(TokenKind.LiteralBigDecimal)
+        case c =>
```
JonathanStarup, line 1013: c here is the result of a call to `peek()`

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -1038,21 +1070,18 @@ object Lexer {
...
-          case _ if isMatch("i8") => error.getOrElse(TokenKind.LiteralInt8)
-          case _ if isMatch("i16") => error.getOrElse(TokenKind.LiteralInt16)
-          case _ if isMatch("i32") => error.getOrElse(TokenKind.LiteralInt32)
-          case _ if isMatch("i64") => error.getOrElse(TokenKind.LiteralInt64)
-          case _ if isMatch("ii") => error.getOrElse(TokenKind.LiteralBigInt)
-          case _ if isMatch("ff") => error.getOrElse(TokenKind.LiteralBigDecimal)
-          case _ =>
-            s.restore(checkpoint)
-            error.getOrElse(TokenKind.LiteralInt32)
+        case c =>
```
JonathanStarup, line 1073: c here is the result of a call to `peek()`

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -594,13 +594,44 @@ object Lexer {
...
+    while (matches && offset < keyword.length) {
+      if (s.src.data(start + offset) != keyword(offset)) {
+        matches = false
+      } else {
+        offset += 1
+      }
+    }
+
+    if (matches) {
+      for (_ <- 0 until keyword.length) {
```
mlutze, line 620: I guess there's nothing like `advance(n)`?

mlutze (approved):

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
JonathanStarup: no - this code is also just copied from the other match, this the -1 adjusted

magnus-madsen (approved):

PR #9584, opened 2024-12-20
fix: don't crash on incomplete unary pattern in Weeder

mlutze:
fixes https://github.com/flix/flix/issues/9583


magnus-madsen (approved):

mlutze:
"Party's over. Let's merge!"

Issue #9561, opened 2024-12-18
Extending `RenameProvider`


LoZander:
Currently, the `RenameProvider` is very limited due limitations on `SymUse`. This basically comes down to `SymUse`s including the entire qualified name in the `SourceLocation`.

LoZander:
For full rename support, it would require some fairly deep changes I think. We'd need to know what ever part of the qualified name is a reference to. Maybe make `NName` into a list of `SymUse` or something. At least at some point, we'd need to create a list of `SymUse` or something like it for each part of the `NName`. This way we'd be able to rename not just things that might appear at the end of a qualified name, but also things that may appear at the start or in the middle of one, like modules, enums, effects, etc.

I think this change might be too big right now though. So I suggest a smaller change. Let's just change the `SouceLocations`s of `SymUse`s to only include the actual symbol and not the entire qualified name. This way we'd get the ability to rename things that may appear at the every end of a qualified name. So we'd get
- Functions
- Enum cases
- Signatures
- Effect ops
- Structs
- Type aliases

We'd still be missing
- Modules
- Enums
- Traits
- Effects

The most important thing, in my opinion, is that we get the ability to rename functions, as I reckon these are what a user would want to rename most often.

Any thoughts on this?

magnus-madsen:
DefnSym is harder to change than I think e.g. EffSym hence the suggestion to start there.

magnus-madsen:
I would go for a minimal PR that extends EffSym with a QName and uses that in the rename provider.

LoZander:
But wouldn't we get a lot by just changing the `SourceLocations` like I suggest?

magnus-madsen:
> But wouldn't we get a lot by just changing the `SourceLocations` like I suggest?

How is that not a change with huge implications? Why is it "the right thing to do"?

LoZander:
Well, my thought was that the "symbol use" is not really the entire qualified name. The symbol is only a part of it. But yeah, changing the sourcelocations for `SymUses` is not a small change. Wouldn't using the QName in the provider to manually adjust sourcelocations be a bit of a hack? Wouldn't that fall outside the responsibility of the provider?

LoZander:
[...]

PR #9586, opened 2024-12-21
feat: extending EffectSymUse with QName

LoZander:
This PR extends `EffectSymUse`s with the `QName`.

Related to #9561.


main/src/ca/uwaterloo/flix/language/ast/shared/SymUse.scala
```diff
@@ -37,7 +38,7 @@ object SymUse {
   /**
     * Represents a use of an effect sym.
     */
-  case class EffectSymUse(sym: Symbol.EffectSym, loc: SourceLocation)
+  case class EffectSymUse(sym: Symbol.EffectSym, qname: QName, loc: SourceLocation)
```
magnus-madsen, line 41: I think - given the earlier confusion - each parameter should be documented.

magnus-madsen: Also consider whether loc is redundant

magnus-madsen:
@LoZander Notice how this PR required only 12 loc change, despite EffUseSym and QName being used _a lot more_ than the highlighter...

LoZander:
> @LoZander Notice how this PR required only 12 loc change, despite EffUseSym and QName being used _a lot more_ than the highlighter...

Yeah, I thought that was odd. But I assumed an issue with regards to this would cause a compile error. I reckon most uses probably just access the members via dot-notation and that effect sym uses just aren't created in many spaces. This might explain why no other uses of it seemingly break. I'll give it another look over to make sure there aren't any other uses of it that have silently broken in (maybe runtime problems?). Do you think that might be the case?

main/src/ca/uwaterloo/flix/language/ast/shared/SymUse.scala
LoZander: Both are good points. For the second point, if we do remove `loc` maybe it would be useful to add a `loc` function that just accesses `qname.loc`, just to keep the pattern of AST nodes having their SourceLocation accessible through `.loc`?

magnus-madsen:
> > @LoZander Notice how this PR required only 12 loc change, despite EffUseSym and QName being used _a lot more_ than the highlighter...
> 
> Yeah, I thought that was odd. But I assumed an issue with regards to this would cause a compile error. I reckon most uses probably just access the members via dot-notation and that effect sym uses just aren't created in many spaces. This might explain why no other uses of it seemingly break. I'll give it another look over to make sure there aren't any other uses of it that have silently broken in (maybe runtime problems?). Do you think that might be the case?

My point was the opposite: if one has the right architecture then changes can be small.

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
```diff
@@ -15,6 +15,7 @@
  */
 package ca.uwaterloo.flix.api.lsp
 
+import ca.uwaterloo.flix.language.ast.Name.QName
```
magnus-madsen, line 18: Why this?

main/src/ca/uwaterloo/flix/language/ast/shared/SymUse.scala
```diff
@@ -36,8 +37,23 @@ object SymUse {
...
     */
-  case class EffectSymUse(sym: Symbol.EffectSym, loc: SourceLocation)
+  case class EffectSymUse(sym: Symbol.EffectSym, qname: QName) {
+    /**
+      * Returns the [[SourceLocation]] of this [[EffectSymUse]]. Note that this is
+      * the same [[SourceLocation]] as that of the [[EffectSymUse]]s `qname`.
+      *
+      * @return The [[SourceLocation]] of the [[EffectSymUse]].
+      */
+    def loc: SourceLocation = qname.loc
```
magnus-madsen, line 55: Is this a good idea?

A use essentially has two (or more) associated locations, right?

Why is loc equal to qname.loc? It might as well be sym.loc?

LoZander:
> > > @LoZander Notice how this PR required only 12 loc change, despite EffUseSym and QName being used _a lot more_ than the highlighter...
> > 
> > 
> > Yeah, I thought that was odd. But I assumed an issue with regards to this would cause a compile error. I reckon most uses probably just access the members via dot-notation and that effect sym uses just aren't created in many spaces. This might explain why no other uses of it seemingly break. I'll give it another look over to make sure there aren't any other uses of it that have silently broken in (maybe runtime problems?). Do you think that might be the case?
> 
> My point was the opposite: if one has the right architecture then changes can be small.

Oh, like that. Yeah. I see what you mean. Yeah, that is nice. Stuff like what I was doing with #9567 shouldn't be necessary. I get so caught up in solving the issue given my understanding of it that I sometimes fail to question my assumptions and the architecture. I have to get better at taking a step back

main/src/ca/uwaterloo/flix/language/ast/shared/SymUse.scala
LoZander: I don't think it would make sense for `sym.loc` to be the SourceLocation of the SymUse, since that is the SourceLocation of the declaration site of the symbol, but you do have a point. 

`qname` contains multiple SourceLocations that might be considered the SourceLocation of the SymUse. For instance, should it be the SourceLocation of the entire qualified name or just that of the symbols occurrence in that qualified name?

So it might be better to leave no `.loc` at all, and let whoever uses `SymUse` decide what `SourceLocation` they care about. Maybe rename `qname` to `use`, to make it clear that that is the qname of the use of the symbol.

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
LoZander: Just a mistake. An import I forgot to remove

main/src/ca/uwaterloo/flix/language/ast/shared/SymUse.scala
LoZander: The def of `loc` that I gave here gives the exact same `SourceLocaiton` for the SymUse as before this PR

LoZander: Do you mean that it doesn't make much sense to choose that specific SourceLocation now that we have the entire QName?

magnus-madsen: > So it might be better to leave no .loc at all, and let whoever uses SymUse decide what SourceLocation they care about. 

Yes. Then the consumer has to pattern match and choose the specific location they want.

LoZander:
I've removed the `.loc` from `EffectSymUse` as you suggested, and to leave uses of `EffectSymUse` unchanged, I've changed occurrences of `symUse.loc` with `symUse.qname.loc`

LoZander:
If it all looks okay, I think this PR is done. I'll move onto adding `QName` to `OpSymUse` as this is also necessary for getting effect renaming in #9561

LoZander:
oops, forgot about copyright headers. I'll quickly update those

LoZander:
Most of the changes I made to most files were trivial, but changes nonetheless, so I assumed I should add my name in the copyright header. There were a couple of files in there that I made changes to that had no copyright header at all. I just left these alone (didn't want to add a copyright header that would be missing the names of the people who made the file).

Anyways. If you're happy with this, I think this PR is done :)

LoZander:
Or, maybe the `qname` member should be renamed to make it clear that it's the qualified name of the use of the symbol. Maybe renaming it to `use` or something like that would make sense?

main/src/ca/uwaterloo/flix/api/lsp/provider/FindReferencesProvider.scala
```diff
@@ -167,7 +168,7 @@ object FindReferencesProvider {
     case SymUse.AssocTypeSymUse(_, loc) => loc.isReal
     case SymUse.CaseSymUse(_, loc) => loc.isReal
     case SymUse.DefSymUse(_, loc) => loc.isReal
-    case SymUse.EffectSymUse(_, loc) => loc.isReal
+    case SymUse.EffectSymUse(_, QName(_, _, loc)) => loc.isReal
```
magnus-madsen, line 171: You use a mixture of styles, in some places its a deeper pattern match and in other places you write `qname.loc`. I think have a slight prefer for the latter, i.e. `qname.loc` instead of the deep match.

magnus-madsen:
Fix the style issue and then we can merge this.

LoZander:
> Fix the style issue and then we can merge this.

Should all be fixed now :)

LoZander:
I have also fixed it in #9602 :)

main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
magnus-madsen: So will you remove it?

main/src/ca/uwaterloo/flix/api/lsp/provider/GotoProvider.scala
```diff
@@ -19,6 +19,7 @@ package ca.uwaterloo.flix.api.lsp.provider
 import ca.uwaterloo.flix.api.lsp.*
 import ca.uwaterloo.flix.api.lsp.acceptors.InsideAcceptor
 import ca.uwaterloo.flix.api.lsp.consumers.StackConsumer
+import ca.uwaterloo.flix.language.ast.Name.QName
```
magnus-madsen, line 22: Ditto

main/src/ca/uwaterloo/flix/api/lsp/provider/HighlightProvider.scala
```diff
@@ -19,6 +19,7 @@ package ca.uwaterloo.flix.api.lsp.provider
 import ca.uwaterloo.flix.api.lsp.acceptors.{FileAcceptor, InsideAcceptor}
 import ca.uwaterloo.flix.api.lsp.consumers.StackConsumer
 import ca.uwaterloo.flix.api.lsp.{Acceptor, Consumer, DocumentHighlight, DocumentHighlightKind, Position, Range, ResponseStatus, Visitor}
+import ca.uwaterloo.flix.language.ast.Name.QName
```
magnus-madsen, line 22: Ditto (and in more files- check the diff)

magnus-madsen (approved):

PR #9588, opened 2024-12-21
feat: rework `concurrency-and-parallelism` examples

magnus-madsen:


magnus-madsen:
We want to fuzz large programs, not small. Hence i removed the one test.

mlutze (approved):

PR #9598, opened 2024-12-23
refactor: separating the finding of occurrences and creation of highlights in HighlightProvider

LoZander:
This PR aims to give the `HighlightProvider` better separation of concerns by first finding all occurrences of the symbol / label in question and then creating a highlights from these, rather than creating the highlights immediately upon finding the occurrences using the `HighlightBuilder`.

This PR then changes the `highlightX` functions to `getXOccurs` and adds the `mkHighlights` function. Additionally we remove the `HighlightBuilder`.

The result is a provider with a more flexsible structure that more closely resembles that of `FindOccurrencesProvider` and `RenameProvider`.

The added flexisibility will also benefit the work being done at #4276


magnus-madsen:
Sounds good. I will have a look later.

magnus-madsen (approved):

magnus-madsen:
Thanks!

PR #9599, opened 2024-12-23
feat: effect highlighting for do

LoZander:
This PR adds effect highlighting for do expressions. This means that when the cursor is on an effect in a function signature, any do expressions that produce said effect will be highlighted.

Related to #4276.


LoZander:
Currently, this is not the prettiest impl, because it doesn't work well with the `HighlightBuilder`. When #9598 goes through, this can be refactored :)

LoZander:
The force push was to rollback some changes that were meant for another PR

magnus-madsen:
Screenshot?

LoZander:
image

magnus-madsen (approved):

PR #9603, opened 2024-12-28
chore: rework `lsp` command

magnus-madsen:
We now have just `lsp` and `lsp-vscode` -- as planned.


magnus-madsen:
The `lsp` command can evolve as we actually implement lsp4j.

mlutze (approved):

Issue #9512, opened 2024-12-14
Annotate all methods in `java.lang.System` with precise effects


magnus-madsen:
Carefully go through all methods in `java.lang.System` and annotate them with primitive effects.

Some of this work has already started.

jaschdoc:
If a method only has `IO`, do we then need to explicitly annotate it? It is automatically given the `IO` effect in `PrimitiveEffects`, no?

magnus-madsen:
> If a method only has `IO`, do we then need to explicitly annotate it? It is automatically given the `IO` effect in `PrimitiveEffects`, no?

Then I think we can leave it unannotated, except if its class or package is annotated with more than `IO`, right?

jaschdoc:
Close this?

Issue #9499, opened 2024-12-08
Split Env?


magnus-madsen:
We need to split Env to separate reading environment variables from mundane information such as line separator.

magnus-madsen:
@jaschdoc We need to discuss this w.r.t. project.

jaschdoc:
There is also a `Sys` and `JVM` distinction we can make.

magnus-madsen:
> There is also a `Sys` and `JVM` distinction we can make.

What's the distinction? In any case, I don't want any specific to the JVM- so it be some more meaningful distinction.

jaschdoc:
Close this?

PR #9621, opened 2025-01-05
feat: annotate methods in `java.lang.System`

jaschdoc:
I think we should discuss some of these annotations more.

I understand `Env` as **read-only** from system properties, but `Sys` as writing. That means that if you have `Sys` you also have `Env` when it interacts with the system properties. Maybe this is not entirely correct.

Note, that the effects file is not sorted alphabetically. I will do that in a subsequent PR. Right now they appear in the order mentioned in the java doc.

Closes https://github.com/flix/flix/issues/9512

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/Bootstrap.scala
+++ b/main/src/ca/uwaterloo/flix/api/Bootstrap.scala
@@ -316,9 +316,14 @@ object Bootstrap {
 
   /**
     * Returns all files in the given path `p` ending with .`ext`.
+    *
+    * @param p   the path from which files a considered.
+    * @param ext the file extension to match. Must not begin with `.`
     */
-  private def getAllFilesWithExt(p: Path, ext: String): List[Path] =
+  private def getAllFilesWithExt(p: Path, ext: String): List[Path] = {
+    require(!ext.startsWith("."), "The file extension must not start with '.' -- This is handled by 'getAllFilesWithExt'.")
     getAllFiles(p).filter(p => p.getFileName.toString.endsWith(s".$ext"))
+  }
 
   /**
     * Returns all files in the given path `p`.


jaschdoc:
I think this also calls for the discussion in https://github.com/flix/flix/issues/9499

magnus-madsen:
CI failures

magnus-madsen:
> I understand `Env` as **read-only** from system properties, but `Sys` as writing. 

I think to a first approximation that is OK. But env is not just properties, it could also be e.g. number of CPUs etc. (which is not an environmental variable).

> That means that if you have `Sys` you also have `Env` when it interacts with the system properties. Maybe this is not entirely correct.

Yes, but there is no subeffecting. So you have to put both. 

> Note, that the effects file is not sorted alphabetically. I will do that in a subsequent PR. Right now they appear in the order mentioned in the java doc.

OK

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
```diff
@@ -8,9 +8,22 @@
     "java.lang.Runtime::maxMemory": "Env, IO",
     "java.lang.Runtime::version": "Env",
     "java.lang.System::console": "IO",
+    "java.lang.System::inheritedChannel": "Sys, IO",
+    "java.lang.System::setSecurityManager": "Env, Sys, IO",
+    "java.lang.System::getSecurityManager": "Env, IO",
```
magnus-madsen, line 13: This is probably also Sys. I mean no harm in putting to broad an effect.

magnus-madsen:
In the case of Runtime and System __it might be worth annotating every method explicitly__ no matter what we decide to do in general. The reasoning being that these classes are very special and we don't want to miss anything.

jaschdoc:
> I think to a first approximation that is OK. But env is not just properties, it could also be e.g. number of CPUs etc. (which is not an environmental variable).

Ah yeah apologies for the confusion. The java documentation treats env vars as system properties along with other things.

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
```diff
@@ -8,9 +8,22 @@
...
     "java.lang.System::currentTimeMillis": "IO",
-    "java.lang.System::getenv": "Env, IO",
     "java.lang.System::nanoTime": "IO",
+    "java.lang.System::getProperties": "Env, IO",
+    "java.lang.System::lineSeparator": "Env, IO",
```
jaschdoc, line 17: Did we agree on `lineSeparator` being only `IO` since it is deterministic once the platform has been determined?

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
```diff
@@ -7,10 +7,31 @@
...
+    "java.lang.System::getProperties": "Env, Sys, IO",
+    "java.lang.System::lineSeparator": "Env, IO",
+    "java.lang.System::setProperties": "Env, Sys, IO",
+    "java.lang.System::getProperty": "Env, Sys, IO",
+    "java.lang.System::setProperty": "Env, Sys, IO",
+    "java.lang.System::clearProperty": "Env, Sys, IO",
+    "java.lang.System::getenv": "Env, IO",
+    "java.lang.System::getLogger": "Env, IO",
+    "java.lang.System::exit": "Sys, IO",
+    "java.lang.System::gc": "Sys, IO",
```
jaschdoc, line 30: `gc` just runs garbage collection but is otherwise safe, so should it just have `IO` even though it interacts (safely) with the JVM?

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
```diff
@@ -7,10 +7,31 @@
...
+    "java.lang.System::arrayCopy": "IO",
+    "java.lang.System::identityHashCode": "IO",
+    "java.lang.System::getProperties": "Env, Sys, IO",
+    "java.lang.System::lineSeparator": "Env, IO",
+    "java.lang.System::setProperties": "Env, Sys, IO",
+    "java.lang.System::getProperty": "Env, Sys, IO",
+    "java.lang.System::setProperty": "Env, Sys, IO",
+    "java.lang.System::clearProperty": "Env, Sys, IO",
+    "java.lang.System::getenv": "Env, IO",
+    "java.lang.System::getLogger": "Env, IO",
```
jaschdoc, line 28: I'm not quite sure what the logger is used for or what mental model to use, so I'm unsure if `Env` is correct. Help! :)

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
jaschdoc: Or even `Pure?`

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
```diff
@@ -17,11 +17,31 @@
...
     "java.lang.Runtime::version": "Env",
-
+    "java.lang.System::arrayCopy": "IO",
+    "java.lang.System::clearProperty": "Env, Sys, IO",
     "java.lang.System::console": "IO",
     "java.lang.System::currentTimeMillis": "IO",
+    "java.lang.System::exit": "Sys, IO",
+    "java.lang.System::gc": "IO",
     "java.lang.System::getenv": "Env, IO",
+    "java.lang.System::getLogger": "Env, Sys, IO",
```
jaschdoc, line 27: Apparently it may use the system class loader so it has `Sys`

jaschdoc: Not sure about `Env`. Thoughts?

jaschdoc: Does not actually seem like `Env` at all tbf

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
```diff
@@ -7,10 +7,31 @@
...
+    "java.lang.System::lineSeparator": "Env, IO",
+    "java.lang.System::setProperties": "Env, Sys, IO",
+    "java.lang.System::getProperty": "Env, Sys, IO",
+    "java.lang.System::setProperty": "Env, Sys, IO",
+    "java.lang.System::clearProperty": "Env, Sys, IO",
+    "java.lang.System::getenv": "Env, IO",
+    "java.lang.System::getLogger": "Env, IO",
+    "java.lang.System::exit": "Sys, IO",
+    "java.lang.System::gc": "Sys, IO",
+    "java.lang.System::runFinalization": "Sys, IO",
```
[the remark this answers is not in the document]
magnus-madsen: I am just saying I think giving a hint to the GC is probably completely safe

magnus-madsen:
@mlutze @JonathanStarup Do we morally agree with these breakages?

mlutze:
> @mlutze @JonathanStarup Do we morally agree with these breakages?

It has my blessing.

magnus-madsen:
I am ambivalent about whether asking for platform newline or CPUs should really be a primitive effect. Hmm. But I suppose it HAS to be some effect.

mlutze:
> I am ambivalent about whether asking for platform newline or CPUs should really be a primitive effect. Hmm. But I suppose it HAS to be some effect.

I am not ambivalent. It is definitely an effect in my book.

JonathanStarup:
Nice if pure code was somewhat immue to "it works on my machine"

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
```diff
@@ -27,10 +27,31 @@
...
+    "java.lang.System::exit": "Sys, IO",
+    "java.lang.System::gc": "IO",
     "java.lang.System::getenv": "Env, IO",
+    "java.lang.System::getLogger": "Sys, IO",
+    "java.lang.System::getProperties": "Env, Sys, IO",
+    "java.lang.System::getProperty": "Env, Sys, IO",
+    "java.lang.System::setSecurityManager": "Env, Sys, IO",
+    "java.lang.System::identityHashCode": "IO",
+    "java.lang.System::inheritedChannel": "Sys, IO",
+    "java.lang.System::lineSeparator": "IO",
```
magnus-madsen: env?

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
```diff
@@ -27,10 +27,31 @@
...
+    "java.lang.System::clearProperty": "Env, Sys, IO",
     "java.lang.System::console": "IO",
     "java.lang.System::currentTimeMillis": "IO",
+    "java.lang.System::exit": "Sys, IO",
+    "java.lang.System::gc": "IO",
     "java.lang.System::getenv": "Env, IO",
+    "java.lang.System::getLogger": "Sys, IO",
+    "java.lang.System::getProperties": "Env, Sys, IO",
+    "java.lang.System::getProperty": "Env, Sys, IO",
+    "java.lang.System::setSecurityManager": "Env, Sys, IO",
```
magnus-madsen: Not sure this is Env

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
```diff
@@ -27,10 +27,31 @@
...
+    "java.lang.System::identityHashCode": "IO",
+    "java.lang.System::inheritedChannel": "Sys, IO",
+    "java.lang.System::lineSeparator": "IO",
+    "java.lang.System::load": "Exec, Sys, IO",
+    "java.lang.System::loadLibrary": "Exec, Sys, IO",
+    "java.lang.System::mapLibraryName": "Env, IO",
```
magnus-madsen, line 67: What does this even do?

JonathanStarup:
> I understand Env as read-only from system properties, but Sys as writing. That means that if you have Sys you also have Env when it interacts with the system properties.

> Not sure this is Env

> What does this even do?

> Does not actually seem like Env at all tbf

Maybe its impractical, but it feels like these annotations should have "doc comments" on them

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
jaschdoc: It checks runtime permissions, so I think that's what I meant by Env, but I agree it shouldn't be Env

https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/System.html#setSecurityManager(java.lang.SecurityManager)

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
jaschdoc: > Maps a library name into a platform-specific string representing a native library.

jaschdoc: https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/System.html#mapLibraryName(java.lang.String)

jaschdoc: Maybe more like `Platform`?

jaschdoc:
Should we do another review of this and see if we can merge it?

jaschdoc:
> Maybe its impractical, but it feels like these annotations should have "doc comments" on them

We could map each function to an object with the shape `{ "effects": effect-str, "comment": str }`

magnus-madsen:
No longer relevant.

PR #9622, opened 2025-01-06
refactor: make dependencies recursive

### LINKED ISSUE #9550: Rewrite `Dependencies` phase to use normal recursive functions
- [ ] Introduce a shared context:

```
private case class SharedContext(errors: ConcurrentMap[(Input, Input), Unit])
```

(We want a multi map, but there is no such concurrent map, so we store pairs of key values).

- [ ] Rewrite the phase to use ordinary recursive functions, e.g. visitDef, visitExp, and so on.

stormckey:
fixes #9550

I introduced a sealed trait `SymUse` to make all the SymUses belong to a same trait, making the implementation more clear.

We could continue to use changeset and change the return type from unit to the untouched input in the following PR.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
@@ -60,7 +60,7 @@ object Dependencies {
     sctx.deps.put((src.sp1.source.input, dst.sp1.source.input), ())
   }
 
-  private def visitDef(defn: TypedAst.Def)(implicit sctx: SharedContext): Unit = {
+  private def visitDef(defn: TypedAst.Def)(implicit sctx: SharedContext): Unit =  {
     visitExp(defn.exp)
     visitSpec(defn.spec)
   }
@@ -125,24 +125,24 @@ object Dependencies {
 
     case Expr.ApplyDef(symUse, exps, itpe, tpe, eff, _) =>
       visitSymUse(symUse)
+      exps.foreach(visitExp)
       visitType(itpe)
       visitType(tpe)
       visitType(eff)
-      exps.foreach(visitExp)
 
     case Expr.ApplyLocalDef(symUse, exps, arrowTpe, tpe, eff, _) =>
       visitSymUse(symUse)
+      exps.foreach(visitExp)
       visitType(arrowTpe)
       visitType(tpe)
       visitType(eff)
-      exps.foreach(visitExp)
 
     case Expr.ApplySig(symUse, exps, itpe, tpe, eff, _) =>
       visitSymUse(symUse)
+      exps.foreach(visitExp)
       visitType(itpe)
       visitType(tpe)
       visitType(eff)
-      exps.foreach(visitExp)
 
     case Expr.Unary(_, exp, tpe, eff, _) =>
       visitExp(exp)
@@ -164,11 +164,11 @@ object Dependencies {
 
     case Expr.LocalDef(bnd, fparams, exp1, exp2, tpe, eff, _) =>
       visitBinder(bnd)
+      fparams.foreach(visitFormalParam)
       visitExp(exp1)
       visitExp(exp2)
       visitType(tpe)
       visitType(eff)
-      fparams.foreach(visitFormalParam)
 
     case Expr.Region(tpe, _) =>
       visitType(tpe)
@@ -198,38 +198,38 @@ object Dependencies {
 
     case Expr.Match(exp, rules, tpe, eff, _) =>
       visitExp(exp)
+      rules.foreach(visitMatchRule)
       visitType(tpe)
       visitType(eff)
-      rules.foreach(visitMatchRule)
 
     case Expr.TypeMatch(exp, rules, tpe, eff, _) =>
       visitExp(exp)
+      rules.foreach(visitTypeMatchRule)
       visitType(tpe)
       visitType(eff)
-      rules.foreach(visitTypeMatchRule)
 
     case Expr.RestrictableChoose(_, exp, rules, tpe, eff, _) =>
       visitExp(exp)
+      rules.foreach(visitRestrictableChooseRule)
       visitType(tpe)
       visitType(eff)
-      rules.foreach(visitRestrictableChooseRule)
 
     case Expr.Tag(sym, exps, tpe, eff, _) =>
       visitSymUse(sym)
+      exps.foreach(visitExp)
       visitType(tpe)
       visitType(eff)
-      exps.foreach(visitExp)
 
     case Expr.RestrictableTag(sym, exps, tpe, eff, _) =>
       visitSymUse(sym)
+      exps.foreach(visitExp)
       visitType(tpe)
       visitType(eff)
-      exps.foreach(visitExp)
 
     case Expr.Tuple(exps, tpe, eff, _) =>
+      exps.foreach(visitExp)
       visitType(tpe)
       visitType(eff)
-      exps.foreach(visitExp)
 
     case Expr.RecordEmpty(tpe, _) =>
       visitType(tpe)
@@ -251,10 +251,10 @@ object Dependencies {
       visitType(eff)
 
     case Expr.ArrayLit(exps, exp, tpe, eff, _) =>
+      exps.foreach(visitExp)
       visitExp(exp)
       visitType(tpe)
       visitType(eff)
-      exps.foreach(visitExp)
 
     case Expr.ArrayNew(exp1, exp2, exp3, tpe, eff, _) =>
       visitExp(exp1)
@@ -280,13 +280,13 @@ object Dependencies {
       visitType(eff)
 
     case Expr.StructNew(_, fields, region, tpe, eff, _) =>
-      visitExp(region)
-      visitType(tpe)
-      visitType(eff)
       fields.foreach{ field =>
         visitSymUse(field._1)
         visitExp(field._2)
       }
+      visitExp(region)
+      visitType(tpe)
+      visitType(eff)
 
     case Expr.StructGet(exp, sym, tpe, eff, _) =>
       visitExp(exp)
@@ -302,9 +302,9 @@ object Dependencies {
       visitType(eff)
 
     case Expr.VectorLit(exps, tpe, eff, _) =>
+      exps.foreach(visitExp)
       visitType(tpe)
       visitType(eff)
-      exps.foreach(visitExp)
 
     case Expr.VectorLoad(exp1, exp2, tpe, eff, _) =>
       visitExp(exp1)
@@ -330,10 +330,10 @@ object Dependencies {
 
     case Expr.UncheckedCast(exp, declaredType, declaredEff, tpe, eff, _) =>
       visitExp(exp)
-      visitType(tpe)
-      visitType(eff)
       declaredType.toList.foreach(visitType)
       declaredEff.toList.foreach(visitType)
+      visitType(tpe)
+      visitType(eff)
 
     case Expr.Unsafe(exp, runEff, tpe, eff, _) =>
       visitExp(exp)
@@ -349,9 +349,9 @@ object Dependencies {
 
     case Expr.TryCatch(exp, rules, tpe, eff, _) =>
       visitExp(exp)
+      rules.foreach(visitCatchRule)
       visitType(tpe)
       visitType(eff)
-      rules.foreach(visitCatchRule)
 
     case Expr.Throw(exp, tpe, eff, _) =>
       visitExp(exp)
@@ -361,31 +361,31 @@ object Dependencies {
     case Expr.TryWith(exp, effUse, rules, tpe, eff, _) =>
       visitExp(exp)
       visitSymUse(effUse)
+      rules.foreach(visitHandlerRule)
       visitType(tpe)
       visitType(eff)
-      rules.foreach(visitHandlerRule)
 
     case Expr.Do(op, exps, tpe, eff, _) =>
       visitSymUse(op)
+      exps.foreach(visitExp)
       visitType(tpe)
       visitType(eff)
-      exps.foreach(visitExp)
 
     case Expr.InvokeConstructor(_, exps, tpe, eff, _) =>
+      exps.foreach(visitExp)
       visitType(tpe)
       visitType(eff)
-      exps.foreach(visitExp)
 
     case Expr.InvokeMethod(_, exp, exps, tpe, eff, _) =>
       visitExp(exp)
+      exps.foreach(visitExp)
       visitType(tpe)
       visitType(eff)
-      exps.foreach(visitExp)
 
     case Expr.InvokeStaticMethod(_, exps, tpe, eff, _) =>
+      exps.foreach(visitExp)
       visitType(tpe)
       visitType(eff)
-      exps.foreach(visitExp)
 
     case Expr.GetField(_, exp, tpe, eff, _) =>
       visitExp(exp)
@@ -424,10 +424,10 @@ object Dependencies {
       visitType(eff)
 
     case Expr.SelectChannel(rules, default, tpe, eff, _) =>
-      visitType(tpe)
-      visitType(eff)
       rules.foreach(visitSelectChannelRule)
       default.toList.foreach(visitExp)
+      visitType(tpe)
+      visitType(eff)
 
     case Expr.Spawn(exp1, exp2, tpe, eff, _) =>
       visitExp(exp1)
@@ -436,10 +436,10 @@ object Dependencies {
       visitType(eff)
 
     case Expr.ParYield(frags, exp, tpe, eff, _) =>
+      frags.foreach(visitParYieldFragment)
       visitExp(exp)
       visitType(tpe)
       visitType(eff)
-      frags.foreach(visitParYieldFragment)
 
     case Expr.Lazy(exp, tpe, _) =>
       visitExp(exp)
@@ -451,14 +451,14 @@ object Dependencies {
       visitType(eff)
 
     case Expr.FixpointConstraintSet(cs, tpe, _) =>
-      visitType(tpe)
       cs.foreach(visitConstrait)
+      visitType(tpe)
 
     case Expr.FixpointLambda(pparams, exp, tpe, eff, _) =>
+      pparams.foreach(visitPParam)
       visitExp(exp)
       visitType(tpe)
       visitType(eff)
-      pparams.foreach(visitPParam)
 
     case Expr.FixpointMerge(exp1, exp2, tpe, eff, _) =>
       visitExp(exp1)
@@ -577,15 +577,15 @@ object Dependencies {
       visitType(tpe)
     case Pattern.Tag(sym, pats, tpe, _) =>
       visitSymUse(sym)
-      visitType(tpe)
       pats.foreach(visitPattern)
+      visitType(tpe)
     case Pattern.Tuple(pats, tpe, _) =>
-      visitType(tpe)
       pats.foreach(visitPattern)
+      visitType(tpe)
     case Pattern.Record(pats, pat, tpe, _) =>
+      pats.foreach(visitRecordLabelPattern)
       visitPattern(pat)
       visitType(tpe)
-      pats.foreach(visitRecordLabelPattern)
     case pat =>
       visitType(pat.tpe)
   }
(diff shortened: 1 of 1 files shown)


main/src/ca/uwaterloo/flix/language/ast/shared/SymUse.scala
```diff
@@ -20,21 +20,21 @@ import ca.uwaterloo.flix.language.ast.Name.{Ident, QName}
 import ca.uwaterloo.flix.language.ast.{SourceLocation, Symbol}
```
mlutze, line 20: Add a license to this file (see other files for examples)

main/src/ca/uwaterloo/flix/language/ast/shared/SymUse.scala
```diff
@@ -20,21 +20,21 @@ import ca.uwaterloo.flix.language.ast.Name.{Ident, QName}
...
 
 object SymUse {
-
+  sealed trait SymUse
```
mlutze, line 23: Move this to the top level of this file.

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
```diff
@@ -1,88 +1,602 @@
 package ca.uwaterloo.flix.language.phase
```
mlutze, line 1: license

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
```diff
@@ -1,88 +1,617 @@
...
+import ca.uwaterloo.flix.util.ParOps
 import ca.uwaterloo.flix.util.collection.MultiMap
 
-object Dependencies {
+import java.util.concurrent.{ConcurrentHashMap, ConcurrentMap}
 
-  /** Checks the safety and well-formedness of `root`. */
+object Dependencies {
+  /**
+    * Updates the dependency graph of the given `root`.
```
JonathanStarup, line 32: i assumed it used old data when it says `Updates`, maybe `Computes` is better here

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
```diff
@@ -1,88 +1,617 @@
...
+    * We visit all the definitions, effects, enums, instances, structs, traits, and type aliases to find the dependencies.
```
JonathanStarup, line 33: Since the code below is so evident, this comment feels redundant and somebody will just forget to update it.

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
```diff
@@ -1,88 +1,617 @@
...
-    // Moreover, we should traverse the AST in parallel and using changeSet.
+    var deps = MultiMap.empty[Input, Input]
+    sctx.deps.forEach((k, _) => deps = deps + k)
+    val dg = DependencyGraph(deps)
+    (root.copy(dependencyGraph = dg), ())
+  }
 
-    object consumer extends Consumer {
-      var deps: MultiMap[Input, Input] = MultiMap.empty
+  private def addToSharedContext(src: SourceLocation, dst: SourceLocation)(implicit sctx: SharedContext): Unit = {
```
JonathanStarup, line 56: maybe just `addDependency`? and also add a doc comment

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
```diff
@@ -1,88 +1,617 @@
...
 
-      /**
-        * Adds a dependency `src -> dst` signifying that if `src` changes then `dst` must be recomputed.
-        */
-      private def addDependency(src: SourceLocation, dst: SourceLocation): Unit = {
-        deps += (src.sp1.source.input, dst.sp1.source.input)
-      }
+  /**
+    * Returns the dependencies of the given type.
+    * One of the base case for the recursive visiting.
```
JonathanStarup, line 62: I think these `One of the base case for the recursive visiting.` can be removed

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
```diff
@@ -1,88 +1,617 @@
...
+    */
+  private def visitType(tpe: Type)(implicit sctx: SharedContext): Unit = tpe match {
```
JonathanStarup, line 64: This function needs to recursively traverse types, it could be like `Type.Apply(Type.Alias(..), ..)`

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
```diff
@@ -1,88 +1,617 @@
...
+  private def visitSymUse(use: SymUse)(implicit sctx: SharedContext): Unit = use match {
+    case SymUse.AssocTypeSymUse(sym, loc) => addToSharedContext(sym.loc, loc)
+    case SymUse.CaseSymUse(sym, loc) => addToSharedContext(sym.loc, loc)
+    case SymUse.DefSymUse(sym, loc) => addToSharedContext(sym.loc, loc)
+    case SymUse.EffectSymUse(sym, qname) => addToSharedContext(sym.loc, qname.loc)
+    case SymUse.OpSymUse(sym, loc) => addToSharedContext(sym.loc, loc)
+    case SymUse.SigSymUse(sym, loc) => addToSharedContext(sym.loc, loc)
+    case SymUse.StructFieldSymUse(sym, loc) => addToSharedContext(sym.loc, loc)
+    case SymUse.TraitSymUse(sym, loc) => addToSharedContext(sym.loc, loc)
+    case _ => ()
```
JonathanStarup, line 84: can this not be removed? you have all cases right?

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
```diff
@@ -1,88 +1,617 @@
...
-  /** Checks the safety and well-formedness of `root`. */
+object Dependencies {
+  /**
+    * Updates the dependency graph of the given `root`.
+    * We visit all the definitions, effects, enums, instances, structs, traits, and type aliases to find the dependencies.
+    *
+    * Most of the time we just recursively visit the sub-nodes. We only add dependencies for three base cases:
+    *   - Type
+    *   - SymUse
+    *   - Instance
```
JonathanStarup, line 38: I think there should be some explanation here, but that is probably for @magnus

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
```diff
@@ -1,88 +1,617 @@
...
+   * One of the base case for the recursive visiting.
+   */
+  private def visitInstances(instances: List[TypedAst.Instance])(implicit sctx: SharedContext): Unit =
+    instances.foreach(instance => addToSharedContext(instance.trt.sym.loc, instance.loc))
 
-      override def consumeDefSymUse(symUse: SymUse.DefSymUse): Unit = {
-        addDependency(symUse.sym.loc, symUse.loc)
-      }
+  private def visitSpec(spec: TypedAst.Spec)(implicit sctx: SharedContext): Unit = {
+    visitType(spec.retTpe)
```
JonathanStarup, line 95: Its a bit hard to see from this code that everything is checked, but if you match on spec (or inline `val Spec(...) = spec` then its easier to see if something is forgotten, and new additions will also give compilation errors which is nice

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
```diff
@@ -1,88 +1,617 @@
...
+    visitType(spec.eff)
+    spec.fparams.foreach(visitFParam)
+    spec.tconstrs.foreach(visitTraitConstraint)
+    spec.econstrs.foreach(visitEqualityConstraint)
+  }
 
-      override def consumeEffectSymUse(symUse: SymUse.EffectSymUse): Unit = {
-        addDependency(symUse.sym.loc, symUse.qname.loc)
-      }
+  private def visitScheme(scheme: Scheme)(implicit sctx: SharedContext): Unit = {
```
JonathanStarup, line 103: like Spec, i think you should match on Scheme here

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
```diff
@@ -1,88 +1,617 @@
...
+
+  private def visitExp(exp: Expr)(implicit sctx: SharedContext): Unit = exp match {
+    case Expr.Cst(_, tpe, _) => visitType(tpe)
+
+    case Expr.Var(_, tpe, _) => visitType(tpe)
+
+    case Expr.Hole(_, tpe, eff, _) => visitType(tpe)
+      visitType(eff)
+
+    case Expr.HoleWithExp(exp, tpe, eff, _) => visitExp(exp)
```
JonathanStarup, line 269: In general we don't write on the => line and below the => line at the same time as far as I know

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
stormckey: I've matched over Scheme and Spec. There are some other types that has less fields could also be changed to use this style, shall we use this style consistently?

To keep the pattern matching more informative, maybe I should change unused fields from "_" to "_doc"?

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
stormckey: No, not all cases.

The current code preserves what the previous implementation did. In the previous implementation, only these 8 SymUse will add dependencies, some cases are left out.

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
stormckey: Ok, I've changed the code to visit sub nodes recursively the same as the visitor do.

Dependencies are only added for Cst and Alias, according to previous implementation.

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
JonathanStarup: i like the `_doc` idea and do it myself sometimes, but it isnt done in the compiler i think

JonathanStarup: for the other things, ill leave it up to you what is clearest

magnus-madsen:
@stormckey I see some compilation errors.

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
```diff
@@ -1,88 +1,719 @@
...
 
-      /**
-        * Adds a dependency `src -> dst` signifying that if `src` changes then `dst` must be recomputed.
-        */
-      private def addDependency(src: SourceLocation, dst: SourceLocation): Unit = {
-        deps += (src.sp1.source.input, dst.sp1.source.input)
-      }
+  /**
+    * Returns the dependencies of the given type.
+    */
```
magnus-madsen, line 65: The order of the file is a bit wonky. Here we have visitType as the first visit function.

Here I would expect these functions: visitDef, visitEff, visitEnum.... to follow the order used in def run.

Then after those we have visitExpr and visitType

and then later we have various helper functions.

We don't have an exact order, but something like the above. You can see how it looks in e.g. Safety.scala - although it might not be perfect.

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
magnus-madsen: What cases are missing? We should at least discuss if some cases need to be handled.

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
```diff
@@ -1,88 +1,719 @@
...
+    case SymUse.TraitSymUse(sym, loc) => addDependency(sym.loc, loc)
+    case _ => ()
+  }
 
-      override def consumeCaseSymUse(symUse: SymUse.CaseSymUse): Unit = {
-        addDependency(symUse.sym.loc, symUse.loc)
-      }
+  /**
+   * Returns the dependencies of the given instance.
+   */
```
magnus-madsen, line 98: Kill comment. It provides no useful information :)

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
```diff
@@ -1,88 +1,719 @@
...
-        addDependency(symUse.sym.loc, symUse.loc)
-      }
+  private def visitSpec(spec: TypedAst.Spec)(implicit sctx: SharedContext): Unit = spec match {
+    case TypedAst.Spec(_, _, _, _, fparams, declaredScheme, retTpe, eff, tconstrs, econstrs) =>
+      visitType(retTpe)
+      visitScheme(declaredScheme)
+      visitType(eff)
+      fparams.foreach(visitFParam)
+      tconstrs.foreach(visitTraitConstraint)
+      econstrs.foreach(visitEqualityConstraint)
```
magnus-madsen, line 109: I think you should visit the elements of each thing in the same order.

So here it would be fparams, declaredScheme, retTpe, and so on.

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
```diff
@@ -1,88 +1,719 @@
...
+  }
 
-      override def consumeEffectSymUse(symUse: SymUse.EffectSymUse): Unit = {
-        addDependency(symUse.sym.loc, symUse.qname.loc)
-      }
+  private def visitScheme(scheme: Scheme)(implicit sctx: SharedContext): Unit = scheme match {
+    case Scheme(_, tconstrs, econstrs, base) =>
+      visitType(base)
+      tconstrs.foreach(visitTraitConstraint)
+      econstrs.foreach(visitBroadEqualityConstraint)
```
magnus-madsen, line 116: Same comment about order.

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
```diff
@@ -1,88 +1,719 @@
...
-      }
+  private def visitBroadEqualityConstraint(ec: BroadEqualityConstraint)(implicit sctx: SharedContext): Unit = {
+    visitType(ec.tpe1)
+    visitType(ec.tpe2)
+  }
 
-      override def consumeTraitSymUse(symUse: SymUse.TraitSymUse): Unit = {
-        addDependency(symUse.sym.loc, symUse.loc)
-      }
+  private def visitBnd(bnd: TypedAst.Binder)(implicit sctx: SharedContext): Unit =
```
magnus-madsen, line 133: -> visitBinder

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
```diff
@@ -1,88 +1,719 @@
...
+    visitType(bnd.tpe)
 
-      override def consumeType(tpe: Type): Unit = tpe match {
-        case Type.Alias(cst, _, _, loc) =>
-          addDependency(cst.sym.loc, loc)
+  private def visitFParam(fparam: TypedAst.FormalParam)(implicit sctx: SharedContext): Unit = {
```
magnus-madsen, line 136: -> visitFormalParam

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
```diff
@@ -1,88 +1,719 @@
...
+    visitBnd(fparam.bnd)
+    visitType(fparam.tpe)
+  }
 
-        case Type.Cst(TypeConstructor.Enum(sym, _), loc) =>
-          addDependency(sym.loc, loc)
+  private def visitMatchRule(matchRule: TypedAst.MatchRule)(implicit sctx: SharedContext): Unit = {
```
magnus-madsen, line 141: We usually use single letters for arguments so `r: TypedAst.MatchRule` is fine.

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
```diff
@@ -1,88 +1,719 @@
...
+    visitPattern(matchRule.pat)
+    visitExp(matchRule.exp)
+    matchRule.guard.toList.foreach(visitExp)
+  }
 
-        case Type.Cst(TypeConstructor.Struct(sym, _), loc) =>
-          addDependency(sym.loc, loc)
+  private def visitPattern(pattern: TypedAst.Pattern)(implicit sctx: SharedContext): Unit = pattern match {
```
magnus-madsen, line 147: Same: p: TypedAst.Pattern is fine.

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
```diff
@@ -1,88 +1,719 @@
...
+      visitType(tpe)
+
+    case Expr.Var(_, tpe, _) =>
+      visitType(tpe)
+
+    case Expr.Hole(_, tpe, eff, _) =>
+      visitType(tpe)
+      visitType(eff)
+
+    case Expr.HoleWithExp(exp, tpe, eff, _) =>
```
magnus-madsen, line 294: Here you have shadowing-- i.e. you have two exp.

We always try to avoid shadowing. For this reason we would often call the argument to visitExp exp0

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
```diff
@@ -1,88 +1,719 @@
...
+
+  private def visitAssocTypeSig(assoc: TypedAst.AssocTypeSig)(implicit sctx: SharedContext): Unit =
+    assoc.tpe.toList.foreach(visitType)
+
+  private def visitSig(sig: TypedAst.Sig)(implicit sctx: SharedContext): Unit = {
+    visitSpec(sig.spec)
+    sig.exp.toList.foreach(visitExp)
+  }
+
+  private def visitExp(exp: Expr)(implicit sctx: SharedContext): Unit = exp match {
```
magnus-madsen, line 283: exp -> exp0

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
```diff
@@ -1,88 +1,719 @@
...
   }
 
+  private def visitTypeAlias(typeAlias: TypedAst.TypeAlias)(implicit sctx: SharedContext): Unit =
+    visitType(typeAlias.tpe)
+
+  /**
+    * Shared context for the dependencies phase.
+    * Ideally, we should use something like a concurrentMultiMap, but we don't have that in the standard library.
+    * So we just use a concurrentMap with a dummy value and keep the key-value pairs in the map.
+    */
```
magnus-madsen, line 717: I would iterate on the comment.

We could say:

We want to compute a set of dependency edges from an input to its dependencies. In other words, we want to compute `Map[Input, Set[Input]]`. However, since we are in a concurrent setting, we prefer to simple compute the
set of edges `Set[(Input, Input)]`. However, since Java has no `ConcurrentSet[t]` we instead use ` ConcurrentMap[(Input, Input), Unit]` to record the edges.

magnus-madsen: Maybe you can polish that text (or ask chatgpt).

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
```diff
@@ -1,88 +1,719 @@
...
+    case RestrictableChoosePattern.Var(bnd, tpe, _) =>
+      visitBnd(bnd)
+      visitType(tpe)
+    case RestrictableChoosePattern.Wild(tpe, _) =>
+      visitType(tpe)
+    case RestrictableChoosePattern.Error(tpe, _) =>
+      visitType(tpe)
+  }
+
+  private def visitField(field: (SymUse.StructFieldSymUse, Expr))(implicit sctx: SharedContext): Unit = {
```
magnus-madsen, line 201: This awkward. Usually we dont pass around pairs like this.

magnus-madsen (commented):
This looks great. I have a lot of comments asking for polish, but with one or two more rounds then I think this can be merged in 👍

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
stormckey: LocalDefSymUse(_, _), RestrictableCaseSymUse(_, _), RestrictableEnumSymUse(_, _)

magnus-madsen: OK; they are OK to skip. But can you write them out and write // do nothing -- local as a comment in each case.

PR #9624, opened 2025-01-06
feat: add debugging utility runWithTimeout

mlutze:


main/src/ca/uwaterloo/flix/language/phase/typer/Debug.scala
```diff
@@ -131,6 +132,19 @@ object Debug {
       ).mkString("\n"))
   }
 
+  /**
+    * Executes the function `f`, timing out after `limitMs` milliseconds.
+    */
+  private def runWithTimeout[A](limitMs: Int)(f: () => A): Option[A] = {
```
magnus-madsen, line 138: You may want to shutdown the executor if you don't want to leak resources.

magnus-madsen (approved):

PR #9630, opened 2025-01-07
fix: correct `Random.handleWithSeed` type and fix region check

JonathanStarup:
- Stop `Random.handleX` from being stateful before the last argument
- Remove unnecessary effects from seeded handler
- Fix region checking for effect with minus, xor, and concrete effects

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Regions.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Regions.scala
@@ -18,11 +18,11 @@ package ca.uwaterloo.flix.language.phase
 import ca.uwaterloo.flix.api.Flix
 import ca.uwaterloo.flix.language.CompilationMessage
 import ca.uwaterloo.flix.language.ast.TypedAst.*
-import ca.uwaterloo.flix.language.ast.{Kind, SourceLocation, Type}
+import ca.uwaterloo.flix.language.ast.{Kind, SourceLocation, Type, TypeConstructor}
 import ca.uwaterloo.flix.language.dbg.AstPrinter.*
 import ca.uwaterloo.flix.language.errors.TypeError
 import ca.uwaterloo.flix.language.phase.unification.Substitution
-import ca.uwaterloo.flix.util.{InternalCompilerException, ParOps, Validation}
+import ca.uwaterloo.flix.util.{CofiniteEffSet, InternalCompilerException, ParOps, Validation}
 
 import scala.collection.immutable.SortedSet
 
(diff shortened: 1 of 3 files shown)


main/src/ca/uwaterloo/flix/language/phase/Regions.scala
```diff
@@ -378,13 +378,16 @@ object Regions {
       * where `trueVars` are the variables ascribed the value TRUE,
       * and all other variables are ascribed the value FALSE.
       */
-    def eval(tpe: Type, trueVars: SortedSet[Type.Var]): Boolean = tpe match {
```
JonathanStarup, line 381: This function was improperly updated to set effects. It crashes for concrete effects and difference and xor. I've fixed this.

This is observable if you have `Abort` across a region scope fx

main/src/library/Random.flix
```diff
@@ -59,17 +59,19 @@ mod Random {
     ///
     /// In other words, re-interprets the `Random` effect using the `NonDet` and `IO` effects.
     ///
-    pub def handle(f: a -> b \ ef): (a -> b \ (ef - Random) + {NonDet, IO}) \ {NonDet, IO} =
-        let rnd = new JRandom(); // Reuse Random instance.
```
JonathanStarup, line 63: I've put the object inside the function such that `handle(f)` is stateless and wont interfere if invoked multiple times

main/src/library/Random.flix
```diff
@@ -84,24 +86,29 @@ mod Random {
...
+            let rnd = unchecked_cast (new JRandom(seed) as _ \ local);
+            run {
+                f(x)
+            } with Random {
+                def randomBool(k)       = k(unchecked_cast (rnd.nextBoolean() as _ \ local))
+                def randomFloat32(k)    = k(unchecked_cast (rnd.nextFloat() as _ \ local))
+                def randomFloat64(k)    = k(unchecked_cast (rnd.nextDouble() as _ \ local))
+                def randomInt32(k)      = k(unchecked_cast (rnd.nextInt() as _ \ local))
+                def randomInt64(k)      = k(unchecked_cast (rnd.nextLong() as _ \ local))
+                def randomGaussian(k)   = k(unchecked_cast (rnd.nextGaussian() as _ \ local))
```
JonathanStarup, line 102: I could've used `unsafely {NonDet, IO} run ..` but i also need to add `local` which we don't have nice syntax for

magnus-madsen (approved):

PR #9632, opened 2025-01-08
refactor: merge enum completer

stormckey:
According to  #9355, we now merge EnumCompleter.

It will provide all three completions for both UndefinedName and UndefinedType:
image
image
image
image
image

Generally, it works fine, but several problems remains:
- Should we keep using the priority rule defined in `getInternalPriority`, which is used by EnumCompleter before.
- Should we provide more information for this completion?
- Despite the usage of priority, the ranking is very strange. The priority for `use ModA.Enum` is Low while the priority for `A.B.C.Enum` is High, but their rank is not we expect from the priority.


main/src/ca/uwaterloo/flix/api/lsp/provider/completion/EnumCompleter.scala
```diff
@@ -0,0 +1,79 @@
...
+    isRoot || isResolved
+  }
+
+  /**
+   * Checks if the definition matches the QName.
+   * Names should match and the definition should be available.
+   */
+  private def matchesEnum(enm: TypedAst.Enum, namespace: List[String], ident: String, uri: String, qualified: Boolean): Boolean = {
+    val isPublic = enm.mod.isPublic && !enm.ann.isInternal
+    val isInFile = enm.sym.loc.source.name == uri
```
stormckey, line 72: I copyed the checking here from a previous implementaion. But I'm not sure if it's correct. Can we access something non-public if we are in the same file?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/EnumCompleter.scala
```diff
@@ -0,0 +1,79 @@
...
+        case enum if matchesEnum(enum, namespace, ident, uri, qualified = false) =>
+          EnumCompletion(enum, ap, qualified = false, inScope = inScope(enum, env))
+      })
+  }
+
+  /**
+   * Checks if the definition is in scope.
+   * If we can find the definition in the scope or the definition is in the root namespace, it is in scope.
+   */
+  private def inScope(decl: TypedAst.Enum, scope: LocalScope): Boolean = {
```
stormckey, line 56: This helper function is also defined in DefCompleter, with just minor differences.

But it's hard to unify them. As you can see that in the pattern match later, we need to specify difference pattern for Enum or Def, which is hard to control through an argument.

So maybe we just define this helper functions for each completer?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/EnumCompleter.scala
```diff
@@ -0,0 +1,79 @@
...
+    })
+    val isRoot = decl.sym.namespace.isEmpty
+    isRoot || isResolved
+  }
+
+  /**
+   * Checks if the definition matches the QName.
+   * Names should match and the definition should be available.
+   */
+  private def matchesEnum(enm: TypedAst.Enum, namespace: List[String], ident: String, uri: String, qualified: Boolean): Boolean = {
```
stormckey, line 70: This helper function is also defined in DefCompleter, with just minor differences.

We can actually unify these two functions by divide the Enum into `mod`, `ann` and `sym`, then this helper function could also be used by later completers, do you think it's a good idea? Or we should just define one for each completer?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/EnumCompleter.scala
magnus-madsen: I suppose so. We can refactor later.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/EnumCompleter.scala
magnus-madsen: Not sure, lets defer.

magnus-madsen (approved):

PR #9634, opened 2025-01-08
feat: make dependencies incremental

### LINKED ISSUE #9628: Make `Dependencies` phase incremental (in addition to being parallel)

stormckey:
fixes #9628
incrementalism json
I have no idea why the ratio is not 1.0, but it's almost totally incremental.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/Flix.scala
+++ b/main/src/ca/uwaterloo/flix/api/Flix.scala
@@ -595,7 +595,7 @@ class Flix {
             val (_, safetyErrors) = Safety.run(afterRedundancy, cachedTyperAst, changeSet)
             errors ++= safetyErrors
 
-            val (afterDependencies, _) = Dependencies.run(afterRedundancy)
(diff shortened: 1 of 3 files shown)


magnus-madsen:
We seem to have tests failures, so maybe we have overlooked something in our understanding?

stormckey:
> We seem to have tests failures, so maybe we have overlooked something in our understanding?

The error message is confusing. I have no idea what's wrong with the code.

magnus-madsen:
@stormckey I think this PR is probably easier to redo from scratch. 

I would suggest starting with just making root.defs incremental and see if it works.

Otherwise we probably have to whiteboard it.

stormckey:
It just works after I rebase and use the new api🤔

Fine, here is the perf figure:
image

magnus-madsen:
@stormckey It works? Do you know what the problem was?

stormckey:
No idea, but now it passes all the tests.

magnus-madsen (approved):

PR #9636, opened 2025-01-08
feat: annotate `java.lang.Character` with primitive effects

jaschdoc:
I have annotated the methods that can throw with `IO` but maybe that is incorrect?

Related to https://github.com/flix/flix/issues/9511


main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
```diff
@@ -1,6 +1,13 @@
 {
   "methods": {
     "java.lang.Boolean::getBoolean": "Env, IO",
+    "java.lang.Character::toString": "IO",
+    "java.lang.Character::codePointAt": "IO",
+    "java.lang.Character::codePointBefore": "IO",
+    "java.lang.Character::toChars": "IO",
+    "java.lang.Character::codePointCount": "IO",
+    "java.lang.Character::offsetByCodePoints": "IO",
+    "java.lang.Character::isDefined": "FsRead, IO",
```
jaschdoc, line 10: Not sure if this is right but the doc mentions a unicode file: https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Character.html#isDefined(char)

jaschdoc: Maybe this is already loaded in the JVM so it is not really `FsRead`

jaschdoc: Same thing for methods below

JonathanStarup:
id say exceptions are outside of what IO is. a pure function can also give stack overflow for example. so imo we just let exceptions be untracked

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
magnus-madsen: I wonder if "The Unicode Datafile" is actually a "file file". I think it is more likely it is something builtin-- but I am not 100% sure.

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
```diff
@@ -1,6 +1,18 @@
 {
   "methods": {
     "java.lang.Boolean::getBoolean": "Env, IO",
+    "java.lang.Character::toString": "IO",
+    "java.lang.Character::codePointAt": "IO",
```
magnus-madsen, line 5: I suppose the reason that codePointAt has IO is because it reads from an array?

magnus-madsen: And so it should probably have a heap effect in that array region, but we have no way to express that...

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
magnus-madsen: If there is actually such a file we should probably discuss it in the paper.

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
magnus-madsen: We probably need to discuss this at the next meeting.

jaschdoc: Ah good point. It was mostly just because it can throw `IndexOutOfBounds`

magnus-madsen: Roger. I think we are ignoring exceptions- at least for now.

magnus-madsen (approved):

PR #9640, opened 2025-01-10
feat: add `QName.endsWithDot`

magnus-madsen:


magnus-madsen:
@mlutze This is not perfect, but its an attempt to create a usable method that can we build on. Later the implementation of this method can then be fixed.

main/src/ca/uwaterloo/flix/language/ast/Name.scala
```diff
@@ -167,6 +167,38 @@ object Name {
...
+     *
+     * We report that the QName `x` ends with a dot.
+     */
+    def endsWithDot: Boolean = {
+      // We return false if this source location is unknown.
+      // This should not happen for code that the programmer is writing.
+      if (loc == SourceLocation.Unknown)
+        false
+      else {
+        val lineNumber = loc.sp2.line
```
mlutze, line 189: Why not just `loc.text.exists(_.lastOption == Some('.'))` ?

mlutze (approved):

magnus-madsen:
@stormckey I have added a new method to QName:  endsWithDot. Can you try to use it instead of your own dot-handling and see if it works?

Issue #9523, opened 2024-12-15
Annotate parts of the Java Class Library with primitive effects


magnus-madsen:

jaschdoc:
- [ ] https://github.com/flix/flix/issues/9753
- [ ] https://github.com/flix/flix/issues/9637

PR #9646, opened 2025-01-10
feat: annotate `java.lang.constant` with primitive effects

jaschdoc:
It performs reflection outside of the `reflect` package. It looks like it is done in a somewhat safe manner, but this is a little side thing we may have easily missed, so I just wanted to highlight it here.

https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/constant/ConstantDesc.html


magnus-madsen:
Thats crazy. Maybe we should mark everything in `java.lang.constant` as `{Sys, IO}`?

magnus-madsen (approved):

Issue #9637, opened 2025-01-08
Annotate `java.lang.Object` with primitive effects


jaschdoc:
There may be some funny methods in there but I have not yet looked.

magnus-madsen:
Yeah. I can imagine.

jaschdoc:
Close this?

PR #9647, opened 2025-01-10
feat: annotate `java.lang.Object` with primitive effects

jaschdoc:
Similar to `java.lang.System`, this is a special class so we should annotate it explicitly. The annotations in this PR are up for debate. Just a sketch / proposal here.

Question: Can effect annotations be sidestepped with `upcast`? E.g., can I do something bad in the `equals` implementation of some anonymous class and then upcast it to `Object` and then call `Object.equals` to just get `IO`?

Closes https://github.com/flix/flix/issues/9637


jaschdoc:
The annotations in this PR are up for debate. Just a sketch / proposal here.

jaschdoc:
@magnus-madsen I think we need to carefully review these changes (maybe together in person)

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
```diff
@@ -56,6 +56,12 @@
     "java.math.BigInteger::probablePrime": "NonDet, IO",
 
     "java.net.http.HttpClient::send": "Net, IO",
-    "java.net.http.HttpClient::sendAsync": "Net, IO"
+    "java.net.http.HttpClient::sendAsync": "Net, IO",
+
+    "java.lang.Object::toString": "NonDet, IO",
+    "java.lang.Object::finalize": "IO",
```
jaschdoc: Seems like `Sys` could be performed here.

jaschdoc:
It seems awfully weird that this is not `NonDet`, when the `toString` method is called.
```flix
    @Test
    def testConsumer(): Bool \ IO = {
        let mkObject = i -> new Object {
            def toString(_this: Object): String = "${i}"
        };
        let st = Ref.fresh(Static, "<none>");
        let stream = Stream.of(mkObject(8));
        stream.forEach((obj: Object) -> let s1 = obj.toString(); Ref.put(s1, st));
        Ref.get(st) == "8"
    }
```

magnus-madsen:
Closing because no longer relevant.

Issue #8759, opened 2024-09-30
Add `StructNewCompleter`


magnus-madsen:
Add a completer for struct new expressions:

See https://doc.flix.dev/structs.html#creating-a-struct

For example:

```
struct Person[r] {
    name: String,
    mut age: Int32,
    mut height: Int32
}

mod Person {
    pub def mkLuckyLuke(rc: Region[r]): Person[r] \ r =
        new Per
}
```

**Note:** __This may be much more difficult than what we have done so far. The problem is that the parser generates:__

```
❌ -- Parse Error (New) -------------------------------------------------- 

>> Expected at least one <ArgumentList>.

9 |         new Per
            ^^^^^^^
            Here
```

and NOT `UndefinedName` which would have made it is. 

Thus the first -- and hard task -- is to explore whether we can even slightly modify the parser to generate a different AST so we can obtain an `UndefinedName` from the `Resolver`.

magnus-madsen:
@stormckey This may be difficult-- bordering on impossible. I suggest you try to poke around in `Parser.scala` / `Weeder2.scala` and report back.

@JonathanStarup May also be able to offer some insights.

magnus-madsen:
I think the starting point is in `ambiguousNewExpr` in `Parser2.scala`.

stormckey:
Thanks to @JonathanStarup , the error message is now a UndefinedJvmClass, which is more specific than the previous ParseError.
<img width="620" alt="image" src="https://github.com/user-attachments/assets/5e482594-9ed9-4c3d-a1e0-85e9a68c0118" />

But it still doesn't works for qualified name
<img width="780" alt="image" src="https://github.com/user-attachments/assets/1824b94c-dd1b-448b-8133-6228f58fb580" />
as it suggests, a UndefinedJvmClass should not handle qualified name. But the problem here is new struct and new java object share the same syntax, and new struct can accept a qualified name. 

I can first go ahead and handle only the unqualified case.

stormckey:
Now the UndefinedJvmClass will also be triggered in threee cases:
- import java.sth
- new StringBuil
- new Pers
And the first two cases are actually for JvmClass. But the third case is actually for new struct.

I think it would be good that we add a new error for the third case and raise it there. Then for UndefinedJvmClass we will call ImportCompleter and a completer for Java classes.

magnus-madsen:
How do we add such an error? I think the problem is that we cannot know the intention.

stormckey:
[...]

PR #9651, opened 2025-01-10
refactor: add `StructCompleter`

stormckey:
One step for #9355
Preview:
image
image
image
image

Now it only applies to UndefinedType.

We also need to provide completions for structs in expr in cases like:
```
    let person = new Per
```
This problem is mention in #8759 and I will start another PR to support that case after this one is merged.


main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
```diff
@@ -16,6 +16,7 @@
 package ca.uwaterloo.flix.api.lsp.provider.completion
 
 import ca.uwaterloo.flix.api.Flix
+import ca.uwaterloo.flix.api.lsp.provider.completion.CompletionUtils.formatTParams
```
magnus-madsen, line 19: I dont like importing functions directly. Can we prefix them so we write:

CompletionUtils.formatTParams?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
```diff
@@ -270,4 +270,15 @@ object CompletionUtils {
...
   }
+
+  /**
+   * Format type params in the right form to be displayed in the list of completions
+   * e.g. "[a, b, c]"
+   */
+  def formatTParams(tparams: List[TypedAst.TypeParam]): String = {
+    tparams match {
+      case Nil => ""
+      case _ => tparams.map(_.name).mkString("[", ", ", "]")
```
magnus-madsen, line 281: I feel like we already have function like this somewhere. Can you search?

Also, perhaps we want snippets so you can easier fill out the params?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
stormckey: Ok, I will change that style. For some former use cases, I also use the imported function directly like fuzzyMatch. Maybe I should also change that to CompletionUtils.fuzzyMatch? In another PR?

stormckey:
Now it provides a snippet:
image

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
magnus-madsen: Lets do it

magnus-madsen:
Let me know when I should merge

stormckey:
It's done.

magnus-madsen (approved):

PR #9671, opened 2025-01-13
feat: introduce partitionOnValues

stormckey:
Introduce partitionOnValues to handle cases like root.Instances.

We will determine if an instance is stale or fresh by looking at the value of the map, instead of the key. I replicate the logic used in `partition` to implement this function.

I think the second filter in 
```
      val fresh = oldMap.filter(kv => !dirty.contains(kv._1.src.input)).filter(kv => newMap.contains(kv._1))
```
is redundant, if it's old and not dirty, it should still be in the newMap. Is there any special case for that?

I implement partitionOnValues in another way to make it more efficient. You can help check its correctness.

If the assertion above, "an item is old and not dirty will also be in the new map", is correct, I think this way works.

The incremental ratio is still 0.5
incrementalism json

Note:
- `mergeMappedList` could be placed in an util package so that it can also be used elsewhere. Maybe just phase/util?
- I make Instance Sourceable and add the body right in its definition. What is the best way to provide this src?

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/ChangeSet.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/ChangeSet.scala
@@ -72,12 +72,11 @@ sealed trait ChangeSet {
       (newMap, Map.empty)
 
     case ChangeSet.Dirty(dirty) =>
-      val (stale, fresh) = newMap.foldLeft((Map.empty[K, List[V]], Map.empty[K, List[V]])){ case ((stale, fresh), (k, vList)) =>
+      newMap.foldLeft((Map.empty[K, List[V]], Map.empty[K, List[V]])){ case ((stale, fresh), (k, vList)) =>
         val oldList = oldMap.getOrElse(k, Nil)
         val (freshList, staleList) = vList.partition(v => oldList.contains(v) &&  !dirty.contains(v.src.input))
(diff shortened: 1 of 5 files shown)


magnus-madsen:
> 🤔 Weird. All tests passed after I changed something that does not matter at all.

So maybe it matters? If you change it again, does the test fail?

magnus-madsen:
In any case: 

Lets split this PR into two: 

- one that adds `partitionOnValues` and
- one that begins to use it in various phases.

Moreover, I think it is time for us to add some unit tests for the changeSet class. I don't recall if it has dependencies, but I think it can be tested in isolation. When you make the PR for partitionOnValues can you write several unit tests?

stormckey:
> Moreover, I think it is time for us to add some unit tests for the changeSet class. I don't recall if it has dependencies, but I think it can be tested in isolation. When you make the PR for partitionOnValues can you write several unit tests?

I think we are right in the PR for partitionOnValues. I've removed the code that applied it to Safety.

I think I will test changeSet by some designed maps? What should the file called? `ChangeSetSuite` or `TestChangeSet`? The tests in language/ast are all called ...Suite while in other folder most are called Test...

magnus-madsen:
TestChangeSet

magnus-madsen (approved):

PR #9675, opened 2025-01-13
feat: introduce TraitCompleter

### LINKED ISSUE #8415: Add completions for `UndefinedTrait`s
We should offer completions when writing

```flix
enum Color with E
```

### LINKED ISSUE #9355: Merge completions from LocalScopeCompleter, ExprCompleter and AutoUseCompleter

stormckey:
fixes #8415
fixes #9355
We now provide completions for the example in #8415 
image

As other merged Completers, trait will handle all three cases.

Note:
- UndefinedTrait will be triggered in several cases:
  - `instance E` in this case we have a InstanceCompleter provide a useful snippet, triggered through ParseError
  - `def foo(a: a) with E` WithCompleter through ParseError too.
  - `enum Color with E` the case in the issue. Only TraitCompleter here.
  Since UndefinedTrait will also be triggered in these cases, sometimes we will provide duplicated completions.
- I think only several traits is derivable, so we are providing some fake completions?

For the first problem, I think we need to trigger different errors, or let UndefinedTrait to take a indicator to tell use the context so that we can provide correct completions. Then we can remove the completers from the ParseError.

For the second problem, if I understand correctly, we just need to provide completions for several known traits.


magnus-madsen:
In the `enum` cases the only valid completions are actually: `Eq`, `Order`, and `ToString`.

I wonder what I was thinking when I created the  ticket... because it seems already have decent coverage.

magnus-madsen:
I am marking this as draft until we can discuss what to do.

magnus-madsen:
I am not really sure what I had in mind here-- because what we have seems to work pretty well?

stormckey:
> I am not really sure what I had in mind here-- because what we have seems to work pretty well?

Yeah, the current completer can provide useful completions, except that they come from ParseError and they don't support things like autoUse or unqualified when already used.

magnus-madsen:
> > I am not really sure what I had in mind here-- because what we have seems to work pretty well?
> 
> Yeah, the current completer can provide useful completions, except that they come from ParseError and they don't support things like autoUse or unqualified when already used.

Ah- so maybe that was the problem. But because they come from parse error we are also kind of stuck, right?

stormckey:
> Ah- so maybe that was the problem. But because they come from parse error we are also kind of stuck, right?

Right, we can not access things like LocalScope etc.

stormckey:
The WithCompleter that works here also relies on regex, as mentioned in #9857

stormckey:
A great update here.

Now TraitCompleter will handle all the cases in UndefinedTrait, including:
derivation with only derivable traits:
image
image

constraint with snippet and auto use:
image
image

instantiation with snippet and auto use:
image
image

I reuse the code to generate instance snippet from former InstanceCompleter. Now it's placed in CompletionUtils.

stormckey:
By the way, it also works for the super trait case
image
image

But there is a bug( the auto use)

main/src/ca/uwaterloo/flix/language/ast/shared/TraitUsageKind.scala
```diff
@@ -23,6 +23,12 @@ sealed trait TraitUsageKind
   * Used to indicate the context of an undefined trait
   */
 object TraitUsageKind {
+  /**
+    * Represents a trait use in a call
+    * e.g. let res = E...
+    */
+  case object Call extends TraitUsageKind
```
magnus-madsen, line 30: Call -> Expr

(You cannot be sure this is a use in a call, it could just be a reference).

magnus-madsen:
Let me know when this is ready for review.

stormckey:
> Let me know when this is ready for review.

Now

magnus-madsen:
What's the plan here? I see lots of conflicts.

stormckey:
> What's the plan here? I see lots of conflicts.

conflicts are solved. All good but the bug mentioned here

> But there is a bug (the auto use). The ticket is here #9974

which should be solved else where.

So just review the code and then merge :)

magnus-madsen (approved):

PR #9689, opened 2025-01-14
fix: disallow qualified name in try-catch

### LINKED ISSUE #9685: Is qualified name allowed in try catch?
First mentioned in #9674, magnus suggests that we should not allow unqualified name in try catch.

Actually the example code is from the book https://doc.flix.dev/exceptions.html?highlight=try%20catch#exceptions.

But the fact is, both qualified name and unqualified name now works fine:
image

image

stormckey:
fixes #9685

Now the parser will not accept qualified name in try catch:
image

But the error message is helpless, how could we improve this?

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
@@ -1784,23 +1784,11 @@ object Weeder2 {
       expect(tree, TreeKind.Expr.TryCatchRuleFragment)
       mapN(pickNameIdent(tree), pickQName(tree), pickExpr(tree)) {
         case (ident, qname, expr) if qname.isUnqualified => CatchRule(ident, qname.ident, expr)
-        case _ =>
-          val error = IllegalQualifiedName(tree.loc)
+        case (_, qname, _) =>
+          val error = IllegalQualifiedName(qname.loc)
           sctx.errors.add(error)
           CatchRule(Name.Ident("error", tree.loc), Name.Ident("error", tree.loc), Expr.Error(error))
       }
-//      pickAll(TreeKind.Ident, tree).map(tokenToIdent) match {
-//        case ident :: className :: Nil =>
-//          println("first case")
-//          mapN(pickExpr(tree)){
-//            expr => CatchRule(ident, className, expr)
-//          }
-//        case _ =>
-//          println("second case")
-//          val error = Malformed(NamedTokenSet.CatchRule, SyntacticContext.Expr.OtherExpr, loc = tree.loc)
-//          sctx.errors.add(error)
-//          Validation.Success(CatchRule(Name.Ident("error", tree.loc), Name.Ident("error", tree.loc), Expr.Error(error)))
-//      }
     }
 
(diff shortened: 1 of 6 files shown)


main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1781,9 +1781,11 @@ object Weeder2 {
 
     private def visitTryCatchRule(tree: Tree)(implicit sctx: SharedContext): Validation[CatchRule, CompilationMessage] = {
       expect(tree, TreeKind.Expr.TryCatchRuleFragment)
-      mapN(pickNameIdent(tree), pickQName(tree), pickExpr(tree)) {
-        (ident, qname, expr) => CatchRule(ident, javaQnameToFqn(qname), expr)
+      val idents = pickAll(TreeKind.Ident, tree).map(tokenToIdent)
+      mapN(pickExpr(tree)){
+        expr => CatchRule(idents.head, idents.last, expr)
```
magnus-madsen, line 1786: I think this can fail. You should match on idents and check that there are two. If not, we have to produce some error.

stormckey: Is there an appropriate error or we should create one?

magnus-madsen: I think we have to create one

magnus-madsen: But you should experiment. Maybe the case cannot happen (?) Try write some partial catch rules and see how the weeder behaves

stormckey: Acctually I now has another idea: what if we parse a qualified name here and then check if it's qualified. This code will looks like:
```
      mapN(pickNameIdent(tree), pickQName(tree), pickExpr(tree)) {
        case (ident, qname, expr) if qname.isUnqualified => CatchRule(ident, qname.ident, expr)
        case _ =>
          val error = IllegalQualifiedName(tree.loc)
          sctx.errors.add(error)
          CatchRule(Name.Ident("error", tree.loc), Name.Ident("error", tree.loc), Expr.Error(error))
      }
``` 
and the error will looks like:
image

stormckey: I created a fake CatchRule when the name is qualified and that triggered the resolution error since that java class doesn't exist. How should we handle this?

stormckey: This way the error when there is only one ident is:
image

stormckey: what do you think @magnus-madsen

magnus-madsen: Let me get back to you :)

stormckey: Beep

stormckey:
Now the error message will be :
image

The second one needs improvement, but not in this PR :)

magnus-madsen (approved):

magnus-madsen:
Conflicted but otherwise good to go

PR #9697, opened 2025-01-15
refactor: use list map everywhere

### LINKED ISSUE #9694: Use ListMap everywhere consistently

stormckey:
fixes #9694

I use ListMap in all the Ast that I can find, is there any other place that we should also replace?

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Deriver.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Deriver.scala
@@ -52,7 +52,7 @@ object Deriver {
     val newInstances = derivedInstances.foldLeft(root.instances) {
       case (acc, inst) =>
         val accInsts = acc.getOrElse(inst.symUse.sym, Nil)
-        acc + (inst.symUse.sym -> (inst :: accInsts))
+        acc ++ (inst.symUse.sym -> (inst :: accInsts))
     }
     (root.copy(instances = newInstances), sctx.errors.asScala.toList)
   }
--- a/main/src/ca/uwaterloo/flix/language/phase/Deriver.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Deriver.scala
@@ -51,8 +51,7 @@ object Deriver {
     val derivedInstances = ParOps.parMap(root.enums.values)(getDerivedInstances(_, root)).flatten
     val newInstances = derivedInstances.foldLeft(root.instances) {
       case (acc, inst) =>
-        val accInsts = acc.getOrElse(inst.symUse.sym, Nil)
-        acc ++ (inst.symUse.sym -> (inst :: accInsts))
+        acc + (inst.symUse.sym -> inst)
     }
     (root.copy(instances = newInstances), sctx.errors.asScala.toList)
   }
--- a/main/src/ca/uwaterloo/flix/language/phase/Monomorpher.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Monomorpher.scala
@@ -403,8 +403,7 @@ object Monomorpher {
     */
   private def mkFastInstanceLookup(instances: ListMap[Symbol.TraitSym, Instance]): Map[(Symbol.TraitSym, TypeConstructor), Instance] = {
     for {
-      (sym, insts) <- instances
-      inst <- insts
+      (sym, inst) <- instances
     } yield ((sym, inst.tpe.typeConstructor.get), inst)
   }
 
--- a/main/src/ca/uwaterloo/flix/language/phase/Monomorpher.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Monomorpher.scala
@@ -402,7 +402,7 @@ object Monomorpher {
     * Creates a table for fast lookup of instances.
     */
   private def mkFastInstanceLookup(instances: ListMap[Symbol.TraitSym, Instance]): Map[(Symbol.TraitSym, TypeConstructor), Instance] = {
-    instances.flatMap{
+    instances.flatMap {
       case (sym, insts) => insts.map(inst => ((sym, inst.tpe.typeConstructor.get), inst))
     }.toMap
   }
--- a/main/src/ca/uwaterloo/flix/language/phase/Monomorpher.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Monomorpher.scala
@@ -402,9 +402,10 @@ object Monomorpher {
     * Creates a table for fast lookup of instances.
     */
   private def mkFastInstanceLookup(instances: ListMap[Symbol.TraitSym, Instance]): Map[(Symbol.TraitSym, TypeConstructor), Instance] = {
-    instances.flatMap {
-      case (sym, insts) => insts.map(inst => ((sym, inst.tpe.typeConstructor.get), inst))
-    }.toMap
+    for {
+      (sym, insts) <- instances
+      inst <- insts
+    } yield ((sym, inst.tpe.typeConstructor.get), inst)
   }
 
   /** Visit a struct field, simplifying its polymorphic type. */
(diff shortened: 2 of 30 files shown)


magnus-madsen:
@mlutze Can you review? You are the expert.

main/src/ca/uwaterloo/flix/language/phase/Deriver.scala
```diff
@@ -52,7 +52,7 @@ object Deriver {
     val newInstances = derivedInstances.foldLeft(root.instances) {
       case (acc, inst) =>
         val accInsts = acc.getOrElse(inst.symUse.sym, Nil)
-        acc + (inst.symUse.sym -> (inst :: accInsts))
+        acc ++ (inst.symUse.sym -> (inst :: accInsts))
```
mlutze, line 55: This looks suspicious. Won't this create too many values?

stormckey: You're right, I think `acc + (inst.symUse.sym -> inst)` is fine

main/src/ca/uwaterloo/flix/util/ParOps.scala
```diff
@@ -90,6 +91,16 @@ object ParOps {
       case (k, v) => (k, f(v))
     }.toMap
 
+  /**
+    * Applies the function `f` to every value of the map `m` in parallel.
+    */
+  def parMapValueList[K, A, B](m: ListMap[K, A])(f: List[A] => List[B])(implicit flix: Flix): ListMap[K, B] =
```
mlutze, line 97: I think it's more straightforward if this function f is just A => B. I am guessing that the functions called `visitInstanceList` just map over the list anyway, so we can get rid of those.

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -100,14 +100,14 @@ object Resolver {
               case () =>
                 ResolvedAst.Root(
                   table.traits,
-                  table.instances.m, // TODO NS-REFACTOR use ListMap elsewhere for this too
+                  table.instances, // TODO NS-REFACTOR use ListMap elsewhere for this too
```
mlutze, line 103: We can get rid of this TODO now :)

main/src/ca/uwaterloo/flix/language/phase/Monomorpher.scala
```diff
@@ -401,11 +401,10 @@ object Monomorpher {
   /**
     * Creates a table for fast lookup of instances.
     */
-  private def mkFastInstanceLookup(instances: Map[Symbol.TraitSym, List[Instance]]): Map[(Symbol.TraitSym, TypeConstructor), Instance] = {
-    for {
-      (sym, insts) <- instances
-      inst <- insts
-    } yield (sym, inst.tpe.typeConstructor.get) -> inst
+  private def mkFastInstanceLookup(instances: ListMap[Symbol.TraitSym, Instance]): Map[(Symbol.TraitSym, TypeConstructor), Instance] = {
+    instances.flatMap{
```
mlutze, line 405: Space before `{`

main/src/ca/uwaterloo/flix/util/collection/ListMap.scala
```diff
@@ -93,4 +93,13 @@ case class ListMap[K, V](m: Map[K, List[V]]) {
    */
   def --(ks: Iterable[K]): ListMap[K, V] = ListMap(m -- ks)
 
+  def values: Iterable[List[V]] = m.values
```
mlutze, line 96: Should this instead get all the values of the list (`Iterable[V]`)?

mlutze (commented):
I think ideally, ListMap should hide a bit more of its internals. So users of the interface shouldn't usually be dealing with lists directly.

This interface is what I mean: https://hackage.haskell.org/package/multi-containers-0.2/docs/Data-Multimap.html

main/src/ca/uwaterloo/flix/util/collection/ListMap.scala
```diff
@@ -93,4 +93,13 @@ case class ListMap[K, V](m: Map[K, List[V]]) {
...
+
+  def keys: Iterable[K] = m.keys
```
magnus-madsen, line 98: All functions should be commented and ordered in a meaningful way.

stormckey: What is a meaningful order🤔 I've put all the functions before operators, should we order all the functions alphabetically?

magnus-madsen:
@stormckey test failures

stormckey:
> @stormckey test failures

I'm on it

stormckey:
The problem is checkInstancesOfTrait used to check over a list of instances for a trait, so it can use a `var heads` to check overlapping. But now, as we changed `ListMap.values` to return `Iterable[V]` instead of `Iterable[List[V]]`, that check fails.
I think we still need a way to return `Iterable[List[V]]` for this use case. Maybe add a `valueLists` method?

magnus-madsen:
> I think we still need a method to return Iterable[List[V]] for this use case. Maybe add a valueLists method?

I will let @mlutze comment on this, and on this PR in general.

mlutze:
> I think we still need a method to return `Iterable[List[V]]` for this use case. Maybe add a `valueLists` method?

That seems reasonable to me.



Overall this is looking good, I just have a few requests to make my `ListMap` dreams come true.
1. `get` should always just return a List, not an option. It can be empty if there is no mapping.
2. `getOrDefault` can then go away, because the `default` is always `Nil`
3. `flatMap` can take a function `f: ((K, V)) => Iterable[A]` because I think we always just iterate over that list afterward anyway
4. `withFilter` can do the same

stormckey:
> 4. `withFilter` can do the same

It's not trivial to rewrite withFilter in this way maybe. As we cannot directly call withFilter over internal m.

We could build a flattened map and call withFilter over that, which is just a stupid way to make the code compile.

Or maybe we should write a dedicated WithFilter for this and things may be fine.

Or we could just keep the old signature.

(I found somewhere that we used to just return `this` for withFilter, but that doest work here as it will produce type mismatch)

magnus-madsen:
@stormckey @mlutze I will let you two discuss a bit more. Then you can ping me :)

mlutze:
> > 4. `withFilter` can do the same
> 
> It's not trivial to rewrite withFilter in this way maybe. As we cannot directly call withFilter over internal m.
> 
> We could build a flattened map and call withFilter over that, which is just a stupid way to make the code compile.
> 
> Or maybe we should write a dedicated WithFilter for this and things may be fine.
> 
> Or we could just keep the old signature.
> 
> (I found somewhere that we used to just return `this` for withFilter, but that doest work here as it will produce type mismatch)

Ok, good points.

IMO flattening may actually be the best. But we can just build a flattened `Iterable` instead of a proper map:
```scala
    for {
      (k, vs) <- m
      v <- vs
      if (f(k, v))
    } yield (k, v)
  }
```

The filtering will be eager, but it's probably worth the cost to avoid the complexity of an entire `WithFilter` implementation.

stormckey:
Many tests failed for the former impl in `monomorpher`, which is:
```
    for {
      (sym, inst) <- instances
    } yield ((sym, inst.tpe.typeConstructor.get), inst)
```
which is fixed after I change this to a map. But I'm not sure what is the problem with this impl. Did I do anything wrong with the `withFilter`?

stormckey:
You're right, thanks! It is safe to return a Seq.

I think I'm done with this PR and you can do a final check if anything needs update.

mlutze (approved):

mlutze:
I'll hand it over to @magnus-madsen now

magnus-madsen (approved):

PR #9703, opened 2025-01-16
fix: Entrypoints handles malformed types

### LINKED ISSUE #9589: EntryPoints must support malformed types
```
ca.uwaterloo.flix.util.InternalCompilerException: Unexpected effect 'Error'.
	at ca.uwaterloo.flix.language.phase.EntryPoints$CheckEntryPoints$.eval(EntryPoints.scala:384)
	at ca.uwaterloo.flix.language.phase.EntryPoints$CheckEntryPoints$.isPrimitiveEffect(EntryPoints.scala:348)
```

JonathanStarup:
fixes #9589

I've removed most of the `throw` cases in the phase by handling malformed/error types explicitly. It is verbose but clear IMO


main/src/ca/uwaterloo/flix/language/phase/EntryPoints.scala
```diff
@@ -568,15 +591,42 @@ object EntryPoints {
...
+      case EvalRes.ErrorOrMalformed => EvalRes.ErrorOrMalformed
+    }
+
+    def flatMap[B](f: T => EvalRes[B]): EvalRes[B] = this match {
+      case EvalRes.Ok(x) => f(x)
+      case EvalRes.ErrorOrMalformed => EvalRes.ErrorOrMalformed
     }
+  }
+
+  object EvalRes {
```
magnus-madsen, line 616: Any reason not to use Option or Result?

magnus-madsen: I think our own Result would fit here.

JonathanStarup: I had separate cases for Error and Malformed previously - but it also avoids option blindness or whatever it is called. You want me to just use Result?

JonathanStarup: In this case it would also be `Result[T, Unit]` or `Option`

magnus-madsen: I think the nicest thing you can do is:

```
case object MyError
Result[t, MyError]
```

Its precise, monadic, and extendable.

JonathanStarup: That's true, but sometimes you also don't like that I have results of options etc. so I don't think its black and white.

In this case I have no error information, so here it would just be
```
case object ErrorOrMalformed
Result[T, ErrorOrMalformed]
```
should I do that change?

magnus-madsen: Seems better than what it is

JonathanStarup: I've implemented it now - I forgot that the type of `ErrorOrMalformed` is `ErrorOrMalformed.type`

magnus-madsen (approved):

magnus-madsen:
I think this turned out well.

magnus-madsen:
@JonathanStarup There is a small conflict, but otherwise this is ready to go.

PR #9709, opened 2025-01-17
feat: require Unit type for sequence expressions

mlutze:


magnus-madsen:
This seems like a language change? In general this is too restrictive, hence why we have MustUse

PR #9711, opened 2025-01-17
test: add test to `TestSetUnification`

magnus-madsen:


magnus-madsen:
@mlutze Lets merge this and then I will extend it over time.

mlutze (approved):
We don't want the old test anymore?

PR #9718, opened 2025-01-19
feat: add `Type.freshEffSlackVar`

magnus-madsen:


main/src/ca/uwaterloo/flix/language/ast/Symbol.scala
```diff
@@ -122,15 +122,15 @@ object Symbol {
...
+  def freshKindedTypeVarSym(text: VarText, kind: Kind, isRegion: Boolean, isSlack: Boolean, loc: SourceLocation)(implicit scope: Scope, flix: Flix): KindedTypeVarSym = {
+    new KindedTypeVarSym(flix.genSym.freshId(), text, kind, isRegion, isSlack, scope, loc)
   }
 
   /**
     * Returns a fresh type variable symbol with the given text.
     */
   def freshUnkindedTypeVarSym(text: VarText, isRegion: Boolean, loc: SourceLocation)(implicit scope: Scope, flix: Flix): UnkindedTypeVarSym = {
-    new UnkindedTypeVarSym(flix.genSym.freshId(), text, isRegion, scope, loc)
+    new UnkindedTypeVarSym(flix.genSym.freshId(), text, isRegion, isSlack = false, scope, loc)
```
magnus-madsen, line 133: I had to make a choice here (and elsewhere) whether to propagate this bit to the arguments. 

But since it can only be set in very select places, I decided instead to not make it an argument, but default it to false.

its only through Type.fressEffSlackVar it is set. And that is only called in 3 places or so.

magnus-madsen: Thoughts?

magnus-madsen:
I think it may make sense to consolidate the two Booleans into some kind of enum. But I suppose I would until I know if this information is useful and then ask a student programmer to help out.

main/src/ca/uwaterloo/flix/language/ast/Symbol.scala
mlutze: Seems reasonable.

mlutze (approved):

JonathanStarup:
Is this for metrics/debugging or for optimization?

magnus-madsen:
> Is this for metrics/debugging or for optimization?

It is for my BU optimization idea.

JonathanStarup:
> > Is this for metrics/debugging or for optimization?
> 
> It is for my BU optimization idea.

sorry if I missed something, but what is the idea?

Issue #9524, opened 2024-12-15
Add support for the `effect-lock` command


magnus-madsen:
Add a command `effect-lock` which generates an `effect.lock` file with content like:

```
github:faisalman/UAParser.parse: String -> UAParser.Browser \ { }
```

I would add a separate class that takes a TypedAst and generates the above file. 

Note: We only want to store signatures for functions (etc) that are actually used.

We could then call this phase-like class from Flix.scala.

Question 1: Do we want to use JSON? (I think so)
Question 2: What should the command be called? `effect-lock` or `lock-effects`? (Perhaps the latter is better?)

jaschdoc:
> Question 2: What should the command be called? `effect-lock` or `lock-effects`? (Perhaps the latter is better?)

What about just `lock`?

jaschdoc:
> Note: We only want to store signatures for functions (etc) that are actually used.

Why is that again? It seems like a malicious upgrade could sneak stuff into a package that way. E.g., I do not use th function `f` from the library `l` and then I upgrade so now `f` is malicious. Then 10 days pass and I discover I need `f` but now I was not notified something wonky happened.

One compromise could be to store all accessible functions from the library (anything marked with `pub`)

magnus-madsen:
> > Note: We only want to store signatures for functions (etc) that are actually used.
> 
> Why is that again? It seems like a malicious upgrade could sneak stuff into a package that way. E.g., I do not use th function `f` from the library `l` and then I upgrade so now `f` is malicious. Then 10 days pass and I discover I need `f` but now I was not notified something wonky happened.
> 
> Another approach could be to store all accessible functions from the library (anything marked with `pub`)

I see these arguments:

- It requires storing a lot of extra information that is mostly unused.
- It makes it much more difficult for the programmer to inspect what is going on. (E.g. do you trust this upgrade of a function you have never heard of? -- that's not good).
- It prevents changes to the library even if that functionality is unused.
- If you think about it, the extra security it offers is very very small.

magnus-madsen:
Fixed in master.

PR #9730, opened 2025-01-20
feat: add `effect-lock` command

jaschdoc:
Closes https://github.com/flix/flix/issues/9524

### DIFF
--- a/main/src/ca/uwaterloo/flix/Main.scala
+++ b/main/src/ca/uwaterloo/flix/Main.scala
@@ -317,6 +317,9 @@ object Main {
         case Command.CompilerMemory =>
           CompilerMemory.run(options)
 
+        case Command.EffectLock =>
+          println("Effect Lock run")
+
       }
     }
 
@@ -401,6 +404,8 @@ object Main {
     case object CompilerPerf extends Command
 
     case object CompilerMemory extends Command
+
+    case object EffectLock extends Command
 
   }
 


main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -825,4 +832,21 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
...
+    check(flix) match {
+      case Validation.Failure(errors) => Validation.Failure(errors)
+      case Validation.Success(root) =>
+        val outputStream = Files.newOutputStream(getEffectLockFile(path))
+        val signatures = mkEffectLockSignatures(root).mkString("\n") // TODO: make JSON
+        Validation.Success(outputStream.write(signatures.getBytes))
+    }
+  }
+
+  def mkEffectLockSignatures(root: TypedAst.Root): List[String] = {
```
magnus-madsen, line 846: I suggest a separate compiler phase. But perhaps even better, we should extend the Safety phase.

jaschdoc: Yes agree. Was exploring code here that I would refactor before marking the pr green

magnus-madsen: Yeah, lets get some code in and then we can iterate.

main/src/ca/uwaterloo/flix/api/EffectLock.scala
```diff
@@ -0,0 +1,348 @@
...
+
+    case Expr.HoleWithExp(exp, _, _, _) =>
+      visitExp(exp)
+
+    case Expr.OpenAs(_, exp, _, _) =>
+      visitExp(exp)
+
+    case Expr.Use(sym, alias, exp, _) => // TODO: look up functions in alias
+      visitExp(exp)
+
```
jaschdoc, line 102: @magnus-madsen does `Resolver` update aliases in `use` or why do we not have to consider aliases?

jaschdoc:
@magnus-madsen maybe we should split the `Serialization` stuff into its own PR?

main/src/ca/uwaterloo/flix/api/effectlock/serialization/STC.scala
```diff
@@ -0,0 +1,112 @@
...
+  case object Bool extends STC
+
+  case object Char extends STC
+
+  case object Float32 extends STC
+
+  case object Float64 extends STC
+
+  case object BigDecimal extends STC
+
```
jaschdoc, line 24: @magnus-madsen we could remove the wrapper object `STC` so the objects are serialized to their names rather than `STC$Name`. E.g., `Void` would be serialized to `Void` instead of `STC$Void`

magnus-madsen: I think that is reasonable.

main/src/ca/uwaterloo/flix/api/effectlock/EffectLock.scala
```diff
@@ -0,0 +1,46 @@
...
+import ca.uwaterloo.flix.language.ast.shared.Scope
+import ca.uwaterloo.flix.language.phase.unification.{EqualityEnv, Unification}
+
+object EffectLock {
+
+  /**
+    * Returns true if `sc1` is unifiable with `sc2` or if `sc1` is a monomorphic downgrade of `sc2`.
+    */
+  def isSafe(sc1: Scheme, sc2: Scheme)(implicit flix: Flix): Boolean = { // TODO: Check if new flix instance can be used locally
+    unifiableSchemes(sc1, sc2) || monomorphicDowngrade(sc1, sc2)
```
jaschdoc, line 14: @magnus-madsen I think I could use an extra check on this. Does this seem right?

jaschdoc: And if so, I am having some issues with checking monomorphic downgrade

magnus-madsen: I think this is where @AStenbaek needs to join

magnus-madsen: Its not entirely correct as it is written, since it allows bi-directional unification.

main/src/ca/uwaterloo/flix/api/effectlock/EffectLock.scala
```diff
@@ -1,100 +1,46 @@
...
+    unification.isDefined
+  }
 
-    case (Type.Cst(tc1, loc1), Type.Cst(tc2, loc2)) =>
-      println(tc1)
-      println(tc2)
-      isSafe(tc1, tc2)
+  private def monomorphicDowngrade(sc1: Scheme, sc2: Scheme): Boolean = {
+    val noQuantifiers = sc1.quantifiers.isEmpty && sc2.quantifiers.isEmpty
+    val sameMonomorphicType = checkSameMonomorphicType(sc1.base, sc2.base)
```
magnus-madsen, line 27: Could this not just be == ?

main/src/ca/uwaterloo/flix/api/effectlock/EffectLock.scala
jaschdoc: Yes, that's also what the tests show

main/src/ca/uwaterloo/flix/api/effectlock/EffectLock.scala
jaschdoc: Are effects not included in `sc1.base`?

magnus-madsen:
We still want this-- but it should come in smaller PRs, so I am closing this for now.

PR #9734, opened 2025-01-21
feat: load main for the standalone lsp when initializing

stormckey:
Change the architect a little bit.

`didOpen` will try to add the opened file to the flix instance if it's a flix source file.


stormckey:
For the new config today, we can receive the didOpen request when starting neovim.

I'll change the code a little bit to utilize this.

magnus-madsen (approved):

PR #9738, opened 2025-01-22
refactor: introduce OpCompleter

stormckey:
OpCompleter will work for two cases:
- UndefinedName:
  image
  image
  image
  image
  image
- UndefinedOp
  image
  image
  Due to the syntax of effect handling, we will never auto use UndefinedOp, neither will we provide completions for qualified name.

Note:
- I add EffectCompleter to UndefinedName so that we can first complete `DivByZero` then its op.
- The completer will not work in this case:
  image
  And the reason is that the completer that should work here is OpCompleter, which should find this is a qualified name and check if there is a qualified op that matches. But it will only check for `A.DivByZero.divByZero` which doesn't match with `DivByZero.divBy`. The problem is that every time we meet a dot, we are not looking into that thing. But calling different completers to check if they have matched qualified name. Here `DivByZero` is in scope, but that is not what the qualified name part of OpCompleter will consider currently.
- The completer will not work in this case:
  image
  As you can see, no resolution error is triggered in this case. We need to improve the parser to fix this.
  I will create a ticket to discuss this.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
@@ -79,6 +79,23 @@ object CompletionUtils {
   }
 
   /**
+    * Generate a snippet which represents calling a function.
+    * Drops the last one or two arguments in the event that the function is in a pipeline
+    * (i.e. is preceeded by `|>`, `!>`, or `||>`)
+    */
+  def getOpHandlerSnippet(name: String, fparams: List[TypedAst.FormalParam])(implicit context: CompletionContext): String = {
+    val functionIsUnit = isUnitFunction(fparams)
+
+    val args = fparams.zipWithIndex.map {
+      case (fparam, idx) => "$" + s"{${idx + 1}:?${fparam.bnd.sym.text}}"
+    } :+ s"$${${fparams.length + 1}:resume}"
+    if (functionIsUnit)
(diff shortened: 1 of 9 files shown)


main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
```diff
@@ -78,6 +78,23 @@ object CompletionUtils {
       name
   }
 
+  /**
+    * Generate a snippet which represents calling a function.
+    * Drops the last one or two arguments in the event that the function is in a pipeline
```
magnus-madsen, line 83: Is this comment true? How does that make sense here?

magnus-madsen (commented):
Looks good. Just one comment

magnus-madsen:
I agree the parser needs improvements. I suggest we look into that in a few weeks once some of the other stuff is in place (lsp, incrementalism, merging the rest of the completers).

magnus-madsen:
Btw, I appreciate the attention to detail with also supporting completion for the handler part 👍

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/CompletionUtils.scala
stormckey: Thanks for pointing that out. I changed the impl but forgot to change the doc. Now it's fixed.

magnus-madsen (approved):

PR #9740, opened 2025-01-22
feat: compile flix for stand alone lsp

stormckey:
Now the standlone lsp will compile flix on open and on change. And diagnostics are properly returned. I also handle the crash situation but I have no immediate example to make it crash.
image

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/Position.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/Position.scala
@@ -21,6 +21,8 @@ import ca.uwaterloo.flix.util.Result.{Err, Ok}
 import org.json4s.JsonDSL.*
 import org.json4s.*
 
+import org.eclipse.lsp4j
+
 /**
   * Companion object for [[Position]].
   */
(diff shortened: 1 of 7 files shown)


magnus-madsen:
Very cool! I will review a bit later.

main/src/ca/uwaterloo/flix/api/lsp/Position.scala
```diff
@@ -68,6 +70,8 @@ case class Position(line: Int, character: Int) extends Ordered[Position] {
   // NB: LSP line and column numbers are zero-indexed, but Flix uses one-indexed numbers internally.
   def toJSON: JValue = ("line" -> (line - 1)) ~ ("character" -> (character - 1))
 
+  def toLsp4j: lsp4j.Position = new lsp4j.Position(line - 1, character - 1)
```
magnus-madsen, line 73: I would copy this comment from above:

// NB: LSP line and column numbers are zero-indexed, but Flix uses one-indexed numbers internally.

magnus-madsen (approved):

magnus-madsen:
Great-- merged! I added two comments, you can consider them for follow-up PR.

PR #9744, opened 2025-01-23
refactor: let HoverProvider return Option[Hover]

stormckey:
Change HoverProvider.processHover() to return Option[Hover] so that it can be further converted to JSON for VSCode LSP or to other formats for other LSP.


magnus-madsen:
In the future its probably best to separate refactorings from LSP4J additions.

In particular, I don't think HoverResult is a good abstraction. 

Why can we not just return an `Option` or a `Result`?

main/src/ca/uwaterloo/flix/api/lsp/HoverResult.scala
```diff
@@ -0,0 +1,34 @@
+package ca.uwaterloo.flix.api.lsp
```
magnus-madsen, line 1: -> Add copyright header for all new files.

main/src/ca/uwaterloo/flix/api/lsp/Position.scala
```diff
@@ -40,6 +40,9 @@ object Position {
   def fromEnd(loc: SourceLocation): Position =
     Position(loc.endLine, loc.endCol)
 
+  def fromLsp4j(pos: lsp4j.Position): Position =
```
magnus-madsen, line 43: Add comment

main/src/ca/uwaterloo/flix/api/lsp/HoverResult.scala
magnus-madsen: (But I don't think we should have this abstraction. I think it is a "false abstraction".)

Instead, we should just use Option or Result

magnus-madsen:
What does LSP say we should return if nothing is there?

stormckey:
> What does LSP spec say we should return if nothing is there?

It just says Hover | null
image
should we just return none for both sever? In that case we could just use Option instead of Result.

magnus-madsen:
> > What does LSP spec say we should return if nothing is there?
> 
> It just says Hover | null image https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_hover should we just return none for both sever? In that case we could just use Option instead of Result.

I suggest to use Option and then return null for LSP. But for LSP VScode we must return an empty result as done today (otherwise it breaks).

stormckey:
Got it.

stormckey:
Should I also remove all the toLsp4j method written for some classes? Leave them for the next PR?

magnus-madsen:
I just want to make 100% sure we do not break in any VSCode stuff.

main/src/ca/uwaterloo/flix/api/lsp/VSCodeLspServer.scala
```diff
@@ -278,7 +277,10 @@ class VSCodeLspServer(port: Int, o: Options) extends WebSocketServer(new InetSoc
       ("id" -> id) ~ HighlightProvider.processHighlight(uri, pos)(root)
 
     case Request.Hover(id, uri, pos) =>
-      ("id" -> id) ~ HoverProvider.processHover(uri, pos)(root, flix)
+      HoverProvider.processHover(uri, pos)(root, flix) match {
+        case Some(hover) => ("id" -> id) ~ hover.toJSON
+        case None => ("id" -> id) ~ ("status" -> ResponseStatus.InvalidRequest) ~ ("result" -> null)
```
magnus-madsen, line 282: Why null here? What did we return before? Will this not trigger a NPE on VSCode?

stormckey: No, no exception is raised when hovering something that doesn't find anything.

We used to return a String indicating that nothing is found, which is just ignored by VSCode.

I could change it back to a plain message like "Nothing found by the HoverProvider" if that is safer.

magnus-madsen: Lets do that. Seems safer.

magnus-madsen (approved):

PR #9746, opened 2025-01-23
refactor: use ListMap for ChangeSet

stormckey:
use ListMap instead of List[k, List[V]]

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/ChangeSet.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/ChangeSet.scala
@@ -16,6 +16,7 @@
 package ca.uwaterloo.flix.language.ast
 
 import ca.uwaterloo.flix.language.ast.shared.{DependencyGraph, Input}
+import ca.uwaterloo.flix.util.collection.ListMap
 
 sealed trait ChangeSet {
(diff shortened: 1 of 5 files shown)


magnus-madsen:
I will ask if @mlutze can spare some cycles before I look

stormckey:
I'm not sure if apply should accept `K, List[V]` or `K, V`. If we want to adhere to the same design pattern, we should make it `K, V`. But
```
1 -> List(1, 2, 3, 4, 5)
```
is easier than
```
1 -> 1, 1 -> 2, 1-> 3, 1 ->4, 1 -> 5
```

magnus-madsen:
@mlutze Can you review and if happy you can just merge? But maybe @stormckey should just add a few unit tests for the most important functions

mlutze (approved):
LGTM

I think the apply taking lists is fine. It's also OK that it's inconsistent with `from` in this case IMO.

mlutze:
> @mlutze Can you review and if happy you can just merge? But maybe @stormckey should just add a few unit tests for the most important functions

Yes, a couple unit tests would be a good idea

stormckey:
I'm adding the tests.

PR #9747, opened 2025-01-23
feat: support hover for the standalone lsp

stormckey:
Preview:
image
image
image

We will return a null when nothing is found and you can see that null is properly processed as the bottom says: "no information available"

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/Hover.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/Hover.scala
@@ -15,6 +15,7 @@
  */
 package ca.uwaterloo.flix.api.lsp
 
+import org.eclipse.lsp4j
 import org.json4s.JObject
 import org.json4s.JsonDSL.*
 
@@ -23,4 +24,12 @@ case class Hover(contents: MarkupContent, range: Range) {
     val result = ("contents" -> contents.toJSON) ~ ("range" -> range.toJSON)
     ("status" -> ResponseStatus.Success) ~ ("result" -> result)
   }
+
+  def toLsp4j: lsp4j.Hover = {
+    val hover = new lsp4j.Hover()
+    hover.setContents(contents.toLsp4j)
+    hover.setRange(range.toLsp4j)
+    hover
+  }
+
 }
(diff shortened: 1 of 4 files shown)


main/src/ca/uwaterloo/flix/api/lsp/Hover.scala
```diff
@@ -23,4 +24,12 @@ case class Hover(contents: MarkupContent, range: Range) {
     val result = ("contents" -> contents.toJSON) ~ ("range" -> range.toJSON)
     ("status" -> ResponseStatus.Success) ~ ("result" -> result)
   }
+
+  def toLsp4j: lsp4j.Hover = {
+    val hover = new lsp4j.Hover()
```
magnus-madsen, line 29: Is this the recommend style? is Hover like an inner class?

main/src/ca/uwaterloo/flix/api/lsp/MarkupContent.scala
```diff
@@ -26,4 +27,11 @@ import org.json4s.*
   */
 case class MarkupContent(kind: MarkupKind, value: String) {
   def toJSON: JValue = ("kind" -> kind.toJSON) ~ ("value" -> value)
+
+  def toLsp4j: lsp4j.MarkupContent = {
+    val markupContent = new org.eclipse.lsp4j.MarkupContent()
```
magnus-madsen, line 32: Here you have to use the FQN because of name class.

magnus-madsen (approved):

magnus-madsen:
I will merge now :) You can consider my comments for future PRs.

main/src/ca/uwaterloo/flix/api/lsp/Hover.scala
stormckey: No, I used this style since the name of the lsp4j class always conflicts with the flix ones.

So it's better to fully qualify them to org.eclipse.lsp4j.*** ?

magnus-madsen: Either style is fine with me. I leave it to you, as long as it is consistent across files.

stormckey: Ok, I will choose using `lsp4j.***` with `import org.eclipse.lsp4j`.

PR #9758, opened 2025-01-24
refactor: introduce tag completer

stormckey:
Preview:
image
image
image
image
image
Also works for UndefinedTag
image
image
image

The previous EnumTagCompleter is also used for `SyntacticContext.Pat`, I write a case to repeat this context, and we can see that `UndefinedTag` is also raised here, so it's safe to remove EnumTagCompleter from this `SyntacticContext.Pat` in this case. But I am not sure if there is another case of `SyntacticContext.Pat`.
image

Note:
- I add EnumCompleter to UndefinedTag so that we can first complete "A.Clr" then "A.Clr.Green"
- Unlike other completers, we don't check if the tag is public since that syntax is not valid. But maybe we should check if the enum is public? The same issue may also apply to other internal constructs like sigs or ops.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
@@ -54,8 +54,8 @@ object CodeActionProvider {
     case ResolutionError.UndefinedTrait(qn, ap,  _, loc) if overlaps(range, loc) =>
       mkUseTrait(qn.ident, uri, ap)
 
-    case ResolutionError.UndefinedTag(name, ap, _, loc) if overlaps(range, loc) =>
-      mkUseTag(name, uri, ap) ++ mkQualifyTag(name, uri, loc)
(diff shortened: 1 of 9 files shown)


magnus-madsen:
> The previous EnumTagCompleter is also used for SyntacticContext.Pat, I write a case to repeat this context, and we can see that UndefinedTag is also raised here, so it's safe to remove EnumTagCompleter from this SyntacticContext.Pat in this case. But I am not sure if there is another case of SyntacticContext.Pat.

I don't think so. Its just for match.

magnus-madsen:
> Unlike other completers, we don't check if the tag is public since that syntax is not valid. But maybe we should check if the enum is public? The same issue may also apply to other internal constructs like sigs or ops.

Visibility rules apply to the enum (instead of the individual cases).

Indeed this is also the same situation for sigs and ops.

magnus-madsen:
It seems this is ready for merge?

stormckey:
> It seems this is ready for merge?

I think so. And we can start a new PR to update the visibility check for sigs, ops and tags.

magnus-madsen (approved):

PR #9759, opened 2025-01-24
ci: add support for external perf server

magnus-madsen:


magnus-madsen:
@mlutze We can merge this and iterate?

A problem is that it runs on every PR (but only works with local branches).

Another problem is that even with 250 runs there are some variance. I do wonder why.

mlutze:
We should disable it for now though because it could interrupt experiments at any time.

mlutze:
Does the run.sh script get put somewhere?

magnus-madsen:
It's just in the home directory

mlutze:
Would be nice to have it under version control.

mlutze:
If you change the trigger then we can merge.

mlutze (approved):

PR #9769, opened 2025-01-27
fix: unseal SetUnification.SolverListener

mlutze:


magnus-madsen:
What's this for? And whats the long term plan for this listener?

mlutze:
My solver uses the listener to listen for progress. I don't know the long term plan, but it's the easiest way for mine to plug in.

magnus-madsen:
> My solver uses the listener to listen for progress. I don't know the long term plan, but it's the easiest way for mine to plug in.

This is exactly what I wanted to avoid.

PR #9771, opened 2025-01-27
fix: check the visibility for completion for sigs, ops and tags

stormckey:
Check the parent trait, effect and enum for visibility


magnus-madsen:
What about things that are in the same module?

stormckey:
We need to know the current namespace where we provide the completion, which is inaccessible for many errors currently. 

I think the `isInFile` check is trying to approximate that? We employ the same checking for all merged completers.

magnus-madsen:
Can you create a ticket? - so we can look into it later.

magnus-madsen (approved):

PR #9789, opened 2025-01-28
feat: introduce Soup.renew

mlutze:


main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolver2.scala
```diff
@@ -132,6 +132,17 @@ object ConstraintSolver2 {
       }
     }
 
+    /**
+      * Creates a new [[Soup]], but reuses this one if the arguments are the same.
+      */
+    private def renew(newConstrs: List[TypeConstraint2], newTree: SubstitutionTree): Soup = {
```
magnus-madsen, line 138: Why is this not called Soup.mk? Is that not the standard?

mlutze: It's a method on an object, rather than a "static method", so in that sense it's unprecedented afaik.

mlutze: I.e. we call `mysoup.renew` rather than `Soup.renew`

mlutze: I'll change if `mk` is better or if you have another suggestion

magnus-madsen: I see OK

PR #9798, opened 2025-01-28
feat: make Regions incremental

### LINKED ISSUE #9658: Add incremental support to `Regions`

stormckey:
fixes #9658
perf figure:
image

I just use cachedTyperAst as the oldRoot, as it's used both before and after Regions. What is the rule to choose the oldRoot?


magnus-madsen (approved):

magnus-madsen:
> I just use cachedTyperAst as the oldRoot, as it's used both before and after Regions. What is the rule to choose the oldRoot?

Thats OK. Just use cachedTyperAst

PR #9807, opened 2025-01-29
feat: introduce SharedContext to Instances

stormckey:
Use SharedContext to store errors.

Will apply the new api to make it incremental in the next PR.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Instances.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Instances.scala
@@ -102,18 +102,25 @@ object Instances {
   }
 
   /**
+    * Returns a Boolean to indicate whether any error is found.
+    *
     * Checks that the instance type is simple:
     * * all type variables are unique
     * * all type arguments are variables
     */
-  private def checkSimple(inst: TypedAst.Instance)(implicit sctx: SharedContext, flix: Flix): Unit = inst match {
+  private def checkSimple(inst: TypedAst.Instance)(implicit sctx: SharedContext, flix: Flix): Boolean = inst match {
     case TypedAst.Instance(_, _, _, trt, tpe, _, _, _, _, _) => tpe match {
-      case _: Type.Cst => ()
-      case _: Type.Var => sctx.errors.add(InstanceError.ComplexInstance(tpe, trt.sym, trt.loc))
+      case _: Type.Cst => false
+      case _: Type.Var =>
+        sctx.errors.add(InstanceError.ComplexInstance(tpe, trt.sym, trt.loc))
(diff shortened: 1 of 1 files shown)


main/src/ca/uwaterloo/flix/language/phase/Instances.scala
```diff
@@ -218,61 +221,58 @@ object Instances {
...
-        // Case 2: Errors. Skip other checks.
-        case errs => errs
-      }
+    insts0.foreach {
+      // check that the instance is on a valid type
+      case inst =>
+        checkSimple(inst)
+        checkOverlap(inst, heads)
+        checkInstance(inst, root, changeSet)
+        heads += (unsafeGetHead(inst) -> inst)
```
stormckey, line 275: The semantic here is changed. Since checkSimple will no longer return a List, we can not rely on that to decide if we should do more checks.

If we want to preserve that semantic, we need to change the return type of checkSimple to something like a Boolean.

stormckey: Ok, it seems that we have to preserve this semantic, or some tests will fail.

mlutze: I think the reason we had the dependent checks is the `unsafeGetHead` function. It will probably crash if we have a complex instance (e.g. `instance Eq[a]`).

Maybe it's easiest if the `getHead` function just returns an option and we match on that?

stormckey: So we can just use inst.tpe.typeConstructor to get the option. And it do eliminates most of the errors!

But there is still one test :
```
  test("Test.ComplexErrorSuppressesOtherErrors.01") {
    val input =
      """
        |trait C[a] {
        |  pub def f(x: a): Bool
        |}
        |
        |instance C[Box[String]] {
        |  pub def g(_x: Box[String]): Bool = false
        |}
        |
        |enum Box[a] {
        |    case Box(a)
        |}
        |""".stripMargin
    expectError[InstanceError.ComplexInstance](result)
    rejectError[InstanceError.ExtraneousDef](result)
  }
```
which is used to test if the error is suppressed. We may want to delete this test if we match over the option and do not suppress errors.

magnus-madsen:
@mlutze If you could do a detailed view first and then I can come in.

mlutze (approved):
Looks reasonable.

I suggest we invert the Boolean in `checkSimple` to return `false` on error, so that it corresponds to the question "Is it simple?"

magnus-madsen:
@mlutze If you are happy then you can merge :)

PR #9811, opened 2025-01-29
feat: require parentheses for datalog-if-guard

stormckey:


JonathanStarup:
Can you check the error message manually for a program like `def example(): Unit = A. // trailing space` and `def example(): Unit = A. B`

magnus-madsen:
Can you also create a PR that updates any relevant examples on: https://github.com/flix/book -- thanks!

stormckey:
They are 
image
and 
image

magnus-madsen:
@JonathanStarup Do we agree this is the right solution/best outcome?

JonathanStarup:
I think so yes, but the reason I wanted to see the errors is that you can see that it says `Expected <Name>, .. before '.'` which is a weird message. It should be something like `Unexpected trailing '.'`. Or if we reuse an error maybe `Expected '.' but found '. '`

stormckey:
> I think so yes, but the reason I wanted to see the errors is that you can see that it says `Expected <Name>, .. before '.'` which is a weird message. It should be something like `Unexpected trailing '.'`. Or if we reuse an error maybe `Expected '.' but found '. '`

You're right, the message makes no sense. But it's us that impl it this way:
```scala
        case TokenKind.DotWhiteSpace =>
          // Trailing dot, stop parsing the qualified name.
          val error = UnexpectedToken(
            expected = NamedTokenSet.FromKinds(kinds),
            actual = Some(TokenKind.Dot),
            sctx = context,
            hint = None,
            loc = currentSourceLocation()
          )
          advanceWithError(error)
          continue = false
        case _ => continue = false
```

The UnexpectedToken here emits the error message of ParseError.

magnus-madsen (approved):

PR #9815, opened 2025-01-30
refactor: simplify `SetFormula.mkInter2` and `SetFormula.mkInter3`

magnus-madsen:
I also deleted some old generator code that is no longer compatible.


mlutze (approved):
Jonathan will be sad.

PR #9821, opened 2025-01-30
feat: use endsWithDot from QName

stormckey:


magnus-madsen:
Does everything work as well as before?

stormckey:
> Does everything work as well as before?

yeah, the logic is the same and the completions work fine.

magnus-madsen (approved):

PR #9832, opened 2025-01-31
refactor: introduce EqualityEnv

mlutze:


magnus-madsen:
@mlutze What is this? Also, I see some failure.

mlutze:
We have a fast lookup for instances, where we use (Trait * TypeHead) as a key. This introduces the same for associated types.

magnus-madsen (approved):

PR #9836, opened 2025-01-31
refactor: introduce Type.Apply.renew

mlutze:


main/src/ca/uwaterloo/flix/language/ast/Type.scala
```diff
@@ -590,6 +586,17 @@ object Type {
     }
 
     override def hashCode(): Int = Objects.hash(tpe1, tpe2)
+
+    /**
+      * Creates a new Type.Apply, reusing this one if they are equivalent.
+      */
+    def renew(newTpe1: Type, newTpe2: Type, newLoc: SourceLocation): Type.Apply = {
```
magnus-madsen, line 593: How is it obvious that "renew" and "apply" are related?

mlutze: You mean `Type.Apply`?

This function is only defined on `Type.Apply`, not on `Type`.

magnus-madsen: > You mean `Type.Apply`?
> 
> This function is only defined on `Type.Apply`, not on `Type`.

Smart. I was tired.

magnus-madsen (approved):

PR #9837, opened 2025-01-31
feat: identify trailing dot in QName

stormckey:
Add a new field `trailingDot` to QName to indicate if there is a trailing dot.

Note:
- I add a new TreeKind named TrailingDot and tryPick that to check
- I add `trailingDot` as the last field so that we can set a default value for it

I'll start using this field in the following PR. But some preliminary tests show that it works fine.

I am not sure if it's possible to process this in another way. When detecting the existence of a trailing dot. we just use all the idents as the namespace and use an empty string as ident.

The benefit is:
- Currently `use AMod.BMod.` will not trigger a resolution error since the QName is AMod as namespace and BMod as ident, which is a valid use. We have to manually check if there is a trailing dot when resolving. However, if we have [AMod, BMod] as namespace and "" as Ident, the resolve will fail and we do not need to check trailing dot manually.
- In completers, we are just using trailingDot to move ident to namespace for further usage(transform ([AMod], BMod) to ([AMod, BMod], "")). And we have to call the function to extract that namespace and ident in every completer, which is tedious. With this change, we can just use the namespace and ident from QName directly.
- Since trailing dot is only possible when the user is writing the code. The change over trailing dot will not break existing code.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/Name.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/Name.scala
@@ -157,11 +157,12 @@ object Name {
   /**
     * Qualified Name.
     *
-    * @param namespace the namespace
-    * @param ident     the identifier.
-    * @param loc       the source location of the qualified name.
+    * @param namespace    the namespace
+    * @param ident        the identifier.
+    * @param loc          the source location of the qualified name.
+    * @param trailingDot  `true` if the qualified name ends with a dot.
     */
-  case class QName(namespace: NName, ident: Ident, loc: SourceLocation) {
+  case class QName(namespace: NName, ident: Ident, loc: SourceLocation, trailingDot: Boolean = false) {
     /**
       * Returns `true` if this name is unqualified (i.e. has no namespace).
       */
--- a/main/src/ca/uwaterloo/flix/language/ast/Name.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/Name.scala
@@ -160,7 +160,7 @@ object Name {
     * @param namespace    the namespace
     * @param ident        the identifier.
     * @param loc          the source location of the qualified name.
-    * @param trailingDot  `true` if the qualified name ends with a dot.
+    * @param trailingDot  `true` if the qualified name ends with a dot (which implies this name is incomplete)
     */
   case class QName(namespace: NName, ident: Ident, loc: SourceLocation, trailingDot: Boolean = false) {
     /**
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -699,12 +699,15 @@ object Parser2 {
             // Trailing dot, stop parsing the qualified name.
             val error = UnexpectedToken(
               expected = NamedTokenSet.FromKinds(kinds),
-              actual = Some(TokenKind.Dot),
+              actual = None,
               sctx = context,
-              hint = None,
+              hint = Some("Expect something after '.'"),
               loc = currentSourceLocation()
             )
-            advanceWithError(error)
+            val mark = open()
+            advance()
+            close(mark, TreeKind.TrailingDot)
+            s.errors.append(error)
             continue = false
           } else {
             advance() // Eat the dot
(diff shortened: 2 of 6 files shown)


main/src/ca/uwaterloo/flix/language/ast/Name.scala
```diff
@@ -157,11 +157,12 @@ object Name {
   /**
     * Qualified Name.
     *
-    * @param namespace the namespace
-    * @param ident     the identifier.
-    * @param loc       the source location of the qualified name.
+    * @param namespace    the namespace
+    * @param ident        the identifier.
+    * @param loc          the source location of the qualified name.
+    * @param trailingDot  `true` if the qualified name ends with a dot.
```
magnus-madsen, line 163: -> `true` if the qualified name ends with a dot (which implies this name is incomplete).

main/src/ca/uwaterloo/flix/language/ast/Name.scala
```diff
@@ -157,11 +157,12 @@ object Name {
...
     */
-  case class QName(namespace: NName, ident: Ident, loc: SourceLocation) {
+  case class QName(namespace: NName, ident: Ident, loc: SourceLocation, trailingDot: Boolean = false) {
     /**
       * Returns `true` if this name is unqualified (i.e. has no namespace).
       */
```
magnus-madsen, line 168: Can we delete/update endsWithDot below now? Then I can experiment with the PR

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -699,12 +699,15 @@ object Parser2 {
             // Trailing dot, stop parsing the qualified name.
             val error = UnexpectedToken(
               expected = NamedTokenSet.FromKinds(kinds),
-              actual = Some(TokenKind.Dot),
+              actual = None,
               sctx = context,
-              hint = None,
+              hint = Some("Expect something after '.'"),
```
magnus-madsen, line 704: Expect ident? or something more precise we can say?

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -717,12 +720,15 @@ object Parser2 {
           // Trailing dot, stop parsing the qualified name.
           val error = UnexpectedToken(
             expected = NamedTokenSet.FromKinds(kinds),
-            actual = Some(TokenKind.Dot),
+            actual = None,
             sctx = context,
-            hint = None,
+            hint = Some("Expect something after '.'"),
```
magnus-madsen, line 725: ident?

magnus-madsen:
Looks reasonable. I would like to play with it.

Thoughts @JonathanStarup?

stormckey:
I've replaced the usage of `endsWithDot` to `trailingDot` so that you can test with it.

If it works fine we can just delete `endsWithDot` in this PR. (Or in the following ones)

magnus-madsen:
I suggest we test a bit to see.

magnus-madsen:
!!!jar

magnus-madsen:
I am working with Jonathans talpin repo. In `EffectSolver.flix`:

I am confused why this does not give me any suggestions:

image

stormckey:
Do you use ***.Subst and then try to complete over Subst?

That is a known problem I am about to solve.

In simple terms, when providing completions for qualified name, we now only checks if it matches with the fully qualified one (***.Subst.compose)

magnus-madsen:
> Do you use ***.Subst and then try to complete over Subst?

My bad. I somehow thought it was a global name, but it is not!

magnus-madsen:
I will experiment a bit more and get back. Also want to hear from @JonathanStarup

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -911,13 +911,14 @@ object Weeder2 {
...
       val idents = pickAll(TreeKind.Ident, tree).map(tokenToIdent)
+      val trailingDot = tryPick(TreeKind.TrailingDot, tree).nonEmpty
       assert(idents.nonEmpty) // Require at least one ident
       val first = idents.head
       val ident = idents.last
       val nnameIdents = idents.dropRight(1)
       val loc = SourceLocation(isReal = true, first.loc.sp1, ident.loc.sp2)
       val nname = Name.NName(nnameIdents, loc)
-      val qname = Name.QName(nname, ident, loc)
+      val qname = Name.QName(nname, ident, loc, trailingDot)
```
JonathanStarup, line 921: As you were saying, we could also here just produce `["A", "B"], ""` it feels like that would make the other code more consistent (i.e. not have the boolean on the qname)

magnus-madsen (approved):

PR #9841, opened 2025-01-31
chore: remove Regions phase

mlutze:


magnus-madsen:
The checks are handled within the typer?

magnus-madsen (approved):

mlutze:
> The checks are handled within the typer?

Yes, escaping corresponds to unifying across region boundaries.

PR #9844, opened 2025-01-31
feat: support partially qualified name for EnumTagCompleter

stormckey:
The first PR try to support partially qualified name(`Clr.Green` below, if we call it this way)

Preview:
image

The general procedure would be:
- find if the namespace(`Clr`) is resolved in the local scope
  - If yes, find the fully qualified name(`AMod.BMod.Clr`) of the namespace 
  - Then use the fully qualified name to find the parent construct from root
  - provide partially qualified completions

Actually we can find the construct in the LocalScope directly. But the construct stored there is from former phases(like NamedAst, ResolvedAst), so we will use the map to get the newest version and use it to provide completions.

The parent construct here is Enum, other possible construct could be Sig, Effect and Module.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/EnumTagCompleter.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/EnumTagCompleter.scala
@@ -17,7 +17,7 @@
 package ca.uwaterloo.flix.api.lsp.provider.completion
 
 import ca.uwaterloo.flix.api.lsp.provider.completion.Completion.EnumTagCompletion
-import ca.uwaterloo.flix.language.ast.NamedAst.Declaration.Case
+import ca.uwaterloo.flix.language.ast.NamedAst.Declaration.{Case, Enum}
 import ca.uwaterloo.flix.language.ast.shared.{AnchorPosition, LocalScope, Resolution}
 import ca.uwaterloo.flix.language.ast.TypedAst
 import ca.uwaterloo.flix.language.errors.ResolutionError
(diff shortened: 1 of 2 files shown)


main/src/ca/uwaterloo/flix/api/lsp/provider/completion/EnumTagCompleter.scala
```diff
@@ -39,21 +39,45 @@ object EnumTagCompleter {
...
       root.enums.values.flatMap(enm =>
         enm.cases.values.collect{
           case tag if matchesTag(enm, tag, namespace, ident, uri, qualified = false) =>
-            EnumTagCompletion(tag, ap, qualified = false, inScope = inScope(tag, env))
+            EnumTagCompletion(tag, Nil, ap, qualified = false, inScope = inScope(tag, env))
         }
       )
   }
 
+  private def fullyQualifiedCompletion(uri: String, ap: AnchorPosition, namespace: List[String], ident: String)(implicit root: TypedAst.Root): Iterable[Completion] = {
```
magnus-madsen, line 52: Docs?

magnus-madsen: Is namespace here the fully-qualified name or something else?

magnus-madsen: Could we use Name.NName?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/EnumTagCompleter.scala
```diff
@@ -39,21 +39,45 @@ object EnumTagCompleter {
...
+  private def partiallyQualifiedCompletions(uri: String, ap: AnchorPosition, env: LocalScope, namespace: List[String], ident: String)(implicit root: TypedAst.Root): Iterable[Completion] = {
+    env.m.getOrElse(namespace.mkString("."), Nil).collect {
+      case Resolution.Declaration(Enum(_, _, _, enumSym, _, _, _, _)) =>
+          root.enums.get(enumSym).map(enm =>
+            enm.cases.values.collect {
+              case tag if matchesTag(enm, tag, namespace, ident, uri, qualified = false) =>
+                EnumTagCompletion(tag, namespace, ap, qualified = true, inScope = true)
+            }
+          ).getOrElse(Nil)
+    }.flatten
```
magnus-madsen, line 78: I find use of both getOrElse and flatten to be a bit of code smell. Is there no better way? 

Perhaps constructing a mutable.ArratBuffer would even be preferable at this point? 

Or the use of flatMap

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/EnumTagCompleter.scala
stormckey: No, we can't. As said in #9837. Here we build the `namespace: List[String]` but looking at the boolean in QName and try to move the ident to the namespace.
If we, as suggested, conduct this moving then identifying a trailing dot, instead of using a boolean to indicate this, then we can just access this namespace from `qn.namespace`.

Maybe I should first change QName this way in another PR and then rebase to use that. The logic here will be more natural :)

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/EnumTagCompleter.scala
stormckey: We can, transform the options to list first and then call flatMap consistently, the code would be:
```scala
    env.m.getOrElse(namespace.mkString("."), Nil).flatMap {
      case Resolution.Declaration(Enum(_, _, _, enumSym, _, _, _, _)) =>
          root.enums.get(enumSym).toList.flatMap(enm =>
            enm.cases.values.collect {
              case tag if matchesTag(enm, tag, namespace, ident, uri, qualified = false) =>
                EnumTagCompletion(tag, namespace, ap, qualified = true, inScope = true)
            }
          )
      case _ => Nil
    }
```

stormckey: Or we can use for comprehension:
```scala
    for {
      resolution <- env.m.get(namespace.mkString(".")).toList
      case Resolution.Declaration(Enum(_, _, _, enumSym, _, _, _, _)) <- resolution
      enm <- root.enums.get(enumSym).toList
      tag <- enm.cases.values
      if matchesTag(enm, tag, namespace, ident, uri, qualified = false)
    } yield EnumTagCompletion(tag, namespace, ap, qualified = true, inScope = true)
```

stormckey:
I prefer the for-yield version and have pushed that :) 

The PR to move ident to namespace is here #9859, with which we can pass QName around instead of List[String]

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/EnumTagCompleter.scala
magnus-madsen: The for-yield looks nice, what do you think?

magnus-madsen (approved):

PR #9859, opened 2025-02-04
feat: move ident to namespace for QName when there is a trailing dot

stormckey:
Move the last ident to namespace when constructing the QName

Prev:
`A.B.` will be QName(["A"], "B", trailingDot = true)

Now:
`A.B.` will be QName(["A", "B"], "")

And the completer works fine after the change:
image
image
image

Note:
- I find the code duplicate in visitQNameExpr and visitQName, and call the latter form the former.
- Next step would be remove `getNamespaceAndIdentFromQName` from all call site. And just pass the qname for completers to use. That will change a lot of code so I left it for the following PR.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/Name.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/Name.scala
@@ -160,6 +160,12 @@ object Name {
     * @param namespace    the namespace
     * @param ident        the identifier.
     * @param loc          the source location of the qualified name.
+    *
+    * Note that the ident could be empty if there is a trailing dot.
+    *
+    * Example:
+    *   - "A.B.Color" -> namespace = ["A", "B"], ident = "Color"
+    *   - "A.B." -> namespace = ["A", "B"], ident = ""
     */
   case class QName(namespace: NName, ident: Ident, loc: SourceLocation) {
     /**
--- a/main/src/ca/uwaterloo/flix/language/ast/Name.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/Name.scala
@@ -160,45 +160,13 @@ object Name {
     * @param namespace    the namespace
     * @param ident        the identifier.
     * @param loc          the source location of the qualified name.
-    * @param trailingDot  `true` if the qualified name ends with a dot (which implies this name is incomplete)
     */
-  case class QName(namespace: NName, ident: Ident, loc: SourceLocation, trailingDot: Boolean = false) {
+  case class QName(namespace: NName, ident: Ident, loc: SourceLocation) {
     /**
       * Returns `true` if this name is unqualified (i.e. has no namespace).
       */
     def isUnqualified: Boolean = namespace.isRoot
 
-    /**
-     * Returns `true` if `this` qualified name is incomplete, i.e. it ends with a dot.
-     *
-     * May return `false` out of an over abundance of caution.
-     *
-     * Note: In some cases this function may give false positives. For example, in:
-     *
-     * {{{
-     *   x.byteValueExact()
-     * }}}
-     *
-     * We report that the QName `x` ends with a dot.
-     */
-    def endsWithDot: Boolean = {
-      // We return false if this source location is unknown.
-      // This should not happen for code that the programmer is writing.
-      if (loc == SourceLocation.Unknown)
-        false
-      else {
-        val lineNumber = loc.sp2.line
-        val columnOffset = loc.sp2.col - 1
-        val line = loc.sp2.source.getLine(lineNumber)
-        if (!(columnOffset < line.length)) {
-          // Out of bounds; return false.
-          false
-        } else {
-          // Within bounds; check if the qname ends with a dot.
-          '.'== line.charAt(columnOffset)
-        }
-      }
-    }
 
     /**
       * Human readable representation.
--- a/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
@@ -909,16 +909,7 @@ object Weeder2 {
     }
 
     private def visitQnameExpr(tree: Tree)(implicit sctx: SharedContext): Expr.Ambiguous = {
-      expect(tree, TreeKind.QName)
-      val idents = pickAll(TreeKind.Ident, tree).map(tokenToIdent)
-      val trailingDot = tryPick(TreeKind.TrailingDot, tree).nonEmpty
-      assert(idents.nonEmpty) // Require at least one ident
-      val first = idents.head
-      val ident = idents.last
-      val nnameIdents = idents.dropRight(1)
-      val loc = SourceLocation(isReal = true, first.loc.sp1, ident.loc.sp2)
-      val nname = Name.NName(nnameIdents, loc)
-      val qname = Name.QName(nname, ident, loc, trailingDot)
+      val qname = visitQName(tree)
       Expr.Ambiguous(qname, qname.loc)
     }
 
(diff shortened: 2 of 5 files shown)


stormckey:
There are several reasons:
- A trailing dot will only occur when the user is coding, so no existing code will break due to this change to QName
- For situations like `use A.B.` the QName(["A", "B"], "") will make the resolver fail and produce the error directly(exactly what we want). Otherwise we need to check the boolean manually(and in any similar situations).
- These function calls are used to move the ident to namespace if there is a trailing dot. 
  image
  And currently we have to call it after every `err.qn`. This change will facilitate this part.
- I think it's intuitive. A and B are both namespace in `A.B.` and we are waiting for the user to type the ident inside B. Some checks we are performing(like `matchesQualifiedName`, which will require the namespace to be identical and the ident to be fuzzyMatched) treat ident and namespace differently.

magnus-madsen:
OK; we should discuss further with @mlutze

JonathanStarup:
I am a fan of this change, seems more sensible than the boolean

mlutze:
This seems reasonable to me.

magnus-madsen:
OK; lets try to meet to discuss in person. I suppose I will be convinced then :)

stormckey:
I can answer it all:
- The source location of the ident will be a zero width one with both the start and end being `idents.last.loc.sp2"
- In the current impl, it will include the trailing dot. But we could include that.
- No, the parser will stop parsing the QName once the token after Dot is not what we expect or it meet a DotWithSpace. Also, when constructing the QName, ident will not include any empty string and we will not append two empty string to it.
- The parser will emit the error, and it will also push all the idents and a TreeKind.TrailingDot to the tree. You can see the first error message in the following screenshot, which is emitted by the parser.
- The resolution error looks like:
  image

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -3080,11 +3071,20 @@ object Weeder2 {
...
-    val nnameIdents = idents.dropRight(1)
-    val loc = SourceLocation(isReal = true, first.loc.sp1, ident.loc.sp2)
-    val nname = Name.NName(nnameIdents, loc)
-    Name.QName(nname, ident, loc, trailingDot)
+    val last = idents.last
+    val loc = SourceLocation(isReal = true, first.loc.sp1, last.loc.sp2)
+
+    // If there is a trailing dot, we use all the idents as namespace and use "" as the ident
+    if (trailingDot) {
+      val nname = Name.NName(idents, loc)
```
magnus-madsen, line 3079: Can we add an example here in a comment. To explain how the generated QName looks like.

magnus-madsen:
@mlutze @JonathanStarup If you believe this is reasonable and safe then we can merge.

magnus-madsen:
@mlutze In particular, you are OK with the Resolver doing lookups on ["A", "B", ""] and producing error messages for that?

stormckey:
Should we include the trailing dot by moving the `qn.sp2.col` one char forward?
And should we move the zero-width ident one char forward to make it follow the dot?

mlutze:
I think the resolution error seems OK.



> Should we include the trailing dot by moving the `qn.sp2.col` one char forward? And should we move the zero-width ident one char forward to make it follow the dot?

Yes, the location should match the whole thing IMO

JonathanStarup:
I dont think the position should be changed in the weeder. The parser tree location should be corrected instead

stormckey:
> I dont think the position should be changed in the weeder. The parser tree location should be corrected instead

But the parser will only build locations for idents. Weeder is the phase where we build locations for QName etc.

But it inspired me with another impl. Instead of inserting a dedicated Treekind.TrailingDot, we can just insert another empty indent to the tree and the Weeder could just work the way it work before.

magnus-madsen:
> > I dont think the position should be changed in the weeder. The parser tree location should be corrected instead
> 
> But the parser will only build locations for idents. Weeder is the phase where we build locations for QName etc.
> 
> But it inspired me with another impl. Instead of inserting a dedicated Treekind.TrailingDot, we can just insert another empty indent to the tree and the Weeder could just work the way it work before.

I think this is dangerous in the sense that the code may look clean, but the next person will be utterly confused. So I would keep it as is.

magnus-madsen:
> I think the resolution error seems OK.
> 
> > Should we include the trailing dot by moving the `qn.sp2.col` one char forward? And should we move the zero-width ident one char forward to make it follow the dot?
> 
> Yes, the location should match the whole thing IMO

The qname loc must span the whole thing.

stormckey:
> > I think the resolution error seems OK.
> > > Should we include the trailing dot by moving the `qn.sp2.col` one char forward? And should we move the zero-width ident one char forward to make it follow the dot?
> > 
> > 
> > Yes, the location should match the whole thing IMO
> 
> The qname loc must span the whole thing.

Now it works as expected, you can see that the dot is underlined.
image

JonathanStarup:
> But the parser will only build locations for idents. Weeder is the phase where we build locations for QName etc.

But parser assigns sourcelocation to each tree (`tree.loc` in Weeder) and the tree location encompasses its children.

I think the issue is here in parser2
```
val mark = open()
close(mark, TreeKind.TrailingDot)
advanceWithError(error)
```
since TrailingDot is just opened and closed, it doesn't "contain" the dot. `A.B.` makes `A.B` a QName thing with a zero-width dot and the last `.` an Error node.

So I think it should be something like:
```
val mark = open()
advance()
close(mark, TreeKind.TrailingDot)
s.errors.append(..)
```

main/src/ca/uwaterloo/flix/language/ast/Name.scala
```diff
@@ -160,45 +160,13 @@ object Name {
     * @param namespace    the namespace
     * @param ident        the identifier.
     * @param loc          the source location of the qualified name.
-    * @param trailingDot  `true` if the qualified name ends with a dot (which implies this name is incomplete)
     */
-  case class QName(namespace: NName, ident: Ident, loc: SourceLocation, trailingDot: Boolean = false) {
```
magnus-madsen, line 165: Lets a sentence to the comment that explain that the ident may be the empty string if there is a trailing dot.

Actually one or two examples would be good too.

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -3080,11 +3071,23 @@ object Weeder2 {
...
+    // If there is a trailing dot, we use all the idents as namespace and use "" as the ident
+    // The resulting QName will be something like QName(["A", "B"], "")
+    if (trailingDot) {
+      val nname = Name.NName(idents, loc)
+      val positionAfterDot = last.loc.sp2.copy(col = (last.loc.sp2.col + 1).toShort)
+      val emptyIdentLoc = SourceLocation(isReal = true, positionAfterDot, positionAfterDot)
+      val emptyIdent = Name.Ident("", emptyIdentLoc)
+      val qnameLoc = SourceLocation(isReal = true, first.loc.sp1, positionAfterDot)
+      Name.QName(nname, emptyIdent, qnameLoc)
+      // Otherwise we use all but the last ident as namespace and the last ident as the ident
```
magnus-madsen, line 3086: This should go inside the else not before it

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -3080,11 +3071,23 @@ object Weeder2 {
...
+    } else{
```
magnus-madsen, line 3087: missing space else{

magnus-madsen:
I am ready to merge this then. But first I want to some more documentation on QName.

(And there are a few other small things to fix.)

magnus-madsen (approved):

PR #9863, opened 2025-02-04
feat: make instances incremental

stormckey:
perf figure:
image
The ratio is super low, and it mat vary from 0 - 0.2 between different runs. Does that indicates a problem?

The check here must be performed over the list of instances, while the former `ParOps.parMapValueList` is performed over a single instance. I reimpl an api to map over lists. But I dont know how to give it a proper name, now it's called `parMapValueList2`

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Instances.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Instances.scala
@@ -41,18 +41,13 @@ object Instances {
   def run(root: TypedAst.Root, oldRoot: TypedAst.Root, changeSet: ChangeSet)(implicit flix: Flix): (TypedAst.Root, List[InstanceError]) = {
     implicit val sctx: SharedContext = SharedContext.mk()
     flix.phaseNew("Instances") {
-      visitInstances(root, oldRoot, changeSet)
-      visitTraits(root)
-      (root, sctx.errors.asScala.toList)
+      val instances = changeSet.updateStaleValueLists(root.instances, oldRoot.instances)(ParOps.parMapValueList2(_)(checkInstancesOfTrait(_, root)))
+      val traits = chaneSet.updateStaleValues(root.traits, oldRoot.traits)(ParOps.parMapValues(_)(visitTrait))
+      (root.copy(instances = instances, traits = traits), sctx.errors.asScala.toList)
     }
   }
 
 
-  /**
-    * Validates all instances in the given AST root.
-    */
-  private def visitTraits(root: TypedAst.Root)(implicit sctx: SharedContext, flix: Flix): Unit =
-    ParOps.parMap(root.traits.values)(visitTrait)
 
   /**
     * Checks that all signatures in `trait0` are used in laws if `trait0` is marked `lawful`.
@@ -74,16 +69,9 @@ object Instances {
   /**
     * Performs validations on a single trait.
     */
-  private def visitTrait(trait0: TypedAst.Trait)(implicit sctx: SharedContext): Unit = {
+  private def visitTrait(trait0: TypedAst.Trait)(implicit sctx: SharedContext): TypedAst.Trait = {
     checkLawApplication(trait0)
-  }
-
-  /**
-    * Validates all instances in the given AST root.
-    */
-  private def visitInstances(root: TypedAst.Root, oldRoot: TypedAst.Root, changeSet: ChangeSet)(implicit sctx: SharedContext, flix: Flix): Unit = {
-    // Check the instances of each trait in parallel.
-    ParOps.parMap(root.instances.valueLists)(checkInstancesOfTrait(_, root, changeSet))
+    trait0
   }
 
   /**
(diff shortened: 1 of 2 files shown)


magnus-madsen:
@mlutze Any thoughts on the incrementality of instances? I would think that the phase should do fine

mlutze:
Can we check what percent of instances are stale/fresh?

Since instances are more complex when it comes to dependency, maybe the dependency phase marks the conservatively?

magnus-madsen:
I suppose some numbers would be useful.

stormckey:
Alright, here are the numbers:
image
As first everything is stale. Later, traits become fresh, while most instances are still stale:
image
The perf figure of this run:
image

stormckey:
Here is the code in Instances.scala for your reference:
```scala
      val (staleInstances, freshInstances) = changeSet.partitionOnValues(root.instances, oldRoot.instances)
      val instances = freshInstances ++ ParOps.parMapValueList2(staleInstances)(checkInstancesOfTrait(_, root))
      val (staleTraits, freshTraits) = changeSet.partition(root.traits, oldRoot.traits)
      val traits = freshTraits ++ ParOps.parMapValues(staleTraits)(visitTrait)
      println (s"stale instance ratio: ${staleInstances.valueLists.map(_.size).sum} / ${root.instances.valueLists.map(_.size).sum}")
      println(s"stale trait ratio: ${staleTraits.size} / ${root.traits.size}")
```

magnus-madsen:
What's the ratio? Is it good that its high? or is that bad?

stormckey:
> What's the ratio? Is it good that its high? or is that bad?

Bad, it's too conservative and treats most instances stale. Meaning that we will revisit them.

magnus-madsen:
> > What's the ratio? Is it good that its high? or is that bad?
> 
> Bad, it's too conservative and treats most instances stale. Meaning that we will revisit them.

Do we know why?

Like if you have a file `foo.Flix` and you change it, does it cause recompilation of all of these instances?

stormckey:
> > > What's the ratio? Is it good that its high? or is that bad?
> > 
> > 
> > Bad, it's too conservative and treats most instances stale. Meaning that we will revisit them.
> 
> Do we know why?
> 
> Like if you have a file `Foo.flix` and you change it, does it cause recompilation of all of these instances?

Mathew is looking into the staleInstances. Maybe we can wait for news from him :)

mlutze:
I think I have found the issue. 

This is a function in `ChangeSet.scala`

```scala
  def partitionOnValues[K, V <: Sourceable](newMap: ListMap[K, V], oldMap: ListMap[K, V]): (ListMap[K, V], ListMap[K, V]) = this match {
    case ChangeSet.Everything =>
      (newMap, ListMap.empty)

    case ChangeSet.Dirty(dirty) =>
      newMap.foldLeft((ListMap.empty[K, V], ListMap.empty[K, V])){ case ((stale, fresh), (k, v)) =>
        if (oldMap.get(k).contains(v) && !dirty.contains(v.src.input))
          (stale, fresh + (k -> v))
        else
          (stale + (k -> v), fresh)
      }
  }
```

The issue is the `.contains(v)` call. The values in the maps are instances, which don't have equality properly defined on them, so the `contains` call almost always returns false.

mlutze:
> The default equality check for a case class is the equality of all fields, right? And that failed most of the time for instance? What makes them different?

I'm not sure exactly which field is causing the difference, since in principle they should be identical to the last phase. But we shouldn't be comparing entire AST nodes anyway. 

> We use `contains` in partition and partitionOnValues, and I think equality is not defined for most constructs including instance. Should we add them for all constructs? How can we determine the equality?

In partition, we only use `contains` on the key. That's ok because the key is always a `Symbol` with well-defined equality/hashcode.


One possible solution to this is to use the types of the instances to do the partition, since the types should be unique and (I think) have well-defined equality.

stormckey:
> In partition, we only use `contains` on the key. That's ok because the key is always a `Symbol` with well-defined equality/hashcode.

You are right! But we can not just compare the keys in the ListMap sine the key could point to both fresh and stale items.

> One possible solution to this is to use the types of the instances to do the partition, since the types should be unique and (I think) have well-defined equality.

So that could be a general problem for partitionOnValues. But since this function is polymorphic(although it's only called over instances), we cannot just take the type field out of the value.

stormckey:
> > So that could be a general problem for partitionOnValues. But since this function is polymorphic(although it's only called over instances), we cannot just take the type field out of the value.
> 
> Yes, good point. Could it take a "idExtractor" function to pull the type out of the value?

OK.

stormckey:
Well it's much better:
image
image

But still not ideal.

stormckey:
The impl is:
image
image

mlutze:
I think it's because of type variables. They get a new ID in every compilation...

stormckey:
> I think it's because of type variables. They get a new ID in every compilation...

Okay, so it's actually changed. What should we use then?

stormckey:
By the way, the discussion is a fix to `partitionOnValues`, I think we can merge the PR and then fix the `partitionOnValues` problem in another one. @magnus-madsen

stormckey:
> > By the way, the discussion is a fix to `partitionOnValues`, I think we can merge the PR and then fix the `partitionOnValues` problem in another one. @magnus-madsen
> 
> You mean to merge this PR, and then open another? I can do that, but we should probably also open a ticket.

That's fine.

magnus-madsen (approved):

PR #9867, opened 2025-02-04
feat: support semantic tokens for standalone lsp

stormckey:
Preview:
image

Note that variables are also highlighted, but their color scheme here is blank:
image

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
@@ -15,7 +15,7 @@
  */
 package ca.uwaterloo.flix.api.lsp
 
-import ca.uwaterloo.flix.api.lsp.provider.{HighlightProvider, HoverProvider}
+import ca.uwaterloo.flix.api.lsp.provider.{HighlightProvider, HoverProvider, SemanticTokensProvider}
 import ca.uwaterloo.flix.api.{CrashHandler, Flix}
 import ca.uwaterloo.flix.api.lsp.{Position, PublishDiagnosticsParams}
 import ca.uwaterloo.flix.language.CompilationMessage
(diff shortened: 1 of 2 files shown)


magnus-madsen:
OK, I still feels like some colors are missing. But I could be wrong.

Can you paste another screenshot?

magnus-madsen:
I suppose next steps could be to do more coloring inside LSP (e.g. literals etc.)

We cannot (yet) do keywords because they are lost.

stormckey:
Some other screenshots:
image
image
image

stormckey:
> I suppose next steps could be to do more coloring inside LSP (e.g. literals etc.)
> 
> We cannot (yet) do keywords because they are lost.

I think for flix in VSCode we are using TextMate to provide syntactic highlight? In nvim this seems to be done by treesitter, which employs a different language.

main/src/ca/uwaterloo/flix/api/lsp/VSCodeLspServer.scala
```diff
@@ -292,6 +292,7 @@ class VSCodeLspServer(port: Int, o: Options) extends WebSocketServer(new InetSoc
       ("id" -> id) ~ FindReferencesProvider.findRefs(uri, pos)(root)
 
     case Request.SemanticTokens(id, uri) =>
+//      ("id" -> id) ~ ("status" -> ResponseStatus.Success) ~ ("result" -> ("data" -> Nil))
```
magnus-madsen, line 295: Remove?

main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
```diff
@@ -95,6 +97,15 @@ object LspServer {
 
       val serverCapabilities = new ServerCapabilities
       serverCapabilities.setHoverProvider(true)
+      serverCapabilities.setSemanticTokensProvider(
+        new SemanticTokensWithRegistrationOptions(
+          new SemanticTokensLegend(
+            List("class", "enum", "enumMember", "function", "interface", "operator", "parameter", "property", "method", "namespace", "type", "typeParameter", "variable").asJava,
```
magnus-madsen, line 103: We have to provide this?

magnus-madsen:
I can merge this but there is a conflict.

main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
stormckey: We have to, the client will use this legend to interpret the returned int encodings.

This could also be seen in https://github.com/flix/vscode-flix/blob/cc0dc4bf281ad298685d708adf371c7c0d1f6f9b/server/src/handlers/handlers.ts#L61

stormckey:
Solved

magnus-madsen (approved):

PR #9869, opened 2025-02-05
refactor: cleanup `SetUnification`

magnus-madsen:


magnus-madsen:
@mlutze You can merge if happy.

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
```diff
@@ -114,16 +115,16 @@ object SetUnification {
...
+      case Result.Ok(s) =>
         // Success: We solved all equations using SVE.
         // Mark all equations as solved and update the substitution.
         state.eqs = Nil
         state.subst = s @@ state.subst
+      case Result.Err(eqs) =>
+        // Failure: We found a conflict.
+        // Store the unsolveable equations and reset the partial substitution.
+        state.eqs = eqs
+        state.subst = SetSubstitution.empty
```
mlutze, line 127: What is the purpose of resetting the substitution?

I.e., what is the invariant relating state.eqs, state.subst, and the input?

magnus-madsen: I cannot establish the proper invariant yet.

magnus-madsen: It will need to be a follow up: https://github.com/flix/flix/issues/9870

mlutze (approved):

main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala
JonathanStarup: in this branch the system is unsolvable so what does incorrect mean?

magnus-madsen: My current understanding is that -- somehow -- the previous simplification rules before SVE produce subst. that do not respect rigidity. But I need to confirm that.

PR #9871, opened 2025-02-05
perf: tweak `ZhegalkinCaches`

magnus-madsen:


mlutze (approved):
Are hashmap sizes supposed to be prime or something?

magnus-madsen:
> Are hashmap sizes supposed to be prime or something?

What? Why?

PR #9872, opened 2025-02-05
perf: tweak `SetSubstitution.apply`

magnus-madsen:


main/src/ca/uwaterloo/flix/util/TwoList.scala
```diff
@@ -35,6 +35,14 @@ case class TwoList[T](x: T, y: T, rest: List[T]) {
     */
   def forall(f: T => Boolean): Boolean = f(x) && f(y) && rest.forall(f)
 
+  /**
+   * Folds `f` over the elements of `this` list.
+   */
+  @inline
+  def foldRight[S](z: S)(f: (T, S) => S): S = {
```
mlutze, line 42: Is there a reason to foldRight instead of left? Left would be tail-recursive, I think.

mlutze (approved):

main/src/ca/uwaterloo/flix/util/TwoList.scala
magnus-madsen: To preserve the order of constraints

magnus-madsen:
For reasons unknown, this PR is slow.

PR #9881, opened 2025-02-05
refactor: introduce StringCursor

JonathanStarup:
Extracting parts of the lexer into StringCursor. We could and should go further but this is a good self-contained and somewhat comprehensible PR I think.

This PR moves all the tracking of line/column into StringCursor


magnus-madsen:
I notice that StringCursor is mutable, but sometimes it used as an immutable value (e.g. in restore). 

Did you consider having two abstractions?

magnus-madsen:
> This PR moves all the tracking of line/column into StringBuilder

Is this a typo or?

JonathanStarup:
> I notice that StringCursor is mutable, but sometimes it used as an immutable value (e.g. in restore).
> 
> Did you consider having two abstractions?

yeah I'm not totally sure how to support it nicely but I am also not making it top priority since IMO the lexer shouldn't backtrack

In this version StringCursor just has the clean concept of a shallow copy, while lexer contains the save/restore

main/src/ca/uwaterloo/flix/util/StringCursor.scala
```diff
@@ -0,0 +1,124 @@
...
+  def getExclusiveEndPosition: (Int, Int) = {
+    if (line <= 0 || column > 0) {
+      (line, column)
+    } else {
+      (line - 1, prevLineMaxColumn + 1)
+    }
+  }
+
+  /** Returns a copy of `this`, pointing to the same underlying array. */
+  def copy(): StringCursor = {
```
magnus-madsen, line 55: It seems the order of functions is:

getter things
advance/peek things

so I would move the copy function last.

main/src/ca/uwaterloo/flix/util/StringCursor.scala
```diff
@@ -0,0 +1,124 @@
+package ca.uwaterloo.flix.util
+
+/**
+  * A class to iterate through an array of characters while maintaining the line and column index
+  * of the cursor.
+  */
+class StringCursor(val data: Array[Char]) {
```
magnus-madsen, line 7: Move this into the end of the Lexer class, make it private, and rename it to ArrayCursor?

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -186,54 +178,26 @@ object Lexer {
...
     */
-  private def previousPrevious()(implicit s: State): Option[Char] = {
-    if (s.current.offset <= 1) {
-      None
-    } else {
-      Some(s.src.data(s.current.offset - 2))
-    }
-  }
+  private def previousPrevious()(implicit s: State): Option[Char] =
+    s.sc.previousPrevious
```
magnus-madsen, line 194: Is this good design? (same comments apply elsewhere)

main/src/ca/uwaterloo/flix/util/StringCursor.scala
JonathanStarup: You don't want a separate file? Isn't that nicer?

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
JonathanStarup: I think it previousPrevious was used for performance? you can skip one of the bound checks if when you know the index I negative. but ill delete this double one at least

main/src/ca/uwaterloo/flix/util/StringCursor.scala
magnus-madsen: No because I don't want it reused outside the Lexer.

JonathanStarup: Okay. I think StringCursor is a better name than ArrayCursor since data is interpreted as a String (I'm also not sure why Source doesn't just have a String instead of converting it into an array)

magnus-madsen (approved):

PR #9883, opened 2025-02-06
feat: add generic CofiniteSet

mlutze:
Here's another attempt at generifying CofiniteSet.

The problem before was that `empty` and `universe` were inefficient because we couldn't reuse _the_ empty set and _the_ universal set -- we had to construct a new one each `empty`/`universe` was called.

The normal Scala solution for that is to make a default `empty` case and make that `X[Nothing]`, where the type parameter is covariant. But SortedSet doesn't allow covariance.

In this PR: I introduce an implicit parameter to `empty` and `universe`, called `Cached`. The `empty` and `universe` function then reuse the single object provided by the `Cached` implementation.

### DIFF
--- a/main/src/ca/uwaterloo/flix/util/CofiniteSet.scala
+++ b/main/src/ca/uwaterloo/flix/util/CofiniteSet.scala
@@ -20,7 +20,7 @@ package ca.uwaterloo.flix.util
 import scala.collection.immutable.SortedSet
 
 /**
-  * Represents a finite or co-finite set with an infinite universe of integers.
+  * Represents a finite or co-finite set with an infinite universe.
   *
   * All sets are either a [[SortedSet]] or a complement of it.
   *
@@ -29,7 +29,6 @@ import scala.collection.immutable.SortedSet
 sealed trait CofiniteSet[T] {
 
   import CofiniteSet.{Compl, Set}
-
 
   /** Returns `true` if `this` is [[CofiniteSet.empty]]. */
   def isEmpty: Boolean = this match {
@@ -56,7 +55,12 @@ object CofiniteSet {
   /**
     * A trait that indicates the empty and universal values of this type have been cached.
     *
-    * We use this to ensure there is ONE empty and ONE universe value (per type) throughout the program.
+    * We use this for performance:
+    * There is one value for `empty` and `universe` per instance of the trait.
+    * As long as the same trait is used as evidence to the [[CofiniteSet.empty]] and [[CofiniteSet.universe]] functions,
+    * the same object will be returned every time [[CofiniteSet.empty]] or [[CofiniteSet.universe]] is called.
+    *
+    * This avoids allocation of new objects every time an empty or universal set is created.
     */
   trait Cached[T] {
     val empty: CofiniteSet[T]
@@ -68,10 +72,10 @@ object CofiniteSet {
     override val universe: CofiniteSet[Int] = Compl(SortedSet.empty)
   }
 
-  /** Represents a finite set of integers. */
+  /** Represents a finite set. */
   case class Set[T](s: SortedSet[T]) extends CofiniteSet[T]
 
-  /** Represents a co-finite set of integers. */
+  /** Represents a co-finite set. */
   case class Compl[T](s: SortedSet[T]) extends CofiniteSet[T]
 
   /** The empty set. */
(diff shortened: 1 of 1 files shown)


mlutze:
> I am concerned about boxing of integers.
> 
> Why do you need it to be generic?

Because there are two copies of it and I need to modify one of them, which is a big pain

mlutze:
> I am concerned about boxing of integers.

https://www.scala-lang.org/api/current/scala/specialized.html maybe

magnus-madsen:
> > I am concerned about boxing of integers.
> > Why do you need it to be generic?
> 
> Because there are two copies of it and I need to modify one of them, which is a big pain

What's the reason that you need to modify this? I thought it was a level below, so to speak.

Does the specialized thing work in Scala 2?

I just don't want this to mess with my current performance experiments. Also the whole point of the new BU solver is to work on integers. But perhaps CoInfset is used in two places for different purposes?

main/src/ca/uwaterloo/flix/util/CofiniteSet.scala
```diff
@@ -0,0 +1,161 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ca.uwaterloo.flix.util
+
+import scala.collection.immutable.SortedSet
+
+/**
+  * Represents a finite or co-finite set with an infinite universe of integers.
```
magnus-madsen, line 22: not integers

main/src/ca/uwaterloo/flix/util/CofiniteSet.scala
```diff
@@ -0,0 +1,161 @@
...
+  *
+  * All sets are either a [[SortedSet]] or a complement of it.
+  *
+  * No finite set is ever equivalent to universe.
+  */
+sealed trait CofiniteSet[T] {
+
+  import CofiniteSet.{Compl, Set}
+
```
magnus-madsen, line 31: extra linebreak

magnus-madsen:
OK; just polish it a bit and then I can merge it.

magnus-madsen (approved):

main/src/ca/uwaterloo/flix/util/CofiniteSet.scala
```diff
@@ -0,0 +1,161 @@
...
+  def isUniverse: Boolean = this match {
+    case Set(_) => false
+    case Compl(s) => s.isEmpty
+  }
+
+}
+
+object CofiniteSet {
+
+  trait Cached[T] {
```
magnus-madsen, line 55: Is cached the right name?

mlutze: Do you have a better idea? Here's what my friend says:

Here are several naming suggestions that might be better aligned with the purpose and context of your implicit class, which caches the empty and universe values for use in CofiniteSet. The goal is to reflect the caching aspect while keeping things clear and descriptive:

    ValueCache: Emphasizes the caching of values like empty and universe.
    SetDefaults: Focuses on the idea that this helper holds the default values for sets.
    CofiniteDefaults: Ties the class more closely to its use in CofiniteSet.
    PredefinedValues: Suggests that the values (empty and universe) are predefined.
    SetContext: Implies that this is part of a context in which these values (empty, universe) are provided.
    ValueProvider: Emphasizes the role of providing the cached values.
    SetCache: A simple and clear name that reflects its caching role specifically for sets.
    CachedSetValues: Very descriptive, indicating it's specifically caching values for a set.
    DefaultSetValues: Suggests this contains the default values for a set, which are cached.
    ImplicitSetHelper: Suggests it’s an implicit helper for set operations (with caching included).
    PrecachedSet: Focuses on the idea that the set's values are precalculated or cached.

Would any of these names suit the class, or do they inspire other ideas?

main/src/ca/uwaterloo/flix/util/CofiniteSet.scala
```diff
@@ -52,6 +53,11 @@ sealed trait CofiniteSet[T] {
 
 object CofiniteSet {
 
+  /**
+    * A trait that indicates the empty and universal values of this type have been cached.
+    *
+    * We use this to ensure there is ONE empty and ONE universe value (per type) throughout the program.
```
magnus-madsen, line 59: I does not really explain why this is important. Is it for semantics or just performance??

magnus-madsen: Also how does it guarantee it

main/src/ca/uwaterloo/flix/util/CofiniteSet.scala
mlutze: I pay for the sins of my ancestors.

PR #9886, opened 2025-02-06
feat: support code action for stand alone lsp

stormckey:
It works fine (if we ignore the broken code action)
image
image

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/Command.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/Command.scala
@@ -15,8 +15,11 @@
  */
 package ca.uwaterloo.flix.api.lsp
 
+import org.eclipse.lsp4j
 import org.json4s.JsonDSL.*
 import org.json4s.*
+
+import scala.jdk.CollectionConverters.*
 
 /**
   * Represents a `Command` in LSP.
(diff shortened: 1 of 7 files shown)


main/src/ca/uwaterloo/flix/api/lsp/Command.scala
```diff
@@ -28,4 +31,6 @@ import org.json4s.*
   */
 case class Command(title: String, command: String, arguments: List[JValue]) {
   def toJSON: JValue = ("title" -> title) ~ ("command" -> command) ~ ("arguments" -> arguments)
+
+  def toLsp4j: lsp4j.Command = new lsp4j.Command(title, command, arguments.map(_.asInstanceOf[Object]).asJava)
```
magnus-madsen, line 35: Why do we need asInstanceOfObject?

main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
```diff
@@ -66,7 +69,7 @@ object LspServer {
     /**
       * The current compilation errors.
       */
-    private var currentErrors: List[CompilationMessage] = Nil
+    var currentErrors: List[CompilationMessage] = Nil
```
magnus-madsen, line 72: Does this need to be public? Is it because its accessed from an inner class or something?

magnus-madsen (approved):

magnus-madsen:
@stormckey I am merging this but please answer my two questions. Thanks!

main/src/ca/uwaterloo/flix/api/lsp/Command.scala
stormckey: It requires a util.List[Object] while JValue is not an object itself.

main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
stormckey: flixTextDocuementService, a member of flixLanguageServer, need to access currentErrors. So I make it public. root and flix are public for the same reason, do you think there is a better way?

PR #9889, opened 2025-02-06
test: allow specifying error locations in tests

mlutze:
This change allows a test writer to specify where an error is expected.

For example, I can do this:
```
  test("TestLeq02") {
    val input =
      """
        |def foo(): List[a] = 21 :: Nil // ERROR
        |
        |enum List[t] {
        |    case Nil,
        |    case Cons(t, List[t])
        |}
      """.stripMargin
    expectError[TypeError](result)
  }
```

Which will ensure the error is on the second line, and fail the test if it is not.


magnus-madsen:
Not exactly pretty, but okay.

mlutze:
> Not exactly pretty, but okay.

I'm always open to beauty tips.

PR #9890, opened 2025-02-06
feat: provide semantic tokens to constants

stormckey:
Now in nvim, constants are colored:
image

We should also remove the syntax highlight for vscode and adjust the SemanticTokenTypes used by the client.

Note:
- maybe we can just use the full list of SemanticTokenTypes and Modifiers so that we dont need to change it every time we add something new:
  image
- I assign Char and Str the String type
- I assign Regex the Regexp type
- I assign number to Int* Float* BigInt BigDecimal
- The rest falls back to Type

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
@@ -333,7 +333,35 @@ object SemanticTokensProvider {
 
     case Expr.Use(_, _, exp, _) => visitExp(exp) // TODO NS-REFACTOR add token for sym
 
-    case cst: Expr.Cst => visitCst(cst)
+    case Expr.Cst(Constant.Str(_), _, loc) => Iterator(SemanticToken(SemanticTokenType.String, Nil, loc))
+
+    case Expr.Cst(Constant.Char(_), _, loc) => Iterator(SemanticToken(SemanticTokenType.String, Nil, loc))
+
+    case Expr.Cst(Constant.Int8(_), _, loc) => Iterator(SemanticToken(SemanticTokenType.Number, Nil, loc))
+
+    case Expr.Cst(Constant.Int16(_), _, loc) => Iterator(SemanticToken(SemanticTokenType.Number, Nil, loc))
+
+    case Expr.Cst(Constant.Int32(_), _, loc) => Iterator(SemanticToken(SemanticTokenType.Number, Nil, loc))
+
+    case Expr.Cst(Constant.Int64(_), _, loc) => Iterator(SemanticToken(SemanticTokenType.Number, Nil, loc))
+
+    case Expr.Cst(Constant.BigInt(_), _, loc) => Iterator(SemanticToken(SemanticTokenType.Number, Nil, loc))
+
+    case Expr.Cst(Constant.Float32(_), _, loc) => Iterator(SemanticToken(SemanticTokenType.Number, Nil, loc))
+
+    case Expr.Cst(Constant.Float64(_), _, loc) => Iterator(SemanticToken(SemanticTokenType.Number, Nil, loc))
+
+    case Expr.Cst(Constant.BigDecimal(_), _, loc) => Iterator(SemanticToken(SemanticTokenType.Number, Nil, loc))
+
+    case Expr.Cst(_: Constant.Regex, _, loc) => Iterator(SemanticToken(SemanticTokenType.Regexp, Nil, loc))
+
+    case Expr.Cst(_: Constant.Bool, _, loc) => Iterator(SemanticToken(SemanticTokenType.Type, Nil, loc))
+
+    case Expr.Cst(Constant.Unit, _, loc) => Iterator(SemanticToken(SemanticTokenType.Type, Nil, loc))
+
+    case Expr.Cst(Constant.Null, _, loc) => Iterator(SemanticToken(SemanticTokenType.Type, Nil, loc))
+
+    case Expr.Cst(Constant.RecordEmpty, _, loc) => Iterator(SemanticToken(SemanticTokenType.Type, Nil, loc))
 
     case Expr.Lambda(fparam, exp, _, _) =>
       visitFormalParam(fparam) ++ visitExp(exp)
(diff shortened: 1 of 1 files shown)


magnus-madsen:
> We should also remove the syntax highlight for vscode and adjust the SemanticTokenTypes used by the client.

Agreed, but we have to sync the release of both. For now we leave it in VSCode and then we can remove it just before a release.

We can create a PR though

magnus-madsen:
> maybe we can just use the full list of SemanticTokenTypes and Modifiers so that we dont need to change it every time we add something new: 

Agreed

magnus-madsen:
Does this PR break semantic highlighting in VSCode then?

main/src/ca/uwaterloo/flix/api/lsp/SemanticTokenType.scala
```diff
@@ -36,11 +36,14 @@ sealed trait SemanticTokenType {
...
-    case SemanticTokenType.TypeParameter => 11
-    case SemanticTokenType.Variable => 12
+    case SemanticTokenType.Regexp => 8
+    case SemanticTokenType.Method => 9
+    case SemanticTokenType.Namespace => 10
+    case SemanticTokenType.Number => 11
+    case SemanticTokenType.String => 12
+    case SemanticTokenType.Type => 13
+    case SemanticTokenType.TypeParameter => 14
+    case SemanticTokenType.Variable => 15
```
magnus-madsen, line 46: I don't think you can change this without changing server\src\handlers.ts so that poses an issue. Hmm

magnus-madsen:
As we discussed, I suggest two PRs that synchronize the lists adding all tokens. (Separately from this PR)

magnus-madsen:
This can now be rebased :)

stormckey:
Done.
image

main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
```diff
@@ -625,6 +625,21 @@ object SemanticTokensProvider {
   private def visitExps(exps0: List[Expr]): Iterator[SemanticToken] =
     exps0.flatMap(visitExp).iterator
 
+  private def visitCst(constant: Expr.Cst): Iterator[SemanticToken] = constant.cst match {
```
magnus-madsen, line 628: I think we should just inline the function, since it is only used in one place. Is that right?

main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
```diff
@@ -625,6 +625,21 @@ object SemanticTokensProvider {
...
+    case _ : Constant.Str |
```
magnus-madsen, line 629: Writing the pattern match this way is not idiomatic. Just match on the constructor and write Str(_, _).

main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
```diff
@@ -625,6 +625,21 @@ object SemanticTokensProvider {
...
+    case _ : Constant.Int8 |
+         _ : Constant.Int16 |
+         _ : Constant.Int32 |
+         _ : Constant.Int64 |
+         _ : Constant.BigInt |
+         _ : Constant.Float32 |
+         _ : Constant.Float64 |
+         _ : Constant.BigDecimal => Iterator(SemanticToken(SemanticTokenType.Number, Nil, constant.loc))
+    case _: Constant.Regex => Iterator(SemanticToken(SemanticTokenType.Regexp, Nil, constant.loc))
+    case _ => Iterator(SemanticToken(SemanticTokenType.Type, Nil, constant.loc))
```
magnus-madsen, line 640: In particular the wildcard case should avoided: Otherwise we won't get errors when the compiler is extended with a new constant.

main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
```diff
@@ -625,6 +625,21 @@ object SemanticTokensProvider {
   private def visitExps(exps0: List[Expr]): Iterator[SemanticToken] =
     exps0.flatMap(visitExp).iterator
 
+  private def visitCst(constant: Expr.Cst): Iterator[SemanticToken] = constant.cst match {
+    case _ : Constant.Str |
+         _ : Constant.Char => Iterator(SemanticToken(SemanticTokenType.String, Nil, constant.loc))
+    case _ : Constant.Int8 |
```
magnus-madsen, line 631: What about Boolean?

stormckey: There is no such type for boolean. It falls to the most generic `Type`.

main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
magnus-madsen: Oh, and also, we dont use | or patterns. They are fragile. Just write out the normal pattern match 👍

main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
```diff
@@ -333,7 +333,35 @@ object SemanticTokensProvider {
 
     case Expr.Use(_, _, exp, _) => visitExp(exp) // TODO NS-REFACTOR add token for sym
 
-    case Expr.Cst(_, _, _) => Iterator.empty
+    case Expr.Cst(Constant.Str(_), _, loc) => Iterator(SemanticToken(SemanticTokenType.String, Nil, loc))
```
magnus-madsen, line 336: Actually, one last change, you should do it like so:

```
case Expr.Cst(c, _, _) => c match {
  case Constant.Char(_) => Iterator(SemanticToken(SemanticTokenType.String, Nil, loc))
```

main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
```diff
@@ -333,7 +333,35 @@ object SemanticTokensProvider {
...
+
+    case Expr.Cst(_: Constant.Regex, _, loc) => Iterator(SemanticToken(SemanticTokenType.Regexp, Nil, loc))
+
+    case Expr.Cst(_: Constant.Bool, _, loc) => Iterator(SemanticToken(SemanticTokenType.Type, Nil, loc))
+
+    case Expr.Cst(Constant.Unit, _, loc) => Iterator(SemanticToken(SemanticTokenType.Type, Nil, loc))
+
+    case Expr.Cst(Constant.Null, _, loc) => Iterator(SemanticToken(SemanticTokenType.Type, Nil, loc))
+
+    case Expr.Cst(Constant.RecordEmpty, _, loc) => Iterator(SemanticToken(SemanticTokenType.Type, Nil, loc))
```
magnus-madsen, line 364: Also please keep the pattern match in the exact same order as the constants are listed in `Constant.scala` -- it helps maintainability.

stormckey:
Just a last reminder of what I found on the VSCode side
syntax highlight will provide syntax highlight for constructs inside a string like this:
image

This is impossible for semantic highlight.

stormckey:
> Just a last reminder of what I found on the VSCode side syntax highlight will provide syntax highlight for constructs inside a string like this: image
> 
> This is impossible for semantic highlight, we will give the whole string a single semantic token and it will be colored with a single color too.

magnus-madsen:
> Just a last reminder of what I found on the VSCode side syntax highlight will provide syntax highlight for constructs inside a string like this: image
> 
> This is impossible for semantic highlight: we will give the whole string a single semantic token and it will be colored with a single color too.

I see. I suppose we can deal with that in the future.

magnus-madsen (approved):

PR #9895, opened 2025-02-07
feat: support code lens for stand alone lsp

stormckey:
Preview:
image
image

Note:
- for nvim, we need to first call `vim.lsp.codelens.refresh()` to get the codelens and then use the shortcut `<leader>cl`(which will call `vim.lsp.codelens.run()` to run the codelens. Usually, it's recommended to add this config to refresh automatically:
  ```lua
        vim.api.nvim_create_autocmd({ "BufEnter", "CursorHold", "InsertLeave" }, {
          pattern = "<buffer>",
          callback = function()
            vim.lsp.codelens.refresh({ bufnr = bufnr })
          end,
        })
   ```
  Should we add this to the minimal config? It makes no sense to just gives a shortcut that doesn't work on its own.
- The commands in the received code lens still need to be implemented. In VSCode, these commands are implemented on the client side, which is not available for nvim. Currently I just add some mock method to handle that on the server side. I think it's possible to handle all of them on the server side so that we do not need to implement them every time we support a new IDE.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/CodeLens.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/CodeLens.scala
@@ -15,6 +15,7 @@
  */
 package ca.uwaterloo.flix.api.lsp
 
+import org.eclipse.lsp4j
 import org.json4s.JsonDSL.*
 import org.json4s.*
 
@@ -26,4 +27,11 @@ import org.json4s.*
   */
 case class CodeLens(range: Range, command: Option[Command]) {
   def toJSON: JValue = ("range" -> range.toJSON) ~ ("command" -> command.map(_.toJSON))
+
+  def toLsp4j: lsp4j.CodeLens = {
+    val codeLens = new lsp4j.CodeLens()
+    codeLens.setRange(range.toLsp4j)
+    command.foreach(cmd => codeLens.setCommand(cmd.toLsp4j))
+    codeLens
+  }
 }
--- a/main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
@@ -16,7 +16,7 @@
 package ca.uwaterloo.flix.api.lsp
 
 import ca.uwaterloo.flix.api.lsp.Range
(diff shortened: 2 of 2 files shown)


magnus-madsen:
> Should we add this to the minimal config? 

I suppose.

In general it will probably be hard to support code lenses properly. So we could also consider leaving the feature out for now.

I am actually more inclined to do that: I would prefer to have less features but that they all work extremely well and are stable.

stormckey:
OK, so what's the plan?

If we'd like to move the executeCommand capability to the server. We could first use several PRs to achieve that. Then all these code lens commands would be supported across IDE. And we'll be able to continue to merge this?

If we want such a shift. The VSCode client should also be changed simultaneously, as it's the provider of these commands now.

magnus-madsen:
> OK, so what's the plan?
> 
> If we'd like to move the executeCommand capability to the server. We could first use several PRs to achieve that. Then all these code lens commands would be supported across IDE. And we'll be able to continue to merge this?
> 
> If we want such a shift. The VSCode client should also be changed simultaneously, as it's the provider of these commands now.

We cannot execute the command on the server -- because it needs to run the program in a new JVM instance.

So I am not really sure what we can do? I think each client (Neovim, etc. etc.) would have to be extended with custom code that knows how to run the command. 

Hence why I am thinking code lenses may not be worth it.

magnus-madsen:
Can we implement some code on the Neovim side such that when "runMain" is received it executes some commands?

stormckey:
Yes, the commands could be implemented on either the client side or the server side. 

I think maybe we could prioritize other functionalities and come back to code lens once others are done?

stormckey:
> But if its possible, you could just map `runMain` to the command `java -jar flix.jar run` in Neovim. (If it is possible).

I think it's possible. I will test that later.

stormckey:
Here is the code required to setup `runMain` in neovim:
image

Result:
image
image

magnus-madsen:
Nice, so we can update this PR, merged it, and then add that stuff to `init.lua`?

stormckey:
> Nice, so we can update this PR, merged it, and then add that stuff to `init.lua`?

Yes, I've removed the exec command part from the server. What else to update?

magnus-madsen:
> > Nice, so we can update this PR, merged it, and then add that stuff to `init.lua`?
> 
> Yes, I've removed the exec command part from the server. What else to update?

I suppose add that code from the screenshot to doc.flix.dev lua config?

Also resolve the conflict.

magnus-madsen (approved):

PR #9896, opened 2025-02-07
feat: use the full SemanticToken legend

stormckey:
for your reference:
image

We use to order them alphabetically. Now I just list them the way spec lists them.


magnus-madsen:
I'll ping you when both are released

magnus-madsen (approved):

PR #9898, opened 2025-02-08
perf: toggle a few performance flags

magnus-madsen:


magnus-madsen:
(This is the optimal configuration.)

mlutze (approved):

PR #9901, opened 2025-02-10
fix: use `System.out` as `PrintStream` in `Main`

jaschdoc:
Why did we use `System.err` before?

- [x] Check it works in vscode

### DIFF
--- a/main/src/ca/uwaterloo/flix/Main.scala
+++ b/main/src/ca/uwaterloo/flix/Main.scala
@@ -123,7 +123,7 @@ object Main {
     // check if command was passed.
     try {
       implicit val formatter: Formatter = Formatter.getDefault
-      implicit val out: PrintStream = System.err
+      implicit val out: PrintStream = System.out
 
       cmdOpts.command match {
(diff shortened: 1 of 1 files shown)


magnus-madsen:
Chestertons fence? Why is this like that?

Maybe it interacts badly with JLine? Where is out even used?

jaschdoc:
> Maybe it interacts badly with JLine?

If it does, we should have a comment on why it is `System.err`

> Where is out even used?

Used in Bootstrap only but it does not seem to interact badly with the terminal or repl, although, I'm not sure.

magnus-madsen:
I am pretty sure there is a reason for this. Its to avoid interacting with the output of the program.

jaschdoc:
> I am pretty sure there is a reason for this. Its to avoid interacting with the output of the program.

Yeah seems like the most obvious pitfall

jaschdoc:
If we use `System.err` for the output stream then we should also use it for printing error messages. Currently we just print using `println` instead of the output stream.

PR #9911, opened 2025-02-11
feat: support rename for standalone lsp

stormckey:
image
image
image

The capability is the same as the vscode client:
image

It's ok to return an empty workspaceEdit when there is no edit:
image

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
@@ -311,6 +311,8 @@ object LspServer {
       val pos = Position.fromLsp4j(params.getPosition)
       RenameProvider.processRename(newName, uri, pos)(flixLanguageServer.root) match {
         case Some(rename) => CompletableFuture.completedFuture(rename.toLsp4j)
+
+        // If nothing is found it's OK to return the empty WorkspaceEdit.
         case None => CompletableFuture.completedFuture(new WorkspaceEdit())
       }
     }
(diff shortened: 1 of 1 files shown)


magnus-madsen:
> It's ok to return an empty workspaceEdit when there is no edit:

Sounds good.

main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
```diff
@@ -126,6 +126,7 @@ object LspServer {
       serverCapabilities.setReferencesProvider(true)
       serverCapabilities.setDefinitionProvider(true)
       serverCapabilities.setImplementationProvider(true)
+      serverCapabilities.setRenameProvider(new RenameOptions(false))
```
magnus-madsen, line 129: Whats the boolean here?

magnus-madsen: Is it the prepareRename thing?

main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
```diff
@@ -304,6 +305,16 @@ object LspServer {
       CompletableFuture.completedFuture(references.map(_.toLsp4j).toList.asJava)
     }
 
+    override def rename(params: RenameParams): CompletableFuture[WorkspaceEdit] = {
+      val newName = params.getNewName
+      val uri = params.getTextDocument.getUri
+      val pos = Position.fromLsp4j(params.getPosition)
+      RenameProvider.processRename(newName, uri, pos)(flixLanguageServer.root) match {
+        case Some(rename) => CompletableFuture.completedFuture(rename.toLsp4j)
+        case None => CompletableFuture.completedFuture(new WorkspaceEdit())
```
magnus-madsen, line 314: I would add that comment from the PR-- that if nothing is found its OK to return the empty workspace edit.

main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
stormckey: Yes

PR #9916, opened 2025-02-12
refactor: make processWorkspaceSymbols return WorkspaceSymbols

stormckey:
I think the go to symbol in workspace relies on the `processWorkspaceSymbols`? Or is there other way to test that it works fine?
image


stormckey:
> There is a command to list all symbols.
> 
> There is also a pane you can bring forth.

The OUTLINE pane shows document symbols. Or could I add another pane?

PR #9918, opened 2025-02-12
feat: support workspace symbol for stand alone lsp

stormckey:
now the content is the same as document symbols
image

Note that I filtered out all the symbols whose uri doesn't starts with "file://", which usually refers to flix standard lib files. nvim will panic over such uris. And these usis do not actaully work on vscode:
image
image

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/SymbolProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/SymbolProvider.scala
@@ -35,7 +35,9 @@ object SymbolProvider {
     val sigs = root.sigs.values.collect { case sig if sig.sym.name.startsWith(query) => mkSigWorkspaceSymbol(sig) }
     val effs = root.effects.values.filter(_.sym.name.startsWith(query)).flatMap(mkEffectWorkspaceSymbol)
     val structs = root.structs.values.filter(_.sym.name.startsWith(query)).flatMap(mkStructWorkspaceSymbol)
-    (traits ++ defs ++ enums ++ sigs ++ effs ++ structs).toList
+    (traits ++ defs ++ enums ++ sigs ++ effs ++ structs).toList.filter{
+      case WorkspaceSymbol(_, _, _, _, loc) => loc.uri.startsWith("file://")
+    }
   }
 
   /**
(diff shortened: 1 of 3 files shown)


main/src/ca/uwaterloo/flix/api/lsp/provider/SymbolProvider.scala
```diff
@@ -35,7 +35,9 @@ object SymbolProvider {
     val sigs = root.sigs.values.collect { case sig if sig.sym.name.startsWith(query) => mkSigWorkspaceSymbol(sig) }
     val effs = root.effects.values.filter(_.sym.name.startsWith(query)).flatMap(mkEffectWorkspaceSymbol)
     val structs = root.structs.values.filter(_.sym.name.startsWith(query)).flatMap(mkStructWorkspaceSymbol)
-    (traits ++ defs ++ enums ++ sigs ++ effs ++ structs).toList
+    (traits ++ defs ++ enums ++ sigs ++ effs ++ structs).toList.filter{
+      case WorkspaceSymbol(_, _, _, _, loc) => loc.uri.startsWith("file://")
```
magnus-madsen, line 39: What's the reason for this?

stormckey: > Note that I filtered out all the symbols whose uri doesn't starts with "file://", which usually refers to flix standard lib files. nvim will panic over such uris. And these usis do not actaully work on vscode: image image

magnus-madsen: And this works in VSCode too?

magnus-madsen: I would assume this is not just for WorkspaceSymbols; so you should open a ticket and then we should deal with it generally.

magnus-madsen (approved):

main/src/ca/uwaterloo/flix/api/lsp/provider/SymbolProvider.scala
stormckey: Ok, I think VSCode just treat uri without file:// as absolute path.

PR #9921, opened 2025-02-13
feat: support load a whole flix project for the standalone lsp

stormckey:
As discussed in this issue https://github.com/microsoft/language-server-protocol/issues/1465. There is no such protocol for the server to load all source files in a project. The server has to manually load all source files given the root of the workspace, and that's what this PR is doing.

Preview:
It works on Jonathan's project, no compiler error here
image
And workspace symbols will include all files:
image

We have hover information for functions defined in other files:
image
and we can jump to the definition:
image

Note:
- I load all the flix files, jars, fpkgs and flix.toml using the pattern in vscode client:
  image
- Now we only load flix.toml as a string in the server.
- The code to add these files is copied from VSCodeLspServer.
- We are still finding `flix.jar` in the root of workspace.
- The client side setting is:
  image

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/Flix.scala
+++ b/main/src/ca/uwaterloo/flix/api/Flix.scala
@@ -30,10 +30,11 @@ import ca.uwaterloo.flix.util.*
 import ca.uwaterloo.flix.util.collection.{Chain, MultiMap}
 import ca.uwaterloo.flix.util.tc.Debug
 
+import java.io.ByteArrayInputStream
 import java.nio.charset.Charset
 import java.nio.file.{Files, Path}
 import java.util.concurrent.ForkJoinPool
-import java.util.zip.ZipFile
+import java.util.zip.{ZipFile, ZipInputStream}
 import scala.collection.mutable
 import scala.collection.mutable.ListBuffer
 import scala.language.implicitConversions
@@ -395,6 +396,23 @@ class Flix {
       throw new IllegalArgumentException(s"'$p' must be a *.pkg file.")
 
     addInput(p.toString, Input.PkgFile(p, sctx))
+    this
+  }
+
+  def addFpkg(uri: String, data: Array[Byte], sources: mutable.Map[String, String]): Flix = {
+    val inputStream = new ZipInputStream(new ByteArrayInputStream(data))
+    var entry = inputStream.getNextEntry
+    while (entry != null) {
+      val name = entry.getName
+      if (name.endsWith(".flix")) {
+        val bytes = StreamOps.readAllBytes(inputStream)
+        val src = new String(bytes, Charset.forName("UTF-8"))
+        addSourceCode(s"$uri/$name", src)(SecurityContext.AllPermissions)
+        sources += (s"$uri/$name" -> src)
+      }
+      entry = inputStream.getNextEntry
+    }
+    inputStream.close()
     this
   }
 
--- a/main/src/ca/uwaterloo/flix/api/Flix.scala
+++ b/main/src/ca/uwaterloo/flix/api/Flix.scala
@@ -30,11 +30,10 @@ import ca.uwaterloo.flix.util.*
 import ca.uwaterloo.flix.util.collection.{Chain, MultiMap}
 import ca.uwaterloo.flix.util.tc.Debug
 
-import java.io.ByteArrayInputStream
 import java.nio.charset.Charset
 import java.nio.file.{Files, Path}
 import java.util.concurrent.ForkJoinPool
-import java.util.zip.{ZipFile, ZipInputStream}
+import java.util.zip.ZipFile
 import scala.collection.mutable
 import scala.collection.mutable.ListBuffer
 import scala.language.implicitConversions
--- a/main/src/ca/uwaterloo/flix/api/Flix.scala
+++ b/main/src/ca/uwaterloo/flix/api/Flix.scala
@@ -399,23 +399,6 @@ class Flix {
     this
   }
 
-  def addFpkg(uri: String, data: Array[Byte], sources: mutable.Map[String, String]): Flix = {
-    val inputStream = new ZipInputStream(new ByteArrayInputStream(data))
-    var entry = inputStream.getNextEntry
-    while (entry != null) {
-      val name = entry.getName
-      if (name.endsWith(".flix")) {
-        val bytes = StreamOps.readAllBytes(inputStream)
-        val src = new String(bytes, Charset.forName("UTF-8"))
-        addSourceCode(s"$uri/$name", src)(SecurityContext.AllPermissions)
-        sources += (s"$uri/$name" -> src)
-      }
-      entry = inputStream.getNextEntry
-    }
-    inputStream.close()
-    this
-  }
-
   /**
     * Removes the given path `p` as a Flix source file.
     */
--- a/main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
@@ -25,7 +25,7 @@ import ca.uwaterloo.flix.language.ast.TypedAst.Root
 import ca.uwaterloo.flix.language.ast.shared.SecurityContext
 import ca.uwaterloo.flix.language.phase.extra.CodeHinter
 import ca.uwaterloo.flix.util.Formatter.NoFormatter
-import ca.uwaterloo.flix.util.Options
+import ca.uwaterloo.flix.util.{FileOps, Options}
 import org.eclipse.lsp4j
 import org.eclipse.lsp4j.*
 import org.eclipse.lsp4j.jsonrpc.messages
@@ -113,23 +113,6 @@ object LspServer {
     }
 
     /**
-      * Returns an iterator of all files in the given path, visited recursively.
-      */
-    private def getRecursiveFilesIterator(path: Path): Iterator[Path] = {
-      if (Files.exists(path) && Files.isDirectory(path))
-        Files.walk(path).iterator().asScala
-      else
-        Iterator.empty
-    }
-
-    /**
-      * Checks if the given path is a regular file with the expected extension.
-      */
-    private def checkExt(p: Path, expectedExt: String): Boolean = {
-      Files.isRegularFile(p) && p.getFileName.toString.endsWith(expectedExt)
-    }
-
-    /**
       * Loads all Flix resources in the workspace, including:
       *   - Flix source files (*.flix, src/**/*.flix, test/**/*.flix).
       *   - JAR files (lib/**/*.jar).
@@ -153,16 +136,17 @@ object LspServer {
       *   - test/**/*.flix
       */
     private def loadFlixSources(path: Path): Unit = {
-      val flixFileCandidates = Files.list(path).iterator().asScala ++
-        getRecursiveFilesIterator(path.resolve("src")) ++
-        getRecursiveFilesIterator(path.resolve("test"))
+      val flixSources = {
+        FileOps.getFilesIn(path, 1) ++
+        FileOps.getFilesIn(path.resolve("src"), Int.MaxValue) ++
+        FileOps.getFilesIn(path.resolve("test"), Int.MaxValue)
+      }
 
-      flixFileCandidates.foreach {
-        case p =>
-          if (checkExt(p, ".flix")) {
-            val source = Files.readString(p)
-            addSourceCode(p.toUri.toString, source)
-          }
+      flixSources.foreach { case p =>
+        if (FileOps.checkExt(p, ".flix")) {
+          val source = Files.readString(p)
+          addSourceCode(p.toUri.toString, source)
+        }
       }
     }
 
@@ -172,13 +156,13 @@ object LspServer {
       *   - lib/**/*.fpkg
       */
     private def loadJarsAndFkgs(path: Path) = {
-      getRecursiveFilesIterator(path.resolve("lib"))
+      FileOps.getFilesIn(path.resolve("lib"), Int.MaxValue)
         .foreach{ case p =>
           // Load all JAR files in the workspace, the pattern should be lib/**/*.jar.
-          if (checkExt(p, ".jar"))
+          if (FileOps.checkExt(p, ".jar"))
             flix.addJar(p)
           // Load all Flix package files in the workspace, the pattern should be lib/**/*.fpkg.
-          if (checkExt(p, ".fpkg")) {
+          if (FileOps.checkExt(p, ".fpkg")) {
             flix.addPkg(p)(SecurityContext.AllPermissions)
           }
         }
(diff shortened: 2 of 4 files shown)


main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
```diff
@@ -90,6 +94,8 @@ object LspServer {
       */
     private var clientCapabilities: ClientCapabilities = _
 
+    private var flixToml: String = _
```
magnus-madsen, line 97: What is this? Is it null? Should be Option then. Can we document it?

main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
```diff
@@ -165,6 +237,7 @@ object LspServer {
       */
     def processCheck(): Unit = {
       try {
+        System.err.println(sources)
```
magnus-madsen, line 240: Remove?

main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
```diff
@@ -98,16 +104,82 @@ object LspServer {
...
       System.err.println(s"initialize: $initializeParams")
 
       clientCapabilities = initializeParams.getCapabilities
 
+      loadFlixProject(initializeParams.getWorkspaceFolders)
+
       CompletableFuture.completedFuture(new InitializeResult(mkServerCapabilities()))
     }
 
+    private def loadFlixProject(roots: util.List[WorkspaceFolder]): Unit = {
```
magnus-madsen, line 120: Add docs with examples

main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
```diff
@@ -98,16 +104,82 @@ object LspServer {
...
+          }
+        }
+
+        getRecursiveFilesIterator(path.resolve("lib"))
+          .foreach{ case p =>
+            // Load all JAR files in the workspace, the pattern should be lib/**/*.jar.
+            if (checkExt(p, ".jar"))
+              flix.addJar(p)
+            // Load all Flix package files in the workspace, the pattern should be lib/**/*.fpkg.
+            if (checkExt(p, ".fpkg")) {
```
magnus-madsen, line 162: Does flix have no add package command?

main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
```diff
@@ -98,16 +104,82 @@ object LspServer {
...
+        root <- roots.asScala
+        path = Paths.get(root.getName)
+        if Files.exists(path) && Files.isDirectory(path)
+      } {
+        // Load the flix.toml file from the root of workspace if it exists.
+        val flixTomlPath = path.resolve("flix.toml")
+        if (Files.exists(flixTomlPath))
+          flixToml = Files.readString(flixTomlPath)
+
+       def getRecursiveFilesIterator(path: Path): Iterator[Path] = {
```
magnus-madsen, line 131: Can we extract these functions into private functions that are not nested?

magnus-madsen (commented):
Good progress. I added a lot of comments. 

I also have some high-level questions:

- How does this affect init.lua? There I seem to remember some code that searched for a .git folder?
- What happens if there is no flix.jar in the folder where the flix.toml is?
- What happens when one of these files is changed on disk? 
- Or when Neovim has two open buffers?

magnus-madsen:
- Does `init.lua` have to be updated? why / why not?

stormckey:
> * How does this affect init.lua? There I seem to remember some code that searched for a .git folder?

Now it's been changed the code that searches for flix.toml

> * What happens if there is no flix.jar in the folder where the flix.toml is?

Now the start will fail and we can see such log:
image
I am figuring out how to emit better error message.

> * What happens when one of these files is changed on disk?

Now we will only update changes on the buffer we open. To track all the changes on the disk, I think we need to impl workspace/didChangeWatchedFiles

> * Or when Neovim has two open buffers?

what do you expect? It will still send a didOpen as we open the first one.
image

main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
stormckey: No, we also do this in VSCodeLspServer

stormckey:
Now we have this error message:
image

through this config:
image

main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
magnus-madsen: I see, but I think that is a mistake. Can we not use flix.addPkg?

main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
```diff
@@ -90,6 +94,12 @@ object LspServer {
       */
     private var clientCapabilities: ClientCapabilities = _
 
+    /**
+      * The flix.toml file content.
+      * We will only try to load flix.toml from the root of the workspace.
+      */
+    private var flixToml: Option[String] = None
```
magnus-madsen, line 101: What is the reason this has to be a field?

main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
```diff
@@ -98,16 +108,89 @@ object LspServer {
...
+      Files.isRegularFile(p) && p.getFileName.toString.endsWith(expectedExt)
+    }
+    /**
+      * Loads all Flix resources in the workspace, including:
+      *   - Flix source files (*.flix, src/**/*.flix, test/**/*.flix).
+      *   - JAR files (lib/**/*.jar).
+      *   - Flix package files (lib/**/*.fpkg).
+      *   - flix.toml file (from the root of the workspace).
+      */
+    private def loadFlixProject(roots: util.List[WorkspaceFolder]): Unit = {
```
magnus-madsen, line 141: I think this function can be further improved by breaking it into smaller parts.

magnus-madsen, line 141: Any reason to pass around Java lists and not Scala lists? We dont usually do that...

main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
```diff
@@ -98,16 +108,89 @@ object LspServer {
...
       System.err.println(s"initialize: $initializeParams")
 
       clientCapabilities = initializeParams.getCapabilities
 
+      loadFlixProject(initializeParams.getWorkspaceFolders)
+
       CompletableFuture.completedFuture(new InitializeResult(mkServerCapabilities()))
     }
 
+    private def getRecursiveFilesIterator(path: Path): Iterator[Path] = {
```
magnus-madsen, line 124: Document all functions

main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
stormckey: I choose to save it as a field.

In the future we will need to watch and update it and refer to it for other functionalities.

Do you think there is a better choice?

magnus-madsen: State is always bad. If it does not need to be a field today then just don't make it into a field.

stormckey: So what should we do with flix.toml. Should we ignore it now?

magnus-madsen: What is the string used for?

magnus-madsen:
Ping me when I should take another look 👍

main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
stormckey: Here is the impl of addPkg:
```scala
  /**
    * Adds the given path `p` as a Flix package file.
    */
  def addPkg(p: Path)(implicit sctx: SecurityContext): Flix = {
    if (p == null)
      throw new IllegalArgumentException(s"'p' must be non-null.")
    if (!Files.exists(p))
      throw new IllegalArgumentException(s"'$p' must be a file.")
    if (!Files.isRegularFile(p))
      throw new IllegalArgumentException(s"'$p' must be a regular file.")
    if (!Files.isReadable(p))
      throw new IllegalArgumentException(s"'$p' must be a readable file.")
    if (!p.getFileName.toString.endsWith(".fpkg"))
      throw new IllegalArgumentException(s"'$p' must be a *.pkg file.")

    addInput(p.toString, Input.PkgFile(p, sctx))
    this
  }
```

and this is how VSCodeLspServer processes addPkg request:
```scala
    case Request.AddPkg(id, uri, data) =>
      // TODO: Possibly move into Input class?
      val inputStream = new ZipInputStream(new ByteArrayInputStream(data))
      var entry = inputStream.getNextEntry
      while (entry != null) {
        val name = entry.getName
        if (name.endsWith(".flix")) {
          val bytes = StreamOps.readAllBytes(inputStream)
          val src = new String(bytes, Charset.forName("UTF-8"))
          addSourceCode(s"$uri/$name", src)
        }
        entry = inputStream.getNextEntry
      }
      inputStream.close()
```

You can see that addPkg doesn't actually load the content of the fpkg, so I think I should add another addFpkg method and move the logic there?

stormckey: But the addSourceCode function is actually accessing the field of the server, so we can not directly move it to Flix.

main/src/ca/uwaterloo/flix/api/Flix.scala
```diff
@@ -398,6 +399,23 @@ class Flix {
     this
   }
 
+  def addFpkg(uri: String, data: Array[Byte], sources: mutable.Map[String, String]): Flix = {
```
magnus-madsen, line 402: I think you misunderstood: There is already a function called addPkg right?

main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
```diff
@@ -98,16 +99,93 @@ object LspServer {
...
 
+      loadFlixProject(initializeParams.getWorkspaceFolders.asScala.toList)
+
       CompletableFuture.completedFuture(new InitializeResult(mkServerCapabilities()))
     }
 
+    /**
+      * Returns an iterator of all files in the given path, visited recursively.
+      */
+    private def getRecursiveFilesIterator(path: Path): Iterator[Path] = {
```
magnus-madsen, line 118: Is it important this is an iterator? I think it can just be a list.

I also suggest to use the more modern Files.walk function.

See FileOps.scala for an example.

magnus-madsen:
@stormckey I added two comments; when they are fixed then we can merge, and iterate.

main/src/ca/uwaterloo/flix/api/Flix.scala
magnus-madsen: It should load the content. Check out `Reader.scala`

magnus-madsen: I think the code in VSCodeLspServer is old and should be refactored.

main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
stormckey: Aren't we just using Files.walk?

```scala
    private def getRecursiveFilesIterator(path: Path): Iterator[Path] = {
      if (Files.exists(path) && Files.isDirectory(path))
        Files.walk(path).iterator().asScala
      else
        Iterator.empty
    }
```

I think changing to List will not change the impl a lot. We just need to change Iterator.empty to List.empty, change the signature etc.

But since we are going to concatenate different candidates from three places, iterator is faster in this operation.

magnus-madsen: We are using Files.walk. My bad. But you should return a list; that is safer.

magnus-madsen: Also, any reason not to filter for Flix files immediately?

main/src/ca/uwaterloo/flix/api/Flix.scala
stormckey: alright, so we just call addPkg for both servers?

magnus-madsen: Lets change the other server in a separate PR

main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
stormckey: It's also used for jar and fpkg.
And we dont want to iterate through files in lib twice for jar and fpkg

magnus-madsen:
Please avoid changes to the VSCode LSP server in this PR.

Also, it seems the tests break? Perhaps using Flix.addPkg does not work or what?

magnus-madsen:
At least for VSCode, you need something like:

```
    case Request.AddPkg(id, uri, data) =>
      val u = new URI(uri)
      flix.addPkg(Paths.get(u))(NoPermissions)
```

main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
magnus-madsen: I see

main/src/ca/uwaterloo/flix/api/lsp/VSCodeLspServer.scala
```diff
@@ -210,20 +207,9 @@ class VSCodeLspServer(port: Int, o: Options) extends WebSocketServer(new InetSoc
...
-          val bytes = StreamOps.readAllBytes(inputStream)
-          val src = new String(bytes, Charset.forName("UTF-8"))
-          addSourceCode(s"$uri/$name", src)
-        }
-        entry = inputStream.getNextEntry
-      }
-      inputStream.close()
+    case Request.AddPkg(id, uri, _) =>
+      val path = Paths.get(URI.create(uri))
+      flix.addPkg(path)(SecurityContext.AllPermissions)
```
magnus-madsen, line 212: But, even if this works, we should not do it in this PR.

main/src/ca/uwaterloo/flix/api/Flix.scala
```diff
@@ -30,10 +30,11 @@ import ca.uwaterloo.flix.util.*
 import ca.uwaterloo.flix.util.collection.{Chain, MultiMap}
 import ca.uwaterloo.flix.util.tc.Debug
 
+import java.io.ByteArrayInputStream
 import java.nio.charset.Charset
 import java.nio.file.{Files, Path}
 import java.util.concurrent.ForkJoinPool
-import java.util.zip.ZipFile
+import java.util.zip.{ZipFile, ZipInputStream}
```
magnus-madsen, line 37: unused leftover imports

main/src/ca/uwaterloo/flix/api/lsp/VSCodeLspServer.scala
```diff
@@ -38,7 +38,7 @@ import org.json4s.native.JsonMethods.parse
 import java.io.ByteArrayInputStream
 import java.net.{InetSocketAddress, URI}
 import java.nio.charset.Charset
-import java.nio.file.Path
+import java.nio.file.{Path, Paths}
```
magnus-madsen, line 41: more unused imports

magnus-madsen:
This looks almost ready and then we can do #9930

magnus-madsen (approved):

PR #9932, opened 2025-02-14
feat: provide semantic tokens for modifiers

stormckey:
image

Note:
- Modifier stored in the AST has no loc, so we can not use it to provide semantic tokens. I use the tokens instead.
- Some Modifier is also a keyword(like pub). I didn't find from the lsp spec what will happen in this case. But in nvim, the first token will shadow the following ones. So I put modifierTokens before keywordTokens.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
@@ -1037,8 +1037,7 @@ object SemanticTokensProvider {
     var prevLine = 0
     var prevCol = 0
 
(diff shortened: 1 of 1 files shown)


stormckey:
Well, I found that it should be okay to provide several semantic tokens for the same token, when I am looking at rust:
image


And it turns out that it's this line of code:
image
that removes duplicated tokens.

stormckey:
Now we can provide several semantic tokens:

And they share the same priority and same fall back, so it doesn't matter.

stormckey:
I think this is used to control overlapping semantic tokens.
image

But now for nvim and vscode, it turns out to be fine.

magnus-madsen (approved):

PR #9951, opened 2025-02-17
feat: add `EffUnification3.simplify`

magnus-madsen:


main/src/ca/uwaterloo/flix/language/phase/unification/EffUnification3.scala
```diff
@@ -392,6 +393,27 @@ object EffUnification3 {
...
+    * Simplifies the given effect `tpe` by conversion to - and back from - a Zhegalkin polynomial.
+    *
+    * WARNING:
+    * - The type `tpe` *MUST* have kind `Eff`.
+    * - The type `tpe` *MUST* be well-kinded. Do not use this function for ill-kinded effects!
+    *
+    * The type `tpe` may contain `Type.Error`.
+    */
+  def simplify(tpe: Type): Type = {
+    implicit val scope: Scope = Scope.Top
```
mlutze, line 406: Comment: we can use an arbirary scope and renv because we don't do any unification.

mlutze (approved):

main/src/ca/uwaterloo/flix/language/phase/unification/EffUnification3.scala
magnus-madsen: Added

PR #9952, opened 2025-02-17
feat: use qname directly in completers

stormckey:
Now we can pass qn where we need it instead of namespace + ident.

Not every completer is tested, but they should all work fine since the change is trivial:
image
image
image
image
image


magnus-madsen (approved):

magnus-madsen:
🚀

PR #9956, opened 2025-02-17
feat: support partially qualified name for SigCompleter

stormckey:
Preview:
image
image
image

Note:
- The algorithm:
  - get the first item of the namespace (we've checked its non-empty beforehand)
  - search for it in the local scope, we only accept Namespace(BMod above) or Trait(Addable above)
  - if found, concatenate its fully qualified form with the rest of the namespace
  - now use the concatenated name to search for trait
  - provide sig completions from the trait
- the resolution conflict problem:
  - If the user uses two different BMod, then there will be two resolution for BMod in the local scope. In this case BMod is ambiguous and we should not provide the completion starts with BMod. But we cannot just check for duplication. For example, in the case above, `Addable` is also resolved to two traits, but it's totally fine. Currently, I am just use headOpt to get one resolution, now defined in env.resolve(name).

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
@@ -359,8 +359,11 @@ sealed trait Completion {
         additionalTextEdits = additionalTextEdit
       )
 
-    case Completion.SigCompletion(sig, ap, qualified, inScope) =>
-      val qualifiedName = sig.sym.toString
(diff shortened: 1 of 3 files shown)


stormckey:
By the way, the current impl in EnumTagCompleter doesn't support case like `BMod.Clr.`. If we adopt this algorithm here, I will update EnumTagCompleter to use it too.

magnus-madsen:
I have been busy, but I will try to get around to it soon.

magnus-madsen (approved):

PR #9962, opened 2025-02-18
fix: require operation decl to be annotated with return type

### LINKED ISSUE #9953: Missing error for unannotated effect operation return type
```
eff E {
    def op()
}
```

This should be a parse error but no error is issued.

stormckey:
fixes #9953
The error message:

image


magnus-madsen:
Can we add a test case? to TestParser?

magnus-madsen:
Or maybe even two or three test cases :)

stormckey:
> Was this only a problem for ops? What about other type of defs? E.g. inside traits or instances?
They are fine, I think
image
image

magnus-madsen (approved):

Issue #9935, opened 2025-02-14
Uri for flix lib files


stormckey:
There are several cases where we could access the uri of a flix lib file.

like commnad+click/go_to_definition or workspace symbols(now symbols from standard lib is filtered).

The problem is that their uri it not standard, it should be something like(assuming it's local)
file://path/to/lib.flix

but now it's lib.flix. 

Without the "file://" prefix, which is called scheme, nvim will raise exception:

<img width="1206" alt="Image" src="https://github.com/user-attachments/assets/03f0b61d-5796-44a2-a7ca-f5f6b782431c" />

VSCode will not, but treat it as "/lib.flix" and says it's not available as the top screenshot.

How should we provide access to these files?

magnus-madsen:
We cannot really provide access to these files since they are not real files on the disk.

At least that is my understanding of the problem.

magnus-madsen:
Perhaps we should instead filter them? So we don't over goto definition etc.

stormckey:
We've already filtered them out in workspace symbols, given that their uri doesn't start with "file://"

I suppose this should also be done for GotoProvider. Any others?

PR #9965, opened 2025-02-18
fix: filter out lib files in GotoProvider

stormckey:
As discussed in #9935
We now filter out standard lib files in the GotoProvider.

I am not not sure if we should do this filtering. Currently if you try to command+click to go to something in the lib, it will say:

image

After filtering, it just keeps silent. (But of course, we can say something, in the plain lsp, we can send something back, in vscode, we can print to the console)


magnus-madsen (approved):

magnus-madsen:
Perhaps we need a general filtering mechanism?

stormckey:
> Perhaps we need a general filtering mechanism?

What does that mean? Stop visiting them or?

Issue #9775, opened 2025-01-27
Finer Visibility Check


stormckey:
Currently the most commonly used visibility check is `isPublic || isInFile` for:
```scala
    val isPublic = effect.mod.isPublic && !effect.ann.isInternal
    val isInFile = effect.sym.loc.source.name == uri
```

The `isInFile` check should be `isInSameNamespace`, which is hard to impl now, since most errors have no access to the current namespace.

Another noticeable thing is that things in the root module should also be visible. This could be done by checking if the namespace is empty. But during the test I found that constructs in the root module seems to be public by default. Is that guaranteed so that we can just rely on that, or should we also check the namespace?

stormckey:
Here is the code that is widely used to check the visibility now:
```scala
  /**
   * Checks if the definition matches the QName.
   * Names should match and the definition should be available.
   */
  private def matchesStruct(struct: TypedAst.Struct, qn: Name.QName, uri: String, qualified: Boolean): Boolean = {
    val isPublic = struct.mod.isPublic && !struct.ann.isInternal
    val isInFile = struct.sym.loc.source.name == uri
    val isMatch = if (qualified)
      matchesQualifiedName(struct.sym.namespace, struct.sym.name, qn)
    else
      fuzzyMatch(qn.ident.name, struct.sym.name)
    isMatch && (isPublic || isInFile)
  }
```

We can divide it into two parts. The visibility check needs `*.mod` `*.ann` and `*.sym.loc`, the matching part needs `*.sym.name` and `*.sym.namspace`. And since there is not base trait for them, we have to write this helper function for every construct, even if the code is identical.

But ideally, we do not need `*.sym.loc`, what we actually need is the namespace where the construct is defined, which should be `*.sym.namespace`

Ideally, we can have a base trait that says a construct have mod and ann fields. And a base trait that says a sym has namespace and name. Then we can use one function for most cases. 

Also, we could just divide the functionalities to two functions. And the first one takes ann, mod and namespace, the second takes one namespace and name.

mlutze:
I think knowing the modifiers and annotations on a symbol seems helpful, but I don't think we should add it directly to the symbol. The reason is that we depend of comparison and equality of symbols, and adding the modifiers and annotations will complicate that.

So I think dividing up the functionalities as you describe is better.

For the namespace stuff, it seems better to put it in the error, and keep the qname clean.

stormckey:
[...]

PR #9969, opened 2025-02-19
feat: introduce Decl for TypedAst constructs

stormckey:
As discussed in #9775

We first introduce available on constructs, so that we can always all it to get the bool instead of writing a function for each.

Next step will be introducing some thing like IsQName(?) to Symbols with namespace and name, so that we can reuse a `match(sym: IsQName, qn: Name.QName, qualified: Boolean`  to handle all matches.

After that we can get rids of all the matchesEnum matchesStruct things.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/TypedAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/TypedAst.scala
@@ -49,14 +49,14 @@ object TypedAst {
                   dependencyGraph: DependencyGraph,
                   tokens: Map[Source, Array[Token]])
 
-  sealed trait Declaration {
+  sealed trait Decl {
     val ann: Annotations
     val mod: Modifiers
   }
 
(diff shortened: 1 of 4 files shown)


magnus-madsen:
@stormckey @mlutze I think we need better name.

Can it be trait Declaration?

I also think that we should remove `available` from the trait. We dont have methods on these marker traits. It can be elsewhere.

stormckey:
Ok

main/src/ca/uwaterloo/flix/language/ast/TypedAst.scala
```diff
@@ -24,6 +24,11 @@ import ca.uwaterloo.flix.util.collection.{ListMap, MultiMap}
 
 import java.lang.reflect.{Constructor, Field, Method}
 
+sealed trait Declaration {
```
magnus-madsen, line 27: Let us move this right after Root and right before Trait

magnus-madsen:
Actually let us also rename Declaration to just Decl

magnus-madsen:
And update the title of the PR

magnus-madsen (approved):

Issue #9652, opened 2025-01-10
EnumCompleter doesn't take type params into account


stormckey:
I think when completing enum in expr, it's okay to complete just the name of the enum, since we will use it like this:
<img width="282" alt="image" src="https://github.com/user-attachments/assets/3f9bf00d-5528-4ad0-a9dc-a030ee6126f0" />
But for enum completions in type, we should provide a snippet accounting the type params.

Is there any use cases that I missed out? Or is there a better solution?

magnus-madsen:
Yes, we should provide a completion that has holes for the type parameters. 

In an expression thats not the case. There we complete tags.

stormckey:
Solved by #9970

PR #9970, opened 2025-02-19
feat: take account type parameters in enum completion

stormckey:
First mentioned in #9652
Now we will consider type parameters (when triggered from undefined type)

image
image

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
@@ -250,21 +250,29 @@ sealed trait Completion {
         additionalTextEdits = additionalTextEdit
       )
 
-    case Completion.EnumCompletion(enm, ap, qualified, inScope) =>
+    case Completion.EnumCompletion(enm, ap, qualified, inScope, withTypeParameters) =>
       val qualifiedName = enm.sym.toString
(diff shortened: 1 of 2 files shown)


main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
```diff
@@ -250,21 +250,29 @@ sealed trait Completion {
...
       val qualifiedName = enm.sym.toString
-      val label = if (qualified) qualifiedName else enm.sym.name
+      val name = if (qualified) qualifiedName else enm.sym.name
       val description = if(!qualified) {
         Some(if (inScope) qualifiedName else s"use $qualifiedName")
       } else None
       val labelDetails = CompletionItemLabelDetails(None, description)
       val additionalTextEdit = if (inScope) Nil else List(Completion.mkTextEdit(ap, s"use $qualifiedName"))
       val priority = if (inScope) Priority.High else Priority.Lower
+      val snippet = if ( withTypeParameters )
```
magnus-madsen, line 262: why extra spaces?

magnus-madsen:
When is withTypeparameters false?

stormckey:
> When is withTypeparameters false?

Depends on where the completer is called from:
image

magnus-madsen (approved):

PR #9972, opened 2025-02-19
feat: introduce QualifiedSym

stormckey:
As discussed in #9775

With IsQualified, we can now check name matching for every IsQualified Symbol, instead of writing a function for each symbol.

We'll do that refactor in next PR.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/Symbol.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/Symbol.scala
@@ -18,7 +18,7 @@ package ca.uwaterloo.flix.language.ast
 
 import ca.uwaterloo.flix.api.Flix
 import ca.uwaterloo.flix.language.ast.Name.{Ident, NName}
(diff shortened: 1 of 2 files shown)


main/src/ca/uwaterloo/flix/language/ast/Symbol.scala
```diff
@@ -320,6 +320,11 @@ object Symbol {
     new OpSym(effectSym, ident.name, ident.loc)
   }
 
+  sealed trait IsQualified{
```
magnus-madsen, line 323: - Rename to `QualifiedSym`
- Move into `ast/shared`

stormckey: The just a trait in ast/shared/QualifiedSym?

magnus-madsen: Yes

magnus-madsen (approved):

Issue #6441, opened 2023-09-10
Rename `MutQueue` to `MutPriorityQueue`


magnus-madsen:

magnus-madsen:
Lets do it.

mlutze:
Why? If I make a queue, I don't expect it to be a priority queue.

magnus-madsen:
Hmm

mlutze:
I'm going to reopen since I think this is worth addressing.

JonathanStarup:
I agree it should be a `MutPriorityQueue`

magnus-madsen:
I agree. Let us rename it back.

PR #9973, opened 2025-02-20
chore: rename MutQueue to MutPriorityQueue

mlutze:
fixes https://github.com/flix/flix/issues/6441


JonathanStarup:
Wait its a priority queue? ive been using it wrong then 😄 for talpin it doesn't matter, but i meant to use a normal queue

magnus-madsen (approved):

PR #9977, opened 2025-02-20
feat: introduce matchesName to CompletionUtils

stormckey:
It works fine:
image
image
image


stormckey:
Do you think it's a good idea to keep this layer?
```scala
  def matchesName(sym: QualifiedSym, qn: Name.QName, qualified: Boolean): Boolean = {
    if (qualified) {
      matchesQualifiedName(sym.namespace, sym.name, qn)
    } else
     fuzzyMatch(qn.ident.name, sym.name)
  }
```

Or should we call fuzzyMatch/matchesQualifiedName directly?

There are some code that calls fuzzyMatch directly, should we keep consistent?

magnus-madsen:
@stormckey Conflicted, sorry :/

magnus-madsen (approved):

PR #9980, opened 2025-02-20
fix: EnumTagCompleter now supports all partially-qualified name

stormckey:
Use the same algorithm used in SignatureCompleter. Now we support the namspace head to be either module or enum:
image
image

Although the algorithm is the same, the pattern matching and field accessing makes it hard to reuse to code:
image


magnus-madsen:
Conflicted.

stormckey:
> Conflicted.

solved.

magnus-madsen (approved):

PR #9984, opened 2025-02-20
experiment: tweak `computeXor`

magnus-madsen:


magnus-madsen:
Need to confirm that this helps.

magnus-madsen:
Maybe this does not work.

magnus-madsen:
Bad idea

PR #9989, opened 2025-02-21
chore: fix typo in name

magnus-madsen:


main/src/ca/uwaterloo/flix/language/phase/unification/zhegalkin/ZhegalkinPerf.scala
```diff
@@ -52,8 +52,8 @@ object ZheglakinPerf {
 
     rq1(r)
     rq2()
-    rq3(r)
     rq4(r)
+    rq3(r)
```
magnus-madsen, line 56: Swapped on purpose because rq3 is so slow

mlutze (approved):

Issue #9976, opened 2025-02-20
Duplicated Completion from ModuleCompleter and others


stormckey:
If a constructs, enum for example, has a non-empty companion namespace, then EnumCompleter and ModuleCompleter will work simultaneously:

<img width="639" alt="Image" src="https://github.com/user-attachments/assets/fe463c6e-dd57-4a8d-8bfc-6372ac3c0721" />

We can not just suppress the completion from construct: when the companion namespace is empty, the module is removed. And only the construct completer can work here.

We can not suppress the module completer trivially. It only knows it's a namespace, with no awareness whether it's a companion namespace.

If we keep empty modules instead of removing them. Then we can just suppress construct completers. The downside of this solution is that they will only have a module symbol, instead of having dedicated symbols from dedicated completers.

Or other solutions. Or you dont think it's a big problem?

magnus-madsen:
> If we keep empty modules instead of removing them. Then we can just suppress construct completers. The downside of this solution is that they will only have a module symbol, instead of having dedicated symbols from dedicated completers.

I think this is OK. Can we do that? What changes are needed?

stormckey:
> I think this is OK. Can we do that? What changes are needed?

Not sure, I need to first find out the code where we remove these modules.

magnus-madsen:
> > I think this is OK. Can we do that? What changes are needed?
> 
> Not sure, I need to first find out the code where we remove these modules.

Is it here: https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/Typer.scala#L64 ?

PR #9994, opened 2025-02-21
feat: keep empty modules

stormckey:
first step to #9976

Now we will have this empty companion modules. So we will has duplicated completion suggestions.

image

Next step will be let ModuleCompleter take the duty of those dedicated completers.

But the Construct Kind is erased: they are all just modules. If we want to keep that, maybe keep a map in the root?

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Typer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Typer.scala
@@ -92,8 +92,8 @@ object Typer {
       val companion_modules =
         for {
           sym <- constructs_with_companion_module
-          modSym = new Symbol.ModuleSym(sym.namespace :+ sym.name)
-        } yield modSym -> Iterable.empty
+          modSym = new Symbol.ModuleSym(sym.namespace :+ sym.name, ModuleKind.Companion)
+        } yield modSym -> List.empty[Symbol]
 
       val groups = syms.groupBy {
         case sym: Symbol.DefnSym => new Symbol.ModuleSym(sym.namespace, ModuleKind.Standalone)
@@ -121,8 +121,8 @@ object Typer {
         case sym: Symbol.HoleSym => throw InternalCompilerException(s"unexpected symbol: $sym", sym.loc)
       }
 
-      groups.map {
-        case (k, v) => (k, v.toList)
+      groups.foldLeft(companion_modules.toMap) {
+        case (acc, (mod, syms)) => acc.updated(mod, syms.toList)
       }
   }
 
--- a/main/src/ca/uwaterloo/flix/language/phase/Typer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Typer.scala
@@ -17,6 +17,7 @@ package ca.uwaterloo.flix.language.phase
 
 import ca.uwaterloo.flix.api.{Flix, FlixEvent}
 import ca.uwaterloo.flix.language.ast.*
+import ca.uwaterloo.flix.language.ast.Symbol.ModuleSym
 import ca.uwaterloo.flix.language.ast.shared.*
 import ca.uwaterloo.flix.language.ast.shared.SymUse.{AssocTypeSymUse, TraitSymUse}
 import ca.uwaterloo.flix.language.dbg.AstPrinter.*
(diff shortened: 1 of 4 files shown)


stormckey:
But wait, if the completer is doing something non-trivial, then it can not just be replaced by the ModuleCompleter.

For example, EnumCompleter knows when to complete with type parameters:
image

This can hardly be done by the ModuleCompleter.

Maybe a better idea is using the map to mute ModuleCompleter if the module is a companion one?

magnus-madsen:
> But the Construct Kind is erased: they are all just modules

I think that is OK

magnus-madsen:
> But wait, if the completer is doing something non-trivial, then it can not just be replaced by the ModuleCompleter.

Right, so the problem is: What's the difference between `Foo.Bar` and `Foo.Bar.Color` where the latter is an enum?

One difference is that we only have to complete `Color[a, b, c]` in a type context. 

We could also try to rank `Foo.Bar.Color[a, b, c]` higher than `Foo.Bar.Color` (module)?

Or we could suppress the module if there is an enum/eff/type-alias

stormckey:
> We could also try to rank `Foo.Bar.Color[a, b, c]` higher than `Foo.Bar.Color` (module)?
I think we always need the type parameter in a type context(or is it?), which means Foo.Bar.Color from module completer is not useful.

We can keep a set of moduleSyms, every of which represents a companion module, in the root and suppress the moduleCompleter over these syms, which I think is the best way here.

magnus-madsen:
We agree: Add a variant type called `shared.ModuleKind = Companion or Standalone`

stormckey:
You can see that CLr is set to a companion Module:
image

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -87,6 +87,14 @@ object Typer {
       }.toSet
       val syms = syms0 ++ namespaces
 
+      val constructs_with_companion_module = traits.keys ++ enums.keys ++ structs.keys ++ effects.keys
```
magnus-madsen, line 90: What happens to the naming? We are not programming C :)

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -87,6 +87,14 @@ object Typer {
...
+      // The companion modules, empty by default.
+      val companion_modules =
+        for {
+          sym <- constructs_with_companion_module
+          modSym = new Symbol.ModuleSym(sym.namespace :+ sym.name, ModuleKind.Companion)
+        } yield modSym -> List.empty[Symbol]
```
magnus-madsen, line 96: I think this whole for-thing is complicated. Can we not do a simple fold or map?

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -113,8 +121,8 @@ object Typer {
         case sym: Symbol.HoleSym => throw InternalCompilerException(s"unexpected symbol: $sym", sym.loc)
       }
 
-      groups.map {
-        case (k, v) => (k, v.toList)
+      groups.foldLeft(companion_modules.toMap) {
```
magnus-madsen, line 124: Perhaps the two maps can be merged with ++? or no?

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
stormckey: My bad, I also wrote some C++ code at that time and I mixed up the convention.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
stormckey: it's in the key. And it turns out that updated will not updated the key.

stormckey: The equal of ModuleSym will only check the ns, not the kind

stormckey:
> Note that there is another problem: image
> 
> If there is something in the module. Then syms in Typer will include `AMod.Clr.foo`, thus it will add Clr to AMod as a standalone module(here it doesn't know Clr is a companion Module), this happens in the `syms.groupBy`.
> 
> As a result, in the use completer. We look into AMod, and will find a EnumSym and ModuleSym there. Then we provide completions for both.
> 
> This duplication happens in UseCompleter. ModuleCompleter will be suppressed as we expect if we check its kind.

I think this problem is hard to solve if we still use the `groupBy`

Maybe we could use `Map[ModuleSym, Set[Symbol]]` internally, add the companion ones first so that they won't be overwritten. Then we add others using fold?

magnus-madsen (approved):

stormckey:
Wait, I can fix this

stormckey:
You want to merge now?

magnus-madsen:
> You want to merge now?

I cancelled. You can fix it

stormckey:
Now it's fixed, every occurrence of Clr is (C)
image

I use a set to keep the symbols and add companion ones first, so that they won't be overwritten.

stormckey:
Well here the compiler wants a Symbol:
image

We need to let QualifiedSym extends Symbol if we want to use the function this way.

But Symbol is sealed and QualifiedSym is in another file.🤔

What do you think? @magnus-madsen 

Or we can cancel this function and inline it in every case.
Or we can move QualifiedSym to Symbol.scala and extends Symbol.

magnus-madsen:
Let's talk more offline

magnus-madsen:
Right now ModuleSyms have a kind-- but only standalone or companion.

Why do we not store what kind of companion it is?

stormckey:
> Let's talk more offline

Ok, but that would be next week.

stormckey:
> Right now ModuleSyms have a kind-- but only standalone or companion.
> 
> Why do we not store what kind of companion it is?

What for?

magnus-madsen:
Lets meet, but it might be a bit.

magnus-madsen:
Failures

stormckey:
I've added quite comprehensive comments for you to understand what's going on here(to remind myself too). We can decide if we should keep these comments, since what we are doing here is not that clear with only the code.

To recap, the goal of this PR is to prevent this duplication:
image

So we need to make sure, for any companion module, it must be marked companion from its parent module. This way, we can filter out companion module symbols and let the EnumSym to work.

What will happen if there is something in the companion module? For example:
```
enum Clr
mod Clr{
    pub def foo(): Unit = ()
}
```
Well, current solution is that this will be a mapping from `ModuleSym(Clr, Standalone)` to `Clr.foo`. It's okay for it to be a standalone module as a key. We just need it to be a companion module from its parent.

Finally let me explain the changes to QualifiedSym.

We used to put in a separate file, but it brings several problems, at least for this PR:
- QualifiedSym is not Symbol. We can not extends Symbol since it's sealed.
- QualifiedSym is not sealed, since it's used in Symbol.scala.

The first point is why the compilation fails previously, the map is from ModuleSym to Symbol, while QualifiedSym is not Symbol(in previous impl).

So I decide to move the trait to Symbol.scala and seal it. But we can discuss.

magnus-madsen:
We need to rework modules.

PR #9996, opened 2025-02-24
feat: emit undefined effect in handler

stormckey:
Currently, we can only see a parse error with withHandler context, and WithHandlerCompleter, which is an evil ctx-dependent completer, will provide completions.

image

So I first change the parser to recover from a handler with only the name(maybe incomplete name).

Then the resolver will find out that it's a undefined effect.

image

And if the name is resolved, there is still an error indicating the missing handler:
image

So the next step:
- add env to UndefinedEff (#9997)
- call EffectCompleter over UndefinedEfffect(yes, currently it's not called) and provide completion with the op snippet. Then we can remove WithHandlerCompleter.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -2488,11 +2488,8 @@ object Parser2 {
           delimiterR = TokenKind.CurlyR,
           context = SyntacticContext.Expr.OtherExpr
         )
-        close(mark, TreeKind.Expr.Handler)
-      } else {
-        val token = nth(0)
-        closeWithError(mark, ParseError.UnexpectedToken(NamedTokenSet.FromKinds(Set(TokenKind.CurlyL)), Some(token), SyntacticContext.WithHandler, loc = currentSourceLocation()))
       }
+      close(mark, TreeKind.Expr.Handler)
     }
 
     private def tryExpr()(implicit s: State): Mark.Closed = {
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -2488,6 +2488,8 @@ object Parser2 {
           delimiterR = TokenKind.CurlyR,
           context = SyntacticContext.Expr.OtherExpr
         )
+      } else {
+        s.errors.append(ParseError.UnexpectedToken(expected = NamedTokenSet.FromKinds(Set(TokenKind.CurlyL)) , actual = Some(nth(0)), sctx = SyntacticContext.Expr.OtherExpr, loc = currentSourceLocation()))
       }
       close(mark, TreeKind.Expr.Handler)
     }
--- a/main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
@@ -1763,7 +1763,6 @@ object Resolver {
             Result.Ok((symUse, rules))
         }
       case Result.Err(error) =>
-        sctx.errors.add(error)
         Validation.Success(Result.Err(error))
     }
   }
--- a/main/test/ca/uwaterloo/flix/language/phase/TestParser.scala
+++ b/main/test/ca/uwaterloo/flix/language/phase/TestParser.scala
@@ -740,6 +740,21 @@ class TestParserRecovery extends AnyFunSuite with TestUtils {
     expectMain(result)
   }
 
+  test("MissingWithBody.01") {
+    val input =
+      """
+        |def foo(): Bool =
+        |    let result = run {
+        |        mutual1(10)
+        |    } with handler AskTell ;
+        |    Assert.eq(Some(84), result)
+        |def main(): Int32 = 123
+        |""".stripMargin
+    val result = check(input, Options.TestWithLibMin)
+    expectErrorOnCheck[ParseError](result)
+    expectMain(result)
+  }
+
   test("MissingCatchBody.01") {
     val input =
       """
--- a/main/test/ca/uwaterloo/flix/language/phase/TestParser.scala
+++ b/main/test/ca/uwaterloo/flix/language/phase/TestParser.scala
@@ -740,21 +740,6 @@ class TestParserRecovery extends AnyFunSuite with TestUtils {
     expectMain(result)
   }
 
-  test("MissingWithBody.01") {
-    val input =
-      """
-        |def foo(): Bool =
-        |    let result = run {
-        |        mutual1(10)
-        |    } with handler AskTell ;
-        |    Assert.eq(Some(84), result)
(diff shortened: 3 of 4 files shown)


magnus-madsen:
I think we still need to manually generate a syntax error that some part is missing.

stormckey:
> I think we still need to manually generate a syntax error that some part is missing.

report that the body/BRACEL is missing? In both cases?

magnus-madsen:
> > I think we still need to manually generate a syntax error that some part is missing.
> 
> report that the body/BRACEL is missing? In both cases?

Something like that? Or that the body is missing

stormckey:
Now we also emit that error no matter if the effect is resolved or not:
image
image

main/test/ca/uwaterloo/flix/language/phase/TestParser.scala
```diff
@@ -740,21 +740,6 @@ class TestParserRecovery extends AnyFunSuite with TestUtils {
...
-        |    let result = run {
-        |        mutual1(10)
-        |    } with handler AskTell ;
-        |    Assert.eq(Some(84), result)
-        |def main(): Int32 = 123
-        |""".stripMargin
-    val result = check(input, Options.TestWithLibMin)
-    expectErrorOnCheck[ParseError](result)
-    expectMain(result)
-  }
```
magnus-madsen, line 756: Can we restore this test?

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -2488,11 +2488,10 @@ object Parser2 {
           delimiterR = TokenKind.CurlyR,
           context = SyntacticContext.Expr.OtherExpr
         )
-        close(mark, TreeKind.Expr.Handler)
       } else {
-        val token = nth(0)
-        closeWithError(mark, ParseError.UnexpectedToken(NamedTokenSet.FromKinds(Set(TokenKind.CurlyL)), Some(token), SyntacticContext.WithHandler, loc = currentSourceLocation()))
+        s.errors.append(ParseError.UnexpectedToken(expected = NamedTokenSet.FromKinds(Set(TokenKind.CurlyL)) , actual = Some(nth(0)), sctx = SyntacticContext.Expr.OtherExpr, loc = currentSourceLocation()))
```
magnus-madsen, line 2492: I would have thought this error should be in the weeder?

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -1763,7 +1763,6 @@ object Resolver {
             Result.Ok((symUse, rules))
         }
       case Result.Err(error) =>
-        sctx.errors.add(error)
```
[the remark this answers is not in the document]
JonathanStarup: Is it because the parser has already given an error? it needs a comment at least yea

main/test/ca/uwaterloo/flix/language/phase/TestResolver.scala
```diff
@@ -1985,4 +1994,23 @@ class TestResolver extends AnyFunSuite with TestUtils {
...
+        |}
+        |
+        |def foo(): Unit = {
+        |    run {
+        |      E.op1()
+        |    } with handler E
+        |}
+        |""".stripMargin
+    val result = check(input, Options.TestWithLibNix)
+    expectErrorOnCheck[ResolutionError.MissingHandlerDef](result)
```
JonathanStarup, line 2013: This feels a bit unexpected to me, is it because you produce "handler E {}" in the parser and then that triggers this error in resolution?

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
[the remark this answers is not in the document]
stormckey: My bad, I forgot to add a comment for this.

The caller of visitHandler will look into Validation.Success and see if there is an error. If there is one, it will add it there.

So at first I found that the error is displayed twice and I remove the one here.

The caller may still receive other Success which brings an error inside? but for this branch. I know it will always be passed to the caller and the error will always be added there.

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
stormckey: Yes the weeder cant.
```scala
    private def visitHandlerExpr(tree: Tree)(implicit sctx: SharedContext): Validation[Expr, CompilationMessage] = {
      expect(tree, TreeKind.Expr.Handler)
      val rules = pickAll(TreeKind.Expr.RunWithRuleFragment, tree)
      mapN(pickQName(tree), /* This qname is an effect */ traverse(rules)(visitRunWithRule)) {
        (eff, handlers) => Expr.Handler(eff, handlers, tree.loc)
      }
    }
``` the weeder only knows if there are any rules. Example {} and Example makes no difference.

main/test/ca/uwaterloo/flix/language/phase/TestResolver.scala
stormckey: There are two errors here:
parser says { is missing
resolver says def is missing

After parser, there is nothing about {}. The Weeder only calls pick all to get all defs and here it gets Nil. As a result, the resolver find out that no def is defined.

magnus-madsen:
!!!jar

main/test/ca/uwaterloo/flix/language/phase/TestResolver.scala
JonathanStarup: @magnus-madsen this again feels like a symptom of making fake sub-parts of expressions. Making new expressions makes a new code-path - avoiding this kind of loss of information.

magnus-madsen:
Does this PR break the current handler completion? I cannot get it to trigger in the above example.

stormckey:
> Does this PR break the current handler completion? I cannot get it to trigger in the above example.

Yes, since this is one of the PRs to replace the old completer by producing more precise errors and then use it to provide completions. 

This is the first step, provide more precise errors.

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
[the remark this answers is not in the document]
magnus-madsen: I understand how. Its because the error is added where visitHandler is called. OK

main/test/ca/uwaterloo/flix/language/phase/TestParser.scala
stormckey: restored

main/test/ca/uwaterloo/flix/language/phase/TestResolver.scala
stormckey: So you think this error should only be emitted in "E {}" case? That way, the weeder should be  able to tell "E{}" from "E "

magnus-madsen:
@stormckey I am merging this so you can continue

PR #10001, opened 2025-02-25
refactor: improve Lexer loc functions

JonathanStarup:
`length` was always 1 and the functions are incorrect if `length != 1` is used. I simplified them and refactored repeated code.


magnus-madsen:
Semantic tokens in VSCode still looks OK?

JonathanStarup:
This is really 100% a refactor - nothing is changed. but i can check

magnus-madsen (approved):

PR #10003, opened 2025-02-25
feat: introduce module kind

stormckey:
As discussed in #9994
We introduce ModuleKind to tell which modules are companion ones.

We will only use it in the completer and the completer will only access the module syms built in Typer.

So the module kind for syms built elsewhere does not matter. I just assume they are all stand alone ones.

It could be complicated to determine the real kind of modules in those places. Or we can introduce a new case Unknown if you prefer

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/Symbol.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/Symbol.scala
@@ -957,7 +957,7 @@ object Symbol {
   /**
     * Module symbol.
     */
-  final class ModuleSym(val ns: List[String], val kind : ModuleKind) extends Symbol {
+  final class ModuleSym(val ns: List[String], val kind: ModuleKind) extends Symbol {
     /**
       * Returns `true` if this is the root module.
       */
--- a/main/src/ca/uwaterloo/flix/language/ast/Symbol.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/Symbol.scala
@@ -18,7 +18,7 @@ package ca.uwaterloo.flix.language.ast
 
 import ca.uwaterloo.flix.api.Flix
 import ca.uwaterloo.flix.language.ast.Name.{Ident, NName}
-import ca.uwaterloo.flix.language.ast.shared.{BoundBy, QualifiedSym, Scope, Source, VarText}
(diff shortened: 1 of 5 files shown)


main/src/ca/uwaterloo/flix/language/ast/shared/ModuleKind.scala
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ca.uwaterloo.flix.language.ast.shared
+
+/**
+  * Represents the kind of module.
+  */
+sealed trait ModuleKind {}
```
magnus-madsen, line 22: Drop extra {}

main/src/ca/uwaterloo/flix/language/ast/shared/ModuleKind.scala
```diff
@@ -0,0 +1,35 @@
...
+
+object ModuleKind {
+  /**
+    * A Companion modules comes with a flix construct.
+    * To be specific, it is a module that is associated with an effect, enum, struct or trait.
```
magnus-madsen, line 27: Replace comment:

A module associated with an effect, enum, struct, or trait.

main/src/ca/uwaterloo/flix/language/ast/shared/ModuleKind.scala
```diff
@@ -0,0 +1,35 @@
...
+    */
+  case object Companion extends ModuleKind
+
+  /**
+    * A Standalone module.
```
magnus-madsen, line 32: Standalone -> standalone

main/src/ca/uwaterloo/flix/language/ast/Symbol.scala
```diff
@@ -957,12 +957,20 @@ object Symbol {
   /**
     * Module symbol.
     */
-  final class ModuleSym(val ns: List[String]) extends Symbol {
+  final class ModuleSym(val ns: List[String], val kind : ModuleKind) extends Symbol {
```
magnus-madsen, line 960: spacing is wrong, should be `val kind: ModuleKind`

magnus-madsen:
@stormckey Many small comments. If you fix them I can merge.

magnus-madsen (approved):

PR #10008, opened 2025-02-26
feat: User facing Type Simplification (and testing)

JonathanStarup:


JonathanStarup:
@magnus-madsen the PR was getting too messy. This PR is just about adding the infrastructure and testing to the user-facing simplified without changing the existing simplifier

main/src/ca/uwaterloo/flix/language/phase/unification/EffUnification3.scala
```diff
@@ -414,9 +414,9 @@ object EffUnification3 {
     *
     * The type `tpe` may contain `Type.Error`.
     */
-  def simplify(tpe: Type): Type = try {
+  def simplify(tpe: Type): Type = {
     // Check if type is too complex to simplify via Zhegalkin polynomials.
-    if (tpe.typeVars.size > SetUnification.MaxVars) {
+    if (tpe.typeVars.size > SetUnification.MaxVars / 2) {
```
JonathanStarup, line 419: This is required to test without stack overflow

JonathanStarup:
@magnus-madsen rfr

main/src/ca/uwaterloo/flix/language/phase/TypeSimplifier.scala
```diff
@@ -0,0 +1,85 @@
...
+import ca.uwaterloo.flix.util.InternalCompilerException
+
+
+/** This type simplification is purely focused on user readability. */
+object TypeSimplifier {
+
+  /**
+    * Simplifies types and effects - intended for user readability.
+    *
+    * This function will never crash and works for any type, no matter how ill-kinded.
```
mlutze, line 31: Bold claim

JonathanStarup: note to the next developer, written as a property 😄

main/src/ca/uwaterloo/flix/language/phase/unification/EffUnification3.scala
```diff
@@ -416,7 +416,7 @@ object EffUnification3 {
     */
   def simplify(tpe: Type): Type = try {
     // Check if type is too complex to simplify via Zhegalkin polynomials.
-    if (tpe.typeVars.size > SetUnification.MaxVars) {
+    if (tpe.typeVars.size > SetUnification.MaxVars / 2) {
```
magnus-madsen, line 419: Why is this not reverted?

main/src/ca/uwaterloo/flix/language/phase/TypeSimplifier.scala
```diff
@@ -0,0 +1,85 @@
...
+    */
+  def simplify(tpe: Type): Type = {
+    try {
+      simplifyInternal(tpe)
+    } catch {
+      case _: InternalCompilerException => tpe
```
magnus-madsen, line 37: This is not acceptable. You can catch a more precise exception, but you cannot catch ICE

JonathanStarup: You also shouldn't crash when trying to display an error - should i just remove the catch?

main/src/ca/uwaterloo/flix/language/phase/unification/EffUnification3.scala
JonathanStarup: Rerunning the solver on the minimized formulas throw stack overflow if this is larger. again note that this only touches type simplification, not regular solving

JonathanStarup:
@magnus-madsen can we merge this now? recap:

TypeSimplifier is the object where user-facing type simplification happens

TestTypeSimplifer makes sure that a) type simplification is not dead code in testing and b) that simplification is somewhat correct, by running std lib with simplified types.

magnus-madsen:
@mlutze Can you review first, thanks.

magnus-madsen:
I need a one sentence description of the goal and one sentence on the means.

main/src/ca/uwaterloo/flix/api/FlixEvent.scala
```diff
@@ -39,7 +41,7 @@ object FlixEvent {
   /**
     * An event that is fired when new type constraints are collected for the given def symbol `sym`.
     */
-  case class NewConstraintsDef(sym: Symbol.DefnSym, tconstrs: List[TypeConstraint]) extends FlixEvent
+  case class NewConstraintsDef(defn: KindedAst.Def, infResult: InfResult, renv: RigidityEnv, tconstrs: List[TraitConstraint], tenv: TraitEnv, eqEnv: EqualityEnv, root: KindedAst.Root) extends FlixEvent
```
magnus-madsen: @mlutze Are we at the point where this needs its own type?

main/src/ca/uwaterloo/flix/language/phase/TypeSimplifier.scala
```diff
@@ -0,0 +1,81 @@
...
+import ca.uwaterloo.flix.language.phase.unification.EffUnification3
+import ca.uwaterloo.flix.util.InternalCompilerException
+
+/** This type simplification is purely focused on user readability. */
+object TypeSimplifier {
+
+  /**
+    * Simplifies types and effects - intended for user readability.
+    *
+    * This function should never crash and works for any type, no matter how ill-kinded.
```
magnus-madsen, line 32: Something about returning an equivalent type (assuming that is true).

And also something about "how" it is simplified.

magnus-madsen: Is it for effects? records what?

main/src/ca/uwaterloo/flix/language/phase/TypeSimplifier.scala
```diff
@@ -0,0 +1,81 @@
...
+    */
+  def simplify(tpe: Type): Type = {
+    // This might throw [[InternalCompilerException]] since it reuses unification infrastructure
```
magnus-madsen: This happens only for ill-kinded types right? Would it not make more sense to simply add that as a precondition and describe it in the docs?

main/src/ca/uwaterloo/flix/language/phase/TypeSimplifier.scala
```diff
@@ -0,0 +1,81 @@
...
+      TypeConstraint.Conflicted(simplify(tpe1), simplify(tpe2), prov)
+  }
+
+  /**
+    * Simplifies all effects in `tpe0`.
+    *
+    * Might throw [[InternalCompilerException]] for non-well-kinded effects because of [[EffUnification3.simplify]].
+    */
+  private def simplifyInternal(tpe0: Type): Type = tpe0 match {
+    case t if t.kind == Kind.Eff => EffUnification3.simplify(t)
```
magnus-madsen: This converts to Zhegalkin form? That can be small, but its not simple or?

Should there not be some extra logic somewhere?

JonathanStarup:
> I need a one sentence description of the goal and one sentence on the means.

The goal is to create a space for user-facing type simplification to live and to add testing that it works

The current implementation (zhegalkin back and forth) is not great but the point of this PR is not to improve simplification, but point is to improve the infrastructure.

On main you can change the simplify method to return Type.Unit and every test will pass

main/src/ca/uwaterloo/flix/language/phase/TypeSimplifier.scala
JonathanStarup: This PR does't change simplification - it is how it is

But follow up work now has better footing to do something better

main/src/ca/uwaterloo/flix/language/phase/TypeSimplifier.scala
JonathanStarup: The point of `simplify` is that all errors call `simplify(tpe)` on whatever types they have. So it is the reponsibility of this function to handle all types.

main/src/ca/uwaterloo/flix/language/phase/TypeSimplifier.scala
JonathanStarup: > Something about returning an equivalent type (assuming that is true).

Good point

> And also something about "how" it is simplified.

I disagree, the contract of the function is just to return more readable types

> Is it for effects? records what?

Anything

main/src/ca/uwaterloo/flix/api/FlixEvent.scala
JonathanStarup: It would just be packed and unpacked again so i don't think there is much point

main/src/ca/uwaterloo/flix/language/phase/TypeSimplifier.scala
magnus-madsen: I am confused. You say the function works for any type, but then you later say it can crash.

So which is it??

main/src/ca/uwaterloo/flix/language/phase/TypeSimplifier.scala
magnus-madsen: I understand

main/src/ca/uwaterloo/flix/api/FlixEvent.scala
magnus-madsen: ConstraintSystem? ConstraintContext?

main/src/ca/uwaterloo/flix/language/phase/TypeSimplifier.scala
JonathanStarup: It is not supposed to crash - but it currently does. This PR doesn't fix that. I wanted to catch the ICE but you said that you would prefer if I didn't.

JonathanStarup: I hope that we agree that formatting a type should not crash. The current implementation just doesn't live up to the interface

magnus-madsen: But that seems to suggest you are introducing behavior that can crash the compiler?

Or is that already the case (?)

JonathanStarup: The behavior is completely unchanged, just moved functions. If you call formatType on a broken type it can crash yea

main/src/ca/uwaterloo/flix/api/FlixEvent.scala
magnus-madsen: I want to hear Matts opinion on this and then we can probably merge.

main/src/ca/uwaterloo/flix/language/phase/TypeSimplifier.scala
magnus-madsen: I would move the comment about throwing from here and down to the call to EffUnification3. Does that make sense?

main/src/ca/uwaterloo/flix/language/phase/TypeSimplifier.scala
```diff
@@ -0,0 +1,83 @@
...
+    case TypeConstraint.Conflicted(tpe1, tpe2, prov) =>
+      TypeConstraint.Conflicted(simplify(tpe1), simplify(tpe2), prov)
+  }
+
+  /**
+    * Simplifies all effects in `tpe0`.
+    *
+    * Might throw [[InternalCompilerException]] for non-well-kinded effects because of [[EffUnification3.simplify]].
+    */
+  private def simplifyInternal(tpe0: Type): Type = tpe0 match {
```
magnus-madsen: -> visitType?

main/src/ca/uwaterloo/flix/api/FlixEvent.scala
JonathanStarup: It feels a bit unrelated to this, but fair

magnus-madsen: True, but now the problem surfaces.

main/src/ca/uwaterloo/flix/language/phase/TypeSimplifier.scala
```diff
@@ -0,0 +1,79 @@
...
+    *
+    * Returns a type that is equivalent to `tpe` but is easier to read.
+    *
+    * This function should never crash and works for any type, no matter how ill-kinded.
+    */
+  def simplify(tpe: Type): Type =
+    visitType(tpe)
+
+  /** Simplifies types inside `tc` with [[simplify]]. */
+  def simplifyConstraint(tc: TypeConstraint): TypeConstraint = tc match {
```
mlutze: Why is this needed? Only for testing?

main/test/ca/uwaterloo/flix/language/TestTypeSimplifier.scala
```diff
@@ -0,0 +1,57 @@
...
+import ca.uwaterloo.flix.language.phase.typer.{ConstraintSolverInterface, InfResult}
+import ca.uwaterloo.flix.language.phase.unification.{EqualityEnv, TraitEnv}
+import ca.uwaterloo.flix.util.Options
+import org.scalatest.funsuite.AnyFunSuite
+
+import java.util.concurrent.ConcurrentLinkedQueue
+
+class TestTypeSimplifier extends AnyFunSuite with TestUtils {
+
+  test("TypeCheckWithSimplifier") {
```
mlutze, line 33: This testing seems very white-box. It seems like it will complicate making changes to the solver.

Why not instead:

1. get all the types in all the constraints
2. make sure each type is equivalent to its simplified version

main/src/ca/uwaterloo/flix/api/FlixEvent.scala
mlutze: I don't think we need to introduce a new type just yet. But I'm also not convinced all these fields are necessary/desirable. See comment below about testing.

main/test/ca/uwaterloo/flix/language/TestTypeSimplifier.scala
JonathanStarup: In step 2, would that be a call to `ConstraintSolver2.fullyUnify` with all variables rigid?

I like it - there will be a function like `check(tpe1): Bool = tpe unifiesWith simplify(tpe)` which is the property, and then you feed it types.

@magnus-madsen do you like this idea?

magnus-madsen: Seems somewhat reasonable ?

main/src/ca/uwaterloo/flix/language/phase/unification/EffUnification3.scala
```diff
@@ -434,7 +434,7 @@ object EffUnification3 {
     implicit val renv: RigidityEnv = RigidityEnv.empty
     implicit val bimap: SortedBimap[Atom, Int] = mkBidirectionalVarMap(Atom.getAtoms(tpe))
 
-    val f0 = toSetFormula(tpe)(withSlack = false, scope, renv, bimap)
+    val f0 = toSetFormula(tpe)(withSlack = true, scope, renv, bimap)
```
JonathanStarup, line 437: Found the first bug - slack variables were always deleted by simplify

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolver2.scala
```diff
@@ -125,6 +125,15 @@ object ConstraintSolver2 {
...
 
+  /**
+    * Returns `true` if `tpe1` and `tpe2` are equivalent types.
+    */
+  def equivalent(tpe1: Type, tpe2: Type)(implicit eqenv: EqualityEnv, flix: Flix): Boolean = {
+    // Mark everything as rigid.
+    val renv = RigidityEnv.ofRigidVars(tpe1.typeVars.map(_.sym) ++ tpe2.typeVars.map(_.sym))
+    ConstraintSolver2.fullyUnify(tpe1, tpe2, Scope.Top, renv)(eqenv, flix).isDefined
+  }
+
```
JonathanStarup, line 135: Everything should be nicer now @mlutze, do you think this function is sensible?

JonathanStarup: oh and also, can this ever crash? I would like this to be safe no matter how terrible the given type is

main/src/ca/uwaterloo/flix/language/phase/unification/EffUnification3.scala
magnus-madsen: But thats OK? or?

magnus-madsen: I guess I mean: NOT OK for type equivalence. OK if simplify is only used for printing.

What to do?

But I guess not documenting it is very bad.

main/test/ca/uwaterloo/flix/language/TestTypeSimplifier.scala
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    implicit val eqEnv: EqualityEnv = root.eqEnv
+
+    // Collect all the types of the AST.
+    val types = mutable.ArrayBuffer.empty[Type]
+    Visitor.visitRoot(
+      root,
+      new Consumer {
+        override def consumeExpr(exp: TypedAst.Expr): Unit = types.append(exp.tpe)
```
magnus-madsen, line 60: // Note: We do not use visitType because it only visits user-facing types. Instead we ...

magnus-madsen:
I like the approach. We just need to hash out a few things.

main/src/ca/uwaterloo/flix/language/phase/unification/EffUnification3.scala
JonathanStarup: I don't think that it is okay, it's arbitrarily hiding information. Why would a simplifier hide anything from the type? `withSlack = true` just means "don't delete variables" which seems like the natural choice

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolver2.scala
```diff
@@ -125,6 +125,15 @@ object ConstraintSolver2 {
     }
   }
 
+  /**
+    * Returns `true` if `tpe1` and `tpe2` are equivalent types.
+    */
+  def equivalent(tpe1: Type, tpe2: Type)(implicit eqenv: EqualityEnv, flix: Flix): Boolean = {
```
mlutze: Nitpick: I prefer functions that are verbs.

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolver2.scala
mlutze: It can probably crash if you have a malformed boolean type like `And[x, y, z]`

mlutze: It looks sensible to me.

main/test/ca/uwaterloo/flix/language/TestTypeSimplifier.scala
```diff
@@ -0,0 +1,70 @@
...
+
+    for (tpe <- getSampleTypes) {
+      val simplifiedType = TypeSimplifier.simplify(tpe)
+      assert(ConstraintSolver2.equivalent(tpe, simplifiedType)(eqEnv, flix), s"\n$tpe\ndoes not unify with\n$simplifiedType")
+    }
+  }
+
+  /** Returns a list of types to use for testing. */
+  private def getSampleTypes: List[Type] = {
+    // To find sample types, check the standard library and pick of the types in the AST.
```
mlutze: "pick of"?

magnus-madsen:
Why is this not merged?

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolver2.scala
magnus-madsen: -> isEquivalent

main/test/ca/uwaterloo/flix/language/TestTypeSimplifier.scala
```diff
@@ -0,0 +1,70 @@
...
+
+    implicit val flix: Flix = new Flix().setOptions(Options.DefaultTest)
+
+    val root = flix.check() match {
```
magnus-madsen: Just use the unsafe .get (it will be easy to debug and this is in a TEST)

magnus-madsen (commented):
@JonathanStarup Fix a handful of things and this can be easily merged.

magnus-madsen:
@JonathanStarup You want to do a pass on this? Otherwise I will close and let someone else pick it up when they work on the effect errors messages.

magnus-madsen:
@JonathanStarup Ping :)

main/test/ca/uwaterloo/flix/language/TestTypeSimplifier.scala
JonathanStarup: but there is also the list of errors, so it'd be a `.get` and `assert(errs.isEmpty)`. You prefer that?

JonathanStarup: we'll probably notice the standard library not compiling

magnus-madsen:
Thanks, lets see how it plays.

PR #10009, opened 2025-02-26
fix: catch `InvalidType` in `EffUnification3`

magnus-madsen:


mlutze (approved):

JonathanStarup:
ah I just fixed this in the other branch. 

why the unused type?

magnus-madsen:
> ah I just fixed this in the other branch.
> 
> why the unused type?

case objects and exceptions do not work well together. Ask Matt

PR #10011, opened 2025-02-26
feat: let EffectCompleter handle UndefinedEffect

stormckey:
WithHandlerCompleter is removed since EffectCompleter now takes over.
Preview:
First it still works well in UndefinedName and UndefinedType:
image
image

In handler, it will offer more:
image
image

I am just not sure if we should add a snippet here? If there are several ops. It may not be a good idea to add anchor to all the ops?

Or we could just add anchor to the first op?

Or just leave it like this?


magnus-madsen:
> I am just not sure if we should add a snippet here? If there are several ops. It may not be a good idea to add anchor to all the ops?

What do you mean  by snippet?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
```diff
@@ -344,20 +344,25 @@ sealed trait Completion {
         additionalTextEdits = additionalTextEdit
       )
 
-    case Completion.EffectCompletion(effect, ap, qualified, inScope) =>
+    case Completion.EffectCompletion(effect, ap, qualified, inScope, inHandler) =>
```
magnus-madsen, line 347: I find it a bit strange that this is one completion

magnus-madsen:
Looks good, but I really think we should separate completion of effect names from completion of entire handlers.

stormckey:
> > I am just not sure if we should add a snippet here? If there are several ops. It may not be a good idea to add anchor to all the ops?
> 
> What do you mean by snippet?

with anchors we can jump through tab.

magnus-madsen:
> > > I am just not sure if we should add a snippet here? If there are several ops. It may not be a good idea to add anchor to all the ops?
> > 
> > 
> > What do you mean by snippet?
> 
> with stops we can jump to through tab.
> 
> I mean, we can set stops at every arg list and body of the required ops. But usually, long after you started editing the first op will you start writing the next. So it could be a little bit useless/ or even annoying to leave a stop that far away?
> 
> But providing stops for the first op makes sense: usually it's just the next place you want to edit.

We can do that-- but lets do it in a follow up PR.

magnus-madsen (approved):

PR #10013, opened 2025-02-26
feat: move global def redundancy check to dependencies

stormckey:
Most relevant component is copied from Redundancy.

The Redundancy phase now stops checking deadDef. Dependencies takes over.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/Flix.scala
+++ b/main/src/ca/uwaterloo/flix/api/Flix.scala
@@ -91,7 +91,6 @@ class Flix {
   private var cachedKinderAst: KindedAst.Root = KindedAst.empty
   private var cachedResolverAst: ResolvedAst.Root = ResolvedAst.empty
   private var cachedTyperAst: TypedAst.Root = TypedAst.empty
-  private var cachedDefDeps: Map[Symbol.DefnSym, SymUse.DefSymUse] = Map.empty
 
   /**
     * A cache of error messages for incremental compilation.
@@ -593,7 +592,7 @@ class Flix {
             val (_, safetyErrors) = Safety.run(afterRedundancy, cachedTyperAst, changeSet)
             errors ++= safetyErrors
 
-            val (afterDependencies, (dependenciesErrors, defDeps)) = Dependencies.run(afterRedundancy, cachedTyperAst, changeSet, cachedDefDeps)
+            val (afterDependencies, dependenciesErrors) = Dependencies.run(afterRedundancy, cachedTyperAst, changeSet)
             errors ++= dependenciesErrors
 
             if (options.incremental) {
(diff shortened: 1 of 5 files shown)


stormckey:
I tried to remove the addDependency for def uses and try to rebuild this edge by from defDeps. But it fails. Do you have any ideas why this will happen?

I've notice that we only add to defDeps if it's not used recursively, does that matter?

stormckey:
I pushed the commit that try to rebuild, and you can see how this commit fails in the incremental check.

Can see the problem of this rebuilding? Am I missing anything? Is it possible for us to rebuild the dependencies?

main/src/ca/uwaterloo/flix/api/Flix.scala
```diff
@@ -91,6 +91,7 @@ class Flix {
   private var cachedKinderAst: KindedAst.Root = KindedAst.empty
   private var cachedResolverAst: ResolvedAst.Root = ResolvedAst.empty
   private var cachedTyperAst: TypedAst.Root = TypedAst.empty
+  private var cachedDefDeps: Map[Symbol.DefnSym, SymUse.DefSymUse] = Map.empty
```
magnus-madsen, line 94: I wanted this map to be part of TypedAst.Root-- or rather DependencyGraph which is part of TypedAst.Root. Is that possible?

main/src/ca/uwaterloo/flix/language/phase/Dependencies.scala
```diff
@@ -717,15 +748,67 @@ object Dependencies {
...
+      */
+    def addDependency(src: SourceLocation, dst: SourceLocation): Unit = {
+      deps.put((src.sp1.source.input, dst.sp1.source.input), ())
+    }
+
+    def buildDependencyGraph(): DependencyGraph = {
+      var allDeps = MultiMap.empty[Input, Input]
+      defDeps.forEach((defn, defnUse) => addDependency(defn.loc, defnUse.loc))
+      deps.forEach((k, _) => allDeps = allDeps + k)
+      DependencyGraph(defDeps.asScala.toMap, allDeps)
```
stormckey, line 785: I keep the logic here instead of inside DependencyGraph since we can still access addDependency here, which is convenient.

main/src/ca/uwaterloo/flix/api/Flix.scala
magnus-madsen: Lets do it 👍

stormckey: Done

magnus-madsen:
> If you have time we can continue with this.

This one will have to wait a bit. Lets put the incremental stuff on ice for a while. I think there are plenty of other things to look into.

PR #10016, opened 2025-02-26
chore: remove ExprCompleter

### LINKED ISSUE #9397: Introduce `ExprCompletion` and `TypeCompletion`
I wonder if it would be useful to implement some marker traits: `ExprCompletion` and `TypeCompletion`?

The goal would be to ensure that we don't accidentally propose some expression in a type context or vise versa.

(There are a few more kinds of completions other than expr and type, but we could start with these.)

### LINKED ISSUE #9569: Remove ExprCompleter and TypeCompleter
We should remove ExprCompleter and TypeCompleter since we are trying to merge their sub-completers with other completers and lift the resulting completers directly to the errors.

A problem remains: ExprCompleter and TypeCompleter are still used in the pattern match for ParseError. It's hard to replace them with a set of merged completers since we can not access the precise error here, SyntacticContext is all we have.

stormckey:
fixes #9569

Finally it's time to remove it.

HoleCompleter is designed to work when there is no errors, good to remove.

The last two we could just inline.

But considering the situations we want the main snippet, it's usually the SyntacticContext of module. So I also remove it:
image

KeywordCompleter is kept for both cases, at least we need it for UndefinedName:
image

For SyntacticContext.Expr, we need more experiments. But it think it's fine for now.


magnus-madsen:
Can we re-add the SnippetCompleter were appropriate? Its only inside a module its relevant.

magnus-madsen:
I would like to still have the completion for `main`. I think this PR removes that, right?

stormckey:
> I would like to still have the completion for `main`. I think this PR removes that, right?

No, I remove it from UndefinedName and SyntacticContext.Expr. It still works for SyntacticContext.Module, which I think is the only case for it to be helpful, or is it?

magnus-madsen:
> > I would like to still have the completion for `main`. I think this PR removes that, right?
> 
> No, I remove it from UndefinedName and SyntacticContext.Expr. It still works for SyntacticContext.Module, which I think is the only case for it to be helpful, or is it?

I see, but your screenshot shows it not working, so I am confused.

stormckey:
Sorry for that, they are all working. I just happened to click somewhere else:
image
image

I didn't even notice that since I am focusing on the error type :)

magnus-madsen (approved):

magnus-madsen:
Conflicted but otherwise ready for merge 👍

stormckey:
solved

PR #10017, opened 2025-02-26
chore: remove unused syntactic context

### LINKED ISSUE #10015: Simplify `SyntacticContext`
We should remove what we don't use.

stormckey:
fixes #10015
Only SyntacticContext used in `CompletionProvider` is reserved:

```scala
        case err: ParseError => err.sctx match {
          // Expressions.
          case SyntacticContext.Expr.Constraint => PredicateCompleter.getCompletions(uri) ++ KeywordCompleter.getConstraintKeywords
          case SyntacticContext.Expr.OtherExpr => ExprCompleter.getCompletions(ctx)

          // Declarations.
          case SyntacticContext.Decl.Enum => KeywordCompleter.getEnumKeywords
          case SyntacticContext.Decl.Effect => KeywordCompleter.getEffectKeywords
          case SyntacticContext.Decl.Module => KeywordCompleter.getModKeywords ++ ExprSnippetCompleter.getCompletions()
          case SyntacticContext.Decl.Struct => KeywordCompleter.getStructKeywords
          case SyntacticContext.Decl.Trait => KeywordCompleter.getTraitKeywords
          case SyntacticContext.Decl.Type => KeywordCompleter.getTypeKeywords

          case _ => Nil
        }
```

But those SyntacticContext is sill used in the Parser/Weeder. For example, it's in the signature of expect:
```scala
  private def expect(kind: TokenKind, context: SyntacticContext, hint: Option[String] = None)(implicit s: State): Unit = {
```

What should we use to replace? SyntacticContext.Unknown?

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
@@ -88,7 +88,7 @@ object CompletionProvider {
         case err: ParseError => err.sctx match {
           // Expressions.
           case SyntacticContext.Expr.Constraint => PredicateCompleter.getCompletions(uri) ++ KeywordCompleter.getConstraintKeywords
-          case _: SyntacticContext.Expr => ExprCompleter.getCompletions(ctx)
+          case SyntacticContext.Expr.OtherExpr => ExprCompleter.getCompletions(ctx)
 
           // Declarations.
           case SyntacticContext.Decl.Enum => KeywordCompleter.getEnumKeywords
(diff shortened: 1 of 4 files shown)


magnus-madsen:
We probably still have to keep SyntacticContext for the KeywordCompleter, but I think some of its cases can be dropped.

stormckey:
> We probably still have to keep SyntacticContext for the KeywordCompleter, but I think some of its cases can be dropped.

Yeah that's what I did.

But the parser is still asking for a SyntacticContext when there is an error. Some of these Context is no longer used by the completer so we remove them. But we still need a placeholder since the parser is always expecting one.

Should we use the SyntacticContext.Unknown or do any other refactoring?

magnus-madsen:
> > We probably still have to keep SyntacticContext for the KeywordCompleter, but I think some of its cases can be dropped.
> 
> Yeah that's what I did.
> 
> But the parser is still asking for a SyntacticContext when there is an error. Some of these Context is no longer used by the completer so we remove them. But we still need a placeholder since the parser is always expecting one context.
> 
> Should we use SyntacticContext.Unknown or do any other refactoring?

Ah yes, then we can use SyntacticContext.Unknown

stormckey:
> Ah yes, then we can use SyntacticContext.Unknown

Done

magnus-madsen (approved):

PR #10028, opened 2025-03-02
feat: remove max comment nesting

### LINKED ISSUE #10023: Remove Lexer `BlockCommentMaxNestingLevel`
* It doesn't make errors better (the lexer immediately fails when the threshold is hit, making all the following `*/` new lexer errors)
* It doesn't help the lexer with anything (it can't loop infinitely or something like that).

Complexity for no benefit as far as I can see.

JonathanStarup:
fixes #10023


magnus-madsen:
tests :)

magnus-madsen (approved):

PR #10029, opened 2025-03-02
feat: require escaping keyword names

### LINKED ISSUE #9213: Decide how to escape keywords
I propose the syntax `$keyword$`, but could this cause problems with Java inner classes?

### LINKED ISSUE #10002: choose and implement keyword function escape
`x.and(y)` is right now parsed as a Java call. but we should require something like `x.$and$(y)` to "escape" the keyword

- choose syntax
- implement syntax

JonathanStarup:
`x.and(y)` now needs to be
```
x.`and`(y)
```
Fixes #10002
Fixes #9213

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/TokenKind.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/TokenKind.scala
@@ -170,6 +170,7 @@ sealed trait TokenKind {
       case TokenKind.NameUpperCase => "<Name>"
       case TokenKind.NameMath => "<math name>"
       case TokenKind.NameGreek => "<greek name>"
+      case TokenKind.NameEscaped => "<Name>"
       case TokenKind.UserDefinedOperator => "<user-defined operator>"
       case TokenKind.Annotation => "<annotation>"
       case TokenKind.BuiltIn => "<built in>"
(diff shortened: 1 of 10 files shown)


JonathanStarup:
PR Lexer: 225,9 ms (07,2%)
Master Lexer: 193,3 ms (06,2%)

magnus-madsen:
@mlutze Can you review and then I will?

JonathanStarup:
I'm not sure this should be merged as is - we should probably think about static names

magnus-madsen:
Hmm. True.

magnus-madsen:
You cannot import methods though

JonathanStarup:
Ah okay, so its mostly `Math.and` or whatever might static things there might be

magnus-madsen:
Marked as draft until completed.

JonathanStarup:
`String.trim` is  parsed the same as `System.out` - allowing static Java things to use escapes means we must allow all(?) QNames to use escapes and then later (after knowing the types) disallow escaping Flix names.

mlutze:
This still alive?

JonathanStarup:
I guess so yeah

magnus-madsen:
My impression is that a clean design is needed... and then some manpower to implement it. 

Could be part of a lexer re-implementation.

PR #10033, opened 2025-03-06
feat: support multiline semantic token

### LINKED ISSUE #10027: Missing comment highlighting
Image

stormckey:
fixes #10027

The issue is that formerly we will filter out multiline tokens. And the multiline-token feature is not supported by all clients(at least nvim doesn't support this by default). So we have to process this manually--split the multiline token into several single-line tokens.

Now the comment block is highlighted properly:
image

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
@@ -21,10 +21,9 @@ import ca.uwaterloo.flix.language.ast.TypedAst.*
 import ca.uwaterloo.flix.language.ast.TypedAst.Predicate.{Body, Head}
 import ca.uwaterloo.flix.language.ast.shared.SymUse.*
 import ca.uwaterloo.flix.language.ast.shared.*
-import ca.uwaterloo.flix.language.ast.{SourceLocation, Symbol, Token, Type, TypeConstructor, TypedAst}
+import ca.uwaterloo.flix.language.ast.{SourceLocation, SourcePosition, Symbol, Token, Type, TypeConstructor, TypedAst}
 import ca.uwaterloo.flix.util.collection.IteratorOps
 
-import scala.collection.immutable.SortedSet
 import scala.collection.mutable.ArrayBuffer
 
 object SemanticTokensProvider {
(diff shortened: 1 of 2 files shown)


main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
```diff
@@ -140,19 +139,34 @@ object SemanticTokensProvider {
...
+    val filteredTokens = allTokens.filter(t => include(uri, t.loc) && !t.loc.isSynthetic)
+
+    // For multiline tokens, we must split them into several single-line tokens
+    val splitTokens = filteredTokens.flatMap {
+      // Do nothing if it's single-line
+      case token@SemanticToken(_, _, loc) if loc.isSingleLine =>
+        token :: Nil
+      // Split the multiline token
+      case SemanticToken(tpe, modifiers, loc) =>
+        val tokens = for (line <- loc.sp1.line to loc.sp2.line) yield {
```
magnus-madsen, line 157: Can we move this into a helper function? (Within this class)

magnus-madsen:
Does the LSP spec say a semantic token cannot span more than one line?

main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
```diff
@@ -1026,6 +1029,26 @@ object SemanticTokensProvider {
...
+    * Splits every multiline token in the given list into single-line tokens.
+    */
+  private def splitToken(tokens: List[SemanticToken]): List[SemanticToken] = {
+    tokens.flatMap {
+      // Do nothing if it's single-line
+      case token@SemanticToken(_, _, loc) if loc.isSingleLine =>
+        token :: Nil
+      // Split the multiline token
+      case SemanticToken(tpe, modifiers, loc) =>
+        val tokens = for (line <- loc.sp1.line to loc.sp2.line) yield {
```
magnus-madsen, line 1042: Can we avoid the shadowing? tokens is already in scope

main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
```diff
@@ -1026,6 +1029,26 @@ object SemanticTokensProvider {
     case _ => false
   }
 
+  /**
+    * Splits every multiline token in the given list into single-line tokens.
+    */
+  private def splitToken(tokens: List[SemanticToken]): List[SemanticToken] = {
+    tokens.flatMap {
```
magnus-madsen, line 1036: flatMap can have complexity O(nˆ2) and this list is very long.

Perhaps we should append to a ArrayBuffer and then convert that to a List at the end

magnus-madsen (approved):

PR #10035, opened 2025-03-06
fix: filter out lib files when publishing diagnostics

stormckey:
Sometimes our code may make the compiler emit some errors for the lib files.

And the uri used there is invalid, which will produce tons of errors in nvim.

So we filter them out.


magnus-madsen (approved):

magnus-madsen:
Merging this, but do see: https://github.com/flix/flix/issues/10036

Issue #10018, opened 2025-02-27
MutCollectable


mlutze:
Collectable cannot be nicely defined for mutable and immutable collections because only mutable collections require a region capability to be constructed.

I propose we add MutCollectable to support mutable collections. We'll also need new syntax for foreach-yield to take a region capability.
I propose:

```scala
foreach (
  x <- y
} yield(rc) {
   x + 1
}

PR #10038, opened 2025-03-07
feat: introduce MutCollectable

mlutze:
related to https://github.com/flix/flix/issues/10018

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/Flix.scala
+++ b/main/src/ca/uwaterloo/flix/api/Flix.scala
@@ -224,6 +224,7 @@ class Flix {
     "Witherable.flix" -> LocalResource.get("/src/library/Witherable.flix"),
     "UnorderedFoldable.flix" -> LocalResource.get("/src/library/UnorderedFoldable.flix"),
     "Collectable.flix" -> LocalResource.get("/src/library/Collectable.flix"),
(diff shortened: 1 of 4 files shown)


mlutze:
> I am not entirely convinced why this is a good abstraction?

I don't know how else to collect into something mutable.

(Also sets us up later to say "Look how ugly this is" when we figure out a better solution)

magnus-madsen:
> I don't know how else to collect into something mutable.

You mean in a polymorphic way? You can already do it non-polymorphically.

> (Also sets us up later to say "Look how ugly this is" when we figure out a better solution)

Sure, but do you have an inkling of how to do that?

mlutze:
> > I don't know how else to collect into something mutable.
> 
> You mean in a polymorphic way? You can already do it non-polymorphically.

Yes but that's sad :(

I want to do

```
let myTrainCars: MutList[Int32, r] = foreach (
  train <- trains
  if train.numCars > 4
  car <- train.cars
) yield(rc) {
  car.number
}
```

> > (Also sets us up later to say "Look how ugly this is" when we figure out a better solution)
> 
> Sure, but do you have an inkling of how to do that?

No :(

magnus-madsen:
yield(rc) :O

magnus-madsen (approved):

magnus-madsen:
I will merge it, but reluctantly :[

PR #10043, opened 2025-03-07
chore: remove unused CompletionContext fields

stormckey:
Now completionContext only has one field: range.

Since we also remove previousWord, we can no longer tell if a def completion comes after |> etc. Relative code is removed too.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
@@ -42,8 +42,8 @@ import ca.uwaterloo.flix.language.phase.Lexer
 object CompletionProvider {
 
   def autoComplete(uri: String, pos: Position, source: String, currentErrors: List[CompilationMessage])(implicit flix: Flix, root: TypedAst.Root): CompletionList = {
-    val ctx = getCompletionContext(source, pos)
-    val completionItems = errorsAt(uri, pos, currentErrors).flatMap({
+    val completionItems = getCompletionContext(source, pos).map { ctx =>
+      errorsAt(uri, pos, currentErrors).flatMap({
         case err: WeederError.UnqualifiedUse =>
           UseCompleter.getCompletions(uri, err)
         case WeederError.UndefinedAnnotation(_, _) => KeywordCompleter.getModKeywords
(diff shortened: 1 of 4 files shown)


main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -124,19 +124,11 @@ object CompletionProvider {
...
-      val prevWord = getSecondLastWord(prefix)
-      val previousWord = if (prevWord.nonEmpty) {
-        prevWord
-      } else lines.slice(y - 1, y).headOption match {
-        case None => ""
-        case Some(s) => getLastWord(s)
-      }
       // Remember positions are one-indexed.
       val range = Range(Position(y + 1, start + 1), Position(y + 1, end + 1))
-      CompletionContext(uri, pos, range, word, previousWord, prefix)
```
magnus-madsen, line 139: All of the above code can now be further simplified right?

magnus-madsen:
I think you can also remove functions like isWordChar and more.

Moreover, don't have to return an option.

stormckey:
I've further simplify it a bit.

I am not sure how to remove isWordChar, inline it?

magnus-madsen:
Can you not also remove getSecondLastWord and getLastWord and so forth?

stormckey:
It do crash over an empty file. So I roll back to the option ver

stormckey:
As for the tokens, we do not have `Source` to lookup the map:
```scala
tokens: Map[Source, Array[Token]])

case class Source(input: Input, data: Array[Char]) extends Sourceable {
```
We only have a string of uri and string of source.

It may not be a good idea to iterate all the tokens.

As for AST visiting. We are not sure what construct is under cursor. So we have to match over all sorts all things, which is tedious.

If there is no elegant way to find a range for all, we should consider asking every error/syntacticContext to bring a correct loc/range.

magnus-madsen (approved):

magnus-madsen:
The lack of a map `URI -> Input` seems to be a recurring problem, but there is no reason we cannot deal with it. Afterall, we control VSCodeLanguageServer.

stormckey:
> The lack of a map `URI -> Input` seems to be a recurring problem, but there is no reason we cannot deal with it. Afterall, we control VSCodeLanguageServer.

Yeah, but it's not that straightforward. And we have another solution: refactor syntacticContext to bring the range.

PR #10044, opened 2025-03-07
refactor: remove `CompletionContext` from `KindCompleter`

stormckey:
Now only change KindCompletion to avoid using CompletionContext.range, as a preview.

This way, we'll need to add range to every Completion, build it from loc(if available), and test if the range from loc is exactly what we want.

Also sometimes we'll meet problems like in KeywordCompletion loc is not available.

Do you think it's okay to impl this way?


magnus-madsen (approved):

magnus-madsen:
Actually, lets get CompletionContext and all the code around it simplified first, and then we can take it from there.

magnus-madsen (approved):

magnus-madsen:
@stormckey I think this is the way to go.

Note that I changed the title.

I suppose we don't need a PR for each provider; we can probably do several at once.

Issue #10046, opened 2025-03-07
Add missing source locations


magnus-madsen:
From `Visitor.scala`: 

- [ ] Add missing loc for `SelectChannelRule`
- [ ] Add missing loc for `HandlerRule`
- [ ] Add missing loc for `MatchRule`
- [ ] Add missing loc for `TypeMatchRule`
- [ ] Add missing loc for `CatchRule`

and update `Visitor.scala` to use them.

PR #10053, opened 2025-03-07
feat: introduce source location to SelectChannelRule

stormckey:
first part for #10046

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
@@ -1938,7 +1938,7 @@ object Weeder2 {
         case (ident, qname, channel :: body :: Nil) => // Shape is correct
           val isRecvFunction = qname.toString == "Channel.recv" || qname.toString == "recv"
           if (isRecvFunction) {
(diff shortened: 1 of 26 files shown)


main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1938,7 +1938,7 @@ object Weeder2 {
         case (ident, qname, channel :: body :: Nil) => // Shape is correct
           val isRecvFunction = qname.toString == "Channel.recv" || qname.toString == "recv"
           if (isRecvFunction) {
-            Result.Ok(SelectChannelRule(ident, channel, body))
+            Result.Ok(SelectChannelRule(ident, channel, body, tree.loc))
```
magnus-madsen, line 1941: Are you 100% sure this has the right span? You might want to set a breakpoint here and check

magnus-madsen (approved):

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
stormckey: This tree is of type SelectRuleFragment, I think that's exactly what we want?

PR #10069, opened 2025-03-10
feat: annotate `java.lang.Number` with primitive effects

jaschdoc:


magnus-madsen:
Do we use this? Do we have to annotate it?

jaschdoc:
> Do we use this? Do we have to annotate it?

No but it is pure, so easy grab.

magnus-madsen (approved):

PR #10072, opened 2025-03-10
feat: introduce SyntacticContext.Instance

### LINKED ISSUE #10056: Add `SyntacticContext.Instance` and use it in `KeywordCompleter`

stormckey:
fixes #10056
Now the three completions can be triggered in instance:
image
image
image

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/shared/SyntacticContext.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/shared/SyntacticContext.scala
@@ -15,9 +15,6 @@
  */
 package ca.uwaterloo.flix.language.ast.shared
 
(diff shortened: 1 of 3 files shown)


main/src/ca/uwaterloo/flix/language/ast/shared/SyntacticContext.scala
```diff
@@ -51,6 +51,8 @@ object SyntacticContext {
     case object OtherExpr extends Expr
   }
 
+  case object Instance extends SyntacticContext
```
magnus-madsen, line 54: An instance is also a declaration, so it should be under Decl

magnus-madsen (approved):

PR #10076, opened 2025-03-10
feat: add LocalScope to Hole

### LINKED ISSUE #10074: Extend `TypedAst.Hole` and `TypedAst.HoleWithName` to have a scope argument
```
    case class Hole(sym: Symbol.HoleSym, sco: LocalScope, tpe: Type, eff: Type, loc: SourceLocation) extends Expr

    case class HoleWithExp(exp: Expr, sco: LocalScope, tpe: Type, eff: Type, loc: SourceLocation) extends Expr
```

stormckey:
fixes #10074

The name convention is env

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
@@ -285,9 +285,9 @@ object Visitor {
     expr match {
       case Expr.Cst(_, _, _) => ()
       case Expr.Var(_, _, _) => ()
-      case Expr.Hole(_, _, _, _) => ()
+      case Expr.Hole(_, _, _, _, _) => ()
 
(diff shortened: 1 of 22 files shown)


magnus-madsen:
> The name convention is env

I find that strange. Perhaps we should rename it, but I suppose we can do so in a followup ticket?

stormckey:
> > The name convention is env
> 
> I find that strange. Perhaps we should rename it, but I suppose we can do so in a followup ticket?

That would be a big project. But yes we can do it later.

magnus-madsen:
> > > The name convention is env
> > 
> > 
> > I find that strange. Perhaps we should rename it, but I suppose we can do so in a followup ticket?
> 
> That would be a big project. I remembered that we inherit this convention from the Resolver impl. But yes we can do it later.

Can you open a ticket? Thanks

magnus-madsen (approved):

Issue #10047, opened 2025-03-07
Resolver does not property visit all sub-expressions in `NamedAst.Expr.Apply` case


magnus-madsen:
Several problems:

- If the first if falls through and the resolution fails then we dont report errors for exps.
- On line 896 we drop the entire apply, not just the base expression. Hence the args get ignored again.

magnus-madsen:
Remember that calling visitExp has side effects, so we should only do it once per expression (including the arguments).

PR #10082, opened 2025-03-10
refactor: stop using context range if possible

stormckey:
Most Completers are now independent of CompletionContext.

The only usage of CompletionContext is KeywordCompleter, PredicateCompleter and ExprSnippetCompleter, the completers on syntacticContext.

In next PR, we should refactor syntacticContext to bring the loc/range. Then we can remove CompletionContext.

I've tested most of the completers after this refactoring and they all work fine now.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
@@ -83,7 +83,7 @@ object CompletionProvider {
             ModuleCompleter.getCompletions(err)
 
         case err: ResolutionError.UndefinedType =>
-          TypeBuiltinCompleter.getCompletions ++
+          TypeBuiltinCompleter.getCompletions(err) ++
             AutoImportCompleter.getCompletions(err) ++
             LocalScopeCompleter.getCompletions(err) ++
             EnumCompleter.getCompletions(err) ++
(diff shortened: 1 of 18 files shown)


magnus-madsen:
This PR is too big for my taste. I am too scared. Can we do one provider at a time?

magnus-madsen:
But before you start, lets talk, because maybe we need the providers to no longer take errors, but just the required information.

stormckey:
> But before you start, lets talk, because maybe we need the providers to no longer take errors, but just the required information.

So what's the plan here? Although it's a big change, I have tested at least one completer for each error to see if the loc it brings is correct.

There are so many completions that we could divide them into tons of PRs.

Or we can do the refactor together with the unwrapping one, since range is also one of the information each completer needs. Do it for every completer in one PR?

magnus-madsen:
I would refactor one provider at a time while also changing the signature not to accept an error, but the necessary argument.

I would also identify the error by using a visitor.

These are several significant changes, so I think it would make sense to do them one at a time and test that provider.

In particular since we don't have any automated tests yet :) :)

stormckey:
> Visiting the AST will enable us to:
> 
> * Have completions for programs that have no errors.
> * Collect more information from the AST that may not be in the error.

OK, but I still need to wait for https://github.com/flix/flix/issues/10047 to impl the second part, right?

magnus-madsen:
We are doing this now in smaller PRs right? So we can close?

PR #10084, opened 2025-03-10
feat: bring SourceLocation with SyntacticContext

stormckey:
Every SyntacticContext now brings a SourceLocation with them except Unknown.

In the parser, we call currentSourceLocation to get the loc.

In Weeder, we use the loc used by nearby exprs(usually tree.loc)


magnus-madsen:
Wait. I don't understand why we need this.

stormckey:
> Wait. I don't understand why we need this.

Because we want to remove CompletionContext. So we need to get the range from SyntacticContext.

magnus-madsen:
I understand now. Its a bigger change than I imagined. Let me think on it.

After all this is only for very few completers.

magnus-madsen:
What completers use this information? Could we not somehow get the same information from the AST or from the tokens?

PR #10088, opened 2025-03-10
feat: annotate `java.lang.ClassLoader` with primitive effects

jaschdoc:


magnus-madsen:
We could also just tag all of them as Sys. Thoughts?

jaschdoc:
> We could also just tag all of them as Sys. Thoughts?

True but it was not that much more work to assign more granular effects. I can still add Sys to each method, though. Also, there are several methods that return `URL`. Should we assign those methods `Net` just to avoid `URL` escaping the net effect?

magnus-madsen:
I just dont think the extra annotations adds that much value.

PR #10090, opened 2025-03-10
feat: annotate `java.lang.StackWalker` with primitive effects

jaschdoc:


main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
```diff
@@ -21,7 +21,7 @@
     "java.lang.Runtime::runFinalization": "Sys, IO",
     "java.lang.Runtime::totalMemory": "Env, NonDet, IO",
     "java.lang.Runtime::version": "Env",
-
+    "java.lang.StackWalker::getInstance": "IO",
```
magnus-madsen, line 24: I wonder if we should not just tag the class and leave it at that.

Like sure you can call getInstance with IO, but very next thing you get the sys effect anyway

jaschdoc: ok

magnus-madsen (approved):

PR #10102, opened 2025-03-11
refactor: `!eof` -> `inBounds`

JonathanStarup:


magnus-madsen:
Not so sure about this one. Mostly because of the name. "eof" seems more specific to me than "withinBounds"

JonathanStarup:
alright - i think its more natural but its up to you

magnus-madsen:
Would you agree to rename to notEof? or better leave it as is?

PR #10107, opened 2025-03-12
refactor: unwrap AutoImportCompleter

stormckey:
We do two things at the same time for one Completer:
- change the signature to get the information it needs directly.
- accept range from the error

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
@@ -98,12 +98,12 @@ sealed trait Completion {
         }
       )
 
-    case Completion.AutoImportCompletion(name, path, ap, labelDetails , priority) =>
+    case Completion.AutoImportCompletion(name, path, range, ap, labelDetails , priority) =>
       CompletionItem(
         label               = name,
         labelDetails        = Some(labelDetails),
         sortText            = Priority.toSortText(priority, name),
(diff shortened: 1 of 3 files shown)


main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
```diff
@@ -593,11 +593,12 @@ object Completion {
     *
     * @param name          the name to be completed under cursor.
     * @param qualifiedName the path of the java class we will auto-import.
+    * @param range         the range of the completion.
     * @param ap            the anchor position.
     * @param labelDetails  to show the namespace of class we are going to import
     * @param priority      the priority of the completion.
     */
-  case class AutoImportCompletion(name:String, qualifiedName: String, ap: AnchorPosition, labelDetails: CompletionItemLabelDetails, priority: Priority) extends Completion
+  case class AutoImportCompletion(name:String, qualifiedName: String, range: Range, ap: AnchorPosition, labelDetails: CompletionItemLabelDetails, priority: Priority) extends Completion
```
magnus-madsen, line 601: Is this the right place to put range in the argument order.

Maybe it is! Just asking!

Usually we have loc: SourceLocation last

magnus-madsen: OK; and that can be done consistently for all completions?

magnus-madsen (approved):

PR #10109, opened 2025-03-12
feat: Improve Hex lexing

JonathanStarup:
- Correctness: `0xFFi32f` is now lexed as a single Error instead of `0xFFi32` followed by `f`.
- Error quality: e.g. `0xi31` now says that the number suffix is wrong and points to `i`.
- Error consistency: errors are - unterminated, Unexpected hex char, and wrong suffix. no special casing for `__` and the like.
  - `TrailingUnderscoreInNumber` -> `UnterminatedHexNumber` (for hex lexing only)
  - `HexLiteralStartsOnUnderscore` -> `MalformedHexNumber`
  - `DoubleUnderscoreInNumber` -> `MalformedHexNumber`
  - (wrong lexing) -> `MalformedHexNumber`
  - (wrong lexing) -> `IncorrectHexNumberSuffix`


JonathanStarup:
(the motivation for this is not hex-centric, it is a method to also apply to number lexing)

magnus-madsen (commented):
I would have expected more test cases :)

Both because there are new errors and because this claims to fix several issues.

JonathanStarup:
I can add more, I just wanted your initial reaction first, if I add more tests for the correctness part, is it then mergeable?

magnus-madsen:
> I can add more, I just wanted your initial reaction first, if I add more tests for the correctness part, is it then mergeable?

OK, I will look and get back

magnus-madsen:
Is this only about numbers or does it also lead into keyword stuff?

JonathanStarup:
just about numbers

magnus-madsen:
test cases?

JonathanStarup:
There is already test cases for hex lexing, they just only expect `LexerError`. The new behaviour (more lexer errors) is tested in the new test

magnus-madsen:
> There is already test cases for hex lexing, they just only expect `LexerError` (before it would be `DoubleUnderscoreInNumber` fx but now its `MalformedHexNumber` so the situation is unchanged in the testing regard).
> 
> The new behaviour (more lexer errors) is tested in the new test

I don't follow. I see lots of new LexerError's (and renamed errors), but I see no specific test cases for them.

We should always have specific test cases.

JonathanStarup:
I agree, what i am saying is that there was no specific tests previously, so this mantains status quo.

I can go through the lexer tests and specify their errors

JonathanStarup:
@magnus-madsen ready for review / merge

main/test/ca/uwaterloo/flix/language/phase/TestLexer.scala
```diff
@@ -295,7 +295,55 @@ class TestLexer extends AnyFunSuite with TestUtils {
          |def f(): Int32 = 0x_1
            """.stripMargin
     val result = compile(input, Options.TestWithLibNix)
-    expectError[LexerError](result)
+    expectError[LexerError.MalformedHexNumber](result)
+  }
+
+  test("LexerError.HexLiteralContinuesAfterSuffix.01") {
```
magnus-madsen, line 301: -> here

magnus-madsen:
What has been baffling me is that the tests are not named after the errors. You forgot to rename them...

In any case, if you ensure that there are 3+ tests for each new lexer error(correctly named) then I will push the button.

(Whether they are new tests or existing tests-- I want three tests per error at least).

JonathanStarup:
> What has been baffling me is that the tests are not named after the errors. You forgot to rename them...

The existing tests don't follow that rule, so how would I know? Do you want me to rename tests according to this for all the hex tests? (they are called `LexerError.Int8.01` etc)

magnus-madsen:
> > What has been baffling me is that the tests are not named after the errors. You forgot to rename them...
> 
> The existing tests don't follow that rule, so how would I know? Do you want me to rename tests according to this for all the hex tests? (they are called `LexerError.Int8.01` etc)

For negative tests, the name of the test should match the error.

magnus-madsen (approved):

PR #10117, opened 2025-03-13
fix: prevent unrelated errors for duplicate names

mlutze:


magnus-madsen:
@mlutze The fuzzer seems unhappy

main/src/ca/uwaterloo/flix/language/phase/Namer.scala
```diff
@@ -177,7 +177,7 @@ object Namer {
       case LookupResult.NotDefined => addDeclToTable(table, ns, name, decl)
       case LookupResult.AlreadyDefined(loc) =>
         mkDuplicateNamePair(name, getSymLocation(decl), loc)
-        SymbolTable.empty
+         table
```
magnus-madsen, line 180: extra space

mlutze:
> @mlutze The fuzzer seems unhappy

I am also unhappy. I can't reproduce it outside the fuzzer :(

mlutze:
> so stuck?

I figured out the issue.

The fuzzer first parses something like

```
enum T {   // incomplete declaration
```

And adds an empty `T` to the context. All is well.

The next time it goes around, it parses
```
enum T {
    case B
}

def foo(): T = T.B
```

and adds `case B` to the context. But only partially (somehow). When we see `T.B` we can resolve it, but `T` itself is still pointing to the old _empty_ `T`.

Invariants are variated in the Typer and a crash happens.

magnus-madsen:
Added on top? That seems crazy

JonathanStarup:
It does yea

magnus-madsen:
Prefix of the file! But it should not add them on top of each other. It should just replace the input file by a growing prefix.

magnus-madsen:
The compiler runs should be independent.

magnus-madsen:
@JonathanStarup How do you cite code? Is that some kind of super power

JonathanStarup:
Just copied the permalink 🤷‍♂️

mlutze:
But does this fuzzer error point to something that could conceivably happen to a real user? I suppose yes?

magnus-madsen:
> But does this fuzzer error point to something that could conceivably happen to a real user? I suppose yes?

Definitely.

magnus-madsen:
Still stuck?

mlutze:
> Still stuck?

Should be fine after https://github.com/flix/flix/pull/10182

magnus-madsen (approved):

PR #10118, opened 2025-03-13
feat: add IDEA inspection profile

magnus-madsen:


JonathanStarup (approved):

mlutze:
Do we need to change .gitignore?

.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -0,0 +1,12 @@
+<component name="InspectionProjectProfileManager">
+  <profile version="1.0">
+    <option name="myName" value="Project Default" />
```
mlutze, line 3: Maybe change the name to Flix Standard or something? Same for the file name.

magnus-madsen:
> Do we need to change .gitignore?

Not sure. It seems to work when its checked in, but I am not Git expert.

mlutze (approved):
We should do the same for some other configuration files.

.idea/inspectionProfiles/Project_Default.xml
magnus-madsen: Am I allowed to? I don't know which names have significance.

mlutze: Changing the value here and file name just does this:

image

magnus-madsen: Can do it next. Let us see what happens when we merge this.

PR #10119, opened 2025-03-13
feat: add weather app

### LINKED ISSUE #9963: Weather App
# Weather App

In order to exercise our effect system,
we'd like to create a Flix application
that provides weather information to the user.

My vision for a minimal first iteration goes something like this:

```
$ java -jar weather.jar
26 degrees, sunny
```

The user launches the app,
and it tells them the current weather conditions.

But once that basic functionality is in place,
there's lots of room for growth and creativity:
ASCII art sunny/cloudy images,
show the week's weather,
etc.

## Design

The design of the application should be _effect-oriented_.
This means that, for example,
accessing the Weather data should be done through an effect,
while calls to the API should be done in an effect handler.

I expect (at least) the following effects:

- `Weather` - for getting the weather data
- `Location` - for getting the user's location
- `Display` - for displaying the information to the user

## API

I have found two reliable API services that I propose we use:

- Open Meteo ([link](https://open-meteo.com)) for weather information
- IP-API ([link](https://ip-api.com/)) for geolocation information

Both are free (for our purposes) and have no sign-up/API key requirements.

stormckey:
fixes #9963 
The first version with only one functionality: print the current location, weather condition and temperature.

It doesn't work now since its dependency: flix-json doesn't compile in the newest version of flix.

Any suggestions to make the implementation more idiomatic? Then I could start adding some other features :)

### DIFF
--- a/examples/apps/weather/flix.toml
+++ b/examples/apps/weather/flix.toml
@@ -1,0 +1,9 @@
+[package]
+name        = "weather"
+description = "show the weather in your current position"
+version     = "0.1.0"
+flix        = "0.58.1"
+authors     = ["Chenhao Gao <chenhaogao123@gmail.com>"]
+
+[dependencies]
+"github:mlutze/flix-json" =  "0.12.17"
--- a/examples/apps/weather/src/Display.flix
+++ b/examples/apps/weather/src/Display.flix
@@ -1,0 +1,17 @@
+mod Display{
+    use Location.Loc
+    use Weather.WeatherInfo
+
+    pub eff Display {
+        def display(loc: Loc, weather: WeatherInfo): Unit
+    }
+
+    pub def displayWithIO(f: Unit -> a \ ef ): a \ {ef - Display + IO} =
+        run {
+            f()
+        } with handler Display {
+            def display(loc, weather, resume) = 
+                println("Weather for ${loc}:\n\t ${weather}") |> resume
+        }
+
+}
--- a/examples/apps/weather/src/Location.flix
+++ b/examples/apps/weather/src/Location.flix
@@ -1,0 +1,35 @@
+mod Location {
+    
+    pub enum Loc(String, String, Float64, Float64)
+
+    instance ToString[Loc] {
+
+        pub def toString(loc: Loc): String = 
+            let Loc.Loc(city, country, lat, lon) = loc;
+            let latDirection = if (lat >= 0.0) "N" else "S";
+            let lonDirection = if (lon >= 0.0) "E" else "W";
+            "${city} (${country}) at ${lat}°${latDirection}, ${lon}°${lonDirection}"
+
+    }
+
+    pub eff Location {
+        def getLocation(): Option[Loc]
+    }
+
+    pub def locationWithHttp(f: Unit -> a \ef ): a \ {ef - Location + HttpWithResult} = {
+        run {
+            f()
+        } with handler Location {
+            def getLocation(resume) = 
+                let url = "http://ip-api.com/json";
+                let locOpt = forM (
+                    resJson <- requestAndParse(url, Map.empty());
+                    lat <- getFloat64FromJObject("lat", resJson);
+                    lon <- getFloat64FromJObject("lon", resJson);
+                    city <- getStringFromJObject("city", resJson);
+                    country <- getStringFromJObject("country", resJson)
+                ) yield Loc.Loc(city, country, lat, lon);
+                resume(locOpt)
+        }
+    }
+}
--- a/examples/apps/weather/src/Util.flix
+++ b/examples/apps/weather/src/Util.flix
@@ -1,0 +1,52 @@
+use Json.JsonElement
+use Json.Parse
+
+instance ToString[JsonElement] {
+
+    pub def toString(json: JsonElement): String = match json {
+        case JsonElement.JsonObject(jobj) => 
+            let map : List[String] = foreach (
+                (k, v) <- jobj
+            ) yield "${k} : ${ToString.toString(v)}, ";
+            let content = List.reduceLeft(String.concat, map) |> Option.getWithDefault("");
+            "{${content}}"
+        case JsonElement.JsonArray(jarray)  => 
+            let array = List.map(e -> ToString.toString(e), jarray);
+            let content = List.reduceLeft(String.concat, array) |> Option.getWithDefault("");
+            "[${content}]"
+        case JsonElement.JsonString(jString) => jString
+        case JsonElement.JsonNumber(jNumber) => ToString.toString(jNumber)
+        case JsonElement.JsonBool(jBool)   => ToString.toString(jBool)
+        case JsonElement.JsonNull          => "Null"
+    }
+}
+
+def requestAndParse(url: String, header: Map[String, List[String]]): Option[JsonElement] \ HttpWithResult = 
+    forM(
+        response <- HttpWithResult.get(url, header) |> Result.toOption ;
+        body <- Parse.parse(Http.Response.body(response))
+    ) yield body
+
+def getFloat64FromJObject(key: String, json: JsonElement): Option[Float64] = match json {
+    case JsonElement.JsonObject(m) => match Map.get(key, m) {
+        case Option.Some(JsonElement.JsonNumber(n)) => BigDecimal.tryToFloat64(n)
+        case _ => None
+    }
+    case _ => None
+}
+
+def getStringFromJObject(key: String, json: JsonElement): Option[String] = match json {
+    case JsonElement.JsonObject(m) => match Map.get(key, m) {
+        case Some(JsonElement.JsonString(s)) => Some(s)
+        case _ => None
+    }
+    case _ => None
+}
+
+def getFromJObject(key: String, json: JsonElement): Option[JsonElement] = match json {
+    case JsonElement.JsonObject(outMap) => match Map.get(key, outMap) {
+        case Some(m) => Some(m)
+        case _ => None
+    }
+    case _ => None
+}
--- a/examples/apps/weather/src/Weather.flix
+++ b/examples/apps/weather/src/Weather.flix
@@ -1,0 +1,50 @@
+mod Weather {
+    use Location.Loc
+    use Json.{JsonElement, Parse}
+
+    pub enum WeatherInfo(String, Float64)
+
+    instance ToString[WeatherInfo] {
+
+        pub def toString(weatherInfo: WeatherInfo): String = 
+            let WeatherInfo.WeatherInfo(condition, temp) = weatherInfo;
+            "${condition}, ${temp}°C"
+
+    }
+
+    def loadJson(f: String): Option[JsonElement] \ FileReadWithResult =
+        forM (
+            descriptions <- FileReadWithResult.read(f) |> Result.toOption;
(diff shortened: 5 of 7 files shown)


examples/apps/weather/src/Util.flix
```diff
@@ -0,0 +1,52 @@
...
+    case _ => None
+}
+
+def getFromJObject(key: String, json: JsonElement): Option[JsonElement] = match json {
+    case JsonElement.JsonObject(outMap) => match Map.get(key, outMap) {
+        case Some(m) => Some(m)
+        case _ => None
+    }
+    case _ => None
+}
```
JonathanStarup, line 52: Nice!

(Nitpick, but your files are lacking a newline at the end)

examples/apps/weather/src/Util.flix
```diff
@@ -0,0 +1,52 @@
+use Json.JsonElement
+use Json.Parse
+
+instance ToString[JsonElement] {
+
+    pub def toString(json: JsonElement): String = match json {
```
mlutze, line 6: This duplicates some code in Json.Write.toPrettyString.

examples/apps/weather/src/Location.flix
```diff
@@ -0,0 +1,35 @@
...
+
+    pub def locationWithHttp(f: Unit -> a \ef ): a \ {ef - Location + HttpWithResult} = {
+        run {
+            f()
+        } with handler Location {
+            def getLocation(resume) = 
+                let url = "http://ip-api.com/json";
+                let locOpt = forM (
+                    resJson <- requestAndParse(url, Map.empty());
+                    lat <- getFloat64FromJObject("lat", resJson);
```
mlutze, line 27: It might be easier to define a `FromJson` instance for `Loc`. The typical pattern is:

```scala
def fromJsonAt(p: Path, json: JsonElement) {
  forM(
    let map = FromJson.fromJsonAt(p, json);             // convert the JSON into a map
    let city = FromJson.getAtKey(p, "city", map);       // get the city key from the map
    ...
  ) yield Loc(city, country, lat, lon)
  ```

That might reduce/eliminate the need for the getXFromJObject code.

File: examples/apps/weather/flix.toml [hunk 4]
```diff
@@ -0,0 +1,9 @@
+[package]
+name        = "weather"
+description = "show the weather in your current position"
+version     = "0.1.0"
+flix        = "0.58.1"
+authors     = ["Chenhao Gao <chenhaogao123@gmail.com>"]
+
+[dependencies]
+"github:mlutze/flix-json" =  "0.12.17"
```
mlutze, line 9: Version 0.12.18 should be up-to-date with master.

mlutze (commented):
```
Weather for Aarhus (Denmark) at 56.1567°N, 10.2153°E:
         Cloudy, 4.8°C
```

Very cool!

examples/apps/weather/src/Location.flix
mlutze: I really need to add documentation to the Json library :D

stormckey: But why does this return None 🤔
image

By the way, this doesn't compile and emits tons of errors, am I doing it wrong?
image
Sometimes I can not annotate "city", like on the lhs of a forM <-

mlutze: > But why does this return None 🤔 

The `path` parameter is confusing. It doesn't actually indicate the path to take, but instead just what path we have already taken so far. It's used for error messages.

> By the way, this doesn't compile and emits tons of errors, am I doing it wrong? Sometimes I can not annotate "city", like on the lhs of a forM <-

The problem there is this: `Json.FromJson[String].fromJsonAt`. We cannot annotate types for instances like that. Instead, the type is just inferred. So it is just `Json.FromJson.fromJsonAt`.

stormckey:
By the way, does dbg still work now?

mlutze:
> By the way, does dbg still work now?

I think it has some issues. Maybe @JonathanStarup knows.

JonathanStarup:
you can debug if you do it like `let sum = dbg(1 + 2); ...`, doint it like `let sum = 1 + 2; dbg(sum); ...` does't work

stormckey:
Now the code will handle all the errors and using fromJson to access elements :)

stormckey:
Instead of creating an error enum for four cases. I think we can also use eff to handle errors like using exceptions. This can either by defining failure effs for every failable operations. Or defining different signatures in a same eff.

Then we can handle this effs in the top level.

Which style do you prefer. This Result based or relying more on eff?

examples/apps/weather/src/Weather.flix
```diff
@@ -0,0 +1,86 @@
...
+    pub enum WeatherInfoPrimitive(Int32, Float64)
+    
+    pub enum WeatherInfo(String, Float64)
+    
+    instance FromJson[WeatherInfoPrimitive] {
+    
+        pub def fromJsonAt(p: Path, json: JsonElement): Result[JsonError, WeatherInfoPrimitive] = 
+            forM (
+                resMap <- fromJsonAt(p, json);
+                currentMap <- getAtKey(p !! "current", "current", resMap);
```
mlutze, line 24: With `getAtKey`, you don't have to give the path of the node you want, just the path you have done so far.

So this can be 
```
currentMap <- getAtKey(p, "current", resMap);
wmoCodeNumber <- getAtKey(p !! "current", "weather_code", currentMap);
```

Sorry for bad documentation :D

mlutze:
> Instead of creating an error enum for four cases. I think we can also use eff to handle errors like using exceptions. This can either by defining failure effs for every failable operations. Or defining different signatures in a same eff.
> 
> Then we can handle this effs in the top level.
> 
> Which style do you prefer. This Result based or relying more on eff?

This kind of decision we leave up to you :) There are advantages and disadvantages to both styles. You should do whatever seems the most natural.

examples/apps/weather/src/Weather.flix
stormckey: Oh I forgot that. Now it's fixed:)

examples/apps/weather/src/Error.flix
```diff
@@ -0,0 +1,24 @@
+mod Error{
```
mlutze, line 1: I think We don't need this namespace because all errors are WeatherError right? So we just need the WeatherError namespace.

examples/apps/weather/src/Error.flix
```diff
@@ -0,0 +1,24 @@
...
+    use Error.WeatherError
+    use Json.JsonError
+
+    pub enum WeatherError{
```
mlutze, line 5: Missing space

examples/apps/weather/src/Util.flix
```diff
@@ -0,0 +1,9 @@
+use Error.WeatherError
+use Json.JsonElement
+use Json.Parse
+
+def getAndParse(url: String, header: Map[String, List[String]]): Result[WeatherError, JsonElement] \ HttpWithResult =
```
mlutze, line 5: Should be in a Util namespace.

examples/apps/weather/src/Error.flix
```diff
@@ -0,0 +1,24 @@
...
+        case ParseWMOCodeError(String)
+    }
+    
+    instance ToString[WeatherError] {
+    
+        pub def toString(we: WeatherError): String = match we {
+            case Error.WeatherError.RequestError(err)      => ToString.toString(err)
+            case Error.WeatherError.JsonError(_err)         => "JsonError"
+            case Error.WeatherError.ParseError               => "ParseError"
+            case Error.WeatherError.ParseWMOCodeError(errString) => errString
```
mlutze, line 18: Arrows should either be aligned have no extra spaces.

mlutze (commented):
The logic in this is looking very nice.

There are some minor stylistic things to address throughout (I didn't mark all of them)

I'd also like to ask you to add brief comments any nontrivial function.

examples/apps/weather/src/Error.flix
```diff
@@ -0,0 +1,26 @@
+mod WeatherError {
+    use Json.JsonError
+
+    ///
+    /// All sorts of errors we may encounter during the request
+    ///
+    pub enum WeatherError {
```
magnus-madsen, line 7: This should be one level out. Otherwise you have WeatherError.WeatherError -- which is nonsense.

examples/apps/weather/src/Weather.flix
```diff
@@ -0,0 +1,91 @@
...
+                    content <- getAtKey(Path.Root !! nString !! "day" , "description", dayDescriptionMap)
+                ) yield content;
+                res |> Result.mapErr( err ->
+                    WeatherError.JsonError(err)
+                )
+        }
+
+    pub eff Weather {
+        def getWeather(loc: Loc): Result[WeatherError, WeatherInfo]
+    }
```
mlutze, line 67: This should move into the companion namespace if possible. (one level out).

examples/apps/weather/src/Display.flix
```diff
@@ -0,0 +1,20 @@
...
+
+    ///
+    /// Handles display effect by printing to stdout
+    ///
+    pub def displayWithIO(f: Unit -> a \ ef): a \ ef - Display + IO =
+        run {
+            f()
+        } with handler Display {
+            def display(loc, weather, resume) = 
+                println("Weather for ${loc}:\n\t ${weather}") |> resume
```
mlutze, line 17: This line seems a little strange. You're piping the result of `println`, which is `Unit`.

I think it's a little clearer if you just do
```
println(...);
resume()
```

examples/apps/weather/src/Location.flix
```diff
@@ -0,0 +1,55 @@
...
+    /// Handles Location effect by sending http request to ip-api.com
+    ///
+    pub def locationWithHttp(f: Unit -> a \ ef): a \ ef - Location + HttpWithResult + IO = {
+        run {
+            f()
+        } with handler Location {
+            def getLocation(resume) = 
+                let url = "http://ip-api.com/json";
+                let loc = Util.getAndParse(url, Map.empty()) |> Result.flatMap ( resJson ->
+                    fromJsonAt(Path.Root, resJson) |> Result.mapErr( err ->
```
mlutze, line 48: This can just use `fromJson`

examples/apps/weather/src/Weather.flix
```diff
@@ -0,0 +1,91 @@
...
+        run {
+            f()
+        } with handler Weather {
+        def getWeather(loc, resume) = 
+            let Loc.Loc(_, _, lat, lon) = loc;
+            let url = "https://api.open-meteo.com/v1/forecast?latitude=${lat}&longitude=${lon}&current=temperature_2m,weather_code";
+            let weather = Util.getAndParse(url, Map.empty()) |> Result.flatMap( resJson ->
+                match fromJson(resJson) {
+                    case Result.Ok(weatherInfoPrimitive)  => 
+                        let WeatherInfoPrimitive(wmo_code, temp) = weatherInfoPrimitive;
```
mlutze, line 82: Flix uses camelcase, so `wmoCode`

mlutze (commented):
Just a few small things.

magnus-madsen:
@mlutze Last chance for comments. Otherwise merge.

mlutze (approved):

PR #10124, opened 2025-03-14
fix: remove multi-line type token for handler

### LINKED ISSUE #10092: Bad Semantic Token in Handler
Image

As you can see, most tokens in the handler is colored green. This is because the lsp returned a HUGE type token that spans the whole op. This used to be suppressed since it's a multi-line token. But now it's rendered.

I debug it and find out that this token is from the fparams in:
https://github.com/flix/flix/blob/baf70ce9c3bceed42efbe1d8209410ce19fd1a28/main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala#L598-L608

Here is fparam include the hole op(see the first screenshot). 

Should we just stop visiting the fparam here to avoid this problem?

stormckey:
fixes #10092
Now it looks like:
image

As you can see, resume is not colored, since it's also included in the fparams:
image

So actually we just need to remove the first one, but what is the rule to filter out that? I am not completely sure about why it's there, the whole handler from 43-50 is treated as a fparam?

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
@@ -602,8 +602,7 @@ object SemanticTokensProvider {
         case (acc, HandlerRule(op, fparams, exp ,loc)) =>
           val st = SemanticToken(SemanticTokenType.Type, Nil, op.loc)
           val t1 = Iterator(st)
-          val t2 = visitFormalParams(fparams)
-          acc ++ t1 ++ t2 ++ visitExp(exp)
+          acc ++ t1 ++ visitExp(exp)
       }
       st1 ++ st2
 


magnus-madsen:
Why is this the right fix? You are just not visiting the formal parameters...

I think its the source location of one of the formal params that have to be "shortened".

magnus-madsen:
@JonathanStarup May be able to help

stormckey:
> I think its the source location of one of the formal params that have to be "shortened".

I think there is one param that refers to the whole handler and we should just ignore that one. But I need to find a way to identify it? Maybe @JonathanStarup knows the best criteria?

stormckey:
There is no need to provide a token for the handler, we just need to visit it and every sub-component will has its own semantic token.

From the screenshot above, the only part that is missing is the resume, which is the second element of fparams. It will also be colored after we filter out the first one.

JonathanStarup:
You have to visit all the fparams, but I've made a PR  #10128 to fix the location of that first fparam

stormckey:
> You have to visit all the fparams, but I've made a PR #10128 to fix the location of that first fparam

Ok, I'll test after that PR is merged. If it works fine, we could just close this one.

PR #10128, opened 2025-03-14
fix: handler unit padding location

### LINKED ISSUE #10092: Bad Semantic Token in Handler

JonathanStarup:
fixes #10092

Also fixes a an issue around people calling their handler continuation `_unit`, e.g. `def op(_unit) = 42`

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
@@ -751,8 +751,8 @@ object Weeder2 {
       }
     }
 
-    def unitFormalParameter(loc: SourceLocation, name: String = "_unit"): FormalParam = FormalParam(
-      Name.Ident(name, SourceLocation.Unknown),
+    def unitFormalParameter(loc: SourceLocation): FormalParam = FormalParam(
+      Name.Ident("_unit", SourceLocation.Unknown),
       Modifiers(List.empty),
       Some(Type.Unit(loc)),
       loc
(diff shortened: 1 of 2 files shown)


main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1793,11 +1793,24 @@ object Weeder2 {
...
+        // `def f(x, y, ..)` is unchanged.
+        fparams0 match {
+          case fparam :: Nil =>
+            // Since a continuation argument must always be there, the underlying function needs a
+            // unit param. For example `def f(k)` becomes `def f(_unit: Unit, k)`.
+
+            // The new param has the zero-width location of the actual argument.
+            val loc = SourceLocation.zeroPoint(isReal = false, fparam.loc.sp1)
+            // We already have an argument, which could be called "_unit", so use a backup.
+            val name = if (fparam.ident.name == "_unit") "_unit1" else "_unit"
```
magnus-madsen, line 1808: ???

JonathanStarup: Turns out expressions don't check for duplicate `_unit` names, so I can omit this.

magnus-madsen (approved):

PR #10129, opened 2025-03-14
feat: unwrap MagicMatchCompleter

stormckey:


main/src/ca/uwaterloo/flix/api/lsp/provider/completion/MagicMatchCompleter.scala
```diff
@@ -44,14 +43,17 @@ object MagicMatchCompleter {
     *  case (Green, Circle) => ???
     *  case (Green, Square) => ???
     * }
+    *
+    * @param baseLocation The base location of the completion.
+    * @param tpe          The type of the expression.
+    * @param range        The location of the completion.
     */
-  def getCompletions(err: TypeError.FieldNotFound)(implicit root: TypedAst.Root): Iterable[Completion] = {
+    def getCompletions(baseLocation: SourceLocation, tpe: Type, range: Range)(implicit root: TypedAst.Root): Iterable[Completion] = {
```
magnus-madsen, line 51: I think baseLocation should be last argument. Also it is a bad name and is not explained.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/MagicMatchCompleter.scala
```diff
@@ -44,14 +43,17 @@ object MagicMatchCompleter {
...
     for {
-      baseExp <- err.base.text
-      patternMatchBody <- mkPatternMatchBody(err.tpe)
+      baseExp <- baseLocation.text
```
magnus-madsen, line 53: Using .text is the same kind of issue as CompletionContext. What else can be done?

stormckey: Can we get a baseQName from the error?

stormckey: It seems the base is an expr. Hmmmm

We can take the expr. Then get the start position of it. Then make an additional insertion to add the "match" before the base expr.

stormckey: What do you think?

stormckey:
I've submitted all the PRs to pass range in to the completer, which means we can remove CompletionContext after all of them are merged.

If we don't want to be blocked for this one, we can merge it first and then update the completer to be more robust in a following PR. Or I could just re-impl it the way I mentioned above

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/MagicMatchCompleter.scala
stormckey: Updated

PR #10141, opened 2025-03-14
feat: implement `BPlusTree`

JumpTwiice:
We have implemented a concurrent mutable B+Tree that functions as a map. It does not currently support deletion, as it is not needed in our project (Datalog v3).


magnus-madsen (commented):
I will give you some very broad feedback before looking at the code in detail. 

- [ ] No TODOs in source code. They should be fixed or should be tickets.
- [ ] The BTree should come with a general documentation comment that explains the overall structure and how its implemented. It could include references to relevant papers etc.
- [ ] Every struct should have a doc comment that explains its invariants.
- [ ] In continuation, every declaration should have comment.
- [ ] The ratio of of code : tests is low. I would have expected more tests than there is code.
- [ ] Remember that small units tests are good; not just complicated tests.
- [ ] I think every unchecked_cast should have a comment explaining why it is required.

Other comments to think about:

- Why is order a field? Do you plan to have non-uniform trees?
- I think its usually called "thread-safe" and not concurrency safe, but I could be wrong.
- In a function like isEmpty you access tree->root without a lock. I don't think this thread-safe is safe.

CasperDN:
Remarks to the 'other comments':
1): Order is a field on the tree struct. A tree therefore has a particular order and is thus non-uniform. All nodes have a 'size' attribute, but that has to do with how many elements are currently inserted in them, not the order of the tree.
2): Checking Java-doc for standard classes reveals that you are correct. We have therefore changed the documentation to use 'thread-safe'
3): The last was indeed an error, with respect to how we have otherwise named functions

main/src/library/BPlusTree.flix
```diff
@@ -0,0 +1,763 @@
...
+                key: k,
+                val: v,
+                insertionPoint: Int32,
+                order: Int32,
+                rc: Region[r]): Option[Node[k, v, r]] \ r with Order[k] =
+            let prev_mid = order / 2;
+            let (mid, insertInLeft) = if (insertionPoint <= prev_mid) (prev_mid, true) else (prev_mid + 1, false);
+            let rightKeys = Array.empty(rc, order);
+            let rightValues = Array.empty(rc, order);
+            Array.copyInto(srcPos = mid, dstPos = 0, len = order - mid, src = leftLeaf->keys, rightKeys);
```
JonathanStarup, line 507: instead of empty and then copyInto, isn't it easier and safer to use `Array.slice`?

main/src/library/BPlusTree.flix
```diff
@@ -0,0 +1,763 @@
...
+
+        ///
+        /// Split `leftLeaf` into two leaves and push the middle key into the parent,
+        /// potentially causing the parent to split and return a new root
+        ///
+        /// Lock invariants:
+        /// - On invocation, the thread has a lock on `leftNode`.
+        /// - On return, the thread only holds a lock if the return value is Some.
+        ///   In that case it holds a lock on the new root and the old root.
+        def splitLeaf(
```
JonathanStarup, line 496: is there no requirements about the size of leftLeft, insertionPoint, and order?

main/src/library/BPlusTree.flix
```diff
@@ -0,0 +1,763 @@
...
+    ///     `lock` is the lock associated with the node. When it is held no other thread may modify the contents
+    ///     The only exception to the previous rule is that the `parent` pointer is owned by the `lock` on the parent
+    ///
+    pub struct Node[k, v, r] {
+        keys:       Array[k, r],
+        children:   Array[Node[k, v, r], r],
+        values:     Array[v, r],
+        lock:       Lock[r],
+        mut size:   Int32,
+        mut parent: Option[Node[k, v, r]],
```
JonathanStarup, line 287: I think there is a conceptual issue here, because before we have a parent, no lock protects this field. This means that two threads could locally update the None parent into two different Some pointers.

I think maybe the code is okay, but conceptually the node lock protects the pointer while it is none, then the parent pointer protects it. Its a tricky situation so we should double double check it.

main/src/library/BPlusTree.flix
```diff
@@ -0,0 +1,763 @@
...
+            rightLeaf->parent = leftLeaf->parent;
+            pushKeyToParent(leftLeaf->parent, leftLeaf, rightLeaf, Array.get(0, rightKeys), order, rc)
+
+
+        ///
+        /// Split `leftNode` into two nodes with `newKey` inserted into the appropriate node
+        /// depending on `insertionPoint`.
+        ///
+        /// Lock invariants:
+        /// - On invocation, the thread has a lock on `parent`.
```
JonathanStarup, line 538: `leftNode` and `insertedChild`?

main/src/library/BPlusTree.flix
```diff
@@ -0,0 +1,763 @@
...
+            Lock.lock(rightLeaf->lock);
+            lockParent(leftLeaf);
+            rightLeaf->parent = leftLeaf->parent;
+            pushKeyToParent(leftLeaf->parent, leftLeaf, rightLeaf, Array.get(0, rightKeys), order, rc)
+
+
+        ///
+        /// Split `leftNode` into two nodes with `newKey` inserted into the appropriate node
+        /// depending on `insertionPoint`.
+        ///
```
JonathanStarup, line 536: preconditions about sizes and order?

main/src/library/BPlusTree.flix
```diff
@@ -0,0 +1,763 @@
...
+                newKey: k,
+                order: Int32,
+                rc: Region[r]): Option[Node[k, v, r]] \ r with Order[k] =
+            Lock.unlock(insertedChild->lock);
+            let prev_mid = order / 2;
+            let (mid, insertInLeft, rightSize) = if (insertionPoint >= prev_mid) (prev_mid + 1, false, order - prev_mid) else (prev_mid, true, order - prev_mid);
+            let rightKeys = Array.empty(rc, order - 1);
+            let rightChildren = Array.empty(rc, order);
+            leftNode->size = mid;
+            let keyOfRight = if (insertInLeft) {
```
JonathanStarup, line 554: instead of making two empty arrays and copying into them, use `let (keyOfright, rightKeys, rightChildren) = ..` and use `Array.slice`

main/src/library/BPlusTree.flix
```diff
@@ -0,0 +1,763 @@
...
+                let keyToPass_ = Array.get(mid - 1, leftNode->keys);
```
JonathanStarup, line 555: remove the `_`

main/src/library/BPlusTree.flix
```diff
@@ -0,0 +1,763 @@
...
+                    Array.put(newKey, childIndexInRight - 1, rightKeys);
+                    // If insertionPoint is the last index we are done, otherwise insert rest of list.
+                    if(childIndexInRight != rightSizeWithoutNew) {
+                        Array.copyInto(srcPos = mid + childIndexInRight - 1, dstPos = childIndexInRight, len = rightSizeWithoutNew - childIndexInRight, src = leftNode->keys, rightKeys);
+                        Array.copyInto(srcPos = mid + childIndexInRight, dstPos = childIndexInRight + 1, len = rightSizeWithoutNew - childIndexInRight, src = leftNode->children, rightChildren)
+                    } else ();
+                    Array.get(mid - 1, leftNode->keys)
+                }
+            };
+            let rightNode = new Node @ rc {
```
JonathanStarup, line 588: Could you make two functions, mkLeaf and mkNonLeaf which only takes the non-empty arguments? easier to read and maintain.

main/src/library/BPlusTree.flix
```diff
@@ -0,0 +1,763 @@
...
+            rightNode->parent = leftNode->parent;
+            pushKeyToParent(leftNode->parent, leftNode, rightNode, keyOfRight, order, rc)
+
+        ///
+        /// Push the key `newKey` to `parent`.
+        /// If `parent` is `Some(node)`, insert `newKey` into node if there is room. Otherwise split
+        /// `node` and return a new root node.
+        ///
+        /// If `parent` is `None`, create a new root node, insert `newKey` into it and return it.
+        ///
```
JonathanStarup, line 614: order/size assumptions?

main/src/library/BPlusTree.flix
```diff
@@ -0,0 +1,763 @@
...
+                };
+                Lock.lock(newRoot->lock);
+                leftChild->parent = Some(newRoot);
+                rightChild->parent = Some(newRoot);
+                Lock.unlock(rightChild->lock);
+                Some(newRoot)
+        }
+
+        ///
+        /// Applies `f` to all key-value pairs `(k, v)` in the tree rooted in `node` where `min <= k <= max`.
```
JonathanStarup, line 668: is `f` applied in ascending order?

main/src/library/BPlusTree.flix
```diff
@@ -0,0 +1,763 @@
...
+                    else ()
+            } else
+                traverseRight(f, index, max, minLeaf)
+
+        ///
+        /// Applies `f` to key-value pairs `(k, v)` starting at `node->keys[index]` until `node->keys[index] > max`.
+        ///
+        /// Uses the `next` pointer to traverse to the next node.
+        ///
+        def traverseRight(
```
JonathanStarup, line 692: this is thread-unsafe

main/src/library/BPlusTree.flix
```diff
@@ -0,0 +1,763 @@
...
+
+        /////////////////////////////////////////////////////////////////////////////
+        // Helper methods                                                          //
+        /////////////////////////////////////////////////////////////////////////////
+
+        ///
+        /// Insert `x` at position `i` in array `arr` as though it currently has
+        /// `len` elements. Shifts elements at position `[i, .., len]` to `[i+1, .., len+1]`.
+        ///
+        def arrayInsert(x: a, i: Int32, len: Int32, arr: Array[a, r]): Unit \ r =
```
JonathanStarup, line 716: couldn't this use `Array.size`?

This takes linear time in the number of elements, is that okay for how its used?

main/src/library/BPlusTree.flix
```diff
@@ -0,0 +1,763 @@
...
+        ///
+        /// Returns `true` if the pointer of `o1` is equal to the pointer of `o2`.
+        ///
+        def ptrEqual(o1: v, o2: v): Bool =
+            // Unchecked_cast is here used to transform the values to Object.
+            // The purpose is to be able to give them to a Java class which performs pointer comparison.
+            let obj1 = unchecked_cast(o1 as Object);
+            let obj2 = unchecked_cast(o2 as Object);
+            // Unchecked_cast is here used to remove the IO effect from interacting with Java.
+            unchecked_cast(Objects.equals(obj1, obj2) as _ \ {})
```
JonathanStarup, line 732: put `(Objects.equals(obj1, obj2): _ \ IO)` on the inside so we know that you are removing IO

(sorry about this stuff, hopefully we will improve it soon)

main/src/library/BPlusTree.flix
```diff
@@ -0,0 +1,763 @@
...
+
+        ///
+        /// Perform a binary search and return the index of `key` in `arr` if it exists.
+        /// Otherwise, return `(-insertionPoint - 1)` where `insertionPoint` is the index
+        /// where `key` would be inserted.
+        ///
+        def binarySearch(key: v, length: Int32, arr: Array[v, r]): Int32 \ r with Order[v] =
```
JonathanStarup, line 739: could this not use `Array.length`?

JumpTwiice: The arrays in the code are dynamic lists that don't expand their capacity. `Array.length` in this case would be the 
list's "capacity" and `len` would be the current number of elements. This is to avoid an extra level of indirection with MutList.

I agree that your suggestion would be more clean, but it adds another level of indirection that affects performance.

main/src/library/BPlusTree.flix
```diff
@@ -0,0 +1,763 @@
...
+        /// Attempts to release `lock`.
+        ///
+        pub def unlock(lock: Lock[r]): Unit \ r =
+            let Lock(l) = lock;
+            unchecked_cast((l.unlock(): _ \ IO) as _ \ r)
+
+        ///
+        /// Returns `true` if `lock` is currently held by a thread.
+        ///
+        pub def isLocked(lock: Lock[r]): Bool \ r =
```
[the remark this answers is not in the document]
CasperDN: It is unused by the implementation of the tree, but it is used in tests to verify that all threads release their locks after operating on the tree.

main/src/library/BPlusTree.flix
```diff
@@ -0,0 +1,763 @@
...
+    /// Lock invariants:
+    ///     `lock` is the lock associated with the node. When it is held no other thread may modify the contents
+    ///     The only exception to the previous rule is that the `parent` pointer is owned by the `lock` on the parent
+    ///
+    pub struct Node[k, v, r] {
+        keys:       Array[k, r],
+        children:   Array[Node[k, v, r], r],
+        values:     Array[v, r],
+        lock:       Lock[r],
+        mut size:   Int32,
```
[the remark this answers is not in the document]
JumpTwiice: See https://github.com/flix/flix/pull/10141#discussion_r2000982791

main/src/library/BPlusTree.flix
```diff
@@ -0,0 +1,763 @@
...
+                root = node,
+                order = order,
+                rootLock = Lock.mkLock(rc),
+                rc = rc
+            }
+
+    ///
+    /// Returns the lock of the root pointer of `tree`.
+    ///
+    pub def lock(tree: BPlusTree[k, v, r]): Lock[r] = tree->rootLock
```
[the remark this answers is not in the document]
CasperDN: It is something we have only used for testing. We could remove the part of the tests that depends on it?
It should definitely not be used by users.

main/src/library/BPlusTree.flix
JonathanStarup: Oh right, they're always `order` long, makes sense. Maybe noting that in relevant doc-comments would be good

main/src/library/BPlusTree.flix
CasperDN: I believe Array.slice creates a new array of however many elements were copied. We need an array of size order.

main/src/library/BPlusTree.flix
```diff
@@ -0,0 +1,763 @@
...
+        pub def insertIntoLeaf(
+                key: k,
+                val: v,
+                order: Int32,
+                rc: Region[r],
+                node: Node[k, v, r]): Option[Node[k, v, r]] \ r with Order[k] =
+            let size = node->size;
+            let index = binarySearch(key, size, node->keys);
+            if (index >= 0)
+                let _ = Array.put(val, index, node->values);
```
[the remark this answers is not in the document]
CasperDN: It was to avoid putting brackets around the else branch, as Flix would not otherwise parse the the expression. It has been changed such that it uses brackets.

main/src/library/BPlusTree.flix
CasperDN: Conceptually it is indeed a problem. In practice only the root will have None. pushKeyToParent keeps the lock on the old root (the node with None as parent) until the root lock has been acquired after which its parent is updated. Any node acquiring the old root's lock will therefore see the change in parent pointer. I have changed the description to explicitly state that the None-parent is protected by 'rootLock'

main/src/library/BPlusTree.flix
CasperDN: See https://github.com/flix/flix/pull/10141#discussion_r2001097399

main/src/library/BPlusTree.flix
```diff
@@ -0,0 +1,763 @@
...
+            Lock.unlock(Node.lock(rootNode));
+            root(tree)
+        }
+
+    ///
+    /// Returns the root of `tree`.
+    ///
+    /// Not thread-safe.
+    ///
+    pub def getRootUnsafe(tree: BPlusTree[k, v, r]): Node[k, v, r] \ r =  tree->root
```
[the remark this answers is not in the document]
JumpTwiice: See https://github.com/flix/flix/pull/10141#discussion_r2001040842

main/src/library/BPlusTree.flix
```diff
@@ -0,0 +1,763 @@
...
+        import java.util.Objects
+        use BPlusTree.Lock
+
+        // Accessor methods
+        /// Provides access to the `children` field of `node`
+        pub def children(node: Node[k, v, r]): Array[Node[k, v, r], r]  = node->children
+        /// Provides access to the `isLeaf` field of `node`
+        pub def isLeaf(node: Node[k, v, r]): Bool  = node->isLeaf
+        /// Provides access to the `keys` field of `node`
+        pub def keys(node: Node[k, v, r]): Array[k, r] = node->keys
```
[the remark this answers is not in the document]
JumpTwiice: See https://github.com/flix/flix/pull/10141#discussion_r2001040842

main/src/library/BPlusTree.flix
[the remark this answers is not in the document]
JonathanStarup: Hmm okay. To avoid this as a public function, you could put it in the test file instead? but maybe magnus thinks differently. could also be @internal if allowed here

main/src/library/BPlusTree.flix
```diff
@@ -0,0 +1,763 @@
...
+        ///
+        /// Internal method to insert `key` and `val` in `node` at `insertionPoint`.
+        /// Increments the `size` of `node`.
+        ///
+        def actualInsertLeaf(
+                key: k,
+                val: v,
+                insertionPoint: Int32,
+                node: Node[k, v, r]): Unit \ r with Order[k] =
+            let size = node->size;
```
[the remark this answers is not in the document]
CasperDN: If we arrive at a point where this is called with a non-leaf value it signals a serious error in the logic of the data-structure. I see no way to recover so we would then throw an error.
We would prefer to avoid an extra branch in every insert. The function will fail anyway, as an attempt to insert into an array of size 0 will make Java throw an exception.

JonathanStarup:
Have you thought about using https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/ReadWriteLock.html ? I think it might make sense here depending on what the engine does. I don't think you use the reentry part of the reentrantlock so it should be somewhat easy to try out once the other code is stable

main/src/library/BPlusTree.flix
JumpTwiice: Yes. The doc comment should be more clear now.

main/src/library/BPlusTree.flix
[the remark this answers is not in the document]
JonathanStarup: In general if you have an expression that uses ; it should be inside {..} as a style choice

main/src/library/BPlusTree.flix
JonathanStarup: Right sorry, that makes sense. I think you should make your own slice-ish helper function then to avoid having empty arrays in more complicated functions.

main/src/library/BPlusTree.flix
JumpTwiice: The alternative is to use a `List` (linked list) but we then lose the ability to binarySearch.

main/src/library/BPlusTree.flix
JonathanStarup: I was just thinking that some inputs would crash the function, but i think its just that order should be >= 3?

main/src/library/BPlusTree.flix
```diff
@@ -0,0 +1,763 @@
...
+                } else
+                    splitLeaf(node, key, val, insertionPoint, order, rc)
+
+        ///
+        /// Attempt to lock the parent of `node`.
+        ///
+        /// Lock invariants:
+        /// - On invocation, the thread has a lock on `node`.
+        /// - On return, the thread holds a lock on `node` and on `parent` given by `node->Some(parent)`
+        def lockParent(node: Node[k, v, r]): Unit \ r = match node->parent {
```
JonathanStarup, line 471: In terms of the "who owns the parent pointer" discussion, this is the place that seems the most tricky in terms of memory model. I think there needs to be an inline comment about why it is okay to look at node->parent without the lock of the parent (which is impossible ofc)

CasperDN:
With respect to the re-entrant part of the locks: As far as we know Java only has re-entrant locks.
We could probably use ReadWriteLocks (they are also re-entrant). The problem is when you write to a leaf you and need to release the read-lock in favour of a write lock. In principle another thread could have split the node, so you would need to be check that the node is still the correct node. A simple check for key<=keys[size] would probably work in most cases. The key might belong between 2 nodes though and we would then need to start traversing right. It could probably be made to work.

main/src/library/BPlusTree.flix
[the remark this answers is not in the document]
JumpTwiice: We are not sure how Flix handles accessing internal data in test files.

main/src/library/BPlusTree.flix
JonathanStarup: sorry, i was too brief. You say that there needs to be a lock on parent, but there is no parent argument, so what nodes do you need locks on?

main/src/library/BPlusTree.flix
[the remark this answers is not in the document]
JonathanStarup: The data inside of enums are public, so you can extract the java lock inside the lock from the test space if you want

main/src/library/BPlusTree.flix
JonathanStarup: A MutList allows both. But my question is just if the linear time is okay (it might be), but this is called quite often right?

CasperDN: I have not met circular arrays before. I will assume it is what is described by: https://en.wikipedia.org/wiki/Circular_buffer.
I do not see how this allows us to write to a particular index such that the data is still sorted when inserts come in random order? We can of course sacrifice the sorted requirement, but would then have to traverse the whole structure each time.
The cost is indeed linear in order and paid each time something is inserted, not just once, but for every split. The implementation should use the System.arraycopy as the backend so the impact is hopefully not that big.

main/src/library/BPlusTree.flix
CasperDN: We have added an inline comment which attempts to argue for why accessing the pointer, as is done, should be okay.

main/src/library/BPlusTree.flix
CasperDN: Ah. That should have been `leftNode`.

main/src/library/BPlusTree.flix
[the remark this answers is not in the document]
CasperDN: That is a leftover from an older version where we wrapped `values` and `children` in `Option`. The match was therefore obligatory at that point.

magnus-madsen:
@JumpTwiice @CasperDN Ping us when we should have another look.

main/src/library/BPlusTree.flix
JonathanStarup: Good point yea

JonathanStarup:
@magnus-madsen what do you think about pub functions that purely exist for tests? a good part of them thread-unsafe. testing is important but its also not nice to pollute the namespace

magnus-madsen:
I think - for now - we need tests.

Later we can think about the exact structure / visibility of those tests.

magnus-madsen:
A public function can still be hidden with `@Internal`. Not ideal of course.

Alternative tests can be placed into the same namespace.

JonathanStarup:
Yeah, I don't know if there exists a perfect solution for "private" tests at the moment. But I think that functions that are purely for tests should be clearly marked.

CasperDN:
> > ... The problem is when you write to a leaf you and need to release the read-lock in favour of a write lock ...
> 
> Ah I didn't think of that - good point. Maybe you could search down with write-locks while you want to insert or do some sort of retry as you said. Do you have an idea of the reading/writing work ratio for the datalog engine?

We would ideally always use read-locks going down as it allows multiple writes to start at the same time. Unless you used fair read-write locks, and depending on the usage of the tree, reads would have an advantage and could potentially starve writes arbitrarily. Another solution for grabbing the write locks in the bottom would be to add a 'time-stamp' to every node and then just check that it has not been changed after acquiring the write lock and if it has retry the locking. You are of course then 'limited' to 2^64 key insertions in a given node. The counter would not even have to be atomic.
In the Datalog implementation we only have reads or writes at any given time for a specific tree. That is part of the purpose of the thread-unsafe read-methods: To allow for the reads to happen without using the locks at all.

CasperDN:
> Yeah, I don't know if there exists a perfect solution for "private" tests at the moment. But I think that functions that are purely for tests should be clearly marked.
> 
> There is also the issue that, even though the test functions are knowingly made thread-unsafe - if its used for concurrent testing, the unsafe access might be stale thus making the test fail.

It depends on whether we are testing concurrently or afterwards. If you join the threads there should be a happens-before edge between the thread-unsafe reads and the previous concurrent reads/writes.
If you test concurrently you should of course always use concurrent methods.

JonathanStarup:
Meeting:
- Make all testing functions private and only make `assertTreeInvariant` public as the only unlawful `@internal` function
- clarify `tree->parent` access as non-happens-before
- make me debug the testing repeat function
- remove the Array.empty calls from the big function

main/src/library/BPlusTree.flix
JumpTwiice: See https://github.com/flix/flix/pull/10141#discussion_r2001313773

magnus-madsen:
Also on the list:

+ Report results of intensive testing
+ Report performance results

Once all that is in, I will take a detailed look at the code :)

CasperDN:
Description of testing:
[Test-report.txt](https://github.com/user-attachments/files/19468037/Test-report.txt)
Description of performance:
[Performance-report.txt](https://github.com/user-attachments/files/19468039/Performance-report.txt)

magnus-madsen:
Quick performance summary:

We performed a number of experiments on the BPlusTree, Java's TreeMap and Flix's RedBlackTree

Most experiments were only done with an order of 10. We expect to retry them with an order of 60 to hopefully get somewhat better performance of the BPlusTree.

All experiments were performed 10 times. For all experiments with time-out we had a 60-second time-out.

We performed the most tests on the BPlusTree to gain an insight in how it performed in different scenarios

The simplest test performed on the BPlusTree was to insert 10_000_000 million uniformly random elements (Int64) and observe the time it took for a varying number of threads. See file ManyInserts.txt

Curiously while 2 threads appear to be faster than 1 and 3 threads faster than 2 this is also more or less the limit. With 4 threads running the speedup was negligible and 6 threads was about as fast as 2 threads. After that the time decreases drastically to be slower than just inserting the elements sequentially.

This trend appears across several experiments: The performance generally increases when using a couple of threads, but at some point the performance decreases to be worse than sequential inserts.


In another experiment we did the following in a loop: Create a tree, have `k` threads insert `n / k` elements each. The loop terminated after 60 seconds and performance was measured in how many elements could be inserted before the time limit was reached. We here observed that the smaller `n` was the closer the peak in performance was to k=2 threads. For n=10_000 inserts the best performance was achieved with k=2 threads and with n=1_000_000 the best performance was observed with k=3 threads.

When performing the above experiment with different order (10, 20, 40, 60) we generally observed that insertions per second was highest for 60 and decreased when decreasing the order. The exception to this was for 20 threads. In this case an order of 40 achieved the best performance. We ascribe this to the fact that the threads will be less likely to attempt to change the same nodes while still being lead down different branches relatively quickly.

Comparison to Java and Flix build-ins: Overall the experiments were performed by repeatedly inserting/reading elements. For the read-experiments the size of the structure was varied (10_000, 100_000 and 1_000_000) and 1_000_000 reads were performed until time-out. For writes the structure was generated and had either 10_000, 100_000 or 1_000_000 inserts in a loop until time-out. All experiments used only 1 thread.

Comparison to Java's TreeMap: Java's TreeMap. Java's TreeMap was both faster for inserts and reads. For inserts the difference was around a factor 4-2, decreasing with the number of inserted elements. For reads Java's TreeMap performed around 2-1.5 more reads in the time frame.

Comparison to Flix's RedBlackTree: For reads: The difference was about the same as with Java, though only around 1.9 to 1.45. For writes the BPlusTree performed around 1.X-2 times more inserts before timeout, increasing with the number of inserted elements.

magnus-madsen:
Test-report:
We have tested the implementation with a mix of small unit-tests and large scale concurrency tests.
The goal of the concurrency tests was to test that after elements were inserted they could be found in the data-structure. Though we have implemented a map, for simplicity we only test that the keys are inserted, not the values. The reasoning being that a key once inserted will stay, but a value might change based on insertion order.

The large scale concurrency tests came in varied in 3 ways.
	How many threads are used
	How the numbers to be inserted were generated
	When membership testing is performed

For inserted elements we had 3 variants:
	Completely random numbers. The purpose is here to generate a lot of splits.
	Random numbers in a small range: The purpose is here to make the threads compete for locks.
	The numbers 1...n in order: The purpose is here to maximize competition between threads

For membership testing we had 2 variants
	After all threads finished insertion verify that all elements they should have inserted is in the tree
	After `k` steps verify item 1 is still in the tree, after `k` more steps verify item 2 is in the tree and so on

The total number of inserted elements (Int64) was always 100_000_000 split between the threads.
The random ranges were either [0..10_000] or [0..100_000]
The number of threads ranged from 10 to 40.

magnus-madsen:
Performance:

- For how long did the experiments run? You say that you run them 10 times. In one experiment you do 10_000_000 insertions. Is this enough to warm the JIT? Whats the absolute running time? Is it more than 5minutes?
- I agree that a BTree will not shine at order 10. I would try 16, 32, and 64.
- Can you confirm that the random numbers were generated _before any inserts_?
- It would be useful to have some graphs...

> Comparison to Java's TreeMap: Java's TreeMap. Java's TreeMap was both faster for inserts and reads. For inserts the difference was around a factor 4-2, decreasing with the number of inserted elements. 

To me, this is unexpected. How can Java's TreeMap be faster when you benefit from both monomorphization and BTree structure? 

- Did you use JProfiler to inspect where the bottle necks are?

CasperDN:
Timewise: The 10 million inserts were executed directly after other experiments, so the JVM should be warm. It is a good point though as the running time of a single experiment was only 5-12 seconds. We will increase the number of inserts.

Technically we cannot guarantee that the random numbers were generated before the experiments, as we did not use them to anything before that point. The repetitions were on the same random numbers each time, and they were only generated once. Since we saw no pattern in the difference in performance  between different repetitions we would claim that the random numbers were indeed created before the experiment started measuring time.

We are still not sure why traverseDown is so expensive when single-threaded. We will investigate it more, but it appears to be tail-call optimized correctly.

JonathanStarup:
To my count, each call to traverseDown makes 3 objects which are not possible to remove without substantial code gen effort

main/src/library/BPlusTree.flix
```diff
@@ -1,282 +1,488 @@
...
-/// is either an internal node or a leaf node. In contrast to binary
-/// search trees, a B+ Tree of order `m` allows for nodes with at most
-/// `m` children and `m-1` keys. A B+ Tree maintains the sorted order
-/// of its keys by storing only keys in internal nodes and additionally
-/// values in leaf nodes. This implementation varies from the standard
-/// definition by also linking internal nodes. This is due to a use-case
-/// within Flix.
+/// A B+ Tree is an m-ary self-balancing search tree where each node is either an internal
+/// node or a leaf node. In contrast to binary search trees, a B+ Tree of order `m` only
+/// allows nodes with at most `m` children and `m-1` keys. A B+ Tree maintains the sorted
```
JonathanStarup, line 23: here you call it `m` and later it is called `order`, I think it is best to either mention that they are the same or use the same name

main/src/library/BPlusTree.flix
```diff
@@ -1,282 +1,488 @@
...
+    /// performed by reading the values and only afterwards verifying that no other
+    /// thread has changed the values. Only annotating a region effect to StampedLock
+    /// is safe, since it can only non-deterministically block a thread for some (
+    /// potentially infinite) amount of time. This is allowed under non-effectful behavior.
+    ///
     ///
+    /// One possible danger, introduced by StampedLock, is handled by a null-check in binarySearch.
+    /// Consider the following example where everything is initialized to 0:
+    /// Thread 1 executes
+    /// {a = 2; b = 3}
```
JonathanStarup, line 77: In this example, what does and b represent? what state in the BTree might be null?

main/src/library/BPlusTree.flix
```diff
@@ -1,282 +1,488 @@
...
 
+
+        /// Attmept to upgrade `stamp` on `lock` to a write-lock on `lock`.
+        /// The returned stamp cannot be invalidated if it was valid when issued.
         ///
-        /// Acquires `lock` only if it is not currently held by another thread.
+        /// This operation is non-blocking.
         ///
-        pub def tryLock(lock: Lock[r]): Bool \ r =
+        pub def tryConvertToWrite(stamp: Int64, lock: Lock[r]): Int64 \ r =
```
JonathanStarup, line 145: For readability, it would be nice if you wrapped the `Int64` values in an `enum Stamp(Int64)` so they're not confused with other numbers.

JonathanStarup: Or maybe `enum WriteStamp(Int64)` and `enum ReadStamp(Int64)`? Then the doc comments also have less to explain

main/src/library/BPlusTree.flix
```diff
@@ -1,282 +1,488 @@
...
+        /// For a better explanation see Netflix's encounter of a similar problem:
+        /// https://netflixtechblog.com/java-21-virtual-threads-dude-wheres-my-lock-3052540e231d?gi=128dba0ad426
+        pub def dumbWait(lock: Lock[r]): Unit \ r =
+            let Lock(l) = lock;
+            let stamp = unchecked_cast((l.readLock(): _ \ IO) as _ \ r);
+            unchecked_cast((l.unlockRead(stamp): _ \ IO) as _ \ r)
+
+        ///
+        /// Returns `true` if `lock` is currently held by a thread.
+        /// Only supposed to be used for testing.
```
JonathanStarup, line 126: newline before this line

main/src/library/BPlusTree.flix
```diff
@@ -1,282 +1,488 @@
...
+        /// Virtual threads can show weird behaviour when using all physical threads
+        /// and using locking. Without this the threads using optimistic reading will never be
+        /// removed from the physical thread. Meanwhile a thread that holds the lock will never
+        /// get to run and we have a deadlock. Progress can happen, but it appears to be
+        /// extremely unlikely. To fix this call a blocking operation which makes the JVM remove
+        /// the virtual threads from the physical threads allowing the lock-holder to unlock.
+        ///
+        /// For a better explanation see Netflix's encounter of a similar problem:
+        /// https://netflixtechblog.com/java-21-virtual-threads-dude-wheres-my-lock-3052540e231d?gi=128dba0ad426
+        pub def dumbWait(lock: Lock[r]): Unit \ r =
```
JonathanStarup, line 119: I understand that this is hack-ish by necessity, but maybe it should be called `yieldBasedOn` or that describes what it is used for

main/src/library/BPlusTree.flix
```diff
@@ -1,282 +1,488 @@
...
     ///
-    pub def containsPairUnsafe(key: k, val: v, tree: BPlusTree[k, v, r]): Bool \ r with Order[k], Eq[v] =
-        let leaf = Node.findLeafUnsafe(key, tree->root);
-        Node.leafContainsPair(key, val, leaf)
+    /// Returns a fresh empty BPlusTree tree of `order` (the maximum keys/values per node).
+    /// The order of the tree must be at least 3.
+    ///
+    @Internal
+    pub def emptyWithOrder(rc: Region[r], order: Int32): BPlusTree[k, v, r] \ r =
+        emptyWithOrderAndSearch(rc, order, Vector#{})
```
JonathanStarup, line 229: We've discussed it before, but i think you should round `order` up to 3 just to be safe

JonathanStarup: Flix users can also use this tree right? I'd think that only assertTreeInvariant would be internal

main/src/library/BPlusTree.flix
```diff
@@ -1,282 +1,488 @@
...
         }
 
     ///
-    /// Returns the root of `tree`.
+    /// Returns `v` if `(key, v)` is in `tree`. Otherwise return `d`.
     ///
-    /// Not thread-safe.
+    pub def getWithDefault(key: k, d: v, tree: BPlusTree[k, v, r]): v \ r with Order[k] =
+        let (leaf, stamp) = BPlusTree.findLeaf(key, tree);
+        let res = Node.getWithDefault(key, stamp, d, leaf, tree);
```
JonathanStarup, line 304: just remove the `let res`

main/src/library/BPlusTree.flix
```diff
@@ -1,282 +1,488 @@
...
-    ///         `children` is of size `order` and `keys` have size `order - 1`
-    ///         `keys[i]` store the smallest value in the subtree store in `children[i+1]`
-    ///         `size` is the number of children currently stored in the leaf
+        "Search: ${Vector.join(" < ", tree->search)}\nTree:\n${Node.toString(0, tree->rc, tree->root)}"
+
+    ///
+    /// Write lock `tree->rootLock` and return the stamp.
+    ///
+    def writeLockRootPointer(tree: BPlusTree[k, v, r]): Int64 \ r =
+        let res = Lock.writeLock(tree->rootLock);
```
JonathanStarup, line 444: remove the `let res`

main/src/library/BPlusTree.flix
```diff
@@ -1,282 +1,488 @@
...
+/// `root` has a separate lock concerning the contents of the node. The lock of `root`
+/// must first be acquired after successfully acquiring `rootLock`.
 ///
-/// `rootLock` is a lock on the `root` node which a thread must first
-/// acquire in order to insert into the tree. As the thread descends into
-/// the tree, `rootLock` is released to allow for other threads to insert.
+/// `rc` is the region for the tree.
 ///
-/// `rc` is the region for the tree which is required for a mutable object.
+/// `search` is a result of the primary usage being for vectors. A `Search` describes the
```
JonathanStarup, line 47: I am still a bit confused about `search`, can you expand on this paragraph? is search used only when `v` is `Vector[_]`?

main/src/library/BPlusTree.flix
JumpTwiice: Alternatively, we could rename `order` to `arity` since `order` usually refers to a way to compare things in Flix. This avoids some ambiguity when also using `emptyWithOrder`.

main/src/library/BPlusTree.flix
JumpTwiice: Unless the enum gets optimized away, this incurs overhead as each read now requires a pattern-match / pointer access.

main/src/library/BPlusTree.flix
JonathanStarup: good point yeah. maybe nodeWith?

main/src/library/BPlusTree.flix
JonathanStarup: true, ill let that be up to you and magnus, but signatures will be prettier with the wrapper types.

@magnus-madsen here is why the enum optimizer matters

main/src/library/BPlusTree.flix
JumpTwiice: `{a = 2; b = 3}` is a Java code block  where `a` and `b` are variables. The entries in the `keys` array can be null.

main/src/library/BPlusTree.flix
JumpTwiice: The internal annotation is a mistake. Regarding rounding `order `up to 3: it would be nicer to explicitly throw an error if `order < 3` since otherwise the user would operate on the basis that they have a tree with order 2 (or 1), even though it is actually 3.

main/src/library/BPlusTree.flix
JumpTwiice: `search` is only used internally for indexes in the Datalog engine and only makes sense if `v` is a `Vector`. It decides in which order to compare the elements in `v`. For normal usage, we pass an empty vector.

main/src/library/BPlusTree.flix
JumpTwiice: I think `arity` fits better since it's common terminology regarding trees.

main/src/library/BPlusTree.flix
JumpTwiice: For now we just round up

main/src/library/BPlusTree.flix
```diff
@@ -1,282 +1,488 @@
...
+    /// Another thread reading may then see the increase of size and try to access that position getting null.
+    /// This has not happened in our experiments, but we wish to handle the case. The risk is in binarySearch and we
+    /// therefore check whether any value read is null and if so 'aborts'.
+    ///
+    /// For Java's memory model see: https://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html#jls-17.4
+    ///
+
     @Internal
-    pub enum Lock[_: Region](JReentrantLock)
+    pub enum Lock[_: Region](JStampedLock)
```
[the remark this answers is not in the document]
JumpTwiice: :D

magnus-madsen:
Ping me when it is my turn to take a look.

JumpTwiice:
@magnus-madsen Ready

magnus-madsen:
Merged 🚀 !

PR #10142, opened 2025-03-14
refactor: unwrap OpCompleter

stormckey:
Besides the normal refactor, also something specific to OpCompleter:
- If isHandler, we can get the fully qualified name of the effect, as a result, we can just lookup the map instead of iterate through it.
- Wee've seen similar code in a lot of places, where we define a sym and use it to lookup the table, should we add a helper function somewhere?
- Although unwrapped, I still keep two entry point for OpCompleter, since they are not completely doing the same thing. 
- Should we add another completer? if isHandler, the completion is a snippet defining the handler, if not isHandler, it's a function call.


magnus-madsen:
I always thought this should be two handlers.

magnus-madsen (approved):

magnus-madsen:
Will you create a ticket to have two handlers? OpCompleter and HandlerCompleter?

Issue #10135, opened 2025-03-14
Improve argument completions


magnus-madsen:
Based on #10133, we want to ensure that only complete arguments if there are none already:

- `Perso -> Person(?fstName, ?lstName)` whereas
- `Pers("M", "M") -> Person("M", "M")`

This means that `EnumTagCompleter` needs to know the number of arguments already present.

Moreover, while using the visitor we need to handle `Expr.Tag` and `Expr.Error` separately.

stormckey:
I try to take a look into this ast. And it turns out that applying a undefine name to a list of args will produce a tree that looks like:
```
ApplyClo(ApplyClo(UndefineName, arg1), arg2)
```
An instant idea is that we can use the same stack-visitor to visit the tree. When the head is an UndefinedName, we count the number of following ApplyClo to get the number of args already applied.

At first, we can just simply say if there is any arg, I will not provide a snippet but just complete the name. But in the future we can check the existing args and figuring out which holes they fit in, and provide a snippet for the missing holes.

I am just a little bit worried about the robustness of the first visiting method. Maybe during the counting we can also check that these ApplyClo are also applying to the same undefinedName to make sure we are counting correctly?

Here is a screenshot for you to have an glimpse into the stack after visiting:

![Image](https://github.com/user-attachments/assets/194934c1-f909-448c-b9de-8dfc3481dfbb)

magnus-madsen:
@JonathanStarup Can tell you how to view the AST

JonathanStarup:
Before going into technical stuff, what should happen for partially applied functions? Shouldn't we just start by disabling the argument insertion for `Pers(..)` no matter what is inside the arguments?

magnus-madsen:
We strip a number of arguments equal to number of applies.

JonathanStarup:
so for `f(x)(y)` you reduce it by two and insert them.. where?

magnus-madsen:
> So for `f(x)(y)` you reduce it by two and insert them.. where? Could be `f(x)(y, ?z)` I guess. I don't see the point in trying to modify arguments that are already partially filled out. What's the probability even of doing something helpful?

I see, so your suggestion is to just not fill any arguments?

stormckey:
> So for `f(x)(y)` you reduce it by two and insert them.. where? Could be `f(x)(y, ?z)` I guess. I don't see the point in trying to modify arguments that are already partially filled out. What's the probability even of doing something helpful?

[...]

PR #10147, opened 2025-03-17
feat: introduce UndefinedNameContext

stormckey:
This context can tell use:

- if there is a pipeline applied to this undefinedName:
  ```
  1 |> fo
  (1, 2) ||> fo
  ```
- How many arguments are directly applied to this undefinedName:
  ```
  fo() \\ 1
  fo(1) \\ 1
  fo(1, 2) \\2
  ```

I use a DFA to visit the stack to build the context.

Now this context is only used in DefCompletion, the result is:
- Completion will only complete the name if it's followed by parentheses:
  image
  image
- Completion is aware of the pipeline:
  image
image
image
image

The Context is now named UndefinedNameContext since the state and DFA is designed for undefinedName case. In cases like a match rule mentioned in #10135, the DFA should transfer over MatchRule, UndefinedTag etc. We can extend it in the future and rename it to something like ApplicationContext.


magnus-madsen (commented):
I am finally getting around to this! I have some high level comments:

- [ ] Get rid of changes to Weeder2 - I don't think they are relevant? Or they can be separate PR.
- [ ] Changes to HoleCompleter can be separate PR?

I think I understand `UndefinedNameContextState` but I would like something that is simpler and also more reusable:

I forgot how the stack is ordered. I will assume the innermost element is first. (If not, then what I propose has to be "reversed").

Given a stack S, we drop the first element (this is typically Expr.Error).

We then need a data type like:

```
case class UndefinedNameContext(applied: Option[Int], pipelined: Option[Int])
```

but I propose just:

```
case class ExprContext(applied: Int, pipelined: Boolean)
```

(Maybe this is not enough-- then you can explain).

I think computing this data type is one of those rare cases where an imperative loop is easier. Just have to local mutable variables and loop through the stack resetting the variables as necessary.

stormckey:
> * [ ]  Get rid of changes to Weeder2 - I don't think they are relevant? Or they can be separate PR.

I think it's actually relative, sorry that I did not explain this part.

The relative change is here:
image
I change the location of this error to be zero-width at the end of the expression.

Otherwise, when visiting, this malformed will be the first element, while we are expecting an UndefinedName. That's why we need to change this.

The change to the whole filed looks quite large, most of which comes from the warnings of our new IDEA setting. I solve all of them.

> * [ ]  Changes to HoleCompleter can be separate PR?

It's just about extracting the code to build the stack. So we can just build the stack once and pass it to difference completers for their own usage. But you are right, I will leave it to another PR, after this.

> I think I understand `UndefinedNameContextState` but I would like something that is simpler and also more reusable:
> 
> I forgot how the stack is ordered. I will assume the innermost element is first. (If not, then what I propose has to be "reversed").
> 
> Given a stack S, we drop the first element (this is typically Expr.Error).
> 
> We then need a data type like:
> 
> ```
> case class UndefinedNameContext(applied: Option[Int], pipelined: Option[Int])
> ```
> 
> but I propose just:
> 
> ```
> case class ExprContext(applied: Int, pipelined: Boolean)
> ```
> 
> (Maybe this is not enough-- then you can explain).
> 
> I think computing this data type is one of those rare cases where an imperative loop is easier. Just have to local mutable variables and loop through the stack resetting the variables as necessary.

Okay. I will try to impl it imperatively.

stormckey:
Now we use imperative style code to parse the context.

magnus-madsen:
Let us do the Weeder2 changes in one or more separate PRs.

I think we need to discuss the context stuff on the whiteboard.

magnus-madsen:
But let us start by splitting this PR and merging some of it.

stormckey:
I've created the PR.

After that one is merged, I will undo and rebase this one

magnus-madsen:
See #10329

PR #10148, opened 2025-03-17
feat: EnumTagCompleter now supports snippet

stormckey:
Preview:
image
image

But there is a problem in displaying the type parameter:
image
image
Ideally they should be replace by a or String. But how could we achieve that?


mlutze:
Ideally they should be replace by a or String. But how could we achieve that?

Just calling `toString` on a type will result in these not-so-nice names. Instead, you can try `FormatType.formatType` to get more user-friendly printing.

Issue #10085, opened 2025-03-10
Stop using env for LocalScope name


stormckey:
EDIT by Magnus: rename `env: LocalScope` to e.g. `sco: LocalScope` and rename `env` elsewhere.

stormckey:
@magnus-madsen

magnus-madsen:
This can probably be done in one or two PRs

PR #10158, opened 2025-03-18
refactor: rename env to scp in resolver

stormckey:
according to #10085

I choose the replace env with scp.

In the arg, to avoid shadowing, I use scp0 consistently.

taenv is left as it is, since it's not of type LocalScope.

Many shadow problem is also solved since we are using the new warning config. To solve them, I rename names in arg with a trailing 0, names inside with a trailing 1, 2 etc.

Somewhere 1, 2 is used for the first and second exp. I use e1_, e2_ to avoid shadowing there.

Resolver is the first place where we use LocalScope a lot. Another place is in the completers, that would be another PR.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
@@ -580,9 +580,9 @@ object Resolver {
   /**
     * Performs name resolution on the given case `caze0` in the given namespace `ns0`.
     */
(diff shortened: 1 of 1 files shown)


magnus-madsen:
Conflicted

magnus-madsen:
@mlutze review and merge

mlutze (approved):
LGTM, just need to address conflicts

stormckey:
It seems that after the new IDEA setting forces shadowing as warnings, I solved the problem in Resolver in this PR while someone else did it prior to me, which result in a lot of conflicts :)

magnus-madsen (approved):

PR #10159, opened 2025-03-18
feat: detect missing `;`

### LINKED ISSUE #10143: Resilient parsing of incomplete expression statements
```
def main(): Unit \ {Net, IO} =
    run {
        let url = "http://example.com/";
        Log                                                                        // <----
        Logger.info("Downloading URL: '${url}'");
        match HttpWithResult.get(url, Map.empty()) {
            case Result.Ok(response) =>
                let file = "data.txt";
                Logger.info("Saving response to file: '${String.toUpperCase(file)}'");
                let body = Http.Response.body(response);
                match FileWriteWithResult.write(str = body, file) {
                    case Result.Ok(_) =>
                        Logger.info("Response saved to file: '${file}'")
                    case Result.Err(err) =>
                        Logger.fatal("Unable to write file: '${err}'")
                }
            case Result.Err(err) =>
                Logger.fatal("Unable to download URL: '${err}'")
        }
    } with FileWriteWithResult.runWithIO
      with HttpWithResult.runWithIO
      with Logger.runWithIO
```

Is there any way we can parse this program? Right now pretty much all information is lost.

JonathanStarup:
Fixes #10143

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/errors/ParseError.scala
+++ b/main/src/ca/uwaterloo/flix/language/errors/ParseError.scala
@@ -212,6 +212,27 @@ object ParseError {
   }
 
   /**
+    * An error raised to indicate that a statement expression is missing a semi colon.
+    *
+    * @param sctx  The syntactic context.
+    * @param loc   The source location.
+    */
+  case class MissingSemi(sctx: SyntacticContext, loc: SourceLocation) extends ParseError {
+    override val kind: CompilationMessageKind.ParseError = CompilationMessageKind.ParseError(sctx)
+
+    def summary: String = s"Runaway expression, possibly missing ';'."
+
+    def message(formatter: Formatter): String = {
+      import formatter.*
+      s""">> Runway expression, possibly missing ';'.
+         |
+         |${code(loc, s"Here")}
+         |
+         |""".stripMargin
+    }
+  }
+
+  /**
     * An error raised to indicate that no items were found in a context where one or more is needed.
     *
     * @param expected Names of the items that are expected at least one of. See [[TokenKind.display]].
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -1366,11 +1366,22 @@ object Parser2 {
       * expression.
       *
       * @param rhsIsOptional If false, the right-hand-side expression is not treated as optional.
+      * @param continueUntil statement parsing will keep looping until an expression is followed by
+      *                      this. E.g. if '}' then `42; 43 44}` will be parsed as a statement of
+      *                      three expressions with an error for the missing `;` after `43`.
       */
-    def statement(rhsIsOptional: Boolean = true)(implicit s: State): Mark.Closed = {
+    def statement(rhsIsOptional: Boolean = true, continueUntil: Option[TokenKind] = None)(implicit s: State): Mark.Closed = {
       var lhs = expression()
-      if (eat(TokenKind.Semi)) {
-        statement()
+      val t = nth(0)
+      if (t == TokenKind.Semi) {
+        advance()
+        statement(continueUntil = continueUntil)
+        lhs = close(openBefore(lhs), TreeKind.Expr.Statement)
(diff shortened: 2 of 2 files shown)


JonathanStarup:
```
def main(): Unit \ {Net, IO} =
    run {
        let url = "http://example.com/";
        Log                                                                        // <----
        Logger.info("Downloading URL: '${url}'");
        match HttpWithResult.get(url, Map.empty()) {
            case Result.Ok(response) =>
                let file = "data.txt";
                Logger.info("Saving response to file: '${String.toUpperCase(file)}'");
                let body = Http.Response.body(response);
                match FileWriteWithResult.write(str = body, file) {
                    case Result.Ok(_) =>
                        Logger.info("Response saved to file: '${file}'")
                    case Result.Err(err) =>
                        Logger.fatal("Unable to write file: '${err}'")
                }
            case Result.Err(err) =>
                Logger.fatal("Unable to download URL: '${err}'")
        }
    } with FileWriteWithResult.runWithIO
      with HttpWithResult.runWithIO
      with Logger.runWithIO
```

gives errors

```
-- Resolution Error --------------------------------------------------

>> Undefined name 'Log'.

4 |         Log                                                                        // <----
            ^^^
            name not found

-- Parse Error (OtherExpr) --------------------------------------------------

>> Runway expression, possibly missing ';'.

4 |         Log                                                                        // <----
                                                                                       ^^^^^^^^
                                                                                       Here

-- Resolution Error --------------------------------------------------

>> Undefined name 'url'.

5 |         Logger.info("Downloading URL: '${url}'");
                                             ^^^
                                             name not found

-- Type Error --------------------------------------------------

>> No instance of the 'ToString' class for the type 't0'.

5 |         Logger.info("Downloading URL: '${url}'");
                                             ^^^
                                             missing instance


-- Resolution Error --------------------------------------------------

>> Undefined name 'url'.

6 |         match HttpWithResult.get(url, Map.empty()) {
                                     ^^^
                                     name not found
```

JonathanStarup:
Im not sure why the undefined name errors occur atm

JonathanStarup:
oh @magnus-madsen `{ let x = 42; 43; 44 }` is a singleton block with a 3-ary let thing, so `{let x = 42; 43 44}` is parsed as `{(let x = 42; 43); 44}` which is why the undefined names pop up

If we need the correct nesting, the let expression needs to know that it is in a context where } is expected, differentiating `def f(): Unit = let x = (); x` and `def f(): Unit = {let x = (); x}`

magnus-madsen:
> oh @magnus-madsen `{ let x = 42; 43; 44 }` is a singleton block with a 3-ary let thing, so `{let x = 42; 43 44}` is parsed as `{(let x = 42; 43); 44}` which is why the undefined names pop up

Ah- thats a problem

> If we need the correct nesting, the let expression needs to know that it is in a context where } is expected, differentiating `def f(): Unit = let x = (); x` and `def f(): Unit = {let x = (); x}`

What does this mean? Sorry I don't understand. How big is the change?

magnus-madsen:
Note that the current code overflows in the fuzz suite.

JonathanStarup:
> What does this mean? Sorry I don't understand. How big is the change?

An extra implicit on every parse function i think ("am in directly inside a block"). We could also finally enforce `{..}` around statement/letbinders

> Note that the current code overflows in the fuzz suite.

I think its very hard to avoid with this strategy, but maybe `statement` could be made tail-recursive

magnus-madsen:
> I think its very hard to avoid with this strategy, but maybe `statement` could be made tail-recursive

It is StackOverflow... or infinite loop...

magnus-madsen:
Conceptually, assume we wanted to allow linebreak to act as semicolon. How would we implement it? And can we not use the same special case here?

JonathanStarup:
> > I think its very hard to avoid with this strategy, but maybe `statement` could be made tail-recursive
> 
> It is StackOverflow... or infinite loop...

no not infinite. it consumes input each loop

JonathanStarup:
> Conceptually, assume we wanted to allow linebreak to act as semicolon. How would we implement it? And can we not use the same special case here?

we would have to require `{}` to not be ambigous

```
def f(y: Int32): Unit IO =
let x = y * 2
println(x)
def g(): Unit \ IO =
println("hello")
println("you")
```

Issue #10138, opened 2025-03-14
AutoComplete Invariants


magnus-madsen:
TBD:

- [ ] Completion on keyword => empty list of suggestions

TODO

stormckey:
> What's next? What easy, maybe that we don't complete ontop of a name, e.g.
> 
> ```
> def foo|
> 
> enum bar|
> ```
> 
> should not offer suggestions

Sounds good, but how should we find out these nodes? IMO they should just be Ident. And we definitely can not say we should not give suggestions after any Ident 🤔

stormckey:
Now I've implemented (or PR) most invariants with empty completions.

For the following invariants, most of them require a place where there are completions.

How can we locate such places? Or we just iterate through all possible places and preserve those do return some completions?

magnus-madsen:
> Sounds good, but how should we find out these nodes? IMO they should just be Ident. And we definitely can not say we should not give suggestions after any Ident 🤔

Here instead of using tokens, we use the typed ast root

stormckey:
The visitor will now skip Expr.Var, so I think we should first create a PR to add that, then we can check invariants over these nodes?

magnus-madsen:
Anyway there many invariants- also tags, ops, sigs etc. etc.

stormckey:
The code in Visitor:
```scala
  private def visitExpr(expr: Expr)(implicit a: Acceptor, c: Consumer): Unit = {
    if (!a.accept(expr.loc)) { return }

    c.consumeExpr(expr)

    expr match {
      case Expr.Cst(_, _, _) => ()
      case Expr.Var(_, _, _) => ()
```

Expr.Var is skipped. So we can not access all the Expr.Var through visitor

magnus-madsen:
Is there not some visitor var things?

Or you can implement consumeExpr

magnus-madsen:
I will create individual tickets from now on.

PR #10160, opened 2025-03-18
feat: introduce TestCompletionProvider

### LINKED ISSUE #10136: Add `TestCompletionProvider`
We want to build some infrastructure for automatic testing. 

I would propose functions like:

- A function that returns the list of source programs want to test on (e.g. could just start with one element, `examples/effects-and-handlers/running...flix`)
- A function to compile a Flix program into a `TypedAst.Root`
- A function to extract all tokens of specific kinds, e.g. keyword tokens.
- A function to take a program and slightly modify it (for example replacing a keyword token "def" by just "de") 
- A function to trigger autocomplete at given source locations and compare them to expected results.

For example, we discussed that triggering auto-complete on keywords should not offer suggestions.

stormckey:
fixes #10136

first invariant for #10138

Here we introduce TestCompletionProvider to check invariants over Completion Provider.

This file could be placed somewhere else if there is a convention.

The first invariant is that no completion should be provided after a complete keyword.

Questions:
- Should we hard code the programs to be tests or we can read them from example files?
- Should I place the helper functions in an other file?
- Any suggestions about the api?


magnus-madsen:
>     * Should we hard code the programs to be tests or we can read them from example files?

That is fine for now. I would suggest to add a few more examples.

Later we will restructure the code to use _all_ our examples (in the example folder).

>     * Should I place the helper functions in an other file?

I think it is good to keep everything in one file for now.

>     * Any suggestions about the api?

I am adding review comments.

magnus-madsen:
What we are trying to do here is called "property-based testing" or "quick checking"- but we are not reusing any library or framework, instead we are building our own DSL.

magnus-madsen:
How long does it take to run this test suite by the way?

main/test/ca/uwaterloo/flix/api/lsp/TestCompletionProvider.scala
```diff
@@ -0,0 +1,120 @@
...
+  /**
+    * Checks the completion provider invariant for the given token filter, string transformer and completion assertion.
+    * For all the programs to be tested, we use the filter to select tokens to transform, apply the transformer to the selected tokens,
+    * call the completion provider with the transformed program and check the completions against the assertion.
+    *
+    * @param tokenFilter          the filter to select tokens to transform.
+    * @param stringTransformer    the transformer to apply to the selected tokens.
+    * @param completionAssertion  the assertion to check the completions against.
+    */
+  private def checkInvariant(tokenFilter: Token => Boolean, stringTransformer: String => String, completionAssertion: CompletionList => Unit ): Unit = {
```
magnus-madsen, line 85: This is like a "god function". In OO one would say that it has too much responsibility.

magnus-madsen: We need to break it down into smaller reusable parts.

main/test/ca/uwaterloo/flix/api/lsp/TestCompletionProvider.scala
```diff
@@ -0,0 +1,120 @@
...
+  private def assertEmptyCompletions(completions: CompletionList): Unit = {
+    assert(completions.items.isEmpty)
+  }
+
+  /**
+    * A function that returns the input string as is.
+    */
+  private def identity(s: String): String = s
+
+  test("No completions after complete keyword"){
```
magnus-madsen, line 113: If we look at the test, what we would like to say is something like

\forall C[keyword], pos \in KeywordPos(C[keyword]). autoComplete(pos) = Nil

So I think we want building like:

- getKeywordTokens
- getPositionsInToken

but we don't need `assertEmptyCompletions`. That is just == Nil.

magnus-madsen: There is also no need to recompile the program or anything.

magnus-madsen:
To help you towards a modular design, I think you should consider two properties:

- Completion on keyword => empty
- Completion on literal => empty

stormckey:
> To help you towards a modular design, I think you should consider two properties:
> 
> * Completion on keyword => empty
> * Completion on literal => empty

We dont actually need any other helper function (except isLiteral from TokenKind) to add the invariant.

stormckey:
> How long does it take to run this test suite by the way?

After adding two other examples and another invariant, now it takes 24s

magnus-madsen:
> > How long does it take to run this test suite by the way?
> 
> After adding two other examples and another invariant, now it takes 24s on my computer

OK; and its only going to grow. So for now, we have to leave it out of the default test suite.

main/test/ca/uwaterloo/flix/api/lsp/LspSuite.scala
```diff
@@ -18,5 +18,6 @@ package ca.uwaterloo.flix.api.lsp
 import org.scalatest.Suites
 
 class LspSuite extends Suites(
-  new VisitorSuite
+  new VisitorSuite,
+  new TestCompletionProvider
```
magnus-madsen, line 22: That is, leave it out here.

magnus-madsen:
In a separate PR, we should add something like: https://github.com/flix/flix/blob/master/.github/workflows/fuzzing.yaml

stormckey:
> In a separate PR, we should add something like: https://github.com/flix/flix/blob/master/.github/workflows/fuzzing.yaml for this specific test class.

OK, in commit d5eff8f, I tried the workflow and it works fine. But now it's removed. I'll add it back in another PR.

It's also removed from the test suit.

main/test/ca/uwaterloo/flix/api/lsp/TestCompletionProvider.scala
magnus-madsen: Will you break this down now or later?

magnus-madsen: Don't add more higher order arguments. The opposite. You want to have small functions that each test case compose.

Issue #9773, opened 2025-01-27
Final cleanup for LocalScope Completer


stormckey:
Now we've arrived at the stage to perform the final cleanup for LocalScope Completer.

Ideally, LocalScopeCompleter is responsible for providing completions for constructs in the localscope. Constructs can appear in the top level of LocalScope as it's naturally included (like var) or it's imported/used.

I've move most constructs out of LocalScopeCompleter.

Here are the things left:
- Can not be used, and can not appear on the top level of LocalScope:
  - StructField
- Can not be used, but can appear on the top level of LocalScope:
  - Var
  - LocalDef
- Can be imported, and can appear on the top level of LocalScope:
  - JavaClass
- Can be used(but should we?), and thus appear on the top level of LocalScope:
  - AssocTypeSig (Trait)
  - AssocTypeDef (Instance)

I think:
- We could keep LocalScopeCompleter for Var, LocalDef and JavaClass
  - Or we can merge JavaClass with AutoImportCompleter, like what we do with auto use and local scope
- We could remove StructField as it has nothing to do with LocalScope
- I'm not sure if we should just remove two Assocs or write a standalone completer for them?

stormckey:
@magnus-madsen

magnus-madsen:
> We could keep LocalScopeCompleter for Var, LocalDef and JavaClass 

Agree.

> We could remove StructField as it has nothing to do with LocalScope

Where would you move it? Its kind of a strange feature because it follows the scope of the companion module.

Not sure about the AssocType things. I think they are local , so I would probably keep them here too?

stormckey:
> What's the goal here? I think most of the things on your list belong in LocalScopeCompleter-- but of course its not entirely clear.

Well LocalScopeCompleter will look up a name in the local scope, if found one, it will suggest to complete that.

For things like def, enum ... This work is done by the dedicated completers. So here we are discussing all the things left.


For struct field, it will only appear in side Struct, instead on the top level map of LocalScope:

<img width="893" alt="Image" src="https://github.com/user-attachments/assets/d2b2a467-49d6-4926-ae43-c425043c50f7" />

Since the code is only searching on the top level, it will never meet struct field, so it's safe to remove it.

For assoc, you can use it, and then you can see it on the top level map:

<img width="884" alt="Image" src="https://github.com/user-attachments/assets/8e77e9da-579c-4df7-b76e-79655f1224b7" />

[...]

PR #10169, opened 2025-03-19
refactor: unwrap LocalScopeCompleter

stormckey:
I also remove the support of localscope completion for structfield, assoctype, assoceff.

Relative discussion in #9773. But it's been hanging there for quite a while. 

I think it's safe to remove these either impossible or uncommon cases.

I preserve two entry points: Expr one and Type one


magnus-madsen:
Conflicted

stormckey:
Solved

PR #10177, opened 2025-03-20
feat: support signature help

### LINKED ISSUE #10166: SignatureHelp
Explore if we can support SignatureHelp from LSP spec

stormckey:
fixes #10166

A preliminary version:

image
image
image

As you can see, there are several errors:
- I can not find doc for parameter, do we support that?
- A bug in locating the active parameter: the compiler seems to always fill in all the args in ApplyDef, for the missing ones, we can tell from its `isSynthetic`. So I use the number of non-synthetic args as the actual number of applied args. But when the list is empty, there is a default Unit and it's not Synthetic, I can not see where this Unit is produced. Maybe @JonathanStarup can help? And can we make it synthetic?

Now it does not support VSCode, since we have to change the vscode client first to enable this feature.


stormckey:
It seems that I checked out this branch from another feat branch instead of from master :( Hopefully I've undone most of the changes from there.

main/src/ca/uwaterloo/flix/api/lsp/SignatureInformation.scala
```diff
@@ -0,0 +1,61 @@
...
+      val eff = eff0 match {
+        case Type.Cst(TypeConstructor.Pure, _) => ""
+        case p => raw" \ " + FormatType.formatType(p)
+      }
+
+      s"(${args.mkString(", ")}): $retTpe$eff"
+  }
+}
+
+case class SignatureInformation(label: String, documentation: Option[String], parameters: List[ParameterInformation], activeParameter: Int) {
```
magnus-madsen, line 53: Why is this here and not in its own file?

magnus-madsen: nvm

magnus-madsen:
Why do some of these class not have toJSON?

magnus-madsen:
We need to add VSCode-LSP support here first.

Then release a new Flix.

And then update the VSCode extension.

magnus-madsen:
You can already create the PR for the VSCode ext

main/src/ca/uwaterloo/flix/api/lsp/SignatureInformation.scala
```diff
@@ -0,0 +1,61 @@
...
+
+object SignatureInformation {
+  def from(defn: TypedAst.Def, activeParameter: Int)(implicit flix: Flix): SignatureInformation = {
+    val label = defn.sym.toString + getLabelForSpec(defn.spec)
+    val documentation = defn.spec.doc.text
+    val parameters = defn.spec.fparams.map(ParameterInformation.from)
+    SignatureInformation(label, Some(documentation), parameters, activeParameter)
+  }
+
+  private def getLabelForSpec(spec: TypedAst.Spec)(implicit flix: Flix): String = spec match {
```
magnus-madsen, line 33: This function does not belong here. These should just be data classes.

stormckey:
> You can already create the PR for the VSCode ext

I'm now testing with it.

main/src/ca/uwaterloo/flix/api/lsp/SignatureInformation.scala
stormckey: So where do we usuallyl place these util functions? Should I add a new Util file under lsp?

stormckey:
It now works fine:

https://github.com/user-attachments/assets/581f05c1-b990-4dd1-b6e5-0d8ff07323da

stormckey:
The PR to the vscode extension is ready here https://github.com/flix/vscode-flix/pull/482.

I can now explain a bit about the impl:
- The front end will always fill in all the arguments, with fake vars, whose loc is marked as synthetic. However, these fake arguments always follows the real ones.
- If there is only parentheses like `foo()` the `()` will be treated a a unit and passes as the first argument.
- So when there is parentheses, we can assumes there will always be leading real args.
- We will find the first arg whose range include ours. Note that this check is inclusive for both ends, which means there are four places valid for the cursor for a argument like `333` whose length is 3.
- I manually add a command to trigger  the signature help in VSCode, when there is a snippet for DefCompletion. See issue https://github.com/microsoft/language-server-protocol/issues/274. This command is specific to VSCode so it's not included in toLsp4j. For users from neovim, they need the help from plugins to achieve this.

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -1562,7 +1562,9 @@ object Resolver {
     val (directArgs, cloArgs) = exps.splitAt(arity)
 
     val fparamsPadding = mkFreshFparams(arity - directArgs.length, loc.asSynthetic)
-    val argsPadding = fparamsPadding.map(fp => ResolvedAst.Expr.Var(fp.sym, loc.asSynthetic))
+    val lastPos = loc.sp2.copy(col = (loc.sp2.col - 1).toShort)
+    val paddingLoc = loc.copy(sp1 = exps.lastOption.map(exp => exp.loc.sp2.copy(col = (exp.loc.sp2.col + 1).toShort)).getOrElse(lastPos), sp2 = lastPos)
+    val argsPadding = fparamsPadding.map(fp => ResolvedAst.Expr.Var(fp.sym, paddingLoc.asSynthetic))
```
magnus-madsen, line 1567: What are these changes?

main/src/ca/uwaterloo/flix/api/lsp/provider/SignatureHelpProvider.scala
```diff
@@ -0,0 +1,52 @@
...
+      Visitor.visitRoot(root, stack, InsideAcceptor(uri, leftPos))
+    }
+
+    stack.getStack.collectFirst {
+      // Find the nearest function application
+      case TypedAst.Expr.ApplyDef(defnSymUse, exps, _, _, _, _) =>
+        // Count the number of arguments applied
+        // The number of arguments applied is the number of non-synthetic expressions
+        val argsNumApplied = exps.indexWhere { exp =>
+          pos.line == exp.loc.sp2.line && pos.character <= exp.loc.sp2.col && pos.character >= exp.loc.sp1.col
```
magnus-madsen, line 44: Do we not already have a helper for this?

main/src/ca/uwaterloo/flix/api/lsp/provider/SignatureHelpProvider.scala
```diff
@@ -0,0 +1,52 @@
...
+  def provideSignatureHelp(uri: String, pos: Position)(implicit root: Root, flix: Flix): Option[SignatureHelp] = {
+    val stack = StackConsumer()
+
+    if (pos.character >= 2) {
+      val leftPos = Position(pos.line, pos.character - 1)
+      Visitor.visitRoot(root, stack, InsideAcceptor(uri, leftPos))
+    }
+
+    stack.getStack.collectFirst {
+      // Find the nearest function application
```
magnus-madsen, line 39: Nearest in what sense? Whats the order of the stack

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
stormckey: The padding args' locs. When we can not find a match for any real args, we will then try to find the first padding arg. Before, the loc for all the padding args is the same -- the whole applyDef, which is actually a 100% successful including check for the cursor.

main/src/ca/uwaterloo/flix/api/lsp/provider/SignatureHelpProvider.scala
stormckey: I think it's inside the visitor. And it's not inclusive for both sides.

main/src/ca/uwaterloo/flix/api/lsp/provider/SignatureHelpProvider.scala
stormckey: It's just from a Visitor with an InsideAcceptor. Given that:
- we must visit a node's parent node before visit it.
- If node A include cursor, parent of A will also include the cursor.

So imo I think the stack is just a path from the leaf node that includes the cursor to the root.

This code is first used in HoleCompleter. I reuse it to find the context for signature help.

main/src/ca/uwaterloo/flix/api/lsp/provider/SignatureHelpProvider.scala
stormckey: Hmmm, I think I should update the logic here.

stormckey: We do have another `contains` in SourceLocation. But it's between two SourceLocations while we have a Position and a SourceLocation here.

Maybe add another function for Position?

magnus-madsen: I would clean up the logic, move it into SourceLocation, and add some tests

main/src/ca/uwaterloo/flix/api/lsp/provider/SignatureHelpProvider.scala
magnus-madsen: Add something like the above as a comment

main/src/ca/uwaterloo/flix/api/lsp/provider/SignatureHelpProvider.scala
stormckey: Got it.

main/src/ca/uwaterloo/flix/api/lsp/provider/SignatureHelpProvider.scala
stormckey: Ok, this stack is also constructed in CompletionProvider, do we think we should create a helper function to be used by both providers?

magnus-madsen: Yes

stormckey:
I'll add tests for position in following PR.

main/src/ca/uwaterloo/flix/api/lsp/Position.scala
```diff
@@ -84,4 +83,12 @@ case class Position(line: Int, character: Int) extends Ordered[Position] {
     val lineComparison = this.line.compare(that.line)
     if (lineComparison != 0) lineComparison else this.character.compare(that.character)
   }
+
+  /**
+    * Returns `true` if `this` position is contained by the given source location `loc`
+    * This check is inclusive for both ends.
+    */
+  def containedBy(loc: SourceLocation): Boolean =
```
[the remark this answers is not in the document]
stormckey: We can define `containedBy` here or `contain` in SourceLocation. And their impl should be 90% similar with a different field accessing. 

First I try to define it in SourcePosition since there is a partial order there. But that comparison is between SourcePositions. To convert a position to SourcePosition, we have to provide extra source to it.

The simplest way, like in my impl, is just to access the field you need.

So we can also impl by direcly checking in SourceLocation. 

Or add lteq, gteq to SourcePosition(besides PartialOrder) then call it from SourceLocation.

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
stormckey: Actually this fix is not very solid, see:
image
We have to figure out a solution here. If you want I can first rollback this change and leave if for next PR

magnus-madsen:
I am OK to merge this PR except for the change to Resolver.scala.

I just don't understand what is happening there and I worry about that change.

If we take that code out does the PR break?

It seems you also want a better solution, not this hack.

Is there something minimal we can do now? E.g. maybe we only have it work for one argument functions and then this PR can be merged.

Then in a followup PR we can address this difficult issue?

stormckey:
> I am OK to merge this PR except for the change to Resolver.scala.
> 
> I just don't understand what is happening there and I worry about that change.
> 
> If we take that code out does the PR break?
> 
> It seems you also want a better solution, not this hack.
> 
> Is there something minimal we can do now? E.g. maybe we only have it work for one argument functions and then this PR can be merged.
> 
> Then in a followup PR we can address this difficult issue?

We can, I'll cancel that.

magnus-madsen (approved):

magnus-madsen:
Next week I will release a new Flix and then update VSCode.

PR #10188, opened 2025-03-24
feat: add new workflow for completion invariant

stormckey:


magnus-madsen:
Can we rename "completion invariant" to "IDE Completion Tests" everywhere

stormckey:
Shall we also rename TestCompletionProvider, the name of the test class?

magnus-madsen (approved):

PR #10189, opened 2025-03-24
refactor: break down completion invariant tests

stormckey:
Programs is lifted to the top, otherwise there will be a warning.

### DIFF
--- a/main/test/ca/uwaterloo/flix/api/lsp/TestCompletionProvider.scala
+++ b/main/test/ca/uwaterloo/flix/api/lsp/TestCompletionProvider.scala
@@ -19,30 +19,12 @@ package ca.uwaterloo.flix.api.lsp
 import ca.uwaterloo.flix.api.Flix
 import ca.uwaterloo.flix.api.lsp.provider.CompletionProvider
 import ca.uwaterloo.flix.language.CompilationMessage
-import ca.uwaterloo.flix.language.ast.Token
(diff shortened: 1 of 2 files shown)


magnus-madsen (approved):

main/test/ca/uwaterloo/flix/api/lsp/TestCompletionProvider.scala
```diff
@@ -103,6 +85,30 @@ class TestCompletionProvider extends AnyFunSuite  {
        |""".stripMargin
   )
 
+  test("No completions after complete keyword"){
+    Programs.foreach{ program =>
+      val (root, flix, errors) = compile(program, Options.Default)
+      val source = mkSource(program)
+      val keywordTokens = root.tokens(source).toList.filter(_.kind.isKeyword)
+      keywordTokens.foreach{ token =>
+        val completions = CompletionProvider.autoComplete(Uri, Position.from(token.sp2), errors)(root, flix)
```
magnus-madsen, line 94: Note that this does not check all locations within in the keyword.

stormckey: So we should also check that?

Currently, for "let", we will check "let|"

So we should also check "l|e|t", how about "|let"?

magnus-madsen: Yes

PR #10191, opened 2025-03-24
feat: better SourceLocation display

stormckey:
I found it useful to show the whole range of the location


magnus-madsen:
Need to be discussed first. Has two many impact points.

stormckey:
I just find myself turning this on every time I want to check the end place of locations. Is this string repr used for any other purpose?

magnus-madsen:
> I just find myself turning this on every time I want to check the end place of locations. Is this string repr used for any other purpose?

Yes. I think it's used in many places for user facing code.

PR #10192, opened 2025-03-25
feat: fresh constraint gen variables all synthetic

JonathanStarup:


JonathanStarup:
@mlutze do you like this? its not the nicest but at least now its harder to forget `loc.asSynthetic`

mlutze:
Would it be better to have something specific in `Type.scala` like `Type.freshMetaVar`? Surely the need for synthetic vars exists outside of `Typer`.

magnus-madsen (approved):

magnus-madsen:
This is a good improvement. Maybe it can be further improved.

PR #10193, opened 2025-03-25
feat: finer invariant check for keyword and literal completion

stormckey:
Now we call autocomplete for every position inside token.

Do you think we should add support for multi-line literals? By adding a function offset to loc to calc the position?

### DIFF
--- a/main/test/ca/uwaterloo/flix/api/lsp/TestCompletionProvider.scala
+++ b/main/test/ca/uwaterloo/flix/api/lsp/TestCompletionProvider.scala
@@ -19,6 +19,7 @@ package ca.uwaterloo.flix.api.lsp
 import ca.uwaterloo.flix.api.Flix
 import ca.uwaterloo.flix.api.lsp.provider.CompletionProvider
 import ca.uwaterloo.flix.language.CompilationMessage
+import ca.uwaterloo.flix.language.ast.Token
 import ca.uwaterloo.flix.language.ast.TypedAst.Root
 import ca.uwaterloo.flix.language.ast.shared.{Input, SecurityContext, Source}
 import ca.uwaterloo.flix.util.Options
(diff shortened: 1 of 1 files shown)


main/test/ca/uwaterloo/flix/api/lsp/TestCompletionProvider.scala
```diff
@@ -91,8 +91,12 @@ class TestCompletionProvider extends AnyFunSuite  {
       val source = mkSource(program)
       val keywordTokens = root.tokens(source).toList.filter(_.kind.isKeyword)
       keywordTokens.foreach{ token =>
-        val completions = CompletionProvider.autoComplete(Uri, Position.from(token.sp2), errors)(root, flix)
-        assert(completions.items.isEmpty)
+        // We will test all possible offsets in the keyword, including the start and end of the keyword
+        (0 to token.text.length).foreach{ offset =>
```
magnus-madsen, line 95: I think this should be extracted into a helper function.

magnus-madsen: This function should take a token and return a list of SourcePositions

stormckey:
Now we have three nested foreach, if you prefer we can write it in a big for:
```scala
    for {
      program <- Programs
      (root, flix, errors) = compile(program, Options.Default)
      source = mkSource(program)
      keywordTokens = root.tokens(source).toList.filter(_.kind.isKeyword)
      token <- keywordTokens
      pos <- getAllPositionsWithinToken(token)
      completions = CompletionProvider.autoComplete(Uri, pos, errors)(root, flix)
    } assert(completions.items.isEmpty)
```

main/test/ca/uwaterloo/flix/api/lsp/TestCompletionProvider.scala
```diff
@@ -143,4 +142,18 @@ class TestCompletionProvider extends AnyFunSuite  {
...
+  /**
+    * Returns all positions within the given token.
+    *
+    * For example, give a token "def", we will return a list of positions:
+    * - |def
+    * - d|ef
+    * - de|f
+    * - def|
+    */
+  private def getAllPositionsWithinToken(token: Token): List[Position] =
```
magnus-madsen, line 155: Are you sure this is inclusive in both directions?

stormckey: yes

magnus-madsen (approved):

PR #10200, opened 2025-03-26
test: introduce ResiliencySuite

mlutze:
Depends on
- https://github.com/flix/flix/pull/10199

### DIFF
--- a/main/test/flix/resiliency/ResiliencySuite.scala
+++ b/main/test/flix/resiliency/ResiliencySuite.scala
@@ -1,0 +1,30 @@
+package flix.resiliency
+
+import ca.uwaterloo.flix.api.Flix
+import ca.uwaterloo.flix.language.ast.shared.SecurityContext.AllPermissions
+import org.scalatest.funsuite.AnyFunSuite
+
+import java.nio.file.{Files, Path}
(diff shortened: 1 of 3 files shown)


magnus-madsen:
What's the point here? And do you want to expand on the code?

mlutze:
The point is that the vast majority of our tests contain correct code or at least valid syntax.

The point of this suite is to be sure that resiliency does not cause problems. So we give it invalid Flix code and ensure that the compiler does not crash.

main/test/flix/resiliency/ResiliencySuite.scala
```diff
@@ -0,0 +1,30 @@
...
+import org.scalatest.funsuite.AnyFunSuite
+
+import java.nio.file.{Files, Path}
+import scala.jdk.CollectionConverters.IteratorHasAsScala
+
+class ResiliencySuite extends AnyFunSuite {
+
+  for {
+    file <- Files.list(Path.of("main", "test", "flix", "resiliency")).iterator().asScala
+    _ = println("hello 1")
```
magnus-madsen, line 14: I don't understand this code.

magnus-madsen (approved):

PR #10208, opened 2025-03-27
feat: must use everything (remove `@MustUse`)

JonathanStarup:
1. You cannot know just from a given type whether it can be ignored - `@MustUse` is broken
2. The default is that you can ignore types - very contrary to Flix design
3. `@MustUse` doesn't work in general (e.g. cant write `def example(f: Unit -> t \ ef): Unit \ ef where t is not @MustUse`)
4. The price to require Unit for statement is very low (just write `discard e`)
  a. `discard` is mostly used for java methods - Flix code generally avoids  such patterns
  b. The standard library only needed 21 discards, 20 of them java related.
5. Requiring Unit for statement position finds bug (already 1 bug found in this PR)
  a. This adheres to Flix's general safety vibe

Previous PR attempt: #9709

### DIFF
--- a/main/src/library/FileWrite.flix
+++ b/main/src/library/FileWrite.flix
@@ -117,7 +117,7 @@ mod FileWrite {
             def write(data, f, k) = {
                 try {
                     let bytes = data#str.getBytes(StandardCharsets.UTF_8);
-                    Files.write(Paths.get(f), bytes, (...{}: Vector[OpenOption]));
+                    discard Files.write(Paths.get(f), bytes, (...{}: Vector[OpenOption]));
                     k(())
                 } catch {
                     case ex: InvalidPathException          => Err(IoError(ErrorKind.InvalidPath, ex.getMessage()))
@@ -146,7 +146,7 @@ mod FileWrite {
                 try {
                     region rc {
                         let bytes = Vector.toArray(rc, data);
-                        Files.write(Paths.get(f), bytes, (...{}: Vector[OpenOption]));
+                        discard Files.write(Paths.get(f), bytes, (...{}: Vector[OpenOption]));
                         k(())
                     }
                 } catch {
@@ -160,7 +160,7 @@ mod FileWrite {
                 try {
                     let bytes = data#str.getBytes(StandardCharsets.UTF_8);
                     let opt: OpenOption = checked_cast(StandardOpenOption.APPEND);
-                    Files.write(Paths.get(f), bytes, ...{opt});
+                    discard Files.write(Paths.get(f), bytes, ...{opt});
                     k(())
                 } catch {
                     case ex: InvalidPathException          => Err(IoError(ErrorKind.InvalidPath, ex.getMessage()))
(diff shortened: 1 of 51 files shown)


main/test/flix/Test.Exp.Struct.Put.flix
```diff
@@ -159,9 +159,10 @@ mod Nested {
     def nestedPut01(): Bool =
         region rc {
             let nested = new Nested @ rc { v = new Nested @ rc { v = new Nested @ rc {v = new Nested @ rc {v = 15} }} };
-            nested->v->v->v->v == 15;
+            let b1 = nested->v->v->v->v == 15;
```
JonathanStarup, line 162: This check was ignored by mistake since Bool was not MustUse

mlutze:
Did it find a bug in your Talpin thing?

magnus-madsen:
I think last this time was discussed the conclusion was that it is not really workable.

I think Java interop shows that is really unworkable.

magnus-madsen:
This level of strictness:

```
    discard frame.getContentPane().add(label);
```

is just not workable. 

We need some (unknown) solution in between.

JonathanStarup:
> I think Java interop shows that is really unworkable.

I think java interop is the only place its annoying, so IMO we should fix it from that perspective (maybe change the rules if the statement expression is a java call?).

But the current state of things is an inconsistent halfway measure so that is also not great

main/src/library/FileWrite.flix
```diff
@@ -117,7 +117,7 @@ mod FileWrite {
             def write(data, f, k) = {
                 try {
                     let bytes = data#str.getBytes(StandardCharsets.UTF_8);
-                    Files.write(Paths.get(f), bytes, (...{}: Vector[OpenOption]));
+                    discard Files.write(Paths.get(f), bytes, (...{}: Vector[OpenOption]));
```
JonathanStarup, line 120: @magnus-madsen I think this might also just be wrong, which is highlighted by the discard requirement. This PR does require some typing, but overall, it's not that much, and it's very good at finding errors.

JonathanStarup: Not looking at the tests, it costs 26-36 `discard` to find 2-6 bugs, depending on how you count. Seems like a pretty good ratio to me

magnus-madsen:
I think this is a big language change, so we need to discuss it in detail.

I understand the merits, but I am just saying there are serious trade-offs here.

JonathanStarup:
Yeah makes sense. I dont know if id call it a serious tradeoff but i also made the pr to see what the consequences would be yea

JonathanStarup:
@magnus-madsen This is already really two decisions - remove the current system and then add the unit requirement. I think requiring Unit might have alternative solutions but i don't think there is a world where we should maintain the current system.

magnus-madsen:
We need to have a spiritual journey before we start thinking about a PR for this.

mlutze:
> We need to have a spiritual journey before we start thinking about a PR for this.

I'll add it to the calendar.

JonathanStarup:
> We need to have a spiritual journey before we start thinking about a PR for this.

The point of this PR is to inform that decision. See the implications. But yeah lets have the discussion.

magnus-madsen:
We may consider a middle-ground. I don't see how this can be done in the presence of Java interop (and a desire for adoption).

PR #10216, opened 2025-03-30
fix: string interpolation after double backslash

### LINKED ISSUE #10215: Wrong string parsing
```flix
def main(): Unit \ IO = println("\\${12}")
```
This prints
```
\${12}
```

stetimi:
```
flix> "${10+2}"
12                                                                              
flix> "\${10+2}"
${10+2}                                                                         
flix> "\\${10+2}"
\12                                                                             
flix> "\\\%{10+2}"
\%{10+2}                                                                        
flix> "\\\\%{10+2}"
\\12                                                                            
```

Fixes #10215

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
@@ -22,6 +22,7 @@ import ca.uwaterloo.flix.language.dbg.AstPrinter.DebugNoOp
 import ca.uwaterloo.flix.language.errors.LexerError
 import ca.uwaterloo.flix.util.{InternalCompilerException, ParOps}
 
+import scala.Function.const
 import scala.collection.mutable
 import scala.util.Random
 
@@ -603,13 +604,21 @@ object Lexer {
     s.sc.advanceWhile(_.isWhitespace)
 
   /**
-    * Moves the current position past all pairs of `\` and any other character.
-    *
-    * This is useful to avoid `\'` and `\"` ending the lexing of literals.
-    */
-  private def consumeSingleEscapes()(implicit s: State): Unit =
-    while (s.sc.advanceIfMatch('\\')) {
+   * Moves the current position past all pairs of `\` and any other character, returning
+   * true if any '\' are seen.
+   *
+   * This is useful to avoid `\'` and `\"` ending the lexing of literals, and to
+   * determine whether a '$' before a '{' is escaped or a string interpolation indicator.
+   */
+  private def consumeSingleEscapes()(implicit s: State): Boolean =
+    if (s.sc.advanceIfMatch('\\')) {
       advance()
+      while (s.sc.advanceIfMatch('\\')) {
+        advance()
+      }
+      true
+    } else {
+      false
     }
 
   /**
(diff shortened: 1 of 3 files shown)


stetimi:
Perhaps though it may be better to have DoubleBackslash as a token?

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -699,10 +699,18 @@ object Lexer {
       // Note: `sc.peek` returns `EOF` if out of bounds, different from `peek`.
```
JonathanStarup, line 699: I think youre right that this is the only way to "locally" fix this issue - but to me it points to a flaw in `consumeSingleEscapes`. Could you instead add an optional argument `ignore: Option[Char] = None` that allows you to call `consumeSingleEscapes(ignore = Some('$'))` here?

JonathanStarup: oh i guess i has to be `consumeSingleEscapes(ignore = Set('$', '%'))`

stetimi: Updated - hope this is what you mean. 

Side note: Should the escaping be documented? I don't see it in https://doc.flix.dev/string-interpolation.html?search=interpolation
Also the %{} notation doesn't appear to be documented.

JonathanStarup: sorry - after skimming through the documentation i realized it is a bit more complicated and i might have led you down the wrong path.

Here is the behaviour I would expect

| input | output |
| --- | --- |
| `"\$"` | Invalid escape `\$` |
| `"\${"` | `"${"` |
| `"\\${"` | Unterminated interpolation |
| `"\\\${"` | `"\${"` |

The problem in the existing code is that once it sees `\${` or `\%{`, the code doesn't know if `\` is escaped by a previous `\`, as you've seen in your first fix of the code.

I think that the cleanest solution is for `consumeSingleEscapes` to return a boolean that is true if any escapes were found, lets call it `esc`.

- If `\${` is found and `esc` is true, then do not go into lexing it as an interpolation
- if `\${` is found and `esc` is false, then `$` is not part of an escape, and therefore it is an interpolation

Does this make sense? I think this handles everything neatly with minimum code complexity. Sorry to make you change it again

JonathanStarup: Changing the function to
```scala
private def consumeSingleEscapes()(implicit s: State): Boolean =
    if (s.sc.advanceIfMatch('\\')) {
      advance()
      while (s.sc.advanceIfMatch('\\')) {
          advance()
      }
      true
    } else {
      false
    }
```

stetimi: I'll try out with your new function. No worries about changing it again - all part of the learning experience.

JonathanStarup: Ah right, I misremember - thats fine then

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -695,14 +704,13 @@ object Lexer {
...
+      val hasEscapes = consumeSingleEscapes()
       // Note: `sc.peek` returns `EOF` if out of bounds, different from `peek`.
       var p = s.sc.peek
       // Check for the beginning of a string interpolation.
-      val prevPrev = previousPrevious()
       val prev = previous()
-      val isInterpolation = !prevPrev.contains('\\') && prev.contains('$') && p == '{'
-      val isDebug = !prevPrev.contains('\\') && prev.contains('%') && p == '{'
+      val isInterpolation = !hasEscapes && prev.contains('$') & p == '{'
+      val isDebug = !hasEscapes &&  prev.contains('%') && p == '{'
```
JonathanStarup, line 713: Other than the extra whitespace here it looks perfect :) thanks

stetimi: Took out the extra space, I'll mark as resolved.

JonathanStarup (approved):
thanks, nice with extra hands :)

PR #10221, opened 2025-03-31
feat: add LangCensus app

stormckey:
To use this app, you need to export your github name and github token to environment variable GITHUB_NAME and GITHUB_TOKEN.

The output looks like:
```
                                                                                
                                                                                
==================== Analysis Result Per Repo ====================


Repository: Foggy-whale/Masked_UnMasked_Face_Recognition
Shell Script: 26 additions, 0 deletions.
Markdown: 59 additions, 0 deletions.
Python: 1788 additions, 614 deletions.
Jupyter Notebook: 5494 additions, 47 deletions.

Repository: ruoxining/HCodeFront
Python: 12 additions, 0 deletions.
Jupyter Notebook: 375 additions, 100 deletions.
TypeScript JSX: 546 additions, 260 deletions.
JavaScript: 3054 additions, 1557 deletions.

...

==================== Analysis Result For User ====================

Vim script: 1 additions, 0 deletions.
OCaml Interface: 1 additions, 146 deletions.
SystemVerilog: 2 additions, 2 deletions.
OPAM: 4 additions, 5 deletions.
Gradle: 7 additions, 0 deletions.
Menhir: 35 additions, 164 deletions.
TypeScript: 49 additions, 79 deletions.
Sass: 66 additions, 0 deletions.
Flix: 69 additions, 69 deletions.
Matlab: 120 additions, 60 deletions.
TeX Class: 139 additions, 0 deletions.
SQL: 186 additions, 0 deletions.
Typst: 246 additions, 306 deletions.
TOML: 375 additions, 63 deletions.
Linker Script: 545 additions, 3 deletions.
TypeScript JSX: 605 additions, 313 deletions.
HTML: 784 additions, 1003 deletions.
Lua: 915 additions, 42 deletions.
OCaml: 1173 additions, 1736 deletions.
Shell Script: 1448 additions, 65 deletions.
CSS: 1559 additions, 481 deletions.
Java: 1568 additions, 158 deletions.
TeX: 2024 additions, 880 deletions.
YAML: 2403 additions, 971 deletions.
Rust: 2757 additions, 2576 deletions.
C++: 4779 additions, 1420 deletions.
Python: 6035 additions, 5461 deletions.
Assembly: 6912 additions, 2446 deletions.
Jupyter Notebook: 7212 additions, 499 deletions.
Scala: 10022 additions, 6900 deletions.
JavaScript: 10565 additions, 9441 deletions.
C: 11993 additions, 1056 deletions.
Markdown: 13539 additions, 3945 deletions.
C Header: 15395 additions, 308 deletions.

==================== Analysis Result By Bytes ====================

RenderScript: 69 Bytes
WebAssembly: 145 Bytes
Standard ML: 936 Bytes
SCSS: 1084 Bytes
MATLAB: 1424 Bytes
Vim Script: 1761 Bytes
Thrift: 2194 Bytes
Lex: 4423 Bytes
CMake: 4537 Bytes
GLSL: 4760 Bytes
Typst: 4904 Bytes
Lean: 5177 Bytes
Yacc: 9787 Bytes
HTML: 11832 Bytes
F*: 12654 Bytes
Handlebars: 15077 Bytes
Makefile: 25920 Bytes
Lua: 28511 Bytes
Shell: 48564 Bytes
Java: 67366 Bytes
CSS: 70753 Bytes
Assembly: 102299 Bytes
Rust: 126705 Bytes
OCaml: 141396 Bytes
JavaScript: 150252 Bytes
TeX: 175875 Bytes
TypeScript: 229536 Bytes
Markdown: 247688 Bytes
Python: 392861 Bytes
C++: 411869 Bytes
Jupyter Notebook: 456581 Bytes
C: 905739 Bytes
Scala: 4382356 Bytes
```

We have a Config eff to get user config from the environment.

a Github eff to get data through github api

an Analyse eff to get statistics through Github

a Display eff to filter and print the result

a Failure eff for all sorts of errors


examples/apps/lancensus/src/Util.flix
```diff
@@ -0,0 +1,72 @@
...
+                        http <- String.stripSuffix({substr = ">; rel=\"next\""}, httpWithoutPrefix)
+                    ) yield getAllPages(http, headers);
+                    let nextPages = Option.getWithDefault(Nil, nextPagesOpt);
+                    thisPage ::: nextPages
+            case Result.Err(err) => AnalysisFailure.raiseRequestFailure(err)
+        }
+    }
+    
+    pub def extractExt(filePath: String): Option[String] = {
+        let generatedPaths = regex".*(/?dist/|/?build/|/?_build/|/?out/|/?target/).*" :: regex".*(package-lock\\.json|yarn\\.lock)$" :: regex".*/?node_modules/.*" :: regex".*(ssl\\.crt|\\.env|\\.config\\.js)$" :: Nil;
```
mlutze, line 60: What's going on here? :0

examples/apps/lancensus/src/Util.flix
```diff
@@ -0,0 +1,72 @@
...
+        match HttpWithResult.get(url, headers) {
+            case Result.Ok(response)  => 
+                if (Http.Response.status(response) != 200)
+                    Nil
+                else 
+                    let thisPage: List[JsonElement] = response |> Http.Response.body |> parse |> fromJson;
+                    let nextPagesOpt = forM (
+                        linkValueList <- Http.Response.headers(response) |> Map.get("link");
+                        linkString <- List.head(linkValueList);
+                        // http <- Regex.(regex"<([^>]*)>; rel=\"next\"", linkString)
```
mlutze, line 47: Commented code

mlutze (commented):
Cool! Can you go through and add some comments? There's some complex stuff going on which could be clarified with documentation.

examples/apps/lancensus/src/Util.flix
stormckey: Filter our generated paths :)

Some tools will generated tons of code that should not be counted.

examples/apps/lancensus/src/Util.flix
stormckey: This is left to be discussed.

Do you think it's a good idea to turn to regex here, or you think it's okay to manipulate the string?

stormckey:
I've added comment to most functions. Feel free to point out any function that is still unclear.

stormckey:
There are still some potential update points:
- Is there a way to get the path separator of the system? Currently I use a hardcoded '/'.
- So failure is not vital, we can resume from them, tbd.
- Is there a method to convert jsonError to some human readable string?

mlutze:
> Is there a way to get the path separator of the system? Currently I use a hardcoded '/'.

Yes, you can use `Environment.getFileSeparator()`

> So failure is not vital, we can resume from them. This will make the implementation more complicated, so I am not sure if we should do this.

What do you mean about resuming? I guess if someone's token is invalid then we should just report an error, right?

> Is there a method to convert jsonError to some human readable string?

There's nothing for it in the library, apparently, oops.

But the error has two components: the path where the error occurred and a set of expected constructs. So you should be able to make something out of that.

stormckey:
> Yes, you can use `Environment.getFileSeparator()`

I've just realized that we get the file path from github api, while git use '/' as the separator, so we should not care about what the separator for the host is.

> > So failure is not vital, we can resume from them. This will make the implementation more complicated, so I am not sure if we should do this.
> 
> What do you mean about resuming? I guess if someone's token is invalid then we should just report an error, right?

Yes, that is the general way we process error now. Actually we will report error and stop analysing if there is a single error in a field accessing to one commit's statistic.

But I am not sure if it will make the program too complicated if we want to handle that case properly. The current failure simplifies the code a lot. Do you think we should do that?

> But the error has two components: the path where the error occurred and a set of expected constructs. So you should be able to make something out of that.

I use this code to print the error message, I think this is what they stand for?
```rust
                let JsonError.JsonError(p, set) = err; 
                println("Failed to access ${set} in path ${p}");
```

mlutze:
> But I am not sure if it will make the program too complicated if we want to handle that case properly. The current failure simplifies the code a lot. Do you think we should do that?

I think a fatal error for invalid JSON is reasonable, because if Github is returning invalid JSON, then we can't really trust anything it's giving us :)

> I use this code to print the error message, I think this is what they stand for?
> 
> ```rust
>                 let JsonError.JsonError(p, set) = err; 
>                 println("Failed to access ${set} in path ${p}");
> ```

Yes, I would write something like "expected one of ${set} at JSON location ${p}"

mlutze:
Is this ready for in-depth review?

stormckey:
> Is this ready for in-depth review?

Almost？Should I first add support for multi-threading before you review?

mlutze:
> > Is this ready for in-depth review?
> 
> Almost？Should I first add support for multi-threading before you review?

That sounds complicated :)

stormckey:
> > > Is this ready for in-depth review?
> > 
> > 
> > Almost？Should I first add support for multi-threading before you review?
> 
> That sounds complicated :)

We an push that later :) So you can take an indepth review now.

examples/apps/langcensus/flix.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name        = "lancensus"
```
mlutze, line 2: update name

examples/apps/langcensus/src/Analyse.flix
```diff
@@ -0,0 +1,71 @@
+use Util.Stat
+use Util.Stats
+
+///
+/// The effect to analyse the a user's github data.
```
mlutze, line 5: "the a" -> "a"

examples/apps/langcensus/src/Analyse.flix
```diff
@@ -0,0 +1,71 @@
...
+    def analyseByByte(repos: List[String]): Map[String, BigInt]
+}
+
+mod Analyse {
+    use Json.JsonElement
+
+    ///
+    /// Handles analysis effect through Github Effect.
+    ///
+    pub def runWithGithub(f: Unit -> a \ ef): a \ ef - Analyse + Github + AnalysisFailure = {
```
mlutze, line 25: The nested folds in this function are hard to follow.

I think you might be able to get something a bit more understandable with a `foreach-yield`, like:

```
let allStats = foreach (
  (repo, shas) <- commits;
  sha <- shas;
  addsDels <- commitToAddsDels(repo, sha);
} yield addsDels;
// now merge all the maps
```

examples/apps/langcensus/src/Analyse.flix
```diff
@@ -0,0 +1,71 @@
...
+                    }, acc, languageStatMap)
+                }, Map.empty(), languageStats) |> resume
+            }
+        }
+    }
+
+    ///
+    /// Gets the adds and dels for each language for a specific commit.
+    ///
+    def commitToAddsDels(stat: Map[String, (BigInt, BigInt)], repo:String, sha: String): Map[String, (BigInt, BigInt)] \ AnalysisFailure + Github = {
```
mlutze, line 54: This function is actually doing two things: getting the data and adding it to the map.

It would be clearer if the function just returned a new map, and then we combine the maps afterward.

examples/apps/langcensus/src/Analyse.flix
mlutze: And something similar for the `analyseByByte` implementation.

examples/apps/langcensus/src/Display.flix
```diff
@@ -0,0 +1,162 @@
...
+    ///
+    def statFilter(stat: Stat): Stat = {
+        Map.filter(lines -> {
+            let (adds, _) = lines;
+            adds > 0ii
+        }, stat)
+    }
+        
+    ///
+    /// Prints the stat for a single language.
```
mlutze, line 64: This comment does not seem right. It looks like we're going over all the languages, right?

examples/apps/langcensus/src/Display.flix
```diff
@@ -0,0 +1,162 @@
...
+                            let (newAdds, newDels) = newData;
+                            Map.insert(language, (adds + newAdds, dels + newDels), acc)
+                    }
+            }
+        }, Map.empty(), stat)
+
+    ///
+    /// Sorts the map by converting it to a list.
+    ///
+    def sortMapByList(map: Map[k, v]): List[(k, v)] with Order[v] = {
```
mlutze, line 93: Maybe `sortMapByValue`

examples/apps/langcensus/src/Github.flix
```diff
@@ -0,0 +1,84 @@
+use Util.Stats
+use Json.JsonElement
+
+///
+/// Gets all sorts of data from github.
+/// Will get the user and token through Config effect.
+///
+pub eff Github {
+    ///
+    /// Gets a list of the full name of reporitories that the user has contributed to, regardless of privacy.
```
mlutze, line 10: `repositories` :)

examples/apps/langcensus/src/Github.flix
```diff
@@ -0,0 +1,84 @@
...
+    /// 
+    /// Returns a map from the repository name to a list of commit shas.
+    ///
+    def getCommits(repos: List[String]): Map[String, List[String]]
+
+    ///
+    /// Gets all the language data from the given repository names.
+    /// 
+    /// Returns a map from the repository name to the json of lanuage data.
+    /// The language data should be a map from laguange name to the byte of code for this languge in this repository.
```
mlutze, line 25: language :D

examples/apps/langcensus/src/Display.flix
```diff
@@ -0,0 +1,162 @@
+use Util.Stats
```
mlutze, line 1: The `foreach` stuff might be useful in this file too.

examples/apps/langcensus/src/Util.flix
```diff
@@ -0,0 +1,110 @@
...
+        match HttpWithResult.get(url, headers) {
+            case Result.Ok(response)  => 
+                if (Http.Response.status(response) != 200)
+                    Nil
+                else 
+                    let thisPage: List[JsonElement] = response |> Http.Response.body |> parse |> fromJson;
+                    let nextPagesOpt = forM (
+                        linkValueList <- Http.Response.headers(response) |> Map.get("link");
+                        linkString <- List.head(linkValueList);
+                        // http <- Regex.(regex"<([^>]*)>; rel=\"next\"", linkString)
```
mlutze, line 73: We agreed to use the regex, right?

examples/apps/langcensus/src/Analyse.flix
stormckey: One of the reasons I don't use foreach is that its error message is so hard to understand, do you know what is wrong here? Are we not allowed to use effectful functions in foreach?

image

mlutze: > One of the reasons I don't use foreach is that its error message is so hard to understand, do you know what is wrong here?

Not sure :(

> Are we not allowed to use effectful functions in foreach?

It should be possible to use effectful functions. It desugars into a pile of `Iterator.flatMap`s I believe.

stormckey:
I've pushed a new version, which solves most of the problems except the foreach ones.

examples/apps/langcensus/src/Analyse.flix
```diff
@@ -0,0 +1,107 @@
...
+    
+    def mergeStats(stats: List[(String, Stat)]): Map[String, Stat] = 
+        List.foldLeft((acc, data) -> {
+            let (repo, stat) = data;
+            match Map.get(repo, acc) {
+                case None => Map.insert(repo, stat, acc)
+                case Some(oldStat) => 
+                    let newStat = mergeStat(oldStat, stat);
+                    Map.insert(repo, newStat, acc)
+            }
```
mlutze, line 93: This can be simplified to `Map.insertWith(mergeStat, repo, stat, acc)`

examples/apps/langcensus/src/Analyse.flix
```diff
@@ -0,0 +1,107 @@
...
+        
+    def mergeStat(oldStat: Stat, stat: Stat): Stat = {
+        Map.foldLeftWithKey((acc, lang, data) -> {
+            let (adds, dels) = data;
+            match Map.get(lang, acc) {
+                case Option.None        => Map.insert(lang, data, acc)
+                case Option.Some(oldData) => 
+                    let (oldAdds, oldDels) = oldData;
+                    Map.insert(lang, (adds + oldAdds, dels + oldDels), acc)
+            }
```
mlutze, line 104: Similarly, this can be simplified with `insertWith`.

You can also do nested pattern matching:

```
case Some((oldAdds, oldDels))
```

examples/apps/langcensus/src/Analyse.flix
mlutze: Here is a fix:

```
                let allStat: List[(String, BigInt)] = foreach (
                    repo <- repos;
                    languageStat = Github.getLanguages(repo);
                    languageStatMap = Util.fromJson(languageStat);
                    (lang, size) <- (languageStatMap: Map[String, BigInt])
                ) yield (lang, size);
                mergeStatsByByte(allStat) |> resume
```


There were 3 problems:

1. mergeStatsByByte was not implemented
2. languageStatMap was unconstrained -- we only knew that it was something `FromJson`.
3. size was unconstrained (again, only `FromJson`)

examples/apps/langcensus/src/Analyse.flix
```diff
@@ -0,0 +1,107 @@
...
+    def commitToAddsDels(repo:String, sha: String): (String, Stat) \ AnalysisFailure + Github = {
+        let stat = Github.getCommit(repo, sha) |> Util.accessJson("files") |> List.foldLeft((acc, file) -> {
+            let filename =  Util.accessJson("filename", file);
+            let extOpt =  filename |> Util.extractExt;
+            match extOpt {
+                case Option.None  => acc
+                case Option.Some(ext) => 
+                        let additions = Util.accessJson("additions", file);
+                        let deletions = Util.accessJson("deletions", file);
+                        match Map.get(ext, acc) {
```
mlutze, line 76: `Map.insertWith`

examples/apps/langcensus/src/Analyse.flix
```diff
@@ -0,0 +1,107 @@
...
+                        case Option.None                   => Map.insert(ext, (additions, deletions), acc)
```
mlutze, line 77: indent

examples/apps/langcensus/src/Analyse.flix
stormckey: Wow, I don't expect to write the type annotation there.

It works!

examples/apps/langcensus/src/Analyse.flix
stormckey: You are right, I didn't use nested patten matching since it doesn't work in some cases.

like this:
image
image
Do you know why?

mlutze: When doing it as a function parameter, the syntax is (I think) `match (x, (y, z)) -> ...`. Or if there are multiple parameters to the function, you need to curry, like `match (x, y) -> match (z, (w, q)) -> ...`

stormckey: Curry do help, thanks!

stormckey:
This new version is much more idiomatic :)

examples/apps/langcensus/src/Analyse.flix
```diff
@@ -0,0 +1,95 @@
...
+        } with handler Analyse {
+            def analyseByRepo(commits, resume) = 
+                let allStats : List[(String, Stat)] = foreach (
+                    (repo, shas) <- commits;
+                    sha <- shas
+                ) yield commitToAddsDels(repo, sha);
+                mergeStats(allStats) |> resume
+
+            def analyseByByte(repos, resume) = {
+                let allStats: List[(String, BigInt)] = foreach (
```
mlutze, line 38: Looks nice :)

examples/apps/langcensus/src/Display.flix
```diff
@@ -0,0 +1,142 @@
...
+                println("\n==================== Analysis Result Per Repo ====================\n");
+                Map.forEach((repo, thisStat) -> {
+                    println("\nRepository: ${repo}");
+                    thisStat |> statFilter |> renameAndMerge |> sortMapByValue |> printEachLanguage              
+                }, stat) |> cont
+            }
+            
+            def displayByUser(stats, resume) = {
+                println("\n==================== Analysis Result For User ====================\n");
+                Map.foldLeft((acc, stat) -> {
```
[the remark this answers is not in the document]
stormckey: We are merging data from different maps, this could also be done using foreach + ref, which is the imperative way to impl this, do you prefer this impl?
```
            def displayByUser(stats, resume) = {
                println("\n==================== Analysis Result For User ====================\n");
                region rc {
                    let map: Ref[Map[String, (BigInt, BigInt)], rc] = Ref.fresh(rc, Map.empty());
                    foreach (
                        (_, stat) <- stats;
                        (ext, data) <- stat
                    ) {
                        Ref.put(
                            Map.insertWith(
                                match (oldAdds, oldDels) -> match (newAdds, newDels) -> (oldAdds + newAdds, oldDels + newDels),
                                ext, data, Ref.get(map)
                            ),
                            map
                        )
                    };
                    Ref.get(map) |> statFilter |> renameAndMerge |> sortMapByValue |> printEachLanguage |> resume
                }
            }
```

examples/apps/langcensus/src/Display.flix
```diff
@@ -0,0 +1,142 @@
...
+    /// Prints the stat for each language.
+    ///
+    def printEachLanguage(stat: List[(String, (BigInt, BigInt))]): Unit \ IO = 
+        List.forEach(match (language, (adds, dels)) -> println("${language}: ${adds} additions, ${dels} deletions."), stat)
+        
+    ///
+    /// Renames the extention to langauges and merge entries of the same language.
+    /// 
+    def renameAndMerge(stat: Stat): Stat = 
+        Map.foldLeftWithKey((acc, ext, data) -> {
```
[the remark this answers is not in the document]
stormckey: It seems `Map.filterMapWithKey` only change the value? Here we need to change the key from ext to language, and we need to merge entries with the same key.

examples/apps/langcensus/src/Display.flix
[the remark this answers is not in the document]
mlutze: Maybe we can use a mutable map instead of a reference to a map, then convert it to immutable at the end.

examples/apps/langcensus/src/Display.flix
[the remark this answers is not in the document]
mlutze: Dang :(

examples/apps/langcensus/src/Main.flix
```diff
@@ -0,0 +1,25 @@
+///
+/// Gets the language usage statistics for the user.
+///
+/// Will print three kinds of statictics:
```
mlutze, line 4: statistics :P

examples/apps/langcensus/src/Main.flix
```diff
@@ -0,0 +1,25 @@
...
+///   - The adds and dels of the user for each language for each repository.
+///   - The total adds and dels of the user for each language.
+///   - The total bytes of code for each language for all the reporitorites the user has paticipated in.
```
mlutze, line 7: participated :P :P

mlutze (approved):

magnus-madsen:
@mlutze Last chance for comments. Otherwise merge.

PR #10229, opened 2025-04-01
feat: new completion invariant: no completions on name definition

stormckey:
name definition invariant in #10138

We collect names from the root and filter out those whose src does not start with "<test>", which means that they are not defined in the tested program, but in the library.

Then I add Sourceable and Locatable to all the Symbols used, so that I can concatenate them together, do the filtering, and map to loc.

Then we just call completion and test its emptiness.

Possible improvements:

- We can modify the name a little bit, the invariant should still hold. But I will wait to reuse the helper functions from other PR to modify the code.
- We can include more name definitions, like parameters, local variable defs, and other bindings.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/Symbol.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/Symbol.scala
@@ -18,7 +18,7 @@ package ca.uwaterloo.flix.language.ast
 
 import ca.uwaterloo.flix.api.Flix
 import ca.uwaterloo.flix.language.ast.Name.{Ident, NName}
-import ca.uwaterloo.flix.language.ast.shared.{BoundBy, ModuleKind, QualifiedSym, Scope, Source, VarText}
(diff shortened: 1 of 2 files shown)


magnus-madsen:
Instead of grouping these together, I would create tests for each

- No completion for VarSym def
- No completion for EnumSym def

and so on.

stormckey:
> Instead of grouping these together, I would create tests for each
> 
> * No completion for VarSym def
> * No completion for EnumSym def
> 
> and so on.

Makes sense.

stormckey:
Now they are separated.

How about these two possible improvements?

> Possible improvements:
> 
> * We can modify the name a little bit, the invariant should still hold. But I will wait to reuse the helper functions from other PR to modify the code.
> * We can include more name definitions, like parameters, local variable defs, and other bindings.

magnus-madsen:
Can we undo changes to Symbol.scala?

magnus-madsen (approved):

PR #10238, opened 2025-04-02
refactor: Polish monomorpher

JonathanStarup:


magnus-madsen:
Perhaps some of this can be resurrect in smaller PRs?

In any case, the next steps for the Monomorpher involves a new MonoType and merging some of lowering into it.

PR #10247, opened 2025-04-04
chore: address unused declarations in Simplifier

mlutze:


main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -517,8 +517,8 @@ object Simplifier {
           case TypeConstructor.Arrow(_) =>
             // Remove the effect from the arrow.
             // Arrow type arguments are ordered (effect, args.., result type).
-            val eff :: targs = tpe.typeArguments
-            val (args, List(res)) = targs.splitAt(targs.length - 1)
+            val args = tpe.arrowArgTypes
+            val res = tpe.arrowResultType
```
JonathanStarup, line 521: this is nicer but i don't think we should traverse twice. we should make a nice function for it.

mlutze: Is your concern efficiency? Because I don't see this as a major bottleneck.

JonathanStarup (approved):

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
JonathanStarup: That's the only upside here, but imo it's just a nice rule of thumb to ask the whole "question" upfront. It usually allows better performance and cleaner code.

PR #10258, opened 2025-04-04
refactor: clean up `specializeSigSym`

JonathanStarup:


main/src/ca/uwaterloo/flix/language/phase/Monomorpher.scala
```diff
@@ -688,49 +688,45 @@ object Monomorpher {
...
-    val defns = instance.defs.filter {
-      d => d.sym.text == sig.sym.name
-    }
+    val instance = instances((sym.trt, tyCon))
+    val defns = instance.defs.filter(_.sym.text == sig.sym.name)
 
     (sig.exp, defns) match {
       // An instance implementation exists. Use it.
-      case (_, defn :: Nil) => specializeDef(defn, tpe)
+      case (_, defn :: Nil) => defn
```
mlutze, line 717: Why don't we specialize the def?

JonathanStarup: now this function just resolves the def, the specializeSigSym calls this function and then calls specializeDef

mlutze: Ok. Please add something to the doc to say that the returned def is not specialized yet.

JonathanStarup: My thought is that everything in this file that specializes something has to be called `specializeX` - that is already mostly how it is, with a few exceptions that i plan to fix. do you think that's enough or should it be written out?

mlutze: Yeah ok

mlutze (approved):

PR #10268, opened 2025-04-08
refactor: introduce Type.decomposeArrow

mlutze:
I'm not totally convinced by this. The three named functions make it clear what we're getting from the type, but returning a tuple means you need to know what order the returned values are in.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/Type.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/Type.scala
@@ -218,13 +218,22 @@ sealed trait Type {
   }
 
   /**
+    * Returns the effect, argument types, and result type of `this` arrow type.
+    *
+    * NB: Assumes that `this` type is an arrow.
+    */
+  def decomposeArrow: (Type, List[Type], Type) = (typeConstructor, typeArguments) match {
+    case (Some(TypeConstructor.Arrow(_)), eff :: (args :+ res)) => (eff, args, res)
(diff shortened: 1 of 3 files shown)


magnus-madsen:
Introduce a case class? DecomposedArrow(..., ..., ...) ?

mlutze:
> Introduce a case class? DecomposedArrow(..., ..., ...) ?

I thought about this, but seems like a lot of overhead...

But if everyone likes it, I'll do it.

JonathanStarup:
> I thought about this, but seems like a lot of overhead...

I guess its okay when comparing to double type traversal? and with a sufficiently smart compiler ... :P

mlutze:
> > I thought about this, but seems like a lot of overhead...
> 
> I guess its okay when comparing to double type traversal? and with a sufficiently smart compiler ... :P

I didn't mean performance, just code complexity.

JonathanStarup:
Right, a record would be nice here

magnus-madsen:
What are you trying to do here and how broadly do you plan to apply it?

Do you want a family of views on types?

mlutze:
I hereby transfer ownership of this idea to Jonathan.

JonathanStarup:
Haha 😄 

I wanted to avoid code that calls
```
Get me arrow effect
Get me arrow args
Get me arrow result
```
And instead have one method that deconstructs the arrow in one go

magnus-madsen:
Let us design then program.

PR #10270, opened 2025-04-08
experiment: introduce LocationVerifier phase

### LINKED ISSUE #10131: Add `LocationVerifier` compiler phase
Add a compiler phase that verifies that source locations are meaningful.

For example, if we have:

```
exp = if (exp1) exp2 else exp3
```

We would expect the source locations of `exp1`, `exp2` and `exp3` to be fully contained within `exp`. 

Moreover, we would expect `exp1` to come before `exp2`, and so forth.

We would also expect `exp` and `exp3` to end in the same place.

This phase would not have any impact of the program being compiled; it would purely exist to help us find errors.

stormckey:
fixes #10131

This phase is used to verify that the locations in the AST meet the invariants we required.

Now it's only a prototype for if then else.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/Flix.scala
+++ b/main/src/ca/uwaterloo/flix/api/Flix.scala
@@ -26,8 +26,8 @@ import ca.uwaterloo.flix.language.phase.optimizer.Optimizer
 import ca.uwaterloo.flix.language.{CompilationMessage, GenSym}
 import ca.uwaterloo.flix.runtime.CompilationResult
 import ca.uwaterloo.flix.tools.Summary
-import ca.uwaterloo.flix.util.Formatter.NoFormatter
 import ca.uwaterloo.flix.util.*
+import ca.uwaterloo.flix.util.Formatter.NoFormatter
 import ca.uwaterloo.flix.util.collection.{Chain, MultiMap}
 import ca.uwaterloo.flix.util.tc.Debug
 
@@ -80,7 +80,7 @@ class Flix {
   /**
     * The set of known Java classes and interfaces.
     */
-  private var availableClasses: AvailableClasses = AvailableClasses(getJavaPlatformClassesAndInterfaces)
+  private var availableClasses: AvailableClasses = AvailableClasses(getJavaPlatformClassesAndInterfaces())
 
   /**
     * A cache of ASTs for incremental compilation.
@@ -825,7 +825,7 @@ class Flix {
   /**
     * Returns all Java classes and interfaces in the current Java Platform.
     */
-  private def getJavaPlatformClassesAndInterfaces: MultiMap[List[String], String] = {
+  private def getJavaPlatformClassesAndInterfaces(): MultiMap[List[String], String] = {
     getPackageContent(ClassList.TheList)
   }
 
--- a/main/src/ca/uwaterloo/flix/language/ast/SourceLocation.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/SourceLocation.scala
@@ -71,15 +71,6 @@ case class SourceLocation(isReal: Boolean, sp1: SourcePosition, sp2: SourcePosit
   }
 
   /**
-    * Returns `true` if `this` [[SourceLocation]] is before `that`, otherwise `false`.
-    *
-    * 'Before' means that `this` ends before or at the same position as `that` begins.
-    */
-  def isBefore(that: SourceLocation): Boolean = {
-    SourcePosition.PartialOrder.lteq(this.sp2, that.sp1)
-  }
-
-  /**
     * Returns the one-indexed line where the entity ends.
     */
   def endLine: Int = sp2.lineOneIndexed
@@ -132,11 +123,6 @@ case class SourceLocation(isReal: Boolean, sp1: SourcePosition, sp2: SourcePosit
     * Returns a string representation of `this` source location with the line and column numbers.
     */
   def format: String = s"${source.name}:$beginLine:$beginCol"
-
-  /**
-    * Returns a string representation of `this` source location with the line and column numbers for the start and end.
-    */
-  def toFullString: String = s"${source.name}:$beginLine:$beginCol-$endLine:$endCol"
 
   /**
     * Returns the source text of the source location.
--- a/main/test/ca/uwaterloo/flix/verifier/LocationVerifier.scala
+++ b/main/test/ca/uwaterloo/flix/verifier/LocationVerifier.scala
@@ -21,8 +21,8 @@ import ca.uwaterloo.flix.language.ast.*
 import ca.uwaterloo.flix.language.ast.TypedAst.Pattern.Record
 import ca.uwaterloo.flix.language.ast.TypedAst.Predicate.Body
 import ca.uwaterloo.flix.language.ast.TypedAst.{Expr, Pattern, RestrictableChoosePattern}
-import ca.uwaterloo.flix.language.errors.LocationError
 import ca.uwaterloo.flix.util.{InternalCompilerException, ParOps}
+import ca.uwaterloo.flix.verifier.errors.LocationError
 
 import java.util.concurrent.ConcurrentLinkedQueue
 
@@ -617,13 +617,22 @@ object LocationVerifier {
   private def verifyAppearanceOrder(locs: List[SourceLocation])(implicit sctx: SharedContext): Unit = {
     locs.sliding(2).foreach {
       case List(prevLoc, currLoc) =>
-        if (!prevLoc.isBefore(currLoc)) {
+        if (!isBefore(prevLoc, currLoc)) {
           sctx.add(
             LocationError.mkAppearanceOrderError(prevLoc, currLoc)
           )
         }
       case _ => ()
     }
+  }
+
+  /**
+    * Returns `true` if `thisLoc` is before `thatLoc`, otherwise `false`.
+    *
+    * 'Before' means that `thisLoc` ends before or in the same position as `thatLoc` begins.
+    */
+  def isBefore(thisLoc: SourceLocation, thatLoc: SourceLocation): Boolean = {
+    SourcePosition.PartialOrder.lteq(thisLoc.sp2, thatLoc.sp1)
   }
 
   /**
--- a/main/test/ca/uwaterloo/flix/verifier/errors/LocationError.scala
+++ b/main/test/ca/uwaterloo/flix/verifier/errors/LocationError.scala
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package ca.uwaterloo.flix.language.errors
+package ca.uwaterloo.flix.verifier.errors
 
 import ca.uwaterloo.flix.language.ast.SourceLocation
 import ca.uwaterloo.flix.util.Formatter.NoFormatter.code
@@ -29,7 +29,7 @@ object LocationError {
     */
   def mkChildOutOfBoundError(parentLoc: SourceLocation, loc: SourceLocation): InternalCompilerException = {
     val message =
-      s""">> The location of the child ${loc.toFullString} is not contained in the location of its parent ${parentLoc.toFullString}.
+      s""">> The location of the child ${toFullString(loc)} is not contained in the location of its parent ${toFullString(parentLoc)}.
          |
          |${code(loc, "Child location.")}
          |""".stripMargin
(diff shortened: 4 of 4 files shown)


main/src/ca/uwaterloo/flix/language/errors/LocationError.scala
```diff
@@ -0,0 +1,86 @@
...
+/**
+  * A common super-type for location errors.
+  */
+trait LocationError extends CompilationMessage {
+  val kind: CompilationMessageKind = CompilationMessageKind.LocationError
+}
+
+object LocationError {
+
+  private def format(loc: SourceLocation): String = s"${loc.source.name}:${loc.beginLine}:${loc.beginCol}-${loc.endLine}:${loc.endCol}"
```
stormckey, line 32: I use this format to print the complete location, since the original format for SourceLocation will not show sp2, but we need sp2 in this phase to debug.

Relative to #10191

stormckey:
We already have tons of errors with just three invariants over one kind of node. And we need to check them one by one to solve them all.

And obviously, we can not merge this PR before we solve all the errors, for all the nodes, which I think is a lot of work and may remain unsolved when I leave.

stormckey:
Another problem is should we check for synthetic locations. I can give an example where it seems reasonable to check them:

```
add(1, |)
```

In the example above, the second argument, which has not yet been filled, is synthetic.

But we actually want to make sure its location is something like:
```
add(1, |)
      ^^
```
so that we could know for which location we should show the signature help for the second parameter.

But yes, we can handle it in other ways like tracing the location of last ',' and the ')'

On the other hand, I've already see some error in this checking for the desugared if-then-else in forM, which makes it harder (or even impossible) to verify the location for them.

main/src/ca/uwaterloo/flix/language/phase/LocationVerifier.scala
```diff
@@ -0,0 +1,83 @@
...
+    (root.copy(defs = defs), List.empty[LocationError])
+  }
+
+  private def visitDef(defn: TypedAst.Def)(implicit sctx: SharedContext): TypedAst.Def = {
+    visitExp(defn.exp)
+    defn
+  }
+
+  private def visitExp(exp0: Expr)(implicit sctx: SharedContext): Unit = exp0 match {
+    case Expr.IfThenElse(exp1, exp2, exp3, _, _, loc) =>
```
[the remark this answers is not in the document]
mlutze: I think for now, we should only consider things that are syntactically present in the program. For if-then-else, the type and effect are not written in the program, they are just inferred. For something like "ascribe", the declared type and effect are written, so they should be visited.

mlutze:
> We already have 1500 lines of errors with just three invariants over one kind of node. And we need to check them one by one to solve them all. (The actual number of errors could be maybe just 3-4, since these errors could be repeated for several times)
> 
> And obviously, we can not merge this PR before we solve all the errors, for all the nodes. But if we dont want to leave every discussion for each error in this PR, we have to figure out a way to handle it.

I think this phase should not be part of the main compiler, but instead activated with a flag or otherwise somehow run as its own test.

Then we can create a separate test in CI (like we do for the fuzzing test) and allow that to fail for now, while we address the issues.

Then a new PR can solve the issues, and then a new PR can add more checks, etc.

mlutze:
> Another problem is should we check for synthetic locations. I can give an example where it seems reasonable to check them:
> 
> ```
> add(1 ,  )
> ```
> 
> In the example above, the second argument, which has not yet been filled, is synthetic.
> 
> But we actually want to make sure its location is something like:
> 
> ```
> add(1 ,  )
>        ^^
> ```
> 
> so that we can identify the correct locations where the signature help should be displayed for the second parameter.
> 
> But yes, we can handle it in other ways like tracing the location of last ',' and the ')'
> 
> On the other hand, I've already see some error in this checking for the desugared if-then-else in forM, which makes it harder (or even impossible) to verify the location for them.

Jonathan and I think that we ought to get rid of most desugarings so that this won't be a problem anymore, but that's a discussion for later.

If synthetic locations cause problems, then as a first step we can ignore them. Later we can look into whether there's a good solution (I suspect no-more-desugaring is the solution)

mlutze:
Good job on this :)

main/src/ca/uwaterloo/flix/language/phase/LocationVerifier.scala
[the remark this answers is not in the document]
stormckey: Makes sense :)

stormckey:
> I think this phase should not be part of the main compiler, but instead activated with a flag or otherwise somehow run as its own test.
> 
> Then we can create a separate test in CI (like we do for the fuzzing test) and allow that to fail for now, while we address the issues.
> 
> Then a new PR can solve the issues, and then a new PR can add more checks, etc.

Sounds great.

stormckey:
> I think this phase should not be part of the main compiler, but instead activated with a flag or otherwise somehow run as its own test.
> 
> Then we can create a separate test in CI (like we do for the fuzzing test) and allow that to fail for now, while we address the issues.
> 
> Then a new PR can solve the issues, and then a new PR can add more checks, etc.

That way we can just make it a special test, that run over a compiled root. It can stay away from the phases. 🤔

JonathanStarup:
i really like the errors, nice!

stormckey:
We may also set the location of the error to point to the parent node so that we can see the whole scene. This will result in 
image
instead of 
image

main/src/ca/uwaterloo/flix/language/ast/SourceLocation.scala
```diff
@@ -70,6 +70,15 @@ case class SourceLocation(isReal: Boolean, sp1: SourcePosition, sp2: SourcePosit
     thatBeginsLater && thatEndsBefore
   }
 
+  /**
+    * Returns `true` if `this` [[SourceLocation]] is before `that`, otherwise `false`.
+    *
+    * 'Before' means that `this` ends before or at the same position as `that` begins.
+    */
+  def isBefore(that: SourceLocation): Boolean = {
```
JonathanStarup, line 78: You can add these file changes as its own PR, good helper function

magnus-madsen:
So I suppose the question is whether this allow us to find small bugs and fix them?

JonathanStarup:
we can add `xverifylocations` and defautl it to false, then add one test that compiles the std lib with the flag enabled

stormckey:
I've turned to ICE. See if you like it.

We can collect all the exceptions through the sctx like what we do before for errors. Or we can just throw it once we find it. Currently I throw it. But that means we can just see one error at one run.

I keep the sctx since the decision has not been made.

main/src/ca/uwaterloo/flix/language/errors/LocationError.scala
JonathanStarup: I think its good to have in a method. maybe we add `SourceLocation.toFullString` or something like that

stormckey: Done

main/src/ca/uwaterloo/flix/Main.scala
```diff
@@ -103,6 +103,7 @@ object Main {
       xfuzzer = cmdOpts.xfuzzer,
       xprinttyper = cmdOpts.xprinttyper,
       xverifyeffects = cmdOpts.xverifyeffects,
+      xverifylocations = cmdOpts.xverifylocations,
```
JonathanStarup, line 106: magnus wanted one flag for all the verifiers, so maybe just call this `xverify`?

main/src/ca/uwaterloo/flix/util/Options.scala
```diff
@@ -47,6 +47,7 @@ object Options {
     xfuzzer = false,
     xprinttyper = None,
     xverifyeffects = false,
+    xverifylocations = false,
     xsubeffecting = Set.empty,
     XPerfN = None,
     XPerfFrontend = false,
```
JonathanStarup, line 53: And then below here (line 61) set the verify flag to true so its used in the tests

stormckey:
I have several problems about this verifier:
- Maybe we don't need to make it incremental, since we are not reusing the same flix from compilation to compilation. Another benefit is that we can call this verifier with just a root from test suites (if added)
- There are some other PR that is designing the parameters (xverify etc), maybe I should wait for them to finish first?
- What is the plan for bug reporting? we just crash using ICE? then this phase dont need to return anything.

mlutze:
> I have several problems about this verifier:
> 
>     * Maybe we don't need to make it incremental, since we are not reusing the same flix from compilation to compilation. Another benefit is that we can call this verifier with just a root from test suites (if added)

I agree. We shouldn't make it incremental.

>     * There is another PR that is designing the parameters (xverify etc), maybe I should wait for them to finish first?

Yes, let's wait for that to be in place.

>     * What is the plan for bug reporting? We just crash using ICE? Then this phase dont need to return anything. The downside it that we may what want to collect more than I bug in one verification, since currently there are many bugs.

Maybe we can collect all the errors and report them, and then crash if there are any.

stormckey:
> I agree. We shouldn't make it incremental.

Now it's not incremental. I modified the `run` function based on `EffectVerifier.run`.

> Yes, let's wait for that to be in place.

Waiting for #10309

> Maybe we can collect all the errors and report them, and then crash if there are any.

We can collect all the ICEs, but we can only throw one of them and that would be the only one we can observe?

stormckey:
> Maybe we just print when we see a problem and keep track if there's any problem. Then at the end we crash if we detected a problem along the way.

Done

stormckey:
Any further improvements for this PR? It's now rebuilt over the event system

main/src/ca/uwaterloo/flix/Main.scala
```diff
@@ -23,8 +23,8 @@ import ca.uwaterloo.flix.language.ast.Symbol
 import ca.uwaterloo.flix.language.phase.unification.zhegalkin.ZhegalkinPerf
 import ca.uwaterloo.flix.runtime.shell.Shell
 import ca.uwaterloo.flix.tools.*
-import ca.uwaterloo.flix.util.Validation.flatMapN
 import ca.uwaterloo.flix.util.*
+import ca.uwaterloo.flix.util.Validation.flatMapN
```
magnus-madsen, line 27: ???

main/src/ca/uwaterloo/flix/api/Flix.scala
```diff
@@ -80,7 +80,7 @@ class Flix {
   /**
     * The set of known Java classes and interfaces.
     */
-  private var availableClasses: AvailableClasses = AvailableClasses(getJavaPlatformClassesAndInterfaces())
+  private var availableClasses: AvailableClasses = AvailableClasses(getJavaPlatformClassesAndInterfaces)
```
magnus-madsen, line 83: separate PR, if you want

main/src/ca/uwaterloo/flix/language/ast/SourceLocation.scala
magnus-madsen, line 78: Dont extend production code to help with testing code

main/src/ca/uwaterloo/flix/language/ast/SourceLocation.scala
```diff
@@ -124,6 +133,11 @@ case class SourceLocation(isReal: Boolean, sp1: SourcePosition, sp2: SourcePosit
     */
   def format: String = s"${source.name}:$beginLine:$beginCol"
 
+  /**
+    * Returns a string representation of `this` source location with the line and column numbers for the start and end.
+    */
+  def toFullString: String = s"${source.name}:$beginLine:$beginCol-$endLine:$endCol"
```
magnus-madsen, line 139: Ditto

main/src/ca/uwaterloo/flix/language/errors/LocationError.scala
```diff
@@ -0,0 +1,68 @@
...
+ * limitations under the License.
+ */
+
+package ca.uwaterloo.flix.language.errors
+
+import ca.uwaterloo.flix.language.ast.SourceLocation
+import ca.uwaterloo.flix.util.Formatter.NoFormatter.code
+import ca.uwaterloo.flix.util.InternalCompilerException
+
+object LocationError {
```
magnus-madsen, line 23: This should not be here. It is test code.

main/src/ca/uwaterloo/flix/Main.scala
stormckey: Oh this is a legacy of a former change, I'll rollback.

stormckey:
> I am not planning to merge this until it has been useful for actually finding and fixing several bugs. Until then its value is unproven.

Ok, after the small updated I will use it to find several errors and create tickets for them. Then you can its impact.

magnus-madsen:
I hope to revisit this at some point in the future.

PR #10273, opened 2025-04-09
fix: filter out non-star type in TypeAliasCompleter

### LINKED ISSUE #10194: Static in TypeCompletion
Image

In Prelude.flix, we have
```scala
///
/// Static denotes the region of global lifetime.
///
pub type alias Static = IO
```

So what is the use case for Static? I think it should not pop up here right?

stormckey:
fixes #10194

We filter out non-star types in type alias completer, so that static is filtered out:

image

Is this a sound filtering? Are we filtering out some types that are actually needed in some UndefinedType errors?


magnus-madsen (approved):

magnus-madsen:
I will merge this as a practical consideration. But indeed it is not exactly right.

PR #10274, opened 2025-04-09
fix: fix broken semantic Token for handler

### LINKED ISSUE #10190: Broken semantic token for handler
Once mentioned in  #10092

Now it seems that this issue remains?
Everything is still green, although the loc of _unit is fixed:

Image

stormckey:
fixes #10190

Preview:
image

Method:
I filter out tokens for binder whose type has a synthetic location.
Also, I stop providing semantic token for all types with a synthetic location. If Synthetic location is not abused, I think that is safe?
I also change the token type for the op name (displayByRepo above) to Function

Known issue:
resume is not colored, the reason is that its location is synthetic and is not precise:
image

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
@@ -600,7 +600,7 @@ object SemanticTokensProvider {
       val st1 = Iterator(t)
       val st2 = rules.foldLeft(Iterator.empty[SemanticToken]) {
         case (acc, HandlerRule(op, fparams, exp ,loc)) =>
-          val st = SemanticToken(SemanticTokenType.Type, Nil, op.loc)
+          val st = SemanticToken(SemanticTokenType.Function, Nil, op.loc)
           val t1 = Iterator(st)
           val t2 = visitFormalParams(fparams)
           acc ++ t1 ++ t2 ++ visitExp(exp)
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
@@ -494,8 +494,8 @@ object SemanticTokensProvider {
           acc ++ Iterator(t) ++ visitType(tpe) ++ visitExp(exp)
       }
 
-    case Expr.RestrictableChoose(_, exp, rules, _, _, _) =>
-      val c = visitExp(exp)
+    case Expr.RestrictableChoose(_, exp1, rules, _, _, _) =>
+      val c = visitExp(exp1)
       rules.foldLeft(c) {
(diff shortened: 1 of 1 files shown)


main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
```diff
@@ -914,9 +918,12 @@ object SemanticTokensProvider {
     */
   private def visitFormalParam(fparam0: FormalParam): Iterator[SemanticToken] = fparam0 match {
     case FormalParam(bnd, _, tpe, _, _) =>
-      val o = getSemanticTokenType(bnd.sym, tpe)
-      val t = SemanticToken(o, Nil, bnd.sym.loc)
-      Iterator(t) ++ visitType(tpe)
+      val bndToken = if (!bnd.tpe.loc.isSynthetic) {
```
magnus-madsen, line 921: Does visitType not already do this?

stormckey: We are trying to provide two semantic tokens here. For the binder, we do not call visitType, so I just check if the type is synthetic here.

magnus-madsen:
@mlutze Can you review and if happy click merge?

magnus-madsen:
(hit wrong button)

mlutze:
I'm not sure this is right.

When we have a formal parameter like 

`def f(x: Foo)`

there are two locations we have to treat _separately_:

- `x`
- `Foo`

Even if `Foo`'s location is synthetic, like this:

```
def f(x)
```

`x`'s location should _not_ be synthetic, so we should use that.

stormckey:
> I'm not sure this is right.
> 
> When we have a formal parameter like
> 
> `def f(x: Foo)`
> 
> there are two locations we have to treat _separately_:
> 
> * `x`
> * `Foo`
> 
> Even if `Foo`'s location is synthetic, like this:
> 
> ```
> def f(x)
> ```
> 
> `x`'s location should _not_ be synthetic, so we should use that.

wait is that allowed? I've just tried:
image

And I think those synthetic types are the ones who colored the whole handler green. If we can not just filter out all of them. I think we have to change its location.

But sometimes in the handler, some type do refer to the whole body and thus range over the whole handler.

mlutze:
> wait is that allowed? 

It is allowed for local definitions (also effect handlers):

```
def foo(): String = {
  def bar(x) = x;
  bar("hi")
}
```


> And I think those synthetic types are the ones who colored the whole handler green. If we can not just filter out all of them. I think we have to change its location. (Or we can special case it for Handler)

In the above example, `x` should not have a synthetic location, so we should not filter it out.
But the _type_ of `x` (`String`) should have a synthetic location (since it's not present in the source), so we _should_ filter the type out.

In short, you should only see if the location of the thing itself is synthetic to tell whether to filter it or not. You should not use the location of other related things.

mlutze (approved):

PR #10277, opened 2025-04-10
feat: Verify Lexer

JonathanStarup:
I was curious about if the lexer was correct, given @stormckey's PR on location checking - everything checks out.


JonathanStarup:
just to have more than one error and fit into the framework. is there is middleway between this and just throwing ice?

magnus-madsen:
> just to have more than one error and fit into the framework. is there is middleway between this and just throwing ice?

I suggest ICE. This is a situation that never happens.

Alternatively, we can make custom output that is simply printed to std out.

main/test/ca/uwaterloo/flix/TestVerifiers.scala
```diff
@@ -0,0 +1,28 @@
+package ca.uwaterloo.flix
+
+import ca.uwaterloo.flix.api.{Flix, FlixEvent, FlixListener}
+import ca.uwaterloo.flix.util.Options
+import ca.uwaterloo.flix.verifier.{TokenVerifier, TypeVerifier}
+import org.scalatest.funsuite.AnyFunSuite
+
+class TestVerifiers extends AnyFunSuite with TestUtils {
+
+  test("RunAllVerifiers") {
```
JonathanStarup, line 10: I think its important to have one test that just runs all the verifiers (or one for each verifier) to make sure that they are always run at least once

JonathanStarup:
@magnus-madsen now in a version a la `TypeVerifier`

magnus-madsen (approved):

magnus-madsen (approved):

magnus-madsen:
Unfortunately this increases test time from 9min to 22min-- so some performance improvents are needed before we can merge.

JonathanStarup:
Damn really? Maybe it helps cache stuff to merge the passes into a single pass, but otherwise, I don't know.

We can also just run it once for the standard library. I don't think there is much benefit to running it on every test.

magnus-madsen:
The GitHub CI box shows the time

JonathanStarup:
Turns out one pass really matters. Now its 11 minutes but as I said, we can also just run it once on the standard lib instead of every test?

magnus-madsen:
Is it not supposed to be 8 or 9min still?

magnus-madsen:
> But as I said, we can also just run it once on the standard lib instead of every test?

I have no strong opinion. I just want these things to be simple and maintainable and separate from production code.

magnus-madsen:
I would like to revisit this in future.

I hope to have someone re-implement the lexer and then they can 'rediscover' these invariants.

JonathanStarup:
> I hope to have someone re-implement the lexer and then they can 'rediscover' these invariants.

Wouldn't it just be extra useful to have in case of a refactor? 

I'd like to finish this up and get it merged. The code is already written and i think it adds value since the lexer is tricky to modify

magnus-madsen:
How do you plan to fix the performance issues?

JonathanStarup:
Just run it once on TestVerifiers like discussed above (ill also get that ready)

JonathanStarup:
@magnus-madsen should be good to merge now - the previous times were also with the listener duplication so it actually quite fast

magnus-madsen (approved):

PR #10281, opened 2025-04-11
fix: drop comments from start of expression in parsing

### LINKED ISSUE #10218: Error location includes comment
```
def reverse(l: List[t]): List[t] = region rc {
    let m = MutDeque.empty(rc);
    let errMsg = "";
    def loop(l ) = match l {
        case Nil => MutDeque.toList(m)
        case n :: m => // Shadowing
            MutDeque.pushFront(x); // Useless Expression
            loop(m)
    };
    loop(l)
}
```

The `// Shadowing` comment has a red underline from the useless expression

stetimi:
closes #10218

This seemed too easy so is probably wrong! 

I minimised to this:
```
def f(): String  = { 
    // test
    MutDeque.pushFront(1);
    ""
}
```

The AST printed as below: the expression included the comment. 
So I called the comments() method at the start of Parser2.expression(). The AST no longer includes the comment node and the underlining in vscode looks correct.

```
                        Expr(
                            Apply(
                                Expr(
                                    CommentList(CommentLine["// test"]),
                                    QName(
                                        Ident(NameUpperCase["MutDeque"]),
                                        Dot["."],
                                        Ident(NameLowerCase["pushFront"])
                                    )
                                ),
```

### DIFF
--- a/examples/functional-style/pure-and-impure-functions.flix
+++ b/examples/functional-style/pure-and-impure-functions.flix
@@ -15,5 +15,5 @@ def twice(f: Int32 -> Int32, x: Int32): Int32 = f(f(x))
 /// We can pass a pure function to twice.
 pub def f(): Int32 = twice(inc, 42)
 
-/// But we *cannot* pass an impure function.
+// But we *cannot* pass an impure function.
 // pub def g(): Int32 = twice(printAndInc, 42)
--- a/main/src/ca/uwaterloo/flix/api/Flix.scala
+++ b/main/src/ca/uwaterloo/flix/api/Flix.scala
@@ -17,6 +17,7 @@
 package ca.uwaterloo.flix.api
 
 import ca.uwaterloo.flix.language.ast.*
+import ca.uwaterloo.flix.language.ast.TokenKind.{CommentBlock, CommentLine}
 import ca.uwaterloo.flix.language.ast.shared.{AvailableClasses, Input, SecurityContext, Source}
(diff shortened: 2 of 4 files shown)


magnus-madsen:
I think this works - but only in a small sense.

The problem is that now we drop comments before an expression. But there are so many other places that comments can appear. How do we handle them all? Surely we cannot insert calls to comment() everywhere, so we need a more systematic approach.

stetimi:
I may be getting out of my depth a little here, but I did find a possible bug...

In Flix.scala, I removed the non-doc comments after lexing, and put them back in again after parsing. I removed code in Parser2 that ignored non doc-comment tokens.

All the tests pass except one: when compiling the example "pure-and-impure-functions.flix", after my change the doc comments are flagged as misplaced.

```
/// But we *cannot* pass an impure function.
// pub def g(): Int32 = twice(printAndInc, 42)
```

magnus-madsen:
I think we need a clear plan before we can make changes. We cannot just add and remove some code. We need a better understanding of the underlying issue-- which is that tree's source location span too much when comments are involved.

I think thats where the fix needs to be. When we ask for the source location of a tree we should not consider comments.

magnus-madsen:
I am going to close this and suggest that you look at Parser2 near line 173.

magnus-madsen:
But perhaps instead of opening a new PR, write in the ticket the proposed fix.

PR #10284, opened 2025-04-11
fix: strict substitution on typematch effect

### LINKED ISSUE #10259: Monomorph nonstrict on effect in typematch
in Monomorph when visiting typematch expression, the constrict substitution is applied to the effect of the expression which is inconsistent and seems like a typo

JonathanStarup:
Fixes #10259


magnus-madsen (approved):

magnus-madsen:
No test case?

JonathanStarup:
I really can't think of a program to observe this change

PR #10295, opened 2025-04-14
feat: remove `Optimizer`

jaschdoc:


magnus-madsen:
Any flags or options we need to remove?

magnus-madsen:
I guess you removed the only relevant flag?

jaschdoc:
Yes I think so

magnus-madsen (approved):

PR #10296, opened 2025-04-14
experiment: introduce Condition, Then, Else treeKinds

### LINKED ISSUE #10219: No completion when writing if
When we are writing if, there are only parse errors:

Image

As a result, we have no completions.

This will not be improved if we add else:

Image

stormckey:
fixes #10219

And it makes the lsp work inside an incomplete if-then-else:
image
image

But still doesnt work here:
image

I manually constructed an Expr.Error in the Weeder, and since there is no placeholder, I randomly use MisplacedComment, just to make it compile.

We could handle that in other means, lke designing a placeholder, or constructing that in Parser2.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -1861,13 +1861,19 @@ object Parser2 {
       val mark = open()
       expect(TokenKind.KeywordIf, SyntacticContext.Expr.OtherExpr)
       expect(TokenKind.ParenL, SyntacticContext.Expr.OtherExpr)
+      val condMark = open()
       expression()
+      close(condMark, TreeKind.Expr.Condition)
       expect(TokenKind.ParenR, SyntacticContext.Expr.OtherExpr)
+      val thenMark = open()
       expression()
+      close(thenMark, TreeKind.Expr.Then)
       if (eat(TokenKind.KeywordElse)) {
         // Only call expression, if we found an 'else'. Otherwise when it is missing, defs might
         // get read as let-rec-defs.
+        val elseMark = open()
         expression()
+        close(elseMark, TreeKind.Expr.Else)
       }
       close(mark, TreeKind.Expr.IfThenElse)
     }
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -1385,9 +1385,9 @@ object Parser2 {
     /**
       * Parse an expression.
       *
-      * @param left         The operator that binds the left-hand-side of the expression.
-      * @param leftIsUnary  If true, the left-hand-side is an unary operator.
-      * @param exitTokens   A set of tokens that will prevent exprDelimited from consuming them.
+      * @param left        The operator that binds the left-hand-side of the expression.
+      * @param leftIsUnary If true, the left-hand-side is an unary operator.
+      * @param exitTokens  A set of tokens that will prevent exprDelimited from consuming them.
       */
     def expression(left: TokenKind = TokenKind.Eof, leftIsUnary: Boolean = false, exitTokens: Set[TokenKind] = Set.empty)(implicit s: State): Mark.Closed = {
       var lhs = exprDelimited(exitTokens)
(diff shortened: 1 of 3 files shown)


main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1261,17 +1261,26 @@ object Weeder2 {
       }
     }
 
+    /**
+      * The condition and then sub-trees are required, but the else sub-tree is optional.
```
JonathanStarup, line 1265: I think this comment should be moved to the `ifThenElseExpr` parser function. That function should say that there is always a cond/then child but only maybe a else child

stormckey:
The current code will produce this tree 
image
for 
image

The problem is `expression` will at least eat one token, as a result, ")" and "else" are eaten by condition and then. 

Is there another helper function that can avoid this eating? @JonathanStarup

JonathanStarup:
> The problem is expression will at least eat one token, as a result, ")" and "else" are eaten by condition and then.

This is a very general problem that we don't have a solution for. In theory the call to expression so be given that we expect a `)` at the end so it knows not to eat it.

stormckey:
> Member

In the first line of expression:
```scala
    def expression(left: TokenKind = TokenKind.Eof, leftIsUnary: Boolean = false)(implicit s: State): Mark.Closed = {
      var lhs = exprDelimited()
```
and in exprDelimited
```scala
    private def exprDelimited()(implicit s: State): Mark.Closed = {
      // If a new expression is added here then add it to FIRST_EXPR too.
      val mark = open()
      nth(0) match {
        case TokenKind.KeywordOpenVariant => openVariantExpr()
        case ....
        case t =>
          val mark = open()
          advance()
          val error = UnexpectedToken(expected = NamedTokenSet.Expression, actual = Some(t), SyntacticContext.Expr.OtherExpr, loc = currentSourceLocation())
          closeWithError(mark, error)
      }
      close(mark, TreeKind.Expr.Expr)
    }
```
The fallback is to eat one token and produce the expected expression error.

It's this line of code that eats ")" and "else". Do you think we should eat one token here?

JonathanStarup:
> It's this line of code that eats ")" and "else". Do you think we can remove the advance here?

No, we had to add the `advance` to avoid infinite loops for some programs. you would have to fix it via external context like my previous comment.

> Or we can match over Else and ParamR and choose to do nothing over them(the list could be extended)

Do you mean that `exprDelimited` sees `else`, then it doesn't advance for the error case?

magnus-madsen:
I will let you continue the discussion, but I just want to say that any larger changes are not warranted by this feature.

If we can fix it (or even just improve it!) with 10 changes line of code -- that is great.

But we are not looking for big changes just for the sake of if-then-else.

magnus-madsen:
> > But we are not looking for big changes just for the sake of if-then-else.
> 
> I get your point, but this small thing is a child of a larger issue in the parser which would improve resilience everywhere. Just to explain why I'm hinting to a parser-wide solution.

Sure, but do you have a specific design in mind? I don't know how to do it nor how to do anything without major (near impossible) changes.

magnus-madsen:
Thats why if we are going to do something specific for IfThenElse lets keep it small.

JonathanStarup:
> Sure, but do you have a specific design in mind?

Right now we have a static guess about when to exit parsing calls/loops (`isFirstXYZ`), instead we should have dynamic token sets that make parsing calls/loops exit out. This would mean that `exprDelimited` and others would have an implicit `exitTokens` argument. This can be added to relevant functions one at a time in small PR's.

> Thats why if we are going to do something specific for IfThenElse lets keep it small.

The above would be a generalizable version of @stormckey's suggestion. Just having it for exprDelimited would probably be a 10 line change.

stormckey:
> > Sure, but do you have a specific design in mind?
> 
> Right now we have a static guess about when to exit parsing calls/loops (`isFirstXYZ`), instead we should have dynamic token sets that make parsing calls/loops exit out. This would mean that `exprDelimited` and others would have an implicit `exitTokens` argument. This can be added to relevant functions one at a time in small PR's.
> 
> > Thats why if we are going to do something specific for IfThenElse lets keep it small.
> 
> The above would be a generalizable version of @stormckey's suggestion. Just having it for exprDelimited would probably be a 10 line change.

You mean we can pass a set of exitTokens, and when nth(0) is in the set, we will not advance in exprDelimited?

stormckey:
See if you like this solution.

It can solve the problem:
image
image

image
image
image
image
image

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -1586,92 +1593,104 @@ object Parser2 {
...
-        case TokenKind.KeywordDebug
-             | TokenKind.KeywordDebugBang
-             | TokenKind.KeywordDebugBangBang => debugExpr()
-        case t =>
-          val mark = open()
-          val error = UnexpectedToken(expected = NamedTokenSet.Expression, actual = Some(t), SyntacticContext.Expr.OtherExpr, loc = currentSourceLocation())
-          advance()
-          closeWithError(mark, error)
+      val token = nth(0)
+      if (exitTokens.contains(token)) {
```
JonathanStarup, line 1605: Exactly my thinking yes, but I would only do it for the case `case _ =>` error case. The difference is mostly philosophical.

The meaning would be `if you find something that you do not expect - exit and give an error if it is in "exitTokens" - continue looking if it is not in "exitTokens"`. The last part is not relevant here but i hope it makes some sense

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -1856,18 +1875,29 @@ object Parser2 {
...
+      */
     private def ifThenElseExpr()(implicit s: State): Mark.Closed = {
       assert(at(TokenKind.KeywordIf))
       val mark = open()
       expect(TokenKind.KeywordIf, SyntacticContext.Expr.OtherExpr)
       expect(TokenKind.ParenL, SyntacticContext.Expr.OtherExpr)
-      expression()
+      val exitTokens: Set[TokenKind] = Set(TokenKind.ParenR, TokenKind.KeywordElse)
+      val condMark = open()
+      expression(exitTokens = exitTokens)
```
JonathanStarup, line 1889: For this call the exitTokens should only be `ParenR` and in the next call it should only be `Then`

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -1856,18 +1869,28 @@ object Parser2 {
...
       expect(TokenKind.KeywordIf, SyntacticContext.Expr.OtherExpr)
       expect(TokenKind.ParenL, SyntacticContext.Expr.OtherExpr)
-      expression()
+      val condMark = open()
+      expression(exitTokens = Set(TokenKind.ParenR))
+      close(condMark, TreeKind.Expr.Condition)
       expect(TokenKind.ParenR, SyntacticContext.Expr.OtherExpr)
-      expression()
+      val thenMark = open()
+      expression(exitTokens = Set(TokenKind.KeywordElse))
```
JonathanStarup, line 1886: This one should be `Then`

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -1856,18 +1869,28 @@ object Parser2 {
...
+      close(thenMark, TreeKind.Expr.Then)
       if (eat(TokenKind.KeywordElse)) {
         // Only call expression, if we found an 'else'. Otherwise when it is missing, defs might
         // get read as let-rec-defs.
+        val elseMark = open()
         expression()
```
JonathanStarup, line 1892: This one can have exit token `Else` to better parse `if (true) 1 + else 2` I believe

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -1382,8 +1382,15 @@ object Parser2 {
...
     }
 
-    def expression(left: TokenKind = TokenKind.Eof, leftIsUnary: Boolean = false)(implicit s: State): Mark.Closed = {
-      var lhs = exprDelimited()
+    /**
+      * Parse an expression.
+      *
+      * @param left        The operator that binds the left-hand-side of the expression.
+      * @param leftIsUnary If true, the left-hand-side is an unary operator.
+      * @param exitTokens  A set of tokens that will prevent exprDelimited from consuming them.
```
JonathanStarup, line 1390: `.. from consuming them` -> `.. from consuming them as unexpected tokens`

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -1586,7 +1593,12 @@ object Parser2 {
       }
     }
 
-    private def exprDelimited()(implicit s: State): Mark.Closed = {
+    /**
+      * Parse a delimited expression.
+      *
+      * @param exitTokens A set of tokens that will prevent exprDelimited from consuming them.
```
JonathanStarup, line 1599: same thing here

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1263,15 +1263,20 @@ object Weeder2 {
...
-        case _ =>
-          val error = UnexpectedToken(NamedTokenSet.FromKinds(Set(TokenKind.KeywordElse)), actual = None, SyntacticContext.Expr.OtherExpr, hint = Some("the else-branch is required in Flix."), tree.loc)
-          sctx.errors.add(error)
-          Validation.Success(Expr.Error(error))
+      val condTreeVal = pick(TreeKind.Expr.Condition, tree)
+      val thenTreeVal = pick(TreeKind.Expr.Then, tree)
+      flatMapN(condTreeVal, thenTreeVal) {
+        case (condTree, thenTree) => tryPick(TreeKind.Expr.Else, tree) match {
+          case Some(elseTree) =>
+            flatMapN(pickExpr(condTree), pickExpr(thenTree), pickExpr(elseTree)) {
```
JonathanStarup, line 1271: This can just be mapN and then remove Success around the inside i think

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1263,15 +1263,20 @@ object Weeder2 {
...
+              (condExpr, thenExpr, elseExpr) => Validation.Success(Expr.IfThenElse(condExpr, thenExpr, elseExpr, tree.loc))
+            }
+          case None =>
+            flatMapN(pickExpr(condTree), pickExpr(thenTree)) {
```
JonathanStarup, line 1275: I think the error behaviour should be the same as before

JonathanStarup (commented):
I like this solution 👍 I think this is a good template for other similar parser issues

stormckey:
Making ElseTree an option will only make the impl more complicated, we can tree all three branches uniformly.

main/test/ca/uwaterloo/flix/language/phase/TestParser.scala
```diff
@@ -517,8 +516,8 @@ class TestParserRecovery extends AnyFunSuite with TestUtils {
   test("BadIfThenElse.03") {
     val input =
       """
-        |def foo(): Int32 = if (false) 123
         |def main(): Unit = ()
+        |def foo(): Int32 = if (false) 123
```
stormckey, line 520: Otherwise main will be considered a local def.

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1263,15 +1263,22 @@ object Weeder2 {
...
+        case (condTree, thenTree) => tryPick(TreeKind.Expr.Else, tree) match {
+          case Some(elseTree) =>
+            mapN(pickExpr(condTree), pickExpr(thenTree), pickExpr(elseTree)) {
+              (condExpr, thenExpr, elseExpr) => Expr.IfThenElse(condExpr, thenExpr, elseExpr, tree.loc)
+            }
+          case None =>
+            mapN(pickExpr(condTree), pickExpr(thenTree)) {
+              val error = UnexpectedToken(NamedTokenSet.FromKinds(Set(TokenKind.KeywordElse)), actual = None, SyntacticContext.Expr.OtherExpr, hint = Some("the else-branch is required in Flix."), tree.loc)
+              sctx.errors.add(error)
+              // This Expr.Error is just a placeholder.
```
JonathanStarup, line 1278: This comment can be deleted

magnus-madsen:
Conclusion:

Need to create a ticket describing the general principle and then come up with a list of expressions to apply it to.

magnus-madsen:
I would like to revisit this at some point in the future.

Issue #10293, opened 2025-04-14
Verifiers


magnus-madsen:
I propose the following:

- [ ] Verifiers live in test/ca/uwaterloo/flix/language/verifier
- [ ] The compiler cannot depend on anything in test. Except Flix.scala may depend on the verifiers.
- [ ] Verifiers must not use CompilationMessage, but _must_ throw ICE.
- [ ] Verifiers, since they are not part of the language proper, are free to print to the terminal or to files.
- [ ] Introduce a flag `--Xverify` that enables all verifiers and use it for tests.

magnus-madsen:
@JonathanStarup @mlutze @stormckey OK?

magnus-madsen:
TODO:

- Rework xnoverify
- Rework xverifyeffects
- Move EffectVerifier
- Move Verifier -> ClassVerifier

magnus-madsen:
Maybe instead of enabling verifiers for all tests, we should just add some test suite/cases that run the verifier on the std lib.

stormckey:
Are we using --Xverify to enable tests or to enable phases?

If the verifiers are just tests. We can just run them directly from IDEA or design a workflow for them.

Another problem is should we just raising an ICE once meet an error? Or accumulate all of them and print them somewhere?

magnus-madsen:
But in any case, lets try to think of verifiers as "test code" that is not part of production code.

mlutze:
I'm not sure about this:

> Introduce a flag --Xverify that enables all verifiers and use it for tests.

We might want more granularity

magnus-madsen:
> I'm not sure about this:
> 
> > Introduce a flag --Xverify that enables all verifiers and use it for tests.
> 
> We might want more granularity

We can drop the flag and just have test suites?

stormckey:
> > I'm not sure about this:
> > > Introduce a flag --Xverify that enables all verifiers and use it for tests.
> > 
> > 
> > We might want more granularity
> 
> We can drop the flag and just have test suites?

We can, If we want to run a specific verifier, we can just run that using IDEA. Also we can design workflows for them.

stormckey:
> > Maybe instead of enabling verifiers for all tests, we should just add some test suite/cases that run the verifier on the std lib.
> 
> I think it makes sense to separate them into cheap (run on all tests) and expensive (just run on std lib) verifications

So if we make it a phase and use a flag to control it, we will be able to run it over all tests, right? That would be a helpful feature.

[...]

PR #10301, opened 2025-04-15
refactor: add `verifier` folder

JonathanStarup:
Related to #10293


magnus-madsen (approved):

JonathanStarup:
But i think production code needs to call it, in order not to duplicate the compiler pipeline

magnus-madsen:
> But i think production code needs to call it, in order not to duplicate the compiler pipeline

We can use a FlixListener

PR #10309, opened 2025-04-17
feat: `verify` option

JonathanStarup:
- `--Xverify` enables all verifiers, none are enabled by default
  - On master `ClassVerifier` is always run, so this is a change for users
- `VerificationOptions` can be used internally to pick and choose any verifier
- `ClassVerifier` is run on all tests, while `TestVerifiers` runs all verifiers on the standard library
  - `EffectVerifier` fails currently, so its disabled for this PR 

Related to #10293

### DIFF
--- a/main/src/ca/uwaterloo/flix/util/Options.scala
+++ b/main/src/ca/uwaterloo/flix/util/Options.scala
@@ -39,13 +39,12 @@ object Options {
     threads = Runtime.getRuntime.availableProcessors(),
     loadClassFiles = true,
     assumeYes = false,
-    xnoverify = false,
     xprintphases = false,
     xnodeprecated = false,
     xsummary = false,
     xfuzzer = false,
     xprinttyper = None,
-    xverifyeffects = false,
+    xverify = VerificationOptions.EnableNone,
     xsubeffecting = Set.empty,
     XPerfN = None,
     XPerfFrontend = false,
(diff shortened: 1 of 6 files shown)


main/src/ca/uwaterloo/flix/util/Options.scala
```diff
@@ -175,3 +179,55 @@ object Subeffecting {
...
+
+  /**
+    * Enable all verifiers.
+    */
+  case object EnableAll extends VerificationOptions
+
+  /**
+    * Enable the listed verifiers.
+    */
+  case class EnableSome(verifiers: Set[Verifier]) extends VerificationOptions
```
JonathanStarup, line 205: @mlutze I kind of want to just have a cofinite set of Verifiers (easy to do all / all except X / just X) is that too weird you think? So the verifier option would just be `Coset[Verifier]`

JonathanStarup: Then, below, I wouldn't have to define `Verifiers.all`

mlutze: Too weird.

JonathanStarup: Yeah, oh well

JonathanStarup:
#10316

PR #10316, opened 2025-04-18
refactor: move `ClassVerifier` now `TypeVerifier` to test

magnus-madsen:


magnus-madsen:
@mlutze @JonathanStarup This is how I want to go forward.

Its clearly separates production code from testing code with no risk that any verifiers depend on or interfere with the production code.

magnus-madsen:
We can merge this PR and then do the rest of the verifiers and remove all the flags.

JonathanStarup (commented):
I was also thinking about that renaming, seems nice

The drawback here is that
- its currently impossible to enable verifiers for all the scala tests
- you cannot verify some arbitrary program for debugging (you have to run it programmatically)

Do you have a plan for those or do they not matter?

magnus-madsen:
>     * its currently impossible to enable verifiers for all the scala tests

The Scala tests are mostly negative tests though...

> 
>     * you cannot verify some arbitrary program for debugging (you have to run it programmatically)

Yeah, you have to install the verifier yourself. But I think that is OK. Its just one line somewhere.

magnus-madsen:
I think this is the better solution. But we can iterate on its design.

JonathanStarup:
I would just think the debugging workflow would be

1. Find bug
2. Run with verifiers
3. ..

And step two requires editting the compiler

JonathanStarup:
The listener part seems fine, but its the test folder and no flag part i think is an issue

magnus-madsen:
> I would just think the debugging workflow would be
> 
>     1. Find bug
> 
>     2. Run with verifiers
> 
>     3. ..
> 
> 
> And step two requires editting the compiler

You can run the test that fails though?

magnus-madsen:
Or add a test ...

JonathanStarup:
yeah for tests, but we cant say to users (or ourselves) to rerun the failing program with `--Xverify`, we would have to rerun it ourselves.

But okay, just wanted to make sure the drawbacks were deliberate.

mlutze (approved):

Issue #10280, opened 2025-04-10
Feature request: JAR resources support


RGBCube:
Like in Java/Kotlin/Scala a way to include resources into the JAR would be nice. Useful for configuring JVM loggers and other libraries that scan resources for global configuration.

A default path `resources/`, similar to `src/` could be used & the user could be able to change it (IF `src` is also changeable, I haven't tinkered with the Flix build system much).

The path would be copied verbatim into the JAR, which is how it is done in other build systems (gradle/maven/sbt)

magnus-madsen:
@jaschdoc I think this is something we can look into.

We just extend build-jar to check if there is a resources folder and then bundle everything in there.

PR #10324, opened 2025-04-18
feat: add support for `resources` dir in package manager

jaschdoc:
Closes https://github.com/flix/flix/issues/10280

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/Bootstrap.scala
+++ b/main/src/ca/uwaterloo/flix/api/Bootstrap.scala
@@ -156,6 +156,11 @@ object Bootstrap {
   private def getClassDirectory(p: Path): Path = getBuildDirectory(p).resolve("./class/")
 
   /**
+    * Returns the path to the artifact directory relative to the given path `p`.
+    */
+  private def getResourcesDirectory(p: Path): Path = p.resolve("./resources/").normalize()
+
+  /**
     * Returns the path to the LICENSE file relative to the given path `p`.
     */
   private def getLicenseFile(p: Path): Path = p.resolve("./LICENSE.md").normalize()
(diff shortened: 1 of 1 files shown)


main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -350,6 +363,10 @@ object Bootstrap {
       Validation.mapN(bootstrap.folderMode())(_ => bootstrap)
     }
   }
+
+  private def snd[A, B](t: (A, B)): B = t match {
```
magnus-madsen, line 367: Remove/inlines

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -73,6 +74,7 @@ object Bootstrap {
     //
     // Create the project directories and files.
     //
+    newDirectoryIfAbsent(resourcesDirectory)
```
magnus-madsen, line 77: I don't think we should require this directory. Can we just make it optional and include it if present?

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -307,6 +314,12 @@ object Bootstrap {
     }
   }
 
+  private def getAllFilesSorted(p: Path): List[(Path, String)] = {
```
magnus-madsen, line 317: Comments?

jaschdoc:
Tested again and works

RGBCube:
An example at `examples/package-manager/jar-with-resources` would be a great addition for people just browsing the repo to see what Flix is capable of

magnus-madsen (approved):

PR #10329, opened 2025-04-22
feat: improve argument completions

magnus-madsen:


magnus-madsen:
@stormckey I propose this instead of #10147. What do you think? It should be easy to tweak and extend.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/DefCompleter.scala
```diff
@@ -53,4 +58,18 @@ object DefCompleter {
...
   }
+
+  /**
+    * Returns the expression context at the given `uri` and position `pos`.
+    */
+  private def getExprContext(uri: String, pos: Position)(implicit root: Root, flix: Flix): ExprContext = {
+    val stack = LspUtil.getStack(uri, pos)
+    // The stack contains the path of expressions from the leaf to the root.
+    stack match {
+      case Expr.Error(_, _, _) :: Expr.ApplyClo(_, _, _, _, _) :: _ =>
```
mlutze, line 69: Should there be ApplyDef and ApplySig here too?

mlutze: A bit more documentation of what's going on (an example) would be nice.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/ExprContext.scala
```diff
@@ -0,0 +1,34 @@
...
+sealed trait ExprContext
+
+object ExprContext {
+
+  /**
+    * Represents an expression in an application context.
+    *
+    * For example, in `e1(e2, e3)` we have that `e1` is `UnderApply` but `e2` and `e3` are not.
+    */
+  case object UnderApply extends ExprContext
```
mlutze, line 27: The name of this is weird, no?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/DefCompleter.scala
magnus-madsen: I am leaving all of this for @chanattan

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/ExprContext.scala
magnus-madsen: Suggestions welcome, but lets leave it for @stormckey

magnus-madsen:
@mlutze This was intended as a template for @stormckey to extend. (If he agree with it)

stormckey:
I think it's quite similar to my previous PR, except three things:
- Here we don't count the number of applied arguments
- Here we don't check if the root Expr.Error is UndefinedName
- Here we don't check if the function is pipelined (|> ||>)

I think we should first determine how exact we want this context to be,(and how different contexts will lead to different completions), then we can design and impl it:)

mlutze (approved):

magnus-madsen:
- We agreed not count the number of arguments
- We agreed to only support `|>`

PR #10335, opened 2025-04-22
feat: add `OccurrenceAnalyzer` pt. 1

jaschdoc:
Adds half of the `OccurrenceAnalyzer` so we can review the documentation and structure of the less critical parts first. I will then open another PR afterwards where we can scrutinize the implementation of `visitExp` and its helper functions.


main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,243 @@
...
+  /**
+    * Performs occurrence analysis on `defn`.
+    */
+  private def visitDef(defn0: OccurrenceAst.Def): (OccurrenceAst.Def, OccurInfo) = {
+
+    /**
+      * Returns true if `expr0` is a function call.
+      */
+    def isDirectCall(expr0: OccurrenceAst.Expr): Boolean = expr0 match {
+      case OccurrenceAst.Expr.ApplyDef(_, _, _, _, _, _) => true
```
jaschdoc, line 131: This could be `ApplyClo` as well

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,243 @@
...
+    def isSelfRecursive(occurInfo: OccurInfo): Boolean = occurInfo.defs.get(defn0.sym) match {
+      case None => false
+      case Some(o) => o match {
+        case Occur.Dead => false
+        case Occur.Once => true
+        case Occur.OnceInLambda => true
+        case Occur.OnceInLocalDef => true
+        case Occur.Many => true
+        case Occur.ManyBranch => true
+        case Occur.DontInline => false
```
jaschdoc, line 147: mark as `true` anyway?

magnus-madsen:
@JonathanStarup Can you do an initial review?

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,242 @@
...
+
+/**
+  * The occurrence analyzer collects information on variable and function usage and calculates the weight of the expressions
+  * Marks a variable or function as Dead if it is not used, Once if it is used exactly once and Many otherwise
+  */
+object OccurrenceAnalyzer {
+
+  private object OccurInfo {
+
+    /**
```
JonathanStarup, line 34: imo `/** doc */` is bettter for these short ones

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,242 @@
...
+import ca.uwaterloo.flix.api.Flix
+import ca.uwaterloo.flix.language.ast.OccurrenceAst.Occur.*
+import ca.uwaterloo.flix.language.ast.OccurrenceAst.{DefContext, Occur}
+import ca.uwaterloo.flix.language.ast.Symbol.{DefnSym, VarSym}
+import ca.uwaterloo.flix.language.ast.{OccurrenceAst, Symbol}
+import ca.uwaterloo.flix.util.ParOps
+
+/**
+  * The occurrence analyzer collects information on variable and function usage and calculates the weight of the expressions
+  * Marks a variable or function as Dead if it is not used, Once if it is used exactly once and Many otherwise
```
JonathanStarup, line 28: I think these should use proper `[[Dead]]` references,

Arent there more options than these 3?

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,242 @@
...
+  */
+object OccurrenceAnalyzer {
+
+  private object OccurInfo {
+
+    /**
+      * Occurrence information for an empty sequence of expressions
+      */
+    val Empty: OccurInfo = OccurInfo(Map.empty, Map.empty, 0, 0)
```
JonathanStarup, line 37: shouldn't `Empty` and `One` be `empty` and `one` like `List.empty`

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,242 @@
...
+  }
+
+  private def increment(occurInfo: OccurInfo): OccurInfo = {
+    occurInfo.copy(size = occurInfo.size + 1)
+  }
+
+  /**
+    * A set of functions that contain masked casts.
+    *
+    * Must be manually maintained since Lowering erases the masked cast.
```
JonathanStarup, line 86: masked cast doesn't exist anymore so this can be removed I would think

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
JonathanStarup: It is not totally clear to me how these Dont objects relate to occurrence. Conceptually it should be (Occur, Danger) since they are two different things. Do you think its practically okay to separate the two? Otherwise I think the docs for DontX need to say "DontX implies that it is not Dead" fx

magnus-madsen:
I agree Jonathans comments. Will add my own soon.

Can you also add comments for Occur as we discussed? i.e. one sentence for each like:

"A dead binder can be removed if the bound expression is pure" etc.

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
magnus-madsen: Yeah, comment seems outdated.

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,242 @@
+/*
+ * Copyright 2022 Anna Krogh, Patrick Lundvig, Christian Bonde, 2025 Jakob Schneider Villumsen
```
magnus-madsen, line 2: Break into two lines

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,242 @@
...
+  /**
+    * The occurrence of `defs` and `vars` inside the body of a `def`
+    * `size` represents the number of expressions in the body of a `def`
+    */
+  case class OccurInfo(defs: Map[DefnSym, Occur], vars: Map[VarSym, Occur], localDefs: Int, size: Int) {
+    def :+(kv: (DefnSym, Occur)): OccurInfo = {
+      this.copy(defs = this.defs + kv)
+    }
+
+    def +(kv: (VarSym, Occur)): OccurInfo = {
```
magnus-madsen, line 54: I don't like that we have both :+ and +.

How much do we need?

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,242 @@
...
+
+    def get(defnSym: DefnSym): Occur = {
+      this.defs.getOrElse(defnSym, Dead)
+    }
+
+    def get(varSym: VarSym): Occur = {
+      this.vars.getOrElse(varSym, Dead)
+    }
+
+    def incrementLocalDefCount: OccurInfo = {
```
magnus-madsen, line 74: Inconsistent - member function

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,242 @@
...
+      this.copy(localDefs = localDefs + 1)
+    }
+  }
+
+  private def increment(occurInfo: OccurInfo): OccurInfo = {
```
magnus-madsen, line 79: Inconsistent - top-level function

magnus-madsen: Names are also inconsistent: incrementLocalDefCount vs increment.

Not good

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,242 @@
...
+
+    def -(varSym: VarSym): OccurInfo = {
+      this.copy(vars = this.vars - varSym)
+    }
+
+    def --(varSyms: Iterable[VarSym]): OccurInfo = {
+      this.copy(vars = this.vars -- varSyms)
+    }
+
+    def get(defnSym: DefnSym): Occur = {
```
magnus-madsen, line 66: Lets have the getters as the first functions in the class.

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,242 @@
...
+  }
+
+  /**
+    * A set of functions that contain masked casts.
+    *
+    * Must be manually maintained since Lowering erases the masked cast.
+    */
+  private val DangerousFunctions: Set[String] = Set("bug!", "Fixpoint.Debugging.notifyPreSolve", "Fixpoint.Debugging.notifyPostSolve", "Fixpoint.Debugging.notifyPreInterpret", "Assert.eq")
+
+  private def toReadableFunction(sym: Symbol.DefnSym): String = {
```
magnus-madsen, line 90: What is this?

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,242 @@
...
+    sym.toString.takeWhile(c => c.toString != Flix.Delimiter)
+  }
+
+  /**
+    * Performs occurrence analysis on the given AST `root`.
+    */
+  def run(root: OccurrenceAst.Root)(implicit flix: Flix): OccurrenceAst.Root = {
```
magnus-madsen, line 97: Usually run is the very first function

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,242 @@
...
+    val defs = visitDefs(root.defs) // visitDefs calls parMap internally and has additional handling
```
magnus-madsen, line 98: That's unusual. Is there a reason not to call parMap here?

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,242 @@
...
+  private def increment(occurInfo: OccurInfo): OccurInfo = {
+    occurInfo.copy(size = occurInfo.size + 1)
+  }
+
+  /**
+    * A set of functions that contain masked casts.
+    *
+    * Must be manually maintained since Lowering erases the masked cast.
+    */
+  private val DangerousFunctions: Set[String] = Set("bug!", "Fixpoint.Debugging.notifyPreSolve", "Fixpoint.Debugging.notifyPostSolve", "Fixpoint.Debugging.notifyPreInterpret", "Assert.eq")
```
magnus-madsen, line 88: Ugh.

What is happening here?

Is this essentially a blacklist of functions not to touch?

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,242 @@
...
+  def run(root: OccurrenceAst.Root)(implicit flix: Flix): OccurrenceAst.Root = {
+    val defs = visitDefs(root.defs) // visitDefs calls parMap internally and has additional handling
+    OccurrenceAst.Root(defs, root.enums, root.structs, root.effects, root.mainEntryPoint, root.entryPoints, root.sources)
+  }
+
+  /**
+    * Performs occurrence analysis on every entry in `defs0` in parallel.
+    * Decorates each Def with occurrence information, i.e., how it appears in the program or if it is unused.
+    */
+  private def visitDefs(defs0: Map[DefnSym, OccurrenceAst.Def])(implicit flix: Flix): Map[DefnSym, OccurrenceAst.Def] = {
```
magnus-madsen, line 106: This is legacy style. We should almost never define functions on maps or lists. Just on the individual elements. Then use map/parMap etc.

magnus-madsen (commented):
Added some comments with more to come.

magnus-madsen:
There seems to be a style issue: some times methods are members of a class and sometimes they are toplevel. 

This is inconsistent.

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
jaschdoc: This is because `+` cannot be overloaded so we need two distinct symbols

magnus-madsen: But thats terrible design no? Two different symbols?

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
jaschdoc: Yes exactly

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,274 @@
...
+
+  /**
+    * Performs occurrence analysis on `defn`.
+    */
+  private def visitDef(defn0: OccurrenceAst.Def): (OccurrenceAst.Def, ExpContext) = {
+
+    /**
+      * Returns true if `expr0` is a function call.
+      */
+    def isDirectCall(expr0: OccurrenceAst.Expr): Boolean = expr0 match {
```
magnus-madsen, line 61: Dont nest functions inside visitDef. Move them elsewhere

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
jaschdoc: This is the long discussion we had on inlining holes and casts. Since it is much harder to figure out where holes and casts are and whether they are safe to inline, we never inline functions that contain these and explicitly ban functions that contain masked casts (because they lie even more to the compiler).

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
jaschdoc: Maybe just an overloaded `add` function?

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
jaschdoc: Ah ok check this out https://github.com/flix/flix/pull/10335#discussion_r2054170912 -- masked casts are gone!

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
jaschdoc: Hmm I see what you mean. The Haskell inliner also has a `DontInline` constructor which also tells the inliner hands off. So it's more of a "whatever value you return does not matter since `DontInline` takes precedence. So if a function marked with `DontInline` gets inlined, then we have a bug. Maybe we should just ignore this (set it to `true`) and check on every inlining decision that its occurrence information is not `DontInline`. If it is, then it should crash. However, they way it is written now also enforces that without crashing.

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,240 @@
...
+      case Occur.ManyBranch => true
+      case Occur.DontInline => true
+      case Occur.DontInlineAndDontRewrite => true
+    }
+  }
+
+  /**
+    * Performs occurrence analysis on `exp0`
+    */
+  private def visitExp(exp0: OccurrenceAst.Expr)(implicit sym0: Symbol.DefnSym): (OccurrenceAst.Expr, ExpContext) = (exp0, ExpContext.empty)
```
magnus-madsen, line 92: Move this right after visitDef

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,240 @@
...
+
+  /**
+    * Combines `expCtx1` and `expCtx2` into a single [[ExpContext]].
+    */
+  private def combineInfoBranch(expCtx1: ExpContext, expCtx2: ExpContext): ExpContext = {
```
magnus-madsen, line 97: There are two types right? Can we reflect that in names:

combineSeq
combine???

magnus-madsen:
A bigger question here is important:

We want this phase to be as fast as possible-- which is not the case previously. 

If we can use more mutation we can speed it up.

The easiest part would be to compute the size parameter in a LocalContext that contains an integer. 

This would save a lot of allocation.

Can we use mutation for anything else?

magnus-madsen:
I think we can also compute the defs in a mutable LocalContext

magnus-madsen:
@jaschdoc I merged the other PR just now. So I will let you rebase and then I can review.

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,230 @@
...
+
+  /**
+    * Combines two occurrences `o1` and `o2` from distinct branches into a single occurrence.
+    */
+  private def combineBranch(o1: Occur, o2: Occur): Occur = (o1, o2) match {
+    case (Dead, _) => o2
+    case (_, Dead) => o1
+    case (Once, Once) => ManyBranch
+    case _ => Many
+  }
```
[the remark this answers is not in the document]
jaschdoc: It seems like these could be merged into the same function if we introduce an `Branch := Same | Distinct` type

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,230 @@
...
+        val defWithContext = defn.copy(context = newContext)
+        macc + (defn.sym -> defWithContext)
+    }
+    OccurrenceAst.Root(defs, root.enums, root.structs, root.effects, root.mainEntryPoint, root.entryPoints, root.sources)
+  }
+
+  /**
+    * Performs occurrence analysis on `defn`.
+    */
+  private def visitDef(defn0: OccurrenceAst.Def): (OccurrenceAst.Def, ExpContext) = {
```
magnus-madsen, line 56: defn0 -> defn

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,230 @@
...
+    val (exp, expCtx) = visitExp(defn0.exp)(defn0.sym)
```
magnus-madsen, line 57: expCtx -> ctx

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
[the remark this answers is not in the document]
jaschdoc: The decision of whether to use `combineSeq` or `combineBranch` entirely depends on which branch `o1` is in compared to `o2`.

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,230 @@
...
+
+  /**
+    * Performs occurrence analysis on `exp0`
+    */
+  private def visitExp(exp0: OccurrenceAst.Expr)(implicit sym0: Symbol.DefnSym): (OccurrenceAst.Expr, ExpContext) = (exp0, ExpContext.empty)
+
+  /**
+    * Combines `expCtx1` and `expCtx2` into a single [[ExpContext]].
+    */
+  private def combineBranch(expCtx1: ExpContext, expCtx2: ExpContext): ExpContext = {
```
magnus-madsen, line 71: expCtx1 -> ctx1
expCtx2 -> ctx2

everywhere

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,230 @@
...
+    * Returns true if `expr0` is a function call.
+    */
+  private def isDirectCall(expr0: OccurrenceAst.Expr): Boolean = expr0 match {
+    case OccurrenceAst.Expr.ApplyDef(_, _, _, _, _, _) => true
+    case OccurrenceAst.Expr.ApplyClo(_, _, _, _, _) => true
+    case _ => false
+  }
+
+  /**
+    * Returns true if `def0` occurs in `expCtx`.
```
magnus-madsen, line 217: def0 -> defn

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,230 @@
...
+    */
+  private def isSelfRecursive(expCtx: ExpContext, defn0: OccurrenceAst.Def): Boolean = expCtx.defs.get(defn0.sym) match {
```
magnus-madsen, line 219: defn0 -> defn

magnus-madsen: Here arguments are switched. Elsewhere we have def before ctx

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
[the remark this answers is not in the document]
magnus-madsen: I would leave everything as-is for now. What you propose moves to a different point in the design space, but it is not clear that it is better.

main/src/ca/uwaterloo/flix/language/phase/OccurrenceAnalyzer.scala
```diff
@@ -0,0 +1,230 @@
...
+  private def combineSeq(expCtx1: ExpContext, expCtx2: ExpContext): ExpContext = {
+    combine(expCtx1, expCtx2, combineSeq)
+  }
+
+  /**
+    * Combines `expCtx1` and `expCtx2` into a single [[ExpContext]].
+    */
+  private def combineSeqOpt(expCtx1: Option[ExpContext], expCtx2: ExpContext): ExpContext = {
+    expCtx1.map(combineSeq(_, expCtx2)).getOrElse(expCtx2)
+  }
```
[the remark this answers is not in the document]
magnus-madsen: We usually prefer free standing functions.

jaschdoc:
Let's merge it

magnus-madsen (approved):

PR #10337, opened 2025-04-22
feat: handle pipeline in expr context

stormckey:
Now we can handle pipeline and () as the def context:
image
image
image
image
image
image

Note that if there is both pipeline and (), () will dominate since it's closer to the error.

As a result, the completion in this mixed context is just what we want:
image
image

So there is no need to consider the state machine(it's only needed if we want the exact arg count in mixed context)

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
@@ -16,7 +16,6 @@
 package ca.uwaterloo.flix.api.lsp.provider.completion
 
 import ca.uwaterloo.flix.api.Flix
-import ca.uwaterloo.flix.api.lsp.provider.completion.ExprContext.UnderPipeline
 import ca.uwaterloo.flix.api.lsp.{Command, CompletionItem, CompletionItemKind, CompletionItemLabelDetails, InsertTextFormat, Position, Range, TextEdit}
 import ca.uwaterloo.flix.language.ast.shared.AnchorPosition
 import ca.uwaterloo.flix.language.ast.{Name, ResolvedAst, SourceLocation, Symbol, Type, TypedAst}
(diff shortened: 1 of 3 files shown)


stormckey:
In next PR I will handle runWith case :)

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
```diff
@@ -163,6 +164,7 @@ sealed trait Completion {
       val label = if (qualified) qualifiedName else decl.sym.name
       val snippet = ectx match {
         case ExprContext.UnderApply => CompletionUtils.getApplySnippet(label, Nil)
+        case UnderPipeline => CompletionUtils.getApplySnippet(label, decl.spec.fparams.dropRight(1))
```
magnus-madsen, line 167: This should be ExprContext.UnderPipeline

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
```diff
@@ -16,6 +16,7 @@
 package ca.uwaterloo.flix.api.lsp.provider.completion
 
 import ca.uwaterloo.flix.api.Flix
+import ca.uwaterloo.flix.api.lsp.provider.completion.ExprContext.UnderPipeline
```
magnus-madsen, line 19: Drop import

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/DefCompleter.scala
```diff
@@ -66,9 +68,12 @@ object DefCompleter {
...
     stack match {
-      case Expr.Error(_, _, _) :: Expr.ApplyClo(_, _, _, _, _) :: _ =>
-        // The leaf is an error followed by an apply expression.
+      case Expr.Error(UndefinedName(_, _, _, _), _, _) :: Expr.ApplyClo(_, _, _, _, _) :: _ =>
+        // The leaf is an error followed by an ApplyClo expression.
         ExprContext.UnderApply
+      case Expr.Error(UndefinedName(_, _, _, _), _, _) :: Expr.ApplyDef(DefSymUse(sym, _), _, _, _, _, _) :: _ if sym.text == "|>" =>
+        // The leaf is an error followed by an ApplyDef expression with the symbol "|>".
+        ExprContext.UnderPipeline
       case _ => ExprContext.Unknown
```
magnus-madsen, line 77: Matt dislikes these two names. Can we make them:

- InsideApply
- InsidePipeline

stormckey: Okay, let's make Matt happy :)

magnus-madsen (approved):

PR #10340, opened 2025-04-22
feat: include most file programs in examples for TestCompletionProvider

### LINKED ISSUE #10326: Add all `examples` to `TestCompletionProvider`
Except for `larger-examples` and `package-manager`

stormckey:
fixes #10326

All files, except ones from large-example, and package-manager are included.

The whole test takes 3 min 3 s on my computer.

Do you think we should first read and compile all the programs for further usage, that would be a huge acceleration.

### DIFF
--- a/main/test/ca/uwaterloo/flix/api/lsp/TestCompletionProvider.scala
+++ b/main/test/ca/uwaterloo/flix/api/lsp/TestCompletionProvider.scala
@@ -25,79 +25,95 @@ import ca.uwaterloo.flix.language.ast.shared.{Input, SecurityContext, Source}
 import ca.uwaterloo.flix.util.Options
 import org.scalatest.funsuite.AnyFunSuite
 
+import java.nio.file.{Files, Paths}
+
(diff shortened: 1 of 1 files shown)


magnus-madsen:
I think we need to reuse the Flix object (to have incremental compilation).

stormckey:
> I think we need to reuse the Flix object (to have incremental compilation).

Currently each invariant will not recompile a program twice, since no one is changing the program.

Later if we want to recompile the program incrementally, we just report the change to flix and call flix.check() again.

stormckey:
Well the change was supposed to accelerate the test, but it didn't.

Any ideas?

stormckey:
OK so the memory is not enough to cache all the artifacts.

magnus-madsen:
In this rare case, the Flix instance object can be a private field on the class.

stormckey:
Now it takes 5 mins on the CI, faster than fuzzing.

magnus-madsen (approved):

PR #10342, opened 2025-04-22
feat: handle runWith in expr context

stormckey:
preview:
image
image

I just choose to complete only the name, no ().

Is it possible that we need to provide a ()? I think their signature should always be `Unit -> a` (this holds for all existing runWith)

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
@@ -163,6 +163,7 @@ sealed trait Completion {
       val label = if (qualified) qualifiedName else decl.sym.name
       val snippet = ectx match {
         case ExprContext.InsideApply => CompletionUtils.getApplySnippet(label, Nil)
+        case ExprContext.InsideRunWith => CompletionUtils.getApplySnippet(label, Nil)
         case ExprContext.InsidePipeline => CompletionUtils.getApplySnippet(label, decl.spec.fparams.dropRight(1))
         case ExprContext.Unknown => CompletionUtils.getApplySnippet(label, decl.spec.fparams)
       }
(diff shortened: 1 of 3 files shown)


JonathanStarup:
Some handlers need info, like `example(x: Int32, f: Unit -> t)` so it should be `run .. with example(42)`

stormckey:
Okay, so I guess we should just drop the rightmost arg here?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
```diff
@@ -163,6 +163,7 @@ sealed trait Completion {
       val label = if (qualified) qualifiedName else decl.sym.name
       val snippet = ectx match {
         case ExprContext.InsideApply => CompletionUtils.getApplySnippet(label, Nil)
+        case ExprContext.InsideRunWith => CompletionUtils.getApplySnippet(label, decl.spec.fparams.dropRight(1))
```
magnus-madsen, line 166: This should be one line down to keep order consistent

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/DefCompleter.scala
```diff
@@ -74,6 +74,9 @@ object DefCompleter {
       case Expr.Error(UndefinedName(_, _, _, _), _, _) :: Expr.ApplyDef(DefSymUse(sym, _), _, _, _, _, _) :: _ if sym.text == "|>" =>
         // The leaf is an error followed by an ApplyDef expression with the symbol "|>".
         ExprContext.InsidePipeline
+      case Expr.Error(UndefinedName(_, _, _, _), _, _) :: Expr.RunWith(_, _, _, _, _) :: _ =>
+        // The leaf is an error followed by an ApplyDef expression.
```
magnus-madsen, line 78: ???

magnus-madsen (approved):

PR #10344, opened 2025-04-22
chore: share stack in CompletionProvider

stormckey:
More and more completers in CompletionProvider will need use the stack, while the stack is read-only and the same across all completers.

We can build the stack in CompletionProvider and pass it to any completer that needs it.

Similar things can be done for ExprContext, which will be shared across several completers. Shall we also share that?

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/LspUtil.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/LspUtil.scala
@@ -35,7 +35,7 @@ object LspUtil {
     *
     * So the stack actually contains a path from the leaf node that contains the given position to the root node, with the leaf node at the top of the stack.
     */
-  def getStack(uri: String, pos: Position)(implicit root: Root, flix: Flix): List[AnyRef] = {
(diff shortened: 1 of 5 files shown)


magnus-madsen:
I agree we should share it, but should it be the first argument?

stormckey:
> I agree we should share it, but should it be the first argument?

How about after qn?

> Also, I don't really like its type List[AnyRef]. So that makes more unsure.

We are reusing the StackConsumer and it's designed to accept any object that is accepted during the visiting.

But we can use another class, say ExprStack, that will only push expr into the stack. What do you think?

stormckey:
@magnus-madsen After merging this PR we can continue to provide the context to other completions.

magnus-madsen:
I am not sure yet whether to do this. Can we continue with the other things and then return to this?

stormckey:
> I am not sure yet whether to do this. Can we continue with the other things and then return to this?

Ok, so first every completer will have their own stack and context if needed. Then we can consider if we should share them.

PR #10346, opened 2025-04-23
fix: report shadowing in formal parameters of handlers

### LINKED ISSUE #10302: Shadowed variable isn't checked for handler fparam and pattern inside it

gaganchandan:
Fixes #10302

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Redundancy.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Redundancy.scala
@@ -691,10 +691,18 @@ object Redundancy {
       sctx.effSyms.put(sym.sym, ())
       rules.foldLeft(Used.empty) {
         case (acc, HandlerRule(_, fparams, body, _)) =>
-          val usedBody = visitExp(body, env0, rc)
           val syms = fparams.map(_.bnd.sym)
-          val dead = syms.filter(deadVarSym(_, usedBody))
-          acc ++ usedBody ++ dead.map(UnusedVarSym.apply)
+          val shadowedFparamVars = syms.map(s => shadowing(s.text, s.loc, env0))
+          val env1 = env0 ++ syms
+          val usedBody = visitExp(body, env1, rc)
+          syms.zip(shadowedFparamVars).foldLeft(acc ++ usedBody) {
+            case (acc, (s, shadow)) =>
+              if (deadVarSym(s, usedBody)) {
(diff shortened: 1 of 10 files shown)


main/src/library/FileRead.flix
```diff
@@ -126,129 +126,129 @@ mod FileRead {
...
                     case ex: IOException          => Err(IoError(ErrorKind.Other, ex.getMessage()))
                 }
             }
 
-            def list(f, k) = {
+            def list(file, k) = {
                 try {
                     // TODO: Use java.nio.file.Files.list when we have support for Java generics.
-                    let file = new File(f);
+                    let file = new File(file);
```
mlutze, line 251: This indicates to me that something is wrong. `file` shadows `file` here.

We probably need to be sure to add the fparams to the environment before visiting the body of the handler.

magnus-madsen:
@mlutze You first then me

main/src/ca/uwaterloo/flix/language/phase/Redundancy.scala
```diff
@@ -691,10 +691,18 @@ object Redundancy {
...
       rules.foldLeft(Used.empty) {
         case (acc, HandlerRule(_, fparams, body, _)) =>
-          val usedBody = visitExp(body, env0, rc)
           val syms = fparams.map(_.bnd.sym)
-          val dead = syms.filter(deadVarSym(_, usedBody))
-          acc ++ usedBody ++ dead.map(UnusedVarSym.apply)
+          val shadowedFparamVars = syms.map(s => shadowing(s.text, s.loc, env0))
+          val env1 = env0 ++ syms
+          val usedBody = visitExp(body, env1, rc)
+          syms.zip(shadowedFparamVars).foldLeft(acc ++ usedBody) {
```
mlutze, line 698: I see that this looks different from what we do with `LocalDef`. Is there a reason for the difference?

gaganchandan: Is it not more or less the same? When checking the body, the environment is extended to include the `fparams`. The accumulator for `foldLeft` contains the results of processing the `usedBody`, similar to the `foldLeft` in `LocalDef` containing the result of processing the inner expressions.

mlutze: You're right. I was confused.

magnus-madsen (approved):

PR #10355, opened 2025-04-23
feat: support Signature Help for Op

stormckey:
Preview:
image
image
image


main/src/ca/uwaterloo/flix/api/lsp/provider/SignatureHelpProvider.scala
```diff
@@ -32,6 +32,10 @@ object SignatureHelpProvider {
         mkSignatureHelp(defnSymUse.sym, root.defs(defnSymUse.sym).spec, exps, pos)
       case TypedAst.Expr.ApplySig(sigSymUse, exps, _, _, _, _) =>
         mkSignatureHelp(sigSymUse.sym, root.sigs(sigSymUse.sym).spec, exps, pos)
+      case TypedAst.Expr.Do(opSymUse, exps, _, _, _) =>
+        val ops = root.effects(opSymUse.sym.eff).ops
+        val op = ops.find(thatOp => thatOp.sym == opSymUse.sym).get
```
magnus-madsen, line 37: This .get is dangerous? Can it not crash if the effect does not exist?

magnus-madsen:
In summary, you should be returning an option from the three cases.

magnus-madsen:
Hmm okay. Can you insert 3 comments?

// The lookup is guaranteed to succeed because otherwise the expression would have been replaced by Expr.Error

Issue #10356, opened 2025-04-24
Let-bind functions that use direct recursion


magnus-madsen:
Given:

```scala
pub def last(l: List[a]): Option[a] = match l {
    case Nil      => None
    case x :: Nil => Some(x)
    case _ :: rs  => last(rs)
}
```

we should write a new compiler phase, probably after `Simplifier`, to transform this into:

```scala
pub def last(l: List[a]): Option[a] = {
    def loop$(l1: List[a])= match l1 {
        case Nil      => None
        case x1 :: Nil => Some(x1)
        case _ :: rs1   => loop$(rs1)
    };
    loop$(l)
}
```

**Proposal:** We do it in a new compiler phase after `Simplifier`.

Later we can discuss how to do it in one of the existing phases.

**Question:** How many traversals does this require? 

I think one traversal to determine that the function is directly recursive. Then one more traversal to unlift it.

magnus-madsen:
Phase name: *Unlifter*?

@jaschdoc Thoughts? I think this can be done cleanly as a new phase.

JonathanStarup:
This sounds like footwork for an inliner, why does the inliner not just convert them to local defs as it is inlining them?

jaschdoc:
It definitely needs to happen before or during inlining so it can eliminate the extra call introduced by this transformation. If we do it before inlining as a separate phase the implementation may be cleaner but increase code size by duplicating the function. If the inliner performs this rewrite when inlining a call to a recursive function, we only pay the price of duplicating the function where it would have been duplicated anyway (so no extra cost in code size). However, it may make the inliner a bit more complex.

Also, for this to really matter, we should analyze which parameters can be captured so the inliner has a chance of specializing captured parameters.

magnus-madsen:
I strongly propose we do a clean standalone implementation _first_.

Then we can discuss how to integrate it with an existing phase.

jaschdoc:
> Also, for this to really matter, we should analyze which parameters can be captured so the inliner has a chance of specializing captured parameters.

This can still be done in a standalone phase

JonathanStarup:
> I strongly propose we do a clean standalone implementation _first_.
> 
> Then we can discuss how to integrate it with an existing phase.

You often say that we should avoid temporary solutions and instead do it the right way - what makes this situation different?

An isolated well-written inliner function can be just as clean and testable - there is a cost of cross-phase logic and if its not on-demand you need heuristics / complexity in the standalone phase.

[...]

PR #10358, opened 2025-04-24
feat: add `LambdaDrop`

jaschdoc:
Adding as separate phase for now. We can always merge it with the inliner later.

Closes https://github.com/flix/flix/issues/10356


jaschdoc:
@magnus-madsen we have something that is close to being done. Still test failures and docs missing but this is the general idea.

magnus-madsen:
Is the point to identify parameters that are unchanged by the recursive call?

main/src/ca/uwaterloo/flix/language/phase/optimizer/RecursionRewriter.scala
```diff
@@ -0,0 +1,400 @@
...
+    val newDefs = ParOps.parMapValues(root.defs)(visitDef)
+    root.copy(defs = newDefs)
+  }
+
+  private def visitDef(defn: MonoAst.Def)(implicit flix: Flix): MonoAst.Def = {
+    // 1. Check that every recursive call is in tail position
+    //    Return a set of alive parameters, i.e, function parameters that are changed in a recursive call (if it is an Expr.Var with the same symbol, then it is dead).
+    implicit val ctx: LocalContext = LocalContext.mk()
+    val allRecursiveCallsInTailPos = checkTailPosition(defn.exp, tailPos = true)(defn.sym, defn.spec.fparams, ctx, flix)
+    val containsRecursiveCall = ctx.isRecursive.get()
```
jaschdoc, line 45: The `isRecursive` predicate in `ctx` could probably be derived from collecting all the self-`ApplyDef` exprs. If that list is non-empty, then at least one recursive call exists.

magnus-madsen:
@jaschdoc What should I do now? Do I look at the code or do you polish first?

jaschdoc:
> @jaschdoc What should I do now? Do I look at the code or do you polish first?

Apologies for editing your comment! I meant to quote reply:


A review now would be nice. Then I can add some tests. So maybe these points from the list above:

- Wording: alive vs. constant vs...
- Documentation
- Code quality and quirks

magnus-madsen:
I think I need documentation of all functions and data types before I can review. I need to know what they are supposed to compose so I can asses.

Also, you should only talk about constant parameters. Remember a constant parameter is _not_ dead. It may be used and typically is. So no reference  to "alive".

magnus-madsen:
I would also requires that you break visitDef into essentially two function calls and check. Try to get it down to 3-5 lines and do the work in helper functions.

jaschdoc:
Sounds good

jaschdoc:
I will add tests first to help with the refactor.

magnus-madsen:
For visitDef, it should be something like:

```
compute
if (something) original else rewrite(...)
```

Or something like that

main/src/ca/uwaterloo/flix/language/phase/optimizer/RecursionRewriter.scala
```diff
@@ -0,0 +1,400 @@
...
+            case Some((fp, _)) => List(fp)
+            case None => throw InternalCompilerException("unexpected empty head", SourceLocation.Unknown)
+          }
+        } else {
+          List.empty
+        }
+    }.toList
+  }
+
+  private def checkTailPosition(exp0: MonoAst.Expr, tailPos: Boolean)(implicit sym0: Symbol.DefnSym, fparams: List[MonoAst.FormalParam], ctx: LocalContext, flix: Flix): Boolean = exp0 match {
```
magnus-madsen, line 87: I propose:

Let us avoid Boolean blindness by introducing these types:

```
TailPosition = Tail | NonTail
Recursion= AllSelf | Other
ParamKind = Const | NonConst
```

I would the change `LocalContext` to:

```
private case class LocalContext(var recur: Recursion, selfTailCalls: mutable.ArrayBuffer[Expr.ApplyDef])
```

I would then rename `checkTailPosition` to `visitExp` with signature:

```
def visitExp(exp0: MonoAst.Expr, pos: TailPosition )(implicit sym0: Symbol.DefnSym, ctx: LocalContext, flix: Flix): Unit
```

In `selfTailCalls` I collect all tail calls to `sym0`.

If I ever see a use of `sym0` that is not in a tail call or not in an ApplyDef then I set `recur` to `Other`.

Note: I can compute constant params by looking at `selfTailCalls`. I can then create a `List[(FormalParam, ParamKind)]` and pass that around.

JonathanStarup: > TailPosition = Tail | NonTail

We already have this type for the TailPos phase, called ExpPosition - maybe its reusable

jaschdoc: One suggestion would be to move the `TailPos` phase earlier (to `RecursionRewriter` right before `Optimizer) and merge it with `RecursionRewriter` / this. Then ApplySelfTail expressions would have to be intrdouced earlier too and extend `RunWith` with an `ExpPosition` field as in `ReducedAst`.

main/src/ca/uwaterloo/flix/language/phase/optimizer/RecursionRewriter.scala
```diff
@@ -0,0 +1,454 @@
...
+  /**
+    * Returns a [[Expr.LocalDef]] that is immediately applied after its declaration. Only the parameters that are alive are declared and applied.
+    * Dead parameters are captured from the containing function.
+    *
+    * @param body   The body of the local def to be created.
+    * @param subst  The variable substitution.
+    * @param params The list of formal parameters and their [[ParameterKind]].
+    */
+  private def mkLocalDefExpr(body: Expr)(implicit subst: Subst, params: List[(MonoAst.FormalParam, ParameterKind)]): Expr = {
+    // Make ApplyLocalDef Expr
```
magnus-madsen, line 381: Useless comment.

main/src/ca/uwaterloo/flix/language/phase/optimizer/RecursionRewriter.scala
```diff
@@ -0,0 +1,454 @@
...
+        }
+    }.toList
+  }
+
+  private def mkFreshLocalDefSym(defn: MonoAst.Def)(implicit flix: Flix): Symbol.VarSym = {
+    Symbol.freshVarSym(defn.sym.text + "Loop", BoundBy.LocalDef, defn.sym.loc)(Scope.Top, flix)
+  }
+
+  /**
+    * Returns a [[Expr.LocalDef]] that is immediately applied after its declaration. Only the parameters that are alive are declared and applied.
```
magnus-madsen, line 373: alive?/dead?

main/src/ca/uwaterloo/flix/language/phase/optimizer/RecursionRewriter.scala
```diff
@@ -0,0 +1,454 @@
...
+    * Dead parameters are captured from the containing function.
+    *
```
magnus-madsen, line 375: I think an example here would be good.

main/src/ca/uwaterloo/flix/language/phase/optimizer/RecursionRewriter.scala
```diff
@@ -0,0 +1,454 @@
...
+    * @param body   The body of the local def to be created.
+    * @param subst  The variable substitution.
+    * @param params The list of formal parameters and their [[ParameterKind]].
+    */
+  private def mkLocalDefExpr(body: Expr)(implicit subst: Subst, params: List[(MonoAst.FormalParam, ParameterKind)]): Expr = {
```
magnus-madsen, line 380: body -> exp0

main/src/ca/uwaterloo/flix/language/phase/optimizer/RecursionRewriter.scala
```diff
@@ -0,0 +1,454 @@
...
+        }
+        invocations.headOption match {
+          case Some((fp, _)) if allConstant => (fp, ParameterKind.Constant)
+          case Some((fp, _)) => (fp, ParameterKind.NonConstant)
+          case None => throw InternalCompilerException("unexpected empty head", SourceLocation.Unknown)
+        }
+    }.toList
+  }
+
+  private def mkFreshLocalDefSym(defn: MonoAst.Def)(implicit flix: Flix): Symbol.VarSym = {
```
magnus-madsen, line 368: docs

main/src/ca/uwaterloo/flix/language/phase/optimizer/RecursionRewriter.scala
```diff
@@ -0,0 +1,454 @@
...
+    Symbol.freshVarSym(defn.sym.text + "Loop", BoundBy.LocalDef, defn.sym.loc)(Scope.Top, flix)
```
magnus-madsen, line 369: lets make it + "$loop$

main/src/ca/uwaterloo/flix/language/phase/optimizer/RecursionRewriter.scala
```diff
@@ -0,0 +1,454 @@
...
+
+    case Expr.VectorLoad(exp1, exp2, _, _, _) =>
+      val t1 = visitExp(exp1, tailPos = TailPosition.NonTail)
+      val t2 = visitExp(exp2, tailPos = TailPosition.NonTail)
+      TailPosition.combine(t1, t2)
+
+    case Expr.VectorLength(exp, _) =>
+      visitExp(exp, tailPos = TailPosition.NonTail)
+
+    case Expr.Ascribe(exp, _, _, _) => // TODO: Is this erased???
```
magnus-madsen, line 162: What AST is this?

main/src/ca/uwaterloo/flix/language/phase/optimizer/RecursionRewriter.scala
```diff
@@ -0,0 +1,454 @@
...
+  }
+
+  private def visitDef(defn: MonoAst.Def)(implicit flix: Flix): MonoAst.Def = {
+    implicit val ctx: LocalContext = LocalContext.mk()
+    if (isRewritable(defn))
+      rewriteDefn(defn)
+    else
+      defn
+  }
+
```
magnus-madsen, line 72: Reorder a bit so:

isRewritable is here
followed by rewriteDefn

main/src/ca/uwaterloo/flix/language/phase/optimizer/RecursionRewriter.scala
```diff
@@ -0,0 +1,454 @@
...
+import ca.uwaterloo.flix.language.ast.shared.{BoundBy, Scope}
+import ca.uwaterloo.flix.language.ast.{MonoAst, SourceLocation, Symbol, Type}
+import ca.uwaterloo.flix.language.dbg.AstPrinter.DebugMonoAst
+import ca.uwaterloo.flix.util.{InternalCompilerException, ParOps}
+
+import scala.collection.mutable
+
+/**
+  * Rewrites functions that recursively call themselves in tail-position to
+  * non-recursive functions with a recursive local def that capture constant
```
magnus-madsen, line 30: -> where constant parameters become closure captured. (or something like this)

main/src/ca/uwaterloo/flix/language/phase/optimizer/RecursionRewriter.scala
```diff
@@ -0,0 +1,454 @@
...
+  * parameters.
+  *
+  * A constant parameter is a parameter that is never updated or changed in
```
magnus-madsen, line 33: A constant parameter is one that has the same value in every recursive call.

main/src/ca/uwaterloo/flix/language/phase/optimizer/RecursionRewriter.scala
```diff
@@ -0,0 +1,454 @@
...
+  *
+  * {{{
+  *   def lastMap(f: a -> b, l: List[a]): Option[b] = match l {
+  *       case Nil      => None
+  *       case x :: Nil => Some(f(x))
+  *       case _ :: xs  => lastMap(f, xs)
+  *   }
+  * }}}
+  *
+  * Will be transformed to
```
magnus-madsen, line 46: is rewritten to

main/src/ca/uwaterloo/flix/language/phase/optimizer/RecursionRewriter.scala
```diff
@@ -0,0 +1,454 @@
...
+  *
+  * {{{
+  *   def lastMap(f: a -> b, l: List[a]): Option[b] =
+  *       def lastMapLoop(l1) = match l1 {
```
magnus-madsen, line 50: lastMap$loop

main/src/ca/uwaterloo/flix/language/phase/optimizer/RecursionRewriter.scala
```diff
@@ -0,0 +1,454 @@
...
+    case Expr.IfThenElse(exp1, exp2, exp3, _, _, _) =>
+      val t1 = visitExp(exp1, tailPos = TailPosition.NonTail)
+      val t2 = visitExp(exp2, tailPos)
+      val t3 = visitExp(exp3, tailPos)
+      TailPosition.combine(TailPosition.combine(t1, t2), t3)
+
+    case Expr.Stm(exp1, exp2, _, _, _) =>
+      val t1 = visitExp(exp1, tailPos = TailPosition.NonTail)
+      val t2 = visitExp(exp2, tailPos)
+      TailPosition.combine(t1, t2)
```
magnus-madsen, line 133: See my comment at the end. This kind of abstraction is non-nonsensical because if exp1 is not in tail position how can t1 be? And if t1 cannot be then why call combine? All these questions arise because we are trying to be too smart.

magnus-madsen (commented):
I added a lot of minor comments to give you something to chew on.

My biggest comment concerns `visitExp`. I don't think its there yet.. but the new version is much improved and gave me an idea. All the option stuff is just too bothersome and unclear.

I propose we change the signature to:

```
def visitExp(exp0: MonoAst.Expr, tailPos: TailPosition)(implicit sym0: Symbol.DefnSym, ctx: LocalContext, flix: Flix): Unit
```

i.e. dropping fparams (are they even used) and changing the return type to Unit.

We then change LocalCtx to:

```
private case class LocalContext(selfCalls: mutable.ArrayBuffer[(Expr.ApplyDef, TailPosition)])
```

This is conceptually a much cleaner thing to compute in `visitExp`.

Moreover, the relevant information can be now be extracted from LocalContext in a way that makes clear what is going on.

main/src/ca/uwaterloo/flix/language/phase/optimizer/RecursionRewriter.scala
magnus-madsen: (Comment at the top actually)

magnus-madsen:
(Perhaps it is more important to progress on the inliner PR before this PR.)

main/src/ca/uwaterloo/flix/language/phase/optimizer/RecursionRewriter.scala
jaschdoc: MonoAst

main/src/ca/uwaterloo/flix/language/phase/optimizer/RecursionRewriter.scala
```diff
@@ -0,0 +1,454 @@
...
+  }
+
+  private case class LocalContext(selfTailCalls: mutable.ArrayBuffer[Expr.ApplyDef])
+
+  private object Subst {
+    def from(old: Symbol.DefnSym, fresh: Symbol.VarSym, vars: Map[Symbol.VarSym, Symbol.VarSym]): Subst = Subst(old, fresh, vars)
+  }
+
+  private case class Subst(old: Symbol.DefnSym, fresh: Symbol.VarSym, vars: Map[Symbol.VarSym, Symbol.VarSym]) {
+    def apply(sym: Symbol.DefnSym): Option[Symbol.VarSym] = {
```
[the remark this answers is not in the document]
jaschdoc: This is used to apply to a `DefnSym` in an `ApplyDef` expr. If the result is `None` then the `ApplyDef` expr is for a different function and if the result is `Some(varSym)` then it is a recursive call that we should rewrite to an `ApplyLocalDef` expr with symbol `varSym`.

main/src/ca/uwaterloo/flix/language/phase/optimizer/RecursionRewriter.scala
magnus-madsen: Neither Ascribe nor Cast is unused... so I am not sure what you mean

magnus-madsen: This is confirmed by throwing ICE and running tests

jaschdoc: > Neither Ascribe nor Cast is unused... so I am not sure what you mean

Used or unused?

magnus-madsen: They occur in real programs

jaschdoc: What I mean by erasure is that it is erased later so at runtime the expression would be in tailposition. Ascribe and Cast only carry type information the compiler needs but should not really have an impact at runtime.

magnus-madsen: No, don't mess with this. Don't make such assumptions.

jaschdoc: Does not matter anymore

jaschdoc:
@magnus-madsen RFR

main/src/ca/uwaterloo/flix/language/phase/optimizer/LambdaDrop.scala
```diff
@@ -0,0 +1,407 @@
...
+  private def visitDef(defn: MonoAst.Def)(implicit flix: Flix): MonoAst.Def = {
+    implicit val ctx: LocalContext = LocalContext.mk()
+    if (isRewritable(defn))
+      rewriteDefn(defn)
+    else
+      defn
+  }
+
+  private def isRewritable(defn: MonoAst.Def)(implicit ctx: LocalContext): Boolean = {
+    if (isHigherOrder(defn.spec.fparams)) { // Only visit exp if it is higher-order
```
magnus-madsen, line 70: Comment seems useless.

main/src/ca/uwaterloo/flix/language/phase/optimizer/LambdaDrop.scala
```diff
@@ -0,0 +1,407 @@
...
+      visitExp(defn.exp)(defn.sym, ctx)
+      val containsRecursiveCall = ctx.recursiveCalls.nonEmpty
```
magnus-madsen, line 72: No need for local variable.

main/src/ca/uwaterloo/flix/language/phase/optimizer/LambdaDrop.scala
```diff
@@ -0,0 +1,407 @@
...
+
+  private def visitDef(defn: MonoAst.Def)(implicit flix: Flix): MonoAst.Def = {
+    implicit val ctx: LocalContext = LocalContext.mk()
+    if (isRewritable(defn))
+      rewriteDefn(defn)
+    else
+      defn
+  }
+
+  private def isRewritable(defn: MonoAst.Def)(implicit ctx: LocalContext): Boolean = {
```
magnus-madsen, line 69: I think a  doc comment that says we only LambdaDrop recursive higher-order functions.

main/src/ca/uwaterloo/flix/language/phase/optimizer/LambdaDrop.scala
```diff
@@ -0,0 +1,407 @@
...
+    if (isHigherOrder(defn.spec.fparams)) { // Only visit exp if it is higher-order
+      visitExp(defn.exp)(defn.sym, ctx)
+      val containsRecursiveCall = ctx.recursiveCalls.nonEmpty
+      containsRecursiveCall
+    } else {
+      false
+    }
+  }
+
+  private def isHigherOrder(fparams: List[MonoAst.FormalParam]): Boolean = {
```
magnus-madsen, line 79: Returns `true` if at least one formal parameter is a function argument.

main/src/ca/uwaterloo/flix/language/phase/optimizer/LambdaDrop.scala
```diff
@@ -0,0 +1,407 @@
...
+        fp.tpe.typeConstructor match {
+          case Some(TypeConstructor.Arrow(_)) => true
+          case Some(TypeConstructor.ArrowWithoutEffect(_)) => true
+          case Some(_) => true
+          case None => false
+        }
+    }
+  }
+
+  private def rewriteDefn(defn: MonoAst.Def)(implicit ctx: LocalContext, flix: Flix): MonoAst.Def = {
```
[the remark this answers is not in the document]
magnus-madsen: rename to lambdaDrop?

main/src/ca/uwaterloo/flix/language/phase/optimizer/LambdaDrop.scala
```diff
@@ -0,0 +1,407 @@
...
+      exps.foreach(visitExp)
+
+    case Expr.ApplyClo(exp1, exp2, _, _, _) =>
+      visitExp(exp1)
+      visitExp(exp2)
+
+    case Expr.ApplyDef(sym, exps, _, _, _, _) =>
+      // Check for recursion
+      if (sym == sym0) {
+        val applyExp = exp0.asInstanceOf[Expr.ApplyDef]
```
magnus-madsen, line 117: instead you can bind with `expr@Expr.ApplyDef`

main/src/ca/uwaterloo/flix/language/phase/optimizer/LambdaDrop.scala
```diff
@@ -0,0 +1,407 @@
...
+    * Returns a tuple with the new local def symbol and a substitution for every variable in `defn`.
+    */
+  private def mkSubst(defn: MonoAst.Def, params: List[(MonoAst.FormalParam, ParamKind)])(implicit flix: Flix): (Symbol.VarSym, Substitution) = {
+    val nonConstantParams = params.filter { case (_, pkind) => pkind == ParamKind.NonConst }
+    val substMap = nonConstantParams.map { case (fp, _) => fp.sym -> Symbol.freshVarSym(fp.sym) }.toMap
+    val freshLocalDefSym = mkFreshLocalDefSym(defn)
+    (freshLocalDefSym, Substitution(substMap))
+  }
+
+  private def paramKinds(calls: Iterable[Expr.ApplyDef], fparams: Iterable[MonoAst.FormalParam]): List[(MonoAst.FormalParam, ParamKind)] = {
```
magnus-madsen, line 326: Docs

magnus-madsen: I would prefer list instead of iterable

main/src/ca/uwaterloo/flix/language/phase/optimizer/LambdaDrop.scala
```diff
@@ -0,0 +1,407 @@
...
+          case Some((fp, _)) if allConstant => (fp, ParamKind.Const)
+          case Some((fp, _)) => (fp, ParamKind.NonConst)
+          case None => throw InternalCompilerException("unexpected empty head", SourceLocation.Unknown)
+        }
+    }.toList
+  }
+
+  /** Returns a fresh [[Symbol.VarSym]] for a local def. */
+  private def mkFreshLocalDefSym(defn: MonoAst.Def)(implicit flix: Flix): Symbol.VarSym = {
+    val text = defn.sym.text + Flix.Delimiter + "Loop"
```
magnus-madsen, line 344: Loop -> loop

magnus-madsen:
Link me functions that are lambda dropped :)

main/src/ca/uwaterloo/flix/language/phase/optimizer/LambdaDrop.scala
```diff
@@ -0,0 +1,407 @@
...
+
+  private def rewriteDefn(defn: MonoAst.Def)(implicit ctx: LocalContext, flix: Flix): MonoAst.Def = {
+    implicit val params: List[(MonoAst.FormalParam, ParamKind)] = paramKinds(ctx.recursiveCalls, defn.spec.fparams)
+    implicit val (newDefnSym, subst): (Symbol.VarSym, Substitution) = mkSubst(defn, params)
+    val rewrittenExp = rewriteExp(defn.exp)(defn.sym, newDefnSym, subst, params)
+    val body = mkLocalDefExpr(rewrittenExp, newDefnSym)
+    defn.copy(exp = body)
+  }
+
+  private def visitExp(exp0: MonoAst.Expr)(implicit sym0: Symbol.DefnSym, ctx: LocalContext): Unit = exp0 match {
```
magnus-madsen, line 99: I think we need docs here to, to clarify that this fills in ctx.

magnus-madsen, line 99: ctx -> lctx by convention

main/src/ca/uwaterloo/flix/language/phase/optimizer/LambdaDrop.scala
```diff
@@ -0,0 +1,407 @@
...
+      Expr.ApplyAtomic(op, es, tpe, eff, loc)
+
+    case Expr.ApplyClo(exp1, exp2, tpe, eff, loc) =>
+      val e1 = rewriteExp(exp1)
+      val e2 = rewriteExp(exp2)
+      Expr.ApplyClo(e1, e2, tpe, eff, loc)
+
+    case Expr.ApplyDef(sym, exps, itpe, tpe, eff, loc) =>
+      if (sym == oldDefnSym) {
+        val es = exps.zip(fparams0).filter { // Exclude constant parameters
```
magnus-madsen, line 209: // Rewrite call to new function symbol and drop constant parameters.

main/src/ca/uwaterloo/flix/language/phase/optimizer/LambdaDrop.scala
```diff
@@ -0,0 +1,407 @@
...
+          case (_, (_, pkind)) => pkind == ParamKind.NonConst
+        }.map {
+          case (e, (_, _)) => rewriteExp(e)
+        }
+        Expr.ApplyLocalDef(newDefnSym, es, tpe, eff, loc)
+      } else {
+        val es = exps.map(rewriteExp)
```
magnus-madsen, line 216: // Preserve call as-is.

main/src/ca/uwaterloo/flix/language/phase/optimizer/LambdaDrop.scala
```diff
@@ -0,0 +1,407 @@
...
+    matrix.map {
+      case invocations =>
+        val allConstant = invocations.forall {
+          case (fp, Expr.Var(sym, _, _)) => fp.sym == sym
+          case _ => false
+        }
+        invocations.headOption match {
+          case Some((fp, _)) if allConstant => (fp, ParamKind.Const)
+          case Some((fp, _)) => (fp, ParamKind.NonConst)
+          case None => throw InternalCompilerException("unexpected empty head", SourceLocation.Unknown)
```
magnus-madsen, line 337: We should check with a recursive functions that takes zero arguments and calls it self.

Or one where all args are constant?

main/src/ca/uwaterloo/flix/language/phase/optimizer/LambdaDrop.scala
```diff
@@ -0,0 +1,407 @@
...
+      rules.foreach(rule => visitExp(rule.exp))
+
+    case Expr.Do(_, exps, _, _, _) =>
+      exps.foreach(visitExp)
+
+    case Expr.NewObject(_, _, _, _, methods, _) =>
+      methods.foreach(m => visitExp(m.exp))
+  }
+
+  private def rewriteExp(expr0: MonoAst.Expr)(implicit oldDefnSym: Symbol.DefnSym, newDefnSym: Symbol.VarSym, subst: Substitution, fparams0: List[(MonoAst.FormalParam, ParamKind)]): MonoAst.Expr = expr0 match {
```
magnus-madsen, line 187: docs

magnus-madsen:
When you write the docs, try to either write something that captures the specification (so that e.g. an LLM conceptually could fill it out) or write something with an example or write something with an invariant.

main/src/ca/uwaterloo/flix/language/phase/optimizer/LambdaDrop.scala
jaschdoc: You have previously said no to this, hence the cast ;)

magnus-madsen:
@jaschdoc Note: Small conflict

main/src/ca/uwaterloo/flix/language/phase/optimizer/LambdaDrop.scala
```diff
@@ -0,0 +1,463 @@
...
+    defn.spec.fparams.exists {
+      fp =>
+        fp.tpe.typeConstructor match {
+          case Some(TypeConstructor.Arrow(_)) => true
+          case Some(TypeConstructor.ArrowWithoutEffect(_)) => true
+          case Some(_) => true
+          case None => false
+        }
+    }
+  }
```
jaschdoc, line 98: @magnus-madsen 
The problem here is that I'm only checking for `Arrow` types. How do we ensure that it is a function type and not a type with a type param like `List`?

main/src/ca/uwaterloo/flix/language/phase/optimizer/LambdaDrop.scala
jaschdoc: Consider this program

```
def main(): Unit =
    loopsForeverNoParams()

def loopsForeverNoParams(): Unit = loopsForeverNoParams()
```
After LambdaDrop:
```
def loopsForeverNoParams(_unit$154822: Unit): Unit = {
    local def loopsForeverNoParams$loop$160865(_unit$160864: Unit): Unit = {
        loopsForeverNoParams$loop$160865(())
    };
    loopsForeverNoParams$loop$160865(_unit$154822)
}

def main(_unit$155809: Unit): Unit = {loopsForeverNoParams(())}
```


Similarly

```
def main(): Unit =
    loopsForeverWithParams((), (), ())

def loopsForeverWithParams(a: Unit, b: Unit, c: Unit): Unit = loopsForeverWithParams(a, b, c)
```
After LambdaDrop:
```
def loopsForeverWithParams(
    a$155937: Unit,
    b$155938: Unit,
    c$155939: Unit
): Unit = {
    local def loopsForeverWithParams$loop$161068(): Unit = {
        loopsForeverWithParams$loop$161068()
    };
    loopsForeverWithParams$loop$161068()
}
```

jaschdoc: So in short, if a functions has all constant parameters it still works. If it has no parameters, the compiler inserts a unit argument which becomes constant. I think maybe @JonathanStarup talked about removing the synthetic `_unit` parameter at one point.

main/src/ca/uwaterloo/flix/language/phase/optimizer/LambdaDrop.scala
jaschdoc: Ah wait no that's `Type.Apply`, right? So what could be the issue here?

jaschdoc: Okay wait there is a `true` in the wildcard case... what a typo

jaschdoc:
@magnus-madsen Okay all ready now. We have docs, tests and I checked that it only applies the transformation on higher-order functions (see https://github.com/flix/flix/pull/10358#discussion_r2066931266 )

magnus-madsen (approved):

PR #10359, opened 2025-04-24
feat: redundancy checking for `unsafely .. run ..`

JonathanStarup:
This is adding nice errors for the common case

We should additionally add a general check, that checks that `ef1 ⊆ ef2` as a fallback, which will give an ugly error.

```
eff Something
pub def f(): Int32 = unsafely {IO, Something} run println("hey")
```

```
-- Redundancy Error --------------------------------------------------

>> Redundant unsafe run, the expression does not have effect 'Something'.

2 | pub def f(): Int32 = unsafely {IO, Something} run println("hey")
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                         redundant unsafe run.
```

This unblocks  #9517


JonathanStarup:
@magnus-madsen do you not like this or just closed it since its old? I think its a good addition

PR #10361, opened 2025-04-24
fix: pass RunWith down the compiler when there is only run

### LINKED ISSUE #10341: No completion inside incomplete run with
Image

If we don't type with, expression inside run will be ignored and no completion is provided.

Also maybe we should provide a snippet for run with? Currently it's not there:

Image

stormckey:
fixes #10341

preview:
image


magnus-madsen:
Is it the same for try?

stormckey:
> Is it the same for try?

same, will fix that too.

JonathanStarup (approved):

magnus-madsen (approved):

PR #10362, opened 2025-04-24
fix: pass TryCatch down the compiler when there is only try

stormckey:
Here we want to fix the problem the way we fix RunWith.

But Expr.TryCatch are expecting `List[CatchRule]`, we don't have something like `CatchRuleError` can use here :(

I think this also applies to all other non-expr constructs.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
@@ -1753,7 +1753,7 @@ object Weeder2 {
         traverse(maybeCatch)(visitTryCatchBody),
       ) {
         // Bad case: try expr
-        case (_, Nil) =>
+        case (expr, Nil) =>
           // Fall back on Expr.Error
           val error = UnexpectedToken(
             expected = NamedTokenSet.FromKinds(Set(TokenKind.KeywordCatch, TokenKind.KeywordWith)),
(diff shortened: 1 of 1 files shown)


stormckey:
By the way, I try to pass a Nil here, the compiler crashes.

magnus-madsen:
@JonathanStarup Suggestions?

JonathanStarup:
Hmm. I mean in some sense we could just return `e` for `try e` since it works out typing wise and everything

stormckey:
> > Hmm. I mean in some sense we could just return `e` for `try e` since it works out typing wise and everything. Only other alternative is to support an empty list all the way through.
> 
> I think supporting an empty list is best. Does it require backend changes?
> 
> @stormckey Can we do that in a separate PR?

The compiler crashed on TypeREconstruction:
```scala
    case KindedAst.Expr.TryCatch(exp, rules, loc) =>
      val e = visitExp(exp)
      val rs = rules map {
        case KindedAst.CatchRule(sym, clazz, body, loc) =>
          val b = visitExp(body)
          val bnd = TypedAst.Binder(sym, Type.mkNative(clazz, SourceLocation.Unknown))
          TypedAst.CatchRule(bnd, clazz, b, loc)
      }
      val tpe = rs.head.exp.tpe
      val eff = Type.mkUnion(e.eff :: rs.map(_.exp.eff), loc)
      TypedAst.Expr.TryCatch(e, rs, tpe, eff, loc)
```

note the calling to head over there.

stormckey:
beep

magnus-madsen:
@JonathanStarup

JonathanStarup:
> > I think supporting an empty list is best. Does it require backend changes?
> 
> It means that we have to give typing meaning to empty tries. We haven't previously done this. The equivalent would be to have `If` be `Expr, Expr, Option[Expr]` and "allow" that. I don't know if we want to go into that kind of thing?

Do you think we should head into this direction? This means that we have to define typing rules for "wrong" expressions

JonathanStarup:
We haven't really answered the overarching issue - how do we represent malformed expressions where Exp.Error doesn't suffice?

PR #10372, opened 2025-04-25
fix: partial application in `Resolver`

### LINKED ISSUE #10367: Partial function application inconsistency
Partially applying a def (local or not) behaves differently than a curried lambda. leaking the fact that they are uncurried internally.

```scala
def defAdd(x: Int32, y: Int32): Int32 = x + y

def main(): Unit \ IO = {
    def localDefAdd(x, y) = x + y;
    let letAdd = (x, y) -> x + y;

    println("partial def");
    let defInc = defAdd(1 !> println);
    println("partial local def");
    let localDefInc = localDefAdd(1 !> println);
    println("partial let");
    let letInc = letAdd(1 !> println);

    println("def");
    println(defInc(1) + defInc(2));
    println("local def");
    println(localDefInc(1) + localDefInc(2));
    println("let");
    println(letInc(1) + letInc(2))
}
```
```
partial def
partial local def
partial let
1
def
1
1
5
local def
1
1
5
let
5
```

JonathanStarup:
now `add(42)` becomes `let x = 42; y -> add(x, y)`

Fixes #10367


magnus-madsen:
Performance implications?

magnus-madsen:
Or some other thing

JonathanStarup:
I just mean we have to merge it even if its slow so it feels a bit redundant

magnus-madsen:
Yeah, but I need to know the impact anyway

JonathanStarup:
`Xperf --n 11` my laptop
```
PR ~~~~ Flix Compiler Performance ~~~~

Throughput (best): 16.763 lines/sec (with 8 threads.)

  min: 13.453, max: 16.763, avg: 15.377, median: 15.877

Finished 11 iterations on 66.667 lines of code in 47 seconds.

Master ~~~~ Flix Compiler Performance ~~~~

Throughput (best): 15.714 lines/sec (with 8 threads.)

  min: 13.074, max: 15.714, avg: 14.804, median: 15.239

Finished 11 iterations on 66.667 lines of code in 49 seconds.
```

magnus-madsen:
I 100% want this, but I need to rerun my paper experiments.

mlutze (approved):

magnus-madsen:
Experiments show this is expensive. We may want to special case so that we introduce few binders.

I propose that we do not let-bind if we know the expression is a variable or constant (those are probably the common cases).

JonathanStarup:
How expensive? just curious

magnus-madsen:
> Fixed now
> 
> How expensive? just curious

I cannot say for sure, maybe 10%, but the frontend has also changed in other ways since my last numbers.

JonathanStarup:
You want me to rerun Xperf? any other experiments?

magnus-madsen:
No, just wait for me to rerun and review.

magnus-madsen:
Performance is OK now

magnus-madsen:
For the code, can we please have some helper functions?

And can we avoid Either?

JonathanStarup:
@magnus-madsen I gave it a go

magnus-madsen (approved):

PR #10380, opened 2025-04-25
refactor: tweak performance of `OccurrenceAnalyzer`

magnus-madsen:
Gain a bit more performance.


main/src/ca/uwaterloo/flix/language/phase/optimizer/Optimizer.scala
```diff
@@ -13,7 +13,7 @@ object Optimizer {
     */
   def run(root: MonoAst.Root)(implicit flix: Flix): MonoAst.Root = flix.phase("Optimizer") {
     var result = ToOccurrenceAst.run(root)
-    for (_ <- 1 to 20) {
+    for (_ <- 1 to 10) {
```
magnus-madsen, line 16: Bring this down to something more comparable to the old stuff.

jaschdoc (approved):

PR #10391, opened 2025-04-28
feat: add no-duplication completion invariant

stormckey:
related to #10138

Let's see the result of the CI first

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/CompletionItem.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/CompletionItem.scala
@@ -89,15 +89,4 @@ case class CompletionItem(
     ci.setCommitCharacters(commitCharacters.asJava)
     ci
   }
-
-  /**
-    * Two completion items are the same if all the immediately visible fields are the same.
-    */
-  override def equals(obj: Any): Boolean = obj match {
-    case that: CompletionItem =>
-      this.label == that.label &&
-      this.labelDetails == that.labelDetails &&
-      this.kind == that.kind
-    case _ => false
-  }
 }
--- a/main/src/ca/uwaterloo/flix/api/lsp/CompletionItem.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/CompletionItem.scala
@@ -89,4 +89,15 @@ case class CompletionItem(
     ci.setCommitCharacters(commitCharacters.asJava)
     ci
   }
+
+  /**
+    * Two completion items are the same if all the immediately visible fields are the same.
+    */
+  override def equals(obj: Any): Boolean = obj match {
+    case that: CompletionItem =>
+      this.label == that.label &&
+      this.labelDetails == that.labelDetails &&
+      this.kind == that.kind
+    case _ => false
+  }
 }
--- a/main/src/ca/uwaterloo/flix/language/ast/SourcePosition.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/SourcePosition.scala
@@ -86,6 +86,17 @@ object SourcePosition {
 /** Represents a physical position inside a source. */
 final class SourcePosition private(val source: Source, val lineOneIndexed: Int, val colOneIndexed: Short) {
 
+  /**
+    * The absolute character offset into `source`, zero-indexed.
+    */
+  def offset: Int = {
+    var offset = 0
+    for (i <- 1 until lineOneIndexed) {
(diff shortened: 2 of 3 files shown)


stormckey:
Yes this is the first invariant that is violated for our completer.

The equals method for CompletionItem has not been overwritten, so an duplicated completions means every field of these two completion item is the same.

When there is a duplication, the assertion will fail, and we can see:
- The whole program we are testing.
- The Expr.Var we are using to test completion
- And the label of the duplicated completion item.

I am not sure how many bugs we need to fix to pass the test.

stormckey:
As for the failure case in the CI, we choose the variable "unsafe1", after deleting one char, it becomes "unsafe" which is a keyword, and thus confuse the compiler.

I think we should skip cases where the altered Var becomes a keyword?

main/src/ca/uwaterloo/flix/language/ast/SourcePosition.scala
```diff
@@ -86,6 +86,17 @@ object SourcePosition {
 /** Represents a physical position inside a source. */
 final class SourcePosition private(val source: Source, val lineOneIndexed: Int, val colOneIndexed: Short) {
 
+  /**
+    * The absolute character offset into `source`, zero-indexed.
+    */
+  def offset: Int = {
```
magnus-madsen, line 92: This should stay in the test since it is only used there.

stormckey:
Now we can pass the invariant test. But we still know that there are some duplicated completions, like the one for companion modules (now on #9994).

I think we still need to enhance the testing in two aspects:
- What should considered duplicated completions? Currently that checking is very strict.
- We could find more occurrence where we can trigger auto complete other than just Expr.Var.

main/test/ca/uwaterloo/flix/api/lsp/TestCompletionProvider.scala
```diff
@@ -262,6 +264,107 @@ class TestCompletionProvider extends AnyFunSuite {
...
+
+  /**
+    * Trims the given program string after the given location `loc` by `n` characters.
+    */
+  private def trimAfter(program: String, loc: SourceLocation, n: Int): String = {
+    val target = program.substring(calcOffset(loc.sp1), calcOffset(loc.sp2))
+    program.substring(0, calcOffset(loc.sp1)) + target.dropRight(n) + program.substring(calcOffset(loc.sp2))
+  }
+
+  private val keywords = Set(
```
stormckey, line 320: The list here is not exhaustive. Ideally we should be able to derive an exhaustive list from the lexer or something.

But it works, since we can only testing over examples, which is finite. So we just need to cover the cases in examples. We can tell from the test result that we did.

magnus-madsen: Or just bail if the program fails to parse?

main/src/ca/uwaterloo/flix/api/lsp/CompletionItem.scala
```diff
@@ -89,4 +89,15 @@ case class CompletionItem(
     ci.setCommitCharacters(commitCharacters.asJava)
     ci
   }
+
+  /**
+    * Two completion items are the same if all the immediately visible fields are the same.
+    */
+  override def equals(obj: Any): Boolean = obj match {
```
magnus-madsen, line 96: Move this to the test suite too and have it take two arguments.

magnus-madsen: (Sidecomment: it is always wrong to implement equals and not hashCode)

main/test/ca/uwaterloo/flix/api/lsp/TestCompletionProvider.scala
magnus-madsen: I think you can just run the lexer on the appropriate string and see what comes back

stormckey: You mean this method?
```scala
  def lex(src: Source): (Array[Token], List[LexerError]) = {
```
We can first pack the string into a Source, then lex it, then see if it's a keyword. It's a little bit roundabout but it's doable.

main/src/ca/uwaterloo/flix/api/lsp/CompletionItem.scala
stormckey: My bad, I will move it.

magnus-madsen (approved):

PR #10396, opened 2025-04-29
feat: use ExprContext for EnumTagCompleter

stormckey:
We've already applied ExprContext to EnumTagCompleter. Similarly, we can use that for EnumTag.

Preview:
image
image
image
image

Now several problems arise during the impl:
- It is possible that the ectx for EnumTagCompleter is `InsideRunWith`, which is actually an invalid context to complete to a EnumTag. What we should do in this case is to stop returning completion for enum tag. But in `toCompletionItem`, we have no choice but have to return one CompletionItem. This problem implies that:
  - Not only can we use this context to determine the content of a completion, but also whether we should offer a completion.
  - We may need a method to suppress the completion in `toCompletionItem`
- EnumTagCompleter is also used for `UndefinedTag`, which mean the current ExprContext, which is detecting `UndefinedName` in the leaf, will only find `ExprContext.Unknown`. The real context for a UndefinedTag is:
image
Pattern.Error -> MatchRule -> Match

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
@@ -47,8 +47,6 @@ object CompletionProvider {
     * Returns all completions in the given `uri` at the given position `pos`.
     */
   private def getCompletions(uri: String, pos: Position, currentErrors: List[CompilationMessage])(implicit root: Root, flix: Flix): List[Completion] = {
-    val stack = LspUtil.getStack(uri, pos)
-    stack.foreach(item => println(s"$item\n"))
     if (currentErrors.isEmpty)
       HoleCompleter.getHoleCompletion(uri, pos).toList
     else
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
@@ -47,6 +47,8 @@ object CompletionProvider {
     * Returns all completions in the given `uri` at the given position `pos`.
     */
   private def getCompletions(uri: String, pos: Position, currentErrors: List[CompilationMessage])(implicit root: Root, flix: Flix): List[Completion] = {
+    val stack = LspUtil.getStack(uri, pos)
+    stack.foreach(item => println(s"$item\n"))
     if (currentErrors.isEmpty)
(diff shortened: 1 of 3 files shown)


main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -47,6 +47,8 @@ object CompletionProvider {
     * Returns all completions in the given `uri` at the given position `pos`.
     */
   private def getCompletions(uri: String, pos: Position, currentErrors: List[CompilationMessage])(implicit root: Root, flix: Flix): List[Completion] = {
+    val stack = LspUtil.getStack(uri, pos)
+    stack.foreach(item => println(s"$item\n"))
```
magnus-madsen, line 51: ???

magnus-madsen:
So is this an improvement or not?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
```diff
@@ -407,13 +407,18 @@ sealed trait Completion {
...
       val qualifiedName = if (namespace.nonEmpty)
         s"$namespace.${tag.sym.name}"
       else
         tag.sym.toString
       val name = if (qualified) qualifiedName else tag.sym.name
-      val snippet = name + CompletionUtils.formatTypesSnippet(tag.tpes)
+      val snippet = ectx match {
+        case ExprContext.InsideApply => name
+        case ExprContext.InsidePipeline => name + CompletionUtils.formatTypesSnippet(tag.tpes.dropRight(1))
+        case ExprContext.InsideRunWith => name + CompletionUtils.formatTypesSnippet(tag.tpes)
```
magnus-madsen, line 419: We could add a comment:

case ExprContext.InsideRunWith => 
   // Technically not possible, but we just revert to default behavior.
  name + CompletionUtils.formatTypesSnippet(tag.tpes)

magnus-madsen:
We can merge this; just fix the print and add the comment.

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/Completion.scala
stormckey: Done, just to note that it's *physically* possible:
image
although it makes no sense.

Another strategy, as I mentioned, is to return None in this branch (and thus changing the interface of ToCompletionItem)

magnus-madsen (approved):

PR #10398, opened 2025-04-29
feat: provide completions over resolved enum tag

### LINKED ISSUE #10133: Extend Tag Completions to work with both `Expr.Tag` and `Expr.Error(UndefinedTag)`
Based on #10132, the idea is that when we lookup the expr, we are interested in both:

- `Expr.Error(UndefinedTag)`
- `Expr.Tag` 

In the latter case the program has no error, but more completions may be available, e.g.:

```
enum Foo {
  case Red(String)
  case Redder(String)
}
```

Asking for autocomplete on `Foo.Red` can then suggest `Foo.Redder`

stormckey:
fixes #10133

Now we can get completions after a resolved enum tag.

Preview:
image
image
image

First step to providing more completions over error-free programs.

The current impl is:
- Visit the AST to build the EnumTagContext
- If the leaf is `SymUse.CaseSymUse`, we are inside a successfully resolved EnumTag node.
- use the SymUse to find the enum and provide further completions.

It also works if we are not at the end of the symuse:
image
image

Or even the namespace:
image
image

But the information we can currently get from the SymUse is still very limited
- We can not access anchor position. Thus we don't know where to insert the extra use.
- We can not access the local scope, so we don't even know if we should insert the extra use.
- We don't know currently the context like if EnumTag is applied, but we can know that from the stack.

I am not sure, but I think we can just suppress the completion which is identical to the current content.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
@@ -47,8 +47,11 @@ object CompletionProvider {
     * Returns all completions in the given `uri` at the given position `pos`.
     */
   private def getCompletions(uri: String, pos: Position, currentErrors: List[CompilationMessage])(implicit root: Root, flix: Flix): List[Completion] = {
+    val stack = LspUtil.getStack(uri, pos)
+    stack.foreach(item => println(s"$item\n"))
     if (currentErrors.isEmpty)
-      HoleCompleter.getHoleCompletion(uri, pos).toList
+      HoleCompleter.getHoleCompletion(uri, pos).toList ++
+        EnumTagCompleter.getFurtherCompletions(uri, pos)
     else
       errorsAt(uri, pos, currentErrors).flatMap {
         case err: WeederError.UndefinedAnnotation => KeywordCompleter.getModKeywords(Range.from(err.loc))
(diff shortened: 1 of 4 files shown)


stormckey:
Oops, the CI test found a bug. Well I should make the error message more readable.

magnus-madsen:
>    We can not access anchor position. Thus we don't know where to insert the extra use.
>    We can not access the local scope, so we don't even know if we should insert the extra use.
>    We don't know currently the context like if EnumTag is applied, but we can know that from the stack.

Ah yes. Never thought of this. We need all of this information somehow.

We can compute some it by collecting more information from the visitor, I suppose.

magnus-madsen:
> > We can filter out non-matched items and identical items to avoid this problem.
> 
> This sounds like something we can fix in a separate PR?

Just to be clear: You are saying we are actually returning nonsensical things, but VSCode just luckily filters them?

stormckey:
> > > We can filter out non-matched items and identical items to avoid this problem.
> > 
> > 
> > This sounds like something we can fix in a separate PR?
> 
> Just to be clear: You are saying we are actually returning nonsensical things, but VSCode just luckily filters them?

Yes the current impl is just returning every tag in the same enum and let vscode do the filtering.

The problem is who should perform the filtering, we or vscode?

If we are going to do the filtering, should we consider the position of the cursor? Like:
```
Clr.R|ed
```
will match both `Clr.Redder` and `Clr.Rad`, since we are only matching `Clr.R`

stormckey:
Well now the update version:
- Is aware of the application.
- Pass the test by filtering out synthetic SymUse
- filter based on the prefix from the begin of the symUse to the position of the cursor.

PR #10416, opened 2025-04-30
feat: add no-duplication completion invariant for defs

stormckey:
Just a duplicate of the former impl for vars.

One thing for the naming: I don't know why,but what inside Expr.Var is not `VarSymUse`, but `VarSym`. `VarSymUse` doesn't even exist and we call it `varOccurs`. But for `Expr.ApplyDef`, we have `DefSymUse`, so we just call it `defSymUse`. Shall we unify the naming? to `xxxSymUse`?

The invariant is violated.

Not sure if we should place the analysis here, but for program

```
...
def test(): Unit \ {Sys, IO} = Environment.handle(() -> {
    let rec = safeConnects("bis", exampleGraph());
    println("Safe:");
    rec#safe |> Vector.sort |> Vector.map(Chalk.green) |> Vector.forEach(println);
    println("Unsafe:");
    rec#unsafe1 |> Vector.sort |> Vector.map(Chalk.red) |> Vector.forEach(println)
})()
```

When we try to delete "|>" to "|".

First, "|>" is treated as a function.
Second, "|" makes the compiler confusing, thus emitting tons of parse errors.
Finally, these parse errors cause duplicated completion.

I would say, most of the time, the duplicated completions come from multiple call to the same KeywordCompleter.

### DIFF
--- a/main/test/ca/uwaterloo/flix/api/lsp/TestCompletionProvider.scala
+++ b/main/test/ca/uwaterloo/flix/api/lsp/TestCompletionProvider.scala
@@ -293,17 +293,17 @@ class TestCompletionProvider extends AnyFunSuite {
   }
 
   test("No duplicated completions for vars") {
-    val charToTrim = 1
+    val numToTrim = 1
     Programs.foreach( program => {
       val (root1, _) = compile(program)
       val varOccurs = getVarSymOccurs()(root1)
       varOccurs.foreach{
-        case varOccur if isValidCodeToTrim(varOccur.text, varOccur.loc, charToTrim) =>
-          val alteredProgram = trimAfter(program, varOccur.loc, charToTrim)
-          val triggerPosition = Position(varOccur.loc.sp2.lineOneIndexed, varOccur.loc.sp2.colOneIndexed - charToTrim)
+        case varOccur if isValidCodeToTrim(varOccur.text, varOccur.loc, numToTrim) =>
+          val alteredProgram = trimAfter(program, varOccur.loc, numToTrim)
+          val triggerPosition = Position(varOccur.loc.sp2.lineOneIndexed, varOccur.loc.sp2.colOneIndexed - numToTrim)
           val (root, errors) = compile(alteredProgram)
           val completions = CompletionProvider.autoComplete(Uri, triggerPosition, errors)(root, Flix)
-          assertNoDuplicatedCompletions(completions, varOccur.text, varOccur.loc, program, charToTrim)
+          assertNoDuplicatedCompletions(completions, varOccur.text, varOccur.loc, program, numToTrim)
         case _ => ()
       }
     })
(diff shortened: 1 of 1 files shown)


magnus-madsen:
We should probably restrict ourselves to ascii variable names.

As for VarSym, just leave it as is for now.

main/test/ca/uwaterloo/flix/api/lsp/TestCompletionProvider.scala
```diff
@@ -265,48 +265,65 @@ class TestCompletionProvider extends AnyFunSuite {
...
-    * Asserts that the given completion list is empty at the given position.
-    */
-  private def assertEmpty(completions: CompletionList, sourceLocation: SourceLocation, pos: Position): Unit = {
-    if (completions.items.nonEmpty) {
-      println(code(sourceLocation, s"Unexpected completions at $pos"))
-      println(s"Found completions: ${completions.items.map(_.label)}")
-      fail(s"Expected no completions at position $pos, but found ${completions.items.toList.length} completions.")
-    }
+  test("No duplicated completions for defs") {
+    val charToTrim = 1
```
magnus-madsen, line 269: We should try trimming up to n-1 chars

magnus-madsen:
Can we extract more common functionality and do it for say other things too? E.g. tags?

main/test/ca/uwaterloo/flix/api/lsp/TestCompletionProvider.scala
stormckey: Scheduled in the next PR for both defs and vars, do you want it right here?

stormckey:
> Can we extract more common functionality and do it for say other things too? E.g. tags?

The type is quite different, one is Expr.Var, while the other is SymUse.DefSymUse.

The current helper functions are already pretty common so that we have to extract specific field from Var/DefSymUse to call them. To make them more generic, we have to define a common trait for those nodes? Or you have some other idea?

stormckey:
Why is the testing out of memory?
image

stormckey:
I found that we will not OOM as long as we comment out this file `//    "examples/misc/type-level-programming/type-level-programming-4bit-adder.flix"`

I've no idea what's happening here. Here it the content:
```scala
///
/// We can use type-level Booleans to represent 4-bit integers.
///
/// We represent a 4-bit integer as 4 bits: a, b, c, d.
///
enum I4[_a: Bool, _b: Bool, _c: Bool, _d: Bool] {
    case I4 // Dummy value
}

///
/// We can specify a function that returns the number 3.
///
def mk3(): I4[false, false, true, true] = I4.I4

///
/// We can define type-level increment on 4-bit integers.
///
/// This is somewhat involved since we have to keep track of overflow.
///
def inc(_: I4[a, b, c, d]):
    I4[a xor (b and c and d), b xor (c and d), c xor d, not d] = I4.I4

///
/// We can now specify a function that returns the number 4.
///
/// The type system ensures that we cannot make a mistake.
///
def mk4(): I4[false, true, false, false] = inc(mk3())

///
/// We can use type inference to compute the number 7.
///
def f(): Unit =
    let _seven = inc(inc(inc(mk4()))); // Hover to see type.
    ()
```

stormckey:
If you are good with this we can merge and then iterate.

stormckey:
There are two following PRs blocked here, one for deleting more than one char, one for the subset checking. It would be great if we could progress on this.

stormckey:
oh I 'll fix the conflict

magnus-madsen (approved):

PR #10417, opened 2025-04-30
draft: add delta set to `Inliner`

jaschdoc:
Now the `Inliner` still visits the functions that it changed last round but excludes any def that it did not touch last round and did not change.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
@@ -68,11 +68,12 @@ import scala.jdk.CollectionConverters.ConcurrentMapHasAsScala
 object Inliner {
 
   /** Performs inlining on the given AST `root`. */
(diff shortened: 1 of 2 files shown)


magnus-madsen:
Are we sure about this? This is the dangerous choice right?

jaschdoc:
No the dangerous choice is letting the occurrence analyzer propagate its own delta set

jaschdoc:
I tried using this on the inliner pr with a `while (delta.nonEmpty)` instead of a for loop in the `Optimizer` and all tests pass. It usually doesn't go more than 2 rounds (bear ind mind that we do not do any beta reduction or fancy stuff yet).

jaschdoc:
Correct. But it does not do that currently. It was just a sanity test.

jaschdoc:
If it does depend on non-local factors, then we cannot do this.

magnus-madsen:
I propose not to do this. Or at least not until we are much further and have a complete understanding.

jaschdoc:
Sounds good

PR #10425, opened 2025-05-01
fix: properly visit inlined bindings

jaschdoc:
I double checked with the reference, and found that we should visit expressions again in the var case


jaschdoc:
!!!rerun

magnus-madsen (approved):

PR #10427, opened 2025-05-01
fix: close `LibLevel.Min`

### LINKED ISSUE #10393: `LibLevel.Min` depends on things outside itself
Can be reproduced by running Monomorpher immediately after Lowering. It seems like there are functions that cannot be found in the `Concurrent` module.

JonathanStarup:
Fixes #10393


JonathanStarup:
Sorry, based it on wip code, ready now

magnus-madsen (approved):

PR #10431, opened 2025-05-02
feat: beta reduce lambdas in `Inliner`

jaschdoc:


main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -407,6 +414,29 @@ object Inliner {
...
+    *   exp1'
+    * }}}
+    * where `sym` is the symbol of formal parameter and `exp1'` is the body of the lambda.
+    *
+    * Lastly, it visits the let-binding, thus possibly removing the binding.
+    */
+  private def betaReduceLambda(exp1: Expr.Lambda, exp2: Expr, loc: SourceLocation, ctx0: LocalContext)(implicit sym0: Symbol.DefnSym, sctx: SharedContext, root: OccurrenceAst.Root, flix: Flix): Expr = {
+    val sym = exp1.fparam.sym // visitExp will refresh the symbol
+    val tpe = exp1.exp.tpe
+    val eff = Type.mkUnion(exp1.exp.eff, exp2.eff, loc)
```
magnus-madsen, line 434: We know that exp1.exp is pure right?

jaschdoc: No it's the body of the lambda so it can have any effect

magnus-madsen (approved):

PR #10433, opened 2025-05-02
feat: only consider functions with non-zero constant parameters in `LambdaDrop`

jaschdoc:


main/src/ca/uwaterloo/flix/language/phase/optimizer/LambdaDrop.scala
```diff
@@ -75,14 +75,17 @@ object LambdaDrop {
...
     *
     * The latter condition can be checked by the [[isHigherOrder]] predicate.
     */
   private def isDroppable(defn: MonoAst.Def)(implicit lctx: LocalContext): Boolean = {
     if (isHigherOrder(defn)) {
       visitExp(defn.exp)(defn.sym, lctx)
-      lctx.recursiveCalls.nonEmpty
+      val hasConstantParameter = paramKinds(lctx.recursiveCalls.toList, defn.spec.fparams)
+        .exists { case (_, pkind) => pkind == ParamKind.Const }
+      lctx.recursiveCalls.nonEmpty && hasConstantParameter
```
jaschdoc, line 88: `hasConstantParameter` implicitly checks that `lctx.recursiveCalls` is non-empty, so we could remove that check.

main/src/ca/uwaterloo/flix/language/phase/optimizer/LambdaDrop.scala
```diff
@@ -75,14 +75,15 @@ object LambdaDrop {
...
+    * with a function type and
+    * (c) has at least one constant parameter.
     *
     * The latter condition can be checked by the [[isHigherOrder]] predicate.
     */
   private def isDroppable(defn: MonoAst.Def)(implicit lctx: LocalContext): Boolean = {
     if (isHigherOrder(defn)) {
       visitExp(defn.exp)(defn.sym, lctx)
-      lctx.recursiveCalls.nonEmpty
+      lctx.recursiveCalls.nonEmpty && hasConstantParameter(lctx.recursiveCalls.toList, defn.spec.fparams)
```
jaschdoc, line 86: `hasConstantParameter` implicitly checks that lctx.recursiveCalls is non-empty, so we could in principle remove that check. However, I think it makes the code more readable, since it corresponds to the documentation. Additionally, if it is empty, then `hasConstantParameter` is never called so we save a few instructions that just return `Nil` and `false` (not that it really matters)

magnus-madsen (approved):

PR #10434, opened 2025-05-02
feat: inline and beta-reduce local defs

jaschdoc:
Only applies to localdefs that occur once


magnus-madsen:
@jaschdoc Conflicted

magnus-madsen:
This is the big one right? This is where we make the big bucks?

magnus-madsen:
Ah no- this is just LocalDef

magnus-madsen:
What happens if the local def is recursive?

magnus-madsen (approved):

magnus-madsen:
FYI: I cannot seem to trigger this.

jaschdoc:
```flix
def main(): List[Int32] = {
    def f(x) = x + 1;
    f(1) :: Nil
}
```

PR #10435, opened 2025-05-02
feat: add deforestation in `Inliner`

jaschdoc:


jaschdoc:
@magnus-madsen initial review?

magnus-madsen:
The code can be improved; but I think this is "nice to have" so lets wait a bit until we have a better grip on the other things.

magnus-madsen:
I looked through some generated code. I did not really see any examples of code that would benefit from deforestation-- except for one specific case: when we use records as "labelled arguments". However, that is somewhat rare.

PR #10436, opened 2025-05-02
fix: ignore completion duplication check for vars

stormckey:
An attempt to eliminate duplicate keyword completions.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
@@ -39,18 +39,18 @@ import ca.uwaterloo.flix.language.errors.{ParseError, ResolutionError, TypeError
 object CompletionProvider {
 
   def autoComplete(uri: String, pos: Position, currentErrors: List[CompilationMessage])(implicit root: Root, flix: Flix): CompletionList = {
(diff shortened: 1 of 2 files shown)


stormckey:
This strategy works! let me clean up.

stormckey:
I think the failure is irrelevant?

magnus-madsen:
Any reason not to look on the specific line?

magnus-madsen:
Did you test this PR a bit manually to see that it works well?

stormckey:
> Did you test this PR a bit manually to see that it works well?

Sadly, it doesn't work that well, let me give you several example:

image

Here the unexpected token is selected, since it's just over the "."

image

Here we choose the rc part instead of the reg part, since that is closer to the cursor.

I would say it's not a good idea to introduce this selection, we'll lost a lot of suggestions.

And I am not sure if we can completely determine the error just from the position.

How about my former suggestion to collect keywordCompleters into a set, call them in one place and then deduplicate them?

magnus-madsen:
Did you consider other definitions of nearest?

magnus-madsen:
Can we fix the CI meanwhile - e.g. by just ignoring the test etc?

stormckey:
> Can we fix the CI meanwhile - e.g. by just ignoring the test etc?

The error is just from No duplicated completions, we can ignore that.

stormckey:
We can change this PR to just ignore the test if you wish.

magnus-madsen:
Let's ignore for now

PR #10438, opened 2025-05-02
feat: Optimize enum wrappers

JonathanStarup:
An enum wrapper is something like `Set`, where it has **one** case with **one** term.
```
enum Set[t](RedBlackTree[t, Unit])
```
Why even have `Set`? Let's delete it and use `RedBlackTree` directly!

# When in the compiler?
- Trait instances are decided based on the enum, not the inner type, so it should be after `Monomorpher` where those are resolved.
- Erasure needs to know about wrappers potentially becoming primitive types, so it should be before `Eraser`/`Reducer`.

# When can it be done?
There are some scenarios where wrappers cannot just be deleted.
- **Recursive**: If the wrapper mentions its own type directly, e.g. `enum E(List[E])`. If we try to remove the type `E`, we would get `List[List[List[..]]]` infinitely - not good.
- **Mutually recursive:** Similarly if the wrapper mentions itself through other wrappers, it is not good, e.g. `enum E(A)` and `enum A(E)`. This leads to infinite types again. Note that you can remove one of the two wrappers, so there are some design choices about when and how.
- **Partial Type:** for `enum E[t, ef](t -> t \ ef)`, if someone uses the type `E[t]`, we cannot directly remove it. We can conceptually transform `E[t]` into a type function `ef ==> t -> t \ ef`, but we do not have that (currently).

## Design Considerations
Sometimes, wrapper enums cannot be optimized, as described above. So, what should we do for the user?
- **hint:** We could give a hint message that the wrapper is not optimized
- **opt-in optimization:** We could have a contract that `@optimize` enums must be optimized and give errors if not able to
- **opt-out optimization:** We could error for wrappers that cannot be optimized unless annotated by `@Unoptimized`

These choices would mean that a frontend phase checks the conditions that allow/disallow optimization.

# Issues

## Native `toString`
`ToString.toString` works perfectly since it relies on instances, but native versions like `dbg!` and `"%{..}"` will observe that the wrappers are not present anymore

## Typematch
```
enum Wrap(Int32)
def main(): Unit \ IO = typematch Wrap(42) {
    case x: Wrap     => println("wrap")
    case x: Int32    => println("int32")
    case _: _        => println("other")
}
```
This program correctly prints `wrap` since typematch is resolved in `Monomorpher`. But what about a more indirect situation.
```
import java.lang.Object
enum Wrap(Int32)
def main(): Unit \ IO = typematch Wrap(42) {
    case v: Bool    => println("bool")
    case v: Char    =>  println("char")
    case v: Int8    =>  println("int8")
    case v: Int16   =>  println("int16")
    case v: Int32   =>  println("int32")
    case v: Int64   =>  println("int64")
    case v: Float32 =>  println("float32")
    case v: Float64 =>  println("float64")
    case obj: _     => println(unchecked_cast(obj as Object).toString())
}
```
Normally, we would be certain that anything that reaches the end must be a non-primitive type. But in this case `Wrap` goes to the bottom case and is resolved in the monomorpher. Then `Wrap` is converted into `Int32` by the optimizer, which makes the `unchecked_cast` result in invalid bytecode :( Very sad. The naive solution, to have `Wrap` match the case of `int32` is also not great since we then cannot have the sensible typematch in the first example.

One possible solution is a new construct specifically for JVM reflection (#10446)

## How often can these be optimized?
The most notable enums in the standard library are `Set`, `Map`, `StringBuilder`, and `Fixpoint.Ast.Datalog.VarSym`.
<details>
<summary> The following enums are optimizable wrappers in the standard library: </summary>

- Concurrent.Condition
- Concurrent.CyclicBarrier
- Concurrent.ReentrantLock
- DelayMap
- Down
- Fixpoint.Ast.Datalog.VarSym
- Fixpoint.Ast.PrecedenceGraph
- Http.Response
- Identity
- Map
- MultiMap
- MutDisjointSets
- ProcessHandle
- Regex.Matcher
- SemiGroup.All
- SemiGroup.Any
- Set
- StringBuilder
- Time.Duration
- Time.Instant

</details>

### Compile time
`Xperf --n 50` went from 72.674 to 71.310 (-1.8 %)

### Run time

Our benchmarks do not use any of the optimized enums, so this result is probably just noise. I'd expect a Datalog benchmark to be more interesting.

| Benchmark | Master | PR | % speed up |
| --- | --- | --- | --- |
| InlinerFilterMap10k | 280572091 | 281539343 | +0.3 % |
| InlinerFilterMap10kOptimized | 254934342 | 256703816 | +0.7 % |
| InlinerMap10k | 508743549 | 527165793 | +3.6 % |
| InlinerMap10kOptimized | 292011855 | 328735342 | +12.6 % |
| List.filter | 488389542 | 542103541 | +11.0 % |
| List.filterMap | 297045669 | 334296876 | +12.5 % |
| List.foldLeft | 453653705 | 454165958 | +0.1 % |
| List.foldRight | 507760589 | 506814197 | -0.2 % |
| List.length | 235668654 | 236036851 | +0.2 % |
| List.map | 492098841 | 489013780 | -0.6 % |
| List.reverse | 241900948 | 241729198 | -0.1 % |


main/src/ca/uwaterloo/flix/api/Flix.scala
```diff
@@ -144,9 +144,6 @@ class Flix {
     // Reflect
     "Reflect.flix" -> LocalResource.get("/src/library/Reflect.flix"),
 
-    // Debug
-    "Debug.flix" -> LocalResource.get("/src/library/Debug.flix"),
```
JonathanStarup, line 148: This change is for typematch reasons

main/src/library/Debug.flix
```diff
@@ -105,6 +105,11 @@ mod Debug {
                 unsafe Arrays.deepToString(unchecked_cast(y as Array[Object, Static]))
             case y: Array[_, _] =>
                 unsafe Arrays.toString(unchecked_cast(y as Array[Object, Static]))
+            case y: Down[_] => stringify(coerce(y))
```
JonathanStarup, line 108: hacking typematch to work

main/src/library/Fixpoint/Boxable.flix
```diff
@@ -32,6 +32,31 @@ mod Fixpoint.Boxable {
         case v: Int64   => Boxed.BoxedInt64(v)
         case v: Float32 => Boxed.BoxedFloat32(v)
         case v: Float64 => Boxed.BoxedFloat64(v)
+        case v: Down[Bool]    => Boxed.BoxedBool(coerce(v))
```
JonathanStarup, line 35: hacking typematch to work

main/src/library/Fixpoint/Boxable.flix
```diff
@@ -73,6 +98,90 @@ mod Fixpoint.Boxable {
             case Boxed.BoxedFloat64(v)   => unchecked_cast(v as a)
             case _                       => unreachable!()
         }
+        case _: Proxy[Down[Bool]] => match x {
```
JonathanStarup, line 101: hacking typematch to work

magnus-madsen:
Did you accidentally merge the wrong branch? This seems to have the JvmReflect stuff in it too.

JonathanStarup:
> Did you accidentally merge the wrong branch? This seems to have the JvmReflect stuff in it too.

This requires JvmReflection. It's described in the main text, but the change of enum representation exposes a coupling of type match and jvm stuff that doesn't hold here.

magnus-madsen:
I assume that with the new MonoType and SimpleType we will need a clean design here?

JonathanStarup:
No. It's unchanged.

magnus-madsen:
I don't want a solution that introduces another typematch-like construct.

I just want a simple AST rewrite.

JonathanStarup:
> I just want a simple AST rewrite.

This is impossible because of type match. I've described it in the main issue but perhaps we can discuss it at the next Flix meeting?

mlutze:
> > I just want a simple AST rewrite.
> 
> This is impossible because of type match. I've described it in the main issue but perhaps we can discuss it at the next Flix meeting?

Adding to the meeting list.

magnus-madsen:
Closing -- but we are re-implementing this in small pieces.

PR #10441, opened 2025-05-03
feat: add lambda inlining heuristics

jaschdoc:
Callsite inlining of lambdas. I will add inlining of top level defs in another PR once this works.


jaschdoc:
We currently only have the heuristic on lambdas that if the body of a lambda is smaller than the call itself, then it is inlined. Combined with beta reduction this gives

```flix
def main(): Int32 = {
    let f = x -> x + 1;
    let g = x -> y -> z -> x + y + z;
    let h = f;
    (y -> (x -> h(x + 2))(3) + g(4)(5)(y) + (x -> f(x + 6))(7))(8)
}
```
becomes
```
def main(_unit$131086: Unit): Int32 = {
    add$131095(
        add$131095(
            add$131095(add$131095(3i32, 2i32), 1i32),
            add$131095(add$131095(4i32, 5i32), 8i32)
        ),
        add$131095(add$131095(7i32, 6i32), 1i32)
    )
}
```

Of course this is still a yes-instance / good case for the heuristic but it shows that it is actually quite effective in this instance.

magnus-madsen:
Do we close this?

PR #10443, opened 2025-05-03
refactor: remove null casting logic from `TypeReconstruction`

JonathanStarup:
This doesn't belong here and makes internal Asts nonwelltyped

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Safety.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Safety.scala
@@ -450,6 +450,9 @@ object Safety {
       )
 
       (Type.eraseAliases(from).baseType, to.map(Type.eraseAliases).map(_.baseType)) match {
+        // Allow all null casts
+        case (Type.Null, _) => ()
+
         // Allow casts where one side is a type variable.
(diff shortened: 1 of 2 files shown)


main/src/ca/uwaterloo/flix/language/phase/Safety.scala
```diff
@@ -450,6 +450,9 @@ object Safety {
       )
 
       (Type.eraseAliases(from).baseType, to.map(Type.eraseAliases).map(_.baseType)) match {
+        // Allow all null casts
+        case (Type.Null, _) => ()
```
JonathanStarup, line 454: This is clearly a bit weird, but this PR is a pure refactor - we can fix this in follow up work

magnus-madsen: What is the fix?

magnus-madsen (approved):

main/src/ca/uwaterloo/flix/language/phase/Safety.scala
JonathanStarup: Just that e.g. `unchecked_cast(null as Int32)` should probably also give a safety error

PR #10446, opened 2025-05-03
feat: JVM reflection

JonathanStarup:


magnus-madsen:
But now we need two constructs or what?

main/src/ca/uwaterloo/flix/api/Flix.scala
```diff
@@ -144,6 +144,7 @@ class Flix {
     "Reflect.flix" -> LocalResource.get("/src/library/Reflect.flix"),
 
     // Debug
+    "JvmReflection.flix" -> LocalResource.get("/src/library/JvmReflection.flix"),
```
mlutze, line 147: Should be under // Reflect instead of // Debug, right?

main/src/ca/uwaterloo/flix/language/phase/Monomorpher.scala
```diff
@@ -560,6 +560,54 @@ object Monomorpher {
           }
       }.get // This is safe since the last case can always match.
 
+    case LoweredAst.Expr.JvmReflection(exp, true, tpe, eff, loc) =>
```
mlutze, line 563: Does pattern matching allow doing `proxy = true` here?

main/src/ca/uwaterloo/flix/language/phase/Monomorpher.scala
```diff
@@ -560,6 +560,54 @@ object Monomorpher {
...
+      val e = specializeExp(exp, env0, subst)
```
mlutze, line 564: I know you are anti-comment, but a high-level comment of what's happening in each of these cases would be nice.

magnus-madsen:
I don't understand why we are introducing two features that can do similar things?

If typematch is more powerful then we should just keep that?

mlutze:
> I don't understand why we are introducing two features that can do similar things?
> 
> If typematch is more powerful then we should just keep that?

I'm hoping this will kill typematch...

main/src/ca/uwaterloo/flix/language/phase/Monomorpher.scala
JonathanStarup: sadly no :L I could add a enum for it.

JonathanStarup:
> I don't understand why we are introducing two features that can do similar things?

Halfway through answering, I realised one i sufficient - changed it now

> If typematch is more powerful then we should just keep that?

Typematch is broken and complicated - I'm just taking away one application of it that a) allows singleton enum optimization and b) makes space for other work to take away the remaining applications.

magnus-madsen:
I would like a write-up of what it is and why it is better.

JonathanStarup:
@magnus-madsen I've updated the main text. can we do it verbally if you want to discuss further?

PR #10449, opened 2025-05-04
feat: make `Input.equals` consistent

JonathanStarup:
Now equality follows the pattern of `Input.Text`

I can't figure out if equals/hashcode is used, but it should be consistent (we can also remove the override of `Input.Text`)

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/shared/Input.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/shared/Input.scala
@@ -54,17 +54,38 @@ object Input {
   /**
     * Represents an input that originates from the filesystem.
     */
-  case class TxtFile(path: Path, sctx: SecurityContext) extends Input
+  case class TxtFile(path: Path, sctx: SecurityContext) extends Input {
+    override def hashCode(): Int = path.hashCode()
+
+    override def equals(obj: Any): Boolean = obj match {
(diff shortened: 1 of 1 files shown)


main/src/ca/uwaterloo/flix/language/ast/shared/Input.scala
```diff
@@ -54,17 +54,38 @@ object Input {
...
 
   /**
     * Represents an input that originates from inside a package.
     */
-  case class FileInPackage(packagePath: Path, virtualPath: String, text: String, sctx: SecurityContext) extends Input
+  case class FileInPackage(packagePath: Path, virtualPath: String, text: String, sctx: SecurityContext) extends Input {
+    override def hashCode(): Int = packagePath.hashCode()
+
+    override def equals(obj: Any): Boolean = obj match {
+      case that: FileInPackage => this.packagePath == that.packagePath
```
mlutze, line 85: We don't care about the other fields here?

JonathanStarup: hmm i think not. Maybe we should just not override the equals method - I'm pretty sure its unused. It was just the one override in one case class that seemed wrong

mlutze:
Seems like this broke community builds?

JonathanStarup:
I give up

PR #10454, opened 2025-05-05
feat: add simple `ApplyDef` inlining

jaschdoc:


main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -181,18 +181,17 @@ object Inliner {
           Expr.ApplyClo(e1, e2, tpe, eff, loc)
       }
 
-    case Expr.ApplyDef(sym, exps, itpe, tpe, eff, loc) =>
-      val es = exps.map(visitExp(_, ctx0))
-      Expr.ApplyDef(sym, es, itpe, tpe, eff, loc)
+    case exp@Expr.ApplyDef(_, _, _, _, _, _) =>
+      callSiteInlineDef(exp, ctx0)
```
magnus-madsen, line 185: Can you move the code here?

Currently call this helper function just to pattern match again. We never use that pattern, instead we aim for structurally recursive functions.

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -515,6 +514,37 @@ object Inliner {
...
+    !ctx0.currentlyInlining && !isRecursive(defn, sym0) && isDirectCall(defn)
+  }
+
+  /** Returns `true` if `defn.sym` is equal to `sym0` */
+  private def isRecursive(defn: OccurrenceAst.Def, sym0: Symbol.DefnSym) = {
+    defn.sym != sym0
+  }
+
+  /** Returns `true` if `defn` is marked as a direct call. */
+  private def isDirectCall(defn: OccurrenceAst.Def): Boolean = {
```
magnus-madsen, line 544: This is really just forwarding a call, so I am not sure if that is helpful.

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -515,6 +514,37 @@ object Inliner {
...
+      val es = exps.map(visitExp(_, ctx0))
+      Expr.ApplyDef(sym, es, itpe, tpe, eff, loc)
+  }
+
+  /** Returns `true` if `defn` is not recursive and is either a higher-order function or is a direct call to another function. */
+  private def shouldInlineDef(defn: OccurrenceAst.Def, ctx0: LocalContext)(implicit sym0: Symbol.DefnSym): Boolean = {
+    !ctx0.currentlyInlining && !isRecursive(defn, sym0) && isDirectCall(defn)
+  }
+
+  /** Returns `true` if `defn.sym` is equal to `sym0` */
```
magnus-madsen, line 538: The comment does not match the implementation.

Also I dont understand the check.

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -464,24 +463,24 @@ object Inliner {
...
+    *   exp
     * }}}
-    * where `symi` is the symbol of the i-th formal parameter and `exp'` is the body of the local def.
+    * where `symi` is the symbol of the i-th formal parameter and `exp` is the body of the function.
     *
     * Lastly, it visits the top-most let-binding, thus possibly removing the bindings.
     */
-  private def betaReduceLocalDef(exp: Expr.LocalDef, exps: List[Expr], loc: SourceLocation, ctx0: LocalContext)(implicit sym0: Symbol.DefnSym, sctx: SharedContext, root: OccurrenceAst.Root, flix: Flix): Expr = {
-    val bindings = exp.fparams.zip(exps).foldRight(exp.exp1) {
+  private def betaReduce(exp: Expr, exps: List[(FormalParam, Expr)], loc: SourceLocation, ctx0: LocalContext)(implicit sym0: Symbol.DefnSym, sctx: SharedContext, root: OccurrenceAst.Root, flix: Flix): Expr = {
```
magnus-madsen, line 482: Just so we are clear: This is not beta reduction.

This is creating an expression that let binds the formal arguments.

magnus-madsen: While it also visits the expression body.

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -515,6 +514,37 @@ object Inliner {
     case _ => false
   }
 
+  /**
+    * Returns an inlined and beta-reduced version of `exp0.sym` if the [[shouldInlineDef]] predicate holds.
+    * Otherwise, returns an [[Expr.ApplyDef]] expression , where the subexpressions have been visited.
+    */
+  private def callSiteInlineDef(exp0: Expr.ApplyDef, ctx0: LocalContext)(implicit sym0: Symbol.DefnSym, sctx: SharedContext, root: OccurrenceAst.Root, flix: Flix): Expr = exp0 match {
+    case Expr.ApplyDef(sym, exps, _, _, _, loc) if shouldInlineDef(root.defs(sym), ctx0) =>
```
magnus-madsen, line 522: This should be inlined up into the bigger pattern match and then you can also drop the guard and just use a plain old if.

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
jaschdoc: True, this is an error

magnus-madsen:
(Removed from queue- want to see test time first)

magnus-madsen:
@jaschdoc This simple PR seems to have already killed compiler performance-- so we will need to fix that first :)

Issue #9067, opened 2024-10-30
`NullPointerException` in compiler when invoking Java methods with wrong arguments


sockmaster27:
Attempting to compile the following program
```flix
import java.nio.file.Files
import java.nio.file.Paths

def main(): Unit \ IO = {
    let path = Paths.get("");
    let exists = Files.exists(path);
    ()
}
```
fails with message
```
java.lang.NullPointerException: Cannot invoke "java.lang.Class.getName()" because the return value of "java.lang.Class.getSuperclass()" is null
	at org.apache.commons.lang3.reflect.MethodUtils.getMatchingAccessibleMethod(MethodUtils.java:354)
	at ca.uwaterloo.flix.language.phase.typer.TypeReduction$.retrieveMethod(TypeReduction.scala:208)
	at ca.uwaterloo.flix.language.phase.typer.TypeReduction$.lookupStaticMethod(TypeReduction.scala:202)
	at ca.uwaterloo.flix.language.phase.typer.TypeReduction$.simplify(TypeReduction.scala:148)
	at ca.uwaterloo.flix.language.phase.typer.ConstraintSolver$.$anonfun$resolveOne$2(ConstraintSolver.scala:262)
	at ca.uwaterloo.flix.language.phase.typer.ConstraintSolver$.resolveOne(ConstraintSolver.scala:261)
	at ca.uwaterloo.flix.language.phase.typer.ConstraintSolver$.tryResolve$1(ConstraintSolver.scala:241)
	at ca.uwaterloo.flix.language.phase.typer.ConstraintSolver$.resolveOneOf(ConstraintSolver.scala:251)
	at ca.uwaterloo.flix.language.phase.typer.ConstraintSolver$.resolve(ConstraintSolver.scala:217)
	at ca.uwaterloo.flix.language.phase.typer.ConstraintSolver$.visitSpec(ConstraintSolver.scala:104)
	at ca.uwaterloo.flix.language.phase.typer.ConstraintSolver$.visitDef(ConstraintSolver.scala:50)
	at ca.uwaterloo.flix.language.phase.Typer$.visitDef(Typer.scala:187)
	at ca.uwaterloo.flix.language.phase.Typer$.$anonfun$visitDefs$1(Typer.scala:168)
	at ca.uwaterloo.flix.util.ParOps$.$anonfun$parTraverseValues$1(ParOps.scala:106)
	at ca.uwaterloo.flix.util.ParOps$.$anonfun$parMap$2(ParOps.scala:64)
[...]

PR #10463, opened 2025-05-06
chore: upgrade commons-lang3 to snapshot

mlutze:
Related to https://github.com/flix/flix/issues/9067

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,6 +9,9 @@ application {
 
 repositories {
     mavenCentral()
+    maven {
+        url = 'https://repository.apache.org/snapshots/'
+    }
 }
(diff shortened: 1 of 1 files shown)


magnus-madsen:
Do you need this? Should we not just wait for the next version?

mlutze:
> Do you need this? Should we not just wait for the next version?

It's unclear when the next version will come out.

magnus-madsen:
> > Do you need this? Should we not just wait for the next version?
> 
> It's unclear when the next version will come out.

We could ping them?

magnus-madsen:
We wait for the next stable release.

PR #10473, opened 2025-05-06
feat: trim every possible number of chars in TestCompletionProvider

stormckey:
The test now fails when we try to trim more than 1 char. And the reason is also duplicate call to keyword completer.

Shall we also ignore this invariant for now?


stormckey:
> We should fix the problem :)

Yeah, but we've not decided the solution for this.

magnus-madsen:
What happens when we trim more than one char? Why does that cause the keyword completer to give duplicates?

stormckey:
The failure case is:
image

We have two errors that will trigger keywordCompleter, thus causing duplicate completions.

The PR that filter by the prefix can fix this specific one, but there is no guarantee that every time we trim the def to a few chars, the extra call to keywordCompleter will be filtered.

stormckey:
> Ah, you _completely remove it_. I was suggesting to consider all prefixes of length at least 1. So never an empty string

Ahh, I do consider that. But I am only ensuring the whole def (`List.range`) is not completely removed.

stormckey:
> > Ah, you _completely remove it_. I was suggesting to consider all prefixes of length at least 1. So never an empty string
> 
> Ahh, I do consider that. But I am only ensuring the whole def (`List.range`) is not completely removed.

But do you think testing no duplication over `List.` makes no sense here? Of course it makes no sense to test that over ""(nothing left). But we can see if we provide duplicate completions in `List.`

magnus-madsen:
I think the parser really struggles when a name is completely omitted. Maybe its OK for e.g. `Foo.`. But I propose to focus now on names that retain at least one char-- that way the program remains syntactically correct.

stormckey:
Now we will only try to trim on the name (namespace excluded) and we will keep at least on char for the name

magnus-madsen (approved):

PR #10476, opened 2025-05-06
feat: pass prefix to KeywordCompleter for filtering

### LINKED ISSUE #10466: `KeywordCompleter` should take an option with the prefix of the `UndefinedName` (if there is such as name)

stormckey:
fixes #10466

Note that in ExprKeyword I split the list into two parts to make sure each list is of the same type.

Or we can change the Completion types to keep things in one list.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
@@ -51,7 +51,7 @@ object CompletionProvider {
       HoleCompleter.getHoleCompletion(uri, pos).toList
     else
       errorsAt(uri, pos, currentErrors).flatMap {
-        case err: WeederError.UndefinedAnnotation => KeywordCompleter.getModKeywords(Range.from(err.loc))
+        case WeederError.UndefinedAnnotation(name, loc) => KeywordCompleter.getModKeywords(Some(name), Range.from(loc))
 
         case err: WeederError.UnqualifiedUse => UseCompleter.getCompletions(err.qn, Range.from(err.loc))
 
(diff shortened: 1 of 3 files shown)


magnus-madsen:
A few things:

- Let us not fuzzy match on keywords!
- If we have a qname `Foo.bar` we should not expect a keyword. Only if the qname has a single name should we call keyword completer.

main/src/ca/uwaterloo/flix/api/lsp/provider/CompletionProvider.scala
```diff
@@ -70,17 +70,20 @@ object CompletionProvider {
           val ident = err.qn.ident.name
           val qn = err.qn
           val range = Range.from(err.loc)
+          val keywordCompletions = if (qn.namespace.nonEmpty)
```
magnus-madsen, line 73: Did you not flip it around here?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/KeywordCompleter.scala
```diff
@@ -149,7 +146,12 @@ object KeywordCompleter {
       Completion.KeywordCompletion("without"     , range, Priority.Default),
       // Y
       Completion.KeywordCompletion("yield"       , range, Priority.Default)
-    )
+    ).filter(c => c.name.startsWith(name.getOrElse(""))) ++ List (
+      Completion.KeywordLiteralCompletion("false", range, Priority.Higher),
```
magnus-madsen, line 150: What is this? Why is it like this?

magnus-madsen: Let us add a field

PR #10477, opened 2025-05-06
feat: completion invariant that auto use/import should be lower

### LINKED ISSUE #10467: TestCompletionProvider: AutoUseCompletion and AutoImportCompletion must be ranked lower than all over completions

stormckey:
fixes #10467
Here we first check this invariant for all DefSymUse.

The error we encounter:
image

Is that the `Console.println` ranks higher than `throw`(which is filtered out)

### DIFF
--- a/main/test/ca/uwaterloo/flix/api/lsp/TestCompletionProvider.scala
+++ b/main/test/ca/uwaterloo/flix/api/lsp/TestCompletionProvider.scala
@@ -342,19 +342,33 @@ class TestCompletionProvider extends AnyFunSuite {
     *
     * Auto use/import completions should be lower than all other completions.
     *
-    * Note that a lower ranking means a larger sortText.
-    *
     * @param completions The completion list to check.
     * @param program     The original program string.
     * @param loc         The source location of the code that we are changing.
     */
   private def assertLowerOrderForUseImport(completions: CompletionList, program: String, loc: SourceLocation): Unit = {
     val (otherCompletions, useImportCompletions) = completions.items.partition(_.additionalTextEdits.isEmpty)
-    val lowestOtherCompletionSortText = otherCompletions.map(_.sortText).max
-    val highestUseImportCompletionSortText = useImportCompletions.map(_.sortText).min
-    if (lowestOtherCompletionSortText > highestUseImportCompletionSortText) {
+    assertPriorityHigher(otherCompletions, useImportCompletions, program, loc)
+  }
+
+  /**
+    * Asserts that the given completion list is ordered correctly.
+    *
+    * All completions in `higherCompletions` should be ranked higher than all completions in `lowerCompletions`.
+    *
+    * Note that higher-ranked completions should have a lower sortText value.
+    *
+    * @param higherCompletions The completion list to check.
+    * @param lowerCompletions  The completion list to check.
+    * @param program           The original program string.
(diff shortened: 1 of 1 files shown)


magnus-madsen:
I think there is an important point I did not make clear.

We are currently testing `CompletionProvider.autoComplete`, but really we should be testing `CompletionProvider.getCompletions` (I know it is private now, but we need to make it accessible) because `getCompletions` has more information (it returns `Completion`).

I would refactor so that `def getCompletions` is the main and only public function in CompletionProvider (in separate PR)

Does it make sense?

magnus-madsen:
> If `getCompletions` become the only entry for it, where should we call `toCompletionItem`? In the lspserver?

Yes, just do that little bit of code where we currently call autoComplete.

magnus-madsen:
> getCompletion returns an Iterable of Completions, which is very generic and has no field defined on it. 

True, we might need some common fields or more pattern matching. In any case, more structure would be good.

stormckey:
>   * Sometimes the fields in specific completions are just materials for them to compute the final label or some other things. If we don't call `toCompletionItem`, we don't do that computation. The raw materials is specific to every construct and it may not be what we want.

How about this problem?

For example:
```
  /**
    * Represents a struct completion
    *
    * @param struct    the struct construct.
    * @param range     the range of the completion.
    * @param ap        the anchor position for the use statement.
    * @param qualified indicate whether to use a qualified label.
    * @param inScope   indicate whether to the enum is inScope.
    */
  case class StructCompletion(struct: TypedAst.Struct, range: Range, ap: AnchorPosition, qualified: Boolean, inScope: Boolean) extends Completion

  /**
    * Represents a trait completion
    *
    * @param trt                trait construct.
    * @param range              the range of the completion.
    * @param ap                 the anchor position for the use statement.
    * @param qualified          indicate whether to use a qualified label.
    * @param inScope            indicate whether to the trait is inScope.
    * @param withTypeParameter  indicate whether to include the type parameter in the completion.
    */
  case class TraitCompletion(trt: TypedAst.Trait, range: Range, ap: AnchorPosition, qualified: Boolean, inScope: Boolean, withTypeParameter: Boolean) extends Completion
```

Currently we can just find out the label from CompletionItem. But for Completion, we have to build them from trait and struct and etc.

magnus-madsen:
Perhaps each test can decide? For example given a `Completion`, a test can decide to convert it to a `CompletionItem` if needed?

stormckey:
> Perhaps each test can decide? For example given a `Completion`, a test can decide to convert it to a `CompletionItem` if needed?

Sounds good, so I first refactor the interface and then change how the tester use it?

magnus-madsen:
> > Perhaps each test can decide? For example given a `Completion`, a test can decide to convert it to a `CompletionItem` if needed?
> 
> Sounds good, so I first refactor the interface and then change how the tester use it?

Yes

magnus-madsen:
> we can write a long pattern match for all constructs' completions whose inScope is false.

I prefer this

stormckey:
Now we use pattern match to find autoUse/autoImport Completions.

We can then decide if we should add priority to all completions, currently we still compare their priority through sortText.

It might also be a good idea to avoid using wildcard in the pattern match so we don't forget to add new completions to the list.

magnus-madsen:
I think all Completions should expose a priority field of type Priority

PR #10482, opened 2025-05-06
feat: Datalog Solver v3

JumpTwiice:


jaschdoc:
That's a big pr :)

magnus-madsen (commented):
A few thoughts on merging this:

- You can create a separate PR for the two examples and then we can quickly merge that one.
- Would it be possible to refactor _the current master_ to erase the use of [v] for everything in Datalog? (e.g. change lowering and the core data types).
- Assuming the previous is possible, could we have both Datalog engines and toggle between them? At least for a short while?

CasperDN:
(1): What are the 2 examples? The [v] stuff or the fix for the test in examples?
(2): It can be done. I'll make the PR tomorrow.
(3): Where/when should the toggle be? Toggling based on something in Fixpoint.Option should be possible, but toggling then requires recompiling the Flix compiler.

magnus-madsen:
> (1): What are the 2 examples? The [v] stuff or the fix for the test in examples?

The two examples with paths and Order

> (2): It can be done. I'll make the PR tomorrow.

This would be very nice and should be easy to get merged.

But how do you deal with Denotation?

> (3): Where/when should the toggle be? Toggling based on something in Fixpoint.Option should be possible, but toggling then requires recompiling the Flix compiler.

Fixpoint.Option is fine. Recompiling is fine.

CasperDN:
Denotation is used by Lowering.flix on non-boxed values. We therefore cannot get rid of its type-parameter entirely. The current fix is to have 2 denotations. One denotation for Lowering.flix and one for the engine. It would probably be cleaner to just have one Denotation-enum and declare `type alias Denotation=ParameterizedDenotation[Boxed]` for internal use.

magnus-madsen:
> Denotation is used by Lowering.flix on non-boxed values. We therefore cannot get rid of its type-parameter entirely. The current fix is to have 2 denotations. One denotation for Lowering.flix and one for the engine. It would probably be cleaner to just have one Denotation-enum and declare `type alias Denotation=ParameterizedDenotation[Boxed]` for internal use.

But `Denotation` is used inside `Constraint` so how do you avoid `Constraint` being polymorphic?

CasperDN:
Lowering.flix is nice with the way it construct Datalog-stuff. All values (and Denotations) are boxed before being inserted. The Denotation in e.g. HeadPredicate is always of type Denotation[Boxed].

magnus-madsen:
OK; then we should be able to drop almost all `[v]`. 

This would bring the two solvers closer in API.

magnus-madsen:
@CasperDN @JumpTwiice I added some thoughts here: https://github.com/flix/flix/issues/10486

magnus-madsen:
Can we close this? I assume we will complete the transition using smaller PRs.

CasperDN:
Sure

PR #10489, opened 2025-05-07
fix: correct Order instance for shortest-path examples

CasperDN:
The previous order defined paths to be equal if they had the same length. This was changed to a real total order.
Should not affect performance.


magnus-madsen (approved):

magnus-madsen:
Thanks! Looking forward to the next PR.

PR #10498, opened 2025-05-07
feat: add `AtomicOp`s for extensible variants

magnus-madsen:


magnus-madsen:
@JonathanStarup My plan is to just to add very small PRs and build from there.

You should check that my definitions are reasonable.

main/src/ca/uwaterloo/flix/language/ast/AtomicOp.scala
```diff
@@ -48,6 +48,12 @@ object AtomicOp {
 
   case class RecordRestrict(label: Name.Label) extends AtomicOp
 
+  case class VariantIs(label: Name.Label) extends AtomicOp
```
JonathanStarup, line 51: I think `ExtensibleIs` is better since both tag operate on variants

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -691,6 +691,12 @@ object GenExpression {
         mv.visitMethodInsn(INVOKEINTERFACE, interfaceType.jvmName.toInternalName, interfaceType.RestrictFieldMethod.name,
           MethodDescriptor.mkDescriptor(BackendObjType.String.toTpe)(interfaceType.toTpe).toDescriptor, true)
 
+      case AtomicOp.VariantIs(label) => ??? // TODO: Ext-Variants
```
JonathanStarup, line 694: Gasp, adding TODO's

main/src/ca/uwaterloo/flix/language/ast/AtomicOp.scala
magnus-madsen: Good point; fixed.

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
magnus-madsen: Yes, but what else? I cannot do everything in one PR.

JonathanStarup: I think the boilerplate is response is that we don't do `TODO`, they should be tickets instead

magnus-madsen:
@JonathanStarup Approve and merge?

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
magnus-madsen: That is only for TODOs that will be left over for a long time.

JonathanStarup: hmm I'll keep that in my back pocket

JonathanStarup (approved):

PR #10500, opened 2025-05-07
feat: add `TypeConstructor.Extensible`

magnus-madsen:


main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -365,6 +365,10 @@ object Simplifier {
             // The row types themselves return monotype records, so we do nothing here.
             visitType(elm)
 
+          case TypeConstructor.Extensible =>
+            // TODO: Monomorphization?
```
JonathanStarup, line 369: two TODO's seem redundant

main/src/ca/uwaterloo/flix/language/fmt/SimpleType.scala
```diff
@@ -439,6 +439,21 @@ object SimpleType {
             case _ :: _ :: _ => throw new OverAppliedType(t.loc)
           }
 
+        case TypeConstructor.Extensible =>
+          val args = t.typeArguments.map(visit)
```
JonathanStarup, line 443: Isn't it better to make this TODO instead of mixing it with Schemas? Now they're displayed the same and can potentially muddle together

magnus-madsen: I made it into a TODO

JonathanStarup (approved):

PR #10501, opened 2025-05-07
refactor: add toggle between old and new engine

CasperDN:


magnus-madsen:
This PR is too big for my taste. How about you leave Solver.flix untouched. Then you add `SolverApi` and make a few changes in `Lowering` to call `SolverApi`. This way it is clear that Solver.flix is not changed.

magnus-madsen (approved):

magnus-madsen:
What's next?

PR #10504, opened 2025-05-07
feat: use contexts in `Inliner`

jaschdoc:


magnus-madsen:
What does this do in words?

jaschdoc:
This just uses the `ExprContext`s at the right position. This will later be used for inlining heuristics but we are at point now where I think we will use them in almost every "important" PR.

magnus-madsen:
But what is it?

jaschdoc:
It models the evaluation context from which the inliner may ask questions that it uses for deciding whether or not to inline a function

magnus-madsen:
I think we can close this?

PR #10509, opened 2025-05-07
refactor: import `MonoAst.Expr` in `OccurrenceAnalyzer`

jaschdoc:


magnus-madsen:
Same treatment for Inliner?

PR #10510, opened 2025-05-07
feat: exit `Optimizer` loop on zero change

jaschdoc:


magnus-madsen:
I would actually suggest to do the structure:

```
for (...)
  if (delta.nonEmpty) {
     ...
   }
```

magnus-madsen (approved):

PR #10517, opened 2025-05-08
feat: consider local def as direct call in `Inliner`

jaschdoc:
This change makes it so that the inliner accounts for the fact that local defs are lifted at a later stage. Thus, it is okay to inline a function that simply defines a local def and immediately invokes it.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
@@ -532,8 +532,9 @@ object Inliner {
     * Returns `true` if `exp0` is a function call with trivial arguments.
     */
   private def isDirectCall(exp0: MonoAst.Expr): Boolean = exp0 match {
-    case MonoAst.Expr.ApplyDef(_, exps, _, _, _, _) => exps.forall(isTrivial)
-    case MonoAst.Expr.ApplyClo(exp1, exp2, _, _, _) => isTrivial(exp1) && isTrivial(exp2)
+    case Expr.ApplyDef(_, exps, _, _, _, _) => exps.forall(isTrivial)
+    case Expr.ApplyClo(exp1, exp2, _, _, _) => isTrivial(exp1) && isTrivial(exp2)
+    case Expr.LocalDef(_, _, _, Expr.ApplyLocalDef(_, exps, _, _, _), _, _, _, _) => exps.forall(isTrivial)
     case _ => false
   }
 


jaschdoc:
This change yields nearly identical ASTs for benchmarks `InlinerMap10K` and `InlinerMap10KOptimized` (the manually optimized version)

magnus-madsen:
Not to be a major pain in the butt, but I think going forward from around here that these PRs should be small (which this is- excellent!) but also come with performance benchmarks (at least for the compiler side), so we don't have to rely on arewefast to tell us that we made the compiler much slower by accident.

jaschdoc:
The raw running time benchmarks in. Basically, `List.map`, `List.length` and `List.reverse` see a big speedup and the rest is as before.

```json
{
  "master": [
    {
      "name": "InlinerFilterMap10k",
      "time": 62985875
    },
    {
      "name": "InlinerFilterMap10kOptimized",
      "time": 46942291
    },
    {
      "name": "InlinerMap10k",
      "time": 82961625
    },
    {
      "name": "InlinerMap10kOptimized",
      "time": 54462834
    },
    {
      "name": "List.filter",
      "time": 63925750
    },
    {
      "name": "List.filterMap",
      "time": 68993416
    },
    {
      "name": "List.foldLeft",
      "time": 42564709
    },
    {
      "name": "List.foldRight",
      "time": 80980291
    },
    {
      "name": "List.length",
      "time": 40761125
    },
    {
      "name": "List.map",
      "time": 62772834
    },
    {
      "name": "List.reverse",
      "time": 50265791
    }
  ],
  "this_branch": [
    {
      "name": "InlinerFilterMap10k",
      "time": 53096709
    },
    {
      "name": "InlinerFilterMap10kOptimized",
      "time": 48189333
    },
    {
      "name": "InlinerMap10k",
      "time": 56197042
    },
    {
      "name": "InlinerMap10kOptimized",
      "time": 54800333
    },
    {
      "name": "List.filter",
      "time": 62987916
    },
    {
      "name": "List.filterMap",
      "time": 68547041
    },
    {
      "name": "List.foldLeft",
      "time": 35430625
    },
    {
      "name": "List.foldRight",
      "time": 74974625
    },
    {
      "name": "List.length",
      "time": 20397375
    },
    {
      "name": "List.map",
      "time": 46996416
    },
    {
      "name": "List.reverse",
      "time": 28895583
    }
  ]
}
```

jaschdoc:
A slight dip in compiler throughput

`master`

```

Throughput (best): 80,213 lines/sec (with 20 threads.)

  min: 60,205, max: 80,213, avg: 70,076, median: 70,565

Finished 100 iterations on 67,496 lines of code in 96 seconds.
```

This pr

```

Throughput (best): 76,877 lines/sec (with 20 threads.)

  min: 52,584, max: 76,877, avg: 67,684, median: 67,569

Finished 100 iterations on 67,496 lines of code in 100 seconds.
```

magnus-madsen (approved):

magnus-madsen:
This seems to have been expensive on arewefast. We may want to pull it back a bit.

But lets see what happens when we get the new isSimple etc.

magnus-madsen:
I actually don't understand why we would do this. It would seem to only lead to code-bloat.

We like to inline when there is a HOF involved or the expression is trivial. But that is not what is happening here.

I think it should probably be undone.

magnus-madsen:
I take some of that back. It looks like there is a benefit to performance. It is just very costly in compiler performance.

PR #10521, opened 2025-05-08
refactor: tweak `Optimizer`

magnus-madsen:


magnus-madsen:
- I determine that 5 is both affordable and has an impact (so let us at least increase it for the moment)
- I tweaked the code style

main/src/ca/uwaterloo/flix/language/phase/optimizer/Optimizer.scala
```diff
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2025 Jakob Schneider Villumsen
```
jaschdoc, line 2: thanks

jaschdoc (approved):

PR #10523, opened 2025-05-08
feat: add abstract ram data-type for the new engine

CasperDN:


magnus-madsen:
@JonathanStarup I will let you do a quick pass. Maybe there is not much to say.

JonathanStarup:
I'm lacking context, what is this?

CasperDN:
It is an abstract description of a Datalog-program which the engine uses internally.

main/src/library/Fixpoint3/Ram.flix
```diff
@@ -0,0 +1,493 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+mod Fixpoint3.Ast.Ram {
+    // We have not updated these imports to be for Fixpoint3 as we have not changed any of the files
```
JonathanStarup, line 20: This seems like it shouldn't be here. But is this just code that is moved or?

CasperDN: It's just an explanation of why we use Fixpoint._ instead of Fixpoint3._, namely that the files are the identical. If we brought our identical copies over and used those, the solver would have to use an adaptor between the 2, as the types would be different. We here just refer to the old engine's definitions instead.

JonathanStarup: Oh so this file is new but you reuse some stuff? i see

main/src/library/Fixpoint3/Ram.flix
```diff
@@ -0,0 +1,493 @@
...
+    use Fixpoint.Boxed
+    use Fixpoint.PredSymsOf
+    use Fixpoint.SubstitutePredSym
+
+    /////////////////////////////////////////////////////////////////////////////
+    // RamProgram                                                              //
+    /////////////////////////////////////////////////////////////////////////////
+
+    ///
+    /// A `RamProgram` consists of a statements, the EDB, predicate information,
```
JonathanStarup, line 32: Grammar stuff. I think the types are good enough that you don't have to spell it out.

main/src/library/Fixpoint3/Ram.flix
```diff
@@ -0,0 +1,493 @@
...
+    /// and the information about the indexes to be constructed.
+    /// 
+    pub enum RamProgram {
+        case Program(RamStmt, Facts, Predicates, IndexInformation)
+    }
+
+    ///
+    /// `Predicate` represents 1 + the maximal indentifier of the full `RelSym`s of the program and set of `RelSym`
```
JonathanStarup, line 40: You mention a program here but there is no program in the type, so maybe just delete that part?

main/src/library/Fixpoint3/Ram.flix
```diff
@@ -0,0 +1,493 @@
...
+    /// The EDB of a program
+    ///
+    pub type alias Facts = Map[RelSym, Vector[Vector[Boxed]]]
+
+    ///
+    /// A `Search` describes the order in which elements are ordered.
+    /// [2, 0, 1] means that the elements are ordered by the third element,
+    /// then the first and finally the second.
+    ///
+    pub type alias Search = Vector[Int32]
```
JonathanStarup, line 54: Is there also some special meaning of Vector.empty here? if so, that should also be mentioned

main/src/library/Fixpoint3/Ram.flix
```diff
@@ -0,0 +1,493 @@
...
+
+    ///
+    /// `ConstructedIndexes` represents a map from list of indexes created for the RelSym.
+    ///
+    pub type alias ConstructedIndexes = Map[RelSym, Vector[Search]]
```
JonathanStarup, line 59: Isn't constructed self-evident? Could just be called Indexes

main/src/library/Fixpoint3/Ram.flix
```diff
@@ -0,0 +1,493 @@
...
+
+    ///
+    /// `IndexPositions` represents a map from RelSym and its place in the vector of ConstructedIndexes
+    /// to its final position in memory.
+    ///
+    pub type alias IndexPositions = Map[(RelSym, Int32), Int32]
```
JonathanStarup, line 65: I think its valid to do it as you have, but richer types (aliases or enums) would help readability of all these integer things

main/src/library/Fixpoint3/Ram.flix
```diff
@@ -0,0 +1,493 @@
...
+    ///
+    /// `IndexInformation` contains the combined information about indexes in a program
+    ///
+    pub type alias IndexInformation = (ConstructedIndexes, IndexPositions)
+
+    /////////////////////////////////////////////////////////////////////////////
+    // RamStmt                                                                 //
+    /////////////////////////////////////////////////////////////////////////////
+
+    ///
```
JonathanStarup, line 76: missing `.` on a couple of these sentences

main/src/library/Fixpoint3/Ram.flix
```diff
@@ -0,0 +1,493 @@
...
+    /// For multi-indexed relations the only guarantee is that all indexes on `deltaRel`
+    /// will contain all facts of `newRel`.
+    ///
+    /// `Purge(rel)` represents removing all facts from all indexes on `rel`.
+    ///
+    /// `Seq(body)` represents a sequence of statments to be executed sequentially
+    ///
+    /// `Par(body)` represents a sequence of statments to be executed in parallel
+    ///
+    /// `Comment(string)` represents a comment for debugging
```
JonathanStarup, line 91: Until is missing in the docs

main/src/library/Fixpoint3/Ram.flix
```diff
@@ -0,0 +1,493 @@
...
+    /// saving them as `rv` and execute `body` if `bools` is true. The index saved at `indexPos` will be used
+    ///
+    /// `Functional(rv, f, input, body, arity)` represents evaluating `f(input)`, for each resutl save it as `rv`
+    /// and execute `body`. `arity` is the size of the output.
+    ///
+    /// `Project(terms, relSym)` represents inserting `terms` as a fact for `relSym`.
+    /// `relSym` should be a new relation.
+    ///
+    /// `If(bools, body)` represents evaluating `body` if `bools` is true.
+
```
JonathanStarup, line 147: I think this line breaks the doc  comment away from the enum

main/src/library/Fixpoint3/Ram.flix
CasperDN: An empty `Search` would translate to all vectors being equal to each other. Except for `EstimateJoinSize` a `Search` will always have the same length as the Vector it defines order on as all orders must be total.

main/src/library/Fixpoint3/Ram.flix
```diff
@@ -0,0 +1,493 @@
...
+            match stmt {
+                case RamStmt.Insert(op) => ToString.toString(op)
+                case RamStmt.Merge(src, dst) => "merge ${src} into ${dst}"
+                case RamStmt.Swap(lhs, rhs) => "Swap ${lhs} and ${rhs}"
+                case RamStmt.Purge(relSym) => "purge ${relSym}"
+                case RamStmt.Seq(xs) => Vector.join(";${nl}", xs)
+                case RamStmt.Until(test, body) =>
+                    let tst = test |> Vector.join(" && ");
+                    "until(${tst}) do${nl}${String.indent(4, "${body}")}end"
+                case RamStmt.Comment(comment) => "// ${comment}"
```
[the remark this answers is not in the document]
CasperDN: `Seq` joins with `;newLine`. Every statement will get its own line so it should be fine?

main/src/library/Fixpoint3/Ram.flix
CasperDN: Yes. The structure is the same as the old, but we have changed/added stuff as needed.

magnus-madsen: Please remove the comment or change it so that it actually makes sense to an outsider.

I think you should just remove it.

main/src/library/Fixpoint3/Ram.flix
magnus-madsen: Please extend the comment.

Please also explain if some vectors are invalid.

main/src/library/Fixpoint3/Ram.flix
magnus-madsen: Agreed

main/src/library/Fixpoint3/Ram.flix
```diff
@@ -0,0 +1,496 @@
...
+    pub enum RamTerm {
+        case Lit(Boxed, RamId)
+        case RowLoad(RowVar, Int32, RelSym)
+        case Meet(Boxed -> Boxed -> Boxed, RamTerm, (RowVar, RelSym), RamId)
+        case App0(Unit -> Boxed, RamId)
+        case App1(Boxed -> Boxed, RamTerm, RamId)
+        case App2(Boxed -> Boxed -> Boxed, RamTerm, RamTerm, RamId)
+        case App3(Boxed -> Boxed -> Boxed -> Boxed, RamTerm, RamTerm, RamTerm, RamId)
+        case App4(Boxed -> Boxed -> Boxed -> Boxed -> Boxed, RamTerm, RamTerm, RamTerm, RamTerm, RamId)
+        case App5(Boxed -> Boxed -> Boxed -> Boxed -> Boxed -> Boxed, RamTerm, RamTerm, RamTerm, RamTerm, RamTerm, RamId)
```
magnus-madsen, line 279: I think it would be more sensible if RamId was the first argument?

main/src/library/Fixpoint3/Ram.flix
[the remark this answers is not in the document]
JonathanStarup: Ah yes, my bad, the code is correct. You could still consider it, since it would make future code less likely to introduce errors but up to you

main/src/library/Fixpoint3/Ram.flix
magnus-madsen: Ok

magnus-madsen (approved):

PR #10529, opened 2025-05-09
feat: require arity for `inject` expressions

mlutze:
depends on 
- https://github.com/flix/flix/pull/10812

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -2903,9 +2903,9 @@ object Parser2 {
         expression()
       }
       expect(TokenKind.KeywordInto)
-      nameUnqualified(NAME_PREDICATE)
+      predicateAndArity()
       while (eat(TokenKind.Comma) && !eof()) {
-        nameUnqualified(NAME_PREDICATE)
+        predicateAndArity()
       }
       close(mark, TreeKind.Expr.FixpointInject)
     }
@@ -2968,6 +2968,22 @@ object Parser2 {
       expect(TokenKind.KeywordWhere)
       expression()
       close(mark, TreeKind.Expr.FixpointWhere)
+    }
+
+    private def predicateAndArity()(implicit s: State): Mark.Closed = {
+      implicit val sctx: SyntacticContext = SyntacticContext.Expr.OtherExpr
+      val mark = open()
+      nameUnqualified(NAME_PREDICATE)
+
+      val hint = "provide a predicate arity such as: Pred/1"
+      // check for shape "/2"
+      expect(TokenKind.Slash, hint = Some(hint))
+      advance()
+
+      expect(TokenKind.LiteralInt, hint = Some(hint))
+      advance()
+
+      close(mark, TreeKind.PredicateAndArity)
     }
 
     private def intrinsicExpr()(implicit s: State): Mark.Closed = {
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -2978,10 +2978,7 @@ object Parser2 {
       val hint = "provide a predicate arity such as: Pred/1"
       // check for shape "/2"
       expect(TokenKind.Slash, hint = Some(hint))
-      advance()
-
       expect(TokenKind.LiteralInt, hint = Some(hint))
-      advance()
 
       close(mark, TreeKind.PredicateAndArity)
     }
--- a/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
@@ -2088,17 +2088,19 @@ object Weeder2 {
 
     private def visitFixpointInjectExpr(tree: Tree)(implicit sctx: SharedContext): Validation[Expr, CompilationMessage] = {
       expect(tree, TreeKind.Expr.FixpointInject)
-      val expressions = pickAll(TreeKind.Expr.Expr, tree)
-      val idents = pickAll(TreeKind.Ident, tree).map(tokenToIdent)
-      mapN(traverse(expressions)(visitExpr)) {
-        case exprs if exprs.length != idents.length =>
+      val expTrees = pickAll(TreeKind.Expr.Expr, tree)
+      val predAndArityTrees = pickAll(TreeKind.PredicateAndArity, tree)
+      val expsVal = traverse(expTrees)(visitExpr)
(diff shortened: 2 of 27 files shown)


main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -2889,6 +2889,36 @@ object Parser2 {
       close(mark, TreeKind.Expr.FixpointWhere)
     }
 
+    private def predicateShape()(implicit s: State): Mark.Closed = {
```
JonathanStarup, line 2892: maybe predicateWithShape or predicateAndShape?

JonathanStarup: shape to me sounds like the `(_, _, ..)` part, but fair

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -2889,6 +2889,36 @@ object Parser2 {
...
+      val mark = open()
+      nameUnqualified(NAME_PREDICATE, SyntacticContext.Expr.OtherExpr)
+
+      // check for shape (_, _, _ ...)
+      if (!at(TokenKind.ParenL)) {
+        closeWithError(open(), UnexpectedToken(
+          expected = NamedTokenSet.FromKinds(Set(TokenKind.ParenL)),
+          actual = Some(nth(0)),
+          sctx = SyntacticContext.Expr.OtherExpr,
+          hint = Some("provide a predicate shape."),
```
JonathanStarup, line 2902: could the hint include the syntax? like `provide a predicate shape like '(_, _)'`

JonathanStarup (approved):

magnus-madsen:
I thought we also discussed `Edge/2` etc. But in any case, the next step in this PR should be to "prove" that you can get rid of Type.mkTuplish?

mlutze:
> I thought we also discussed `Edge/2` etc.

Jonathan said `(_, _, ...)` is prettier.


> But in any case, the next step in this PR should be to "prove" that you can get rid of Type.mkTuplish?

Proof of concept here: https://github.com/flix/flix/pull/10532

JonathanStarup:
I think that there is no flix-like way to have a number here. `Pred/N`, `Pred-N`, etc would be novel syntax. `(_, _, ..)` reads nicely and is in the language of Flix syntax. Typing is a bit verbose for more than 2 things though yeah.

magnus-madsen:
`Pred/X` is actually used in the hoverprovider. It is also standard from logic programming.

JonathanStarup:
It doesn't contradict that this isn't used in flix anywhere. but it does allow less keystrokes yeah.

mlutze:
## Consistent syntax
Flix aims to have consistent and predictable syntax. As an example, we try to have the syntax of types mirror that of expressions

magnus-madsen:
TODO:
- Discuss syntax
- Verify that this solves tuplish
- Brainstorm alternatives and conclude they are all horrible
- Talk to Ondrej

Lastly, if approved:
- PRs for book
- PRs for website
- Update playground
- Update slides

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
magnus-madsen: predicateWithArity

mlutze:
@JonathanStarup aled

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -2968,6 +2969,40 @@ object Parser2 {
       close(mark, TreeKind.Expr.FixpointWhere)
     }
 
+    private def predicateAndArity()(implicit s: State): Mark.Closed = {
+      implicit val sctx: SyntacticContext = SyntacticContext.Expr.OtherExpr
+      val mark = open()
+      nameUnqualified(NAME_PREDICATE)
+
+      // check for shape "/2"
+      if (!at(TokenKind.Slash)) {
```
JonathanStarup: Is't this the same as `Expect(TokenKind.Slash, hint = ...)` ?

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -3364,6 +3377,20 @@ object Weeder2 {
...
 
+  /**
+    * Picks out the first token of a specific [[TokenKind]].
+    */
+  private def pickToken(kind: TokenKind, tree: Tree, synctx: SyntacticContext = SyntacticContext.Unknown): Validation[Token, CompilationMessage] = {
+    tree.children.collectFirst {
+      case token: Token if token.kind == kind => token
+    } match {
+      case Some(t) => Validation.Success(t)
+      case _ =>
```
JonathanStarup, line 3398: This is technically a duplicate error. If parer doesn't give you a LiteralInt then it has already thrown an error, but this kind of parser<->weeder dependency is also not nice so maybe its fine

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -3247,6 +3249,16 @@ object Weeder2 {
     }
   }
 
+  private def visitPredicateAndArity(tree: Tree)(implicit sctx: SharedContext): Validation[PredicateAndArity, CompilationMessage] = {
+    val identVal = pickNameIdent(tree)
+    val arityTokenVal = pickToken(TokenKind.LiteralInt, tree)
+    mapN(identVal, arityTokenVal) {
+      case (ident, arityToken) =>
+        val arity = arityToken.text.toInt
```
JonathanStarup: I have unfortunate news 🧑‍⚕️

You use LiteralInt, which both allows `0x1` and `4_2` or a combination thereof. But this will crash `toInt`. You can just call `tryParseInt` in weeder if you want to handle these cases

magnus-madsen:
Do we need any additional tests?

JonathanStarup:
> Do we need any additional tests?

We have zero tests for `inject` weirdly enough, but I think its unfair to put that burden on Matt. This PR is still strictly better without any tests, it'll just maintain the lack of tests

magnus-madsen:
> > Do we need any additional tests?
> 
> We have zero tests for `inject` weirdly enough, but I think its unfair to put that burden on Matt. This PR is still strictly better without any tests, it'll just maintain the lack of tests

OOof

magnus-madsen:
Well, Jakob can write some tests. I cannot merge this before #10756 anyway

magnus-madsen:
Which arity PR is the right one? And is it ready for review?

mlutze:
Both are the right one. Both are ready for review/merge.

This one only adds the syntax and throws it away.

The other one builds on this one and adds the arity up to typing.

magnus-madsen (approved):

PR #10535, opened 2025-05-09
fix: do copy propagation on simple exps

jaschdoc:


magnus-madsen:
Was this even working? We saw those variables that were not propagated?

jaschdoc:
Yes,

```flix
def main(): (Option[Int32], Option[Option[Int32]], Option[Option[Int32]]) =
    let x = Some(1);
    let y = Some(x);
    (x, y, y)
```
now becomes
```

def main(
    _unit$131794: Unit
): (Option[Int32], Option[Option[Int32]], Option[Option[Int32]]) = {
    (
        Option.Some(1i32),
        Option.Some(Option.Some(1i32)),
        Option.Some(Option.Some(1i32))
    )
}
```

magnus-madsen (approved):

PR #10536, opened 2025-05-09
refactor: rework `bug!`

magnus-madsen:


magnus-madsen:
Interesting example where Monomorphization copies too much monomorphic code.

magnus-madsen:
@mlutze @JonathanStarup I wonder if this could be a small research idea.

Note that writing a local letdef does not help. Monomorph will still duplicate it.

main/src/library/Prelude.flix
```diff
@@ -206,15 +206,17 @@ pub def touch(_: Region[r]): Unit \ r = checked_ecast(())
 ///
 /// Crashes the current process with the given message `m`.
 ///
-pub def bug!(m: String): a = unchecked_cast({
+pub def bug!(m: String): a = unchecked_cast({ printBugMsg(m); ?panic } as _ \ {})
+
+@Internal
+pub def printBugMsg(m: String): Unit \ IO = {
```
JonathanStarup, line 212: This can just be private right?

jaschdoc (approved):

main/src/library/Prelude.flix
magnus-madsen: We are at the prelude

JonathanStarup: Right, would have to add a mod to hide it

PR #10537, opened 2025-05-09
refactor: cleanup `isSingleAction`

magnus-madsen:


magnus-madsen:
Added Index and Untag

jaschdoc:
Untag is not created until simplifier so that is dead code

magnus-madsen:
> Untag is not created until simplifier so that is dead code

Ah. Hmm

magnus-madsen:
Neither is index.

jaschdoc (approved):

Issue #10540, opened 2025-05-09
Optimizer: Compute reachable functions in a SharedContext in OA


magnus-madsen:
And see if it can speedup the Optimizer.

magnus-madsen:
Perhaps we should call these live functions instead of reachable.

PR #10542, opened 2025-05-09
feat: compute live defs in `OccurrenceAnalyzer`

jaschdoc:
Closes https://github.com/flix/flix/issues/10540


magnus-madsen:
Returning the right/updated AST root should make this just work out of the box, right?

magnus-madsen:
Then you can report results with `Xperf --n 50`

Issue #10373, opened 2025-04-25
Optimizer: Introduce `@Inline` annotation


magnus-madsen:
This annotation is only for debugging and testing. We will decide later if it should become a language feature.

Note: `@Inline` on a recursive function is allowed and just unroll it once.

magnus-madsen:
We may also want a `@DontInline` annotation.

PR #10546, opened 2025-05-09
feat: add `@Inline` and `@DontInline` annotations

jaschdoc:
Maybe the `inlined` variable should be copied over to `sctx` at the beginning of `Inliner.run`.

Closes https://github.com/flix/flix/issues/10373

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/shared/Annotation.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/shared/Annotation.scala
@@ -154,7 +154,7 @@ object Annotation {
     * An AST node that represents an `@Inline` annotation.
     *
     * A function marked with `@Inline` is guaranteed to be inlined by the compiler.
-    * If it is recursive, it is only ever unrolled once.
+    * If it is recursive, it is only ever unrolled once into the same function.
     *
     * @param loc the source location of the annotation.
     */
--- a/main/src/ca/uwaterloo/flix/language/ast/shared/Annotation.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/shared/Annotation.scala
@@ -68,7 +68,8 @@ object Annotation {
     * An AST node that represents an `@Inline` annotation.
     *
     * A function marked with `@Inline` is guaranteed to be inlined by the compiler.
-    * If it is recursive, it is only ever unrolled once into the same function.
+    * If it is recursive, it may be unrolled any number of times but never inside
+    * its own definition.
     *
     * @param loc the source location of the annotation.
     */
--- a/main/src/ca/uwaterloo/flix/language/ast/shared/Annotation.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/shared/Annotation.scala
@@ -150,4 +150,27 @@ object Annotation {
     override def toString: String = "@" + name
   }
 
+  /**
+    * An AST node that represents an `@Inline` annotation.
+    *
+    * A function marked with `@Inline` is guaranteed to be inlined by the compiler.
+    * If it is recursive, it is only ever unrolled once.
+    *
+    * @param loc the source location of the annotation.
+    */
+  case class Inline(loc: SourceLocation) extends Annotation {
+    override def toString: String = "@Inline"
+  }
+
+  /**
+    * An AST node that represents a `@DontInline` annotation.
+    *
+    * A function marked with `@DontInline` is guaranteed to never be inlined by the compiler.
+    *
+    * @param loc the source location of the annotation.
+    */
+  case class DontInline(loc: SourceLocation) extends Annotation {
+    override def toString: String = "@DontInline"
+  }
+
 }
--- a/main/src/ca/uwaterloo/flix/language/ast/shared/Annotation.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/shared/Annotation.scala
@@ -36,6 +36,17 @@ object Annotation {
   }
 
   /**
+    * An AST node that represents a `@DontInline` annotation.
+    *
+    * A function marked with `@DontInline` is guaranteed to never be inlined by the compiler.
+    *
+    * @param loc the source location of the annotation.
+    */
+  case class DontInline(loc: SourceLocation) extends Annotation {
+    override def toString: String = "@DontInline"
+  }
+
+  /**
     * An annotation that marks a construct as experimental.
     *
     * @param loc the source location of the annotation.
@@ -51,6 +62,18 @@ object Annotation {
     */
   case class Export(loc: SourceLocation) extends Annotation {
     override def toString: String = "@Export"
+  }
+
+  /**
+    * An AST node that represents an `@Inline` annotation.
+    *
+    * A function marked with `@Inline` is guaranteed to be inlined by the compiler.
+    * If it is recursive, it is only ever unrolled once into the same function.
+    *
+    * @param loc the source location of the annotation.
+    */
+  case class Inline(loc: SourceLocation) extends Annotation {
+    override def toString: String = "@Inline"
   }
 
   /**
@@ -148,29 +171,6 @@ object Annotation {
     */
   case class Error(name: String, loc: SourceLocation) extends Annotation {
     override def toString: String = "@" + name
-  }
-
-  /**
-    * An AST node that represents an `@Inline` annotation.
-    *
-    * A function marked with `@Inline` is guaranteed to be inlined by the compiler.
-    * If it is recursive, it is only ever unrolled once into the same function.
-    *
-    * @param loc the source location of the annotation.
-    */
-  case class Inline(loc: SourceLocation) extends Annotation {
-    override def toString: String = "@Inline"
-  }
-
-  /**
-    * An AST node that represents a `@DontInline` annotation.
-    *
-    * A function marked with `@DontInline` is guaranteed to never be inlined by the compiler.
-    *
-    * @param loc the source location of the annotation.
-    */
-  case class DontInline(loc: SourceLocation) extends Annotation {
-    override def toString: String = "@DontInline"
   }
 
 }
--- a/main/src/ca/uwaterloo/flix/language/ast/shared/Annotations.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/shared/Annotations.scala
@@ -37,6 +37,11 @@ case class Annotations(annotations: List[Annotation]) {
   def isDeprecated: Boolean = annotations exists (_.isInstanceOf[Annotation.Deprecated])
 
   /**
+    * Returns `true` if `this` sequence contains the `@DontInline` annotation.
+    */
+  def isDontInline: Boolean = annotations.exists(_.isInstanceOf[Annotation.DontInline])
+
+  /**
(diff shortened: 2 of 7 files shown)


main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -497,8 +501,11 @@ object Inliner {
     * @param exps the arguments to the function.
     * @param ctx0 the local context.
     */
-  private def shouldInlineDef(defn: MonoAst.Def, exps: List[Expr], ctx0: LocalContext): Boolean = {
-    !ctx0.currentlyInlining && !defn.spec.defContext.isSelfRef &&
+  private def shouldInlineDef(defn: MonoAst.Def, exps: List[Expr], ctx0: LocalContext)(implicit sym0: Symbol.DefnSym, inlined: ConcurrentHashMap[Symbol.DefnSym, ConcurrentHashMap[Symbol.DefnSym, Unit]]): Boolean = {
+    val isInlinedHere = inlined.getOrDefault(defn.sym, new ConcurrentHashMap()).containsKey(sym0)
+    !defn.spec.ann.isDontInline &&
+      ((defn.spec.ann.isInline && defn.spec.defContext.isSelfRef && !isInlinedHere) || !defn.spec.defContext.isSelfRef || (defn.spec.ann.isInline && !defn.spec.defContext.isSelfRef)) &&
+      !ctx0.currentlyInlining &&
```
jaschdoc, line 508: This is btw not atomic so a race condition could inline `defn` multiple places

magnus-madsen: Why do we need the map?

jaschdoc: We need to know if it has been unrolled once already in `sym0`

jaschdoc: The new solution is somewhat verbose but that's the nature of java and null checks I guess...

jaschdoc:
For some really weird reason it doesn't print the `Optimizer` AST anymore. Thoughts?

magnus-madsen:
> For some really weird reason it doesn't print the `Optimizer` AST anymore. Thoughts?

No idea.

magnus-madsen:
I was actually thinking something much simpler here. Putting `@Inline` should just mean that our criteria are disabled.

We should not worry about only unfolding or whatever.

(Let me also clarify that I am not necessarily adding these annotations to the language--for now they are just to help us).

eyestorefivem:
Very clean code with good test coverage. This is a great example of best practices.

fivemservereyestore:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
```diff
@@ -632,6 +632,25 @@ object WeederError {
     override def explain(formatter: Formatter): Option[String] = None
   }
 
+  case class InlineAndDontInline(inlineLoc: SourceLocation, dontInlineLoc: SourceLocation) extends WeederError {
```
magnus-madsen, line 635: Add test case

jaschdoc:
Contains the change from https://github.com/flix/flix/pull/10567 since the bug breaks inlining here.

main/src/ca/uwaterloo/flix/language/ast/shared/Annotation.scala
```diff
@@ -150,4 +150,27 @@ object Annotation {
     override def toString: String = "@" + name
   }
 
+  /**
```
magnus-madsen, line 153: The annotations are kept sorted (it seems)

main/src/ca/uwaterloo/flix/language/ast/shared/Annotation.scala
```diff
@@ -150,4 +150,27 @@ object Annotation {
...
+    * An AST node that represents an `@Inline` annotation.
+    *
+    * A function marked with `@Inline` is guaranteed to be inlined by the compiler.
+    * If it is recursive, it is only ever unrolled once into the same function.
```
magnus-madsen, line 157: -> If the function is recursive, it may be unrolled in undetermined number of times.

magnus-madsen: (Right-- because each round would unroll it once?)

main/src/ca/uwaterloo/flix/language/ast/shared/Annotations.scala
```diff
@@ -85,4 +85,16 @@ case class Annotations(annotations: List[Annotation]) {
     * Returns `true` if `this` sequence contains the `@Test` annotation.
     */
   def isTest: Boolean = annotations exists (_.isInstanceOf[Annotation.Test])
+
+
+  /**
```
magnus-madsen, line 90: Same comment about order

main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
magnus-madsen, line 635: Docs

main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
```diff
@@ -632,6 +632,25 @@ object WeederError {
...
+    override def summary: String = "Annotated with both `@Inline` and `@DontInline`"
+
+    override def message(formatter: Formatter): String = {
+      import formatter.*
+      s""">> Annotated with both `@Inline` and `@DontInline`.
```
magnus-madsen, line 640: -> A def cannot be marked both ... 

(same change above)

main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
```diff
@@ -632,6 +632,25 @@ object WeederError {
...
+         |${code(inlineLoc, "the `@Inline` occurs here")}
+         |
+         |${code(dontInlineLoc, "the `@DontInline` occurs here")}
+         |
+         |""".stripMargin
+    }
+
+    override def explain(formatter: Formatter): Option[String] = None
+
+    override def loc: SourceLocation = inlineLoc
```
magnus-madsen, line 651: Min of the two locs gives us the first occurrence

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -668,6 +684,8 @@ object Weeder2 {
         case "@Skip" => Skip(loc)
         case "@Test" | "@test" => Test(loc)
         case "@TailRec" => TailRecursive(loc)
+        case "@Inline" => Inline(loc)
```
magnus-madsen, line 687: sorted

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -492,13 +493,49 @@ object Inliner {
...
+      }
+
+      return true
+    }
+
+    !defn.spec.defContext.isSelfRef && !ctx0.currentlyInlining &&
       (isSingleAction(defn.exp) || isSimple(defn.exp) || hasKnownLambda(exps))
   }
 
+  private def atomicallyCheckShouldInlineRecursive(defn: MonoAst.Def)(implicit sym0: Symbol.DefnSym, sctx: SharedContext): Boolean = {
```
magnus-madsen, line 521: What is all this? I just want it to inline.

magnus-madsen: `@Inline` should simply mean: always inline when possible - i.e. disable any checks.

jaschdoc: We still only want to unroll a recursive function once at a call site.

magnus-madsen: Not something that requires more complexity.

jaschdoc:
The program
```flix
def main(): List[Int32] = {
    mapper(1 :: 2 :: 3 :: Nil)
}

@Inline
def mapper(l: List[Int32]): List[Int32] = match l {
    case Nil     => Nil
    case x :: xs => id(x) :: mapper(xs)
}

@DontInline
def id(x: a): a = x
```
gives
```flix
def id$135976(x$136070: Int32): Int32 = {x$136070}

def main(_unit$135952: Unit): List[Int32] = {
    match List.Cons(1i32, List.Cons(2i32, List.Cons(3i32, List.Nil))) {
        case List.Nil =>
            List.Nil
        case List.Cons(x$136739, xs$136740) =>
            List.Cons(
                id$135976(x$136739),
                match xs$136740 {
                    case List.Nil =>
                        List.Nil
                    case List.Cons(x$136750, xs$136751) =>
                        List.Cons(
                            id$135976(x$136750),
                            match xs$136751 {
                                case List.Nil =>
                                    List.Nil
                                case List.Cons(x$136754, xs$136755) =>
                                    List.Cons(
                                        id$135976(x$136754),
                                        match xs$136755 {
                                            case List.Nil =>
                                                List.Nil
                                            case List.Cons(
                                                x$136758,
                                                xs$136759
                                            ) =>
                                                List.Cons(
                                                    id$135976(x$136758),
                                                    mapper(xs$136759)
                                                )
                                        }
                                    )
                            }
                        )
                }
            )
    }
}

def main$(_unit$135949: Unit): Unit \ IO = {println$135965(main(()))}

def mapper(l$135950: List[Int32]): List[Int32] = {
    match l$135950 {
        case List.Nil =>
            List.Nil
        case List.Cons(x$136708, xs$136710) =>
            List.Cons(id$135976(x$136708), mapper(xs$136710))
    }
}
```

main/src/ca/uwaterloo/flix/language/ast/shared/Annotation.scala
```diff
@@ -53,6 +64,18 @@ object Annotation {
     override def toString: String = "@Export"
   }
 
+  /**
+    * An AST node that represents an `@Inline` annotation.
+    *
+    * A function marked with `@Inline` is guaranteed to be inlined by the compiler.
+    * If it is recursive, it is only ever unrolled once into the same function.
```
magnus-madsen, line 71: The latter statement is no longer true.

main/src/ca/uwaterloo/flix/language/phase/optimizer/Optimizer.scala
```diff
@@ -17,8 +17,11 @@ package ca.uwaterloo.flix.language.phase.optimizer
 
 import ca.uwaterloo.flix.api.Flix
 import ca.uwaterloo.flix.language.ast.MonoAst
+import ca.uwaterloo.flix.language.ast.Symbol
 import ca.uwaterloo.flix.language.dbg.AstPrinter.DebugMonoAst
 
+import java.util.concurrent.ConcurrentHashMap
```
magnus-madsen, line 23: Old changes

magnus-madsen (approved):

PR #10548, opened 2025-05-09
refactor: simplify and improve `Inliner`

magnus-madsen:


magnus-madsen:
@jaschdoc This PR:

- Removes unused code. (Let me know if there is more unused code that can be removed).*
- It adds a small trick to the inliner. The idea is that we do not inline an ApplyDef until the symbol did not change in the last round. The idea is for defs to stabilize before they get inlined. Note that this may require more rounds to achieve the same effect as the old code. But I did not change the amount of rounds because the results seemed to be 99% similar. 

* I am not saying we won't need this code. Just that we should do clean PRs ontop of a clean baseline.

mlutze (approved):

PR #10563, opened 2025-05-10
refactor: remove redundant call to `isTrivial` in `Inliner`

jaschdoc:


main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -579,10 +579,10 @@ object Inliner {
...
     case Expr.ApplyAtomic(AtomicOp.Unary(_), exps, _, _, _) => exps.forall(isTrivial)
     case Expr.ApplyAtomic(AtomicOp.Binary(_), exps, _, _, _) => exps.forall(isTrivial)
-    case Expr.ApplyAtomic(AtomicOp.Tag(_), exps, _, _, _) => exps.forall(e => isTrivial(e) || isSimple(e))
-    case Expr.ApplyAtomic(AtomicOp.Tuple, exps, _, _, _) => exps.forall(e => isTrivial(e) || isSimple(e))
-    case Expr.ApplyAtomic(AtomicOp.ArrayLit, exps, _, _, _) => exps.forall(isTrivial)
-    case Expr.ApplyAtomic(AtomicOp.StructNew(_, _), exps, _, _, _) => exps.forall(isTrivial)
+    case Expr.ApplyAtomic(AtomicOp.Tag(_), exps, _, _, _) => exps.forall(isSimple)
+    case Expr.ApplyAtomic(AtomicOp.Tuple, exps, _, _, _) => exps.forall(isSimple)
+    case Expr.ApplyAtomic(AtomicOp.ArrayLit, exps, _, _, _) => exps.forall(isSimple)
+    case Expr.ApplyAtomic(AtomicOp.StructNew(_, _), exps, _, _, _) => exps.forall(isSimple)
```
magnus-madsen, line 585: I would argue these should all be isTrivial. The larger pattern is not generated by the compiler and could be very big.

jaschdoc:
Ok now it is just a pure refactor :)

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -579,8 +579,8 @@ object Inliner {
     case Expr.Lambda(_, _, _, _) => true
     case Expr.ApplyAtomic(AtomicOp.Unary(_), exps, _, _, _) => exps.forall(isTrivial)
     case Expr.ApplyAtomic(AtomicOp.Binary(_), exps, _, _, _) => exps.forall(isTrivial)
-    case Expr.ApplyAtomic(AtomicOp.Tag(_), exps, _, _, _) => exps.forall(e => isTrivial(e) || isSimple(e))
-    case Expr.ApplyAtomic(AtomicOp.Tuple, exps, _, _, _) => exps.forall(e => isTrivial(e) || isSimple(e))
+    case Expr.ApplyAtomic(AtomicOp.Tag(_), exps, _, _, _) => exps.forall(isSimple)
+    case Expr.ApplyAtomic(AtomicOp.Tuple, exps, _, _, _) => exps.forall(isSimple)
     case Expr.ApplyAtomic(AtomicOp.ArrayLit, exps, _, _, _) => exps.forall(isTrivial)
     case Expr.ApplyAtomic(AtomicOp.StructNew(_, _), exps, _, _, _) => exps.forall(isTrivial)
     case Expr.Cast(exp, _, _, _) => isSimple(exp)
```
magnus-madsen, line 586: I have a meta comment here. What makes an inlining good? I hope we agree that it is good if:

- Programs run faster (yay!)
- Code size is smaller
- Compiler runs faster

Now, we might say it makes sense to inline nested tags. But I am not so sure. For example, if a function constructs a 10 element list do we want to inline it? If we have deforestation I can see the answer being a big YES. But if we don't (and master does not) then I would say no. Inlining a bit list literal bloats the code, but does lead to speedsups. In particular, the overhead of one call compared to constructing the big list is negligible. 

You might say, why then inline one tag at all? Well, here we are targeting program patterns we know the compiler generates. We surely don't want to create X functions that all return just Nil-- but that can happen. In other words, the main purpose of isSimple is to inline these very shallow functions.

TLDR: I would require all arguments to be trivial.

magnus-madsen: I think we are also a point where we have the performance we want, but now we wish for the code to be smaller and for the compiler to run faster. So what changes can we make that don't compromise performance?

magnus-madsen (approved):

Issue #10283, opened 2025-04-11
Unify IntelliJ style document


mlutze:
Followup to https://github.com/flix/flix/issues/10093

We previously unified the warnings we get in IntelliJ. We can do the same for the autoformatter.

mlutze:
@magnus-madsen @JonathanStarup @jaschdoc @stormckey 

Please post your IntelliJ Scala format file. It should be at:

`./idea/codeStyles/Project.xml`

mlutze:
```xml
<component name="ProjectCodeStyleConfiguration">
  <code_scheme name="Project" version="173">
    <ScalaCodeStyleSettings>
      <option name="USE_SCALADOC2_FORMATTING" value="true" />
      <option name="MULTILINE_STRING_CLOSING_QUOTES_ON_NEW_LINE" value="true" />
    </ScalaCodeStyleSettings>
  </code_scheme>
</component>
```

magnus-madsen:
```xml
<component name="ProjectCodeStyleConfiguration">
  <code_scheme name="Project" version="173">
    <ScalaCodeStyleSettings>
      <option name="USE_SCALADOC2_FORMATTING" value="true" />
    </ScalaCodeStyleSettings>
  </code_scheme>
</component>
```

stormckey:
```
<component name="ProjectCodeStyleConfiguration">
  <code_scheme name="Project" version="173">
    <ScalaCodeStyleSettings>
      <option name="MULTILINE_STRING_CLOSING_QUOTES_ON_NEW_LINE" value="true" />
    </ScalaCodeStyleSettings>
  </code_scheme>
</component>
```

jaschdoc:
```xml
<component name="ProjectCodeStyleConfiguration">
  <code_scheme name="Project" version="173">
    <ScalaCodeStyleSettings>
      <option name="MULTILINE_STRING_CLOSING_QUOTES_ON_NEW_LINE" value="true" />
    </ScalaCodeStyleSettings>
  </code_scheme>
</component>
```

JonathanStarup:
```
<component name="ProjectCodeStyleConfiguration">
  <code_scheme name="Project" version="173">
    <option name="RIGHT_MARGIN" value="100" />
    <option name="SOFT_MARGINS" value="80,100" />
    <ScalaCodeStyleSettings>
      <option name="USE_SCALADOC2_FORMATTING" value="true" />
[...]

PR #10571, opened 2025-05-12
chore: introduce global and apply global style

mlutze:
Here we can see what IntelliJ will do if we introduce the style in #10283 

Depends on https://github.com/flix/flix/pull/10570

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
@@ -120,7 +120,7 @@ object LspServer {
       *   - Flix source files (*.flix, src/**/*.flix, test/**/*.flix).
       *   - JAR files (lib/**/*.jar).
       *   - Flix package files (lib/**/*.fpkg).
-      *     */
+      */
     private def loadFlixProject(roots: List[WorkspaceFolder]): Unit = {
       for {
         root <- roots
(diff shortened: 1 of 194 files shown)


main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
```diff
@@ -120,7 +120,7 @@ object LspServer {
       *   - Flix source files (*.flix, src/**/*.flix, test/**/*.flix).
       *   - JAR files (lib/**/*.jar).
       *   - Flix package files (lib/**/*.fpkg).
-      */
+      *     */
```
JonathanStarup, line 123: whats happening here?

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/KeywordCompleter.scala
```diff
@@ -90,74 +90,74 @@ object KeywordCompleter {
   def getExprKeywords(range: Range): List[Completion] =
     List(
       // A
-      Completion.KeywordCompletion("and"         , range, Priority.Default),
+      Completion.KeywordCompletion("and", range, Priority.Default),
```
JonathanStarup, line 93: I think this part is easier to read with the non-standard whitespace. Is this change a must-do setting or just configuring the autoformat button?

magnus-madsen:
I think such a big PR is too much. Can we not do it gradually? Also it seems some files should not be reformatted, as pointed out by Jonathan

main/src/ca/uwaterloo/flix/api/lsp/provider/completion/KeywordCompleter.scala
mlutze: I cannot find an option to turn it off.

mlutze:
I'll do one directory at a time.

PR #10575, opened 2025-05-13
refactor: generalize `DependencyGraph`

jaschdoc:
@magnus-madsen Maybe this is useful for `Optimizer`. Maybe not? :)


magnus-madsen:
Not sure. Is this the exact abstraction we need? Its only one function...

JonathanStarup:
I think it makes it easier to read in itself

jaschdoc:
I wanted to prepare it for adding it to the `Inliner` where I may have to add more functions

magnus-madsen:
> I wanted to prepare it for adding it to the `Inliner` where I may have to add more functions

I would rather you start with a separate implementation and then we can see later if it is the same abstraction.

PR #10577, opened 2025-05-13
feat: extensible variants backend

JonathanStarup:
Lowering is still missing


magnus-madsen:
A single test case? Did you test?

magnus-madsen:
Could there be clashes if we have ext. variants and enums that have the same tag names but different arity/types?

JonathanStarup:
No the program is well typed, so theyre used disjointly

PR #10576, opened 2025-05-13
feat: add static jvm method for control pure functions

jaschdoc:
Closes https://github.com/flix/flix/issues/10480

PR #10607, opened 2025-05-14
feat: add `DirectContext` in `GenExpression`

jaschdoc:
This is a step towards static method calls and should result in a small decrease in code size. Direct context is used for functions that are control pure, so in the future static methods can use this context as they will not have access to `pc` and `pcLabels` etc.


JonathanStarup:
@jaschdoc oops made reviews to the wrong PR  #10576

But idea is - it should be about effect vs direct context, without adding anything about the static def

JonathanStarup:
looks good, now its just filling out the `???`?

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -36,21 +36,48 @@ object GenExpression {
...
+      }
+    }
+
+  }
+
+  case class EffectContext(clazz: JvmType.Reference,
                            entryPoint: Label,
                            lenv: Map[Symbol.LabelSym, Label],
                            newFrame: InstructionSet, // [...] -> [..., frame]
                            setPc: InstructionSet, // [..., frame, pc] -> [...]
```
JonathanStarup, line 65: Oh i forgot this part. You can see that setPc expects frame and PC on the stack, so we have to avoid those in the direct case

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -1198,7 +1240,18 @@ object GenExpression {
       }
       mv.visitVarInsn(ALOAD, 0)
```
JonathanStarup, line 1241: So this ALOAD and compileINt is a part of the setPc call essentially

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -1166,22 +1200,30 @@ object GenExpression {
...
+              mv.visitJumpInsn(GOTO, afterUnboxing)
+
+              mv.visitLabel(pcPointLabel)
+              printPc(mv, pcPoint)
+              mv.visitVarInsn(ALOAD, 1)
+
+              mv.visitLabel(afterUnboxing)
+
+            case DirectContext(_, _, _, _) =>
+              throw InternalCompilerException("unexpected static method context in control impure function", loc)
```
JonathanStarup, line 1225: static -> direct

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -1214,9 +1261,8 @@ object GenExpression {
       mv.visitLabel(ifEnd)
 
     case Expr.Branch(exp, branches, _, _, _) =>
-      // Calculating the updated jumpLabels map
```
JonathanStarup, line 1217: I think just keep the unrelated code to make the PR prettier (e.g. this comment)

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -1384,69 +1430,81 @@ object GenExpression {
           INVOKESTATIC(BackendObjType.Handler.InstallHandlerMethod)
       }
       ins(new BytecodeInstructions.F(mv))
-      // handle value/suspend/thunk if in non-tail position
```
JonathanStarup, line 1387: I wish you could edit the git diff :L

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
JonathanStarup, line 1225: ICE is usually capitalized, i think you have a few lowercase ones

JonathanStarup:
@magnus-madsen The idea is to toggle effect instrumentation on/off via two different method contexts. The git diff is a mess but I vouch for the detailed stuff. You can just review big picture if you prefer.

JonathanStarup:
@magnus-madsen I think this is a good PR even without any follow up work and all the backend stuff i'm doing makes it annoying to maintain. Not saying that this PR is urgent, but it'd be nice if this PR be merged before my backend PR's

JonathanStarup:
@jaschdoc I've merged here https://github.com/jaschdoc/flix/pull/4

jaschdoc:
Thanks!

PR #10611, opened 2025-05-14
refactor: implicit parser context

JonathanStarup:
Makes the code a lot easier to read and reduces chance of bugs (and fixes a few)

+400 -300 but if you count chars then this removes code

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -17,7 +17,7 @@ package ca.uwaterloo.flix.language.phase
 
 import ca.uwaterloo.flix.api.Flix
 import ca.uwaterloo.flix.language.CompilationMessage
-import ca.uwaterloo.flix.language.ast.{shared, *}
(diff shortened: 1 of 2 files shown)


magnus-madsen:
I would have thought the mistake was to give it a default parameter?

JonathanStarup:
default parameter is a mistake yea, i guess this is one part fixing that and one part just realizing that the context follows the function structure, so the context is easier written as an implicit

magnus-madsen:
I am not really convinced. I could be convinced it should be explicit?

But also, we also most not using syntactic context anymore... Hmm

magnus-madsen:
I think it is only used for keyword completion?

JonathanStarup:
oh I mean deleting it is also fine by me. What points do you disagree with?

JonathanStarup:
This PR fixes bugs, prevents bugs, and makes code clearer, what more should a PR do?

magnus-madsen:
But would be better to get rid of it-- but I suppose we cannot? We would have to drop keyword completions?

JonathanStarup:
even deleting syntactic context is easier if we merge this

PR #10614, opened 2025-05-15
chore: autoformat lsp/*

mlutze:


main/src/ca/uwaterloo/flix/api/lsp/CompletionItem.scala
```diff
@@ -48,18 +48,18 @@ object CompletionItem {
   *                         will be ignored.
   */
 case class CompletionItem(
-  label: String,
```
mlutze, line 51: I don't like this formatting. Should we add a rule?

magnus-madsen: But I think it is standard. Its like that elsewhere.

Issue #10620, opened 2025-05-16
Refactor library tests


magnus-madsen:
Files to look at:

- [ ] TestList
- [ ] TestOption
- [ ] TestMap
- [ ] ...

What to do:

- Check that every functions in the corresponding module has a few tests
- Change `@test` to `@Test`
- Reindent correctly

magnus-madsen:
Fixed.

PR #10634, opened 2025-05-19
refactor: change @test to @Test and fix indenting

CadeMichael:
fixes issue #10620


magnus-madsen:
This PR is too big for us to handle. Can we work one module at a time starting with List?

CadeMichael:
Yep I will remove the changes for all other modules and readd them one at a time.

PR #10645, opened 2025-05-21
feat: extend `Priority` with two additional priorities

magnus-madsen:


magnus-madsen:
@JonathanStarup @jaschdoc If either of you could approve and click merge. Thanks

PR #10647, opened 2025-05-21
fix: avoid spurious expression keyword suggestions

magnus-madsen:
If the user has typed "le" we suggest "let" but not "match". 

The fix for expressions is easy, but the general fix in other cases is hard and will have to wait.


JonathanStarup (approved):
how did this work before, without any filtering?

magnus-madsen:
> how did this work before, without any filtering?

VSCode does its own magic filtering and ranking. So it mostly worked 👍

PR #10652, opened 2025-05-22
chore: reorder `ExtPattern` in ASTs

magnus-madsen:


magnus-madsen:
@jaschdoc I just did a bit of reordering for consistency.

jaschdoc (approved):
Shouldn't the commit tag be `style` instead of `chore`?

PR #10654, opened 2025-05-22
feat: memoize `ParamKind`s in `LambdaDrop`

jaschdoc:


magnus-madsen:
What is this?

jaschdoc:
Instead of computing each matrix twice, we can simply store the result after compute it :)

jaschdoc:
On second thought, this doesn't really do much. The matrices are often very small (something like 2x3)

PR #10660, opened 2025-05-23
refactor: backend Ins without functions

JonathanStarup:
I was hoping this would
- speed up the backend
- with that speedup, permit full backend conversion

PR: 456.3 ms (25.0%)
Master: 297.1 ms (17.2%)

amazing


JonathanStarup:
@magnus-madsen this is not important right now - but removing the functional part of InstructionSet is 50% slower

PR #10682, opened 2025-05-25
feat: drop usage of delta set in `Inliner`

magnus-madsen:
This PR drops usage of the delta set in the inliner.

This makes it simpler to understand the inliner.

In theory this could increase work because inlining is repeated. 

In practice this seems not to be the case:

```
Master

Throughput (best): 72.277 lines/sec (with 24 threads.)

  min:  9.088, max: 72.277, avg: 60.475, median: 62.102

Finished 100 iterations on 69.994 lines of code in 123 seconds.


PR

Throughput (best): 72.698 lines/sec (with 24 threads.)

  min:  8.575, max: 72.698, avg: 67.482, median: 70.518

Finished 100 iterations on 69.994 lines of code in 111 seconds.
```


magnus-madsen:
My theory is that this improve parallelism (even if at the cost of more work)

jaschdoc (approved):

magnus-madsen:
(This of course changes your baseline)

PR #10683, opened 2025-05-26
refactor: simplify `GetFunctionInterfaceType`

JonathanStarup:


JonathanStarup:
@magnus-madsen These PR avoid creating a JvmType just for everybody to only care about the `.name` of it.

magnus-madsen (approved):

PR #10684, opened 2025-05-26
fix: perform checked cast on struct put and get

jaschdoc:
It is safer to cast to struct type locally rather than relying on the variable being cast to struct from object somewhere else in the method. This is also important for static methods since the type of the parameters will be erased, i.e., primitive or `Object`, which means struct put and get should cast the object to the desired type.
Also, they are *atomic* operations so they should be self-contained, right? :)

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
@@ -790,6 +790,8 @@ object GenExpression {
         val structType = BackendObjType.Struct(JvmOps.instantiateStruct(sym, targs))
         // evaluating the `base`
         compileExpr(exp)
+        // Cast operand to struct type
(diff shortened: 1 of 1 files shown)


JonathanStarup:
> This is also important for static methods since the type of the parameters will be erased

Do they have to be? I think we can keep them exact since the static function is not a part of any general interface

JonathanStarup:
In general, casting should be eager (documented in Eraser) - which means that the casts should be unnecessary here. Casting is a bit of a mess to figure out but i don't think erased args can fit into the current setup (you could cast and store into different vars up front)

magnus-madsen:
Does this happen because of erasure?

jaschdoc:
What exactly do you mean? :)

jaschdoc:
@JonathanStarup should I close this now? I see you have made a lot of changes to the backend

JonathanStarup:
Yeah probably. Now you should be able to trust the output of compileExpr more

PR #10691, opened 2025-05-27
experiment: first iteration of blackjack without user input or backtracking

CadeMichael:
If anyone has tips on how to make this more effect style please let me know, I find that I'm falling into branching pattern matches, I'm not sure if `forM` would help as it would still wrap the output in a result.

### DIFF
--- a/examples/effects-and-handlers/Black-Jack.flix
+++ b/examples/effects-and-handlers/Black-Jack.flix
@@ -1,0 +1,115 @@
+// Player Actions
+enum Action {
+    case Hit,
+    case Stand
+}
+
+// Card Ranks
+enum Card with ToString {
+    case Two,
+    case Three,
+    case Four,
+    case Five,
+    case Six,
+    case Seven,
+    case Eight,
+    case Nine,
+    case Ten,
+    case Jack,
+    case Queen,
+    case King,
+    case Ace
+}
+
+// Deck Operations
+eff Deck {
+    // generate and shuffle a 52 card deck
+    def genDeck() : List[Card]
+    // pull card
+    def pullCard(deck: List[Card]): (Option[Card], List[Card])
+}
+
+eff Terminal {
+    def println(s: String): Unit
+}
+
+def cardValue(card: Card): Int32 = match card {
+   case Card.Two => 2
+   case Card.Three => 3
+   case Card.Four => 4
+   case Card.Five => 5
+   case Card.Six => 6
+   case Card.Seven => 7
+   case Card.Eight => 8
+   case Card.Nine => 9
+   case Card.Ace => 11
+   case _ => 10 // 10 and face cards
+}
+def gameOver(p_val: Int32, d_val: Int32): Bool \ Terminal = match (p_val, d_val) {
+    case (p, _) if p == 21 => Terminal.println("player -> ${p}, Player wins."); true
+    case (p, _) if p > 21 => Terminal.println("player -> ${p}, Dealer wins."); true
+    case (_, d) if d == 21 => Terminal.println("dealer -> ${d}, Dealer wins."); true
+    case (_, d) if d > 21 => Terminal.println("dealer -> ${d}, Player wins."); true
+    case (_, _) => false
+}
+
+def handValue(cards: List[Card]): Int32 =
+    List.foldLeft((x, y) -> x + y, 0, List.map(cardValue, cards))
+
+def gameLoop(deck: List[Card], player: List[Card], dealer: List[Card]): Unit \ {Terminal, Deck} = {
+    match Deck.pullCard(deck) { // player draws
+        case (Some(p_card), afterPlayer)  => {
+            Terminal.println("Player => ${List.toString(p_card::player)}");
+            // players hand with new card
+            let p_val = handValue(p_card::player);
+            // dealer never hits after 16
+            if (handValue(dealer) > 16) {
+                if (gameOver(p_val, handValue(dealer))) { // check winner or loser
+                    Terminal.println("Done.")
+                } else {
+                    gameLoop(afterPlayer, p_card::player, dealer)
+                }
+            } else {
+                match Deck.pullCard(afterPlayer) { // dealer draws
+                    case (Some(d_card), afterDealer) => {
+                        Terminal.println("Dealer => ${List.toString(d_card::dealer)}");
+                        let d_val = handValue(d_card::dealer);
+                        if (gameOver(p_val, d_val)) { // check winner or loser
+                            Terminal.println("Done.")
+                        } else {
+                            gameLoop(afterDealer, p_card :: player, d_card :: dealer)
+                        }
+                    }
+                    case (None, _) => Terminal.println("No more cards")
+                }
+            }
+        }
+        case (None, _) => Terminal.println("No more cards")
+    }
+}
+
+def main(): Unit \ {NonDet, IO} =
+    run {
+        let deck = Deck.genDeck();
+        gameLoop(deck, Nil, Nil)
+    } with handler Deck {
+        def genDeck(resume) = {
+            // type of cards
+            let cards = List#{
+                Card.Ace, Card.Two, Card.Three, Card.Four, Card.Five,
+                Card.Six, Card.Seven, Card.Eight, Card.Nine, Card.Ten,
+                Card.Jack, Card.Queen, Card.King
+            };
+            // 4 of each card per deck
+            let deck = List.flatten(List.map(card -> List.repeat(4, card), cards));
+            // shuffle
+            resume(List.shuffle(deck))
+        }
+        // present the next card
+        def pullCard(deck, resume) = match deck {
+            case Nil => Terminal.println("No more cards"); resume((None, Nil)),
+            case (card :: tail) => resume((Some(card),tail))
+        }
+    } with handler Terminal {
+        def println(s, resume) = {println(s); resume()}
+    }


examples/effects-and-handlers/Black-Jack.flix
```diff
@@ -0,0 +1,115 @@
...
+def cardValue(card: Card): Int32 = match card {
+   case Card.Two => 2
+   case Card.Three => 3
+   case Card.Four => 4
+   case Card.Five => 5
+   case Card.Six => 6
+   case Card.Seven => 7
+   case Card.Eight => 8
+   case Card.Nine => 9
+   case Card.Ace => 11
```
JonathanStarup, line 45: Can't ace also be 1?

examples/effects-and-handlers/Black-Jack.flix
```diff
@@ -0,0 +1,115 @@
...
+    case Nine,
+    case Ten,
+    case Jack,
+    case Queen,
+    case King,
+    case Ace
+}
+
+// Deck Operations
+eff Deck {
```
JonathanStarup, line 25: I wouldn't have thought that this would be an effect but it might make sense

examples/effects-and-handlers/Black-Jack.flix
```diff
@@ -0,0 +1,115 @@
...
+    // generate and shuffle a 52 card deck
+    def genDeck() : List[Card]
+    // pull card
+    def pullCard(deck: List[Card]): (Option[Card], List[Card])
+}
+
+eff Terminal {
```
JonathanStarup, line 32: If the example doesn't have to be self-contained, i'd just reuse Console

examples/effects-and-handlers/Black-Jack.flix
```diff
@@ -0,0 +1,115 @@
...
+            let p_val = handValue(p_card::player);
+            // dealer never hits after 16
+            if (handValue(dealer) > 16) {
+                if (gameOver(p_val, handValue(dealer))) { // check winner or loser
+                    Terminal.println("Done.")
+                } else {
+                    gameLoop(afterPlayer, p_card::player, dealer)
+                }
+            } else {
+                match Deck.pullCard(afterPlayer) { // dealer draws
```
JonathanStarup, line 73: Couldn't you call gameLoop here directly?

CadeMichael: I could use a tracker for dealer vs player for who pulls next and call there, would that be better?

examples/effects-and-handlers/Black-Jack.flix
CadeMichael: I'm thinking of a clever way to introduce this, maybe something like

```
def handValue(cards: List[Card]): Int32 =
    let value = List.foldLeft((x, y) -> x + y, 0, List.map(cardValue, cards));
    if (value > 21 and List.exists(x -> x == Card.Ace, cards)) value - 10 else value
```

examples/effects-and-handlers/Black-Jack.flix
JonathanStarup: hmm I would've thought you could do a player loop like you do, and then when the player stands and didn't lose, you'd calls `doDealerStuff` and that function would handle the dealer logic

magnus-madsen:
@mlutze You can take over reviewing.

examples/effects-and-handlers/Black-Jack.flix
```diff
@@ -0,0 +1,171 @@
+import java.lang.System
+import java.io.BufferedReader;
+import java.io.InputStreamReader;
```
magnus-madsen: Where are these used? It would be best if the program could be pure Flix.

examples/effects-and-handlers/Black-Jack.flix
```diff
@@ -0,0 +1,171 @@
...
+
+/// Player Actions
+enum Action with ToString {
+    case Hit,
+    case Stand
+}
+
+/// Player input
+eff Play {
+    def hitOrStand(): Result[String,Action]
```
magnus-madsen: Why do we return a result here? Seems overly complicated?

examples/effects-and-handlers/Black-Jack.flix
```diff
@@ -0,0 +1,171 @@
...
+}
+
+/// Card Ranks
+enum Card with ToString, Eq {
```
magnus-madsen, line 19: Eq before ToString (by convention)

examples/effects-and-handlers/Black-Jack.flix
```diff
@@ -0,0 +1,171 @@
...
+    case Nine,
+    case Ten,
+    case Jack,
+    case Queen,
+    case King,
+    case Ace
+}
+
+/// Deck Operations
+eff Deck {
```
magnus-madsen: OK; so working with decks is also an effect. I guess that is OK-- but will we have more than one implementation?

magnus-madsen: In some sense this is completely orthogonal to the playing part right?

examples/effects-and-handlers/Black-Jack.flix
```diff
@@ -0,0 +1,171 @@
...
+    // generate and shuffle a 52 card deck
+    def genDeck() : List[Card]
+    // pull card
+    def pullCard(deck: List[Card]): (Option[Card], List[Card])
```
magnus-madsen: This is actually not an effect. Its just an operation on the data.

examples/effects-and-handlers/Black-Jack.flix
```diff
@@ -0,0 +1,171 @@
...
+}
+
+eff Terminal {
+    def println(s: String): Unit
```
magnus-madsen: Indeed reuse Console

examples/effects-and-handlers/Black-Jack.flix
```diff
@@ -0,0 +1,171 @@
...
+        // present the next card
+        def pullCard(deck, resume) = match deck {
+            case Nil => Terminal.println("No more cards"); resume((None, Nil)),
+            case (card :: tail) => resume((Some(card),tail))
+        }
+    } with handler Terminal {
+        def println(s, resume) = {println(s); resume()}
+    } with handler Play {
+        def hitOrStand(resume) = 
+            let reader = new BufferedReader(new InputStreamReader(System.in));
```
magnus-madsen: Console can do this for you.

mlutze:
> @mlutze You can take over reviewing.

I'll wait for this iteration. Ping me when it's RFR.

mlutze:
Is this ready for another review?

examples/effects-and-handlers/Black-Jack.flix
```diff
@@ -0,0 +1,178 @@
...
+    case (p, _) if p > 21 => println("player -> ${p}, Dealer wins."); true
+    case (_, d) if d == 21 => println("dealer -> ${d}, Dealer wins."); true
+    case (_, d) if d > 21 => println("dealer -> ${d}, Player wins."); true
+    case (_, _) => false
+}
+
+///
+/// Calculate the value of a hand.
+///
+def handValue(cards: List[Card]): Int32 =
```
mlutze, line 89: Would it be clearer if `cardValue` returned a list?

examples/effects-and-handlers/Black-Jack.flix
```diff
@@ -0,0 +1,178 @@
...
+    let value = List.foldLeft((x, y) -> x + y, 0, List.map(cardValue, cards));
```
mlutze: I think there's `List.sum`

examples/effects-and-handlers/Black-Jack.flix
```diff
@@ -0,0 +1,178 @@
...
+    // 4 of each card per deck
+    let deck = List.flatten(List.map(card -> List.repeat(4, card), cards));
+    // shuffle
+    List.shuffle(deck)
+}
+
+///
+/// Return the result of the game if the Player or Dealer has won.
+///
+def gameOver(p_val: Int32, d_val: Int32): Bool \ IO = match (p_val, d_val) {
```
mlutze: `camelCase` instead of `snake_case` by convention

examples/effects-and-handlers/Black-Jack.flix
```diff
@@ -0,0 +1,178 @@
...
+/// Player Action.
+///
+eff Play {
+    def hitOrStand(): Result[String,Action]
+}
+
+///
+/// Card Ranks.
+///
+enum Card with ToString, Eq {
```
mlutze, line 19: I see that suit is not _really_ used in Blackjack, but a typical video game would show suits nonetheless.

examples/effects-and-handlers/Black-Jack.flix
```diff
@@ -0,0 +1,178 @@
...
+   case Card.Eight => 8
+   case Card.Nine => 9
+   case Card.Ace => 11
+   case _ => 10 // 10 and face cards
+}
+
+///
+/// Pop a card from the top of the Deck.
+///
+def pullCard(deck: List[Card]): Option[(Card, List[Card])] \ IO = match deck {
```
mlutze: In an effect-oriented program, I would expect IO, NonDet, etc. only to show up in the handlers.

IO would be delegated to some user interface effect, which could be implemented either with the console or graphical UI.

mlutze (commented):
This is a reasonable implementation, but it is not totally effect-oriented yet. As much as possible, real effects (IO, NonDet) and other errors (Result) should be pushed toward the edges -- into handlers that are used at the entrypoint.

examples/effects-and-handlers/Black-Jack.flix
CadeMichael: `def readGuess(): Result[String, String] \ IO = `

same basic structure as the first demo in the documentation. I'm not sure how I would account for improper input. I'm trying the following without Result but there are some errors

```
    } with handler Play {
        def hitOrStand(resume) = 
            let line = Console.readln();
            match line { // get Player action
                case "y" => resume(Action.Hit)
                case "n" => resume(Action.Stand)
                case _ => resume(Play.hitOrStand())
            }
    } with Console.runWithIO
```

examples/effects-and-handlers/Black-Jack.flix
CadeMichael: I think that would only apply to Ace?

mlutze: Yes, but the statement "a card may have more than one value" is generally true.

I'm not sure the subtraction is correct. For example, what happens if we have 3 or 4 aces?

PR #10703, opened 2025-05-28
fix: remove ad hoc erasure in the backend

JonathanStarup:
This removes ad hoc erasure of a few things, but notably arrays. This PR should fix our long standing issue around array types in the backend.

As a bonus, this PR conceptually eliminates JvmType - but follow up PRs will actually weed it out

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
@@ -566,7 +566,7 @@ object GenExpression {
         val cases = JvmOps.instantiateEnum(root.enums(sym.enumSym), targs)
         val termTypes = cases(sym)
         compileUntag(exp, idx, termTypes)
-        AsmOps.castIfNotPrim(mv, JvmOps.getJvmType(tpe))
+        mv.visitByteIns(BytecodeInstructions.castIfNotPrim(BackendType.toBackendType(tpe)))
 
       case AtomicOp.Index(idx) =>
         val List(exp) = exps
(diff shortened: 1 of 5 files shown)


main/src/ca/uwaterloo/flix/language/phase/jvm/JvmOps.scala
```diff
@@ -28,54 +28,6 @@ import java.nio.file.{Files, LinkOption, Path}
...
-    case MonoType.Int16 => JvmType.PrimShort
-    case MonoType.Int32 => JvmType.PrimInt
-    case MonoType.Int64 => JvmType.PrimLong
-    case MonoType.BigInt => JvmType.BigInteger
-    case MonoType.String => JvmType.String
-    case MonoType.Regex => JvmType.Regex
-    case MonoType.Region => JvmType.Object
-    case MonoType.Null => JvmType.Object
-    // Compound
-    case MonoType.Array(_) => JvmType.Object
```
JonathanStarup, line 63: In this function you can see that even though `getJvmType` is supposed to return the exact type of `tpe`, there is some erasure. This is very ad hoc and makes code gen hard to get right. This PR instead uses BackendType for casting - doing no ad-hoc erasure

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -1636,21 +1633,7 @@ object GenExpression {
     // Evaluate arguments left-to-right and push them onto the stack.
     for ((arg, argType) <- args.zip(signature)) {
       compileExpr(arg)
-      if (!argType.isPrimitive) {
-        // NB: Really just a hack because the backend does not support array JVM types properly.
-        mv.visitTypeInsn(CHECKCAST, asm.Type.getInternalName(argType))
-      } else {
-        arg.tpe match {
-          // NB: This is not exhaustive. In the new backend we should handle all types, including multidim arrays.
```
JonathanStarup, line 1644: Notice that this hack is now removed - arrays are just fully represented in BackendType

magnus-madsen:
I need to read and think a bit-- so expect a short delay.

What does it conceptually mean to remove JvmType? Does it mean that MonoType is sufficient to generate the right bytecode?

Can we then also eventually remove MonoType in favor of ClassDesc?

PR #10709, opened 2025-05-29
refactor: Split `F` from `BackendObjType`

JonathanStarup:
This PR splits code generations (public, final, implementation) from the method/constructor/etc references that are used for name lookup

This leads the way for
- `F` -> classFile API
- `BackendObjType` -> `ClassDesc`

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
@@ -188,7 +188,7 @@ object BackendObjType {
     def Constructor: ConstructorMethod = ConstructorMethod(this.jvmName, List(JavaObject.toTpe))
 
     /** `[] --> return` */
-    private def constructorIns: InstructionSet = {
+    private def constructorIns: InstructionSet =
       withName(1, JavaObject.toTpe)(exp =>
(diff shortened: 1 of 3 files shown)


main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
```diff
@@ -872,76 +898,80 @@ object BackendObjType {
         thisLoad() ~ GETFIELD(BeginColField) ~ INVOKEVIRTUAL(StringBuilder.AppendInt32Method) ~
         // create the string
         INVOKEVIRTUAL(JavaObject.ToStringMethod) ~ ARETURN()
-    ))
+    }
   }
 
   case object Global extends BackendObjType with Generatable {
     def genByteCode()(implicit flix: Flix): Array[Byte] = {
```
JonathanStarup, line 905: In this PR - all code related to code generation is within the `genByteCode: Unit => Array[Byte]` function. Which means that all code in this file could be rewritten in the classfile lib without touching any other code

magnus-madsen: OK; so this might be starting point for @CadeMichael

PR #10716, opened 2025-05-30
docs: tic-tac-toe on the command line

gaganchandan:
This is my first attempt at the tic-tac-toe app. I've made it for the CLI since I wasn't able to make windows persist with Swing. However, the interface is written as an effect separate from the game loop and logic to make it easier to implement a GUI in the future if possible.

### DIFF
--- a/examples/apps/tic-tac-toe/flix.toml
+++ b/examples/apps/tic-tac-toe/flix.toml
@@ -1,0 +1,6 @@
+[package]
+name        = "tic-tac-toe"
+description = "test"
+version     = "0.1.0"
+flix        = "0.59.0"
+authors     = ["Gagan Chandan <gagan@gaganchandan.com>"]
--- a/examples/apps/tic-tac-toe/src/Board.flix
+++ b/examples/apps/tic-tac-toe/src/Board.flix
@@ -1,0 +1,78 @@
+pub struct Board[r: Region] {
+    r: Region[r],
+    arr: Array[Symbol, r]
+}
+
+mod Board {
+
+    pub def newBoard(rc: Region[r]): Board[r] \ r = {
+        new Board @ rc {r = rc, arr = Array.repeat(rc, 9, Symbol.Empty)}
+    }
+
+    pub def getArr(board: Board[r]): Array[Symbol, r] = {
+        board->arr
+    }
+
+    pub def makeMove(board: Board[r], symbol: Symbol, index: Int32): (Board[r], Bool) \ r = {
+        if (Array.get(index, board->arr) == Symbol.Empty) {
+            Array.put(symbol, index, board->arr);
+            (board, true)
+        }
+        else {
+           (board, false)
+        }
+    }
+
+    pub def checkWin(board: Board[r], symbol: Symbol): Bool \ r  = {
+        if ((Array.get(0, board->arr) == symbol)
+		and (Array.get(1, board->arr) == symbol)
+		and (Array.get(2, board->arr) == symbol)) {
+            true
+        }
+        else if ((Array.get(3, board->arr) == symbol)
+		and (Array.get(4, board->arr) == symbol)
+		and (Array.get(5, board->arr) == symbol)) {
+            true
+        }
+        else if ((Array.get(6, board->arr) == symbol)
+		and (Array.get(7, board->arr) == symbol)
+		and (Array.get(8, board->arr) == symbol)) {
+            true
+        }
+        else if ((Array.get(0, board->arr) == symbol)
+		and (Array.get(3, board->arr) == symbol)
+		and (Array.get(6, board->arr) == symbol)) {
+            true
+        }
+        else if ((Array.get(1, board->arr) == symbol)
+		and (Array.get(4, board->arr) == symbol)
+		and (Array.get(7, board->arr) == symbol)) {
+            true
+        }
+        else if ((Array.get(2, board->arr) == symbol)
+		and (Array.get(5, board->arr) == symbol)
+		and (Array.get(8, board->arr) == symbol)) {
+            true
+        }
+        else if ((Array.get(0, board->arr) == symbol)
+		and (Array.get(4, board->arr) == symbol)
+		and (Array.get(8, board->arr) == symbol)) {
+            true
+        }
+        else if ((Array.get(2, board->arr) == symbol)
+		and (Array.get(4, board->arr) == symbol)
+		and (Array.get(6, board->arr) == symbol)) {
+            true
+        }
+
+        else {
+            false
+        }
+
+    }
+
+    pub def boardFilled(board: Board[r]): Bool \ r = {
+        Array.forAll(sym -> sym != Symbol.Empty, board->arr)
+    }
+
+}
--- a/examples/apps/tic-tac-toe/src/Interface.flix
+++ b/examples/apps/tic-tac-toe/src/Interface.flix
@@ -1,0 +1,46 @@
+mod Interface {
+
+    import java.io.BufferedReader;
+    import java.io.InputStreamReader;
+    import java.lang.System;
+
+    pub eff Interface {
+        def display(board: List[Symbol], msg: String): Unit
+        def getMove(): Result[String, String]
+    }
+
+    pub def interfaceWithIO(f: Unit -> a \ ef): a \ ef - Interface + IO = 
+        run {
+            f()
+        }
+        with handler Interface {
+            def display(board, msg, resume) = region rc {
+                if (not List.isEmpty(board)) {
+                    let arr = List.toArray(rc, board);
+                    println("");
+                    println("       |       |       ");
+                    println("   " ++ Symbol.toString(Array.get(0, arr)) ++ "   |   " ++ Symbol.toString(Array.get(1, arr)) ++ "   |   " ++ Symbol.toString(Array.get(2, arr)) ++ "   " );
+                    println(" - - - | - - - | - - - ");
+                    println("   " ++ Symbol.toString(Array.get(3, arr)) ++ "   |   " ++ Symbol.toString(Array.get(4, arr)) ++ "   |   " ++ Symbol.toString(Array.get(5, arr)) ++ "   " );
+                    println(" - - - | - - - | - - - ");
+                    println("   " ++ Symbol.toString(Array.get(6, arr)) ++ "   |   " ++ Symbol.toString(Array.get(7, arr)) ++ "   |   " ++ Symbol.toString(Array.get(8, arr)) ++ "   " );
+                    println("       |       |       ");
+                    println("");
+                    println(msg);
+                    resume()
+                } else {
+                    println("");
+                    println(msg);
+                    resume()
+                }
+            }
+            def getMove(resume) = {
+                let reader = new BufferedReader(new InputStreamReader(System.in));
+                let line = reader.readLine();
+                if (Object.isNull(line)) 
+                    resume(Result.Err("no input"))
+                else 
+                    resume(Result.Ok(line))
+            }
+        }
+}
--- a/examples/apps/tic-tac-toe/src/Main.flix
+++ b/examples/apps/tic-tac-toe/src/Main.flix
@@ -1,0 +1,59 @@
+def readAndParseMove(): Result[String, Int32] \ {Interface.Interface} = 
+    forM(m <- Interface.Interface.getMove();
+         n <- Int32.parse(10, m)
+    ) yield n
+
+def gameLoop(board: Board[r], symbol: Symbol): Unit \ {IO + r + Interface.Interface} = {
+    Interface.Interface.display(Array.toList(Board.getArr(board)), "Enter a move (0 - 8): ");
+    match readAndParseMove() {
+        case Result.Ok(index) =>
+            if(not(index >= 0 and index <= 8)) {
+                Interface.Interface.display(List.Nil, "Invalid move");
+                gameLoop(board, symbol) 
+            } else {
+                let (newBoard, moveSuccess) = Board.makeMove(board, symbol, index);
+                if (moveSuccess) {
+                    if(Board.checkWin(newBoard, symbol)) {
+                        match symbol {
+                            case Symbol.X => Interface.Interface.display(Array.toList(Board.getArr(board)), "Player One wins!!!")
+                            case Symbol.O => Interface.Interface.display(Array.toList(Board.getArr(board)), "Player Two wins!!!")
+                            case _ => ()
+                        }
+                    } else if(Board.boardFilled(newBoard)) {
+                        Interface.Interface.display(Array.toList(Board.getArr(board)), "Draw")
+                    } else {
+                        gameLoop(newBoard, Symbol.complement(symbol))
+                    }
+                } else {
+                    Interface.Interface.display(List.Nil, "Cell already filled");
+                    gameLoop(board, symbol) 
+                }
+            }
+        case Result.Err(_) => Interface.Interface.display(List.Nil, "Invalid move")
+    }
+}
+
+def main(): Unit \ IO = region rc1 {
+    run {
+        println("");
+        println("_______ _          _______             _______           ");
+        println(" |__   __(_)        |__   __|           |__   __|        ");
+        println("    | |   _  ___ ______| | __ _  ___ ______| | ___   ___ ");
+        println("    | |  | |/ __|______| |/ _` |/ __|______| |/ _ \\ / _ \\");
+        println("    | |  | | (__       | | (_| | (__       | | (_) |  __/");
+        println("    |_|  |_|\\___|      |_|\\__,_|\\___|      |_|\\___/ \\___|");
+        println("");
+        println("Instructions:");
+        println("The cells are numbered as follows:\n");
+        println("   " ++ "0" ++ "   |   " ++ "1" ++ "   |   " ++ "2" ++ "   " );
+        println(" - - - | - - - | - - - ");
+        println("   " ++ "3" ++ "   |   " ++ "4" ++ "   |   " ++ "5" ++ "   " );
+        println(" - - - | - - - | - - - ");
+        println("   " ++ "6" ++ "   |   " ++ "7" ++ "   |   " ++ "8" ++ "   " );
+        println("       |       |       ");
+        println("Enter your move as one these numbers corresponding to a cell");
+        println("");
+        let board = Board.newBoard(rc1);
+        gameLoop(board, Symbol.X)      
+    } with Interface.interfaceWithIO
+}
--- a/examples/apps/tic-tac-toe/src/Symbol.flix
+++ b/examples/apps/tic-tac-toe/src/Symbol.flix
@@ -1,0 +1,20 @@
+pub enum Symbol with Eq {
+    case X,
+    case O,
+    case Empty
+}
+
+mod Symbol {
+
+    pub def complement(symbol: Symbol): Symbol  = match symbol {
+        case Symbol.X       => Symbol.O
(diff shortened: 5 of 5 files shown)


examples/apps/tic-tac-toe/src/Interface.flix
```diff
@@ -0,0 +1,46 @@
+mod Interface {
+
+    import java.io.BufferedReader;
+    import java.io.InputStreamReader;
+    import java.lang.System;
+
+    pub eff Interface {
+        def display(board: List[Symbol], msg: String): Unit
+        def getMove(): Result[String, String]
```
magnus-madsen: What's the return type here? Could it be simplified.

For example, we could leave error handling as the responsibility of the handler. See e.g. Http effect.

examples/apps/tic-tac-toe/src/Interface.flix
```diff
@@ -0,0 +1,46 @@
+mod Interface {
+
+    import java.io.BufferedReader;
+    import java.io.InputStreamReader;
+    import java.lang.System;
```
magnus-madsen: We would like a pure Flix program.

Can the Console effect be reused here?

examples/apps/tic-tac-toe/src/Interface.flix
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    pub def interfaceWithIO(f: Unit -> a \ ef): a \ ef - Interface + IO = 
+        run {
+            f()
+        }
+        with handler Interface {
+            def display(board, msg, resume) = region rc {
+                if (not List.isEmpty(board)) {
+                    let arr = List.toArray(rc, board);
```
magnus-madsen: I would isolate pretty printing to some specialized functions.

examples/apps/tic-tac-toe/src/Interface.flix
gaganchandan: Are you suggesting that `getMove` not return a `Result` type but rather just the move (`Int32`)?

examples/apps/tic-tac-toe/src/Interface.flix
gaganchandan: Yes, this can be done. I'll amend the commit.

examples/apps/tic-tac-toe/src/Interface.flix
gaganchandan: I designed it like this with a single display function because the initial plan was to implement it with Swing. Doing it like this would mean that `Main.flix` would be completely independent of `Interface` should it be possible to do it with Swing in the future. I can change it if you would prefer that.

examples/apps/tic-tac-toe/src/Interface.flix
magnus-madsen: I think so, yes.

examples/apps/tic-tac-toe/src/Interface.flix
magnus-madsen: I am just saying to add some helper functions so it is not one big blob :)

gaganchandan: Okay.

examples/apps/tic-tac-toe/src/Board.flix
```diff
@@ -0,0 +1,87 @@
...
+        if (Array.get(index, board->arr) == Symbol.Empty) {
+            Array.put(symbol, index, board->arr);
+            (board, true)
+        }
+        else {
+           (board, false)
+        }
+    }
+
+    pub def checkWin(board: Board[r], symbol: Symbol): Bool \ r  = {
```
magnus-madsen: Do you see anyway to improve the formatting here? or shorten the function?

Maybe it is not possible.

examples/apps/tic-tac-toe/src/Board.flix
```diff
@@ -0,0 +1,87 @@
...
+
+    }
+
+    pub def boardFilled(board: Board[r]): Bool \ r = {
+        Array.forAll(sym -> sym != Symbol.Empty, board->arr)
+    }
+
+    pub def toString(board: Board[r]): String \ r =
+                    "       |       |       "
+                    ++ "   " ++ Symbol.toString(Array.get(0, board->arr)) ++ "   |   " ++ Symbol.toString(Array.get(1, board->arr)) ++ "   |   " ++ Symbol.toString(Array.get(2, board->arr)) ++ "   "
```
magnus-madsen: Perhaps string interpolation can be used, ie.. "${foo}"

examples/apps/tic-tac-toe/src/Interface.flix
magnus-madsen: Is this refactoring possible?

examples/apps/tic-tac-toe/src/Interface.flix
```diff
@@ -0,0 +1,35 @@
+mod Interface {
```
magnus-madsen, line 17: Can we come up with a better name than interface?

examples/apps/tic-tac-toe/src/Main.flix
```diff
@@ -0,0 +1,60 @@
+def readAndParseMove(): Result[String, Int32] \ {Interface.Interface} =
+    forM(m <- Interface.Interface.getMove();
+         n <- Int32.parse(10, m)
+    ) yield n
+
+def gameLoop(board: Board[r], symbol: Symbol): Unit \ {IO + r + Interface.Interface} = {
```
magnus-madsen: This function seems to have a lot of different effects.

What do you think about that?

examples/apps/tic-tac-toe/src/Board.flix
```diff
@@ -0,0 +1,87 @@
+pub struct Board[r: Region] {
+    r: Region[r],
+    arr: Array[Symbol, r]
+}
+
+
+mod Board {
+
+    pub def newBoard(rc: Region[r]): Board[r] \ r = {
```
magnus-madsen: Since this is an example, let us document each function, even if it is just:

```
///
/// Returns an empty tic-tac-toe board.
///
```

examples/apps/tic-tac-toe/src/Interface.flix
gaganchandan: I'm not quite sure. In the Http effect, the Java exceptions are caught and handled in a `try..catch` block which doesn't seem to be applicable here.

magnus-madsen: Chat with Matt. It is about the game not worrying about invalid moves. From the game point of view, there can only be valid moves.

magnus-madsen:
@mlutze Gagan is using a mutable array, but I think for any kind of backtracking to be feasible we will want a purely functional implementation, right?

examples/apps/tic-tac-toe/src/Interface.flix
gaganchandan: The only way I can think of right now is for `getMove` to be recursive so it can ask for a move again in case an invalid one is given. But I don't think handlers can be recursive, no?

examples/apps/tic-tac-toe/src/Board.flix
gaganchandan: Should I still keep each row separate for better readability?

examples/apps/tic-tac-toe/src/Interface.flix
gaganchandan: Terminal?

gaganchandan: It might be too similar to Console though.

examples/apps/tic-tac-toe/src/Main.flix
gaganchandan: Hmm. `IO` doesn't seem to be necessary anymore. I'm not sure if that was always the case or whether it's a result of the latest changes.

mlutze:
> @mlutze Gagan is using a mutable array, but I think for any kind of backtracking to be feasible we will want a purely functional implementation, right?

Yes, agreed. If we want to do a minimax thing later, we ought to use immutable data structures.

examples/apps/tic-tac-toe/src/Interface.flix
magnus-madsen: We should simply not consider the possibility of invalid moves in the domain game logic.

@mlutze Can you explain? Thanks

mlutze: You're right that the handler itself should not call "getMove()".

But a recursive helper function would be appropriate! For example:

```
def getMoveFromConsole(): Move \ Console = {
   let moveString = // read line, maybe prompt
   match parseMove(moveString) {
       case None =>
           // print "invalid move, must by XYZ"
           getMoveFromConsole() // recursive call to try again
       case Some(move) =>
           move
}
```

The point is that the handler knows about errors, but the tic-tac-toe game itself does not need to care about them.

gaganchandan: It seems so simple now that you've given me the answer :(

gaganchandan:
I have replaced Arrays with Vectors based on @mlutze's suggestion

examples/apps/tic-tac-toe/src/Board.flix
```diff
@@ -0,0 +1,110 @@
...
+        }
+        else {
+           (board, false)
+        }
+    }
+
+    ///
+    /// Checks if the player with the specified symbol has won the game on the given board.
+    ///
+    pub def checkWin(board: Board[r], symbol: Symbol): Bool = {
```
mlutze: I suggest we DRY this up a little by creating a list of the winning patterns (`(0, 1, 2), (0, 3, 6), ...`) and then just iterating over the list.

examples/apps/tic-tac-toe/src/Board.flix
```diff
@@ -0,0 +1,110 @@
...
+    /// Returns a new, empty tic-tac-toe board.
+    ///
+    pub def newBoard(rc: Region[r]): Board[r] \ r = {
+        new Board @ rc {vec = Vector.repeat(9, Symbol.Empty)}
+    }
+
+    ///
+    /// Returns the current state of the board as a vector.
+    ///
+    pub def getVec(board: Board[r]): Vector[Symbol] = {
```
mlutze: Can this be made private?

examples/apps/tic-tac-toe/src/Interface.flix
```diff
@@ -0,0 +1,72 @@
+///
+/// Effect representing the user interface for the game.
+///
+pub eff Interface {
+
+    ///
+    /// Prints the current state of the board and a message.
+    ///
+    def display(board: String, msg: String): Unit
```
mlutze: Taking the board as a string here doesn't seem right.

Imagine we have GUI interface with images etc. We'd want to interpret the board some other way.

examples/apps/tic-tac-toe/src/Main.flix
```diff
@@ -0,0 +1,55 @@
...
+
+///
+/// The main entry point of the application.
+///
+/// Creates a new game board and starts the game loop.
+///
+def main(): Unit \ IO = region rc {
+    run {
+        println("");
+        println("_______ _          _______             _______           ");
```
mlutze: Fun :)

Is this line off-center?

examples/apps/tic-tac-toe/src/Main.flix
```diff
@@ -0,0 +1,55 @@
...
+        println(" |__   __(_)        |__   __|           |__   __|        ");
+        println("    | |   _  ___ ______| | __ _  ___ ______| | ___   ___ ");
+        println("    | |  | |/ __|______| |/ _` |/ __|______| |/ _ \\ / _ \\");
+        println("    | |  | | (__       | | (_| | (__       | | (_) |  __/");
+        println("    |_|  |_|\\___|      |_|\\__,_|\\___|      |_|\\___/ \\___|");
+        println("");
+        println("Instructions:");
+        println("The cells are numbered as follows:\n");
+        println("   " ++ "0" ++ "   |   " ++ "1" ++ "   |   " ++ "2" ++ "   " );
```
mlutze: Is there a reason you concatenate the strings instead of just writing it?

examples/apps/tic-tac-toe/src/Board.flix
```diff
@@ -0,0 +1,110 @@
...
+        board->vec
+    }
+
+    ///
+    /// Makes a move on the board at the specified index with the given symbol.
+    ///
+    /// Returns a tuple containing the updated board and a boolean indicating whether the move was successful.
+    ///
+    pub def makeMove(board: Board[r], symbol: Symbol, index: Int32, rc: Region[r]): (Board[r], Bool) \ r = {
```
mlutze: Returning a Boolean for success is a bit unusual for effect-oriented programming, I suspect.

What happens if we instead introduce an error for this case?

examples/apps/tic-tac-toe/src/Interface.flix
```diff
@@ -0,0 +1,72 @@
...
+}
+
+mod Interface {
+
+    ///
+    /// Recursive helper function to read and return only valid moves.
+    ///
+    def getMoveFromConsole(): Int32 \ Console = {
+        let line = Console.readln();
+        if(Object.isNull(line)) {
```
mlutze, line 24: Is this possible? :O

examples/apps/tic-tac-toe/src/Board.flix
gaganchandan: I think it's just left over from some previous iteration. It can be removed entirely.

examples/apps/tic-tac-toe/src/Interface.flix
gaganchandan: What would you suggest? I don't think I can pass `Board[r]` because the effect handler `display(..)` can't have an effect, correct? Perhaps the vector would work?

examples/apps/tic-tac-toe/src/Main.flix
gaganchandan: It is, good catch.

examples/apps/tic-tac-toe/src/Interface.flix
magnus-madsen: Board should probably be immutable and so not have a region.

examples/apps/tic-tac-toe/src/Main.flix
gaganchandan: Oh no! I reused the the old print statements from `toString` from `Board.flix` and didn't pay much attention to it. Let me fix it before anyone else sees this :(

examples/apps/tic-tac-toe/src/Interface.flix
gaganchandan: I have been using a struct for it. Perhaps with the move to `Vector` I could simply use a type alias?

examples/apps/tic-tac-toe/src/Interface.flix
gaganchandan: Do you mean the line being null? If so, that happens if you hit `Ctrl + D`.

examples/apps/tic-tac-toe/src/Interface.flix
magnus-madsen: Yeah or an enum (we dont have pure structs yet, so those are the choices)

mlutze:
@gaganchandan please ping me when this is ready for another review

examples/apps/tic-tac-toe/src/Interface.flix
gaganchandan: I've switched to an enum in the latest commit.

examples/apps/tic-tac-toe/src/Board.flix
gaganchandan: I've tweaked it in the latest commit, using the `Result` type instead of a Boolean.

examples/apps/tic-tac-toe/src/Board.flix
```diff
@@ -0,0 +1,71 @@
...
+    ///
+    /// Checks if the player with the specified symbol has won the game on the given board.
+    ///
+    pub def checkWin(board: Board, symbol: Symbol): Bool = match board {
+        case Board(vec) => {
+            let patterns = (0, 1, 2) :: (3, 4, 5) :: (6, 7, 8) :: (0, 3, 6) :: (1, 4, 7) :: (2, 5, 8) :: (0, 4, 8) :: (2, 4, 6) :: Nil;
+            List.foldLeft(
+                (acc: Bool, pattern: (Int32, Int32, Int32)) -> {
+                    match pattern {
+                        case (a, b, c) => acc or
```
mlutze: Instead of `acc or` you can do `List.exists`

examples/apps/tic-tac-toe/src/Interface.flix
```diff
@@ -0,0 +1,68 @@
...
+                    Console.println("Invalid move. Move must be a positive integer between 0 and 8");
+                    getMoveFromConsole()
+                }
+                case Result.Ok(move) => {
+                    if(not(move >= 0 and move <= 8)) {
+                        Console.println("Invalid move. Move must be in between 0 and 8");
+                        getMoveFromConsole()
+                    } else {
+                        move
+                    }
```
mlutze: Can this be collapsed into one case with a guard?

```
case Result.Ok(move) if (0 <= move and move <= 8) =>
```

examples/apps/tic-tac-toe/src/Board.flix
```diff
@@ -0,0 +1,71 @@
...
+    pub def newBoard(): Board = {
+        Board(Vector.repeat(9, Symbol.Empty))
+    }
+
+    ///
+    /// Makes a move on the board at the specified index with the given symbol.
+    ///
+    /// Returns the updated board if the move is valid, otherwise returns the original board.
+    ///
+    pub def makeMove(board: Board, symbol: Symbol, index: Int32): Result[Board, Board]  = match board {
```
mlutze: It looks like the value of the error case is never used. Should this be an Option instead?

examples/apps/tic-tac-toe/flix.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name        = "tic-tac-toe"
+description = "test"
```
mlutze: Short description please :)

mlutze (commented):
Just a few final touches and then I think we can merge this.

examples/apps/tic-tac-toe/src/Board.flix
gaganchandan: Yes, that seems more appropriate.

gaganchandan:
@mlutze I've made the changes.

examples/apps/tic-tac-toe/src/Interface.flix
```diff
@@ -0,0 +1,61 @@
...
+    ///
+    /// In other words, re-interprets the Interface effect using the IO and Console effects.
+    ///
+    pub def interfaceWithIOAndConsole(f: Unit -> a \ ef): a \ ef - Interface + IO + Console =
+        run {
+            f()
+        }
+        with handler Interface {
+            def display(board, msg, resume) = {
+                println("");
```
mlutze: If we change this to `Console.println` then we only have the `Console` effect, right?

mlutze (commented):
Looks good. Just that one fix and we'll merge.

examples/apps/tic-tac-toe/src/Interface.flix
gaganchandan: I've made the fix

mlutze (approved):

magnus-madsen:
🚀

PR #10719, opened 2025-06-01
feat: add `Priority`s to some `Completion`s

magnus-madsen:
Refactor


magnus-madsen:
`Completion.scala` is not formatted. It needs to be, but should be independent PR.

PR #10726, opened 2025-06-02
refactor: turn `ZhegalkinCache` into a class

magnus-madsen:


main/src/ca/uwaterloo/flix/language/phase/unification/EffUnification3.scala
```diff
@@ -429,6 +429,8 @@ object EffUnification3 {
     implicit val renv: RigidityEnv = RigidityEnv.empty
     implicit val bimap: SortedBimap[Atom, Int] = mkBidirectionalVarMap(Atom.getAtoms(tpe))
 
+    implicit val alg: ZhegalkinAlgebra.type = ZhegalkinAlgebra
```
magnus-madsen, line 432: I use .type here and elsewhere because ZhegalkinAlgebra is an object.

Next step is to make it into a class too.

magnus-madsen:
No measurable impact on performance.

PR #10740, opened 2025-06-03
refactor: BOT Defn

JonathanStarup:


main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
```diff
@@ -654,6 +655,8 @@ object BackendObjType {
     }
   }
 
+  case class Defn(sym: Symbol.DefnSym) extends BackendObjType
```
magnus-madsen, line 658: Why are we adding such empty classes-- should we not be moving in the opposite direction?

magnus-madsen: Like this also costs object allocation

JonathanStarup:
Yeah this is for uniformity - i will extract the three remaining "old" classes (fun / namespace / effects) into the BOT system. Then the backend is fully uniform, without erasure, and ready to be rewritten 😄

PR #10741, opened 2025-06-03
refactor: remove `BackendType.asErasedBackendType`

JonathanStarup:
This function is the same as `toBackendType` but with an assert that the type is an erased type. This was merely a stepping stone since `toBackendType` is newer.

This is a step towards the backend to be erasure oblivious

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/jvm/GenEffectClasses.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/jvm/GenEffectClasses.scala
@@ -48,7 +48,7 @@ import org.objectweb.asm.Opcodes.*
   */
 object GenEffectClasses {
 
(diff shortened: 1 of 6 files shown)


main/src/ca/uwaterloo/flix/language/phase/jvm/GenEffectClasses.scala
```diff
@@ -86,7 +86,7 @@ object GenEffectClasses {
     mv.visitEnd()
   }
 
-  private def genFieldAndMethod(visitor: ClassWriter, effectName: JvmName, op: Op): Unit = {
```
JonathanStarup, line 89: These implicit root arguments are used to convert Enum types. Now that we don't assume erased types, these are required (but unused currently because the types are indeed erased)

magnus-madsen (approved):

PR #10659, opened 2025-05-23
feat: add static method calls

jaschdoc:
Depends on https://github.com/flix/flix/pull/10607
Closes https://github.com/flix/flix/issues/10480

PR #10751, opened 2025-06-03
fix: check control purity of target function in `Reducer`

jaschdoc:
Related to https://github.com/flix/flix/pull/10659

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
@@ -88,14 +88,14 @@ object Reducer {
         Expr.ApplyAtomic(op, es, tpe, purity, loc)
 
       case Expr.ApplyClo(exp1, exp2, ct, tpe, purity, loc) =>
-        if (ct == ExpPosition.NonTail && Purity.isControlImpure(purity)) lctx.addPcPoints(1)
+        if (ct == ExpPosition.NonTail && Purity.isControlImpure(purity)) lctx.addPcPoint()
         val e1 = visitExpr(exp1)
         val e2 = visitExpr(exp2)
         Expr.ApplyClo(e1, e2, ct, tpe, purity, loc)
 
       case Expr.ApplyDef(sym, exps, ct, tpe, purity, loc) =>
         val defn = root.defs(sym)
-        if (ct == ExpPosition.NonTail && Purity.isControlImpure(defn.expr.purity)) lctx.addPcPoints(1)
+        if (ct == ExpPosition.NonTail && Purity.isControlImpure(defn.expr.purity)) lctx.addPcPoint()
         val es = exps.map(visitExpr)
         Expr.ApplyDef(sym, es, ct, tpe, purity, loc)
 
@@ -146,7 +146,7 @@ object Reducer {
         Expr.TryCatch(e, rs, tpe, purity, loc)
 
       case Expr.RunWith(exp, effUse, rules, ct, tpe, purity, loc) =>
-        if (ct == ExpPosition.NonTail) lctx.addPcPoints(1)
+        if (ct == ExpPosition.NonTail) lctx.addPcPoint()
         val e = visitExpr(exp)
         val rs = rules.map {
           case HandlerRule(op, fparams, body) =>
@@ -156,7 +156,7 @@ object Reducer {
         Expr.RunWith(e, effUse, rs, ct, tpe, purity, loc)
 
       case Expr.Do(op, exps, tpe, purity, loc) =>
-        lctx.addPcPoints(1)
+        lctx.addPcPoint()
         val es = exps.map(visitExpr)
         Expr.Do(op, es, tpe, purity, loc)
 
@@ -190,9 +190,9 @@ object Reducer {
     /**
       * Adds n to the private [[pcPoints]] field.
       */
-    def addPcPoints(n: Int): Unit = {
+    def addPcPoint(): Unit = {
       if (isControlImpure) {
-        pcPoints += n
+        pcPoints += 1
       }
     }
 


main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -724,10 +724,10 @@ object GenExpression {
         val innerType = tpe.asInstanceOf[MonoType.Array].tpe
         val backendType = BackendType.toBackendType(innerType)
         val fillMethod = ClassMaker.StaticMethod(JvmName.Arrays, "fill", mkDescriptor(BackendType.Array(backendType.toErased), backendType.toErased)(VoidableType.Void))
-        pushExpr(exp1) ~                                                      // default
-          pushExpr(exp2) ~                                                    // default, length
```
JonathanStarup: It'd be nice if this PR was without re-formatting (you can make another PR if you want)

main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
```diff
@@ -74,7 +74,7 @@ object Reducer {
       Def(ann, mod, sym, cparams, fparams, ls, pcPoints, e, tpe, unboxedType, loc)
   }
 
-  private def visitExpr(exp0: Expr)(implicit lctx: LocalContext, ctx: SharedContext): Expr = {
+  private def visitExpr(exp0: Expr)(implicit lctx: LocalContext, root: Root, ctx: SharedContext): Expr = {
```
JonathanStarup, line 77: why is the root needed for this?

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -1169,7 +1169,8 @@ object GenExpression {
               mv.visitLabel(afterUnboxing)
 
             case DirectContext(_, _, _) =>
-              throw InternalCompilerException("Unexpected direct method context in control impure function", loc)
+              // Required for functions that use purity casts
+              mv.visitByteIns(BackendObjType.Result.unwindSuspensionFreeThunk("in pure function call", loc))
```
JonathanStarup: Is this because a control-pure function can call a control-impure function under a cast?

jaschdoc: Exactly. See `Assert.eq`

main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
jaschdoc: To look up `sym` from `ApplyDef` in root

jaschdoc:
!!!rerun

magnus-madsen:
What does this do?

magnus-madsen:
And why is so much code changed? :-/

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
JonathanStarup: its a bit annoying but i think we should look into that problem to see if we can fix it instead of introducing a band-aid here

main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
```diff
@@ -94,7 +94,8 @@ object Reducer {
         Expr.ApplyClo(e1, e2, ct, tpe, purity, loc)
 
       case Expr.ApplyDef(sym, exps, ct, tpe, purity, loc) =>
-        if (ct == ExpPosition.NonTail && Purity.isControlImpure(purity)) lctx.pcPoints += 1
```
JonathanStarup, line 97: @magnus-madsen the current code looks at the whole effect of the function call, i.e. the effects of the functions __and__  the effects of the arguments. This PR is changing it to look at the actual function effect. The current code thinks more things are control-impure than there actually are

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
jaschdoc: Agreed. It seems to just be `Assert.eq`. We should rewrite it to handle its effects locally and then cast to pure, but that does not solve the problem that any program with `unchecked_cast(effect() as _ \ {})` crashes the compiler.

jaschdoc: However, it seems like a compiler crash is preferable to weird byte code.

jaschdoc: Maybe we could raise an exception here or in the reducer?

jaschdoc: Or maybe we disallow purity casts altogether and only use `checked_ecast` for that? Looking through the code, it seems we can get away with that. Maybe `checked_ecast` would have to be a bit more expressive, i.e., allowing a desired effect and then checking that it is only and upcast to primitive effects! Additionally, a test should always have IO due to its natural interaction when reporting pass or fail. I think also we can easily have the Environment effect handler not have `Sys` but only `IO`.

main/src/library/Assert.flix
```diff
@@ -25,11 +25,13 @@ mod Assert {
             true
         else {
             use Chalk.{green, red};
-            println("Assertion Error");
-            println("  Expected: ${green(expected)}");
-            println("  Actual:   ${red(actual)}");
-            println("");
-            ?assertEq
+            Environment.runWithIO(() -> {
```
magnus-madsen: Why this change?

magnus-madsen:
@JonathanStarup If you are happy, you can merge.

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
JonathanStarup: For `Assert.eq`, that function is just wrong and should be fixed (although its not great that it results in a compiler crash

For `unchecked_cast(effect() as _ \ {})` i will think a bit

main/src/library/Assert.flix
JonathanStarup: Assert.eq has unhandled effects (`Chalk.green` has effect `Environment`) so it should be handled into a primitive effect before it can be ignored.

@jaschdoc this fix should probably be its own PR, and should also add a testcase that uses a failing call to `Assert.eq`. Not saying you should do it but still.

jaschdoc: Makes sense

main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -1161,7 +1161,7 @@ object GenExpression {
               mv.visitLabel(afterUnboxing)
 
             case DirectContext(_, _, _) =>
-              throw InternalCompilerException("Unexpected direct method context in control impure function", loc)
+              throw InternalCompilerException(s"Unexpected direct method context in control impure function: $sym", loc)
```
JonathanStarup: so this should be undwind suspension free thunk

JonathanStarup: but it also means that reducer should not give pcpoints to control-impure things inside control-pure functions

main/src/ca/uwaterloo/flix/language/phase/Reducer.scala
```diff
@@ -176,15 +177,30 @@ object Reducer {
...
     *
     * @param lparams the bound variables in the def.
     */
-  private case class LocalContext(lparams: mutable.ArrayBuffer[LocalParam], var pcPoints: Int)
+  private case class LocalContext(lparams: mutable.ArrayBuffer[LocalParam], private var pcPoints: Int, private val isControlImpure: Boolean) {
+
+    /**
+      * Adds n to the private [[pcPoints]] field.
+      */
+    def addPcPoints(n: Int): Unit = {
```
JonathanStarup: isn't this always 1? i'd prefer `addPcPoint()`

jaschdoc: ok

JonathanStarup:
@magnus-madsen This PR works - you can review for style or just merge as is

magnus-madsen (approved):

magnus-madsen:
> @magnus-madsen This PR works - you can review for style or just merge as is

I will just have to trust you on that :)

PR #10754, opened 2025-06-03
refactor: introduce Zhegalkin `Domain`

magnus-madsen:


magnus-madsen:
Near term I will merge Domain and ZhegalkinCst.

JonathanStarup (approved):

PR #10763, opened 2025-06-04
fix: purity cast in `Assert.flix` after handling AE

jaschdoc:
The cast comes **after** control impure effects have been handled so now only effect "annotations", i.e., the primitive effects are cast to pure.

### DIFF
--- a/main/src/library/Assert.flix
+++ b/main/src/library/Assert.flix
@@ -25,11 +25,13 @@ mod Assert {
             true
         else {
             use Chalk.{green, red};
-            println("Assertion Error");
-            println("  Expected: ${green(expected)}");
-            println("  Actual:   ${red(actual)}");
-            println("");
-            ?assertEq
+            Environment.runWithIO(() -> {
+                println("Assertion Error");
+                println("  Expected: ${green(expected)}");
+                println("  Actual:   ${red(actual)}");
+                println("");
+                ?assertEq
+            })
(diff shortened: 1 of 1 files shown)


JonathanStarup:
is it possible to add a test where `Assert.eq` runs this code? just to avoid future regression

jaschdoc:
> is it possible to add a test where `Assert.eq` runs this code? just to avoid future regression

Wait I'm confused. What do you mean "run this code"?

JonathanStarup:
add a test to TestAssert (if that exists) which expect `Assert.eq` to fail - thus verifying that the fail code doesn't crash the compiler

magnus-madsen:
A test would be good

jaschdoc:
A scala test or flix test? A Flix test seems weird since it unwraps the program which always causes a test failure

JonathanStarup:
I was thinking a flix program but youre right that it would just crash. 

I guess we just merge, no test

jaschdoc:
roger

magnus-madsen:
What no?

magnus-madsen:
You can add a test to a Scala suite that checks the compiler does not crash.

magnus-madsen:
The bytecode should still be valid, right?

jaschdoc:
roger

magnus-madsen (approved):

PR #10764, opened 2025-06-04
refactor: Ns class in BOT

JonathanStarup:


JonathanStarup:
In a follow up i can remove the erasure from the Ns class. I don't think there is any reason to use erasure here

magnus-madsen (approved):

PR #10765, opened 2025-06-04
refactor: rename `BoolAlg` to `FreeBoolAlg`

magnus-madsen:


magnus-madsen:
I admit that I am not 100% sure this is the right name.

The point is that we will have both BoolAlg and FreeBoolAlg trait (CofiniteSet is an instance of the former whereas ZhegalkinAlgebra is an instance of the latter).

jaschdoc (approved):

magnus-madsen:
Changed my mind.

PR #10773, opened 2025-06-05
feat: add `Bool2` lattice

magnus-madsen:


main/src/ca/uwaterloo/flix/language/phase/unification/shared/Bool2.scala
```diff
@@ -0,0 +1,51 @@
...
+
+package ca.uwaterloo.flix.language.phase.unification.shared
+
+/**
+  * Represents the two-valued Boolean algebra B2.
+  */
+sealed trait Bool2
+
+object Bool2 {
+  case object False extends Bool2
```
JonathanStarup, line 25: You can't use inbuilt bools?

magnus-madsen: We can, but it makes no difference.

magnus-madsen: In fact it is probably more cumbersome because of boxing etc.

main/src/ca/uwaterloo/flix/language/phase/unification/shared/CofiniteIntSet.scala
```diff
@@ -50,7 +50,6 @@ object CofiniteIntSet {
     if (s.isEmpty) Empty else Set(s)
   }
 
-  /** An instance for [[CofiniteIntSet]]. */
```
mlutze, line 53: ?

mlutze (approved):

main/src/ca/uwaterloo/flix/language/phase/unification/shared/Bool2.scala
JonathanStarup: It just seems a bit unnecessary, you have to write `t1 == True || t2 == True` instead of `t1 || t2` etc.

> but it makes no difference.

I'd be impressed if thats true for scala

PR #10776, opened 2025-06-06
refactor: backend of the backend

JonathanStarup:
This separates generating classes and loading/outputting classes

This allows @CadeMichael's prototype to replace `JvmBackend` easier and reuse `Loader`


main/src/ca/uwaterloo/flix/language/phase/jvm/JvmBackend.scala
```diff
@@ -153,66 +149,9 @@ object JvmBackend {
...
-        JvmOps.writeClass(flix.options.output.get.resolve("class/"), jvmClass)
-      }
-    }
-
-    // Collect code size for performance tracking.
-    val outputBytes = allClasses.map(_._2.bytecode.length).sum
-
-    if (flix.options.loadClassFiles) {
-      val main = Loader.load(allClasses)
-      new CompilationResult(root, main, getCompiledDefs(root), flix.getTotalTime, outputBytes)
```
JonathanStarup, line 171: Its really weird to call `flix.getTotalTime` inside of a call to `flix.phase` - that's why I introduced `Loader.LoaderResult`

main/src/ca/uwaterloo/flix/api/Flix.scala
```diff
@@ -681,7 +681,9 @@ class Flix {
     val eraserAst = Eraser.run(tailPosAst)
     val reducerAst = Reducer.run(eraserAst)
     val varOffsetsAst = VarOffsets.run(reducerAst)
-    val result = JvmBackend.run(varOffsetsAst)
+    val (backendAst, classes) = JvmBackend.run(varOffsetsAst)
+    val (loadedAst, loadRes) = Loader.run(backendAst, classes)
+    val result = new CompilationResult(loadedAst, loadRes.main, loadRes.defs, flix.getTotalTime, loadRes.byteSize)
```
JonathanStarup, line 687: now `flix.getTotalTime` is outside of a phase.

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmBackend.scala
magnus-madsen: Seems reasonable (the change)

magnus-madsen:
Conceptually writing the class files is not a loading/linking thing.

Would it make sense to rename the phase or split it into two?

JonathanStarup:
Yeah `Loader` is not the best name. Both are fine, but splitting them would create very small phases.

I thought to call it `BytecodeHandler` but yeah this phase is just "handling the bytecode", don't know what a good name is

magnus-madsen:
> Yeah `Loader` is not the best name. Both are fine, but splitting them would create very small phases.
> 
> I thought to call it `BytecodeHandler` but yeah this phase is just "handling the bytecode", don't know what a good name is

Small phases; but no computational cost? I would split them in two. ClassWriter and Loader?

magnus-madsen:
`ClassWriter`, `ClassLoader`
`BytecodeWriter`, `BytecodeLoader`

JonathanStarup:
fixed

magnus-madsen:
@JonathanStarup CI failures

JonathanStarup:
I've renamed them away from `ClassLoader`/`ClassWriter` to avoid so many name clashes

magnus-madsen (approved):

magnus-madsen:
@JonathanStarup I am thinking of undoing this PR unless the performance difference can be explained.

EDIT: I see that the time is not included. It was clearly bugged.

JonathanStarup:
The code was not taking JvmBackend into account for the total time - it is now

magnus-madsen:
> The code was not taking JvmBackend into account for the total time - it is now

Yeah, see my edit.

magnus-madsen:
We should still not treat ClassLoader and ClassWriter as phases though.

magnus-madsen:
Unfortunately this shows that the backend is extremely slow compared to the rest of the compiler :/

PR #10786, opened 2025-06-07
feat: exclude class loading/writing in total compilation time

JonathanStarup:


magnus-madsen (approved):

magnus-madsen:
@JonathanStarup I don't think this fixed the problem on arewefast

JonathanStarup:
I'm not sure there is a problem, I think the previous time was excluding JvmBackend, which also matches that throughput fell around 50%

magnus-madsen:
> I'm not sure there is a problem, I think the previous time was excluding JvmBackend, which also matches that throughput fell around 50%

Why was there then a phase called JvmBackend that was going up and down with Jakobs changes?

magnus-madsen:
Make JvmLoader and JvmWriter into non-phases and the problem goes away.

JonathanStarup:
@magnus-madsen the total time is only used for throughput - all the other graphs uses different numbers

The problem is that the previous code called `getTotalTime` before the `JvmBackend` timer had been added to the list of phases

PR #10788, opened 2025-06-07
feat: check for concurrent phases in `Flix`

JonathanStarup:
We don't usually do this, @magnus-madsen you can merge or close as you wish


magnus-madsen:
No, we cannot just add such adhoc checks

Issue #10772, opened 2025-06-05
Implement InlayHintProvider to show effects


magnus-madsen:
The simple idea is that inlay hints about effects should be provided on the right:

```
def main(): Unit \ {Net, IO} =
    run {
        let url = "http://example.com/";
        Logger.info("Downloading URL: '${url}'");                                 // {Logger}
        match HttpWithResult.get(url, Map.empty()) {                          // {HttpWithResult}
            case Result.Ok(response) =>
                let file = "data.txt";
                Logger.info("Saving response to file: '${file}'");               // {Logger}
                let body = Http.Response.body(response);
                match FileWriteWithResult.write(str = body, file) {
                    case Result.Ok(_) =>
                        Logger.info("Response saved to file: '${file}'")
                    case Result.Err(err) =>
                        Logger.fatal("Unable to write file: '${err}'")
                }
            case Result.Err(err) =>
                Logger.fatal("Unable to download URL: '${err}'")
        }
    } with FileWriteWithResult.runWithIO
      with HttpWithResult.runWithIO
      with Logger.runWithIO
```

and so on.

The starting point is `InlayHintProvider.scala`

magnus-madsen:
@gaganchandan Does this make sense? 

For simplicitly, you can use our `Visitor` to find all EffSymUse in the current file.

Then you create InlayHints on the right.

To test, you can start by just creating an inlay hint on line 1 at offset e.g. 100

gaganchandan:
I have been looking into it and there is a challenge. I'm using the `Visitor` with a consumer defined as follows:

```scala
    object opSymUseConsumer extends Consumer {
      override def consumeOpSymUse(opSymUse: SymUse.OpSymUse): Unit = {
        opSymUses = opSymUse :: opSymUses
        val line = opSymUse.loc.beginLine
        var hint = InlayHint(
          position = Position(line, 100),
          label = opSymUse.sym.eff.name,
          kind = Some(InlayHintKind.Type),
          textEdits = List.empty,
          tooltip = s"Effect: ${opSymUse.sym.eff.name}",
          paddingLeft = true,
          paddingRight = true
        )
      inlayHints = hint :: inlayHints
      }


```

This works perfectly for operations defined in the effect's definition, but not for other effectful operations defined in the module with the same name as the effect. For example, consider `Logger.flix`:
```scala
pub eff Logger {

    ///
    /// Logs the given message `m` at the given severity `s`.
    ///
    def log(s: Severity, m: String): Unit

}

[...]

PR #10792, opened 2025-06-08
feat: inlay hints to show effects

gaganchandan:
Adds inlay hints to show effects as discussed in #10772. Currently, in only displays hints for operations defined in the effect's definition. Example screenshot:

image

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
@@ -15,13 +15,33 @@
  */
 package ca.uwaterloo.flix.api.lsp.provider
 
-import ca.uwaterloo.flix.api.lsp.{InlayHint, Range}
+import ca.uwaterloo.flix.api.lsp.acceptors.{FileAcceptor}
+import ca.uwaterloo.flix.language.ast.shared.SymUse
+import ca.uwaterloo.flix.language.ast.TypedAst.{Expr, Root}
+import ca.uwaterloo.flix.api.lsp.{Consumer, InlayHint, InlayHintKind, Position, Range}
+import ca.uwaterloo.flix.api.lsp.Visitor
 
 object InlayHintProvider {
+  def getInlayHints(uri: String, range: Range)(implicit root: Root): List[InlayHint] = {
+    var inlayHints: List[InlayHint] = Nil
+    object opSymUseConsumer extends Consumer {
+      override def consumeOpSymUse(opSymUse: SymUse.OpSymUse): Unit = {
+        val line = opSymUse.loc.beginLine
+        var hint = InlayHint(
+          position = Position(line, 100),
+          label = opSymUse.sym.eff.name,
+          kind = Some(InlayHintKind.Type),
+          textEdits = List.empty,
+          tooltip = s"Effect: ${opSymUse.sym.eff.name}",
+          paddingLeft = true,
+          paddingRight = true
+        )
+      inlayHints = hint :: inlayHints
+      }
+    }
 
-  def getInlayHints(uri: String, range: Range): List[InlayHint] = {
-    // We currently do not support any inlay hints.
-    Nil
+    Visitor.visitRoot(root, opSymUseConsumer, FileAcceptor(uri))
+
+    inlayHints
   }
-
 }
--- a/main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/LspServer.scala
@@ -410,7 +410,7 @@ object LspServer {
     override def inlayHint(params: InlayHintParams): CompletableFuture[util.List[InlayHint]] = {
       val uri = params.getTextDocument.getUri
       val range = Range.fromLsp4j(params.getRange)
-      val hints = InlayHintProvider.getInlayHints(uri, range)
+      val hints = InlayHintProvider.getInlayHints(uri, range)(flixLanguageServer.root)
       CompletableFuture.completedFuture(hints.map(_.toLsp4j).asJava)
     }
 
--- a/main/src/ca/uwaterloo/flix/api/lsp/VSCodeLspServer.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/VSCodeLspServer.scala
@@ -123,7 +123,7 @@ class VSCodeLspServer(port: Int, o: Options) extends WebSocketServer(new InetSoc
     parseRequest(data) match {
       case Ok(request) =>
         val t = System.nanoTime()
-        val result = processRequest(request)(ws)
+        val result = processRequest(request)(ws, root)
         if (ws.isOpen) {
           val jsonCompact = JsonMethods.compact(JsonMethods.render(result))
           // val jsonPretty = JsonMethods.pretty(JsonMethods.render(result))
@@ -212,7 +212,7 @@ class VSCodeLspServer(port: Int, o: Options) extends WebSocketServer(new InetSoc
   /**
     * Process the request.
     */
-  private def processRequest(request: Request)(implicit ws: WebSocket): JValue = request match {
+  private def processRequest(request: Request)(implicit ws: WebSocket, root: Root): JValue = request match {
 
     case Request.AddUri(id, uri, src) =>
       addSourceCode(uri, src)


main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
```diff
@@ -15,13 +15,33 @@
...
+import ca.uwaterloo.flix.api.lsp.Visitor
 
 object InlayHintProvider {
+  def getInlayHints(uri: String, range: Range)(implicit root: Root): List[InlayHint] = {
+    var inlayHints: List[InlayHint] = Nil
+    object opSymUseConsumer extends Consumer {
+      override def consumeOpSymUse(opSymUse: SymUse.OpSymUse): Unit = {
+        val line = opSymUse.loc.beginLine
+        var hint = InlayHint(
+          position = Position(line, 100),
```
magnus-madsen: What happens if there is already code at col 100?

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
```diff
@@ -15,13 +15,33 @@
...
+import ca.uwaterloo.flix.api.lsp.acceptors.{FileAcceptor}
+import ca.uwaterloo.flix.language.ast.shared.SymUse
+import ca.uwaterloo.flix.language.ast.TypedAst.{Expr, Root}
+import ca.uwaterloo.flix.api.lsp.{Consumer, InlayHint, InlayHintKind, Position, Range}
+import ca.uwaterloo.flix.api.lsp.Visitor
 
 object InlayHintProvider {
+  def getInlayHints(uri: String, range: Range)(implicit root: Root): List[InlayHint] = {
+    var inlayHints: List[InlayHint] = Nil
+    object opSymUseConsumer extends Consumer {
```
magnus-madsen: I would separate the two concerns: finding where to create an inline hint and creating the InlayHint. 

Here you have them in the same code. But they are really different things. 

Here is an example: https://github.com/flix/flix/blob/master/main/test/ca/uwaterloo/flix/api/lsp/TestCompletionProvider.scala#L589

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
```diff
@@ -15,13 +15,33 @@
...
+      override def consumeOpSymUse(opSymUse: SymUse.OpSymUse): Unit = {
+        val line = opSymUse.loc.beginLine
+        var hint = InlayHint(
+          position = Position(line, 100),
+          label = opSymUse.sym.eff.name,
+          kind = Some(InlayHintKind.Type),
+          textEdits = List.empty,
+          tooltip = s"Effect: ${opSymUse.sym.eff.name}",
```
magnus-madsen: What text should we actually show here for it to be readable?

magnus-madsen:
It would be nice if the columns aligned.

magnus-madsen:
Next you could check ApplyDef to catch those calls that indirectly cause effects.

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
```diff
@@ -15,13 +15,33 @@
...
 package ca.uwaterloo.flix.api.lsp.provider
 
-import ca.uwaterloo.flix.api.lsp.{InlayHint, Range}
+import ca.uwaterloo.flix.api.lsp.acceptors.{FileAcceptor}
+import ca.uwaterloo.flix.language.ast.shared.SymUse
+import ca.uwaterloo.flix.language.ast.TypedAst.{Expr, Root}
+import ca.uwaterloo.flix.api.lsp.{Consumer, InlayHint, InlayHintKind, Position, Range}
+import ca.uwaterloo.flix.api.lsp.Visitor
 
 object InlayHintProvider {
```
magnus-madsen, line 31: Because this feature - at least to begin with - is for us, let us add a boolean:

private val EnableEffectHints: Boolean = false

and  then use that in the body of getInlayHints

magnus-madsen:
But this is some good progress.

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
gaganchandan: The hint replaces the code :(. But I've just been using 100 as a placeholder.

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
gaganchandan: In the tooltip? Maybe something like `Operation with effect xyz performed here`? Maybe it could have the name of the enclosing function as well so the user could match performed and declared effects.

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
magnus-madsen: Can we fix that? Or compute the maximum width?

gaganchandan: Using `consumeExpr` and matching against `Expr.Do` instead of `consumeOpSymUse` gives the end of the line. It looks like this:
image
The hint is always at the end of the line regardless of the column.

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
gaganchandan: Do you want me to collect every location that needs a hint in one place in one method and use another to create all the hints at a time? Or should I create a separate method (something like `mkInlayHint` and call it to construct the hint each time a suitable location is found?

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
magnus-madsen: I would like:

Alignment
Not to run into text

Can we do that?

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
magnus-madsen: Compute locations in one place and convert them to hints in another place

gaganchandan: Something like this?:

```scala
  def getInlayHints(uri: String, range: Range)(implicit root: Root): List[InlayHint] = {
    var inlayHints: List[InlayHint] = Nil
    if(EnableEffectHints) {
      var opSymUses: Set[(SymUse.OpSymUse, SourceLocation)] = getOpSymUses(uri)
      opSymUses.foreach { case (opSymUse, loc) =>
          var hint = InlayHint(
            position = Position(loc.endLine, loc.endCol + 1),
            label = opSymUse.sym.eff.name,
            kind = Some(InlayHintKind.Type),
            textEdits = List.empty,
            tooltip = s"Effect: ${opSymUse.sym.eff.name}",
            paddingLeft = true,
            paddingRight = true
          )
          inlayHints = hint :: inlayHints
      }
    }
    inlayHints
  }

  def getOpSymUses(uri: String)(implicit root: Root): Set[(SymUse.OpSymUse, SourceLocation)] = {
    var opSymUses: Set[(SymUse.OpSymUse, SourceLocation)] = Set.empty
    object opSymUseConsumer extends Consumer {
      override def consumeExpr(expr: Expr): Unit = {
        expr match {
          case Expr.Do(opSymUse, _, _, _, loc) =>
            opSymUses += ((opSymUse, loc))
          case _ => ()
        }
      }
    }
    Visitor.visitRoot(root, opSymUseConsumer, FileAcceptor(uri))
    opSymUses
  }
```

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
gaganchandan: Alignment as in having each hint starting at the same column?

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
magnus-madsen: Sure close, but opSymUses.foreach can be in its own function too right?

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
magnus-madsen: Maybe it should just be { effect1, effect2 } to start with

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
magnus-madsen: Easier to comment on the code when its in the PR

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
magnus-madsen: Yep.

So compute the longest column and then offset a bit from there?

gaganchandan: That would work. However, I'm not sure if it's the best idea. If there was to be really long line then no hints would be visible without having to scroll, and you would have to scroll every time you want to see the hint even for shorter lines.

magnus-madsen: So, can you do better? You want to compute the max col for the function?

magnus-madsen: I think my proposal is a decent starting point

magnus-madsen: Actually, I would propose to start with: min(120, computedMaxColWidth)

gaganchandan: > Actually, I would propose to start with: min(120, computedMaxColWidth)

This makes sense. I'll implement this for now.

gaganchandan: This would only be for lines ending before column 120, right? Otherwise the hint would replace the code again. What about for those greater than 120? Should the hint be at the end of the line again?

gaganchandan: There seems to another issue with the way VSCode renders the hints. It seems like it can only render hints where there is already text or right after. For instance, if we have a line of say 80 characters, it doesn't seem to be able to render a hint at column 120 because there's nothing between columns 80 and 120. If there were spaces until then it works, but not otherwise.

magnus-madsen: > There seems to another issue with the way VSCode renders the hints. It seems like it can only render hints where there is already text or right after. For instance, if we have a line of say 80 characters, it doesn't seem to be able to render a hint at column 120 because there's nothing between columns 80 and 120. If there were spaces until then it works, but not otherwise.

Are you sure?  What does the spec say? 

if true that seems disappointing.

gaganchandan: I haven't been able to find anything about this specifically in the spec. I'm trying to look into the code to see if there's any indication. For now, it is just what the behaviour seems to indicate.

gaganchandan:
> Next you could check ApplyDef to catch those calls that indirectly cause effects.

I've added this in the latest commit. Updated screenshot:
image

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
```diff
@@ -15,33 +15,41 @@
...
+  private val EnableEffectHints: Boolean = false
 
   def getInlayHints(uri: String, range: Range)(implicit root: Root): List[InlayHint] = {
     var inlayHints: List[InlayHint] = Nil
     if(EnableEffectHints) {
-      var opSymUses: Set[(SymUse.OpSymUse, SourceLocation)] = getOpSymUses(uri)
-      opSymUses.foreach { case (opSymUse, loc) => mkHintFromOpSymUse(opSymUse, loc) :: inlayHints }
+      var effects: Set[(String, Int, Int)] = (getOpSymUses(uri) ++ getDefSymUses(uri)).filter {
+        case (eff, _, _) =>
+          eff != "Pure"
```
magnus-madsen: This is not good :) You must never do string comparisons-- instead you can match on the effect.

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
```diff
@@ -15,33 +15,41 @@
...
+      lineToEffectsMap.foreach { case (line, effects) =>
+        inlayHints ::= mkHintFromEffects(line, effects, maxColWidth + 1)
+      }
     }
     inlayHints
   }
 
-  def getOpSymUses(uri: String)(implicit root: Root): Set[(SymUse.OpSymUse, SourceLocation)] = {
-    var opSymUses: Set[(SymUse.OpSymUse, SourceLocation)] = Set.empty
+  def getOpSymUses(uri: String)(implicit root: Root): Set[(String, Int, Int)] = {
```
magnus-madsen: Avoid use of String and Int, you want to return the symbol and source location

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
```diff
@@ -50,13 +58,29 @@ object InlayHintProvider {
     opSymUses
   }
 
-  def mkHintFromOpSymUse(opSymUse: SymUse.OpSymUse, loc: SourceLocation): InlayHint = {
+  def getDefSymUses(uri: String)(implicit root: Root): Set[(String, Int, Int)] = {
```
magnus-madsen: same comment

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
```diff
@@ -15,33 +15,41 @@
...
+          eff != "Pure"
+      }
+      var maxColWidth: Int = effects.map { case (_, _, col) => col }.maxOption.getOrElse(0)
+      var lineToEffectsMap: Map[Int, Set[String]] = Map.empty
+      effects.foreach { case (eff, line, col) =>
+        lineToEffectsMap = lineToEffectsMap.updated(line, lineToEffectsMap.getOrElse(line, Set.empty) + eff)
+      }
+      lineToEffectsMap.foreach { case (line, effects) =>
+        inlayHints ::= mkHintFromEffects(line, effects, maxColWidth + 1)
+      }
```
magnus-madsen, line 69: I would introduce a handful of functions for the above functionality. Also try to avoid imperative programming, but use functional programming.

magnus-madsen (commented):
I added some comments for the code

magnus-madsen:
> > Next you could check ApplyDef to catch those calls that indirectly cause effects.
> 
> I've added this in the latest commit. Updated screenshot: image

Looks good, but there is no alignment on the right?

gaganchandan:
> I added some comments for the code

I've tried fixing them in the latest commit. Let me know if it's better.

magnus-madsen:
> > I added some comments for the code
> 
> I've tried fixing them in the latest commit. Let me know if it's better.

I will let @mlutze give some feedback on the code style

magnus-madsen:
When computing the column width, you could also take this approach:

- Find the def we are inside. Get its expression body. Get that ones source location.
- This source location spans the whole function body.
- By iterating through each line, you could find the maximum width.

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
```diff
@@ -13,15 +13,86 @@
...
+          case _ => ()
+        }
+      }
+    }
+    Visitor.visitRoot(root, defSymUseConsumer, FileAcceptor(uri))
+    defSymUses
+  }
+
+  def mkHintsFromEffects(lineToEffectsMap: Map[Int, Set[Type]], maxColWidth: Int): List[InlayHint] = {
+    lineToEffectsMap.map({
```
mlutze: avoid extra paren:

```
lineToEffectsMap.map {
```

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
```diff
@@ -13,15 +13,86 @@
...
+      var lineToEffectsMap: Map[Int, Set[Type]] = Map.empty
+      effects.foreach { case (eff, loc) =>
+        lineToEffectsMap = lineToEffectsMap.updated(loc.endLine, lineToEffectsMap.getOrElse(loc.endLine, Set.empty) + eff)
+      }
+      inlayHints = mkHintsFromEffects(lineToEffectsMap, maxColWidth)
+    }
+    inlayHints
   }
 
+  def getOpSymUses(uri: String)(implicit root: Root): Set[(Type, SourceLocation)] = {
```
mlutze: `private` for everything that's not part of the interface

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
```diff
@@ -13,15 +13,86 @@
...
+
+  def mkHintsFromEffects(lineToEffectsMap: Map[Int, Set[Type]], maxColWidth: Int): List[InlayHint] = {
+    lineToEffectsMap.map({
+      case (line, effs) =>
+        mkHint(effs, line, maxColWidth)
+      }).toList
+  }
+
+  def mkHint(effs: Set[Type], line: Int, col: Int): InlayHint = {
+    var effectString: String = effs.mkString(" + ")
```
mlutze: `val`

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
```diff
@@ -13,15 +13,86 @@
...
+  }
+
+  def mkHintsFromEffects(lineToEffectsMap: Map[Int, Set[Type]], maxColWidth: Int): List[InlayHint] = {
+    lineToEffectsMap.map({
+      case (line, effs) =>
+        mkHint(effs, line, maxColWidth)
+      }).toList
+  }
+
+  def mkHint(effs: Set[Type], line: Int, col: Int): InlayHint = {
```
mlutze: Documentation on the shape that this makes would be nice.

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
```diff
@@ -13,15 +13,86 @@
...
+            defSymUses += ((eff, loc))
+          case _ => ()
+        }
+      }
+    }
+    Visitor.visitRoot(root, defSymUseConsumer, FileAcceptor(uri))
+    defSymUses
+  }
+
+  def mkHintsFromEffects(lineToEffectsMap: Map[Int, Set[Type]], maxColWidth: Int): List[InlayHint] = {
```
mlutze: Documentation on what `maxColWidth` precisely means. (E.g. does it include the hint?)

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
```diff
@@ -13,15 +13,86 @@
...
+  def getInlayHints(uri: String, range: Range)(implicit root: Root): List[InlayHint] = {
+    var inlayHints: List[InlayHint] = Nil
+    if(EnableEffectHints) {
+      var effects: Set[(Type, SourceLocation)] = (getOpSymUses(uri) ++ getDefSymUses(uri)).filter {
+        case (eff, loc) => eff match {
+          case Type.Pure => false
+          case _ => true
+        }
+      }
+      var maxColWidth: Int = Math.min(120, effects.map { case (_, loc) => loc.endCol }.maxOption.getOrElse(120))
```
mlutze: `val`

gaganchandan:
> When computing the column width, you could also take this approach:
> 
>     * Find the def we are inside. Get its expression body. Get that ones source location.
> 
>     * This source location spans the whole function body.
> 
>     * By iterating through each line, you could find the maximum width.

Do we still need the column width, seeing as it's not possible to use it the way we expected?

magnus-madsen:
> > When computing the column width, you could also take this approach:
> > ```
> > * Find the def we are inside. Get its expression body. Get that ones source location.
> > 
> > * This source location spans the whole function body.
> > 
> > * By iterating through each line, you could find the maximum width.
> > ```
> 
> Do we still need the column width, seeing as it's not possible to use it the way we expected?

No, since its not possible to use.

magnus-madsen:
@gaganchandan When you were experimenting with this, did you modify leftPadding or position?

gaganchandan:
> @gaganchandan When you were experimenting with this, did you modify leftPadding or position?

Yes. I tried modifying position by hardcoding the column number as well, but it behaved the same way it does now with the hint at the end of the line. `paddingLeft` simply adds a space between the last character and the hint to separate it visually.

magnus-madsen:
> > @gaganchandan When you were experimenting with this, did you modify leftPadding or position?
> 
> Yes. I tried modifying position by hardcoding the column number as well, but it behaved the same way it does now with the hint at the end of the line. `paddingLeft` simply adds a space between the last character and the hint to separate it visually.

So can you use the last source position of each line and pad it appropriately?

gaganchandan:
> > > @gaganchandan When you were experimenting with this, did you modify leftPadding or position?
> > 
> > 
> > Yes. I tried modifying position by hardcoding the column number as well, but it behaved the same way it does now with the hint at the end of the line. `paddingLeft` simply adds a space between the last character and the hint to separate it visually.
> 
> So can you use the last source position of each line and pad it appropriately?

I don't think the size of the padding can be modified. It is only a Boolean.

magnus-madsen:
> > > > @gaganchandan When you were experimenting with this, did you modify leftPadding or position?
> > > 
> > > 
> > > Yes. I tried modifying position by hardcoding the column number as well, but it behaved the same way it does now with the hint at the end of the line. `paddingLeft` simply adds a space between the last character and the hint to separate it visually.
> > 
> > 
> > So can you use the last source position of each line and pad it appropriately?
> 
> I don't think the size of the padding can be modified. It is only a Boolean.

Ah- its a Boolean!

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
```diff
@@ -13,15 +13,87 @@
...
+import scala.collection.immutable.Map
 
 object InlayHintProvider {
+  private val EnableEffectHints: Boolean = false
 
-  def getInlayHints(uri: String, range: Range): List[InlayHint] = {
-    // We currently do not support any inlay hints.
-    Nil
+  def getInlayHints(uri: String, range: Range)(implicit root: Root): List[InlayHint] = {
+    var inlayHints: List[InlayHint] = Nil
```
mlutze: Is mutation needed here? Looks like we can just do an if-else with Nil in one branch.

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
```diff
@@ -13,15 +13,87 @@
...
+    if(EnableEffectHints) {
+      val effects: Set[(Type, SourceLocation)] = (getOpSymUses(uri) ++ getDefSymUses(uri)).filter {
+        case (eff, loc) => eff match {
+          case Type.Pure => false
+          case _ => true
+        }
+      }
+      var locationToEffectsMap: Map[SourceLocation, Set[Type]] = Map.empty
```
mlutze: We generally prefer functional style for simple cases like this.

```scala
effects.foldLeft(Map.empty[SourceLocation, Set[Type]]) {
   ...
}
```

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
```diff
@@ -13,15 +13,87 @@
...
 object InlayHintProvider {
+  private val EnableEffectHints: Boolean = false
 
-  def getInlayHints(uri: String, range: Range): List[InlayHint] = {
-    // We currently do not support any inlay hints.
-    Nil
+  def getInlayHints(uri: String, range: Range)(implicit root: Root): List[InlayHint] = {
+    var inlayHints: List[InlayHint] = Nil
+    if(EnableEffectHints) {
+      val effects: Set[(Type, SourceLocation)] = (getOpSymUses(uri) ++ getDefSymUses(uri)).filter {
```
mlutze: This set seems suspicious to me. I am not sure that hashing types is well-defined.

I would expect instead perhaps a list of `(EffSym, SourceLocation)`.

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
```diff
@@ -13,15 +13,87 @@
...
+      var locationToEffectsMap: Map[SourceLocation, Set[Type]] = Map.empty
+      effects.foreach { case (eff, loc) =>
+        locationToEffectsMap = locationToEffectsMap.updated(loc, locationToEffectsMap.getOrElse(loc, Set.empty) + eff)
+      }
+      inlayHints = mkHintsFromEffects(locationToEffectsMap)
+    }
+    inlayHints
   }
 
+  private def getOpSymUses(uri: String)(implicit root: Root): Set[(Type, SourceLocation)] = {
```
mlutze: Why do we return a type here instead of just the `OpSym`?

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
mlutze: I see that it's perhaps unavoidable with the consumer/acceptor stuff, but here we can avoid it.

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
```diff
@@ -13,15 +13,87 @@
...
+            opSymUses += ((eff, loc))
+          case _ => ()
+        }
+      }
+    }
+    Visitor.visitRoot(root, opSymUseConsumer, FileAcceptor(uri))
+    opSymUses
+  }
+
+  private def getDefSymUses(uri: String)(implicit root: Root): Set[(Type, SourceLocation)] = {
```
mlutze: Same here; why do we return a type?

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
mlutze: Also, returning a set seems unnecessary; a list avoids comparison/hashing operations.

mlutze (commented):
I want to clarify: we're only looking for the concrete effects, right? Not variables?

gaganchandan:
Yep, only concrete ones

magnus-madsen:
> I want to clarify: we're only looking for the concrete effects, right? Not variables?

It would be good to cover rigid variables- but I could be a follow up. Perhaps after you have had a chance to discuss.

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
magnus-madsen: Whats the alternative?

magnus-madsen (commented):
Before sure to:

- Document each method with Scala comments /**/
- Reformat the file using intellij ideas formatter

magnus-madsen:
Oh, and post a screenshot of how it looks in the current form.

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
mlutze: at least rename the function :)

Better yet, keep this function, return the def sym uses, then transform them into the used effects in a separate function IMO.

main/src/ca/uwaterloo/flix/api/lsp/provider/InlayHintProvider.scala
gaganchandan: We can get an `EffSym` from `OpSymUse`, but I'm not sure if we can from `DefSymUse`. I thought it would be best to keep it uniform. Would a list of `(Type, SourceLocation)` work?

mlutze: I would write something like the following:

```scala
val opSyms = getOpSyms(...)
val opEffSyms = opSyms.map(... extract the eff syms ...)
val defSyms = getDefSyms(...)
val defEffSyms = defSyms.map(... extract the eff syms ...)
val effSyms = opEffSyms ++ defEffSyms
```


Extracting the effSyms from the defs is more complicated since you need to get it out of the effect of the function. But you cannot just use the effect itself, because that can be a big effect formula with variables, etc.

For example, if you have `Console + ef + ef2 + r`, you need to extract just the `Console` sym from this.

gaganchandan: Okay. I'll see if I can make that happen.

gaganchandan: Hmm. I'll see if I can make this happen.

magnus-madsen:
Hi @gaganchandan Lets resurrect. If you do one more pass on the code (improve best as you can and leave the rest as is) then I will merge this.

magnus-madsen (approved):

magnus-madsen:
@gaganchandan Looks good, but some build failures.

gaganchandan:
> @gaganchandan Looks good, but some build failures.

Yes, it's quite strange. I'll fix that and and some more changes suggested by Matt shortly.

gaganchandan:
Updated screenshot:
image

magnus-madsen (approved):

Issue #10722, opened 2025-06-02
Reuse and Cleanup `TypeError`


magnus-madsen:
Delete:
- IrreducibleAssocType
- MismatchedCaseSets
- MismatchedBools
- NonRecordType
- NonSchemaType
- RegionVarEscapes

Repurpose / reintroduce (try to follow this order):
- MissingInstanceArrow 
- UnexpectedEffect
- MismatchedEffects
- MismatchedArrowEffects
- UndefinedLabel (talk to Matt, extra label too?)
- UndefinedPred
- PossibleCheckedTypeCast
- (OccursCheck -- do this very last)

Ensure that there test cases for each error. Write them if necessary.

Future work:
- Split MissingInstanceEq/MissingInstanceOrder/MissingInstanceToString into two (ask Matt - variable vs. concrete)

magnus-madsen:
One PR for each. You can delete all the ones to be deleted in one PR though.

magnus-madsen:
@gaganchandan What's the status on this?

It should be easy to get the unused errors deleted-- will you create a PR?

As for UndefinedLabel/ExtraLabel/MissingLabel did you have a chance to look at that?

gaganchandan:
> [@gaganchandan](https://github.com/gaganchandan) What's the status on this?
> 
> It should be easy to get the unused errors deleted-- will you create a PR?
> 
> As for UndefinedLabel/ExtraLabel/MissingLabel did you have a chance to look at that?

I've just opened the PR. I'll look into `UndefinedLabel` next.

gaganchandan:
I have some doubts about undefined/missing/extra label. Consider the following program:

```scala
def undefinedLabel1(): {x = Int32, y = Int32}  = {
    let r = {x = 1, z = 3};
    r
}
```
Should this be an instance of both missing and extra labels? 

The way I understand it is that the follwing is an instance of missing/undefined label:
```scala
def undefinedLabel2(): {x = Int32, y = Int32}  = {
    let r = {y = 1};
    r
}
```

And this is an instance of extra label:
```scala
def undefinedLabel3(): {x = Int32, y = Int32}  = {
    let r = {x = 1, y = 2, z = 3};
    r
}
```

magnus-madsen:
> I have some doubts about undefined/missing/extra label. Consider the following program:
> 
> def undefinedLabel1(): {x = Int32, y = Int32}  = {
>     let r = {x = 1, z = 3};
>     r
> }
> 

[...]

PR #10793, opened 2025-06-08
refactor: remove unused type errors

gaganchandan:
Removes the unneeded type errors mention in #10722 .

`TestTyper.scala`, however, contains test cases with the titles indicating that they are for `RegionVarEscapes`. However, they expect any type error and not the specific one. I ran some of them and they throw other errors such as `UnexpectedArg`. Should I rename/repurpose these tests or delete them entirely since there already test cases for these other type errors?

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/errors/TypeError.scala
+++ b/main/src/ca/uwaterloo/flix/language/errors/TypeError.scala
@@ -34,31 +34,6 @@ sealed trait TypeError extends CompilationMessage {
 object TypeError {
 
   /**
-    * Irreducible associated type error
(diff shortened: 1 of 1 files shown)


magnus-madsen:
> TestTyper.scala, however, contains test cases with the titles indicating that they are for RegionVarEscapes. However, they expect any type error and not the specific one. I ran some of them and they throw other errors such as UnexpectedArg. Should I rename/repurpose these tests or delete them entirely since there already test cases for these other type errors?

We leave the tests and deal with them later.

magnus-madsen (approved):

PR #10801, opened 2025-06-08
feat: add `FiniteSet`

magnus-madsen:


magnus-madsen:
This will allow us to get rid of CaseSetUnification

JonathanStarup:
Why not just use set? Seems like a redundant object and Set is also finite

mlutze:
> Why not just use set? Seems like a redundant object and Set is also finite

I guess because Set does not have the universe baked in.

main/src/ca/uwaterloo/flix/language/phase/unification/shared/FiniteSet.scala
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+package ca.uwaterloo.flix.language.phase.unification.shared
+
+/**
+  * Represents a Boolean algebra over a finite set.
+  */
+object FiniteSet {
+
+  case class FiniteSet[T](s: Set[T])
```
mlutze, line 23: We should be doing companion object style, right?

magnus-madsen: Not sure, look at Bool2 and tell me how this should look

mlutze: > Not sure, look at Bool2 and tell me how this should look

That one does companion object:

```
sealed trait Bool2

object Bool2 {
```

Why not do the same?

magnus-madsen: There is no trait here??

PR #10809, opened 2025-06-08
feat: simplify ReifiedSourceLocation

JonathanStarup:
Instead of storing 5 fields, we just construct the string immediately.

Additionally, the end line and column were unused.


magnus-madsen:
This I am actually somewhat against, if we ever want to do anyform of dynamic analysis or instrumentation then we would want exact information

PR #10812, opened 2025-06-09
refactor: introduce LiteralInt and LiteralFloat Lexer tokens

mlutze:


magnus-madsen:
> We need LiteralInt for the `Pred/4` syntax, and we need LiteralFloat so that we don't all get confused.

Why? Are you not just creating literals that are overlapping? What is "123" today?

mlutze:
> > We need LiteralInt for the `Pred/4` syntax, and we need LiteralFloat so that we don't all get confused.
> 
> Why? Are you not just creating literals that are overlapping? What is "123" today?

Before: `LiteralInt32` is both `123i32` and `123`.

After: `LiteralInt32` s `123i32` and `LiteralInt` is `123`

JonathanStarup:
I like this change - its weird that 123 is a literalInt32

magnus-madsen:
OK good enough

magnus-madsen (approved):

PR #10816, opened 2025-06-09
experiment: ClassFileMigration

JonathanStarup:
@magnus-madsen here is a draft of the class file migration path. Most things are probably more complicated than this but it seems like a good path to take


main/src/ca/uwaterloo/flix/language/phase/jvm/ClassFileMigration.scala
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+  * This object enriches [[org.objectweb.asm.MethodVisitor]] to have similar methods to those of
+  * java.lang.classFile.CodeBuilder
+  * [[https://docs.oracle.com/en/java/javase/24/docs/api/java.base/java/lang/classfile/CodeBuilder.html]]
+  * from Java 24.
+  *
+  * This is intended to replace [[ca.uwaterloo.flix.language.phase.jvm.BytecodeInstructions.RichMethodVisitor]].
+  */
+object ClassFileMigration {
```
JonathanStarup, line 13: The outer object does nothing, but you cant make top-level things implicit 🤷

main/src/ca/uwaterloo/flix/language/phase/jvm/ClassFileMigration.scala
```diff
@@ -0,0 +1,29 @@
...
+
+  implicit class ClassFileMethodVisitor(mv: MethodVisitor) {
```
magnus-madsen: Would this be called CodeBuilder?

main/src/ca/uwaterloo/flix/language/phase/jvm/ClassFileMigration.scala
```diff
@@ -0,0 +1,29 @@
...
+
+    def aconst_null(): MethodVisitor =
```
magnus-madsen: Should it not return CodeBuilder?

magnus-madsen:
Could  you point out exactly where the ClassFile API is mirrored?

main/src/ca/uwaterloo/flix/language/phase/jvm/ClassFileMigration.scala
JonathanStarup: yeah that's probably a better name

main/src/ca/uwaterloo/flix/language/phase/jvm/ClassFileMigration.scala
JonathanStarup: Hmm I was debating it. I think the zero-allocation version of implicit class needs this because CodeBuilder/ClassFileMethodVisitor essentially doesn't exist for more than one call.

Returning MethodVistor means that the CodeBuilder/ClassFileMethodVisitor is only used instantaniously.

On the use-site the two are identical, except returning MethodVisitor is a bit more flexible

JonathanStarup:
> Could you point out exactly where the ClassFile API is mirrored?

In the code or here? `ClassFileMethodVisitor.aconst_null()` corresponds to `CodeBuilder.aconst_null()`

main/src/ca/uwaterloo/flix/language/phase/jvm/ClassFileMigration.scala
magnus-madsen: I thought the point was to match the CodeBuilder API?

magnus-madsen:
> > Could you point out exactly where the ClassFile API is mirrored?
> 
> In the code or here? `ClassFileMethodVisitor.aconst_null()` corresponds to `CodeBuilder.aconst_null()`

OK; maybe it should just be called code builder then?

main/src/ca/uwaterloo/flix/language/phase/jvm/ClassFileMigration.scala
JonathanStarup: Everything just returns MethodVisitor instead of CodeBuilder/ClassFileMethodVisitor. This is not really a meaningful difference, just an `implicit class` thing

The content/arguments of the calls are the interesting part, the return is just the pipe/builder pattern

magnus-madsen: My high level question is: Swapping to CodeBuilder will now still require lots of changes right? 

Why not avoid that? Is there any downside?

JonathanStarup: You can swap it without changing any of the code (given that it uses these methods) there is no downside to this choice

magnus-madsen: Just so we are on the same page:

It would be nice if one could write - say - the code for ArrayLoad and then just swap in the CodeBuilder.

But note that CodeBuilder from Java is a builder-- You have to chain calls on it.

Does that make sense?

magnus-madsen: This is their suggested style

```
    codeBuilder.iload(1)
        .ifeq(label1)
        .aload(0)
        .iload(2)
        .invokevirtual(ClassDesc.of("Foo"), "foo", MethodTypeDesc.of(CD_void, CD_int))
        .goto_(label2)
        .labelBinding(label1)
        .aload(0)
        .iload(2)
        .invokevirtual(ClassDesc.of("Foo"), "bar", MethodTypeDesc.of(CD_void, CD_int))
        .labelBinding(label2);
        .return_();
```

JonathanStarup: Yes, and you can do the same here. `mv.aconst_null().aconst_null()` still works, the `MethodVisitor` is just implicitly wrapped at every call. Re-using the same `CodeBuilder` class for multiple calls might make Scala keep it around.

magnus-madsen: But when the switch to CodeBuilder happens the code would have to be rewritten to use dot-chaining right?

JonathanStarup: As far as I can its also fine to write it `v.aconst_null(); v.aconst_null()` since the CodeBuilder is mutable.

Either way it doesn't influence this design, only use-sites

magnus-madsen: Are you 100% sure-- because they always show the other style? 

But in any case, why not use that style?

JonathanStarup: it is in that style - I don't understand. Whether it says MethodVisitor or not, you can still write in in pipeline style. 

Returning MethodVisitor reduces allocations and allows it to be more flexible while converting code. There is no drawbacks.

magnus-madsen: My point is that the code should use the Code builder in idiomatic style such that it is accessible to everyone. What's the point of writing in a style that does not match the examples.

JonathanStarup: To adhere to the implicit class pattern. And you can write exactly like the examples.

JonathanStarup: I can change it if you insist, it doesn't matter that much, but i think this is cleaner.

magnus-madsen: I would insist

main/src/ca/uwaterloo/flix/language/phase/jvm/ClassFileMigration.scala
```diff
@@ -0,0 +1,34 @@
...
+  implicit class CodeBuilder(mv: MethodVisitor) {
+
+    def aconst_null(): CodeBuilder =
+      builder(_.visitInsn(Opcodes.ACONST_NULL))
+
+    // Private functions.
+
+    private def builder(f: MethodVisitor => Unit): CodeBuilder = {
+      f(mv)
+      CodeBuilder(mv)
```
JonathanStarup: You can see here how it breaks the idea of an implicit class, since `this` now outlives the call.

The idea of the implicit class is to treat `MethodVisitor` as `CodeBuilder` via enrichment, not to actually replace the object.

magnus-madsen:
Probably easier to discuss in person.

magnus-madsen:
Instead of CodeBuilder being an implicit class, could it not be a proper argument?

JonathanStarup:
i think itll just bring more pain while migrating, but in the end, sure

magnus-madsen:
I would like to revisit this in the future.

Issue #10798, opened 2025-06-08
Add `Ping` effects


magnus-madsen:
Add new effects with default handlers:

```
eff Ping {
    def isReachable(host: String, timeout: Int32): Bool
}

eff PingWithResult {
    def isReachable(host: String, timeout: Int32): Result[IoError, Bool]
}
```

Based on https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/net/InetAddress.html#isReachable(int)

magnus-madsen:
Possibly we should use `IpAddr` from #10799

CadeMichael:
Just to make sure I'm understanding the `Ping` functionality vs `PingWithResult`

`Ping` will have similar functionality to this java (using the Inet library)

```java
InetAddressurl = InetAddress.getByName(ipAddress);

if (url.isReachable(5000))
      true;
    else
      false;
```

and `PingWithResult` will have an enum defining all possible errors and return them accordingly

magnus-madsen:
@CadeMichael Before we close this, I think one question is if we should introduce: IpAddr (see DNS ticket) and require that as an argument to Ping and PingWithResult.

Maybe this will become more clear as you implement the DNS effect.

magnus-madsen:
I will leave this ticket open until we make that decision.

magnus-madsen:
I think the proposal was to change the argument from host: String to host: IpAddr. What do you think?

PR #10818, opened 2025-06-09
feat: add `Ping` effect

CadeMichael:
addresses #10798.

next push will contain PingWithResult.

### DIFF
--- a/main/src/library/IoError.flix
+++ b/main/src/library/IoError.flix
@@ -39,7 +39,7 @@ mod IoError {
         case PermissionDenied,
         case UnexpectedEof,
         case Unsupported,
-        case UknownHost,
+        case UnknownHost,
         case Other
     }
 
--- a/main/src/library/IoError.flix
+++ b/main/src/library/IoError.flix
@@ -39,6 +39,7 @@ mod IoError {
         case PermissionDenied,
         case UnexpectedEof,
         case Unsupported,
+        case UknownHost,
         case Other
     }
 
--- a/main/src/library/Ping.flix
+++ b/main/src/library/Ping.flix
@@ -15,7 +15,7 @@
  */
 
 ///
-/// The effect used to connect to an address
+/// The effect used to determine if an address is reachable.
 ///
 eff Ping {
     def isReachable(host: String, timeout: Int32): Bool
--- a/main/src/library/Ping.flix
+++ b/main/src/library/Ping.flix
@@ -1,0 +1,50 @@
+/*
+ *  Copyright 2025 Cade Lueker
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+///
+/// The effect used to connect to an address
+///
+eff Ping {
+    def isReachable(host: String, timeout: Int32): Bool
+}
+
+mod Ping {
+    use IoError.ErrorKind
+    use IoError.IoError
+
+    import java.net.InetAddress
+    import java.io.IOException
+    import java.net.UnknownHostException
+
+    pub def handle(f: a -> b \ ef): a -> Result[IoError, b] \ (ef - Ping) + {Net, IO} = x ->
+        run {
+            Ok(f(x))
+        } with handler Ping {
+            def isReachable(host, timeout, k) = {
+                try {
+                    let addr: InetAddress = InetAddress.getByName(host);
+                    if (addr.isReachable(timeout)) {
+                        k(true)
+                    } else {
+                        k(false)
+                    }
+                } catch {
+                    case ex: IOException => Err(IoError(ErrorKind.Other, ex.getMessage()))
+                    case ex: UnknownHostException => Err(IoError(ErrorKind.UnknownHost, ex.getMessage()))
+                }
+            }
+        }
+}
--- a/main/src/library/Ping.flix
+++ b/main/src/library/Ping.flix
@@ -39,17 +39,10 @@ mod Ping {
         run {
             Ok(f(x))
         } with handler Ping {
-            def isReachable(host, timeout, k) = {
+            def isReachable(host, timeout, u, k) = {
                 try {
-                    let timeoutMs = match u {
-                           case TimeUnit.Days => timeout * 86400000,
-                           case TimeUnit.Hours => timeout * 3600000,
-                           case TimeUnit.Microseconds => timeout / 1000,
-                           case TimeUnit.Milliseconds => timeout,
-                           case TimeUnit.Minutes=> timeout * 6000,
-                           case TimeUnit.Nanoseconds=> timeout / 1000000,
-                           case TimeUnit.Seconds => timeout * 1000
-                    };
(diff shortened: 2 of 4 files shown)


CadeMichael:
One thing to note is that on non windows machines InetAddress will fail without sudo. I tried a java version that also returns false for googles dns "8.8.8.8"

```java
import java.net.InetAddress;

public class PingTest {
    public static void main(String[] args) {
        try {
            String host = "8.8.8.8";
            int timeout = 5000;
            InetAddress inet = InetAddress.getByName(host);

            boolean reachable = inet.isReachable(timeout);
            System.out.println(reachable);
        } catch (Exception e) {
            System.out.println(e.getMessage());
        }
    }
}
```

i'm not sure the best approach as something like a process builder would need specific commands based on the OS.

magnus-madsen:
> next push will contain PingWithResult.

Please do that in separate PR.

main/src/library/IoError.flix
```diff
@@ -39,6 +39,7 @@ mod IoError {
         case PermissionDenied,
         case UnexpectedEof,
         case Unsupported,
+        case UknownHost,
```
magnus-madsen: Spelling :D

main/src/library/Ping.flix
```diff
@@ -0,0 +1,49 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+///
+/// The effect used to connect to an address
```
magnus-madsen: -> The effect used to determine if an address is reachable.

main/src/library/Ping.flix
```diff
@@ -0,0 +1,49 @@
...
+                } catch {
+                    case ex: IOException => Err(IoError(ErrorKind.Other, ex.getMessage()))
+                    case ex: UnknownHostException => Err(IoError(ErrorKind.UnknownHost, ex.getMessage()))
+                }
+            }
+        }
+
+    pub def runWithIO(f: Unit -> a \ ef): Result[IoError, a] \ ef - Ping + {Net, IO} =
+        handle(f)()
+}
```
magnus-madsen, line 63: missing line break at the end

main/src/library/Ping.flix
```diff
@@ -0,0 +1,49 @@
...
+
+mod Ping {
+    use IoError.ErrorKind
+    use IoError.IoError
+
+    import java.net.InetAddress
+    import java.io.IOException
+    import java.net.UnknownHostException
+
+    pub def handle(f: a -> b \ ef): a -> Result[IoError, b] \ (ef - Ping) + {Net, IO} = x ->
```
magnus-madsen, line 38: Add docs (see the style used for other effects)

main/src/library/Ping.flix
```diff
@@ -0,0 +1,49 @@
...
+                    let addr = InetAddress.getByName(host);
+                    k(addr.isReachable(timeout))
+                } catch {
+                    case ex: IOException => Err(IoError(ErrorKind.Other, ex.getMessage()))
+                    case ex: UnknownHostException => Err(IoError(ErrorKind.UnknownHost, ex.getMessage()))
+                }
+            }
+        }
+
+    pub def runWithIO(f: Unit -> a \ ef): Result[IoError, a] \ ef - Ping + {Net, IO} =
```
magnus-madsen: Add docs (see the style used for other effects)

main/src/library/Ping.flix
```diff
@@ -0,0 +1,49 @@
...
+        run {
+            Ok(f(x))
+        } with handler Ping {
+            def isReachable(host, timeout, k) = {
+                try {
+                    let addr = InetAddress.getByName(host);
+                    k(addr.isReachable(timeout))
+                } catch {
+                    case ex: IOException => Err(IoError(ErrorKind.Other, ex.getMessage()))
+                    case ex: UnknownHostException => Err(IoError(ErrorKind.UnknownHost, ex.getMessage()))
```
magnus-madsen, line 49: UnknownHostException  extends IOException so it has to be above that.

main/src/library/Ping.flix
```diff
@@ -0,0 +1,49 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+///
+/// The effect used to connect to an address
+///
+eff Ping {
+    def isReachable(host: String, timeout: Int32): Bool
```
magnus-madsen: I think we should add `u: TimeUnit` as the last argument to the signature.

magnus-madsen:
Looks good! If you polish a bit more then we can merge and continue with PingWithResult.

main/src/library/Ping.flix
```diff
@@ -0,0 +1,71 @@
...
+    ///
+    /// In other words, re-interprets the `Ping` effect using the `Net` and `IO` effects.
+    ///
+    pub def handle(f: a -> b \ ef): a -> Result[IoError, b] \ (ef - Ping) + {Net, IO} = x ->
+        run {
+            Ok(f(x))
+        } with handler Ping {
+            def isReachable(host, timeout, k) = {
+                try {
+                    let timeoutMs = match u {
```
magnus-madsen: Where is u bound?

main/src/library/Ping.flix
```diff
@@ -0,0 +1,71 @@
...
+                           case TimeUnit.Days => timeout * 86400000,
+                           case TimeUnit.Hours => timeout * 3600000,
+                           case TimeUnit.Microseconds => timeout / 1000,
+                           case TimeUnit.Milliseconds => timeout,
+                           case TimeUnit.Minutes=> timeout * 6000,
+                           case TimeUnit.Nanoseconds=> timeout / 1000000,
+                           case TimeUnit.Seconds => timeout * 1000
```
magnus-madsen: We will probably want to move the conversion somewhere else so we can reuse it, maybe into TimeUnit module?

magnus-madsen: Also, I would avoid magic constants but instead have the multiplications there.

Oh and use _ for thousand separator, e.g. 1_000

magnus-madsen (commented):
Very good. I think this is almost ready. Just a few TODOs and remember to the file to Flix.scala.

magnus-madsen:
Once that is done, I propose that we merge, and then we can continue with PingWithResult in separate PR.

magnus-madsen:
Remember to add to Flix.scala

main/src/library/TimeUnit.flix
```diff
@@ -26,3 +26,22 @@ enum TimeUnit with Eq {
...
     case Seconds
 }
+
+mod TimeUnit {
+
+    ///
+    /// Returns the number of miliseconds for timeunit `u`
+    ///
+    pub def toMilliseconds(u: TimeUnit): Int32 =
+        let ms = match u {
```
magnus-madsen: No need for the extra local variable

just do

def toMilliseconds(u: TimeUnit): Int32 = match u {
 ...

main/src/library/TimeUnit.flix
```diff
@@ -26,3 +26,22 @@ enum TimeUnit with Eq {
...
+           case TimeUnit.Days => 24 * 60 * 60 * 1_000,
```
magnus-madsen: We also usually align the arrows =>

```
     => 24 * 60 * 60 * 1_000
     => 60 * 60 * 1_000
```

magnus-madsen:
@CadeMichael Once my last comments are fixed, lets merge this, and then we can iterate.

magnus-madsen:
@CadeMichael Remember to add to the new files to Flix.scala

magnus-madsen:
Merged! Now we can move onto PingWithResult

PR #10822, opened 2025-06-09
refactor: remove BOT.Thread

JonathanStarup:


main/src/ca/uwaterloo/flix/language/phase/jvm/ClassMaker.scala
```diff
@@ -311,4 +311,25 @@ object ClassMaker {
       InstanceMethod(JvmName.String, "concat", mkDescriptor(BackendType.String)(BackendType.String))
   }
 
+  object Thread {
+
+    def StartMethod: InstanceMethod =
```
JonathanStarup, line 316: Once these ClassMaker things are seperated from BOT, they can also become `val`, which i think should give some speedup

magnus-madsen (approved):

PR #10826, opened 2025-06-09
feat: test that `KeywordCompletions` are ranked lowest

magnus-madsen:


magnus-madsen:
As usual- this is a small step forwards that does not get us there entirely.

PR #10828, opened 2025-06-09
refactor: remove JvmType

JonathanStarup:


magnus-madsen:
What about JvmName- what is the plan for that?

magnus-madsen (approved):

JonathanStarup:
> AsmOps and JvmOps are pretty small- maybe then can be folded together or removed all together?

Yea idea is that they will shrink to nothing

> What about JvmName- what is the plan for that?

I think JvmName is fine - its a very simple abstraction to allows some functions to avoid messy string manipulation

PR #10831, opened 2025-06-09
feat: improve `JvmName.ofClass`

JonathanStarup:
Now it crashes for classes that would otherwise just create invalid `JvmName` objects.

As for the string manipulation - `Class` doesn't have any direct way to access it, and asm also uses string manipulation. This seems to be state of the art


magnus-madsen:
Ultimately of course we should not use class-- because that means we will actually have loaded the class.

magnus-madsen (approved):

PR #10836, opened 2025-06-09
refactor: update GenAnonClasses

JonathanStarup:


main/src/ca/uwaterloo/flix/language/phase/jvm/GenAnonymousClasses.scala
```diff
@@ -19,147 +19,92 @@ package ca.uwaterloo.flix.language.phase.jvm
...
 object GenAnonymousClasses {
 
-  /**
-    * Returns the set of anonymous classes for the given set of objects
-    */
-  def gen(objs: List[AnonClass])(implicit root: Root, flix: Flix): Map[JvmName, JvmClass] = {
-    //
-    // Generate an anonymous class for each object and collect the results in a map.
-    //
-    ParOps.parAgg(objs, Map.empty[JvmName, JvmClass])({
```
JonathanStarup, line 41: I've removed ParOps here because of uniformity, I also don't think it matters much since there are so few

main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
```diff
@@ -1765,6 +1765,7 @@ object BackendObjType {
       crashIfSuspension(errorHint, loc)
       CHECKCAST(Value.jvmName) // Cannot fail
       GETFIELD(Value.fieldFromType(tpe))
+      castIfNotPrim(tpe)
```
JonathanStarup, line 1704: A missing cast in `unwindSuspensionFreeThunkToType`. It was only used in erased ways so it was not observable

main/src/ca/uwaterloo/flix/language/phase/jvm/GenAnonymousClasses.scala
JonathanStarup: We should instead generate all classes in parallel

JonathanStarup:
!!!rerun

magnus-madsen (approved):

PR #10838, opened 2025-06-09
feat: parallel backend

JonathanStarup:
We can generate all classes in parallel, except function classes, effect classes, and anon classes (because of their coding style, not anything deep)

This list is usually between 90 and 300 classes during the tests so it'll give more once those three generators are rewritten.


JonathanStarup:
PR `715,8 ms (20,3%)`
Master `804,6 ms (21,3%)`

Pretty noisy, just my PC, but could give a little performance

magnus-madsen:
I think this significantly increases the brittleness of the code. Is there really not a better way to do it?

More importantly, does it not enable any speeds-ups? If not, I don't think we should do it at increase cost to code complexity.

JonathanStarup:
> I think this significantly increases the brittleness of the code

Why? To me its as bad as it was before

> Is there really not a better way to do it?

I hope so, but its tough because its basically just a long heterogeneous list of unrelated calls.

> More importantly, does it not enable any speeds-ups?

I think most of the work is in GenExpression, and those classes are not included yet. I'd add them once Jakobs PR is in

PR #10845, opened 2025-06-10
feat: transformer from Flix PredSym to Fixpoint PredSym

CasperDN:
This phase simply transforms PredSym as Flix understands them into PredSyms with unique integer parts as our solver uses this.
To keep ensure the old Datalog engine and ours can work on `Datalog.flix` at the same time we have modified the old engine's `Solver.flix`.

### DIFF
--- a/main/src/library/Fixpoint3/Phase/RenamePredSyms.flix
+++ b/main/src/library/Fixpoint3/Phase/RenamePredSyms.flix
@@ -15,14 +15,13 @@
  * limitations under the License.
  */
 
-//
-/// For the rest of the compiler we expect a PredSym, `PredSym(name, id)`, to have unique `id`.
-/// This phase passes over the Datalog-Ast and replaces the id and after everything is done
+///
+/// Purpose of phase: For `PredSym(name, id)` we want `id` to be unique, for the whole program.
+/// This phase passes over the Datalog-Ast and replaces the `id` and after everything is done
 /// recreates the old `PredSym`s for the resulting model.
-/// 
-/// It also contains method for recreating the original PredSyms such that the model given to Flix
-/// is correct.
-//
+///
+/// This is necessary as Flix uses the id-part for generating local predicates, not as unique identifiers.
+///
 mod Fixpoint3.Phase.RenamePredSyms {
     use Fixpoint.Ast.Datalog.{Datalog, Constraint, BodyPredicate, BodyTerm, HeadTerm, Polarity, VarSym}
     use Fixpoint.Ast.Datalog.Datalog.{Datalog, Model, Join}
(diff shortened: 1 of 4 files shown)


main/src/library/Fixpoint3/Phase/RenamePredSyms.flix
```diff
@@ -0,0 +1,122 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+//
+/// For the rest of the compiler we expect a PredSym, `PredSym(name, id)`, to have unique `id`.
+/// This phase passes over the Datalog-Ast and replaces the id and after everything is done
```
JonathanStarup: > replaces the id

from what? to what?

main/src/library/Fixpoint3/Phase/RenamePredSyms.flix
```diff
@@ -0,0 +1,122 @@
...
+    }
+    
+    ///
+    /// Maps `PredSym`s of `body` according to `transformer`.
+    ///
+    def remapBody(transformer: PredSym -> PredSym \ r, body: BodyPredicate): BodyPredicate \ r = match body {
+        case BodyAtom(pred, den, pol, fixed, terms) => BodyAtom(transformer(pred), den, pol, fixed, terms)
+        case _ => body
+    }
+}
```
JonathanStarup: newline issue

main/src/library/Fixpoint3/Phase/RenamePredSyms.flix
```diff
@@ -0,0 +1,122 @@
...
+            Model
+        case _ => bug!("Bad argument of type 'Join' in PredSym renaming")
+    }
+    
+    ///
+    /// Maps `pred` back to its original `PredSym` according to `reversing`.
+    ///
+    def mapBack(reversing: FixInformation, pred: PredSym): PredSym = 
+        let PredSym(_, id) = pred;
+        getOrCrash(Map.get(getOrCrash(Int64.tryToInt32(id)), reversing))
```
JonathanStarup, line 100: Why are both IDs not the same integer type?

main/src/library/Fixpoint3/Phase/RenamePredSyms.flix
JonathanStarup: So the goal of this module is to make `PredSyms` have unique integers.

What are they in the input? are they not unique already? Can you expand on what the meaning of the input PredSym, since they already have an integer attached

main/src/library/Fixpoint3/Phase/RenamePredSyms.flix
```diff
@@ -0,0 +1,122 @@
...
+        case Join(d1, d2) =>
+            Join(renamePredSymsInternal(d1, mapping), renamePredSymsInternal(d2, mapping))
+        case _ => bug!("PredSym error")
+    }
+
+    ///
+    /// Applies `fixing` to the `d` to restore the `PredSym`s to the original values.
+    ///
+    @Internal
+    pub def renamePredSymToFlixStyle(d: Datalog, fixing: FixInformation): Datalog = match d {
```
JonathanStarup: having FlixStyle in the name made me a bit confused - "what is Flix style?".

This function is just applying the fixing mapping to the predSyms of `d` right? So this function should just be called `renamePredSyms` and then its usage can be commented in the doc comment if needed

main/src/library/Fixpoint3/Phase/RenamePredSyms.flix
CasperDN: Flix appears to assign every PredSym 0, unless local predicates are used. See https://doc.flix.dev/local-predicates.html
If local predicates are used then the first appear to get 1, the second 2 and so on.

main/src/library/Fixpoint3/Phase/RenamePredSyms.flix
CasperDN: `UniqueInts` only generate `Int32`. `PredSym` uses `Int64`. This is only used for 'our' `PredSym`s, which are named 0 to n and should be well within the `Int32` space.

JonathanStarup: I see. I think you should explain that in a comment above the `getOrCrash`

main/src/library/Fixpoint3/Phase/RenamePredSyms.flix
JonathanStarup: Right, you could add a paragraph that in the input you interpret `(name, id)` to identity the predicate while in the output, because of the conversions `id` is enough to identify the predicate.

Since you reeuse the same object `PredSym` in input and output I think it's important to be precise in how the interpretation changes.

JonathanStarup (approved):

magnus-madsen:
Ready for me to look at?

CasperDN:
@magnus-madsen yes

magnus-madsen (approved):

PR #10846, opened 2025-06-10
feat: add parallel foreach to BPlusTree

CasperDN:
We have added a parallel foreach method to the BPlusTree. It is meant for internal usage by the Datalog engine, though e.g. merge could also use it.

### DIFF
--- a/main/src/library/BPlusTree.flix
+++ b/main/src/library/BPlusTree.flix
@@ -360,6 +360,7 @@ mod BPlusTree {
     @Internal
     @Parallel
     pub def parForEach(f: k -> v -> Unit \ r, t: BPlusTree[k, v, r]): Unit \ r with Order[k] = region rc {
+        // `threadNum` is the number of spawned threads.
         let threadNum = if(Node.computeHeight(t->root) >= 2) {
             threads()
         } else {
@@ -370,13 +371,13 @@ mod BPlusTree {
         } else {
             let minLeaf = Node.leftMostChild(t->root);
             unchecked_cast(Vector.range(0, threadNum) |>
-            Vector.forEach(i -> {
+            (Vector.forEach(i -> {
                 let _: Unit = spawn unchecked_cast(
                     // Flix does not allow region effect in spawns. Remove the effect. 
                     (Node.traverseRightUnconditionalInc(f, i, threadNum, minLeaf): _ \ r) as _ \ IO
                 ) @ rc;
                 ()
-            }) as _ \ r)
+            }): _ \ {}) as _ \ r)
         }
     }
 
@@ -784,14 +785,16 @@ mod BPlusTree {
         ///
         @Internal
         pub def computeHeight(node: Node[k, v, r]): Int32 \ r with Order[k] =
+            // Every leaf is at the same height, so just use the left most path.
             def loop(count, n) = {
                 match n->isLeaf {
                     case true => count
-                    case false => if(n->size == 0) {
-                        count
-                    }  else {
-                        loop(count + 1, Array.get(0, n->children))
-                    }
+                    case false => 
+                        if(n->size == 0) {
+                            count
+                        }  else {
+                            loop(count + 1, Array.get(0, n->children))
+                        }
                 }
             };
             loop(1, node)
--- a/main/src/library/BPlusTree.flix
+++ b/main/src/library/BPlusTree.flix
@@ -60,6 +60,7 @@ pub struct BPlusTree[k, v, r] {
 }
 
 mod BPlusTree {
+    import java.lang.Runtime
     import java.util.concurrent.locks.{StampedLock => JStampedLock}
 
     @Internal
@@ -350,6 +351,36 @@ mod BPlusTree {
         Node.getWithDefault(k, stamp, d, leaf, t)
 
     ///
+    /// Applies `f` to all mappings `k => v` in `t` in parallel.
+    ///
+    /// This method unsafely removes the `r` effect when spawning threads.
+    ///
+    /// Not thread-safe.
+    ///
+    @Internal
+    @Parallel
+    pub def parForEach(f: k -> v -> Unit \ r, t: BPlusTree[k, v, r]): Unit \ r with Order[k] = region rc {
+        let threadNum = if(Node.computeHeight(t->root) >= 2) {
+            threads()
+        } else {
+            Int32.max(threads(), Node.size(t->root))
+        };
+        if (threadNum <= 1) {
+            forEach(f, t)
+        } else {
+            let minLeaf = Node.leftMostChild(t->root);
+            unchecked_cast(Vector.range(0, threadNum) |>
+            Vector.forEach(i -> {
+                let _: Unit = spawn unchecked_cast(
+                    // Flix does not allow region effect in spawns. Remove the effect. 
+                    (Node.traverseRightUnconditionalInc(f, i, threadNum, minLeaf): _ \ r) as _ \ IO
+                ) @ rc;
+                ()
+            }) as _ \ r)
+        }
+    }
+
+    ///
     /// Updates the tree `t` with the mapping `k => v`. Replaces any
     /// existing mapping.
     ///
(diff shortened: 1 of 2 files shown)


magnus-madsen:
@JonathanStarup I will let you review first

main/src/library/BPlusTree.flix
```diff
@@ -349,6 +350,36 @@ mod BPlusTree {
         let (leaf, stamp) = BPlusTree.findLeaf(k, t);
         Node.getWithDefault(k, stamp, d, leaf, t)
 
+    ///
+    /// Applies `f` to all mappings `k => v` in `t` in parallel.
```
JonathanStarup: something about `f` should modify `t`?

main/src/library/BPlusTree.flix
```diff
@@ -349,6 +350,36 @@ mod BPlusTree {
...
+    ///
+    /// This method unsafely removes the `r` effect when spawning threads.
+    ///
+    /// Not thread-safe.
+    ///
+    @Internal
+    @Parallel
+    pub def parForEach(f: k -> v -> Unit \ r, t: BPlusTree[k, v, r]): Unit \ r with Order[k] = region rc {
+        let threadNum = if(Node.computeHeight(t->root) >= 2) {
```
JonathanStarup, line 365: I'm still a bit iffy on the memory model aspect of this:

So you can only use this in single threaded cases where you have had a synchronizing operation on the multithreaded work?

main/src/library/BPlusTree.flix
```diff
@@ -732,6 +777,25 @@ mod BPlusTree {
...
+        /// Not thread-safe.
+        ///
+        @Internal
+        pub def computeHeight(node: Node[k, v, r]): Int32 \ r with Order[k] =
+            def loop(count, n) = {
+                match n->isLeaf {
+                    case true => count
+                    case false => if(n->size == 0) {
+                        count
+                    }  else {
```
JonathanStarup: the formatting here is a bit non-standard

main/src/library/BPlusTree.flix
CasperDN: The idea is to have a function which does a forEach in parallel. The programmer should ensure that the function `f` is safe to execute in parallel.

The current implementation of `merge` could for example use this function instead of `forEach`.

main/src/library/BPlusTree.flix
```diff
@@ -732,6 +777,25 @@ mod BPlusTree {
...
+                        loop(count + 1, Array.get(0, n->children))
```
JonathanStarup: What is the reason that you only have to look at the left-most child?

main/src/library/BPlusTree.flix
JonathanStarup, line 365: Could you add a short comment about the `threadNum` number? im not sure what this ITE ensures

Also, for efficiency, it seems potentially wasteful to compute the full height only to check that it is `>= 2`. I guess they're not tall in practice.

JonathanStarup: Oh I see that computeHeight is `O(Height)`

main/src/library/BPlusTree.flix
CasperDN: It is not thread safe in the sense that you should not modify `t` when the function executes.

For our sake `f` could do anything, with any effect it wants, as long as the user guarantees that it is thread-safe. We do not know how to express this with effects, and we only use it for something modifying something from the same region, so we simplified the effect to this.

main/src/library/BPlusTree.flix
```diff
@@ -1367,5 +1437,24 @@ mod BPlusTree {
                 case Some(next) => traverseRightUnconditional(f, 0, next)
                 case None => ()
             }
+        ///
+        /// Apply `f` to every `stepSize` mapping in `node` and continues in right of `node`.
+        ///
+        /// Not thread-safe.
+        ///
+        pub def traverseRightUnconditionalInc(
```
JonathanStarup, line 1449: what does Unconditional refer to here? is this not just `TraverseRightStep`

main/src/library/BPlusTree.flix
CasperDN: With 1 billion elements they are at most of height 6.

main/src/library/BPlusTree.flix
```diff
@@ -349,6 +350,36 @@ mod BPlusTree {
...
+        } else {
+            let minLeaf = Node.leftMostChild(t->root);
+            unchecked_cast(Vector.range(0, threadNum) |>
+            Vector.forEach(i -> {
+                let _: Unit = spawn unchecked_cast(
+                    // Flix does not allow region effect in spawns. Remove the effect. 
+                    (Node.traverseRightUnconditionalInc(f, i, threadNum, minLeaf): _ \ r) as _ \ IO
+                ) @ rc;
+                ()
+            }) as _ \ r)
```
JonathanStarup: It would be nice if you allow follow the annotate -> cast pattern here, so you can see what is removed

main/src/library/BPlusTree.flix
CasperDN: Yes, but `traverseRight` traverses until it finds a `max` value, so `unconditional` was used to mean, without an upper bound. In hindsight renaming `traverseRight` to `traverseRightUntil` or something similar might have been a better choice.

main/src/library/BPlusTree.flix
CasperDN: Every leaf is at the same level.

main/src/library/BPlusTree.flix
JonathanStarup: yeah but this function access the whole tree without aquiring locks, so java memory model says that these values might be out of sync with other thread changes right?

main/src/library/BPlusTree.flix
JonathanStarup: Right, maybe a comment for forgetful people like me?

main/src/library/BPlusTree.flix
JonathanStarup: Depdending on the size of the nodes I suppose, but yeah that's fine

> Could you add a short comment about the threadNum number? im not sure what this ITE ensures

But back to this, could you add a comment about what this ITE does?

main/src/library/BPlusTree.flix
JonathanStarup: Oh right so it is unconditional compared to `traverseRight`. hmm I see. I would still drop the unconditional part of the name but I understand the reason. More of a subjective thing so ill let you and magnus decide.

main/src/library/BPlusTree.flix
JonathanStarup: I think that deserves to be explicit even though it is a bit obvious yes.

main/src/library/BPlusTree.flix
CasperDN: Yes, so no other thread should modify this tree while this method executes.

If we merge t1 into t2 then we can look through t1 in parallel and insert every element from t1 into t2.

magnus-madsen:
Ping me when you have resolved your differences :)

main/src/library/BPlusTree.flix
CasperDN: We could change `traverseRight` to `traverseRightUntil` and rename this to `traverseRightStep`. At the same time the current `traverseRightUnconditional` could simply be `traverseRightStep(stepsize=1)`. This is more of a refactoring though.

main/src/library/BPlusTree.flix
CasperDN: I believe one major problem is that we in principle would prefer this function to take any thread-safe `f`, but we do not know how to type this. We could extend the type to allow all effect that Flix allows in `spawn` statement.
We limited the effect to `r`, with the purpose that it should only affect thread-safe stuff in the region, namely other B+ trees, as that was our usecase.

JonathanStarup: I think its accepted that the `f` function is the caller's reponsibility in regards to thread-safety, but i was just pondering what thread-safe really means especially in regards to the memory model. But after mulling it over I think this is fine 😄 sorry about the detour

JonathanStarup:
@magnus-madsen I'm happy :)

main/src/library/BPlusTree.flix
CasperDN: No problem. The memory model is quite tricky.

Issue #10799, opened 2025-06-08
Add `Dns` effects


magnus-madsen:
Here is some skeleton code:

```
enum IpAddr { // in its own file
    case IPv4(Int32, Int32, Int32, Int32),
    case IPv6(...)
}

eff Dns {
    def lookup(host: String): IpAddr 
}

eff DnsWithError {
    def lookup(host: String): Result[IoError, IpAddr ]
}
```

CadeMichael:
should each effect be it's own file?

magnus-madsen:
> should each effect be it's own file?

Yes

magnus-madsen:
@CadeMichael We should have a look at https://doc.rust-lang.org/std/net/index.html and see if we want to follow their design a bit more closely.

magnus-madsen:
We will continue in new tickets.

PR #10850, opened 2025-06-10
feat: Dns effect for creating IpAddr

CadeMichael:
Addresses #10799.

### DIFF
--- a/main/src/library/Dns.flix
+++ b/main/src/library/Dns.flix
@@ -1,0 +1,69 @@
+/*
+ *  Copyright 2025 Cade Lueker
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+///
+/// an effect used to create an IpAddr
+///
+eff Dns {
+    def lookup(host: String): IpAddr
+}
+
+mod Dns {
+    use IoError.ErrorKind
+    use IoError.IoError
+
+    import java.io.IOException
+    import java.lang.IllegalArgumentException
+    import java.net.InetAddress
+
+    ///
+    /// Handles the `Dns` effect of the given function `f`.
+    ///
+    /// In other words, re-interprets the `Dns` effect using the `Net` and `IO` effects.
+    ///
+    pub def handle(f: a -> b \ ef): a -> Result[IoError, b] \ (ef - Dns) + {Net, IO} = x ->
+        run {
+            Ok(f(x))
+        } with handler Dns {
+            def lookup(host, k) = {
+                try {
+                    let address = InetAddress.getByName(host);
+                    let addrBytes = Array.toList(address.getAddress());
+                    let ipAddr = match addrBytes {
+                        case b1 :: b2 :: b3 :: b4 :: Nil => {
+                            IpAddr.IPv4(b1, b2, b3, b4)
+                        }
+                        case b1 :: b2 :: b3 :: b4 :: b5 :: b6 :: b7 :: b8 :: Nil => {
+                            IpAddr.IPv6(b1, b2, b3, b4, b5, b6, b7, b8)
+                        }
+                        case _ => unreachable!()
+                    };
+                    k(ipAddr)
+                } catch {
+                    case ex: IllegalArgumentException => Err(IoError(ErrorKind.InvalidInput, ex.getMessage()))
+                    case ex: IOException              => Err(IoError(ErrorKind.Other, ex.getMessage()))
+                }
+            }
+        }
+
+    ///
+    /// Handles the `Dns` effect of the given function `f`.
+    ///
+    /// In other words, re-interprets the `Dns` effect using the `Net` and `IO` effects.
+    ///
+    pub def runWithIO(f: Unit -> a \ ef): Result[IoError, a] \ ef - Dns + {Net, IO} = handle(f)()
+
+}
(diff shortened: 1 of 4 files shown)


CadeMichael:
- I used int8 because that's what the Inet function returns
- I think Ipv6 is 16 int8's 

from some of the tests

```
pub def main(): Result[IoError, Unit] \ {Net, IO}=
    run {
        let host = "localhost";
        println("looking up ${host}");
        println(Dns.lookup(host))
    } with Dns.runWithIO
```

```
looking up "localhost"                                                            
IPv4(127, 0, 0, 1)
Ok(())

looking up "local"                                                              
Err(IoError(Other, local: nodename nor servname provided, or not known))

looking up "8.8.8.8"                                                            
IPv4(8, 8, 8, 8)
Ok(())
```

CadeMichael:
if it is 16 Int8's then I will add the changes from my tests which gave me

```
flix> :eval main()
looking up "2606:4700:4700::1111"                                               
38 :: 6 :: 71 :: 0 :: 71 :: 0 :: 0 :: 0 :: 0 :: 0 :: 0 :: 0 :: 0 :: 0 :: 17 :: 17 :: Nil
IPv6(38, 6, 71, 0, 71, 0, 0, 0, 0, 0, 0, 0, 0, 0, 17, 17)
Ok(())
```

main/src/library/Dns.flix
```diff
@@ -0,0 +1,69 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+///
+/// an effect used to create an IpAddr
```
magnus-madsen: I would say something like for resolving a domain name to an ip address

main/src/library/Dns.flix
```diff
@@ -0,0 +1,69 @@
...
+    ///
+    /// In other words, re-interprets the `Dns` effect using the `Net` and `IO` effects.
+    ///
+    pub def handle(f: a -> b \ ef): a -> Result[IoError, b] \ (ef - Dns) + {Net, IO} = x ->
+        run {
+            Ok(f(x))
+        } with handler Dns {
+            def lookup(host, k) = {
+                try {
+                    let address = InetAddress.getByName(host);
```
magnus-madsen, line 60: What about [UnknownHostException](https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/net/UnknownHostException.html)

main/src/library/Dns.flix
```diff
@@ -0,0 +1,69 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+///
+/// an effect used to create an IpAddr
+///
+eff Dns {
+    def lookup(host: String): IpAddr
```
magnus-madsen, line 21: I think we should have both

```
def lookup(host: String): IpAddr
and
def lookupAll(host: String): List[IpAddr]
```

main/src/library/Dns.flix
CadeMichael: I don't believe it throws that exception, because the host could be unknown but the Ip could be valid. So it is only concerned with a valid string being passed in.

main/src/library/Dns.flix
CadeMichael: would it need a `List[String]` as input as each string returns one Ip?

main/src/library/Dns.flix
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+    ///
+    /// Handles the `Dns` effect of the given function `f`.
+    ///
+    /// In other words, re-interprets the `Dns` effect using the `Net` and `IO` effects.
+    ///
+    pub def runWithIO(f: Unit -> a \ ef): Result[IoError, a] \ ef - Dns + {Net, IO} = handle(f)()
+
+}
```
magnus-madsen, line 95: I think files typically end with a new line. This goes for all the other files too.

main/src/library/Ping.flix
```diff
@@ -46,9 +46,9 @@ mod Ping {
                     let addr = InetAddress.getByName(host);
                     k(addr.isReachable(timeoutMs))
                 } catch {
-                    case ex: UnknownHostException => Err(IoError(ErrorKind.UnknownHost, ex.getMessage()))
-                    case ex: IllegalArgumentException  => Err(IoError(ErrorKind.InvalidInput, ex.getMessage()))
-                    case ex: IOException => Err(IoError(ErrorKind.Other, ex.getMessage()))
+                    case ex: UnknownHostException     => Err(IoError(ErrorKind.UnknownHost, ex.getMessage()))
+                    case ex: IllegalArgumentException => Err(IoError(ErrorKind.InvalidInput, ex.getMessage()))
+                    case ex: IOException              => Err(IoError(ErrorKind.Other, ex.getMessage()))
```
magnus-madsen, line 51: Normally this would be a separate PR-- but OK

main/src/library/Dns.flix
magnus-madsen: According to the doc it can or am I reading the docs wrong?

magnus-madsen: What happens if you lookup doesnotexist.com? (I should probably be clear that UHEx extends IOEx so it probably still works- I just thought it best to be explicit).

main/src/library/Dns.flix
magnus-madsen: Actually a single domain name can be mapped to multiple IPs

magnus-madsen: That is why Java has this: https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/net/InetAddress.html#getAllByName(java.lang.String)

magnus-madsen:
Minor adjustments and then we can merge this 🚀

main/src/library/Dns.flix
CadeMichael: got it! looking through docs more thoroughly now, when testing with dummy addresses it didn't throw but I was not triggering it properly. Will add.

main/src/library/Dns.flix
CadeMichael: got it ! I'll add asap

magnus-madsen (approved):

magnus-madsen:
Merged!

PR #10791, opened 2025-06-08
fix: add `IllegalReservedName`

wizard7377:
Made it so that inputs also have whether they are part of the standard library or not.
This allows for builtins to trigger an error in user files, in response to #2238

PR #10852, opened 2025-06-10
refactor: rename types in `BoxingType` to avoid use of reserved names

wizard7377:
Split from #10791


magnus-madsen (approved):

magnus-madsen:
Merged! Thanks

PR #10857, opened 2025-06-11
feat: remove local variable field erasure on def/closure objects

JonathanStarup:
The local variables are stored and read to local fields to enable effect resumption. These fields are not a part of any interface and are truly local, so it does't make any sense to erase them - pure downside (runtime casts and compiletime complexity)

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/jvm/GenFunAndClosureClasses.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/jvm/GenFunAndClosureClasses.scala
@@ -109,7 +109,7 @@ object GenFunAndClosureClasses {
     *   // locals variables
     *   public int l0;
     *   public char l1;
-    *   public Object l2;
+    *   public String l2;
     *   // function arguments
     *   public Object arg0;
     *   public int arg1
@@ -120,7 +120,7 @@ object GenFunAndClosureClasses {
     *     // fields are put into the local frame according to symbol data
     *     int ? = this.l0;
     *     char ? = this.l1;
-    *     Object ? = this.l2;
+    *     String ? = this.l2;
     *
     *     EnterLabel:
     *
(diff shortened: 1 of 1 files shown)


main/src/ca/uwaterloo/flix/language/phase/jvm/GenFunAndClosureClasses.scala
```diff
@@ -413,39 +413,41 @@ object GenFunAndClosureClasses {
...
+    m.visitFieldInsn(GETFIELD, className.toInternalName, name, fieldType.toDescriptor)
     // cast the value and store it
-    val bType = BackendType.toBackendType(tpe)
-    BytecodeInstructions.castIfNotPrim(bType)
-    BytecodeInstructions.xStore(bType, localIndex)
+    castTo match {
+      case Some(targetType) =>
+        BytecodeInstructions.castIfNotPrim(targetType)
+        BytecodeInstructions.xStore(targetType, localIndex)
+      case None =>
```
JonathanStarup, line 426: Here is the saved CHECKCAST

main/src/ca/uwaterloo/flix/language/phase/jvm/GenFunAndClosureClasses.scala
```diff
@@ -345,12 +346,12 @@ object GenFunAndClosureClasses {
     implicit val m: MethodVisitor = visitor.visitMethod(ACC_PUBLIC + ACC_FINAL, applyMethod.name, applyMethod.d.toDescriptor, null, null)
     val localOffset = 2 // [this: Obj, value: Obj, ...]
 
-    val lparams = defn.lparams.zipWithIndex.map { case (lp, i) => (s"l$i", lp.sym.getStackOffset(localOffset), lp.sym.isWild, lp.tpe) }
-    val cparams = defn.cparams.zipWithIndex.map { case (cp, i) => (s"clo$i", cp.sym.getStackOffset(localOffset), false, cp.tpe) }
-    val fparams = defn.fparams.zipWithIndex.map { case (fp, i) => (s"arg$i", fp.sym.getStackOffset(localOffset), false, fp.tpe) }
+    val lparams = defn.lparams.zipWithIndex.map { case (lp, i) => (s"l$i", lp.sym.getStackOffset(localOffset), lp.sym.isWild, BackendType.toBackendType(lp.tpe), None) }
```
JonathanStarup, line 349: Here is where lparams say that they are non-erased

JonathanStarup:
The code here is not great. But it is a tangled ball of yarn and trying to really untagle it will balloon the PR to be impractical. This PR will allow later code to be more succinct because the generating code now only needs to know the actual type of the lparams, not both the erased and un-erased

magnus-madsen (approved):

PR #10860, opened 2025-06-11
refactor: remove BOT FlixError

JonathanStarup:
@magnus-madsen I don't know if you agree with that approach

this `FlixError` object basically corresponds to the old `GenFlixError` class. These objects have no reason to be a part of BOT so i will extract them to be their own ClassMaker thing, makes both simpler in the process. 

This means that most of `BOT` gets extracted into `ClassMaker` which is maybe fine but i could also add `ClassConstants` to cut in in two, or have `FlixError` be its own file.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/jvm/BackendObjType.scala
@@ -53,7 +53,6 @@ sealed trait BackendObjType {
     case BackendObjType.Record => JvmName(RootPackage, mkClassName("Record"))
     case BackendObjType.ReifiedSourceLocation => JvmName(DevFlixRuntime, mkClassName("ReifiedSourceLocation"))
     case BackendObjType.Global => JvmName(DevFlixRuntime, "Global") // "Global" is fixed in source code, so should not be mangled and $ suffixed
-    case BackendObjType.FlixError => JvmName(DevFlixRuntime, mkClassName("FlixError"))
     case BackendObjType.HoleError => JvmName(DevFlixRuntime, mkClassName("HoleError"))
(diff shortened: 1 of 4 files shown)


JonathanStarup:
Following this idea to its conclusion, BackendType wil just become

```
case object Bool extends BackendType
case object Char extends BackendType
case object Int8 extends BackendType
case object Int16 extends BackendType
case object Int32 extends BackendType
case object Int64 extends BackendType
case object Float32 extends BackendType
case object Float64 extends BackendType
case class Array(tpe: BackendType) extends BackendType
case class Reference(ref: JvmName/ClassDesc/Whatever) extends BackendType
```
which is just `JvmType` with arrays, which is probably a good place to be

magnus-madsen:
> Following this idea to its conclusion, BackendType wil just become
> 
> ```
> case object Bool extends BackendType
> case object Char extends BackendType
> case object Int8 extends BackendType
> case object Int16 extends BackendType
> case object Int32 extends BackendType
> case object Int64 extends BackendType
> case object Float32 extends BackendType
> case object Float64 extends BackendType
> case class Array(tpe: BackendType) extends BackendType
> case class Reference(ref: JvmName/ClassDesc/Whatever) extends BackendType
> ```
> 
> which is just `JvmType` with arrays, which is probably a good place to be

JvmType is gone right?

Does it mean that BOT could begin to hold a reference to a ClassDesc and we can build on that?

magnus-madsen (approved):

magnus-madsen:
I think having ClassConstants would be good.

JonathanStarup:
> JvmType is gone right?

Yea, BackendType is / will be JvmType 2.0 (where arrays can be represented properly)

> Does it mean that BOT could begin to hold a reference to a ClassDesc and we can build on that?

BOT will be a lot simpler yeah. A simple data type.

Issue #10795, opened 2025-06-08
Rework `ProcessHandle`


magnus-madsen:
We current have a `Process` effect that returns a `ProcessHandle`:

```
enum ProcessHandle({
    exitValue = Unit -> Int32 \ {Exec, IO},
    isAlive = Unit -> Bool \ {Exec, IO}, 
    stop = Unit -> Unit \ {Exec, IO}
    })
```

We to change this to:

```
import java.lang.{ProcessHandle => JProcessHandle}
enum ProcessHandle(JProcessHandle)
```

So `ProcessHandle` wraps the Java handle.

Next we need to introduce a `ProcessControl` effect:

```
eff ProcessControl {
    def exitValue(ph: ProcessHandle): Int32
    def isAlive(ph: ProcessHandle): Bool
    def stop(ph: ProcessHandle): Bool
}
```

magnus-madsen:
In the above design, we have two effects: `Process` and `ProcessControl`. Another design might be to combine them into one effect.

magnus-madsen:
Forget about ProcessControl as a separate effect. I think the next step should be to move these functions (e.g. exitValue) into the Process effect itself.

PR #10862, opened 2025-06-11
refactor: use `java.lang.Process` as `JProcess`

CadeMichael:
Related to #10795


magnus-madsen:
Looks good. The next step is that we need to turn exit etc. into effect operations. I will write more in the ticket.

PR #10863, opened 2025-06-11
refactor: introduce ClassConstants

JonathanStarup:


main/src/ca/uwaterloo/flix/language/phase/jvm/ClassConstants.scala
```diff
@@ -0,0 +1,194 @@
...
+import ca.uwaterloo.flix.api.Flix
+import ca.uwaterloo.flix.language.phase.jvm.ClassMaker.{ConstructorMethod, InstanceMethod, InterfaceMethod, StaticMethod}
+import ca.uwaterloo.flix.language.phase.jvm.ClassMaker.Visibility.IsPublic
+import ca.uwaterloo.flix.language.phase.jvm.JvmName.MethodDescriptor
+import ca.uwaterloo.flix.language.phase.jvm.JvmName.MethodDescriptor.mkDescriptor
+import org.objectweb.asm.MethodVisitor
+
+object ClassConstants {
+
+  // Flix Constants.
```
magnus-madsen, line 12: Getting this separated seems good, but what is the end goal?

- I assume there should be no behavior here-- i.e. no genByteCode
- What should Constructor and StaticMethod etc be replaced by? 

The invoke API is: https://docs.oracle.com/en/java/javase/24/docs/api/java.base/java/lang/classfile/CodeBuilder.html#invoke(java.lang.classfile.Opcode,java.lang.constant.ClassDesc,java.lang.String,java.lang.constant.MethodTypeDesc,boolean)

So perhaps MethodTypeDesc? But that seems to loose some information.

main/src/ca/uwaterloo/flix/language/phase/jvm/ClassConstants.scala
```diff
@@ -0,0 +1,194 @@
...
+
+  object FlixError {
+
+    def Constructor: ConstructorMethod = ConstructorMethod(JvmName.FlixError, List(BackendType.String))
```
magnus-madsen: Why are these methods and not fields?

JonathanStarup: There was issues with mutual references in BOT, but maybe the cycle is broken now - let me check

main/src/ca/uwaterloo/flix/language/phase/jvm/ClassConstants.scala
JonathanStarup: > I assume there should be no behavior here-- i.e. no genByteCode

Conceptually no - genByteCode can use these constants but is independent. For now I've just put them here in a lack of a better place, but `FlixError.genByteCode` can be placed anywhere

> What should Constructor and StaticMethod etc be replaced by?

It could be kept as is - but replacing `MethodDescriptor` by `MethodTypeDesc` potentially

I think it'll be clearer both what this abstraction offers and whether we should keep it as the clean up progresses

magnus-madsen: Not sure why we want an extra layer of abstraction (?)

JonathanStarup: StringBuilder.appendString is used 22 places. Calling this method looks like

```
mv.visitMethodInstruction(Opcodes.INVOKEVIRTUAL, JvmName.StringBuilder, "append", mkDescriptor(JvmName.String)(JvmName.StringBuilder), isInterface = false)
```
The fact that the method is called `"append"`, is not an interface method, and has a `String -> StringBuilder` signature would be hardcoded and duplicated 22 places without `StringBuilder.AppendStringMethod`.

The ClassMaker abstractions just serve to have that information in one place which i think is very obviously good here.

magnus-madsen: I suppose one could wonder why StringBuilder is used in 22 places in the first place?

JonathanStarup: I mean StringBuilder specifically doesn't matter, unless you believe that no function will be referenced more than once.

Such a collection of methods also serve as a singular place where we list all of our assumptions about classes present in the standard library

JonathanStarup: I think we should merge this even if you don't agree with it. Right now I'm just trying to split the ClassMaker stuff away from BOT. We can delete or change things later. This PR is not introducing a new abstraction, its just refactoring and cleaning up

main/src/ca/uwaterloo/flix/language/phase/jvm/ClassConstants.scala
```diff
@@ -0,0 +1,194 @@
+package ca.uwaterloo.flix.language.phase.jvm
```
magnus-madsen, line 1: Add copyright header.

main/src/ca/uwaterloo/flix/language/phase/jvm/ClassConstants.scala
magnus-madsen: I think we can merge this-- it is a step in the right direction.

magnus-madsen (approved):

magnus-madsen:
You can add the copyright header in a separate PR

PR #10866, opened 2025-06-12
refactor: make `ProcessHandle` functions `Process` effects

CadeMichael:
Addresses #10795.

### DIFF
--- a/main/src/library/Process.flix
+++ b/main/src/library/Process.flix
@@ -26,6 +26,22 @@ eff Process {
     /// Immediately executes the command `cmd` with the arguments `args`, by the path `cwd` and with the environmental `env`.
     ///
     def execWithCwdAndEnv(cmd: String, args: List[String], cwd: Option[String], env: Map[String, String]): ProcessHandle
+
+    ///
+    /// Returns the exit value of the process.
+    ///
+    def exitValue(ph: ProcessHandle): Int32
+
+    ///
+    ///  Returns status of the process, if it alive or not.
+    ///
+    def isAlive(ph: ProcessHandle): Bool
+
+    ///
+    /// Stops the process.
+    ///
+    def stop(ph: ProcessHandle): Unit
+
 }
 
 mod Process {
@@ -89,6 +105,43 @@ mod Process {
                     case ex: IOException               => Err(IoError(ErrorKind.Other, ex.getMessage()))
                 }
             }
+
+            def exitValue(ph, k) =
+                try {
+                    match ph {
+                        case ProcessHandle.ProcessHandle(jProc) => k(jProc.exitValue())
+                    }
+                } catch {
+                    case ex: IllegalArgumentException  => Err(IoError(ErrorKind.InvalidInput, ex.getMessage()))
+                    case ex: NullPointerException      => Err(IoError(ErrorKind.InvalidInput, ex.getMessage()))
+                    case ex: SecurityException         => Err(IoError(ErrorKind.PermissionDenied, ex.getMessage()))
+                    case ex: IOException               => Err(IoError(ErrorKind.Other, ex.getMessage()))
+                }
+
+            def isAlive(ph, k) =
+                try {
+                    match ph {
+                        case ProcessHandle.ProcessHandle(jProc) => k(jProc.isAlive())
+                    }
+                } catch {
+                    case ex: IllegalArgumentException  => Err(IoError(ErrorKind.InvalidInput, ex.getMessage()))
+                    case ex: NullPointerException      => Err(IoError(ErrorKind.InvalidInput, ex.getMessage()))
+                    case ex: SecurityException         => Err(IoError(ErrorKind.PermissionDenied, ex.getMessage()))
+                    case ex: IOException               => Err(IoError(ErrorKind.Other, ex.getMessage()))
+                }
+
+            def stop(ph, k) =
+                try {
+                    match ph {
+                        case ProcessHandle.ProcessHandle(jProc) => k(jProc.destroy())
+                    }
+                } catch {
+                    case ex: IllegalArgumentException  => Err(IoError(ErrorKind.InvalidInput, ex.getMessage()))
+                    case ex: NullPointerException      => Err(IoError(ErrorKind.InvalidInput, ex.getMessage()))
+                    case ex: SecurityException         => Err(IoError(ErrorKind.PermissionDenied, ex.getMessage()))
+                    case ex: IOException               => Err(IoError(ErrorKind.Other, ex.getMessage()))
+                }
+
         }
 
     ///
--- a/main/src/library/ProcessHandle.flix
+++ b/main/src/library/ProcessHandle.flix
@@ -21,28 +21,3 @@ import java.lang.{Process=> JProcess}
 ///
 enum ProcessHandle (JProcess)
 
-mod ProcessHandle {
-
-    ///
-    /// Returns the exit value of the process.
-    ///
-    pub def exitValue(ph: ProcessHandle): Int32 \ {Exec, IO} = match ph {
-        case ProcessHandle.ProcessHandle(jProc) => jProc.exitValue()
-    }
-
-    ///
-    ///  Returns status of the process, if it alive or not.
-    ///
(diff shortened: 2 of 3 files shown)


main/src/library/Process.flix
```diff
@@ -89,6 +105,43 @@ mod Process {
...
             }
+
+            def exitValue(ph, k) =
+                try {
+                    match ph {
+                        case ProcessHandle.ProcessHandle(jProc) => k(jProc.exitValue())
+                    }
+                } catch {
+                    case ex: IllegalArgumentException  => Err(IoError(ErrorKind.InvalidInput, ex.getMessage()))
+                    case ex: NullPointerException      => Err(IoError(ErrorKind.InvalidInput, ex.getMessage()))
```
magnus-madsen: As a rule, we don't check for nullptrs.

main/src/library/Process.flix
```diff
@@ -89,6 +105,43 @@ mod Process {
...
                 }
             }
+
+            def exitValue(ph, k) =
+                try {
+                    match ph {
+                        case ProcessHandle.ProcessHandle(jProc) => k(jProc.exitValue())
+                    }
+                } catch {
+                    case ex: IllegalArgumentException  => Err(IoError(ErrorKind.InvalidInput, ex.getMessage()))
```
magnus-madsen: I am not sure that can happen here. exitValue takes to no arguments, right?

magnus-madsen: same comments about the other methods.

magnus-madsen:
Can you do some manual testing just to check that it works.

Are there any other operations we should support?

main/src/library/Process.flix
```diff
@@ -26,6 +26,22 @@ eff Process {
     /// Immediately executes the command `cmd` with the arguments `args`, by the path `cwd` and with the environmental `env`.
     ///
     def execWithCwdAndEnv(cmd: String, args: List[String], cwd: Option[String], env: Map[String, String]): ProcessHandle
+
+    ///
+    /// Returns the exit value of the process.
```
mlutze, line 31: What happens if the process has not finished?

CadeMichael: hmm good point maybe it should be an Option

magnus-madsen: It blocks right? What does the Java doc say?

CadeMichael: from the docs
[IllegalThreadStateException](https://docs.oracle.com/javase/8/docs/api/java/lang/IllegalThreadStateException.html) - if the subprocess represented by this Process object has not yet terminated

CadeMichael: Maybe this is why there was the illegalArgument exception

Class IllegalThreadStateException

 - [java.lang.Object](https://docs.oracle.com/javase/8/docs/api/java/lang/Object.html)
 - [java.lang.Throwable](https://docs.oracle.com/javase/8/docs/api/java/lang/Throwable.html)
- [java.lang.Exception](https://docs.oracle.com/javase/8/docs/api/java/lang/Exception.html)
- [java.lang.RuntimeException](https://docs.oracle.com/javase/8/docs/api/java/lang/RuntimeException.html)
- [java.lang.IllegalArgumentException](https://docs.oracle.com/javase/8/docs/api/java/lang/IllegalArgumentException.html)

java.lang.IllegalThreadStateException

CadeMichael: Should I change back to `IllegalArgument` or add `IllegalThread` ?

mlutze: I'd lean toward returning an option for this operation, but let's see what @magnus-madsen has to say.

main/src/library/Process.flix
```diff
@@ -84,11 +100,42 @@ mod Process {
...
+                    case ex: IOException        => Err(IoError(ErrorKind.Other, ex.getMessage()))
+                }
+
+            def stop(ph, k) =
+                try {
+                    match ph {
+                        case ProcessHandle.ProcessHandle(jProc) => k(jProc.destroy())
+                    }
+                } catch {
+                    case ex: SecurityException  => Err(IoError(ErrorKind.PermissionDenied, ex.getMessage()))
```
magnus-madsen: We should not have security exception anywhere

main/src/library/Process.flix
magnus-madsen: How does Java process to use exitValue correctly? How does one wait for the process to finish?

I suppose we can return an Option

CadeMichael: based on the docs it just errors out, you'd have to wait for `isAlive` to return `false` before you can get the exit value. To make it more user friendly I could add a call to `isAlive` and if it returns `true` you return `None` otherwise `Some(exitvalue)`

CadeMichael: kindof like this

```
            def exitValue(ph, k) = 
                try {
                    match ph {
                        case ProcessHandleDev.ProcessHandleDev(jProc) => k(Some(jProc.exitValue()))
                    }
                } catch {
                    case _: IllegalThreadStateException => k(None)
                    case ex: IOException                => Err(IoError(ErrorKind.Other, ex.getMessage()))
                }
```

- where `IllegalThreadStateException` is thrown while the process is still running

mlutze: Yes, I think this is the right solution.

magnus-madsen: And the matches can be come just let ProcessHandle(...) = ...

magnus-madsen: I think we also need to expose waitFor

magnus-madsen: Instead of calling isAlive. I would just call exitValue() and catch the exception.

magnus-madsen:
I think- eventually- the next step is to have ProcessHandle implement Reader/Writer

EDIT: In separate PR

magnus-madsen:
We should also brainstorm some primitives to build in top. E.g. we can look at https://doc.rust-lang.org/std/process/struct.Command.html#method.output

main/src/library/Process.flix
```diff
@@ -84,11 +99,38 @@ mod Process {
...
                     case ex: IllegalArgumentException  => Err(IoError(ErrorKind.InvalidInput, ex.getMessage()))
-                    case ex: NullPointerException      => Err(IoError(ErrorKind.InvalidInput, ex.getMessage()))
-                    case ex: SecurityException         => Err(IoError(ErrorKind.PermissionDenied, ex.getMessage()))
                     case ex: IOException               => Err(IoError(ErrorKind.Other, ex.getMessage()))
                 }
             }
+
+            def exitValue(ph, k) =
+                try {
+                    match ph {
```
magnus-madsen: Let us rewrite all the match ph to simple let bindings, let ProcessHandle.ProcessHandle(jproc). Maybe also use ProcessHandle so its shorter.

CadeMichael:
some of the tests

```
def main(): Result[IoError, Unit] \ {Exec, IO} =
    run {
        let p = Process.exec("ls", Nil);
        Process.waitFor(p);
        let ex_val = Process.exitValue(p);
        println("done");
        println(ex_val)
    } with Process.runWithIO

def main(): Unit \ {Exec, IO} =
    run {
        match ProcessWithResult.exec("ls", Nil) {
            case Ok(p) =>
                let _ = ProcessWithResult.waitFor(p);
                let ex_val = ProcessWithResult.exitValue(p);
                println("done");
                println(ex_val)
            case Err(e) => println(e)
        }
    } with ProcessWithResult.runWithIO
```

main/src/library/Process.flix
```diff
@@ -26,16 +26,38 @@ eff Process {
     /// Immediately executes the command `cmd` with the arguments `args`, by the path `cwd` and with the environmental `env`.
     ///
     def execWithCwdAndEnv(cmd: String, args: List[String], cwd: Option[String], env: Map[String, String]): ProcessHandle
+
+    ///
+    /// Returns the exit value of the process.
+    ///
+    def exitValue(ph: ProcessHandle): Int32
```
mlutze: We'll chage this to Option[Int32], right?

magnus-madsen: Yes

PR #10867, opened 2025-06-12
chore: remove dead code from Parser

mlutze:


main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -3748,14 +3748,4 @@ object Parser2 {
       close(mark, kind)
     }
   }
-
-  /** Returns a textual representation of a [[SyntaxTree.Tree]] meant for debugging. */
-  private def syntaxTreeToDebugString(tree: SyntaxTree.Tree, nesting: Int = 1): String = {
```
JonathanStarup, line 3753: Also note that after this was creasted, AstPrinter can now show the syntex tree in a similar manner

magnus-madsen (approved):

Issue #10865, opened 2025-06-12
Add effect for accepting TCP connection


rywiese:
Proposal:

```
import java.net.Socket

enum TcpSocket(Socket)

eff TcpAccept {
    def accept(server: TcpServer): TcpSocket
}
```

magnus-madsen:
Seems reasonable.

PR #10868, opened 2025-06-12
feat: add TcpAccept effect

rywiese:
fixes https://github.com/flix/flix/issues/10865


main/src/library/IoError.flix
```diff
@@ -37,6 +38,7 @@ mod IoError {
         case NotFile,
         case NotDirectory,
         case PermissionDenied,
+        case SocketTimeout,
```
magnus-madsen: In the interest of re usability perhaps this should just be called Timeout

main/src/library/IoError.flix
```diff
@@ -29,6 +29,7 @@ mod IoError {
     pub enum ErrorKind with Eq, ToString {
         case AlreadyExists,
         case ConnectionFailed,
+        case IllegalBlockingMode,
```
magnus-madsen: What is this?

main/src/library/TcpAccept.flix
```diff
@@ -0,0 +1,62 @@
...
+    ///
+    /// In other words, re-interprets the `TcpAccept` effect using the `Net` and `IO` effects.
+    ///
+    pub def handle(f: a -> b \ ef): a -> Result[IoError, b] \ {ef, IO, Net} - TcpAccept = x ->
+        run {
+            Ok(f(x))
+        } with handler TcpAccept {
+            def accept(server, resume) = {
+                try {
+                    let socket = match server {
```
magnus-madsen: You and @CadeMichael both have this type of pattern match. However, you can just write:

let TcpServer.TcpServer(serverSocket) = server

main/src/library/TcpAccept.flix
```diff
@@ -0,0 +1,62 @@
...
+                        case TcpServer.TcpServer(serverSocket) => serverSocket.accept()
+                    };
+                    resume(TcpSocket.TcpSocket(socket))
+                } catch {
+                    case ex: IOException            => Err(IoError(ErrorKind.Other, ex.getMessage()))
```
magnus-madsen, line 48: Alignment

magnus-madsen, line 48: IOException should typically be last because otherwise it may mask the other errors if they are subtypes.

main/src/library/IoError.flix
rywiese: From [the docs for accept](https://docs.oracle.com/javase/8/docs/api/java/net/ServerSocket.html#accept--), it can be thrown:

> if this socket has an associated channel, the channel is in non-blocking mode, and there is no connection ready to be accepted

rywiese: I'm trying to figure out if a user could set the channel to nonblocking mode. if not, maybe we don't need to account for this error

rywiese: Though if this error occurs, it means the user is not using the `ServerSocket` interface properly so maybe a crash is okay. I guess this exception shouldn't be handled at runtime via the `Result` type. Should I get rid of it?

rywiese: Yeah I think I've convinced myself that we don't need to handle this. User would have to go in and deliberately change the server to be nonblocking, which is not really the intended way this effect is supposed to be used. I've removed it

magnus-madsen: I don't think we need to handle it, indeed.

magnus-madsen:
🚀

PR #10870, opened 2025-06-12
feat: make InputStream instance of Reader

rywiese:
fixes https://github.com/flix/flix/issues/6402

### DIFF
--- a/main/src/library/Reader.flix
+++ b/main/src/library/Reader.flix
@@ -13,11 +13,22 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+use IoError.ErrorKind
+use IoError.IoError
+
+import java.io.InputStream
+import java.io.IOException
 
 ///
 /// A trait for types which are resources that can be accessed as bytes.
 ///
 trait Reader[t] {
+
+    ///
+    /// Effect associated with reading bytes from the underlying resource.
+    ///
+    type Aef: Eff
+
     ///
     /// Reads `k` bytes from the underlying resource and writes them into `b`.
     ///
@@ -27,5 +38,19 @@ trait Reader[t] {
     ///
     /// Returns `Err(e)` if some underlying I/O error occurs.
     ///
-    pub def read(b: Array[Int8, r], r: t): Result[IoError, Int32] \ {r, IO}
+    pub def read(b: Array[Int8, r], r: t): Result[IoError, Int32] \ {r, Reader.Aef[t]}
+
+}
+
+instance Reader[InputStream] {
+
+    type Aef = IO
+
+    pub def read(b: Array[Int8, r], r: InputStream): Result[IoError, Int32] \ {r, IO} =
+        try {
+            checked_ecast(Ok(r.read(b)))
+        } catch {
+            case ex: IOException => Err(IoError(ErrorKind.Other, ex.getMessage()))
+        }
+
 }


main/src/library/Reader.flix
```diff
@@ -13,11 +13,22 @@
...
 
 ///
 /// A trait for types which are resources that can be accessed as bytes.
 ///
 trait Reader[t] {
+
+    ///
+    /// Effect associated with reading bytes from the underlying resource.
+    ///
+    type Eff: Eff
```
magnus-madsen: Eff -> Aef

magnus-madsen: By convention

main/src/library/Reader.flix
```diff
@@ -27,5 +38,19 @@ trait Reader[t] {
...
-    pub def read(b: Array[Int8, r], r: t): Result[IoError, Int32] \ {r, IO}
+    pub def read(b: Array[Int8, r], r: t): Result[IoError, Int32] \ {r, Reader.Eff[t]}
+
+}
+
+instance Reader[InputStream] {
+
+    type Eff = IO
+
+    pub def read(b: Array[Int8, r], r: InputStream): Result[IoError, Int32] \ {r, IO} =
```
magnus-madsen, line 49: @mlutze Is this syntax considered deprecated or?

main/src/library/Reader.flix
```diff
@@ -27,5 +38,19 @@ trait Reader[t] {
...
+        try {
+            checked_ecast(Ok(r.read(b)))
```
magnus-madsen, line 51: I suppose this is because otherwise we do not get the r effect.

magnus-madsen:
Cool- so soon this can be linked up with the tcp server

main/src/library/Reader.flix
rywiese: Yeah

rywiese: I was wondering about this, bc ideally the java `InputStream::read` would be known to flix as having this `r` effect, since it reads from the region. In the flix docs it says:

> The Flix compiler ships with a built-in database that maps classes, constructors, and methods in the Java Class Library to primitive effects.

and maybe it's not possible to add `r` to that mapping bc it's not primitive (and also depends on the flix type of `b` argument)?

main/src/library/Reader.flix
rywiese: What syntax?

mlutze: I don't know that we made an official decision, but I prefer the effects written as:

```
r + IO
```

instead of

```
{r, IO}
```

The reason is that `r` is actually a set, so theoretically we should be `unioning` it with another set. Then `IO` is just syntactic sugar for `{IO}`

main/src/library/Reader.flix
```diff
@@ -27,5 +38,19 @@ trait Reader[t] {
...
+
+}
+
+instance Reader[InputStream] {
+
+    type Aef = IO
+
+    pub def read(b: Array[Int8, r], r: InputStream): Result[IoError, Int32] \ {r, IO} =
+        try {
+            checked_ecast(Ok(r.read(b)))
```
mlutze, line 51: I guess closing the resource is not the responsibility of this trait, right?

mlutze (approved):

main/src/library/Reader.flix
magnus-madsen: Its not yet possible. We can only describe primitive effects.

main/src/library/Reader.flix
magnus-madsen: No, especially because we are only reading a few bytes at a time. Don't want to close after reading the first batch.

main/src/library/Reader.flix
magnus-madsen: Lets do r + IO (we can do it in the next PR)

PR #10871, opened 2025-06-12
feat: self-contained class loading

JonathanStarup:


main/src/ca/uwaterloo/flix/language/phase/jvm/JvmBackend.scala
```diff
@@ -151,7 +152,19 @@ object JvmBackend {
...
+      case defn =>
+        val nsType = BackendObjType.Namespace(defn.sym.namespace)
+        BytecodeAst.Test(nsType.jvmName, nsType.ShimMethod(defn).name, defn.ann.isSkip)
+    }
+    val main = root.mainEntryPoint.map{
+      case _ =>
+        val mainType = BackendObjType.Main
+        BytecodeAst.Def(mainType.jvmName, mainType.MainMethod.name)
+    }
+    BytecodeAst.Root(classMap, tests, main, root.sources)
```
JonathanStarup, line 167: The Backend now describes its own structure, so the main of main and the lists of tests is returned instead of relying on cross-codebase knowledge

main/src/ca/uwaterloo/flix/language/ast/BytecodeAst.scala
```diff
@@ -0,0 +1,20 @@
+package ca.uwaterloo.flix.language.ast
+
+import ca.uwaterloo.flix.language.ast.shared.Source
+import ca.uwaterloo.flix.language.phase.jvm.{JvmClass, JvmName}
+
+
+object BytecodeAst {
+
+  case class Root(
+                   classes: Map[JvmName, JvmClass],
```
JonathanStarup, line 31: Ideally these would use ClassDesc, but they don't have `_.toBinaryName` so i prefer to leave this as a follow up PR

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmLoader.scala
```diff
@@ -29,133 +30,94 @@ object JvmLoader {
 
   case class LoaderResult(
                            main: Option[Array[String] => Unit],
-                           defs: Map[Symbol.DefnSym, () => AnyRef],
-                           byteSize: Int
+                           tests: Map[Symbol.DefnSym, (() => AnyRef, Boolean)],
```
JonathanStarup: It's a bit awkward with a boolean here, it is because the later stuff needs to know whether to skip at test or not.

I should probably add `LoaderResult.Test`

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmLoader.scala
```diff
@@ -29,133 +30,94 @@ object JvmLoader {
...
     */
-  private def load(classes: List[JvmClass])(implicit flix: Flix, root: Root): (Option[Array[String] => Unit], Map[Symbol.DefnSym, Method]) = {
+  private def load(root: Root)(implicit flix: Flix): (Option[Array[String] => Unit], Map[Symbol.DefnSym, (() => AnyRef, Boolean)]) = {
     // Load each class into the JVM in a fresh class loader.
-    implicit val loadedClasses: Map[JvmName, Class[?]] = loadAll(classes)
+    implicit val loadedClasses: Map[JvmName, Class[?]] = loadAll(root.classes.values)
 
-    // Computes a map from classes and method names to method objects.
-    // TODO: We should not load all method objects here. Only a subset. Need some notion of entry points.
-    implicit val allMethods: Map[Class[?], Map[String, Method]] = loadedClasses.foldLeft(Map.empty[Class[?], Map[String, Method]]) {
```
JonathanStarup, line 101: This map has been eliminated, the methodsOf(clazz) lookup is just done on demand

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmLoader.scala
```diff
@@ -29,133 +30,94 @@ object JvmLoader {
...
-      case (macc, (_, clazz)) => macc + (clazz -> methodsOf(clazz))
+    val tests = MapOps.mapValues(root.tests) {
+      case defn => (wrapTest(loadMethod(defn.className, defn.methodName)), defn.isSkip)
+    }
+    val main = root.main.map {
+      case defn => wrapMain(loadMethod(defn.className, defn.methodName))
     }
 
-    val entrypoints = root.defs.filter { case (sym, _) => root.entryPoints.contains(sym) }
```
JonathanStarup, line 105: This ad-hoc filtering is now contained in JvmBackend

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmLoader.scala
```diff
@@ -29,133 +30,94 @@ object JvmLoader {
...
-
-    loadMethod(ns.jvmName, methodName)
+    (main, tests)
   }
 
   /** Returns the [[Method]] object of the main function of `root` if it is defined. */
-  private def loadMain(root: Root)(implicit loadedClasses: Map[JvmName, Class[?]], allMethods: Map[Class[?], Map[String, Method]]): Option[Array[String] => Unit] = {
-    if (root.getMain.isEmpty) return None
-
-    val mainMethod = loadMethod(BackendObjType.Main.jvmName, BackendObjType.Main.MainMethod.name)
```
JonathanStarup, line 123: This name is now present in the AST instead of being a magic constant

main/src/ca/uwaterloo/flix/runtime/CompilationResult.scala
```diff
@@ -16,45 +16,44 @@
...
+  * @param sources   the sources of the program.
+  * @param classes   the JVM classes of the program.
   * @param totalTime the total compilation time, excluding class writing/loading.
-  * @param codeSize  the number of bytes the compiler generated.
   */
-class CompilationResult(root: Root,
-                        main: Option[Array[String] => Unit],
-                        defs: Map[Symbol.DefnSym, () => AnyRef],
-                        val totalTime: Long,
-                        val codeSize: Int) {
```
JonathanStarup, line 35: I've removed `codeSize` and added `classes`. This means that we avoid summing the byte size unless we ask for it (benchmarking)

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmLoader.scala
```diff
@@ -29,133 +30,94 @@ object JvmLoader {
...
+  /** Returns the [[Method]] object for `className.methodName`. */
+  private def loadMethod(className: JvmName, methodName: String)(implicit loadedClasses: Map[JvmName, Class[?]]): Method = {
     val mainClass = loadedClasses.getOrElse(className, throw InternalCompilerException(s"Class not found: '${className.toBinaryName}'.", SourceLocation.Unknown))
-    val mainMethods = allMethods.getOrElse(mainClass, throw InternalCompilerException(s"methods for '${className.toBinaryName}' not found.", SourceLocation.Unknown))
-    mainMethods.getOrElse(methodName, throw InternalCompilerException(s"Cannot find '$methodName' method of '${className.toBinaryName}'", SourceLocation.Unknown))
+    findMethod(mainClass, methodName).getOrElse(throw InternalCompilerException(s"Cannot find '$methodName' method of '${className.toBinaryName}'", SourceLocation.Unknown))
   }
 
-  /** Returns a map from names to method objects for the given class `clazz`. */
-  private def methodsOf(clazz: Class[?]): Map[String, Method] = {
```
JonathanStarup, line 149: Instead of creating a map and then lookup up into it, we now just ask for the method directly

main/src/ca/uwaterloo/flix/runtime/CompilationResult.scala
```diff
@@ -16,45 +16,41 @@
...
-    defs.collect {
-      case (sym, run) if root.defs(sym).ann.isTest =>
-        val skip = root.defs(sym).ann.isSkip
-        sym -> TestFn(sym, skip, run)
-    }
-  }
+class CompilationResult(main: Option[Array[String] => Unit],
+                        tests: Map[Symbol.DefnSym, TestFn],
+                        sources: Map[Source, SourceLocation],
+                        classes: Map[JvmName, JvmClass],
```
magnus-madsen: This makes the compilation result closely coupled to JvmBackend. Not sure why that is desirable.

JonathanStarup: The coupling to JvmBackend is unchanged. It is now explicit rather than hidden in the source code.

JonathanStarup: Or do you just mean the `classes` field? It could be `bytecodeSize: Unit => Int` instead if you prefer

JonathanStarup:
This PR makes it easier to plug in Cade's backend fx because you can make your own decisions about what classes you generate and what they are called.

main/src/ca/uwaterloo/flix/runtime/CompilationResult.scala
```diff
@@ -16,45 +16,41 @@
...
-    * Returns all the test functions in the program.
-    */
-  def getTests: Map[Symbol.DefnSym, TestFn] = {
-    defs.collect {
-      case (sym, run) if root.defs(sym).ann.isTest =>
-        val skip = root.defs(sym).ann.isSkip
-        sym -> TestFn(sym, skip, run)
-    }
-  }
+class CompilationResult(main: Option[Array[String] => Unit],
```
magnus-madsen, line 32: I would like this class restored to its old definition, so that it is independent of JvmName, BytecodeAst, etc.

magnus-madsen: Later those changes could come or not, but not in this PR

JonathanStarup: The point of this PR is to introduce `tests` and `main`, do you mean the whole class or just `classes`?

magnus-madsen: I would like CompilationResult to be independent of the backend. I would also restore codesize. Why would it not be free to compute?

JonathanStarup: > I would also restore codesize. Why would it not be free to compute?

I was trying to be consistent with getTotalLines and avoid collecting metrics eagerly. I think that is better. But i've reverted it.

> I would like CompilationResult to be independent of the backend

Is it good now?

JonathanStarup: Maybe to reiterate what this PR achieves here. CompilationResult is now independent of ReducedAst. It is fully its own thing. There is no assumptions about the internals of JvmBackend. Everything is directly here.

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmLoader.scala
```diff
@@ -88,74 +71,56 @@ object JvmLoader {
...
-    JvmUtils.getMethods(clazz).foldLeft(Map.empty[String, Method]) {
-      case (macc, method) =>
-        if (method.isSynthetic) macc
-        else macc + (method.getName -> method)
+  /** Returns a Method for `clazz.methodName` if possible. */
+  private def findMethod(clazz: Class[?], methodName: String): Option[Method] = {
+    JvmUtils.getMethods(clazz).find {
```
magnus-madsen, line 117: @JonathanStarup This here will still load all methods as reflective objects, AFAIK.

So it still consumes a lot of memory.

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmLoader.scala
```diff
@@ -29,53 +31,34 @@ object JvmLoader {
...
-    *   - for every entrypoint, there should exist a matching namespace function (see [[BackendObjType.Namespace.ShimMethod]])
-    *   - if the main function is defined, a main class should be defined (see [[BackendObjType.Main.MainMethod]])
     */
-  def run(root: Root, classes: List[JvmClass])(implicit flix: Flix): (Root, LoaderResult) = {
-    implicit val r: Root = root
-
-    // Collect code size for performance tracking.
-    val outputBytes = classes.map(_.bytecode.length).sum
-
+  def run(root: Root)(implicit flix: Flix): LoaderResult = {
```
magnus-madsen, line 43: This is a phase so maybe it should call flix.phase?

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmLoader.scala
magnus-madsen: But perhaps you are now only looking at a very small subset of classes?

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmLoader.scala
```diff
@@ -88,74 +71,56 @@ object JvmLoader {
...
-  /** Returns a map from names to method objects for the given class `clazz`. */
-  private def methodsOf(clazz: Class[?]): Map[String, Method] = {
-    JvmUtils.getMethods(clazz).foldLeft(Map.empty[String, Method]) {
-      case (macc, method) =>
-        if (method.isSynthetic) macc
-        else macc + (method.getName -> method)
+  /** Returns a Method for `clazz.methodName` if possible. */
+  private def findMethod(clazz: Class[?], methodName: String): Option[Method] = {
+    JvmUtils.getMethods(clazz).find {
+      case method => method.getName == methodName && !method.isSynthetic
```
magnus-madsen, line 118: Is the synthetic check even required?

magnus-madsen:
@JonathanStarup I merged this because it seems like a good step forward-- but please do address my comments (in text or follow-up PR).

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmLoader.scala
JonathanStarup: Afair you said to remove it to avoid loading/writing being included in arewefast

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmLoader.scala
JonathanStarup: Hmm yeah it seems bad to ask for the whole list just to select one

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmLoader.scala
JonathanStarup: not sure

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmLoader.scala
JonathanStarup: but the whole reflection look up is a bit hacky throughout the compiler - it could use a compiler wide overhaul

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmLoader.scala
magnus-madsen: Hmm. Ah because its not "real" - in the sense of how the compiler usually works.

magnus-madsen: But maybe now its small? You only load a few classes? or?

JonathanStarup: If you set `loadClassFiles = false` in options its instantaneous

main/src/ca/uwaterloo/flix/language/phase/jvm/JvmLoader.scala
magnus-madsen: Yes, but of course the two uses cases are widely different. We could probably refactor this part of the backend without messing with the typing part.

PR #10874, opened 2025-06-12
feat: safe zip

JonathanStarup:
And by safe I mean crashing - safer than silently removing list elements.

Cannot be merged before
- #10882
- #10881
- #10880
- #10878

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/CodeActionProvider.scala
@@ -23,6 +23,7 @@ import ca.uwaterloo.flix.language.ast.TypedAst.Root
 import ca.uwaterloo.flix.language.ast.shared.AnchorPosition
 import ca.uwaterloo.flix.language.ast.{Name, SourceLocation, SourcePosition, Symbol}
 import ca.uwaterloo.flix.language.errors.ResolutionError
+import ca.uwaterloo.flix.util.collection.ListOps
 
 /**
   * The CodeActionProvider offers quickfix suggestions.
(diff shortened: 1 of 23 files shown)


magnus-madsen:
How about we resurrect your ListOps and then much more gradually introduce zip and zipOption?

I would hate to introduce 100 new places for crashes at once. But doing it gradually would be very sensible.

JonathanStarup:
> I would hate to introduce 100 new places for crashes at once. But doing it gradually would be very sensible.

There is the crashes listed above. I think its much more sensible to fix those few places and then do it at once. Doing it gradually means that someone has to push that wagon and keep it going.

magnus-madsen:
> > I would hate to introduce 100 new places for crashes at once. But doing it gradually would be very sensible.
> 
> There is the crashes listed above. I think its much more sensible to fix those few places and then do it at once. Doing it gradually means that someone has to push that wagon and keep it going.

I just don't have time to review a big PR with so many critical changes.

magnus-madsen:
Like make it bite-sized

JonathanStarup:
> I just don't have time to review a big PR with so many critical changes.

I don't think this pr is scary since it reduces a subtle bug into an eager crash. Chopping it up also doesn't reduce the time cost, just postpones it.

In the interest of saving you time, maybe Matt (he said ok) could review it and we bubble up anything important to you?

magnus-madsen:
If someone is willing to review it and promise that they will not cause a single new crash then I am happy.

mlutze:
> If someone is willing to review it and promise that they will not cause a single new crash then I am happy.

kill-them-crash

mlutze:
I'll review when this is rebased.

magnus-madsen:
Uh oh. I don't remember saying that ?

mlutze:
> Uh oh. I don't remember saying that ?

It's been in my meme template folder since March.

magnus-madsen:
Not against this, but we need small PRs since each risk introducing a crash.

PR #10886, opened 2025-06-13
fix: InputStream Reader implementation returns 0 instead of -1

rywiese:
fixes a small mistake i overlooked in https://github.com/flix/flix/issues/6402

### DIFF
--- a/main/src/library/Reader.flix
+++ b/main/src/library/Reader.flix
@@ -34,7 +34,7 @@ trait Reader[t] {
     ///
     /// Returns `Ok(k)` to signify that `k` bytes was successfully read and written to `b`.
     ///
-    /// Guarantees that `0 <= k < length(b)`.
+    /// Guarantees that `0 <= k <= length(b)`.
     ///
     /// Returns `Err(e)` if some underlying I/O error occurs.
     ///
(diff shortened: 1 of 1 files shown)


main/src/library/Reader.flix
```diff
@@ -34,7 +34,7 @@ trait Reader[t] {
     ///
     /// Returns `Ok(k)` to signify that `k` bytes was successfully read and written to `b`.
     ///
-    /// Guarantees that `0 <= k < length(b)`.
+    /// Guarantees that `0 <= k <= length(b)`.
```
magnus-madsen, line 37: I am confused. This suggested that we could read bytes past the length of the array.

main/src/library/Reader.flix
```diff
@@ -48,7 +48,9 @@ instance Reader[InputStream] {
 
     pub def read(b: Array[Int8, r], r: InputStream): Result[IoError, Int32] \ {r, IO} =
         try {
-            checked_ecast(Ok(r.read(b)))
+            let result = r.read(b);
+            let numRead = if (result == -1) 0 else result;
```
magnus-madsen, line 52: -1 seems to indicate end of stream in Java
In Rust -- which this trait is based of -- https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read -- it seems 0 can mean end of stream. What would make sense to do here?

main/src/library/Reader.flix
rywiese: If we read every byte in the array, then the number of bytes read would be equal to the length of the array right? Since `k` is quantity and not index of last read byte. [Rust docs](https://doc.rust-lang.org/std/io/trait.Read.html#required-methods) use `<=` as well:

> If the return value of this method is [Ok(n)](https://doc.rust-lang.org/std/result/enum.Result.html#variant.Ok), then implementations must guarantee that 0 <= n <= buf.len()

main/src/library/Reader.flix
rywiese: Continuing the convo in https://github.com/flix/flix/issues/10895

PR #10888, opened 2025-06-13
feat: remove closure argument erasure

JonathanStarup:
Similarly to local arguments, there is no reason that captured closure args should be erased since they are specific to each closure

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
@@ -188,10 +188,11 @@ object GenExpression {
         mv.visitMethodInsn(INVOKESPECIAL, jvmName.toInternalName, JvmName.ConstructorMethod, MethodDescriptor.NothingToVoid.toDescriptor, false)
         // Capturing free args
         for ((arg, i) <- exps.zipWithIndex) {
-          val erasedArgType = BackendType.toErasedBackendType(arg.tpe)
+          val argType = BackendType.toBackendType(arg.tpe)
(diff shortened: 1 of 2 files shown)


main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -188,10 +188,11 @@ object GenExpression {
         mv.visitMethodInsn(INVOKESPECIAL, jvmName.toInternalName, JvmName.ConstructorMethod, MethodDescriptor.NothingToVoid.toDescriptor, false)
         // Capturing free args
         for ((arg, i) <- exps.zipWithIndex) {
-          val erasedArgType = BackendType.toErasedBackendType(arg.tpe)
+          val argType = BackendType.toBackendType(arg.tpe)
           mv.visitInsn(DUP)
           compileExpr(arg)
-          mv.visitFieldInsn(PUTFIELD, jvmName.toInternalName, s"clo$i", erasedArgType.toDescriptor)
+          BytecodeInstructions.castIfNotPrim(argType)
```
JonathanStarup, line 194: When applying the closure argument, ideally we shouldn't have to cast after calling compileExpr, but the casting protocol of the backend still has some holes which forces us to cast here. Over time this should go away

JonathanStarup: This means that in this PR, closure arguments are casted once upon capture, instead of casting them once at each call to the closure, reducing the total number of casts

magnus-madsen (approved):

PR #10889, opened 2025-06-13
fix: bug in ford fulkerson

CasperDN:
Back edges were treated wrongly, but the old engine never created them.
Furthermore many tests assumed a specific shortest path when there are multiple.
All such errors have not necessarily been identified, but finding them is a quite tedious process.

### DIFF
--- a/examples/larger-examples/datalog/ford-fulkerson.flix
+++ b/examples/larger-examples/datalog/ford-fulkerson.flix
@@ -78,7 +78,7 @@ mod FordFulkerson {
     def minCapacity(p: Path[t], g: m[(t, Int32, Int32, t)]): Int32 \ Foldable.Aef[m] with Foldable[m], Order[t] =
         let onPath = (s, d) -> isForwardEdge(s, d, p) or isBackEdge(s, d, p);
         let optMin = g |> Foldable.filter(match (s, _, _, d) -> onPath(s, d))
-            |> List.map(match (_, u, f, _) -> u - f)
+            |> List.map(match (s, u, f, d) -> if (isForwardEdge(s, d, p)) u - f else f)
             |> List.minimum;
         match optMin {
             case Some(u) => u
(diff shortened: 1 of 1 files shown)


CasperDN:
Also with parallelism the way lattices have been used so far is `NonDet`. See e.g. `Path`.
For example when the order of comparison matters the engine is free to do either leading to different results.

magnus-madsen (approved):

magnus-madsen:
Thanks-- perhaps this is not so great an example.

examples/larger-examples/datalog/ford-fulkerson.flix
```diff
@@ -123,20 +123,20 @@ mod FordFulkerson {
...
+    instance Order[Path[a]] with Order[a] {
         pub def compare(x: Path[a], y: Path[a]): Comparison = match (x, y) {
-            case (Bot, Bot)           => Comparison.EqualTo
-            case (Bot, _)             => Comparison.LessThan
-            case (_, Bot)             => Comparison.GreaterThan
-            case (Path(xs), Path(ys)) => List.length(xs) <=> List.length(ys)
+            case (Bot, Bot)                 => Comparison.EqualTo
+            case (Bot, _)                   => Comparison.LessThan
+            case (_, Bot)                   => Comparison.GreaterThan
+            case (Path(list1), Path(list2)) => list1 <=> list2
```
magnus-madsen: Why rename the variables here? xs and ys are standard names.

CasperDN: Sorry, `Path` was 'fixed' based on the code in `single-source-shortest-path-arbitrary.flix`. It had a slightly different enum, and I simply adapted the code, which resulted in the renaming. I'll undo it.

CasperDN:
I think the fundamental problem with these examples is that they use lattice semantics with non-lattice elements. As in `Path` is not really a lattice.
It works due to details in the implementation, but could easily break. They do not, fundamentally, break in our engine, but changing the way comparisons are performed in the engine could lead to an infinite loop.

magnus-madsen:
>  It works due to details in the implementation, but could easily break. They do not, fundamentally, break in our engine, but changing the way comparisons are performed in the engine could lead to an infinite loop.

How do you get an infinite loop?

CasperDN:
I use `leastUpperBound` in reverse order, thereby producing 'new' elements in each iteration as we now always pick the latest shortest path produced. Due to `lessEqual` it might actually not run infinitely, but verifying it becomes a lot harder.

CasperDN:
I guess that we actually form a lattice as long as we agree that according to the lattice all paths beginning in `s` and ending in `t` of length `k` are identical. Least upper bound of paths starting in `s` and `s'` are still undefined though...

PR #10890, opened 2025-06-14
wip add type parameters to effect declarations

mlutze:


magnus-madsen:
What about kinds? For my purpose it would be OK if in `Throw[t]`, `t` is required to be a `Type`. I am also OK with at most one type parameter.

We can always lift these restrictions later.

mlutze:
I'm just going to add support for the most general case. We can add restrictions once we know what they are.

magnus-madsen:
I think one more test to be dropped.

magnus-madsen:
But we should ensure that we don't break Flix for end-users while we experiment. What do you suggest?

We leave in the frontend check?

magnus-madsen:
You know "do no harm"

mlutze:
> But we should ensure that we don't break Flix for end-users while we experiment. What do you suggest?
> 
> We leave in the frontend check?

Behind a flag or what?

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -1250,12 +1250,8 @@ object Parser2 {
       expect(TokenKind.KeywordEff)
       nameUnqualified(NAME_EFFECT)
 
-      // Check for illegal type parameters.
       if (at(TokenKind.BracketL)) {
-        val mark = open()
-        val loc = currentSourceLocation()
         Type.parameters()
-        closeWithError(mark, WeederError.IllegalEffectTypeParams(loc))
       }
```
magnus-madsen: Let us re-add this logic for now.

magnus-madsen:
@mlutze I think this is ready to be merged. Just re-add the check in the Parser2.

magnus-madsen (approved):

PR #10892, opened 2025-06-14
experiment: `while`

JonathanStarup:
```
while (Ref.get(cantSleep)) {
    addWhileLoop()
//  ^^^^^^^^^^^^^^ useless expression
}
```

- Good for user programs (I've implemented my own wonky while at least thrice)
- Good for backend rewrites (Optimizer could probably use it)
- Good for showing that effects are cool (error when body is pure)


main/src/ca/uwaterloo/flix/language/phase/Redundancy.scala
```diff
@@ -441,6 +441,19 @@ object Redundancy {
       val us3 = visitExp(exp3, env0, rc)
       us1 ++ us2 ++ us3
 
+    case Expr.While(exp1, exp2, _, _, _) =>
+      val us1 = visitExp(exp1, env0, rc)
+      val us2 = visitExp(exp2, env0, rc)
+      val pureCondition = exp1 match {
```
JonathanStarup, line 447: This is the only interesting part. Should we warn users about pure conditions? `while (true) ..` might be used when e.g. an effect will break the loop or the loop is truly infinite (server perhaps).

This version allows `while(true)` but doesn't allow `while(0 < 1)` since that is probably user error

main/src/ca/uwaterloo/flix/language/phase/optimizer/OccurrenceAnalyzer.scala
```diff
@@ -161,6 +161,18 @@ object OccurrenceAnalyzer {
           (Expr.IfThenElse(e1, e2, e3, tpe, eff, loc), ctx4)
         }
 
+      case Expr.While(exp1, exp2, tpe, eff, loc) =>
+        val (e1, ctx1) = visitExp(exp1)
+        val (e2, ctx2) = visitExp(exp2)
+        // TODO: No clue what to do here
+        //       Maybe OnceInLambda is kind of the right thing?
```
JonathanStarup: @jaschdoc when you have time, do you know what we should do here about the occurance of the while body?

jaschdoc: I think we need a new Occur variant for this

JonathanStarup: you can think of while as being `while(() -> cond)(() -> body)`, doesnt that justify the lambda occurance?

jaschdoc: Right,  but that seems hacky or lazy. I think either we add a proper constructor and support it fully. It may have the same semantics currently, but it could change in the future and for readabillity / maintainability it is much more pure to have a type for it. Otherwise, the compiler should unroll the while loop to `while(() -> cond)(() -> body)` but that defeats the point of having the expression.

JonathanStarup: So add `OnceInLoop`, and have it behave similarly to `OnceInLambda` in regards to the lattice?

jaschdoc: Yes :)

JonathanStarup: Will you look the inliner part over and see if you like it?

jaschdoc: Sure

main/src/ca/uwaterloo/flix/language/ast/MonoAst.scala
```diff
@@ -198,6 +200,15 @@ object MonoAst {
...
 
+    /**
+      * Represents a binder that occurs exactly once and that occurrence is inside a while loop (condition or body).
+      *
+      * If the let-binding is pure, then it is safe to move its definition to the occurrence.
+      *
+      * Doing so strictly decreases code size but may increase work.
+      */
+    case object OnceInWhile extends Occur
+
```
jaschdoc: Can you move this below `OnceInLocalDef` so it is sorted?

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -392,6 +402,7 @@ object Inliner {
 
       case Occur.Once
            | Occur.OnceInLambda
+           | Occur.OnceInWhile
```
jaschdoc: same here

JonathanStarup:
I've changed it to feat since it is a feature and ready to review since the code is done

main/src/ca/uwaterloo/flix/language/ast/MonoAst.scala
JonathanStarup: The list is already not sorted, but i can swap them sure

JonathanStarup: done now

jaschdoc: Ah yeah true sorry

jaschdoc: Thanks

jaschdoc: They are pseudo sorted haha

magnus-madsen:
This is not to be merged until there has been further discussion.

jaschdoc:
> This is not to be merged until there has been further discussion.

Yeah, didn't mean to approve for merge. Just wanted to approve specific changes :)

magnus-madsen:
I think there is a broader context here to consider: 

(A) What is the interaction with mutable local variables, and  
(B) Why can this not just be a higher-order function.

JonathanStarup:
> What is the interaction with mutable local variables, and

We don't have mutable variables. It interacts with structs/ref/array like everything else does

> Why can this not just be a higher-order function.

```
def while(cond: Unit -> Bool \ ef1, body: Unit -> ??? \ ef2): Unit \ ef1 + ef = ..
def example(): Unit =
    while (_ -> something.. ) ( _ -> something ..)
```
- It is ugly
- It doesn't work with `@MustUse` because in the above, you either have to make body return anything `t` or `Unit`. Both are worse than what a statement allows. (MustUse is broken, so ignore this if you accept that it should be removed)
- it doesn't allow custom user support, e.g. "your while loop perfoms no side-effects"

I really think we should add this, I thought that would be non-controversial. It is obviously something users want but it also sells Flix - it allows us to showcase why effects are good! We can warn the user that a while body is pure, or that the condition has no side effects.

I think that the only world in which it would be acceptable to not have while as an inbuilt thing, would be if I could do the following
```
def while(cond: Unit -> Bool \ ef1, body: Unit -> Unit \ ef2): Unit \ ef1 + ef2 where ef1 !~ {} and ef2 !~ {} = ..
def example(): Unit =
    while{something..}{something ..}
```
I.e. we have non-purity constraints `ef !~ {}` and we have thunk application syntax `f{something} === f(_ -> something)`

jaschdoc:
> We can warn the user that a while body is pure, or that the condition has no side effects.

And that variables in the condition do not occur in an effectful expression, e.g. `Ref.put`

PR #10899, opened 2025-06-17
feat: add predicate tracking

CasperDN:


main/src/library/Fixpoint3/Predicates.flix
```diff
@@ -0,0 +1,248 @@
...
+mod Fixpoint3.Predicates {
+    use Fixpoint.Ast.Shared.PredSym
+    use Fixpoint.Ast.Shared.PredSym.PredSym
+    use Fixpoint.Ast.Datalog.Datalog.Datalog
+    use Fixpoint.Ast.Datalog.{Datalog, Constraint, BodyPredicate, BodyTerm, HeadPredicate, HeadTerm, Polarity, VarSym}
+    use Fixpoint.Ast.Datalog.Constraint.Constraint
+    use Fixpoint.Ast.Datalog.HeadPredicate.HeadAtom
+    use Fixpoint.Ast.Datalog.BodyPredicate.{BodyAtom}
+    use Fixpoint3.Ast.Ram.{RamStmt, RelOp, RelSym, RamTerm, BoolExp, RowVar, Facts, Predicates, RamProgram, toId}
+    use Fixpoint3.Boxed
```
JonathanStarup, line 51: I think it'd be nice if this list is sorted

main/src/library/Fixpoint3/Predicates.flix
```diff
@@ -0,0 +1,248 @@
...
+
+    @Internal
+    pub enum PredType with Eq {
```
JonathanStarup, line 63: Can you add a doc comment to this?

main/src/library/Fixpoint3/Predicates.flix
```diff
@@ -0,0 +1,248 @@
...
+        if(id <= max) {
+            PredType.Full
+        } else if (id <= max * 2i64) {
+            PredType.Delta
+        } else {
+            PredType.New                
+        }
+
+    ///
+    /// Collect all preciate information given a Datalog program `program` and EDB `db`.
```
JonathanStarup: I think its nice to avoid acronyms without the full name. This is also the only place you call db a EDB, is this on purpose?

CasperDN: Not really. We normally consider it the EDB, so I didn't think about it.

JonathanStarup: I'd think it'd be most consistent just to remove the word EDB here

CasperDN: That was also what I ended up doing. Well I replaced it with 'database'.

main/src/library/Fixpoint3/Predicates.flix
```diff
@@ -0,0 +1,253 @@
...
+    use Fixpoint.Ast.Datalog.Constraint.Constraint
+    use Fixpoint.Ast.Datalog.HeadPredicate.HeadAtom
+    use Fixpoint.Ast.Datalog.BodyPredicate.{BodyAtom}
+    use Fixpoint3.Ast.Ram.{BoolExp, Facts, Predicates, RamProgram, RamStmt, RamTerm, RelOp, RelSym, RowVar, toId}
+    use Fixpoint3.Boxed
+
+    ///
+    /// `PredType` describes the three types of predicates that exists in a RAM program.
+    /// The `Full` relations contains all facts for the relation, the `Delta` is for fixpoint iterations
+    /// and the `New` relations are the relations that values are inserted into in project statements.
```
JonathanStarup: Maybe I'm just slow, but I don't fully understand the New explanation

magnus-madsen: I think the description can indeed be better. Also, make it into 3 bullet points

CasperDN: Better?

main/src/library/Fixpoint3/Predicates.flix
```diff
@@ -0,0 +1,257 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ 
+///
+/// This files keeps track predicates (`PredSym`). It keeps information about which `PredSym`s exist
+/// and which predicates represent full, delta and new relations.
```
magnus-madsen: The sentence has some grammar issues.

I am also not sure how a file can keep track of anything since there is no state in Flix.

main/src/library/Fixpoint3/Predicates.flix
```diff
@@ -0,0 +1,257 @@
...
+///
+/// It uses the following convention: Assume that there are `k` predicates. The original predicate
+/// `PredSym(name, id)` will be the full predicate while `PredSym(name, id + k)` will be the 
+/// delta-predicate and `PredSym(name, id + 2k)` will be the new-predicate.
```
magnus-madsen: I don't fully understand. I think you need to describe the overall goal-- which seems to be some integer encoding of predicate symbols, right?

Perhaps giving an example would be useful.

If the top text here makes sense then I am sure the rest of the file follows.

main/src/library/Fixpoint3/Predicates.flix
```diff
@@ -0,0 +1,257 @@
...
+///
+/// Other code should not know or rely on this convention.
+///
+mod Fixpoint3.Predicates {
```
magnus-madsen: We typically don't give modules names in plural.

main/src/library/Fixpoint3/Predicates.flix
CasperDN: Better?

main/src/library/Fixpoint3/Predicates.flix
CasperDN: I have rewritten the description and added an example. Is the new version more readable?

main/src/library/Fixpoint3/Predicates.flix
```diff
@@ -0,0 +1,268 @@
...
+ 
+///
+/// This file provides an interface for interacting with `Fixpoint.Ast.Ram.Predicates`.
+/// Core features include providing a list of all `RelSym` in `Predicates`,
+/// provide the `PredType` of a `RelSym` and transforming `RelSym` of one
+/// `PredType` into a `RelSym` of another `PredType`.
+///
+/// The three `PredType`s originate from semi-naive evaluation and are `Full`, `Delta` and `New`
+/// (delta').
+///
```
JonathanStarup, line 26: I like it. I think here, I'd add that:

Instead of adding an enum type (Full/Delta/New) to the predicate, we embed the type into the identifier, which we can do since we have an upper bound on the integers used.

magnus-madsen:
@JonathanStarup Ping me when you are happy.

JonathanStarup:
@magnus-madsen didn't to full detail, but i am happy

Issue #10756, opened 2025-06-04
Rework `inject` to `inject/X`


magnus-madsen:
See #10529 

Very simply `inject` must now be given an arity, e.g. `inject edges into Edge/2`. 

Hence, we need PRs to update the syntax on
- [ ] Flix book
- [ ] Flix website
- [ ] Flix playground (I suppose this can be pulled from master when the above PR is merged)
- [ ] Any material in https://github.com/magnus-madsen/advprog that use `inject`.

jaschdoc:
Will you rebuild the playground when Matt's PR is in?

magnus-madsen:
> Will you rebuild the playground when Matt's PR is in?

Done.

magnus-madsen:
I think we are done with this right? Other than the check

jaschdoc:
Yes I believe so

PR #10532, opened 2025-05-09
feat: add arity to predicates

mlutze:
depends on
- https://github.com/flix/flix/pull/10529

PR #10900, opened 2025-06-17
test: add tests for `inject`

jaschdoc:
Depends on https://github.com/flix/flix/pull/10532
Related to https://github.com/flix/flix/issues/10756

### DIFF
--- a/main/test/flix/Test.Exp.Fixpoint.Inject.flix
+++ b/main/test/flix/Test.Exp.Fixpoint.Inject.flix
@@ -1,0 +1,58 @@
+mod Test.Exp.Fixpoint.Inject {
+
+    @test
+    def inject01(): #{ Pred(Bool) | r } = {
+        inject (Nil: List[Bool]) into Pred/1
+    }
+
+    @test
+    def inject02(): #{ Pred(Char) | r } = {
+        inject (Nil: List[Char]) into Pred/1
+    }
+
+    @test
+    def inject03(): #{ Pred(Float32) | r } = {
+        inject (Nil: List[Float32]) into Pred/1
+    }
+
+    @test
+    def inject04(): #{ Pred(Float64) | r } = {
+        inject (Nil: List[Float64]) into Pred/1
+    }
+
+    @test
+    def inject05(): #{ Pred(BigDecimal) | r } = {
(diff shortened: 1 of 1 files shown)


magnus-madsen:
Two things: Could we do end-to-end like so:

(A)

let l = List#{(1, 2), (2, 3), (3, 4)};
let pr = inject l into A/2;
let r = query pr select (x, y) from A(x, y);
l == r

(B)

I see we have some mixture of primitive types. That is good. But we should also have on non-primitive, e.g. string.

magnus-madsen:
Testing arity 0 ... n is very good. Testing that more than one inject works is less important because it is really just doing the same thing twice.

jaschdoc:
> Testing arity 0 ... n is very good. Testing that more than one inject works is less important because it is really just doing the same thing twice.

Alright. My idea was to try to cover some of the combinations of types, arities, foldable instances etc.

magnus-madsen:
> > Testing arity 0 ... n is very good. Testing that more than one inject works is less important because it is really just doing the same thing twice.
> 
> Alright. My idea was to try to cover some of the combinations of types, arities, foldable instances etc.

I think that is good. The only thing I would say is that more than 1 relation in inject is probably not worth (and there are so many dimensions to try).

jaschdoc:
> > > Testing arity 0 ... n is very good. Testing that more than one inject works is less important because it is really just doing the same thing twice.
> > 
> > 
> > Alright. My idea was to try to cover some of the combinations of types, arities, foldable instances etc.
> 
> I think that is good. The only thing I would say is that more than 1 relation in inject is probably not worth (and there are so many dimensions to try).

Sounds good. Can we keep the ones I have already added?

magnus-madsen:
> > > > Testing arity 0 ... n is very good. Testing that more than one inject works is less important because it is really just doing the same thing twice.
> > > 
> > > 
> > > Alright. My idea was to try to cover some of the combinations of types, arities, foldable instances etc.
> > 
> > 
> > I think that is good. The only thing I would say is that more than 1 relation in inject is probably not worth (and there are so many dimensions to try).
> 
> Sounds good. Can we keep the ones I have already added?

I leave it to you.

PR #10903, opened 2025-06-17
feat: inline negated forward function calls

jaschdoc:
This allows the inliner to inline functions like
`pub def nonEmpty(l: List[a]): Bool = not isEmpty(l)`


magnus-madsen:
I can kind of see why this would be useful- although I don't know if it matters in practice? You have examples?

My concern is just when I look at the heuristics again my head explodes: isTrivial, isSimple, isSimpleCall, ... Argh :)

What can be done?

magnus-madsen:
(Did not mean to close)

magnus-madsen:
Maybe its ok. Let me sit on it.

magnus-madsen:
I decided against this-- for now.

magnus-madsen:
Let us let the inliner rest for a bit-- it deserves it.

PR #10906, opened 2025-06-17
fix: enforce predicate arity greater than zero

### LINKED ISSUE #10901: Disallow `inject .. into ../0` in Weeder

magnus-madsen:
Fixes #10901

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
+++ b/main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
@@ -495,6 +495,23 @@ object WeederError {
   }
 
   /**
+    * An error raised to indicate an illegal predicate arity.
+    *
+    * @param loc the location where the error occurs.
+    */
+  case class IllegalPredicateArity(loc: SourceLocation) extends WeederError {
+    override def summary: String = "Illegal predicate arity."
+
+    override def message(formatter: Formatter): String = {
+      import formatter.*
+      s""">> Illegal predicate arity. Arity must be an integer larger than zero.
+         |
+         |${code(loc, "illegal arity.")}
+         |""".stripMargin
+    }
+  }
+
+  /**
     * An error raised to indicate an illegal private declaration.
     *
     * @param ident the name of the declaration.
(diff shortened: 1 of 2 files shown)


magnus-madsen:
@jaschdoc I am reusing the same error-- which I think is OK.

main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
```diff
@@ -494,6 +494,23 @@ object WeederError {
...
+    * An error raised to indicate an illegal predicate arity.
+    *
+    * @param loc the location where the error occurs.
+    */
+  case class IllegalPredicateArity(loc: SourceLocation) extends WeederError {
+    override def summary: String = "Illegal predicate arity."
+
+    override def message(formatter: Formatter): String = {
+      import formatter.*
+      s""">> Illegal predicate arity. Arity must be an integer larger than zero.
```
JonathanStarup, line 507: If I write `inject .. into Pred/1_000` the error says `Arity must be an integer larger than zero`

JonathanStarup: I'd just add a `hint: String` argument to `IllegalPredicateArity`

magnus-madsen: Because 1_000 is a different token or what?

JonathanStarup: `1_000` and `1000` and `0xFF` are all just LiteralInt

`case None` in weeder is hit because `token.text` is either a hex number and/or a number with `_`.

The old error was also not super good but I think just having one be `hint = "Arity must be larger than zero."` and the other be `hint = "Arity can not use '_' or be a hex-number"` could be nice

magnus-madsen: You can merge and then create a follow up PR

JonathanStarup: To me it makes sense to do here, is it just a time budget thing?

magnus-madsen: Yes. But feel free to create a new PR or a ticket.

For me I think the interesting case is when you forget the arity. What happens there?

JonathanStarup: parser says "unexpected xyz, expected literal int" i think

JonathanStarup (approved):

PR #10908, opened 2025-06-18
feat: add skeleton for `ErrorCode`

magnus-madsen:


magnus-madsen:
@JonathanStarup @mlutze Does this sparkle with the boys? Does it bring joy.

This is based on the Rust design

main/src/ca/uwaterloo/flix/language/errors/ErrorCode.scala
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ca.uwaterloo.flix.language.errors
+
+class ErrorCode(val number: Int)
+
+object ErrorCode {
+
+  case object E51895 extends ErrorCode(51895) // UNUSED
```
JonathanStarup, line 22: Don't know how we feel about OOP, but it seems easy to have a `def code(ec: ErrorCode): Int = match ec ..` function and avoid this inheritance stuff

I think calling it MismatchedArity (or whatever) is better than E51895. In this version it's hard to look at `ParserError.MismatchedArity` and see whether its correct or not. "does E51895 belong to MIsmatchedArity? who knows"

mlutze:
> @JonathanStarup @mlutze Does this sparkle with the boys? Does it bring joy.
> 
> This is based on the Rust design

I'm OK with this though it breaks the "it should be doable in Flix" policy.

main/src/ca/uwaterloo/flix/language/errors/ErrorCode.scala
magnus-madsen: Do you mean just one pattern match on CompilationMessage?

magnus-madsen: Its simple, elegant, but a long function with no guarantee that the RHS would be unique.

JonathanStarup: Easier to show https://github.com/flix/flix/pull/10934

> but a long function with no guarantee that the RHS would be unique.

The "long function" always have to exist, your version just lives inside `extends ErrorCode(..)`

magnus-madsen:
I like this design; but this is not a priority for now.

Issue #10894, opened 2025-06-16
Extend `Process` and `ProcessWithResult`


magnus-madsen:
Add the following operations to Process:
 
- [ ] `stdout` - ProcessOutputStream
- [ ] `stderr` - ProcessOutputStream
- [ ] `stdin` - ProcessInputStream
- [ ] `output` - returns Result[IoError, String]
- [ ] `pid`
- [ ] `waitForWithTimeout` - similar design to Ping
- [ ] stop or destroy?

Essentially ProcessOutputStream wraps OutputStream while implementing the `Reader` trait.

CadeMichael:
the java api for `waitFor` with the timeout takes a long (`Int64`) and allows the user to choose a timeUnit. leveraging the new timeunit function I have to milliseconds from a few PR's ago I have this. Does it make sense or should I refactor?

```
def wait(ph, time, unit, k) =
    try {
        let timeMS = time * Int32.toInt64(TimeUnit.toMilliseconds(unit));
        let ProcessHandleDev(jProc) = ph;
        let exitVal = jProc.waitFor(timeMS, JTimeUnit.MILLISECONDS);
        k(exitVal)
    } catch {
        case ex: InterruptedException => Err(IoError(ErrorKind.Interrupted, ex.getMessage()))
        case ex: IOException          => Err(IoError(ErrorKind.Other, ex.getMessage()))
    }
```

I can hardcode milliseconds if that would be cleaner

CadeMichael:
> 
> Essentially ProcessOutputStream wraps OutputStream while implementing the `Reader` trait.

Should `ProcessOutput` be their own Modules or withing `Process.flix`? I could make two new modules `InputStream` and `OutputStream` that implement the needed traits and use them in the `Process` operations.

I see `Reader` has an instance for `InputStream`

```
instance Reader[InputStream] {

    type Aef = IO

    pub def read(b: Array[Int8, r], r: InputStream): Result[IoError, Int32] \ r + IO =
        try {
            let result = r.read(b);
            let numRead = if (result == -1) 0 else result;
            checked_ecast(Ok(numRead))
        } catch {
            case ex: IOException => Err(IoError(ErrorKind.Other, ex.getMessage()))
        }

}
```

magnus-madsen:
[...]

PR #10909, opened 2025-06-18
feat: handle process streams with Flix

CadeMichael:
Addresses #10894.

Handling process streams with flix wrappers and trait implementations.

### DIFF
--- a/main/src/library/Process.flix
+++ b/main/src/library/Process.flix
@@ -100,7 +100,7 @@ mod Process {
     use Process.StdErr.StdErr
     use Process.StdOut.StdOut
 
-    pub enum StdIn(JOutputStream)
+    pub enum StdIn(OutputStream)
     instance Writer[StdIn] {
         type Aef = IO
         pub def write(b: Array[Int8, r], r: StdIn): Result[IoError, Int32] \ r + IO =


main/src/library/Process.flix
```diff
@@ -73,7 +93,48 @@ mod Process {
     import java.lang.ProcessBuilder
     import java.lang.Runtime
     import java.io.{File => JFile}
+    import java.io.{InputStream => JInputStream}
+    import java.io.{OutputStream => JOutputStream}
```
magnus-madsen: I suppose we don't need to rename these anymore?

main/src/library/Process.flix
```diff
@@ -73,7 +93,48 @@ mod Process {
...
     import java.util.concurrent.{TimeUnit => JTimeUnit}
+    use Process.StdErr.StdErr
+    use Process.StdIn.StdIn
+    use Process.StdOut.StdOut
+
+    pub enum StdIn(JOutputStream)
+
+    instance Writer[StdIn] {
+
```
magnus-madsen: Remove linebreak

main/src/library/Process.flix
```diff
@@ -73,7 +93,48 @@ mod Process {
...
+        type Aef = IO
+
```
magnus-madsen: Remove linebreak

main/src/library/Process.flix
```diff
@@ -73,7 +93,48 @@ mod Process {
...
+        pub def write(b: Array[Int8, r], r: StdIn): Result[IoError, Int32] \ r + IO =
+            let StdIn(pr) = r;
+            Writer.write(b, pr)
+
```
magnus-madsen: Remove linebreak

main/src/library/Process.flix
magnus-madsen: Just to keep things tight. I think you can do the same for the other two instances.

main/src/library/Process.flix
```diff
@@ -149,6 +210,33 @@ mod Process {
                     case ex: IOException                   => Err(IoError(ErrorKind.Other, ex.getMessage()))
                 }
 
+            def stderr(ph, k) =
+                try {
+                    let ProcessHandle(jProc) = ph;
+                    k(StdErr.StdErr(jProc.getErrorStream()))
```
magnus-madsen: I would `use StdErr.StdErr` and so on

main/src/library/Process.flix
```diff
@@ -42,6 +42,26 @@ eff Process {
...
+
+    ///
+    /// Returns stdin InputStream of the process.
+    ///
+    def stdin(ph: ProcessHandle): Process.StdIn
+
+    ///
+    /// Returns stdout OutputStream of the process.
+    ///
+    def stdout(ph: ProcessHandle): Process.StdOut
```
magnus-madsen, line 53: For good style, I think the order should be:

stdin
stdout
stderr

Or is this because you are keeping them alphabetical?

main/src/library/Process.flix
```diff
@@ -149,6 +210,33 @@ mod Process {
...
+                } catch {
+                    case ex: IOException => Err(IoError(ErrorKind.Other, ex.getMessage()))
+                }
+
+            def stdout(ph, k) =
+                try {
+                    let ProcessHandle(jProc) = ph;
+                    k(StdOut.StdOut(jProc.getInputStream()))
+                } catch {
+                    case ex: IOException => Err(IoError(ErrorKind.Other, ex.getMessage()))
```
magnus-madsen: Can this exception happen here? When calling getInputStream? Or is it not only when reading from the stream?

magnus-madsen (commented):
All of this looks good. My comments are only pedantic nitpicks and one question.

main/src/library/Process.flix
CadeMichael: I was just doing alphabetic but I can change it.

main/src/library/Process.flix
CadeMichael: I think you're right

```
Returns:
    the input stream connected to the normal output of the subprocess 
```

so I think it will just be Null but doesn't throw anything. I'll remove the catch, should I wrap in an Option?

magnus-madsen: When will it be null?

magnus-madsen (approved):

PR #10919, opened 2025-06-19
chore: rename Weeder2 to Weeder

mlutze:
There was a reason we didn't do this before, but I don't remember what it was, so I am trying again.


magnus-madsen:
Its so we can review it- and we dont have time for that now

mlutze:
I suspect that we will never have time for it.

Issue #10923, opened 2025-06-19
Make TcpSocket an instance of Reader and Writer


rywiese:
`TcpSocket` wraps a java [Socket](https://docs.oracle.com/javase/8/docs/api/java/net/Socket.html), which exposes both an `InputStream` and an `OutputStream`. These streams are already instances of `Reader` and `Writer` respectively. `TcpSocket` can delegate its read/write calls to these instances.

This does mean that each `read` call has to call `getInputStream` before invoking the underlying `read`, but I think we've decided this is a non-issue, as the underlying java implementation should be caching that input stream. So that call should be cheap/safe.

magnus-madsen:
When you create the TcpSocket you could make it into a 3-tuple where you have already extracted the two streams, right?

rywiese:
Yes we can go that route too

PR #10924, opened 2025-06-19
feat: make `TcpSocket` an instance of `Reader` and `Writer`

rywiese:
fixes https://github.com/flix/flix/issues/10923

### DIFF
--- a/main/src/library/TcpSocket.flix
+++ b/main/src/library/TcpSocket.flix
@@ -14,9 +14,40 @@
  *  limitations under the License.
  */
 
+import java.io.InputStream
+import java.io.OutputStream
 import java.net.Socket
 
 ///
 /// Represents a TCP socket
 ///
-enum TcpSocket(Socket)
+enum TcpSocket(Socket, InputStream, OutputStream)
+
+instance Reader[TcpSocket] {
+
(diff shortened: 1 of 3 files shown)


main/src/library/TcpSocket.flix
```diff
@@ -20,3 +20,44 @@ import java.net.Socket
 /// Represents a TCP socket
 ///
 enum TcpSocket(Socket)
```
magnus-madsen: How about making this have 3 arguments? The socket and its streams.

main/src/library/TcpSocket.flix
```diff
@@ -14,9 +14,52 @@
...
 /// Represents a TCP socket
 ///
-enum TcpSocket(Socket)
+enum TcpSocket(Socket, InputStream, OutputStream)
+
+instance Reader[TcpSocket] {
+
+    type Aef = IO
+
+    pub def read(
```
magnus-madsen: I think this can just be one line. Also I dont think the line breaks are needed. Keep it short and sweet.

(Same comment applies elsewhere)

magnus-madsen:
Excellent 🚀

Issue #10912, opened 2025-06-18
Generate pseudo/provenance constraints for effect sources


jaschdoc:
In `Expr.Do` in `ConstraintGen`:
```scala
val pvar = Type.freshVar(Kind.Effect)
c.sourceType(pvar, effTpe) // unify source *new method*
val resEff = Type.mkUnion(pvar :: op.spec.eff :: effs, loc)

case class Source(...) extends Provenance
```

For sinks `expectType` should be called, for other expressions they should just be equal.

@mlutze @gaganchandan

gaganchandan:
> Should also happen for `ApplySig` and `Op`? Someone remembers this?

I think it's for `ApplyDef`. But I'm not sure how to identify the concrete effect in that case.

mlutze:
For ApplyDef, you'll need to extract it from the signature. `defn.spec.eff` contains an effect you can use.

gaganchandan:
> For ApplyDef, you'll need to extract it from the signature. `defn.spec.eff` contains an effect you can use.

Currently, the following constraint is generated:
https://github.com/flix/flix/blob/d904fc4807f46ab7954f7e33396e40a0424e8d90/main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala#L102

`evar` is then returned as `resEff`.

I think I should create a new constraint:
```scala
val pvar = Type.freshVar(Kind.Eff, loc)
c.unifyType(pvar, defn.spec.eff, loc1)
```

Then replace `defn.spec.eff` in the union from the original constraint (the snippet linked above) with `pvar` and then create that constraint. Is that correct?

mlutze:
That sounds about right to me.

magnus-madsen:
Please make small PRs when you do. Makes it easier to review and understand.

PR #10931, opened 2025-06-20
feat: add-source-provenance

gaganchandan:
Adds the new `Source` provenance type as discussed in #10912.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/typer/TypeConstraint.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/typer/TypeConstraint.scala
@@ -114,6 +114,11 @@ object TypeConstraint {
     case class Match(tpe1: Type, tpe2: Type, loc: SourceLocation) extends Provenance
 
     /**
+      * The constraint indicates that the left effect is a source representing the concrete effect on the right.
+      */
+    case class Source(eff1: Type, eff2: Type, loc: SourceLocation) extends Provenance
+
+    /**
       * Indicates the conflict arose from a timeout
       */
     // TODO this is an abuse of provenance. We should instead have a separate "conflict reason" type.
--- a/main/src/ca/uwaterloo/flix/language/phase/typer/TypeConstraint.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/typer/TypeConstraint.scala
@@ -114,9 +114,9 @@ object TypeConstraint {
     case class Match(tpe1: Type, tpe2: Type, loc: SourceLocation) extends Provenance
 
     /**
-      * The constraint indicates that the left effect is a source representing the concrete effect on the right.
+      * The constraint indicates that the left effect is a variable representing the source effect on the right.
       */
-    case class Source(eff1: Type, eff2: Type, loc: SourceLocation) extends Provenance
+    case class Source(eff1: Type.Var, eff2: Type, loc: SourceLocation) extends Provenance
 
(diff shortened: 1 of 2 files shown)


main/src/ca/uwaterloo/flix/language/phase/typer/TypeContext.scala
```diff
@@ -193,6 +193,18 @@ class TypeContext {
...
   }
 
+  /** Generates a constraints representing the source of a concrete effect.
+   *
+   *  For an effect variable effVar and concrete effect concreteEff, generates:
+   * {{
+   *  effVar ~ concreteEff
+   * }}
+   */
+  def unifySource(effVar: Type, concreteEff: Type, loc: SourceLocation): Unit = {
```
mlutze: Is this guaranteed to be a concrete effect? It seems it might sometimes involve variables if we are not careful, right?

gaganchandan: Investigation after I opened the PR has shown that it can be a variable, or sometimes the union of a concrete effect and an associated effect.

main/src/ca/uwaterloo/flix/language/phase/typer/TypeConstraint.scala
```diff
@@ -113,6 +113,11 @@ object TypeConstraint {
       */
     case class Match(tpe1: Type, tpe2: Type, loc: SourceLocation) extends Provenance
 
+    /**
+      * The constraint indicates that the left effect is a source representing the concrete effect on the right.
+      */
+    case class Source(eff1: Type, eff2: Type, loc: SourceLocation) extends Provenance
```
magnus-madsen: I would change eff1 to Type.Var

main/src/ca/uwaterloo/flix/language/phase/typer/TypeContext.scala
```diff
@@ -193,6 +193,18 @@ class TypeContext {
     }
   }
 
+  /** Generates a constraints representing the source of a concrete effect.
```
magnus-madsen: Linebreak

main/src/ca/uwaterloo/flix/language/phase/typer/TypeContext.scala
```diff
@@ -193,6 +193,18 @@ class TypeContext {
...
+   *
+   *  For an effect variable effVar and concrete effect concreteEff, generates:
```
magnus-madsen: Indentation?

main/src/ca/uwaterloo/flix/language/phase/typer/TypeContext.scala
gaganchandan: Should I call it a rigid effect or something like that?

gaganchandan: Or do I restrict it to only have a concrete effect for now? The restriction can be enforced in `ConstraintGen`.

mlutze: I think you cannot restrict it, since it is part of the constraint system.

I would just call them "effVar" and "sourceEff" or something.

magnus-madsen:
Remember to mark conversations as resolved-- once they are resolved.

magnus-madsen (approved):

PR #10932, opened 2025-06-20
fix: match kind in expect type

### LINKED ISSUE #10910: `TypeContext.expectType` should match on kind
Solves the bug in `ConstraintGen` in the `Expr.Ascribe` case 
https://github.com/flix/flix/blob/b9cebcb22cd22e02aae40d79620975531901b3ac/main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala#L687

The function `expectType` should match on the kind of the expected and actual types and generate `Provenance.ExpectEffect` if their kinds are effects and use the `Provenance.ExpectType` if their kinds are types (which it currently assumes).
Different kinds should be an error?

https://github.com/flix/flix/blob/b9cebcb22cd22e02aae40d79620975531901b3ac/main/src/ca/uwaterloo/flix/language/phase/typer/TypeContext.scala#L169

gaganchandan:
Fixes #10910.


mlutze (approved):
Looks good. Can we also (in this or a follow-up) get rid of expectKind, since it is now redundant.

magnus-madsen (approved):

PR #10933, opened 2025-06-20
feat: add compiler

CasperDN:
The compiler is mostly a copy of the one from `Fixpoint/Phase/Compiler/flix`, changed to fit our data types.

### DIFF
--- a/main/src/library/Fixpoint3/Phase/Compiler.flix
+++ b/main/src/library/Fixpoint3/Phase/Compiler.flix
@@ -1,0 +1,578 @@
+/*
+ * Copyright 2021 Benjamin Dahse
+ * Copyright 2025 Casper Dalgaard Nielsen
+ *                Adam Yasser Tallouzi
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+mod Fixpoint3.Phase.Compiler {
+    use Fixpoint3.Ast.Ram.{arityOfNonLat, BoolExp, Facts, Predicates, RamId, RamProgram, RamStmt, RamTerm, RelOp, RelSym, RowVar}
+    use Fixpoint.Ast.Datalog.{BodyPredicate, BodyTerm, Constraint, Datalog, Fixity, HeadTerm, Polarity, VarSym}
+    use Fixpoint.Ast.Datalog.Datalog.{Datalog}
+    use Fixpoint.Ast.Datalog.Constraint.Constraint
+    use Fixpoint.Ast.Datalog.HeadPredicate.HeadAtom
+    use Fixpoint.Ast.Datalog.BodyPredicate.{BodyAtom, Functional, Guard0, Guard1, Guard2, Guard3, Guard4, Guard5}
+    use Fixpoint.Ast.Shared.{BoxedDenotation => Denotation}
+    use Fixpoint.Ast.Shared.Denotation.{Relational, Latticenal}
+    use Fixpoint.Ast.Shared.isRelational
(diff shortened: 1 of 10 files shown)


JonathanStarup:
> The compiler is mostly a copy of the one from Fixpoint/Phase/Compiler/flix

Can you point to the differences? just to ease review

CasperDN:
The function `isImpossible` and the removal of `Guard0`. The introduction of `Par` blocks. The function `constructMerges`. A global counter for assigning `RowVar`s passed between functions. `idbSet` passed to `compileRuleIncr` to avoid creating dead code. `compileBoolForLatticeenal`. `compileFacts`.

Perhaps it would be more correct to say that the structure is mostly the same...

main/src/library/Fixpoint3/Phase/Compiler.flix
```diff
@@ -0,0 +1,584 @@
...
+                    MutList.push(codeBlock, stmts)
+                } else match List.head(runInParallel) {
+                    case Some(v) => MutList.push(v, stmts)
+                    case None => ()
+                }
+            });
+            let stmt = MutList.toVector(stmts) |> RamStmt.Seq;
+            RamProgram.Program(stmt, compiledFacts, predicates, (Map#{}, Map#{}))
+        }
+        case _ => bug!("Datalog Boxing bug")
```
JonathanStarup: Would this not be a lowering bug?

main/src/library/Fixpoint3/Phase/Compiler.flix
```diff
@@ -0,0 +1,584 @@
...
+    }
+
+    ///
+    /// Collect all rules in `d` and assign them a number, deterministically.
+    ///
+    @Internal
+    pub def enumerateRules(d: Datalog): Vector[(Int32, Constraint)] = match d {
+        case Datalog(_, rules) => Vector.mapWithIndex(i -> r -> (i, r), rules)
+        case _ => unreachable!()
```
JonathanStarup: This is probably not you I'd guess, but it would be nice to have functions like this mention in the doc that `d` must be a `Datalog(_, _)`

main/src/library/Fixpoint3/Phase/Compiler.flix
```diff
@@ -0,0 +1,584 @@
...
+    }
+
+    ///
+    /// Functional relations should be treated as `Relational`, but they do not posses any `PredSym`, indicated
+    /// by assigning them -1.
+    ///
+    @Internal
+    pub def functionalRelSym(arity: Int32): RelSym = RelSym.Symbol(PredSym.PredSym("", -1i64), arity, Relational)
```
JonathanStarup, line 85: This looks a bit scary to me, can you add explanation why its okay for all of these to reuse the same "-1" sym

main/src/library/Fixpoint3/Phase/Compiler.flix
```diff
@@ -0,0 +1,584 @@
...
+                // Remove p(;bot) from the facts
+                if(v `leq` bot) {
+                    acc
+                } else {
+                    Vector.append(k, Vector#{v}) :: acc
+                }
+            , Nil, values)
+        }, db);
+
+        facts |> Vector.foldRight(constraint -> acc -> {
```
JonathanStarup: With these foldleft/foldRight calls, I find it easier to read with ||>. e.g. `m |> foldleft(f, zero)` becomes `(m, zero) ||> foldleft(f)`

main/src/library/Fixpoint3/Phase/Compiler.flix
CasperDN: It looks like it has at some point been copied from another file. The old compiler said 'Datalog normalization bug'

main/src/library/Fixpoint3/Phase/Compiler.flix
CasperDN: This is part of our provenance implementation, which we are not merging at the moment. I have removed it for now.

main/src/library/Fixpoint3/Phase/Compiler.flix
CasperDN: I have extended the description about its use. It it better now?

main/src/library/Fixpoint3/Phase/Compiler.flix
CasperDN: I agree. We were not aware of the `||>` operator...

magnus-madsen:
Is this PR still "live"?

JumpTwiice:
We are waiting for a final review

magnus-madsen:
Merged 🚀

PR #10934, opened 2025-06-21
error code draft

JonathanStarup:


main/src/ca/uwaterloo/flix/language/errors/TypeError.scala
```diff
@@ -41,6 +42,8 @@ object TypeError {
     * @param loc  the location where the error occured.
     */
   case class ConstructorNotFound(clazz: Class[?], tpes: List[Type], renv: RigidityEnv, loc: SourceLocation) extends TypeError {
+    def errorCode: ErrorCode = ErrorCode.ConstructorNotFound
```
JonathanStarup, line 45: Here it is easy to check - you just verify that the name makes sense

main/src/ca/uwaterloo/flix/language/errors/ErrorCode.scala
```diff
@@ -0,0 +1,29 @@
+package ca.uwaterloo.flix.language.errors
+
+sealed trait ErrorCode {
+  import ErrorCode.*
+  def id: Int = this match {
+    case ConstructorNotFound => 47261
```
JonathanStarup, line 6: Here it is easy to check, since they are just one long list. You can also freely change the implementation to look at at config file or whatever you want.

The job of finding unique ID's is completely contained in this method

magnus-madsen:
I prefer the Rust design where each error is its own EXYZ value. In any case, I decided this is not a priority for now.

Issue #10855, opened 2025-06-11
Rework `IpAddr`


magnus-madsen:
Rework `IpAddr` to the following types (in separate files and with their own modules as needed):

```
enum IpAddr {
case V4(Ipv4Addr)
case V6(Ipv6Addr)
}

enum Ipv4Addr {
  case Ipv4Addr(...)
}

same for Ipv6Addr
```

magnus-madsen:
We will want to implement Eq, Order, and our own ToString on these types.

magnus-madsen:
@CadeMichael Can you take a look at https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html and propose what constants and functions we would want? I don't want all that stuff, just the basics.

magnus-madsen:
We can do these improvements in small PRs

CadeMichael:
for **Ipv4** do these constants and methods seem like a good starting point?

- constants
    - bits (32)
    - broadcast (225, 225, 225, 225)
    - _don't need_ localhost (Inet gives us localhost)
- methods
    - `toBytes`
    - `fromBytes` (we have)
    - `toIpV6`
    - `toString`
    - `Eq`

**Ipv6**

+ constants
  * bits
+ methods
  * `toBytes`
  * `fromBytes`
  * `fromString`
  * `toString`
  * `toIpv4`
  * `Eq`

magnus-madsen:
I think we would still want localhost right?

Instead of fromString, we can implement the FromString trait for both types.

The rest look good.

magnus-madsen:
Perhaps bits is not necessary.

CadeMichael:
> I think we would still want localhost right?
> 
> Instead of fromString, we can implement the FromString trait for both types.
> 
> The rest look good.

sounds good.

I omitted localhost because if you pass "localhost" into the `getAddress` Inet function it returns the correct IP. I can add an explicit Flix function if that would make it more intuitive.

magnus-madsen:
I think an explicit function would be useful.

rywiese:
@CadeMichael btw, I added a `toByteArray` function to this file, so I would maybe suggest renaming `toBytes`/`fromBytes` to `toByteList`/`fromByteList` so that all byte conversion methods are explicit abt what collection they convert

magnus-madsen:
We should also consider not fixing ourselves to List. Have a look at Foldable and Collectable.

PR #10944, opened 2025-06-23
feat: add functionality to and refactor `IpAddr`

CadeMichael:
Addresses #10855.

First push separates `IpAddr` into two enums for v6 and v4. Next step is adding functionality.

### DIFF
--- a/main/src/library/IpAddr.flix
+++ b/main/src/library/IpAddr.flix
@@ -70,11 +70,11 @@ mod IpAddr {
     ///
     pub def fromString(s: String): Option[IpAddr] = try {
         match Ipv6Addr.fromString(s) {
-          case Some(v6) => Some(IpAddr.V6(v6))
-          case None => match Ipv4Addr.fromString(s) {
-            case Some(v4) => Some(IpAddr.V4(v4))
-            case None => None
-          }
+            case Some(v6) => Some(IpAddr.V6(v6))
+            case None => match Ipv4Addr.fromString(s) {
+              case Some(v4) => Some(IpAddr.V4(v4))
+              case None => None
+            }
         }
         } catch {
             case _ : UnknownHostException => None
--- a/main/src/library/IpAddr.flix
+++ b/main/src/library/IpAddr.flix
@@ -14,20 +14,26 @@
  *  limitations under the License.
  */
 
+
+
 ///
-/// Represents an IP address
+/// Represents an IP address.
 ///
 enum IpAddr with Eq, ToString {
-  case V4(Ipv4Addr),
-  case V6(Ipv6Addr)
+    case V4(Ipv4Addr),
+    case V6(Ipv6Addr)
+}
+
+instance FromString[IpAddr] {
+    pub def fromString(x: String): Option[IpAddr] = IpAddr.fromString(x)
 }
 
 mod IpAddr {
 
-    import java.net.UnknownHostException
-
     use Ipv4Addr.Ipv4Addr
     use Ipv6Addr.Ipv6Addr
+
+    import java.net.UnknownHostException
 
     ///
     /// Converts a `List[Int8]` into `IpAddr`
@@ -35,10 +41,10 @@ mod IpAddr {
     /// an Ipv4 is 4 bytes and Ipv6 is 16 bytes
     ///
     pub def fromBytes(l: List[Int8]): IpAddr = match l {
-        case b1 :: b2 :: b3 :: b4 :: Nil => {
+        case b1::b2::b3::b4::Nil => {
             IpAddr.V4(Ipv4Addr(b1, b2, b3, b4))
         }
-        case b1 :: b2 :: b3 :: b4 :: b5 :: b6 :: b7 :: b8 :: b9 :: b10 :: b11 :: b12 :: b13 :: b14 :: b15 :: b16 :: Nil => {
+        case b1::b2::b3::b4::b5::b6::b7::b8::b9::b10::b11::b12::b13::b14::b15::b16::Nil => {
             IpAddr.V6(Ipv6Addr(b1, b2, b3, b4, b5, b6, b7, b8, b9, b10, b11, b12, b13, b14, b15, b16))
         }
         case _ => unreachable!()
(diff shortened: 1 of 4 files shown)


CadeMichael:
I'm wondering about the `FromString` implementation because for `IpAddr` how will it know if it's a v4 or v6? I have `FromString` for both individually, but will need to add an implementation for `IpAddr` that can distinguish which version to return. Right now I see two approaches

1. Try one, because if it fails it returns `None` and if we get a `None` try the other (v4 should fail with 128 bits and v6 with 32)

2. Count the number of `Chars` in the string and based on the length determine which version to return.

main/src/library/Ipv4Addr.flix
```diff
@@ -0,0 +1,48 @@
...
+    import java.net.InetAddress
+    import java.net.UnknownHostException
+
+    use Ipv4Addr.Ipv4Addr
+
+    ///
+    /// Parse the string `s` as an Ipv4Addr.
+    /// A successful parse is wrapped with `Some(x)`, a parse failure is indicated by `None`.
+    ///
+    pub def fromString(s: String): Option[Ipv4Addr] = try {
```
magnus-madsen: I would put the unsafe block before try, so:

```
unsafe try {
...
}
```

main/src/library/Ipv4Addr.flix
```diff
@@ -0,0 +1,48 @@
...
+}
+
+mod Ipv4Addr {
+    import java.net.InetAddress
+    import java.net.UnknownHostException
+
+    use Ipv4Addr.Ipv4Addr
+
+    ///
+    /// Parse the string `s` as an Ipv4Addr.
```
magnus-madsen: linebreak after

magnus-madsen: I would rephrase the documentation to:

```
Attempts to parse the given String `s` as an `Ipv4Addr`.

Returns `Some(addr)` if the string was successfully parsed. Otherwise returns `None`.
```

main/src/library/Ipv6Addr.flix
```diff
@@ -0,0 +1,59 @@
...
+enum Ipv6Addr with Eq, ToString {
+  case Ipv6Addr( // 128 bits
+    Int8, Int8, Int8, Int8, // 32 bits
+    Int8, Int8, Int8, Int8, // 32 bits
+    Int8, Int8, Int8, Int8, // 32 bits
+    Int8, Int8, Int8, Int8  // 32 bits
+  )
+}
+
+mod Ipv6Addr {
```
magnus-madsen, line 33: same comments as I made for Ipv4Addr

main/src/library/Ipv4Addr.flix
```diff
@@ -0,0 +1,48 @@
...
+    /// A successful parse is wrapped with `Some(x)`, a parse failure is indicated by `None`.
+    ///
+    pub def fromString(s: String): Option[Ipv4Addr] = try {
+          let addr = unsafe InetAddress.getByName(s);
+          match unsafe Array.toList(addr.getAddress()) {
+            case b1::b2::b3::b4::Nil => Some(Ipv4Addr(b1, b2, b3, b4))
+            case _ => None
+          }
+        } catch {
+          case _ : UnknownHostException => None
```
magnus-madsen: Is IllegalArgumentException possible?

main/src/library/Ipv4Addr.flix
```diff
@@ -0,0 +1,48 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+///
+///Represents a V4 Ip Address.
+///
+enum Ipv4Addr with Eq, ToString { // 32 bits
+  case Ipv4Addr(Int8, Int8, Int8, Int8)
```
magnus-madsen: four spaces indentation (applies everywhere)

main/src/library/Ipv4Addr.flix
```diff
@@ -0,0 +1,49 @@
...
+            match unsafe Array.toList(addr.getAddress()) {
+              case b1::b2::b3::b4::Nil => Some(Ipv4Addr(b1, b2, b3, b4))
+              case _ => None
+            }
+        } catch {
+            case _ : UnknownHostException => None
+        }
+}
+
+instance FromString[Ipv4Addr] {
```
magnus-madsen: instances should be right after the enum declaration

main/src/library/Ipv4Addr.flix
CadeMichael: from the docs


Throws:
    [UnknownHostException](https://docs.oracle.com/javase/8/docs/api/java/net/UnknownHostException.html) - if no IP address for the host could be found, or if a scope_id was specified for a global IPv6 address.
    [SecurityException](https://docs.oracle.com/javase/8/docs/api/java/lang/SecurityException.html) - if a security manager exists and its checkConnect method doesn't allow the operation

main/src/library/IpAddr.flix
```diff
@@ -45,11 +57,28 @@ mod IpAddr {
...
         }
 
+    ///
+    /// Attempts to parse the given String `s` as an `IpAddr`.
+    ///
+    /// Returns `Some(addr)` if the string was successfully parsed. Otherwise returns `None`.
+    ///
+    pub def fromString(s: String): Option[IpAddr] = try {
+        match Ipv6Addr.fromString(s) {
+          case Some(v6) => Some(IpAddr.V6(v6))
```
magnus-madsen: Formatting is still off. It should be 4 spaces :)

main/src/library/Ipv4Addr.flix
```diff
@@ -0,0 +1,50 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+///
+///Represents a V4 Ip Address.
```
magnus-madsen: Missing space between /// and comment

magnus-madsen:
Looks good; there are just still some issues with formatting. (We use 4 space indent in flix files)

magnus-madsen (approved):

magnus-madsen:
What's next? Using these in Ping etc?

CadeMichael:
> What's next? Using these in Ping etc?

I think some extra module methods for Ipv4/6 such as converting between them if possible.

Issue #10207, opened 2025-03-27
Type errors for different types of same name


JonathanStarup:
Doesn not handle `Som.T ~ Oth.T` good - it says `T not T` ish

wizard7377:
I am interested in working on this issue

magnus-madsen:
This issue is a bit difficult to start with. It might be better to pick something easier.

magnus-madsen:
Unless perhaps @JonathanStarup can point to where the change has to be made?

JonathanStarup:
No I don't know - haven't looked into it

wizard7377:
On second thought probably a good idea 😅

wizard7377:
Ok, I'm going to try fixing this now

magnus-madsen:
OK; you probably need some example where it happens. I don't know if we have one ready?

wizard7377:
I have one already prepared, I'll share it and the corresponding error when I get home

wizard7377:
```

trait Collection[t] {
    type Elm
    pub def empty(): t
    pub def insert(x: Collection.Elm[t], c: t): t
    pub def toList(c: t): List[Collection.Elm[t]]
}

mod TestA {
    pub enum Term with ToString {
        case Lambda
    }
}
mod TestB {
    pub enum Term with ToString {
        case Product
    }
    pub enum ManyTerm with ToString{
        case Yeild(Term, ManyTerm)
        case Nil
    }
    instance Collection[ManyTerm] {
        type Elm = Term
        pub def empty(): ManyTerm = ManyTerm.Nil
        pub def insert(x: Term, xs: ManyTerm): ManyTerm = ManyTerm.Yeild(x, xs)
        pub def toList(input: ManyTerm): List[Term] = match input {
            case ManyTerm.Nil => List.empty()
            case ManyTerm.Yeild(x, xs) => x :: Collection.toList(xs)
        }
    }
}
def myFun(): (TestA.Term,TestB.Term) =
    // This will cause a compilation error because `Term` is defined in both TestA and TestB.
    let x: TestB.Term = TestA.Term.Lambda;
    let y: TestA.Term = TestB.Term.Product;
    (x,y)
def myFun2() : TestB.ManyTerm where Collection.Elm[TestB.ManyTerm] ~ TestA.Term =
    let x = Collection.empty();
    let y: TestB.Term = TestB.Term.Product;
    Collection.insert(y, x)
def main(): Unit \ IO =
    let x = myFun() ;
[...]

PR #10947, opened 2025-06-23
fix: qualify errors when types have same name

wizard7377:
responds to #10207


magnus-madsen:
Can you write some text to describe how this works? Thanks!

main/src/ca/uwaterloo/flix/language/fmt/SimpleType.scala
```diff
@@ -343,10 +343,48 @@ object SimpleType {
   case class NonPredFieldType(name: String, tpe: SimpleType) extends PredicateFieldType
 
   /**
-    * Creates a simple type from the well-kinded type `t`.
+    * From a given `Type` return a `Symbol` if it does indeed have a symbol
+    *
+    * @param t The type
+    */
+  private def getSymbol(t : Type): Option[Symbol] = t match {
```
magnus-madsen: I see two issues here:

- A type can have more than one symbol.
- You do not handle Type.Apply

mlutze: I would expect this function to be something like `TypeConstructor -> Option[QualifiedSym]`.

magnus-madsen:
I think there is too much string manipulation going on. Also too much extension of symbols.

Maybe we need to break this down into smaller pieces (and PRs):

1. We want a function to compute all the symbols in a type.
2. We want a function that computes the symbols that are ambiguous without prefixes.
3. We want to refactor the pretty print function to a list (set?) of ambiguous symbols. If a symbol is ambiguous we print its full name.

main/src/ca/uwaterloo/flix/language/fmt/SimpleType.scala
```diff
@@ -343,10 +343,46 @@ object SimpleType {
...
+      case x0 : QualifiedSym => List(x0)
+      case _ => List()
+    }} map { t1 =>
+      if (t0.qname == t1.qname && t0.qnamespace != t1.qnamespace) {
+        //If the names are the same, but the ''true'' full names are also diffent
+        t0.qnamespace zip t1.qnamespace indexWhere { case (x0, x1) => x0 != x1 } match {
+          case -1 => t0 //If the namespaces are (somehow) the same, return the original type (this should not happen)
+          case i => //Otherwise, return a new type with the namespace up to the first difference
+            val newNamespace = t0.qnamespace.take(i + 1)
+            AName.mkQName(newNamespace, t0.qname,t0.loc)
```
magnus-madsen: We try avoid this kind of string / list manipulation. We want to work a higher level.

magnus-madsen:
For (1), I think we want a `case class SymbolSet(enums: Set[Symbol.EnumSym], ...` and so on.

I suggest open a new PR that starts there.

wizard7377:
Ok, the `SymbolSet` logic is current in the (very WIP) #10985

magnus-madsen:
I think the next task should be:

We want a function that computes the symbols that are ambiguous without prefixes.

which can be a single function on SymbolSet.

I would request to add that in a separate PR.

wizard7377:
#11010

magnus-madsen:
The diff looks very messy. I would suggest the next step is to create a fresh PR that changes FormatType to take a `amb: SymbolSet` and then work from there.

wizard7377:
Yeah probably for the best

magnus-madsen:
I will close this and let you open a new PR

PR #10958, opened 2025-06-24
fix: CompilerPerf with zero runs

JonathanStarup:
This function returns one instantaneous run instead of no runs


magnus-madsen:
Is this to avoid div-by-zero or something?

JonathanStarup:
> Is this to avoid div-by-zero or something?

Yes, but it ruins the maximum throughput number since there is now instantanious runs

Issue #10854, opened 2025-06-11
Refactor `Ping` and `PingWithResult` to take an `IpAddr`


magnus-madsen:

magnus-madsen:
Fixed in master.

PR #10959, opened 2025-06-24
refactor: make `isReachable` take `IpAddr` instead of `String`

CadeMichael:
Addresses #10854.

example main function

```
def main(): Result[IoError, Unit] \ {Net, IO} = run {
        let host = IpAddr.fromString("localhost");
        match host {
            case Some(ip) => {
                let reachable = Ping.isReachable(ip, 100, TimeUnit.Milliseconds);
                println("${host} is ${reachable}")
            }
            case None => println("bad address")
        }
    } with Ping.runWithIO

Some(V4(Ipv4Addr(127, 0, 0, 1))) is true                                        
Ok(())
```

### DIFF
--- a/main/src/library/Ping.flix
+++ b/main/src/library/Ping.flix
@@ -18,7 +18,7 @@
 /// The effect used to determine if an address is reachable.
 ///
 eff Ping {
-    def isReachable(host: String, timeout: Int32, u: TimeUnit): Bool
+    def isReachable(host: IpAddr, timeout: Int32, u: TimeUnit): Bool
 }
 
 mod Ping {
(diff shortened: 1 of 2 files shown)


main/src/library/Ping.flix
```diff
@@ -43,8 +43,11 @@ mod Ping {
                 try {
                     let ms = TimeUnit.toMilliseconds(u);
                     let timeoutMs = ms * timeout;
-                    let addr = InetAddress.getByName(host);
-                    k(addr.isReachable(timeoutMs))
+                    region rc {
+                        let ipBytes = IpAddr.toByteArray(rc, host);
```
magnus-madsen, line 47: In the follow up PR this just becomes a vector.

magnus-madsen (approved):

PR #10960, opened 2025-06-24
fix: CompilerPerf

JonathanStarup:


main/src/ca/uwaterloo/flix/tools/CompilerPerf.scala
```diff
@@ -393,7 +391,8 @@ object CompilerPerf {
     */
   private def perfBaseLine(N: Int, o: Options): IndexedSeq[Run] = {
     // Note: The Flix object is created _for every iteration._
-    (0 until N).map { _ =>
+    (0 until N).map { i =>
+      println(s"Base Iteration ${(i+1).toString.padTo(3, ' ')} / $N")
```
JonathanStarup: I left these in by accident, but honestly is super nice to actually see progress. can I leave them in? @magnus-madsen

magnus-madsen: I think it breaks some scripts (?)

magnus-madsen: At least use stderr

JonathanStarup: Scripts surely use the json output right? It feels like printing to err would break things worse since its not an error

JonathanStarup: I can also just remove it again, up to you

magnus-madsen: If you change it to stderr we can keep it

PR #10962, opened 2025-06-24
refactor: Systematic build folder

JonathanStarup:
We have
```
@param output the optional output directory where to place JVM bytecode.
output: Option[Path]
```
as an option but
- Some places use this as the general output folder
- Some places always use `./build/`
- Some places use this if possible, but `./build/` as a fall back

This is not great and makes it hard to change the name / location of the build folder (like `.toml` files want to)

In this PR:
- The output path is always defined, no ad-hoc fall back paths
- The output path is always used
- There is one singular default for the output path in general `Options.Default.outputPath` and one singular default for project outputPath `Bootstrap.buildDirectoryRaw`. Easy to change.


main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -514,7 +535,7 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
     */
   def build(flix: Flix): Validation[CompilationResult, BootstrapError] = {
     // Configure a new Flix object.
-    val newOptions = flix.options.copy(output = Some(Bootstrap.getBuildDirectory(projectPath)))
+    val newOptions = flix.options.copy(outputJvm = true, outputPath = Bootstrap.getBuildDirectory(projectPath))
```
JonathanStarup, line 538: Don't know if we actually want `outputJvm = true` here, the previous code just didn't have another option.

main/src/ca/uwaterloo/flix/language/phase/HtmlDocumentor.scala
```diff
@@ -46,7 +46,7 @@ object HtmlDocumentor {
   /**
     * The directory where to write the ouput.
     */
-  val OutputDirectory: Path = Paths.get("./build/doc")
+  def OutputDirectory(implicit flix: Flix): Path = flix.options.outputPath.resolve("doc/")
```
JonathanStarup, line 49: Now if you generate docs for your flix project, it'll respect your output path

JonathanStarup:
@magnus-madsen @mlutze Oh i forgot this in the miscellaneous discussion

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -90,9 +90,9 @@ object Bootstrap {
       s"""*.fpkg
          |*.jar
          |.GITHUB_TOKEN
-         |artifact/
-         |build/
-         |lib/
+         |$artifactDirectoryRaw
```
mlutze, line 93: Why doesn't this use `getArtifactDirectory`?

mlutze (commented):
Where is `outputPath` and `outputJvm` actually set by the user?

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
JonathanStarup: since that returns a path

JonathanStarup:
> Where is `outputPath` and `outputJvm` actually set by the user?

Only through the toml. neither has a CmdOpts entrypoint. This is just the current state of things, unrelated to this PR

magnus-madsen (commented):
Why do this:

getArtifactDirectory
artifactDirectoryRaw

This always always causes breakages and pain later.

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
magnus-madsen: Just get the name from the path!

magnus-madsen:
Please don't merge before this has been cleaned up.

Also there must be a better PR title

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
JonathanStarup: it just makes the code worse to read. How do i get `artifacts/` out from a path, exactly that format? `getArtifactDirectory` uses the raw string, so there is only one source of truth

JonathanStarup: also note that `artifactDirectoryRaw` is private

magnus-madsen: Use `getFileName`?

JonathanStarup: That doesn't work if the path is changed to be nested. it should really be the path relative to the build folder.

my point is just that this just works - no unstable formatting needed

JonathanStarup:
@magnus-madsen you can look it over again. The main text explains the change

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
magnus-madsen: We do, otherwise you cannot do :build followed by :build-jar.

magnus-madsen:
🫡

PR #10966, opened 2025-06-25
refactor: avoid file loading on `new Flix`

JonathanStarup:
Currently `new Flix()` loads all the files of the standard library. In Flix language it has effect `IO + IoException` which seems bad.

Now the standard lib is read at every call to `check()`.

Locally the time of TestAll is unchanged


magnus-madsen:
The exception cannot happen right? Only if the JAR is corrupted? Similarly to if a `.class` file was missing?

magnus-madsen:
I suppose we should move this file list into a separate object somewhere.

JonathanStarup:
Yeah it doesn't happen unless you mess inside the compiler.

I just think the flix object should be as stupid as possible, loading the standard library doesn't belong here. Listing the standard library, perhaps.

magnus-madsen:
How about:

object Library {
// that map from the library loaded eagerly.
}

JonathanStarup:
just a static field? that seems a bit un-flix like

Issue #10961, opened 2025-06-24
Refactor `Ipv4Addr.fromString` and `Ipv6Addr.fromString` to parse the string


magnus-madsen:
We cannot use `InetAddress` because it also does a lookup.

CadeMichael:
got it so it'll split the string on `.` into a `List[String]` and then map `Int8.fromString` over the list and use that to create the `IpAddr`

magnus-madsen:
> got it so it'll split the string on `.` into a `List[String]` and then map `Int8.fromString` over the list and use that to create the `IpAddr`

Yes, for Ipv4. For Ipv6 I suppose something else is needed.

PR #10967, opened 2025-06-25
refactor: remove `Inet` usage from `Ipv4Addr`

CadeMichael:
partially addresses #10961. Getting a parse for `Ipv6Addr` in progress.

### DIFF
--- a/main/src/library/Ipv4Addr.flix
+++ b/main/src/library/Ipv4Addr.flix
@@ -32,29 +32,16 @@ mod Ipv4Addr {
 
     use Ipv4Addr.Ipv4Addr
 
-    def byteToInt8(b: Int32): Option[Int8] =
-        if (b > 255 or b < -255)
-            None
-        else
-        try {
-            if (b > 127)
-                let bs = Int32.toString(b-256);
-                Int8.fromString(bs)
-            else
-                let bs = Int32.toString(b);
-                Int8.fromString(bs)
-        } catch {
-            case _: NumberFormatException => None
-        }
-
     def parseInt8NoLeading0(b: String): Option[Int8] =
         if (String.length(b) > 1 and String.startsWith({prefix = "0"}, b))
             None
         else
             forM(
-                b32 <- Int32.fromString(b);
-                b8 <- byteToInt8(b32)
-            ) yield b8
+                b32 <- Int32.fromString(b); if b32 < 256 and b32 > -256;
+                b8 <- if (b32 > 127) Int32.tryToInt8(b32-256) else Int32.tryToInt8(b32)
+            ) yield {
+                b8
+            }
 
     ///
     /// Attempts to parse the given String `s` as an `Ipv4Addr`.
--- a/main/src/library/Ipv4Addr.flix
+++ b/main/src/library/Ipv4Addr.flix
@@ -28,8 +28,33 @@ instance FromString[Ipv4Addr] {
 mod Ipv4Addr {
     import java.net.InetAddress
     import java.net.UnknownHostException
+    import java.lang.NumberFormatException
 
     use Ipv4Addr.Ipv4Addr
+
+    def byteToInt8(b: Int32): Option[Int8] =
+        if (b > 255 or b < -255)
+            None
+        else
+        try {
+            if (b > 127)
+                let bs = Int32.toString(b-256);
+                Int8.fromString(bs)
+            else
+                let bs = Int32.toString(b);
+                Int8.fromString(bs)
+        } catch {
+            case _: NumberFormatException => None
+        }
+
+    def parseInt8NoLeading0(b: String): Option[Int8] =
+        if (String.length(b) > 1 and String.startsWith({prefix = "0"}, b))
+            None
+        else
+            forM(
+                b32 <- Int32.fromString(b);
+                b8 <- byteToInt8(b32)
+            ) yield b8
 
     ///
     /// Attempts to parse the given String `s` as an `Ipv4Addr`.
@@ -39,9 +64,11 @@ mod Ipv4Addr {
     pub def fromString(s: String): Option[Ipv4Addr] =
         let bytes = s
             |> String.split({regex = "[.]" })
-            |> List.traverse(Int8.fromString);
+            |> List.traverse(parseInt8NoLeading0);
         match bytes {
-            case Some(b1 :: b2 :: b3 :: b4 :: Nil) => Some(Ipv4Addr(b1, b2, b3, b4))
+            case Some(b1::b2::b3::b4::Nil) => {
+                Some(Ipv4Addr(b1, b2, b3, b4))
+            }
             case _ => None
         }
 
--- a/main/src/library/Ipv4Addr.flix
+++ b/main/src/library/Ipv4Addr.flix
@@ -37,14 +37,15 @@ mod Ipv4Addr {
     /// Returns `Some(addr)` if the string was successfully parsed. Otherwise returns `None`.
     ///
     pub def fromString(s: String): Option[Ipv4Addr] =
-        unsafe try {
-            let addr = InetAddress.getByName(s);
-            match Array.toList(addr.getAddress()) {
-              case b1::b2::b3::b4::Nil => Some(Ipv4Addr(b1, b2, b3, b4))
-              case _ => None
+        let bytes = s
+            |> String.split({regex = "[.]" })
+            |> List.traverse(b -> Int8.fromString(b));
+        match bytes {
+            case Some(b1::b2::b3::b4::Nil) => {
+                Some(Ipv4Addr(b1, b2, b3, b4))
             }
-        } catch {
-            case _ : UnknownHostException => None
+            case _ => None
         }
+
 }
 
--- a/main/src/library/Ipv4Addr.flix
+++ b/main/src/library/Ipv4Addr.flix
@@ -17,12 +17,28 @@
 ///
 /// Represents a V4 Ip Address.
 ///
-enum Ipv4Addr with Eq, ToString { // 32 bits
+enum Ipv4Addr with Eq { // 32 bits
     case Ipv4Addr(Int8, Int8, Int8, Int8)
 }
 
 instance FromString[Ipv4Addr] {
     pub def fromString(x: String): Option[Ipv4Addr] = Ipv4Addr.fromString(x)
+}
+
+instance ToString[Ipv4Addr] {
+    pub def toString(x: Ipv4Addr): String =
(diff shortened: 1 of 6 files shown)


magnus-madsen:
I think now would be a good time to start adding some unit tests. You can search for TestList.flix and put in a TestIpv4Addr

magnus-madsen:
ChatGPT can probably generate interesting correct and incorrect inputs.

main/src/library/Ipv4Addr.flix
```diff
@@ -37,14 +37,15 @@ mod Ipv4Addr {
...
     ///
     pub def fromString(s: String): Option[Ipv4Addr] =
-        unsafe try {
-            let addr = InetAddress.getByName(s);
-            match Array.toList(addr.getAddress()) {
-              case b1::b2::b3::b4::Nil => Some(Ipv4Addr(b1, b2, b3, b4))
-              case _ => None
+        let bytes = s
+            |> String.split({regex = "[.]" })
+            |> List.traverse(b -> Int8.fromString(b));
```
magnus-madsen: You dont need b -> Int8.fromString, You can just pass Int8.fromString

main/src/library/Ipv4Addr.flix
```diff
@@ -37,14 +37,15 @@ mod Ipv4Addr {
...
+        match bytes {
+            case Some(b1::b2::b3::b4::Nil) => {
```
magnus-madsen, line 66: Can be one line, no need for { linebreak etc.

magnus-madsen: Also add space before and after ::

CadeMichael:
> I think now would be a good time to start adding some unit tests. You can search for TestList.flix and put in a TestIpv4Addr

So a new file as opposed to adding tests to `TestList.flix` ?

magnus-madsen:
> > I think now would be a good time to start adding some unit tests. You can search for TestList.flix and put in a TestIpv4Addr
> 
> So a new file as opposed to adding tests to `TestList.flix` ?

Yes new file, I was just trying to point to the directory (without actually finding it)

magnus-madsen:
I think you accidentally merged in the ipv6 changes too.

CadeMichael:
> I think you accidentally merged in the ipv6 changes too.

I was thinking of changing both in one PR but I can do one at a time aswell.

CadeMichael:
> ChatGPT can probably generate interesting correct and incorrect inputs.

I can also test it against `Inet` to make sure they're giving the same outputs.

main/src/library/Ipv4Addr.flix
```diff
@@ -28,9 +28,34 @@ instance FromString[Ipv4Addr] {
 mod Ipv4Addr {
     import java.net.InetAddress
     import java.net.UnknownHostException
+    import java.lang.NumberFormatException
 
     use Ipv4Addr.Ipv4Addr
 
+    def byteToInt8(b: Int32): Option[Int8] =
```
magnus-madsen: I think we already have something like this in the Int8 or perhaps Int32 module. Something clamp

CadeMichael:
```
pub def clampToInt8(min: {min = Int8}, max: {max = Int8}, x: Int32): Int8 =
```
this doesn't return an option, is that an issue? shouldn't it fail with None if it is out of bounds?

CadeMichael:
```
Running 9 tests...                                                              

   FAIL  TestIpv4AddrDev.fromString0 (Assertion Error)
   PASS  TestIpv4AddrDev.fromString1 76.6us
   PASS  TestIpv4AddrDev.fromString3 54.3us
   FAIL  TestIpv4AddrDev.fromString4 (Assertion Error)
   FAIL  TestIpv4AddrDev.fromString5 (Assertion Error)
   PASS  TestIpv4AddrDev.fromString6 50.7us
   FAIL  TestIpv4AddrDev.fromString7 (Assertion Error)
   FAIL  TestIpv4AddrDev.fromString8 (Assertion Error)
   PASS  TestIpv4AddrDev.fromString9 49.3us
```

it breaks some of the tests

CadeMichael: putting bounds in the yield makes a few more pass but I think it's how Inet handles ip addresses. I had to write the custom function to have IpAddr represent Ip's with the same byte values that Inet uses so Ping and Dns would work as they use Inet under the hood.

magnus-madsen: What about the tryToInt8 functions?

CadeMichael: the issue is that Ip's are represented as u8's but Inet uses i8, so it wraps around 128 instead of going up to 256, so I had to take the u8 to a i8.

main/test/ca/uwaterloo/flix/library/TestIpv4Addr.flix
```diff
@@ -0,0 +1,72 @@
...
+
+    @Test
+    def fromString4(): Bool =
+        Ipv4Addr.fromString("192.168.1.300") |> Option.isEmpty
+
+    @Test
+    def fromString5(): Bool =
+        let parseIp = Ipv4Addr.fromString("192.168.1.1");
+        let ip = Some(Ipv4Addr(-64i8, -88i8, 1i8, 1i8));
+        parseIp == ip
```
magnus-madsen: These tests are a bit hard to read. How about we implement ToString on the Ipv4Addr and then we can do 

let s = ""192.168.1.1""
Ipv4Addr.fromString(s) == "Some(${s})"

magnus-madsen: Or something like that

main/src/library/Ipv4Addr.flix
magnus-madsen: Can we not subtract and then use tryToInt8?

main/test/ca/uwaterloo/flix/library/TestIpv4Addr.flix
CadeMichael: so the `toString` should represent the Ip as u8 instead of the Inet i8? I'll add that because right now the `toString` for that Ip would look like

`"Ipv4AddrDev(-64, -88, 1, 1)"`

magnus-madsen: > so the `toString` should represent the Ip as u8 instead of the Inet i8? I'll add that because right now the `toString` for that Ip would look like
> 
> `"Ipv4AddrDev(-64, -88, 1, 1)"`

Indeed. The above string format is not very readable.

CadeMichael: I got this working but I converted the IpAddr toByteArray -> toByteVec,

this breaks DNS I can fix for this PR or make the toString function use the old Array Implementation

```
instance ToString[Ipv4AddrDev] {
    pub def toString(x: Ipv4AddrDev): String =
        let u8Bytes = IpAddrDev.V4(x)
            |> IpAddrDev.toByteVec
            |> Vector.map(b -> {
                if (b < 0i8)
                    Int8.toInt32(b) + 256
                else
                    Int8.toInt32(b)
            });
        match Vector.toList(u8Bytes) {
            case b1 :: b2 :: b3 :: b4 :: Nil => "${b1}.${b2}.${b3}.${b4}"
            case _ => ""
        }
}
```

magnus-madsen: You have the four bytes already, right?

CadeMichael:
It's failing for files I don't have 

`TcpConnect.flix` which it says uses the function I changed `toByteArray`

magnus-madsen:
I think some branches got tangled again.

CadeMichael:
> I think some branches got tangled again.

I put the `toByteArray` back in maybe it is good to have both?

magnus-madsen:
Look at the diff on GitHub. Lets try to focus on one change at a time-- here to avoid the use of InetAddress and to add some tests for Ipv4

magnus-madsen:
We can do the subsequent changes in followup PRs

CadeMichael:
> Look at the diff on GitHub. Lets try to focus on one change at a time-- here to avoid the use of InetAddress and to add some tests for Ipv4

- I can undo the changes to `Tcp`
- The issue with removing `toByteVec` is the toString desired I added the `toByteVec` so the toString won't use the old `toByteArray` function. If it makes sense to have both or remove `toByteArray` later I can push both, if I remove `toByteVec` then I'll have to change the `toString` later when we take out `toByteArray`

main/src/library/Ipv4Addr.flix
```diff
@@ -17,34 +17,63 @@
...
+enum Ipv4Addr with Eq { // 32 bits
     case Ipv4Addr(Int8, Int8, Int8, Int8)
 }
 
 instance FromString[Ipv4Addr] {
     pub def fromString(x: String): Option[Ipv4Addr] = Ipv4Addr.fromString(x)
 }
 
+instance ToString[Ipv4Addr] {
+    pub def toString(x: Ipv4Addr): String =
```
magnus-madsen, line 29: I was thinking the implementation would just be

```
match x {
    case Ipv4Addr(b1, b2, b3, b4) => "${Int8.toInt32(b1) + 128}.${Int8.toInt32(b2) + 128}.${Int8.toInt32(b3) + 128}.${Int8.toInt32(b4) + 128}"
}
```

But maybe I am missing something?

CadeMichael: it's more for this function, but I can use a region with the `toBytesArray` method

```
instance ToString[Ipv4AddrDev] {
    pub def toString(x: Ipv4AddrDev): String =
        let u8Bytes = IpAddrDev.V4(x)
            |> IpAddrDev.toByteVec /// here
            |> Vector.map(b -> {
                if (b < 0i8)
                    Int8.toInt32(b) + 256
                else
                    Int8.toInt32(b)
            });
        match Vector.toList(u8Bytes) {
            case b1 :: b2 :: b3 :: b4 :: Nil => "${b1}.${b2}.${b3}.${b4}"
            case _ => ""
        }
}
```

CadeMichael: the thing is not every byte get's 128 added, only negative ones, but I can have a lambda do that

magnus-madsen:
Thanks!

magnus-madsen:
@CadeMichael I was taking a look at the code again. 

I was thinking that u8Conv can be simplified to simply ANDing with 0xff. 

As for, parseInt8NoLeading0, I wonder if it can also not be simplified.

What do you think?

PR #10972, opened 2025-06-26
feat: generate pseudo constraints with Source provenance

gaganchandan:
Generate pseudo constraints as discussed in #10912.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
@@ -89,7 +89,7 @@ object ConstraintGen {
       case Expr.ApplyDef(DefSymUse(sym, loc1), exps, itvar, tvar, evar, loc2) =>
         val defn = root.defs(sym)
 
-        // Psuedo variable for source to flow into
+        // Pseudo variable for source to flow into
         val pvar = Type.freshVar(Kind.Eff, loc1)
 
         val (tconstrs1, econstrs1, declaredType, _) = Scheme.instantiate(defn.spec.sc, loc1.asSynthetic)
@@ -816,7 +816,7 @@ object ConstraintGen {
         val op = lookupOp(symUse.sym, symUse.loc)
         val effTpe = Type.Cst(TypeConstructor.Effect(symUse.sym.eff), loc)
 
-        // Psuedo variable for source to flow into
+        // Pseudo variable for source to flow into
         val pvar = Type.freshVar(Kind.Eff, loc)
 
         // length check done in Resolver


main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -88,18 +88,41 @@ object ConstraintGen {
...
         val constrs1 = tconstrs1.map(_.copy(loc = loc2))
         val declaredEff = declaredType.arrowEffectType
         val declaredArgumentTypes = declaredType.arrowArgTypes
         val declaredResultType = declaredType.arrowResultType
         val (tpes, effs) = exps.map(visitExp).unzip
+
+        // Substitute the source effect with the psueo variable.
+        var sourceFound = false
+        val substitutedEffs =
+          (declaredEff :: effs).map {
```
mlutze: I don't think we need to go over the effs here, because those come from nested subexpressions. Instead we should only consider the declared effect.

mlutze: We'll still need to unify `effs` at the end, but it shouldn't be part of the source.

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -810,15 +833,19 @@ object ConstraintGen {
...
         val effs = visitOpArgs(op, exps)
 
         // specialize the return type of the op if needed
         val opTpe = getDoType(op)
 
         c.unifyType(opTpe, tvar, loc)
+        c.unifySource(pvar, effTpe, loc)
         val resTpe = tvar
-        val resEff = Type.mkUnion(effTpe :: op.spec.eff :: effs, loc)
+        val resEff = Type.mkUnion(pvar :: op.spec.eff :: effs, loc)
```
mlutze: We can get rid of `op.spec.eff` in this line, because it's the same as `effTpe`

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
mlutze: Ultimately this should look very similar to what you did for `Do` below.

gaganchandan: Fixed in latest commit.

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
gaganchandan: Fixed in latest commit.

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -88,18 +88,31 @@ object ConstraintGen {
...
+
         c.unifyType(itvar, declaredType, loc2)
         c.expectTypeArguments(sym, declaredArgumentTypes, tpes, exps.map(_.loc))
         c.addClassConstraints(constrs1, loc2)
         c.addEqualityConstraints(econstrs1, loc2)
         c.unifyType(tvar, declaredResultType, loc2)
-        c.unifyType(evar, Type.mkUnion(declaredEff :: effs, loc2), loc2)
+
+        // If source effect was found, generate a source constraint
+        if(defn.spec.eff == declaredEff) {
```
mlutze: Can you explain why this check is needed?

gaganchandan: Ah, I think that's just left over from the original commit where there was a comparison. I believe we don't need `defn.spec.eff` at all can just use `declaredEff`.

They seem to differ in the case of effect handlers (different effect variables).

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -88,18 +88,24 @@ object ConstraintGen {
 
       case Expr.ApplyDef(DefSymUse(sym, loc1), exps, itvar, tvar, evar, loc2) =>
         val defn = root.defs(sym)
+
+        // Psuedo variable for source to flow into
```
mlutze: psuedo -> pseudo :)

mlutze (approved):

PR #10975, opened 2025-06-26
fix: correctly load primitive package effects

jaschdoc:
Before, it returned `Map(null -> effset)`, just choosing the last occuring effect set in the `json` file.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.scala
@@ -139,7 +139,7 @@ object PrimitiveEffects {
     val m = json \\ "packages" match {
       case JObject(l) => l.map {
         case (packageName, JString(s)) =>
-          val pkg = ClassLoader.getPlatformClassLoader.getDefinedPackage(packageName)
+          val pkg = Package.getPackages.filter(p => p.getName == packageName).head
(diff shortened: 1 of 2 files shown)


magnus-madsen:
Sorry  what? I am totally confused.

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.scala
```diff
@@ -139,9 +139,9 @@ object PrimitiveEffects {
     val m = json \\ "packages" match {
       case JObject(l) => l.map {
         case (packageName, JString(s)) =>
-          val clazz = ClassLoader.getPlatformClassLoader.getDefinedPackage(packageName)
+          val pkg = Package.getPackages.filter(p => p.getName == packageName).head
```
magnus-madsen, line 142: Does .head always exist? What is Package.getPackages?

jaschdoc: https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Package.html#getPackages()

jaschdoc: Head should always exist unless we wrote some garbage in the `json` file

magnus-madsen (approved):

PR #10986, opened 2025-06-28
refactor: Ipv4Addr to use `bitwiseAnd`

CadeMichael:
partially addresses #10961. using `bitwiseAnd` to get u8 -> i8. I don't think it will work for the parse function because going the otherway requires really complicated bit operations.

### DIFF
--- a/main/src/library/Ipv4Addr.flix
+++ b/main/src/library/Ipv4Addr.flix
@@ -36,18 +36,18 @@ mod Ipv4Addr {
 
     use Ipv4Addr.Ipv4Addr
 
-    def toUnsigned8(x: Int8): Int32 = Int32.bitwiseAnd(Int8.toInt32(x), 0xff)
+    def toUnsigned8(x: Int8): Int32 = Int32.bitwiseAnd(Int8.toInt32(x), 0xFF)
 
     def toSigned8(x: Int32): Option[Int8] =
         forM(
-            b32 <- Some(x); if x < 256 and x > -256;
-            b8 <- Int32.tryToInt8(
+            bu8 <- Some(x); if 0 <= x and x < 256;
+            bi8 <- Int32.tryToInt8(
                     Int32.bitwiseXor(
-                        Int32.bitwiseAnd(b32, 0xFF),
+                        Int32.bitwiseAnd(bu8, 0xFF),
                         0x80
                     ) - 0x80
                 )
-        ) yield b8
+        ) yield bi8
 
     def parseInt8NoLeading0(b: String): Option[Int8] =
         forM(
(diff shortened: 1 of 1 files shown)


main/src/library/Ipv4Addr.flix
```diff
@@ -42,16 +36,22 @@ mod Ipv4Addr {
 
     use Ipv4Addr.Ipv4Addr
 
+    pub def toString(x: Ipv4Addr): String =
+        let u8Conv = b -> Int32.bitwiseAnd(Int8.toInt32(b), 0xff);
```
magnus-madsen: Lets do:

u8Conv -> toUnsigned8

main/src/library/Ipv4Addr.flix
```diff
@@ -42,16 +36,22 @@ mod Ipv4Addr {
...
+        match x {
```
magnus-madsen: This match can just be a single let-pattern

main/src/library/Ipv4Addr.flix
```diff
@@ -42,16 +36,22 @@ mod Ipv4Addr {
...
+            case Ipv4Addr.Ipv4Addr(b1, b2, b3, b4) =>
+                "${u8Conv(b1)}.${u8Conv(b2)}.${u8Conv(b3)}.${u8Conv(b4)}"
+            case _ => ""
```
magnus-madsen: This is unreachable, so I am surprised it is even allowed.

main/src/library/Ipv4Addr.flix
```diff
@@ -42,16 +36,22 @@ mod Ipv4Addr {
 
     use Ipv4Addr.Ipv4Addr
 
+    pub def toString(x: Ipv4Addr): String =
```
magnus-madsen: I would move the toString function after the fromString function

magnus-madsen:
I added some detailed comments.

I think it must be possible to simplify parseInt8NoLeading0. It currently has lots of branches and monadic style. It should be fun to simplify it.

CadeMichael:
> I added some detailed comments.
> 
> I think it must be possible to simplify parseInt8NoLeading0. It currently has lots of branches and monadic style. It should be fun to simplify it.

I'm not sure, everything I see when searching for u8->i8 suggests the subtraction solution. I can't find anything with regards to bitwise operations taking u8->i8.
To be fair I am not very familiar with bitwise operations, but there doesn't seem to be any examples of people doing it that way whereas there are a multitude of simple subtraction solutions.
I think the monadic style is a result of the `tryToInt8` and `Int32.fromString` which is unavoidable. if there is a bitwise solution it will most likely be a chain of bit operations and then a `tryToInt8` which will avoid the `forM` style but add logical complexity.

magnus-madsen:
Remember to push the update

magnus-madsen (approved):

PR #10987, opened 2025-06-28
refactor: Ipv6 `toString`, `fromString` w/o Inet, Tests, and bitwise

CadeMichael:
Addresses addresses #10961.

### DIFF
--- a/main/src/library/Ipv6Addr.flix
+++ b/main/src/library/Ipv6Addr.flix
@@ -15,9 +15,9 @@
  */
 
 ///
-/// Represents a V6 Ip Address.
+///Represents a V6 Ip Address.
 ///
-enum Ipv6Addr with Eq, ToString {
+enum Ipv6Addr with Eq {
     case Ipv6Addr( // 128 bits
       Int8, Int8, Int8, Int8, // 32 bits
       Int8, Int8, Int8, Int8, // 32 bits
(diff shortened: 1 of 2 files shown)


main/src/library/Ipv6Addr.flix
```diff
@@ -15,9 +15,9 @@
  */
 
 ///
-/// Represents a V6 Ip Address.
+///Represents a V6 Ip Address.
```
magnus-madsen: Why remove the space?

main/src/library/Ipv6Addr.flix
```diff
@@ -30,33 +30,124 @@ instance FromString[Ipv6Addr] {
...
+
 mod Ipv6Addr {
+    import java.lang.Integer
+    import java.lang.NumberFormatException
     import java.net.InetAddress
     import java.net.UnknownHostException
 
     use Ipv6Addr.Ipv6Addr
 
+    def toUnsigned8(x: Int8): Int32 = Int32.bitwiseAnd(Int8.toInt32(x), 0xFF)
```
magnus-madsen, line 72: I would move the private helper functions last in the module, so the public functions come first.

magnus-madsen: We also need to document each function

main/src/library/Ipv6Addr.flix
```diff
@@ -30,33 +30,124 @@ instance FromString[Ipv6Addr] {
...
+
+    def byteToHex(b1: Int8, b2: Int8): String = {
+        let combined = (toUnsigned8(b1) * 256) + toUnsigned8(b2);
+        let hexString = String.toUpperCase(Integer.toString(combined, 16));
```
magnus-madsen, line 79: this can be inlined?

main/src/library/Ipv6Addr.flix
```diff
@@ -30,33 +30,124 @@ instance FromString[Ipv6Addr] {
...
+    import java.lang.Integer
+    import java.lang.NumberFormatException
     import java.net.InetAddress
     import java.net.UnknownHostException
 
     use Ipv6Addr.Ipv6Addr
 
+    def toUnsigned8(x: Int8): Int32 = Int32.bitwiseAnd(Int8.toInt32(x), 0xFF)
+
+    def byteToHex(b1: Int8, b2: Int8): String = {
```
magnus-madsen, line 77: We dont have something like this in Int8?

magnus-madsen:
What's next? SocketAddr etc?

main/src/library/Ipv6Addr.flix
CadeMichael: `String.toUpperCase(Integer.toString((toUnsigned8(b1) * 256) + toUnsigned8(b2)), 16)`

like this?

main/src/library/Ipv6Addr.flix
CadeMichael: I don't see any hex functions in Int8

main/src/library/Ipv6Addr.flix
magnus-madsen: Nevermind, I misread the code. Just leave it as is.

CadeMichael:
> What's next? SocketAddr etc?

I believe so. I have two open IpAddr PR's. I will add one more for "localhost" and Eq, should be very small.

PR #10998, opened 2025-06-30
chore: rename Do to ApplyOp

mlutze:
The distinction between ops and defs/sigs is become less important. This change makes things more regular.


magnus-madsen:
I agree with the rename
I also agree with moving ApplyOp up
But you should also do it in the phases then.

PR #10999, opened 2025-06-30
refactor: improve type errors for record types

gaganchandan:
Adds undefined label and extra label errors as discussed in #10722. 

A couple of comments:
i) Currently, it produces the same old error in the case of open records. 

ii) It seems difficult to fit the code into the design of the `toTypeErrors` function creating a list from a `TypeError` since there can now be multiple error messages generated from the same constraint. So `mkErrorsFromRecords` returns the list.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/errors/TypeError.scala
+++ b/main/src/ca/uwaterloo/flix/language/errors/TypeError.scala
@@ -412,6 +412,33 @@ object TypeError {
   }
 
   /**
+    * Extra label error.
+    *
+    * @param label      the name of the extra label.
+    * @param labelType  the type of the extra label.
+    * @param recordType the record type where the label is missing.
+    * @param renv       the rigidity environment.
+    * @param loc        the location where the error occurred.
+    */
+  case class ExtraLabel(label: Name.Label, labelType: Type, recordType: Type, renv: RigidityEnv, loc: SourceLocation)(implicit flix: Flix) extends TypeError {
+    def summary: String = s"Extra label '$label' of type '$labelType'."
+
+    def message(formatter: Formatter): String = {
+      import formatter.*
+      s""">> Extra label '${red(label.name)}' of type '${cyan(formatType(labelType, Some(renv)))}'.
+         |
+         |${code(loc, "extra label.")}
+         |
+         |The record type:
+         |
+         |  ${formatType(recordType, Some(renv))}
+         |
+         |contains the extra label '${red(label.name)}' of type ${cyan(formatType(labelType, Some(renv)))}.
+         |""".stripMargin
+    }
+  }
+
+  /**
     * Undefined predicate error.
     *
     * @param pred       the missing predicate.
--- a/main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolverInterface.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolverInterface.scala
@@ -19,12 +19,13 @@ import ca.uwaterloo.flix.api.Flix
 import ca.uwaterloo.flix.language.ast.Type.JvmMember
 import ca.uwaterloo.flix.language.ast.shared.SymUse.{AssocTypeSymUse, TraitSymUse}
 import ca.uwaterloo.flix.language.ast.shared.{AssocTypeDef, EqualityConstraint, Scope, TraitConstraint}
-import ca.uwaterloo.flix.language.ast.{KindedAst, RigidityEnv, SourceLocation, Symbol, Type, TypeConstructor, Kind}
+import ca.uwaterloo.flix.language.ast.{KindedAst, Name, RigidityEnv, SourceLocation, Symbol, Type, TypeConstructor, Kind}
 import ca.uwaterloo.flix.language.errors.TypeError
 import ca.uwaterloo.flix.language.phase.typer.TypeConstraint.Provenance
 import ca.uwaterloo.flix.language.phase.unification.{EqualityEnv, Substitution, TraitEnv}
 import ca.uwaterloo.flix.util.collection.ListMap
 import ca.uwaterloo.flix.language.phase.util.PredefinedTraits
+
 
 import scala.annotation.tailrec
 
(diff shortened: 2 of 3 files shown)


magnus-madsen:
We probably want a bunch of extra test cases in TestTyper

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolverInterface.scala
```diff
@@ -125,7 +126,12 @@ object ConstraintSolverInterface {
       List(TypeError.UnexpectedArg(sym, num, expected = subst(expected), actual = subst(actual), renv, loc))
 
     case TypeConstraint.Equality(tpe1, tpe2, Provenance.Match(baseTpe1, baseTpe2, loc)) =>
```
mlutze: Since we're looking at Provenance.Match here, there's not a reference point to say that something is "missing" or "extra", since there's not an "expected" type.

I think we should look for Provenance.Expect instead.

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolverInterface.scala
```diff
@@ -192,6 +198,54 @@ object ConstraintSolverInterface {
...
+  /**
+    * Creates appropriate errors for the mismatched record types.
+    */
+  private def mkErrorsFromRecords(tpe1: Type, tpe2: Type, renv: RigidityEnv, loc: SourceLocation)(implicit flix: Flix): List[TypeError] = {
+    // TODO: Open records
+    if (tpe1.typeVars.nonEmpty || tpe2.typeVars.nonEmpty) {
+      return List(mkMismatchedTypesOrEffects(tpe1, tpe2, tpe1, tpe2, renv, loc))
+    } else {
+      val tpe1Labels = getRecordLabels(tpe1, Set.empty[(Name.Label, Type)])
+      val tpe2Labels = getRecordLabels(tpe2, Set.empty[(Name.Label, Type)])
```
mlutze: It's dangerous (unpredictable behavior) to put types in sets.

gaganchandan:
> Based on the way the unification of records works, I would actually expect _everything_ remaining after unification to be extra or missing. (Not sure though.)

Everything as in every label on either side of the constraint?

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolverInterface.scala
gaganchandan: Yes, you're right. I'll fix it.

`Provenance.ExpectArgument` as well?

mlutze:
> > Based on the way the unification of records works, I would actually expect _everything_ remaining after unification to be extra or missing. (Not sure though.)
> 
> Everything as in every label on either side of the constraint?

Yes

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolverInterface.scala
mlutze: > Yes, you're right. I'll fix it.
> 
> `Provenance.ExpectArgument` as well?

Yes

main/src/ca/uwaterloo/flix/language/errors/TypeError.scala
```diff
@@ -505,39 +532,39 @@ object TypeError {
   }
 
   /**
-   * Unresolved constructor type error.
-   * This is a dummy error used in Java constructor type reconstruction for InvokeConstructor.
-   */
+    * Unresolved constructor type error.
```
magnus-madsen: Please avoid unrelated changes. If reformatting is required, you can make a separate PR.

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolverInterface.scala
magnus-madsen: But you are allowed to put Name.Label in a set, if that helps.

magnus-madsen: But maybe sets are not what you want, since we support scoped labels.

magnus-madsen:
I think a good next step-- before  you iterate on the implementation-- would be to add some tests. I would expect 3-5 tests per new TypeError. You will want some with duplicate labels.

gaganchandan:
> > > Based on the way the unification of records works, I would actually expect _everything_ remaining after unification to be extra or missing. (Not sure though.)
> > 
> > 
> > Everything as in every label on either side of the constraint?
> 
> Yes

It's not always the case. Consider the following program:
```
def undefinedLabel7(): Unit  = {
    let r: {x = Int32, y = Int32} = {y = 1, z = 3};
    ()
}
```
The unsolved constraint is something like `{ x = Int32, y = Int32 } ~ { a = Int32, y = Int32 }`. Here we need to filter out `y`.

main/src/ca/uwaterloo/flix/language/errors/TypeError.scala
gaganchandan: Sorry about that, I ran the IDEA formatter and it did those.

mlutze:
Thanks for the example; this is a good point.

What happens for this program? I just moved the `y` so it's first.

```
def undefinedLabel8(): Unit  = {
    let r: {y = Int32, x = Int32} = {y = 1, z = 3};
    ()
}
```

gaganchandan:
> Thanks for the example; this is a good point.
> 
> What happens for this program? I just moved the `y` so it's first.
> 
> ```
> def undefinedLabel8(): Unit  = {
>     let r: {y = Int32, x = Int32} = {y = 1, z = 3};
>     ()
> }
> ```

The constraint is `{ x = Int32 } ~ { z = Int32 }`

mlutze:
> > Thanks for the example; this is a good point.
> > What happens for this program? I just moved the `y` so it's first.
> > ```
> > def undefinedLabel8(): Unit  = {
> >     let r: {y = Int32, x = Int32} = {y = 1, z = 3};
> >     ()
> > }
> > ```
> 
> The constraint is `{ x = Int32 } ~ { z = Int32 }`

Ok, makes sense.

gaganchandan:
> I think a good next step-- before you iterate on the implementation-- would be to add some tests. I would expect 3-5 tests per new TypeError. You will want some with duplicate labels.

How shall I create tests for programs with both undefined and extra labels? Is there a way to pass multiple errors to `expectError`? If not, should I make two test cases for each such program where one has `expectError(TypeError.UndefinedLabel)` and the other has `expectError(TypeError.ExtraLabel)` or simply use the generic `expectError(TypeError)`?

mlutze:
> > I think a good next step-- before you iterate on the implementation-- would be to add some tests. I would expect 3-5 tests per new TypeError. You will want some with duplicate labels.
> 
> How shall I create tests for programs with both undefined and extra labels? Is there a way to pass multiple errors to `expectError`? If not, should I make two test cases for each such program where one has `expectError(TypeError.UndefinedLabel)` and the other has `expectError(TypeError.ExtraLabel)` or simply use the generic `expectError(TypeError)`?

You can repeat the expectError on the result.

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolverInterface.scala
```diff
@@ -192,6 +205,59 @@ object ConstraintSolverInterface {
...
+              } else Nil
+            }
+          }
+        } else if (tpe2Labels.isEmpty && !tpe1Labels.isEmpty) {
+          tpe1Labels.toList.map {
+            case (label, labelTpe) =>
+              TypeError.UndefinedLabel(label, labelTpe, tpe2, renv, loc)
+          }
+          // Both undefined and extra labels.
+        } else {
```
mlutze, line 239: I think we only need this else-branch, right? Because this covers the empty cases.

gaganchandan: We'll still need the branch that checks subsets, but the other two are covered.

magnus-madsen:
How are we dealing with duplicate labels?

magnus-madsen:
@mlutze Let me know when I should review

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolverInterface.scala
mlutze: I'm still confused.

I would expect it to be essentially:
- `extraLabels = tpe1Labels - tpe2Labels`
- `missingLabels = tpe2Labels - tpe1Labels`

Which is what you have below. All the empty/subset checking is handled by this, no?

magnus-madsen:
@mlutze Hi. How about be merge this and then iterate? (If you agree, you can merge).

For next iteration: Cleanup the set logic and deal with duplicate labels. Matt can explain.

magnus-madsen:
@gaganchandan Can you connect with Matt when you are both in the office? Thanks!

magnus-madsen:
@mlutze I suggest you create a ticket together.

PR #11001, opened 2025-06-30
feat: add join profiler

JumpTwiice:
The join profiler estimates the sizes of relations and joins in Datalog programs. The estimates are used to optimize join orders in rules and are necessary since relations are dynamic (their sizes increase) throughout fixpoint computation, meaning static analysis is insufficient to determine optimal join orders.

### DIFF
--- a/main/src/library/Fixpoint3/Debugging.flix
+++ b/main/src/library/Fixpoint3/Debugging.flix
@@ -17,32 +17,16 @@
  */
 
 mod Fixpoint3.Debugging {
-    use Fixpoint3.Ast.Ram
+    use Fixpoint3.Ast.Ram.RamProgram
     use Fixpoint3.Options
-
-    ///
-    /// Write `s` to standard file. Overwrite if `overwrite` is true, otherwise append.
-    ///
-    def flush(s: String, overwrite: Bool): Unit \ IO =
-        if (Options.enableDebugToFile()) {
-            let res =
-                if (overwrite)
-                    FileWrite.runWithIO(() -> FileWrite.write(str = s, Options.debugFileName()))
-                else
-                    FileWrite.runWithIO(() -> FileWrite.append(str = s, Options.debugFileName()));
-            match res {
-                case Ok(_) => ()
-                case Err(e) => println(e)
-            }
-        } else println(s)
 
     ///
     /// Invoked after every phase during the fixpoint computation.
     ///
-    /// Note: Returns `d` to ensure it is not erased.
+    /// Note: Returns `program` to ensure it is not erased.
     ///
     @Internal
-    pub def notifyPreLowering(phase: String, program: Ram.RamProgram): Ram.RamProgram = match program {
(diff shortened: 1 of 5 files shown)


main/src/library/Fixpoint3/Debugging.flix
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+mod Fixpoint3.Debugging {
+    use Fixpoint3.Ast.Ram
+    use Fixpoint3.Options
+
+    ///
+    /// Write `s` to standard file. Overwrite if `overwrite` is true, otherwise append.
+    ///
+    def flush(s: String, overwrite: Bool): Unit \ IO =
```
magnus-madsen: This is a private helper, so move it later in the file.

main/src/library/Fixpoint3/Debugging.flix
```diff
@@ -0,0 +1,64 @@
...
+            match res {
+                case Ok(_) => ()
+                case Err(e) => println(e)
+            }
+        } else println(s)
+
+    ///
+    /// Invoked after every phase during the fixpoint computation.
+    ///
+    /// Note: Returns `d` to ensure it is not erased.
```
magnus-madsen: What is d?

main/src/library/Fixpoint3/Debugging.flix
```diff
@@ -0,0 +1,64 @@
...
+    ///
+    @Internal
+    pub def notifyPreLowering(phase: String, program: Ram.RamProgram): Ram.RamProgram = match program {
```
magnus-madsen: Why not use Ram.RamProgram like elsewhere

magnus-madsen:
I need to do more code review. But perhaps this is the least important optimization? 

I seem to remember that the results were somewhat inconclusive?

JumpTwiice:
You are correct - it is an optional phase that can be disabled easily. At least in our experiments, join-order optimization resulted in a rather small speedup in general. However, it is useful to smooth a program's performance out by alleviating the programmer of finding the optimal join order in their rules themselves.

magnus-madsen:
> You are correct - it is an optional phase that can be disabled easily. At least in our experiments, join-order optimization resulted in a rather small speedup in general. However, it is useful to smooth a program's performance out by alleviating the programmer of finding the optimal join order in their rules themselves.

OK; lets leave it as the last PR to be merged then. In general I am a bit skeptical about the approach.

main/src/library/Fixpoint3/Debugging.flix
JumpTwiice: The function argument. We copied the doc comments from the old debugging file.

magnus-madsen:
If the join optimizer is not enabled then we simply rely on the rule order, right? And build the necessary indexes?

Hence- in this way- the programmer has some control over join-ordering?

JumpTwiice:
Yes.

magnus-madsen:
I have decided against this-- so I will close the PR for now.

Issue #10785, opened 2025-06-07
Add `EffectProvenance` object (Iteration 1)


magnus-madsen:
Here is a program with a bug:

```
def main(): Unit \ {Net, IO} =
    run {
        let url = "http://example.com/";
        Logger.info("Downloading URL: '${url}'");
        match HttpWithResult.get(url, Map.empty()) {
            case Result.Ok(response) =>
                let file = "data.txt";
                Logger.info("Saving response to file: '${file}'");
                let body = Http.Response.body(response);
                Console.println("fdsa"); // oops
                match FileWriteWithResult.write(str = body, file) {
                    case Result.Ok(_) =>
                        Logger.info("Response saved to file: '${file}'")
                    case Result.Err(err) =>
                        Logger.fatal("Unable to write file: '${err}'")
                }
            case Result.Err(err) =>
                Logger.fatal("Unable to download URL: '${err}'")
        }
    } with FileWriteWithResult.runWithIO
      with HttpWithResult.runWithIO
      with Logger.runWithIO
```

The error is detected near `ConstraintSolver2.scala` line 464.

Here is the value of `eqConstrs` at that point:

```
Net + IO ~ e79776
e79828 ~ Pure
e79833 ~ Logger + e79828
e79842 ~ Pure
e79845 ~ HttpWithResult + e79842
e79888 ~ Pure
e79908 ~ Logger + e79888
e79939 ~ Pure
e80046 ~ Pure
e80057 ~ Logger + e80046
e80075 ~ Pure
e80080 ~ Logger + e80075
e80112 ~ Pure
e80120 ~ Logger + e80112
e80135 ~ (e128155 - FileWriteWithResult) + IO
e79797 ~ e80135
e79833 + e79845 + e79908 + e79939 + Console + FileWriteWithResult + e80057 + e80080 + e80120 ~ e128155
e80152 ~ (e128162 - HttpWithResult) + Net + IO
e79782 ~ e80152
e79797 ~ e128162
e80161 ~ (e128182 - Logger) + IO
e79776 ~ e80161
e79782 ~ e128182
```

What is notable is that the constraint:

```
Net + IO ~ e79776
```

asserts that we ***expect*** `e79776` to be `Net + IO`. 

What I suggest is to add a class `EffectProvenance.scala` with the following initial interface:

```
object EffectProvenance {
  def debug(constrs0: List[TypeConstraint]): Unit = ... // We will return Unit for now and simply print information to the console. Later the goal will be to print something more meaningful.
[...]

PR #11007, opened 2025-07-01
feat: effect provenance

gaganchandan:
**WIP**: First iteration of effect provenance discussed in #10785 using the new `Source` provenance.


main/src/ca/uwaterloo/flix/language/phase/typer/EffectProvenance.scala
```diff
@@ -0,0 +1,198 @@
...
+  // First, we find all equality constraints that have a provenance of type
+  // Provenance.Match with a concrete effect on one side and an effect variable
+  // on the other side.
+  //
+  // For each such variable and set of concrete effects , we track a path
+  // through the constraints until we reach a constraint that has a
+  // provenance of type Provenance.ExpectEffect and an effect in the
+  // in the concrete set is not in the expected effects.
+  def debug(constrs0: List[TypeConstraint]): Unit = {
+    if (enableEffectProvenance) {
```
magnus-madsen: Can we break some of this code down into smaller helper functions? Preferably in a functional style.

main/src/ca/uwaterloo/flix/language/phase/typer/EffectProvenance.scala
```diff
@@ -0,0 +1,198 @@
...
+      constrs0.foreach { constr =>
+        constr match {
+          case TypeConstraint.Equality(_, _, prov) => {
+            prov match {
+              case Provenance.Source(eff1, eff2, loc) => {
+                (eff1, eff2) match {
+                  case _ => {
```
mlutze: I guess this case is not needed?

main/src/ca/uwaterloo/flix/language/phase/typer/EffectProvenance.scala
```diff
@@ -0,0 +1,198 @@
...
+  // through the constraints until we reach a constraint that has a
+  // provenance of type Provenance.ExpectEffect and an effect in the
+  // in the concrete set is not in the expected effects.
+  def debug(constrs0: List[TypeConstraint]): Unit = {
+    if (enableEffectProvenance) {
+      constrs0.foreach { constr =>
+        constr match {
+          case TypeConstraint.Equality(_, _, prov) => {
+            prov match {
+              case Provenance.Source(eff1, eff2, loc) => {
```
mlutze: We can simplify this with a nested match:

```suggestion
          case TypeConstraint.Equality(_, _, Provenance.Source(eff1, eff2, loc)) => {
```

main/src/ca/uwaterloo/flix/language/phase/typer/EffectProvenance.scala
mlutze: Let's invert the if and turn this into a guard.

main/src/ca/uwaterloo/flix/language/phase/typer/EffectProvenance.scala
```diff
@@ -0,0 +1,198 @@
...
+        println(msg)
+      }
+    }
+  }
+
+  // Dump all the constraints with their provenance.
+  private def printConstrs(constrs: List[TypeConstraint]): Unit = {
+    constrs.foreach {
+      case TypeConstraint.Equality(eff1, eff2, prov) => prov match {
+        case _ => println(s"$eff1 ~ $eff2 at ${prov.loc} with provenance $prov")
```
mlutze: Unneeded match?

main/src/ca/uwaterloo/flix/language/phase/typer/EffectProvenance.scala
```diff
@@ -0,0 +1,198 @@
...
+          case TypeConstraint.Equality(_, _, prov) => {
+            prov match {
+              case Provenance.Source(eff1, eff2, loc) => {
+                (eff1, eff2) match {
+                  case _ => {
+                    eff2 match {
+                      case Type.Pure => findPath(eff1, Symbol.mkEffectSym("Pure"), constr :: Nil, constrs0.filterNot(_ == constr))
+                      case _ => {
+                        val (add, _) = findConcreteEffects(eff2)
+                        eff2.effects.filter(add.contains(_)).foreach {
```
mlutze: A couple inline comments here would be nice, since this is a bit complex.

main/src/ca/uwaterloo/flix/language/phase/typer/EffectProvenance.scala
```diff
@@ -0,0 +1,198 @@
...
+    if (enableEffectProvenance) {
+      constrs0.foreach { constr =>
+        constr match {
+          case TypeConstraint.Equality(_, _, prov) => {
+            prov match {
+              case Provenance.Source(eff1, eff2, loc) => {
+                (eff1, eff2) match {
+                  case _ => {
+                    eff2 match {
+                      case Type.Pure => findPath(eff1, Symbol.mkEffectSym("Pure"), constr :: Nil, constrs0.filterNot(_ == constr))
```
mlutze: Making the effect symbol `Pure` seems brittle.

Would it make sense instead to create an `Option`-like type?

```scala
sealed trait XYZ

object XYZ {
    case class Effect(effectSym: EffSym) extends XYZ
    case class Pure extends XYZ
}
```

main/src/ca/uwaterloo/flix/language/phase/typer/EffectProvenance.scala
```diff
@@ -0,0 +1,198 @@
...
+                        val (add, _) = findConcreteEffects(eff2)
+                        eff2.effects.filter(add.contains(_)).foreach {
+                          findPath(eff1, _, constr :: Nil, constrs0.filterNot(_ == constr))
+                        }
+                      }
+                    }
+                  }
+                }
+              }
+              case _ => ()
```
mlutze: These dangling wildcard cases are hard to match up. Can we simplify by introducing one or more helper functions?

main/src/ca/uwaterloo/flix/language/phase/typer/EffectProvenance.scala
```diff
@@ -0,0 +1,198 @@
...
+              if (eff1.typeVars.contains(effVar)) {
+                val (add, sub) = findConcreteEffects(eff1)
+                if (concreteEff == Symbol.mkEffectSym("Pure") && !add.isEmpty) {
+                  ()
+                } else if (!sub.contains(concreteEff)) {
+                  eff2.typeVars.foreach { tpeVar =>
+                    tpeVar match {
+                      case Type.Var(_, _) =>
+                        findPath(tpeVar, concreteEff, constr :: path, constrs.filterNot(_ == constr))
+                      case _ => ()
```
mlutze: This case cannot happen, right?

main/src/ca/uwaterloo/flix/language/phase/typer/EffectProvenance.scala
```diff
@@ -0,0 +1,198 @@
...
+  // on the other side.
+  //
+  // For each such variable and set of concrete effects , we track a path
+  // through the constraints until we reach a constraint that has a
+  // provenance of type Provenance.ExpectEffect and an effect in the
+  // in the concrete set is not in the expected effects.
+  def debug(constrs0: List[TypeConstraint]): Unit = {
+    if (enableEffectProvenance) {
+      constrs0.foreach { constr =>
+        constr match {
```
mlutze: Scala allows simplifying functions where we immediately match on the argument:

```suggestion
      constrs0.foreach {
```

main/src/ca/uwaterloo/flix/language/phase/typer/EffectProvenance.scala
```diff
@@ -0,0 +1,198 @@
...
+                  }
+                }
+              }
+            }
+
+            case Provenance.ExpectEffect(expected, actual, loc) => {
+              if (expected.typeVars.isEmpty && actual.typeVars.contains(effVar)) {
+                expected match {
+                  // If all actual effects are not expected, we have a path.
+                  case _ => {
```
mlutze: Redundant match?

main/src/ca/uwaterloo/flix/language/phase/typer/EffectProvenance.scala
```diff
@@ -0,0 +1,198 @@
...
+    constrs.foreach {
+      case TypeConstraint.Equality(eff1, eff2, prov) => prov match {
+        case _ => println(s"$eff1 ~ $eff2 at ${prov.loc} with provenance $prov")
+      }
+      case _ => ()
+    }
+  }
+
+  // Represents a path through the constraints from a source to a sink.
+  object EffectProvenancePath {
```
mlutze: Will there be more than one kind of path in the future? If not, we can just turn this into a case class directly.

main/src/ca/uwaterloo/flix/language/phase/typer/EffectProvenance.scala
gaganchandan: I made it explicit since we refer to `constr` inside the `foreach` block.

magnus-madsen:
@gaganchandan What's the status on this?

gaganchandan:
> @gaganchandan What's the status on this?

I'm working on the changes suggested by Matt

magnus-madsen:
@mlutze How can we connect the computed paths to some error message?

mlutze:
> @mlutze How can we connect the computed paths to some error message?

Maybe we can start with an unexpected effect error, where we show both endpoints of the path?

magnus-madsen:
> > @mlutze How can we connect the computed paths to some error message?
> 
> Maybe we can start with an unexpected effect error, where we show both endpoints of the path?

What I mean is; how does Gagan connect his code to the compiler?

mlutze:
> > > @mlutze How can we connect the computed paths to some error message?
> > 
> > 
> > Maybe we can start with an unexpected effect error, where we show both endpoints of the path?
> 
> What I mean is; how does Gagan connect his code to the compiler?

Maybe we can move this analysis closer to the type error reporting site?

There's somewhere we convert all the leftover constraints to errors directly.
We can probably check that the problem is an effect problem there and then create a specific error message.

gaganchandan:
I've cleaned up the code a little by introducing helpers and removing some unnecessary stuff. I've also introduced a new `Option`-esque type to represent source effects.

mlutze:
Ready for review?

gaganchandan:
> Ready for review?

Yes.

main/src/ca/uwaterloo/flix/language/phase/typer/EffectProvenance.scala
```diff
@@ -0,0 +1,228 @@
...
+import scala.collection.immutable.SortedSet
+
+sealed trait SourceEffect {
+  def equals(eff: SourceEffect): Boolean = eff match {
+    case SourceEffect.Pure => this == SourceEffect.Pure
+    case SourceEffect.Effect(sym) => this match {
+      case SourceEffect.Pure => false
+      case SourceEffect.Effect(thisSym) => thisSym == sym
+    }
+  }
```
mlutze, line 33: Do we need to do this? I think this definition is just the default one that Scala would generate, no?

main/src/ca/uwaterloo/flix/language/phase/typer/EffectProvenance.scala
```diff
@@ -0,0 +1,228 @@
...
+        // Kick off path searching from source constraints.
+        case TypeConstraint.Equality(_, _, Provenance.Source(eff1, eff2, loc)) => {
+          eff2 match {
+            case Type.Pure => findPath(eff1, SourceEffect.Pure, constr :: Nil, constrs0.filterNot(_ == constr))
+            case _ => {
+              val (add, _) = findConcreteEffects(eff2)
+              // For each concrete effect introduced by the source effect,
+              // we begin the search for a path.
+              eff2.effects.filter(add.contains(_)).toList.flatMap { eff =>
+                findPath(eff1, SourceEffect.Effect(eff), constr :: Nil, constrs0.filterNot(_ == constr))
```
mlutze, line 79: There's a lot going on in these lines of code.

The difference between the two branches of the case seems to be the set of source effects. We can pull that out with something like:

```scala
sources = eff2 match {
    case Type.Pure => List(SourceEffect.Pure)
    case _ => findConcreteEffects(eff2)
}
// now map over sources
```

main/src/ca/uwaterloo/flix/language/phase/typer/EffectProvenance.scala
```diff
@@ -0,0 +1,228 @@
...
+        // If the concrete effect is Pure and new effects are not introduced, we can traverse further.
+        case SourceEffect.Pure if add.isEmpty => {
+          eff2.typeVars.toList.flatMap { tpeVar =>
+            findPath(tpeVar, concreteEff, constr :: path, constrs.filterNot(_ == constr))
+          }
+        }
+        case SourceEffect.Effect(sym) if !sub.contains(sym) => {
+          // If the concrete effect is not in the subtracted effects, we can traverse further.
+          eff2.typeVars.toList.flatMap { tpeVar =>
+            findPath(tpeVar, concreteEff, constr :: path, constrs.filterNot(_ == constr))
```
mlutze, line 115: These branches do the same thing right?

main/src/ca/uwaterloo/flix/language/phase/typer/EffectProvenance.scala
```diff
@@ -0,0 +1,228 @@
...
+        (expected, concreteEff) match {
+          case (Type.Pure, SourceEffect.Pure) => {
+            // If the expected effect is Pure and the concrete effect is Pure, we have a path.
+            val newPath = EffectProvenancePath(concreteEff, path.reverse.head, constr :: path, constr)
+            newPath :: Nil
+          }
+          case (_, SourceEffect.Effect(sym)) if !expected.effects.contains(sym) => {
+            // If the expected effect does not contain the concrete effect, we have a path.
+            val newPath = EffectProvenancePath(concreteEff, path.reverse.head, constr :: path, constr)
+            newPath :: Nil
```
mlutze, line 136: These branches do the same thing, right?

mlutze:
I'll merge this so we can iterate.

mlutze (approved):

Issue #11005, opened 2025-07-01
Introduce NormalType


mlutze:
Monomorpher has a lot of invariants on its types, which are not encoded in `Type.scala`.

I'd like to introduce a new type, `NormalType`, that lies between `Type` and `MonoType`, which enforces these invariants.

mlutze:
First some renaming:

* SimpleType -> DisplayType
* MonoType -> SimpleType
* Then NormalType becomes the new MonoType

magnus-madsen:
Fixed

mlutze:
This is not yet where I'd like it; not all the invariants are enforced. This stalled due to type match but now that it's gone, maybe we can look into it again.

magnus-madsen:
Can you write up what its about?

magnus-madsen:
@mlutze Maybe our friend Claude?

mlutze:
> Can you write up what its about?

I want a unique representation of types that we can use with monomorphization. The representation will have well-defined equality and hashcode. This way, we have type-level assurances that types are normalized, and we don't have to worry all the time about using equality/hash on types in monomorphization.

mlutze:
> [@mlutze](https://github.com/mlutze) Maybe our friend Claude?

Maybe. But we wait until the monomorpher2 is in place.

mlutze:
By the way, this is a helper for indexed effects IIRC, since we really wanted to put types in a set/map/something.

PR #11013, opened 2025-07-02
chore: rename SimpleType to DisplayType

mlutze:
I decided against `FormattedType` because it's confusing with our typical terminology, where `formatX` means turn it into a string. Plus then we would need a `formatFormattedType` function which seemed silly to me.

related to #11005


magnus-madsen:
I think calling it TypeView is misleading. It is not a view.

Moreover, the nice thing about FormattedType or at least something like it is that then you know you should not compute with it.

mlutze:
Changed to DisplayType

Issue #10965, opened 2025-06-25
Support `where` constraints on trait instance declarations


rywiese:
Flix supports `where` constraints on function declarations, [for example](https://doc.flix.dev/associated-types.html#equality-constraints)
```
def sum(c: t): Int32 with Collection[t] where Collection.Elm[t] ~ Int32 = 
    Collection.toList(c) |> List.sum
```

It should also support such constraints on `instance` declarations, such as
```
instance SomeTrait[t] with Collection[t] where Collection.Elm[t] ~ Int32 { ... }
```

### Use case
I need this for a specific use case (which i have simplified for readability):

I have a trait:
```
trait Readable[t] {

    type Aty

    pub def read(reader: t): Option[Readable.Aty[t]]

}
```

and I want to write an adapter that wraps an `Int8` `Readable` and implements a `Char` `Readable`:
```
struct DecodingReader[t, r] {
    charset: Charset,
    byteReader: t
}
```
Here `t` must be a `Readable` instance with an associated type of `Int8` (the `Charset` can only convert bytes), but this constraint does not live in the struct.

I can implement the `read` function with the constraint that `t` is a `Char` `Readable`:
```
mod DecodingReader {

    pub def read(reader: DecodingReader[t, r]): Option[Char] with Readable[t] where Readable.Aty[t] ~ Int8 = ???

}
```

But there is currently no way to package this function up in an instance of `Readable`, like:
```
instance Readable[DecodingReader[t, r]] with Readable[t] /*where Readable.Aty[t] ~ Int8*/ {

    type Aty = Char

    pub def read(reader: DecodingReader[t, r]): Option[Char] = DecodingReader.read(reader)

}
```

where I can only construct the instance if `Readable[t]` (possible today) _and_ `Readable.Aty[t] ~ Int8*` (not possible today).

magnus-madsen:
Very cool! @mlutze Can we give Ry this super power?

mlutze:
> Very cool! @mlutze Can we give Ry this super power?

I gave an overview today. Barring the inevitable unforeseen circumstances, he has the tools he needs to implement it himself.

magnus-madsen:
> > Very cool! [@mlutze](https://github.com/mlutze) Can we give Ry this super power?
> 
> I gave an overview today. Barring the inevitable unforeseen circumstances, he has the tools he needs to implement it himself.

Even better! He can become a super hero himself. Very expedient

PR #11021, opened 2025-07-02
feat: support equality constraints on instance defs

rywiese:
Fixes https://github.com/flix/flix/issues/10965


main/test/flix/Test.A.Ry.RenameMe3.flix
```diff
@@ -0,0 +1,30 @@
...
+                |> Option.flatMap(Iterator.next)
+                |> Option.sum; // Requires Option[Int32]
+            popped
+        }
+}
+
+pub def ry(): Unit \ IO =
+    region rc {
+        let mList: MutList[Option[Int32], rc] = MutList.empty(rc) !> MutList.push(Some(1));
+        let popped: Option[Option[Int32]] = Popable.pop(mList);
```
rywiese: This fails here and it shouldn't

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolver2.scala
```diff
@@ -622,6 +625,10 @@ object ConstraintSolver2 {
     case TraitConstraint(head, arg, _) => TypeConstraint.Trait(head.sym, arg, loc)
   }
 
+  def equalityConstraintToTypeConstraint(constr: EqualityConstraint, loc: SourceLocation): TypeConstraint = constr match {
+    case EqualityConstraint(_, tpe1, tpe2, _) => TypeConstraint.Equality(tpe1, tpe2, Provenance.Match(tpe1, tpe2, loc))
```
mlutze: There's your problem :)


You need to construct an associated type on the left using the symbol from the equality constraint.

main/src/ca/uwaterloo/flix/language/phase/Instances.scala
```diff
@@ -245,6 +246,19 @@ object Instances {
                     }
                   }
               }
+              superInst.econstrs.foreach {
+                econstr =>
+                  EqualityEnvironment.entail(econstrs.map(subst.apply), subst(econstr), root.traitEnv, eqEnv) match {
+                    case Result.Ok(_) => Nil
+                    case Result.Err(errors) => errors.foreach {
```
rywiese, line 261: Open to feedback on these compiler errors.

An example of the errors in action:

Screenshot 2025-07-03 at 4 49 20 PM

Ideally, I'd want to assume that the constraints on `A` are correct/intended, and tell the user that the constraint on the instance for `B` is invalid.

This isn't quite doable though, because we add all the constraints on `B` as assumptions in the environment and then (one by one) ensure that the constraints on `A` are entailed. We fail here if we find a constraint on `A` that is not entailed. So there's a specific constraint on `A` to fail on, but none in `B`.

`ConflictingEqualityConstraint` highlights the specific failed constraint on `A`.
`ConflictingEqualityConstraints` highlights that there are constraints on `B` that don't entail the constraint on `A`. But we don't have a specific culprit (and generally, there might not be just one).

I waffled on which of these errors to throw, but figured maybe it's good to have both?

Also, since we can't point at a specific constraint in `B`, the second error just points to the instance name. It might be better to put a red squiggly under the equality constraints in `B`, but I think this will require computing the source position start/end based on the positions of each constraint. Doable but I didn't want to do it before hearing other people's thoughts.

mlutze: I think indeed there are two different errors here:

1. Missing equality constraint (supertrait constraint not entailed)
2. Conflicting equality constraints

It seems to me that the first one is essential for correctness. The second one is not essential because any declaration with conflicting constraints cannot be used anyway.

I propose we address the second one in a separate PR.

rywiese: If I understand you right, both of the errors I have here address 1. I don't have any logic to address 2.

I meant "conflict" as in: the subtrait constraints conflict with the supertrait constraints. Maybe "missing" would be better. But the question I'm asking is whether this error should be thrown at the subtrait, supertrait, or both
1. From a user's perspective, I think erroring on the subtrait makes more sense as this is likely where the problem was introduced.
2. By the nature of the entailment check, the error arises on some constraint in the supertrait which is not entailed. So if we error on this supertrait constraint, we can actually point to that specific constraint and give the user more information to debug.

I thought maybe it would be helpful to throw an error in both locations (as pictured)?

main/src/ca/uwaterloo/flix/language/ast/Scheme.scala
```diff
@@ -140,11 +140,16 @@ object Scheme {
...
-  def lessThanEqual(sc1: Scheme, sc2: Scheme, tenv0: TraitEnv, eenv0: EqualityEnv)(implicit scope: Scope, flix: Flix): Boolean = {
+  def lessThanEqual(sc1: Scheme, sc2: Scheme, tenv0: TraitEnv, globalEqEnv0: EqualityEnv, localEconstrs: List[EqualityConstraint])(implicit scope: Scope, flix: Flix): Boolean = {
 
     // Instantiate sc2, creating [T/α₂]π₂ and [T/α₂]τ₂
     // We use the top scope because this function is only used for comparing schemes, which are at top-level.
-    val (cconstrs2_0, econstrs2_0, tpe2_0, _) = Scheme.instantiate(sc2, SourceLocation.Unknown)(Scope.Top, flix)
+    val (cconstrs2_0, econstrs2_0, tpe2_0, substMap) = Scheme.instantiate(sc2, SourceLocation.Unknown)(Scope.Top, flix)
+
+    // TODO: Do I need to recreate Scheme.instantiate/visitType (local fun) instead of relying on subst0.apply?
+    // It seems visitType recreates Substitution logic but also replaces locs.
```
magnus-madsen: What do you say @mlutze ?

main/src/ca/uwaterloo/flix/language/errors/InstanceError.scala
```diff
@@ -271,6 +272,50 @@ object InstanceError {
     })
   }
 
+  /**
```
magnus-madsen: I believe this file is alphabetically sorted, so these two error messages should be moved up.

main/src/ca/uwaterloo/flix/language/errors/InstanceError.scala
```diff
@@ -271,6 +272,50 @@ object InstanceError {
...
+    * An error indicating that the equality constraint of an instance declaration conflicts with the equality constraints of a sub instance declaration due to a type mismatch.
+    *
```
magnus-madsen: I would give an example here, to aid understanding.

main/src/ca/uwaterloo/flix/language/errors/InstanceError.scala
```diff
@@ -271,6 +272,50 @@ object InstanceError {
...
+      s""">> Constraint conflicts with constraints on instance for '${subTrait.name}'. Unable to unify the types '${formatType(fullType1, None, minimizeEffs = true)}' and '${formatType(fullType2, None, minimizeEffs = true)}'.
+         |
+         |${code(loc, s"conflicting equality constraint")}
+         |""".stripMargin
+    }
+  }
+
+  /**
+    * An error indicating that the equality constraints of an instance declaration conflict with an equality constraint of a super instance declaration due to a type mismatch.
+    *
```
magnus-madsen: same about an example

main/src/ca/uwaterloo/flix/language/errors/InstanceError.scala
```diff
@@ -271,6 +272,50 @@ object InstanceError {
...
+    * @param baseType1 the first base type.
+    * @param baseType2 the second base type.
+    * @param fullType1 the first full type.
+    * @param fullType2 the second full type.
+    * @param supTrait  the trait whose equality constraint conflicts with the equality constraints on the sub instance.
+    * @param loc       the location of the sub trait whose equality constraints conflict with the equality constraints on the super instance.
+    */
+  case class ConflictingEqualityConstraints(baseType1: Type, baseType2: Type, fullType1: Type, fullType2: Type, supTrait: Symbol.TraitSym, loc: SourceLocation)(implicit flix: Flix) extends InstanceError {
```
magnus-madsen: Why is this plural??

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -394,49 +396,72 @@ object Typer {
...
-            if (matches) {
-              ()
-            } else {
-              val renv = tparams.map(_.sym).foldLeft(RigidityEnv.empty)(_.markRigid(_))
-              val error = TypeError.MissingTraitConstraint(trtSym, arg, renv, loc)
-              sctx.errors.add(error)
-              ()
-            }
+      // check that they are all covered by the type constraints
+      tpes.flatMap(getAssocTypes).foreach(checkAssocType(_, tparams, extraTconstrs ::: tconstrs, tenv))
```
magnus-madsen: I would suggest some old style loops here because flatMap followed by foreach is a bit a mixture of styles. The first flatMap suggests we are doing pure FP, but that is actually not the case. So I think two for(each) loops are better.

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -394,49 +396,72 @@ object Typer {
...
+  /**
+    * Verifies that all the associated types in the instance are resolvable, according to the declared type constraints.
+    */
+  private def checkAssocTypes(inst: KindedAst.Instance, tenv: TraitEnv)(implicit sctx: SharedContext, flix: Flix): Unit = inst match {
+    case KindedAst.Instance(_, _, _, _, tparams, tpe, tconstrs, econstrs, _, _, _, _) =>
+      // get all the associated types in the instance
+      val tpes = tpe :: econstrs.flatMap(getTypes)
+
+      // check that they are all covered by the type constraints
+      tpes.flatMap(getAssocTypes).foreach(checkAssocType(_, tparams, tconstrs, tenv))
```
magnus-madsen: same comment as above

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -394,49 +396,72 @@ object Typer {
...
+        ()
+      } else {
+        val renv = tparams.map(_.sym).foldLeft(RigidityEnv.empty)(_.markRigid(_))
+        val error = TypeError.MissingTraitConstraint(trtSym, arg, renv, loc)
+        sctx.errors.add(error)
+        ()
+      }
+
+    // Resiliency: If the associated type is ill-formed, then we ignore it.
+    case _ => ()
```
magnus-madsen: @mlutze Are you OK with the code above?

main/src/ca/uwaterloo/flix/language/phase/unification/TraitEnvironment.scala
```diff
@@ -17,14 +17,11 @@
 package ca.uwaterloo.flix.language.phase.unification
```
magnus-madsen: The changes in this file seem unnecessary. Undo?

main/test/ca/uwaterloo/flix/language/phase/TestTyper.scala
```diff
@@ -315,6 +357,26 @@ class TestTyper extends AnyFunSuite with TestUtils {
...
+        |
+        |enum Adapter[t, a, b](t, a -> b)
+        |
+        |instance A[Adapter[t, a, b]] where A.Typ[t] ~ Int8 {
+        |    type Typ = b
+        |    pub def foo(adapter: Adapter[t, a, b]): b = ???
+        |}
+        |""".stripMargin
+    val result = compile(input, Options.TestWithLibMin)
+    expectError[TypeError](result)
```
magnus-madsen: Expect the precise type error?

main/test/ca/uwaterloo/flix/language/phase/TestTyper.scala
```diff
@@ -140,6 +140,48 @@ class TestTyper extends AnyFunSuite with TestUtils {
...
+        |
+        |instance A[Adapter[t, a, b]] with A[t] where A.Typ[t] ~ Int8 {
+        |    type Typ = b
+        |    pub def foo(adapter: Adapter[t, a, b]): b =
+        |        let Adapter.Adapter(x, f) = adapter;
+        |        f(A.foo(x))
+        |}
+        |""".stripMargin
+    val result = compile(input, Options.TestWithLibMin)
+    expectError[TypeError](result)
```
magnus-madsen: Expect the precise type error?

main/test/ca/uwaterloo/flix/language/phase/TestTyper.scala
```diff
@@ -140,6 +140,48 @@ class TestTyper extends AnyFunSuite with TestUtils {
...
+        |
+        |instance A[Adapter[t, a, b]] with A[t] {
+        |    type Typ = b
+        |    pub def foo(adapter: Adapter[t, a, b]): b =
+        |        let Adapter.Adapter(x, f) = adapter;
+        |        f(A.foo(x))
+        |}
+        |""".stripMargin
+    val result = compile(input, Options.TestWithLibMin)
+    expectError[TypeError](result)
```
magnus-madsen: Expect the precise type error?

main/test/ca/uwaterloo/flix/language/phase/TestTyper.scala
```diff
@@ -315,6 +357,26 @@ class TestTyper extends AnyFunSuite with TestUtils {
...
+        |    type Typ = b
+        |    pub def foo(adapter: Adapter[t, a, b]): b = ???
+        |}
+        |""".stripMargin
+    val result = compile(input, Options.TestWithLibMin)
+    expectError[TypeError](result)
+  }
+
   test("NoMatchingInstance.Location.01") {
     val input =
```
magnus-madsen: I think a handful more of negative tests would be appropriate.

Moreover, these should be named after the specific new error(s) and expect them

main/test/flix/Test.Instance.Eq.Constraint.Super.flix
```diff
@@ -0,0 +1,52 @@
+mod Test.Instance.Eq.Constraint.Super {
+
+    trait Popable[t] {
+        type Elm
```
magnus-madsen: I would kind this- just for readability

main/test/flix/Test.Instance.Eq.Constraint.Super.flix
```diff
@@ -0,0 +1,52 @@
...
+            let popped = Popable.pop(mList);
+            popped == Some(Some(1))
+        }
+
+    @test
+    pub def testReplace(): Bool \ IO =
+        region rc {
+            let mList = MutList.empty(rc) !> MutList.push(Some(1));
+            let old = Replaceable.replace(Some(2), mList);
+            let noo = Popable.pop(mList);
```
magnus-madsen, line 48: What is with the local variable naming?

main/test/flix/Test.Instance.Eq.Constraint.flix
```diff
@@ -0,0 +1,76 @@
+mod Test.Instance.Eq.Constraint {
+
+    trait Producer[p] {
+        type Elm
```
magnus-madsen: : Type

main/test/flix/Test.Instance.Eq.Constraint.flix
```diff
@@ -0,0 +1,76 @@
...
+        type Aef: Eff
+        pub def produce(producer: p): Producer.Elm[p] \ Producer.Aef[p]
+    }
+
+    trait Consumer[c] {
+        type Elm
```
magnus-madsen: : Type

magnus-madsen:
@mlutze I think we can move towards merging this soon? 

We could do better error messages in a follow up PR. But of course we could also try to nail them down now?

main/src/ca/uwaterloo/flix/language/ast/Scheme.scala
mlutze: I believe you can just use subst0.apply.

Type locations don't matter in this function, since it just returns a Boolean (unlike fx. an error that has a location)

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
mlutze: Or rather, one for-loop in Scala syntax:

```
for {
 t <- tpes
 assocs <- getAssocTypes(t)
} {
    checkAssocType(...)
}
```

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
```diff
@@ -394,49 +396,72 @@ object Typer {
...
+  }
 
-          // Resiliency: If the associated type is ill-formed, then we ignore it.
-          case _ => ()
-        }
-    }
+  /**
+    * Verifies that all the associated types in the instance are resolvable, according to the declared type constraints.
+    */
+  private def checkAssocTypes(inst: KindedAst.Instance, tenv: TraitEnv)(implicit sctx: SharedContext, flix: Flix): Unit = inst match {
```
mlutze: We try to avoid overloading. In Scala it often leads to confusing error messages.

main/test/flix/Test.Instance.Eq.Constraint.Super.flix
mlutze: `new` is a keyword :P

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
mlutze: It looks like it's just a refactoring of existing code.

@rywiese Can we extract this refactoring to a separate PR so it's isolated?

main/src/ca/uwaterloo/flix/language/phase/Instances.scala
mlutze: I agree that the error location should be on the subtrait. Since it's possible the user is extending a trait from a library they don't own, we shouldn't point at that as the problem.

I propose we just have one error location, but we give the constraint that is missing rather than the `unable to unify ...` string.

rywiese: But we don't have the constraint that is missing really... we just have a constraint on the supertrait that isn't entailed. as far as i can tell we can't point at any specific constraint on the subtrait that breaks the entailment. so we either point to a specific constraint on the supertrait or just all the constraints on the subtrait

mlutze: "isn't entailed" and "missing" are pretty much the same, no?

rywiese: No because the one that's "missing" isn't the one that "isn't entailed". The one that isn't entailed is present and on the super trait. The one that's missing is on the subtrait, and, if present, would entail the one on the super. But the one on the super is the only one we know about specifically

rywiese: General case:
```
trait A[t] {}

trait B[t] with A[t] where c_b1, ..., c_bn {} // c_bi is a constraint

trait C[t] with B[t] where c_c1, ..., c_cm {} // c_cj is a constraint
```

When checking `C`, we loop through constraints on `B` and check one by one that they are entailed. So if we encounter a `c_bi` that is not entailed, we can
1. error on `B` and say "here's a specific constraint, c_bi, that is not entailed" and put a red squiggly under that constraint
2. error on `C` and say " `c_c1`, ..., `c_cm` don't entail `c_bi`, but we don't know which one" and put a red squiggly under the whole thing

1 is unintuitive as the error is on the super, but it has a benefit of giving more information. The problem here is that we don't get any information back from the constraint solver about which of the assumptions caused `c_bi` to not be entailed

mlutze: Ok, but it's a reasonable approximation to say that it's missing, right?

In other words, if `A.T[a] ~ String` is on the supertrait, but not the subtrait, then the only reasonable solution is just to add `A.T[a] ~ String` to the subtrait.

I guess there's some alpha renaming to be done, but that's all, no?

rywiese: Oh sure that makes sense. I maybe got hung up overthinking the red squiggly. This should be doable, I'll look into the alpha renaming

rywiese: New error:
Screenshot 2025-07-09 at 12 47 30 PM

main/src/ca/uwaterloo/flix/language/errors/InstanceError.scala
rywiese: Replaced with `MissingEqConstraint`

main/src/ca/uwaterloo/flix/language/errors/InstanceError.scala
rywiese: Replaced with `MissingEqConstraint`

main/src/ca/uwaterloo/flix/language/errors/InstanceError.scala
rywiese: Replaced with `MissingEqConstraint`

main/src/ca/uwaterloo/flix/language/phase/Typer.scala
rywiese: Done

main/test/ca/uwaterloo/flix/language/phase/TestTyper.scala
rywiese: I've added them to `TestInstances`, as those new errors are `InstanceError`s.

rywiese:
I've addressed comments, fixed the error message, and taken it out of draft mode. I think it's ready for final review @magnus-madsen @mlutze

main/src/ca/uwaterloo/flix/language/phase/Instances.scala
mlutze: I trust you get `Aty[sadf]` when it's called `B[Wrapper[sadf]]`?

main/src/ca/uwaterloo/flix/language/ast/Scheme.scala
```diff
@@ -140,11 +140,14 @@ object Scheme {
     * ---------------------------------------
     * Θₚ ⊩ (∀α₁.π₁ ⇒ τ₁) ≤ (∀α₂.π₂ ⇒ τ₂)
     */
-  def lessThanEqual(sc1: Scheme, sc2: Scheme, tenv0: TraitEnv, eenv0: EqualityEnv)(implicit scope: Scope, flix: Flix): Boolean = {
+  def lessThanEqual(sc1: Scheme, sc2: Scheme, tenv0: TraitEnv, globalEqEnv0: EqualityEnv, localEconstrs: List[EqualityConstraint])(implicit scope: Scope, flix: Flix): Boolean = {
```
mlutze, line 147: Let's add a comment explaining what we're doing here -- local constraints need to be instantiated.

main/src/ca/uwaterloo/flix/language/phase/unification/EqualityEnvironment.scala
```diff
@@ -0,0 +1,47 @@
...
+import ca.uwaterloo.flix.language.ast.shared.{EqualityConstraint, Scope}
+import ca.uwaterloo.flix.language.phase.typer.TypeConstraint.Provenance
+import ca.uwaterloo.flix.language.phase.typer.{ConstraintSolver2, ConstraintSolverInterface, SubstitutionTree, TypeConstraint}
+import ca.uwaterloo.flix.util.Result
+
+object EqualityEnvironment {
+
+  /**
+    * Returns success iff equality constraints `econstrs0` entail type constraint `econstr0`.
+    * That is, `econstr0` is true if all of `tconstrs0` are true.
```
mlutze, line 30: Something about what it returns if it's not successful.

main/src/ca/uwaterloo/flix/language/phase/unification/EqualityEnvironment.scala
```diff
@@ -0,0 +1,47 @@
...
+    */
+  def entail(econstrs0: List[EqualityConstraint], econstr0: EqualityConstraint, traitEnv: TraitEnv, eqEnv: EqualityEnv)(implicit scope: Scope, flix: Flix): Result[Unit, List[TypeConstraint]] = {
+    val econstr = econstr0 match {
+      case EqualityConstraint(symUse, tpe1, tpe2, loc) =>
```
mlutze, line 36: Don't we have a helper for this somewhere?

rywiese: I think there was a private helper in another file, but I couldn't find anything public

main/src/ca/uwaterloo/flix/language/phase/Instances.scala
rywiese: I do now:
Screenshot 2025-07-09 at 2 33 00 PM

main/src/ca/uwaterloo/flix/language/ast/Scheme.scala
rywiese: Let me know if what I said is accurate and makes sense... i understand the example I was working with but it's a little tricky to describe abstractly

rywiese: I also added a comment on 153 where it's used

magnus-madsen (approved):

magnus-madsen:
Merged! Yay! 🚀 

@mlutze If you have any follow-up concerns, I suggest to create a ticket and put them there.

main/src/ca/uwaterloo/flix/language/phase/Instances.scala
```diff
@@ -224,18 +225,24 @@ object Instances {
...
     * and that the constraints on `inst` entail the constraints on the super instance.
     */
-  private def checkSuperInstances(inst: TypedAst.Instance, root: TypedAst.Root)(implicit sctx: SharedContext, flix: Flix): Unit = inst match {
-    case TypedAst.Instance(_, _, _, trt, tpe, tconstrs, _, _, _, _) =>
+  private def checkSuperInstances(inst: TypedAst.Instance, root: TypedAst.Root, eqEnv: EqualityEnv)(implicit sctx: SharedContext, flix: Flix): Unit = inst match {
+    case TypedAst.Instance(_, _, _, trt, tpe, tconstrs, econstrs, _, _, _, _) =>
+      // Rigidify sub-instance constraint vars in the substitution to ensure that they appear as-written in compiler errors
+      val evars = econstrs.flatMap(econstr => econstr.tpe1.typeVars ++ econstr.tpe2.typeVars).map(_.sym)
+      val tvars = tconstrs.flatMap(tconstr => tconstr.arg.typeVars).map(_.sym)
+      val rigidityEnv = RigidityEnv.ofRigidVars(evars ::: tvars)
```
mlutze, line 233: I think creating this rigidityenv should be in the logic of `getInstanceForType` instead of here.

Issue #10897, opened 2025-06-17
Rework `IpAddr.toByteArray`


magnus-madsen:
- Return type could be Vector instead of Array -- then no region is needed.
- Name could be just toBytes

PR #11025, opened 2025-07-02
refactor: make `toBytes` return a Vector instead of an Array

CadeMichael:
Addresses #10897.

### DIFF
--- a/main/src/library/Ping.flix
+++ b/main/src/library/Ping.flix
@@ -43,8 +43,9 @@ mod Ping {
                 try {
                     let ms = TimeUnit.toMilliseconds(u);
                     let timeoutMs = ms * timeout;
+                    let ipVec = IpAddr.toBytes(host);
                     region rc {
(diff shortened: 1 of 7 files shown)


main/src/library/Ping.flix
```diff
@@ -43,8 +43,9 @@ mod Ping {
                 try {
                     let ms = TimeUnit.toMilliseconds(u);
                     let timeoutMs = ms * timeout;
+                    let ipVec = IpAddr.toBytes(host);
                     region rc {
-                        let ipBytes = IpAddr.toByteArray(rc, host);
+                        let ipBytes = Vector.toArray(rc, ipVec);
```
magnus-madsen: Is this needed? I think maybe you can just call getByAddress with the vector

CadeMichael: yep sorry I assumed the types wouldn't match up pushing change now.

PR #11029, opened 2025-07-03
feat: add hoisting phase

CasperDN:


main/src/library/Fixpoint3/Phase/Hoisting.flix
```diff
@@ -0,0 +1,620 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+///
+/// One hard dependency of the current implementation is that for `BoolExp.Eq(x, y)` `x` and `y` can
```
magnus-madsen: I feel like I need a few sentences before this to explain the goal.

"The purpose of the hoisting phase is to ..."

main/src/library/Fixpoint3/Phase/Hoisting.flix
```diff
@@ -0,0 +1,620 @@
...
+/// either be `Lit`, or `RowLoad(_ ,_, _)`.
+///
+/// The hoisting in general happens with a `MutDisjointSets`, called `equalitySets`,
+/// which is populated by `BoolExp.Eq(RowLoad, RowLoad)`. In this way all `RowLoad`'s
+/// which are equal will be treated as one.
+///
+/// `termMap` contains a map from the representing `RowLoad`, with respect to `equalitySets`,
+/// to the `RamTerm` highest in the AST where the value of `Rep` is known.
+///
+/// For
```
magnus-madsen: For example, given:

main/src/library/Fixpoint3/Phase/Hoisting.flix
```diff
@@ -0,0 +1,620 @@
...
+///     search b ∈ B do
+///         search c ∈ C do
+///             if(b[0] = a[0] ∧ c[0] = a[0], b[1] = 1, b[2] = c[1]) then
+///                 project (a[0]) into R
+///             end
+///         end
+///     end
+/// end
+/// ```
+/// `equalitySets` could be `{a[0] => a[0], b[0] => a[0], c[0] => a[0], b[1] => 1, b[2] => c[1], c[1] => c[1]}`.
```
magnus-madsen: could be or is?

main/src/library/Fixpoint3/Phase/Hoisting.flix
```diff
@@ -0,0 +1,620 @@
...
+    /// if(not (A = ∅) ∧ not (B = ∅) not (C = ∅)) then
+    ///     search a ∈ A do
+    ///         query {b ∈ B | b[0] = a[0]} do
+    ///             query {c ∈ C | c[0] = a[0]} do
+    ///                 project (a[0]) into R
+    ///             end
+    ///         end
+    ///     end
+    /// end
+    /// ```
```
magnus-madsen: Thanks for the example 👍

magnus-madsen: I wonder if this example and doc would not be more appropriate higher op?

main/src/library/Fixpoint3/Phase/Hoisting.flix
```diff
@@ -0,0 +1,620 @@
...
+                        };
+                    Some(newStmt)
+            }
+        case RamStmt.MergeInto(_, _) => Some(stmt)
+        case RamStmt.Swap(_, _) => Some(stmt)
+        case RamStmt.Purge(_) => Some(stmt)
+        case RamStmt.Seq(xs) =>
+            let children =
+                Vector.filterMap(hoistStmt(rc, termMap, equalitySets, impossible), xs);
+            if (Vector.isEmpty(children))
```
magnus-madsen: A pattern match:

```
match Vector.length(children) {
  case 0 => 
  case 1 => 
  case _ = > 
}
```

might be cleaner. I leave it to you.

main/src/library/Fixpoint3/Phase/Hoisting.flix
```diff
@@ -0,0 +1,620 @@
...
+                None
+            else if (Vector.length(children) == 1)
+                Vector.head(children)
+            else Some(RamStmt.Seq(children))
+        case RamStmt.Par(xs) =>
+            let children =
+                Vector.filterMap(hoistStmt(rc, termMap, equalitySets, impossible), xs);
+            if (Vector.isEmpty(children)) None
```
magnus-madsen: same comment about pattern match

main/src/library/Fixpoint3/Phase/Hoisting.flix
magnus-madsen: unfortunately I don't think we have pattern matching on vectors, do we?

main/src/library/Fixpoint3/Phase/Hoisting.flix
```diff
@@ -0,0 +1,620 @@
...
+    ///
+    /// Returns the set of `RowVar` in `term`. See `collectBoolVariables`.
+    ///
+    def collectTermVariables(termMap: MutMap[(RowVar, Int32), RamTerm, r], equalitySets: MutDisjointSets[(RowVar, Int32), r], term: RamTerm): Set[RowVar] \ r =
+        let collectTerm = collectTermVariables(termMap, equalitySets);
+        let getRepVar = getRepRowVar(termMap, equalitySets);
+        match term {
+            case RamTerm.Lit(_, _)               => Set#{}
+            case RamTerm.RowLoad(rv, i, _)       => getRepVar(rv, i)
+            case RamTerm.Meet(_, t1, (rv, relSym), _) => Set.union(collectTerm(t1), getRepVar(rv, arityOfNonLat(relSym)))
```
magnus-madsen: I think you wanted => to be aligned, but here it is not

main/src/library/Fixpoint3/Phase/Hoisting.flix
```diff
@@ -0,0 +1,620 @@
...
+    ///
+    /// Ignores `Eq`, as `Eq`'s are seen as a special case and will always be a comparison between loads and
+    /// loads/constants.
+    ///
+    /// As an example `Guard2(x[0], y[3])` will return `{x}` if `termMap` is e.g. `[y[3] => 8]`.
+    /// If `termMap` contained `y[3]=> z[3]` then `{x, z}` will be returned.
+    ///
+    /// The purpose is to return the `RowVar`'s in the AST which are possible places `bool` can be hoisted to.
+    ///
+    def collectBoolVariables(termMap: MutMap[(RowVar, Int32), RamTerm, r], equalitySets: MutDisjointSets[(RowVar, Int32), r], bool: BoolExp): Set[RowVar] \ r =
```
magnus-madsen: While `bool` is a valid variable/fparam name... it is perhaps not the best name. Perhaps bexp? etc

(same comment applies elsewhere)

main/src/library/Fixpoint3/Phase/Hoisting.flix
```diff
@@ -0,0 +1,620 @@
...
+        case BoolExp.Eq(RamTerm.Lit(val, id), RamTerm.RowLoad(rv, i, _)) =>
+            MutMap.putWith(_ -> list -> (val, id) :: list, (rv, i), (val, id) :: Nil, constEqualities)
+        case _ => ()
+    }
+
+    ///
+    /// Add `(rowVar, i)` for all `i` between `0` (inclusive) and `arity` (exlusive)
+    /// to `equalitySets`.
+    ///
+    def collectLoads(rowVar: RowVar, arity: Int32, equalitySets: MutDisjointSets[(RowVar, Int32), r]): Unit \ r =
```
magnus-madsen: I think - at some point - we need to come up with a naming scheme where we do not use "collect" for functions that sometimes return their result and other times stores their result by mutation into an argument.

main/src/library/Fixpoint3/Phase/Hoisting.flix
CasperDN: Could be. We use Flix's disjoint-set data structure and which value ends up being the representative is not important for us.

main/src/library/Fixpoint3/Phase/Hoisting.flix
CasperDN: I have moved it up above the 'hard dependencies'.

main/src/library/Fixpoint3/Phase/Hoisting.flix
CasperDN: I don't think so. At least it complained when I just tried
```
match children {
  case Vector#{x} => ???
}
```

main/src/library/Fixpoint3/Phase/Hoisting.flix
CasperDN: Perhaps `register` or `add` could be used for mutation?

main/src/library/Fixpoint3/Phase/Hoisting.flix
magnus-madsen: OK; can we clarify that

main/src/library/Fixpoint3/Phase/Hoisting.flix
magnus-madsen: or collectInto?

magnus-madsen:
I will let @JonathanStarup also take a look

main/src/library/Fixpoint3/Phase/Hoisting.flix
CasperDN: For this file it is only used in this way for `collectLoads`. `CollectIntoLoads` sounds weird. `CollectLoadsInto`?

Also should we rename `unifyEqualitiesX` to `collectIntoX`? It would follow the proposed convention, but it also loses the intuitive meaning given by `unify`

magnus-madsen: I am not sure. We can consider it later. We dont have to fix it for this PR

CasperDN: 👍

main/src/library/Fixpoint3/Phase/Hoisting.flix
```diff
@@ -0,0 +1,642 @@
...
+///         query {b ∈ B | b[0] = a[0]} do
+///             query {c ∈ C | c[0] = a[0]} do
+///                 project (a[0]) into R
+///             end
+///         end
+///     end
+/// end
+/// ```
+///
+/// One hard dependency of the current implementation is that for `BoolExp.Eq(x, y)` `x` and `y` can
```
JonathanStarup: must either be .. sounds more suitable to me unless I misunderstand

main/src/library/Fixpoint3/Phase/Hoisting.flix
```diff
@@ -0,0 +1,642 @@
...
+///             if(b[0] = a[0] ∧ c[0] = a[0], b[1] = 1, b[2] = c[1]) then
+///                 project (a[0]) into R
+///             end
+///         end
+///     end
+/// end
+/// ```
+/// `equalitySets` could be `{a[0] => a[0], b[0] => a[0], c[0] => a[0], b[1] => b[1], b[2] => c[1], c[1] => c[1]}`.
+///
+/// Since `equalitySets` is only used to get representatives for loads the exact value of the representative
```
JonathanStarup: for loads, the

?

main/src/library/Fixpoint3/Phase/Hoisting.flix
```diff
@@ -0,0 +1,642 @@
...
+///
+/// `termMap` will be initialized with `{b[1] => 1}`.
+///
+/// When arriving at `search b ∈ B` `termMap` will be `{b[1] => 1, a[0] => a[0]}`.
+///
+/// When arriving at `search c ∈ C` `termMap` will be `{b[1] => 1, a[0] => a[0], c[1] => b[2]}`.
+///
+/// When arriving at `if(...)` and `project(...)` `termMap` will remain unchanged.
+///
+mod Fixpoint3.Phase.Hoisting {
```
JonathanStarup, line 85: could be nice if we had `@Internal` modules

main/src/library/Fixpoint3/Phase/Hoisting.flix
```diff
@@ -0,0 +1,642 @@
...
+                case Some(hoistedStmt) => RamProgram.Program(hoistedStmt, facts, meta, index)
+            }
+        }
+    }
+
+    ///
+    /// Hoist `If` ops in `stmt` and merge them with `Search` ops when relevant.
+    ///
+    /// `termMap` is used to look up which term a given `(RowVar, Int32)` is represented by and
+    /// whether they are currently considered ground.
```
JonathanStarup: Could you expand a little on what "currently considered" means?

main/src/library/Fixpoint3/Phase/Hoisting.flix
```diff
@@ -0,0 +1,642 @@
...
+    ///
+    /// `impossible` contains the set of `RowVar`'s/rules which have been determined impossible to satisfy.
+    ///
+    def hoistStmt(
+        rc: Region[r],
+        termMap: MutMap[(RowVar, Int32), RamTerm, r],
+        equalitySets: MutDisjointSets[(RowVar, Int32), r],
+        impossible: Set[RowVar],
+        stmt: RamStmt
+    ): Option[RamStmt] \ r = match stmt {
```
JonathanStarup, line 147: I think you should explain when None is returned

main/src/library/Fixpoint3/Phase/Hoisting.flix
```diff
@@ -0,0 +1,642 @@
...
+    ///
+    /// `termMap` maps loads, e.g. `x[0]`, to a literals or other loads, bound higher in the AST.
+    /// More specifically the representing load for `x[0]` will be used, where representing stems from
+    /// `equalitySets`.
+    ///
+    /// `equalitySets` maps loads, e.g. `x[0]`, to a load that represents a transitively equivalent RowLoad
+    /// If we have x[0] = x[1] = y[4] then looking up any of these will consistently return the same value/load.
+    ///
+    /// `impossible` is a set of `RowVar` that are involved in unsatisfying conditions. If they are encountered
+    /// `None` can be returned immediately.
```
JonathanStarup: "can be" is also a bit odd here. will it? or only sometimes?

JonathanStarup: its also explained at the bottom, but "will" is better i think

main/src/library/Fixpoint3/Phase/Hoisting.flix
```diff
@@ -0,0 +1,642 @@
...
+    ///
+    def hoistOp(
+        rc: Region[r],
+        termMap: MutMap[(RowVar, Int32), RamTerm, r],
+        equalitySets: MutDisjointSets[(RowVar, Int32), r],
+        impossible: Set[RowVar],
+        idToBool: MutMap[RowVar, MutList[BoolExp, r], r],
+        independentBool: MutList[BoolExp, r],
+        op: RelOp
+    ): Option[RelOp] \ r =
```
JonathanStarup, line 250: This is a very long and nested function. Is there any parts that could be extracted to make it more readable?

main/src/library/Fixpoint3/Phase/Hoisting.flix
CasperDN: Agreed

main/src/library/Fixpoint3/Phase/Hoisting.flix
```diff
@@ -0,0 +1,642 @@
...
+    ///
+    /// Ignores `Eq`, as `Eq`'s are seen as a special case and will always be a comparison between loads and
+    /// loads/constants.
+    ///
+    /// As an example `Guard2(x[0], y[3])` will return `{x}` if `termMap` is e.g. `[y[3] => 8]`.
+    /// If `termMap` contained `y[3]=> z[3]` then `{x, z}` will be returned.
+    ///
+    /// The purpose is to return the `RowVar`'s in the AST which are possible places `bool` can be hoisted to.
+    ///
+    def collectBoolVariables(
```
JonathanStarup, line 455: I don't know how hot-path this function is, but It'll probably be faster in tail-recursive form (with a work list and accumulator)

main/src/library/Fixpoint3/Phase/Hoisting.flix
```diff
@@ -0,0 +1,642 @@
...
+    }
+
+    ///
+    /// Replace terms in `bool` by their representative.
+    ///
+    /// In particular `RowLoad(rv, i, _)` will be replaced by termMap[equalitySet[(rv, i)]].
+    /// This ensures that we all loads refer to variables as far up in the AST as possible,
+    /// and therefore variables which are still bound after hoisting.
+    ///
+    /// `Meet` is excluded from this treatment.
```
JonathanStarup: maybe this could be explained a bit?

main/src/library/Fixpoint3/Phase/Hoisting.flix
CasperDN: I have broken it into a couple of pieces. Is it more readable now?

main/src/library/Fixpoint3/Phase/Hoisting.flix
CasperDN: It is called once for every guard in the program. It should not be that hot-path.

The function itself is tail-recursive, right? It only has one recursive call...

Most of the other calls are to `collectTermVariables`.

main/src/library/Fixpoint3/Phase/Hoisting.flix
CasperDN: Better?

main/src/library/Fixpoint3/Phase/Hoisting.flix
CasperDN: Better?

magnus-madsen:
@JonathanStarup Ping me when happy

main/src/library/Fixpoint3/Phase/Hoisting.flix
JonathanStarup: > It is called once for every guard in the program. It should not be that hot-path.

Right, clearer code is probably more important here

> The function itself is tail-recursive, right? It only has one recursive call...

Yea, i suppose its more aimed at collectTermVariables

But yeah, makes sense to keep it as is

JonathanStarup:
@magnus-madsen 🏓

JumpTwiice:
Im fixing punctuation and some typos in comments

PR #11030, opened 2025-07-03
fix: remove dependency on determinism in old Datalog engine

CasperDN:
The tests were hard-coded to use the paths the old engine found.

They have here been generalized to accept any shortest paths that exist in the graphs.

The usage of `Assert.eq(true, shortestPathFound)` is unfortunate, as errors are hard to track, but I am unaware of a better way to assert this.

### DIFF
--- a/examples/larger-examples/datalog/single-source-shortest-paths.flix
+++ b/examples/larger-examples/datalog/single-source-shortest-paths.flix
@@ -265,24 +265,93 @@ mod ShortestPath {
 
     @Test
     def testExampleGraphShortestPaths01(): Bool =
-        Assert.eq(Map#{1 => Vector#{0, 1}, 2 => Vector#{0, 1, 2}, 3 => Vector#{0, 3}, 4 => Vector#{0, 3, 4}, 5 => Vector#{0, 3, 4, 5}}, sssp(0, exampleGraph01()))
+        sssp(0, exampleGraph01()) |> 
+        Map.forAll(k -> v -> match k {
+            case 1 => Assert.eq(Vector#{0, 1}, v)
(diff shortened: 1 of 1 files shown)


examples/larger-examples/datalog/single-source-shortest-paths.flix
```diff
@@ -265,24 +265,93 @@ mod ShortestPath {
...
+            case 2 => Assert.eq(Vector#{0, 1, 2}, v)
+            case 3 => Assert.eq(Vector#{0, 3}, v)
+            case 4 =>
+                let shortestPathFound = v == Vector#{0, 3, 4} or v == Vector#{0, 1, 4};
+                Assert.eq(true, shortestPathFound)
+            case 5 =>
+                let shortestPathFound = v == Vector#{0, 1, 2, 5} or v == Vector#{0, 3, 4, 5} or v == Vector#{0, 1, 4, 5};
+                Assert.eq(true, shortestPathFound)
+            case _ => unreachable!()
+        })
```
magnus-madsen, line 280: How many shortest paths are there? Would it be possible to enumerate them and then just check that the found path is in the set of shortest paths?

I kind of assume this is what you are doing, but instead you are using a somewhat complex Map.forAll

CasperDN: The code finds the shortest path from 0 to every other node. It then checks that the path from 0 to e.g. 1 is X, from 0 to 2 is Y and so on.

magnus-madsen: Ah. I see.

magnus-madsen: I suppose its OK, if a bit overwhelming.

PR #11032, opened 2025-07-03
chore: add `Bool.runIf`

### LINKED ISSUE #11008: `Bool.runIf`
```
def runIf(f: Unit -> Unit \ ef, cond: Bool): Unit \ ef = if (cond) f() else ()
```

The applications are

```
l |> List.map(..) |> List.forAll(..) |> runIf(..)
```

wizard7377:
Fixes #11008

### DIFF
--- a/main/src/library/Bool.flix
+++ b/main/src/library/Bool.flix
@@ -60,7 +60,7 @@ mod Bool {
     pub def <==>(x: Bool, y: Bool): Bool = x == y
 
     ///
-    /// Guards the side effect `f` if `cond` is true.
+    /// Calls `f` if `cond` is true.
     ///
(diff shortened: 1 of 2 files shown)


magnus-madsen:
If you could add 3 tests (every standard library function must have at least 3 tests).

main/src/library/Bool.flix
```diff
@@ -59,4 +59,9 @@ mod Bool {
     ///
     pub def <==>(x: Bool, y: Bool): Bool = x == y
 
+    ///
+    /// Guards the side effect `f` if `cond` is true.
```
JonathanStarup: This language seems a bit non-standard to me. I'd just say ```Calls `f` if `cond` is true.``` or refer to `ef` as the side effect, not `f`

Issue #8439, opened 2024-08-26
add `Iterable[String]` instance


JonathanStarup:

magnus-madsen:
Strings being viewed as a sequence of chars is considered a mistake by many. 
I think we need to fully explore the issue before we commit to this.

JonathanStarup:
You're right, but the choice has already been made in terms of the standard library. Its a bit weird to draw the line now. 
```
def charAt(i: Int32, s: String): Char
def dropWhile(f: Char -> Bool, s: String): String
def enumerator(rc: Region[r], s: String): Iterator[(Int32, Char), r, r] \ r
def exists(f: Char -> Bool \ ef, s: String): Bool \ ef
def findIndexOfLeft(f: Char -> Bool \ ef, s: String): Option[Int32] \ ef
def foldLeft(f: b -> (Char -> b \ ef), x: b, s: String): b \ ef
def forAll(f: Char -> Bool \ ef, s: String): Bool \ ef
def init(f: Int32 -> Char \ ef, len: Int32): String \ ef
def iterator(rc: Region[r], s: String): Iterator[Char, r, r] \ r
def map(f: Char -> Char \ ef, s: String): String \ ef
def nth(i: Int32, s: String): Option[Char]
def toList(s: String): List[Char]
```

magnus-madsen:
Hmmm... My plan is still to enforce that Strings are not containers. But then you have me, and it seems that I should remove some of the above functions. So I suppose I will do that.

PR #11033, opened 2025-07-03
chore: make `String` instance of `Iterable`

wizard7377:
#8439


magnus-madsen:
See discussion in https://github.com/flix/flix/issues/8439

magnus-madsen:
We continue discussion first.

PR #11035, opened 2025-07-04
feat: extend `FixpointSolve` in ASTs with `SolveMode` 

JumpTwiice:
See https://github.com/flix/flix/issues/11022

Issue #11022, opened 2025-07-02
Add `psolve` expression


magnus-madsen:
- [ ] PR1: Add `TokenKind.KeywordPSolve`
- [ ] PR2: (a) Extend `Parser2.fixpointSolveExpr` to parse both `solve` and `psolve` (I suggest to pass a Boolean) and (b) Extend `Weeder2.visitFixpointSolveExpr` to support both `solve` and `psolve`.
- [ ] PR3: Extend all ASTs and phases all the way to `Lowering` to work with something like:
```scala
case class FixpointSolve(exp: Expr, mode: SolveMode, loc: SourceLocation) extends Expr

sealed trait SolveMode // in ast.shared package
object SolveMode{
  case object Default extends SolveMode
  case object WithProvenance extends SolveMode
}
```
- [ ] Extend `Lowering` to pass a Boolean depending on the provenance mode.
- [ ] (Later) Add provenance support to the Datalog engine

magnus-madsen:
@CasperDN @JumpTwiice A proposal for a path towards `psolve` (-- thats the easy one, `pquery` will be more challenging.)

@JumpTwiice I can assign you the ticket when you join the flix org.

magnus-madsen:
Remember to add to `KeywordCompleter`

magnus-madsen:
@JumpTwiice We still need to enable this in the frontend right?

JumpTwiice:
Already enabled: https://github.com/flix/flix/commit/7c959664e1c7d210144f1fa972f821c6f49a85e7

magnus-madsen:
Okay. I just got some errors, but maybe I was doing it wrong. I will let you add some tests.

magnus-madsen:
Just needs tests

magnus-madsen:
Fixed.

PR #11037, opened 2025-07-04
feat: add unimplemented `runSolverWithProvenance` to Datalog solver and match `SolveMode` in `Lowering`

JumpTwiice:
This allows `Lowering` to handle the `solveMode` in the pattern-match in `visitExp` for when https://github.com/flix/flix/pull/11035 is merged.

See https://github.com/flix/flix/issues/11022.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
@@ -46,6 +46,7 @@ object Lowering {
     val version: String = ""
     lazy val Box: Symbol.DefnSym = Symbol.mkDefnSym(s"Fixpoint${version}.Boxable.box")
     lazy val Solve: Symbol.DefnSym = Symbol.mkDefnSym(s"Fixpoint${version}.Solver.runSolver")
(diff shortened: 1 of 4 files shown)


main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -785,7 +786,11 @@ object Lowering {
       LoweredAst.Expr.ApplyDef(defn.sym, argExps, Types.MergeType, resultType, eff, loc)
 
     case TypedAst.Expr.FixpointSolve(exp, _, eff, loc) =>
-      val defn = Defs.lookup(Defs.Solve)
+      val solveMode: SolveMode = SolveMode.Default
```
magnus-madsen: solveMode -> mode 
(and drop the type annotation)

JumpTwiice: The annotation is an upcast from SolveMode.Default => SolveMode. Scala won't compile without it due to the pattern match right below.

JumpTwiice: It is temporary anyway since we get it from the `Expr.FixpointSolve` patmatch when https://github.com/flix/flix/pull/11035 is merged. I will make another quick PR when that happens.

magnus-madsen:
@JumpTwiice Conflicted but otherwise ready.

PR #11040, opened 2025-07-05
feat: Add "Go to definition" for stdlib

saviorand:
Adds the "Go to definition" functionality in the LSP for stdlib code. 
This is done via extracting the sources to a given location with a gradle task and pointing to respective files in the sources.

Pasting the message with more details from gitter:

> On this one, I checked how Scala Metals [does it](https://github.com/scalameta/metals/blob/c9f074fac10084f2e04dc814f72255796c8013ae/metals/src/main/scala/scala/meta/internal/metals/MetalsLspService.scala#L1647), looks like when you click on Go To Defintion in a stdlib construct they open it in the sources (that also means sources should be readily available), e.g
`file:/Users/myuser/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-library/2.13.16/scala-library-2.13.16-sources.jar!/scala/Predef.scala`
> For Flix that path would look different, but the approach should work as well. 
I tried adjusting Flix's LSP code and the gradle task to extract sources and look them up for stdlib files, see PR here:
> We will also need to change [this glob](https://github.com/flix/vscode-flix/blob/465e587ec00fdf9f786f814cc9dc4cee3859fb7d/client/src/extension.ts#L37C1-L40C2) in the VSCode extension to add a new directory where we extract the sources to, something like /.flix or /.flix-stdlib .
> Currently going to definition works in my PR but it refuses to e.g syntax highlight in the stdlib source files because of this rule in the extension code.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -128,7 +128,19 @@ task testIDECompletion(dependsOn: ['testClasses'], type: JavaExec) {
 
 test.dependsOn testAll
 
-task vscode(type: Copy, dependsOn: [clean, jar]) {
+task extractStdlibSources(type: Copy) {
+    description = 'Extracts Flix standard library source files'
+
+    from 'main'
+    include '**/*.flix'
+    into project.findProperty('flix.stdlib.output') ?: 'build/stdlib-sources'
+
+    doLast {
+        println "Flix stdlib sources extracted to: ${destinationDir}"
+    }
+}
+
+task vscode(type: Copy, dependsOn: [clean, jar, extractStdlibSources]) {
     from 'build/libs/flix.jar'
     into providers.gradleProperty('dev.flix.vscode.project')
 }
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/GotoProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/GotoProvider.scala
@@ -36,7 +36,95 @@ object GotoProvider {
 
     gotoRight
       .orElse(gotoLeft)
-      .filter(_.targetUri.startsWith("file://")) // We do not support goto for non-file URIs, which is the case for the standard library.
+      .map(convertToValidUri)
+  }
+
+  /**
+   * Converts non-file URIs to file URIs by searching for the corresponding source file.
+   * If the URI is already a file URI, returns it unchanged.
+   * If the URI points to a standard library file that cannot be found as a regular file,
+   * attempts to locate it in the standard library sources directory.
+   */
+  private def convertToValidUri(link: LocationLink): LocationLink = {
+    if (link.targetUri.startsWith("file://")) {
+      link
+    } else {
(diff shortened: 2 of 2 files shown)


magnus-madsen:
Hello @saviorand and thanks for the PR. I agree the feature would be useful, but I wonder about a more general solution:

- A solution that would work for any Flix package (e.g. a zip-file .fpkg file containing Flix sources)
- A solution hat is more robust in that it does not change the build
- A solution that can give VSCode support (e.g. semantic tokens). (Note: Reloading the library as you propose would trigger duplicate definition errors everywhere).

My suggestion would be some mechanism to tell VSCode to open a "virtual path" with some internal name that we can map onto the Flix standard library (and external packages). Is that possible?

magnus-madsen:
Maybe something like:

https://code.visualstudio.com/api/extension-guides/virtual-documents
https://github.com/microsoft/vscode-extension-samples/tree/main/virtual-document-sample

magnus-madsen:
I think here:

https://github.com/flix/vscode-flix/blob/master/server/src/handlers/handlers.ts#L205

We could add some infrastructure to load a virtual document

magnus-madsen:
Or perhaps we can simply refer to the file inside the zip/jar/fpkg like: `jar:file:/path/to/my.zip!/data/document.txt`

saviorand:
Right, I wanted to just access .jar directly but couldn't figure it out initially. Was able to do this now. 
The issue was that VSCode doesn't support `jar:` URIs natively, and also cannot read into ZIP file contents (there's an open issue [here](https://github.com/microsoft/vscode/issues/75559)). So I to make some changes to the VSCode plugin, and now it works for me locally.  Opened a new PR here:
https://github.com/flix/vscode-flix/pull/492

Also reverted the changes in this PR, now it just removes the filter. We might make this a bit more secure but using a `flixstdlib://` custom scheme I defined in the other PR doesn't seem to work, the actual paths I'm seeing in VSCode just show up as e.g `Prelude.flix` , so without any prefixes 🤔 

Another open point is I don't see any syntax highlight on the stdlib files that open now when I go to the definition. Not sure how to fix that one, could that also be just a restriction somewhere on the language server side?

In any case, thanks for the feedback! Hope we can finish this one and get that functionality merged. Would be really nice

magnus-madsen:
I have suggestion. What if we give the library files (and later package files) some special name:

```
virtualname://List.flix (or whatever)
```

On the VSCode side we can intercept these names and open them. 

Ideally VSCode then sends back some SemanticTokenProvider request for the above virtual path and we can respond do it normally. 

You can check Flix.scala-- thats where the library files get their names.

Could you explore this idea and report back? Thanks!

saviorand:
@magnus-madsen thanks, I'll probably look into this later this week. Also saw the comments in the other PR. Will let you know how it goes

Issue #10938, opened 2025-06-22
Introduce `SocketAddr`


magnus-madsen:
https://doc.rust-lang.org/beta/std/net/enum.SocketAddr.html

CadeMichael:
following the Rust implementation

- Enum with 2 members
  - SocketAddrV4
  - SocketAddrV6
- Functions
  - `parse_ascii`
  - `ip`
  - `set_ip` instead of mutation probably return new socket with passed in ip
  - `port`
  - then trait implementations Eq, ToString, FromString

magnus-madsen:
I think we want the `ip`, `port`, and `parse` functions. The rest can wait.

PR #11042, opened 2025-07-05
feat: Add `SocketAddr` to standard library, first push

CadeMichael:
Addresses #10938.

### DIFF
--- a/main/src/library/SocketAddr.flix
+++ b/main/src/library/SocketAddr.flix
@@ -1,0 +1,70 @@
+/*
+ *  Copyright 2025 Cade Lueker
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+///
+/// Represents a Socket Address.
+///
+enum SocketAddr with Eq {
+    case V4(SocketAddrV4)
+    case V6(SocketAddrV6)
+}
+
+instance FromString[SocketAddr] {
+    pub def fromString(x: String): Option[SocketAddr] = SocketAddr.parse_ascii(x)
+}
+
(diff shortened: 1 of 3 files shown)


CadeMichael:
If this looks alright I'll add some tests. I don't know if `parse_ascii` should be `from_string` but I followed the rust api naming.

magnus-madsen:
> If this looks alright I'll add some tests. I don't know if `parse_ascii` should be `from_string` but I followed the rust api naming.

I think it should be FromString

main/src/library/SocketAddr.flix
```diff
@@ -0,0 +1,70 @@
...
+
+    use SocketAddrV4.SocketAddrV4
+    use SocketAddrV6.SocketAddrV6
+
+    ///
+    /// Attempts to parse the given String `x` as a `SocketAddr`.
+    ///
+    /// Returns `Some(socket)` if the string was successfully parsed. Otherwise returns `None`.
+    ///
+    pub def parse_ascii(x: String): Option[SocketAddr] =
```
magnus-madsen: fromString?

magnus-madsen:
Looks good. You can add some tests 👍

main/src/library/SocketAddr.flix
CadeMichael: got it i'll make the change

CadeMichael:
> Looks good. You can add some tests 👍

is this a good way to test port and Ip?

```
    @Test
    def socketAddrPortIp0(): Bool =
        match SocketAddr.fromString("192.168.1.1:8080") {
            case Some(s) => {
                let ip = SocketAddr.ip(s);
                let port = SocketAddr.port(s);
                port == 8080 and "V4(192.168.1.1)" == "${ip}"
            }
            case None => false
        }
```

CadeMichael:
I was able to get it to parse ipv4 and ipv6, for v6 the ip must be surrounded by [], using this invariant I check for occurances of "]", if it exists it can't be a v4 and should be a v6 so i remove the first char of the string, split on "]:" and then process using `IpAddr.fromString` 

```
    def socketAddrPortIp1(): Bool =
        match SocketAddr.fromString("[2001:db8::1]:443") {
            case Some(s) => {
                let ip = SocketAddr.ip(s);
                let port = SocketAddr.port(s);
                port == 443 and "V6(2001:0DB8:0000:0000:0000:0000:0000:0001)" == "${ip}"
            }
            case None => false
        }
```

magnus-madsen:
Looks reasonable.

magnus-madsen (approved):

magnus-madsen:
I merged this. Next PRs should be:

- tests
- using SocketAddr in the library

PR #11051, opened 2025-07-07
feat: add Boxing

CasperDN:


CasperDN:
@magnus-madsen Provenance is threaded through this program more intricately than the other parts we have submitted, I have therefore left some of it in the program. This is primarily in the form of the `withProv` parameter. I can remove it if you deem it necessary, though it will complicate the process slightly.

main/src/library/Fixpoint3/Boxing.flix
```diff
@@ -0,0 +1,778 @@
...
+        // need to be extended by RamId(-1).
+        let nonExtendedPositions = Vector.range(0, arityOf(relSym)) |>
+            Vector.map(i -> getOrCrash(Map.get(RamId.RelPos(relSymId, i), mapping)));
+        let positions =
+            if (not withProv) nonExtendedPositions
+            else Vector.append(nonExtendedPositions, 
+                Vector#{getOrCrash(Map.get(RamId.Id(-1), mapping)), getOrCrash(Map.get(RamId.Id(-1), mapping))});
+        let threadNum = threads();
+        let resultArr = Array.empty(rc1, numOfFacts);
+        // Map the facts in parallel, unless there are relatively few.
```
magnus-madsen, line 176: So you have some parallelism even before Datalog evaluation begins?

magnus-madsen:
The maps that end up holding the values, when do they get deallocated by the GC? Is that when Datalog computation ends?

main/src/library/Fixpoint3/Boxing.flix
CasperDN: Yes. We observed that it took some time to go through all the facts sequentially. As mentioned in a comment this part of the program will be heavily simplified later on.

CasperDN:
If the map is `RamIdToPos` then yes. It is needed in the last stage of the interpreter.

main/src/library/Fixpoint3/Boxing.flix
```diff
@@ -0,0 +1,778 @@
...
+///
+/// Consider the following program:
+/// ```
+/// A(x, y) :- B(x), C(y).
+/// ```
+/// We know that A[0] and B[0] should share the same value, and if we project the
+/// representing `Int64` from `B` to `A` then we should be able to compute the
+/// `Boxed` value. We therefore unify the 2, effectively declaring that the types
+/// are the same. The same is done for `A[1]` and `C[0]`. On the other hand unifying
+/// `A[0]` and `A[1]` is not secure, as it could mix 2 unrelating types.
```
JonathanStarup: "mix 2 unrelating types" -> unrelated

main/src/library/Fixpoint3/Boxing.flix
```diff
@@ -0,0 +1,778 @@
...
+        spawnFuncs(funcsToBeRun);
+        let newBoxing = fixBoxingAfterInit(rc1, counters, boxingInfo);
+        (Array.toMap(newFacts), newBoxing)
+    }
+
+    ///
+    /// Returns a correct `Boxing` constructed from `counters` and `boxingInfo`.
+    ///
+    /// After the simple initialization `boxingInfo` is missing the map back from `Int64`
+    /// to `Boxed`. This simply creates the it from the old `boxingInfo`.
```
JonathanStarup: "This simply creates the it from the old `boxingInfo`"?

main/src/library/Fixpoint3/Boxing.flix
```diff
@@ -0,0 +1,778 @@
...
+        let newMutLists = newBoxedArr |> Vector.map(Array.toMutList(rc));
+        (newMutLists, boxToRepVec, typeInfo, locks)
+    }
+
+    ///
+    /// Spawns all functions in `funcs` and await their termination.
+    ///
+    def spawnFuncs(funcs: Vector[Unit -> Unit \ ef]): Unit \ ef = region rc2 {
+        unchecked_cast({
+            let _: Unit = spawn unchecked_cast(
```
JonathanStarup: This function just spawns one thread and runs them all there right? Should be `forEach` and then `spawn` inside?

main/src/library/Fixpoint3/Boxing.flix
```diff
@@ -0,0 +1,778 @@
...
+        relSym: RelSym,
+        facts: Vector[Vector[Boxed]],
+        boxingInfo: Boxing[r],
+        counters: Vector[AtomicCounter[r]],
+        mapping: RamIdToPos,
+        withProv: Bool
+    ): Vector[(Vector[Int64], Boxed)] \ r = {
+        let relSymId = Ram.toId(relSym);
+        let numOfFacts = Vector.length(facts);
+        // We need to test for whether there are provenance facts. If `numOfFacts > 0` and
```
JonathanStarup: "for whether" -> whether

main/src/library/Fixpoint3/Boxing.flix
```diff
@@ -0,0 +1,778 @@
...
+        // `arity != Vector.length(facts[0])` then there are provenance facts and positions
+        // need to be extended by RamId(-1).
```
JonathanStarup: need -> that need

main/src/library/Fixpoint3/Boxing.flix
```diff
@@ -0,0 +1,778 @@
...
+            case Types.TyFloat32   => BoxedFloat32(Float.intBitsToFloat(getOrCrash(Int64.tryToInt32(v))))
+            case Types.TyFloat64   => BoxedFloat64(Double.longBitsToDouble(v))
+            case Types.TyObject    => deMarshalObject(v, index, info)
+            case Types.TyUnknown   => bug!("Unormalizing value, which has never been normalized")
+        }
+
+    ///
+    /// Construct a map from `RamId` to `UnifiedTypePos`.
+    ///
+    /// See the description of the module.
```
JonathanStarup: of the Boxing module?

main/src/library/Fixpoint3/Boxing.flix
```diff
@@ -0,0 +1,778 @@
...
+            let insert = registerRamId(set, map, counter);
+            match op {
+                case RelOp.Search(rv, RelSym.Symbol(PredSym.PredSym(_, predSym), arity, _), body) =>
+                    let usedArity = getProvSafeIdArity(arity, withProv);
+                    Vector.forEach(i -> insert(RamId.TuplePos(rv, i)), Vector.range(0, usedArity));
+                    Vector.forEach(i -> insert(RamId.RelPos(predSym, i)), Vector.range(0, usedArity));
+                    recurse(body)
+                case RelOp.Query(rv, RelSym.Symbol(PredSym.PredSym(_, predSym), arity, _), bools, _, body) =>
+                    let usedArity = getProvSafeIdArity(arity, withProv);
+                    Vector.forEach(i -> insert(RamId.TuplePos(rv, i)), Vector.range(0, usedArity));
```
JonathanStarup: Usually I find that doing iteration like this looks nice but is quite a bit slower (cache stuff perhaps?)

Could these foreach calls be merged into one for each case?

main/src/library/Fixpoint3/Boxing.flix
CasperDN: Yes

main/src/library/Fixpoint3/Boxing.flix
magnus-madsen: What would be slow here?

CasperDN: The lists are probably of size <6, but sure.

magnus-madsen: Then I don't  think it matters.

magnus-madsen:
@CasperDN Ping me when I should take another look.

CasperDN:
@magnus-madsen You can take another look now.

Issue #11023, opened 2025-07-02
Add `pquery` expression


magnus-madsen:
- [x] PR1: Add `TokenKind.KeywordPQuery`
- [x] PR2: Add `TreeKind.FixpointPQuery`
- [x] PR3: Add `WeededAst.Expr.FixpointPQuery`:
```scala
// pquery mm select Path("Toronto", "Chicago") with {Road , Bridge}
case class FixpointPQuery(exps: List[Expr], select: Predicate.Head, with: List[Name.Pred], loc: SourceLocation) extends Expr
```
- [x] PR4: Add support for `WeededAst.Expr.FixpointPQuery` to `Parser2` and `Weeder2`.
- [x] PR5: Add `FixpointPQuery` to all ASTs and phase up to `ConstraintGen`.
- [ ] ...

magnus-madsen:
Remember to add to `KeywordCompleter`

magnus-madsen:
Given:

```
pquery db, pr select Path(1, 3) from Edge, Path
```

The hard part left is now for `Lowering` to generate a function like (in pseudo code):

```
def mkExtVar(name: String, tuple: Vector[Boxed]): <ExtVar row type> = match name {
  case "Edge/2" => xvar Edge(unbox(tuple[0]), unbox(tuple[1]))
  case "Path/2" => xvar Path(...)
  case _ => bug!(...)
}
```

The idea is that this function can then be passed to the Datalog engine and used to construct the path.

JumpTwiice:
I think this can be closed for now.

magnus-madsen:
Fixed in master.

PR #11054, opened 2025-07-07
feat: add support for `pquery` in `Parser2` and `Weeder2`

JumpTwiice:

I tested `pquery` in a program by adding it to the lexer, and it mostly works fine but causes a lot of crashing when it's not written out fully.  There is also an issue with the predicate symbols in the `with { ... }` part; it will accept anything but an `Ident` and I can't figure out why.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -2947,7 +2947,7 @@ object Parser2 {
         case TokenKind.CurlyL => zeroOrMore(
           namedTokenSet = NamedTokenSet.FromKinds(NAME_PREDICATE),
           getItem = () => nameUnqualified(NAME_PREDICATE),
-          checkForItem = _.isPredicateName,
+          checkForItem = NAME_PREDICATE.contains,
           breakWhen = _.isRecoverExpr,
           delimiterL = TokenKind.CurlyL,
           delimiterR = TokenKind.CurlyR
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -1679,6 +1679,7 @@ object Parser2 {
         case TokenKind.KeywordPSolve => fixpointSolveExpr(isPSolve = true)
         case TokenKind.HashCurlyL => fixpointConstraintSetExpr()
         case TokenKind.HashParenL => fixpointLambdaExpr()
+        case TokenKind.KeywordPQuery => fixpointPQueryExpr()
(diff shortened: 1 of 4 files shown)


magnus-madsen:
I suggest we:

- Look at the code together.
- Add some tests (even if the tests cannot be merged yet-- they will be valuable later).

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -2914,6 +2915,54 @@ object Parser2 {
...
+    private def fixpointPQueryWith()(implicit s: State): Mark.Closed = {
+      implicit val sctx: SyntacticContext = SyntacticContext.Expr.OtherExpr
+      assert(at(TokenKind.KeywordWith))
+      val mark = open()
+      expect(TokenKind.KeywordWith)
+      nth(0) match {
+        case TokenKind.CurlyL => zeroOrMore(
+          namedTokenSet = NamedTokenSet.FromKinds(NAME_PREDICATE),
+          getItem = () => nameUnqualified(NAME_PREDICATE),
+          checkForItem = _.isPredicateName,
```
magnus-madsen: This should be NAME_PREDICATE.contains

magnus-madsen:
Code looks OK. What's the example where it goes wrong?

magnus-madsen:
Code looks OK. What's the example where it goes wrong?

JumpTwiice:
Any `pqquery` exp with a nonempty `with` part. In the example below, I get an `Expected at least one <Ident>` error.
```
let p = #{ }
pquery p select Path(1, 2) with { Path }
```
To actually test an example, add `case _ if isKeyword("pquery") => TokenKind.KeywordPQuery` to `scanToken` in `Lexer.scala`

magnus-madsen:
@JumpTwiice Here you go:

```scala
    private def visitFixpointQueryWithProvenanceExpr(tree: Tree)(implicit sctx: SharedContext): Validation[Expr, CompilationMessage] = {
      expect(tree, TreeKind.Expr.FixpointQueryWithProvenance)
      val expressions = traverse(pickAll(TreeKind.Expr.Expr, tree))(visitExpr)
      val select = flatMapN(pick(TreeKind.Expr.FixpointSelect, tree))(Predicates.pickHead)
      val withh = mapN(pick(TreeKind.Expr.FixpointWith, tree))(
        withTree => pickAll(TreeKind.Ident, withTree).map(tokenToIdent)
      )
      mapN(expressions, select, withh) {
        (expressions, select, withh) =>
          Expr.FixpointQueryWithProvenance(expressions, select, withh.map(Name.mkPred), tree.loc)
      }
    }
```

JumpTwiice:
@magnus-madsen should be ready now

Issue #10997, opened 2025-06-30
Extend Diagnostic with DiagnosticRelatedInformation


magnus-madsen:
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnosticRelatedInformation

magnus-madsen:
@gaganchandan Did you have a chance to look into this? Need any help?

gaganchandan:
> [@gaganchandan](https://github.com/gaganchandan) Did you have a chance to look into this? Need any help?

I have spent a little time on it. However I'm not sure how to extract related info from `CompilationMessage` seeing as it currently only has a single location and message. Shall I open a PR with a sort of skeleton of the extension that I have so far and we can continue from there?

magnus-madsen:
> > [@gaganchandan](https://github.com/gaganchandan) Did you have a chance to look into this? Need any help?
> 
> I have spent a little time on it. However I'm not sure how to extract related info from `CompilationMessage` seeing as it currently only has a single location and message. Shall I open a PR with a sort of skeleton of the extension that I have so far and we can continue from there?

Please do open a draft PR.

I think we will extend `CompilationMessage` with `locs: List[SourceLocation] = Nil` and then go from there. Makes sense?

magnus-madsen:
Fixed in master.

PR #11055, opened 2025-07-08
feat: extend Diagnostic with DiagnosticRelatedInformation

gaganchandan:
Implementation for #10997.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/DiagnosticRelatedInformation.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/DiagnosticRelatedInformation.scala
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2025 Gagan Chandan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


main/src/ca/uwaterloo/flix/api/lsp/DiagnosticRelatedInformation.scala
```diff
@@ -0,0 +1,37 @@
...
+/**
+  * Represents `DiagnosticRelatedInformation` in LSP.
+  */
+
+sealed trait DiagnosticRelatedInformation {
+  def toLsp4j: lsp4j.DiagnosticRelatedInformation
+}
+
+object DiagnosticRelatedInformation {
+  case class DiagnosticRelatedInformation(location: lsp4j.Location, message: String) {
```
magnus-madsen: When there is just one type, we dont need the trait-object thing. Just a single case class will suffice.

magnus-madsen (approved):

magnus-madsen:
Next, add `locs: List[SourceLocation]` to `CompilationMessage`

PR #11057, opened 2025-07-08
feat: extend ASTs with `FixpointQueryWithProvenance`

JumpTwiice:

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/DesugaredAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/DesugaredAst.scala
@@ -205,6 +205,8 @@ object DesugaredAst {
 
     case class FixpointMerge(exp1: Expr, exp2: Expr, loc: SourceLocation) extends Expr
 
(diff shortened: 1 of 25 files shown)


main/src/ca/uwaterloo/flix/language/ast/DesugaredAst.scala
```diff
@@ -205,6 +205,8 @@ object DesugaredAst {
 
     case class FixpointMerge(exp1: Expr, exp2: Expr, loc: SourceLocation) extends Expr
 
+    case class FixpointQueryWithProvenance(exps: List[Expr], select: Predicate.Head, withh: List[Name.Pred], loc: SourceLocation) extends Expr
```
magnus-madsen, line 208: Do we need to call it `withh` here? is `with` reserved in Scala?

magnus-madsen (approved):

main/src/ca/uwaterloo/flix/language/ast/DesugaredAst.scala
JumpTwiice: 2x yes.

magnus-madsen:
@JumpTwiice I suggest - for fun - you give it an hour or two with the typing. After that I can join in.

Issue #10976, opened 2025-06-27
Rename `Reader` to `Readable` and give it an associated type


rywiese:
https://github.com/flix/flix/blob/master/main/src/library/Reader.flix should look like:

```
trait Readable[t] {

    type Aty

    type Aef: Eff

    pub def read(buffer: Array[Readable.Aty[t], r], reader: t): Result[IoError, Int32] \ r + Readable.Aef[t]

}
```

rywiese:
Should I call it something other than `Aty`? Maybe `Elm`? `Elm` didn't seem quite right since it's not a collection. Open to suggestions.

rywiese:
It seems this is blocked by https://github.com/flix/flix/issues/10965 as well.

Because this change ripples into [BufReader](https://github.com/flix/flix/blob/6bed995a9ffcf4e330b5169b688a77a0a5153340/main/src/library/BufReader.flix#L38). I naively changed this to:
```
pub struct BufReader[t: Type, r: Region] {
    // buffer: Array[Int8, r],
    buffer: Array[Reader.Aty[t], r],
    mut cursor: Int32,
    mut size: Int32,
    reader: t
}
```

which crashed the compiler actually.

So the solution I think is to add a type variable:

```
pub struct BufReader[a, Type, t: Type, r: Region] {
    // buffer: Array[Int8, r],
    // buffer: Array[Reader.Aty[t], r],
    buffer: Array[a, r],
    mut cursor: Int32,
    mut size: Int32,
    reader: t
}
```

which now forces:
```
instance Readable[BufReader[a, t, r] with Readable[t] where Readable.Aty[t] ~ a { ... }
```

magnus-madsen:
> Should I call it something other than `Aty`? Maybe `Elm`? `Elm` didn't seem quite right since it's not a collection. Open to suggestions.

Yeah. I think Aty is too abstract.  There is already some concern about "Aef". Maybe Elm is okish?

magnus-madsen:
> which crashed the compiler actually.

Can you make a separate ticket with the crash? Thanks

magnus-madsen:
I think documenting the members (so associated type/effect and signature) is important to help understand the trait.

magnus-madsen:
@mlutze We still require the trait name prefix here:

```
    pub def read(buffer: Array[Readable.Aty[t], r], reader: t): Result[IoError, Int32] \ r + Readable.Aef[t]
```

right? One cannot yet write just Aef?

[...]

PR #11071, opened 2025-07-09
feat: rename `Reader` to `Readable` and add an associated type

rywiese:
fixes https://github.com/flix/flix/issues/10976

### DIFF
--- a/main/src/library/BufReader.flix
+++ b/main/src/library/BufReader.flix
@@ -15,38 +15,40 @@
  */
 
 ///
-/// A `Reader` implementation that wraps an underlying `Reader` instance
-/// and caches the bytes read from the underlying instance in an internal buffer.
+/// A `Readable` implementation that wraps an underlying `Readable` instance
+/// and stores the items read from the underlying instance in an internal buffer.
 ///
-/// A call to `read` will read directly from the buffer if bytes are available, and
+/// A call to `read` will read directly from the buffer if items are available, and
 /// will trigger a `read` from the underlying `reader` when the buffer is empty.
 ///
 /// This has the effect of reducing the overall number of `read`s to the underlying
-/// `Reader`.
+/// `Readable`.
 ///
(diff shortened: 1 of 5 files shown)


main/src/library/BufReader.flix
```diff
@@ -15,78 +15,80 @@
...
+        withCapacity(rc, defaultBufferCapacity(), defaultValue, reader)
 
     ///
-    /// Constructs a `BufReader` wrapping `reader` with the specified `capacity` in bytes.
+    /// Constructs a `BufReader` wrapping `reader` with the specified `capacity` in items.
     ///
     /// If `capacity <= 0`, the `BufReader` will delegate directly to `reader` without buffering.
     ///
-    pub def withCapacity(rc: Region[r], capacity: Int32, reader: t): BufReader[t, r] \ r =
+    pub def withCapacity(rc: Region[r], capacity: Int32, defaultValue: a, reader: t): BufReader[a, t, r] \ r with Readable[t] =
```
rywiese, line 69: The defaultValue doesn't seem ideal, i just need to initialize an empty array

magnus-madsen: Check out `Reflect` module

magnus-madsen: Or `MutList`

magnus-madsen:
This is almost just a pure refactor right?

rywiese:
> This is almost just a pure refactor right?

Yeah. Except adding the associated type. I called it `feat` instead of `chore` because it adds the associated type which is useful for adding things we don't have yet, but maybe calling it a refactor would be better.

main/src/library/BufReader.flix
magnus-madsen: Actually, you can just use Array.empty and pass the capacity

magnus-madsen: I will merge this, but please fix the Array.empty :-)

magnus-madsen (approved):

PR #11084, opened 2025-07-11
feat: add Debugging and Solver

CasperDN:


CasperDN:
The `Solver` is mostly just a copy paste from the old `Solver` while `Debugging` has seen some changes.

PR #11088, opened 2025-07-11
chore: use `Vector` in `injectIntoX` and `factsOf` and exclude tuples with lattice element bot

JumpTwiice:
Store tuples as `Vector` rather than first as a list followed by a `List.toVector`. Also exclude latticenal tuples mapping to bottom.

### DIFF
--- a/main/src/library/Fixpoint3/Solver.flix
+++ b/main/src/library/Fixpoint3/Solver.flix
@@ -19,6 +19,7 @@
 mod Fixpoint3.Solver {
     import dev.flix.runtime.Global
 
+    use Fixpoint3.{Boxed, Debugging, Interpreter, Options, Phase, Predicate, PredSymsOf, SubstitutePredSym }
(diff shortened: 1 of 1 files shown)


main/src/library/Fixpoint3/Solver.flix
```diff
@@ -741,17 +809,20 @@ mod Fixpoint3.Solver {
...
-                        let it = Iterator.range(rc, 0, Vector.length(tuple) + 1);
-                        let vs = it |> Iterator.map(i -> {
-                            if (i < Vector.length(tuple))
-                                Vector.get(i, tuple)
-                            else
-                                lat
-                        }) |> Iterator.toList;
-                        MutList.push(f(vs), pFacts)
+                        if (not leq(lat, bot)) {
+                            let arity = Vector.length(tuple);
```
magnus-madsen, line 816: Can vector.init be used here?

PR #11089, opened 2025-07-11
feat: desugar `FixpointQueryWithProject`

JumpTwiice:
Rewrite 
```
pquery p1, p2, p3 select ... with { ... }
```
to 
```
pquery p1 <+> p2 <+> p3 select ... with { ...},
```
The remaining ASTs then operate on a single exp.


magnus-madsen:
We use to rely a lot on desugarings-- it makes our life as compiler writers easier. 

Unfortunately it turns out that all of these rewrites are bad for tooling.

Hence the new approach is not to perform desugaring and over time our goal get rid of all desugarings.

Of course you had no way to know this.

TLDR: We don't need  this PR. It is fine as is.

magnus-madsen:
Instead simplifications should happen in Lowering.

PR #11094, opened 2025-07-13
feat: allow keyword escape in new java defs

JonathanStarup:
allows
```
new Runnable {
    def `run`(_this: Runnable): Unit \ IO = println("Cool!")
}
```

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -2742,7 +2742,11 @@ object Parser2 {
       assert(at(TokenKind.KeywordDef))
       val mark = open()
       expect(TokenKind.KeywordDef)
-      nameUnqualified(NAME_JAVA)
+      if (eat(TokenKind.InfixFunction)) {
+        ()
+      } else {
+        nameUnqualified(NAME_JAVA)
+      }
       Decl.parameters()
       expect(TokenKind.Colon)
       Type.typeAndEffect()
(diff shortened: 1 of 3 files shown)


JonathanStarup:
@mlutze probably outside of the realm of mergeable things without @magnus-madsen?

But man it really sucks not being able to construct `Runnable` objects, we need this asap imo

JonathanStarup:
!!!jar

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -2742,7 +2742,11 @@ object Parser2 {
       assert(at(TokenKind.KeywordDef))
       val mark = open()
       expect(TokenKind.KeywordDef)
-      nameUnqualified(NAME_JAVA)
+      if (eat(TokenKind.InfixFunction)) {
```
mlutze, line 2745: Seems like a hack to reuse this token, right?

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1966,14 +1966,27 @@ object Weeder2 {
...
+          case Token(TokenKind.InfixFunction, _, _, _, _, _) => true
+          case _ => false
+        } match {
+          case Some(token@Token(_, _, _, _, _, _)) =>
+            val n = token.text.stripPrefix("`").stripSuffix("`")
+            Name.Ident(n, token.mkSourceLocation())
+          case _ =>
+            // Parser always either gives Ident tree or infix token.
+            throw InternalCompilerException("Parser gave no Jvm Method Name", tree.loc)
+        }
```
mlutze, line 1981: This is just a collectFirst I think.

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
JonathanStarup: hmm i mean in `def `hello`` the lexing of hello is completely seperate from def, so it has to be the same as 
```
`Add.add`
```

PR #11098, opened 2025-07-14
test: remove redundant `Redundancy` test

jaschdoc:


mlutze:
In what sense is this redundant?

jaschdoc:
There is an identical test earlier in the match-related test. This just seems like a copy paste mistake

mlutze (approved):

Issue #11075, opened 2025-07-10
DisplayType should have its own types for extensible variants


magnus-madsen:
Here we reuse schema types, but we should introduce our own types and pretty print them differently:

```
   case TypeConstructor.Extensible =>
          val args = t.typeArguments.map(visit)
          args match {
            // Case 1: No args. { ? }
            case Nil => SchemaConstructor(Hole)
            // Case 2: One row argument. Extract its values.
            case tpe :: Nil => tpe match {
              case SchemaRow(fields) => Schema(fields)
              case SchemaRowExtend(fields, rest) => SchemaExtend(fields, rest)
              case nonSchema => SchemaConstructor(nonSchema)
            }
            // Case 3: Too many args. Error.
            case _ :: _ :: _ => throw new OverAppliedType(t.loc)
          }
```

magnus-madsen:
I reopen to discuss syntax

magnus-madsen:
We should probably also rename these:

- ExtSchemaConstructor
- ExtSchema
- ExtSchemaExtend
- ExtSchemaRow
- ExtSchemaRowExtend


@jaschdoc You introduce `ExtSchemaRow` and `ExtSchemaRowExtend` but they are not used. What are they?

jaschdoc:
I don't quite know yet. I added it because I copied over from the other schema types but I think they can be deleted since they are never constructed

PR #11099, opened 2025-07-14
feat: add Extensible `DisplayType`

jaschdoc:
Formats extensible types as `#|{ ... }|`. Notice the vertial bars `|`.

Closes https://github.com/flix/flix/issues/11075


mlutze (approved):
We can revisit what the best syntax is later.

Issue #11101, opened 2025-07-15
Add Peekable trait


rywiese:
Buffered readers give two benefits:
1. If you do lots of small reads (such as reading line by line), buffering is more efficient because you don't have to do IO on every read, only when you need to refill the buffer. This is a performance benefit that just makes the existing `Readable` interface more efficient.
2. You can look ahead at bytes that have been read from the underlying buffer but not yet read from the buffer. This enables conditional reading, with functions like `readWhile`/`readUntil`. This is useful for reading lines, where you scan through the buffer looking for a newline and then stop. This is new functionality not possible to implement with the existing `Readable` interface alone, as `Readable.read` requires you to specify the buffer size in advance.

Both of these benefits enable reading lines. One is a performance benefit, one is new functionality. The purpose of this ticket is to write a new trait that serves as the interface for such buffered readers that provides this new functionality.

I propose calling it `Peekable` because, from an API perspective, being able to `peek` is what I see as the new functionality enabled by buffering (in addition to performance benefit). I'm open to naming suggestions though. Rust calls this [BufRead](https://doc.rust-lang.org/std/io/trait.BufRead.html), and I'd be fine naming it that instead. It just seems to me like peeking/conditional reading is the "what" and buffering is the "how".

In theory, I think `readWhile` could be implemented in terms of `peek`/`skip`. Therefore I think that, conceptually, `readWhile` belongs here. In practice, such a default implementation of `readWhile` would be very ineficient so I propose not having it, and forcing implementers to write their own. But if they can `peek` and `skip`, they should be able to `readWhile` also.

```
trait Peekable[t] with Readable[t] {

    pub def peek(r: t): Result[IoError, Option[Readable.Elm[t]]] \ Readable.Aef[t]

    pub def skip(n: Int32, r: t): Result[IoError, Int32] \ Readable.Aef[t]

    pub def readWhile(p: Readable.Elm[t] -> Bool, r: t): Result[IoError, Vector[Readable.Elm[t]]] \ Readable.Aef[t]

}
```

`BufReader` should be an instance of this.

rywiese:
Maybe `ConditionallyReadable`?

Issue #10895, opened 2025-06-16
Decide on return value for Reader read


rywiese:
There was a discussion on [this PR](https://github.com/flix/flix/pull/10886/files#r2146622267) around whether `Reader read(b: Array[Int8, r], r: InputStream): Result[IoError, Int32]` (which returns number of bytes read) should return 0 or -1 on EOF. The [rust version](https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read)  returns 0, which can mean EOF or that the provided array is empty. The [java version](https://docs.oracle.com/javase/8/docs/api/java/io/Reader.html#read-char:A-) returns -1 for EOF and 0 for empty array.

Question is what we should do in the Flix reader. **Currently**, we specify return value of 0 for both cases, as they both indicate 0 bytes read (with no error) (like the Rust version).

This issue can track the decision as well as the code change (if we decide to make one).

One thing I noticed is that the only [abstract read method](https://docs.oracle.com/javase/8/docs/api/java/io/Reader.html#read-char:A-int-int-) in Java Reader has signature:

```
read(char[] cbuf, int off, int len)
```

This method is public also, which to me _suggests_ (tho i could be way off base) that this API is designed around a use case of a loop that repeatedly `read`s into the same array, incrementing the offset each time. Such a loop would break at EOF or at array full (sub-array size 0) and proceed accordingly, and in such a loop it would be convenient to case over the result of the read to determine next steps. `read(char[] cbuf)` is just a convenience method (and naturally has the same return value contract as the offset version).

[Rust's abstract read method](https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read), similar to flix's, doesn't do this offset business:
```
read(&mut self, buf: &mut [[u8](https://doc.rust-lang.org/std/primitive.u8.html)])
```

so it is maybe more reasonable to avoid adding complexity to the return value by just expecting the client to be responsible for checking array size if it's relevant.

With that in mind,
* I'm inclined to leave it as-is (0 for both cases).
* If we decide (now or at some point in the future) that we do want to make this distinction, I'd be inclined to try and capture this distinction in the return type, rather than expecting clients and instances to read the documentation carefully enough to consider both 0 and -1.

Maybe something like:
```
pub def read(b: Array[Int8, r], r: t): Result[IoError, Option[Int32]] \ {r, Reader.Aef[t]}
```

JonathanStarup:
[...]

PR #11103, opened 2025-07-15
feat: add Peekable trait and instance for BufReader

rywiese:
fixes https://github.com/flix/flix/issues/11101

As I mention in the ticket, i'm open to suggestions on the naming of things

### DIFF
--- a/main/src/library/BufReader.flix
+++ b/main/src/library/BufReader.flix
@@ -53,6 +53,16 @@ instance Readable[BufReader[a, t, rr]] with Readable[t] where Readable.Elm[t] ~ 
 
 }
 
+instance Peekable[BufReader[a, t, r]] with Readable[t] where Readable.Elm[t] ~ a {
+
+    pub def peek(bufReader: BufReader[a, t, r]): Result[IoError, Option[Readable.Elm[t]]] \ Readable.Aef[t] + r = BufReader.peek(bufReader)
+
+    pub def skip(n: Int32, bufReader: BufReader[a, t, r]): Result[IoError, Int32] \ Readable.Aef[t] + r = BufReader.skip(n, bufReader)
+
+    pub def readWhile(p: Readable.Elm[t] -> Bool, bufReader: BufReader[a, t, r]): Result[IoError, Vector[Readable.Elm[t]]] \ Readable.Aef[t] + r = BufReader.readWhile(p, bufReader)
+
+}
+
 mod BufReader {
 
     ///
@@ -89,6 +99,131 @@ mod BufReader {
             readFromCursor(0, dest, bufReader)
         } else {
             Readable.read(dest, bufReader->reader)
+        }
+
+    ///
+    /// Returns `Ok(Some(x))` where x is the next value to be read, or `Ok(None)` if the `bufReader` is empty.
+    /// Does not advance the `bufReader`.
+    ///
+    /// Returns `Err(err)` if there is an IO error.
+    ///
+    pub def peek(bufReader: BufReader[a, t, r]): Result[IoError, Option[Readable.Elm[t]]] \ Readable.Aef[t] + r with Readable[t] where Readable.Elm[t] ~ a =
+        if (bufReader->size != 0) {
+            Ok(Some(Array.get(bufReader->cursor, bufReader->buffer)))
+        } else {
+            forM(
+                itemsFilled <- fillBuffer(bufReader)
+            ) yield {
+                if (itemsFilled != 0) {
+                    Some(Array.get(bufReader->cursor, bufReader->buffer))
+                } else {
+                    None
+                }
+            }
+        }
+
+    ///
+    /// Advances the reader by `n`, skipping over the elements without reading them.
+    ///
+    /// Returns `Ok(k)` where `k` is the number of items skipped.
+    ///
+    /// Guarantees that `0 <= k <= max(n, 0)`. If `k < n`, then EOF has been reached.
+    ///
+    /// Returns `Err(err)` if there is an IO error.
+    ///
+    pub def skip(n: Int32, bufReader: BufReader[a, t, r]): Result[IoError, Int32] \ Readable.Aef[t] + r with Readable[t] where Readable.Elm[t] ~ a =
+        if (n <= 0) {
+            // Done
+            Ok(0)
+        } else if (bufReader->size != 0) {
+            // Nonempty Buffer: process items
+            let itemsSkipped = skipFromBuffer(n, bufReader);
+
+            // recurse into Done or Empty Buffer branch
+            forM(
+                rest <- skip(n - itemsSkipped, bufReader)
+            ) yield itemsSkipped + rest
+        } else {
+            // Empty Buffer: refill it
+            forM(
+                itemsFilled <- fillBuffer(bufReader);
+                skipped <- if (itemsFilled == 0) {
+                    // Done: Eof
+                    Ok(0)
+                } else {
+                    // Nonempty Buffer: continue processing
+                    skip(n, bufReader)
+                }
+            ) yield skipped
+        }
+
+    ///
+    /// Advances the reader by `n` or the size of the buffer, whichever is smaller, skipping over the elements without reading them.
+    ///
+    /// If the buffer runs out, does NOT trigger a read to the underlying reader.
+    ///
+    /// Guarantees that `0 <= k <= min(n, size(buffer))`.
+    ///
+    /// Assumes `n > 0`.
+    ///
+    def skipFromBuffer(n: Int32, bufReader: BufReader[a, t, r]): Int32 \ r with Readable[t] where Readable.Elm[t] ~ a =
+        let itemsToSkip = Int32.min(n, bufReader->size);
+        bufReader->cursor = bufReader->cursor + itemsToSkip;
+        bufReader->size = bufReader->size - itemsToSkip;
+        itemsToSkip
+
+    ///
+    /// Reads from `reader` until an element is reached that does not satisfy predicate `p`.
+    ///
+    /// Returns `Ok(vec)` where `vec` is a vector of the first `k` elements where `k` is the index of the first element that does not satisfy `p`.
+    /// This element is not included in the vector, and will be read on the next read.
+    ///
+    /// All elements of `vec` are guaranteed to satisfy `p`.
+    ///
+    /// Returns `Err(err)` if there is an IO error.
+    ///
+    pub def readWhile(p: a -> Bool, bufReader: BufReader[a, t, r]): Result[IoError, Vector[a]] \ Readable.Aef[t] + r with Readable[t] where Readable.Elm[t] ~ a =
+        region rc {
+            let mutList = MutList.empty(rc);
+            forM(
+                _ <- readWhileInto(rc, p, mutList, bufReader)
+            ) yield mutList |> MutList.toVector
+        }
+
+    ///
+    /// Like `readWhile`, but appends results into `results` and returns `Ok(())` on success.
+    ///
+    def readWhileInto(rc: Region[rl], p: a -> Bool, results: MutList[a, rl], bufReader: BufReader[a, t, rb]): Result[IoError, Unit] \ Readable.Aef[t] + rb + rl with Readable[t] where Readable.Elm[t] ~ a =
+        if (bufReader->size != 0) {
+            ///
+            /// Finds the index of the first element in the buffer that does not match predicate `p`.
+            /// If all elements in the buffer match, returns `i + 1` where `i = bufReader->cursor + bufReader->size`
+            /// is the last element in the logical buffer.
+            ///
+            def findIndex(i) = {
+                if (i < bufReader->cursor + bufReader->size) {
+                    let x = Array.get(i, bufReader->buffer);
+                    if (p(x)) findIndex(i + 1) else i
+                } else {
+                    i
+                }
+            };
+            let i = findIndex(bufReader->cursor);
+            if (i == bufReader->cursor) {
+                Ok(())
+            } else {
+                let vec = Array.slice(rc, start = bufReader->cursor, end = i, bufReader->buffer) |> Array.toVector;
+                MutList.append(vec, results);
+                let itemsRead = Vector.length(vec);
+                bufReader->cursor = bufReader->cursor + itemsRead;
+                bufReader->size = bufReader->size - itemsRead;
+                readWhileInto(rc, p, results, bufReader)
+            }
+        } else {
+            forM(
+                itemsFilled <- fillBuffer(bufReader);
+                unit <- if (itemsFilled == 0) Ok(()) else readWhileInto(rc, p, results, bufReader)
+            ) yield unit
         }
 
     ///
--- a/main/src/library/Peekable.flix
+++ b/main/src/library/Peekable.flix
@@ -1,0 +1,66 @@
+/*
+ *  Copyright 2025 Ry Wiese
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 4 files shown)


main/src/library/BufReader.flix
```diff
@@ -91,6 +101,90 @@ mod BufReader {
             Readable.read(dest, bufReader->reader)
         }
 
+    ///
+    /// Returns `Ok(Some(x))` where x is the next value to be read, or `Ok(None)` if the `reader` is empty.
+    /// Does not read `x` from the `reader`.
```
mlutze: Maybe "does not advance the reader"

main/src/library/BufReader.flix
```diff
@@ -91,6 +101,90 @@ mod BufReader {
...
+    ///
+    /// Returns `Err(err)` if there is an IO error.
+    ///
+    pub def peek(bufReader: BufReader[a, t, r]): Result[IoError, Option[Readable.Elm[t]]] \ Readable.Aef[t] + r with Readable[t] where Readable.Elm[t] ~ a =
+        if (bufReader->size != 0) {
+            Ok(Some(Array.get(bufReader->cursor, bufReader->buffer)))
```
mlutze, line 112: What happens if the cursor is at the end of the buffer? Is that possible?

main/src/library/BufReader.flix
```diff
@@ -91,6 +101,90 @@ mod BufReader {
...
+        } else {
+            forM(
+                itemsFilled <- fillBuffer(bufReader)
+            ) yield if (itemsFilled != 0) {
+                Some(Array.get(bufReader->cursor, bufReader->buffer))
+            } else {
+                None
+            }
```
mlutze: Formatting: put the if in another brace. Right now it looks like the `else` is connected to the `forM`

main/src/library/BufReader.flix
```diff
@@ -91,6 +101,90 @@ mod BufReader {
...
+        }
+
+    ///
+    /// Returns `Ok(k)` where `k` is the number of items skipped.
+    ///
+    /// Guarantees that `0 <= k <= n`.
```
mlutze: Technically false if e.g. `n = -1` :)

main/src/library/BufReader.flix
```diff
@@ -91,6 +101,90 @@ mod BufReader {
...
+                itemsFilled <- fillBuffer(bufReader)
+            ) yield if (itemsFilled != 0) {
+                Some(Array.get(bufReader->cursor, bufReader->buffer))
+            } else {
+                None
+            }
+        }
+
+    ///
+    /// Returns `Ok(k)` where `k` is the number of items skipped.
```
mlutze, line 128: Missing a line about what this actually does.

main/src/library/BufReader.flix
```diff
@@ -91,6 +101,90 @@ mod BufReader {
...
+    pub def skip(n: Int32, bufReader: BufReader[a, t, r]): Result[IoError, Int32] \ Readable.Aef[t] + r with Readable[t] where Readable.Elm[t] ~ a =
+        if (n <= 0) {
+            Ok(0)
+        } else if (bufReader->size != 0) {
+            let itemsToSkip = Int32.min(n, bufReader->size);
+            bufReader->cursor = bufReader->cursor + itemsToSkip;
+            bufReader->size = bufReader->size - itemsToSkip;
+            forM(
+                rest <- skip(n - itemsToSkip, bufReader)
+            ) yield itemsToSkip + rest
```
mlutze: Some comments here would be good. I'm confused about why we recurse. I guess that forces taking the `else` branch the second time?

main/src/library/BufReader.flix
```diff
@@ -91,6 +101,90 @@ mod BufReader {
...
+        }
+
+    ///
+    /// Returns `Ok(k)` where `k` is the number of items skipped.
+    ///
+    /// Guarantees that `0 <= k <= n`.
+    ///
+    /// Returns `Err(err)` if there is an IO error.
+    ///
+    pub def skip(n: Int32, bufReader: BufReader[a, t, r]): Result[IoError, Int32] \ Readable.Aef[t] + r with Readable[t] where Readable.Elm[t] ~ a =
```
mlutze, line 134: Does it mean anything in particular if `n != k`?

mlutze: As a user, do I just keep calling this in a loop until we've skipped as much as I want?

main/src/library/BufReader.flix
```diff
@@ -91,6 +101,90 @@ mod BufReader {
...
+    /// Returns `Ok(vec)` where `vec` is a vector of the first `k` elements where `k` is the index of the first element that does not satisfy `p`.
+    /// This element is not included in the vector, and will be read on the next read.
+    ///
+    /// All elements of `vec` are guaranteed to satisfy `p`.
+    ///
+    /// Returns `Err(err)` if there is an IO error.
+    ///
+    pub def readWhile(p: a -> Bool, bufReader: BufReader[a, t, r]): Result[IoError, Vector[a]] \ Readable.Aef[t] + r with Readable[t] where Readable.Elm[t] ~ a =
+        if (bufReader->size != 0) region rc {
+            def findIndex(i) = {
```
mlutze, line 203: Nice to abstract this. a quick doc comment would help here

main/src/library/Peekable.flix
```diff
@@ -0,0 +1,64 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+///
+/// Extends `Readable` with the abillity to `peek` at values that have not yet been `read`,
```
mlutze: s/abillity/ability/

main/src/library/Readable.flix
mlutze, line 1: Not relevant to this PR, right?

main/src/library/BufReader.flix
rywiese: The cursor never wraps back around. Every time we refill the buffer we set the cursor back to zero. So if the cursor is at the end of the buffer, then `bufReader->size == 0`

main/src/library/Readable.flix
rywiese: Right, i just forgot to rename the file in my last PR which renamed the actual trait. I can do this in a separate PR if that's preferred?

mlutze: Yes, I'm trying to enforce my "atomic PR" agenda :P

main/src/library/BufReader.flix
rywiese: Or the first if branch. Maybe I'll extract this branch into a `peekFromBuffer` function so that the logic is more clear.

That's kind of what I did for the `read` function as well, the structure is similar. It's basically mutual recursion between `fillBuffer` and `readFromBuffer`/`peekFromBuffer`, except that I pulled the recursive calls out of those functions so they could be more reusable.

main/src/library/Readable.flix
rywiese: https://github.com/flix/flix/pull/11117 will rebase this one

main/src/library/BufReader.flix
rywiese: Formally no, in practice it likely means EOF was hit after `k` skips. Though we don't specify any meaning to reads/skips where `0 < k < n`. 0 means EOF. Maybe my comments need to be more explicit about that part.

Probably not in a loop, since you pass in `n`. Maybe just check that you skipped the number you want; if not, try again with the remainder and if you get 0 you know to stop. My use case is actually just skipping one at a time (read until `\n` or `\r`, skip over it, peek for `\r` (windows newline is `\n\r`), skip over that, etc). But supporting `n` instead of just 1 seemed easy enough to implement.

main/src/library/BufReader.flix
```diff
@@ -91,6 +101,112 @@ mod BufReader {
             Readable.read(dest, bufReader->reader)
         }
 
+    ///
+    /// Returns `Ok(Some(x))` where x is the next value to be read, or `Ok(None)` if the `reader` is empty.
+    /// Does not advance the `reader`.
```
mlutze: `s/reader/bufReader/`

main/src/library/BufReader.flix
mlutze: > in practice it likely means EOF was hit after k skips

Likely or definitely?

Yes, more documentation here would help.

main/src/library/BufReader.flix
```diff
@@ -91,6 +101,112 @@ mod BufReader {
...
+        } else {
+            forM(
+                itemsFilled <- fillBuffer(bufReader);
+                skipped <- if (itemsFilled == 0) {
+                    Ok(0)
+                } else {
+                    skip(n, bufReader)
+                }
+            ) yield skipped
+        }
```
mlutze: Ok, I think I understand what's going on now:

1. We clear the buffer (up to the max)
2. We refill the buffer
3. We recurse with n - whatever we skipped so far

Maybe something smart about EOF needs to happen, but I think this will be clearer if we just make it those 3 steps. The "recurse to the other if-branch" is quite confusing.

main/src/library/BufReader.flix
```diff
@@ -91,6 +101,112 @@ mod BufReader {
...
+    ///
+    /// Advances the reader by `n` or the size of the buffer, whichever is smaller, skipping over the elements without reading them.
+    ///
+    /// If the buffer runs out, does NOT trigger a read to the underlying reader.
+    ///
+    /// Guarantees that `0 <= k <= min(n, size(buffer))`.
+    ///
+    /// Assumes `n > 0`.
+    ///
+    def skipFromBuffer(n: Int32, bufReader: BufReader[a, t, r]): Int32 \ r with Readable[t] where Readable.Elm[t] ~ a =
```
mlutze, line 169: This is really nice for simplifying the other function.

main/src/library/BufReader.flix
```diff
@@ -91,6 +101,112 @@ mod BufReader {
...
+            };
+            let i = findIndex(bufReader->cursor);
+            if (i == bufReader->cursor) {
+                Ok(Vector.empty())
+            } else {
+                let vec = Array.slice(rc, start = bufReader->cursor, end = i, bufReader->buffer) |> Array.toVector;
+                let itemsRead = Vector.length(vec);
+                bufReader->cursor = bufReader->cursor + itemsRead;
+                bufReader->size = bufReader->size - itemsRead;
+                forM(rest <- readWhile(p, bufReader)) yield Vector.append(vec, rest)
```
mlutze: `Vector.append` seems fishy. Doesn't that make it n-squared?

mlutze: Maybe it would be better to append to a mutable list and convert it at the end.

main/src/library/BufReader.flix
rywiese: Java `read` at the interface level returns the number of things read, but does not formally specify what it means if the number returned is less than the number requested.

Take for example the decoding reader from the other PR: you request `n` chars, `DecodingReader` reads some bytes from the underlying byte reader, decodes them to chars, returns them. But how many bytes should decoding reader ask for? Some chars are multibytes, so if we ask for `n` bytes, we may end up decoding `< n` chars. This would be an example of valid behavior where `0 < k < n` and we are not at EOF.

That being said, the actual behavior in that PR is to read `4 * n` bytes since there are at most 4 bytes per char, at least in utf8 (some get stored as leftovers). And I think java implementations make a best attempt to give you as many as you ask for, but it's not formally guaranteed. What if you have an obscure encoding with unbounded bytes per char? I just don't want to specify guarantees that java doesn't even specify

mlutze: Ok. I just thought it was strange that O always means EOF, but nonzero could sometimes mean other things.

rywiese: Yes. And java also makes this distinction where 0 means 0 bytes read and -1 means EOF. But rust does not do this, 0 means EOF. I have a ticket somewhere to maybe think harder about that but I couldn't come up with a strong enough reason to change the existing interface in Flix (which is that 0 means EOF)

main/src/library/BufReader.flix
rywiese: FYI `readFromCursor` is written in the same structure, but I can change that one too in a follow up if that's preferred, it would be nice for them to have consistent structure.

Something like this?
```
pub def skip(n: Int32, bufReader: BufReader[a, t, r]): Result[IoError, Int32] \ Readable.Aef[t] + r with Readable[t] where Readable.Elm[t] ~ a =
        if (n <= 0) {
            Ok(0)
        } else {
           let itemsSkipped = skipFromBuffer(n, bufReader);
           forM(
               rest <- if (bufReader->size == 0) {
                    forM(
                        itemsFilled <- fillBuffer(bufReader);
                        rest <- if (itemsFilled == 0) {
                            Ok(0)
                        } else {
                            skip(n - itemsSkipped, bufReader)
                        }
                    ) yield rest
               } else Ok(0)
           ) yield itemsSkipped + rest
        }
```

The recursing between if branches logic is a bit unconventional i guess but it has benefits:
* fewer nesting levels of if/forM
* skipping and filling/eof handling are separated into their own branches
* if we call skip and the buffer is empty, we enter that branch right away

rywiese: But I can change it to that snippet or something similar if you like that better :)

mlutze: Hmm, I thought it would be prettier. But I suppose maybe I was assuming some invariants about when you can skip less than you request, which Java doesn't seem to uphold.

Is there no way to avoid all the nesting? It's just a crazy state machine?

rywiese: > Is there no way to avoid all the nesting?

I don't think so. And i don't think the logic itself is all that crazy, it's just that we have two checks that need to happen (whether to refill the buffer, whether we hit EOF on the refill), but everything is monadic so there's `forM`s at every layer.

I spent some time trying to make it as pretty as possible in the first PR for this class with `readFromCursor`, and this is what I came up with 🤷. I guess you can think of it as a state machine that's either done/buffer-ready/buffer-empty, with each branch handling a case and then continuing recursively

rywiese: But `readFromCursor` and `readWhile` are also written in this way, is the problem with the if/else structure generally or is it less intuitive to understand with the skip logic specifically?

mlutze: Ok, I see that the complexity is just inherent to all these things dealing with the buffer. Maybe in another world we could represent EOF in the monad, but we shouldn't complicate the types here.

To help the reader understand what's going on, please add in-line comments for each branch. For example, `reading returned 0, which means XYZ...`

We should also be very clear in documentation about these arcane secrets.

rywiese: Okay I realized I misspoke about `k < n` wrt `skip`. Everything I said is true for `read`, which is java's contract. But in `skip` we read in a loop that breaks only when `n` or EOF is reached. So for skip, `k < n` does imply EOF which I will document here/on the trait (along with requested in-line comments).

For `read`, `k` can be less than `n` without EOF. So there's no way to restructure this code based on invariants. I will clarify the subtleties of the API in some comments in a separate PR. I will also add this as context in https://github.com/flix/flix/issues/10895, since it seems like there may be value in pursuing a richer return type for `read`.

rywiese: Done

main/src/library/BufReader.flix
rywiese: Done

mlutze (approved):
Seems reasonable. Maybe we can land on a better design for the types after some discussion.

Issue #10977, opened 2025-06-27
Add `DecodingReader`


rywiese:
This wraps a `byteReader` with `Readable[t]` where `Readable.Aty[t] ~ Int8`, and is an instance of `Readable` where the associated type is `Char`. It is an adapter from byte reader to char reader.

Note: this cannot be implemented without instance-level `where` constraints. This motivates and is blocked by https://github.com/flix/flix/issues/10965. But maybe we can start for now and figure out a short term workaround.

```
struct DecodingReader[t, r] {
    charset: Charset,
    byteReader: t
}

mod DecodingReader {

    pub def read(buffer: Array[Char, rb], reader: DecodingReader[t, rr]): Result[IoError, Int32] \ rb + Readable.Aef[t] + rr with Readable[t] where Readable.Aty[t] ~ Int8 = ???

}

instance Readable[DecodingReader[t, rr]] with Readable[t] where Readable.Aty[t] ~ Int8 {

    type Aty = Char

    type Aef = Readable.Aef[t] + rr

    pub def read(buffer: Array[Char, rb], reader: DecodingReader[t, rr]): Result[IoError, Int32] \ rb + Readable.Aef[t] + rr = DecodingReader.read(buffer, reader)

}
```

magnus-madsen:
Is DecodingReader an immutable struct? Perhaps it should be an enum then (or pure struct-- but we dont have those yet, @JonathanStarup )

JonathanStarup:
@magnus-madsen back at you :P I can do it if the syntax is decided https://github.com/flix/flix/issues/10856

rywiese:
> Is DecodingReader an immutable struct? Perhaps it should be an enum then (or pure struct-- but we dont have those yet, [@JonathanStarup](https://github.com/JonathanStarup) )

Yeah. So maybe for now something like:

```
enum DecodingReader[t]({charset = Charset, byteReader = t})
```
?

PR #11104, opened 2025-07-16
feat: add DecodingReader

rywiese:
fixes https://github.com/flix/flix/issues/10977

### DIFF
--- a/main/src/library/DecodingReader.flix
+++ b/main/src/library/DecodingReader.flix
@@ -1,0 +1,124 @@
+/*
+ *  Copyright 2025 Ry Wiese
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+import java.nio.charset.CharsetDecoder
+import java.nio.ByteBuffer
+import java.nio.CharBuffer
+
+///
+/// Adapts a `Readable` of bytes to a `Readable` of chars.
+///
+/// `byteReader`: the underlying byte reader, an instance of `Readable where Readable.Elm[t] ~ Int8`.
+/// `decoder`: used to decode bytes read from `byteReader`.
+/// `leftovers`: stores bytes that have been consumed from `byteReader` but have not yet been decoded;
+///              either because they are lead bytes whose continuation bytes have not been read, or
+///              because `DecodingReader` overestimated the number of bytes needed to fill a char buffer.
+///
+struct DecodingReader[t, r] {
+    rc: Region[r],
+    byteReader: t,
+    decoder: CharsetDecoder,
+    mut leftovers: Array[Int8, r]
+}
+
+instance Readable[DecodingReader[t, rr]] with Readable[t] where Readable.Elm[t] ~ Int8 {
+
+    type Elm = Char
+
+    type Aef = Readable.Aef[t] + IO + rr
+
+    pub def read(buffer: Array[Char, rb], reader: DecodingReader[t, rr]): Result[IoError, Int32] \ rb + rr + Readable.Aef[t] + IO = DecodingReader.read(buffer, reader)
+
+}
+
+mod DecodingReader {
+    import java.nio.charset.CharsetDecoder
+    import java.nio.charset.CodingErrorAction
+    import java.nio.ByteBuffer
+    import java.nio.CharBuffer
+
+    ///
+    /// Wraps `byteReader` in a `DecodingReader`.
+    ///
+    pub def wrap(rc: Region[rr], charset: CharacterSet, byteReader: t): DecodingReader[t, rr] \ rr + IO =
+        let CharacterSet.CharacterSet(cs) = charset;
+        new DecodingReader @ rc {
+            rc = rc,
+            decoder = cs.newDecoder().onMalformedInput(CodingErrorAction.REPLACE).onUnmappableCharacter(CodingErrorAction.REPLACE),
+            byteReader = byteReader,
+            leftovers = Array.empty(rc, 0)
+        }
+
+    ///
+    /// Reads bytes from the underlying byteReader and decodes the result into `dest`.
+    ///
+    /// Returns `Ok(k)` to signify that `k` chars were successfully read and written to `dest`.
+    ///
+    /// Guarantees that `0 <= k <= length(dest)`.
+    ///
+    /// Returns `Err(e)` if some underlying error occurs.
+    ///
+    pub def read(dest: Array[Char, rb], reader: DecodingReader[t, rr]): Result[IoError, Int32] \ IO + rr + rb + Readable.Aef[t] with Readable[t] where Readable.Elm[t] ~ Int8 =
+        forM(
+            (src, endOfInput) <- {
+                // At most 4 bytes per char, guarantees we can fill dest if there are enough bytes to do so
+                let totalBytesNeeded = 4 * Array.length(dest);
+                readBytes(totalBytesNeeded, reader)
+            };
+            charsRead <- {
+                let charsRead = decode(src, dest, endOfInput, reader);
+
+                // charsRead == 0 would imply end of input to the caller.
+                // block again until end of input or we have at least one char to return
+                if (not endOfInput and charsRead == 0) read(dest, reader) else Ok(charsRead)
+            }
+        ) yield charsRead
+
+    ///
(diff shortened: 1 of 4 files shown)


rywiese:
~~Moving to draft cause I need to rethink some things~~

done rethinking

main/src/library/DecodingReader.flix
```diff
@@ -0,0 +1,126 @@
...
+ *  limitations under the License.
+ */
+
+import java.nio.charset.CharsetDecoder
+import java.nio.ByteBuffer
+import java.nio.CharBuffer
+
+///
+/// An adapter that wraps a byteReader, which is an instance of `Int8` `Readable`, and is itself
+/// an instance of `Char` `Readable`. Uses the `decoder` to convert the read bytes to chars.
```
mlutze: This is a confusing sentence...

main/src/library/DecodingReader.flix
```diff
@@ -0,0 +1,126 @@
...
+                let charsRead = decode(src, dest, endOfInput, reader);
+
+                // charsRead == 0 would imply end of input to the caller.
+                // block again until end of input or we have at least one char to return
+                if (not endOfInput and charsRead == 0) read(dest, reader) else Ok(charsRead)
+            }
+        ) yield charsRead
+
+    ///
+    /// Returns `Ok(k, bytes)` where `bytes` is the leftover bytes from the previous read prepended to `k` bytes read from the underlying reader.
```
mlutze: This documentation seems out of date.

main/src/library/DecodingReader.flix
```diff
@@ -0,0 +1,126 @@
...
+    ///
+    /// The total number of bytes read (number of leftovers + k) is equal to
```
mlutze: And this line seems incomplete :)

main/src/library/DecodingReader.flix
```diff
@@ -0,0 +1,126 @@
...
+
+///
+/// An adapter that wraps a byteReader, which is an instance of `Int8` `Readable`, and is itself
+/// an instance of `Char` `Readable`. Uses the `decoder` to convert the read bytes to chars.
+///
+struct DecodingReader[t, r] {
+    rc: Region[r],
+    decoder: Decoder,
+    byteReader: t,
+    mut leftovers: Array[Int8, r]
```
mlutze, line 34: A comment about the point of this would be good.

main/src/library/DecodingReader.flix
```diff
@@ -0,0 +1,126 @@
...
+    /// Decodes as far as possible from the `ByteBuffer`, skipping over any malformed/unmappable bytes and decoding them as '�'.
+    ///
+    def decodeBuffer(byteBuffer: ByteBuffer, charBuffer: CharBuffer, endOfInput: Bool, reader: DecodingReader[t, r]): Unit \ IO with Readable[t] where Readable.Elm[t] ~ Int8 =
+        let Decoder.Decoder(decoder) = reader->decoder;
+        let result = decoder.decode(byteBuffer, charBuffer, endOfInput);
+        if (result.isError()) {
+            charBuffer.put('�');
+            byteBuffer.position(byteBuffer.position() + result.length());
+            decodeBuffer(byteBuffer, charBuffer, endOfInput, reader)
+        } else ()
```
mlutze: It seems like this stuff is handled by the Java Decoder itself: https://docs.oracle.com/javase/8/docs/api/java/nio/charset/CharsetDecoder.html#onMalformedInput-java.nio.charset.CodingErrorAction-

Is there a reason we need to do it?

rywiese: fixed but i had to tweak things a bit in order to customize the decoder without exposing java types in the api

rywiese: Ah it looks like I broke a community build because I added a java wrapper type for `Charset`, which was already defined in https://github.com/stephentetley/charset-locale

tagging @JonathanStarup cause I see that you've contributed to that repo 😎 

should i not be adding this type?

mlutze: Maybe we can promote Stephen's Charset to the standard library?

rywiese: Was wondering that too, but felt like i was missing some context on what that project is and why it's separate

mlutze: I will summon him to discuss.

mlutze: Actually, I guess we ought to wait for Magnus before we consider merging in a whole library.

Maybe for now we use the full name `CharacterSet`. Later we can decide whether to merge in Stephen's library.

main/src/library/DecodingReader.flix
```diff
@@ -0,0 +1,123 @@
...
+    import java.nio.CharBuffer
+
+    ///
+    /// Wraps `byteReader` in a `DecodingReader`.
+    ///
+    pub def wrap(rc: Region[rr], charset: Charset, byteReader: t): DecodingReader[t, rr] \ rr + IO =
+        let Charset.Charset(cs) = charset;
+        new DecodingReader @ rc {
+            rc = rc,
+            decoder = cs.newDecoder().onMalformedInput(CodingErrorAction.REPLACE),
```
mlutze: Should we also do `onUnmappableCharacter`?

main/src/library/DecodingReader.flix
```diff
@@ -0,0 +1,123 @@
...
+            (src, endOfInput) <- {
+                // At most 4 bytes per char, guarantees we can fill dest if there are enough bytes to do so
+                let totalBytesNeeded = 4 * Array.length(dest);
+                readBytes(totalBytesNeeded, reader)
+            };
+            charsRead <- {
+                let charsRead = decode(src, dest, endOfInput, reader);
+
+                // charsRead == 0 would imply end of input to the caller.
+                // block again until end of input or we have at least one char to return
```
mlutze, line 85: This is not a busy wait/block because the underlying reader is not a busy waiter, right?

main/src/library/DecodingReader.flix
```diff
@@ -0,0 +1,123 @@
...
+                if (not endOfInput and charsRead == 0) read(dest, reader) else Ok(charsRead)
+            }
+        ) yield charsRead
+
+    ///
+    /// Returns `Ok(bytes, endOfInput)` where `bytes` are the next bytes to be decoded.
+    ///
+    /// `bytes` contains `leftovers` from the last read prepended to new bytes read from the `byteReader` (if necessary).
+    ///
+    /// `length(bytes) == numBytes` if `endOfInput` is false. `length(bytes) <= numBytes` otherwise.
```
mlutze: What if I ask for 10 bytes but only 5 are left in the input?

main/src/library/DecodingReader.flix
```diff
@@ -0,0 +1,123 @@
...
+    ///
+    def readBytes(numBytes: Int32, reader: DecodingReader[t, r]): Result[IoError, (Array[Int8, r], Bool)] \ r + Readable.Aef[t] with Readable[t] where Readable.Elm[t] ~ Int8 =
+        let bytesToRead = numBytes - Array.length(reader->leftovers);
+        if (bytesToRead > 0) {
+            let src = Array.empty(reader->rc, bytesToRead);
+            forM(
+                bytesRead <- Readable.read(src, reader->byteReader)
+            ) yield {
+                let newBytes = Array.take(reader->rc, bytesRead, src);
+                let bytes = Array.append(reader->rc, reader->leftovers, newBytes);
```
mlutze, line 106: I wonder if there's something to be done about reusing existing arrays instead of these linear-tie operations.

But I don't think we should worry about that now.

main/src/library/DecodingReader.flix
rywiese: It can be, ex reading from an inputstream on a tcp connection will block until there's some data. This is probably a sign of needing a better return type; since we have to do extra reading in order to comply with the "0 means EOF" contract.

main/src/library/DecodingReader.flix
rywiese: I'll remove this comment and just say `0 <= length(bytes) <= numBytes`, since (as we've already discussed) I can't reason formally about exact number of bytes read from underlying reader.

But to answer your question, `endOfInput` would be false in that case because we read 5 bytes from the input. `endOfInput` means we hit EOF at the beginning of the read (ie read back 0).

If we return `(byteArray, true)` then byteArray came entirely from the leftovers, and I can add a comment to clarify that.

The `endOfInput` ultimately is passed into the decoder so that it knows to turn any straggler lead bytes into `�` instead of leaving them in the buffer waiting for continuation bytes. In your example, any such chars would be put back into the buffer as leftovers, since `endOfInput` is false. The next read would detect `endOfInput` as true and those chars would get converted then.

main/src/library/DecodingReader.flix
rywiese: Agreed... the `ByteBuffer` and `CharBuffer` in java are designed to be reused, and doing so would probably be the most "standard" way of solving this. So store a `ByteBuffer` and a `CharBuffer` on the struct. I didn't do this in the first pass because the caller could request to read into a buffer bigger than ours, in which case we'd have to reallocate (or find some way to deal with it) and I didn't want to solve that in the first pass.

mlutze (approved):

Issue #11074, opened 2025-07-10
Instance equality constraints followup


mlutze:
I think creating this rigidityenv should be in the logic of `getInstanceForType` instead of here.

_Originally posted by @mlutze in https://github.com/flix/flix/pull/11021#discussion_r2196687939_

PR #11110, opened 2025-07-18
chore: refactor rigidity env construction when checking super instances

rywiese:
fixes https://github.com/flix/flix/issues/11074

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Instances.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Instances.scala
@@ -19,7 +19,7 @@ package ca.uwaterloo.flix.language.phase
 import ca.uwaterloo.flix.api.Flix
 import ca.uwaterloo.flix.language.ast.ops.TypedAstOps
 import ca.uwaterloo.flix.language.ast.shared.SymUse.TraitSymUse
-import ca.uwaterloo.flix.language.ast.shared.{Instance, Scope, TraitConstraint}
+import ca.uwaterloo.flix.language.ast.shared.{EqualityConstraint, Instance, Scope, TraitConstraint}
 import ca.uwaterloo.flix.language.ast.{ChangeSet, RigidityEnv, Scheme, Symbol, Type, TypeConstructor, TypedAst}
 import ca.uwaterloo.flix.language.dbg.AstPrinter.DebugTypedAst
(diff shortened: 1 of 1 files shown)


magnus-madsen:
@mlutze You can review and merge if happy.

main/src/ca/uwaterloo/flix/language/phase/Instances.scala
```diff
@@ -210,11 +210,15 @@ object Instances {
...
     */
-  private def findInstanceForType(tpe: Type, trt: Symbol.TraitSym, root: TypedAst.Root, rigidityEnv: RigidityEnv)(implicit flix: Flix): Option[(Instance, Substitution)] = {
+  private def findInstanceForType(tpe: Type, tconstrs: List[TraitConstraint], econstrs: List[EqualityConstraint], trt: Symbol.TraitSym, root: TypedAst.Root)(implicit flix: Flix): Option[(Instance, Substitution)] = {
     val instOpt = root.traitEnv.getInstance(trt, tpe)
     // lazily find the instance whose type unifies and save the substitution
     instOpt.flatMap {
       superInst =>
+        // Rigidify sub-instance constraint vars in the substitution to ensure that they appear as-written in compiler errors
+        val evars = econstrs.flatMap(econstr => econstr.tpe1.typeVars ++ econstr.tpe2.typeVars).map(_.sym)
+        val tvars = tconstrs.flatMap(tconstr => tconstr.arg.typeVars).map(_.sym)
```
mlutze: I believe we don't actually have to extract these vars. The variables in the type constraints and equality constraints should be a subset of those in the type itself.

So I think we can just use `tpe.typeVars`.

mlutze: Or maybe instance has `tparams`? I don't remember if we added that :)

rywiese: > Or maybe instance has `tparams`? I don't remember if we added that :)

The shared instance does but the one at the call site does not.

rywiese: but tpe.typeVars works and [this error](https://github.com/flix/flix/pull/11021#discussion_r2194912186) that motivated the change still renders as expected

PR #11111, opened 2025-07-18
fix: invoke build command before building the jar

### LINKED ISSUE #11102: `:build-jar` creates an empty jar file if there is no `build/` directory
The `:build-jar` command yields to an empty jar file if it is invoked without prior building the project as stated in the documentation here https://doc.flix.dev/build.html . 
I suggest two approaches that might lead to a better usability:

1. Invoke the `build` process in case there is no existing `build/` directory.
2. Warn the user that an empty jar file was created

This must also be implemented for the `build-fatjar`.

I tend towards the first approach as it is cleaner from a user perspective. Feel free to add your ideas :)

gwydd12:
In this PR we find two important changes:
- Refactoring the `buildJar` and `buildFatJar` into a single function as both have the exact same structure only that `buildFatJar` includes dependencies. 
- This now simply invokes the build command before building the jar file. 

Of course, the refactoring is not essential for the fix and can be separated into another PR. However, I thought it would directly improve further development in the case we improve the build commands even further, maybe with the logic Matthew suggested in the issue. 

This comment raises a question: `Note: As the buildJar does the same, this build doesn't erase previous jar-file if existing.`. As they are not aligned in their behavior, it is unclear to me if the code is correct or the comment? I think the code has the correct behavior.

```scala
// Check whether it is safe to write to the file.
if (Files.exists(jarFile) && !Bootstrap.isJarFile(jarFile)) {
  return Validation.Failure(BootstrapError.FileError(s"The path '${formatter.red(jarFile.toString)}' exists and is not a jar-file. Refusing to overwrite."))
}
```
Thanks :) 

This PR fixes #11102.

### DIFF
--- a/main/src/ca/uwaterloo/flix/Main.scala
+++ b/main/src/ca/uwaterloo/flix/Main.scala
@@ -187,7 +187,7 @@ object Main {
             bootstrap =>
               val flix = new Flix().setFormatter(formatter)
               flix.setOptions(options.copy(loadClassFiles = false))
-              bootstrap.buildJar(flix, includeDependencies = true)
+              bootstrap.buildFatJar(flix)
           }.toResult match {
             case Result.Ok(_) =>
               System.exit(0)
--- a/main/src/ca/uwaterloo/flix/api/Bootstrap.scala
+++ b/main/src/ca/uwaterloo/flix/api/Bootstrap.scala
@@ -530,10 +530,8 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
 
   /**
     * Builds a jar package for the project.
-    * In the case the :fatjar command is invoked the inlcudeDependencies must be true to include all dependencies from
-    * the lib folder.
     */
-  def buildJar(flix: Flix, includeDependencies: Boolean = false)(implicit formatter: Formatter): Validation[Unit, BootstrapError] = {
+  def buildJar(flix: Flix)(implicit formatter: Formatter): Validation[Unit, BootstrapError] = {
     // Build the project before building the jar
     val buildResult = build(flix)
     buildResult match {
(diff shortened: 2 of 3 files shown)


main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -530,124 +530,83 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
...
-      // Add all resources, again sorting by relative file name
-      val resourcesDir = Bootstrap.getResourcesDirectory(projectPath)
-      for ((resource, fileNameWithSlashes) <- Bootstrap.getAllFilesSorted(resourcesDir)) {
-        Bootstrap.addToZip(zip, fileNameWithSlashes, resource)
-      }
-
-    } match {
-      case Success(()) => Validation.Success(())
-      case Failure(e) => Validation.Failure(BootstrapError.GeneralError(List(e.getMessage)))
+  def buildJar(flix: Flix, includeDependencies: Boolean = false)(implicit formatter: Formatter): Validation[Unit, BootstrapError] = {
```
mlutze: In order to keep the interface the same (1 Bootstrap function per command), I propose we have something like the following:

```scala
def buildJar(...) = buildJarHelper(..., includeDependencies = true)
def buildFatJar(...) = buildJarHelper(..., includeDependencies = false)
def buildJarHelper(..., includeDepenences: Boolean) = /* this function */

File: main/src/ca/uwaterloo/flix/api/Bootstrap.scala [hunk 2]
```diff
@@ -530,124 +530,83 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
...
-        Using(new ZipInputStream(Files.newInputStream(dep))) {
-          zipIn =>
-            var entry = zipIn.getNextEntry
-            while (entry != null) {
-              // Get the class files except module-info and META-INF classes which are specific to each library.
-              if (entry.getName.endsWith(".class") && !entry.getName.equals("module-info.class") && !entry.getName.contains("META-INF/")) {
-                // Write extracted class file to zip.
-                val classContent = zipIn.readAllBytes()
-                Bootstrap.addToZip(zipOut, entry.getName, classContent)
+      if(includeDependencies) {
```
mlutze: Space after `if`

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
mlutze: Ideally this refactoring should be in a separate PR, as you say :)

main/src/ca/uwaterloo/flix/Main.scala
```diff
@@ -170,7 +170,10 @@ object Main {
 
         case Command.BuildJar =>
           flatMapN(Bootstrap.bootstrap(cwd, options.githubToken)) {
-            bootstrap => bootstrap.buildJar()
+            bootstrap =>
+              val flix = new Flix().setFormatter(formatter)
+              flix.setOptions(options.copy(loadClassFiles = false))
```
mlutze, line 175: What is the reason for setting `loadClassFiles` to false?

gwydd12: Based on this comment in the `Options.scala`: `loads the generated class files into the JVM.`.
I do not think we need to load the class files into the JVM while building the jar file. Additionally, the `build` command sets the same option.

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
gwydd12: That makes total sense, I'll add this to the refactoring branch. I'll create a new PR to add the refactoring to avoid the cluttering, thanks :)

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
gwydd12: This is also noted for the refactoring branch, thank you :)

main/src/ca/uwaterloo/flix/Main.scala
JonathanStarup: Hmm yeah its a bit unexpected that `loadClassFiles = true` is the default.

But I'd also vote to handle this part in another PR/Issue

mlutze (approved):

magnus-madsen:
@gwydd12 Could you please update the documentation on doc.flix.dev-- thanks

PR #11113, opened 2025-07-19
refactor: add helper function for buildJar and buildFatJar

gwydd12:
The PR includes a refactoring of the `buildJar` and `buildFatJar` functions to merge overlapping functionalities. Therefore, the PR should simplify implementing future enhancements discussed in the issue #11102 .

This PR is discussed in another PR #11111.

EDIT:
This PR depends on the PR #11111.


main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -529,21 +529,33 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
   }
 
   /**
-    * Builds a jar package for the project.
+    * Builds the jar or the fatjar
     */
-  def buildJar()(implicit formatter: Formatter): Validation[Unit, BootstrapError] = {
+  private def buildJarBase(flix: Flix, includeDependencies: Boolean = false)(implicit formatter: Formatter): Validation[Unit, BootstrapError] = {
```
mlutze: Let's make `includeDependencies` non-default. We try to avoid default parameters in general.

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -568,90 +580,52 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
...
+  /**
+    * Builds a jar package for the project.
+    */
+  def buildJar(flix: Flix)(implicit formatter: Formatter): Validation[Unit, BootstrapError] = {
+    buildJarBase(flix)
+  }
+
   /**
     * Builds a fatjar package for the project.
     * This function relies essentially on the same pattern as used in the buildJar function.
```
mlutze: This documentation line doesn't make sense anymore since we don't have a copy-pasted function.

main/src/ca/uwaterloo/flix/Main.scala
```diff
@@ -170,7 +170,10 @@ object Main {
 
         case Command.BuildJar =>
           flatMapN(Bootstrap.bootstrap(cwd, options.githubToken)) {
-            bootstrap => bootstrap.buildJar()
+            bootstrap =>
```
mlutze: I think this file should be unchanged. I'm not sure what the implications of changing `loadClassFiles` are, so we should look into that separately.

gwydd12: If I understand correctly this file will be unchanged after we merge the PR #11111 into the master or atleast this was the idea. I've only added those changes for better reproducibility. And I agree that we need to discuss the `loadClassFiles` separately in another issue. Thanks :)

mlutze: Oh, I see! I thought we were doing this one first and then the other.

We often put "depends on #XYZ" in the PR description to indicate that another PR should be merged first.

gwydd12: Oh I'm sorry that makes definitely more sense. I thought the indication that this is a draft was my way of telling something like that, but makes definitely sense to write this in the description. Thanks for the information :)

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
gwydd12: This should now be resolved. We might want to add this information to the `STYLE.md`, might help someone at some point.

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
gwydd12: You are totally right. I'll removed the extra parts and reduced it to one line, hope this is correct now. :)

mlutze (approved):

mlutze:
Merged. Thanks!

PR #11115, opened 2025-07-20
chore: named argument in FloatX.pow

### LINKED ISSUE #7123: Improve `FloatX.pow`
`Float64.pow(x: Float64, n: Float64): Float64`

It should used named parameter

CasperDN:
Fixed #7123

### DIFF
--- a/main/src/library/Float32.flix
+++ b/main/src/library/Float32.flix
@@ -47,7 +47,7 @@ mod Float32 {
     ///
     /// Returns the maximum number representable by a `Float32`.
     ///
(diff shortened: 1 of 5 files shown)


mlutze:
The standard is that we leave the last argument unnamed. That helps us support partial application and especially piping.

There's also a question here about whether `pow` is regularly used infix, which means that we shouldn't name the arguments.

CasperDN:
> There's also a question here about whether `pow` is regularly used infix, which means that we shouldn't name the arguments.

I have no strong feelings with regards to the second. An argument for the change would be that we avoid code like
```2f32 + -2f32 `Float32.pow` 2f32```
While the operator precedence ensures that the result is as expected, `6.0`, it has nothing to do with the fact that `Float32.pow` is the power operator.

PR #11116, opened 2025-07-20
chore: List.groupBy should return List[Nel[a]]

### LINKED ISSUE #10298: `List.groupBy` should return `List[Nel[a]]`

CasperDN:
Fixes #10298

### DIFF
--- a/main/src/library/List.flix
+++ b/main/src/library/List.flix
@@ -887,35 +887,32 @@ mod List {
     ///
     /// The function `f` must be pure.
     ///
-    pub def groupBy(f: (a, a) -> Bool, l: List[a]): List[List[a]] =
-        def loop(ll, k) = match ll {
-            case Nil     => k(Nil)
+    pub def groupBy(f: (a, a) -> Bool, l: List[a]): List[Nel[a]] =
+        (Nil, l) ||> List.foldLeft(acc -> value -> {
+            groupByHelper(value, f, acc)
+        }) |> List.map(Nel.reverse)
+
+    ///
+    /// Handles the 2 cases:
+    ///
+    /// 1. `value` can be added to a list. `value` is added as the first value to the `Nel`.
+    /// 2. `value` cannot be added to any list. Returns `List.reverse(currentNels) ::: (Nel(value, Nil) :: Nil)`.
+    ///
+    /// In either case the created `Nel`'s are reversed.
+    ///
+    def groupByHelper(value: a, f: (a, a) -> Bool, currentNels: List[Nel[a]]): List[Nel[a]] =
+        // Record seen elements in reverse order in `acc`.
+        def loop(cur, acc) = match cur {
+            case Nil => (Nel.singleton(value) :: acc) |> List.reverse
             case x :: xs =>
-                let (r1, r2) = extractHelper(f, xs, x :: Nil);
-                loop(r2, ks -> k(r1 :: ks))
+                if (Nel.forAll(v -> f(value, v) and f(v, value), x)) {
+                    let newNel = Nel.cons(value, x);
+                    List.reverse(newNel :: acc) ::: xs
+                } else {
+                    loop(xs, x :: acc)
+                }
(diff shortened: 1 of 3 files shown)


mlutze:
@JonathanStarup this gets rid of an `?unreachable` in your code!

main/src/library/List.flix
```diff
@@ -887,35 +887,34 @@ mod List {
...
-    def agreeHelper(f: (a, a) -> Bool, a: a, l: List[a]): Bool = match l {
-        case Nil     => true
-        case x :: xs => if (f(x, a) and f(a, x)) agreeHelper(f, a, xs) else false
-    }
+    def groupByHelper(value: a, f: (a, a) -> Bool, currentNels: List[Nel[a]]): List[Nel[a]] =
+        // Record seen elements in reverse order in `acc`.
+        def loop(cur, acc) = match cur {
+            case Nil => (Nel.singleton(value) :: acc) |> List.reverse
+            case x :: xs =>
+                if (Nel.forAll(v -> f(value, v) and f(v, value), x)) {
```
mlutze, line 910: This looks suspicious to me. Why do we need to check all of the elements of the Nel? Can't we just check the head?

main/src/library/List.flix
```diff
@@ -887,35 +887,34 @@ mod List {
...
-    pub def groupBy(f: (a, a) -> Bool, l: List[a]): List[List[a]] =
-        def loop(ll, k) = match ll {
-            case Nil     => k(Nil)
-            case x :: xs =>
-                let (r1, r2) = extractHelper(f, xs, x :: Nil);
-                loop(r2, ks -> k(r1 :: ks))
-        };
-        loop(l, identity)
+    pub def groupBy(f: (a, a) -> Bool, l: List[a]): List[Nel[a]] =
+        (Nil, l) ||> List.foldLeft(acc -> value -> {
```
mlutze, line 891: There's a lot of reversing going on. Can we get rid of all the reversing by using a foldRight instead? Since `f` is pure, evaluation order does not matter.

main/src/library/List.flix
CasperDN: The function might be f(x,y)=x!=y

main/src/library/List.flix
CasperDN: If we simply wanted a groupBy I think all reverses could be removed. The documentation of groupBy is, at the moment, specific about the order of the result and so are the tests.

main/src/library/List.flix
mlutze: That's a good point. But is a wacky function to group by... I'm not sure it merits changing the time complexity of this function to support.

Haskell requires `f` to be an equivalence relation in its documentation: `https://hackage.haskell.org/package/base-4.10.1.0/docs/Data-List.html#v:groupBy`. We should probably just do the same.

main/src/library/List.flix
mlutze: Right. I'm not saying we change the order. I'm suggesting that if we traverse the list in reverse order (using foldRight), then maybe we don't have to reverse everything at the end.

main/src/library/List.flix
CasperDN: I can make the change, but then we are changing the meaning in this simple refactor.

main/src/library/List.flix
CasperDN: This will only be true if we restrict the supplied functions as discussed above.

main/src/library/List.flix
mlutze: Ah, I didn't read the previous implementation closely enough. I thought you were introducing this change.

I agree to keep it as it is and we can look into restricting it later.

main/src/library/List.flix
mlutze: Got it. I'll merge this and create a separate PR to discuss.

mlutze (approved):

PR #11119, opened 2025-07-21
chore: make taenv implicit in Kinder

mlutze:


main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
```diff
@@ -87,20 +87,20 @@ object Kinder {
   /**
     * Performs kinding on the given enum.
     */
-  private def visitEnum(enum0: ResolvedAst.Declaration.Enum, taenv: Map[Symbol.TypeAliasSym, KindedAst.TypeAlias], root: ResolvedAst.Root)(implicit sctx: SharedContext, flix: Flix): KindedAst.Enum = enum0 match {
+  private def visitEnum(enum0: ResolvedAst.Declaration.Enum, root: ResolvedAst.Root)(implicit taenv: Map[Symbol.TypeAliasSym, KindedAst.TypeAlias], sctx: SharedContext, flix: Flix): KindedAst.Enum = enum0 match {
```
JonathanStarup, line 90: taenv is a part of the shared context of kinding across threads, maybe it could be a part of SharedContext? Or do we only put mutable things in there?

mlutze: I think we only use it for errors. Plus there are some places where taenv is not used or available yet but we can still have errors.

PR #11120, opened 2025-07-21
kinder separate specs

mlutze:


magnus-madsen:
@mlutze Conflicted

main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
```diff
@@ -344,482 +380,492 @@ object Kinder {
...
+          case ExtPattern.Var(sym, _, _) => sym
+          case ExtPattern.Error(_, _) => ??? // crash
+        }
+        val exp2 = r2.exp
+        val sym2 = r2.pats.head match {
+          case ExtPattern.Wild(_, loc1) => Symbol.freshVarSym("wildExtPattern", BoundBy.Pattern, loc1)
+          case ExtPattern.Var(sym, _, _) => sym
+          case ExtPattern.Error(_, _) => ??? // crash
+        }
+        KindedAst.Expr.ExtMatch(label, e, sym1, exp1, sym2, exp2, tvar, loc)
```
magnus-madsen: Above needs to be rebased

main/src/ca/uwaterloo/flix/language/phase/Kinder.scala
```diff
@@ -1732,4 +1785,19 @@ object Kinder {
     */
   private case class SharedContext(errors: ConcurrentLinkedQueue[KindError])
 
+  // MATT docs
```
magnus-madsen: What about this?

magnus-madsen:
I would like to request:

- A few sentences to describe the overall strategy
- A measurement of any performance impact with Xperf
- There are some TODOs about missing comments
- There is some code related to extensible variants that has to be carefully managed

magnus-madsen:
At somepoint I proposed a new `case class Instantiation` will that not be useful and more extensible?

magnus-madsen:
I guess the performance benchmark should wait until all the PRs are merged.

mlutze:
This PR is the start of a chain of events intended to lead us to supporting parametric effects, while making our phases more principled along the way.

Monomorphization makes use of sets of effect symbols as the canonical form of monomorphized effect sets. When we add parameters to effects, it means we now need sets of `(effect, type)`. Of course, equality and hashcode on types is not well-defined, so we need a new type, `MonoType`.

Monomorphization makes use of `unifyTypes` to determine how defs and sigs must be specialized. If we have `MonoType`, this would involve unifying `MonoType`s with `Type`s, which would be a huge pain. A solution is to track instantiation information during typing instead of recalculating it later.

This instantiation must be done in `Kinder` rather than `Typer` because `Typer` only returns a substitution, rather than a new AST.

There are several places where we could have taken a shortcut, but I think there are several benefits to this bigger change:

- no more calling unification in the back-end
- beautiful canonical types in the back-end (maybe gets rid of $n$ ICEs)
- one step toward defeating my nemesis, Type Schemes
- the instantiation information in TypedAst could be used for some LSP stuff

mlutze:
> At somepoint I proposed a new `case class Instantiation` will that not be useful and more extensible?

Maybe, but the instantiation information is different depending on what you're instantiating.

If you're instantiating a sig, you need both the class and the sig type parameters, whereas for a def, you only need the def parameters.

Later we'll need to do this for effect ops, where there is only the effect type parameter.

Further down the line, it would be nice to make this uniform and do it for enums as well, where there is only the enum type parameter.

So the data structure would be something like:

```scala
case class Instantiation(targ: Option[Type], targs: List[Type])
```

Or maybe for sigs we have 2 instantiations and it's just

```scala
case class Instantiation(targs: List[Type])
```

Note that we cannot use a map due to problems with incrementality -- symbols are not preserved.

magnus-madsen:
All of the above sounds good 👍

mlutze:
Performance:

```
./benchmark -a "Xperf --n 100 --par --frontend" mlutze/kinder-separate-specs master
Building 'mlutze/kinder-separate-specs'
Building 'master'
On ref 'mlutze/kinder-separate-specs' executing: flix Xperf --n 100 --par --frontend

Throughput (best): 169,774 lines/sec (with 20 threads.)

  min: 24,147, max: 169,774, avg: 147,529, median: 148,723

Finished 100 iterations on 79,091 lines of code in 57 seconds.


On ref 'master' executing: flix Xperf --n 100 --par --frontend

Throughput (best): 172,532 lines/sec (with 20 threads.)

  min: 24,251, max: 172,532, avg: 140,339, median: 144,998

Finished 100 iterations on 79,091 lines of code in 61 seconds.

```

mlutze:
Note that there may be some performance loss in incrementality as scheme kinding is no longer incremental. But it's still incremental on expressions, so it shouldn't be too bad.

magnus-madsen (approved):

PR #11122, opened 2025-07-21
chore: introduce named argument in `pow` in Int* and Big*

CasperDN:


main/src/library/BigDecimal.flix
```diff
@@ -42,10 +42,10 @@ mod BigDecimal {
         unsafe x.abs()
 
     ///
-    /// Returns `x` raised to the power of `n`.
+    /// Returns `base` raised to the power of `n`.
     ///
-    pub def pow(x: BigDecimal, n: Int32): BigDecimal =
-        unsafe x.pow(n)
+    pub def pow(base: {base = BigDecimal}, n: Int32): BigDecimal =
```
mlutze, line 47: Strange that the second parameter is an Int32 here, but I guess that's just because of the Java API.

Issue #6508, opened 2023-09-29
Add `Vector.search` / Binary Search


jaschdoc:
Now that we have `Vector`, I think it would be great to have a search function that uses binary search in addition to the find functions. Why? Well, if we have a pure indexable data structure, we may as well benefit from it!

```scala
///
/// Finds the element `x` in `v`.
///
/// The implementation uses binary search so it may often be faster than `find`
/// if the order of appearance is not important.
///
pub def search(x: a, v: Vector[a]): Option[a] with Order[a] = ???
```

This could probably also apply to `Array`.

magnus-madsen:
I think we can implement this for both Array and Vector, but:

- We should probably call it `binarySearch`, and
- We should be very careful specify that it requires the underlying array/vector to be sorted!

magnus-madsen:
(so it is not an alternative to find).

PR #11123, opened 2025-07-21
feat: implement binarySearch for `Vector`

CasperDN:
Simple proposal for a binary search for vector. See #6508

Could be changed to have a range, e.g. `binarySearch(x: a, from: {from = Int32}, to: {to = Int32}, v: Vector[a])`

### DIFF
--- a/main/src/library/Vector.flix
+++ b/main/src/library/Vector.flix
@@ -146,7 +146,7 @@ mod Vector {
     ///
     /// If `v` contains multiple `x`-values then the index of one of them will be returned.
     ///
-    pub def binarySearch(x: a, v: Vector[a]): Option[Int32]  with Order[a] =
+    pub def binarySearch(x: a, v: Vector[a]): Option[Int32] with Order[a] =
         def f(l, r) = {
             if (l <= r) {
                 let m = (l + r) `Int32.rightShift` 1;
@@ -157,8 +157,8 @@ mod Vector {
                 }
             } else None
         };
-        let length = length(v);
-        f(0, length - 1)
+        let len = length(v);
+        f(0, len - 1)
 
     ///
     /// Compares `a` and `b` lexicographically.
--- a/main/src/library/Vector.flix
+++ b/main/src/library/Vector.flix
@@ -137,6 +137,30 @@ instance ToFlix[Vector[a]] {
 mod Vector {
 
     ///
+    /// Perform a binary search for `x` on `v` and returns the position of `x`.
+    ///
+    /// Returns `None` if `v` does not contain `x`. Otherwise returns `Some(i)`,
+    /// where `Vector.get(i, v) == x`.
+    ///
+    /// Assumes `v` is sorted. If this is not the result is undefined.
+    ///
+    /// If `v` contains multiple `x`-values then one of the indexes will be returned.
+    ///
+    pub def binarySearch(x: a, v: Vector[a]): Option[Int32]  with Order[a] =
+        def f(l, r) = {
+            if (l <= r) {
+                let m = (l + r) `Int32.rightShift` 1;
+                match get(m, v) <=> x {
+                    case Comparison.LessThan => f(m+1, r)
+                    case Comparison.GreaterThan => f(l, m-1)
+                    case Comparison.EqualTo => Some(m)
+                }
+            } else None
+        };
+        let length = length(v);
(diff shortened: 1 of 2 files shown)


main/src/library/Vector.flix
```diff
@@ -136,6 +136,30 @@ instance ToFlix[Vector[a]] {
 
 mod Vector {
 
+    ///
+    /// Perform a binary search for `x` on `v` and returns the position of `x`.
```
JonathanStarup: I think it should say
```
the sorted vector `v`
```
to really make sure the reader gets the message

CasperDN:
We could also go the Java/Scala way and return the insertion point of the element `x` if `x` is not in `v`.

main/src/library/Vector.flix
```diff
@@ -136,6 +136,30 @@ instance ToFlix[Vector[a]] {
...
+    /// where `Vector.get(i, v) == x`.
+    ///
+    /// Assumes `v` is sorted. If this is not the case the result is undefined.
+    ///
+    /// If `v` contains multiple `x`-values then the index of one of them will be returned.
+    ///
+    pub def binarySearch(x: a, v: Vector[a]): Option[Int32]  with Order[a] =
+        def f(l, r) = {
+            if (l <= r) {
+                let m = (l + r) `Int32.rightShift` 1;
```
mlutze: Why not just `/ 2`?

main/src/library/Vector.flix
```diff
@@ -136,6 +136,30 @@ instance ToFlix[Vector[a]] {
...
+    /// Perform a binary search for `x` on the sorted vector `v` and returns the position of `x`.
+    ///
+    /// Returns `None` if `v` does not contain `x`. Otherwise returns `Some(i)`,
+    /// where `Vector.get(i, v) == x`.
+    ///
+    /// Assumes `v` is sorted. If this is not the case the result is undefined.
+    ///
+    /// If `v` contains multiple `x`-values then the index of one of them will be returned.
+    ///
+    pub def binarySearch(x: a, v: Vector[a]): Option[Int32]  with Order[a] =
```
mlutze: Extra space before `with`

main/src/library/Vector.flix
```diff
@@ -136,6 +136,30 @@ instance ToFlix[Vector[a]] {
...
+            if (l <= r) {
+                let m = (l + r) `Int32.rightShift` 1;
+                match get(m, v) <=> x {
+                    case Comparison.LessThan => f(m+1, r)
+                    case Comparison.GreaterThan => f(l, m-1)
+                    case Comparison.EqualTo => Some(m)
+                }
+            } else None
+        };
+        let length = length(v);
```
mlutze: Maybe change to `len` so we don't have this shadowing.

mlutze:
> We could also go the Java/Scala way and return the insertion point of the element `x` if `x` is not in `v`.

I think it's more clear to return an option. Returning the negative position is not very Flixlike. We don't like puns :P

main/src/library/Vector.flix
CasperDN: Performance. The method is from our B+ tree and we found this to be more performant.

This was before the new inliner and we also observed that this got inlined whereas division did not.

CasperDN:
> > We could also go the Java/Scala way and return the insertion point of the element `x` if `x` is not in `v`.
> 
> I think it's more clear to return an option. Returning the negative position is not very Flixlike. We don't like puns :P

I was more thinking about [Searching](https://www.scala-lang.org/api/2.12.x/scala/collection/Searching$.html). As in return an enum with either the insertion point or the index.

main/src/library/Vector.flix
mlutze: Ok. Please add a comment saying that this is demonstrably faster, or else someone like me will come along and replace it later

mlutze:
> > > We could also go the Java/Scala way and return the insertion point of the element `x` if `x` is not in `v`.
> > 
> > 
> > I think it's more clear to return an option. Returning the negative position is not very Flixlike. We don't like puns :P
> 
> I was more thinking about [Searching](https://www.scala-lang.org/api/2.12.x/scala/collection/Searching$.html). As in return an enum with either the insertion point or the index.

Yes, we could consider that. Let's start with the option for now.

PR #11127, opened 2025-07-22
refactor: remove `flix.subtask` and `sample` param from `progressbar.observe`

### LINKED ISSUE #10939: Remove `flix.subtask`
We are not really using it any more and the compiler is now very fast-- so let us drop it.

We should also change ProgressBar.observe not to take the sample parameter.

gwydd12:
This PR includes:
- removal of `flix.subtask`
- removal of `sample` parameter in the `progressbar.observe` function

This PR fixes #10939 .

### DIFF
--- a/main/src/library/Array.flix
+++ b/main/src/library/Array.flix
@@ -220,7 +220,7 @@ mod Array {
     ///
     /// Returns `a` as a MutDeque.
     ///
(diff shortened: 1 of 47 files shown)


mlutze (approved):
Looks good. Thanks!

PR #11129, opened 2025-07-22
chore: improve performance of tuple comparison

CasperDN:
Removed `lazy` and `thenCompare` from tuple comparison.

The following very unscientific program was used to evaluate the performance (it prints every 5000000 comparison):
```
mod TupleOrder {
    pub def compare(t1: (a1, a2), t2: (a1, a2)): Comparison with Order[a1], Order[a2] =
        let (x1, x2) = t1;
        let (y1, y2) = t2;
        match x1 <=> y1 {
            case Comparison.EqualTo => x2 <=> y2
            case res => res
        }

    pub def compare2(t1: (a1, a2), t2: (a1, a2)): Comparison with Order[a1], Order[a2] =
        let (x1, x2) = t1;
        let (y1, y2) = t2;
        let cmp1 = x1 <=> y1;
        if (cmp1 != Comparison.EqualTo) cmp1
        else x2 <=> y2

}

def mainBTreeExperiment(): Unit \ IO = region rc {
    let t1 = (1, 2);
    let t2 = (1, 2);
    // let t1 = (1, 2);
    // let t2 = (3, 2);
    spawn runForeverOld(0i64, t1, t2) @ rc;
    spawn runForeverNew(0i64, t1, t2) @ rc;
    spawn runForeverNew2(0i64, t1, t2) @ rc;
    ()
}

def runForeverOld(i: Int64, t1: (a, b), t2: (a, b)): Unit \ IO with Order[a], Order[b] = 
    if(Int64.modulo(i, 5000000i64) == 0i64) {
        println("Old: ${i}")
    } else {
        ()
    };
    if (t1 <=> t2 == Comparison.EqualTo) {
        runForeverOld(i + 1i64, t1, t2)
    } else {
        runForeverOld(i + 1i64, t1, t2)
    }

def runForeverNew(i: Int64, t1: (a, b), t2: (a, b)): Unit \ IO with Order[a], Order[b] = 
    if(Int64.modulo(i, 5000000i64) == 0i64) {
        println("New: ${i}")
    } else {
        ()
    };
    if (TupleOrder.compare(t1, t2) == Comparison.EqualTo) {
        runForeverNew(i + 1i64, t1, t2)
    } else {
        runForeverNew(i + 1i64, t1, t2)
    }


def runForeverNew2(i: Int64, t1: (a, b), t2: (a, b)): Unit \ IO with Order[a], Order[b] = 
    if(Int64.modulo(i, 5000000i64) == 0i64) {
        println("New2: ${i}")
    } else {
        ()
    };
    if (TupleOrder.compare2(t1, t2) == Comparison.EqualTo) {
        runForeverNew2(i + 1i64, t1, t2)
    } else {
        runForeverNew2(i + 1i64, t1, t2)
    }
```

### DIFF
--- a/main/src/library/Order.flix
+++ b/main/src/library/Order.flix
@@ -374,11 +374,12 @@ instance Order[(a1, a2)] with Order[a1], Order[a2] {
     /// Compares `t1` and `t2` lexicographically.
     ///
     pub def compare(t1: (a1, a2), t2: (a1, a2)): Comparison =
-        use Order.thenCompare;
         let (x1, x2) = t1;
         let (y1, y2) = t2;
-             (x1 <=> y1) `thenCompare`
-        lazy (x2 <=> y2)
+        match x1 <=> y1 {
+            case Comparison.EqualTo => x2 <=> y2
+            case res => res
+        }
 
 }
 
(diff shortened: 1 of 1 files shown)


CasperDN:
For me the match version is quite a bit faster when the first values disagree. This is even true for the proposed new if-else version.

I have no idea why this would be the case.

mlutze:
I wonder if we ought to do something similar for enum comparison.

Issue #11034, opened 2025-07-04
Going from `Map` to `B+ trees` in the Datalog enums


CasperDN:
The most time consuming part of the new engine is writing all computed values to a `Map` to create a `Model` as below:
```
    pub enum Datalog {
        case Datalog(Vector[Constraint], Vector[Constraint])
        case Model(Map[RelSym, Map[Vector[Boxed], Boxed]])
        case Join(Datalog, Datalog)
    }
```
Replacing the inner `Map` in `Model` with a `B+ tree` should yield significant speedup. Unfortunately `B+ trees` are mutable and must have a region. To solve this the enum `Datalog` could have a region or type casts and `Static` must be used. With `Static` the new enum would be:
```
    pub enum Datalog {
        case Datalog(Vector[Constraint], Vector[Constraint])
        case Model(Map[RelSym, BPlusTree[Vector[Boxed], Boxed, Static]])
        case Join(Datalog, Datalog)
    }
```

PR #11131, opened 2025-07-24
refactor: Replace `Map` with `B+` tree in Datalog

CasperDN:
Initial draft to address #11034.

### DIFF
--- a/main/src/library/Fixpoint3/Ast/Datalog.flix
+++ b/main/src/library/Fixpoint3/Ast/Datalog.flix
@@ -44,7 +44,7 @@ mod Fixpoint3.Ast.Datalog {
     @Internal
     pub enum Datalog {
         case Datalog(Vector[Constraint], Vector[Constraint])
-        case Model(Map[RelSym, Map[Vector[Boxed], Boxed]])
+        case Model(Map[RelSym, BPlusTree[Vector[Boxed], Boxed, Static]])
         case Join(Datalog, Datalog)
     }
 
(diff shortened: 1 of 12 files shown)


main/src/library/Fixpoint3/Ast/Datalog.flix
```diff
@@ -92,18 +92,20 @@ mod Fixpoint3.Ast.Datalog {
...
+                                    case Relational => StringBuilder.append("${relSym}(${tupleString}).${String.lineSeparator()}", builder)
+                                    case Latticenal(_, _, _, _) => StringBuilder.append("${relSym}(${tupleString}; %{lat}).${String.lineSeparator()}", builder)
                                 }
                             })
-                        })
-                        |> Iterator.join(String.lineSeparator())
-                        + String.lineSeparator()
+                        );
+                        StringBuilder.toString(builder)
+                    // Remove the `IO` effect from dealing with `Static`.
```
mlutze: I think we have an `unsafely` expression for removing exactly one effect.

CasperDN:
> I'm probably not Datalog-smart enough to judge most of this, but if you extract the random cleanup stuff, I can at least merge that separately.

I have created #11145 for them.

main/src/library/Fixpoint3/Ast/Datalog.flix
CasperDN: Did not know about that one.

CasperDN: I ended up with a weird error. See #11146

magnus-madsen:
@CasperDN Looking at the diff, I think there are still several unrelated changes?

I see some AtomicCounter that is gone and I see some uses of unsafely run IO. Are all of these changes required for this specific PR? Can we break it into smaller pieces?

CasperDN:
The removal of `AtomicCounter` is due to its only usage disappearing.

I only see one way to break it up: Introduce the change in `Datalog`, `Solver.flix`, but remove it from `Ram` and all phases. This would be followed by the change to all phases. Should we move forward with this approach?

CasperDN:
The `unsafely IO run` is for whenever we have to e.g. combine 2 `B+`-trees or read from one.

magnus-madsen:
> The removal of `AtomicCounter` is due to its only usage disappearing.
> 
> I only see one way to break it up: Introduce the change in `Datalog`, `Solver.flix`, but remove it from `Ram` and all phases. This would be followed by the change to all phases. Should we move forward with this approach?

That sounds good.

Issue #11130, opened 2025-07-24
`Parser.oneOrMore`


mlutze:
There are at least two problems with `Parser.oneOrMore`:

1. It returns an Option with an error, which is accidentally ignored in multiple places, leading to crashes such as in https://github.com/flix/flix/issues/11128
2. The error it gives is bad. For example:

_Input_
```scala
def foo(): String = {
    try 123 catch
}
```

_Error_
```
-- Parse Error (OtherExpr) -------------------------------------------------- <playground>

>> Expected 'catch' or 'with'.

2 |     try 123 catch
        ^^^^^^^^^^^^^
        Here

-- Parse Error (OtherExpr) -------------------------------------------------- <playground>

>> Expected at least one <catch-rule>.


Here
```

I'm inclined to think that we should get rid of `Parser.oneOrMore` and just let `Weeder` handle the zero case. We already do this to some extent. For example:

```scala
      flatMapN(pickExpr(tree), traverse(rules0)(visitMatchRule)) {
        // Case: no valid match rule found in match expr
        case (expr, Nil) =>
          val error = NeedAtleastOne(NamedTokenSet.MatchRule, SyntacticContext.Expr.OtherExpr, loc = expr.loc)
          // Fall back on Expr.Error. Parser has reported an error here.
          Validation.Success(Expr.Error(error))
        case (expr, rules) => Validation.Success(Expr.Match(expr, rules, tree.loc))
      }
```

Here we end up redoing the logic for checking "at least one" because we need to construct the error node. But we explicitly avoid adding an error to the context because we already added one in the parser (the bad one).

Related:
- https://github.com/flix/flix/issues/8089
- https://github.com/flix/flix/issues/11128

magnus-madsen:
I think that could be reasonable.

CasperDN:
Fixed in master

PR #11133, opened 2025-07-25
refactor: move 'atLeastOne' checking to Weeder for type parameters

mlutze:
Related to #11130

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
@@ -3126,8 +3126,11 @@ object Weeder2 {
                   val error = MismatchedTypeParameters(tparamsTree.loc)
                   sctx.errors.add(error)
                   tparams
+                // No type parameters. Issue an error and return an empty list.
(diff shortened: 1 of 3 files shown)


magnus-madsen:
Would it make sense to add some test cases while you are doing this?

magnus-madsen:
I think that would be very sensible.

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -3126,8 +3126,11 @@ object Weeder2 {
                   val error = MismatchedTypeParameters(tparamsTree.loc)
                   sctx.errors.add(error)
                   tparams
+                // No type parameters. Issue an error and return an empty list.
```
magnus-madsen, line 3129: We usually have these comments under the case no?

magnus-madsen (approved):

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
mlutze: Not in my experience

PR #11144, opened 2025-07-29
chore: rename sym: XSymUse to symUse: XSymUse

### LINKED ISSUE #9650: name `XSymUse` consistently
In the Asts there is different name schemes for `XSymUse`
- `op: OpSymUse`
- `sym: RestrictableCaseSymUse`
- `symUse: RestrictableEnumSymUse`

and on the use-site its also inconsistent
- `case ResolvedAst.Expr.OpenAs(sym, ..`
- `case ResolvedAst.Expr.ApplyLocalDef(symUse, ..`
- `case KindedAst.Expr.StructGet(exp0, field, ..` (field here is a XsymUse)
- `case Expr.TryWith(exp, effUse, ..`

I propose always using `symUse` and if multiple occur then using `effectSymUse` or other prefix qualifiers

mlutze:
Fixes #9650

I searched for the regexes:

```regex
sym:\s*\S*SymUse                      # sym: XyzSymUse
[Ss]ym\.sym                           # sym.sym or Sym.sym
(?<!Use):\s*(SymUse\.)?\w*SymUse\b    # x: XyzSymUse where x is not "Use"
```

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/Consumer.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/Consumer.scala
@@ -56,7 +56,7 @@ trait Consumer {
 
   def consumeCase(cse: Case): Unit = ()
 
-  def consumeCaseSymUse(sym: CaseSymUse): Unit = ()
+  def consumeCaseSymUse(symUse: CaseSymUse): Unit = ()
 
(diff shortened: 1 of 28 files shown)


magnus-madsen:
@mlutze Conflicted

Issue #11065, opened 2025-07-09
Add extensive Datalog tests


magnus-madsen:
We should brainstorm a long list here. We want both human and machine created tests.

- [ ] Add some tests that check that we at least scale to 1M facts.

JumpTwiice:
- [x] Tests with if-guards containing x number of variables
- [x] Tests with functionals
- [x] Tests with application
- [x] Several tests on the same programs but with different orderings of body atoms
- [x] Test stratification
- [x] Correctness of lattice semantics in Datalog
- [x] Tests on same program with different term orderings in atoms
- [x] Test that `solve` does not mutate internal B+ Trees
- [x] Test `pquery` 
- [x] Test `pquery` order of term types (https://github.com/flix/flix/pull/11249#issuecomment-3163311031) 

Feel free to add more. We have already begun creating some of these

magnus-madsen:
Do we have any more specific tests in mind?

JumpTwiice:
I think we still need the large-scale tests

magnus-madsen:
There is a PR

PR #11154, opened 2025-07-30
feat: add extensive Datalog tests

JumpTwiice:

This is one of several PRs that will add more extensive testing of the Fixpoint module.


JumpTwiice:
On second thought, this specific PR should probably be split into several PRs.

PR #11155, opened 2025-07-30
feat: add tests for application in Datalog

JumpTwiice:


main/test/ca/uwaterloo/flix/library/Fixpoint/TestFixpoint.Application.flix
mlutze, line 1: Since this isn't testing the library, it should go with our Language tests instead.

Something like `main/test/flix/Test.Exp.Fixpoint.Application.flix`

mlutze (commented):
Can we add a few more tests:

1. Where the first argument uses an application
2. Where both arguments use an application
3. Application with lattices (is that allowed?)

CasperDN:
> 3. Application with lattices (is that allowed?)

We haven't touched lattice-tests yet, but that is planned (also for if-guards).

Of course the `fix` keyword is needed for applications, unless the application is to a lattice-relation's lattice-value.

JumpTwiice:
We will add more lattice tests soon. This is a good starting point however.

magnus-madsen (approved):

magnus-madsen:
Thanks!

PR #11156, opened 2025-07-30
feat: add tests for functionals in Datalog

JumpTwiice:


magnus-madsen (approved):

magnus-madsen:
Looks good and I will merge it for now. But I think it can be extended a bit:

- Dimension one: tests with 1, 2, 3 functionals
- Dimension two: tests with different types (e.g. vector of int32, vector of string) and so on.

PR #11157, opened 2025-07-30
feat: add tests for if-guards in Datalog

JumpTwiice:


CasperDN:
> Is the if-guard always last in the Datalog body? And can you have more than one?

You can have more than one if-guard and they do not need to be last.

PR #11159, opened 2025-07-31
feat: add tests with different body atom orderings in Datalog

JumpTwiice:

### DIFF
--- a/main/test/flix/Test.Exp.Fixpoint.Permutations.flix
+++ b/main/test/flix/Test.Exp.Fixpoint.Permutations.flix
@@ -1,0 +1,863 @@
+/*
+ * Copyright 2025 Casper Dalgaard Nielsen
+ *                Adam Yasser Tallouzi
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+mod Test.Exp.Fixpoint.Permutations {
+
+    /////////////////////////////////////////////////////////////////////////////
+    // Test permuting body atom order with Int32                               //
+    /////////////////////////////////////////////////////////////////////////////
+
+    // May not be bigger than 12.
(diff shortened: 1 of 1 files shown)


main/test/flix/Test.Exp.Fixpoint.Permutations.flix
```diff
@@ -0,0 +1,863 @@
...
+    def smallN(): Int32 = 10
+
+    def resultForFirstInt32(n: Int32): Vector[(Int32, Int32, Int32, Int32)] = {
+        Vector.range(0, n - 2) |> Vector.map(x -> (x, x + 1, x + 2, x + 3))
+    }
+
+    @Test
+    def testPermutation01(): Bool = {
+        let n = smallN();
+        let a_ = inject Vector.range(0, n) |> Vector.map(x -> (x, x + 1)) into A/2;
```
magnus-madsen: I would avoid variables that end with underscores. To easy to confused with variables that start with underscore (i.e. unused variables)

main/test/flix/Test.Exp.Fixpoint.Permutations.flix
```diff
@@ -0,0 +1,863 @@
...
+        let b_ = inject Vector.range(0, n) |> Vector.map(x -> (x, x + 1)) into B/2;
+        let c_ = inject Vector.range(0, n) |> Vector.map(x -> (x, x + 1)) into C/2;
+        let p = #{
+            D(a, b, c, d) :- A(a, b), B(b, c), C(c, d).
```
magnus-madsen: Would it be nicer if the predicates are called A, B, C,... and so on.

But the variable symbols are drawn from x, y, z, and so on. 

That way-- there is no mental confusion between A and a.

Or is there in fact some relationship between them?

main/test/flix/Test.Exp.Fixpoint.Permutations.flix
```diff
@@ -0,0 +1,863 @@
...
+    /////////////////////////////////////////////////////////////////////////////
+    // Test permuting body atom order with Int32                               //
+    /////////////////////////////////////////////////////////////////////////////
+
+    // May not be bigger than 12.
+    def smallN(): Int32 = 10
+
+    def resultForFirstInt32(n: Int32): Vector[(Int32, Int32, Int32, Int32)] = {
+        Vector.range(0, n - 2) |> Vector.map(x -> (x, x + 1, x + 2, x + 3))
+    }
```
magnus-madsen, line 32: I think a few comments here in the top of the file to outline the strategy would be useful.

main/test/flix/Test.Exp.Fixpoint.Permutations.flix
CasperDN: I think this was just a case of generating names alphabetically was easiest

PR #11161, opened 2025-07-31
chore: improve performance of `IndexSelection`

CasperDN:
This replaces a `Map` in `IndexSelection` by an array.

### DIFF
--- a/main/src/library/Fixpoint3/Ast/Ram.flix
+++ b/main/src/library/Fixpoint3/Ast/Ram.flix
@@ -442,11 +442,11 @@ mod Fixpoint3.Ast.Ram {
             case (RowVar.Named(i1), RowVar.Named(i2)) => i1 <=> i2
         }
     }
-
+    
     /////////////////////////////////////////////////////////////////////////////
     // RamId                                                                   //
     /////////////////////////////////////////////////////////////////////////////
-    
+
     ///
     /// `RamId` represents a unique identifier for literals, positions in relations,
     /// positions in tuples or input to a function.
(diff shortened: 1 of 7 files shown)


magnus-madsen:
IndexSelection is part of the compilation Datalog to RAM pipeline. Why is that slow-- given that programs are typically small. Is it because it also processes facts?

It seems this PR trades clarify for performance-- which is okay and natural-- but I am just surprised it is worth optimizing on the compiler itself.

CasperDN:
It does not process facts. We are fairly comfortable with the cost of compilation for programs that have large EDBs. We profiled the program:
```
def testFunctionalF1X2_01(): Bool =
let ff = _x -> Vector#{};
    let pr = #{
        A(1).
        R(a, b) :- A(x), let (a, b) = ff(x).
    };
    let actual = query pr select (x, y) from R(x, y);
    Assert.eq((Vector#{}: Vector[(Int32, Int32)]), actual)

def runForever(): Unit \ IO = {
    if(testFunctionalF1X2_01()) runForever() else runForever()
}
```
After ignoring `Lowering.flix` (as we expect this to be changed significantly) and `waitForSingleObject`, which has nothing to do with us, `IndexSelection` is the most expensive operation (30%). A large part of the time is used in `UniqeInts` and its backing `Map`. With this it goes down to 20%, and `AutomaticIndexSelection` is now using a larger part of this share.

magnus-madsen:
Let us discuss in person.

magnus-madsen:
Was this related to the test time or no?

PR #11163, opened 2025-07-31
chore: replace `Map` by `BPlusTree` in the frontend of the Datalog engine

CasperDN:
Frontend of #11131. Works towards a fix of #11034.


CasperDN:
Probably clashes with #11161 in some minor way.

magnus-madsen (approved):

PR #11168, opened 2025-07-31
fix: use zeroOrMore for parsing 'without' expressions

mlutze:
related to #11130 

This also improves recovery by using the nested expression directly instead of creating an error.


magnus-madsen:
Conflicted

Issue #11177, opened 2025-07-31
Add `filterAdjDups` and `filterAdjDupsWith`


magnus-madsen:
- [x] Add  `filterAdjDups` and `filterAdjDupsWith` to `List`
- [x] Add  `filterAdjDups` and `filterAdjDupsWith` to `Vector`

Note `filterAdjDups` is just `filterAdjDupsWith` passing `==`.

Add at least 3 tests for each function.

More info here: https://hackage.haskell.org/package/list-duplicate-0.1.0.0/docs/Data-List-Duplicate.html#v:deleteAdjDups

magnus-madsen:
@CasperDN @JumpTwiice If I could ask either of you to implement this. Its very relevant for `pquery`. 

For example, we want to do:

```scala
let db = #{ Edge(1, 2). Edge(2, 3). Edge(3, 4). /* ... */ };
let pr = #{
    Path(x, y) :- Edge(x, y).
    Path(x, z) :- Path(x, y), Edge(y, z).
};
let mm = psolve db, pr;
let pp = pquery mm select Path(1, 4) with {Edge};
List.map(ematch Edge^(s, d) -> "${s} -> ${d}") |> List.filterAdjDups |> println
```

(will) print `1 :: 2 :: 3 :: 4 :: Nil`

(without `filterAdjDups` it (will) print `1 :: 2 :: 2 :: 3 :: 3 :: 4 :: Nil`)

magnus-madsen:
EDIT: Maybe these have to be called `removeAdjDups` and `removeAdjDupsWith` because `filter` leaves the impression that the duplicates are retained.

magnus-madsen:
Fixed in master.

PR #11188, opened 2025-08-01
feat: Implement `removeAdjDups(With)` in `List`

CasperDN:
First part of #11177.

### DIFF
--- a/main/src/library/List.flix
+++ b/main/src/library/List.flix
@@ -526,6 +526,15 @@ mod List {
     }
 
     ///
+    /// Helper function for `permutations`.
+    ///
+    def removeIndex(i: Int32, l: List[a]): List[a] = match (i, l) {
+        case (_, Nil) => l
+        case (0, _ :: xs) => xs
+        case (p, x :: xs) => x :: removeIndex(p - 1, xs)
+    }
+
+    ///
     /// Returns `l` without adjacent duplicates according to their `Eq` instance.
     ///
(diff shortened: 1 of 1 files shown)


main/src/library/List.flix
```diff
@@ -525,6 +525,37 @@ mod List {
...
+    ///
+    pub def removeAdjDups(l: List[a]): List[a] with Eq[a] = removeAdjDupsWith(Eq.eq, l)
+
+    ///
+    /// Returns `l` without adjacent duplicates according to the function `f`.
+    /// Elements `x` and `y` are duplicates if and only if `f(x, y) = true`.
+    ///
+    /// The first occurence in a chain of duplicates is kept.
+    ///
+    /// `f` is assumed to be symmetric and transitive.
```
magnus-madsen: `f` must defined an equivalence relation on the elements of the list.

main/src/library/List.flix
```diff
@@ -525,6 +525,37 @@ mod List {
         case _ => unreachable!()
     }
 
+    ///
```
magnus-madsen: It seems this was added in the middle of some private helper functions for permutations?

Maybe move it a bit down or up. Somewhere where it logically fits.

CasperDN: Sure. Why are the public functions not ordered alphabetically btw?

magnus-madsen: Historical reasons. Not sure what would be best. Alphabetical ordering is also sometimes strange.

PR #11193, opened 2025-08-01
feat: add support for `pquery` in Lowering

JumpTwiice:

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
@@ -816,13 +816,13 @@ object Lowering {
           val argExps = visitExp(exp) :: acc :: Nil
           val resultType = Types.Datalog
           val itpe = Types.MergeType
-          LoweredAst.Expr.ApplyDef(defn.sym, argExps, itpe, resultType, exp.eff, loc)
+          LoweredAst.Expr.ApplyDef(defn.sym, argExps, List.empty, itpe, resultType, exp.eff, loc)
       }
       val (goalPredSym, goalTerms) = select match {
         case TypedAst.Predicate.Head.Atom(pred, _, terms, _, loc1) =>
           val loweredTerms = terms.map(visitExp)
           val boxedTerms = loweredTerms.map {
-            t => LoweredAst.Expr.ApplyDef(Defs.Box, List(t), Type.mkPureArrow(t.tpe, Types.Boxed, loc), Types.Boxed, t.eff, loc)
+            t => LoweredAst.Expr.ApplyDef(Defs.Box, List(t), List.empty, Type.mkPureArrow(t.tpe, Types.Boxed, loc), Types.Boxed, t.eff, loc)
           }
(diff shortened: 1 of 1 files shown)


JumpTwiice:
I'm stuck on extracting the information I need from the type of `FixpointQueryWIthProvenance`. An example type is 
`Extensible[#(P(String), Q(Int32, Int64))]` from which I want `(Name.Pred("P"...), Type.Str :: Nil) :: (Name.Pred("Q", ...), Type.Int32 :: Type.Int64 :: Nil) :: Nil` of type `List[(Name.Pred, List[Type])]`. I cant see an obvious way to do this, since the `exp` just has a general type.

magnus-madsen:
The expression just has `tpe: Type` but because this is after type inference you are permitted to assume more.

That is, you are allowed to write a pattern match on type (or use .typeConstructor / .typeArguments) to take the type apart. You know its expected shape (but it is a bit difficult to navigate). I suggest you add a breakpoint in the IDE and open the debugger. Alternatively, you can print the type. 

So, you write some code to take a part the type. If you see some type that you don't expected (i.e. in all other cases), you can just throw an InternalCompilerException.

magnus-madsen:
For this program:

```
def main(): Unit =
    let db = #{ A(1). B(2, 3). };
    let _ = pquery db select B(2, 3) with {A, B};
    ()
```

The type of `db` is something like:

```
Apply(
  Cst(Schema,foo.flix:2:14),
    Apply(
      Apply(Cst(SchemaRowExtend(A),foo.flix:2:18),
        Apply(Cst(Relation(1),foo.flix:2:18),Cst(Int32,unknown:1:1),foo.flix:2:18),foo.flix:2:18
      ),
      Apply(Apply(Cst(SchemaRowExtend(B),foo.flix:2:24),
        Apply(Apply(Cst(Relation(2),foo.flix:2:24),Cst(Int32,unknown:1:1),foo.flix:2:24),Cst(Int32,unknown:1:1),foo.flix:2:24),foo.flix:2:24),
      Var(tvar$170265,foo.flix:2:24),foo.flix:2:24),foo.flix:2:18
     ),foo.flix:2:14
)
```

Not super nice, but with some helper functions and recursion you can easily extract what you need.

magnus-madsen:
I think I would do:

- One function that unwraps the outer Apply on Schema
- A helper function that unwraps one Apply on SchemaRowExtend -- extracting the label
- A helper function that extracts the arguments of SchemaRowExtend  into a list of types

magnus-madsen:
The very first function looks approximately like:

```
def unwrapSchema(tpe: Type): ... = tpe match {
  case Type.Apply(Type.Cst(TypeConstructor.Schema), row)) => unwrapRow(row)
  case _ => throw InternalCompilerException(...)
}
```

magnus-madsen:
Program:

```
def main(): Unit \ IO = 
    let x = pquery #{ A("he"). B("hi", "ho"). } select A("he") with { A, B };
    println(Vector.length(x))
```

JumpTwiice:
This branch is so cooked I dont know how to fix it.

JumpTwiice:
Making a new oen

Issue #11106, opened 2025-07-16
Add `EncodingWriter`


rywiese:
An adapter that wraps an `Int8` writer and an `Encoder`, and is itself a `Char` writer:
```
enum Encoder(CharsetEncoder)

enum EncodingWriter[t]({encoder = Encoder, byteWriter = t})

instance Writeable[EncodingWriter[t]] with Writeable[t] where Writeable.Elm[t] ~ Int8 {

    type Elm = Char

    type Aef = Writeable.Aef[t] + IO

    pub def write(buffer: Array[Char, r], writer: EncodingWriter[t]): Result[IoError, Int32] \ r + Writeable.Aef[t] + IO = EncodingWriter.write(buffer, writer)

}
```

PR #11194, opened 2025-08-01
feat: adds EncodingWriter

rywiese:
fixes https://github.com/flix/flix/issues/11106

### DIFF
--- a/main/src/library/EncodingWriter.flix
+++ b/main/src/library/EncodingWriter.flix
@@ -28,9 +28,9 @@ instance Writable[EncodingWriter[t]] with Writable[t] where Writable.Elm[t] ~ In
 
     type Elm = Char
 
-    type Aef = Writable.Aef[t] + IO
+    type Aef = Writable.Aef[t]
 
-    pub def write(buffer: Array[Char, r], writer: EncodingWriter[t]): Result[IoError, Int32] \ r + Writable.Aef[t] + IO = EncodingWriter.write(buffer, writer)
+    pub def write(buffer: Array[Char, r], writer: EncodingWriter[t]): Result[IoError, Int32] \ r + Writable.Aef[t] = EncodingWriter.write(buffer, writer)
 
 }
 
@@ -59,7 +59,7 @@ mod EncodingWriter {
     ///
     /// Returns `Err(e)` if some underlying I/O error occurs.
     ///
-    pub def write(src: Array[Char, r], writer: EncodingWriter[t]): Result[IoError, Int32] \ IO + r + Writable.Aef[t] with Writable[t] where Writable.Elm[t] ~ Int8 =
+    pub def write(src: Array[Char, r], writer: EncodingWriter[t]): Result[IoError, Int32] \ r + Writable.Aef[t] with Writable[t] where Writable.Elm[t] ~ Int8 =
         let EncodingWriter.EncodingWriter({ encoder, byteWriter }) = writer;
         region rc {
             let dest = encoder |> encode(rc, src);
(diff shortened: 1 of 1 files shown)


main/src/library/EncodingWriter.flix
```diff
@@ -0,0 +1,115 @@
...
+    ///
+    def encodeBuffered(rc: Region[rc], charBuffer: CharBuffer, byteBuffer: ByteBuffer, encoder: CharsetEncoder): Array[Int8, rc] \ IO + rc =
+        let result = encoder.encode(charBuffer, byteBuffer, false);
+        let bytesEncoded = byteBuffer.position();
+        let dest = byteBuffer.array() |> Array.take(rc, bytesEncoded);
+        if (result.isOverflow()) {
+            charBuffer.flip();
+            byteBuffer.clear();
+            let rest = encoder |> encodeBuffered(rc, charBuffer, byteBuffer);
+            Array.append(rc, dest, rest)
```
rywiese: This branch is only possible if the encoding can have more than 4 bytes per char which is super edge case.

I figured it's good to have this branch just for correctness.

But there's a couple of points to be raised such as:
* i will need to write a custom implementation of a decoder that supports 5 byte chars just to test this (probably doable, essentially just a mock)
* this array.append is not efficient, but introducing a mutable map would have performance cost for regular encodings, which i don't think is worth while

rywiese: I could actually do

```
let dest: Array[Int8, rc] = Array.empty(rc, numChars * encoding.maxBytesPerChar());
```

on line 76 instead of hardcoding 4. i'll look into this more. that returns a float though, which gives me pause because what to do about nans and infinities. I have to specify something, even if just to satisfy the `clampToInt32` function... it would maybe be nice to have some kind of illegal argument exception here for these kinds of edge cases

magnus-madsen:
@mlutze Can you do initial review and then I will review?

main/src/library/EncodingWriter.flix
mlutze: Yeah it seems crazy that `maxBytesPerChar` returns a float instead of an `int`...

main/src/library/EncodingWriter.flix
```diff
@@ -0,0 +1,115 @@
...
+        } else {
+            dest
+        }
+
+    ///
+    /// Writes all bytes from `dest` into `byteWriter`.
+    ///
+    def writeBytes(rc: Region[r], dest: Array[Int8, r], byteWriter: t): Result[IoError, Unit] \ r + Writable.Aef[t] with Writable[t] where Writable.Elm[t] ~ Int8 =
+        def loop(arr) = {
+            if (arr |> Array.length == 0) {
```
mlutze: Do we not have `Array.isEmpty`?

main/src/library/EncodingWriter.flix
```diff
@@ -0,0 +1,115 @@
...
+    ///
+    /// Guarantees that `0 <= k <= length(b)`.
+    ///
+    /// Returns `Err(e)` if some underlying I/O error occurs.
+    ///
+    pub def write(src: Array[Char, r], writer: EncodingWriter[t]): Result[IoError, Int32] \ IO + r + Writable.Aef[t] with Writable[t] where Writable.Elm[t] ~ Int8 =
+        let EncodingWriter.EncodingWriter({ encoder, byteWriter }) = writer;
+        region rc {
+            let dest = encoder |> encode(rc, src);
+            forM(_ <- byteWriter |> writeBytes(rc, dest)) yield src |> Array.length
```
mlutze: I think we tend to prefer `forM` stuff on separate lines.

main/src/library/EncodingWriter.flix
```diff
@@ -0,0 +1,115 @@
...
+    ///
+    pub def wrap(charset: CharacterSet, byteWriter: t): EncodingWriter[t] \ IO =
+        let CharacterSet.CharacterSet(cs) = charset;
+        EncodingWriter({
+            encoder = cs.newEncoder().onMalformedInput(CodingErrorAction.REPLACE).onUnmappableCharacter(CodingErrorAction.REPLACE),
+            byteWriter = byteWriter
+        })
+
+    ///
+    /// Reads `k` items from `b` and writes them into the underlying resource.
```
mlutze: `b` should be `src`?

main/src/library/EncodingWriter.flix
```diff
@@ -0,0 +1,115 @@
...
+    /// Writes all bytes from `dest` into `byteWriter`.
+    ///
+    def writeBytes(rc: Region[r], dest: Array[Int8, r], byteWriter: t): Result[IoError, Unit] \ r + Writable.Aef[t] with Writable[t] where Writable.Elm[t] ~ Int8 =
+        def loop(arr) = {
+            if (arr |> Array.length == 0) {
+                Ok(())
+            } else {
+                forM(
+                    bytesWritten <- Writable.write(arr, byteWriter);
+                    _ <- loop(arr |> Array.drop(rc, bytesWritten))
```
mlutze, line 122: This is a similar situation to before, where we only get into this part if the writer misbehaves, right?

mlutze: Because if this is a possibly frequent occurrence, copying the array feels really bad.

mlutze (commented):
Does an encoding writer need to have an IO effect? Say I am conceptually writing from one array to another: it doesn't have to do IO, right? Or is encoding IO by itself?

I assume the same question concerns EncodingReader

main/src/library/EncodingWriter.flix
rywiese: Right, happens if for some reason not all bytes were consumed on the first read.

Looking at the java [writer](https://docs.oracle.com/javase/8/docs/api/java/io/Writer.html) docs, the `write` methods return void rather than number of bytes written (like `read` does). So it's not clear to me how java might communicate to a caller that only some bytes were written. It seems likely that java does have stronger guarantees for writes than for reads, since writing doesn't block waiting for input like reading does. But i'm not sure, and I don't want to code against that assumption here without more investigation. i can make a ticket to investigate that and maybe formalize our guarantees/return types a little better on the write side as well. (ex. does it make sense to return `bytesWritten` if it's guaranteed to be the input size?)

but at any rate - it is not frequent, and i don't know how to get around copying the array. since we need to say "okay just copy the last `n` elements from this existing array". It's not like we can pass a pointer to the middle of the array right? we would need a mutable list due to dynamic sizing, but then we'd have to convert to arrays anyways. Or we could add a method like java's [write](https://docs.oracle.com/javase/8/docs/api/java/io/Writer.html#write-char:A-int-int-)(char[] cbuf, int off, int len) to the writer trait but i don't think that's worth doing for this ticket.

rywiese:
> Does an encoding writer need to have an IO effect? Say I am conceptually writing from one array to another: it doesn't have to do IO, right? Or is encoding IO by itself?
> 
> I assume the same question concerns EncodingReader

Java `CharsetEncoder.encode` method has the `IO` effect

main/src/library/EncodingWriter.flix
rywiese: seems that possibly it's due to the fact that `averageBytesPerChar` does return a float and they wanted api consistency like if someone is using both values in an equation together. tho i found a JDK ticket talking about fractional values here. I think maybe the best bet is allocate: `numChars * maxBytesPerChar |> ceil |> clampToInt32(nanValue = 4)` (4 is the greatest `maxBytesPerChar` in the java `StandardCharsets`).

It seems like an unrealistic edge case that there would be a charset that breaks this, and then the blame would probably be on the charset and not on us. And, this would be good because we would not need to allocate `4 * numChars` for a charset with, for example, `maxBytesPerChar` of 2.

Thoughts?

mlutze:
> > Does an encoding writer need to have an IO effect? Say I am conceptually writing from one array to another: it doesn't have to do IO, right? Or is encoding IO by itself?
> > I assume the same question concerns EncodingReader
> 
> Java `CharsetEncoder.encode` method has the `IO` effect

Right, I guess it's more of a philosophical question: is encoding effectful? Can we safely cast it away? What say ye, @magnus-madsen ?

main/src/library/EncodingWriter.flix
mlutze: Seems reasonable.

main/src/library/EncodingWriter.flix
mlutze: Ok. Let's move ahead with this design, and we can investigate the array-plus-offset style later if performance is an issue.

magnus-madsen:
> > > Does an encoding writer need to have an IO effect? Say I am conceptually writing from one array to another: it doesn't have to do IO, right? Or is encoding IO by itself?
> > > I assume the same question concerns EncodingReader
> > 
> > 
> > Java `CharsetEncoder.encode` method has the `IO` effect
> 
> Right, I guess it's more of a philosophical question: is encoding effectful? Can we safely cast it away? What say ye, @magnus-madsen ?

I leave it to you guys.

magnus-madsen:
@mlutze If you are happy you can merge.

rywiese:
FYI i just pushed a fixup commit because i had accidentally named the tests `readxx` instead of `writexx`

rywiese:
> is encoding effectful? Can we safely cast it away?

> I leave it to you guys.

Maybe leave it for now, and if it proves tedious then remove it? Seems safer than the other way around.

rywiese:
But maybe an argument in favor of removing it is that: for both encoding and decoding, we have configuration to map bad bytes/chars into U+FFFD/�, so these operations never throw io errors.

mlutze:
I think, in principle encoding and decoding are pure functions, so we should be able to cast away the effect safely. Like you say, errors are not possible so it's just a question of side-effects or nondeterminism, which there should not be, right?

rywiese:
Right. Yeah that makes sense then. I'll cast it away (and do the same for encoding reads in another PR).

rywiese:
@mlutze I did the cast in a second commit which I can squash once you give it the 👍. I casted the whole function body in one go, which didn't seem ideal... (as opposed to just putting a cast around the function call where the effect comes from).

But... every `buffer.position()`, `buffer.flip()`, etc had `IO` so I had casts on every line. There was also an issue where (at least my hunch is that):
```
let _ unchecked_cast(charBuffer.flip() as _ \ {});
```
was being optimized away/not called and everything broke

mlutze:
> @mlutze I did the cast in a second commit which I can squash once you give it the 👍. I casted the whole function body in one go, which didn't seem ideal... (as opposed to just putting a cast around the function call where the effect comes from).
> 
> But... every `buffer.position()`, `buffer.flip()`, etc had `IO` so I had casts on every line. There was also an issue where (at least my hunch is that):
> 
> ```
> let _ unchecked_cast(charBuffer.flip() as _ \ {});
> ```
> 
> was being optimized away/not called and everything broke

Yeah, better to cast at the top since we need the intermediate effects.

PR #11197, opened 2025-08-02
feat: adds readLine functionality to Peekable

rywiese:


main/src/library/Peekable.flix
```diff
@@ -63,4 +63,32 @@ trait Peekable[t] with Readable[t] {
...
+        forM(
+            line <- Peekable.readUntil(c -> c == '\n' or c == '\r', reader);
+            delim1 <- Peekable.peek(reader);
+            _ <- if (delim1 == Some('\n') or delim1 == Some('\r')) Peekable.skip(1, reader) else Ok(0);
+            _ <- if (delim1 == Some('\r')) {
+                forM(
+                    delim2 <- Peekable.peek(reader);
+                    res <- if (delim2 == Some('\n')) Peekable.skip(1, reader) else Ok(0)
+                ) yield res
+            } else Ok(0)
```
mlutze: There's some subtlety going on here; maybe it's clearer if we extract it into a helper?

Something like `skipChar(c: Char): Result[IoError, Unit]` which does nothing if it's not the char we want.

Then we have:

```
forM(
  ...
  _ <- skip('\r')
  _ <- skip('\n')
  ...
)
```

That seems to cover the 3 cases.

main/src/library/Peekable.flix
```diff
@@ -63,4 +63,55 @@ trait Peekable[t] with Readable[t] {
...
+    ///
+    /// Returns `Err(err)` if there is an IO error.
+    ///
+    pub def readLine(reader: t): Result[IoError, String] \ Readable.Aef[t] where Readable.Elm[t] ~ Char =
+        forM(
+            line <- Peekable.readUntil(c -> c == '\n' or c == '\r', reader);
+            _ <- Peekable.skipElem('\r', reader);
+            _ <- Peekable.skipElem('\n', reader)
+        ) yield region rc {
+            let builder = StringBuilder.empty(rc);
```
mlutze: Java has a built-in char array-to-string function. That's probably much faster than a foreach

main/src/library/Peekable.flix
```diff
@@ -63,4 +63,55 @@ trait Peekable[t] with Readable[t] {
...
+            peeked <- reader |> Peekable.peek;
+            n <- match peeked {
+                case Some(next) => if (p(next)) {
+                    reader |> Peekable.skip(1)
+                } else {
+                    Ok(0)
+                }
+                case None => Ok(0)
+            }
+        ) yield if (n == 0) false else true
```
mlutze: Is the redundant if/else just for clarity?

rywiese: I... don't know why i did that

main/src/library/Peekable.flix
```diff
@@ -63,4 +65,49 @@ trait Peekable[t] with Readable[t] {
...
+            peeked <- reader |> Peekable.peek;
+            n <- match peeked {
+                case Some(next) => if (p(next)) {
+                    reader |> Peekable.skip(1)
+                } else {
+                    Ok(0)
+                }
+                case None => Ok(0)
+            }
+        ) yield not (n == 0)
```
mlutze: You could even use `n != 0` if you want :D

rywiese: done. for some reason i thought flix wouldn't let me do that. but maybe `!=` is allowed but `!b` generally isn't, and you have to use `not`? doesn't really matter i guess

PR #11205, opened 2025-08-03
feat: improve performance of `TypeContext`

magnus-madsen:


mlutze:
Does this really make a difference?

mlutze (approved):

magnus-madsen:
> Does this really make a difference?

In my experiments yes.

magnus-madsen:
> > Does this really make a difference?
> 
> In my experiments yes.

Conceptually: You would expect ConstraintGen to produce on the order of N new data objects-- i.e. approx one constraint per size of the AST. This is not really true because more than not there are more than one constraint. But anyway, the overhead here means that it was a constant factor more object allocations than it should be- which is part of why ConstraintGen was more expensive than one would expect.

Issue #11016, opened 2025-07-02
Rework `foreach`


magnus-madsen:
- [x] Introduce the `ForEach` trait:
```scala
trait ForEach[t] {
    type Elm: Type
    type Aef: Eff = {}
    pub def forEach(f: ForEach.Elm[t] -> Unit \ ef, t: t): Unit \ (ef + ForEach.Aef[t])
}
```
- [x] Add instances for everything (i.e. all collections + range)
- [x] Rewrite `Desugar.desugarForEach` to use `ForEach.forEach` (instead of `Iterable` and `Iterator`).
- [x] Update https://doc.flix.dev/foreach.html to discuss the `ForEach` trait and not `Iterable`.
- [x] Add `WithIndex`:
```scala
mod ForEach {
    pub enum WithIndex[t](t)

    instance ForEach[WithIndex[t]] with ForEach[t] {
        type Elm = (Int32, ForEach.Elm[t])
        type Aef = ForEach.Aef[t]
        pub def forEach(f: ((Int32, ForEach.Elm[t])) -> Unit \ ef, t: WithIndex[t]): Unit \ ef + ForEach.Aef[t]  = ???
    }
    
    pub def withIndex(m: t): WithIndex[t] with ForEach[t] = WithIndex.WithIndex(m)
}
```
- [x] Add more views like `WithFilter`, `WithMap` etc.
- [x] Allow irrefutable pattern matches `foreach((fst, snd) <- pairs)`.

Each bullet can be a separate PR.

magnus-madsen:
Phase 2: Keep `foreach` as an AST node all the way up to TypedAst (to improve IDE support)

rywiese:
> Rewrite Desugar.desugarForEach to use ForEach.forEach (instead of Iterable and Iterator).

@magnus-madsen should `desugarForEachYield` be a bullet also? That one uses `Collectable.collect` which takes an iterator so it might be slightly less straightforward, and worth doing in its own PR. But worth doing?

magnus-madsen:
> > Rewrite Desugar.desugarForEach to use ForEach.forEach (instead of Iterable and Iterator).
> 
> [@magnus-madsen](https://github.com/magnus-madsen) should `desugarForEachYield` be a bullet also? That one uses `Collectable.collect` which takes an iterator so it might be slightly less straightforward, and worth doing in its own PR. But worth doing?

No. ForEachYield is a completely different thing and will die soon for unrelated causes.

[...]

PR #11209, opened 2025-08-04
chore: foreach uses ForEach trait instead of Iterators

rywiese:
https://github.com/flix/flix/issues/11016

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
@@ -1115,7 +1115,7 @@ object Desugar {
   }
 
   /**
-    * Rewrites a `ForEach` loop into a series of `Iterator.forEach` calls.
+    * Rewrites a `ForEach` loop into a series of `ForEach.forEach` calls.
     * {{{
     *   foreach (
     *           x <- xs;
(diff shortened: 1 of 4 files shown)


main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
```diff
@@ -1155,10 +1151,8 @@ object Desugar {
         DesugaredAst.Expr.Match(e1, List(matchRule), loc1.asSynthetic)
     }
 
-    val scope = DesugaredAst.Expr.Scope(regIdent, foreachExp, loc0)
-
     // We add an ascription to Unit because inference across region boundaries is limited.
```
rywiese, line 1154: Is this relevant now that we don't need a region?

main/test/flix/Test.Dec.Assoc.Type.Eff.flix
```diff
@@ -32,7 +32,8 @@ mod Test.Dec.Assoc.Type.Eff {
     instance Runner[Vector[a]] with Runner[a] {
         pub type E = Runner.E[a]
         pub def exec(x: Vector[a]): Unit \ Runner.E[a] =
-            foreach (a <- x) Runner.exec(a)
+            let f: a -> Unit \ Runner.E[a] = Runner.exec;
```
rywiese, line 35: Without this annotation, the compiler times out.

I think this is a bug that is uncovered here but not introduced. Justification: if i check out master and manually desugar this `foreach` loop into a `ForEach.forEach`, I get the same timeout. So it's not related to the actual `foreach` construct and it's not introduced in this PR.

Also, the [PR](https://github.com/flix/flix/pull/11195) that introduced `ForEach` did not touch the compiler, only added flix code to stdlib.

Is this a known issue or should I make a ticket? I don't think it should block merging this.

main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
magnus-madsen: No, we can drop the scope.

main/test/flix/Test.Dec.Assoc.Type.Eff.flix
magnus-madsen: Interesting... You can make a ticket

main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
rywiese: @magnus-madsen I was referring to the ascription

main/test/flix/Test.Dec.Assoc.Type.Eff.flix
rywiese: https://github.com/flix/flix/issues/11213

main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
magnus-madsen: Ah, I think that can be dropped too?

Issue #11203, opened 2025-08-03
Datalog: Tweaks to style


magnus-madsen:
- [x] Use `foreach`-loop (once it lands)
- [x] Instead of `List.append` just use `++`
- [x] Replace `foldLeft` by `foldMap` when possible:
  -  e.g. `Vector.foldLeft(acc -> x -> collectRowVarRelSymStmt(x) ++ acc, Nil, stmts)` becomes
  - Vector.foldMap(collectRowVarRelSymStmt, stmts)
- [x] Align pattern matches on `=>` (e.g. all the arrows are on the same column) -- of course only when body is a single line
- [x] Rename "collectX' functions to "Xof"
  - e.g. `collectRowVarRelSymOp` becomes `rowVarRelSymOf`
- [x] Align multi-line pipelines not like this:
```
            minChainCover(searches) |>
            List.map(chain -> List.zip(chain, Vector.empty() :: chain)) |>
            List.map(List.map(match (s1, s2) -> Vector.filter(x -> not Vector.memberOf(x, s2), s1))) |>
            List.map(List.fold)
```
but like this:
```
            minChainCover(searches) 
                |> List.map(chain -> List.zip(chain, Vector.empty() :: chain)) 
                |> List.map(List.map(match (s1, s2) -> Vector.filter(x -> not Vector.memberOf(x, s2), s1))) 
                |> List.map(List.fold)
```

CasperDN:
Is there a reason for `++` instead of `:::`?

We have used `:::` in 2 locations should I also change these to `++`?

CasperDN:
> Align multi-line pipelines not like this:

What about for single pipelines? As an example:
```
let nonExtendedPositions = Vector.range(0, arityOf(relSym)) |>
    Vector.map(i -> getOrCrash(Map.get(RamId.RelPos(relSymId, i), mapping)));
```

magnus-madsen:
> Is there a reason for `++` instead of `:::`?
> 
> We have used `:::` in 2 locations should I also change these to `++`?

::: is better if its for lists

magnus-madsen:
> > Align multi-line pipelines not like this:
> 
> What about for single pipelines? As an example:
> 
> ```
> let nonExtendedPositions = Vector.range(0, arityOf(relSym)) |>
>     Vector.map(i -> getOrCrash(Map.get(RamId.RelPos(relSymId, i), mapping)));
> ```

Here I have no opinion

CasperDN:
[...]

PR #11211, opened 2025-08-04
chore: consistent use of pipeline

CasperDN:
See #11203

### DIFF
--- a/main/src/library/Fixpoint3/Provenance.flix
+++ b/main/src/library/Fixpoint3/Provenance.flix
@@ -1,0 +1,392 @@
+/*
+ * Copyright 2025 Casper Dalgaard Nielsen
+ *                Adam Yasser Tallouzi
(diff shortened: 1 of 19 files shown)


main/src/library/Fixpoint3/Provenance.flix
```diff
@@ -0,0 +1,392 @@
+/*
```
magnus-madsen: This file is new :)

magnus-madsen (approved):

main/src/library/Fixpoint3/Provenance.flix
magnus-madsen: I removed it from the PR :)

CasperDN: Copy pasting will be the end of me at some point.

PR #11215, opened 2025-08-05
chore: remove `Eq[v]` requirement from `BPlusTree`

CasperDN:
The only negative side effect will be that a lock is actually grabbed when inserting duplicate values.


magnus-madsen:
But Order implies Eq, so I don't follow?

CasperDN:
For a `B+` tree, `BPlusTree[k, v, r]`, we required `Eq[v]` and `Order[k]`. This just removes the requirement on `v`, not on `k`.

magnus-madsen (approved):

PR #11226, opened 2025-08-05
feat: add functionality for writing lines to Writable

rywiese:


main/src/library/Writable.flix
```diff
@@ -47,6 +47,45 @@ trait Writable[t] {
...
+            ) yield ()
+        }
+
+    ///
+    /// Writes a new line separator to `writer`.
+    ///
+    /// The new line separator is platform dependent, and is determined by the "line.separator" system property.
+    ///
+    pub def writeNewLine(writer: t): Result[IoError, Unit] \ IO + Sys + Writable.Aef[t] with Writable[t] where Writable.Elm[t] ~ Char =
+        let newLine = System.getProperty("line.separator");
```
mlutze, line 73: I'm wondering whether this should use the abstract `Environment` effect instead.

mlutze: On one hand, this is a maybe a lowish level API, but on the other hand, we tend to prefer the high-level effects whenever possible...

I'll need another second opinion from the chief @magnus-madsen

magnus-madsen: Not sure

magnus-madsen: What's pros/cons?

mlutze: Pros:

- abstract
- can make a pure writer in theory?

Cons:

- performance
- requires another handler

magnus-madsen: I say we leave out the environment for now, but create a ticket.

magnus-madsen: Ohw ait, but we dont want the Sys and IO effect right?

magnus-madsen:
@mlutze @rywiese You two can make a decision and merge.

main/src/library/Writable.flix
magnus-madsen: What if we took the line separator as an argument?

rywiese: > What if we took the line separator as an argument?

I'd be concerned that people just hardcode `\n` instead of making that system property call

magnus-madsen: We can leave a documentation comment.

magnus-madsen: My proposal is also a bit of a cop out; deferring the decision until later.

rywiese: I think it would be an odd user experience if a function requires an argument, but the docs tell me exactly what the argument has to be. and, assuming the user does `writeNewLine(System.getProperty("line.separator"), writer: t)` the same effects are emitted at the call site, so do we achieve anything?

I think if we went this route it might be cleaner to just have a `writeString` function and the caller has to put the newlines in themselves. Otherwise the `writeNewLine(newLine: String)` function becomes identical to the `writeString(str: String)` function.

rywiese: And for that matter, `writeLine` is basically a convenience for `writeString` then `writeNewLine`. So if `Sys` is burdensome for some reason, `writeString` is always an option

rywiese: But that doesn't really answer the `Sys` vs `Environment` question, and i'm not sure i'm familiar enough yet with the idioms here to have a strong opinion

mlutze: We'll go with this for now, and can revisit later.

PR #11227, opened 2025-08-06
chore: allow nullary functionals in Datalog

### LINKED ISSUE #11011: Functional predicates in Datalog crashing at runtime
Following the [example](https://doc.flix.dev/functional-predicates.html) with a function to generate primes, `primesInRange` we could write the following:
```
Primes(x) :- let x = primesInRange(0, 100).
```
This fails though. `Lowering` crashes with `mkFunctional` hitting the first if-branch which declares that the number of variables 'in variable', should be at least 1 (and at most 5). The following works fine though
```
Dummy(100).
Primes(x) :- Dummy(x), let x = primesInRange(0, x).
```
It appears that removing
```
if (numberOfInVars == 0) {
  throw InternalCompilerException("Requires at least one in variable.", loc)
}
```
from `Lowering.flix` and adding
```
@Internal
pub def lift0X1(f: Vector[(o1)]): Vector[Boxed] -> Vector[Vector[Boxed]]
    with Order[o1] =
    (_: Vector[Boxed]) ->
        let g = match (o1) -> Vector#{box(o1)};
        Vector.map(g, f)
```
to `Fixpoint/Boxable.flix` allows this program to run.

CasperDN:
Fixes #11011.

Simply removes the requirement of at least 1 free variable and add functions to handle this in both engines.


magnus-madsen (approved):

magnus-madsen:
Very nice. Thanks.

PR #11232, opened 2025-08-06
chore: replace `X.forEach(...)` with `foreach(...) {...}` in Boxing

CasperDN:
Replace most `X.forEach` with the new foreach loops.

There are 2 kinds of exceptions to the replacement in this draft. The first is partial application:
```
Vector.forEach(computeMappingTermRec, terms)
```
The second is small one liners:
```
Vector.forEach(i -> insert(RamId.RelPos(predSym, i)), Vector.range(0, arity))
```

Style choice: I have always used the brackets, even for one-liners. Argument: Readability

### DIFF
--- a/main/src/library/Fixpoint3/Boxing.flix
+++ b/main/src/library/Fixpoint3/Boxing.flix
@@ -80,19 +80,17 @@ mod Fixpoint3.Boxing {
         let RamProgram.Program(_, facts, _, (indexes, _)) = program;
         let (f1, f2) = facts;
         let newFacts = MutMap.empty(rc1);
-        f1 |>
-            Map.forEach(relSym -> relFacts -> {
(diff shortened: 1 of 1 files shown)


magnus-madsen:
What are your personal thoughts on readability? is this better or not?

magnus-madsen:
I think it should also be `foreach<SPACE>(`. Because its `if (foo)` not `if(foo)`.

CasperDN:
I think its more natural. As can be seen in the old code with the pipe lines we prefer to know what the data is before reading the loop. With `foreach` we naturally have the data structure on the first line.

It also helps with a simpler/more consistent use of indentation.

magnus-madsen:
Okay; let's try to use `foreach` everywhere in the Datalog solver where it makes sense.

PR #11234, opened 2025-08-06
fix: bug in example `ShortestPath`

CasperDN:
Hopefully last error in these examples.


CasperDN:
This error has been introduced by me, so I guess I can't complain about this one.

magnus-madsen (approved):

PR #11235, opened 2025-08-06
chore: remove unnecessary ascription from desugared foreach

rywiese:
This is a follow up from https://github.com/flix/flix/pull/11209#discussion_r2250942848

The region boundary in question no longer exists because we are no longer using iterators. So it maybe makes sense to remove this in the interest of minimizing cruft.

That being said, i don't fully understand the problem that motivated this ascription in the first place, so maybe this is a chesterson fence and it's safer to leave it be. Bringing it up to the team to consider, i don't have strong feelings about it


magnus-madsen:
You are right. This can be removed. Thanks!

PR #11239, opened 2025-08-06
chore: add tests for `solve` in Datalog

JumpTwiice:

### DIFF
--- a/main/test/flix/Test.Exp.Fixpoint.KeywordSolve.flix
+++ b/main/test/flix/Test.Exp.Fixpoint.KeywordSolve.flix
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2025 Casper Dalgaard Nielsen
+ *                Adam Yasser Tallouzi
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+mod Test.Exp.Fixpoint.KeywordSolve {
+
+    /////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


main/test/flix/Test.Exp.Fixpoint.KeywordSolve.flix
```diff
@@ -0,0 +1,69 @@
...
+            ASolve01(42).
+        };
+        let otherProg = #{
+            ASolve01(-42).
+        };
+        let k = solve solved, otherProg;
+        let res1 = query solved select x from ASolve01(x);
+        let res2 = query k select x from ASolve01(x);
+        res1 `Assert.eq` Vector#{42} and res2 `Assert.eq` Vector#{-42, 42}
+    }
```
magnus-madsen, line 36: I think these tests would be easier to read if it was simply:

db1
db2

and/or mm1 and mm2.

That way the text aligns better.

I would also just call the predicate `A`.

main/test/flix/Test.Exp.Fixpoint.KeywordSolve.flix
```diff
@@ -0,0 +1,69 @@
...
+        };
+        let otherProg = #{
+            ASolve03(-x) :- BSolve03(x).
+        };
+        let k = solve solved, otherProg;
+        let res1 = query solved select x from ASolve03(x);
+        let res2 = query k select x from ASolve03(x);
+        res1 `Assert.eq` Vector#{1, 2} and res2 `Assert.eq` Vector#{-2, -1, 1, 2}
+    }
+
```
magnus-madsen, line 92: I feel like we are missing some tests that facts from one program does not leak into another. Abstractly:

db1 = #{A(1)}
db2 = #(B(1)}
solve db1, db2
solve db1 // should not contain B fact
solve db2 // should not contain A fact

main/test/flix/Test.Exp.Fixpoint.KeywordSolve.flix
JumpTwiice: The issue with calling the predicate `A` is that it might not stratify if another test/program uses the same predicate in a negative atom.

magnus-madsen: I see

magnus-madsen: But I think tests are independent, so only within the same file.

magnus-madsen (approved):

Issue #11066, opened 2025-07-09
Rework Datalog engine to use the new `foreach` construct


magnus-madsen:
Depends on #11016

I think, style-wise, we should use *nested* foreach loops to emphasize the loop structure. 

A good file to start with is `Fixpoint3.Phase.Stratifier`  to ensure we agree on the desired style/outcome.

magnus-madsen:
Fixed in master.

PR #11244, opened 2025-08-07
chore: use new `foreach` statement in Datalog engine

CasperDN:
See #11066.

Missing:
1. `ForEach` on `BPlusTree`.
2. A zip view?

### DIFF
--- a/main/src/library/Fixpoint3/Interpreter.flix
+++ b/main/src/library/Fixpoint3/Interpreter.flix
@@ -149,17 +149,16 @@ mod Fixpoint3.Interpreter {
             case Some((_, v)) => v + 1
         };
         let indexes = Array.empty(rc, indexNum);
-        indexMap |>
-            Map.forEach(relSym -> searches -> {
-                let tuples = Map.getWithDefault(relSym, BPlusTree.empty(rc), facts);
-                searches |> Vector.forEachWithIndex(num -> search -> {
-                    let pos = match Map.get((relSym, num), posMap) {
-                        case Some(v) => v
-                        case None => bug!("In Fixpoint.Interpreter.interpretWithInput: ${relSym} has no index!")
-                    };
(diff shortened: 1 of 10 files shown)


magnus-madsen:
- You are welcome to add a `ForEach` instance on your BTree
- @rywiese We don't know how to zip. Could you provide us some examples where you need it?

main/src/library/Fixpoint3/Interpreter.flix
```diff
@@ -149,17 +149,16 @@ mod Fixpoint3.Interpreter {
...
-                    let pos = match Map.get((relSym, num), posMap) {
-                        case Some(v) => v
-                        case None => bug!("In Fixpoint.Interpreter.interpretWithInput: ${relSym} has no index!")
-                    };
-                    storeIndex(mkIndex(rc, relSym, search, tuples), pos, indexes)
-                })
-            });
+        foreach ((relSym, searches) <- indexMap) {
+            let tuples = Map.getWithDefault(relSym, BPlusTree.empty(rc), facts);
+            searches |> Vector.forEachWithIndex(num -> search -> {
```
magnus-madsen, line 154: Why no foreach here?

CasperDN: That's a mistake. I'll fix it when I do the `BPlusTree`.

magnus-madsen: You can do a directory search for "forEach"

CasperDN:
> * @rywiese We don't know how to zip. Could you provide us some examples where you need it?

Do you mean where we would like to zip? Then here
```
foreach ((rule, ruleNum) <- List.zip(stratum, rulesWithId)) {
    compileRule(inserts, rule, ruleNum, predicates, counter)
};
```
I would simply use it to avoid creating a new list, just go through the two lists at once.

PR #11246, opened 2025-08-07
feat: switch to Datalog3 engine

CasperDN:


magnus-madsen (approved):

magnus-madsen:
@stephentetley FYI: This PR replaces the old Datalog engine by a new one that is about 10x-20x faster and much better tested.

stephentetley:
Thanks Magnus, I'll try to pick up Flix again in the next couple of weeks.

magnus-madsen:
> Thanks Magnus, I'll try to pick up Flix again in the next couple of weeks.

@stephentetley 

Welcome back :-)

You might enjoy the new `pquery`. It still being polished and tested, but should work. 

Documentation to come, but here is an example:

https://github.com/flix/flix/blob/master/examples/fixpoints/train-schedule.flix

PR #11256, opened 2025-08-07
refactor: use instantiation information in monomorpher

mlutze:


mlutze:
Currently a problem with wildcard type parameters, as demonstrated by


```
    def hasCookie(x: {| _}): Bool = typematch x {
        case _: {cookie = _ | _} => true
        case _: _ => false
    }

    @Test
    def testCookie01(): Bool = hasCookie({cookie = "woohoo"})
```

That fails, but the following succeeds.

```
    def hasCookie(x: {| r}): Bool = typematch x {
        case _: {cookie = _ | _} => true
        case _: _ => false
    }

    @Test
    def testCookie01(): Bool = hasCookie({cookie = "woohoo"})
```

I think we don't always add wildcard type parameters to the list of type parameters. Combine that with unsafe zip, and we have a bad time.

@JonathanStarup one more point in your point bowl for safe zip.

mlutze:
Problem seems to be that `_` gets kind `???`.

Kind inference should be able to resolve it, but types `_` get turned into fresh type var syms _sometimes_.

We should restructure namer to:

- visit explicit type parameters first
- visit and transform the spec under the type parameter environment
  - implicit parameters then get fresh names and are added to the env (only wildcards if explicit tparams are given)

mlutze:
Issue is now that instances defs and signatures can have type parameters that don't line up due to patterns like:

```
    pub def iterator(rc: Region[r], t: t): Iterator[Iterable.Elm[t], r + aef, r] \ (r + aef) where Iterable.Aef[t] ~ aef
```

mlutze:
I think we may be able to fix this by adding instantiation information also to instance definitions.

The assumption before was that instantiation information would just be the instance type, but due to the issue above, that is not exactly true.

Moreover, the current solution assumes that ordering of type parameters is consistent between instance and trait, but that's not the case in general.

PR #11258, opened 2025-08-07
feat: cli find project's root directory

### LINKED ISSUE #11150: CLI: detect toml in parent directory
If I `cd` into the `src` directory and execute `flix build`, the compiler will gladly build everything and put it in the `src` folder, because it searches for `*.flix`.

Maybe it should look up one directory and see that there's a `flix.toml` there and use that as the working directory instead.

gwydd12:
This PR implements the suggestion to locate the root working directory containing the flix.toml file. This enhancement ensures that the current working directory (cwd) is consistently set to the project's root, regardless of where the command is executed. A new helper function findRootWorkingDirectory has been added to facilitate traversing the filesystem hierarchy.

This PR fixes #11150

### DIFF
--- a/main/src/ca/uwaterloo/flix/Main.scala
+++ b/main/src/ca/uwaterloo/flix/Main.scala
@@ -28,7 +28,7 @@ import ca.uwaterloo.flix.util.*
 
 import java.io.{File, PrintStream}
 import java.net.BindException
-import java.nio.file.Paths
+import java.nio.file.{Files, Path, Paths}
 
 /**
   * The main entry point for the Flix compiler and runtime.
@@ -40,7 +40,11 @@ object Main {
     */
   def main(argv: Array[String]): Unit = {
 
-    val cwd = Paths.get(".").toAbsolutePath.normalize()
+    val iwd = Paths.get(".").toAbsolutePath.normalize()
+    val cwd = findRootWorkingDirectory(iwd) match {
+      case Some(wd) => wd
+      case None => iwd
+    }
 
     // parse command line options.
     val cmdOpts: CmdOpts = parseCmdOpts(argv).getOrElse {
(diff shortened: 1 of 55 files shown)


magnus-madsen:
Thanks-- @jaschdoc and I will take a look once he is back.

main/src/ca/uwaterloo/flix/Main.scala
```diff
@@ -600,4 +601,22 @@ object Main {
...
+    val validFlixToml = "flix.toml"
+
+    def getParentDirectories(path: Path): Iterator[Path] =
+      Iterator.iterate(path)(_.getParent).takeWhile(_ != null)
+
+    getParentDirectories(cwd)
+      .map(_.resolve(validFlixToml))
+      .find(Files.exists(_))
+  }
+
```
jaschdoc: What happens if the current project path `.` has a valid `flix.toml` but I accidentally put a `flix.toml` in `..`? It seems like it should *first* check if `flix.toml` exists in `.` and then only check parent dirs if it is not found.

Additionally, maybe there should be some limit on how far up the directory tree it should look to avoid checking to `/`? Thoughts on this @magnus-madsen ?

gwydd12: Hey, it should have exactly this behavior. It searches all the directories from the `cwd`, which is `.`, and all parent directories. Therefore, it should find the `flix.toml` inside the `.`. 

Regarding the limitation of looking up the directory tree to a certain point, I'm unsure whether this is beneficial. If I'm not mistaken, `cargo` also looks up the directory tree until it reaches `/`. A restriction might lead to limitation in project structures.

Thanks for the review :)

jaschdoc: > Regarding the limitation of looking up the directory tree to a certain point, I'm untsure whether this is beneficial. If I'm not mistaken, `cargo` also looks up the directory tree until it reaches `/`. A restriction might lead to limitation in project structures.

I agree, it seems like a bad restriction that you can only open subdirectories up to a specific depth, but maybe in a future iteration, the user could specify a number for the depth.

gwydd12:
@jaschdoc I found a mistake in my code. Silly me accidentally returned the entire path, including the file name. It should be fixed now. 
An idea that came up while finding my mistake was having testcases that execute the possible shell commands. As for this mistake, the `build-jar` command threw an exception.

jaschdoc:
Hey @gwydd12 apologies for the long response time. We appreciate the contribution, but we have started overhauling the cli and package manager, so issues like this will be worked out during the overhaul. I'm therefore closing this PR, but feel free to contribute in some other way.
For the same reason I will also close https://github.com/flix/flix/pull/9912
Have a nice day :)

gwydd12:
This makes total sense, thanks for the response. And if I can help in anyway feel free to assign me to a task if you have one in mind :) .

Issue #10930, opened 2025-06-20
Implement Tcp Echo Server using the new effects


magnus-madsen:
https://en.wikipedia.org/wiki/Echo_Protocol

magnus-madsen:
I guess we also need an effect to CONNECT to remote server :)

rywiese:
@magnus-madsen is the thought this should be an example app? or maybe two example apps (client and server)? not something to be added to std lib?

and is there any support for cli args (for ip/port) or is it cool to just hardcode those things

magnus-madsen:
Yeah, two example apps

magnus-madsen:
You can hard code ip/port

PR #11263, opened 2025-08-08
feat: add echo server and client example apps

rywiese:
fixes https://github.com/flix/flix/issues/10930


examples/apps/echo-server/src/Server.flix
```diff
@@ -0,0 +1,41 @@
+mod Server {
+    import java.nio.charset.StandardCharsets
+
+    use Peekable.readLine
+    use Writable.writeLine
+
+    pub def launch(ipAddr: IpAddr, port: Int16): Result[IoError, Unit] \ TcpBindWithResult + TcpAcceptWithResult + Logger + Sys + Net + IO =
```
mlutze: Ideally, we'd only have abstract effects here, then all the primitive effects show up in main (potentially via some handler).

I guess Sys is from some of line separator stuff we saw before. What's causing the `Net` effect?

mlutze: (Ditto for client)

rywiese: it was not needed. removed

mlutze (commented):
Looks good.

I already regret the Sys + IO stuff for decoding...


I just suggest that we wrap the `main` so it does a nice message if there's a `Result.Err` on binding.

PR #11238, opened 2025-08-06
chore: add tests for Fixpoint stratification

JumpTwiice:
Related to https://github.com/flix/flix/issues/11065.

PR #11267, opened 2025-08-08
chore: reformat Fixpoint tests

JumpTwiice:
Fixes https://github.com/flix/flix/pull/11238#discussion_r2262021387.


JumpTwiice:
I will also make a separate PR that fixes the flipped use of `Assert.eq`.

PR #11268, opened 2025-08-08
fix: track wildcards properly in Resolver

mlutze:
We used to track wildcard types in the Namer, but we did it improperly, so that new type variables would be generated in the Resolver, which don't correspond to those in the Namer.
This meant that type parameters for wildcards didn't line up with the types in formal parameters, etc. For example:

```scala
def length(l: List[_]): Int32 = ???
```

This function implicitly has one type parameter `_`.

Namer identified this and made it look like:

```scala
def length[_$1](l: List[_]): Int32 = ???
```

But then Resolver came along and did this:

```scala
def length[_$1](l: List[_$2]): Int32 = ???
```

Mismatch!

This PR changes so that we ignore wildcards in the Namer, and track the new implicit type parameters as they are generated in the Resolver.

Namer leaves it how it found it:

```scala
def length(l: List[_]): Int32 = ???
```

While Resolver adds the correct type parameter.

```scala
def length[_$1](l: List[_$1]): Int32 = ???
```

The controversial part of this is that I introduced a mutable parameter `RecordWild` that tracks types as they are visited.

I did this to avoid returning a list from visitType, since that's used in `visitExpression`, so it would be rather cumbersome.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
@@ -2025,6 +2025,8 @@ object Resolver {
 
   /**
     * Performs name resolution on the given formal parameter `fparam0` in the given namespace `ns0`.
+    *
+    * Wildness may be set to `RecordWild` in nonlocal definitions and signatures to track the symbols of wildcard types.
     */
   private def resolveFormalParam(fparam0: NamedAst.FormalParam, wildness: Wildness, scp0: LocalScope, taenv: Map[Symbol.TypeAliasSym, ResolvedAst.Declaration.TypeAlias], ns0: Name.NName, root: NamedAst.Root)(implicit scope: Scope, sctx: SharedContext, flix: Flix): Validation[ResolvedAst.FormalParam, ResolutionError] = {
     val tVal = traverseOpt(fparam0.tpe)(resolveType(_, Some(Kind.Star), wildness, scp0, taenv, ns0, root))
@@ -2956,6 +2958,8 @@ object Resolver {
           // We use Top scope because these lookups only occur at top level
           Result.Ok(LowerType.Var(Symbol.freshUnkindedTypeVarSym(VarText.SourceText(ident.name), ident.loc)))
         case Wildness.RecordWild(syms) =>
+          // ALERT!
+          // Here we mutate the RecordWild list because we are visiting a wildcard type!
           val sym = Symbol.freshUnkindedTypeVarSym(VarText.SourceText(ident.name), ident.loc)
           syms.append((ident -> sym))
           Result.Ok(LowerType.Var(sym))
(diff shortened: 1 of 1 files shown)


mlutze:
This helps support changes to monomorpher to support indexed effects.

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -3662,13 +3669,26 @@ object Resolver {
...
-    * Enum indicating whether a variable may be a wildcard.
+    * Enum indicating whether a type variable may be a wildcard.
     */
   private sealed trait Wildness
 
   private object Wildness {
+    /**
+      * Indicates that wildcard type variables should be recorded.
+      *
+      * This is used to track variables that must be added to a type signature.
```
magnus-madsen, line 3684: Can you add some comment about it being mutable and where it is mutated.

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -2952,6 +2955,10 @@ object Resolver {
         case Wildness.AllowWild =>
           // We use Top scope because these lookups only occur at top level
           Result.Ok(LowerType.Var(Symbol.freshUnkindedTypeVarSym(VarText.SourceText(ident.name), ident.loc)))
+        case Wildness.RecordWild(syms) =>
+          val sym = Symbol.freshUnkindedTypeVarSym(VarText.SourceText(ident.name), ident.loc)
+          syms.append((ident -> sym))
```
magnus-madsen, line 2964: Add some big warning

main/src/ca/uwaterloo/flix/language/phase/Resolver.scala
```diff
@@ -2023,8 +2026,8 @@ object Resolver {
   /**
     * Performs name resolution on the given formal parameter `fparam0` in the given namespace `ns0`.
     */
-  private def resolveFormalParam(fparam0: NamedAst.FormalParam, scp0: LocalScope, taenv: Map[Symbol.TypeAliasSym, ResolvedAst.Declaration.TypeAlias], ns0: Name.NName, root: NamedAst.Root)(implicit scope: Scope, sctx: SharedContext, flix: Flix): Validation[ResolvedAst.FormalParam, ResolutionError] = {
-    val tVal = traverseOpt(fparam0.tpe)(resolveType(_, Some(Kind.Star), Wildness.AllowWild, scp0, taenv, ns0, root))
+  private def resolveFormalParam(fparam0: NamedAst.FormalParam, wildness: Wildness, scp0: LocalScope, taenv: Map[Symbol.TypeAliasSym, ResolvedAst.Declaration.TypeAlias], ns0: Name.NName, root: NamedAst.Root)(implicit scope: Scope, sctx: SharedContext, flix: Flix): Validation[ResolvedAst.FormalParam, ResolutionError] = {
```
magnus-madsen, line 2031: Add warning to the doc

magnus-madsen:
I have approved the PR, but to avoid the taint of accepting mutation, you will have to press the merge button yourself.

https://en.wikipedia.org/wiki/Hand_washing

mlutze:
> I have approved the PR, but to avoid the taint of accepting mutation, you will have to press the merge button yourself.
> 
> https://en.wikipedia.org/wiki/Hand_washing

I'm pretty sure Pontius Pilatus didn't go to the Good Place.

Issue #11058, opened 2025-07-08
Datalog3 Master Issue


magnus-madsen:
- [ ] #11026
- [ ] #11023
- [ ] #11022
- [ ] #11011
- [ ] #4505

CasperDN:
The JProfiler is a bit problematic. It seems to have a hard time with parallelism/virtual threads. We could disable it, but it would skew results somewhat.

magnus-madsen:
> The JProfiler is a bit problematic. It seems to have a hard time with parallelism/virtual threads. We could disable it, but it would skew results somewhat.

Do you use sampling?

CasperDN:
We did in those tests yes. We could try instrumentation.

CasperDN:
Preliminary performance tests show that the vast majority of time spent on compilation is in `Lowering` / computing the `Boxing` information. I see this as a hint that we could benefit greatly from #11086.

magnus-madsen:
> Preliminary performance tests show that the vast majority of time spent on compilation is in `Lowering` / computing the `Boxing` information. I see this as a hint that we could benefit greatly from [#11086](https://github.com/flix/flix/issues/11086).

Is the time in Lowering spent on fact or rule compilation? Most programs are small, but perhaps some work is done for every fact and that is expensive?

CasperDN:
For a simple program (nearly no facts) 41% is used for unification of types and 15% for compiling fact(s).
The tested program was:
```
let ff = _x -> Vector#{};
let pr = #{
    A(1).
    R(a, b) :- A(x), let (a, b) = ff(x).
};
let actual = query pr select (x, y) from R(x, y);
```
I admit that this is not an interesting or representative program, but the idea was simply to look at a 'minimal' case and see how we could minimize time usage. The idea being that Datalog might also be useful for small computation/logic tasks and we wanted to look at the overhead of compiling the program.

magnus-madsen:
Fixed in master.

PR #11270, opened 2025-08-08
chore: make `pquery` return an open row

JumpTwiice:
Related to https://github.com/flix/flix/issues/11058.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/typer/SchemaConstraintGen.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/typer/SchemaConstraintGen.scala
@@ -93,9 +93,9 @@ object SchemaConstraintGen {
         val (withRow, resultRow) = withh.foldRight((mkAnySchemaRowType(loc1), mkAnySchemaRowType(loc1))) {
           case (pred, (acc1, acc2)) =>
             val relType = Type.freshVar(Kind.Predicate, loc1)
-            val row1 = Type.mkSchemaRowExtend(pred, relType, acc1, loc1)
-            val row2 = Type.mkSchemaRowExtend(pred, relType, acc2, loc1)
-            (row1, row2)
+            val withRow1 = Type.mkSchemaRowExtend(pred, relType, acc1, loc1)
+            val resultRow1 = Type.mkSchemaRowExtend(pred, relType, acc2, loc1)
+            (withRow1, resultRow1)
         }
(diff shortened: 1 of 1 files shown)


main/src/ca/uwaterloo/flix/language/phase/typer/SchemaConstraintGen.scala
```diff
@@ -90,14 +90,14 @@ object SchemaConstraintGen {
...
-        val (openRow, resultRow) = withh.foldRight((mkAnySchemaRowType(loc1), Type.mkSchemaRowEmpty(loc1))) {
+        val (withRow, resultRow) = withh.foldRight((mkAnySchemaRowType(loc1), mkAnySchemaRowType(loc1))) {
           case (pred, (acc1, acc2)) =>
             val relType = Type.freshVar(Kind.Predicate, loc1)
-            val openRow = Type.mkSchemaRowExtend(pred, relType, acc1, loc1)
-            val closedRow = Type.mkSchemaRowExtend(pred, relType, acc2, loc1)
-            (openRow, closedRow)
+            val row1 = Type.mkSchemaRowExtend(pred, relType, acc1, loc1)
+            val row2 = Type.mkSchemaRowExtend(pred, relType, acc2, loc1)
+            (row1, row2)
```
magnus-madsen: since we are now entering the polish phase, I think some more meaningful names here would be useful :-)

magnus-madsen: Maybe we can just call them withRow1 and resultRow1 or something

JumpTwiice: Without the 1 should be fine

magnus-madsen: Will that not introduce shadowing? We try to avoid that

JumpTwiice:
> I would also add three tests somewhere: that we can ematch on a row with a larger ematch than necessary.

Separate PR after this.

main/src/ca/uwaterloo/flix/language/phase/typer/SchemaConstraintGen.scala
magnus-madsen: Or at least look like shadowing

JumpTwiice: with 1 it is

magnus-madsen (approved):

Issue #11272, opened 2025-08-08
Datalog: Add `WeederError2.IllegalLatticeProvenance`


magnus-madsen:
Given an illegal provenance query like:

```
pquery pr select R(123;456) with {A};
```

We should not crash in SchemaConstraintGen, but instead we should add a check to Weeder2 to give a nice error message.

PR #11277, opened 2025-08-09
chore: add error in Weeder2 for illegal lattice provenance

JumpTwiice:
Fixes https://github.com/flix/flix/issues/11272.

The error is only for the `select` atom and not potential lattice relations appearing in `with`.


magnus-madsen (approved):

magnus-madsen:
Excellent- thanks.

Issue #11172, opened 2025-07-31
ExtVar: Ensure uniform treatment of nullary constructors for both `ExtTag` and `ExtMatch`


magnus-madsen:

magnus-madsen:
There may be some design questions around how this integrates with Datalog atoms- which cannot be nullary, but can take Unit.

PR #11290, opened 2025-08-11
feat: add support for nullary extensible variant constructors

jaschdoc:
Closes https://github.com/flix/flix/issues/11172

### DIFF
--- a/main/test/flix/Test.Exp.ExtMatch.flix
+++ b/main/test/flix/Test.Exp.ExtMatch.flix
@@ -347,4 +347,15 @@ mod Test.Exp.ExtMatch {
             case Nullary => true
         }
     }
+
+    @Test
+    def testExtMatchDatalog01(): Bool = {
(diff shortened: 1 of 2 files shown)


main/test/flix/Test.Exp.ExtMatch.flix
```diff
@@ -338,4 +338,24 @@ mod Test.Exp.ExtMatch {
...
+            case Nullary => true
+        }
+    }
+
+    @Test
+    def testExtMatchDatalog01(): Bool = {
+        let p = #{
+            A(1).
+        };
+        let v = pquery p select A(1) with {A};
```
jaschdoc: I'm somewhat confused by the meaning of this. What does `with {A}` mean? `A` is not a nullary constructor and definitely not a unit constructor either but `A()` succesfully matches `Vector.get(0, v)`

Issue #11283, opened 2025-08-10
Datalog: Program unexpectedly type checks


magnus-madsen:
I would not expect this program to type check:

```
def main(): Unit \ IO = 
    let p = #{
        A(1).
    };
    let v = pquery p select A(1) with {A};
    ematch Vector.get(0, v) {
        case A() => println("a")
    }
```

I am not sure if the problem is in pquery or ematch?

jaschdoc:
Probably `ematch`, since there are still issues with nullary and unit constructors

magnus-madsen:
@jaschdoc Will you report back with your findings?

jaschdoc:
Yes. I will do some experimentation and investigation today.

jaschdoc:
Okay it was a simple issue. The weeder inserted a wildcard pattern if the tuple pattern was empty. Clearly a mistake.

```scala
    private def visitExtTagTermsPat(tree: Tree, seen: collection.mutable.Map[String, Name.Ident])(implicit sctx: SharedContext): Validation[Nel[ExtPattern], CompilationMessage] = {
      expect(tree, TreeKind.Pattern.Tuple)
      val patterns = pickAll(TreeKind.Pattern.Pattern, tree)
      mapN(traverse(patterns)(visitPattern(_, seen))) {
        case Nil => Nel(ExtPattern.Wild(tree.loc), Nil)
        case x :: xs => Nel(restrictToVarOrWild(x), xs.map(restrictToVarOrWild))
      }
    }
```
https://github.com/flix/flix/blob/e83e069faeb56ca466ef5b2eefdba89eb987e22c/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala#L2437

magnus-madsen:
@jaschdoc What was the conclusion here? Can we fix the type issues separately from the question of nullary madness?

magnus-madsen:
I guess maybe not?

magnus-madsen:
In any case, do we wait until the new pattern PR is in?

jaschdoc:
This is related to nullary constructors, and the meaning of `A()`. I have a fix for this in the linked PR, which inserts a unit pattern (this is the same behavior as functions ans enums), so `A()` becomes `A(())`. As mentioned in my previous comment, the weeder would insert a wildcard pattern instead of unit so `A()` is translated to `A(_)` which unifies with `A(1)`, hence the bug.

Issue #11284, opened 2025-08-10
Datalog: Nullary confusion


magnus-madsen:
This program crashes:

```
def main(): Unit \ IO = 
    let p = #{
        A().
        R() :- A().
    };
    let v = pquery p select R() with {A};
    ematch Vector.get(0, v) {
        case A => println("")
    }
```

with

```
ca.uwaterloo.flix.util.InternalCompilerException: Expected Type.Apply(_, _, _), but got Relation() (/Users/au216465/tmp/flix/Main.flix:6:13)
        at ca.uwaterloo.flix.language.phase.Lowering$.f$1(Lowering.scala:1777)
        at ca.uwaterloo.flix.language.phase.Lowering$.termTypesOfRelation(Lowering.scala:1779)
        at ca.uwaterloo.flix.language.phase.Lowering$.predicatesOfSchemaRow(Lowering.scala:1764)
        at ca.uwaterloo.flix.language.phase.Lowering$.predicatesOfExtVar(Lowering.scala:1752)
        at ca.uwaterloo.flix.language.phase.Lowering$.visitExp(Lowering.scala:829)
        at ca.uwaterloo.flix.language.phase.Lowering$.visitExp(Lowering.scala:444)
        at ca.uwaterloo.flix.language.phase.Lowering$.visitExp(Lowering.scala:445)
```

Not sure whether this is pquery or ematch or both.

jaschdoc:
Probably `ematch`, since there are still issues with nullary and unit constructors

magnus-madsen:
This one could be ematch, but I am leaning towards lowering. But how about you both take a look.

jaschdoc:
True, I now think this is related to `pquery`

magnus-madsen:
But I also think there could simple be a mismatch in concepts that can only be resolved by making some design choices.

magnus-madsen:
It seems extensible variants support all of:

```
ematch ? {
  case P => 
  case Q() => 
  case Q(u) => // u is unit
}
```

@jaschdoc Is that right? How is each different from the others? Or are P and Q the same?

Datalog on the other hand supports:

```
A()
A(())
```

what should they map to?

jaschdoc:
Extensible variants have the same semantics around constructors as normal variants.

In short:
1. `A != A()`
2. `A() == A(())`

[...]

PR #11291, opened 2025-08-11
fix: ensure ematch scrutinee type matches rules

jaschdoc:
The bug was that the weeder inserted a wildcard into `()`. In other words, the pattern `A()` became `A(_)`.
It now inserts a unit pattern in `A()` so it effectively becomes `A(())`. This is the same behavior as normal enums. I have added some more tests to verify that this is the case. Note that the user is still only allowed to write variables or wildcards in the pattern. The unit pattern is completely synthetic.

Related to https://github.com/flix/flix/issues/11284
Closes https://github.com/flix/flix/issues/11283


magnus-madsen:
A few questions:

- Why do we need to extend the patterns with Unit--- could we not just keep the empty list around?
- Assuming we do have to extend ExtPattern; do we have to do it in the Parser2, could we not just defer it all the way to Lowering?

I feels like shooting sparrows with railguns, but I could be wrong.

jaschdoc:
> Why do we need to extend the patterns with Unit--- could we not just keep the empty list around?

Not quite, since `A` and `A()` become ambiguous then. We could, however, wrap the pattern in an option. Then `A` has `None` for its pattern, `A()` has `Some(Nil)` and `A(x)` has `Some(x :: Nil)`. It only becomes weird from a user perspective when `A(()) == A()` for normal enums but `A(()) != A()` for extensible variants. I do not like such an inconsistency. I think it stems from the fact that we cannot declare / construct (normal) enums without a unit parameter:

```scala
// Invalid enum
enum A {
    case A()
}

// Valid enum
enum A {
    case A
}

// Valid enum
enum A {
    case A(Unit)
}
```

> Assuming we do have to extend ExtPattern; do we have to do it in the Parser2, could we not just defer it all the way to Lowering?

See the solution above using `Option` to disambiguate. This also removes `ExtPattern.Unit`.

jaschdoc:
I will leave this until we resolve the discussion around `A, A(), A(())`.

jaschdoc:
> Why do we need to extend the patterns with Unit--- could we not just keep the empty list around?

I have given this some more thought. I believe having a `Unit` type is the proper way to deal with this issue instead of having an empty list.
1. It is mirrors what normal patterns do, thus removing some implicit assumption that empty list means unit.
2. If an empty list is used, more special casing is needed in `ConstraintGen`, i.e., checking if the list is empty and unifying with the `Unit` type. Again, if we add the type, then the types guide the implementation and it becomes more readable and fool proof.
3. In the long run, if we ever were to merge match and extmatch, they now have the same structure, so merging should be smoother.

jaschdoc:
Where did we land on this?

magnus-madsen:
I think we are still in a holding pattern- but with the tentative plan to merge in the pattern refactor (and wildcard stuff first) and then circle back to this and probably adopt your proposal.

magnus-madsen:
@jaschdoc No longer blocked right?

What happens if you type `A((), ()) => ... `. What should happen?

jaschdoc:
Correct, we can proceed now.

`A((), ())` is a tag with two unit arguments. However, it is _not_ a valid pattern. You must still write `A(x, y)`

magnus-madsen:
> Correct, we can proceed now.
> 
> `A((), ())` is a tag with two unit arguments. However, it is _not_ a valid pattern. You must still write `A(x, y)`

"Where" is this checked?

jaschdoc:
I can add some tests :)

magnus-madsen:
Lets add some tests, indeed.

main/src/ca/uwaterloo/flix/language/ast/LoweredAst.scala
```diff
@@ -188,6 +188,8 @@ object LoweredAst {
 
     case class Wild(tpe: Type, loc: SourceLocation) extends ExtTagPattern
 
+    case class Unit(tpe: Type, loc: SourceLocation) extends ExtTagPattern
```
magnus-madsen: As a matter of consistent, I  think we always have the order: Wildcards, Variables, and then Unit.

The point being that Wild and Var are often treated similar.

So it is strange to have Unit in between them.

jaschdoc: Sounds good. I just went with this pattern in order of increasing complexity.

magnus-madsen (approved):

magnus-madsen:
Yay!

PR #11297, opened 2025-08-11
fix: issue error when `with` misses `{}`

### LINKED ISSUE #11280: Datalog: Parsing of `pquery` is too lenient
This program should issue a parse error or weeder error:

```
def main(): Unit = 
    let p = #{
        Edge(1, 2).
        Edge(y, x) :- Edge(x, y).
    };
    let _ = pquery p select A() with ;
    ()
```

it should not just compile without any error.

CasperDN:
Fixes #11280.

The error can be kind of hard to see in the editor. Closing the with-mark causes another error which makes it hard to see the problem.


CasperDN:
Result:
image

magnus-madsen:
Can you add a test case to `TestParser`-- Thanks!

PR #11302, opened 2025-08-11
chore: replace 'type class' terminology with 'trait' in Scala source files

magnus-madsen:


main/src/ca/uwaterloo/flix/util/tc/Debug.scala
```diff
@@ -19,7 +19,7 @@ import ca.uwaterloo.flix.api.Flix
 import ca.uwaterloo.flix.language.dbg.AstPrinter
 
 /**
-  * Type class for values that can be debugged.
+  * Trait for values that can be debugged.
```
mlutze, line 22: This one is referring to a Scala trait that operates like a type class (Flix trait). It's also in a directory called `tc`. I dunno what the rules are for that. :)

mlutze (approved):

main/src/ca/uwaterloo/flix/util/tc/Debug.scala
magnus-madsen: Yeah, fair point. But its still a trait.

Issue #11292, opened 2025-08-11
ExtVar: Consistenly parse patterns


jaschdoc:
From https://github.com/flix/flix/issues/11162#issuecomment-3174170693

Patterns are parsed separately and emit a pattern but also an `ExtPattern`. It should either only emit tokens related to `ExtPat` or `Pattern`. The latter approach would attempt to reuse as much code as possible around patterns in `Parser2` and `Weeder2` and "convert" the pattern to `ExtPat` in `Weeder2` while emitting appropriate errors. I like this approach since it reuses the resilient parsing around patterns and treats them equally from a user-perspective.
Only emitting `ExtPat` tokens in the parser requires more parsing functions, and more code duplication in the weeder too. However, the syntactic structure may be less confusing maybe?

PR #11316, opened 2025-08-12
feat: add `Tag` extensible pattern

jaschdoc:
This makes parsing less weird and improves consistency with normal patterns. This also sets the stage for allowing wildcard patterns and makes it easier to merge ext patterns and normal patterns in the future, e.g., if we merge it into a single match expression.

Related to https://github.com/flix/flix/issues/11173
Closes https://github.com/flix/flix/issues/11292

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
@@ -477,8 +477,8 @@ object ConstraintGen {
       case e: Expr.RestrictableChoose => RestrictableChooseConstraintGen.visitRestrictableChoose(e)
 
       case Expr.ExtMatch(exp, rules, loc) =>
-        val (tpe, eff) = visitExp(exp)
-        val (patTypes, ruleBodyTypes, effs) = rules.map(visitExtMatchRule).unzip3
+        val (scrutineeType, scrutineeEff) = visitExp(exp)
+        val (patTypes, ruleBodyTypes, ruleBodyEffs) = rules.map(visitExtMatchRule).unzip3
         val expectedRowType =
           patTypes.collect { case Left(tag) => tag }
             .foldRight(Type.mkSchemaRowEmpty(loc.asSynthetic)) {
@@ -488,10 +488,10 @@ object ConstraintGen {
             }
         val expectedExtensibleType = Type.mkExtensible(expectedRowType, loc.asSynthetic)
         val defaultPatternTvars = patTypes.collect { case Right(tvar) => tvar }
-        c.unifyAllTypes(tpe :: expectedExtensibleType :: defaultPatternTvars, loc)
+        c.unifyAllTypes(scrutineeType :: expectedExtensibleType :: defaultPatternTvars, loc)
         c.unifyAllTypes(ruleBodyTypes, loc)
         val resTpe = ruleBodyTypes.head // Note: We are guaranteed to have one rule.
-        val resEff = Type.mkUnion(eff :: effs, loc)
+        val resEff = Type.mkUnion(scrutineeEff :: ruleBodyEffs, loc)
         (resTpe, resEff)
 
       case KindedAst.Expr.Tag(symUse, exps, tvar, loc) =>


main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -478,12 +478,8 @@ object ConstraintGen {
...
       case Expr.ExtMatch(exp, rules, loc) =>
         val (tpe, eff) = visitExp(exp)
-        val (caseTpes, tpes, effs) = rules.map(visitExtMatchRule).unzip3
-        val expectedRowType = caseTpes.foldRight(Type.mkSchemaRowEmpty(loc)) {
-          case ((name, patTpes), acc) => Type.mkSchemaRowExtend(name, Type.mkRelation(patTpes, loc), acc, loc)
-        }
-        val expectedSchemaType = Type.mkExtensible(expectedRowType, loc)
-        c.unifyType(tpe, expectedSchemaType, exp.loc)
+        val (patTypes, tpes, effs) = rules.map(visitExtMatchRule).unzip3
+        c.unifyAllTypes(tpe :: patTypes, loc.asSynthetic)
```
jaschdoc: Two problems with this approach:
1. There is no way to enforce closed rows.
2. The scrutinee may not appear in any pattern but unify succesfully anyway.

I need to rethink this a bit.

jaschdoc: Example of problem 2:
```flix
ematch xvar A() {
    B() => ???
}
```

magnus-madsen: I think you have to build the open or closed row using a fold. No way around it.

main/src/ca/uwaterloo/flix/language/ast/WeededAst.scala
```diff
@@ -299,11 +299,17 @@ object WeededAst {
 
   object ExtPattern {
 
-    case class Wild(loc: SourceLocation) extends ExtPattern
+    sealed trait VarOrWild {
```
magnus-madsen: This does not extend ExtPattern. The structure here is a bit confusing to me.

I suppose there is only one ExtPattern, namely Tag. which then has VarOrWild?

What about rearranging:

```
case class ExtTag(label: Name.Label, pats: List[ExtTagPat], loc: SourceLocation) extends ExtPattern

sealed trait ExtTagPat  // could also be called ExtTagFragment

object ExtTagPat {
 // Wild, Var, and Error
}
```

magnus-madsen: Having ExtPattern as a trait makes little sense when it only has one implementation.

main/src/ca/uwaterloo/flix/language/ast/ops/TypedAstOps.scala
```diff
@@ -462,6 +458,21 @@ object TypedAstOps {
...
+  /**
+    * Returns the free variables in the given extensible pattern `pat0`.
+    */
+  private def freeVars(pat0: ExtPattern): Set[Symbol.VarSym] = pat0 match {
+    case ExtPattern.Wild(_, _) => Set.empty
+    case ExtPattern.Tag(_, pats, _, _) => pats.toSet.flatMap((v: ExtPattern.VarOrWild) => freeVars(v))
+    case ExtPattern.Error(_, _) => Set.empty
+  }
+
+  private def freeVars(v: ExtPattern.VarOrWild): Set[Symbol.VarSym] = v match {
```
magnus-madsen: Missing comment.

magnus-madsen:
I think the structure is the most important to discuss.

main/src/ca/uwaterloo/flix/language/ast/WeededAst.scala
jaschdoc: Ah yes but after this, `Wild` will be a valid pattern too

magnus-madsen: Is it not? Its really because the structure of patterns has two cases and the structure of arguments two patterns have two cases?

magnus-madsen: Can you propose an alternate structure?

magnus-madsen:
Lets try to progress (at least on the design). What do you think of this (for all ASTs except the first):

We exploit that -- after Weeding -- there can only be zero or one default (wildcard) case. But let us call it default.

```
case class ExtMatch(exp: Expr, rules: List[ExtMatchRule], default: Option[Expr], tpe: Type, eff: Type, loc: SourceLocation) extends Expr

case class ExtMatchRule(label: Name.Label, pats: List[ExtPattern], exp: Expr, loc: SourceLocation)

sealed trait ExtPattern

object ExtPattern {
 // Wild, Var, and Error
}
```

@jaschdoc Can we do this?

magnus-madsen:
Here ExtPattern might need to be extended with Unit which would only be allowed in ExtMatchRule with 1 pattern.

main/src/ca/uwaterloo/flix/language/ast/WeededAst.scala
jaschdoc: Yes, true it can be flattened like you propose. I just did not want to pollute the asts

jaschdoc:
I think the following structure is more flexible and true to the syntactic structure (basically outlining `VarOrWild` and renaming to `ExtTagPat`):

```scala

  case class ExtMatchRule(pat: ExtPattern, exp: Expr, loc: SourceLocation)

  sealed trait ExtPattern {
    def loc: SourceLocation
  }

  object ExtPattern {

    sealed trait VarOrWild {
      def loc: SourceLocation
    }

    case class Wild(loc: SourceLocation) extends ExtPattern

    case class Var(ident: Name.Ident, loc: SourceLocation) extends ExtPattern

    case class Tag(label: Name.Label, pats: List[ExtTagPattern], loc: SourceLocation) extends ExtPattern

    case class Error(loc: SourceLocation) extends ExtPattern
  }
  
  sealed trait ExtTagPattern {
    def loc: SourceLocation
  }
  
  object ExtTagPattern {

    case class Wild(loc: SourceLocation) extends ExtPattern with ExtTagPattern

    case class Var(ident: Name.Ident, loc: SourceLocation) extends ExtTagPattern
    
    case class Error(loc: SourceLocation) extends ExtPattern with ExtTagPattern
    
  }

```

main/src/ca/uwaterloo/flix/language/ast/WeededAst.scala
magnus-madsen: OK; shall we go with my proposal then?

jaschdoc:
> We exploit that -- after Weeding -- there can only be zero or one default (wildcard) case. But let us call it default.

Normal pattern matches allow multiple default cases in flix so maybe we should follow the same pattern here for consistency and simplicity

magnus-madsen:
> > We exploit that -- after Weeding -- there can only be zero or one default (wildcard) case. But let us call it default.
> 
> Normal pattern matches allow multiple default cases in flix so maybe we should follow the same pattern here for consistency and simplicity

That seems broken-- or old. I would disallow such things in the Weeder.

magnus-madsen:
I guess it could be argued for resilience, but it still just seems to so bad. I think it is over complicating things. Also, extensible variants are really not ADTs. They are a different animal that plays by different rules.

jaschdoc:
> Also, extensible variants are really not ADTs. They are a different animal that plays by different rules.

Yes, but from a syntactic user perspective, writing the pattern match should *feel* the same way.

jaschdoc:
> > > We exploit that -- after Weeding -- there can only be zero or one default (wildcard) case. But let us call it default.
> > 
> > 
> > Normal pattern matches allow multiple default cases in flix so maybe we should follow the same pattern here for consistency and simplicity
> 
> That seems broken-- or old. I would disallow such things in the Weeder.

Or exhaustiveness?

jaschdoc:
> I guess it could be argued for resilience, but it still just seems to so bad. I think it is over complicating things. Also, extensible variants are really not ADTs. They are a different animal that plays by different rules.

How do you propose we minimize complexity? It seems to me, this is the minimal way / path of least resistance.

magnus-madsen:
I would go with this:

```
case class ExtMatch(exp: Expr, rules: List[ExtMatchRule], default: Option[Expr], tpe: Type, eff: Type, loc: SourceLocation) extends Expr

case class ExtMatchRule(label: Name.Label, pats: List[ExtPattern], exp: Expr, loc: SourceLocation)

sealed trait ExtPattern

object ExtPattern {
 // Wild, Var, and Error
}
```

magnus-madsen:
> > Also, extensible variants are really not ADTs. They are a different animal that plays by different rules.
> 
> Yes, but from a syntactic user perspective, writing the pattern match should _feel_ the same way.

I am not really sure about that. These are different features. Like restrictable variants with their `choose` (which we may call `rmatch`).

jaschdoc:
> I would go with this:
> 
> ```
> case class ExtMatch(exp: Expr, rules: List[ExtMatchRule], default: Option[Expr], tpe: Type, eff: Type, loc: SourceLocation) extends Expr
> 
> case class ExtMatchRule(label: Name.Label, pats: List[ExtPattern], exp: Expr, loc: SourceLocation)
> 
> sealed trait ExtPattern
> 
> object ExtPattern {
>  // Wild, Var, and Error
> }
> ```

Problem is that when we add default cases, we can't always construct `ExtMatchRule`.

magnus-madsen:
> > I would go with this:
> > ```
> > case class ExtMatch(exp: Expr, rules: List[ExtMatchRule], default: Option[Expr], tpe: Type, eff: Type, loc: SourceLocation) extends Expr
> > 
> > case class ExtMatchRule(label: Name.Label, pats: List[ExtPattern], exp: Expr, loc: SourceLocation)
> > 
> > sealed trait ExtPattern
> > 
> > object ExtPattern {
> >  // Wild, Var, and Error
> > }
> > ```
> 
> Problem is that when we add default cases, we can't always construct `ExtMatchRule`.

The ExtMatch has the one optional default case.

Only the Weeder needs to deal with more than one default case.

jaschdoc:
Aha... I can agree to that, but it seems more confusing than just having the patterns in the AST. Is it your opinion that this is simpler?

magnus-madsen:
Yes

magnus-madsen:
It's just the default expression that's not in a match rule.

jaschdoc:
If we do it your way, then typing is a bit more awkward since the list of rules may be empty in which case the type (after reconstruction) must be a variable which shouldn't make sense

jaschdoc:
> I am not really sure about that. These are different features. Like restrictable variants with their `choose` (which we may call `rmatch`).

Btw, I agree that we shouldn't use funny scala inheritance with `VarOrWild`, but choose rules use the same `VarOrWild` trick. That's where I got the inspiration from.

magnus-madsen:
> > I am not really sure about that. These are different features. Like restrictable variants with their `choose` (which we may call `rmatch`).
> 
> Btw, I agree that we shouldn't use funny scala inheritance with `VarOrWild`, but choose rules use the same `VarOrWild` trick. That's where I got the inspiration from.

Sure, but let us make it better going forward.

magnus-madsen:
Why not just allow a wildcard pattern?

Then a default rule is one where the pattern is wildcard?

jaschdoc:
Here is the design we agreed upon
```haskell
Expr      ::= ...     | ExtMatch Expr (List ExtPat)
ExtPat    ::= Default | Tag (List ExtTagPat) | Err
ExtTagPat ::= Wild    | Var                  | Err
```

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
```diff
@@ -1022,14 +1022,14 @@ object Simplifier {
     val extVar = SimplifiedAst.Expr.Var(extName, tagType, exp.loc)
     val errorExp = SimplifiedAst.Expr.ApplyAtomic(AtomicOp.MatchError, List.empty, tpe, Purity.Impure, loc)
     val iftes = rules.foldRight(errorExp: SimplifiedAst.Expr) {
-      case (MonoAst.ExtMatchRule(label, pats, exp1, loc1), branch2) =>
+      case (MonoAst.ExtMatchRule(MonoAst.ExtPattern.Tag(label, pats, _, _), exp1, loc1), branch2) =>
```
magnus-madsen, line 1025: Maybe I am blind, but what happens if it is not `MonoAst.ExtPattern.Tag`

Is there not a missing case here? How does this compile? :-)

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -2438,22 +2446,22 @@ object Weeder2 {
       }
     }
 
-    private def visitExtTagTermsPat(tree: Tree, seen: collection.mutable.Map[String, Name.Ident])(implicit sctx: SharedContext): Validation[Nel[ExtPattern], CompilationMessage] = {
+    private def visitExtTagTermsPat(tree: Tree, seen: collection.mutable.Map[String, Name.Ident])(implicit sctx: SharedContext): Validation[Nel[ExtTagPattern], CompilationMessage] = {
```
magnus-madsen: We use `Nel` here , but not later?

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -2346,26 +2346,34 @@ object Weeder2 {
...
+    private def visitExtPattern(tree: Tree, seen: collection.mutable.Map[String, Name.Ident] = collection.mutable.Map.empty)(implicit sctx: SharedContext): Validation[ExtPattern, CompilationMessage] = {
       expect(tree, TreeKind.Pattern.Pattern)
-      val extTagPattern = tryPick(TreeKind.Pattern.ExtTag, tree)
-      extTagPattern match {
-        case Some(pat) => visitExtTagPattern(pat, seen)
-        case None =>
-          val error = UnexpectedToken(NamedTokenSet.ExtPattern, actual = None, SyntacticContext.Unknown, loc = tree.loc)
-          sctx.errors.add(error)
-          Validation.Failure(error)
+      tree.children.headOption match {
```
magnus-madsen, line 2351: What is the reason not to use one of the pickXs here? Is it not possible?

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -2346,26 +2346,34 @@ object Weeder2 {
...
+        case Some(subtree: Tree) => subtree.kind match {
+          case TreeKind.Pattern.Tag => visitExtTagPattern(subtree, seen)
+          // Avoid double reporting errors by returning a success here
+          case TreeKind.ErrorTree(_) => Validation.Success(ExtPattern.Error(subtree.loc))
```
magnus-madsen, line 2355: "Feels" like some map/flatMap operation would be more appropriate?

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -1066,18 +1071,41 @@ object ConstraintGen {
...
+  }
+
+  /**
+    * Generates constraints for the patterns inside the ext tag pattern.
     *
     * Returns the type of the pattern.
     */
-  private def visitExtPattern(pat0: KindedAst.ExtPattern)(implicit c: TypeContext): Type = pat0 match {
-    case ExtPattern.Wild(tvar, _) => tvar
+  def visitExtTagPattern(pat0: KindedAst.ExtTagPattern)(implicit c: TypeContext): Type = pat0 match {
```
magnus-madsen: private?

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -1116,12 +1144,11 @@ object ConstraintGen {
     *
     * Returns the name of the constructor, the types of the constructor, the type of the expression body, and its effect.
     */
-  private def visitExtMatchRule(rule: KindedAst.ExtMatchRule)(implicit c: TypeContext, root: KindedAst.Root, flix: Flix): ((Name.Pred, List[Type]), Type, Type) = rule match {
-    case KindedAst.ExtMatchRule(label, pats, exp, _) =>
-      val name = Name.Pred(label.name, label.loc)
-      val patTypes = pats.map(visitExtPattern)
+  private def visitExtMatchRule(rule: KindedAst.ExtMatchRule)(implicit c: TypeContext, scope: Scope, root: KindedAst.Root, flix: Flix): (Either[(Name.Pred, List[Type]), Type.Var], Type, Type) = rule match {
```
magnus-madsen: Can we reorder so the functions appear in a logical order:

ExtMatchRule
ExtPattern
ExtTagPattern

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -1066,18 +1071,41 @@ object ConstraintGen {
   }
 
   /**
-    * Generates constraints for the patterns inside the record label pattern.
+    * Generates constraints for the patterns inside the ext pattern.
+    *
+    * Returns the type of the pattern.
```
magnus-madsen: Uhm. I think some explanation is needed.

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -478,14 +478,19 @@ object ConstraintGen {
...
+        val (patTypes, branchExpTypes, effs) = rules.map(visitExtMatchRule).unzip3
+        val expectedRowType =
+          patTypes.collect { case Left(value) => value }
+            .foldRight(Type.mkSchemaRowEmpty(loc.asSynthetic)) {
+              case ((pred, tpes), acc) =>
+                val relation = Type.mkRelation(tpes, pred.loc.asSynthetic)
+                Type.mkSchemaRowExtend(pred, relation, acc, pred.loc.asSynthetic)
+            }
+        val expectedExtensibleType = Type.mkExtensible(expectedRowType, loc.asSynthetic)
+        val tvars = patTypes.collect { case Right(value) => value }
```
magnus-madsen: I think some better naming in the above code. For example, here tvars is really defaultTypes or similar.

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintGen.scala
```diff
@@ -478,14 +478,19 @@ object ConstraintGen {
...
         val (tpe, eff) = visitExp(exp)
-        val (caseTpes, tpes, effs) = rules.map(visitExtMatchRule).unzip3
-        val expectedRowType = caseTpes.foldRight(Type.mkSchemaRowEmpty(loc)) {
-          case ((name, patTpes), acc) => Type.mkSchemaRowExtend(name, Type.mkRelation(patTpes, loc), acc, loc)
-        }
-        val expectedSchemaType = Type.mkExtensible(expectedRowType, loc)
-        c.unifyType(tpe, expectedSchemaType, exp.loc)
-        c.unifyAllTypes(tpes, loc)
-        val resTpe = tpes.head // Note: We are guaranteed to have one rule.
+        val (patTypes, branchExpTypes, effs) = rules.map(visitExtMatchRule).unzip3
```
magnus-madsen: branchExpTypes -> ruleBodyTypes etc?

magnus-madsen:
What about test cases for wildcards? I can help make some with Claude, if needed.

main/src/ca/uwaterloo/flix/language/phase/Simplifier.scala
jaschdoc: After Lowering, we only have `Tag` so it is exhaustive. I will add full support for `Default` in a subsequent PR.

jaschdoc:
> What about test cases for wildcards? I can help make some with Claude, if needed.

Would be awesome, but the idea here is not to add default pattern just yet. This PR adds the infrastructure and reuses parsing from patterns and in another PR I will add support for `Default`.

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
jaschdoc: To avoid double reporting an error, i.e., both a parser error and then an `IllegalExtPattern` error. This mirrors the structure of `visitPattern`. But if we don't care about that, then we can just use `tryPick`.

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
jaschdoc: Hmm... not sure. You still want the `ICE` in there to signal that something went very wrong with parsing. If we use `map` then you still have to call `.getOrElse`. It feels like that, yes, but I think this might be less complicated in the end.

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
magnus-madsen: OK no that is fine

magnus-madsen:
Wildcards next? 🚀

jaschdoc:
> Wildcards next? 🚀

Yes!

PR #11319, opened 2025-08-12
fix: improved case set unification performance

Alex1005a:
Before unifying set formulas, you can try to get rid of a large number of trivial cases when the structure of formulas is similar. For example, for formulas `x & ~{Var}` and `y & ~{Var}`, substitution `[x -> y]` is suitable.

Related to https://github.com/flix/flix/pull/8753

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/unification/CaseSetUnification.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/unification/CaseSetUnification.scala
@@ -164,6 +164,12 @@ object CaseSetUnification {
     mkOr(mkAnd(p, mkNot(q)), mkAnd(mkNot(p), q))
 
   /**
+    * Translate [[Xor]] using the equation: a ⊕ b = (a ∪ b) - (a ∩ b) = (a ∪ b) ∩ ¬(a ∩ b)
+    */
+  private def elimXor(xor: Xor)(implicit univ: Set[Int]): SetFormula =
+    mkAnd(mkOr(xor.f1, xor.f2), mkNot(mkOr(xor.f1, xor.f2)))
+
+  /**
     * An atom is a constant or a variable.
     */
   private sealed trait Atom
@@ -248,6 +254,7 @@ object CaseSetUnification {
     case Not(tpe) => nnfNot(tpe)
     case Or(tpe1, tpe2) => Nnf.Union(nnf(tpe1), nnf(tpe2))
     case And(tpe1, tpe2) => Nnf.Intersection(nnf(tpe1), nnf(tpe2))
+    case xor: Xor => nnf(elimXor(xor))
   }
 
   /**
@@ -267,6 +274,7 @@ object CaseSetUnification {
       nnf(mkNot(tpe1)),
       nnf(mkNot(tpe2))
     )
+    case xor: Xor => nnf(mkNot(elimXor(xor)))
   }
 
   /**
--- a/main/src/ca/uwaterloo/flix/language/phase/unification/CaseSetZhegalkinUnification.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/unification/CaseSetZhegalkinUnification.scala
@@ -72,6 +72,7 @@ object CaseSetZhegalkinUnification {
     case SetFormula.Not(f1) => ZhegalkinExpr.mkCompl(toZhegalkin(f1))
     case SetFormula.And(x, y) => ZhegalkinExpr.mkInter(toZhegalkin(x), toZhegalkin(y))(alg, lat)
     case SetFormula.Or(x, y) => ZhegalkinExpr.mkUnion(toZhegalkin(x), toZhegalkin(y))(alg, lat)
+    case SetFormula.Xor(x, y) => ZhegalkinExpr.mkXor(toZhegalkin(x), toZhegalkin(y))(alg, lat)
   }
 
   /** Returns the given Zhegalkin expression as a SetFormula. */
@@ -94,10 +95,7 @@ object CaseSetZhegalkinUnification {
         // `c ⊕ t1 ⊕ t2 ⊕ ... ⊕ tn`
         terms.foldLeft(visitCst(cst): SetFormula) {
           case (acc, term) =>
-            val ft = visitTerm(term)
-            // To represent xor, the following equality is used:
-            // `x ⊕ y = (xᶜ ∩ y) ∪ (x ∩ yᶜ)`
-            SetFormula.mkOr(SetFormula.mkAnd(SetFormula.mkNot(acc), ft), SetFormula.mkAnd(acc, SetFormula.mkNot(ft)))
+            SetFormula.mkXor(acc, visitTerm(term))
         }
     }
   }
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/SemanticTokensProvider.scala
@@ -891,6 +891,7 @@ object SemanticTokensProvider {
     case TypeConstructor.CaseComplement(_) => false
     case TypeConstructor.CaseUnion(_) => false
     case TypeConstructor.CaseIntersection(_) => false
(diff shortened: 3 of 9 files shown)


magnus-madsen:
While this commendable, I think we have a much simpler and stronger approach in mind. We want to use the Zhegalkin polynomials: https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/unification/zhegalkin/ZhegalkinAlgebra.scala

I very recently made this implementation polymorphic in the underlying `BoolLattice`.

For restrictable variants, the `BoolLattice` should simple be based on the set of constructor (one for each restrictable variant). 

When we started to use Zhegalkin for effect sets, we gained massive speed-ups. I believe the same would play out here.

magnus-madsen:
Of course this refactoring is much more complex. Essentially CaseSetUnification have be replaced by something like `EffUnification3`. Formulas have to be mapped back and forth. It is really messy. But it has the right foundation.

Alex1005a:
But wouldn't such elimination of trivial cases still be faster? For example, in SetUnification there is a [trivial](https://github.com/flix/flix/blob/6d390c64be52b0395746868f5f331f4c4ad32c30/main/src/ca/uwaterloo/flix/language/phase/unification/set/SetUnification.scala#L258) function that is responsible for this.

Alex1005a:
It seems that using Zhegalkin polynomials gives a good performance boost. I'll try to add elimination of trivial cases and see if it makes sense.

magnus-madsen:
> It seems that using Zhegalkin polynomials gives a good performance boost. I'll try to add elimination of trivial cases and see if it makes sense.

Do you have any concrete numbers?

magnus-madsen:
I was also thinking we can reuse SetUnification-- but I forget if it has been made polymorphic yet.

Alex1005a:
> Do you have any concrete numbers?

Specifically for the example from https://github.com/flix/flix/pull/8753 everything is clear without measurements, since with the old unification the `Typer` phase did not end on my computer in any acceptable time, while using Zhegalkin polynomials the phase takes about a second of time.

magnus-madsen:
OK; lets try to merge this in. Even if as an intermediate step. Ill add some comments.

Is your name already in AUTHORS.md?

main/src/ca/uwaterloo/flix/language/phase/unification/CaseSetSubstitution.scala
```diff
@@ -135,4 +135,11 @@ case class CaseSetSubstitution(m: Map[Int, SetFormula]) {
     }
     Substitution(map)
   }
+
+  /**
+    * Check that `this` substitution variables do not overlap with `that` substitution variables
```
magnus-madsen: Can we make the statement more precise because a substitution has variables in both its domain and range (i.e. keys and values).

main/src/ca/uwaterloo/flix/language/phase/unification/CaseSetUnification.scala
```diff
@@ -35,29 +35,48 @@ object CaseSetUnification {
...
+    /// Try get rid of trivial cases.
+    /// If failed, then run the expensive boolean unification algorithm.
     ///
+    tryFastUnify(input1, input2, renv)
+      .orElse(booleanUnification(input1, input2, renv, univ)).map {
+        case subst => subst.toTypeSubstitution(enumSym, env)
+      }
+  }
+
+  private def tryFastUnify(tpe1: SetFormula, tpe2: SetFormula, renv: Set[Int]): Option[CaseSetSubstitution] = {
```
magnus-madsen, line 66: Add documentation

main/src/ca/uwaterloo/flix/language/phase/unification/CaseSetZhegalkinUnification.scala
```diff
@@ -0,0 +1,80 @@
+package ca.uwaterloo.flix.language.phase.unification
```
magnus-madsen, line 16: Add copyright header. If there is a bit copy-paste then retain the original names and add your own name.

main/src/ca/uwaterloo/flix/language/phase/unification/CaseSetZhegalkinUnification.scala
```diff
@@ -0,0 +1,80 @@
...
+    val f2 = toZhegalkin(input2)(renv, alg, lat)
+    val res = SveAlgorithm.unify(f1, f2, renv)(alg)
+    res.map(subst => subst.m.foldLeft(IntMap.empty[SetFormula]) {
+      case (acc, (x1, e)) => acc.updated(x1, toSetFormula(e)(univ))
+    }).map {
+      CaseSetSubstitution(_).toTypeSubstitution(enumSym, env)
+    }
+  }
+
+  private def toZhegalkin(f: SetFormula)(implicit renv: Set[Int], alg: ZhegalkinAlgebra[FiniteSet[Int]], lat: BoolLattice[FiniteSet[Int]]): ZhegalkinExpr[FiniteSet[Int]] = f match {
```
magnus-madsen, line 69: Add documentation.

The key here thing here to document is that the universe is finite and known from the declaration of res. var.

main/src/ca/uwaterloo/flix/language/phase/unification/CaseSetZhegalkinUnification.scala
```diff
@@ -0,0 +1,80 @@
...
+    case SetFormula.Cst(s) => ZhegalkinExpr(FiniteSet(s), Nil)
+    case SetFormula.Var(x) => ZhegalkinExpr(lat.Bot, List(ZhegalkinTerm(lat.Top, SortedSet(ZhegalkinVar(x, flexible = !renv.contains(x))))))
+    case SetFormula.Not(f1) => ZhegalkinExpr.mkCompl(toZhegalkin(f1))
+    case SetFormula.And(x, y) => ZhegalkinExpr.mkInter(toZhegalkin(x), toZhegalkin(y))(alg, lat)
+    case SetFormula.Or(x, y) => ZhegalkinExpr.mkUnion(toZhegalkin(x), toZhegalkin(y))(alg, lat)
+  }
+
+  private def toSetFormula(z: ZhegalkinExpr[FiniteSet[Int]])(implicit univ: Set[Int]): SetFormula = {
```
magnus-madsen, line 79: Add docs

magnus-madsen:
@Alex1005a Remember to mark comments as resolved. Thanks!

magnus-madsen:
Also, would this allow us to deal some old unification code?

Alex1005a:
Do you mean remove old `CaseSetUnification` code? I think it is possible. This module is currently used only in tests.

magnus-madsen:
> Do you mean remove old `CaseSetUnification` code? I think it is possible. This module is currently used only in tests.

Yes. But let us do it in a follow-up PR.

magnus-madsen:
Changes look good. Just waiting for CI to complete.

magnus-madsen:
Unfortunately something seems to have happened to the tests. Usually they take 14min, but now the duration is 21+ minutes?

Alex1005a:
It seems to have fixed the speed of passing tests. Apparently the formulas `x ⊕ y = (xᶜ ∩ y) ∪ (x ∩ yᶜ)` were too big. To fix this, I added `Xor` to `SetFormula`. I didn't change anything in the parser, so the user himself will not be able to use xor, but he can see it, for example, in the error message.

magnus-madsen (approved):

magnus-madsen:
@Alex1005a Thanks! Congratulations on landing the first PR.

Will you look into if there is now some old code that can be removed?

If you want to push further:

- The new unification algorithm is based on Zhegalkin (which you know use) and `set.SetUnification.scala` which is not yet used.

We should work towards using both components for both `Bool2` (i.e. type level Booleans) and restrictable variants.

magnus-madsen:
Can we remove CaseSetUnification now? Or at least a lot of code in it?

magnus-madsen:
Maybe we can also get rid of SetFormula in favor of set.SetFormula. (Yes, the naming is terribly confusing.)

Alex1005a:
Yes, we can still get rid of CaseSetUnification. It is only used in tests.

PR #11325, opened 2025-08-13
fix: renamePredSyms failing to rename rules in `Provenance` enum

### LINKED ISSUE #11321: Datalog: `psolve` should be idempotent
Instead this crashes:

```
def testPSolveIdempotent(): Bool =
    let db = #{
        Edge(1, 2). Edge(2, 3). Edge(3, 1). Edge(1, 4).
    };
    let pr = #{
        Path(x, y) :- Edge(x, y).
        Cycle(x)   :- Path(x, x).
    };
    let pm = psolve (psolve db, pr);
    let pp = pquery pm select Cycle(2) with {Edge};
    Vector.map(v -> ematch v { case Edge(x, y) => (x, y) }, pp) `Assert.eq` Vector#{(2, 3), (3, 1), (1, 2)}
```

CasperDN:
Should fix #11321

In the current implementation successive calls to `solve` and `psolve` should be idempotent. Is this what we want? As in do we want that
```psolve (psolve X) == solve (psolve X)```
I would not say it is a guaranteed property, just a consequence of the current `Solver.runWithStratification` being lazy.

An argument against this property being allowed would be that it allows people to confuse the semantics and could lead to people depending on it.

### DIFF
--- a/main/src/library/Fixpoint3/Phase/RenamePredSyms.flix
+++ b/main/src/library/Fixpoint3/Phase/RenamePredSyms.flix
@@ -69,12 +69,14 @@ mod Fixpoint3.Phase.RenamePredSyms {
                 })
                 |> Model
         case Provenance(rules, facts) =>
+            let transformer = mapForward(mapping);
+            let newRules = remapConstraints(rules, transformer);
             (Map#{}, facts)
                 ||> Map.foldLeftWithKey(acc -> match RelSym.Symbol(predSym, arity, den) -> v ->
                     let newSym = RelSym.Symbol(mapForward(mapping, predSym), arity, den);
                     Map.insert(newSym, v, acc)
                 )
-                |> Provenance(rules)
+                |> Provenance(newRules)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by CasperDN** [comment]
An argument for keeping the property would be that it is in line with 'only composition destroys provenance information'.

**Review by magnus-madsen — APPROVED** [review]

**Comment by magnus-madsen** [comment]
Does `psolve` throw away the rules? E.g. `solve (psolve db, pr) <+> db2` cannot introduce new facts other than those in `db2`?

**Comment by CasperDN** [comment]
`psolve` does not throw away the rules. Composition does, though. We should have that
```
solve (psolve db, pr) <+> db2 === solve (solve db, pr) <+> db2
```
Not because of `psolve`, but due to the fact that during composition a `Provenance` is treated as though it is a `Model` i.e. we ignore the rules. In particular the implementation of union says:
```
case (Provenance(_, _), _) => union(provenanceToModel(d1), d2)
```

Issue #11275, opened 2025-08-09
Datalog: Ensure argument order to Assert.eq


magnus-madsen:
We seem to have often flipped expected and actual. I have done this myself. We should  go through the tests and fix that.

I would almost say that - except for simple tests - it might even be worth it to write:

```
let expect = ...
let actual = ...
```

mlutze:
Named arguments for the function?

magnus-madsen:
Maybe yeah

magnus-madsen:
I think this is low enough priority that we can leave it for later and for when look into the Assert effect.

PR #11335, opened 2025-08-13
chore: reformat `Test.Exp.Fixpoint.Application`

JumpTwiice:
Related to https://github.com/flix/flix/issues/11275.


magnus-madsen:
Thanks-- but let me suggest we defer some of the reformatting till later; given that there still some bugs to chase :)

PR #11344, opened 2025-08-13
test: add tests for `query` and `pquery` on large minimal models

magnus-madsen:


magnus-madsen:
@jaschdoc Do we know what is happening here with this error? Have we seen it before?

mlutze (approved):

jaschdoc:
On the face of it, it looks like two types aren't properly unified... my hunch is that during lowering an incorrect type is used to construct the result of the `pquery`. I will have to investigate more. It does not seem like an issue with `ematch` but more that `ExTag` has an open row

jaschdoc:
Wow okay it seems your tests never touches `ExtTag` or `ExtMatch` in `ConstraintGen`. I think it's all related to `FixpointQueryWithProvenance`.

jaschdoc:
Okay after a bit more investigation my hunch is that there is a mismatch between constraintgen for pquery and extensible variants so that when even though lowering generates the correct expressions, the types are incorrect, so the type verifier reports an error. I will have to investigate more later but for now that's my idea.

magnus-madsen:
OK; so a bug in inference or Lowering or both?

magnus-madsen:
@jaschdoc Could it be a bug in the TypeVerifier? The programs actually run correctly it seems.

magnus-madsen:
Or its because we generate incorrect types, but somehow it does not harm us.

jaschdoc:
> Or its because we generate incorrect types, but somehow it does not harm us.

Yes, I definitely think it's "just" a type mismatch, since the generated expressions (from Lowering) are correct.

magnus-madsen:
OK; I will let you do some digging.

magnus-madsen:
@jaschdoc I fixed the bug. It was insidious.

jaschdoc:
Ouch yes that would have taken me a while

jaschdoc:
Nicely spotted though :)

magnus-madsen:
Yeah. It was my own bug thought. But pretty bad.

PR #11345, opened 2025-08-14
test: add tests for `pquery` that use functional predicates

magnus-madsen:


magnus-madsen:
Depends on @CasperDN and #11304

magnus-madsen:
Already merged.

PR #11348, opened 2025-08-14
fix: allow unused variables in functional predicates

CasperDN:
When this has been merged I will make a PR for #11304.

### DIFF
--- a/main/test/flix/Test.Exp.Fixpoint.Functional.flix
+++ b/main/test/flix/Test.Exp.Fixpoint.Functional.flix
@@ -180,7 +180,7 @@ mod Test.Exp.Fixpoint.Functional {
     }
 
     @Test
-    def testFunctionalIgnoreRes01(): Bool = {
+    def testFunctionalUnusedVar01(): Bool = {
         let db = #{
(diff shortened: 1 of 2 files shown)


magnus-madsen:
What does this do? I mean what is ignored?

CasperDN:
It would previously crash if you did not use the bound variable. As in the following would die:
```
let pr = #{
    R(x) :- A(x), let y = Vector#{3}.
};
```
because `y` is not used.

main/test/flix/Test.Exp.Fixpoint.Functional.flix
```diff
@@ -179,6 +179,39 @@ mod Test.Exp.Fixpoint.Functional {
         Vector#{1f64} `Assert.eq` query p select (x) from NonEmpty(x)
     }
 
+    @Test
+    def testFunctionalIgnoreRes01(): Bool = {
```
magnus-madsen: Can we rename this to: 

testFunctionalUnusedVar

magnus-madsen: and so on

PR #11349, opened 2025-08-14
chore: cleanup tests related to extensible variants

magnus-madsen:


main/test/flix/Test.Exp.ExtMatch.flix
```diff
@@ -437,8 +437,9 @@ mod Test.Exp.ExtMatch {
     }
 
     ///////////////////////////////////////////////////////////////////////////
-    /// Other.
+    /// Unsorted.
```
mlutze, line 440: Should probably avoid this term and use "miscellaneous" instead.

mlutze (approved):

Issue #11343, opened 2025-08-13
Datalog: Allow longer provenance paths


magnus-madsen:
We currently error out at 2K:

```
    @Test
    def testPQueryLargePath01(): Bool =
        let pr = #{
            R(0).
            R(x + 1) :- R(x), if (x < 2048).
        };
        let result = pquery pr select R(2048) with {R};
        Assert.eq(2, Vector.length(result))
```

gives stackoverflow in `buildProof`

magnus-madsen:
I am OK with some bound, but 2048 seems too low.

magnus-madsen:
Not sure what is best. Options include:

- CPS
- Mutation
- Use effects and handlers

CasperDN:
Close this?

PR #11353, opened 2025-08-14
fix: support arbitrary size provenance paths

CasperDN:
Fix for #11343

Should we add some tests for this? They will be somewhat slow as they will need >2000 iterations of fixpoint computations to show it works? The simplest (performant) test I can imagine is:
```
@Test
def testPQueryLargePath01(): Bool =
    let db = inject Vector.range(0, 5000) |> Vector.map(x -> (x, x + 1)) into R/2;
    let pr = #{
        A(0).
        A(y) :- A(x), R(x, y).
    };
    let result = pquery db, pr select A(5000) with {R};
    Assert.eq(5000, Vector.length(result))
```

### DIFF
--- a/main/src/library/Fixpoint3/Provenance.flix
+++ b/main/src/library/Fixpoint3/Provenance.flix
@@ -117,15 +117,15 @@ mod Fixpoint3.Provenance {
     /// respect to `ProvIDB`, and it is not a ground atom. Furthermore `subproofs`
     /// contains the proofs/facts of the body which (could have) derived `P(tuple)`.
     ///
-    enum ProofTree {
+    enum ProofTree[r: Region] {
         case EDB(PredSym, Vector[Boxed])
         case Negative(PredSym, Vector[Boxed])
-        case IDB(PredSym, Vector[Boxed], Vector[ProofTree])
+        case IDB(PredSym, Vector[Boxed], Array[ProofTree[r], r])
     }
 
     mod ProofTree {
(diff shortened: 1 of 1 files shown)


CasperDN:
The 'fix' here uses a worklist and creates uninitialized arrays which are filled when they are handled in the worklist.

main/src/library/Fixpoint3/Provenance.flix
```diff
@@ -213,34 +209,52 @@ mod Fixpoint3.Provenance {
...
+                        Vector.filterMap(bodyAtom -> match bodyAtom {
+                            case BodyPredicate.BodyAtom(pred, _, _, _, _) => Some(pred)
+                            case _ => None
+                        });
+                    let ref = Ref.fresh(rc, tail);
+                    let childProofs = Array.empty(rc, Array.length(bodyAtoms));
+                    foreach((i, atom) <- ForEach.withIndex(relevantPreds)) {
+                        let pred = Array.get(i, bodyAtoms);
+                        Ref.put((i, childProofs, atom, pred) :: Ref.get(ref), ref)
+                    };
```
CasperDN, line 268: Having to make a `Ref` for this is kind of annoying.

magnus-madsen:
I was adding some tests-- so I will expand on those.

magnus-madsen:
I would have expected some documentation that explains why we have to build the tree in this cumbersome way.

Issue #11340, opened 2025-08-13
Datalog: Move desugaring of `FixpointQueryWithSelect` to `Lowering`


magnus-madsen:
The same as #11224, but now for the more complex:

```
case class FixpointQueryWithSelect(exps: List[Expr], selects: List[Expr], from: List[Predicate.Body], where: List[Expr], loc: SourceLocation) extends Expr
```

PR #11354, opened 2025-08-14
chore: defer desugaring of `FixpointQueryWithSelect` to Lowering

JumpTwiice:
Fixes https://github.com/flix/flix/issues/11340


magnus-madsen:
We talked about this being difficult. To you see a way to do it partially? E.g. delaying some of the desugaring but not all?

JumpTwiice:
I think we can get away with just creating the `$Result` relation with the corresponding rule before `TypeReconstruction` and then desugar the rest later

magnus-madsen:
> I think we can get away with just creating the `$Result` relation with the corresponding rule before `TypeReconstruction` and then desugar the rest later

OK. I would follow the "something is better than nothing".

JumpTwiice:
I'm closing this and opening a new PR.

PR #11355, opened 2025-08-14
chore: rework `Benchmarks.flix`

### LINKED ISSUE #11015: Rescale and cleanup benchmarks
- Drop outdated ones (e.g. Inliner10k)
- Rescale so the benchmarks are closer together:

Image

magnus-madsen:
Fixes #11015

The idea is simple: The benchmarks do the same thing on the same sized thing.


magnus-madsen:
The `Set` and `Map` collections are gone because they are too high-level and too slow.

mlutze (approved):

Issue #11264, opened 2025-08-08
Datalog: Add infrastructure for provenance suppression


magnus-madsen:
Imagine that we have the following program:

```scala
def main1(): Unit \ IO =
    let db = #{
        Edge(1, 2).
        Edge(2, 1).

        Edge(2, 3).
        Edge(3, 2).
        
        Edge(3, 4).
    };
    let pr = #{
        Path(x, y) :- Edge(x, y).
        Path(x, z) :- Path(x, y), Edge(y, z).
        Result(x, y) :- Path(x, y), Path(y, x).
    };
    query db, pr select (x, y) from Result(x, y) |> println
```

I would like to ask for provenance like so:

```
    let pp = pquery db, pr select Result(1, 3) with {Edge};
    Vector.forEach(v -> ematch v {
        case Edge(x, y) => println("${x} - ${y}")
    }, pp)
```

which gives:

```
1 - 2                                                                           
2 - 3
3 - 2
2 - 1
```

But I would like to suppress the second part. 

Thus I would like to spiritually write:

```
Result(x, y) :- Path(x, y), @Suppress Path(y, x).
```

This ticket is asking for infrastructure to support such provenance suppressions. How to support it at the language level will come later.

magnus-madsen:
@CasperDN @JumpTwiice Will this be difficult?

CasperDN:
Question: Does `@Suppress P(...)` mean do not show the proof for `P(...)` or does it mean show neither `P(...)` nor the proof for it. In other words is `P(...)` included in the provenance?

During solving: No. It simply means that we will need to expand the Datalog enum to receive the information and mostly ignore it.
During provenance reconstruction: No. Simply check whether the flag is set and if so report it as an EDB (or ignore?).

Parsing/everything else: No sure. Adam has the most experience there. I don't think that it would be horribly complicated.

magnus-madsen:
> Question: Does `@Suppress P(...)` mean do not show the proof for `P(...)` or does it mean show neither `P(...)` nor the proof for it. In other words is `P(...)` included in the provenance?

Not sure, what do you think is best?

> During solving: No. It simply means that we will need to expand the Datalog enum to receive the information and mostly ignore it. During provenance reconstruction: No. Simply check whether the flag is set and if so report it as an EDB (or ignore?).
> 
> Parsing/everything else: No sure. Adam has the most experience there. I don't think that it would be horribly complicated.

[...]

PR #11356, opened 2025-08-14
feat: add provenance suppression to Datalog engine

CasperDN:
Proposal for #11264

Will need merging with #11353.


CasperDN:
Other than testing it (which is hard without frontend support) this should work.

The only thing that will need to be changed is `mkSuppressed` in `Lowering.flix`.

magnus-madsen:
When everything else is done, rebase this onto master so there are no conflicts.

CasperDN:
So that was not the way to rebase. I again blame Intellij. I'll make a new branch instead.

Issue #11281, opened 2025-08-10
Hovering on wildcards (to see their type) no longer works


magnus-madsen:

mlutze:
FWIW, this is likely unrelated to my change, as that only related to wildcard types (which still seem to work)

magnus-madsen:
Ah okay. But it used to work?

mlutze:
I think it's been a problem for a while

mlutze:
I think it's something to do with that `let _ = 123` gets desugared into a match rather than a let-binding.

PR #11360, opened 2025-08-15
fix: enable hover information on wildcards

mlutze:
fixes https://github.com/flix/flix/issues/11281


magnus-madsen:
Does it work?

mlutze:
Yes, I tested. I'm not sure we should really have wildcard patterns as separate from variables, but that's a different issue.

magnus-madsen (approved):

PR #11361, opened 2025-08-15
chore: simplify `Fixpoint/Phase/IndexSelection.flix`

CasperDN:
Mostly removal of mutation and prevention of a relatively expensive algorithm in the common (simple) case.


magnus-madsen (approved):

magnus-madsen:
Excellent- We should always favor functional programming over imperative programming-- unless its performance critical.

CasperDN:
I actually did this because I saw `Set.remove` in the JProfiler...

magnus-madsen:
As a rule of thumb I would say:

Any code that touches rules can be slow (because there are few rules).
Any code that touches facts has to be fast (because there can be many facts).

Not sure how that applies here.

PR #11366, opened 2025-08-15
fix: add `ExtMatch` to `StackConsumer`

jaschdoc:


magnus-madsen:
What about the visitor? Do they not need to be in sync?

jaschdoc:
Yes, they are but I forgot to add this part. The functions here are also defined on `Consumer` with a default implementation hence no compiler error

PR #11374, opened 2025-08-18
fix: resilient parsing of empty tuple types

magnus-madsen:


mlutze:
Test?

magnus-madsen:
> Test?

Cannot reproduce

magnus-madsen:
But TLDR: Recall that we are going to replace all ICEs with soft errors.

PR #11381, opened 2025-08-19
refactor: move typing from `Fixpoint/Phase/Lowering` to `Fixpoint/Phase/`Compiler`

CasperDN:
Hopefully removes some overhead relating to compiling Datalog programs. Simplifies `Lowering`, `Boxing` and `Ram` somewhat.

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/Flix.scala
+++ b/main/src/ca/uwaterloo/flix/api/Flix.scala
@@ -288,6 +288,7 @@ class Flix {
     "Fixpoint3/ReadWriteLock.flix" -> LocalResource.get("/src/library/Fixpoint3/ReadWriteLock.flix"),
     "Fixpoint3/Solver.flix" -> LocalResource.get("/src/library/Fixpoint3/Solver.flix"),
     "Fixpoint3/SubstitutePredSym.flix" -> LocalResource.get("/src/library/Fixpoint3/SubstitutePredSym.flix"),
+    "Fixpoint3/TypeInfo.flix" -> LocalResource.get("/src/library/Fixpoint3/TypeInfo.flix"),
     "Fixpoint3/UniqueInts.flix" -> LocalResource.get("/src/library/Fixpoint3/UniqueInts.flix"),
     "Fixpoint3/Util.flix" -> LocalResource.get("/src/library/Fixpoint3/Util.flix"),
 
--- a/main/src/library/Fixpoint3/Ast/ExecutableRam.flix
+++ b/main/src/library/Fixpoint3/Ast/ExecutableRam.flix
@@ -22,8 +22,9 @@
 mod Fixpoint3.Ast.ExecutableRam {
     use Fixpoint3.Ast.Ram.{Predicates, RelSym, IndexInformation}
     use Fixpoint3.Ast.Shared.{BoxedDenotation => Denotation}
-    use Fixpoint3.Boxed;
-    use Fixpoint3.BoxingType.{UnifiedTypePos, RamIdToPos};
+    use Fixpoint3.Boxed
+    use Fixpoint3.BoxingType.UnifiedTypePos
+    use Fixpoint3.TypeInfo.TypeInformation
 
     /////////////////////////////////////////////////////////////////////////////
     // RamProgram                                                              //
@@ -34,7 +35,7 @@ mod Fixpoint3.Ast.ExecutableRam {
     /// by the interpreter.
     ///
     pub enum RamProgram[r: Region] {
-        case Program(RamStmt, Facts[r], Predicates, IndexInformation, (Arities, ConstWrites), RamIdToPos)
+        case Program(RamStmt, Facts[r], Predicates, IndexInformation, (Arities, ConstWrites), TypeInformation)
     }
     ///
     /// The extensible database (EDB) of a program.
--- a/main/src/library/Fixpoint3/Ast/Ram.flix
+++ b/main/src/library/Fixpoint3/Ast/Ram.flix
@@ -22,6 +22,7 @@ mod Fixpoint3.Ast.Ram {
     use Fixpoint3.Boxed
     use Fixpoint3.PredSymsOf
     use Fixpoint3.SubstitutePredSym
+    use Fixpoint3.TypeInfo.{provType, RelType, Type, TypeInformation}
 
     /////////////////////////////////////////////////////////////////////////////
     // RamProgram                                                              //
@@ -32,7 +33,7 @@ mod Fixpoint3.Ast.Ram {
     ///
     @Internal
     pub enum RamProgram {
-        case Program(RamStmt, FactTuple, Predicates, IndexInformation)
+        case Program(RamStmt, FactTuple, Predicates, IndexInformation, TypeInformation)
     }
 
     ///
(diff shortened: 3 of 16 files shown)


magnus-madsen:
What's the short story? What is the impact on time on `gradlew test` on the server for master versus this PR?

CasperDN:
Story: For very small programs `Lowering`/`Boxing` dominates the time usage. This eliminates that (more or less). The small programs do not take a lot of type so it will likely not affect test time significantly.

It should be noted that this is primarily based on what I can see with `JProfiler`. The problem here is that with parallelism its precision degrades significantly and manually removing parallelism from the interpreter means that its cost of spawning threads is also discarded, so precision is still bad. Tomorrow I will be testing whether it helps to replace
```
BPlusTree.parForEach(parEvalSearchOrQueryOp(rc, tuplePos, writes, body, ctxt, parLevel - 1), tree)
```
with
```
BPlusTree.parForEachWhen(
    parEvalSearchOrQueryOp(rc, tuplePos, writes, body, ctxt, parLevel - 1),
    evalSearchOrQueryOp(rc, tuplePos, writes, body, ctxt, parLevel),
    n, tree)
```
for some smallish `n` (e.g. 10).

Design wise we are now dealing with a typed AST/Ram, at least as typed as the Datalog engine cares for. If we ever provide the types to the engine only `TypeInfo` will have to be modified. Furthermore the `RowVar`'s can now be used as position in memory meaning that `Lowering` can get rid of `IdToIndex` which will simplify it somewhat.

I believe that limiting thread spawning with `parForEachWhen` will have a much higher impact than this.

Conclusion: Started with being a performance measure, which somewhat failed, but I believe it will make the engine 'nicer'.

magnus-madsen:
What's the ultimate impact on `gradlew test`.

Note: For short running programs profiling is often very tricky... You get a lot of misleading noise.

magnus-madsen:
Did you attach JProfiler to gradlew test?

CasperDN:
> Note: For short running programs profiling is often very tricky... You get a lot of misleading noise.

Sure. I usually use the following if I want to profile a singular test:
```
def runForever(): Unit \ IO = {
    if (someTest()) runForever() else runForever()
```

CasperDN:
> Did you attach JProfiler to gradlew test?

Around 50-60% of the time is used in `TypeVerifier`, more for Fixpoint programs. I do not know how `TypeVerifier` works, so I have no idea how one would improve that...

magnus-madsen:
> > Did you attach JProfiler to gradlew test?
> 
> Around 50-60% of the time is used in `TypeVerifier`, more for Fixpoint programs. I do not know how `TypeVerifier` works, so I have no idea how one would improve that...

Hmm yeah. Any where specific in it?

Thats probably because Datalog programs just include a lot of code and running the  type verifier on that takes longer.

Is the conclusion now that there is no specific problem in the Datalog engine and its just some type verifier stuff?

CasperDN:
> Hmm yeah. Any where specific in it?

Just `visitExpr`, which is the main method of the file.

> Thats probably because Datalog programs just include a lot of code and running the type verifier on that takes longer.

Agreed. You are basically including a couple of thousand lines whenever you use the engine.

Most of the time is, apparently, spent compiling/typing/generating code and so on.

> Is the conclusion now that there is no specific problem in the Datalog engine and its just some type verifier stuff?

As mentioned the profiler is kind of bad when profiling parallel stuff, which the interpreter is. I'll experiment a bit with turning it of and report back.

I can say that an easy way to gain some performance when running tests will be to turn down the test sizes in `BPlusTree`. This will also make it harder for the tests to catch actual errors, though.

magnus-madsen:
On the server: 

- What is the test time with and without Type Verifier?
- What is the test time with and without Test*Fixpoint*?

CasperDN:
| Datalog tests\\`TypeVerifier` | Yes | No |
| -- | -------- | ------- |
| Yes | 4m 41s  | 3m 10s |
| No |  2m 27s  | 2m 7s |

Note that I understood the removal of _Fixpoint_ as the removal of `Test.Exp.Fixpoint.*`, `Test.Predicate.*`, `Test.Stratification.flix`, `Test.Term.*` and `examples/larger-examples/datalog/*`.

The test suite I ran was whatever is included in `./gradlew testAll`. I used the latest master as blueprint.

magnus-madsen:
@JonathanStarup What do you make of the above table?

JonathanStarup:
uhm i think im missing context, running less stuff is faster?

magnus-madsen:
@JonathanStarup The question is more: Why is the type verifier so expensive (as it relates to Datalog)? Does it have some bad n^2 behavior somewhere you can recall? Does it deal with rows or not?

JonathanStarup:
Its a pretty direct traversal of the AST and its types (including rows, but no quadratic stuff)

It checks java subtyping, perhaps that is slow?

magnus-madsen:
> Its a pretty direct traversal of the AST and its types (including rows, but no quadratic stuff)
> 
> It checks java subtyping, perhaps that is slow?

Unlikely. I wonder if there is some n^2 behavior.

JonathanStarup:
My guess is just that traversing types takes a long type

magnus-madsen:
I think we abandon this? It turns out there was a bug in the test framework-- so the numbers are probably not even right.

CasperDN:
I believe the design is better than the current (computing 'type' on the `Datalog` instead of `Ram`), but we can close it.

magnus-madsen:
> I believe the design is better than the current (computing 'type' on the `Datalog` instead of `Ram`), but we can close it.

Better in what sense? it is a cleaner design? I do notice it is shorter.

CasperDN:
The Datalog engine currently tracks something called `RamID` for the sole purpose of 'typing' the program. It then at some point has to 'type' the program based on the RAM during lowering.

Instead this change would decide 'types' based on the much simpler `Datalog` enum in the beginning and instead pass
the types during the all phases (previously `RamID` was passed). This means that we can get rid of a complex 'type'-inference in lowering/boxing (so far it has been prone to errors) and the `RamID` which is only used to 'type' the program.

Note that when I say 'types' the Datalog engine represent a type by the position it can store the boxed values, as that it is currently all it is interested in.

magnus-madsen:
> The Datalog engine currently tracks something called `RamID` for the sole purpose of 'typing' the program. It then at some point has to 'type' the program based on the RAM during lowering.
> 
> Instead this change would decide 'types' based on the much simpler `Datalog` enum in the beginning and instead pass the types during the all phases (previously `RamID` was passed). This means that we can get rid of a complex 'type'-inference in lowering/boxing (so far it has been prone to errors) and the `RamID` which is only used to 'type' the program.
> 
> Note that when I say 'types' the Datalog engine represent a type by the position it can store the boxed values, as that it is currently all it is interested in.

OK. That sounds like a better solution. Let us implement that.

CasperDN:
@magnus-madsen The Git history got a bit messed up, but this should be fine. The only 'problem' I see is that some unrelating changes to formatting (indent) in `Ram`. I can remove them, but they should be included in some later PR then.

If you want to make the PR smaller I guess we could make `TypeInfo` a PR by itself. Then a PR where it is used and passed the type through the phases as another and finally a PR where the type is used instead of the current Lowering/Boxing approach.

They will still be quite big PRs.

magnus-madsen:
The PR is too large for my test. Can you break it down without too much effort? Thanks!

PR #11383, opened 2025-08-19
refactor: parse extpat as pattern

jaschdoc:
Supersedes https://github.com/flix/flix/pull/11316

Closes https://github.com/flix/flix/issues/11292

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Redundancy.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Redundancy.scala
@@ -610,7 +610,7 @@ object Redundancy {
             RedundancyError.DuplicateExtPattern(rule1.label, rule1.label.loc, rule2.label.loc)
         }
 
-      Used(Set.empty, duplicatePatterns.toSet) ++ usedMatch ++ usedRules.reduceLeft(_ ++ _)
+      Used(Set.empty, duplicatePatterns.toSet) ++ usedMatch ++ usedRules.foldLeft(Used.empty)(_ ++ _)
 
     case Expr.Tag(CaseSymUse(sym, _), exps, _, _, _) =>
       val us = visitExps(exps, env0, rc)


jaschdoc:
This also further complicates default patterns later since you have to split the list of patterns at the default pattern before knowing / weeding which pattern is which.

jaschdoc:
> This also further complicates default patterns later since you have to split the list of patterns at the default pattern before knowing / weeding which pattern is which.

If we go with https://github.com/flix/flix/pull/11316 then it's a simpler check that can be implemented in `Redundancy` (and we can implement it for ADTs too!)

jaschdoc:
Reposting this here: https://github.com/flix/flix/pull/11316#issuecomment-3200719040

If we do it your way, then typing is a bit more awkward since the list of rules may be empty in which case the type (after reconstruction) must be a variable which shouldn't make sense

magnus-madsen:
@jaschdoc What if KindedAst.Expr.ExtMatch simply had a tvar: Type.Var like some other expression have?

magnus-madsen:
> This also further complicates default patterns later since you have to split the list of patterns at the default pattern before knowing / weeding which pattern is which.

I don't understand.

jaschdoc:
> @jaschdoc What if KindedAst.Expr.ExtMatch simply had a tvar: Type.Var like some other expression have?

That does not solve the problem, since there is nothing to unify the variable with.

jaschdoc:
> > This also further complicates default patterns later since you have to split the list of patterns at the default pattern before knowing / weeding which pattern is which.
> 
> I don't understand.

Assuming you want to disallowing more than one default pattern or disregard anything that comes after, then weeding default patterns is harder this way since you first have to assemble the `WeededAst` version of the `ematch`.
It seems easier to emit an error in `Redunancy` instead and let weeding just assemble the ast. Besides, there is nothing *syntactically* wrong with
```
ematch exp {
case _ => exp
case A => exp
case _ => exp
}
```

magnus-madsen:
> > @jaschdoc What if KindedAst.Expr.ExtMatch simply had a tvar: Type.Var like some other expression have?
> 
> That does not solve the problem, since there is nothing to unify the variable with.

Unify it with the bodies of the rules and the optional expression?

magnus-madsen:
OK; so if you really want to preserve the multiple default cases can you sketch a design where you use ADTs. (But no Scala inheritance that does not correspond to a regular ADT). Does it make sense what I am saying?

jaschdoc:
We agreed on https://github.com/flix/flix/pull/11316#issuecomment-3205604425

Issue #11372, opened 2025-08-17
Compiler crash Inline of pre inlined variable with subeffecting


NeonOxide:
I was playing around with trying to trick the compiler into believing a function was pure when it was not by using continuations but in the midst of it I found a valid program that crashes at execution time. I minimized the example as much as I could. I found that removing the IO effect annotation from testing_parent_handler fixes the issue but aside from that, I do not know why this may happen. It is also important to note that I run the compiler with the following flags: --Xsubeffecting mod-defs,ins-defs,lambdas. I was unable to make the example work without the flags by substituting with checked_ecast while keeping IO in the function signature. This does not seem like an urgent issue to address but I figured that it is better to leave it at least documented. This is the code block that raises the exception: 

```
eff MyPureFnGenerator{
    def leakage(): (Unit -> Unit) 
}
def testing_parent_handler() : Int32  \ IO = 
    run {
        0xDEAD
    } with handler MyPureFnGenerator {
        def leakage(resume) = {
            resume(_ -> {
                let _ = resume(identity);
                ()
                })
        }
    } 

def main(): Unit \ IO  = 
    let f : Int32   = testing_parent_handler();
    println("Hello World " ++ ToString.toString(f))
```
This is the crash report:

 An unexpected error has been detected by the Flix compiler:

   unexpected call site inline of pre-inlined variable f$146891 (/home/neo/code_projects/flix/example/Main.flix:7:12)

 This is a bug in the Flix compiler. Please report it here:

 https://github.com/flix/flix/issues

# -- Flix Compiler --
#
# Flix Version : 0.60.0
#   incremental: All
#
# -- Java Virtual Machine --
#
# JVM Version  : 21.0.7 (2025-04-15)
# JVM Vendor   : N/A
# JAVA_HOME    : /nix/store/jjknlm78gfms2gc6fnjzpb6963yk6v81-openjdk-21.0.7+6/lib/openjdk
# System       : Linux (6.12.10-76061203-generic)
#
# -- Stack Trace --
ca.uwaterloo.flix.util.InternalCompilerException: unexpected call site inline of pre-inlined variable f$146891 (/home/neo/code_projects/flix/example/Main.flix:7:12)
	at ca.uwaterloo.flix.language.phase.optimizer.Inliner$.shouldInlineVar(Inliner.scala:558)
	at ca.uwaterloo.flix.language.phase.optimizer.Inliner$.useSiteInline(Inliner.scala:540)
[...]

PR #11384, opened 2025-08-19
fix: handle inliner crash on subeffecting

jaschdoc:
Closes https://github.com/flix/flix/issues/11372


magnus-madsen:
Tricky... But makes good sense.

Issue #11342, opened 2025-08-13
Datalog: Move desugaring of `FixpointSolveWithProject` to `Lowering`


magnus-madsen:
Similar to #11340

PR #11387, opened 2025-08-19
chore: defer desugaring of `FixpointSolveWithProject` to Lowering

JumpTwiice:
Fixes https://github.com/flix/flix/issues/11342.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
@@ -833,6 +833,12 @@ object Lowering {
       LoweredAst.Expr.ApplyDef(defn.sym, argExps, List.empty, itpe, tpe, eff, loc)
 
     case TypedAst.Expr.FixpointSolveWithProject(exps0, optPreds, mode, _, eff, loc) =>
+      // Rewrites
+      //     solve e₁, e₂, e₃ project P₁, P₂, P₃
+      // to
+      //     let tmp% = solve e₁ <+> e₂ <+> e₃;
+      //     merge (project P₁ tmp%, project P₂ tmp%, project P₃ tmp%)
+      //
       val defn = mode match {
         case SolveMode.Default => Defs.lookup(Defs.Solve)
         case SolveMode.WithProvenance => Defs.lookup(Defs.SolveWithProvenance)


main/src/ca/uwaterloo/flix/language/ast/TypedAst.scala
```diff
@@ -274,7 +274,7 @@ object TypedAst {
 
     case class FixpointQueryWithProvenance(exps: List[Expr], select: Predicate.Head, withh: List[Name.Pred], tpe: Type, eff: Type, loc: SourceLocation) extends Expr
 
-    case class FixpointSolve(exp: Expr, tpe: Type, eff: Type, mode: SolveMode, loc: SourceLocation) extends Expr
+    case class FixpointSolveWithProject(exps: List[Expr], mode: SolveMode, optPreds: Option[List[Name.Pred]], tpe: Type, eff: Type, loc: SourceLocation) extends Expr
```
magnus-madsen: `mode: SolveMode` is some data we attach, but not something the user writes. So to keep the user stuff first, I would change the order of arguments to:

```
case class FixpointSolveWithProject(exps: List[Expr], optPreds: Option[List[Name.Pred]], mode: SolveMode, tpe: Type, eff: Type, loc: SourceLocation)
```

main/src/ca/uwaterloo/flix/language/ast/KindedAst.scala
```diff
@@ -219,7 +219,7 @@ object KindedAst {
 
     case class FixpointQueryWithProvenance(exps: List[Expr], select: Predicate.Head, withh: List[Name.Pred], tvar: Type, loc: SourceLocation) extends Expr
 
-    case class FixpointSolve(exp: Expr, mode: SolveMode, loc: SourceLocation) extends Expr
+    case class FixpointSolveWithProject(exps: List[Expr], mode: SolveMode, optPreds: Option[List[Name.Pred]], tvar: Type.Var, loc: SourceLocation) extends Expr
```
magnus-madsen: `mode: SolveMode` is some data we attach, but not something the user writes. So to keep the user stuff first, I would change the order of arguments to:

```
    case class FixpointSolveWithProject(exps: List[Expr], optPreds: Option[List[Name.Pred]], mode: SolveMode, tvar: Type.Var, loc: SourceLocation) extends Expr
```

main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
```diff
@@ -1454,7 +1455,7 @@ object Desugar {
     val mergeExp = es.reduceRight[DesugaredAst.Expr] {
       case (e, acc) => DesugaredAst.Expr.FixpointMerge(e, acc, loc0)
     }
-    val modelExp = DesugaredAst.Expr.FixpointSolve(mergeExp, mode, loc0)
+    val modelExp = ??? // DesugaredAst.Expr.FixpointSolve(mergeExp, mode, loc0)
```
magnus-madsen: What happened here?

magnus-madsen: I suppose this whole function `desugarFixpointSolveWithProject` can be removed?

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -2127,8 +2127,8 @@ object Weeder2 {
       val idents = pickAll(TreeKind.Ident, tree).map(tokenToIdent)
       mapN(traverse(expressions)(visitExpr)) {
         exprs =>
-          val optIdents = if (idents.isEmpty) None else Some(idents)
-          Expr.FixpointSolveWithProject(exprs, solveMode, optIdents, tree.loc)
+          val optPreds = if (idents.isEmpty) None else Some(idents.map(Name.mkPred))
```
magnus-madsen, line 2130: Good, we might as well go for Name.mkPred as early as possible 👍

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -1731,6 +1736,14 @@ object Lowering {
         LoweredAst.Expr.ApplyDef(defn.sym, argExps, List.empty, itpe, resultType, exp.eff, loc)
     }
 
+  private def filterExp(predSym: LoweredAst.Expr, datalog: LoweredAst.Expr, loc: SourceLocation)(implicit root: TypedAst.Root): LoweredAst.Expr = {
```
magnus-madsen: Documentation? Example of what it constructs.

magnus-madsen: Also filter is not the best name

magnus-madsen: And the argument names are also not great.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -832,14 +832,19 @@ object Lowering {
...
       }
-      val argExps = visitExp(exp) :: Nil
-      val resultType = Types.Datalog
-      LoweredAst.Expr.ApplyDef(defn.sym, argExps, List.empty, Types.SolveType, resultType, eff, loc)
+      val exps = exps0.map(visitExp)
+      val mergedExp = mergeExps(exps, loc)
+      val argExps = mergedExp :: Nil
+      val solvedExp = LoweredAst.Expr.ApplyDef(solveDefn.sym, argExps, List.empty, Types.SolveType, Types.Datalog, eff, loc)
+      optPreds match {
+        case Some(preds) => mergeExps(preds.map(pred => filterExp(mkPredSym(pred), solvedExp, loc)), loc)
```
magnus-madsen: Are we sure this is correct? I would worry that we also lack test cases.

What I mean is, if you filter A and then B and then C, won't you end up with nothing?

magnus-madsen: Maybe go back and compare to what happened in desugar.

There we introduce a local variable tmp to hold the result, filter and merge.

main/src/ca/uwaterloo/flix/language/phase/typer/SchemaConstraintGen.scala
```diff
@@ -102,20 +102,34 @@ object SchemaConstraintGen {
...
+        //  solve exp project P1, P2, ... : #{ P1(...), P2(...), ... }
         //
-        val (tpe, eff) = visitExp(exp)
-        c.unifyType(tpe, Type.mkSchema(mkAnySchemaRowType(loc), loc), loc)
-        val resEff = eff
-        val resTpe = tpe
+        val (tpes, effs) = exps.map(visitExp).unzip
+        val freshSchemaRow = Type.freshVar(Kind.SchemaRow, loc)
+        c.unifyAllTypes(Type.mkSchema(freshSchemaRow, loc) :: tpes, loc)
+        val resultSchemaRow = optPreds match {
```
magnus-madsen, line 113: Let us add a helper function that encapsulates this computation, optPreds -> row

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
JumpTwiice: We apply `Solver.projectSym` on the Datalog expression `solvedExp` separately for each relation. Finally, you merge the separate Datalog expressions (one for each relation).

main/src/ca/uwaterloo/flix/language/phase/Desugar.scala
JumpTwiice: Yes, it is not needed anymore

main/src/ca/uwaterloo/flix/language/phase/typer/SchemaConstraintGen.scala
JumpTwiice: Wouldn't it be better if it was `List[Name.Pred] -> Row` ? That way `visitFixpointQueryWithProvenance` could use the same function.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
JumpTwiice: About the tests, `Test.Exp.Fixpoint.Solve.Project.flix` is pretty comprehensive and does catch exactly what you're worried about.

main/src/ca/uwaterloo/flix/language/phase/typer/SchemaConstraintGen.scala
magnus-madsen: Yes, let us make it reusable. Will you do that and ping me?

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
magnus-madsen: OK; great

magnus-madsen: But how many times do you call solve? It seems you create the solve expression and use it inside a .map.

Will that not cause the fixpoint computation to run more than once?

main/src/ca/uwaterloo/flix/language/phase/typer/SchemaConstraintGen.scala
JumpTwiice: @magnus-madsen Already done.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
JumpTwiice: We call solve once, giving us a minimal model `solvedExp`. We then filter the minimal model for each predicate.

JumpTwiice: Actually, you're right. We need a tmp variable.

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -832,14 +832,26 @@ object Lowering {
...
-    case TypedAst.Expr.FixpointSolve(exp, _, eff, mode, loc) =>
+    case TypedAst.Expr.FixpointSolveWithProject(exps0, optPreds, mode, _, eff, loc) =>
       val defn = mode match {
         case SolveMode.Default => Defs.lookup(Defs.Solve)
         case SolveMode.WithProvenance => Defs.lookup(Defs.SolveWithProvenance)
       }
-      val argExps = visitExp(exp) :: Nil
-      val resultType = Types.Datalog
-      LoweredAst.Expr.ApplyDef(defn.sym, argExps, List.empty, Types.SolveType, resultType, eff, loc)
+      val exps = exps0.map(visitExp)
```
magnus-madsen, line 846: I would like a comment - like on other places - that shows pseudo code for the generated code.

PR #11389, opened 2025-08-20
ci: increase compiler fuzzing time

jaschdoc:
We have significantly increased the time taken by the compiler test suite, so this adds a bit more compiler fuzzing "for free".

### DIFF
--- a/main/test/flix/fuzzers/FuzzDeleteLines.scala
+++ b/main/test/flix/fuzzers/FuzzDeleteLines.scala
@@ -28,7 +28,7 @@ class FuzzDeleteLines extends AnyFunSuite with TestUtils {
   /**
     * Number of variants to make for each file. Each variant has a single line deleted.
     */
-  private val N = 30
+  private val N = 60
 
   test("simple-card-game") {
     val filepath = Paths.get("examples/larger-examples/simple-card-game.flix")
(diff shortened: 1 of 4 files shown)


jaschdoc:
Now runs fuzzing for ~12 minutes and tests for ~14 minutes

magnus-madsen:
We are trying to bring down the test time though-- so. I guess I will leave the PR until either that happens or I conclude it cannot be done (and then we can merge this PR).

mlutze:
I disagree with this on principle. If the goal is that this should run as long as the other test suite runs, then we should somehow actually connect them.

As it is, this just seems inflationary (X takes more time so Y can take more time so X can take more time...)

magnus-madsen:
Fuzzers are usually run on separate gear 24/7. Typically not connected to PRs.

magnus-madsen:
We are not using the server right now, so if someone wants to write some infrastructure to run it for a month-- we could.

magnus-madsen:
On the other hand, perhaps we have enough bug reports already?

mlutze:
GitHub actions allow cron triggers

jaschdoc:
> I disagree with this on principle. If the goal is that this should run as long as the other test suite runs, then we should somehow actually connect them.

I agree that we could fuzz in a cron job too or set it up on the server for the latest release.

> As it is, this just seems inflationary (X takes more time so Y can take more time so X can take more time...)

I disagree with this, though. As it is now fuzzing should *always* take less time than compiler tests. So if we decrease compiler test time, the fuzzing suite time should be decreased too.

magnus-madsen:
Lets postpones this. We have plenty of other high priority tasks.

PR #11391, opened 2025-08-20
fix: use iterated locs in `TypeContext.expectTypeArguments`

### LINKED ISSUE #11310: ExtVar: Source locations wrong
```scala
def main(): Unit \ IO = {
    let p = #{
        A(1).
    };
    let v = pquery p select A(1) with {A};
    let res = Vector.get(0, v);
    ematch res {
        case A() => println("a")
    }
}
```
errors with

```
-- Type Error -------------------------------------------------- 

>> Expected argument of type 'Vector[#| A(Unit) |#]', but got 'Vector[#| A(Int32) |#]'.

87 |     let res = Vector.get(0, v);
                              ^
                              expected: 'Vector[#| A(Unit) |#]'

The function 'Vector.get' expects its 2nd argument to be of type 'Vector[#| A(Unit) |#]'.

Expected: Vector[#| A(Unit) |#]
  Actual: Vector[#| A(Int32) |#]
```
It is correct behavior to produce an error, but the location is not where `0` is. The location should be on `v` instead, i.e., the last argument to `Vector.get`.

jaschdoc:
In `master` / the buggy version, it would always use the head of all locs. This seems like just a typo.

Closes #11310

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/typer/TypeContext.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/typer/TypeContext.scala
@@ -193,7 +193,7 @@ class TypeContext {
     while (es.nonEmpty && as.nonEmpty) {
       val expectedType = es.head
       val actualType = as.head
-      val loc = actualLocs.head
+      val loc = ls.head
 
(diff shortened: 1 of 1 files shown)


magnus-madsen:
Great catch!

Issue #11362, opened 2025-08-15
Datalog: Interaction between provenance and stratification


CasperDN:
What does the following program print?
```
def main(): Unit \ IO = {
    let db = #{
        A(1). A(2). A(3). A(4). A(5).
    };
    let pr = #{
        B(1) :- not A(-1).
        B(x + 1) :- B(x), A(x).
        C(-1) :- not B(-2).
        C(x) :- B(x).
        C(x) :- A(x), C(-1).
    };
    let pp = pquery db, pr select C(5) with {A, B, C};
    Vector.map(v -> ematch v {
        case A(x) => "A(${x})"
        case B(x) => "B(${x})"
        case C(x) => "C(${x})"
    }, pp) |> println
}
```
<details><summary>Result</summary> Vector#{C(5), B(5), B(4), B(3), B(2), B(1), A(1), A(2), A(3), A(4)} </details>

CasperDN:
Explanation of the above: Stratification: `C -> B -> A`. The engine will compute all `A` facts first. It will then compute the `B` facts. Finally it will compute the `C` facts.
In the first iteration of the `C`-loop it will evaluate all `C(x) :- B(x)` and `C(-1) :- not B(-2)`, while `C(x) :- A(x), C(-1).` cannot activate as `C(-1)` has not been computed.
This means that the provenance of the `C`-facts will use the unnecessarily long path through `B`, as it takes 1 iteration of `C`-computation, whereas the shortest path takes 2 rounds of `C`-computation.

CasperDN:
Summarizing the above: The current Datalog engine has the following properties:
1. For any given strata the engine will compute a shortest proof
2. For any given strata we treat IDB and EDB facts from previous strata as EDB facts

Together this means that we treat a long derivation in a previous strata as a proof of height 1. In other words we do not (in general) compute the shortest derivation when there are multiple strata.

magnus-madsen:
Hypothetically, if we wanted to guarantee the property that the proof tree has minimal height, what would be the options? Do we have to give up performance?

CasperDN:
We have to give up some performance. As the example shows we would potentially have to do more computation in a single strata to ensure that we get the shortest path.

The real problem is that the interpreter at the moment is almost completely unaware of the provenance elements and so are the `BPlusTree`s. We would need some way to ensure that recomputed old facts with lower height would be re-handled. I can see some ways to do this, but most sacrifice performance further.

magnus-madsen:
[...]

PR #11395, opened 2025-08-21
feat: provenance should always compute the shortest proof

CasperDN:
See #11362.

Proof of concept that provenance can compute the shortest path with (relatively) few changes to the engine.

### DIFF
--- a/main/test/flix/Test.Exp.Fixpoint.PQuery.flix
+++ b/main/test/flix/Test.Exp.Fixpoint.PQuery.flix
@@ -969,4 +969,48 @@ mod Test.Exp.Fixpoint.PQuery {
         }, pp) `Assert.eq` Vector#{"A(3)"}
     }
 
+    /////////////////////////////////////////////////////////////////////////////
+    /// Provenance should compute the shortest path.
+    /////////////////////////////////////////////////////////////////////////////
+
+    @Test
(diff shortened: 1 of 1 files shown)


CasperDN:
If we decide to go forward with this I will split this up in smaller pieces (starting with `BPlusTree`)

magnus-madsen:
I would say it is 80% like we need this. But perhaps first I need more details. So maybe we discuss it on Monday and go from there?

CasperDN:
Sure

magnus-madsen:
@CasperDN We agreed that our goal is to merge this after one more review by @CasperDN 

Maybe split into smaller PRs.

CasperDN:
Summarizing the performance considerations:
Propagating facts multiple times, as we have talked about.
When computing a new fact we need to defer the check of whether fact `x` has been computed until all variables have been bound. Currently this proposal pushes it to right before insertion. Some programs might be more expensive. Artificial example:
```
A(x) :- B(x), C(y), D(z), E(a).
```
Currently after binding `x` in `B(x)` we will test whether `A(x)` exists and otherwise terminate. The new solution has to do the check after binding `y`, `z` and `a`, as it must compare depths, which also depends on `y`, `z` and `a`. This means that we might miss some early-termination if `x` is already in `A`.

Issue #10518, opened 2025-05-08
Analysis of CPS-usage in the Standard Library


magnus-madsen:
I would like a table with the following information:

Each row is the name of a function that uses CPS internally (e.g. `List.findRight`, `List.foldRight`, etc.). In addition:

- If that function can trivially be rewritten to use e.g. `List.reverse` instead. (Some can, but some cannot.)
- If that function uses an `unchecked_cast`.

jaschdoc:
Here is `List.flix`:

| Function | Can be rewritten |	Unchecked cast |
| -------- | ---------------- | -------------- |
| List.findRight |	Yes |	No |
| List.scanLeft |	Yes |	No |
| List.scanRight |	Yes |	No |
| List.mapWithIndex |	Yes |	No |
| List.update |	Maybe |	No |
| List.patch |	Yes |	No |
| List.removeIndex (not cps) |	Maybe |	No |
| List.applyHelper |	Yes |	No |
| List.intersperse |	Yes |	No |
| List.intercalate |	Yes |	No |
| List.foldRight |	Yes |	No |
| List.reduceRight |	Yes |	No |
| List.flatten |	Yes |	No |
| List.filter |	Yes |	No |
| List.init |	Yes |	No |
| List.slice |	Yes |	No |
| List.partition |	Yes |	No |
| List.span |	Yes |	No |
| List.take |	Yes |	No |
| List.groupBy |	Yes |	No |
| List.extractHelper |	Yes |	No |
| List.zip |	Yes |	No |
| List.zipWith |	Yes |	No |
| List.zipWithA |	No |	No |
| List.unzip |	Yes |	No |
| List.zipWith3 |	Yes |	No |
| List.unzip3 |	Yes |	No |
| List.foldRigh2 |	Yes |	No |
| List.filterMap |	Yes |	No |
| List.unfold |	Yes |	No |
| List.unfoldWithIter |	Yes |	No |
| List.unfoldWithOkIter |	Yes |	No |
| List.distinct |	Yes |	No |
| List.distinctWith |	Yes |	No |
| List.sequence |	Yes |	No |
| List.traverse |	Yes |	No |
| List.merge |	Yes |	No |

jaschdoc:
| Function | Can Be Rewritten | Unchecked Cast |
| --------- | ----------------- | ----------------- |
| Chain.viewLeft | Yes | No |
| Chain.viewRight | Yes | No |
| Chain.zipWithA |Yes | No |
[...]

PR #11399, opened 2025-08-21
chore: remove CPS from `List`

CasperDN:
Changes most CPS uses to use an accumulator which is often reversed afterwards. Added `reverseAppend` for that purpose. The changes are strictly related to rewriting from CPS to accumulation, smart rewrites have not been done. Related to #10518.

Not updated: `List.sequence` and.`List.traverse`.

Performance has not been a strict focus, but it appears to be better for the tested `update`, though the tests are too small to see this improvement.

### DIFF
--- a/main/src/library/List.flix
+++ b/main/src/library/List.flix
@@ -956,7 +956,7 @@ mod List {
     def groupByHelper(value: a, f: (a, a) -> Bool, currentNels: List[Nel[a]]): List[Nel[a]] =
         // Record seen elements in reverse order in `acc`.
         def loop(cur, acc) = match cur {
-            case Nil => reverse((Nel.singleton(value) :: acc))
+            case Nil => reverse(Nel.singleton(value) :: acc)
             case x :: xs =>
                 if (Nel.forAll(v -> f(value, v) and f(v, value), x)) {
                     let newNel = Nel.cons(value, x);
--- a/main/src/library/List.flix
+++ b/main/src/library/List.flix
@@ -219,6 +219,14 @@ mod List {
         foldRight((x, acc) -> x :: acc, l2, l1)
 
     ///
+    /// Returns `l2` appended to `reverse(l1)`.
+    ///
+    /// More efficient implementation than `append(reverse(l1), l2)` as it has 1 less reverse.
+    ///
+    def reverseAppend(l1: List[a], l2: List[a]): List[a] =
+        foldLeft((acc, x) -> x :: acc, l2, l1)
+
+    ///
     /// Returns `true` if and only if `l` contains the element `x`.
     ///
     pub def memberOf(a: a, l: List[a]): Bool with Eq[a] = match l {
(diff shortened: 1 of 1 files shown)


magnus-madsen:
We should carefully check that the order effects is unchanged. (We probably do not have test cases for that).

CasperDN:
As in write a test for all changed functions or inspect the changes?

Most of the changes go through the list left-to-right and end with `reverse` or `reverseAppend`.

Exceptions include `foldRight`, `reduceRight`, `scanRight` and `foldRight2`. All of these should be traversed right-to-left so they should be fine.

magnus-madsen:
> As in write a test for all changed functions or inspect the changes?

No no. Too much work to write test cases. But reading through the changes - both of us - while thinking about evaluation order would be useful.

> Most of the changes go through the list left-to-right and end with `reverse` or `reverseAppend`.

That is good.

> Exceptions include `foldRight`, `reduceRight`, `scanRight` and `foldRight2`. All of these should be traversed right-to-left so they should be fine.

That is also good.

CasperDN:
Here's my overview of the evaluation order of changed functions, where the evaluation order might matter. The evaluation order is what it will be with the proposed changes. 

| Function | Evaluation order |
| -------- | ------- |
| scanLeft  | left-to-right |
| scanRight | right-to-left |
| map | left-to-right |
| mapWithIndex | left-to-right |
| update | N/A |
| patch | N/A |
| removeAdjDupsWith | left-to-right |
| applyHelper | N/A |
| intersperse | N/A|
| intercalate | N/A |
| foldRight | right-to-left |
| reduceRight | right-to-left |
| flatten | N/A |
| filter | left-to-right |
| init | N/A |
| slice | N/A |
| partition | left-to-right |
| span | left-to-right |
| take | N/A |
| takeWhile | left-to-right |
| zip | N/A |
| zipWith | left-to-right |
| zipWithIndex | N/A |
| unzip | N/A |
| zipWith3 | left-to-right |
| unzip3 | N/A |
| foldRight2 | right-to-left |
| filterMap | left-to-right |
| unfold | left-to-right |
| unfoldWithIter | left-to-right |
| unfoldWithOkIter | left-to-right |
| distinct | left-to-right |
| distinctWith | left-to-right |
| merge | N/A |

main/src/library/List.flix
```diff
@@ -353,10 +361,10 @@ mod List {
     ///
     pub def map(f: a -> b \ ef, l: List[a]): List[b] \ ef =
         def loop(ll, acc) = match ll {
-            case Nil     => acc
+            case Nil     => reverse(acc)
```
magnus-madsen, line 364: I think this breaks the inliner-- because now the tail call has to happen inside the loop. The previous code was better.

magnus-madsen: @jaschdoc What are your thoughts? What happens when loop has to capture reverse like this?

magnus-madsen (commented):
I added one comment that we need to think about before we can address the rest.

main/src/library/List.flix
CasperDN: Rip. I though it would allow the function to finish directly instead of having to return to `map` for the `reverse` call. If this is worse I'll change it everywhere, as it is a common pattern in all functions which end with `reverse`.

magnus-madsen: I am not sure, but now reverse is free inside `loop` which might mess with inliner/optimizer/lambda lift.

magnus-madsen:
It might be better to do small, more focused PRs. E.g. like the one with foldRight

EDIT: I am not saying one PR per function, but perhaps just a couple of functions (that are in the same style) per PR.

CasperDN:
Added in smaller PRs

PR #11400, opened 2025-08-22
test: add test for `ExtMatch`

jaschdoc:


magnus-madsen:
What are we testing here?

jaschdoc:
I was testing unification. It seems odd to me that extensible variants can be closed. Shouldn't `#| A(Int32) |#` always be open?
```
ematch generatesA() {
            case B(i) => i == 1
            case A(i) => i == 1
        }

    def generatesA(): #| A(Int32) |# = xvar A(1)
```

magnus-madsen:
In your example, you have explicitly closed the row, so I guess it is closed and hence fails to compile. 

We have had discussions about closed vs. open before, but I am not sure if any specific consensus emerged.

Implicitly opening rows also has some downsides: spooky action at a distance and warnings about type variables that the programmer has not written.

jaschdoc:
Yeah, ok maybe we just leave this for now then.

jaschdoc:
I added some more tests for function calls.

magnus-madsen (approved):

Issue #11323, opened 2025-08-13
Brainstorm `Assert` effect


magnus-madsen:
Claude suggests:

```scala

///
/// An effect used to perform assertions during program execution.
///
pub eff Assert {

    ///
    /// Asserts that the given condition `cond` is `true` with an optional message `msg`.
    ///
    /// If the assertion fails, the behavior depends on the effect handler.
    ///
    def asrt(cond: Bool, msg: String): Unit

}

mod Assert {

    ///
    /// Asserts that the given condition `cond` is `true`.
    ///
    pub def assertTrue(cond: Bool): Unit \ Assert =
        Assert.asrt(cond, "Assertion failed: expected true")

    ///
    /// Asserts that the given condition `cond` is `true` with the given message `msg`.
    ///
    pub def assertTrueWithMsg(cond: Bool, msg: String): Unit \ Assert =
        Assert.asrt(cond, msg)

    ///
    /// Asserts that the given condition `cond` is `false`.
    ///
    pub def assertFalse(cond: Bool): Unit \ Assert =
        Assert.asrt(not cond, "Assertion failed: expected false")

    ///
    /// Asserts that the given condition `cond` is `false` with the given message `msg`.
    ///
    pub def assertFalseWithMsg(cond: Bool, msg: String): Unit \ Assert =
        Assert.asrt(not cond, msg)

    ///
    /// Asserts that the given values `actual` and `expected` are equal.
    ///
    pub def assertEq(actual: a, expected: a): Unit \ Assert with Eq[a], ToString[a] =
        Assert.asrt(actual == expected, "Assertion failed: expected `${expected}`, but got `${actual}`")

    ///
    /// Asserts that the given values `actual` and `expected` are equal with the given message `msg`.
    ///
    pub def assertEqWithMsg(actual: a, expected: a, msg: String): Unit \ Assert with Eq[a] =
        Assert.asrt(actual == expected, msg)

    ///
    /// Asserts that the given values `actual` and `unexpected` are not equal.
    ///
    pub def assertNeq(actual: a, unexpected: a): Unit \ Assert with Eq[a], ToString[a] =
        Assert.asrt(actual != unexpected, "Assertion failed: expected values to be different, but both were `${actual}`")

    ///
    /// Asserts that the given values `actual` and `unexpected` are not equal with the given message `msg`.
    ///
    pub def assertNeqWithMsg(actual: a, unexpected: a, msg: String): Unit \ Assert with Eq[a] =
[...]

PR #11402, opened 2025-08-22
feat: preliminary assert effect related to #11323

NeonOxide:
Created a preliminary version of the Assert effect and some utility functions like assertTrue, assertEq, etc.

There are some extra things to consider as: 

- Should isSome et al. print require toString and print the unexpected contents?
- I also think we should expose some interface where the message is generated from a function with the offending element as argument, in order to not force users to use their toString implementation.
- Should runWithIO expose the Sys Effect or the Environment effect from Chalk?
- I removed the hole present in the old eq in order to avoid the super exception and just returned false and it seems to work fine, but I do not know if there was any reason for the hole's existence or it was just left there. 
- Right now, handleWithCount returns a tuple (result, number_of_errors) and handleWithCollection returns a Result[errorList, result). I think both of these should be consistent with each other. However, I do not know which one is preferable (or even using Validation instead of result). What do you think?
- Right now runWithIO throws an Assertion error that gives a very ugly stacktrace, I guess once the code generation around the Assert effect is done, it can be managed that the handler just returns early with false in order to give the users what they expect. However then the signature of runWithIO would require to take an a -> Bool instead of a -> b. But given its use I think it merits its own function to fit with the Test framework.
- I have been researching other similar frameworks and one essential operation that would be really nice to support is something like assertThrows or assertNotThrows (which in our case could be generalized to assertCallsEffect or assertDoesNotCallEffect). 
 However, I have not been able to find a way of emulating this at Flix user level (at compiler level it would just need to introduce a handler that for every function either fails when called or fails at the end if not called). The user interface can be done quite seamlessly with something like assertCallsEffect(function, (ef: Eff[effect_equation])). Where Eff or another name is just an enum with a phantom effect kind. 
 - With typematch you could also do some assertIsPure or assertHasEffect that instead of simply not compiling use typematch and return true or false depending on the effects. However, I fail to see why an user would prefer to obtain these errors at test time instead of at compile time. However, I do not know how typematch works with subeffecting, if a pure function is treated as IO by subeffecting, when used with typematch will it match on Pure or on IO?
 
 Thanks in advance for your time! This is the first of a series of pull requests that will be done regarding Assert, once this is done, I will go into allowing Assert at the top level of tests and including a default handler for it if that is the case.

### DIFF
--- a/main/src/library/Assert.flix
+++ b/main/src/library/Assert.flix
@@ -13,7 +13,136 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+///
+/// An effect used to perform assertions during program execution.
+///
+pub eff Assert {
+
+    ///
+    /// Asserts that the given condition `cond` is `true` with a message `msg`.
+    /// The provided functions in the associated module treat this message as
+    /// the reason of failure, but it does not necessarily have to be that.
+    ///
+    /// If the assertion fails, the behavior depends on the effect handler.
+    ///
+    def assert(cond: Bool, msg: String): Unit
+
+}
+
 mod Assert {
+    import java.lang.AssertionError;
+
+    ///
+    /// Asserts that the given condition `cond` is `true`.
+    ///
+    pub def assertTrue(cond: Bool): Unit \ Assert =
+        Assert.assert(cond, "Assertion failed: expected true")
+
+    ///
+    /// Asserts that the given condition `cond` is `true` with the given message `msg`.
+    ///
+    pub def assertTrueWithMsg(cond: Bool, msg: String): Unit \ Assert =
+        Assert.assert(cond, msg)
+
+    ///
+    /// Asserts that the given condition `cond` is `false`.
+    ///
+    pub def assertFalse(cond: Bool): Unit \ Assert =
+        Assert.assert(not cond, "Assertion failed: expected false")
+
+    ///
+    /// Asserts that the given condition `cond` is `false` with the given message `msg`.
+    ///
+    pub def assertFalseWithMsg(cond: Bool, msg: String): Unit \ Assert =
+        Assert.assert(not cond, msg)
+
+    ///
+    /// Asserts that the given values `expected` and `actual` are equal.
+    ///
+    pub def assertEq(exp: {expected = a}, actual: a): Unit \ Assert with Eq[a], ToString[a] =
+        Assert.assert(actual == exp#expected, "Assertion failed: expected `${exp#expected}`, but got `${actual}`")
+
+    ///
+    /// Asserts that the given values `expected` and `actual` are equal with the given message `msg`.
+    ///
+    pub def assertEqWithMsg(exp: {expected = a}, actual: a, msg: String): Unit \ Assert with Eq[a] =
+        Assert.assert(actual == exp#expected, msg)
+
+    ///
+    /// Asserts that the given values `unexpected` and `actual` are not equal.
+    ///
+    pub def assertNeq(unexp: {unexpected = a}, actual: a): Unit \ Assert with Eq[a], ToString[a] =
+        Assert.assert(actual != unexp#unexpected, "Assertion failed: expected values to be different, but both were `${actual}`")
+
+    ///
+    /// Asserts that the given values `unexpected` and `actual` are not equal with the given message `msg`.
+    ///
+    pub def assertNeqWithMsg(unexp: {unexpected = a}, actual: a, msg: String): Unit \ Assert with Eq[a] =
+        Assert.assert(actual != unexp#unexpected, msg)
+
+    ///
+    /// Asserts that the given Option `o` is `Some(v)`.
+    ///
+    pub def assertSome(o: Option[a]): Unit \ Assert =
+        Assert.assert(Option.nonEmpty(o), "Assertion failed: expected Some(_), but got None")
+
+    ///
+    /// Asserts that the given Option `o` is `None`.
+    ///
+    pub def assertNone(o: Option[a]): Unit \ Assert =
+        Assert.assert(Option.isEmpty(o), "Assertion failed: expected None, but got Some(_)")
+
+    ///
+    /// Asserts that the given Result `r` is `Ok(v)`.
+    ///
+    pub def assertOk(r: Result[e, a]): Unit \ Assert =
+        Assert.assert(Result.isOk(r), "Assertion failed: expected Ok(_), but got Err(_)")
+
+    ///
+    /// Asserts that the given Result `r` is `Err(e)`.
+    ///
+    pub def assertErr(r: Result[e, a]): Unit \ Assert =
+        Assert.assert(Result.isErr(r), "Assertion failed: expected Err(_), but got Ok(_)")
+
+    ///
+    /// Unconditionally fails with the given message `msg`.
+    ///
+    pub def fail(msg: String): Unit \ Assert =
+        Assert.assert(false, msg)
+
+    ///
+    /// Handles the `Assert` effect of the given function `f` by throwing exceptions on assertion failures.
+    ///
+    /// In other words, re-interprets the `Assert` effect using the `IO` effect.
+    ///
+    pub def handle(f: a -> b \ ef): (a -> b \ (ef - Assert) + IO + Sys) =
+        x ->
+        run {
+            f(x)
+        } with handler Assert {
+            def assert(cond, msg, resume) =
+            {
+                if (cond)
+                    resume(())
+                else {
+                    use Chalk.{red, bold};
+                    run {
+                        println("");
+                        println("${red("Assertion Failed:")} ${bold(msg)}");
+                        println("");
+                        throw new AssertionError(msg)
+                    } with Environment.runWithIO
+                }
+            }
+        }
+
+    ///
+    /// Runs the `Assert` effect of the given function `f` by throwing exceptions on assertion failures.
+    ///
+    /// In other words, re-interprets the `Assert` effect using the `IO` effect.
+    ///
+    pub def runWithIO(f: Unit -> a \ ef): a \ (ef - Assert) + IO + Sys = handle(f)()
 
     ///
     /// Asserts that `expected` must equal `actual`.
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -83,3 +83,4 @@ By adding your name to this document, you agree to release all your contribution
 - [Ry Wiese](https://github.com/rywiese)
 - [Michał Kukieła](https://github.com/kukimik)
 - [Alex Asafov](https://github.com/Alex1005a)
+- [Daniel Neo López Martínez](https://github.com/NeonOxide)


main/src/library/Assert.flix
```diff
@@ -13,7 +13,228 @@
...
+/// An effect used to perform assertions during program execution.
+///
+pub eff Assert {
+
+    ///
+    /// Asserts that the given condition `cond` is `true` with an optional message `msg`.
+    ///
+    /// If the assertion fails, the behavior depends on the effect handler.
+    ///
+    def assert(cond: Bool, msg: String): Unit
```
magnus-madsen, line 29: Do we want the msg to be optional?

main/src/library/Assert.flix
```diff
@@ -13,7 +13,228 @@
...
+    ///
+    /// Asserts that the given condition `cond` is `false` with the given message `msg`.
+    ///
+    pub def assertFalseWithMsg(cond: Bool, msg: String): Unit \ Assert =
+        Assert.assert(not cond, msg)
+
+    ///
+    /// Asserts that the given values `actual` and `expected` are equal.
+    ///
+    pub def assertEq(actual: a, expected: a): Unit \ Assert with Eq[a], ToString[a] =
```
magnus-madsen: I think it is reasonable to require ToString[a].

- I would change the signature to make the expected value the first.
- I would change the signature to require a record argument as the first argument, e.g. `assertEq(expected: {expected = String}, actual: a)`

main/src/library/Assert.flix
```diff
@@ -13,7 +13,228 @@
...
+        Assert.assert(actual == expected, "Assertion failed: expected `${expected}`, but got `${actual}`")
```
magnus-madsen: Later we can think about how to color expected and actual.

main/src/library/Assert.flix
```diff
@@ -13,7 +13,228 @@
...
+    pub def assertNeq(actual: a, unexpected: a): Unit \ Assert with Eq[a], ToString[a] =
+        Assert.assert(actual != unexpected, "Assertion failed: expected values to be different, but both were `${actual}`")
+
+    ///
+    /// Asserts that the given values `actual` and `unexpected` are not equal with the given message `msg`.
+    ///
+    pub def assertNeqWithMsg(actual: a, unexpected: a, msg: String): Unit \ Assert with Eq[a] =
+        Assert.assert(actual != unexpected, msg)
+
+
```
magnus-madsen: Why extra line breaks? Just trim them :)

main/src/library/Assert.flix
```diff
@@ -13,7 +13,228 @@
...
+
+    ///
+    /// Asserts that the given element `x` is a member of the given collection `iter`.
+    ///
+    pub def assertMemberOf(x: a, iter: iter[a]): Unit \ Assert + aef with ToString[a], Eq[a], Foldable[iter] where Foldable.Aef[iter] ~ aef = {
```
magnus-madsen: iter -> fa
iter[a] -> f[a]

main/src/library/Assert.flix
```diff
@@ -13,7 +13,228 @@
...
+    ///
+    /// Asserts that the given values `actual` and `unexpected` are not equal with the given message `msg`.
+    ///
+    pub def assertNeqWithMsg(actual: a, unexpected: a, msg: String): Unit \ Assert with Eq[a] =
+        Assert.assert(actual != unexpected, msg)
+
+
+
+    ///
+    /// Asserts that the given element `x` is a member of the given collection `iter`.
```
magnus-madsen: I would say ... is a member of foldable data type.

main/src/library/Assert.flix
```diff
@@ -13,7 +13,228 @@
...
+    ///
+    pub def assertMemberOf(x: a, iter: iter[a]): Unit \ Assert + aef with ToString[a], Eq[a], Foldable[iter] where Foldable.Aef[iter] ~ aef = {
+        Assert.assert(Foldable.memberOf(x, iter), "Assertion failed: `${x}` is not a member of the given collection")
+    }
+
+    ///
+    /// Asserts that the given element `x` is a member of the given collection `iter`.
+    ///
+    pub def assertNotMemberOf(x: a, iter: iter[a]): Unit \ Assert + aef with ToString[a], Eq[a], Foldable[iter] where Foldable.Aef[iter] ~ aef = {
```
magnus-madsen: Is the plan to have negation of every assert? Just a question

main/src/library/Assert.flix
```diff
@@ -13,7 +13,228 @@
...
+    ///
+    pub def assertOk(r: Result[e, a]): Unit \ Assert =
+        Assert.assert(Result.isOk(r), "Assertion failed: expected Ok(_), but got Err(_)")
+
+    ///
+    /// Asserts that the given Result `r` is `Err(e)`.
+    ///
+    pub def assertErr(r: Result[e, a]): Unit \ Assert =
+        Assert.assert(Result.isErr(r), "Assertion failed: expected Err(_), but got Ok(_)")
+
```
magnus-madsen, line 107: I think it is reasonable to have special logic for Option and Result 👍

magnus-madsen: I would move this up before the isMemberOf

main/src/library/Assert.flix
```diff
@@ -13,7 +13,228 @@
...
+    pub def assertEmpty(iter: iter[a]): Unit \ Assert + aef with Foldable[iter] where Foldable.Aef[iter] ~ aef = {
+        Assert.assert(Foldable.isEmpty(iter), "Assertion failed: expected empty collection, but got non-empty collection")
+    }
+
+    ///
+    /// Asserts that the given collection `iter` is not empty.
+    ///
+    pub def assertNonEmpty(iter: iter[a]): Unit \ Assert + aef with Foldable[iter] where Foldable.Aef[iter] ~ aef = {
+        Assert.assert(not Foldable.isEmpty(iter), "Assertion failed: expected non-empty collection, but got empty collection")
+    }
```
magnus-madsen: missing line break

main/src/library/Assert.flix
```diff
@@ -13,7 +13,228 @@
...
+    pub def assertErr(r: Result[e, a]): Unit \ Assert =
+        Assert.assert(Result.isErr(r), "Assertion failed: expected Err(_), but got Ok(_)")
+
+    ///
+    /// Asserts that the given collection `iter` is empty.
+    ///
+    pub def assertEmpty(iter: iter[a]): Unit \ Assert + aef with Foldable[iter] where Foldable.Aef[iter] ~ aef = {
+        Assert.assert(Foldable.isEmpty(iter), "Assertion failed: expected empty collection, but got non-empty collection")
+    }
+
```
magnus-madsen: In general what collection/foldable queries should we have? Perhaps we should start without them and only have the ones for Option/Result?

main/src/library/Assert.flix
```diff
@@ -13,7 +13,228 @@
...
+
+
+
+    ///
+    /// Handles the `Assert` effect of the given function `f` by collecting assertion failures.
+    ///
+    /// Returns `Ok(result)` if all assertions pass, or `Err(failures)` with a list of failure messages.
+    ///
+    pub def handleWithCollection(f: a -> b \ ef): a -> Result[List[String], b] \ ef - Assert = x ->
```
magnus-madsen: I like it, but maybe a better name? It does not have to be called handleWIth? Or does it? What do you think?

main/src/library/Assert.flix
```diff
@@ -13,7 +13,228 @@
...
+            else
+                Err(failureList)
+        }
+
+    ///
+    /// Handles the `Assert` effect of the given function `f` by counting assertion failures.
+    ///
+    /// Returns a tuple of (result, failureCount).
+    ///
+    pub def handleWithCount(f: a -> b \ ef): a -> (b, Int32) \ ef - Assert = x ->
```
magnus-madsen: I would remove this

main/src/library/Assert.flix
```diff
@@ -26,11 +247,10 @@ mod Assert {
         else {
             use Chalk.{green, red};
             Environment.runWithIO(() -> {
-                println("Assertion Error");
                 println("  Expected: ${green(expected)}");
                 println("  Actual:   ${red(actual)}");
                 println("");
-                ?assertEq
+                false
             })
```
magnus-madsen: Please keep the above function intact until we refactor and remove it.

magnus-madsen (commented):
I will give high-level comments now and then we can focus on details after one iteration.

- Remember to add your name to AUTHORS.md

magnus-madsen:
I think this is a good start. I suggest to reduce the scope a bit for the first PR. (E.g. drop the Foldable things and the handler with count. We can add them in later PRs.)

magnus-madsen:
We typically prefer many small PRs than can be merged without too many iterations.

main/src/library/Assert.flix
NeonOxide: I think not given that an assertion handler that only receives a boolean and no message cannot do much more than just say failed or not failed with resuming as it does not have access to the context of where the boolean received comes from. 

This behavior could be useful when you just want to count the errors or just fail in case of error without any information. However, I think that even in this cases the messages also work as documentation so I do not think it is necessary to wrap every other call to assert with a Some in order to support what I consider a bad practice. What do you think?

main/src/library/Assert.flix
NeonOxide: I agree. I also think that given that assertEq and assertNeq are basically the same, assertNeq should also be given a first record argument with unexpected.

main/src/library/Assert.flix
NeonOxide: I think that is what they usually do in most assertion libraries so I would say yes. I think not only does it show the intention behind the assertion itself, it also allows us to provide nice default messages for the majority of these cases.

main/src/library/Assert.flix
NeonOxide: Totally agree, I did not know what to call the data type so after a couple of tries I just put collection but foldable is much more precise.

main/src/library/Assert.flix
NeonOxide: I agree the Option and Result should have special treatment as they do now given that they are quite ubiquitous, but would it make more sense to print their contents when the assertions fail or not to print them? That was my main concern.

main/src/library/Assert.flix
NeonOxide: Ok. I removed that because without it the error message is quite ugly and just returning false seemed to provide the desired behavior, eg. just printing the Expected Actual message without a really big stacktrace with the internals of Flix. But I will put it back

NeonOxide:
Accidentally closed this by updating my own repo with the changes upstream from Flix. I will reopen

magnus-madsen:
@NeonOxide Remember to mark comments as resolved.

main/src/library/Assert.flix
magnus-madsen: OK

main/src/library/Assert.flix
magnus-madsen: But does it scale?

main/src/library/Assert.flix
magnus-madsen: Good question, I have no answer.

magnus-madsen:
@NeonOxide Congratulations on landing your first PR!

Issue #11403, opened 2025-08-22
`pquery` crash in Lowering when extracting non-existent relations


JumpTwiice:
The program
```
def main(): Unit = 
    let _ = pquery #{ A(3) } select A(3) with { B };
    ()
```
crashes with
```
#
# An unexpected error has been detected by the Flix compiler:
#
#   Expected Type.Apply(_, _, _), but got '3576267 (C:\Users\adamt\Desktop\flix\playground2\main.flix:2:13)
#
# This is a bug in the Flix compiler. Please report it here:
#
# https://github.com/flix/flix/issues
#
# -- Flix Compiler --
#
# Flix Version : 0.60.0
#   incremental: All
#
# -- Java Virtual Machine --
#
# JVM Version  : 21.0.3 (2024-04-16)
# JVM Vendor   : Oracle Corporation
# JAVA_HOME    : C:\Program Files\Java\jdk-21
# System       : Windows 11 (10.0)
#
# -- Stack Trace --
ca.uwaterloo.flix.util.InternalCompilerException: Expected Type.Apply(_, _, _), but got '3576267 (C:\Users\adamt\Desktop\flix\playground2\main.flix:2:13)
	at ca.uwaterloo.flix.language.phase.Lowering$.f$1(Lowering.scala:1818)
	at ca.uwaterloo.flix.language.phase.Lowering$.termTypesOfRelation(Lowering.scala:1821)
	at ca.uwaterloo.flix.language.phase.Lowering$.predicatesOfSchemaRow(Lowering.scala:1805)
	at ca.uwaterloo.flix.language.phase.Lowering$.predicatesOfExtVar(Lowering.scala:1793)
	at ca.uwaterloo.flix.language.phase.Lowering$.visitExp(Lowering.scala:829)
	at ca.uwaterloo.flix.language.phase.Lowering$.visitExp(Lowering.scala:483)
	at ca.uwaterloo.flix.language.phase.Lowering$.visitDef(Lowering.scala:189)
	at ca.uwaterloo.flix.language.phase.Lowering$.$anonfun$run$2(Lowering.scala:163)
	at ca.uwaterloo.flix.util.ParOps$.$anonfun$parMapValues$1(ParOps.scala:91)
	at ca.uwaterloo.flix.util.ParOps$.$anonfun$parMap$2(ParOps.scala:65)
	at java.base/java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1423)
[...]

PR #11404, opened 2025-08-22
fix: `pquery` crash in Lowering when using non-existent relations

JumpTwiice:
Fixes https://github.com/flix/flix/issues/11403.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
@@ -1805,7 +1805,7 @@ object Lowering {
       (pred, termTypesOfRelation(rel, loc2)) :: predicatesOfSchemaRow(tpe2, loc1)
     case Type.Var(_, _) => Nil
     case Type.SchemaRowEmpty => Nil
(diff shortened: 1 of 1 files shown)


JumpTwiice:
With this fix, `B` becomes an extensible variant with 0 arguments, which I think is the sanest approach. The program
```
def main(): Unit \ IO = 
    let xx = pquery #{ A(3) } select A(3) with { B };
    Vector.forEach(x -> ematch x {
        case B() => println("hi")
    }, xx)
```
type checks and does not print `hi`.

magnus-madsen:
The problem here is that this solves the program for a monomorphic program (a program without polymorphism), but for a polymorphic program it is probably wrong. Unfortunately there is no way for us to fix this until we perform monomorphization and lowering at the same time.

JumpTwiice:
Oh yeah. Forgot about this.

magnus-madsen:
The problem is that we have a type variable X in hand. We cannot know which of two case are true:

- Will X be replaced by an actual row type during Monomorphization? or 
- Will X be left free during monomorphization and hence we can replace it by whatever we want (e.g. Unit).

PR #11407, opened 2025-08-22
feat: fill out 'output' function

CadeMichael:
addresses #10894.

Example of running in a Main.flix

```
def main(): Result[IoError, Unit] \ {Exec, IO} =
    run {
        let p = ProcessDev.exec("ls", Nil);
        ProcessDev.waitFor(p);
        let str = ProcessDev.output(p);
        match str {
            case Ok(s) => println(s)
            case Err(_) => println("error")
        }
    } with ProcessDev.runWithIO
```


magnus-madsen:
Fixed in main.

PR #11410, opened 2025-08-23
fix: a few uses of `???`

magnus-madsen:


main/src/library/Process.flix
```diff
@@ -213,9 +208,6 @@ mod Process {
                 let ProcessHandle(jProc) = ph;
                 k(StdErr(jProc.getErrorStream()))
 
-            def output(ph, k) = ???
```
magnus-madsen, line 216: Better to remove than crash. There is a PR that adds some of this stuff back- but it will have to come later.

PR #11411, opened 2025-08-23
chore: change Datalog nullary semantics to `A != A() == A(())`

CasperDN:
This allows `A` everywhere. Note that previously this program was allowed:
```
def main(): Unit \ IO = {
    let program: #{A, C(Int32)} = #(A, C) -> #{
        B(()).
        C(1) :- A, B(()).
    };
    println(query program select () from A, C(1));
    ()
}
```
Only declaring `A.` as a fact was disallowed, apparently. Note that typewise `A()` is still considered to be `A` schema-rows. I do not know whether this is wanted.

The change will make the program
```
def main(): Unit \ IO = {
    let program = #{
        A(()).
        B().
        C.
        D(0) :- A().
        D(1) :- A(()).
        D(2) :- B().
        D(3) :- B(()).
        D(4) :- C.
        D(5) :- E.
    };
    println(query program select i from D(i));
    println(query program select () from A());
    println(query program select () from A(()));
    println(query program select () from C);
    println(query program select () from E);
    ()
}
```
print
```
Vector#{0, 1, 2, 3, 4}
Vector#{()}
Vector#{()}
Vector#{()}
Vector#{}
```

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
@@ -2835,724 +2835,724 @@ object Weeder2 {
         case (Nil, latTerm) => (List(Pattern.Cst(Constant.Unit, tree.loc)), latTerm)
         case (nonEmpty, latTerm) => (nonEmpty, latTerm)
       }
-  }
-
-  private def visitGuard(tree: Tree)(implicit sctx: SharedContext): Validation[Predicate.Body.Guard, CompilationMessage] = {
-    expect(tree, TreeKind.Predicate.Guard)
-    mapN(Exprs.pickExpr(tree))(Predicate.Body.Guard(_, tree.loc))
-  }
-
-  private def visitFunctional(tree: Tree)(implicit sctx: SharedContext): Validation[Predicate.Body.Functional, CompilationMessage] = {
-    expect(tree, TreeKind.Predicate.Functional)
-    val idents = pickAll(TreeKind.Ident, tree).map(tokenToIdent)
-    mapN(Exprs.pickExpr(tree))(Predicate.Body.Functional(idents, _, tree.loc))
-  }
-
-  def visitParam(tree: Tree)(implicit sctx: SharedContext): Validation[PredicateParam, CompilationMessage] = {
-    expectAny(tree, List(TreeKind.Predicate.Param, TreeKind.Predicate.ParamUntyped))
-    val types0 = pickAll(TreeKind.Type.Type, tree)
-    val maybeLatTerm = tryPickLatticeTermType(tree)
-    mapN(pickNameIdent(tree), traverse(types0)(Types.visitType), maybeLatTerm) {
-      case (ident, Nil, _) => PredicateParam.PredicateParamUntyped(Name.mkPred(ident), tree.loc)
-      case (ident, types, None) => PredicateParam.PredicateParamWithType(Name.mkPred(ident), Denotation.Relational, types, tree.loc)
-      case (ident, types, Some(latTerm)) => PredicateParam.PredicateParamWithType(Name.mkPred(ident), Denotation.Latticenal, types :+ latTerm, tree.loc)
     }
-  }
 
-  private def tryPickLatticeTermExpr(tree: Tree)(implicit sctx: SharedContext): Validation[Option[Expr], CompilationMessage] = {
-    traverseOpt(tryPick(TreeKind.Predicate.LatticeTerm, tree))(Exprs.pickExpr)
-  }
-
-  private def tryPickLatticeTermType(tree: Tree)(implicit sctx: SharedContext): Validation[Option[Type], CompilationMessage] = {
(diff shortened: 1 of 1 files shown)


magnus-madsen:
Could we add some tests. Test.Exp.Fixpoint.Predicate.Nullary (or similar) to illustrate what this allows?

EDIT: Essentially your great example could be broken down into a handful of test cases.

magnus-madsen:
> Only declaring A. as a fact was disallowed, apparently. Note that typewise A() is still considered to be A schema-rows. I do not know whether this is wanted.

What does this mean?

magnus-madsen:
@jaschdoc @JonathanStarup Do you have some cents to share here?

JonathanStarup:
seems sensible

CasperDN:
> Could we add some tests. Test.Exp.Fixpoint.Predicate.Nullary (or similar) to illustrate what this allows?

Sure. I just wanted to see whether we would want this before I wrote a bunch of tests. I'll add some tests.

> > Only declaring A. as a fact was disallowed, apparently. Note that typewise A() is still considered to be A schema-rows. I do not know whether this is wanted.
> 
> What does this mean?

It means that I could type the example program with: `#{A(Unit), B(Unit), C, E(), D(Int32)}`. In other words `E()` type-wise is still equivalent to `E`, not `E(Unit)`.

I am unsure whether this is desired and whether it should be fixed here or in a later PR, if it is undesired.

jaschdoc:
> It means that I could type the example program with: `#{A(Unit), B(Unit), C, E(), D(Int32)}`. In other words `E()` type-wise is still equivalent to `E`, not `E(Unit)`.

Yes, I think the syntax should also reflect this change so `A: #{ A }` and `A() : #{ A(Unit) }`. Just like with ADTs, the type `#{ A() }` should be invalid.

Other than that I think it looks good.

magnus-madsen:
Could we add some negative tests to show me what is no longer expected to type check?

magnus-madsen:
@jaschdoc If this is merged, do we then need to make changes to ExtVars?

Does this solve all problems-- while still enforcing that inject only works for n>0 relations?

jaschdoc:
> @jaschdoc If this is merged, do we then need to make changes to ExtVars?

No, ExtVars already behave like this.

> Does this solve all problems-- while still enforcing that inject only works for n>0 relations?

Yes, I believe so.

CasperDN:
@magnus-madsen I believe this is done. I will make another PR proposing the invalidation of `${ A() }` and `#| A() |#`.

magnus-madsen:
OK; I just need to catch up.

magnus-madsen:
Make it look a bit like visitTagPat in  Weeder

CasperDN:
@magnus-madsen Its a bit more like `visitTagPat` now. If you want I could change the matching from
```
X match {
  case None =>
  case Some((terms, None))
  case Some((terms, Some(latTerm))
}
```
to a new matching on the instead. The pair is the patterns/terms and the potential lattice value.

magnus-madsen:
@CasperDN What happened to Weeder2? Maybe it was reformatted. But if so, that needs to be a separate PR.

I suggest to check out the original file and then backport your changes so only the relevant lines are touched.

CasperDN:
@magnus-madsen Agreed. I blame Intellij.

magnus-madsen:
> @magnus-madsen Agreed. I blame Intellij.

:-/ but we have an .idea folder that sets the style

CasperDN:
I would guess it/I formatted at some point where the parenthesis were messed up.

magnus-madsen:
@CasperDN Unfortunately I thinks PR broke some parsing. For example, now this parses:

```
    let pr = #{
        Path(x, y) :- Edge(x, y).
        Pat
    };
```

Note the missing dot.

This also means that auto-complete fails.

magnus-madsen:
@CasperDN Will you look into a fix or should I undo the PR?

CasperDN:
I will look into it.

CasperDN:
I am unsure what the problem is exactly. The following program runs fine in the release 0.61 (before the PR). As far as I know the last dot has been voluntary for some time?
```
def tessst(): Unit \ IO = {
    let pr = #{
        Edge(1,2).
        Path(x, y) :- Edge(x, y).
        Pat()
    };
    println(query pr select () from Pat)
}
```

CasperDN:
The only problem I can see is that nullary constraints autocomplete to `/1`, but this was also the case in 0.61.

magnus-madsen:
There is a missing dot. (I am not sure if that bug was already present?)

magnus-madsen:
Hmm. It seems master before PR also allowed that missing dot :-/

magnus-madsen:
Since this issue was already present, I suppose it is not directly related to this PR. I have created a ticket here: 

https://github.com/flix/flix/issues/11439

but I won't require you to fix it :)

CasperDN:
I honestly thought it was a feature that the last dot was optional.

PR #11414, opened 2025-08-23
refactor: remove `foreach-yield`

magnus-madsen:


magnus-madsen:
@jaschdoc Anything I forgot to remove here?

jaschdoc:
I don't think so. It was handled in `Desugar` so any removal of that and the following consequences should hit all the reachable stuff. I think there are some markdown files in the `docs/` directory that mention `foreach-yield`, e.g., `DIDYOUKNOW.md`.

jaschdoc (approved):

PR #11423, opened 2025-08-25
fix: unify `ExtPattern.Tag` with closed row

jaschdoc:


magnus-madsen:
What's the consequence here? Did this allow ill-typed programs?

magnus-madsen:
If so, we would probably want some tests.

jaschdoc:
I'm not actually sure the type is used anywhere.

jaschdoc:
This is how the pattern is used in simplifier and other than that it's just carried through all phases.
```scala
      case (MonoAst.ExtMatchRule(MonoAst.ExtPattern.Tag(label, pats, _, _), exp1, loc1), branch2) =>
        val e1 = visitExp(exp1)
        val is = SimplifiedAst.Expr.ApplyAtomic(AtomicOp.ExtIs(label), List(extVar), SimpleType.Bool, Purity.Pure, extVar.loc)
        val termTypes = SimpleType.findExtensibleTermTypes(label, tagType)
```

It didn't matter during constraint gen either since the type of the tag is not used anywhere. So maybe it's better to just not have a type on the ext patterns (but that seems off too).

magnus-madsen:
It seems like `ExtPattern.Tag.tvar` is redundant?

jaschdoc:
Yes

magnus-madsen:
Can we remove `def tpe: Type` from `ExtPattern`? That would simplify everything I suppose.

magnus-madsen:
I am asking Claude. Let us see how it goes.

PR #11424, opened 2025-08-25
chore: make `#| A() |#` illegal

CasperDN:
This makes `#| A() |#` and `#{ A() }#` (as a type) illegal following discussion on #11411.

I'm unsure how this could be made nicer. Currently the program
```
def main() : #| A(2) |# = xvar A
```
reports
```
Illegal syntax: empty argument list of predicate.(Syntax Error)
Expected <type> before LiteralInt.(Parse Error (Unknown))
```
Also as long as the error is getting thrown linting seems to halt here...

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -3575,17 +3575,22 @@ object Parser2 {
         arguments()
         close(mark, TreeKind.Type.PredicateWithAlias)
       } else {
-        zeroOrMore(
-          namedTokenSet = NamedTokenSet.Type,
-          getItem = () => ttype(),
-          checkForItem = _.isFirstType,
-          breakWhen = _.isRecoverType,
-          optionallyWith = Some((TokenKind.Semi, () => {
-            val mark = open()
-            ttype()
-            close(mark, TreeKind.Predicate.LatticeTerm)
-          }))
-        )
+
+        if(at(TokenKind.ParenL)) {
+          val argMark = open()
+          zeroOrMore(
+            namedTokenSet = NamedTokenSet.Type,
+            getItem = () => ttype(),
+            checkForItem = _.isFirstType,
+            breakWhen = _.isRecoverType,
+            optionallyWith = Some((TokenKind.Semi, () => {
+              val mark = open()
+              ttype()
+              close(mark, TreeKind.Predicate.LatticeTerm)
+            }))
+          )
+          close(argMark, TreeKind.Type.ArgumentList)
+        }
         close(mark, TreeKind.Type.PredicateWithTypes)
       }
     }
--- a/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
@@ -3013,7 +3013,8 @@ object Weeder2 {
               pickAll(TreeKind.Type.Type, typeTree) match {
                 case Nil =>
                   val error = IllegalEmptyPredicateArgument(typeTree.loc)
-                  Failure(error)
+                  sctx.errors.add(error)
+                  Success(Type.Error(typeTree.loc))
                 case types =>
                   val maybeLatTerm = tryPick(TreeKind.Predicate.LatticeTerm, typeTree)
                   mapN(pickQName(tree), traverse(types)(Types.visitType), traverseOpt(maybeLatTerm)(Types.pickType)) {
--- a/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
@@ -3004,13 +3004,24 @@ object Weeder2 {
           }
 
         case (tree, acc) if tree.kind == TreeKind.Type.PredicateWithTypes =>
-          val types = pickAll(TreeKind.Type.Type, tree)
-          val maybeLatTerm = tryPick(TreeKind.Predicate.LatticeTerm, tree)
-          mapN(pickQName(tree), traverse(types)(Types.visitType), traverseOpt(maybeLatTerm)(Types.pickType)) {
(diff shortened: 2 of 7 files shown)


main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -3575,17 +3575,22 @@ object Parser2 {
...
+            getItem = () => ttype(),
+            checkForItem = _.isFirstType,
+            breakWhen = _.isRecoverType,
+            optionallyWith = Some((TokenKind.Semi, () => {
+              val mark = open()
+              ttype()
+              close(mark, TreeKind.Predicate.LatticeTerm)
+            }))
+          )
+          close(argMark, TreeKind.Type.ArgumentList)
```
jaschdoc, line 3592: Seems like you want to use `delimiterL = TokenKind.ParenL, delimiterR = TokenKind.ParenR` in your call to `zeroOrMore`.

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -3575,17 +3575,22 @@ object Parser2 {
...
-            ttype()
-            close(mark, TreeKind.Predicate.LatticeTerm)
-          }))
-        )
+
+        if(at(TokenKind.ParenL)) {
+          val argMark = open()
+          zeroOrMore(
+            namedTokenSet = NamedTokenSet.Type,
+            getItem = () => ttype(),
```
jaschdoc, line 3583: this can just be `getItem = ttype`

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -3004,13 +3004,24 @@ object Weeder2 {
...
+          tryPick(TreeKind.Type.ArgumentList, tree) match {
+            case None =>
+              mapN(pickQName(tree))(
+                qname => Type.SchemaRowExtendByTypes(qname.ident, Denotation.Relational, Nil, acc, tree.loc)
+              )
+            case Some(typeTree) =>
+              pickAll(TreeKind.Type.Type, typeTree) match {
+                case Nil =>
+                  val error = IllegalEmptyPredicateArgument(typeTree.loc)
+                  Failure(error)
```
jaschdoc: Here you can return `Success(Type.Error(typeTree.loc))` so the remaining compiler phases can run.

jaschdoc: And then push the `error` to `sctx.errors`. You can see how it's done in many other places in the weeder :)

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
CasperDN: The call to `zeroOrMore` is a verbatim copy of the old call and I don't think you usually write the default parameters explicitly.

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
CasperDN: Not according to Scala:
```
found   : ca.uwaterloo.flix.language.ast.TokenKind => ca.uwaterloo.flix.language.phase.Parser2.Mark.Closed
required: () => ca.uwaterloo.flix.language.phase.Parser2.Mark.Closed
```

jaschdoc: wow

CasperDN: I guess the explicit case makes Scala use the default parameter, while the other case is more likely be a programmer error and is reported as that, instead if just using the default.

jaschdoc: Ah yeah makes sense.

magnus-madsen:
You write:

````
def main() : #| A(2) |# = xvar A
```

Did you mean?

```
def main() : #| A() |# = xvar A
```

magnus-madsen:
When say making it illegal as a type, what do you mean?

CasperDN:
> When say making it illegal as a type, what do you mean?

It is not really the type that is illegal it is the syntax, in the same way that the empty tuple-type is illegal, syntactically.

> You write:
> 
> ```
> def main() : #| A(2) |# = xvar A
> ```
> 
> Did you mean?
> 
> ```
> def main() : #| A() |# = xvar A
> ```
I mean the second is what is disallowed. What I wanted to point out was simply that the first results in 2 errors. The first that `2` is not a type and the second that no types are provided.

I am unsure whether we should fix that or just accept that it is less than perfect at the moment.

jaschdoc:
It _should_ result in two errors, no?

CasperDN:
> It _should_ result in two errors, no?

I guess its a matter of taste. I think simply reporting that an integer is passed instead of a type gives sufficient information to the programmer.

If they try to fix all problems reported starting with `empty argument list of predicate` and see a clearly non-empty list of arguments then surely confusion will follow. It will of course help highlight that there is a problem here, but the first error succinctly explains what the problem is.

magnus-madsen:
Does this PR resolve this problem:

```

def main(): Unit \ IO =
    let db = #{
        A.
    };
    let pr = #{
        R :- A.
    };
    let pp = pquery db, pr select R with {A};
    pp |> Vector.forEach(ematch A -> println("Oh boy!"))

```

CasperDN:
This PR has nothing to do with any problems relating to Datalog. It was simply a question of the allowed syntax for types.

I have not really taken a deep dive into Flix' representation of types, but I believe the bug you highlight in that program is due to `Lowering.termTypesOfRelation` expecting the predicates to have arity > 0. Simply adding:
```
      case Type.Cst(TypeConstructor.Relation(_), _) => Nil
```
to the match makes it work, but it would probably be better to just use this as a base case. I'll make a PR for it in a bit.

magnus-madsen:
Is there any reason to make this type illegal now? (Is it only for consistency with ADTs?)

CasperDN:
It is only based on what I understood as the consensus from Gitter. We can close this if we don't want to disallow it.

Issue #10921, opened 2025-06-19
Fix Java field reading and writing


jaschdoc:
We forgot to put both back in. They exist in `ResolvedAst` but not before.

JonathanStarup:
Putfield is `exp.name = exp` and PutStatic is `QName = exp` where resolver has to resolve the QName (with one dot, maybe that doesn't matter) to a java static field

PR #11425, opened 2025-08-25
fix: add `PutField` for java interop

jaschdoc:
Related to https://github.com/flix/flix/issues/10921

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
@@ -695,7 +695,7 @@ object Visitor {
   }
 
   private def visitFormalParam(fparam: FormalParam)(implicit a: Acceptor, c: Consumer): Unit = {
-    val FormalParam(bnd, _, tpe, _, loc) = fparam
+    val FormalParam(bnd, tpe, _, loc) = fparam
     if (!a.accept(loc)) {
       return
     }
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/FindReferencesProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/FindReferencesProvider.scala
@@ -153,7 +153,7 @@ object FindReferencesProvider {
     case TypedAst.RestrictableCase(_, _, _, loc) => loc.isReal
     case TypedAst.Constraint(_, _, _, loc) => loc.isReal
     case TypedAst.ConstraintParam(_, _, loc) => loc.isReal
-    case TypedAst.FormalParam(_, _, _, _, loc) => loc.isReal
+    case TypedAst.FormalParam(_, _, _, loc) => loc.isReal
     case TypedAst.PredicateParam(_, _, loc) => loc.isReal
     case TypedAst.JvmMethod(_, _, _, _, _, loc) => loc.isReal
     case TypedAst.CatchRule(_, _, _, _) => true
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/GotoProvider.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/GotoProvider.scala
@@ -154,7 +154,7 @@ object GotoProvider {
     case TypedAst.RestrictableCase(_, _, _, loc) => loc.isReal
     case TypedAst.Constraint(_, _, _, loc) => loc.isReal
     case TypedAst.ConstraintParam(_, _, loc) => loc.isReal
-    case TypedAst.FormalParam(_, _, _, _, loc) => loc.isReal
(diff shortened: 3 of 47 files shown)


magnus-madsen:
Very cool!

magnus-madsen:
How certain do we feel about the parsing when the LHS is a complex expression, e.g. `new Foo().x=123` or `foo().x = 123` or whatever?

jaschdoc:
I will add tests

jaschdoc:
Seems like we can support this for now and generalize it later

magnus-madsen:
What is the TLDR of the parse issue?

jaschdoc:
We can parse `expr.field = expr` but not `expr = expr` where the left hand side evaluates to a Java field reference. It _might_ be more complicated in terms of types and code gen but I'm not totally sure how type inference for the Java parts works so I would have to study that a bit more to give a better answer.

main/test/flix/Test.Exp.Jvm.PutField.flix
```diff
@@ -0,0 +1,42 @@
...
+        new Point(42, 43).x=1
+
+    @Test
+    def testPutField06(): Unit \ IO =
+        { rp = { p = new Point(42, 43) } }#rp#p.y=100
+
+    // TODO: Requires: `exp = exp` parsing, constraints and code gen. Current impl uses `expr.field = expr`.
+    // @Test
+    // def testPutField07(): Unit \ IO =
+    //     (new Point(42, 43).x) = 1
```
magnus-madsen, line 49: What about (new Point(42, 43)).x = 1?

jaschdoc:
Added test

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -1412,11 +1412,18 @@ object Parser2 {
...
             nameUnqualified(Set(TokenKind.NameLowerCase))
-            // `exp.f` is a Java field lookup and `exp.f(..)` is a Java method invocation.
+            // Check if method invocation, field assignment or field read.
             if (at(TokenKind.ParenL)) {
+              // expr.method()
               arguments()
               lhs = close(mark, TreeKind.Expr.InvokeMethod)
-            } else {
+            } else if (eat(TokenKind.Equal)) {
+              // expr.field = expr
```
JonathanStarup, line 1421: Hmm is it better to treat `=` as an binary operator instead?

JonathanStarup: This solution doesn't allow `obj.x = helper()` right?

main/test/flix/Test.Exp.Jvm.PutField.flix
```diff
@@ -0,0 +1,46 @@
...
+        (new Point(42, 43)).x=1
+
+    @Test
+    def testPutField07(): Unit \ IO =
+        { rp = { p = new Point(42, 43) } }#rp#p.y=100
+
+    // TODO: Requires: `exp = exp` parsing, constraints and code gen. Current impl uses `expr.field = expr`.
+    // @Test
+    // def testPutField08(): Unit \ IO =
+    //     (new Point(42, 43).x) = 1
```
JonathanStarup, line 49: Huh I thought this would work without parens since `new Point(42, 43)` is a delimited expr?

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
JonathanStarup: Can't you just write expression instead of expressionDelimited as the rhs call?

magnus-madsen:
The reason I am holding of on this is that I don't want to extend the language with the "wrong" grammar. We should figure out something that works, or at least the trade-offs. But this is not a priority right now compared to the package manager.

main/test/flix/Test.Exp.Jvm.PutField.flix
jaschdoc: these tests work
```
    def testPutField05(): Unit \ IO =
        new Point(42, 43).x=1

    def testPutField06(): Unit \ IO =
        (new Point(42, 43)).x=1
```

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
JonathanStarup: There is also the question of static field writes (`System.out = null`) could be nice if they were supported somewhat uniformly

jaschdoc: True yeah maybe this requires a bit more thought.

magnus-madsen:
Closed?

jaschdoc:
Yeah it seems like we want a more robust solution that also handles static fields. At least we have this tracked in the issue now.

PR #11447, opened 2025-08-28
feat: add provenance suppression to Datalog engine

CasperDN:
Proposal for #11264.

A not-messed-up version of #11356.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
@@ -1167,12 +1167,12 @@ object Lowering {
     * Lowers the given body predicate `p0`.
     */
   private def visitBodyPred(cparams0: List[TypedAst.ConstraintParam], p0: TypedAst.Predicate.Body)(implicit scope: Scope, root: TypedAst.Root, flix: Flix): LoweredAst.Expr = p0 match {
-    case TypedAst.Predicate.Body.Atom(pred, den, polarity, fixity, terms, _, loc) =>
+    case TypedAst.Predicate.Body.Atom(pred, den, polarity, fixity, provenance, terms, _, loc) =>
       val predSymExp = mkPredSym(pred)
       val denotationExp = mkDenotation(den, terms.lastOption.map(_.tpe), loc)
       val polarityExp = mkPolarity(polarity, loc)
       val fixityExp = mkFixity(fixity, loc)
-      val provenanceExp = mkProvenance(loc)
+      val provenanceExp = mkProvenance(provenance, loc)
       val termsExp = mkVector(terms.map(visitBodyTerm(cparams0, _)), Types.BodyTerm, loc)
       val innerExp = List(predSymExp, denotationExp, polarityExp, fixityExp, provenanceExp, termsExp)
       mkTag(Enums.BodyPredicate, "BodyAtom", innerExp, Types.BodyPredicate, loc)
@@ -1361,8 +1361,11 @@ object Lowering {
   /**
     * Constructs a `Fixpoint/Ast/Datalog.Provenance`.
     */
-  private def mkProvenance(loc: SourceLocation): LoweredAst.Expr =
-    mkTag(Enums.Provenance, "Enabled", Nil, Types.Provenance, loc)
+  private def mkProvenance(provenance: Provenance, loc: SourceLocation): LoweredAst.Expr = provenance match {
+    case Provenance.Enabled  => mkTag(Enums.Provenance, "Enabled",  Nil, Types.Provenance, loc)
+    case Provenance.Disabled => mkTag(Enums.Provenance, "Disabled", Nil, Types.Provenance, loc)
+  }
+    
 
   /**
     * Constructs a `Fixpoint/Ast/Shared.PredSym` from the given predicate `pred`.
--- a/main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/Visitor.scala
@@ -857,7 +857,7 @@ object Visitor {
 
     p match {
       case Predicate.Head.Atom(_, _, terms, _, _) => terms.foreach(visitExpr)
-      case Predicate.Body.Atom(_, _, _, _, terms, _, _) => terms.foreach(visitPattern)
+      case Predicate.Body.Atom(_, _, _, _, _, terms, _, _) => terms.foreach(visitPattern)
       case Predicate.Body.Functional(_, exp, _) => visitExpr(exp)
       case Predicate.Body.Guard(exp, _) => visitExpr(exp)
     }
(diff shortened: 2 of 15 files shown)


main/src/library/Fixpoint3/Ast/Datalog.flix
```diff
@@ -345,6 +349,16 @@ mod Fixpoint3.Ast.Datalog {
         case Fixed
     }
 
+    /////////////////////////////////////////////////////////////////////////////
+    // ProvSuppressed                                                          //
+    /////////////////////////////////////////////////////////////////////////////
+
+    @Internal
+    pub enum ProvSuppressed {
```
magnus-madsen: I would like to rename this enum to:

```
enum Provenance {
    case Enabled
    case Disabled
}
```

Can you also add it to `ast.shared` and `LoweredAst`?

main/src/ca/uwaterloo/flix/language/phase/Lowering.scala
```diff
@@ -120,6 +121,7 @@ object Lowering {
     lazy val Denotation: Type = Type.mkEnum(Enums.Denotation, Boxed :: Nil, SourceLocation.Unknown)
     lazy val Polarity: Type = Type.mkEnum(Enums.Polarity, Nil, SourceLocation.Unknown)
     lazy val Fixity: Type = Type.mkEnum(Enums.Fixity, Nil, SourceLocation.Unknown)
+    lazy val ProvSuppressed: Type = Type.mkEnum(Enums.ProvSuppressed, Nil, SourceLocation.Unknown)
```
magnus-madsen: This should just be Provenance (see below)

main/src/library/Fixpoint3/Phase/RenamePredSyms.flix
```diff
@@ -145,7 +145,7 @@ mod Fixpoint3.Phase.RenamePredSyms {
     /// Maps `PredSym`s of `body` according to `transformer`.
     ///
     def remapBody(transformer: PredSym -> PredSym \ r, body: BodyPredicate): BodyPredicate \ r = match body {
-        case BodyAtom(pred, den, pol, fixed, terms) => BodyAtom(transformer(pred), den, pol, fixed, terms)
+        case BodyAtom(pred, den, pol, fixed, suppressed, terms) => BodyAtom(transformer(pred), den, pol, fixed, suppressed, terms)
```
magnus-madsen: suppressed -> provenance (everywhere)

main/src/library/Fixpoint3/Provenance.flix
```diff
@@ -257,14 +257,20 @@ mod Fixpoint3.Provenance {
...
-                            case BodyPredicate.BodyAtom(pred, _, _, _, _) => Some(pred)
+                            case BodyPredicate.BodyAtom(pred, _, _, _, sup, _) => Some((pred, sup))
                             case _ => None
                         });
                     let ref = Ref.fresh(rc, tail);
-                    let childProofs = Array.empty(rc, Array.length(bodyAtoms));
-                    foreach((i, atom) <- ForEach.withIndex(relevantPreds)) {
-                        let pred = Array.get(i, bodyAtoms);
-                        Ref.put((i, childProofs, atom, pred) :: Ref.get(ref), ref)
+                    let suppressedAtomCount = Vector.count(v -> match v {
```
magnus-madsen: Add some documentation to explain the intuition.

main/src/library/Fixpoint3/Ast/Datalog.flix
magnus-madsen: (In LoweredAst you can just set it to Enabled)

magnus-madsen (commented):
Added some comments. (Sorry for the delay).

main/src/library/Fixpoint3/Ast/Datalog.flix
CasperDN: > Can you also add it to `ast.shared` and `LoweredAst`?

I am unsure what you want me to do here. `LoweredAst` does not contain `FixpointX`/`Constraint` so it will not need the enum/class in Scala.

I could add it to `ast.shared` and create the `TypedAst` with the `Enabled` as standard and then implement the logic in `Lowering` to map to the correct Datalog-enum. Is this what you meant?

CasperDN:
With the new naming we have an `Enum`, and two modules called `Provenance`. Should we aim to rename some of these? The reconstruction module could be something like `ReconstructProvenance` and the phase could be `ProvenanceAugment` or something like that?

main/src/library/Fixpoint3/Ast/Datalog.flix
magnus-madsen: Yes sorry that is what i meant.

magnus-madsen:
> With the new naming we have an `Enum`, and two modules called `Provenance`. Should we aim to rename some of these? The reconstruction module could be something like `ReconstructProvenance` and the phase could be `ProvenanceAugment` or something like that?

Oh that is bad. What is the "other" provenance enum? What is its definition?

Maybe we need something like `PredicateProvenance`etc?

CasperDN:
Sorry if I was unclear. We do not have other enums called `Provenance`, but we have 2 modules: The reconstruction and the augmentation.

main/src/library/Fixpoint3/Ast/Datalog.flix
CasperDN: Done. `TypeReconstruction` now just passes `Enabled`, and the rest works as it should. I mean I cannot test it, but it should work.

CasperDN:
@magnus-madsen Is it on purpose that `zeroOrMore` cannot parse `{ | {} }` as there must be at least one element before the `optionallyWith` can be parsed? If so we can remove a check I introduced in the weeder.

magnus-madsen:
> @magnus-madsen Is it on purpose that `zeroOrMore` cannot parse `{ | {} }` as there must be at least one element before the `optionallyWith` can be parsed? If so we can remove a check I introduced in the weeder.

This seems unrelated to this PR?

magnus-madsen:
> Sorry if I was unclear. We do not have other enums called `Provenance`, but we have 2 modules: The reconstruction and the augmentation.

So whats the clash? Between the data type and a phase or what?

CasperDN:
> > @magnus-madsen Is it on purpose that `zeroOrMore` cannot parse `{ | {} }` as there must be at least one element before the `optionallyWith` can be parsed? If so we can remove a check I introduced in the weeder.
> 
> 
> 
> This seems unrelated to this PR?

Should have been in the fix fuel PR.

CasperDN:
> > Sorry if I was unclear. We do not have other enums called `Provenance`, but we have 2 modules: The reconstruction and the augmentation.
> 
> 
> 
> So whats the clash? Between the data type and a phase or what?

There is no direct clash, but calling 3 separate things the same could create confusion.

magnus-madsen:
Do you have some better ideas for names? Give me some suggestions :)

CasperDN:
Could be as simple as naming the modules `ProvenanceReconstruct` and `ProvenanceAugment`.

magnus-madsen:
> Could be as simple as naming the modules `ProvenanceReconstruct` and `ProvenanceAugment`.

These are module names, so a collection of functions?

Meanwhile Provenance would still be the enum?

CasperDN:
> > Could be as simple as naming the modules `ProvenanceReconstruct` and `ProvenanceAugment`.
> 
> 
> 
> These are module names, so a collection of functions?
> 
> 
> 
> Meanwhile Provenance would still be the enum?

Yes

magnus-madsen:
Sounds good. Want to give it a go?

CasperDN:
If you want me to make a separate PR for it I would like to finish/update #11381 first. It will probably first be done tomorrow.

magnus-madsen:
Sounds good.

magnus-madsen:
We leave as future work.

PR #11448, opened 2025-08-29
refactor: remove CPS from right-to-left List folds

CasperDN:
Fixes the right-to-left CPS functions from #11399

### DIFF
--- a/main/src/library/List.flix
+++ b/main/src/library/List.flix
@@ -346,13 +346,13 @@ mod List {
     /// That is, the result is of the form: `... f(xn-1, f(xn, s)) :: f(xn, s) :: s`.
     ///
     pub def scanRight(f: (a, b) -> b \ ef, s: b, l: List[a]): List[b] \ ef =
-        def loop(ll, k) = match ll {
-            case Nil     => k(s :: Nil)
-            case x :: xs => loop(xs, ks -> match ks {
-                case ss :: _ => k(f(x, ss) :: ks)
-                case _       => unreachable!()})
+        def loop(ll, accRes, acc) = match ll {
+            case Nil     => accRes
+            case x :: xs =>
+                let y = f(x, acc);
+                loop(xs, y :: accRes, y)
         };
-        loop(l, x -> checked_ecast(x))
+        loop(reverse(l), s :: Nil, s)
 
     ///
     /// Returns the result of applying `f` to every element in `l`.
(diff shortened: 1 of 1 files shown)


main/src/library/List.flix
```diff
@@ -749,14 +749,14 @@ mod List {
...
-            case x :: xs => loop(xs, ks -> k(match ks {
-                case None    => Some(x)
-                case Some(v) => Some(f(x, v))
-            }))
+        def loop(ll, acc) = match ll {
+            case Nil     => Some(acc)
+            case x :: xs => loop(xs, f(x, acc))
         };
-        loop(l, x -> checked_ecast(x))
+        match reverse(l) {
```
magnus-madsen, line 756: Why this match? Is it not redundant?

main/src/library/List.flix
```diff
@@ -346,13 +346,13 @@ mod List {
     /// That is, the result is of the form: `... f(xn-1, f(xn, s)) :: f(xn, s) :: s`.
     ///
     pub def scanRight(f: (a, b) -> b \ ef, s: b, l: List[a]): List[b] \ ef =
-        def loop(ll, k) = match ll {
-            case Nil     => k(s :: Nil)
-            case x :: xs => loop(xs, ks -> match ks {
-                case ss :: _ => k(f(x, ss) :: ks)
-                case _       => unreachable!()})
+        def loop(ll, accRes, acc) = match ll {
```
magnus-madsen: I would move accRes as the last argument.

magnus-madsen: Perhaps the names: bacc and acc would be better

main/src/library/List.flix
```diff
@@ -712,11 +712,11 @@ mod List {
     /// That is, the result is of the form: `f(x1, ...f(xn-1, f(xn, s))...)`.
     ///
     pub def foldRight(f: (a, b) -> b \ ef, s: b, l: List[a]): b \ ef =
```
magnus-madsen, line 714: reverse followed by foldLeft?

main/src/library/List.flix
```diff
@@ -749,14 +749,14 @@ mod List {
     /// Returns `None` if `l` is empty.
     ///
     pub def reduceRight(f: (a, a) -> a \ ef, l: List[a]): Option[a] \ ef =
```
magnus-madsen, line 751: reverse followed by reduceLeft?

main/src/library/List.flix
CasperDN: argument order of `f` is reversed for `foldLeft`. Could wrap in `flip`, but the purpose of the PR has mostly been getting rid of CPS, not rewriting `List`.

main/src/library/List.flix
CasperDN: Same as `foldRight`.

main/src/library/List.flix
CasperDN: We need to start with the last element? As in we need to find an `acc`.

PR #11449, opened 2025-08-29
refactor: remove CPS from simple left-to-right List folds

CasperDN:
Fixes the simple right-to-left CPS functions from #11399

Simple here means that the loop is (mostly) of the form: `loop(restOfList, accumulatedResultToBeReversed) = ???`


magnus-madsen:
Thanks!

PR #11451, opened 2025-08-29
feat: wrap tests in default Assert handler for :test 

NeonOxide:
Added logic to wrap tests in a default Assert handler for the :test command. Right now it is using Assert.runWithIO but this can be easily changed in the future. This also opens the possibility of adding more default handlers for other effects following the same method.

Related to #11323

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+Version 0.61.1:
+- Compiler: Improve type error messages (thanks Magnus!)
+
 Version 0.61.0:
 - Compiler: Reworked `foreach` to use the `ForEach` trait (thanks Ry!)
 - Library: Added completely new Datalog engine with much improved performance (thanks Adam and Casper!)
--- a/main/src/ca/uwaterloo/flix/api/Version.scala
+++ b/main/src/ca/uwaterloo/flix/api/Version.scala
@@ -20,7 +20,7 @@ object Version {
   /**
     * Represents the current version of Flix.
     */
-  val CurrentVersion: Version = Version(major = 0, minor = 61, revision = 0)
+  val CurrentVersion: Version = Version(major = 0, minor = 61, revision = 1)
 }
 
 /**
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -2872,13 +2872,14 @@ object Parser2 {
         getItem = fixpointConstraint,
         delimiterL = TokenKind.HashCurlyL,
         delimiterR = TokenKind.CurlyR,
-        separation = Separation.Required(TokenKind.DotWhiteSpace, allowTrailing = true),
+        separation = Separation.None,
         breakWhen = _.isRecoverExpr,
       )
       close(mark, TreeKind.Expr.FixpointConstraintSet)
     }
 
     private def fixpointConstraint()(implicit s: State): Mark.Closed = {
+      implicit val sctx: SyntacticContext = SyntacticContext.Expr.Constraint
       val mark = open()
       Predicate.head()
       if (eat(TokenKind.ColonMinus)) {
@@ -2887,6 +2888,7 @@ object Parser2 {
           Predicate.body()
         }
       }
+      expect(TokenKind.DotWhiteSpace, hint = Some("Datalog constraints must end with a '.'"))
       close(mark, TreeKind.Expr.FixpointConstraint)
     }
 
--- a/main/test/ca/uwaterloo/flix/language/phase/TestParser.scala
+++ b/main/test/ca/uwaterloo/flix/language/phase/TestParser.scala
@@ -792,6 +792,33 @@ class TestParserRecovery extends AnyFunSuite with TestUtils {
     expectMain(result)
   }
 
+  test("MissingDotInDatalogConstraint.01") {
+    val input =
+      """def main(): Unit =
+        |    let _ = #{
+        |        Edge(1, 2)
+        |    };
+        |    ()
+        |"""".stripMargin
+    val result = check(input, Options.TestWithLibMin)
+    expectErrorOnCheck[ParseError](result)
+    expectMain(result)
+  }
+
+  test("MissingDotInDatalogConstraint.02") {
+    val input =
+      """def main(): Unit =
+        |    let _ = #{
+        |        Edge(1, 2).
+        |        Path(x, y) :- Edge(x, y)
+        |    };
+        |    ()
+        |"""".stripMargin
+    val result = check(input, Options.TestWithLibMin)
+    expectErrorOnCheck[ParseError](result)
+    expectMain(result)
+  }
+
   test("MissingWithInPQuery") {
     val input =
       """def main(): Unit =
--- a/main/test/flix/Test.Exp.Fixpoint.Application.flix
+++ b/main/test/flix/Test.Exp.Fixpoint.Application.flix
@@ -38,7 +38,7 @@ mod Test.Exp.Fixpoint.Application {
         let n = 50;
         let r = #{
             Succ(0).
-            Succ(x + 1) :- Succ(x), if(x < n)
+            Succ(x + 1) :- Succ(x), if(x < n).
         };
         let expected = Vector.range(0, n + 1);
         let actual = query r select x from Succ(x);
@@ -117,7 +117,7 @@ mod Test.Exp.Fixpoint.Application {
         let r = #{
             Two(2).
             SuccTwo(0).
-            SuccTwo(x + y) :- SuccTwo(x), Two(y), if(x < n)
+            SuccTwo(x + y) :- SuccTwo(x), Two(y), if(x < n).
         };
         let expected = Vector#{0, 2, 4, 6, 8, 10, 12, 14, 16, 18, 20};
         let actual = query r select x from SuccTwo(x);
--- a/main/test/flix/Test.Exp.Fixpoint.If.flix
+++ b/main/test/flix/Test.Exp.Fixpoint.If.flix
@@ -276,7 +276,7 @@ mod Test.Exp.Fixpoint.If {
         let f = inject tuples into Edge/2;
         let r = #{
             Path(x, y; Down.Down(1)) :- Edge(x, y).
-            Path(x, z; l + Down.Down(1)) :- Path(x, y; l), Edge(y, z), if (l == Down.Down(0))
+            Path(x, z; l + Down.Down(1)) :- Path(x, y; l), Edge(y, z), if (l == Down.Down(0)).
         };
         let res = query f, r select (x, y) from Path(x, y; l);
         res `Assert.eq` tuples
--- a/main/test/flix/Test.Exp.Fixpoint.Solve.Lattice.flix
+++ b/main/test/flix/Test.Exp.Fixpoint.Solve.Lattice.flix
@@ -263,7 +263,7 @@ mod Test.Exp.Fixpoint.Solve.Lattice {
             A(1).
             A(2).
             B(;x) :- A(x).
-            R(x) :- fix B(;x)
+            R(x) :- fix B(;x).
         };
         let mm = query pr select x from R(x);
         Vector#{2} `Assert.eq` mm
@@ -274,7 +274,7 @@ mod Test.Exp.Fixpoint.Solve.Lattice {
             A(1).
             A(2).
             B(;x) :- A(x).
-            R(x) :- fix B(;x)
+            R(x) :- fix B(;x).
         };
         let mm = query pr select x from R(x);
         Vector#{2} `Assert.eq` mm


magnus-madsen:
!!!jar

main/src/ca/uwaterloo/flix/language/ast/Symbol.scala
```diff
@@ -40,6 +40,7 @@ object Symbol {
   val Net: EffSym = mkEffSym(Name.RootNS, Ident("Net", SourceLocation.Unknown))
   val NonDet: EffSym = mkEffSym(Name.RootNS, Ident("NonDet", SourceLocation.Unknown))
   val Sys: EffSym = mkEffSym(Name.RootNS, Ident("Sys", SourceLocation.Unknown))
+  val Assert: EffSym = mkEffSym(Name.RootNS, Ident("Assert", SourceLocation.Unknown))
```
magnus-madsen: Keep sorted

main/src/ca/uwaterloo/flix/language/ast/Type.scala
```diff
@@ -399,6 +399,11 @@ object Type {
     */
   val Sys: Type = Type.Cst(TypeConstructor.Effect(Symbol.Sys, Kind.Eff), SourceLocation.Unknown)
 
+  /**
+    * Represents the Assert effect.
+    */
+  val Assert: Type = Type.Cst(TypeConstructor.Effect(Symbol.Assert, Kind.Eff), SourceLocation.Unknown)
```
magnus-madsen: Keep sorted

main/src/ca/uwaterloo/flix/language/phase/EntryPoints.scala
```diff
@@ -57,7 +58,8 @@ object EntryPoints {
```
magnus-madsen, line 58: Reformat this file using Intellij IDEA, please- thanks :)

main/src/ca/uwaterloo/flix/language/phase/EntryPoints.scala
```diff
@@ -57,7 +58,8 @@ object EntryPoints {
 
   def run(root: TypedAst.Root)(implicit flix: Flix): (TypedAst.Root, List[EntryPointError]) = flix.phaseNew("EntryPoints") {
     val (root1, errs1) = resolveMain(root)
-    val (root2, errs2) = CheckEntryPoints.run(root1)
+    val root_with_wrapped_tests = WrapTest.run(root1)
```
magnus-madsen: Scala is camelCase, so rootWithWrappedTests

main/src/ca/uwaterloo/flix/language/phase/EntryPoints.scala
magnus-madsen: In general, improve the formatting.

main/src/ca/uwaterloo/flix/language/phase/EntryPoints.scala
```diff
@@ -597,6 +599,65 @@ object EntryPoints {
...
+      * mkTest takes a test function and creates a new test function with a
+      * default Assert handler by using Assert.runWithIO.
+      *
+      * Takes def oldTest(arg1:t1, ...) : ret \ ef = exp
+      * Returns def oldTest(arg1:t1, ...) : ret \ IO + Sys + (ef - Assert) = Assert.runWithIO(_ -> exp)
+      */
+    private def mkTest(oldTest: TypedAst.Def, root: TypedAst.Root)(implicit flix: Flix): TypedAst.Def = {
+      // The new type is the same as the wrapped test with an effect set of
+      // `IO + Sys + (eff - Assert)` where `eff` is the effect set of the old test.
+      val eff = Type.mkUnion(Type.IO, Type.Sys, Type.mkDifference(oldTest.spec.eff, Type.Assert, SourceLocation.Unknown)
```
magnus-madsen: Using SourceLocation.Unknown should be avoided as much as possible. Better to reuse a source location, e.g. of the current effect.

magnus-madsen: This comment applies everywhere you use Unknown.

main/src/ca/uwaterloo/flix/language/phase/EntryPoints.scala
```diff
@@ -597,6 +599,65 @@ object EntryPoints {
...
+                            , SourceLocation.Unknown)
+      val tpe = Type.mkCurriedArrowWithEffect(oldTest.spec.fparams.map(_.tpe), eff, oldTest.spec.retTpe, SourceLocation.Unknown)
+      val spec = oldTest.spec.copy(
+        declaredScheme = oldTest.spec.declaredScheme.copy(base=tpe),
+        eff = eff,
+      )
+      // Get Assert.runWithIO symbol
+      val runWithIOSym = DefSymUse(root.defs(new Symbol.DefnSym(namespace=List("Assert"), text="runWithIO", loc=SourceLocation.Unknown, id=None)).sym, SourceLocation.Unknown)
```
magnus-madsen: introduce a local variable for the DefnSym.

main/src/ca/uwaterloo/flix/language/phase/EntryPoints.scala
```diff
@@ -597,6 +599,65 @@ object EntryPoints {
...
+      // Create _ -> exp
+      val innerLambda = TypedAst.Expr.Lambda(fparam=TypedAst.FormalParam(
+                                                          bnd = TypedAst.Binder(sym = Symbol.freshVarSym("_", FormalParam, SourceLocation.Unknown), tpe = Type.Unit),
+                                                          mod = Modifiers(Nil),
+                                                          tpe = Type.Unit,
+                                                          src = TypeSource.Inferred,
+                                                          loc=SourceLocation.Unknown
```
magnus-madsen: We typically don't use labelled arguments.

magnus-madsen: (This comment applies everywhere)

main/src/ca/uwaterloo/flix/language/phase/EntryPoints.scala
```diff
@@ -57,7 +58,8 @@ object EntryPoints {
 
   def run(root: TypedAst.Root)(implicit flix: Flix): (TypedAst.Root, List[EntryPointError]) = flix.phaseNew("EntryPoints") {
     val (root1, errs1) = resolveMain(root)
-    val (root2, errs2) = CheckEntryPoints.run(root1)
+    val root_with_wrapped_tests = WrapTest.run(root1)
+    val (root2, errs2) = CheckEntryPoints.run(root_with_wrapped_tests)
```
magnus-madsen: What's reasoning for doing the checks after transforming the AST?

magnus-madsen:
Do you have some examples of what works and what does not?

If I try:

```
def test04(): Unit \ Assert = 
    Assert.assertEq(expected = 42, 42)
```

I get a type error:

```
Type One: Chan + Env + Sys + NonDet + Net + IO + FsWrite + FsRead + Exec
Type Two: (Assert + IO) ⊕ ((~(Assert + IO)) & e0)
```

magnus-madsen:
I note that this:

```
@Test
def test04(): Unit \ Assert = 
    Assert.assertEq(expected = 42, 42);
    Assert.assertEq(expected = 42, 1);
    Assert.assertEq(expected = 42, 2)
```

when run with `:test` produces the right output, but ALSO:

```

Passed: 4, Failed: 1. Skipped: 0. Elapsed: 1,8ms.
Tester Error <---- what is this?
```

magnus-madsen:
Further comments: 

- We probably want some tests
- I also see some CI test failures

NeonOxide:
> Do you have some examples of what works and what does not?
> 
> If I try:
> 
> ```
> def test04(): Unit \ Assert = 
>     Assert.assertEq(expected = 42, 42)
> ```
> 
> I get a type error:
> 
> ```
> Type One: Chan + Env + Sys + NonDet + Net + IO + FsWrite + FsRead + Exec
> Type Two: (Assert + IO) ⊕ ((~(Assert + IO)) & e0)
> ```

This is what I discussed in the Element chat I think. Right now the default handler is only added when called with :test as I thought that was the most logical way of doing it. I could also add it to eval but I think that that is treating a function specially, as far as I know calling eval should call the function as it is, not do any transformations but I can add the logic to eval too.

NeonOxide:
> I note that this:
> 
> ```
> @Test
> def test04(): Unit \ Assert = 
>     Assert.assertEq(expected = 42, 42);
>     Assert.assertEq(expected = 42, 1);
>     Assert.assertEq(expected = 42, 2)
> ```
> 
> when run with `:test` produces the right output, but ALSO:
> 
> ```
> 
> Passed: 4, Failed: 1. Skipped: 0. Elapsed: 1,8ms.
> Tester Error <---- what is this?
> ```

You are right. I was looking at the result of the tests but I did not notice that extra line at the end, if I had to say it is due to throwing exceptions in the handler (as it does not appear in tests that pass). I will look further into it

main/src/ca/uwaterloo/flix/language/phase/EntryPoints.scala
NeonOxide: The reasoning is that the checks for example for main and test check whether the entry points only have primitive effects. By doing the transformations before, we do not need to special case Assert in the checks for entry points, by doing checkPrimitiveEffectsOrAssert. Which is not the end of the world, but I found it easier to do. The visitTest only checks that there are no type variables, no unit argument and no other primitive effects, and all of these invariants are kept after the transformation. However it is true that some errors may be more confusing this way. Do you think I should just special case the Assert effect and perform the transformations after checking?

main/src/ca/uwaterloo/flix/language/phase/EntryPoints.scala
NeonOxide: Okay. Do I just use the default Reformat code of Intellij or do you guys use any other additional settings?

main/src/ca/uwaterloo/flix/language/ast/Symbol.scala
NeonOxide: Will do

main/src/ca/uwaterloo/flix/language/ast/Type.scala
NeonOxide: Will also do

magnus-madsen:
> This is what I discussed in the Element chat I think. Right now the default handler is only added when called with :test as I thought that was the most logical way of doing it. I could also add it to eval but I think that that is treating a function specially, as far as I know calling eval should call the function as it is, not do any transformations but I can add the logic to eval too.

Assuming both buttons in VSCode should work, how would you design that?

NeonOxide:
> > I note that this:
> > ```
> > @Test
> > def test04(): Unit \ Assert = 
> >     Assert.assertEq(expected = 42, 42);
> >     Assert.assertEq(expected = 42, 1);
> >     Assert.assertEq(expected = 42, 2)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > when run with `:test` produces the right output, but ALSO:
> > ```
> > 
> > Passed: 4, Failed: 1. Skipped: 0. Elapsed: 1,8ms.
> > Tester Error <---- what is this?
> > ```
> 
> You are right. I was looking at the result of the tests but I did not notice that extra line at the end, if I had to say it is due to throwing exceptions in the handler (as it does not appear in tests that pass). I will look further into it

I have looked into it and it also happens in the current build of Flix with Assert.eq so it is more related to either not throwing internal exceptions (which I think is the way we should go but it requires more thought as there are some non trivial choices to be made) or making the tester catch them and not fail

main/src/ca/uwaterloo/flix/language/phase/EntryPoints.scala
NeonOxide: I swapped the order in the end as it gives better error messages and is more consistent with the pipeline

NeonOxide:
I corrected most of the issues given (the other ones are due to me thinking they belong to another PR or need more discussion). The description of all the changes is in the commit itself. It passed the AllTest task on my machine, but the vscode task that failed before is not a Gradle task as far as I know and I have not been able to test it locally. If there is a way to do it please let me know.

Thanks in advance for your time!

magnus-madsen:
> I have looked into it and it also happens in the current build of Flix with Assert.eq so it is more related to either not throwing internal exceptions (which I think is the way we should go but it requires more thought as there are some non trivial choices to be made) or making the tester catch them and not fail

I see. Hmm. Another reason we need to rework things.

magnus-madsen:
!!!jar

magnus-madsen:
You are now using real source locations-- that is good. But you need to mark them as synthetic, so they are treated appropriately by VSCode:

CleanShot 2025-08-30 at 13 46 38@2x

magnus-madsen:
Looks like we have approximately the right semantics. I will look over the code soon.

main/src/ca/uwaterloo/flix/language/ast/Symbol.scala
```diff
@@ -48,6 +53,13 @@ object Symbol {
     Chan, Env, Exec, FsRead, FsWrite, IO, Net, NonDet, Sys
   ))
 
+  /**
+    * The set of effects allowed in tests.
+    */
+  val TestEffs: SortedSet[EffSym] = SortedSet.from(List(
+    Chan, Env, Exec, FsRead, FsWrite, IO, Net, NonDet, Sys, Assert
```
magnus-madsen: Would it be better to define this as:

PrimitiveEffs + Assert

? Why or why not?

main/src/ca/uwaterloo/flix/language/phase/EntryPoints.scala
```diff
@@ -367,6 +371,33 @@ object EntryPoints {
...
+          None
+        case Result.Ok(false) =>
+          Some(EntryPointError.IllegalEntryPointEffect(eff, eff.loc))
+        case Result.Err(ErrorOrMalformed) =>
+          // Do not report an error, since previous phases should have done already.
+          None
+      }
+    }
+
+    private def isTestEffect(eff: Type): Result[Boolean, ErrorOrMalformed.type] = {
```
magnus-madsen, line 401: Add doc comment. (You can copy from above and use the same style).

main/src/ca/uwaterloo/flix/language/phase/EntryPoints.scala
```diff
@@ -367,6 +371,33 @@ object EntryPoints {
...
+      * Returns `None` if `defn` has an effect that is a subset of the set of allowed Test effects
+      * (e.g. `IO + Console + Assert`). Returns an error otherwise.
+      */
+    private def checkTestEffects(defn: TypedAst.Def)(implicit flix: Flix): Option[EntryPointError] = {
+      val eff = defn.spec.eff
+      isTestEffect(eff) match {
+        case Result.Ok(true) =>
+          None
+        case Result.Ok(false) =>
+          Some(EntryPointError.IllegalEntryPointEffect(eff, eff.loc))
```
magnus-madsen, line 389: I think might need a new custom error. Check this program and the error message produced:

```
@Test
pub def foo(): Unit \ {Abort, Assert} = 
    Assert.assertEq(expected = "abc", "def");
    Abort.abort("fds")
```

I find that very confusing.

main/src/ca/uwaterloo/flix/language/phase/EntryPoints.scala
```diff
@@ -597,6 +628,68 @@ object EntryPoints {
...
   }
 
+  private object WrapTest {
+
+
+    /**
+      * WrapTest takes all tests and wraps them in a default handler for Assert
+      *
+      * Then, it returns a new root with these new tests
+      *
```
magnus-madsen, line 651: The documentation comment is unclear. Can we do better? :)

main/src/ca/uwaterloo/flix/language/phase/EntryPoints.scala
```diff
@@ -597,6 +628,68 @@ object EntryPoints {
...
+      */
+    def run(root: TypedAst.Root)(implicit flix: Flix): TypedAst.Root = {
+      val updatedDefs = ParOps.parMapValues(root.defs)(defn =>
```
magnus-madsen: -> wrappedDefs?

-> defsWithAssertHandler?

main/src/ca/uwaterloo/flix/language/phase/EntryPoints.scala
```diff
@@ -597,6 +628,68 @@ object EntryPoints {
...
+        if (isTest(defn))
```
magnus-madsen: It is OK to put such a short if-then-else on a single line.

main/src/ca/uwaterloo/flix/language/phase/EntryPoints.scala
```diff
@@ -597,6 +628,68 @@ object EntryPoints {
...
+      root.copy(defs = updatedDefs)
+    }
+
+    /**
+      * mkTest takes a test function and creates a new test function with a
+      * default Assert handler by using Assert.runWithIO.
+      *
+      * Takes def oldTest(arg1:t1, ...) : ret \ ef = exp
+      * Returns def oldTest(arg1:t1, ...) : ret \ IO + Sys + (ef - Assert) = Assert.runWithIO(_ -> exp)
+      */
```
magnus-madsen, line 672: You can write the above much better. :-)

Use {{{ code }}} to show how code is transformed. 

Also be a bit more precise with names:

```
def f(arg1: tpe1, ...): tpe \ ef = exp 
```

We use tpe for Type. Not t or ret.

main/src/ca/uwaterloo/flix/language/phase/EntryPoints.scala
```diff
@@ -597,6 +628,68 @@ object EntryPoints {
...
+    private def mkTest(oldTest: TypedAst.Def, root: TypedAst.Root)(implicit flix: Flix): TypedAst.Def = {
```
magnus-madsen: mkTest -> wrapDefWithAssertHandler
oldTest -> currentDef

main/src/ca/uwaterloo/flix/language/phase/EntryPoints.scala
```diff
@@ -597,6 +628,68 @@ object EntryPoints {
...
+      // `IO + Sys + (eff - Assert)` where `eff` is the effect set of the old test.
+      val eff = Type.mkUnion(Type.IO, Type.Sys, Type.mkDifference(oldTest.spec.eff, Type.Assert, oldTest.spec.eff.loc)
+        , oldTest.spec.eff.loc)
+      val tpe = Type.mkCurriedArrowWithEffect(oldTest.spec.fparams.map(_.tpe), eff, oldTest.spec.retTpe, oldTest.spec.declaredScheme.base.loc)
+      val spec = oldTest.spec.copy(
+        declaredScheme = oldTest.spec.declaredScheme.copy(base = tpe),
+        eff = eff,
+      )
+      // Get Assert.runWithIO symbol
+      val runWithIOSym = new Symbol.DefnSym(None, List("Assert"), "runWithIO", SourceLocation.Unknown)
```
magnus-madsen: This symbol is a constant right? So could go somewhere in the top of the file?

main/src/ca/uwaterloo/flix/runtime/shell/Shell.scala
```diff
@@ -313,7 +313,7 @@ class Shell(bootstrap: Bootstrap, options: Options) {
         val src =
           s"""def ${main.name}(): Unit \\ $effString =
              |checked_ecast(
-             |  println($s)
+             |  Assert.runWithIO(_ -> println($s))
```
magnus-madsen, line 316: This actually makes Assert a default handler for the terminal. I suppose that is OK (??)

main/test/ca/uwaterloo/flix/language/phase/TestSafety.scala
```diff
@@ -729,7 +729,7 @@ class TestSafety extends AnyFunSuite with TestUtils {
         |@test
         |def f(x: Int32): Int32 = x
       """.stripMargin
-    val result = compile(input, Options.TestWithLibNix)
+    val result = compile(input, Options.TestWithLibAll)
```
magnus-madsen, line 732: This is because @Test not requires a definition of Assert right?

magnus-madsen (commented):
Added more detailed comments.

magnus-madsen:
I think one more pass over this PR to polish the code and fix the last few issues.

Then we can merge and start thinking about extensions, e.g. default handlers.

main/src/ca/uwaterloo/flix/runtime/shell/Shell.scala
NeonOxide: This is the easy way of making it work also for eval, I basically mimicked what is already done for Main entry point. In normal runs they are wrapped after checking as I did. But for eval of arbitrary functions that wrapping happens in this source block

main/src/ca/uwaterloo/flix/language/phase/EntryPoints.scala
NeonOxide: You are right, I think so too

main/src/ca/uwaterloo/flix/language/phase/EntryPoints.scala
NeonOxide: I did not know that notation existed for comments. It is much more better I agree

main/src/ca/uwaterloo/flix/language/phase/EntryPoints.scala
NeonOxide: This is the same error already present before these changes (eg. if you change Assert for IO in the vscode live version you will receive the same confusing message).

magnus-madsen: But the error message suggests that `Assert` is unhandled-- that now seems misleading?

NeonOxide: I have been looking at this and it seems non trivial to solve. You at the top level obtain an effect equation, how do you calculate a neat set of only the additional effects they have provided to the user given the current infrastructure? (it is the same problem with main currently on live)

NeonOxide:
```
pub def main(): Unit \ IO + Clock = checked_ecast(())
```
Gives the same problem. I agree this should be solved, but given its reaching implications I think it belongs in another PR

main/test/ca/uwaterloo/flix/language/phase/TestSafety.scala
NeonOxide: Yeah, given that we are wrapping @test with the default runWithIO handler it needs to have access to Assert. I considered moving Assert into LibMin, but then Assert requires Chalk, and Chalk requires Environment... So I thought it was better to just extend it to LibAll for all tests that require test.

main/src/ca/uwaterloo/flix/language/ast/Symbol.scala
NeonOxide: I think it is, I will change it like that. At least for now it makes sense, in the future we may want to disallow some primitive effects in tests, but then it is just as easy as coming here and just changing this one line, no other reaching changes

NeonOxide:
New commit. Same as last time, changes are in the commit. There are some unresolved comments that I think need either more thought or just belong in another PR.

main/src/ca/uwaterloo/flix/language/phase/EntryPoints.scala
magnus-madsen: We know the effect cannot be polymorphic right?

magnus-madsen:
Does this infrastructure play well with `FlixSuite`? What happens if you change one test -- in this PR -- in `TestList.flix` to use Assert?

NeonOxide:
I have done that and it works. The only thing that would cause problems is if the tests are currently performed with a limited set of the standard library. Another thing that we should consider in the future is the possibility of Assert returning a boolean in order to be more intuitive with Tests. We recommend people to use Bool in their tests, but if you use assert you have to change the return type of your functions to Unit. I don't know if the base assert should return a boolean (if we get polymorphic effects I think it should be able to return any value to allow for modification of values in tests). But the wrappers around it would be more convenient that way in my opinion. However, then it remains the choice of what boolean to return, I think returning false in case of failure and true otherwise is a good choice that mimics the current @Test choice

NeonOxide:
Following that same logic we could do a runWithBool that takes only f: Unit -> bool \ ef + Assert and returns bool with ef. That way throwing exceptions is not required anymore and we swiftly integrate with the Test framework. This new function should be the one used to wrap Tests, however the standard runWithIO should be kept for evaluating in my opinion, as to not change the semantics. Either that or check if the return type of the top level function is Bool and choose accordingly which runWith to use for wrapping the inner expression, which is not a bad idea either.

magnus-madsen:
> I have done that and it works. The only thing that would cause problems is if the tests are currently performed with a limited set of the standard library. 

Right. For now, we could focus on library tests, e.g. TestList. They obviously load the entire library. 

> Another thing that we should consider in the future is the possibility of Assert returning a boolean in order to be more intuitive with Tests. We recommend people to use Bool in their tests, but if you use assert you have to change the return type of your functions to Unit. 

I initially designed tests to return a Boolean (or actually any non-false value). I now think this was a big mistake. 

To me, a test should have one or more asserts (giving us the Assert effect) and return Unit.

This is simple, clean, and explicit.

(Allowing tests to return any value has many absurd consequences. For example, is it ok for a test to return `NaN`? Also, when a test fails, we have no idea why, because there was no explicit assert.)

magnus-madsen:
> Following that same logic we could do a runWithBool that takes only f: Unit -> bool \ ef + Assert and returns bool with ef. That way throwing exceptions is not required anymore and we swiftly integrate with the Test framework. 

We could, but I see no reason to have tests return a Boolean. Effects and handlers give us all the handles we need.

magnus-madsen (approved):

magnus-madsen:
I merged this and created a ticket for the error message: https://github.com/flix/flix/issues/11455

NeonOxide:
I agree with that. However, we need to change the test runner to be able to differentiate between tests that fail and are correct. How do you do that with just a simple Unit return value? One alternative would be to catch the AssertionError exceptions but then you are tying the logic of the Test runner to Java world. Another problem with that is that if we have multiple asserts it would stop at the first one (unless we made a collecting runWithIO which is possible but again, requires a little bit of thought about what to do) I think the way forward is tests being able to differentiate between failure and correctness by their return value ( by a boolean or other value) but restrict the return type of test functions to that value. What do you think?

magnus-madsen:
> I agree with that. However, we need to change the test runner to be able to differentiate between tests that fail and are correct. How do you do that with just a simple Unit return value? 

It seems you implicitly assume that the test runner has to be written in Scala and we have to cross some FFI boundary. 

Do we have to do all that? Why not move all that code into Flix and write a very short adapter (if needed at all).

magnus-madsen:
And if we do that--  maybe we can keep the test framework pluggable / swappable?

NeonOxide:
> > I agree with that. However, we need to change the test runner to be able to differentiate between tests that fail and are correct. How do you do that with just a simple Unit return value?
> 
> It seems you implicitly assume that the test runner has to be written in Scala and we have to cross some FFI boundary.
> 
> Do we have to do all that? Why not move all that code into Flix and write a very short adapter (if needed at all).

I was implicitly assuming that. I agree that it should be perfectly swappable with Flix. However, we are not able to access annotations in Flix right? How would you collect all the functions needed to run?

NeonOxide:
In fact I open another problem. How would you handle running a collection of functions that all may have different sets of effects?

magnus-madsen:
> I was implicitly assuming that. I agree that it should be perfectly swappable with Flix. However, we are not able to access annotations in Flix right? How would you collect all the functions needed to run?

Excellent point. What do you think? (Remember we control the entire compiler).

NeonOxide:
Because if not we could simply do accept a function def myTestRunner(funs: List[Unit -> Unit \ ef + Assert]): Unit \ ef or something like that and plug via the compiler all the functions with @Test.

NeonOxide:
Because as far as I know it is safe to be covariant over effects in a list of functions, (eg. List[a -> b \ Pure, a -> b \ IO, a -> b \ Sys]) could be coerced to List[a -> b \ IO + Sys] safely. If we do that, the Test runner would need to take a List[Unit -> Unit \ PrimitiveEffects + Assert] and that would solve all problems. But I don't know how feasible this coercion is

NeonOxide:
With subeffecting it should work seamlessly I think

PR #11457, opened 2025-08-31
test: cleanup `TestRef.flix`

magnus-madsen:


magnus-madsen:
@JonathanStarup @mlutze @jaschdoc Other than the potentially dangerous combination of region and effect handlers, this is the style I am thinking about for library tests going forward. Notably:

- The signature is `Unit -> Unit \ Assert`
- The tests are called e.g. `transform03`. 
- Each test has a explicit asserts (can be more than one per test, but at least one).

mlutze:
> @JonathanStarup @mlutze @jaschdoc Other than the potentially dangerous combination of region and effect handlers, this is the style I am thinking about for library tests going forward. Notably:
> 
>     * The signature is `Unit -> Unit \ Assert`
>     * The tests are called e.g. `transform03`.
>     * Each test has a explicit asserts (can be more than one per test, but at least one).

* no license

PR #11458, opened 2025-08-31
test: cleanup `TestMutDisjointSets.flix`

magnus-madsen:


magnus-madsen:
@mlutze I think some of the test code is sub optimal, but this PR is only about changing the surroundings not the tests themselves.

PR #11459, opened 2025-08-31
refactor: remove `Assert.eq`

magnus-madsen:


magnus-madsen:
@jaschdoc We have removed `Assert.eq`. You can use the `Assert.assertEq` effect (which is WIP) or you can go without.

magnus-madsen:
@jaschdoc Could approve this PR and the others? Thanks!

mlutze (approved):

PR #11460, opened 2025-08-31
test: reduce usage of `Assert.eq` in handler tests

magnus-madsen:


magnus-madsen:
For now, my goal is to get rid of `Assert.eq`. Later I will introduce usage of the algebraic effect Assert.

magnus-madsen:
@jaschdoc I used Claude for this PR- so please randomly inspect a few of the changes. I will do the same.

PR #11461, opened 2025-09-01
chore: remove dead `Inline` keyword

jaschdoc:


magnus-madsen:
Ah- the keyword is dead? But not the annotation- is that right?

jaschdoc:
Exactly. This does not touch the annotations.

magnus-madsen:
Good catch 👍

PR #11468, opened 2025-09-01
fix: selectively apply the default `Assert` handler

magnus-madsen:


main/src/ca/uwaterloo/flix/language/phase/EntryPoints.scala
```diff
@@ -119,6 +119,10 @@ object EntryPoints {
   private def isTest(defn: TypedAst.Def): Boolean =
     defn.spec.ann.isTest
 
+  /** Returns `true` if `defn` has the Assert effect. */
+  private def hasAssert(defn: TypedAst.Def): Boolean =
+    defn.spec.eff.effects.contains(Symbol.Assert)
```
mlutze, line 124: This is a syntactic check. Things might go wrong if I have a function `foo(): Unit \ Assert - Assert`

mlutze: We should at least add a comment if this is acceptable.

mlutze (approved):

PR #11469, opened 2025-09-01
fix: remove slow permutation tests

magnus-madsen:


magnus-madsen:
@NeonOxide If you are interested in the performance aspect, these tests take 30 sec to run with effect handlers (!)

NeonOxide:
> @NeonOxide If you are interested in the performance aspect, these tests take 30 sec to run with effect handlers (!)

I can take a look into it, but realistically it should not have added that much overhead given that the handlers are not multishot right? They should be translated to something close to simple function calls.

magnus-madsen:
> > @NeonOxide If you are interested in the performance aspect, these tests take 30 sec to run with effect handlers (!)
> 
> I can take a look into it, but realistically it should not have added that much overhead given that the handlers are not multishot right? They should be translated to something close to simple function calls.

Should be :-) but they are not.

PR #11488, opened 2025-09-04
test: refactor `TestInt8` to use assertions

magnus-madsen:


mlutze:
In other test files, you removed the `////// mod //////` stuff.

PR #11497, opened 2025-09-05
test: refactor `TestFoldable` to use assertions

magnus-madsen:


main/test/ca/uwaterloo/flix/library/TestFoldable.flix
```diff
@@ -669,6 +636,6 @@ mod TestFoldable {
                 resume(x)
             }
         };
-        List#{1, 2} == filtered and List#{3, 4} == MutList.toList(effects)
+        assertTrue(List#{1, 2} == filtered and List#{3, 4} == MutList.toList(effects))
```
mlutze, line 639: Should be two asserts.

magnus-madsen: I plan to do a pass later over all assertions. For example, we will have e.g. assertNonEmpty.

But since the API is not yet there, I am satisfied with anything that uses the effect.

PR #11527, opened 2025-09-07
feat: add `RichString`

magnus-madsen:


magnus-madsen:
@mlutze Thoughts on API?

magnus-madsen:
image

mlutze (commented):
API seems reasonable.

I don't love that the colors are all lies...

magnus-madsen:
> API seems reasonable.
> 
> I don't love that the colors are all lies...

What do you mean?

mlutze:
> > API seems reasonable.
> > I don't love that the colors are all lies...
> 
> What do you mean?

```
    pub def green(x: a): RichString with Formattable[a] =
        withColor(Color.Rgb(13, 188, 121), x)
```

To me, green is `(0, 255, 0)`.

These colors seem adapted to a particular terminal background.

mlutze (approved):

mlutze:
I approved so I'm not blocking merging, but disabled automerge so you can think about what you've done.

Issue #10227, opened 2025-04-01
Add `clean` command for flix repl


jaschdoc:
Simply clean the `build` directory

mlutze:
The higher-ups are opposed to commands that delete things.

magnus-madsen:
I will allow you to add it provided that:

- We first scan the build directory for files. If the directory contains any file that does not end with .class, .flixir, .jar (and what else)? we immediately abort. Only then will we remove the directory.

Just to be clear: We do not delete a single file in the directory until we have determined that it only contains files of the expected sort.

mlutze:
Can we get something for lib cleaning as well?

magnus-madsen:
> Can we get something for lib cleaning as well?

How would it work?

mlutze:
When Flix downloads a new version of a library (as a result of TOML stuff) it should remove the old version.

mlutze:
I don't see the difficulty, at least for a naive solution. When I change TOML, I go from one dependency graph to another. Add the new stuff and remove the difference.

jaschdoc:
> I will allow you to add it provided that:
> 
>     * We first scan the build directory for files. If the directory contains any file that does not end with .class, .flixir, .jar (and what else)? we immediately abort. Only then will we remove the directory.
> 
> 
> Just to be clear: We do not delete a single file in the directory until we have determined that it only contains files of the expected sort.

What does Gradle do? I think many package managers just delete the entire thing

mlutze:
> What does Gradle do? I think many package managers just delete the entire thing

Gradle has a cache in some central location (like the home directory) to avoid downloading the same dependency multiple times, so it's easy for them to delete and then copy from the local cache as needed. I don't think we want to manipulate files outside the project directory right now, so we can't use that same system. If we delete everything then we have to re-download everything, which might be annoying for large projects.

jaschdoc:
> > What does Gradle do? I think many package managers just delete the entire thing
> 
> Gradle has a cache in some central location (like the home directory) to avoid downloading the same dependency multiple times, so it's easy for them to delete and then copy from the local cache as needed. I don't think we want to manipulate files outside the project directory right now, so we can't use that same system. If we delete everything then we have to re-download everything, which might be annoying for large projects.

Ah I was only talking about the `build/class` directory

magnus-madsen:
Lets split this ticket into two: one concerns clean (which affects build folder) and one about libraries.

jaschdoc:
[...]

PR #11538, opened 2025-09-08
feat: add `clean` command

jaschdoc:
Refactor into steps.

Closees https://github.com/flix/flix/issues/10227


main/src/ca/uwaterloo/flix/Main.scala
```diff
@@ -207,6 +207,20 @@ object Main {
               System.exit(1)
           }
 
+        case Command.Clean =>
+          implicit val console: java.io.Console = System.console()
```
magnus-madsen, line 211: No sure what this is. I think we just System.out / System.err elsewhere.

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -528,6 +529,72 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
...
+      val deletionFiles = files.map(p => p.toString).mkString("- ", System.lineSeparator() + "- ", "")
+      val deletionSummary =
+        s"""
+           |Removing the following files:
+           |
+           |$deletionFiles
+           |""".stripMargin
+
+      out.println(deletionSummary)
+      val confirmation = in.readLine("Do you want to continue? [y/N] ").toLowerCase.strip()
```
magnus-madsen, line 584: I see. But we usually don't do that for other things (i.e. we have nowhere else in the compiler where we prompt).

I don't think we should begin in this PR; just abort.

Ideally this situation never happens anyway!

magnus-madsen:
I am not necessarily against prompting, I just don't want to add it ad-hocly now. I would rather we do a pass over all build commands and then after that we can add prompting.

jaschdoc:
Yeah I'm just exploring what it should feel like

jaschdoc:
Don't look at this yet please :)

magnus-madsen:
> Yeah I'm just exploring what it should feel like

 Sounds good.

magnus-madsen:
I will look away 🌅

PR #11539, opened 2025-09-08
refactor: remove type from `KindedAst.Enum`

### LINKED ISSUE #8720: Remove `Enum.tpe` up to `KindedAst`
Follow up from #8719

You can reconstruct `Enum.tpe` from `Enum.sym` and `Enum.tparams`

CasperDN:
Closes #8720.

Removes `tpe: Type` from `KindedAst.Enum` and instead reconstruct it as needed in `Deriver`.

Also as a side note: What is `ResolvedAst.TypeParam.Implicit` and why can it occur in `Enum`?

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Deriver.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Deriver.scala
@@ -16,6 +16,7 @@
 package ca.uwaterloo.flix.language.phase
 
 import ca.uwaterloo.flix.api.Flix
+import ca.uwaterloo.flix.language.ast.KindedAst.TypeParam
 import ca.uwaterloo.flix.language.ast.shared.*
 import ca.uwaterloo.flix.language.ast.shared.SymUse.*
 import ca.uwaterloo.flix.language.ast.{Kind, KindedAst, Name, Scheme, SemanticOp, SourceLocation, Symbol, Type, TypeConstructor}
@@ -58,10 +59,25 @@ object Deriver {
   }
 
   /**
+    * Reconstructs the type of the enum.
+    * @param sym The [[Symbol]] of the enum.
+    * @param tparams The list of [[TypeParam]] of the enum.
+    */
+  private def deriveEnumType(sym: Symbol.EnumSym, tparams: List[TypeParam]): Type = {
+    val tvars = tparams.map(tparam => Type.Var(tparam.sym, tparam.loc.asSynthetic))
+    val kinds = tvars.map(symm => symm.kind)
+//    See `Phase.Kinder.visitEnum`.
+    val kind = kinds.foldRight(Kind.Star: Kind) {
+      case (tparam, acc) => tparam ->: acc
+    }
+    Type.mkApply(Type.Cst(TypeConstructor.Enum(sym, kind), sym.loc.asSynthetic), tvars, sym.loc.asSynthetic)
+  }
+
+  /**
     * Builds the instances derived from this enum.
     */
   private def getDerivedInstances(enum0: KindedAst.Enum, root: KindedAst.Root)(implicit sctx: SharedContext, flix: Flix): List[KindedAst.Instance] = enum0 match {
-    case KindedAst.Enum(_, _, _, enumSym, _, derives, cases, _, _) =>
+    case KindedAst.Enum(_, _, _, enumSym, _, derives, cases, _) =>
 
       derives.traits.flatMap {
         case Derivation(traitSym, loc) if cases.isEmpty =>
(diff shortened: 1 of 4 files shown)


magnus-madsen:
@mlutze If you can review

main/src/ca/uwaterloo/flix/language/phase/Deriver.scala
```diff
@@ -57,11 +58,26 @@ object Deriver {
     (root.copy(instances = newInstances), sctx.errors.asScala.toList)
   }
 
+  /**
+    * Reconstructs the type of the enum.
+    * @param sym The [[Symbol]] of the enum.
+    * @param tparams The list of [[TypeParam]] of the enum.
+    */
+  private def deriveEnumType(sym: Symbol.EnumSym, tparams: List[TypeParam]): Type = {
```
mlutze: Let's just call it `getEnumType` since `derive` refers to the creation of new trait instances.

mlutze: Move this helper to the bottom of the file.

main/src/ca/uwaterloo/flix/language/phase/Deriver.scala
```diff
@@ -57,11 +58,26 @@ object Deriver {
...
+    val tvars = tparams.map(tparam => Type.Var(tparam.sym, tparam.loc.asSynthetic))
+    val kinds = tvars.map(symm => symm.kind)
+//    See `Phase.Kinder.visitEnum`.
+    val kind = kinds.foldRight(Kind.Star: Kind) {
+      case (tparam, acc) => tparam ->: acc
```
mlutze: `tparam` is a kind; Calling it a `tparam` is confusing.

main/src/ca/uwaterloo/flix/language/phase/Deriver.scala
```diff
@@ -57,11 +58,26 @@ object Deriver {
...
 
+  /**
+    * Reconstructs the type of the enum.
+    * @param sym The [[Symbol]] of the enum.
+    * @param tparams The list of [[TypeParam]] of the enum.
+    */
+  private def deriveEnumType(sym: Symbol.EnumSym, tparams: List[TypeParam]): Type = {
+    val tvars = tparams.map(tparam => Type.Var(tparam.sym, tparam.loc.asSynthetic))
+    val kinds = tvars.map(symm => symm.kind)
+//    See `Phase.Kinder.visitEnum`.
```
mlutze: I'm not sure what this comment means. If there's something worth commenting, it's better to put the actual comment here instead of redirecting.

mlutze:
> Also as a side note: What is `ResolvedAst.TypeParam.Implicit` and why can it occur in `Enum`?

This refers to implicit type parameters, which can occur in defs. Compare:

```scala
// implicit type parameter
def id(x: a): a = x

// explicit type parameter
def id[a: Type](x: a): a = x
```

It also covers wildcard type parameters:

```scala
// implicit type parameter
def length(x: List[_]): Int32 = ???
```

These cases cannot happen with enums, however, because they require explicit type parameters.

main/src/ca/uwaterloo/flix/language/phase/Deriver.scala
CasperDN: It was just a reference to where I (more or less) copied the code from.

CasperDN:
@mlutze Do we want to enforce this in the Kinder? `getEnumKind` currently ends up in `getKindEnvFromTypeParam` which would accept the `Implicit`. I imagine this happens because we want to reuse old code, but still.

mlutze:
> @mlutze Do we want to enforce this in the Kinder? `getEnumKind` currently ends up in `getKindEnvFromTypeParam` which would accept the `Implicit`. I imagine this happens because we want to reuse old code, but still.

I think it's fine to allow it in the Kinder. It's checked in an earlier phase, and it keeps the Kinder code more regular.

main/src/ca/uwaterloo/flix/language/phase/Deriver.scala
```diff
@@ -963,6 +975,20 @@ object Deriver {
...
+  /**
+    * Reconstructs the type of the enum.
+    * @param sym The [[Symbol]] of the enum.
+    * @param tparams The list of [[TypeParam]] of the enum.
+    */
+  private def getEnumType(sym: Symbol.EnumSym, tparams: List[TypeParam]): Type = {
+    val tvars = tparams.map(tparam => Type.Var(tparam.sym, tparam.loc.asSynthetic))
+    val kinds = tvars.map(symm => symm.kind)
+    val kind = kinds.foldRight(Kind.Star: Kind) {
+      case (kindParam, acc) => kindParam ->: acc
```
mlutze: Actually, it's simpler just to use Kind.mkArrow(List[Kind]) here.

PR #11541, opened 2025-09-09
test: refactor `TestDelayList` to use assertions

magnus-madsen:


magnus-madsen:
@mlutze If you could merge this.

The rest will have to be done manually.

PR #11544, opened 2025-09-09
feat: use `RichString` in `Assert`

magnus-madsen:


magnus-madsen:
I plan to refactor more asserts later, but I am taking an incremental approach.

mlutze (approved):

PR #11545, opened 2025-09-09
ci: trim community builds

magnus-madsen:
I am removing community builds that are small. 

I think apps should be at least 500 lines of non-trival code whereas reusable libraries can be a bit smaller.

### DIFF
--- a/.github/workflows/community-build.yaml
+++ b/.github/workflows/community-build.yaml
@@ -40,16 +40,11 @@ jobs:
       max-parallel: 1
       matrix:
         repo:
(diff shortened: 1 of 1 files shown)


mlutze:
I'm not sure I agree with this.

The point of the community builds is to have a breadth of real-world programs to test on. Size shouldn't be a factor.

magnus-madsen:
> I'm not sure I agree with this.
> 
> The point of the community builds is to have a breadth of real-world programs to test on. Size shouldn't be a factor.

These programs are very small and don't add much.

mlutze (approved):

PR #11548, opened 2025-09-09
ci: rename jobs

magnus-madsen:
Trying to give more sensible names for the CI GitHub box

### DIFF
--- a/.github/workflows/fuzzing-completer.yml
+++ b/.github/workflows/fuzzing-completer.yml
@@ -1,4 +1,4 @@
-name: IDE Completion Tests
+name: Fuzz Autocomplete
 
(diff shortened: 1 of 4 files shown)


.github/workflows/fuzzing-completer.yml
```diff
@@ -21,8 +21,6 @@ jobs:
         with:
           distribution: 'temurin'
           java-version: '21'
-      - name: Build with gradle
-        run: ./gradlew clean compileTestScala --console=plain
```
magnus-madsen, line 25: Useless code (AFAIK)

magnus-madsen:
> You might need to modify the "required actions" after this.

Where is this?

PR #11549, opened 2025-09-09
feat: better number parsing

JonathanStarup:
- [Fix] Treats e.g. `42i32x` as a single wrong number (previously int `42i32` and variable `x`)
- [Fix] Allows `12e+12`.
- [Feat] Errors are better
  - `32i31` says that `i31` is an invalid suffix (previously "Undefined name 'i31'")
  - `12e12e` says that a digit was expected instead of `e` (second occurrence) (previously "malformed float")
  - `12.4i32` says that decimals cannot have integer suffixes (previously "malformed float")
- [Test] More tests
- [Docs] Documented a regex describing valid numbers.
- [Refactor] The regex based function makes it easier to read
- [Performance] `--Xbenchmark-phases`
  - `    PR: 164,5 ms (07,2%)` 
  - `Master: 165,6 ms (07,2%)`

Follow up work: sort `LexerError` and `TestLexer`

### DIFF
--- a/main/test/ca/uwaterloo/flix/language/phase/TestWeeder.scala
+++ b/main/test/ca/uwaterloo/flix/language/phase/TestWeeder.scala
@@ -1051,18 +1051,6 @@ class TestWeeder extends AnyFunSuite with TestUtils {
     expectError[WeederError.InlineAndDontInline](result)
   }
 
-  test("MalformedFloat64.01") {
-    val input = "def f(): Float64 = 1.7976931348623158e+308"
-    val result = compile(input, Options.TestWithLibNix)
-    expectError[WeederError.MalformedFloat](result)
-  }
-
-  test("MalformedFloat64.02") {
-    val input = "def f(): Float64 = -1.7976931348623158e+308"
-    val result = compile(input, Options.TestWithLibNix)
-    expectError[WeederError.MalformedFloat](result)
-  }
-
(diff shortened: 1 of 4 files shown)


main/test/ca/uwaterloo/flix/language/phase/TestWeeder.scala
```diff
@@ -1051,18 +1051,6 @@ class TestWeeder extends AnyFunSuite with TestUtils {
     expectError[WeederError.InlineAndDontInline](result)
   }
 
-  test("MalformedFloat64.01") {
-    val input = "def f(): Float64 = 1.7976931348623158e+308"
```
JonathanStarup, line 1055: Previously these were lexed as '1.7976931348623158e', '+', and '308' which caused the malformed error. But this is a valid float so im not sure why this test was added

JonathanStarup:
Just to show an example of how errors are simpler and better at the same time:
```
def main(): String = ToString.toString(
    3e3e3e3e3
)
```
PR
```
>> Malformed number, found 'e'.

2 |     3e3e3e3e3
           ^
           Number was correct up to here
```
Master
```
>> Number has two scientific notation indicators.

2 |     3e3e3e3e3
               ^
               Second 'e' is here.
```

magnus-madsen (approved):

magnus-madsen:
Great. Thanks!

PR #11550, opened 2025-09-10
feat: add new handlers for `Assert`

NeonOxide:
Added some new handlers for the Assert effect with common functionality that could be desired (logging errors, printing to stderr, ...).

I did not remove the red Assertion failed from the previous handler in order to not mix commits. But right now that handler prints Assertion Failed twice due to Assertion Failed also being contained in the assertion message.

I also did not create runWithLogger, runWithStdout, ... Because I did not know if we desired those runners. In my opinion, they are nice to have. But they are also really easy to add.

I am creating another pull request because I accidentally closed the other one thinking I was not in master branch when doing some hard resets. Lesson learnt, no risky git actions first thing out from bed.


NeonOxide:
Okay, there were some changes to use RichString that I hadn't seen. I'll fix them right away

main/src/library/Assert.flix
```diff
@@ -157,6 +157,133 @@ mod Assert {
             }
         }
 
+    ///
+    /// Handles the `Assert` effect by ignoring all assertions (no-op).
+    ///
+    pub def handleNoOp(f: a -> b \ ef): (a -> b \ (ef - Assert)) = x ->
```
magnus-madsen: -> handleWithNoOp?

main/src/library/Assert.flix
```diff
@@ -157,6 +157,133 @@ mod Assert {
...
+    ///
+    pub def handleWithLogger(f: a -> b \ ef): (a -> b \ (ef - Assert) + Logger) = x ->
+        run {
+            f(x)
+        } with handler Assert {
+            def assert(cond, msg, resume) = {
+                if (cond)
+                    resume(())
+                else {
+                    Logger.fatal(msg);
```
magnus-madsen: It is not fatal since we resume. So I would make it error.

main/src/library/Assert.flix
```diff
@@ -157,6 +157,133 @@ mod Assert {
...
+    ///
+    pub def handleWithStdoutPrint(f: a -> b \ ef): (a -> b \ (ef - Assert) + IO + Sys) = x ->
+        run {
+            f(x)
+        } with handler Assert {
+            def assert(cond, msg, resume) = {
+                if (cond)
+                    resume(())
+                else {
+                    run {
```
magnus-madsen: We can just call println directly. It has the IO effect.

magnus-madsen: We dont need the environmental handler.

main/src/library/Assert.flix
```diff
@@ -157,6 +157,133 @@ mod Assert {
...
+        run {
+            f(x)
+        } with handler Assert {
+            def assert(_cond, _msg, resume) = resume(())
+        }
+
+    ///
+    /// Handles the `Assert` effect by logging assertion failures using the `Logger` effect.
+    ///
+    /// Assertions continue execution after logging the failure.
```
magnus-madsen: Let us not use the word "failure" but either the full "assertion failure" or "assertion violation"

(This comment applies elsewhere too.)

main/src/library/Assert.flix
```diff
@@ -157,6 +157,133 @@ mod Assert {
...
+        } with handler Assert {
+            def assert(cond, msg, resume) = {
+                if (cond)
+                    resume(())
+                else {
+                    use Console.eprintln;
+                    run {
+                        eprintln(RichString.toAnsiString(msg))
+                    } with Environment.runWithIO
+                      with Console.runWithIO;
```
magnus-madsen: Same comments as above.

main/src/library/Assert.flix
```diff
@@ -157,6 +157,133 @@ mod Assert {
...
+                }
+            }
+        }
+
+    ///
+    /// Handles the `Assert` effect by printing assertion failures to stdout.
+    ///
+    /// Assertions continue execution after printing the failure.
+    ///
+    pub def handleWithStdoutPrint(f: a -> b \ ef): (a -> b \ (ef - Assert) + IO + Sys) = x ->
```
magnus-madsen: Idea for better name? handleWithStdOut?

main/src/library/Assert.flix
```diff
@@ -157,6 +157,133 @@ mod Assert {
...
+                }
+            }
+        }
+
+    ///
+    /// Handles the `Assert` effect by converting assertion failures into abort operations.
+    ///
+    /// When an assertion fails, the program will abort with the assertion message.
+    ///
+    pub def handleWithAbort(f: a -> b \ ef): (a -> b \ (ef - Assert) + Abort) = x ->
```
magnus-madsen: Lets skip this one. Abort is not yet "functional".

main/src/library/Assert.flix
```diff
@@ -157,6 +157,133 @@ mod Assert {
...
+            def assert(cond, msg, resume) = {
+                if (cond)
+                    resume(())
+                else {
+                    use Console.{print, readln};
+                    run {
+                        println("");
+                        println(RichString.toAnsiString(msg));
+                        println("");
+                        print(RichString.toAnsiString(yellow("Continue execution? (y/n): ")));
```
magnus-madsen: There should be a default-- which I think should be Y. Then everything other than "n" will continue.

main/src/library/Assert.flix
```diff
@@ -157,6 +157,133 @@ mod Assert {
...
+                        if (not String.isEmpty(input))
+                                let trimmed = String.trim(input);
+                                if (String.toLowerCase(trimmed) == "y" or String.toLowerCase(trimmed) == "yes")
+                                    resume(())
+                                else if (String.toLowerCase(trimmed) == "n" or String.toLowerCase(trimmed) == "no")
+                                    Abort.abort("Execution aborted by user after assertion failure")
+                                else
+                                    Abort.abort("Execution aborted after obtaining malformed input following assertion failure.")
+                        else
+                            Abort.abort("Could not read user input after assertion failure")
```
magnus-madsen: No abort? We should just return?

main/src/library/Assert.flix
NeonOxide: There is no error level. The previous one to fatal is warning. Do we want that one?

magnus-madsen: Yeah

main/src/library/Assert.flix
NeonOxide: We cannot return without a value. Same as runWithIO. We could simply throw an exception. That is part of why I think tests should return Bool or some kind of validation. Then from all of these handlers we could return true or false or the equivalent.

main/src/library/Assert.flix
NeonOxide: Sounds good to me

NeonOxide: I also changed StdErr to be consistent

main/src/library/Assert.flix
NeonOxide: Sounds good

main/src/library/Assert.flix
```diff
@@ -157,6 +157,107 @@ mod Assert {
...
+        run {
+            f(x)
+        } with handler Assert {
+            def assert(cond, msg, resume) = {
+                if (cond)
+                    resume(())
+                else {
+                    use Console.eprintln;
+                    run {
+                        eprintln(RichString.toAnsiString(msg))
```
magnus-madsen, line 223: This is because we have no eprintln in prelude. But perhaps we should have that instead?

main/src/library/Assert.flix
```diff
@@ -157,6 +157,107 @@ mod Assert {
...
+    ///
+    /// Assertions continue execution after logging the assertion failure.
+    ///
+    pub def handleWithLogger(f: a -> b \ ef): (a -> b \ (ef - Assert) + Logger) = x ->
+        run {
+            f(x)
+        } with handler Assert {
+            def assert(cond, msg, resume) = {
+                if (cond)
+                    resume(())
```
magnus-madsen, line 181: We can just call resume() right? Or am I wrong?

magnus-madsen:
@NeonOxide I merged this, but also added two comments for you to consider.

main/src/library/Assert.flix
NeonOxide: I mean, you need to only log the failures. That is why the conditional is there. I don't think I get your point.

magnus-madsen: The syntax. You don't need (()), just (), right?

PR #11551, opened 2025-09-10
test: refactor `TestDayList` to use `Assert` effect

### LINKED ISSUE #11543: Refactor `TestDelayList` to use assertions
This PR #11541 refactors `TestDelayList` to use assertions-- but I was not able to get all the functions refactored. 

This ticket is to finish the rest in the same style (e.g. using Assert.assertXYZ) and returning `Unit`.

jaschdoc:
Closes #11543

### DIFF
--- a/.github/workflows/compiler-build-and-test.yaml
+++ b/.github/workflows/compiler-build-and-test.yaml
@@ -1,14 +1,14 @@
-name: Compiler Fuzzing
+name: Compiler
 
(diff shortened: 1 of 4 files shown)


magnus-madsen:
Nice! Can you do a grep for `(): Bool` there should be no such functions.

jaschdoc:
`Bool` does not occur anywhere in the file :)

PR #11553, opened 2025-09-10
ci: combine authors and semantic commit checks

mlutze:


.github/workflows/check-pr-data.yaml
```diff
@@ -0,0 +1,32 @@
+name: Check PR Data
```
magnus-madsen: Can we call it "Style" -- that way it appears last.

mlutze: I changed the C to С so it sorts last.

magnus-madsen: What? How does that even work?

magnus-madsen: If this is some other C then please no 🤡

mlutze: Knew you would love it.

PR #11557, opened 2025-09-10
test: add `TestStandardLibrary`

magnus-madsen:


magnus-madsen:
@JonathanStarup @NeonOxide 

This PR compiles ALL library unit tests into one big program. 
Note incremental compilation is disabled, so intermediate ASTs are not stored.

Memory usage is a problem. 

Up to type inference we only use 600MBish which is great. 
Up to just before EffectBinder we use 1100MBish which is acceptable.
Up to just before JvmBackend we use 2500MBish which is a lot.
By the end of the compiler pipeline we use 3400MBish which does not fit.

magnus-madsen:
I have identified the following bottlenecks:

- SimpleType can be cached.
- The use of lists in SimpleType consumes a lot of space. (We may want our own Vector type.)
- The use of lists in ReducedAst consumes a lot of space. (We may want our own Vector type.)
- The implementation of Flix.codeGen means that we hold onto old ASTs even though they are no longer used.
- We hold on to _ALL_ Method objects (even for methods that are not entry points or tests).
- The generated bytecode uses 600MB.

mlutze:
> We may want our own Vector type.

Maybe of interest: https://www.scala-lang.org/api/current/scala/collection/immutable/ArraySeq.html

magnus-madsen:
I think `ArraySeq` fits the bill. We should use it more.

magnus-madsen:
@mlutze Ready. The setup here is simple. The new goal is to compile *ALL* standard library tests in one go.

magnus-madsen:
This change has the upshot of also testing the Flix compiler under serious load.

main/test/ca/uwaterloo/flix/StandardLibrarySuite.scala
```diff
@@ -0,0 +1,85 @@
...
+    case ex: Throwable =>
+      // We create a fictitious test to ensure that something shows up.
+      test("StandardLibrary -- COMPILATION FAILED.") {
+        fail(ex)
+      }
+  }
+
+  private def runTests(r: CompilationResult): Unit = {
+    // Group the tests by namespace.
+    val testsByNamespace = r.getTests.groupBy(_._1.namespace)
```
mlutze: prefer matching instead of `_._1`

main/test/ca/uwaterloo/flix/StandardLibrarySuite.scala
```diff
@@ -0,0 +1,85 @@
...
+
+    // Sort the namespaces.
+    val testsByNamespaceSorted = testsByNamespace.toList.sortBy(p => p._1.mkString("."))
```
mlutze: idem

main/test/ca/uwaterloo/flix/StandardLibrarySuite.scala
```diff
@@ -0,0 +1,85 @@
...
+    for ((_, tests) <- testsByNamespaceSorted) {
+      // Sort the tests by name.
+      val testsByName = tests.toList.sortBy(_._1.name)
+
+      // Dynamically create a ScalaTest unit test for each @Test function.
+      for ((sym, TestFn(_, skip, run)) <- testsByName; if !skip) {
+        test(sym.toString) {
+          run()
+        }
+      }
```
mlutze: We can record that there are skipped tests like so:

```suggestion
      for ((sym, TestFn(_, skip, run)) <- testsByName) {
        if (skip) {
          ignore(sym.toString) {}
        } else {
          test(sym.toString) {
            run()
          }
        }
      }
```

main/test/ca/uwaterloo/flix/StandardLibrarySuite.scala
```diff
@@ -0,0 +1,85 @@
...
+  /** The default options. */
+  private val Opts = Options.DefaultTest.copy(incremental = false, outputJvm = false)
+
+  private def init(): Unit = try {
+    // Create a new Flix compiler.
+    val flix = new Flix
+    flix.setOptions(Opts)
+
+    // Find and add all test suites.
+    val paths = FileOps.getFlixFilesIn(Paths.get(Path), 1)
```
mlutze, line 41: What's `1`?

magnus-madsen: Ask Jakob, he changed the API :( Its the search depth.

magnus-madsen:
@mlutze Fixed

mlutze:
!!!rerun

main/test/ca/uwaterloo/flix/StandardLibrarySuite.scala
```diff
@@ -62,20 +62,29 @@ class StandardLibrarySuite extends AnyFunSuite {
...
       // Sort the tests by name.
       val testsByName = tests.toList.sortBy(_._1.name)
 
       // Dynamically create a ScalaTest unit test for each @Test function.
-      for ((sym, TestFn(_, skip, run)) <- testsByName; if !skip) {
-        test(sym.toString) {
-          run()
+      for ((sym, TestFn(_, skip, run)) <- testsByName) {
+        val testName = sym.toString
+        if (skip){
```
mlutze, line 82: missing space

PR #11558, opened 2025-09-11
feat: reduce memory usage in `Flix.codeGen`

magnus-madsen:


mlutze:
@JonathanStarup find a more beautiful solution please

PR #11562, opened 2025-09-11
refactor: use `ArraySeq` in `ReducedAst`

magnus-madsen:


magnus-madsen:
I am not really sure about this. Is this something we want to pursue?

magnus-madsen:
@JonathanStarup Your comments are also welcome.

How far would we take this? Use ArraySeq everywhere in ReducedAst? Everywhere in the compiler?

magnus-madsen:
I feel like we have to do it:

CleanShot 2025-09-12 at 08 39 17

JonathanStarup:
how much does it reduce memory by?

mlutze:
> @JonathanStarup Your comments are also welcome.
> 
> How far would we take this? Use ArraySeq everywhere in ReducedAst? Everywhere in the compiler?

It's probably the right thing to use almost everywhere. We make little use of the properties of lists.

magnus-madsen:
> how much does it reduce memory by?

Don't know yet.

PR #11566, opened 2025-09-11
Experiment: Controlled .class output

JonathanStarup:
This is not really the correct solution. but it shows which class do want jvm output and which doesn't

It is awkward since `build` is used both as the `:build` command and as a generic way to build the project.


JonathanStarup:
@magnus-madsen this technically fixes the .class loading issue, but its not so clean

magnus-madsen:
Yeah, this is a solution, but not a good one.

I think we need to rework this as part of the package manager overhaul.

In particular, I think that something like we only emit class files when we are trying to build a jar or fat-jar.

So possibly the build commands in Bootstrap need to track some flags.

magnus-madsen:
We are remaking the package manager.

PR #11567, opened 2025-09-11
refactor: pipe without null

JonathanStarup:
@mlutze you like it?

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/Flix.scala
+++ b/main/src/ca/uwaterloo/flix/api/Flix.scala
@@ -700,12 +700,6 @@ class Flix {
 
   /**
     * Compiles the given typed ast to an executable ast.
-    *
-    * Note: The `codeGen` method has a long execution time, and its local variables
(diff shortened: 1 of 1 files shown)


mlutze:
Hooray! Is the memory footprint saved?

JonathanStarup:
(didn't check but I assume so)

magnus-madsen:
It is nice, but given Scala compiler magic and closures, we have to check that memory is not leaked.

Insert a Thread.sleep somewhere in the middle of the backend pipeline, open JProfiler, and check that the "objects in memory" (or whatever it is called)  does not list objects from previous ASTs in the middle of the pipeline.

JonathanStarup:
> It is nice, but given Scala compiler magic and closures, we have to check that memory is not leaked.

I think Flix would make them long lived because of the EffectBinder transformation. Something to think about on our end

JonathanStarup:
@mlutze you can resurrect if you want

PR #11568, opened 2025-09-11
refactor: remove unused function

JonathanStarup:


magnus-madsen:
Flix would have complained :)

PR #11569, opened 2025-09-11
chore: rename class to trait in comments and param

kukimik:
See https://github.com/flix/flix/pull/11390.

Cleanup to make future devs' life a little bit easier.


magnus-madsen:
Many thanks!

Issue #11436, opened 2025-08-27
Meta Issue: Tracking the Future of the Flix Package Manager


jaschdoc:
This meta issue tracks all work, bugs, and proposals related to the future of the Flix package manager.

---

## Design Guidelines & Philosophy

- **User Experience First:** All design decisions should prioritize user experience.
- **Self-aware Actions:** The package manager should be smart enough to perform all necessary sub-actions for a given command. For example, `flix run` should check for and download missing dependencies automatically.
- **Transparency:** Inform the user about which sub-actions will be performed and report on the progress of each.
- **Safety:** Actions that delete files or download dependencies should be performed safely. Always validate before deleting anything; if validation fails, prompt the user for permission.
- **Detailed Errors:** Error messages should be clear and actionable, similar to compiler errors, so users understand what went wrong and what to do next.

---

## Current Behavior

Current behavior and limitations of the package manager.

### Commands

| Command         | Description                                                         |
|-----------------|---------------------------------------------------------------------|
| `build`         | Compiles the current project.                                       |
| `build-jar`     | Builds a jar-file from the current project.                         |
| `build-fatjar`  | Builds a fatjar-file from the current project (includes dependencies). |
| `build-pkg`     | Builds a Flix package (`fpkg`) file from the current project.       |
| `doc`           | Generates API documentation for the project.                        |
| `run`           | Runs main entry point of the project.                               |
| `test`          | Runs the project's test suite.                                      |
| `check`         | Checks the project for errors (type, syntax, etc).                  |
| `init`          | Initializes a new Flix project in the current directory.            |
| `release`       | Publishes a release of the current project to GitHub.               |
| `outdated`      | Shows dependencies that have newer versions available.              |
| `repl`          | Starts a REPL for the current project or Flix source files.         |

---

## Bugs

- Old dependency versions are not removed. When upgrading a dependency, the old version is not removed. The package manager should suggest removing obsolete versions.
- Name clashes can happen. Since all namespaces `mod`s are visible, packages can declare the same module names and potentially cause compilation errors by declaring the same functions (even with different types). This should be resolved by prefixing or otherwise distinguishing modules.

---

## Stage 1 – Clarification

- [x] Come up with clear semantics of each command, including
  - Preconditions, e.g., structure of directory or required files
  - Output format or goal
  - Required sub-tasks, i.e., other commands.
- [x] `build`, `build-jar`, `build-fatjar`, `build-pkg` are unclear. We should clarify and figure out what users expect from `build`.
[...]

PR #11574, opened 2025-09-12
refactor: split `Bootstrap` into step functions

jaschdoc:
Related to #11436


main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -844,4 +816,85 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
       Validation.Success(true)
     }
   }
+
+  private object Steps {
```
magnus-madsen, line 769: Good its private, so this prevents the steps from leaking outside.

magnus-madsen:
I think it is important that this PR does not grow to a 1000 line diff. Can we do it more stepwise? (pun intended :))

jaschdoc:
Hehe yeah sure

jaschdoc:
@magnus-madsen I think now is a good time to review things. The entry level commands are bigger than what a quick refactor can handle, so I think those should be refactored in separate PRs one by one (`buildJarBase`, `buildPkg`, `release` `outdated`).

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -409,40 +410,17 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
...
-      case Err(e) => return Validation.Failure(BootstrapError.FlixPackageError(e))
-    }
-    MavenPackageManager.installAll(manifests, projectPath) match {
-      case Ok(l) => mavenPackagePaths = l
-      case Err(e) => return Validation.Failure(BootstrapError.MavenPackageError(e))
-    }
-    JarPackageManager.installAll(manifests, projectPath) match {
-      case Ok(l) => jarPackagePaths = l
-      case Err(e) => return Validation.Failure(BootstrapError.JarPackageError(e))
+    flatMapN(Steps.parseManifest) {
```
magnus-madsen: Since there is only one argument does Scalas foreach loop work here or no?

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -844,4 +739,200 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
...
+      */
+    def addLocalLibs(): List[List[Path]] = {
+      addLocalMavenJars() :: addLocalJars() :: addLocalFlixLibs() :: Nil
+    }
+
+    /**
+      * Returns and caches all `.jar` files from `lib/cache/`.
+      * The cached result is stored in [[mavenPackagePaths]].
+      */
+    private def addLocalMavenJars(): List[Path] = {
```
magnus-madsen, line 820: When is something private and when is it not? Should all steps not be private?

magnus-madsen (commented):
Structure looks reasonable. Now add comments?

Did you check that everything still seemingly works? Especially from the REPL and VSCode?

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
jaschdoc: For the dependency "stuff", it seems to me that it's nice to have helper functions but that if you want to add local libraries or install dependencies, you always want to do it for all three eco systems.

Also, the `Steps` object is private, but its members should be public. That way only `Bootstrap` may call its functions.

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
jaschdoc: We do not have `map` and `flatMap` implemented on the `Validation` trait itself, so I do not think that it's possible currently.

jaschdoc:
Seems to work fine from commandline, vscode and repl.

jaschdoc:
I will add doc for all step functions now.

jaschdoc:
RFR again

PR #11575, opened 2025-09-12
refactor: remove `Bootstrap` dependency from `Reader`

jaschdoc:
Related to #11436 

Removes `Bootstrap` from `Reader`. Added a helper function in `Flix` that checks for valid `.fpkg` files.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/ReducedAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/ReducedAst.scala
@@ -92,9 +92,13 @@ object ReducedAst {
 
     case class JumpTo(sym: Symbol.LabelSym, tpe: SimpleType, purity: Purity, loc: SourceLocation) extends Expr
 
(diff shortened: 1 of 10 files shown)


jaschdoc:
Amazing. I think I managed to break the Scala parser?

jaschdoc:
@magnus-madsen have you seen this error before?

magnus-madsen:
No never, seen it before.

magnus-madsen:
Can you say a few sentences about the refactor. What responsibility will go where?

jaschdoc:
Yes. The `Reader` phase depends on `Bootstrap.isPkgFile` to determine whether a `.fpkg` file can be read. It seems to me that the compilation pipeline must not depend on `Bootstrap`, but `Bootstrap` requires the compiler interface to work. I think the reason for this was that the author of `isPkgFile` knew that it existed and used it for a check in `Reader`, when in fact `flix` should perform those checks / be the source of truth.

jaschdoc:
So basically, this PR removes any mention of `Bootstrap` in `Reader` and adds the safety check in `Flix`. In fact, `Flix` already did most of the checks when adding inputs. The only thing missing was to check that the input was a valid zip archive.

magnus-madsen:
Thanks for the explanation.

PR #11579, opened 2025-09-12
feat: introduce colors for `Assert` messages

magnus-madsen:


magnus-madsen:
@mlutze Feel free to review in more detail.

One question is: Do we require `ToString` on `Result[e, t]` so we can print the unexpected value?

magnus-madsen:
I think we should require `ToString`-- we require it for the asserts and it seems to be the common case.

PR #11580, opened 2025-09-12
refactor: reduce memory usage of `VarSym`

magnus-madsen:


magnus-madsen:
@JonathanStarup @mlutze This saves about 1,000,000 instances of `Some` on the standard library test suite. 

There are still 250,000 instances left (probably from ascribe).

PR #11583, opened 2025-09-12
refactor: split ascribe

JonathanStarup:


magnus-madsen:
As it turns out, this was not the bottle neck for Some-memory usage. (That was my mistaken belief).

I still like the idea of splitting these.

But now I wonder about LSP interaction?

JonathanStarup:
If it doesn't do anything with performance lets just redo this more well-thought out at some other point

magnus-madsen:
> If it doesn't do anything with performance lets just redo this more well-thought out at some other point

Yeah. I was wrong. Must have misunderstood the debugger.

JonathanStarup:
all good, i also thought it'd be easier to split than it was

PR #11585, opened 2025-09-12
refactor: include `ByteCodeAst.Root` in `CompilationResult`

jaschdoc:
Includes the `ByteCodeAst` representation of the program, i.e., the compiled output in the `CompilationResult` type. The reason for this change is that `Bootstrap` may call `flix.compile()` without emitting class files. Instead, it directly obtains the byte code representation and can write it directly to a jar.

Related to #11436

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/Flix.scala
+++ b/main/src/ca/uwaterloo/flix/api/Flix.scala
@@ -766,8 +766,7 @@ class Flix {
     val loaderResult = JvmLoader.run(bytecodeAst)
 
     // Construct the compilation result.
-    val totalSize = bytecodeAst.classes.values.map(_.bytecode.length).sum
-    val result = new CompilationResult(loaderResult.main, loaderResult.tests, loaderResult.sources, totalTime, totalSize)
+    val result = new CompilationResult(loaderResult.main, loaderResult.tests, loaderResult.sources, bytecodeAst, totalTime)
 
     // Shutdown fork-join thread pool.
     shutdownForkJoinPool()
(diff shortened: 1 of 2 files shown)


magnus-madsen:
The problem with this is the amount of memory it holds on to.

magnus-madsen:
Is there any reason the compiler cannot just emit the classes as it currently does?

jaschdoc:
No, except it's a bit more impure and not what I understood from how we wanted to design the `Bootstrap` class. Did it previously hold on to the root? In other words, was it removed from `CompilationResult` recently in your recent efforts to reduce memory usage?

jaschdoc:
But it is certainly still doable if it first writes to the file system

magnus-madsen:
I guess the question is who should be responsible for writing bytecode to disk. 

I will think a bit about it. But for now, I don't think the status quo presents any obstacles?

jaschdoc:
Sounds good

JonathanStarup:
A potential option is to add a `FlixEvent` for the JvmBackend phase

jaschdoc:
> A potential option is to add a `FlixEvent` for the JvmBackend phase

Seems like a good solution. I can try to code that up. What do you think @magnus-madsen

magnus-madsen:
The EventListener mechanism should not be used for semantics.

magnus-madsen:
I think your goal is to bypass the filesystem entirely. But I am not convinced this is a good idea. 

In any case, it seems unnecessary. The build step will emit class files and build-jar / build-fatjar will bundle them together.

magnus-madsen:
Overall, I am neutral on the question of where to emit class files. I just don't think this is the right time to move that responsibility out of the Flix compiler. In particular, I think we can refactor the package manager (at least iteration 1) without making invasive changes to the rest of the compiler.

jaschdoc:
Sounds good. I will close this for now then.

PR #11587, opened 2025-09-12
refactor: cleanup `Assert`

magnus-madsen:


main/src/library/Assert.flix
```diff
@@ -163,32 +163,35 @@ mod Assert {
...
                 if (cond)
-                    resume(())
+                    resume()
                 else {
-                    run {
-                        println("");
-                        println(RichString.toAnsiString(msg));
-                        println("");
-                        throw new AssertionError(RichString.toString(msg))
-                    } with Environment.runWithIO
```
magnus-madsen, line 179: This effect was not used, so it was just contributing extra code for no reason.

PR #11591, opened 2025-09-13
test: refactor library unit tests to use better assertions

magnus-madsen:


magnus-madsen:
@mlutze I know some of these can be further improved. I am just trying to take small steps and do several passes.

mlutze (approved):

PR #11594, opened 2025-09-13
refactor: remove `tpe` field from `ReducedAst.Cst`

magnus-madsen:


main/src/ca/uwaterloo/flix/language/phase/jvm/GenExpression.scala
```diff
@@ -98,14 +98,13 @@ object GenExpression {
...
   def compileExpr(exp0: Expr)(implicit mv: MethodVisitor, ctx: MethodContext, root: Root, flix: Flix): Unit = exp0 match {
-    case Expr.Cst(cst, tpe, loc) => cst match {
+    case Expr.Cst(cst, loc) => cst match {
       case Constant.Unit =>
         BytecodeInstructions.GETSTATIC(BackendObjType.Unit.SingletonField)
 
       case Constant.Null =>
         import BytecodeInstructions.*
         ACONST_NULL()
-        castIfNotPrim(BackendType.toBackendType(tpe))
```
magnus-madsen, line 108: No cast is needed because at the JVM bytecode level null is an instance of every reference type.

(And a cast to a primitive would be a verifier error.)

JonathanStarup (approved):

PR #11595, opened 2025-09-13
feat: remove Interpolation nesting threshold

JonathanStarup:
- No reason to arbitrarily limit it
- other places in the lexer don't have this recursion counter (block comments, numbers, keywords, ..)

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/errors/LexerError.scala
+++ b/main/src/ca/uwaterloo/flix/language/errors/LexerError.scala
@@ -134,24 +134,6 @@ object LexerError {
   }
 
   /**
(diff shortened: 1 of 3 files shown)


magnus-madsen:
There was a reason for it. I just cannot remember... Maybe the reason is gone.

But Chesterton's fence?

JonathanStarup:
I was trying to think of its reason but could only come up with it being a guard against runaway recursion

magnus-madsen:
> I was trying to think of its reason but could only come up with it being a guard against runaway recursion

Can that happen now?

JonathanStarup:
No each loop always calls advance

magnus-madsen (approved):

PR #11601, opened 2025-09-14
feat: end col lexer verification

JonathanStarup:
For a program like
```
123
42
```
This verifies that `123` has range `[1:1, 1:4[` instead of `[1:1, 2:1[`. The lexer does this to make ranges touch as few lines as possible.

This is accepted with full FlixSuite enabled


magnus-madsen (approved):

magnus-madsen:
👍

PR #11603, opened 2025-09-15
fix: replace backticks with single quotes in prints

jaschdoc:
`Bootstrap` uses `'` quoted strings in all printed messages except here.


magnus-madsen:
Indeed we always use single quotes. Also in compiler errors.

PR #11604, opened 2025-09-15
feat: check file extension input in `Bootstrap`

jaschdoc:
The function `getAlFilesWithExt` could easily be misunderstood, so I added a check to ensure it is used correctly. This can also be done via a `Validation` type. What do you think is best?

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/Bootstrap.scala
+++ b/main/src/ca/uwaterloo/flix/api/Bootstrap.scala
@@ -317,8 +317,10 @@ object Bootstrap {
   /**
     * Returns all files in the given path `p` ending with .`ext`.
     */
-  private def getAllFilesWithExt(p: Path, ext: String): List[Path] =
+  private def getAllFilesWithExt(p: Path, ext: String): List[Path] = {
+    require(!ext.startsWith("."), "The file extension must not start with '.' -- This is handled by `getAllFilesWithExt`.")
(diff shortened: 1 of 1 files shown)


main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -316,9 +316,14 @@ object Bootstrap {
 
   /**
     * Returns all files in the given path `p` ending with .`ext`.
+    *
+    * @param p   the path from which files a considered.
+    * @param ext the file extension to match. Must not begin with `.`
     */
-  private def getAllFilesWithExt(p: Path, ext: String): List[Path] =
+  private def getAllFilesWithExt(p: Path, ext: String): List[Path] = {
+    require(!ext.startsWith("."), "The file extension must not start with '.' -- This is handled by `getAllFilesWithExt`.")
```
magnus-madsen: Scala uses require not assert?

magnus-madsen:
Not Validation, this is a programming bug. So assert/expect is fine.

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
jaschdoc: `require` is used for preconditions on parameters.

jaschdoc: Throws invalid argument exception if the conditions does not hold.

magnus-madsen (approved):

PR #11606, opened 2025-09-15
refactor: merge file operations in `Bootstrap` with `FileOps`

jaschdoc:
This PR removes file operations from `Bootstrap` and merges them into `FileOps`. The reason is threefold:
1. `Bootstrap` had many functions related to file operations. Removing them from the class allows it to focus more cleanly on what it must achieve, i.e., less clutter more meaningful code.
2. `Bootstrap` and `FileOps` had code duplication, so this PR improves code reuse.
3. `isZipArchive` was used in both `Bootstrap` and `Flix` but defined twice. As it seems to be a predicate that relies solely on a file and its first four bytes, it seems natural to move it to `FileOps`. This improves code reuse again.


magnus-madsen:
Good idea.

PR #11610, opened 2025-09-15
feat: avoid array clone in lexer fuzzing

JonathanStarup:
There is no reason to copy the array, it is a local array so it can just be mutated.

This makes fuzzing a constant time operation

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
@@ -142,8 +142,8 @@ object Lexer {
       val (results, errors) = ParOps.parMap(staleByDecreasingSize) {
         src =>
           val (tokens, errors) = lex(src)
-          val fuzzedTokens = fuzz(tokens)
-          (src -> fuzzedTokens, errors)
+          fuzz(tokens)
+          (src -> tokens, errors)
       }.unzip
 
       // Construct a map from each source to its tokens.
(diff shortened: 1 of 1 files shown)


JonathanStarup:
@magnus-madsen is this actually used? its never set to `true` in this repo (CI or otherwise) and it seems redundant when `FuzzerSuite` exists?

magnus-madsen:
But its fine to copy the tokens. I don't understand. Why micro optimize?

magnus-madsen:
As for its use, I think it is waiting for the time when we get a server dedicated to fuzzing.

JonathanStarup:
> But its fine to copy the tokens. I don't understand. Why micro optimize?

micro optimize makes it sound like a make the function worse to optimize performance. This change makes the function more general and faster. Why not do it?

magnus-madsen:
> > But its fine to copy the tokens. I don't understand. Why micro optimize?
> 
> micro optimize makes it sound like a make the function worse to optimize performance. This change makes the function more general and faster. Why not do it?

Because it makes the correctness of the code less clear? And there is no measurable performance gain here right?

magnus-madsen:
Its written in a functional-style right?

JonathanStarup:
its not more or less clear to me, but now its simpler and more efficient.

JonathanStarup:
I can also add back the call to `array.clone()` on the outside. I just wanted to make the fuzzer function itself as clean and simple as possible.

magnus-madsen:
I want `fuzz` to be a function.

JonathanStarup:
That is hard to argue against.

Can I counter propose removing it since its 1) quite simple, 2) dead code, and 3) is testing code that exists in core logic?

Issue #11514, opened 2025-09-05
Add missing `case` completion suggestion when in match expression


LoZander:
When in a match expression, such as
```flix
match x {
  c|
}
```
where | denotes the cursor, only the `catch` keyword is suggested and not `case`.

While it would also be nice for `catch` to not be suggested in this case (since I believe it would never be applicable in this context), I feel this is being the scope of this issue.

LoZander:
Unless keyword completion has changed considerably since I worked on this, I believe the solution is as simple as adding `case` to the list of keyword completions returned by `KeywordCompleter.getExprKeywords`. I could probably do PR sometime this weekend.

LoZander:
Or maybe monday

LoZander:
A further improvement would be to utilise syntactic context to only suggest `case` when on the left-hand side of a rule in a match expression (this would also allow ruling out other suggestions that do not make sense here. But with  this PR, `case` will at least be suggested to begin with within the match expression.

LoZander:
As discussed in #11613, the proper solution is to add a syntactic context for match rules and add a `case` completion when in that context. I might look into that when I find the time, unless someone else gets to it first ofc :)

magnus-madsen:
The situation should be very similar to e.g. the enum case.

PR #11613, opened 2025-09-15
fix: added missing `case` expression keyword completion

LoZander:
Related to issue #11514

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/KeywordCompleter.scala
+++ b/main/src/ca/uwaterloo/flix/api/lsp/provider/completion/KeywordCompleter.scala
@@ -88,6 +88,7 @@ object KeywordCompleter {
       // A
       Completion.KeywordCompletion("and"         , range, Priority.Lowest(0)),
       // C
+      Completion.KeywordCompletion("case"        , range, Priority.Lowest(-1)),
       Completion.KeywordCompletion("catch"       , range, Priority.Lowest(0)),
       // D
       Completion.KeywordCompletion("def"         , range, Priority.Lowest(-1)),


LoZander:
Since everything else in the file has the absolute priority of "lowest", I gave this keyword completion that same. I wasn't exactly sure what to set the relative priority to, but I figured it should have priority over `catch`, so I set it to `-1` since I saw that had been done elsewhere in the completer.

magnus-madsen:
@LoZander `case` is not a syntactic expression; rather it is part of a MatchRule (among others) and should have its own syntactic category. Adding it to expression will cause incorrect completions.

magnus-madsen:
We have an enum context. In the same style, we could have a match context.

LoZander:
> @LoZander `case` is not a syntactic expression; rather it is part of a MatchRule (among others) and should have its own syntactic category. Adding it to expression will cause incorrect completions.

@magnus-madsen. Right, better to be conservative with the suggestion rather than too liberal with it, since incorrect suggestions are misleading.

Would it be a big job to add a match context, do you think? I could try looking into that

magnus-madsen:
> > @LoZander `case` is not a syntactic expression; rather it is part of a MatchRule (among others) and should have its own syntactic category. Adding it to expression will cause incorrect completions.
> 
> @magnus-madsen. Right, better to be conservative with the suggestion rather than too liberal with it, since incorrect suggestions are misleading.
> 
> Would it be a big job to add a match context, do you think? I could try looking into that

It should not bee too big.

LoZander:
> > > @LoZander `case` is not a syntactic expression; rather it is part of a MatchRule (among others) and should have its own syntactic category. Adding it to expression will cause incorrect completions.
> > 
> > 
> > @magnus-madsen. Right, better to be conservative with the suggestion rather than too liberal with it, since incorrect suggestions are misleading.
> > Would it be a big job to add a match context, do you think? I could try looking into that
> 
> It should not bee too big.

alright, cool. Then I might take a look at that sometime soon, when I find the time :)

PR #11614, opened 2025-09-15
refactor: remove simple alias

JonathanStarup:
I think its more readable with the stringcurser call

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
@@ -145,7 +145,7 @@ object Lexer {
   def lex(src: Source): (Array[Token], List[LexerError]) = {
     implicit val s: State = new State(src)
     while (!eof()) {
-      consumeWhitespace()
+      s.sc.advanceWhile(_.isWhitespace)
       if (!eof()) {
         s.resetStart()
         val k = scanToken()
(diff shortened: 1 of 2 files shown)


jaschdoc:
Not sure I agree. Why do you think it is more readable?

magnus-madsen:
I think the current code is more readable.

JonathanStarup:
if you read comsumeWhitespace, you don't know what that function does. If you read s.sc.advanceWhile(_.isWhitespace) you know exactly what it does. Normally the fault in this logic is the size of the function code but here they are the same size.

magnus-madsen:
@mlutze What is your vote?

magnus-madsen:
The nice thing about having one function is that you can see the definition in one place. No chance of drift.

JonathanStarup:
In this case though, the function is just hiding the call and affect on `s.sc`, and the whole thing with Lexer is about the state of the stringcursor.

Matt went from against to neutral after I presented my points 😄

I made this change because i read the `consumeWhitespace` call and had to go away to read what it actually did. I think it easy to forget that the reader of the code does not neccessarily know/remember that it is a one line function. I don't like these functions that adds clutter to selfdescribing code.

jaschdoc:
I still think `s.sc.advanceWhile(_.isWhitespace)` is a lower level implementation of *how* to consume whitespace. When adding new functionality, it's much more cumbersome to *learn how* to consume whitespace rather than *discovering* a function that tells me *what* it does.

magnus-madsen (approved):

magnus-madsen:
I disagree. But let us not overthink it. Since you are the one working on the Lexer atm, I will let it go.

JonathanStarup:
> I still think `s.sc.advanceWhile(_.isWhitespace)` is a lower level implementation of _how_ to consume whitespace. When adding new functionality, it's much more cumbersome to _learn how_ to consume whitespace rather than _discovering_ a function that tells me _what_ it does.

but "lower" is relative, all the lexer functions operate on that level (do xyz to the string cursor)

> I disagree. But let us not overthink it. Since you are the one working on the Lexer atm, I will let it go.

Yeah i was also gonna suggest that we conclude no matter the judgement 😄

mlutze:
> @mlutze What is your vote?

I was originally opposed but Jonathan brought me to neutrality

PR #11617, opened 2025-09-16
feat: add `Debug` effect and prepare code for new type rules

magnus-madsen:


magnus-madsen:
I also fixed a lot of Intellij IDEA warnings.

main/src/library/Debug.flix
```diff
@@ -11,6 +32,24 @@ mod Debug {
...
+
+    ///
+    /// Prints `x` to the standard out followed by a new line.
+    ///
+    pub def dprintln(x: a): Unit \ Debug with ToString[a] =
+        unchecked_cast({
+            let s: String = ToString.toString(x);
+            System.out.println(s)
+        } as _ \ Debug)
+
```
mlutze, line 52: These should print to stderr, not stdout.

Rust agrees: https://doc.rust-lang.org/std/macro.dbg.html

magnus-madsen: Can we debate that in a ticket? :)

magnus-madsen: https://github.com/flix/flix/issues/11618

mlutze (approved):

PR #11621, opened 2025-09-16
feat: allow implicit `debug` effect in type signatures

magnus-madsen:


magnus-madsen:
@mlutze Do give suggestions for how to improve the code.

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolverInterface.scala
```diff
@@ -102,7 +102,28 @@ object ConstraintSolverInterface {
...
-      (subst, mkTypeErrors(leftovers, subst, renv, root))
+      leftovers match {
+        case Nil =>
+          // All constraints solved. Yay!
+          (subst, Nil)
+        case _ =>
+          // We have one or more type errors. We need to report them, but first we have special logic for the [[Debug]] effect.
+
+          // We solve a new constraint system where [[Debug]] is allowed by the effect signature.
+          val declaredEffConstrWithDebug = TypeConstraint.Equality(Type.mkUnion(eff, Type.Debug, loc), infEff, Provenance.ExpectEffect(expected = eff, actual = infEff, loc))
```
mlutze: Let's use the unioned type for the provenance expected.

mlutze:
I wonder if it would be nice to extract some of it to a helper function.

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolverInterface.scala
magnus-madsen: Fixed.

magnus-madsen:
> I wonder if it would be nice to extract some of it to a helper function.

Yeah, but how? it is also only two (complex) lines.

PR #11623, opened 2025-09-16
refactor: remove special keyword lexing

JonathanStarup:
This special casing means that unlike all other keywords, you cannot write `x.null`, `x.true` or `x.false`. This seems like a needless distinction.


magnus-madsen:
Could you explain more? I don't understand the situation before nor the situation after.

JonathanStarup:
Sorry I just changed it. This makes it such that `true`, `false`, and `null` are treated like all the other keywords.

So in this PR you can write `something.true()` as a java method call, just like `something.not()`. Keyword lexing is not perfect but making these three keywords similar to all the others make sense to me.

As the comment says "We are assuming no literal keyword needs to be imported" but this assumption does not help us in any way so there is no point in making it.

PR #11624, opened 2025-09-16
feat: add `FILE` and `LINE` intrinsics

magnus-madsen:


magnus-madsen:
@JonathanStarup @mlutze The implementation is simple enough and does not really interact much with the language at all.

On the other hand, it seems cumbersome to use.

JonathanStarup:
Alternative:

Add `$POS$` and have weeder turn that into `Debug.prettyPos($FILE$, $Line$)`

We control the fomatting and can do it in Flix

mlutze:
Jonathan has promised to clean this up later.

PR #11626, opened 2025-09-16
feat: harden keywords in lexer

JonathanStarup:
current keywords are interpreted as names if prefixed by `.`. This means that `something.and` does not see `and` as a keyword. I propose that we remove this edge case since we have `$and` to allow keywords as names.

| Input | Before | PR | Fix |
| --- | --- | --- | --- |
| `x.not(y)` |  accepted | `expected <name> instead of 'not'` | `x.$not(y)` |

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
@@ -445,12 +445,11 @@ object Lexer {
 
   /**
     * Check that the potential keyword is sufficiently separated.
-    * A keyword is separated if it is surrounded by anything __but__ a character, digit a dot or
-    * underscore.
+    * A keyword is separated if it is surrounded by anything __but__ a character, digit, or underscore.
     * Note that __comparison includes current__.
     */
   private def isSeparated(keyword: String)(implicit s: State): Boolean = {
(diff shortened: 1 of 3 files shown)


main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -445,12 +445,11 @@ object Lexer {
...
-    * A keyword is separated if it is surrounded by anything __but__ a character, digit a dot or
-    * underscore.
+    * A keyword is separated if it is surrounded by anything __but__ a character, digit, or underscore.
     * Note that __comparison includes current__.
     */
   private def isSeparated(keyword: String)(implicit s: State): Boolean = {
-    def isSep(c: Char) = !(c.isLetter || c.isDigit || c == '_' || c == '.')
+    def isSep(c: Char) = !(c.isLetter || c.isDigit || c == '_')
 
     s.sc.nthIsPOrOutOfBounds(-2, isSep) && s.sc.nthIsPOrOutOfBounds(keyword.length - 1, isSep)
```
JonathanStarup, line 454: This PR means that in one more PR we can avoid the `-2` lookup for keywords

JonathanStarup: (which is one of the hottest spots)

PR #11628, opened 2025-09-16
refactor: separate user op lexing special case

JonathanStarup:
pulling out the weird special case

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
@@ -427,13 +427,13 @@ object Lexer {
       case '0' if s.sc.peekIs(_ == 'x') => acceptHexNumber()
       case c if c.isDigit => acceptNumber()
       // User defined operators.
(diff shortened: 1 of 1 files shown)


JonathanStarup:
Can just close if you disagree, I don't mind

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -427,13 +427,13 @@ object Lexer {
       case '0' if s.sc.peekIs(_ == 'x') => acceptHexNumber()
       case c if c.isDigit => acceptNumber()
       // User defined operators.
+      case '<' if s.sc.peekIs(_ == '>') && peekPeek().flatMap(isUserOp).isEmpty =>
```
magnus-madsen, line 430: Whats the order in all of these cases?

magnus-madsen (approved):

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
JonathanStarup: not sure, seems a bit ad hoc. I was considering reordering them but i will do the more "pure" stuff first

PR #11630, opened 2025-09-16
refactor: avoid EOF peeks

JonathanStarup:
`peekIs(_ == ..)` is better since EOF is explicitly handled, instead of operating on the null char

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
@@ -445,12 +445,11 @@ object Lexer {
 
   /**
     * Check that the potential keyword is sufficiently separated.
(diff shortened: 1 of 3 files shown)


main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -768,8 +766,7 @@ object Lexer {
     while (!eof()) {
       consumeSingleEscapes()
       // Note: `sc.peek` returns `EOF` if out of bounds, different from `peek`.
```
mlutze: Is this note needed?

JonathanStarup: good point, deleted

PR #11631, opened 2025-09-16
refactor: split StringCursor.advance

JonathanStarup:
The advance that returns the char is somewhat unsafe because it returns a null char.

- peekAndAdvance is unsafer and is only used 2 places
- advance is safe and is used 20 places

I think its good to isolate the unsafer uses.


magnus-madsen:
@JonathanStarup Conflicted.

PR #11637, opened 2025-09-16
Default handlers for effects

NeonOxide:
Added default handlers for effects and removed some special treatment for Assert. 

Everything kind of works more or less. However there are a couple of things still not working, so this is more of a draft PR, also quite big, so maybe it should be split at some point.

The two main issues right now are:

- Default handlers do not work with :eval. Already found the root issue, it has to do with typing happening before EntryPoints.scala
- For some reason I have been unable to pinpoint, when a correct default handler definition is made, VSCode turns the whole definition to a yellowish color. No idea why.

The requirements for defining a default handler are:

- Must be in the companion module of its effect.
- There must not be more than one default effect handler per effect.
- Must be annotated with @DefaultHandler
- Its signature must be (without having to choose the same names): def myFunc(f: Unit -> a \ ef): a \ (ef - handledEffect) + primEff1 + primEff2 + ... It is important to note that the final effect set is checked syntactically so it must be written in that order.  

The error messages have been chosen a little bit arbitrarily, I am quite open to suggestions. I think that this PR will take a couple of iterations to get off the ground.


NeonOxide:
There may be some style issues given that I had to do quite a big merge but I think I cleaned up almost everything (at least everything I saw)

NeonOxide:
Also, as an starting point to test the feature. I use this code:

```
use A.TestEff


mod A{
    pub eff TestEff{
        def doSomething(): Unit
    }

    mod TestEff {
    pub eff B
    @DefaultHandler
    def _func(f: Unit -> a \ ef): a \ (ef - TestEff) + IO + Sys  = 
        checked_ecast(
            run {
                f()
            } with handler TestEff {
                def doSomething(resume) = {
                    println("DEFAULT BEHAVIOUR");
                    resume()
                }
                
            }
        )
}

}

def main(): Unit \ IO + TestEff = 
    TestEff.doSomething();
    println("Hello world")


@Test
def myTest(): Unit \ TestEff = TestEff.doSomething()
```

NeonOxide:
I may had forgotten to run the tests after everything and I did not notice that I had not added a default handler for Assert. That should be fixed now. I should also at some point revisit the logic for visitTest and remove the special logic for Assert (the compiler was throwing unhandled effect because Tests relied on Assert being handled, which they now are, but there is no need for leaving the door open to future unhandled effects) as it is no longer needed but it is 1 am, I will do it tomorrow. 

I also thought of a (non quite efficient) but quite convenient way of making shell1 work (spoiler, it is kind of the same thing I did with Assert but for every handler)

NeonOxide:
I do not know what the fuzz autocomplete error mean, I will solve it tomorrow as well if I can

magnus-madsen:
> Default handlers do not work with :eval. Already found the root issue, it has to do with typing happening before EntryPoints.scala

My suggestion is to change the REPL to create:

```
s"""def ${main.name}(): Unit \\ $effString =
   |def f() = { println($s) };
   |checked_ecast(())
   |""".stripMargin
```

then you can rewrite that in `EntryPoints`.

> For some reason I have been unable to pinpoint, when a correct default handler definition is made, VSCode turns the whole definition to a yellowish color. No idea why.

This means your source locations are incorrect (or you have source locations for generated code that is not marked synthetic)

> The requirements for defining a default handler are:

Looks good!

magnus-madsen:
I suggest we break this down into smaller PRs that can we can merge and iterate on quickly. Here is my proposal:

1. Add `@DefaultHandler` + AnnotationCompleter and all code that touches on the annotation. This should be simple and clean.
2. Add EntryPointErrors. This will allow us to understand the design space of what can go wrong and bikeshed on the messages.

From here we can either:
3. Add new checking code + tests, or 
4. Add new wrapping code.

Sounds good?

EDIT: We can also do a PR that refactors how the Shell generates code. This would not touch handlers, but just prepare the ground for how to add them. That way refactoring can be kept separate from new functionality.

NeonOxide:
> > For some reason I have been unable to pinpoint, when a correct default handler definition is made, VSCode turns the whole definition to a yellowish color. No idea why.
> 
> This means your source locations are incorrect (or you have source locations for generated code that is not marked synthetic)

Are you sure there cannot be another issue that generates this? I'll look into it this evening after work, but it seems weird to me that the error is happening in the handler definition where no code rewriting is happening.

NeonOxide:
Also in general, what is the best approach of debugging VsCode and Lsp issues?

magnus-madsen:
> Also in general, what is the best approach of debugging VsCode and Lsp issues?

I forget how, but you can ask VScode to show you the semantic tokens it gets.

NeonOxide:
> I suggest we break this down into smaller PRs that can we can merge and iterate on quickly. Here is my proposal:
> 
>     1. Add `@DefaultHandler` + AnnotationCompleter and all code that touches on the annotation. This should be simple and clean.
> 
>     2. Add EntryPointErrors. This will allow us to understand the design space of what can go wrong and bikeshed on the messages.
> 
> 
> From here we can either: 3. Add new checking code + tests, or 4. Add new wrapping code.
> 
> Sounds good?
> 
> EDIT: We can also do a PR that refactors how the Shell generates code. This would not touch handlers, but just prepare the ground for how to add them. That way refactoring can be kept separate from new functionality.

Sounds good, I will get onto it and update how it goes. The shell generation of code however I think it will require much more work just due to the sheer flexibility of it.

magnus-madsen:
(PR outdated)

PR #11638, opened 2025-09-17
feat: add preliminary support for debug strings

magnus-madsen:


magnus-madsen:
@mlutze I would like this merged now-- so I can make a new Flix release :)

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -360,6 +360,7 @@ object Lexer {
       case _ if isKeyword("Map#") => TokenKind.MapHash
       case _ if isKeyword("List#") => TokenKind.ListHash
       case _ if isKeyword("Vector#") => TokenKind.VectorHash
+      case _ if isMatchPrev("d\"") => acceptDebugString()
```
mlutze, line 363: I can accept the single-letter prefix if this becomes a DebugString type. Otherwise it needs to be uglier to discourage abuse.

magnus-madsen: Create a ticket with some proposals :)

magnus-madsen: This feature is by no means finalized. I just want something we can experiment with.

PR #11643, opened 2025-09-17
feat: add `Build` option to `Options`

magnus-madsen:


main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -327,9 +327,9 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
   /**
     * Builds (compiles) the source files for the project.
     */
-  def build(flix: Flix): Validation[CompilationResult, BootstrapError] = {
+  def build(flix: Flix, build: Build = Build.Development): Validation[CompilationResult, BootstrapError] = {
```
magnus-madsen, line 330: This is really just a first attempt. As you rework the package manager, you will have to keep track of this more precisely.

But I think it should be fairly straight forward.

jaschdoc (approved):

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
mlutze: default parameters bad

PR #11646, opened 2025-09-17
feat: refactor `Bootstrap.buildJarBase` into step functions

jaschdoc:
Related to https://github.com/flix/flix/issues/11436

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/Bootstrap.scala
+++ b/main/src/ca/uwaterloo/flix/api/Bootstrap.scala
@@ -348,11 +348,10 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
       _ <- Steps.compile(flix)
       _ <- Steps.validateJarFile(jarFile)
       contents = (zip: ZipOutputStream) => {
-        Steps.addManifestToZip(zip)
         Steps.addClassFilesFromDirToZip(Bootstrap.getClassDirectory(projectPath), zip)
         Steps.addResourcesFromDirToZip(Bootstrap.getResourcesDirectory(projectPath), zip)
       }
-      _ <- Steps.createZip(jarFile, contents)
+      _ <- Steps.createJar(jarFile, contents)
     } yield {
       ()
     }
@@ -373,12 +372,11 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
       _ <- Steps.validateDirectory(libDir)
       _ <- Steps.validateJarFilesIn(libDir)
       contents = (zip: ZipOutputStream) => {
-        Steps.addManifestToZip(zip)
         Steps.addClassFilesFromDirToZip(Bootstrap.getClassDirectory(projectPath), zip)
         Steps.addResourcesFromDirToZip(Bootstrap.getResourcesDirectory(projectPath), zip)
         Steps.addJarsFromDirToZip(libDir, zip)
       }
-      _ <- Steps.createZip(jarFile, contents)
+      _ <- Steps.createJar(jarFile, contents)
     } yield {
       ()
     }
@@ -702,7 +700,7 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
     /**
       * Adds a `META-INF/MANIFEST.MF` file to `zip`.
       */
-    def addManifestToZip(zip: ZipOutputStream): Unit = {
+    private def addManifestToZip(zip: ZipOutputStream): Unit = {
       // META-INF/MANIFEST.MF
       val manifest =
         """Manifest-Version: 1.0
@@ -767,13 +765,21 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
     }
 
     /**
-      * Writes `contents` to the zip file located at `zip`.
+      * Writes `contents` to the jar file located at `jar`.
       *
-      * Creates the zip file if it does not exist, and truncates it if it already exists.
+      * This function also adds a manifest to the jar file.
+      *
+      * Creates the jar file if it does not exist, and truncates it if it already exists.
+      *
+      * @see [[Steps.addManifestToZip]]
       */
-    def createZip(zip: Path, contents: ZipOutputStream => Unit): Result[Unit, BootstrapError.FileError] = {
-      Files.createDirectories(zip.getParent.normalize())
-      Result.fromTry(Using(new ZipOutputStream(Files.newOutputStream(zip)))(contents))
+    def createJar(jar: Path, contents: ZipOutputStream => Unit): Result[Unit, BootstrapError.FileError] = {
+      Files.createDirectories(jar.getParent.normalize())
+      val contentsWithManifest = (zip: ZipOutputStream) => {
+        Steps.addManifestToZip(zip)
+        contents(zip)
+      }
+      Result.fromTry(Using(new ZipOutputStream(Files.newOutputStream(jar)))(contentsWithManifest))
         .mapErr(e => BootstrapError.FileError(e.getMessage))
     }
 
@@ -939,7 +945,7 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
     /**
       * Returns `Ok(())` if all files ending with `.jar` in `dir` are valid jar files.
       *
-      * @see [[validateJarFile]]
+      * @see [[Steps.validateJarFile]]
       */
     def validateJarFilesIn(dir: Path): Result[Unit, BootstrapError] = {
       Result.traverse(FileOps.getFilesWithExtIn(dir, EXT_JAR, Int.MaxValue))(Steps.validateJarFile).map(_ => ())


main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -811,6 +814,18 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
...
+      val zip: (ZipOutputStream => Unit) => Try[Unit] = Using(new ZipOutputStream(Files.newOutputStream(zipFile)))
+      val callback = (contents: ZipOutputStream => Unit) => {
+        zip(contents) match {
+          case Success(()) => Validation.Success(())
+          case Failure(exception) => Validation.Failure(BootstrapError.FileError(exception.getMessage))
+        }
+      }
+      Validation.Success(callback)
+    }
+
```
jaschdoc: Should this return a validation or just a callback? I.e.,
```scala
Validation[(ZipOutputStream => Unit) => Validation[Unit, BootstrapError], BootstrapError]
```
or just
```scala
(ZipOutputStream => Unit) => Validation[Unit, BootstrapError]
```
since the result is always `Validation.Success`?

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -839,7 +854,7 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
       * If they have, they are added to flix. Then updates the timestamps
       * map to reflect the current source files and packages.
       */
-    def updateStaleSources(flix: Flix): Validation[Flix, BootstrapError] = {
+    def updateStaleSources(flix: Flix): Validation[Unit, BootstrapError] = {
```
jaschdoc: Should this be a Validation since the result is always Success?

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -869,8 +884,28 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
...
+      if (Files.exists(dir)) {
+        if (!Files.isDirectory(dir)) {
+          return Validation.Failure(BootstrapError.FileError(s"The path '${formatter.red(dir.toString)}' is not a directory."))
+        }
+        if (!Files.isReadable(dir)) {
+          return Validation.Failure(BootstrapError.FileError(s"The path '${formatter.red(dir.toString)}' is not readable."))
+        }
+      }
+      Validation.Success(())
+    }
```
jaschdoc: Should this be a Validation since the result is always Success?

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -869,8 +884,28 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
...
+
+    def validateOutJarFile(jarFile: Path)(implicit formatter: Formatter): Validation[Unit, BootstrapError] = {
+      // Check whether it is safe to write to the file.
+      if (Files.exists(jarFile) && !Bootstrap.isJarFile(jarFile)) {
+        Validation.Failure(BootstrapError.FileError(s"The path '${formatter.red(jarFile.toString)}' exists and is not a jar-file. Refusing to overwrite."))
+      } else {
+        Validation.Success(())
+      }
+    }
```
jaschdoc: Should this be a Validation since the result is always Success?

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
magnus-madsen: Why does it need such a complex type signature?

jaschdoc: It wraps the `Using` type in a `Validation` so it can be used in a flatmap. It partially applies `Using` which returns a function that gives access to the zip file. However, the result of that function is a `Try` which is then mapped the a `Validation`.

magnus-madsen: But I don't think such a design is good...

jaschdoc: `mkZipFile` could also mirror the interface of `Using`:
```scala
def mkZipFile(file: => ZipOutputStream)(contents: ZipOutputStream => Unit): Validation[Unit, BootstrapError] = {
  Using(file)(contents) match {
    case Success(()) => Validation.Success(())
    case Failure(e) => Validation.Failure(BootstrapError.FileError(e.getMessage))
  }
}
```

magnus-madsen: I would just avoid HOFs here if possible.

jaschdoc: And then just call `Using` directly at the use site?

jaschdoc:
I still need to add docs, but a review would be nice :)

jaschdoc:
Note that in the future, build jar will always include dependencies, so all the special casing with `if (includeDependencies) ...` will go away.

magnus-madsen:
I am not really sure what we are trying to achieve here. When I look at the old code, it almost seems cleaner to me?

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -343,84 +343,38 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
...
-          Using(new ZipInputStream(Files.newInputStream(dep))) {
-            zipIn =>
-              var entry = zipIn.getNextEntry
-              while (entry != null) {
-                // Get the class files except module-info and META-INF classes which are specific to each library.
-                if (entry.getName.endsWith(s".$EXT_CLASS") && !entry.getName.equals(s"module-info.$EXT_CLASS") && !entry.getName.contains("META-INF/")) {
-                  // Write extracted class files to zip.
-                  val classContent = zipIn.readAllBytes()
-                  FileOps.addToZip(zip, entry.getName, classContent)
+    flatMapN(Steps.updateStaleSources(flix)) {
```
magnus-madsen: Can we not use a for-comprehension?

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -343,84 +343,38 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
...
+      _ =>
+        flatMapN(Steps.configureJarOutput(flix)) {
+          _ =>
+            flatMapN(Steps.compile(flix)) {
+              _ =>
+                val jarFile = Bootstrap.getJarFile(projectPath)
```
magnus-madsen: It seems the steps are interleaved with code?

jaschdoc:
> I am not really sure what we are trying to achieve here. When I look at the old code, it almost seems cleaner to me?

If leave out the dependency stuff, then it reduces to:
```scala
  /**
    * Builds the jar or the fatjar
    */
  private def buildJarBase(flix: Flix, includeDependencies: Boolean)(implicit formatter: Formatter): Validation[Unit, BootstrapError] = {
    val jarFile = Bootstrap.getJarFile(projectPath)
    flatMapN(Steps.updateStaleSources(flix)) {
      _ =>
        flatMapN(Steps.configureJarOutput(flix)) {
          _ =>
            flatMapN(Steps.compile(flix)) {
              _ =>
                flatMapN(Steps.validateJarFile(jarFile)) {
                  _ =>
                    Files.createDirectories(getArtifactDirectory(projectPath))
                    val outputStream = new ZipOutputStream(Files.newOutputStream(jarFile))
                    val contents = sequence(Seq(
                      Steps.addManifestToZip,
                      Steps.addClassFilesFromDirToZip(Bootstrap.getClassDirectory(projectPath)),
                      Steps.addResourcesFromDirToZip(Bootstrap.getResourcesDirectory(projectPath)),
                    ))
                    toValidation(Using(outputStream)(contents))
                }
            }
        }
    }
  }
```

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
jaschdoc: No, Validation does not implement the required interface(s).

jaschdoc:
Ok updated.

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
magnus-madsen: Right- I suppose because its parallel? What about Result?

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -869,8 +908,28 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
...
+          return Validation.Failure(BootstrapError.FileError(s"The path '${formatter.red(dir.toString)}' is not a directory."))
+        }
+        if (!Files.isReadable(dir)) {
+          return Validation.Failure(BootstrapError.FileError(s"The path '${formatter.red(dir.toString)}' is not readable."))
+        }
+      }
+      Validation.Success(())
     }
 
+    def validateJarFile(jarFile: Path)(implicit formatter: Formatter): Validation[Unit, BootstrapError] = {
```
magnus-madsen: Maybe this is also going too far? This is not really a step, is it? Its more like part of the job of building a JAR.

(Not saying it could not be a helper function, but you have put it into Steps)

jaschdoc: Agreed, many of these "steps" are too small. It seems that once the refactor is done, there is a better overview of which functions depend on which. Then these smaller functions will either go away or be private helper functions that are called by the actual steps.

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
jaschdoc: Yeah I think that would be better. I think we should try to merge this when possible and then refactor to Result.

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
magnus-madsen: OK, but I am a bit hesitant to merge in this PR as it looks ATM.

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -655,8 +605,63 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
...
+      * Adds all class files from `dir` to `zip`.
+      */
+    def addClassFilesFromDirToZip(dir: Path)(zip: ZipOutputStream): Unit = {
+      // Add all class files.
+      // Here we sort entries by relative file name to apply https://reproducible-builds.org/
+      val classFiles = FileOps.getFilesWithExtIn(dir, EXT_CLASS, Int.MaxValue)
+      for ((buildFile, fileNameWithSlashes) <- FileOps.sortPlatformIndependently(dir, classFiles)) {
+        FileOps.addToZip(zip, fileNameWithSlashes, buildFile)
+      }
+    }
```
jaschdoc: It already seems we can move some of the validation of dirs and files to functions like these instead of validating at the call site. Doing so, would remove some of the "nano" steps.

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -336,106 +334,60 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
...
+      contents = sequence(Seq(
+        Steps.addManifestToZip,
+        Steps.addClassFilesFromDirToZip(Bootstrap.getClassDirectory(projectPath)),
+        Steps.addResourcesFromDirToZip(Bootstrap.getResourcesDirectory(projectPath))
+      ))
+      _ <- toResult(Using(new ZipOutputStream(Files.newOutputStream(jarFile)))(contents))
+    } yield {
+      ()
+    }
+    result.toValidation
```
magnus-madsen, line 358: Why the local var?

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -655,8 +605,66 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
...
+    case Success(v) => Ok(v)
+    case Failure(e) => Err(BootstrapError.FileError(e.getMessage))
+  }
+
+  /**
+    * Applies all functions in `fs` to `x`.
+    */
+  private def sequence[A](fs: Seq[A => Unit])(x: A): Unit = {
+    fs.foreach(f => f(x))
+  }
```
magnus-madsen: This you should avoid

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -655,8 +605,66 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
     }
   }
 
+  /**
+    * Converts `t` to a [[Validation]].
+    * If `t` is a failure then a [[BootstrapError.FileError]] is returned.
+    */
+  private def toResult[A](t: Try[A]): Result[A, BootstrapError] = t match {
```
magnus-madsen: This belongs somewhere in Result

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
jaschdoc: To convert to validation. You can't invoke the method directly on the for-comprehension

magnus-madsen: I see. Not even after the brace? You would need parens?

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -336,106 +334,60 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
...
-                }
-                entry = zipIn.getNextEntry
-              }
-          }
-        })
-      }
-    } match {
-      case Success(()) => Validation.Success(())
-      case Failure(e) => Validation.Failure(BootstrapError.GeneralError(List(e.getMessage)))
+    Steps.compile(flix) match {
```
magnus-madsen: Do we not have a .toValidation?

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -336,106 +334,60 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
...
   def buildJar(flix: Flix)(implicit formatter: Formatter): Validation[Unit, BootstrapError] = {
-    buildJarBase(flix, includeDependencies = false)
+    val jarFile = Bootstrap.getJarFile(projectPath)
+    Steps.updateStaleSources(flix)
+    val result = for {
+      _ <- Steps.configureJarOutput(flix)
+      _ <- Steps.compile(flix)
+      _ <- Steps.validateJarFile(jarFile)
+      _ = Files.createDirectories(getArtifactDirectory(projectPath))
+      contents = sequence(Seq(
```
magnus-madsen: I think sequence should already exist

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -336,106 +334,60 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
...
+        Steps.addManifestToZip,
+        Steps.addClassFilesFromDirToZip(Bootstrap.getClassDirectory(projectPath)),
+        Steps.addResourcesFromDirToZip(Bootstrap.getResourcesDirectory(projectPath))
+      ))
+      _ <- toResult(Using(new ZipOutputStream(Files.newOutputStream(jarFile)))(contents))
```
magnus-madsen: What does this line do? Perhaps a helper function here would be more clear.

magnus-madsen (commented):
Added some more comments.

First off, kudos for bringing this back a much more reasonable state 👍

magnus-madsen:
I think there is still one confusion. For a function return type we always want the most precise type right? E.g. String is better than Object.

When it comes to Result vs. Validation-- If you can return a Result do it! It means that the function can only fail with single error. Only return a Validation if there can be multiple errors.

Yes, the call sites may want a Validation but then they should do the conversion. 

In the same way that if you need some specific type, you typically don't change the callers.
 You just deal with the conversion locally.

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
jaschdoc: Why? Perhaps you can explain, so I can avoid the pattern in the future :)

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
jaschdoc: Yes, missed this one.

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
jaschdoc: Yes exactly.

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
magnus-madsen: Ask yourself: What does this abstraction provide? What are its laws?

magnus-madsen:
@jaschdoc Ping me when I should take another look.

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
jaschdoc: This sequence is this one https://github.com/flix/flix/pull/11646#discussion_r2379481504

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
jaschdoc: Ah I see what you mean. It just makes the call site prettier IMO, since you avoid explicitly creating a lambda by using partial application.

jaschdoc: From
```scala
      contents = sequence(Seq(
        Steps.addManifestToZip,
        Steps.addClassFilesFromDirToZip(Bootstrap.getClassDirectory(projectPath)),
        Steps.addResourcesFromDirToZip(Bootstrap.getResourcesDirectory(projectPath))
      ))
```
to (if `sequence` is removed)
```
      contents = (zip: ZipOutputStream) => {
        Steps.addManifestToZip(zip)
        Steps.addClassFilesFromDirToZip(Bootstrap.getClassDirectory(projectPath))(zip)
        Steps.addResourcesFromDirToZip(Bootstrap.getResourcesDirectory(projectPath))(zip)
      }
```

magnus-madsen: To my mind this just looks like a boilerplate. Just write the calls.

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -655,8 +600,57 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
...
+  private def sequence[A](fs: Seq[A => Unit])(x: A): Unit = {
+    fs.foreach(f => f(x))
+  }
+
   private object Steps {
 
+    /**
+      * Adds all class files from `dir` to `zip`.
+      */
+    def addClassFilesFromDirToZip(dir: Path)(zip: ZipOutputStream): Unit = {
```
magnus-madsen: Why is this curried?

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -336,106 +334,55 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
...
-    buildJarBase(flix, includeDependencies = true)
+    val jarFile = Bootstrap.getJarFile(projectPath)
+    val libDir = Bootstrap.getLibraryDirectory(projectPath)
+    Steps.updateStaleSources(flix)
+    val result = for {
+      _ <- Steps.configureJarOutput(flix)
+      _ <- Steps.compile(flix)
+      _ <- Steps.validateJarFile(jarFile)
+      _ <- Steps.validateDirectory(libDir)
+      _ <- Result.traverse(FileOps.getFilesWithExtIn(libDir, EXT_JAR, Int.MaxValue))(Steps.validateJarFile)
```
magnus-madsen: What is this line supposed to do?

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
jaschdoc: Due to the `sequence` function, but it's being removed as we speak ;)

magnus-madsen:
!!!jar

magnus-madsen:
@jaschdoc Could you also do some manual testing? Esp in folderMode?

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -729,13 +739,50 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
...
+    /**
+      * Writes `contents` to the jar file located at `jar`.
+      *
+      * This function also adds a manifest to the jar file.
+      *
+      * Creates the jar file if it does not exist, and truncates it if it already exists.
+      *
+      * @see [[Steps.addManifestToZip]]
+      */
+    def createJar(jar: Path, contents: ZipOutputStream => Unit): Result[Unit, BootstrapError.FileError] = {
```
magnus-madsen, line 776: Why does this function need to be higher order?

jaschdoc: Because of `Using`

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
```diff
@@ -868,8 +910,45 @@ class Bootstrap(val projectPath: Path, apiKey: Option[String]) {
...
+      Ok(())
+    }
+
+    /**
+      * Returns `Ok(())` if all files ending with `.jar` in `dir` are valid jar files.
+      *
+      * @see [[Steps.validateJarFile]]
+      */
+    def validateJarFilesIn(dir: Path): Result[Unit, BootstrapError] = {
+      Result.traverse(FileOps.getFilesWithExtIn(dir, EXT_JAR, Int.MaxValue))(Steps.validateJarFile).map(_ => ())
```
magnus-madsen, line 951: This is not consistent with the original semantics.

We are trying to create a new JAR file, possibly by overwriting an existing file, we want to be sure that we are not overwriting some other file.

But that is not what this does.

magnus-madsen: I mean perhaps this check is meaningful? But the error message you create would not make any sense!

jaschdoc: No, this is to validate that all jar files in the library are valid jars. These jars will be included in the new jar. None of them will be overwritten.

jaschdoc:
> @jaschdoc Could you also do some manual testing? Esp in folderMode?

Yes, the flix benchmarks seem to work just fine with this version. Project mode also works fine :)

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
magnus-madsen: Check the error message you create...

jaschdoc: Yeah ok, the error message should be updated then.

magnus-madsen (approved):

magnus-madsen:
@jaschdoc I am merging this-- then you can fix the error message in a quick follow-up PR.

main/src/ca/uwaterloo/flix/api/Bootstrap.scala
jaschdoc: `s"The path '${jarFile.toString}' exists and is not a jar-file. Refusing to overwrite."` -> `s"The path '${jarFile.toString}' exists and is not a jar-file."`

PR #11647, opened 2025-09-17
feat: allow string interpolations under debug

magnus-madsen:


magnus-madsen:
> What happens if I try to match on a Debug string?

Its a string.

mlutze:
> > What happens if I try to match on a Debug string?
> 
> Its a string.

Sorry, wrong preposition.

What happens if I do:

```
def foo(): Unit = match ??? {
    case d"hello ${adsf}" => ???
}
```

magnus-madsen:
> > > What happens if I try to match on a Debug string?
> > 
> > 
> > Its a string.
> 
> Sorry, wrong preposition.
> 
> What happens if I do:
> 
> ```
> def foo(): Unit = match ??? {
>     case d"hello ${adsf}" => ???
> }
> ```

It is a parse error. DebugInterpolators are now expressions, but not literals-- so they cannot occur in constant/literal/pattern position. Its a parse error.

PR #11649, opened 2025-09-17
feat: Pretty print lexer output

JonathanStarup:


main/src/ca/uwaterloo/flix/language/dbg/Doc.scala
```diff
@@ -244,12 +244,26 @@ object Doc {
...
+    foldLeft({case (x, y) => f(y, x)}, d.reverse)
+  }
+
+  /**
+    * Left fold of `d` with `f`.
+    */
+  private def foldLeft(f: (Doc, Doc) => Doc, d: List[Doc]): Doc = d match {
     case Nil => empty
     case x :: Nil => x
-    case x :: xs => f(x, fold(f, xs))
```
JonathanStarup, line 252: Changed this to a tail recursive version and added a comment of why foldRIght was used to begin with

PR #11650, opened 2025-09-17
refactor: uniform intrinsic weeding

### LINKED ISSUE #11632: Jonathan has promised to clean this up later.
Jonathan has promised to clean this up later.

_Originally posted by @mlutze in https://github.com/flix/flix/issues/11624#issuecomment-3298875187_

JonathanStarup:
Fixes #11632


main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
```diff
@@ -1108,31 +1108,13 @@ object WeederError {
...
-    }
-  }
-
-  /**
-    * An error raised to indicate an unapplied intrinsic.
-    *
-    * @param intrinsic name of the intrinsic.
-    * @param loc       the location where the illegal intrinsic occurs.
-    */
-  case class UnappliedIntrinsic(intrinsic: String, loc: SourceLocation) extends WeederError {
```
JonathanStarup, line 1128: I've removed this since some intrinsics are unapplied. Now there is one vague error for all intrinsics. I think that is appropriate given their use.

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -2196,136 +2180,145 @@ object Weeder2 {
...
-        case ("CHANNEL_GET", e1 :: Nil) => Validation.Success(Expr.GetChannel(e1, loc))
-        case ("CHANNEL_PUT", e1 :: e2 :: Nil) => Validation.Success(Expr.PutChannel(e1, e2, loc))
-        case ("CHANNEL_NEW", e :: Nil) => Validation.Success(Expr.NewChannel(e, loc))
-        case ("ARRAY_NEW", e1 :: e2 :: e3 :: Nil) => Validation.Success(Expr.ArrayNew(e1, e2, e3, loc))
-        case ("ARRAY_LENGTH", e1 :: Nil) => Validation.Success(Expr.ArrayLength(e1, loc))
-        case ("ARRAY_LOAD", e1 :: e2 :: Nil) => Validation.Success(Expr.ArrayLoad(e1, e2, loc))
-        case ("ARRAY_STORE", e1 :: e2 :: e3 :: Nil) => Validation.Success(Expr.ArrayStore(e1, e2, e3, loc))
-        case ("VECTOR_GET", e1 :: e2 :: Nil) => Validation.Success(Expr.VectorLoad(e1, e2, loc))
-        case ("VECTOR_LENGTH", e1 :: Nil) => Validation.Success(Expr.VectorLength(e1, loc))
+      Validation.Success((intrinsic, args) match {
```
magnus-madsen: Are you matching on Success? I don't get it.

JonathanStarup: No its just wrapping the output of the match in Success, instead of repeating it in every case.

JonathanStarup:
> I agree the code is more messy in master, but it just seems strange to have one error message for both missing and under applied?

`$VECTOR_GET$(1, 2)` says "undefined intrinsic" currently so I'd say this PR is comparable to status quo.

I can match in two steps (first the name, then the args) which you allow a precise "$Example$ needs 7 arguements" errors. The halfway precision is not possible like this.

I think this PR is a good change as is, but i can do a follow up with better errors if you want to?

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
magnus-madsen: Oh, I misread.

mlutze: I don't like it.

I'd say bind it to a result and then wrap that in success.

JonathanStarup: Your wish is my command

Issue #11483, opened 2025-09-04
Add support for default effect handlers


magnus-madsen:
- [ ] Add an annotation `@DefaultHandler` that can be placed on a function.
- [ ] A function with a default handler must have a signature of the form `def func(f: Unit -> a \ ef): a \ (ef - X) + Y + Z ...` where `X` is the effect it handles and `Y + Z + ...` are primitive effects. Add a check for this in e.g. `EntryPoints`.
- [ ] There can only be one default handler per effect `X`. Add a check for this in e.g. `EntryPoints`.
- [ ] If effect `X` has a default handler `func`  then main (and every test entrypoint) is automatically wrapped in that handler.
- [ ] Refactor `Assert` to have a default handler.

magnus-madsen:
@NeonOxide I drafted an outline of how we can implement this. What do you think?

magnus-madsen:
Fixed in master.

PR #11652, opened 2025-09-18
feat: add new EntryPointErrors for default handlers

NeonOxide:
Added new EntryPointErrors for default handlers. related to: #11483


magnus-madsen:
We agree to boil it down to just 4 errors...

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -222,4 +222,150 @@ object EntryPointError {
...
+         |""".stripMargin
+    }
+  }
+  /**
+    * An error raised to indicate that there are multiple default handlers for the same effect.
+    *
+    * @param loc the location of the defn.
+    * @param eff the effect associated with this default handler
+    */
+  case class MultipleDefaultHandlers(loc: SourceLocation, eff: Type) extends EntryPointError {
```
magnus-madsen: The source location parameter is by convention always last.

instead of the type, can we pass the effect sym?

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -222,4 +222,150 @@ object EntryPointError {
...
+    def summary: String = s"Multiple default handlers for $eff."
```
magnus-madsen: '$eff'.

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -222,4 +222,150 @@ object EntryPointError {
...
+
+    def message(formatter: Formatter): String = {
+      import formatter.*
+      s""">> Multiple default handlers for $eff found.
```
magnus-madsen: Remove found, so its the same line as the summary.

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
magnus-madsen: Can you look at some of the other duplicate errors? I think we should want the source location of both handlers.

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -222,4 +222,150 @@ object EntryPointError {
          |""".stripMargin
     }
   }
+
+  /**
+    * An error raised to indicate that the effect for default handler was not able to be found.
```
magnus-madsen: not able to be found? Just not found?

But perhaps this needs to be reworded entirely. Something like: 

DefaultHandlerNotInEffectModule

with an error like "A default handler must be in the companion module of an effect."

Then expand with 

"Module Foo does not have a companion effect Foo" etc.

magnus-madsen: I suggest you try to remake this error.

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -222,4 +222,150 @@ object EntryPointError {
...
+         |""".stripMargin
+    }
+  }
+  /**
+    * An error raised to indicate that the amount of arguments of a default handler is wrong.
+    *
+    * @param loc the location of the defn.
+    * @param eff the symbol of the effect associated with this default handler
+    */
+  case class WrongNumberOfArgsForDefaultHandler(loc: SourceLocation, eff: Symbol.EffSym) extends EntryPointError {
```
magnus-madsen: Delete

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -222,4 +222,150 @@ object EntryPointError {
...
+         |""".stripMargin
+    }
+  }
+  /**
+    * An error raised to indicate that the expected type of the argument of a default handler is a function.
+    *
+    * @param loc the location of the defn.
+    * @param eff the symbol of the effect associated with this default handler
+    */
+  case class ExpectedFunctionArgForDefaultHandler(loc: SourceLocation, eff: Symbol.EffSym) extends EntryPointError {
```
magnus-madsen: Delete

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -222,4 +222,150 @@ object EntryPointError {
...
+         |""".stripMargin
+    }
+  }
+  /**
+    * An error raised to indicate that the expected type of the effect set of the function argument of a default handler is wrong.
+    *
+    * @param loc the location of the defn.
+    * @param eff the symbol of the effect associated with this default handler
+    */
+  case class ExpectedEffectVarForDefaultHandler(loc: SourceLocation, eff: Symbol.EffSym) extends EntryPointError {
```
magnus-madsen: Delete

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -222,4 +222,150 @@ object EntryPointError {
...
+         |""".stripMargin
+    }
+  }
+  /**
+    * An error raised to indicate that the expected type of the argument of a default handler or its return type is wrong.
+    *
+    * @param loc the location of the defn.
+    * @param eff the symbol of the effect associated with this default handler
+    */
+  case class ExpectedTypeVarArgForDefaultHandler(loc: SourceLocation, eff: Symbol.EffSym) extends EntryPointError {
```
magnus-madsen: Delete

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -222,4 +222,150 @@ object EntryPointError {
...
+    }
+  }
+  /**
+    * An error raised to indicate that a default handler produces non primitive effects.
+    *
+    * @param loc the location of the defn.
+    * @param eff the symbol of the effect associated with this default handler
+    * @param wrongEff the symbol of the non primitive effect produced
+    */
+  case class NonPrimitiveEffectForDefaultHandler(loc: SourceLocation, eff: Symbol.EffSym, wrongEff: Symbol.EffSym) extends EntryPointError {
```
magnus-madsen: I think you argued this was useful to keep.

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -222,4 +222,150 @@ object EntryPointError {
...
+      s""">> Effect for Default handler not found. Default handlers must be in their respective effect's companion module.
+         |
+         |${code(loc, "default handler.")}
+         |
+         |""".stripMargin
+    }
+  }
+
+  /**
+    * An error raised to indicate that the signature of a default handler is wrong.
```
magnus-madsen: Good

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -222,4 +222,150 @@ object EntryPointError {
...
+    *
+    * @param loc the location of the defn.
+    * @param eff the symbol of the effect associated with this default handler
+    */
+  case class WrongSignatureForDefaultHandler(loc: SourceLocation, eff: Symbol.EffSym) extends EntryPointError {
```
magnus-madsen: loc always goes last.

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -222,4 +222,150 @@ object EntryPointError {
...
+    def summary: String = s"Wrong signature for $eff's default handler found.\nExpected 'def handler(f: Unit -> a \\ ef) : a \\ (ef - ${eff.name}) + PrimEf1 + ...'"
```
magnus-madsen: I think the summary needs to be short, so just "Illegal signature for default handler" or something

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -222,4 +222,150 @@ object EntryPointError {
...
+
+    def message(formatter: Formatter): String = {
+      import formatter.*
+      s""">> Wrong signature for $eff's default handler found. Expected 'def handler(f: Unit -> a \\ ef) : a \\ (ef - ${eff.name}) + PrimEf1 + ...'
```
magnus-madsen: Repeat the summary, then add some more lines like:

The default handler must have the exact signature:

  pub def handler(f: Unit -> a \\ ef) : a \\ (ef - ${eff.name}) + PrimEf1 + ...

then the code

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -222,4 +222,150 @@ object EntryPointError {
...
+         |
+         |${code(loc, "default handler.")}
```
magnus-madsen: illegal default handler signature etc.

magnus-madsen (commented):
Looks good, but lets do a pass and really polish it! :)

magnus-madsen:
Remember to keep the file sorted too.

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
NeonOxide: But you will get an error for each one already. The problem with this is that where do we stop if we have for example 3 repeated handlers?

NeonOxide: Also, remember that the only way there are multiple default handlers for the same effect is if there are multiple effects marked with DefaultHandler in the companion module. So maybe what we can do is say that in the hint?

Something like: this error is generated because there is more than one function marked with @DefaultHandler in '$moduleName'. Keep only one of them.

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
NeonOxide: I can agree. However, the message is actually not super big (it is just two lines, most other errors I get in Flix are bigger), and I found while debugging myself that it is much more useful to see by putting the mouse over the red line something like: hey this is bad, we expect this (as in normal type errors) than just say the signature is wrong, go figure out why.

I can change it if you want (after all it is just removing a line). But I think this is actually better. Removing this would be the equivalent of having the summary of a type error be, Int32 is not the expected type, without saying which one is the actually expected type. What do you think?

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
NeonOxide: Yeah, I just did it out of convenience. I'll change it

magnus-madsen: We do all pairs, I guess you get at most nˆ2?

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -205,6 +223,53 @@ object EntryPointError {
     override def loc: SourceLocation = SourceLocation.Unknown
   }
 
+  /**
+    * An error raised to indicate that there are multiple default handlers for the same effect.
+    *
+    * @param eff the symbol of the effect associated with this default handler
+    * @param loc the location of the defn.
+    */
+  case class MultipleDefaultHandlers(eff: Symbol.EffSym, loc: SourceLocation) extends EntryPointError {
```
magnus-madsen: eff -> sym

magnus-madsen: (because eff means the grammar of effects)

magnus-madsen: The usual structure here should be DuplicateDefaultHandler (with two source locations)

Take a look at some of the other duplicate errors.

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -205,6 +223,53 @@ object EntryPointError {
...
+  }
+
+  /**
+    * An error raised to indicate that a default handler produces non-primitive effects.
+    *
+    * @param eff the symbol of the effect associated with this default handler
+    * @param wrongEff the symbol of the non-primitive effect produced
+    * @param loc the location of the defn.
+    */
+  case class NonPrimitiveEffectForDefaultHandler(eff: Symbol.EffSym, wrongEff: Symbol.EffSym, loc: SourceLocation) extends EntryPointError {
```
magnus-madsen: eff -> sym

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -30,6 +30,24 @@ sealed trait EntryPointError extends CompilationMessage {
 
 object EntryPointError {
 
+  /**
+    * An error raised to indicate that a default handler is not in the companion module of an effect.
```
magnus-madsen: linebreak

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -30,6 +30,24 @@ sealed trait EntryPointError extends CompilationMessage {
...
+    * @param module the symbol of the module containing the default handler
```
magnus-madsen: missing .

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -30,6 +30,24 @@ sealed trait EntryPointError extends CompilationMessage {
...
+    * @param loc the location of the defn.
```
magnus-madsen: the location of the default handler?

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -30,6 +30,24 @@ sealed trait EntryPointError extends CompilationMessage {
...
+    */
+  case class DefaultHandlerNotInEffectModule(module: Symbol.ModuleSym,loc: SourceLocation) extends EntryPointError {
```
magnus-madsen: module -> sym

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -30,6 +30,24 @@ sealed trait EntryPointError extends CompilationMessage {
...
+    def summary: String = s"A default handler must be in the companion module of an effect."
```
magnus-madsen: Good 👍

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -30,6 +30,24 @@ sealed trait EntryPointError extends CompilationMessage {
...
+
+    def message(formatter: Formatter): String = {
+      import formatter.*
+      s""">> A default handler must be in the companion module of an effect. Module '$module' does not have a companion effect.
```
magnus-madsen: This will not fit on 80 char terminal. You can make it two lines or incorporate the text in the code block.

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -30,6 +30,24 @@ sealed trait EntryPointError extends CompilationMessage {
...
+         |
+         |${code(loc, "default handler.")}
```
magnus-madsen, line 49: is there some better text we can say here?

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
magnus-madsen: I wonder if we want to call this an orphan handler? Maybe not?

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -205,6 +223,53 @@ object EntryPointError {
...
+         |
+         |""".stripMargin
+    }
+
+    override def explain(formatter: Formatter): Option[String] = Some({
+      s"""
+         |This error is generated because there is more than one function marked with @DefaultHandler in module '$eff'.
+         |Fix it by reducing the number to either one or zero default handlers.
```
magnus-madsen: This sentence does not add much. Its grammar is also strange.

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -205,6 +223,53 @@ object EntryPointError {
...
+         |
+         |""".stripMargin
+    })
+  }
+
+  /**
+    * An error raised to indicate that a default handler produces non-primitive effects.
```
magnus-madsen: produces -> has

Otherwise very good 👍

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -205,6 +223,53 @@ object EntryPointError {
...
+    *
+    * @param eff the symbol of the effect associated with this default handler
+    * @param wrongEff the symbol of the non-primitive effect produced
+    * @param loc the location of the defn.
+    */
```
magnus-madsen, line 265: Update the above comments.

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -205,6 +223,53 @@ object EntryPointError {
...
+  case class NonPrimitiveEffectForDefaultHandler(eff: Symbol.EffSym, wrongEff: Symbol.EffSym, loc: SourceLocation) extends EntryPointError {
+    def summary: String = s"Illegal signature for '$eff's default handler.\nNon Primitive effect '$wrongEff' generated."
+
+    def message(formatter: Formatter): String = {
+      import formatter.*
+      s""">> Illegal signature for '$eff's default handler. Non Primitive effect '$wrongEff' generated.
+         |
+         |${code(loc, "default handler.")}
```
magnus-madsen: -> illegal signature

magnus-madsen: or maybe -> illegal handler signature.

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -222,4 +287,26 @@ object EntryPointError {
...
   }
+
+  /**
+    * An error raised to indicate that the signature of a default handler is wrong.
+    *
+    * @param eff the symbol of the effect associated with this default handler
+    * @param loc the location of the defn.
+    */
+  case class WrongSignatureForDefaultHandler(eff: Symbol.EffSym, loc: SourceLocation) extends EntryPointError {
+    def summary: String = s"Illegal signature for '$eff's default handler .\nExpected 'def handler(f: Unit -> a \\ ef) : a \\ (ef - ${eff.name}) + PrimEf1 + ...'"
```
magnus-madsen: Summary has to be 1 line. It cannot have linebreaks. Also it should be short.

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -222,4 +287,26 @@ object EntryPointError {
...
+
+    def message(formatter: Formatter): String = {
+      import formatter.*
+      s""">> Illegal signature for '$eff's default handler.
+         | The default handler must have this exact signature (names can be chosen arbitrarily):
```
magnus-madsen, line 310: Add linebreak between.

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
```diff
@@ -222,4 +287,26 @@ object EntryPointError {
...
+         |'def handler(f: Unit -> a \\ ef) : a \\ (ef - ${eff.name}) + PrimEf1 + ...'
+         |
+         |${code(loc, "default handler.")}
```
magnus-madsen: illegal signature or the like (see above)

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
NeonOxide: I mean we could. But then I think that makes its use less clear. What is exactly an orphan handler? It could mean that the handler is not in the correct module, that the handler is not in the companion module of an effect (current meaning), it could also mean that it is a handler for orphan elements (whatever that would mean, something like unhandled effects?). Conclusion being that I think it is less clear, what do you think?.

main/src/ca/uwaterloo/flix/language/errors/EntryPointError.scala
NeonOxide: I don't really know. The error is kind of just that. We can add something along the lines of, a default handler does not make sense if there is no effect to handle. Could also say something like, handler is in module without a companion Effect. I reworded it slightly, but I do not think it has changed that much.

magnus-madsen: Roger

magnus-madsen (approved):

magnus-madsen:
Thanks!

PR #11654, opened 2025-09-18
feat: Lex keywords with prefix tree

JonathanStarup:
- [feat] faster lexer (`--benchmark-phases` once on laptop)
  - `    PR Lexer 135,2 ms (02,8%)`
  - `Master Lexer 273,7 ms (05,2%)`
  - See #11655 for the full version of this idea
- [feat] keywords are accessible as data, useful for documentation or checking.


JonathanStarup:
I only did this for keywords to keep the PR smaller, but if "end" benchmarking is needed I can do the same for operators in this PR

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -397,13 +318,132 @@ object Lexer {
...
+  private def acceptKeyword()(implicit s: State): Option[TokenKind] = {
+    def advanceIfSome[T](offset: Int, opt: Option[T]): Option[T] = {
+      opt.foreach(_ => s.sc.advanceN(offset))
+      opt
+    }
+
+    @tailrec
+    def search(offset: Int, node: MutPrefixTree.Node[TokenKind]): Option[TokenKind] = {
+      s.sc.nth(offset) match {
+        case Some(c) =>
```
magnus-madsen: For extra performance, you probably want to avoid options in the search

JonathanStarup: yeah, a lot of options everywhere. But i tried to write it without those sorts of optimizations initially.

magnus-madsen: I think you should keep those options everywhere-- except for where you eat one char at time.

magnus-madsen: I suppose we can still make a clean PR and merge it. Then you can optimize in a subsequent PR.

magnus-madsen:
Any reason the prefix tree has to be mutable? It seems to me its construction could be mutable and then it becomes immutable.

JonathanStarup:
> Any reason the prefix tree has to be mutable? It seems to me its construction could be mutable and then it becomes immutable.

Yeah that'd be the proper thing to do - but also requires adding another version of prefix tree so I left it here to keep it smaller.

magnus-madsen:
> > Any reason the prefix tree has to be mutable? It seems to me its construction could be mutable and then it becomes immutable.
> 
> Yeah that'd be the proper thing to do - but also requires adding another version of prefix tree so I left it here to keep it smaller.

A smart constructor?

JonathanStarup:
Oh encapsulate the building of it. yeah that would work.

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
JonathanStarup: Yea 👍

JonathanStarup:
@magnus-madsen It's more polished now and has an immutable interface

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -453,13 +460,44 @@ object Lexer {
...
+    // A potential match must be:
+    //   - The longest match in the keyword map.
+    //   - Followed by a char that is non-letter-like (or EOF).
+
+    /**
+      * Keep peeking input while the prefix is contained in the tree.
+      * At end of file or non-matching non-letter-like char, return the optional current mapping.
+      * At non-matching letter-like char, return none.
+      */
+    @tailrec
```
magnus-madsen, line 474: Good... Embrace the darkside.. erh I mean FP side.

magnus-madsen:
Thanks-- looks good!

magnus-madsen:
I think if some of the options (in tight loops) are dropped then this could be very fast.

PR #11656, opened 2025-09-19
feat: disallow implicit `Debug` in production mode

magnus-madsen:


main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolverInterface.scala
```diff
@@ -107,9 +108,13 @@ object ConstraintSolverInterface {
           // All constraints solved. Yay!
           (subst, Nil)
         case _ =>
-          // We have one or more type errors. We need to report them, but first we have special logic for the [[Debug]] effect.
+          // We have one or more type errors. We check whether we are in development or production mode.
+          if (flix.options.build == Build.Production) {
```
mlutze: A match would be better here for when we add `Build.SemiProdWithDevFoo` mode.

mlutze (approved):

main/src/ca/uwaterloo/flix/language/phase/typer/ConstraintSolverInterface.scala
magnus-madsen: Fixed.

PR #11660, opened 2025-09-21
chore: improve performance of `BPlusTree` tests

CasperDN:
Changes:
1. The number of inserted elements in some tests has been decreased.
2. `put05` does no longer re-compute of the same thing for each round.
3. Deleted some redundant `concurrencyXX` tests.


magnus-madsen (approved):

magnus-madsen:
Awesome!

PR #11661, opened 2025-09-21
refactor: rename `fixpoint` to `datalog`

magnus-madsen:


magnus-madsen:
I want to use the name "Datalog" going forward. Will also rename in docs.

mlutze (approved):

PR #11662, opened 2025-09-21
ci: enable gradle cache

magnus-madsen:


magnus-madsen:
This is an experiment to see if we can speed-up the build.

.github/workflows/compiler-build-and-test.yaml
```diff
@@ -21,8 +21,9 @@ jobs:
         with:
           distribution: 'temurin'
           java-version: '21'
+          cache: gradle
       - name: Build with gradle
-        run: ./gradlew clean compileTestScala --console=plain
```
magnus-madsen, line 25: I don't know if removing clean can be a problem. We have to see.

magnus-madsen:
Not really sure why the cache is not being reused. Does it have to be merged first?

magnus-madsen:
@mlutze Perhaps we can merge it and see what happens.

mlutze (approved):

PR #11663, opened 2025-09-21
test: speed-up some tests

magnus-madsen:


main/test/ca/uwaterloo/flix/language/TestFlixErrors.scala
```diff
@@ -54,59 +54,4 @@ class TestFlixErrors extends AnyFunSuite with TestUtils {
     expectRuntimeError(result, BackendObjType.HoleError.jvmName.name)
   }
 
-  test("SpawnedThreadError.01") {
```
magnus-madsen, line 57: Removed these tests because they interfere with a lot of tooling by printing error.

mlutze (approved):

PR #11669, opened 2025-09-22
feat: add `@CompileTest` annotation

magnus-madsen:


magnus-madsen:
I think this works well. We can always discuss whether we should have fewer (or no `@CompileTest`) -- but this allows us to begin flagging them.

PR #11670, opened 2025-09-22
test: cleanup and remove old code

magnus-madsen:


main/test/flix/Test.Equality.Map.flix
```diff
@@ -1,125 +0,0 @@
-mod Test.Equality.Map {
```
magnus-madsen, line 1: Already tested in TestMap

main/test/flix/Test.Equality.Set.flix
```diff
@@ -1,169 +0,0 @@
-mod Test.Equality.Set {
```
magnus-madsen, line 1: Already tested in TestSet

PR #11675, opened 2025-09-23
fix: parsing of debug interpolator

magnus-madsen:


magnus-madsen:
This is more correct, but still only allows strings and string interpolations due to the lexing.

JonathanStarup:
> only allows strings and string interpolations

Isn't that all its supposed to allow?

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1623,12 +1621,12 @@ object Weeder2 {
     private def visitLiteralRecordExpr(tree: Tree)(implicit sctx: SharedContext): Validation[Expr, CompilationMessage] = {
       expect(tree, TreeKind.Expr.RecordOperation)
       // `{ +x = expr }` is not allowed.
-      pickAll(TreeKind.Expr.RecordOpExtend, tree).foreach{t =>
+      pickAll(TreeKind.Expr.RecordOpExtend, tree).foreach { t =>
```
mlutze, line 1624: blah

Issue #11676, opened 2025-09-23
Remove `~~~`


JonathanStarup:
It is currently unused as a Token - what was it used for?

magnus-madsen:
Bitwise negation. Its friends are <<< and >>>.

magnus-madsen:
We should keep it. If its not used thats a bug.

JonathanStarup:
```
>> Undefined name '~~~'.

1 | def main(): Int32 = ~~~32
                        ^^^^^
                        name not found
```

JonathanStarup:
> Bitwise negation. Its friends are <<< and >>>.

But maybe they should just be functions that uses a `$BITWISE_NEGATION$`?

`>>>`, `<<<`, and `~~~` all just crash after parsing

JonathanStarup:
I think this is just leftover stuff from #5397

magnus-madsen:
OK; just nuke the tokens then.

JonathanStarup:
`>>>` and `<<<` could be reintroduced in Prelude if wanted (`~` is not allowed in user ops)

PR #11677, opened 2025-09-23
feat: prefix tree the inbuilt operators

JonathanStarup:


main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -154,6 +154,33 @@ object Lexer {
       ("{", TokenKind.CurlyL),
       ("|#", TokenKind.BarHash),
       ("}", TokenKind.CurlyR),
+      ("~", TokenKind.Tilde),
```
JonathanStarup, line 157: Tilde can be added here now since the operators can be put before the simple tokens. If #11676 is done then there is no dependency between the sets.

magnus-madsen: What does this mean?

magnus-madsen:
Can I merge this if I still want ˜˜~

JonathanStarup:
Yes 👍

PR #11678, opened 2025-09-23
refactor: proper zero-point source location

JonathanStarup:
Since end is exclusive, the end point should be one further out from the start. It is still zero width.


JonathanStarup:
Matt caught my thought cod

PR #11684, opened 2025-09-23
refactor: name char sets

JonathanStarup:
Make commonly used char classes explicit and reused.


main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -471,7 +471,7 @@ object Lexer {
 
   /** Advance the current position past a keyword if any keyword completely matches the current word. */
   private def acceptIfKeyword()(implicit s: State): Option[TokenKind] =
-    advanceIfInTree(Keywords, c => !(c.isLetter || c.isDigit || c == '_'))
+    advanceIfInTree(Keywords, c => !isNameChar(c))
```
JonathanStarup, line 474: behaviour here is slightly changed, now `def$` or `def!` is a name instead of a keyword followed by some symbol

PR #11685, opened 2025-09-23
refactor: allow empty cparams for closures

### LINKED ISSUE #11586: Avoid spurious unit arguments in `LambdaLift`
There is a million of these:

https://github.com/flix/flix/blob/master/main/src/ca/uwaterloo/flix/language/phase/LambdaLift.scala#L116-L118

JonathanStarup:
Fixes #11586

There is the possibility that `isClosure = false && cparams.nonEmpty` would cause bugs, but thats also basically how it was before. So this PR does not worsen the situation.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/LiftedAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/LiftedAst.scala
@@ -32,7 +32,7 @@ object LiftedAst {
                   entryPoints: Set[Symbol.DefnSym],
                   sources: Map[Source, SourceLocation])
 
-  case class Def(ann: Annotations, mod: Modifiers, sym: Symbol.DefnSym, cparams: List[FormalParam], fparams: List[FormalParam], exp: Expr, tpe: SimpleType, loc: SourceLocation)
+  case class Def(ann: Annotations, mod: Modifiers, sym: Symbol.DefnSym, cparams: List[FormalParam], isClosure: Boolean, fparams: List[FormalParam], exp: Expr, tpe: SimpleType, loc: SourceLocation)
 
   case class Enum(ann: Annotations, mod: Modifiers, sym: Symbol.EnumSym, tparams: List[TypeParam], cases: Map[Symbol.CaseSym, Case], loc: SourceLocation)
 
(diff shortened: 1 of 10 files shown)


JonathanStarup:
Splitting closures and functions into two classes would be more efficient and clean but would require hundreds of lines of code more.

main/src/ca/uwaterloo/flix/language/ast/LiftedAst.scala
```diff
@@ -32,7 +32,7 @@ object LiftedAst {
                   entryPoints: Set[Symbol.DefnSym],
                   sources: Map[Source, SourceLocation])
 
-  case class Def(ann: Annotations, mod: Modifiers, sym: Symbol.DefnSym, cparams: List[FormalParam], fparams: List[FormalParam], exp: Expr, tpe: SimpleType, loc: SourceLocation)
+  case class Def(ann: Annotations, mod: Modifiers, sym: Symbol.DefnSym, cparams: List[FormalParam], isClosure: Boolean, fparams: List[FormalParam], exp: Expr, tpe: SimpleType, loc: SourceLocation)
```
magnus-madsen, line 35: Could we add an assertion to the body of the case class?

magnus-madsen:
I need to check the impact on bytecode size. I would expect it to be substantial.

magnus-madsen:
!!!jar

magnus-madsen:
I think something maybe wrong? Not sure, it has zero impact right now.

main/src/ca/uwaterloo/flix/language/ast/LiftedAst.scala
JonathanStarup: Could add `if (!isClosure) assert(cparams.isEmpty)`

JonathanStarup:
> I need to check the impact on bytecode size. I would expect it to be substantial.

up to one field per closure class doesn't sound like too much to me

magnus-madsen:
It is also one argument per function right?

magnus-madsen:
Including code to pass that argument.

JonathanStarup:
right, per non-capturing closure

main/src/ca/uwaterloo/flix/language/ast/LiftedAst.scala
jaschdoc: With `else` branch too?

PR #11688, opened 2025-09-23
feat: change `$INTRINSIC$` to `%%INTRINSIC%%`

JonathanStarup:
- Simpler: $ is used in other syntax like keyword escaped `$run`, names `Map$Entry` and user operators `def &(x: Int32): Int32`. `%` is not.
- Hard to reach: `%INTRINSIC%` syntax would give users writing `x % 2 == 0` an ugly error. The repetition puts it out of reach.
- Unique: easier to search for compared to $

Syntax errors in intrinsics are not great but also not unchanged.


magnus-madsen:
Any alternatives for %%?

JonathanStarup:
it is the only unused keyboard accessible ascii symbols i could find.

And if you get a good idea after this PR it is now easy to search and replace 😁

magnus-madsen:
@mlutze Thoughts?

magnus-madsen:
What's the benefit here?

mlutze:
> @mlutze Thoughts?

I am convinced by Jonathan's arguments. We overload symbols too much.

JonathanStarup:
Cost
- Longer syntax for built-ins - but they shouldn't be used much (should be used once each imo)

Benefit
- Easier Lexing
- Better errors, i.e. its harder to stumble into the lexer path that goes inside intrinsics and get unterminated intrinsic for example
- Easier searchable

JonathanStarup:
Other than the arithmetic symbols `+-*/` and the relations `><=`, this is the current state of keyboard symbols (ish)

| Symbol | Used for |
| --- | --- |
| `!`  | user ops and inside namex |
| `"`  | Strings | 
| `#`  | Datalog and list/vector/map literals |
| `$`  | Escaped names, user ops, and names etc. |
| `%`  | - |
| `&`  | user ops |
| `'` |  Chars |
| `?`  | holes |
| `@`  | regions and annotations |
| `~`  | negation |
| `^`  | user ops |

PR #11691, opened 2025-09-23
feat: use Datalog types instead of RamId

CasperDN:
Third part of #11381.

Switches to use the `Type`s instead of `RamId` during lowering. After this the only thing left is cleanup, e.g. deleting `RamId` and removing it from the ast.

To show that `RamId` truly is not used I've hardcoded them to `RamId.Id(42)` in the compiler.


magnus-madsen:
Thanks!

PR #11694, opened 2025-09-23
feat: remove `<=>`

JonathanStarup:
Unused token


JonathanStarup:
Nevermind, just only used via toString

magnus-madsen:
We are keeping the spaceship :D

PR #11698, opened 2025-09-23
refactor: add `#` as simple token

JonathanStarup:
The tree uses longest match, so `#` can also be added


magnus-madsen:
Do we need a comment or is it blindingly obvious?

JonathanStarup:
there is a comment in the search function

PR #11699, opened 2025-09-23
refactor: make `<>` more consistent

JonathanStarup:
`<>` is an empty case set, and this PR avoids a hack that lexes `<>` and two tokens.

This PR treats it like other special combinations (like `:::` etc)

(As another option, we can require `< >` for empty case sets, allowing us to delete this code completely)


magnus-madsen:
This is for restrictable variants, right?

Since this is not their final syntax (if they will ever have one), what is the simplest implementation (i.e. what cleans the code the most? Would that be to require `< >`? If so, I would go for that).

JonathanStarup:
Would that be to require < >? If so, I would go for that). Yes this requires zero code. Ill make a new pr

PR #11702, opened 2025-09-24
refactor: use `ConcurrentCache` in `SimpleType`

magnus-madsen:


mlutze:
See https://github.com/flix/flix/pull/11704 and avoid casts.

mlutze (approved):

PR #11704, opened 2025-09-24
refactor: improve ConcurrentCache API

mlutze:


main/src/ca/uwaterloo/flix/util/ConcurrentCache.scala
```diff
@@ -31,8 +31,8 @@ class ConcurrentCache[V] {
     *
     * The value `v` must correctly implement the `equals` and `hashCode` methods.
     */
-  def getCanonicalValue(v: V): V = {
-    cache.computeIfAbsent(v, (_: V) => v)
+  def getCanonicalValue[T <: V](v: T): T = {
+    cache.computeIfAbsent(v, (_: V) => v).asInstanceOf[T]
```
magnus-madsen, line 35: This is unsound right?

mlutze: It is unsound if we have equality defined such that different types can be equal.

magnus-madsen: I think they can? With effects? rows?

mlutze: But surely we wouldn't define Scala equality on those in such a way, right?

magnus-madsen: Type.Apply could become Type.Var right?

magnus-madsen: I am saying it could fail with an innocent mistake due to equivalences.

mlutze: > Type.Apply could become Type.Var right?

I don't see how it's possible.

For anything like this to happen, one would have to override the default equals implementation to match on a different type.

I claim that if this fails then your equality function is ill-defined.

magnus-madsen:
I claim this is unsound :)

main/src/ca/uwaterloo/flix/util/ConcurrentCache.scala
magnus-madsen: x and x replaced by x

mlutze:
> I claim this is unsound :)

I claim that your claim is unfounded.

> x and x replaced by x

The equivalence in this class is determined by the `equals` and `hashcode` function, as suggested by the documentation comment 

```
The value `v` must correctly implement the `equals` and `hashCode` methods.
```

The type `x and x` is not equal to `x` according to the `equals` function defined on `Type.Apply`.

If two values are equal, then they must support the same API, so it is safe to cast one to another. Equality that does not have this property is unsound.

If we want a ConcurrentCache to be modulo some other form of equivalence, then I propose that it should be a different class.

Also @JonathanStarup agrees with me.

JonathanStarup:
I think the correctness condition is that structural equality is used right?

mlutze:
> I think the correctness condition is that structural equality is used right? 

That's a possible condition, though I think any type with a well-designed API should work.

(I'm cheating of course, since I can just say anything that doesn't work is poorly designed.)

JonathanStarup:
But fx If i have
```
sealed trait Example
case object A extends Example
case object B extends Example
case object C extends Example
```
That are all equal to eachother and nothing else, then this `asInstanceOf` might still go wrong. Thats why i feel it must be structural

mlutze:
> But fx If i have
> ```
> sealed trait Example
> case object A extends Example
> case object B extends Example
> case object C extends Example
> ```
> That are all equal to eachother and nothing else, then this `asInstanceOf` might still go wrong. Thats why i feel it must be structural

And I claim that is not a reasonable equality function :)

Flix also can go wrong if you define equality like that.

mlutze:
My claim is "If two values are equal, then they should have the same type."

Types violating this property should be private or otherwise not exposed to the API of the type.

magnus-madsen:
Look at you two go 🍿 🍿 🍿

JonathanStarup:
> My claim is "If two values are equal, then they should have the same type."

Ah right, im sold. Seems like a good sentence to put on the cache if this is merged.

magnus-madsen:
> > My claim is "If two values are equal, then they should have the same type."
> 
> Ah right, im sold. Seems like a good sentence to put on the cache if this is merged.

But why should this be a requirement? We have types that have some equivalence.

JonathanStarup:
I vote neutral

mlutze:
> But why should this be a requirement? We have types that have some equivalence.

Please give an example.

If your example is `Chain`:

It has 3 cases (each with its own type) and equality is defined across these types. However, the interface correctly hides these cases. From the outside, the type is always `Chain`.

If your example is `Type`:

Equivalence among types does not use the `equals` and `hashcode` methods, because usually we don't expect `equals` and `hashcode` to be exponential operations. So we need to change the implementation to do something else.

Unfortunately, since our general cache class only knows that the element has some type `V`, the `Object` methods are all we have.

mlutze:
Claude's input:

```
Is the conclusion that there are no such types?

● Looking at my analysis, the conclusion is actually no - there are no problematic cases where different types have equal values in this codebase.

  What I found was:

  1. The DependencyEdge case: While DependencyEdge.Weak and DependencyEdge.Strong are different types that could theoretically have the same field values, their equals method is implemented correctly with type
  checking - it only returns true when comparing objects of the exact same type with matching field values.
  2. All other equality implementations in the codebase follow the same pattern - they use pattern matching to first check the type, then compare field values only within the same type.

  For example, in the Type hierarchy:
  - Type.Var only equals other Type.Var instances
  - Type.Cst only equals other Type.Cst instances
  - Type.Apply only equals other Type.Apply instances

  And in Symbol classes:
  - Each symbol type only compares equal to the same symbol type
  - Cross-type comparisons always return false

  So while there are cases where you could have foo: TypeA and bar: TypeB with identical field values, the equality implementations prevent them from being considered equal because they first check that both objects
  are of the same concrete type.

  Conclusion: No, there are no classes/objects in this codebase where equality of values does not imply equality of types. All equality implementations are type-safe.
```

magnus-madsen:
It seems you demand structural equality following the type hierarchy. But why is such a hidden implicit requirement better than the explicit casts (that are only sometimes needed)?

mlutze:
> It seems you demand structural equality following the type hierarchy. But why is such a hidden implicit requirement better than the explicit casts (that are only sometimes needed)?

Those explicit casts require knowledge of the properties of both the cache and the type that they are casting.

I want to provide users of the interface with a guarantee.

The explicit casts to me seem more like "I know how the cache interacts with this particular type so I can cast it."

If it is important that we allow weird definitions of equality, then I propose we should _never_ rely on it returning the same type, and the definitions in SimpleType should change to

```scala
  def mkArray(tpe: SimpleType): SimpleType = {
```

PR #11707, opened 2025-09-24
feat: add `SemiGroup` and `Monoid` instances to `RichString`

kukimik:
I'm writing a simple CLI app and would find these useful (so that I can use `Monoid.fold` with a `List[RichString]`).


magnus-madsen:
Thanks!

magnus-madsen:
Let us know if you have other suggestions for how to improve RichString. I think its pretty neat and plan to use it to remove our old Chalk style thing.

PR #11708, opened 2025-09-24
feat: better lexer error locations

JonathanStarup:
Instead of pointing to the start of a token, locations now point to the whole token in question. This is also better documented in the lexer errors.

This makes the errors more consistent with the rest of the compiler (pointing to full things instead of their starting point).

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
@@ -919,6 +919,7 @@ object Lexer {
     mkErrorKind(LexerError.UnterminatedBlockComment(sourceLocationFromStart()))
   }
 
+
   /** Creates a [[SourceLocation]] of the inclusive `start` and exclusive `end` in the current source. */
   private def mkSourceLocation(start: SourcePosition, end: SourcePosition)(implicit s: State): SourceLocation =
     SourceLocation(isReal = true, s.src, start, end)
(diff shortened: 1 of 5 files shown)


magnus-madsen:
> Instead of pointing to the start of a token, locations now point to the whole token in question. This is also better documented in the lexer errors.

Is this always what we want? I would think that for lexing specifically pointing to the start of the problem might be better?

JonathanStarup:
> Is this always what we want? I would think that for lexing specifically pointing to the start of the problem might be better?

The starting point is still here, this is just more information. Some examples showcase longer tokens where just the start position is lacking.

```
// Old Error
-2i32a
  ^ Wrong number sufffix

// New Error
-2i32a
  ^^^^ Wrong number suffix
```
```
def example(): String = "multiline
    string"

>> Missing '"' in string.

1 | def example(): String = "multiline
                            ^
                            String starts here.
```
```
>> Missing '*/' in block-comment.

1 | /* Returns a cool thing, crashing for "/*" */
    ^
    Block-comment starts here.
```

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -689,19 +689,13 @@ object Lexer {
...
   /** Returns `true` if `c` is recognized by `[0-9a-z._]`. */
   private def isNumberLikeChar(c: Char): Boolean =
     c.isDigit || c.isLetter || c == '.' || c == '_'
 
-  /** Consumes the remaining [[isNumberLikeChar]] characters and returns `error`. */
-  private def wrapAndConsume(error: LexerError)(implicit s: State): TokenKind = {
-    s.sc.advanceWhile(isNumberLikeChar)
-    mkErrorKind(error)
-  }
-
```
JonathanStarup, line 704: I've inlined this since it was kind of odd to begin with and now it would need two variants

magnus-madsen (approved):

PR #11711, opened 2025-09-24
refactor: collect common lexer cases per char

JonathanStarup:
Instead of a flat pattern match, combine cases that match on the same char (and is exhaustive in combination)

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
@@ -345,19 +345,20 @@ object Lexer {
           TokenKind.Dot
         }
       case '%' if s.sc.peekIs(_ == '%') => acceptBuiltIn()
-      case '$' if s.sc.peekIs(isFirstNameChar) =>
-        // Don't include the $ sign in the name.
-        s.resetStart()
-        acceptName(isUpper = false)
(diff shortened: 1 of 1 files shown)


main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -390,8 +391,9 @@ object Lexer {
             acceptUserDefinedOp()
           } else TokenKind.Underscore
         } else TokenKind.Underscore
-      case '0' if s.sc.peekIs(_ == 'x') => acceptHexNumber()
-      case c if c.isDigit => acceptNumber()
+      case c if c.isDigit =>
+        if (c == '0' && s.sc.peekIs(_ == 'x')) acceptHexNumber()
+        else acceptNumber()
```
magnus-madsen: Stylewise, I wish the outcomes would be aligned, e.g.

```
if (c == '0' && s.sc.peekIs(_ == 'x')) 
  acceptHexNumber()
else 
  acceptNumber()
```

so its easier to scan the code.

magnus-madsen: Can we do that? Otherwise the PR seems fine.

JonathanStarup:
@magnus-madsen fixed

PR #11712, opened 2025-09-24
refactor: collect lexer errors eagerly

JonathanStarup:


main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -256,11 +259,7 @@ object Lexer {
     s.resetStart()
     addToken(TokenKind.Eof)
 
-    val errors = s.tokens.collect {
-      case Token(TokenKind.Err(err), _, _, _, _, _) => err
-    }
```
JonathanStarup, line 261: Profiling says that this iteration takes quite a while.

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -298,6 +297,9 @@ object Lexer {
     val (endLine, endColumn) = if (s.start.offset != s.sc.getOffset) s.sc.getExclusiveEndPosition else s.sc.getPosition
     val e = SourcePosition.mkFromZeroIndexed(endLine, endColumn)
     s.tokens.append(Token(kind, s.src, s.start.offset, s.sc.getOffset, b, e))
+    kind match {
+      case TokenKind.Err(error) => s.errors.append(error)
+    }
     s.resetStart()
```
JonathanStarup, line 300: To be more performant, each error-site should add the error directly to avoid inspection of each token. But in this way you can't forget to add it.

magnus-madsen:
Neat!

PR #11715, opened 2025-09-24
refactor: prefer advanceIfMatch to peekIs

JonathanStarup:
Slightly simpler and faster

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
@@ -325,7 +325,7 @@ object Lexer {
         } else {
           TokenKind.Dot
         }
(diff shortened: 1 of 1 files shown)


main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -442,7 +442,6 @@ object Lexer {
 
   /** Moves current position past a built-in function (e.g. "%%BUILT_IN%%"). */
   private def acceptBuiltIn()(implicit s: State): TokenKind = {
-    s.sc.advance() // Consume `%`.
```
JonathanStarup, line 445: I promise to make a PR that makes these cursor position assumptions clear

magnus-madsen (approved):

PR #11716, opened 2025-09-24
refactor: move `advanceIfInTree` to StringCursor

JonathanStarup:
This saves an option allocation for each tree lookup.

Not sure if its worth it - feel free to merge or close.


magnus-madsen:
I suppose there is some bug...

magnus-madsen:
Not really sure about this one...

JonathanStarup:
Yeah same, seems like the code cost is too high for the payoff

PR #11718, opened 2025-09-25
refactor: explicit parser operators

JonathanStarup:
Instead of reusing token kinds as operators, introduce its own custom type.

`--Xbenchmark-phases`
- `    PR 137,2 ms (06,8%)`
- `Master 161,8 ms (08,0%)`

This has the extra benefit that the Lexer Tokens are separated from parser operators, i.e. lexing of infix operators can be made sensible.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -2382,9 +2382,9 @@ object Parser2 {
              | (TokenKind.NameLowerCase, TokenKind.Equal)
              | (TokenKind.Plus, TokenKind.NameLowerCase)
              | (TokenKind.Minus, TokenKind.NameLowerCase) =>
-          // Either `{ +y = expr | expr }` or `{ x = expr }`.
-          // Both are parsed the same and the difference is handled in Weeder.
-          recordOperation()
+            // Either `{ +y = expr | expr }` or `{ x = expr }`.
+            // Both are parsed the same and the difference is handled in Weeder.
+            recordOperation()
         case _ => block()
       }
     }
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -328,6 +328,26 @@ object Parser2 {
     }
   }
 
+  /**
+    * Look-ahead `lookahead` tokens.
+    *
+    * Consume one fuel and throws [[InternalCompilerException]] if the parser is out of fuel but
+    * does not consume fuel if parser has hit end-of-file. This lets the parser return whatever
+    * errors were produced from a deep call-stack without running out of fuel.
+    */
+  private def nthToken(lookahead: Int)(implicit s: State): Option[Token] = {
+    if (s.fuel == 0) {
+      throw InternalCompilerException(s"Parser is stuck", currentSourceLocation())
+    }
+
+    if (s.position + lookahead >= s.tokens.length - 1) {
+      None
+    } else {
+      s.fuel -= 1
+      Some(s.tokens(s.position + lookahead))
+    }
+  }
+
   /** Checks if the parser is at a token of a specific `kind`. */
   private def at(kind: TokenKind)(implicit s: State): Boolean = {
     nth(0) == kind
@@ -1399,7 +1419,7 @@ object Parser2 {
       lhs
     }
 
-    def expression(left: TokenKind = TokenKind.Eof, leftIsUnary: Boolean = false)(implicit s: State): Mark.Closed = {
+    def expression(leftOpt: Option[Op] = None)(implicit s: State): Mark.Closed = {
       implicit val sctx: SyntacticContext = SyntacticContext.Expr.OtherExpr
       var lhs = exprDelimited()
       // Handle chained calls and record lookups.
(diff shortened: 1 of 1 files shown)


main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -1515,58 +1543,136 @@ object Parser2 {
...
+
+      case object UserOperator extends Op
+    }
+
+    private def rightBindsTighter(left: Op, right: Op): Boolean = {
+      if (left.precedence == right.precedence && left.isRightAssoc) {
+        true
+      } else {
+        right.precedence > left.precedence
+      }
```
JonathanStarup: Im still a bit fuzzy on how this works, but it is totally unchanged from before

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -1515,58 +1543,137 @@ object Parser2 {
...
 
-    private def rightBindsTighter(left: TokenKind, right: TokenKind, leftIsUnary: Boolean): Boolean = {
-      def tightness(kind: TokenKind, opKind: OpKind = OpKind.Binary): Int = {
-        PRECEDENCE.indexWhere { case (k, l) => k == opKind && l.contains(kind) }
+      def rightBindsTighter(left: Op, right: Op): Boolean = {
+        if (left.precedence == right.precedence && left.isRightAssoc) {
+          true
+        } else {
+          right.precedence > left.precedence
+        }
```
JonathanStarup, line 1636: Im still a bit fuzzy on how this works, but it is totally unchanged from before

magnus-madsen:
I would need some more information. Also not sure why the number fiddling is better.

JonathanStarup:
> I would need some more information.

Current we lex
```
`Add.add`
```
As one infix token, which forces the weeder to ad-hoc parse the string inside (`token.text.split(".")..` etc). Currently this means that you can write `Example.$.$.$` as a name inside infix ticks fx (its easy to be inconsistent with actual name parsing).

This cannot be fixed without decoupling parser operators from tokens themselves. So this PR is a requirement for sensible infix parsing.

> Also not sure why the number fiddling is better.

Sure, but that function is private to `Op` so it can be rewritten however fit. Also note that these numbers also existed in the old code, just via `Array.indexWhere`. On master, retrieving the precedence of `not` requires 41 array lookups and 82 calls to `.equals`. Here it is one match

magnus-madsen:
What's the reasoning for having unary and binary operators in one enum?

magnus-madsen:
It is kind of unclear what the different ops are. E.g. DatalogCompose vs And. 

Knowing makes it clear, but by itself is seems unclear.

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -1515,58 +1543,137 @@ object Parser2 {
...
+      token.kind match {
+        case TokenKind.AngledEqual => Some(Op.Compare)
+        case TokenKind.AngledPlus => Some(Op.DatalogCompose)
+        case TokenKind.AngleL => Some(Op.Less)
+        case TokenKind.AngleLEqual => Some(Op.LessEq)
+        case TokenKind.AngleR => Some(Op.Greater)
+        case TokenKind.AngleREqual => Some(Op.GreaterEq)
+        case TokenKind.BangEqual => Some(Op.Neq)
+        case TokenKind.ColonColon => Some(Op.Cons)
+        case TokenKind.EqualEqual => Some(Op.Eq)
```
magnus-madsen: Like why are the names different?

JonathanStarup: Left is syntactic and right is semantic. My thinking is that semantic names would make the precedence list easier to read.

JonathanStarup:
> What's the reasoning for having unary and binary operators in one enum?

You can split them into a hierachy, but their precedence is compared across. Also - just to illustrate the difference towards the current code - TokenKind.Eof was used as operator. I am not opposed to doing better but I think its already an improvement.

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
JonathanStarup: The closest thing to semantics names is in Weeder. I can copy those names?
```
case "+" => Validation.Success(mkApply("Add.add"))
            case "-" => Validation.Success(mkApply("Sub.sub"))
            case "*" => Validation.Success(mkApply("Mul.mul"))
            case "/" => Validation.Success(mkApply("Div.div"))
            case "<" => Validation.Success(mkApply("Order.less"))
            case "<=" => Validation.Success(mkApply("Order.lessEqual"))
            case ">" => Validation.Success(mkApply("Order.greater"))
            case ">=" => Validation.Success(mkApply("Order.greaterEqual"))
            case "==" => Validation.Success(mkApply("Eq.eq"))
            case "!=" => Validation.Success(mkApply("Eq.neq"))
            case "<=>" => Validation.Success(mkApply("Order.compare"))
            // SEMANTIC OPS
            case "and" => Validation.Success(Expr.Binary(SemanticOp.BoolOp.And, e1, e2, tree.loc))
            case "or" => Validation.Success(Expr.Binary(SemanticOp.BoolOp.Or, e1, e2, tree.loc))
            // SPECIAL
            case "::" => Validation.Success(Expr.FCons(e1, e2, tree.loc))
            case ":::" => Validation.Success(Expr.FAppend(e1, e2, tree.loc))
            case "<+>" => Validation.Success(Expr.FixpointMerge(e1, e2, tree.loc))
```

magnus-madsen: How are they semantic? We have trait- so we don't know their semantics yet. At least not for all of them.

JonathanStarup: Semantic in the way that `<=>` is called compare since its a comparison and not AngleEqual.

I can use their syntactic name again if you want. Im indifferent

magnus-madsen:
I am OK with this but I would reuse the token names. E.g. `Slash` not `Div`.

I think translating `Slash` to `Div` gives the wrong impression.

What is the specific reason that we cannot reuse TokenKind?

JonathanStarup:
> I would reuse the token names. E.g. Slash not Div.

Fixed

> What is the specific reason that we cannot reuse TokenKind?

If we do reuse tokens, infix operators have to be one token, which means that we can't parse them uniformly with other names.
```
`Add.add`
```

Additionally, this allows all operators to be lexed uniformly (if we want) `+` could be a generic operator just like `++` since there is this middle step where precedence is computed, separated from what the actual token is.

PR #11719, opened 2025-09-25
refactor: move `.{` logic out of lexer

JonathanStarup:
It seems ill-guided to treat `.{` in the lexer instead of just handling those tokens in Parser


main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -689,21 +690,28 @@ object Parser2 {
     while (continue && !isTail && !eof()) {
       nth(0) match {
         case TokenKind.Dot =>
+          println(nth(1))
```
magnus-madsen: oops

JonathanStarup: Got distracted 😄

PR #11720, opened 2025-09-25
refactor: better Lexer map union

JonathanStarup:
Instead of converting the existing map into an iterator and building a new map, add the new keys into the existing.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
@@ -202,12 +202,16 @@ object Lexer {
     /** A string cursor on `src.data`. */
     val sc: StringCursor = new StringCursor(src.data)
 
-    /** `start` is the first position of the token that is currently being lexed. */
-    var start: Position = new Position(sc.getLine, sc.getColumn, sc.getOffset)
+    /** The first position of the token that is currently being lexed. */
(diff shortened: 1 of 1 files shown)


main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -243,8 +243,8 @@ object Lexer {
       }.unzip
 
       // Construct a map from each source to its tokens.
-      val all = results ++ fresh
-      (all.toMap, errors.flatten.toList)
+      val all = fresh.concat(results)
+      (all, errors.flatten.toList)
```
magnus-madsen, line 248: Since you are here, is flatten.toList not wasteful?

JonathanStarup: Hmm not sure, `errors` is an `Iterable` so no clue whats best. `parMap` returns an array typed as `Iterable` which is then called `unzip` on.

JonathanStarup: Maybe `errors.iterator.flatten.toList` is faster?

magnus-madsen: Hmm. Why is it an iterable?

magnus-madsen: Ah from the unzip.

magnus-madsen (approved):

PR #11724, opened 2025-09-25
feat: constrain chars to ascii

JonathanStarup:


JonathanStarup:
Actually, this program just works, pretty cool @magnus-madsen @mlutze 
```
def ƕ(): Int32 = 1 + ٥٥
def main(): Int32 = ƕ()
```
So I don't see the harm in allowing it (`٥` is "Arabic-Indic Digit Five", `ƕ` is a lowercase letter)

magnus-madsen:
I cannot reopen this, but if I could, I would. I really don't think we should allow this.

Seems like an attack vector.

JonathanStarup:
To trick the reader visually?

magnus-madsen:
> To trick the reader visually?

Yeah-- there have been some attacks

JonathanStarup:
Probably the right-to-left ones could be tricky at least.

I do really like the idea of allowing other numbers/letters but also don't know enough unicode to avoid such attacks. You want me to reopen?

magnus-madsen:
Please reopen and lock it down. 

Silly to have an effect system unable to be defeated by UTF8.

PR #11725, opened 2025-09-25
refactor: `private StringCursor.advanceUnsafe`

JonathanStarup:
A lot of double bounds checking is going on, so use `advanceUnsafe` inside StringCursor

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
@@ -1051,7 +1051,7 @@ object Lexer {
     def peekAndAdvance(): Char = {
       if (this.inBounds) {
         val c = data(offset)
-        advance()
+        advanceUnsafe()
         c
       } else {
         EOF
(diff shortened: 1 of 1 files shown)


main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -1136,7 +1139,7 @@ object Lexer {
       }
 
       for (_ <- 0 until s.length) {
-        advance()
+        advanceUnsafe()
```
JonathanStarup, line 1142: The bounds here are checked at the beginning of the function

magnus-madsen:
I am not sure whether this is worth it. Safe > performance.

JonathanStarup:
Yea I agree in general - but when its kept internal to StringCursor, which is also quite a simple class (183 lines) I think it's debatable. Ill see if it gives noticable upside

magnus-madsen:
Looking at the code, I would be slightly against this.

PR #11726, opened 2025-09-25
refactor: faster lexer while loops

JonathanStarup:


magnus-madsen:
I don't like these micro optimizations without comments. Do we even know if they have a measurable impact?

JonathanStarup:
Yeah good point. This and the `advanceUnsafe` one are better together, so ill test that to see if it has noticable impact

magnus-madsen:
Looking at the code, I feel there is almost no chance this matters.

PR #11739, opened 2025-09-28
refactor: remove `Env`, `Exec`, `FsRead`, `FsWrite`, `Net`, `Sys`

magnus-madsen:
This PR returns us to the status quo ante.

We type everything in Java world with the `IO` effect. Going forward everything else should be typed with effect handlers.

(We still retain the Chan and NonDet effects-- for now.)

### DIFF
--- a/main/test/ca/uwaterloo/flix/library/TestRedBlackTree.flix
+++ b/main/test/ca/uwaterloo/flix/library/TestRedBlackTree.flix
@@ -141,13 +141,13 @@ mod TestRedBlackTree {
         assertTrue(checkRedParentInvariant(tree))
 
     @Test
-    def testRedParentInvariant10(): Unit \ {Assert, NonDet, IO} = assertTrue(checkRedParentInvariant(random(10)))
+    def testRedParentInvariant10(): Unit \ {Assert, IO} = assertTrue(checkRedParentInvariant(random(10)))
 
     @Test
-    def testRedParentInvariant11(): Unit \ {Assert, NonDet, IO} = assertTrue(checkRedParentInvariant(random(100)))
+    def testRedParentInvariant11(): Unit \ {Assert, IO} = assertTrue(checkRedParentInvariant(random(100)))
 
     @Test
-    def testRedParentInvariant12(): Unit \ {Assert, NonDet, IO} = assertTrue(checkRedParentInvariant(random(1000)))
+    def testRedParentInvariant12(): Unit \ {Assert, IO} = assertTrue(checkRedParentInvariant(random(1000)))
 
     ///
     /// Returns `true` if and only if the binary search tree invariant holds for `tree`.


main/src/library/Time/Instant.flix
```diff
@@ -35,7 +35,7 @@ mod Time {
         /// Returns the current time from the system clock.
         ///
         @Experimental
-        pub def now(): Time.Instant \ { NonDet, IO } =
+        pub def now(): Time.Instant \ { IO } =
```
mlutze, line 38: Elsewhere we keep NonDet, right?

magnus-madsen:
@JonathanStarup @mlutze @rywiese @stephentetley Sorry for the breakages.

But I think this is the right way forward. We push everything to use effects and handlers. 

Then only those handlers that touch Java will have the `IO` effect.

main/src/library/Time/Instant.flix
magnus-madsen: We keep it mostly for channel stuff.

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Classes.json
```diff
@@ -3,50 +3,11 @@
     "java.lang.Boolean": "",
     "java.lang.Byte": "",
     "java.lang.Character": "",
-    "java.lang.Character$Subset": "",
```
JonathanStarup, line 6: Why this change?

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
```diff
@@ -1,61 +1,5 @@
...
-    "java.lang.ClassLoader::resources": "FsRead, IO",
-    "java.lang.ClassLoader::setClassAssertionStatus": "Sys, IO",
-    "java.lang.ClassLoader::setDefaultAssertionStatus": "Sys, IO",
-    "java.lang.ClassLoader::setPackageAssertionStatus": "Sys, IO",
-
-    "java.lang.Runtime::addShutdownHook": "Exec, Sys, IO",
-    "java.lang.Runtime::availableProcessors": "Env, IO",
-    "java.lang.Runtime::exec": "Env, Exec, IO",
-    "java.lang.Runtime::exit": "Sys, IO",
-    "java.lang.Runtime::freeMemory": "Env, NonDet, IO",
```
JonathanStarup, line 40: A NonDet was removed here. Wouldn't it make sense for this PR not to touch NonDet? it seems a bit muddled on top of the removal stuff

main/src/library/Time/Instant.flix
JonathanStarup: Channels already have `Chan` so I don't understand halfway deleting NonDet.

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Classes.json
magnus-madsen: Unused  by the current standard library.

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
magnus-madsen: What do you mean exactly? I think the only muddle is that I left non-det for java.util.Random.

Otherwise I removed everything except purity annotations for the wrapper types.

JonathanStarup: On master `java.lang.Runtime::freeMemory` produces `NonDet`, with this change it doesn't.

Can't this PR just avoid touching NonDet if we are not removing it?

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Classes.json
JonathanStarup: So all the remaining annotations are required for std lib compilation?

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
magnus-madsen: This PR removes all annotations of Java except for wrappers (and a few misc things)

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Classes.json
jaschdoc: Isn't it better to have explicit effect annotations still? What if someone uses it in their own library / code?

magnus-madsen: > So all the remaining annotations are required for std lib compilation?

Yes

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
JonathanStarup: I think the three things of this PR would be better separate but ill let it go as it goes

magnus-madsen:
Can we merge this?

main/src/ca/uwaterloo/flix/language/phase/typer/PrimitiveEffects.Methods.json
magnus-madsen: One issue here is that for the paper it would be easier to "revert" or rebase on top of one PR.

mlutze (approved):

PR #11741, opened 2025-09-28
fix: disallow redefining trait parameter

### LINKED ISSUE #2114: Disallow repeating class tparams in sigs
This is currently allowed:

```
lawless class C[a] {
    pub def f[a: Type](x: a): Int
}
```

But (probably?) shouldn't be.

Related: https://github.com/flix/flix/issues/2111

CasperDN:
Closes #2114. The program

```
trait C[a] {
    pub def f[a: Type](x: a): Int32
}
```
Gives
```
Syntax Error  ----


>> Redefinition of type parameter 'a'.

1 | trait C[a] {
            ^
            the original definition was here.

2 |     pub def f[a: Type](x: a): Int32
                  ^
                  the redefinition was here.


Resolution Error ----

>> Unexpected signature 'f' which does not mention the type variable of the trait.

2 |     pub def f[a: Type](x: a): Int32
                ^
                unexpected signature.

```

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/ast/WeededAst.scala
+++ b/main/src/ca/uwaterloo/flix/language/ast/WeededAst.scala
@@ -460,13 +460,16 @@ object WeededAst {
 
   case class SelectChannelRule(ident: Name.Ident, exp1: Expr, exp2: Expr, loc: SourceLocation)
 
-  sealed trait TypeParam
+  sealed trait TypeParam {
+    def ident: Name.Ident
+    def loc: SourceLocation
+  }
 
   object TypeParam {
 
-    case class Unkinded(ident: Name.Ident) extends TypeParam
+    case class Unkinded(ident: Name.Ident, loc: SourceLocation) extends TypeParam
 
-    case class Kinded(ident: Name.Ident, kind: Kind) extends TypeParam
+    case class Kinded(ident: Name.Ident, kind: Kind, loc: SourceLocation) extends TypeParam
 
   }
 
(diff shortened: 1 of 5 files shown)


magnus-madsen:
Thanks @CasperDN ! @mlutze is the expert here, so I will let him review.

mlutze:
This is good, but I think it's happening in the wrong place. I'd expect instead this to be a redundancy check for shadowing type parameter.

CasperDN:
@mlutze Agreed. I've moved the check to `Redundancy`. I haven't used `ParOps` as I doubt it makes sense for the type of check I do.

main/src/ca/uwaterloo/flix/language/phase/Redundancy.scala
```diff
@@ -299,6 +304,17 @@ object Redundancy {
       }
   }
 
+  /**
+    * Finds unused type parameters.
+    */
+  private def findShadowedTraitTypeParams(traitt: Trait): Used = {
```
mlutze: The documentation and name of this function don't seem to agree.

main/src/ca/uwaterloo/flix/language/phase/Redundancy.scala
```diff
@@ -299,6 +304,17 @@ object Redundancy {
...
+    val traitParam = traitt.tparam
+    val traitParamName = traitParam.name.name
+    val dummyEnv = Env.empty + (traitParamName, traitParam.loc)
```
mlutze: Just `env` is fine :)

PR #11747, opened 2025-09-29
feat: add `PackageTrust`

jaschdoc:
Related to https://github.com/flix/flix/issues/11436

### DIFF
--- a/main/src/ca/uwaterloo/flix/api/PackageTrust.scala
+++ b/main/src/ca/uwaterloo/flix/api/PackageTrust.scala
@@ -55,83 +55,285 @@ object PackageTrust {
   }
 
   private def visitExp(expr0: TypedAst.Expr)(implicit sctx: SharedContext): Unit = expr0 match {
-    case Expr.Cst(cst, tpe, loc) => ???
-    case Expr.Var(sym, tpe, loc) => ???
-    case Expr.Hole(sym, scp, tpe, eff, loc) => ???
-    case Expr.HoleWithExp(exp, scp, tpe, eff, loc) => ???
-    case Expr.OpenAs(symUse, exp, tpe, loc) => ???
-    case Expr.Use(sym, alias, exp, loc) => ???
-    case Expr.Lambda(fparam, exp, tpe, loc) => ???
-    case Expr.ApplyClo(exp1, exp2, tpe, eff, loc) => ???
-    case Expr.ApplyDef(symUse, exps, targs, itpe, tpe, eff, loc) => ???
-    case Expr.ApplyLocalDef(symUse, exps, arrowTpe, tpe, eff, loc) => ???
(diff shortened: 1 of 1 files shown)


main/src/ca/uwaterloo/flix/api/PackageTrust.scala
```diff
@@ -0,0 +1,354 @@
...
+import ca.uwaterloo.flix.language.ast.TypedAst
+import ca.uwaterloo.flix.language.ast.TypedAst.Expr
+import ca.uwaterloo.flix.util.ParOps
+
+import java.util.concurrent.ConcurrentLinkedQueue
+import scala.jdk.CollectionConverters.CollectionHasAsScala
+
+object PackageTrust {
+
+  def run(root: TypedAst.Root)(implicit flix: Flix): List[SuspiciousExpr] = {
```
magnus-madsen, line 27: I would really encourage you to add this to the Safety phase. And just have it be Safety errors.

magnus-madsen:
@jaschdoc Added a comment, just want to make sure you see it. 

I think building on the Safety phase is simplest and best.

magnus-madsen:
I would focus on just one unsafe feature at a time, e.g. get unchecked_cast to work.

Then adding the rest will be easy.

jaschdoc:
Okay that seems like a good idea, but it seems like a code smell to hard code it to functions located in libraries. Should we maybe also consider defining a trust level for the package we are defining? Not sure how that would interact with library code then, so maybe that's a thing for the not-so-near-future.

jaschdoc:
Also, if you want it in `Safety` how do you suggest passing the list of library trust values (from the toml file) to the Safety phase? It seems to me like we are mixing two different concepts i) language errors and ii) package management.

jaschdoc:
I can open another PR that refactors `SecurityContext` to the trust levels we agreed upon and we can go from there. It seems like this is the intention for `SecurityContext` but it never really went anywhere.

magnus-madsen:
> I can open another PR that refactors `SecurityContext` to the trust levels we agreed upon and we can go from there. It seems like this is the intention for `SecurityContext` but it never really went anywhere.

Yes.

magnus-madsen:
> Okay that seems like a good idea, but it seems like a code smell to hard code it to functions located in libraries. Should we maybe also consider defining a trust level for the package we are defining? Not sure how that would interact with library code then, so maybe that's a thing for the not-so-near-future.

What do you mean? When you add a dependency you specify its trust-level/security-context/whatever...

magnus-madsen:
> Also, if you want it in `Safety` how do you suggest passing the list of library trust values (from the toml file) to the Safety phase? It seems to me like we are mixing two different concepts i) language errors and ii) package management.

Each source location knows where it comes from and hence its security context.

PR #11748, opened 2025-09-30
feat: add `MutHashMap`

magnus-madsen:


magnus-madsen:
@mlutze This is ready. The core algorithm looks OK. The use of `Option` is sub-optimal for performance, but clean. 

I think we agreed that its better for the compiler to deal with options that doing hacky things.

main/src/library/MutHashMap.flix
```diff
@@ -0,0 +1,817 @@
...
+    /// Returns the value associated with key `k`, or `default` if not found.
+    ///
+    pub def getWithDefault(k: k, default: v, m: MutHashMap[k, v, r]): v \ r with Eq[k], Hash[k] =
+        Option.getWithDefault(default, get(k, m))
+
+    ///
+    /// Converts hash code to a valid bucket index.
+    /// Handles negative hashes correctly using bitwise AND.
+    ///
+    def hash2bucket(hash: Int32, numBuckets: Int32): Int32 =
```
JonathanStarup: seems like a pretty odd name to me

main/src/library/MutHashMap.flix
```diff
@@ -0,0 +1,817 @@
...
+        pub def getKey(e: Entry[k, v, r]): k = e->key
+        pub def getValue(e: Entry[k, v, r]): v \ r = e->value
+        pub def setValue(v: v, e: Entry[k, v, r]): Unit \ r = e->value = v
+        pub def getHash(e: Entry[k, v, r]): Int32 = e->hash
+        pub def getBucketNext(e: Entry[k, v, r]): Option[Entry[k, v, r]] \ r = e->bucketNext
+        pub def setBucketNext(next: Option[Entry[k, v, r]], e: Entry[k, v, r]): Unit \ r = e->bucketNext = next
+        pub def getOrderPrev(e: Entry[k, v, r]): Option[Entry[k, v, r]] \ r = e->orderPrev
+        pub def setOrderPrev(prev: Option[Entry[k, v, r]], e: Entry[k, v, r]): Unit \ r = e->orderPrev = prev
+        pub def getOrderNext(e: Entry[k, v, r]): Option[Entry[k, v, r]] \ r = e->orderNext
+        pub def setOrderNext(next: Option[Entry[k, v, r]], e: Entry[k, v, r]): Unit \ r = e->orderNext = next
```
JonathanStarup, line 90: imo these should have space between

main/src/library/MutHashMap.flix
```diff
@@ -0,0 +1,817 @@
...
+                if (k == Entry.getKey(entry))
+                    Some(entry)
+                else
+                    findInBucket(k, Entry.getBucketNext(entry))
+        }
+
+    ///
+    /// Returns `true` if key `k` exists in `m`.
+    ///
+    pub def memberOf(k: k, m: MutHashMap[k, v, r]): Bool \ r with Eq[k], Hash[k] =
```
JonathanStarup, line 216: don't we have `pub` before other methods generally?

main/src/library/MutHashMap.flix
```diff
@@ -0,0 +1,817 @@
...
+    /// Returns an iterator over keys in `m` in insertion order.
+    ///
+    pub def iteratorKeys(rc: Region[r2], m: MutHashMap[k, v, r]): Iterator[k, r2 + r, r2] \ {r, r2} =
+        iterator(rc, m) |> Iterator.map(fst)
+
+    ///
+    /// Returns an iterator over values in `m` in insertion order.
+    ///
+    pub def iteratorValues(rc: Region[r2], m: MutHashMap[k, v, r]): Iterator[v, r2 + r, r2] \ {r, r2} =
+        iterator(rc, m) |> Iterator.map(snd)
```
JonathanStarup, line 511: Given our move away from iterators, maybe these two shorthands are a bit unneccesary

main/src/library/MutHashMap.flix
```diff
@@ -0,0 +1,817 @@
...
+            case Some(entry) =>
+                Entry.setValue(f(Entry.getKey(entry), Entry.getValue(entry)), entry);
+                loop(Entry.getOrderNext(entry))
+        };
+        loop(m->head)
+
+    ///
+    /// Removes all entries from `m` where `f(value)` returns false.
+    ///
+    pub def refine(f: v -> Bool \ ef, m: MutHashMap[k, v, r]): Unit \ {ef, r} with Eq[k], Hash[k] =
```
JonathanStarup, line 554: Should be called `filter`

main/src/library/MutHashMap.flix
```diff
@@ -0,0 +1,817 @@
...
+                let newAcc = if (f(v)) acc else k :: acc;
+                collectKeys(Entry.getOrderNext(entry), newAcc)
+        };
+        let keysToRemove = collectKeys(m->head, Nil);
+        List.forEach(k -> remove(k, m), keysToRemove)
+
+    ///
+    /// Removes all entries from `m` where `f(key, value)` returns false.
+    ///
+    pub def refineWithKey(f: (k, v) -> Bool \ ef, m: MutHashMap[k, v, r]): Unit \ {ef, r} with Eq[k], Hash[k] =
```
JonathanStarup, line 560: also filter

JonathanStarup:
I don't think I buy into intersion-order hashmap. For the Flix compiler, using iteration-order hashmap instead of the normal hashmap would fix no problems. It seems like an aesthetic feature to me.

main/src/library/MutHashMap.flix
mlutze: The use of "2" for "to" always makes me feel like I'm back in 2008 reading a text.

main/src/library/MutHashMap.flix
```diff
@@ -0,0 +1,817 @@
...
+    ///
+    pub def getWithDefault(k: k, default: v, m: MutHashMap[k, v, r]): v \ r with Eq[k], Hash[k] =
+        Option.getWithDefault(default, get(k, m))
+
+    ///
+    /// Converts hash code to a valid bucket index.
+    /// Handles negative hashes correctly using bitwise AND.
+    ///
+    def hash2bucket(hash: Int32, numBuckets: Int32): Int32 =
+        Int32.bitwiseAnd(hash, numBuckets - 1)
```
mlutze: I don't understand what's happening here. Why not `mod`?

main/src/library/MutHashMap.flix
magnus-madsen: Hah. Suggestions?

magnus-madsen: getBucket? lookupBucket?

magnus-madsen: indexOfBucket?

main/src/library/MutHashMap.flix
magnus-madsen: Between the defs or where?

main/src/library/MutHashMap.flix
magnus-madsen: What do you mean?

main/src/library/MutHashMap.flix
magnus-madsen: Some of the API is to mirror MutMap. We can remove them, but lets do it together. So rather have them in now and them kill them everywhere if that is what we want.

main/src/library/MutHashMap.flix
magnus-madsen: Actually not, because its return type is Unit.

Same for transform not being called map.

You can check MutMap API.

main/src/library/MutHashMap.flix
magnus-madsen: Recall: The hash may be negative. The modulo with a negative number can be negative!

mlutze: The hash can be negative but the number of buckets cannot, right?

Flix distinguishes between modulo (always nonnegative) and remainder (can be negative), so we should be able to use modulo here.

magnus-madsen: I am not an expert, but I just tried modulo and it became negative.

magnus-madsen:
```
println(Int32.modulo(5, -11))
```

magnus-madsen: Google suggests that bitwise and is OK as long as the length (number of buckets) is a power of two.

main/src/library/MutHashMap.flix
mlutze: All of those sound fine to me.

main/src/library/MutHashMap.flix
mlutze: > ```
> println(Int32.modulo(5, -11))
> ```

That's probably because you're doing it backward.

Here's the text of the modulo documentation:

```
Returns the Euclidean modulo of x and n. The result is always non-negative.
```

Clearly you just proved the documentation wrong, but as long as the base (2nd argument) is positive, the result will always be positive.

main/src/library/MutHashMap.flix
magnus-madsen: @JonathanStarup Which do you prefer?

main/src/library/MutHashMap.flix
JonathanStarup: I agree, should be mod

main/src/library/MutHashMap.flix
JonathanStarup: public and private methods are mixed one after another

main/src/library/MutHashMap.flix
JonathanStarup: yea line breaks

main/src/library/MutHashMap.flix
JonathanStarup: bucketIndex

JonathanStarup:
> I don't think I buy into intersion-order hashmap. For the Flix compiler, using iteration-order hashmap instead of the normal hashmap would fix no problems. It seems like an aesthetic feature to me.

@magnus-madsen any comment to this? I feel like its a surface level fix to the problems of hashmap iteration and its "costs" code and complexity.

main/src/library/MutHashMap.flix
magnus-madsen: I can change it to mod. But it should be fine with bitwiseAnd. It just picks the last `i` bits which is fine for power-two.

magnus-madsen:
> > I don't think I buy into intersion-order hashmap. For the Flix compiler, using iteration-order hashmap instead of the normal hashmap would fix no problems. It seems like an aesthetic feature to me.

I mean; it would have fixed that issue with Boolean unification variable unification order being unstable leading to non-deterministic crashes...

A linked hash-map has many advantages:

- Iteration order is well-defined, so it can implement everything from `forEach` to `foldLeft`.
- Iteration order is preserved: so adding one element and removing the same element does not change the order of elements.

The cost is slightly higher insertion cost, but on the other hand it has much faster traversal. 

Finally, there are many hashing strategies, e.g. linear probing or cuckoo hashing. If you need a very specific set of performance characteristics (e.g. the weights between the different operations: insert/find/traverse) you have to implement your own.

main/src/library/MutHashMap.flix
magnus-madsen: Ah yes. I tried to figure that out too. It seems somewhat inconsistent. In other modules the private functions sometimes come right after they are used. I know we have been using that scheme too, so I kept it here.

main/src/library/MutHashMap.flix
magnus-madsen: Fixed.

main/src/library/MutHashMap.flix
magnus-madsen: Fixed

main/src/library/MutHashMap.flix
magnus-madsen: Fixed.

magnus-madsen:
I think I can easily sum it up now: A non-linked hash table is leaky abstraction.

mlutze:
> I think I can easily sum it up now: A non-linked hash table is leaky abstraction.

I'm convinced.

JonathanStarup:
> I think I can easily sum it up now: A non-linked hash table is leaky abstraction.

That's a nice thing to say, but i still don't believe it fixes anything. It's like `zip` that ignores elements silently to have a "nice" definition.

magnus-madsen:
So can we merge this? Then I can do MutHashSet

JonathanStarup:
> I mean; it would have fixed that issue with Boolean unification variable unification order being unstable leading to non-deterministic crashes...

No? If we had derministic insertion order then a regular hashmap also has deterministic iteration, which is kind of my point

JonathanStarup:
> I mean; it would have fixed that issue with Boolean unification variable unification order being unstable leading to non-deterministic crashes...

i dont believe so. thats only true if we have deterministic insertion order which im pretty sure we dont have

mlutze:
> So can we merge this? Then I can do MutHashSet

Maybe Jonathan can handle it, but it's too big for me to review atm. That's why I proposed splitting it up.

JonathanStarup:
Also, all my critique goes away if this is called `MutLinkedHashMap` like in Java. I think in general we should be more specific about what things are

main/src/library/MutHashMap.flix
JonathanStarup: Looks unchanged to me

JonathanStarup (approved):

Issue #11740, opened 2025-09-28
Lenient parsing of trait type parameters


CasperDN:
The following is executable and runs just fine printing `2`.
```
trait C[m, a] {
    pub def f(x: m): Int32
}

instance C[Int32] {
    pub def f(x: Int32): Int32 = x
}

def main4(): Unit \ IO = {
    println(C.f(2))
}
```
I believe `Parser2.traitDecl` should avoid using `Type.parameters()` as the rest of the engine expects exactly 1 type parameter. Alternatively the weeder should catch and disallow this (might lead to better parsing and errors of bad programs).

JonathanStarup:
IMO, the way to have good error messages is to avoid lexer/parser errors for things that look right if you squint your eyes. So i'd say this should be a Weeder check.

PR #11749, opened 2025-09-30
fix: require exactly one trait parameter

### LINKED ISSUE #8038: Traits allow more than one parameter type parameter
Traits should only allow one type parameter. Yet `trait MyTrait[t1, t2] {}` compiles without issues.

CasperDN:
Fixes #8038 and duplicate #11740.


mlutze (commented):
Looks reasonable. Just needs a test :)

magnus-madsen:
And this does not crash the compiler later? (Letting more than one or zero params through?)

CasperDN:
I haven't changed anything in what the compiler (weeder) produces, only added errors when the number of parameters is not 1.

mlutze (approved):

PR #11750, opened 2025-09-30
fix: allow comments between `use`/`import`

### LINKED ISSUE #7925: Import parsing error
The following code throws an error:

```scala
import javax.swing.SwingUtilities;
//import javax.swing.SwingUtilities;
import java.lang.Runnable;

def main(): Unit \ IO =
    println("hello")
```

```
-- Parse Error (OtherDecl) -------------------------------------------------- /home/csok/IdeaProjects/flix/test.flix

>> Expected <declaration> before 'import'.

3 | import java.lang.Runnable;
    ^^^^^^
    Here
```
It doesn't happen if the first line of the program is a comment, only for in-between comments in imports.

CasperDN:
Fixes #7925

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -811,6 +811,7 @@ object Parser2 {
     val mark = open(consumeDocComments = false)
     var continue = true
     while (continue && !eof()) {
+      comments()
       nth(0) match {
         case TokenKind.KeywordUse =>
           use()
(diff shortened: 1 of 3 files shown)


magnus-madsen:
Does this break `java -jar flix.jar doc`? I.e. would it eat the doc comments or is it fine?

CasperDN:
> Does this break `java -jar flix.jar doc`? I.e. would it eat the doc comments or is it fine?

`comments` has `consumeDocComments = false` as default.

Running `java -jar flix.jar doc` with the new jar creates a `build/doc`. The 2 files I looked at were fine, I think.

PR #11751, opened 2025-09-30
refactor: remove `BiasedCoin` and `RandomCoin`

magnus-madsen:


magnus-madsen:
These were really just some early effects-- so that we had _something_. But now we have several real effects- and these seem to have low value and would belong in a library anyway.

PR #11752, opened 2025-09-30
refactor: remove `Duration`, `Epoch`, `Instant`, `Thread`, `Time`

magnus-madsen:


main/src/library/Time/Epoch.flix
```diff
@@ -1,45 +0,0 @@
...
-
-mod Time.Epoch {
-    import java.lang.System
-
-    ///
-    /// Returns the number of nano seconds since the Unix epoch.
-    ///
-    @Experimental
-    pub def nanoseconds(): Int64 \ IO =
-        System.nanoTime()
```
mlutze, line 25: I'm pretty sure this isn't epoch time anyway.

PR #11753, opened 2025-10-01
feat: parse bad names better

### LINKED ISSUE #9534: Parser is stuck: capital lambda
```
def bar(): String = 
    Abc -> foo()
```

CasperDN:
Change: Allow `nameUnqualified` to eat any name-like tokens instead of the specified on error. Inspired by and closes #9534.

Question: Should `NAME_LIKE` include more 'names'?

Reduces the errors from 18 to 6 in the following program:
```
mod Bar {
    pub def _(): Int32 = 123

    pub def a(): Unit = {
        let _ = #{
            A :- not b.
        };
        ()
    }

    type alias c = Int32

    pub enum d {
        case d(Int32)
    }

    instance e[d] {}

    trait e[x] {}
}
```

Also why is `trait e[x] {}` allowed? You cannot implement it...

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -428,7 +428,8 @@ object Parser2 {
 
   /**
     * Advance past current token if it is of kind in `kinds`. Otherwise wrap it in an error.
-    * Optionally returns the error mark
+    * Returns true if it successfully advances.
+    * Returns false otherwise.
     */
   private def expectAny(kinds: Set[TokenKind], hint: Option[String] = None)(implicit sctx: SyntacticContext, s: State): Boolean = {
     if (eatAny(kinds)) {


JonathanStarup:
I've been thinking for a while about how to handle casing errors better. This seems like a good small change 👍

CasperDN:
I would argue that the parsing should be done without relying on what the names are, at all. Of course an error would still be issued for a bad name and the parser could still be responsible for it, but so could the weeder.

JonathanStarup:
> I would argue that the parsing should be done without relying on what the names are, at all. Of course an error would still be issued for a bad name and the parser could still be responsible for it, but so could the weeder.

I agree yea

The Java stuff is the most tricky in that regard, fx `System.out.println()`.

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -426,10 +426,13 @@ object Parser2 {
     closeWithError(mark, error)
   }
 
-  /** Advance past current token if it is of kind in `kinds`. Otherwise wrap it in an error. */
-  private def expectAny(kinds: Set[TokenKind], hint: Option[String] = None)(implicit sctx: SyntacticContext, s: State): Unit = {
+  /**
+    * Advance past current token if it is of kind in `kinds`. Otherwise wrap it in an error.
+    * Optionally returns the error mark
```
JonathanStarup: comment doesn't match the behaviour

CasperDN: Forgot to update after deciding that a bool was sufficient.

magnus-madsen (approved):

magnus-madsen:
Thanks!

PR #11756, opened 2025-10-01
refactor: introduce `Shuffle` effect

magnus-madsen:


main/src/library/Array.flix
```diff
@@ -1634,21 +1633,15 @@ mod Array {
...
-                loop(i - 1)
-            }
-            else
-                ()
-        };
-        loop(len - 1)
-    } as _ \ {r, NonDet})
+        let perm = Shuffle.permutation(len);
+        // Apply the permutation: result[i] = a[perm[i]]
+        let result = Vector.init(i -> Array.get(Vector.get(i, perm), a), len);
```
magnus-madsen, line 1645: This is necessary to avoid "double swaps". Otherwise we are not actually applying the permutation.

magnus-madsen:
@JonathanStarup @mlutze Thoughts on this? I am still leaning towards just removing shuffle.

magnus-madsen:
Missing, if we want this:

- handlers
- fix tests

magnus-madsen:
@mlutze I decided - for now - to go with a Shuffle effects. I think its better than NonDet (i.e. the library should use effects that can be handled). I agree in time that it would be better to have some Shuffle module etc.

mlutze (approved):

PR #11758, opened 2025-10-01
refactor: remove `sum`, `sumWith`, `sumKeys`, `sumValues` from sets and maps

magnus-madsen:


magnus-madsen:
I think we added a lot of these for the purity reflection work, but I am not sure its a good idea to have e.g. `Set.sum`. Seems dangerous. Moreover, these functions seem really niche.

JonathanStarup (approved):
As flix matures, `@deprecated` should probably be used. Also avoids community builds giving red x'es

magnus-madsen:
> As flix matures, `@deprecated` should probably be used. Also avoids community builds giving red x'es

Yeah, post 1.0 we will have to do that.

PR #11761, opened 2025-10-01
experiment: add `Console.promptInt32`

magnus-madsen:


magnus-madsen:
@mlutze Thoughts on this?

main/src/library/Console.flix
```diff
@@ -50,6 +50,24 @@ mod Console {
...
+    ///
+    /// Repeatedly prints `msg` and reads input until a valid integer in the given range `[start; end)` is entered.
+    ///
+    pub def promptInt32(start: {start = Int32}, end: {end = Int32}, msg: a): Int32 \ Console with ToString[a] =
+        Console.println("${msg}");
+        match readInt32() {
+            case Some(i) if (start#start <= i and i < end#end) => i
+            case _                                             => promptInt32(start, end, msg)
+        }
+
```
JonathanStarup, line 70: I think the generalized version is also pretty nice. This also to do allow `quit` to exit the loop fx
```flix
///
/// Reads a parsed input from the console with a prompt.
///
/// Repeatedly prints `msg` and reads input until a valid input is entered.
///
pub def prompt(parse: String -> Option[t] \ ef, msg: a): t \ Console + ef with ToString[a] =
    Console.println("${msg}");
    match Console.readln() |> parse {
        case Some(v) => v
        case _       => prompt(parse, msg)
    }
```
```flix
prompt(Int32.fromString >> (i -> start <= i and i < end))
```

main/src/library/Console.flix
```diff
@@ -50,6 +50,24 @@ mod Console {
...
+    ///
+    pub def readInt32(): Option[Int32] \ Console =
+        Console.readln() |> Int32.fromString
+
+    ///
+    /// Reads a 32-bit integer from the console with a prompt.
+    ///
+    /// Repeatedly prints `msg` and reads input until a valid integer in the given range `[start; end)` is entered.
+    ///
+    pub def promptInt32(start: {start = Int32}, end: {end = Int32}, msg: a): Int32 \ Console with ToString[a] =
```
mlutze, line 64: I think this pattern encourages bad UX.

It's rare that the right thing to do is prompt the user with the same exact message without telling them what's wrong with the original.

mlutze: I don't love the msg being `a` either. Seems fishy to prompt the user with e.g. a Boolean.

mlutze (commented):
Seems reasonable but I think the API isn't quite right.

main/src/library/Console.flix
JonathanStarup: > I think this pattern encourages bad UX.

Also my thinking, the current code also has no way to quit the loop or cancel.

In the generalized version `parse: String -> Result[String, t]` could be used to give feedback

Issue #11590, opened 2025-09-13
Problematic reuse of `Pattern.pattern` for let parsing.


CasperDN:
Current code:
```
    private def letMatchExpr()(implicit s: State): Mark.Closed = {
      implicit val sctx: SyntacticContext = SyntacticContext.Expr.OtherExpr
      assert(at(TokenKind.KeywordLet))
      val mark = open()
      expect(TokenKind.KeywordLet)
      Pattern.pattern()
      if (eat(TokenKind.Colon)) {
        Type.ttype()
      }
      expect(TokenKind.Equal)
      statement(rhsIsOptional = false)
      close(mark, TreeKind.Expr.LetMatch)
    }
```
The problem is the `Pattern.pattern()`. It is to permissive. It allows the following to parse, though it is absurd:
```
def main(): Unit \ IO =
    // Problematic
    let "-" = ???;
    let 2 = ???;
    let _ :: Nil = ???;
```
All of them reports `Non-exhaustive match. Missing case: '_'` or `Non-exhaustive match. Missing case: 'Nil'` for the last one.

While the reuse of `Pattern.pattern()` is nice the error reporting is bad also shown by the current error of #8261. Simplifying the let there to `let- = ???;` we get `Malformed Int` for `- =` and `Expected '=' before HoleAnonymous` for `???`.

magnus-madsen:
What's the proposal? We do want let to be able to destruct.

CasperDN:
1. Duplicating and modifying `pattern` and relating functions to e.g. `patternNoConstants`.
2. Detecting constants in the pattern in the weeder and rejecting here.
3. Make `pattern` and relating functions take a bool which defaults to true: `allowConstants: Bool = true`

The first result in some problematic code duplication and maintenance.

The weeder solution might okay-ish. It would need to check the pattern tree recursively for `TreeKind.Pattern.Literal`, `TreeKind.Pattern.Unary` and `TreeKind.Pattern.FCons`, possibly more.

The last is also okay-ish. `pattern` becomes a bit more complex. It would mean that we reject the pattern during parsing which might cause an avalanche of errors (`Enum.Case(1, x, y, z)` does not _parse_). On the other hand as #8261 shows the current parsing is already brittle on simple programs. I guess we could also just issue an error, but attempt to still parse the pattern.

JonathanStarup:
I like `Detecting constants in the pattern in the weeder and rejecting here.` because it gives better errors as you say. but 3 is also an improvement

[...]

PR #11762, opened 2025-10-01
feat: prohibit constants in destructuring

CasperDN:
Disallow constants in destructuring patterns.

Overall change: For destructuring assignments, except `match`, constants will be reported as an error and replaced by `Pattern.Error` to prevent further errors (concretely `Non-exhaustive match`).

What will users experience?

1. Better error messages
2. Expressions of the form `let A.A = ???` are now illegal independent of how the enum `A` looks.

Reasoning:

1. The current error reporting is bad. See #11590.
2. Destructuring with constants is useless. It can be replaced by destructuring to `_`. The only exception would be to show what the value is, but I think that is a niche concern.

Not handled: Destructuring with `x :: xs`. It will always be reported as non-exhaustive, but we do not throw an error. It can be caught by the following, but the error is bad:
```
if (qname.ident.name == "Cons" && qname.namespace.idents.map(_.name) == List("List")) {
  // `_ :: _ `. Reason: Not exhaustive.
  sctx.errors.add(IllegalConstantPattern(loc))
  return Pattern.Error(loc)
}
```

With this change all of the following will be reported as errors:
```
enum A {
    case A
}
def f(): Unit \ IO = {
    let {a = A.A | _} = {a = A.A};
    let A.A = A.A;
    foreach(A.A <- A.A :: Nil) {
        println("")
    };
    let _ = forM(A.A <- A.A :: Nil) yield 2;
    let _ = forA(A.A <- A.A :: Nil) yield 2;
    let _ = par(A.A <- A.A) yield Nil;
    let 2 = ???;
    let (a, 2) = ???;
    let (b, (c, 3)) = ???;
    ()
}
```

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
+++ b/main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
@@ -701,13 +701,13 @@ object WeederError {
     * @param loc the location where the constant pattern occurs.
     */
   case class IllegalConstantPattern(loc: SourceLocation) extends WeederError {
-    def summary: String = "Illegal constant pattern"
+    def summary: String = "Unexpected constant pattern"
 
     def message(formatter: Formatter): String = {
       import formatter.*
-      s""">> Illegal constant pattern.
+      s""">> Unexpected constant pattern.
          |
-         |${code(loc, "Constants not allowed in destructuring pattern.")}
+         |${code(loc, "Constants are not allowed in let or lambda matches.")}
          |""".stripMargin
     }
   }
--- a/main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
+++ b/main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
@@ -696,6 +696,23 @@ object WeederError {
   }
 
   /**
+    * An error raised to indicate an illegal constant pattern.
+    *
+    * @param loc the location where the constant pattern occurs.
+    */
+  case class IllegalConstantPattern(loc: SourceLocation) extends WeederError {
+    def summary: String = "Illegal constant pattern"
+
+    def message(formatter: Formatter): String = {
+      import formatter.*
+      s""">> Illegal constant pattern.
+         |
+         |${code(loc, "Constants not allowed in destructuring pattern.")}
+         |""".stripMargin
+    }
+  }
+
+  /**
     * An error raised to indicate illegal syntax: empty tuple type.
     *
     * @param loc the location where the error occurs.
--- a/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
@@ -2495,6 +2495,11 @@ object Weeder2 {
         ExtTagPattern.Error(pat.loc)
     }
 
+    /**
+      * Attempts to verify that there are no constants in `pat`.
+      *
+      * Constants are reported as errors to `sctx` and replaced by `Pattern.Error`.
+      */
     def restrictToNonConstant(pat: Pattern)(implicit sctx: SharedContext): Pattern = pat match {
       case Pattern.Cst(_, loc) =>
         sctx.errors.add(IllegalConstantPattern(loc))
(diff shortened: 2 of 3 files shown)


jaschdoc:
Why do you want to disallow single-constructor enums? These will always succeed and saves you a match expression.

CasperDN:
I'm not disallowing single-constructor enums. I'm disallowing 0-ary cases. The following will still be fine:
```
enum B {
    case B(Int32)
}
def f1(): Unit \ IO = {
    let B.B(x) = B.B(2);
    println(x)
}
```

CasperDN:
Should the used `loc` in the `Pattern.Error` be created with `loc.asSynthetic`?

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -2495,6 +2495,29 @@ object Weeder2 {
         ExtTagPattern.Error(pat.loc)
     }
 
+    def restrictToNonConstant(pat: Pattern)(implicit sctx: SharedContext): Pattern = pat match {
```
JonathanStarup: `private def`

Even though the weeder is not rich in doc comments, I think it'd be nice to add one here, just noting that errors are pushed to `sctx` and replaced by Pattern.Error

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -2495,6 +2495,29 @@ object Weeder2 {
...
+      case Pattern.Cst(_, loc) =>
+        sctx.errors.add(IllegalConstantPattern(loc))
+        Pattern.Error(loc)
+      case Pattern.Tag(qname, pats, loc) =>
+        if (pats.isEmpty) {
+          // Disallow `A.A`
+          sctx.errors.add(IllegalConstantPattern(loc))
+          return Pattern.Error(loc)
```
JonathanStarup: I think a regular else branch without the early return is better

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -2495,6 +2495,29 @@ object Weeder2 {
...
+        }
+        Pattern.Tag(qname, pats.map(restrictToNonConstant), loc)
+      case Pattern.Tuple(pats, loc) =>
+        Pattern.Tuple(pats.map(restrictToNonConstant), loc)
+      case Pattern.Record(pats, restPat, loc) =>
+        val newPats = pats.map(innerPat => Pattern.Record.RecordLabelPattern(
+            label = innerPat.label,
+            pat = innerPat.pat.map(restrictToNonConstant),
+            loc = innerPat.loc
```
JonathanStarup: this might be prettier with `innerpat.copy(pat = ..)`

main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
```diff
@@ -695,6 +695,23 @@ object WeederError {
...
+
+    def message(formatter: Formatter): String = {
+      import formatter.*
+      s""">> Illegal constant pattern.
+         |
+         |${code(loc, "Constants not allowed in destructuring pattern.")}
+         |""".stripMargin
+    }
+  }
+
```
jaschdoc, line 714: Can we change the wording from `illegal` to `unexpected`? It's less frightening for the user. Also I think "Constants not allowed in destructuring pattern." should be changed to something like "Constants are not allowed in let-patterns" or maybe "let-bound patterns".
Generally, too machine-like and not human centered.

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
jaschdoc: or add a helper function

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -2495,6 +2495,29 @@ object Weeder2 {
...
+          ))
+        Pattern.Record(newPats, restPat, loc)
```
jaschdoc: Just as a naive question, why is `restPat` not visited? Is it always restricted to variables or default `_`?

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
CasperDN: Agreed. Originally there were 2 `if` constructs before the final returned value.

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
jaschdoc: Ah it is ok, but maybe can rename it to `rowPattern`?

main/src/ca/uwaterloo/flix/language/errors/WeederError.scala
CasperDN: I used `destructuring` because that's what JS calls them.

The error is also used for functions with `match`. I forgot to add it to the list of examples.

CasperDN: I've changed it to `Constants are not allowed in let or lambda matches.`.

main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
CasperDN: Not `private def`, right? `Patterns` is private, but the method is called from outside.

JonathanStarup: Ah right, can you move it up to the other non-private defs then?

CasperDN: Sure. I placed it there since it was similar to `restrictToVarOrWild`.

magnus-madsen (approved):

magnus-madsen:
Great; thanks!

PR #11763, opened 2025-10-02
refactor: Improve StringCursor interface

JonathanStarup:


main/src/ca/uwaterloo/flix/util/StringCursor.scala
```diff
@@ -0,0 +1,232 @@
...
+      Some(chars(offset))
+    } else {
+      None
+    }
+  }
+
+  /**
+    * Returns `p(this.nth(n).get)` if the cursor is inbounds.
+    *
+    * Returns `false` if the cursor is out of bounds.
```
mlutze: ):<

JonathanStarup:
If you have time to review @mlutze I'd say this can go in without taking @magnus-madsen time

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -1047,143 +1054,157 @@ object Lexer {
...
-    def nthIsPOrOutOfBounds(n: Int, p: Char => Boolean): Boolean = {
-      val index = offset + n
-      if (0 <= index && index < data.length) {
-        p(data(index))
+    /**
+      * Returns `p(this.nth(n).get)` if the cursor is inbounds.
+      *
+      * Returns `outOfBounds` if the cursor is out of bounds.
+      */
+    def nthIs(n: Int, p: Char => Boolean, outOfBounds: Boolean): Boolean = {
```
mlutze, line 1202: IMO the out-of-bounds parameter is confusing.

Can we just add an `nthIsOOB`. Then this one returns false and we use nthIs(...) || nthIsOOB(...) for those cases.

JonathanStarup: but what does `nthIs(...)` return for oob?

JonathanStarup: > IMO the out-of-bounds parameter is confusing.

in what way?

> Then this one returns false

The upside with this version is that it is very fool proof. oob is explicit and cannot be forgotten (`nthIs(...) || nthIsOOB(...)` relies on the user)

mlutze: Hmmm fair point

magnus-madsen (approved):

main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
JonathanStarup: I also don't think its beautiful but it was the lessest demon i found 😄

PR #11765, opened 2025-10-02
feat: simplify `Reflect.default`

JonathanStarup:
The doc says `The value returned should not be depended on` so we should use null for all object types

### DIFF
--- a/main/test/ca/uwaterloo/flix/library/TestReflect.flix
+++ b/main/test/ca/uwaterloo/flix/library/TestReflect.flix
@@ -24,7 +24,7 @@ mod TestReflect {
 
     @Test
     def defaultBigDecimal01(): Unit \ Assert =
-        assertEq(expected = 0.0ff, (Reflect.default(): BigDecimal))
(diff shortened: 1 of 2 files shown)


main/test/ca/uwaterloo/flix/library/TestReflect.flix
```diff
@@ -24,7 +24,7 @@ mod TestReflect {
 
     @Test
     def defaultBigDecimal01(): Unit \ Assert =
-        assertEq(expected = 0.0ff, (Reflect.default(): BigDecimal))
+        assertTrue(Object.isNull((Reflect.default(): BigDecimal)))
```
mlutze, line 27: Why in blazes do we have tests testing a property that we're not supposed to rely on?

mlutze (approved):

magnus-madsen (approved):

main/test/ca/uwaterloo/flix/library/TestReflect.flix
JonathanStarup: 🤷‍♂️

PR #11766, opened 2025-10-02
refactor: rename `SecurityContext` constructors

jaschdoc:
Related to https://github.com/flix/flix/pull/11747


magnus-madsen:
If you are doing a pure rename, should we do that first in its own PR?

PR #11771, opened 2025-10-02
feat: fix infix parsing

JonathanStarup:
As it is now, backtick is whitespace _insensitive_, meaning that (backtick is changed to ' for md formatting)
| Input | Before | Now |
| --- | --- | --- |
| `'Add.add'` | ✔ | ✔ |
| `' Add.add '` | ❌ | ✔ |
| `' Add. add'` | ❌ | ❌ |
| `'f(g)'` | ❌ | ❌ |
| `' /* hello */ Add.add'` | ❌ | ❌ |
| `'Add..add'` | ✔(Undefined name) | ❌ |
| `'Add.$run'` | ✔(Undefined name `Add.$run`) | ✔(Undefined name `Add.run`) |
| `'Add.add.'` | ✔ | ❌ |
| `'4.2'` | ✔(Undefined name) | ❌ |
| `'$.$'` | ✔(Undefined name) | ❌ |

As an infrastructure benefit, infix names are parsed as regular names, sharing code branch.

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -1643,7 +1643,7 @@ object Parser2 {
       close(mark, TreeKind.Operator)
     }
 
-    /** Returns the unary operator type of the current token if applicable. */
+    /** Returns the unary operator type of the current token, or `None` if the current token is not a unary operator. */
     private def peekUnaryOp(token: Token): Option[UnaryOp] = {
       token.kind match {
         case TokenKind.KeywordDiscard => Some(UnaryOp.Discard)


magnus-madsen:
Just to be sure we agree: it is only qualified names that can appear infix right?  Not entire expressions.

JonathanStarup:
@magnus-madsen You can see the updated main text now. TLDR more correct but now whitespace insensitive

magnus-madsen:
Why is it a good idea to allow whitespace here?

JonathanStarup:
> Why is it a good idea to allow whitespace here?

Just the easier choice. Disallowing whitespace means that `TokenKind.Tick` must become `TokenKind.SpaceTick`, `TokenKind.TickSpace`, and `TokenKind.SpaceTickSpace`. The parser can then react accordingly.

If the lexer maintained whitespace this would be easier. Just like the `TokenKind.Dot`, `TokenKind.DotSpace`, and `TokenKind.SpaceDot` situation.

I can add whitespace sensitivity but it'll probably cost 40 LOC in the lexer and 150 LOC in the parser. That feels a bit expensive to me

magnus-madsen:
Are we moving towards the lexer keeping whitespace? Perhaps this should wait until then?

JonathanStarup:
> Perhaps this should wait until then?

I'd prefer to merge this now. I think accepting infix functions with whitespace is a lesser evil than the nonsense we allow now.

> Are we moving towards the lexer keeping whitespace?

I vote yes. clearly Flix is whitespace sensitive. Only unknowns are 1) performance and 2) how to handle `TokenKind.Whitespace` in the parser. 

1) given that lexer is now even faster, maybe even a 2x performance cost would be acceptable 🤷‍♂️
2) the parser already handles floating comment tokens in a uniform way, so perhaps handling whitespace similarly is not much issue

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -1578,33 +1576,89 @@ object Parser2 {
...
-        case TokenKind.KeywordOr => Some(BinaryOp.Or)
-        case TokenKind.Minus => Some(BinaryOp.Minus)
-        case TokenKind.NameMath => Some(BinaryOp.NameMath)
-        case TokenKind.Plus => Some(BinaryOp.Plus)
-        case TokenKind.Slash => Some(BinaryOp.Slash)
-        case TokenKind.Star => Some(BinaryOp.Star)
-        case TokenKind.ColonColonColon => Some(BinaryOp.TripleColon)
-        case TokenKind.GenericOperator => Some(BinaryOp.UserDefinedOperator)
-        case _ => None
+    /** Returns the binary operator type of the current token if applicable. */
```
magnus-madsen, line 1565: When would it not be applicable, so to speak. When is none returned?

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -1578,33 +1576,89 @@ object Parser2 {
...
+            case TokenKind.Star => Some(BinaryOp.Star)
+            case TokenKind.Tick => Some(BinaryOp.InfixFunction)
+            case TokenKind.ColonColonColon => Some(BinaryOp.TripleColon)
+            case TokenKind.GenericOperator => Some(BinaryOp.UserDefinedOperator)
+            case _ => None
+          }
+      }
+    }
+
+    private val FIRST_BINARY_OP: Set[TokenKind] = Set(
```
magnus-madsen, line 1596: Are these constants usually scattered throughout the file?

magnus-madsen (approved):

magnus-madsen:
I approved -- so you can merge. But you are welcome to fix the two comments before (or not)

main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
JonathanStarup: Yes just before the function they are used in

PR #11772, opened 2025-10-02
fix: lex `<=>` distinctly

JonathanStarup:
Somewhere `<=>` got lost as an operator, but because `Weeder` does string matching, it only resulted in inproper precedence


main/src/ca/uwaterloo/flix/language/phase/Lexer.scala
```diff
@@ -183,6 +185,8 @@ object Lexer {
       ("<+>", TokenKind.AngledPlus),
       ("<-", TokenKind.ArrowThinL),
       ("<=", TokenKind.AngleLEqual),
+      ("<=>", TokenKind.AngledEqual),
+      ("=", TokenKind.Equal),
```
JonathanStarup, line 186: This was moved because it was mis-places before (alphabetized lines)

PR #11774, opened 2025-10-02
refactor: avoid operator string matching in Weeder

JonathanStarup:


main/src/ca/uwaterloo/flix/language/phase/Weeder2.scala
```diff
@@ -1314,10 +1311,11 @@ object Weeder2 {
                   sctx.errors.add(error)
                   Validation.Success(Expr.Error(error))
               }
-            // UNRECOGNIZED
-            case id =>
-              val ident = Name.Ident(id, op.loc)
+            case token@Token(TokenKind.UserDefinedOperator, _, _, _, _, _) =>
+              val ident = Name.Ident(token.text, op.loc)
               Validation.Success(Expr.Apply(Expr.Ambiguous(Name.QName(Name.RootNS, ident, ident.loc), op.loc), List(e1, e2), tree.loc))
+            case _ =>
```
JonathanStarup, line 1317: Before, if the operator was `TokenKind.KeywordDef` you would here just create a QName called `def`. This explicitly crashes for unexpected operators.

Otherwise pure refactor

magnus-madsen: Why is it good to crash? Why not error out instead?

JonathanStarup: Whenever something crashes because of parser assumptions, ICE is used. Just keeping consistent with the rest of the weeder.

In master, any syntax tree is just converted to a string and thrown into a QName. ICE is not great but definitely better

magnus-madsen: Right, but we were trying to move away from ICEs -- whenever possible.

JonathanStarup: But we don't have anything else atm. I'd have to add a custom WeederError just for this. Can we merge this and leave the ICE rewrite to its own PR?

JonathanStarup: @magnus-madsen ?

PR #11775, opened 2025-10-02
feat: `Reflect.reflect`

JonathanStarup:


main/src/library/Fixpoint3/Boxable.flix
```diff
@@ -23,18 +23,21 @@ mod Fixpoint3.Boxable {
...
+            case JvmValue.JvmBool(v)  => Boxed.BoxedBool(v)
+            case JvmValue.JvmChar(v)  => Boxed.BoxedChar(v)
+            case JvmValue.JvmInt8(v)  => Boxed.BoxedInt8(v)
+            case JvmValue.JvmInt16(v) => Boxed.BoxedInt16(v)
+            case JvmValue.JvmInt32(v) => Boxed.BoxedInt32(v)
+            case JvmValue.JvmInt64(v) => Boxed.BoxedInt64(v)
+            case JvmValue.JvmFloat32(v) => Boxed.BoxedFloat32(v)
+            case JvmValue.JvmFloat64(v) => Boxed.BoxedFloat64(v)
+            case JvmValue.JvmObject(v) =>
+                let cmp = (o1, o2) -> Order.compare(unchecked_cast(o1 as a), unchecked_cast(o2 as a));
```
JonathanStarup, line 39: This is a really interesting case of combining traits and jvm representation. This might be a problem for enum optimization but not for the mission to remove typematch

main/src/library/Reflect.flix
```diff
@@ -17,6 +18,33 @@ mod Reflect {
...
+            case i: Int16 => JvmValue.JvmInt16(i)
+            case i: Int32 => JvmValue.JvmInt32(i)
+            case i: Int64 => JvmValue.JvmInt64(i)
+            case f: Float32 => JvmValue.JvmFloat32(f)
+            case f: Float64 => JvmValue.JvmFloat64(f)
+            case _: _ => JvmValue.JvmObject(unchecked_cast(x as JObject))
+        }
+
+    /// A reflected value, reflecting its erased runtime representation.
+    pub enum JvmValue {
```
magnus-madsen: Can you move the type up before the function, ditto for enum Purity

main/src/library/Fixpoint3/Boxable.flix
magnus-madsen: Will I lose Datalog performance if I merge this now?

JonathanStarup: Yes its not compiletime evaluated. I am working on a optimizer thing so we can wait for that if you prefer

jaschdoc:
`Reflect.reflect` will never be inlined, right?

magnus-madsen:
> `Reflect.reflect` will never be inlined, right?

Of course it will. Why should it not? 

To avoid the fiasco with the previous debug expression, _every function must be inlineable_.

jaschdoc:
> > `Reflect.reflect` will never be inlined, right?
> 
> Of course it will. Why should it not?
> 
> To avoid the fiasco with the previous debug expression, _every function must be inlineable_.

You're right that in principle it is okay to inline, but in reality it does not qualify for inlining. Just wondering if we should somehow catch this case so `Reflect.reflect` is always inlined?

jaschdoc:
`isSingleAction(defn.exp) || isSimple(defn.exp) || hasKnownLambda(exps)` never holds for `Reflect.reflect`.

main/src/library/Fixpoint3/Boxable.flix
CasperDN: As a note I think `box` is only used with respect to the boundary of the Datalog engine/interaction with Flix.

With this I mean that it should currently only be used for lifted functions (functionals, applications, guards and lattice functions). Of course also once for data entry/`InjectIntoY`.

JonathanStarup:
@magnus-madsen as @jaschdoc says, `Reflect.reflect` is not inlined, which hinders the collapse of the pattern match. At optimization time `Reflect.reflect` is
```flix
import java.lang.{Object => JObject}
def reflect$174358(x$176101: String): JvmValue =
    JvmValue.JvmObject(unchecked_cast(x$176101 as JObject))
```

Should we add logic to cover this case or? We can also add `@Inline` to `reflect`, that also works.

jaschdoc:
`@Inline` may be removed in the future so that's not good to rely on. I think we should update the `isTrivial` predicate to recursively include casts: `Expr.Cast(exp) => isTrivial(exp)`

JonathanStarup:
IMO inline annotations are vital to a system with inlining to make users feel like they can understand/shape the system.

Is the meaning of `isTrivial` that it makes "no computation" (ignoring simple allocation)?

JonathanStarup:
@magnus-madsen I've now added the change is @jaschdoc suggested, `Reflect.reflect` is completely inlined

magnus-madsen:
@JonathanStarup @jaschdoc What's the status on the inline annotation? Does it exist? Does it work? Under what conditions?

JonathanStarup:
It worked for this application when I tried it. But this version doesn't use the inline annotation

main/src/library/Fixpoint3/Boxable.flix
```diff
@@ -23,18 +23,21 @@ mod Fixpoint3.Boxable {
...
-        case v: Int8    => Boxed.BoxedInt8(v)
-        case v: Int16   => Boxed.BoxedInt16(v)
-        case v: Int32   => Boxed.BoxedInt32(v)
-        case v: Int64   => Boxed.BoxedInt64(v)
-        case v: Float32 => Boxed.BoxedFloat32(v)
-        case v: Float64 => Boxed.BoxedFloat64(v)
-        case _: _ =>
-            let cmp = (o1, o2) -> Order.compare(unchecked_cast(o1 as a), unchecked_cast(o2 as a));
-            Boxed.BoxedObject(unchecked_cast(x as Object), cmp)
+    pub def box(x: a): Boxed with Order[a] = {
```
magnus-madsen, line 27: Mark as inline?

main/src/library/Object.flix
```diff
@@ -22,17 +22,9 @@ mod Object {
     /// Returns `true` if the given value `x` is `null`.
     ///
     pub def isNull(x: a): Bool =
```
magnus-madsen, line 25: Use inline ann

jaschdoc:
> Is the meaning of `isTrivial` that it makes "no computation" (ignoring simple allocation)?

Yes. :)
Trivial expressions are either constants or variables. In those cases we know for sure that computation is bounded.

jaschdoc:
> @JonathanStarup @jaschdoc What's the status on the inline annotation? Does it exist? Does it work? Under what conditions?

Both `Inline` and `DontInline` annotations are supported and work.

magnus-madsen:
Upwards and onwards! 🚀

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -873,12 +873,14 @@ object Inliner {
...
     * A trivial expression is one of the following:
     *   - [[Expr.Var]]
+    *   - Is a [[Expr.Cast]] of a trivial expression.
     *   - Any expression where [[isCst]] holds.
     *
     * A pure and trivial expression can always be inlined even without duplicating work.
     */
   private def isTrivial(exp0: Expr): Boolean = exp0 match {
     case Expr.Var(_, _, _) => true
+    case Expr.Cast(exp, _, _, _) => isTrivial(exp)
```
magnus-madsen, line 883: @JonathanStarup I think this caused a performance regression: 

https://flix.cs.au.dk/d/agddciz4k/main?orgId=1&from=now-7d&to=now&timezone=browser

PR #11777, opened 2025-10-02
experiment: add `Diffable`

magnus-madsen:


magnus-madsen:
@NeonOxide @mlutze @JonathanStarup Thoughts on this?

What if we require it for `assertEq`? I suppose it would such that you cannot assert on things that are not diffable? 

(Optional type classes?)

main/src/library/Diffable.flix
```diff
@@ -0,0 +1,87 @@
...
+///
+pub trait Diffable[a] with Eq[a], ToString[a] {
+
+    ///
+    /// Returns a `RichString` showing the differences between `expected` and `actual`.
+    ///
+    /// The diff is formatted with colors:
+    /// - Red for elements only in `expected` (removed)
+    /// - Green for elements only in `actual` (added)
+    /// - Yellow/Blue for changed values
```
JonathanStarup, line 30: yellow or blue?

main/src/library/Diffable.flix
```diff
@@ -0,0 +1,87 @@
...
+    pub def formatHeader(expected: String, actual: String): RichString =
+        text("Expected: ") + text(expected) + text("\n") +
+        text("Actual:   ") + text(actual) + text("\n") +
+        text("\nDiff:\n")
+
+    ///
+    /// Returns "No differences" message when values are equal.
+    ///
+    pub def formatNoDiff(): RichString =
+        green("No differences")
```
JonathanStarup, line 70: `Green for elements only in actual`

main/src/library/Diffable.flix
```diff
@@ -0,0 +1,87 @@
...
+    ///
+    /// Formats a changed element (different values for the same key/position).
+    ///
+    pub def formatChanged(s: String): RichString =
+        yellow("  ~ ") + text(s)
+
+    ///
+    /// Builds a diff header showing the expected and actual values.
+    ///
+    pub def formatHeader(expected: String, actual: String): RichString =
```
JonathanStarup, line 61: The header thing doesn't belong to this name space i my head. Feels like its too contextual

main/src/library/Diffable.flix
```diff
@@ -0,0 +1,87 @@
...
+}
+
+instance Diffable[Set[a]] with Order[a], ToString[a] {
+    pub def diff(expected: {expected = Set[a]}, actual: Set[a]): RichString =
+        if (expected#expected == actual)
+            Diffable.formatNoDiff()
+        else {
+            let removed = Set.difference(expected#expected, actual);
+            let added = Set.difference(actual, expected#expected);
+            let header = Diffable.formatHeader(ToString.toString(expected#expected), ToString.toString(actual));
```
JonathanStarup, line 81: Same way I don't think header should be here, it should just output the diff string, nothing else

main/src/library/Diffable.flix
```diff
@@ -0,0 +1,87 @@
...
+
+mod Diffable {
+
+    use RichString.{green, red, text, yellow};
+
+    ///
+    /// Formats an added element (present in actual but not in expected).
+    ///
+    pub def formatAdded(s: String): RichString =
+        green("  + ") + text(s)
```
JonathanStarup, line 44: You use this `+-~` thing, maybe it should be a part of the effect spec? seems like a good requirement for monochrome output

JonathanStarup:
I think its cool

> What if we require it for assertEq?

I'd vote no. Too restrictive and annoying for custom types. Have its own assert function

mlutze:
I agree with what Jonathan says

PR #11778, opened 2025-10-02
work

JonathanStarup:


JonathanStarup:
@magnus-madsen @jaschdoc Isn't it pretty cool :D

jaschdoc:
Yeah that's quite nice. How much does compilation time increase?

JonathanStarup:
`--Xbenchmark-phases` with `N = 50`
- `    PR: Optimizer 83,7 ms (07,0%)`
- `Master: Optimizer 79,9 ms (07,1%)`

JonathanStarup:
@jaschdoc the code is a lot better now, maybe you can give it a look?

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -332,7 +334,7 @@ object Inliner {
     case Expr.Match(exp, rules, tpe, eff, loc) =>
       val e = visitExp(exp, ctx0)
       val rs = rules.map(visitMatchRule(_, ctx0))
-      Expr.Match(e, rs, tpe, eff, loc)
+      reduceMatch(e, rs, tpe, eff, loc)
```
JonathanStarup, line 338: Like with `IfThenElse` the `visitExp` could be done "lazily" in case there are dead code. But this only seems like a performance consideration and not a correctness concern. Is that right @jaschdoc?

Doing it lazily seems quite cumbersome

magnus-madsen:
Are there some test cases we should add?

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -371,6 +373,188 @@ object Inliner {
       Expr.NewObject(name, clazz, tpe, eff, methods, loc)
   }
 
+  /** Evaluate match if possible */
```
magnus-madsen: I would like a description in words of what this does and does not do.

JonathanStarup: I've added some documentation. But i've tried to make the code as self-describing as possible.

JonathanStarup:
> Are there some test cases we should add?

I could add a small suite of trivial pattern matches, maybe that'd be good?

JonathanStarup:
Actually most of our match tests are hit by the optimizer, fx
```flix
@test
    def testMatchNested03(): Bool = match Some(Some(123)) {
        case None           => false
        case Some(None)     => false
        case Some(Some(_))  => true
    }
```

jaschdoc:
> Actually most of our match tests are hit by the optimizer, fx
> 
> ```flix
> 
> @test
> 
>     def testMatchNested03(): Bool = match Some(Some(123)) {
> 
>         case None           => false
> 
>         case Some(None)     => false
> 
>         case Some(Some(_))  => true
> 
>     }
> 
> ```

So this is optimized to `true`?

magnus-madsen:
Is there any need for "negative tests"?

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -371,6 +373,195 @@ object Inliner {
       Expr.NewObject(name, clazz, tpe, eff, methods, loc)
   }
 
+  /**
+    * Evaluate match if possible.
+    *
+    * The match rules are iterated top to bottom, at each step either:
```
magnus-madsen, line 382: iterated? checked?matched?

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -371,6 +373,195 @@ object Inliner {
       Expr.NewObject(name, clazz, tpe, eff, methods, loc)
   }
 
+  /**
+    * Evaluate match if possible.
```
magnus-madsen: More precise? Like how much evaluation does it do? For example, does it simplify a match?

e.g.

```
match Some(UNKNOWN) { ... } 
```

I suppose not, having not read further, so I would say something:

Statically evaluates a match expression to one of its match rules if the scrutinee is known (or sufficiently known?)

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -371,6 +373,195 @@ object Inliner {
...
+    *
+    * The match rules are iterated top to bottom, at each step either:
+    *   - Removing the rule if it can never match (either the patten cannot match or the guard is guaranteed false)
+    *   - Removing the match if a rule is known to match (either `Some(12)` matching `Some(x)` or a lenient pattern like `_`)
+    *   - Returning the rules as is if the rule cannot be determined statically.
```
magnus-madsen: What does this mean? Is it more like:

Otherwise leaves the match expression as it were?

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -371,6 +373,195 @@ object Inliner {
...
+    */
+  @tailrec
+  private def reduceMatch(exp: MonoAst.Expr, rules: List[MonoAst.MatchRule], tpe: Type, eff: Type, loc: SourceLocation)(implicit sym0: Symbol.DefnSym, sctx: SharedContext): Expr = {
+    rules match {
+      case MonoAst.MatchRule(pat, guardOpt, ruleExp) :: rest =>
+        matchRule(exp, pat, guardOpt) match {
+          case MatchResult.Match(binders) =>
+            // Guaranteed rule - convert to let binders.
```
magnus-madsen: -> Guaranteed match

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -371,6 +373,195 @@ object Inliner {
...
+            sctx.changed.putIfAbsent(sym0, ())
+            bindPatterns(binders.toSeq, ruleExp, loc)
+          case MatchResult.NoMatch =>
+            // Impossible rule - delete and continue.
```
magnus-madsen: Impossible match

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -371,6 +373,195 @@ object Inliner {
...
+            sctx.changed.putIfAbsent(sym0, ())
+            reduceMatch(exp, rest, tpe, eff, loc)
+          case MatchResult.Unknown =>
+            // Match is unknown - do nothing.
```
magnus-madsen: Why not recurse on the rest? Is it not possible to simplify the rest of the pattern?

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -371,6 +373,195 @@ object Inliner {
...
+    case object True extends FuzzyBool
+
+    case object False extends FuzzyBool
+
+    case object Unknown extends FuzzyBool
+
+  }
+
+  /** Returns the fuzzy evaluation of `exp` as a boolean. */
+  private def evalBoolExpression(exp: MonoAst.Expr): FuzzyBool = exp match {
```
magnus-madsen, line 427: I suppose its OK, but Option[Boolean] would also work right? Is it worth it. I don't see that you have any abstract operations on fuzzy bool like and or . So maybe not?

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -371,6 +373,195 @@ object Inliner {
...
+    case Expr.Cst(Constant.Bool(true), _, _) => FuzzyBool.True
+    case Expr.Cst(Constant.Bool(false), _, _) => FuzzyBool.False
+    case _ => FuzzyBool.Unknown
+  }
+
+  private sealed trait MatchResult
+
+  private object MatchResult {
```
magnus-madsen, line 435: I would expect some comments on the case classes/objects

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -371,6 +373,195 @@ object Inliner {
...
+
+    case class Match(binders: Chain[(Option[Pattern.Var], MonoAst.Expr)]) extends MatchResult
```
magnus-madsen, line 448: -> ExactMatch?

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -371,6 +373,195 @@ object Inliner {
...
+
+    case object NoMatch extends MatchResult
+
+    case object Unknown extends MatchResult
+
+    def emptyMatch(): MatchResult =
```
magnus-madsen, line 465: I dont understand what this is

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -371,6 +373,195 @@ object Inliner {
...
+      * Concatenates two match results.
+      *
+      * If either result is [[NoMatch]], the output is [[NoMatch]].
+      *
+      * Then if either result is [[Unknown]], the output is [[Unknown]].
+      *
+      * Then (both results now being [[Match]]) the binders are concatenated.
+      */
+    def concat(mr1: MatchResult, mr2: MatchResult): MatchResult = (mr1, mr2) match {
+      case (MatchResult.NoMatch, _) | (_, MatchResult.NoMatch) =>
```
magnus-madsen, line 498: We dont use or patterns anywhere else in the compiler; let us not start here.

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -371,6 +373,195 @@ object Inliner {
...
+        MatchResult.NoMatch
+      case (MatchResult.Unknown, _) | (_, MatchResult.Unknown) =>
+        MatchResult.Unknown
+      case (MatchResult.Match(binders1), MatchResult.Match(binders2)) =>
+        MatchResult.Match(binders1 ++ binders2)
```
magnus-madsen, line 503: Something here about preservation of evaluation order.

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -371,6 +373,195 @@ object Inliner {
...
+    }
+
+  }
+
+  /** Returns the match result of `exp` against `pat` with the guard `guardOpt`. */
```
magnus-madsen, line 508: This function and the next sounds like the ones where you want an inline code example.

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -371,6 +373,195 @@ object Inliner {
...
+      case (_, FuzzyBool.False) | (MatchResult.NoMatch, _) =>
+        MatchResult.NoMatch
+      case (MatchResult.Unknown, _) | (MatchResult.Match(_), FuzzyBool.Unknown) =>
+        MatchResult.Unknown
+      case (resultMatch@MatchResult.Match(_), FuzzyBool.True) =>
+        resultMatch
+    }
+  }
+
+  /** Returns the match result of `exp` against `pat`. */
```
magnus-madsen, line 521: There is no mention anywhere of linearity. I would assume that to be mentioned since it is a requirement for correctness.

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -371,6 +373,195 @@ object Inliner {
...
+    case Pattern.Tuple(pats, _, _) =>
+      exp match {
+        case Expr.ApplyAtomic(AtomicOp.Tuple, exps, _, _, _) =>
+          matchPatterns(exps, pats.toList)
+        case _ =>
+          MatchResult.Unknown
+      }
+
+    case Pattern.Record(_, _, _, _) =>
+      MatchResult.Unknown
```
magnus-madsen, line 557: Too bad; we have a lot of labelled arguments where this might help :)

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -371,6 +373,195 @@ object Inliner {
...
+
+  }
+
+  /**
+    * Returns the match result for matching `exps` to the patterns of `pats`.
+    *
+    * If the two lists are of different lengths, [[MatchResult.Unknown]] is returned.
+    */
+  private def matchPatterns(exps: List[MonoAst.Expr], pats: List[MonoAst.Pattern]): MatchResult = {
+    if (exps.lengthCompare(pats) != 0) {
```
magnus-madsen, line 569: When does this happen?

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -371,6 +373,195 @@ object Inliner {
...
+      MatchResult.Unknown
+    } else {
+      val res = pats.zip(exps).foldLeft(MatchResult.emptyMatch()) {
```
magnus-madsen, line 576: I am confused why this is not a foldRight (the natural fold on lists)

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -371,6 +373,195 @@ object Inliner {
...
+    import Constant.*
+    import MatchResult.*
+    (cst1, cst2) match {
+      case (Unit, Unit) => emptyMatch()
+      case (Null, Null) => emptyMatch()
+      case (Bool(v1), Bool(v2)) => matchFromBool(v1 == v2)
+      case (Char(v1), Char(v2)) => matchFromBool(v1 == v2)
+      case (Float32(v1), Float32(v2)) => matchFromBool(v1 == v2)
+      case (Float64(v1), Float64(v2)) => matchFromBool(v1 == v2)
+      case (BigDecimal(_), BigDecimal(_)) => Unknown // Avoiding static checking of object types.
```
magnus-madsen, line 599: Great 👍

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -371,6 +373,195 @@ object Inliner {
...
+    */
+  private def matchConstant(cst1: Constant, cst2: Constant): MatchResult = {
+    import Constant.*
+    import MatchResult.*
+    (cst1, cst2) match {
+      case (Unit, Unit) => emptyMatch()
+      case (Null, Null) => emptyMatch()
+      case (Bool(v1), Bool(v2)) => matchFromBool(v1 == v2)
+      case (Char(v1), Char(v2)) => matchFromBool(v1 == v2)
+      case (Float32(v1), Float32(v2)) => matchFromBool(v1 == v2)
```
magnus-madsen, line 597: Thoughts on your best friend, NaN?

magnus-madsen (commented):
Overall looks good. Just a lot of pendatory and a bit of concern about evaluation order.

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
JonathanStarup: all valid, i think iterated is best 🤷‍♂️

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
JonathanStarup: I've added the overall logic of the function. I think documenting all details will just go out of date immediately and not give much benefit.

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
JonathanStarup: clarified

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
JonathanStarup: The logic has to change, because rules following this unknown case cannot 'match' statically, they can only unmatch statically.

This code focuses on matches that can statically match, not be statically pruned.

This could easily be done in follow up PR

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
JonathanStarup: You usually don't like that i use options so I preempted that critique. But I do think FuzzyBool makes it nicer to read.

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
JonathanStarup: added

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
JonathanStarup: ExactMatch seem to imply an inexact match, so i think match is fine

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
JonathanStarup: https://github.com/flix/flix/blob/4df8a6dd7f28273951977de25bae073a2a02f2c4/main/src/ca/uwaterloo/flix/language/ast/TokenKind.scala#L394-L404

https://github.com/flix/flix/blob/4df8a6dd7f28273951977de25bae073a2a02f2c4/main/src/ca/uwaterloo/flix/language/ast/SimpleType.scala#L150-L153

https://github.com/flix/flix/blob/4df8a6dd7f28273951977de25bae073a2a02f2c4/main/src/ca/uwaterloo/flix/language/phase/Monomorpher.scala#L281-L283

https://github.com/flix/flix/blob/4df8a6dd7f28273951977de25bae073a2a02f2c4/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala#L1614-L1630

etc etc etc

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
JonathanStarup: In this function the binders are concatenated as documented, but ill add a note at the use site

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
JonathanStarup: IMO its documented (with the new changes) in `MatchResult` itself

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
JonathanStarup: linearity of what?

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
JonathanStarup: wanted to keep the PR small, the concrete matching precision can always be increased.

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
JonathanStarup: Just compiler bugs.

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
JonathanStarup: The intuitive traversal of rules is top-to-bottom, left-to-right. Following the evaluation of it.

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
JonathanStarup: `v1 == v2` is using float equality, just like a pattern match would do. So no matter what happens it should be consistent.

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
JonathanStarup: Makes the data more self-documenting instead of relying so much on comments

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
magnus-madsen: I bow before you.

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
magnus-madsen: I means a variable can only occur once.

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
magnus-madsen: Would ICE not be better here? It should never happen because of the type system, right?

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
magnus-madsen: OK

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
```diff
@@ -371,6 +373,235 @@ object Inliner {
...
+  /** Returns the match result of `exp` against `pat`. */
+  private def matchPat(exp: MonoAst.Expr, pat: MonoAst.Pattern): MatchResult = pat match {
+    case Pattern.Wild(_, _) =>
+      // Preserve the expression in case it is impure.
+      MatchResult.singleMatch(None, exp)
+
+    case v@Pattern.Var(_, _, _, _) =>
+      MatchResult.singleMatch(Some(v), exp)
+
+    case Pattern.Cst(cst, _, _) =>
```
magnus-madsen: I would write => exp match { 

to save some line breaks, but whatever.

magnus-madsen:
I added some comments, but the only real important one is whether:

exps.lengthCompare(pats) != 0

should not be removed or turned into ICE

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
JonathanStarup: Ah right

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
JonathanStarup: made it ice now

main/src/ca/uwaterloo/flix/language/phase/optimizer/Inliner.scala
JonathanStarup: fixed

JonathanStarup:
I've updated the PR with the changes. I think ICE is good yeah.

In regards to the name, i just wanted to make sure we are on the same page. This PR can reduce match expressions even if the scrutinee is not known.

```
match Some(Random.nextInt()) {
    case Some(x) if true => x
    case _ => 42
}
```
becomes
```
let x = Random.nextInt();
x
```
for example

JonathanStarup:
@magnus-madsen can this be merged or?

magnus-madsen:
Yes, I approved it.

PR #11782, opened 2025-10-03
chore: clean up Type.scala

mlutze:


mlutze:
!!!jar

stealing compute so I don't have to do it myself

magnus-madsen (approved):

PR #11783, opened 2025-10-03
ci: set gradle project name to 'flix'

mlutze:


magnus-madsen:
VSCode repo does not like

magnus-madsen (approved):

PR #11785, opened 2025-10-03
fix: annotations and modifiers require decl

### LINKED ISSUE #8416: Writing just `@Dep` should be an error
This program should not be valid:

```
@Dep
```

CasperDN:
Fixes #8416

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -917,14 +917,7 @@ object Parser2 {
         case TokenKind.KeywordStruct => structDecl(mark)
         case TokenKind.KeywordType => typeAliasDecl(mark)
         case TokenKind.KeywordEff => effectDecl(mark)
-        case TokenKind.Eof =>
-          if (wasAtEofBeforeAnnotations) {
-            close(mark, TreeKind.CommentList) // Last tokens in the file were comments.
-          } else {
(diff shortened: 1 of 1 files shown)


main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -916,7 +917,14 @@ object Parser2 {
         case TokenKind.KeywordStruct => structDecl(mark)
         case TokenKind.KeywordType => typeAliasDecl(mark)
         case TokenKind.KeywordEff => effectDecl(mark)
-        case TokenKind.Eof => close(mark, TreeKind.CommentList) // Last tokens in the file were comments.
+        case TokenKind.Eof =>
+          if (wasAtEofBeforeAnnotations) {
+            close(mark, TreeKind.CommentList) // Last tokens in the file were comments.
+          } else {
```
JonathanStarup: Isn't this else branch the same as the `case at =>` branch? maybe just adding `case TokenKind.Eof if wasAtEofBeforeAnnotations => close(mark, TreeKind.CommentList) // Last tokens in the file were comments.` will keep the logic more uniform?

magnus-madsen:
Seems a bit adhoc. Is there no better way @JonathanStarup ?

JonathanStarup:
I can't think of something better, and considering its a two line addition I think its fine

magnus-madsen:
Why is this only a problem for annotations though?

CasperDN:
I think its due to annotations being allowed at to level (Decls). For most of the parsing EOF would signal an error.

magnus-madsen (approved):

PR #11786, opened 2025-10-03
fix: type alias must have body

### LINKED ISSUE #8357: Missing error for undefined type alias
```
type alias Magic
def f(): Magic = ???
def main(): Int32 = 32
```
This works and runs

CasperDN:
Fixes #8357

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -1252,9 +1252,8 @@ object Parser2 {
       if (at(TokenKind.BracketL)) {
         Type.parameters()
       }
-      if (eat(TokenKind.Equal)) {
-        Type.ttype()
-      }
+      expect(TokenKind.Equal)
+      Type.ttype()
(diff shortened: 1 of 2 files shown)


main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
```diff
@@ -1252,9 +1252,8 @@ object Parser2 {
       if (at(TokenKind.BracketL)) {
         Type.parameters()
       }
-      if (eat(TokenKind.Equal)) {
-        Type.ttype()
-      }
+      expect(TokenKind.Equal)
+      Type.ttype()
```
JonathanStarup: I think the idea of `if (eat(TokenKind.Equal))` is to avoid two parser errors, perhaps this is best of both worlds?
```scala
if (eat(TokenKind.Equal)) {
  Type.ttype()
} else {
  expect(TokenKind.Equal)
}
```

magnus-madsen:
Does this problem happen elsewhere?

JonathanStarup:
I would guess almost certainly. The cause of this is places in the Weeder that says "parser will have done X" and many arent true

PR #11788, opened 2025-10-03
fix: disallow capital name functions

CasperDN:
With the change to `nameUnqualified` a more restricted set of `TokenKind` can be supplied to generate the needed errors by the parser while breaking parsing.

Disallows every function declaration in the following (except `def a()`):
```
trait A[a] {
    pub def A(_: a): Int32 = 0
}

trait B[a] {
    law A:forall() false
}

instance A[Int32] {
    pub def A(_: Int32): Int32 = 2
    pub redef B(_: Int32): Int32 = 2
}

pub eff C {
    def A(): Unit
}

def a(): Unit = {
    def A() = ();
    A()
}
```

### DIFF
--- a/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
+++ b/main/src/ca/uwaterloo/flix/language/phase/Parser2.scala
@@ -618,6 +618,7 @@ object Parser2 {
     */
   private val NAME_LIKE: Set[TokenKind] = Set(TokenKind.NameLowercase, TokenKind.NameUppercase, TokenKind.NameMath, TokenKind.Underscore)
   private val NAME_DEFINITION: Set[TokenKind] = Set(TokenKind.NameLowercase, TokenKind.NameUppercase, TokenKind.NameMath, TokenKind.GenericOperator)
+  private val NAME_FUNCTION: Set[TokenKind] = Set(TokenKind.NameLowercase, TokenKind.NameMath, TokenKind.GenericOperator)
   private val NAME_PARAMETER: Set[TokenKind] = Set(TokenKind.NameLowercase, TokenKind.NameMath, TokenKind.Underscore)
   private val NAME_VARIABLE: Set[TokenKind] = Set(TokenKind.NameLowercase, TokenKind.NameMath, TokenKind.Underscore)
   private val NAME_JAVA: Set[TokenKind] = Set(TokenKind.NameLowercase, TokenKind.NameUppercase)
@@ -1023,7 +1024,7 @@ object Parser2 {
       implicit val sctx: SyntacticContext = SyntacticContext.Decl.Module
       assert(at(TokenKind.KeywordDef))
       expect(TokenKind.KeywordDef)
-      nameUnqualified(NAME_DEFINITION)
+      nameUnqualified(NAME_FUNCTION)
       if (at(TokenKind.BracketL)) {
         Type.parameters()
       }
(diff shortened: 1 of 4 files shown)


JonathanStarup:
This makes the parser more case-sensitive, isn't that the wrong direction to go? To me it feels cleaner that Weeder verifies the name casing. Is the current weeder missing some checks or what is the issue?

CasperDN:
I guess it depends on your point of view.

The parser is, by this change, no more case sensitive in the sense that its behavior, with respect to parsing, is unchanged. It just issues some errors for unexpected names. That is what I was trying to show with the tests. The parsing is the same and it only issues the errors for unexpected names.

I can move the checks to the weeder if we want them there. The weeder currently only checks casing for `TreeKind.Decl.Def` in `visitDefinitionDecl`.

magnus-madsen:
Either way, we should get this fixed soon 👍

JonathanStarup:
I think - long term - case checking should be done systematically in the weeder. This removes the noise from the parser and keeps errors good.

Short term 🤷‍♂️.

CasperDN:
@JonathanStarup Would you be against this being merged and then moved to weeding later with the rest of the checks? The checks from `nameUnqualified` should probably be moved while the parser should start using some `eatNameLike` function which does not care about how the name looks. With this the move should be simpler as the token set already exist and so do the tests.

JonathanStarup:
Uhm sure yeah, but sounds like a @magnus-madsen decision.

Bonus note, `name.isUpper` is kinda of broken with `_names` so its probably nice to use it less.

https://github.com/flix/flix/blob/7a96559304f2f4f7828d01f35a221130d08ec3a5/main/src/ca/uwaterloo/flix/language/ast/Name.scala#L90-L100

magnus-madsen:
I skimmed the discussion. I am not sure what the benefit is here. What's the overall goal here?

I am partial to Jonathan's comment that making the parser more case sensitive may not be what we want.

CasperDN:
The goal is to enforce the requirement that functional names must not be upper cased. If we do not want to enforce that then this PR is of course meaningless.

I would still argue that the parser is not more sensitive in the sense that if it parsed before it will parse now. It just might also produce an error mark. As far as I know an error mark will not produce any problems during weeding as long as the ident is also produced, which it is.

One check is moved to the parser (`Weeder2.visitDefinitionDecl`'s `isUpper` check) for consistency (at the moment). I agree that in general checks should go the other way, but as @JonathanStarup mentioned it used the broken `isUpper`.

One option towards the overall goal of moving name verification to the weeder would be to make `Weeder2.tokenToIdent` take a set of `TokenKind` and emit an error if the `TokenKind` did not match the identifier (exactly what `nameUnqualified` does today).

I see this as mostly orthogonal, but with the nice property that the tests now exists (though they will be moved to the weeder at some point) and so does the name class that should be used.

magnus-madsen:
> The goal is to enforce the requirement that functional names must not be upper cased. If we do not want to enforce that then this PR is of course meaningless.

I definitely want to enforce that. Not just for functions though-- signatures and operators too.

@JonathanStarup What do you think? This is a bit beyond me.

JonathanStarup:
This code is a positive change. Even though later it might be overridden by a systemic treatment

magnus-madsen:
OK; then I will defer to you and merge this. 

Is there anywhere we have written up the long-term plan?

JonathanStarup:
@CasperDN once this is non-conflicting it can be merged :)

magnus-madsen (approved):

magnus-madsen:
@JonathanStarup I don't fully understand whats the plan here, but if you say its better, I will merge it.

